PR #4203, opened 2022-09-02
Increase timeouts of expensive tests

nilsdeppe:
## Proposed changes

We've had some timeouts of these tests over the last month. Since they already
all increase the timeout, it's better to bump it a bit more and not have CI
failing.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Evolution/Systems/GeneralizedHarmonic/Test_Constraints.cpp
+++ b/tests/Unit/Evolution/Systems/GeneralizedHarmonic/Test_Constraints.cpp
@@ -1332,7 +1332,7 @@ void test_compute_tags() {
 }
 }  // namespace
 
-// [[TimeOut, 20]
+// [[TimeOut, 10]]
 SPECTRE_TEST_CASE("Unit.Evolution.Systems.GeneralizedHarmonic.Constraints",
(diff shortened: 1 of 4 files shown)


tests/Unit/Evolution/Systems/GeneralizedHarmonic/Test_Constraints.cpp
```diff
@@ -1332,7 +1332,7 @@ void test_compute_tags() {
 }
 }  // namespace
 
-// [[TimeOut, 20]
+// [[TimeOut, 30]
```
wthrowe: This has (and already had) imbalanced brackets.  Is this timeout actually being applied?

wthrowe (changes requested):

tests/Unit/Evolution/Systems/GeneralizedHarmonic/Test_Constraints.cpp
nilsdeppe: Hah, it wasn't! Good catch! Decreased to 10s, corrected brackets

wthrowe (approved):

PR #4212, opened 2022-09-07
Make TCI status a simple tag with integer values

yoonso0-0:
## Proposed changes

This is the first part of a change to make TciStatus more informative, by assigning different integer values to TciStatus depending on which particular one among TCI checks failed.

* Currently `rdmp_tci` and `two_mesh_rdmp_tci` returns `bool`. Change this return type from `bool` to `int`. The returned value is `0` if all checks are passed, and is a positive integer `j` if j-th component of inputs makes the check fail (see the added documentation). Internal automatic conversion by C++ for `0`->`False` and `j`->`True` prevents other related parts of subcell codes from being broken.

* Change `TciStatus` from a compute tag to a simple tag.
 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (approved):
LGTM! Nice work :D

PR #4214, opened 2022-09-08
Fix vector size bug in FreeOutflow boundary condition

yoonso0-0:
## Proposed changes

Current implementation of DG FreeOutflow condition in GRMHD is loading spatial metric and its sqrt determinant from Databox as gridless tags, which means they are retrieved without being projected to an element face. This raises 'Vector size does not match` type of bug when computing conservatives from primitives.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/BoundaryConditions/FreeOutflow.cpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/BoundaryConditions/FreeOutflow.cpp
@@ -72,19 +72,20 @@ std::optional<std::string> FreeOutflow::dg_ghost(
 
     const tnsr::I<DataVector, 3, Frame::Inertial>& interior_shift,
     const Scalar<DataVector>& interior_lapse,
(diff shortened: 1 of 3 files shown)


src/Evolution/Systems/GrMhd/ValenciaDivClean/BoundaryConditions/FreeOutflow.cpp
```diff
@@ -101,6 +105,12 @@ std::optional<std::string> FreeOutflow::dg_ghost(
     }
   }
 
+  determinant_and_inverse(make_not_null(&interior_sqrt_det_spatial_metric),
```
nilsdeppe, line 113: Maybe add a comment as to why this isn't just retrieved? I think I commented on the original PR asking about it :P

nilsdeppe (changes requested):
LGTM, one small request, please squash immediately :)

yoonso0-0:
Rebased and squashed the fix :)

nilsdeppe (approved):

PR #4216, opened 2022-09-09
Add TensorExpression support for std::complex<double> and ComplexDataVector

macedo22:
## Proposed changes

This PR (1) refactors `TensorExpression` code and tests to be more easily extendable to other types and (2) adds support for using `std::complex<double>`, `Tensor<std::complex<double>>`, and `Tensor<ComplexDataVector>` terms in `TensorExpression`s.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
+++ b/src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
@@ -39,7 +39,9 @@ namespace detail {
 ///
 /// \tparam X the arithmetic data type
 template <typename X>
-struct is_supported_number_datatype : std::is_same<X, double> {};
+struct is_supported_number_datatype
+    : std::disjunction<std::is_same<X, double>,
+                       std::is_same<X, std::complex<double>>> {};
 
 template <typename X>
 constexpr bool is_supported_number_datatype_v =
@@ -58,8 +60,9 @@ constexpr bool is_supported_number_datatype_v =
 /// \tparam X the `Tensor` data type
 template <typename X>
 struct is_supported_tensor_datatype
-    : std::disjunction<std::is_same<X, double>, std::is_same<X, DataVector>,
-                       std::is_same<X, ComplexDataVector>> {};
+    : std::disjunction<
+          std::is_same<X, double>, std::is_same<X, std::complex<double>>,
+          std::is_same<X, DataVector>, std::is_same<X, ComplexDataVector>> {};
 
 template <typename X>
 constexpr bool is_supported_tensor_datatype_v =
@@ -164,6 +167,9 @@ struct is_assignable;
 /// Can assign a type to itself
 template <typename LhsDataType, typename RhsDataType>
 struct is_assignable : std::is_same<LhsDataType, RhsDataType> {};
+/// Can assign a complex numeric type to its underlying real-valued numeric type
+template <typename X>
+struct is_assignable<std::complex<X>, X> : std::true_type {};
 /// Can assign the LHS `VectorImpl` to the RHS `VectorImpl` if `VectorImpl`
 /// allows it
 template <typename ValueType1, typename VectorType1, typename ValueType2,
@@ -230,6 +236,18 @@ struct get_binop_datatype_impl<VectorImpl<ValueType, VectorType>,
   using type = VectorType;
 };
 /// @{
+/// A binary operation between a type `T` and `std::complex<T>` yields a
+/// `std::complex<T>`
+template <typename T>
+struct get_binop_datatype_impl<T, std::complex<T>> {
+  using type = std::complex<T>;
+};
+template <typename T>
+struct get_binop_datatype_impl<std::complex<T>, T> {
+  using type = std::complex<T>;
+};
+/// @}
+/// @{
 /// A binary operation between a `VectorImpl` and its underlying value type
 /// yields the `VectorImpl`, e.g. `DataVector OP double = DataVector`
 template <typename ValueType, typename VectorType>
@@ -254,6 +272,32 @@ template <typename ValueType, typename VectorType>
 struct get_binop_datatype_impl<
     ValueType, VectorImpl<std::complex<ValueType>, VectorType>> {
   using type = VectorType;
+};
+/// @}
+/// @{
+/// A binary operation between a real-valued `VectorImpl` and the complex-valued
+/// partner to the `VectorImpl`'s underlying type yields the complex partner
+/// type of the `VectorImpl`, e.g.
+/// `std::complex<double> OP DataVector = ComplexDataVector`
+///
+/// \note Blaze supports multiplication between a `std::complex<double>` and a
+/// `DataVector`, but does not support addition, subtraction, or division
+/// between these two types. This specialization of `get_binop_datatype_impl`
+/// simply defines that the result of any of the binary operations should be
+/// `ComplexDataVector`. Because Blaze doesn't support addition, subtraction,
+/// and division between these two types, the `AddSub` and `Divide` classes
+/// disallow this type combination in their class definitions to prevent these
+/// operations. That way, if Blaze support is later added for e.g. division,
+/// we simply need to remove the assert in `Divide` that prevents it.
+template <typename ValueType, typename VectorType>
+struct get_binop_datatype_impl<VectorImpl<ValueType, VectorType>,
+                               std::complex<ValueType>> {
+  using type = typename get_complex_datatype<VectorType>::type;
+};
+template <typename ValueType, typename VectorType>
+struct get_binop_datatype_impl<std::complex<ValueType>,
+                               VectorImpl<ValueType, VectorType>> {
+  using type = typename get_complex_datatype<VectorType>::type;
 };
 /// @}
 /// @{
--- a/src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
+++ b/src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
@@ -1,0 +1,389 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+/// \file
+/// Defines which types are allowed, whether operations with certain types are
+/// allowed, and other type-specific properties and configuration for
+/// `TensorExpression`s
+///
+/// \details
+/// To add support for a data type, modify the templates in this file and the
+/// arithmetic operator overloads as necessary. Then, add tests as appropriate.
+
+#pragma once
+
+#include <cstddef>
+#include <limits>
+#include <type_traits>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Expressions/TensorExpression.hpp"
+#include "DataStructures/VectorImpl.hpp"
+#include "Utilities/NoSuchType.hpp"
+
+namespace tenex {
+template <typename DataType>
+struct NumberAsExpression;
+
+namespace detail {
+/// @{
+/// \brief Whether or not `TensorExpression`s supports using a given type as a
+/// numeric term
+///
+/// \details
+/// To make it possible to use a new numeric data type as a term in
+/// `TensorExpression`s, add the type to this alias and adjust other templates
+/// in this file, as necessary.
+///
+/// \tparam X the arithmetic data type
+template <typename X>
+struct is_supported_number_datatype : std::is_same<X, double> {};
+
+template <typename X>
+constexpr bool is_supported_number_datatype_v =
+    is_supported_number_datatype<X>::value;
+/// @}
+
+/// @{
+/// \brief Whether or not `Tensor`s with the given data type are currently
+/// supported by `TensorExpression`s
+///
+/// \details
+/// To make it possible to use a new data type in a `Tensor` term in
+/// `TensorExpression`s, add the type to this alias and adjust other templates
+/// in this file, as necessary.
+///
+/// \tparam X the `Tensor` data type
+template <typename X>
+struct is_supported_tensor_datatype
+    : std::disjunction<std::is_same<X, double>, std::is_same<X, DataVector>> {};
+
+template <typename X>
+constexpr bool is_supported_tensor_datatype_v =
+    is_supported_tensor_datatype<X>::value;
+/// @}
+
+/// \brief If the given type is a derived `VectorImpl` type, get the base
+/// `VectorImpl` type, else return the given type
+///
+/// \tparam T the given type
+/// \tparam IsVector whether or not the given type is a `VectorImpl` type
+template <typename T, bool IsVector = is_derived_of_vector_impl_v<T>>
+struct upcast_if_derived_vector_type;
+
+/// If `T` is not a `VectorImpl`, the `type` is just the input
+template <typename T>
+struct upcast_if_derived_vector_type<T, false> {
+  using type = T;
+};
+/// If `T` is a `VectorImpl`, the `type` is the base `VectorImpl` type
+template <typename T>
+struct upcast_if_derived_vector_type<T, true> {
+  using upcasted_type = typename T::BaseType;
+  // if we have a derived VectorImpl, get base type, else T is a base VectorImpl
+  // and we use that
+  using type =
+      tmpl::conditional_t<std::is_base_of_v<MarkAsVectorImpl, upcasted_type>,
+                          upcasted_type, T>;
+};
+
+/// \brief Whether or not a given type is assignable to another within
+/// `TensorExpression`s
+///
+/// \details
+/// This is used to define which types can be assigned to which when evaluating
+/// the result of a `TensorExpression`. For example, you can assign a
+/// `DataVector` to a `double`, but not vice versa.
+///
+/// To enable assignment between two types that is not yet supported, modify a
+/// current template specialization or add a new one.
+///
+/// \tparam LhsDataType the type being assigned
+/// \tparam RhsDataType the type to assign the `LhsDataType` to
+template <typename LhsDataType, typename RhsDataType>
+struct is_assignable;
+
+/// Can assign a type to itself
+template <typename LhsDataType, typename RhsDataType>
+struct is_assignable : std::is_same<LhsDataType, RhsDataType> {};
+/// Can assign a `VectorImpl` to its value type, e.g. can assign a `DataVector`
+/// to a `double`
+template <typename ValueType, typename VectorType>
+struct is_assignable<VectorImpl<ValueType, VectorType>, ValueType>
+    : std::true_type {};
+
+/// \brief Whether or not a given type is assignable to another within
+/// `TensorExpression`s
+///
+/// \details
+/// See `is_assignable` for which assignments are permitted
+template <typename LhsDataType, typename RhsDataType>
+constexpr bool is_assignable_v = is_assignable<
+    typename upcast_if_derived_vector_type<LhsDataType>::type,
+    typename upcast_if_derived_vector_type<RhsDataType>::type>::value;
+
+/// \brief Get the data type of a binary operation between two data types
+/// that may occur in a `TensorExpression`
+///
+/// \details
+/// This is used to define the resulting types of binary arithmetic operations
+/// within `TensorExpression`s, e.g. `double OP double = double` and
+/// `double OP DataVector = DataVector`.
+///
+/// To enable binary operations between two types that is not yet supported,
+/// modify a current template specialization or add a new one.
+///
+/// \tparam X1 the data type of one operand
+/// \tparam X2 the data type of the other operand
+template <typename X1, typename X2>
+struct get_binop_datatype_impl;
+
+/// No template specialization was matched, so it's not a known pairing
+template <typename X1, typename X2>
+struct get_binop_datatype_impl {
+  using type = NoSuchType;
+};
+/// A binary operation between two terms of the same type will yield a result
+/// with that type
+template <typename X>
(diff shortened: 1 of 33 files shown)


src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -0,0 +1,497 @@
...
+/// `VectorImpl` type, else return the given type
+///
+/// \tparam T the given type
+template <typename T, typename = void>
+struct upcast_if_derived_vector_type {
+  using type = T;
+};
+template <typename T>
+struct upcast_if_derived_vector_type<
+    T, typename std::enable_if<is_vector<T>::value>::type> {
```
macedo22: `git` warns to use `Requires` instead of `std::enable_if`, but I wasn't able to figure out how to get it to work in a template specialization. Let me know how I can use `Requires` here or suppress the warning!

wthrowe: I would do this using a plain specialization (untested):
```c++
// Can wrap the bool in a bool_constant if you need to manipulate this in metaprogramming
template <typename T, bool IsVector = is_vector<T>::value>
struct upcast_if_derived_vector_type {
  using type = T;
};
template <typename T>
struct upcast_if_derived_vector_type<T, true> {
  //...
};
```

src/DataStructures/Tensor/Expressions/Evaluate.hpp
```diff
@@ -351,6 +367,13 @@ void evaluate_impl(
       "Cannot assign a tensor expression to a LHS tensor with generic "
       "indices that would be contracted, e.g. evaluate<ti::A, ti::a>.");
 
+  if constexpr (tenex::detail::is_vector<X>::value) {
+    ASSERT(get_size((*lhs_tensor)[0]) > 0,
+           "Tensors with vector components must be sized before calling "
+           "tenex::evaluate<...>("
+           "\tgsl::not_null<Tensor<VectorType, ...>*>, number).");
```
wthrowe: Is the tab character intentional here?  Seems an odd place for one.

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -79,6 +82,47 @@ struct upcast_if_derived_vector_type<
 };
 /// @}
 
+/// \brief Whether or not a given type is the complex-valued parter to another
```
wthrowe: partner

tests/Unit/DataStructures/Tensor/Expressions/Test_EvaluateComplex.cpp
```diff
@@ -0,0 +1,463 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+// \file
+// Tests evaluation of `TensorExpression`s with complex-valued LHS `Tensor`s
+// and RHS expressions that may contain real-valued terms, complex-valued terms,
+// or both
+//
+// \details
+// The tests in this file are designed to test that these funamentally work:
```
wthrowe: fundamentally

wthrowe (changes requested):

src/DataStructures/Tensor/Expressions/Evaluate.hpp
macedo22: I put it here in the hopes that it would make the function more readable in an error message, but looking at this I realize I forgot to add newlines in front of the line above it and in front of the tab, so this would look weird as is.

I'll add newlines unless you think I should scrap the newlines and tabs altogether

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
macedo22: Thanks! I'll try this

src/DataStructures/Tensor/Expressions/Evaluate.hpp
wthrowe: That makes sense.  Tabs and newlines sound fine.

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
macedo22: I also added a definition for the `upcast_if_derived_vector_type<T, false>` specialization when it's not a vector just to make it really clear what happens when

wthrowe: Fine with me.

wthrowe:
Squash.

wthrowe (dismissed):

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -0,0 +1,387 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+/// \file
+/// Defines which types are allowed, whether operations which certain types are
```
nilsdeppe: `which certain` -> `with certain`

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -0,0 +1,387 @@
...
+/// numeric term
+///
+/// \details
+/// To make it possible to use a new numeric data type as a term in
+/// `TensorExpression`s, add the type to this alias and adjust other templates
+/// in this file, as necessary.
+///
+/// \tparam X the arithmetic data type
+template <typename X>
+using is_supported_number_datatype =
```
nilsdeppe: Would it be easier to make these `static constexpr bool is_supported_number_datatype_v = ...::value;`? Or add that? That way you don't need to do `::value` in a bunch of places.

Related question: `std::is_same` inherits off of `std::bool_constant` (ish) and so do you need to write it this way at all? Can you not just do `std::is_same<X, double>`?

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -0,0 +1,387 @@
...
+template <typename X>
+using is_supported_tensor_datatype =
+    std::bool_constant<std::is_same_v<X, double> or
+                       std::is_same_v<X, DataVector>>;
+
+/// \brief Whether or not the given type is a `VectorImpl` type
+///
+/// \tparam T the given type
+template <typename T>
+using is_vector = std::bool_constant<std::is_base_of_v<MarkAsVectorImpl, T>>;
```
nilsdeppe: Isn't `is_derived_of_vector_impl<T>` good enough (defined in VectorImpl.hpp)? Same question about making it `is_vector_v` (or adding that)

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -0,0 +1,387 @@
...
+  using type = T;
+};
+/// If `T` is a `VectorImpl`, the `type` is the base `VectorImpl` type
+template <typename T>
+struct upcast_if_derived_vector_type<T, true> {
+  using upcasted_type = typename T::BaseType;
+  // if we have a derived VectorImpl, get base type, else T is a base VectorImpl
+  // and we use that
+  using type =
+      tmpl::conditional_t<std::is_base_of_v<MarkAsVectorImpl, upcasted_type>,
```
nilsdeppe, line 92: `is_derived_of_vector_impl_v<T>` or the `is_vector_v` you (will) define above?

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -0,0 +1,387 @@
...
+///
+/// \tparam X1 the data type of one operand
+/// \tparam X2 the data type of the other operand
+template <typename X1, typename X2>
+struct get_binop_datatype_impl;
+
+/// No template specialization was matched, so it's not a known pairing
+template <typename X1, typename X2>
+struct get_binop_datatype_impl {
+  using type = std::bool_constant<false>;
```
nilsdeppe: we also have a `NoSuchType` in `Utilities/NoSuchType.hpp` to signal that some sort of processing failed/is invalid :) brigand uses the same approach (we stole it from them, actually :P )

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -0,0 +1,387 @@
...
+///
+/// \details
+/// See `get_binop_datatype_impl` for which data type combinations have a
+/// defined result type
+///
+/// \tparam X1 the data type of one operand
+/// \tparam X2 the data type of the other operand
+template <typename X1, typename X2>
+struct binop_datatypes_are_supported {
+  using type = std::bool_constant<not std::is_same_v<
```
nilsdeppe: why not inherit from `std::is_same<...>`? That'll add `::type` automatically and is normally how this sort of thing is handled.

Apply this comment to anywhere else in this file that you can, please :)

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -82,6 +85,47 @@ struct upcast_if_derived_vector_type<T, true> {
...
+/// No template specialization was matched, so it's not a known pairing
+template <typename MaybeComplexDataType, typename OtherDataType>
+struct is_complex_datatype_of_impl : std::false_type {};
+/// ComplexDataVector is the complex type to DataVector
+template <>
+struct is_complex_datatype_of_impl<typename ComplexDataVector::BaseType,
+                                   typename DataVector::BaseType>
+    : std::true_type {};
+
+/// \brief Whether or not a given type is the complex-valued parter to another
```
nilsdeppe: `parter` -> `partner`

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
```diff
@@ -98,9 +142,12 @@ struct upcast_if_derived_vector_type<T, true> {
 template <typename LhsDataType, typename RhsDataType>
 struct lhs_datatype_is_assignable_to_rhs_datatype_impl;
 
-/// No template specialization was matched, so it's not a known pairing
+/// If the LHS type is the complex-valued partner to the RHS type, we can assign
+/// the LHS type to the RHS type
 template <typename LhsDataType, typename RhsDataType>
-struct lhs_datatype_is_assignable_to_rhs_datatype_impl : std::false_type {};
+struct lhs_datatype_is_assignable_to_rhs_datatype_impl
+    : std::bool_constant<(
```
nilsdeppe: once these things all inherit from `std::BLAH` you can remove the `bool_constant` stuff here too. That's why the inheritance approach is preferred, it cascades to all derived type traits

src/DataStructures/VectorImpl.hpp
```diff
@@ -33,6 +33,46 @@
...
+/// \details
+/// This is used to define which types can be assigned to one another. For
+/// example, you can assign a `ComplexDataVector` to a `DataVector`, but not
+/// vice versa.
+///
+/// To enable assignments between more types, modify a current template
+/// specialization or add a new one.
+///
+/// \tparam LhsDataType the type being assigned
+/// \tparam RhsDataType the type to convert to `LhsDataType`
```
nilsdeppe, line 53: Can all this be checked with expression SFINAE? Like:
```cpp
template <typename T, typename S, std::void_t<>> is_assignable : std::false_type {};

template <typename T, typename S, std::void_t<decltype(std::declval<T>() = std::declval<S>())>> is_assignable :  std::true_type {};
```
and for that matter, https://en.cppreference.com/w/cpp/types/is_assignable might work even. The downside of `std::is_assignable` is that you need full type definitions at places where it's checked, but that should be fine. I'm not sure if it silently fails if the full type definition isn't available, but I assume so and is probably the correct behavior.

tests/Unit/DataStructures/Tensor/Expressions/Test_DataTypeSupport.cpp
```diff
@@ -115,7 +115,7 @@ SPECTRE_TEST_CASE("Unit.DataStructures.Tensor.Expression.DataTypeSupport",
   test_is_supported_number_datatype<double, true>();
```
nilsdeppe: Generalyquestion for this file: this is a very large number of similar checks. Is it possible to generalize them to use things like `tmpl::for_each`-type loops to do the checks? E.g. you could have a `tmpl::list<tmpl::list<std::complex<double>, std::true_type>, ...>`? Or, there might be an even easier way to generalize it, not sure :) But given the length of this file, it might be worth thing about :)

nilsdeppe (changes requested):
Overall looks great! Just some thoughts on type trait definitions, etc. I'd suggest reading through all the comments before starting changes, just to make sure some later comment doesn't cause design collisions :)

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
macedo22: Definitely didn't see this defined there, so yes mine should be scrapped

macedo22: Wait sorry, mine differs in that it evaluates to `true` for both base `VectorImpl` and derived types, not just derived. So I think I will need to keep mine, if I understand `is_derived_of_vector` correctly

nilsdeppe: In what case are you dealing with `VectorImpl` directly?

macedo22: Sorry, just removed my comment. Based on the alias' name, I had thought that `VectorImpl<double, DataVector>` would not be `true` because it isn't a _derived_ type of `VectorImpl`, but I just tested it and it's good. My `is_vector` is redundant, so I'll remove it.

macedo22: I'm dealing with `VectorImpl` directly in a few places where I wanted to extract the underlying value type and vector type, like here:

```
template <typename ValueType, typename VectorType>
struct lhs_datatype_is_assignable_to_rhs_datatype_impl<
    VectorImpl<std::complex<ValueType>, VectorType>, ValueType>
    : std::true_type {};
```

but I'm still reading through your comments, so maybe after I am done reading them, this won't be true anymore

macedo22: But in terms of when checking if something is a vector with `is_vector`, I don't think I'm actually using `VectorImpl`s directly, if that's what you're asking.

Regardless, my `is_vector` can be scrapped because it is equivalent to the existing`is_derived_of_vector_impl<T>`.

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
macedo22: I think this needs to stay as is because the key part here is that this `type` should be the base `VectorImpl` type. Since something like `is_derived_of_vector_impl_v<VectorImpl<double, DataVector>>` will be `true`, this would incorrectly assign this `type` to be the base blaze type like `blaze::CustomVector<...>`, which is not what I want

Let me know if I am misunderstanding. I basically want this to be able to take a type and if it is a derived `VectorImpl`, I want the base `VectorImpl` type, otherwise do nothing to the type.

tests/Unit/DataStructures/Tensor/Expressions/Test_DataTypeSupport.cpp
macedo22: That's a good idea. I will try to make it more concise and general!

src/DataStructures/VectorImpl.hpp
macedo22: I will look into this and use it if it works!

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
macedo22: After thinking about this more, I don't think I follow how this could inherit from `std::is_same` since I need to check if something is `not std::is_same`. The best I can think of is keeping the `std::bool_constant` and inheriting from that like:

```
template <typename X1, typename X2>
struct binop_datatypes_are_supported
    : std::bool_constant<not std::is_same_v<
          typename get_binop_datatype_impl<
              typename upcast_if_derived_vector_type<X1>::type,
              typename upcast_if_derived_vector_type<X2>::type>::type,
          NoSuchType>> {};
```

Did you have something different in mind that I'm not thinking of or is this fine?

nilsdeppe: nah that's fine :)

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
macedo22: Since I have a boolean expression `std::is_same_v<...> or std::is_same_v<...>`, I think I think I can't get rid of `std::bool_constant` because I need to wrap the expression with it. However, I imagine I can inherit from `std::bool_constant`. Does the following work for you or did you have something else in mind that I'm not following?:
```
template <typename X>
struct is_supported_number_datatype
    : std::bool_constant<std::is_same_v<X, double> or
                         std::is_same_v<X, std::complex<double>>> {};
```
Alternatively, I could define a specialization for each and do this instead where we _can_ inherit off of `std:is_same`:
```
template <typename X>
struct is_supported_number_datatype;

template <typename X>
struct is_supported_number_datatype : std::false_type {};
template <>
struct is_supported_number_datatype : std::is_same<X, double>> {};
template <>
struct is_supported_number_datatype : std::is_same<X, std::is_complex<double>>> {};
```

Let me know if you have a preference or a different idea

nilsdeppe: https://en.cppreference.com/w/cpp/types/disjunction

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
nilsdeppe: https://en.cppreference.com/w/cpp/types/negation

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
macedo22: Wow, never seen this but this is a really nice thing to know about

src/DataStructures/Tensor/Expressions/DataTypeSupport.hpp
macedo22: Thanks, I haven't heard of this either, but I'll do the negation

src/DataStructures/VectorImpl.hpp
macedo22: I tried these but the issue I ran into is the opposite of what I would have expected - that I think more assignments would be permitted than we want. Apparently, you can assign a `DataVector` to a `ComplexDataVector`, but I imagine that we don't want that because when would we want that? And I figure in most cases if someone does that, it's an accident and the mistake would be hidden.

Because of this, I did not end up using either of the suggestions here in my fixups.

tests/Unit/DataStructures/Tensor/Expressions/Test_DataTypeSupport.cpp
macedo22: I was able to condense parts of the file a bit but not a ton

macedo22:
@nilsdeppe First fixup addresses your review requests and the second fixup adds support for `std::complex<double> * Tensor<DataVector>()`. When I was working with Alex Carpenter on psi4, it was convenient to be able to do this operation for coding the product `iy` where `y` is a `Tensor<DataVector>`, but not sure if it's something we want to support or not

I grouped all of the fixups in this way for reviewing but when I eventually squash, I will break them up and squash the pieces into the original commits they should belong to.

tests/Unit/DataStructures/Tensor/Expressions/Test_EvaluateComplex.cpp
```diff
@@ -386,60 +409,68 @@ void test_evaluate_large_expressions(
...
+                        64.0 * get(complex_scalar) * complex_number);
 }
 }  // namespace
 
 SPECTRE_TEST_CASE("Unit.DataStructures.Tensor.Expression.EvaluateComplex",
                   "[Unit][DataStructures]") {
   MAKE_GENERATOR(generator);
 
-  const size_t vector_size = 3;
+  const size_t vector_size = 2;
```
nilsdeppe, line 474: Why the size change?

nilsdeppe (commented):
LGTM, please rebase and squash :)

tests/Unit/DataStructures/Tensor/Expressions/Test_EvaluateComplex.cpp
macedo22: I just figured there were a lot of FLOPs happening in this test and this just made it slightly less, but there is not reason it can't still be 3

nilsdeppe: That's fine, was just curious.

nilsdeppe (approved):

nilsdeppe:
@wthrowe would you mind giving feedback on this since the last time you looked? Thank you!

wthrowe (approved):

PR #4218, opened 2022-09-09
ObserveCenters: use Euclidean area element.

markscheel:
## Proposed changes

Uses the Euclidean area element in computation of the Strahlkorper center in the inertial frame.
The center is computed by averaging the coords over the Strahlkorper, where the averaging
is done by a surface integral (with the Euclidean area element).

The Euclidean area element is probably more appropriate, and matches better the behavior of other algorithms to compute the Strahlkorper center in a different frame (such as computing a full Strahlkorper in a different frame and then computing the center; that process is independent of the metric).

Also fixes a bug in how centers were computed, and adds Tags declarations that were missing.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2 (approved):
I've tested this out and the centers that are outputted are now correct. Previously, they were slightly off from what I would have expected.

wthrowe (approved):

PR #4223, opened 2022-09-13
Allow TimescaleTuner to be initialized with arbitary number of timescales

knelli2:
## Proposed changes

This will be necessary when we have shape control so we don't have to type out a bunch of the same number in an input file.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/TimescaleTuner.hpp
+++ b/src/ControlSystem/TimescaleTuner.hpp
@@ -4,6 +4,11 @@
 #pragma once
 
 #include <array>
+#include <cstddef>
+#include <limits>
+#include <optional>
+#include <utility>
+#include <variant>
 
 #include "DataStructures/DataVector.hpp"
 #include "Options/Options.hpp"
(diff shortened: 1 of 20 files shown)


src/ControlSystem/TimescaleTuner.hpp
```diff
@@ -45,14 +50,33 @@ class er;
  */
 
 class TimescaleTuner {
+ private:
+  struct TimescaleCreator {
```
wthrowe: Move this class to an anonymous namespace in the cpp file.

tests/Unit/ControlSystem/Test_Tags.cpp
```diff
@@ -29,7 +31,46 @@ struct LabelA {};
 using system = control_system::TestHelpers::System<
     2, LabelA, control_system::TestHelpers::Measurement<LabelA>>;
 
-struct MetavarsEmpty {};
+struct MetavarsEmpty {
+  static constexpr size_t volume_dim = 3;
+};
+
+class FakeCreator : public DomainCreator<3> {
```
wthrowe: Can this test use the `FakeCreator` class added in the previous commit?  They look the same except the other one has option-parsing stuff.

wthrowe (changes requested):

knelli2:
Squashed suggested changes directly and rebased.

wthrowe (dismissed):

knelli2:
@wthrowe I realized I introduced a bug with these changes when I was testing it with the shape map. The `control_system::Tags::Controller` tag uses a timescale tuner when it is created from options, but that tuner wasn't initialized if the number of components hadn't been set yet. The fixup fixes this.

wthrowe:
OK, squash.

wthrowe (approved):

PR #4232, opened 2022-09-17
Add Shape control system

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Shape.hpp
+++ b/src/ControlSystem/ControlErrors/Shape.hpp
@@ -1,0 +1,152 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cmath>
+#include <cstddef>
+#include <pup.h>
+#include <string>
+
+#include "ApparentHorizons/ObjectLabel.hpp"
+#include "ControlSystem/ApparentHorizons/Measurements.hpp"
+#include "ControlSystem/Protocols/ControlError.hpp"
+#include "ControlSystem/Tags.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/SpherepackIterator.hpp"
+#include "Options/Options.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ForceInline.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+/// \cond
+namespace domain::Tags {
+template <size_t Dim>
+struct Domain;
+struct FunctionsOfTime;
+}  // namespace domain::Tags
+namespace Frame {
+struct Grid;
+}  // namespace Frame
+/// \endcond
+
+namespace control_system {
+namespace ControlErrors {
+namespace detail {
+template <::ah::ObjectLabel Horizon>
+std::string excision_sphere_name() {
+  return "Object"s + ::ah::name(Horizon) + "ExcisionSphere"s;
+}
+
+template <::ah::ObjectLabel Horizon>
+std::string size_name() {
+  return "Size"s + ::ah::name(Horizon);
+}
+}  // namespace detail
+
+/*!
+ * \brief Control error in the
+ * \link domain::CoordinateMaps::TimeDependent::Shape Shape \endlink coordinate
+ * map
+ *
+ * \details Computes the error for each \f$ l,m \f$ mode of the shape map except
+ * for \f$ l=0 \f$ and \f$ l=1 \f$. This is because the \f$ l=0 \f$ mode is
+ * controlled by characteristic speed control, and the \f$ l=1 \f$ mode is
+ * redundant with the \link
+ * domain::CoordinateMaps::TimeDependent::Translation Translation \endlink map.
+ * The equation for the control error is Eq. (77) from \cite Hemberger2012jz
+ * which is
+ *
+ * \f{align}
+ * Q_{lm} &= -\frac{\sqrt{\frac{\pi}{2}} r_\mathrm{EB} -
+ * Y_{00}\lambda_{00}(t)}{Y_{00}S_{00}} S_{lm} - \lambda_{lm}(t), \quad l>1
+ * \label{eq:shape_control_error}
+ * \f}
+ *
+ * where \f$ r_\mathrm{EB} \f$ is the radius of the excision boundary in the
+ * grid frame, \f$ \lambda_{00}(t) \f$ is the size map parameter, \f$
+ * \lambda_{lm}(t) \f$ for \f$ l>1 \f$ are the shape map parameters, and \f$
+ * S_{lm}\f$ are the coefficients of the harmonic expansion of the apparent
+ * horizon. The coefficients \f$ \lambda_{lm}(t) \f$ (including \f$ l=0 \f$) and
+ * \f$ S_{lm}\f$ are stored as the real-valued coefficients \f$ a_{lm} \f$ and
+ * \f$ b_{lm} \f$ of the SPHEREPACK spherical-harmonic expansion (in
+ * YlmSpherepack) as opposed to complex coefficients \f$ A_{lm} \f$ of the
+ * standard \f$ Y_{lm} \f$ decomposition. The representation does not matter
+ * here except for the term involving \f$ r_\mathrm{EB} \f$. Because \f$ a_{00}
+ * = \sqrt{\frac{2}{\pi}}A_{00} \f$, there is an extra factor of \f$
+ * \sqrt{\frac{2}{\pi}} \f$ in the above formula.
+ *
+ * Requirements:
+ * - This control error requires that there be at least one excision surface in
+ *   the simulation
+ * - Currently this control error can only be used with the \link
+ *   control_system::Systems::Shape Shape \endlink control system
+ */
+template <::ah::ObjectLabel Horizon>
+struct Shape : tt::ConformsTo<protocols::ControlError> {
+  using options = tmpl::list<>;
+  static constexpr Options::String help{
+      "Computes the control error for shape control. This should not take any "
+      "options."};
+
+  void pup(PUP::er& /*p*/) {}
+
+  template <typename Metavariables, typename... TupleTags>
+  DataVector operator()(const Parallel::GlobalCache<Metavariables>& cache,
+                        const double time,
+                        const std::string& function_of_time_name,
+                        const tuples::TaggedTuple<TupleTags...>& measurements) {
+    const auto& domain = get<domain::Tags::Domain<3>>(cache);
+    const auto& functions_of_time = get<domain::Tags::FunctionsOfTime>(cache);
+    const DataVector lambda_lm_coefs =
+        functions_of_time.at(function_of_time_name)->func(time)[0];
+    const double lambda_00_coef =
+        functions_of_time.at(detail::size_name<Horizon>())->func(time)[0][0];
+
+    const auto& ah =
+        get<control_system::QueueTags::Strahlkorper<Frame::Grid>>(measurements);
+    const auto& ah_coefs = ah.coefficients();
+
+    ASSERT(lambda_lm_coefs.size() == ah_coefs.size(),
+           "Number of coefficients for shape map '"
+               << function_of_time_name << "' (" << lambda_lm_coefs.size()
+               << ") does not match the number of coefficients in the AH "
+                  "Strahlkorper ("
+               << ah_coefs.size() << ").");
+
+    const auto& excision_spheres = domain.excision_spheres();
+    // See above docs for why we have the sqrt(pi/2)
+    const double radius_excision_sphere_grid_frame =
+        sqrt(0.5 * M_PI) *
+        excision_spheres.at(detail::excision_sphere_name<Horizon>()).radius();
+
+    const double Y00 = sqrt(0.25 / M_PI);
+    SpherepackIterator iter{ah.l_max(), ah.m_max()};
+    const double relative_size_factor =
+        (radius_excision_sphere_grid_frame / Y00 - lambda_00_coef) /
+        ah_coefs[iter.set(0, 0)()];
+
+    // The map parameters are in terms of SPHEREPACK coefficients (just like
+    // strahlkorper coefficients), *not* spherical harmonic coefficients, thus
+    // the control error for each l,m is in terms of SPHEREPACK coefficients
+    // and no extra factors of sqrt(2/pi) are needed
+    DataVector Q = -relative_size_factor * ah_coefs - lambda_lm_coefs;
+
+    // Shape control is only for l > 1 so enforce that Q=0 for l=0,l=1. These
+    // components of the control error won't be 0 automatically because the AH
+    // can freely have nonzero l=0 and l=1 coefficients so we have to set the
+    // control error components to be 0 manually.
+    Q[iter.set(0, 0)()] = 0.0;
+    Q[iter.set(1, -1)()] = 0.0;
+    Q[iter.set(1, 0)()] = 0.0;
+    Q[iter.set(1, 1)()] = 0.0;
+
+    return Q;
+  }
+};
+}  // namespace ControlErrors
+}  // namespace control_system
--- a/tests/Unit/ControlSystem/ControlErrors/CMakeLists.txt
+++ b/tests/Unit/ControlSystem/ControlErrors/CMakeLists.txt
@@ -5,6 +5,7 @@ set(LIBRARY_SOURCES
   ${LIBRARY_SOURCES}
   ControlErrors/Test_Expansion.cpp
   ControlErrors/Test_Rotation.cpp
+  ControlErrors/Test_Shape.cpp
   ControlErrors/Test_SizeControlStates.cpp
   ControlErrors/Test_Translation.cpp
   PARENT_SCOPE)
--- a/tests/Unit/ControlSystem/ControlErrors/Test_Shape.cpp
+++ b/tests/Unit/ControlSystem/ControlErrors/Test_Shape.cpp
@@ -1,0 +1,184 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <array>
+#include <cmath>
+#include <cstddef>
+#include <limits>
+#include <memory>
+#include <random>
+#include <string>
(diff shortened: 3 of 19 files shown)


src/ControlSystem/ControlErrors/Shape.hpp
```diff
@@ -0,0 +1,149 @@
...
+namespace Frame {
+struct Grid;
+}  // namespace Frame
+/// \endcond
+
+namespace control_system {
+namespace ControlErrors {
+namespace detail {
+template <::ah::ObjectLabel Horizon>
+std::string excision_sphere_name() {
```
wthrowe, line 41: `#include <string>`

src/ControlSystem/ControlErrors/Shape.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+template <::ah::ObjectLabel Horizon>
+std::string size_name() {
+  return "Size"s + ::ah::name(Horizon);
+}
+
+// This is the Y00 spherical harmonic expressed in terms of SPHEREPACK
+// coefficients which has an extra factor of sqrt(2/pi) in front of the
+// original coefficient sqrt(1/4pi)
+SPECTRE_ALWAYS_INLINE double y00_spherepack_coef() { return sqrt(0.5) / M_PI; }
```
wthrowe: `#include "Utilities/ForceInline.hpp"`

src/ControlSystem/ControlErrors/Shape.hpp
```diff
@@ -0,0 +1,149 @@
...
+    const auto& domain = get<domain::Tags::Domain<3>>(cache);
+    const auto& functions_of_time = get<domain::Tags::FunctionsOfTime>(cache);
+    const DataVector lambda_lm_coefs =
+        functions_of_time.at(function_of_time_name)->func(time)[0];
+    const double lambda_00_coef =
+        functions_of_time.at(detail::size_name<Horizon>())->func(time)[0][0];
+
+    const auto& ah =
+        get<control_system::QueueTags::Strahlkorper<Frame::Grid>>(measurements);
+    const auto ah_coefs = ah.coefficients();
```
wthrowe: Reference?

src/ControlSystem/ControlErrors/Shape.hpp
```diff
@@ -0,0 +1,149 @@
...
+
+    ASSERT(lambda_lm_coefs.size() == ah_coefs.size(),
+           "Number of coefficients for shape map '"
+               << function_of_time_name << "' (" << lambda_lm_coefs.size()
+               << ") does not math the number of coefficients in the AH "
```
wthrowe: "math" -> "match"

tests/Unit/ControlSystem/ControlErrors/CMakeLists.txt
```diff
@@ -6,5 +6,6 @@ set(LIBRARY_SOURCES
   ControlErrors/Test_Expansion.cpp
   ControlErrors/Test_Rotation.cpp
   ControlErrors/Test_SizeControlStates.cpp
+  ControlErrors/Test_Shape.cpp
```
wthrowe, line 8: 'h' < 'i'

tests/Unit/ControlSystem/ControlErrors/Test_Shape.cpp
```diff
@@ -0,0 +1,187 @@
...
+}  // namespace Frame
+
+namespace control_system {
+namespace {
+using FoTMap = std::unordered_map<
+    std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>;
+// We use the distorted frame here because that is where the apparent horizons
+// will be found, however, for this test it doesn't really matter because we
+// randomize the initial data
+using Strahlkorper = Strahlkorper<Frame::Grid>;
```
wthrowe: This comment doesn't seem to match the code.

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+
+template <typename Generator, typename Metavars, size_t DerivOrder>
+void test_shape_control(
+    const gsl::not_null<Generator*> generator,
+    const gsl::not_null<SystemHelper<Metavars>*> system_helper,
+    const double initial_time, const double final_time, const size_t l_max,
+    const domain::FunctionsOfTime::PiecewisePolynomial<DerivOrder>&
+        ah_coefs_function_of_time,
+    const double func_eps, const double deriv_eps) {
+  // First 3 zeros are for translation, rotation, and expansion
```
wthrowe: What is this comment referring to?

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+  using system = typename metavars::shape_system;
+
+  // Responsible for running all control system checks
+  SystemHelper<metavars> system_helper{};
+  SpherepackIterator iter{l_max, l_max};
+  const size_t num_ah_coeffs = iter.spherepack_array_size();
+  std::uniform_real_distribution<double> coef_dist{-0.1, 0.1};
+  // Hard code the radius to 1 because shape doesn't depend on the size of the
+  // strahlkorper
+  const double ah_radius = 1.0;
```
wthrowe: If the value is irrelevant don't use 1.

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+  const double initial_time = 0.0;
+  const double final_time = 150.0;
+
+  // Setup initial shape map coefficients. In the map the coefficients are
+  // stored as the negative of the actual spherical harmonic coefficients
+  // because that's just how the map is defined. But since these are random
+  // numbers it doesn't matter for initial data. Also set initial size map
+  // coefficient. Here we only test non-changing size. A test with a changing
+  // size parameter can be added in later in needed.
```
wthrowe: "in" (second) -> "if"

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+  {
+    INFO(deriv_order_string + "Stationary spherical AH");
+    // Set all coefs back to zero first
+    initial_ah_coefs = zero_array;
+    initial_ah_coefs[0][iter.set(0, 0)()] = ah_radius;
+    expected_ah_coefs = {initial_time, initial_ah_coefs,
+                         std::numeric_limits<double>::infinity()};
+    test_shape_control(generator, make_not_null(&system_helper), initial_time,
+                       final_time, l_max, expected_ah_coefs, deriv_eps,
+                       deriv_eps);
```
wthrowe: Most of these calls use `deriv_eps` twice.  Is this intentional?

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+    // Ensure that radius of AH is positive and constant
+    initial_ah_coefs[0][iter.set(0, 0)()] = ah_radius;
+    initial_ah_coefs[1] = zero_dv;
+    initial_ah_coefs[2] = zero_dv;
+    expected_ah_coefs = {initial_time, initial_ah_coefs,
+                         std::numeric_limits<double>::infinity()};
+    test_shape_control(generator, make_not_null(&system_helper), initial_time,
+                       final_time, l_max, expected_ah_coefs, deriv_eps,
+                       deriv_eps);
+    (void)func_eps;
```
wthrowe: Remove.

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+    expected_ah_coefs = {initial_time, initial_ah_coefs,
+                         std::numeric_limits<double>::infinity()};
+    test_shape_control(generator, make_not_null(&system_helper), initial_time,
+                       final_time, l_max, expected_ah_coefs, deriv_eps,
+                       deriv_eps);
+  }
+  system_helper.reset();
+  {
+    INFO(deriv_order_string +
+         "AH coefficients quadratically increasing/decreasing");
```
wthrowe, line 335: I don't understand this case.  It looks like it will have a cubic part if DerivOrder is 3.  Also, why the division by `cube(i + 5)`?

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+      if (DerivOrder > 2) {
+        initial_ah_coefs[3][iter()] =
+            -cube(freq) * cos(freq * initial_time + offset);
+      }
+    }
+    // Ensure that radius of AH is positive and constant
+    initial_ah_coefs = amplitude * initial_ah_coefs;
+    initial_ah_coefs[0][iter.set(0, 0)()] = ah_radius;
+    initial_ah_coefs[1][iter.set(0, 0)()] = 0.0;
+    initial_ah_coefs[2][iter.set(0, 0)()] = 0.0;
```
wthrowe, line 400: Should this set `[3]` as well?

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+  // based off the constant offset in the control error.
+  const double custom_approx_func_eps = 5.0e-3;
+  // This epsilon is mostly necessary for deriv order 3. For most of deriv order
+  // 2, we can use an epsilon of 5.0e-10, but occasionally there are cases where
+  // the error for deriv order 2 is larger than 5.0e-10 so to be safe we just
+  // use 5.0e-5 all the time.
+  const double custom_approx_deriv_eps = 5.0e-5;
+
+  std::vector<size_t> ells_to_test{2, 5, 7};
+  // std::vector<size_t> ells_to_test{2};
```
wthrowe: Remove.

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+}
+
+// [[Timeout, 15]]
+SPECTRE_TEST_CASE("Unit.ControlSystem.Systems.Shape", "[ControlSystem][Unit]") {
+  MAKE_GENERATOR(generator);
+  domain::FunctionsOfTime::register_derived_with_charm();
+
+  test_names();
+
+  // For some of the faster changing AHs, (quadratic and sinusoid) the control
```
wthrowe: What counts as fast?  I wouldn't usually consider a sinusoid with a period of at least 150 to be rapidly varying, although I'm not sure what the relevant timescale to compare to is.

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -0,0 +1,513 @@
...
+  // system isn't able to perfectly match the expected map parameters, but it is
+  // able to get the control error to be a small constant offset (rather than
+  // 0). This larger epsilon accounts for the difference in the map parameters
+  // (not their derivatives though which it is able to more accurately match)
+  // based off the constant offset in the control error.
+  const double custom_approx_func_eps = 5.0e-3;
+  // This epsilon is mostly necessary for deriv order 3. For most of deriv order
```
wthrowe: Is it expected that a higher-order control system performs significantly worse?

wthrowe (changes requested):

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
knelli2: Yes because `deriv_eps` is stricter and for these tests that are stationary or linear, both the function and derivative can be checked using this smaller epsilon.

But I agree, the name is sort of confusing. Maybe I'll rename them to `stricter_eps` and `looser_eps` to avoid tying their meaning to func vs deriv.

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
knelli2: Not necessarily faster, but I noticed the constant offset for the quadratic and sinusoidal coefficients of the AHs, and I've seen this behavior in SpEC sometimes.

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
knelli2: These numbers are pretty old. I don't think they are true anymore. I'll remove the confusing comment

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -178,16 +174,16 @@ void test_suite(const gsl::not_null<Generator*> generator, const size_t l_max,
   std::uniform_real_distribution<double> coef_dist{-0.1, 0.1};
   // Hard code the radius to 1 because shape doesn't depend on the size of the
   // strahlkorper
-  const double ah_radius = 1.0;
+  const double ah_radius = 1.63;
```
wthrowe: Update comment.

wthrowe (changes requested):
Squash.

wthrowe (dismissed):

src/ControlSystem/ControlErrors/Shape.hpp
```diff
@@ -0,0 +1,151 @@
...
+
+template <::ah::ObjectLabel Horizon>
+std::string size_name() {
+  return "Size"s + ::ah::name(Horizon);
+}
+
+// This is the Y00 spherical harmonic expressed in terms of SPHEREPACK
+// coefficients which has an extra factor of sqrt(2/pi) in front of the
+// original coefficient sqrt(1/4pi)
+SPECTRE_ALWAYS_INLINE double y00_spherepack_coef() { return sqrt(0.5) / M_PI; }
```
markscheel: The Y00 spherical harmonic is a function, not a coefficient, so I don't really understand the comment. If you really need this quantity, please name it something like y00_times_sqrt_two_over_pi (although that name is ambiguous too; there must be a better name).

src/ControlSystem/ControlErrors/Shape.hpp
```diff
@@ -0,0 +1,151 @@
...
+ * for \f$ l=0 \f$ and \f$ l=1 \f$. This is because the \f$ l=0 \f$ mode is
+ * controlled by characteristic speed control, and the \f$ l=1 \f$ mode is
+ * redundant with the \link
+ * domain::CoordinateMaps::TimeDependent::Translation Translation \endlink map.
+ * The equation for the control error is Eq. (77) from \cite Hemberger2012jz
+ * which is
+ *
+ * \f[ Q_{lm} = -\frac{r_\mathrm{EB} - Y_{00}\lambda_{00}(t)}{Y_{00}S_{00}}
+ * S_{lm} - \lambda_{lm}(t), \quad l>1
+ * \f]
```
markscheel: Please put the factor of sqrt(pi/2) here explicitly. Here Y_00 isn't really Y_00 because lambda_00 and S_00 aren't really the spherical harmonic coefficients.
I think if you multiply `r_\mathrm{EB}` by `sqrt(pi/2)` and leave the rest of the equation as-is, then it is correct, with the standard definition of Y_00.  I wouldn't call something Y_00 if it is not actually Y_00.

And I think the note below should be replaced with a statement right here, explaining where the sqrt(pi/2) comes from.  Something like
```
* where \f$ r_\mathrm{EB} \f$ is the radius of the excision boundary in the
* grid frame, \f$ \lambda_{00}(t) \f$ is the size map parameter, \f$ \lambda_{lm}(t) \f$ for
* \f$l!=0\f$ are the shape map parameters,
*  and \f$ S_{lm}\f$ are the coefficients of the harmonic expansion of the apparent horizon.
* The coefficients \f$ \lambda_{lm}(t) \f$ (including \f$l=0\f$) and \f$ S_{lm}\f$ are stored as 
* the real-valued coefficients
* \f$a_{lm}\f$ and \f$b_{lm}$ of the SPHEREPACK spherical-harmonic expansion 
*  (equation 1 in YlmSpherepack.hpp [can doxygen reference equations in other files
*  without explicitly giving equation numbers?  And can it link to them?]) as opposed to complex coefficients
* \f$A_{lm}$ of the standard \f$Y_{lm}\f$ decomposition. 
* The representation does not matter here except for the term
* involving  \f$r_\mathrm{EB}\f$. Because \f$a_{00} = \sqrt(2/pi)A_{00}\f$, there is an extra
* \f$sqrt(2/pi)\f$ in Eq. BLA above.
```

I will submit a new PR that will add more documentation to YlmSpherepack, where I will add formulas for transformations between the Spherepack coefficients and the standard Ylm decomposition, and also real-valued spherical harmonics which we sometimes use.

src/ControlSystem/ControlErrors/Shape.hpp
```diff
@@ -0,0 +1,151 @@
...
+               << function_of_time_name << "' (" << lambda_lm_coefs.size()
+               << ") does not match the number of coefficients in the AH "
+                  "Strahlkorper ("
+               << ah_coefs.size() << ").");
+
+    const auto& excision_spheres = domain.excision_spheres();
+    const double radius_excision_sphere_grid_frame =
+        excision_spheres.at(detail::excision_sphere_name<Horizon>()).radius();
+
+    // Y00 coefficient expressed as SPHEREPACK coefficient
```
markscheel: Please reword; Y00 is not a coefficient.

markscheel (changes requested):

src/ControlSystem/ControlErrors/Shape.hpp
knelli2: Couldn't figure out equation referencing with doxygen so I just linked the whole `YlmSpherepack` doc instead.

markscheel (dismissed):
Looks good! Pls squash.

wthrowe (changes requested):
The code changes look fine, but the Unit.ControlSystem.Systems.Shape test is timing out sometimes on CI even with a limit of 15 seconds.

Skimming the code, I see that, in addition to whatever the control system test helper is doing, this test is doing 6 runs with 10000 time steps each.  Can that be reduced?

knelli2:
@wthrowe That was just for calculating an expected function of time to compare against for that case. I used a lot of updates just so it would be smooth but I guess I went a bit overboard. Reduced the number of updates/time steps for that by a factor of 10.

wthrowe:
Yes, it looks like my guess was wrong and that wasn't the problem.  We're still getting timeouts.

knelli2:
Having larger $\ell$ can slow things down because of the number of components. Should I remove the largest $\ell$ I test and only have the first two?

wthrowe:
Yes, let's try that.

wthrowe (approved):
Test time looks much better now.

PR #4236, opened 2022-09-17
Treat deadlock/hangs in iterable actions as failed execution

nilsdeppe:
## Proposed changes

Only commit `Check that all components have terminated before exiting` is new.

Does a reduction over each component to check that `terminate` is set to `true` (the component has explicitly acknowledged it is done). All components are checked and it is reported which ones did not terminate cleanly. Currently a component is marked as `terminate==true` if there are no iterable actions in the algorithm. Since all simple, threaded, and reduction actions are guaranteed to have completed before we enter the `Exit` phase, this really can only check against deadlocks of iterable actions. However, this covers most use-cases.

Update:
I've added a commit that allows running simple actions to analyze why a deadlock occurs.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/Main.hpp
+++ b/src/Parallel/Main.hpp
@@ -46,10 +46,16 @@
 #include "Utilities/TMPL.hpp"
 #include "Utilities/TaggedTuple.hpp"
 #include "Utilities/TypeTraits/CreateGetTypeAliasOrDefault.hpp"
+#include "Utilities/TypeTraits/CreateIsCallable.hpp"
 
 #include "Parallel/Main.decl.h"
 
 namespace Parallel {
+namespace detail {
+CREATE_IS_CALLABLE(run_deadlock_analysis_simple_actions)
+CREATE_IS_CALLABLE_V(run_deadlock_analysis_simple_actions)
+}  // namespace detail
+
 /// \ingroup ParallelGroup
 /// The main function of a Charm++ executable.
 /// See [the Parallelization documentation](group__ParallelGroup.html#details)
@@ -127,6 +133,10 @@ class Main : public CBase_Main<Metavariables> {
   /// This allows detecting deadlocks in iterable actions, but not simple or
   /// reduction action.
   void did_all_elements_terminate(bool all_elements_terminated);
+
+  /// Prints exit info and stops the executable with failure if a deadlock was
+  /// detected.
+  void post_deadlock_analysis_termination();
 
  private:
   // Return the dir name for the next Charm++ checkpoint as well as the pieces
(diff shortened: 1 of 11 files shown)


nilsdeppe:
Rebased on develop, no longer dependent

src/Parallel/Main.hpp
```diff
@@ -713,8 +722,8 @@ void Main<Metavariables>::execute_next_phase() {
   }
 
   if (Parallel::Phase::Exit == current_phase_) {
-    Informer::print_exit_info();
-    sys::exit();
+    did_all_elements_terminate(true);
```
knelli2: At first, I was a bit confused about why you were calling a reduction target of Main *on* Main. But then I realized that it won't actually do anything in `did_all_elements_terminate` except call the `contribute_termination_status_to_main()` entry on the chare corresponding to `current_termination_check_index_` (which should be 0 at this point).

How do you feel about adding a private member function to Main called `check_if_next_component_terminated_properly` that does
```cpp
auto* global_cache = Parallel::local_branch(global_cache_proxy_);
  ASSERT(global_cache != nullptr, "Could not retrieve the local global cache.");

  tmpl::for_each<component_list>([global_cache, this](auto component_tag_v) {
    using component_tag = tmpl::type_from<decltype(component_tag_v)>;
    if (tmpl::index_of<component_list, component_tag>::value ==
        current_termination_check_index_) {
      Parallel::get_parallel_component<component_tag>(*global_cache)
          .contribute_termination_status_to_main();
    }
  });
```

And calling that here and at the end of `did_all_elements_terminate` so you don't overuse a reduction target? You could even add an ASSERT/ERROR here that checks that `current_termination_check_index_ == 0`

src/Parallel/Main.hpp
```diff
@@ -799,6 +808,60 @@ void Main<Metavariables>::add_exception_message(std::string exception_message) {
       });
 }
 
+template <typename Metavariables>
+void Main<Metavariables>::did_all_elements_terminate(
+    const bool all_elements_terminated) {
+  if (not all_elements_terminated) {
+    if (UNLIKELY(current_termination_check_index_ == 0)) {
```
knelli2: If you make the above change about `check_if_next_component_terminated_properly`, you can get rid of this check here.

src/Parallel/Main.hpp
```diff
@@ -119,6 +120,8 @@ class Main : public CBase_Main<Metavariables> {
   /// quiescence occurs soon after the exception is reported.
   void add_exception_message(std::string exception_message);
 
+  void did_all_elements_terminate(bool all_elements_terminated);
```
knelli2: Docs

knelli2 (commented):
Implementation is ok. These comments are more organizational to make things a bit easier to understand conceptually and facilitate code readability as I was initially confused. They are optional if you want to leave it as is.

knelli2 (approved):
Looks good!

wthrowe (approved):

PR #4237, opened 2022-09-19
Put GR variables on right grid for FixToAtmosphere

yoonso0-0:
## Proposed changes
Merging [Will's commit](https://github.com/wthrowe/spectre/commit/f5e156c7cd1c0e1c723a8356e5066ee13d12353b) which swaps the execution order of two initialization actions for GRMHD subcell.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


yoonso0-0:
I actually have a question for this : why do we need to put GR variables on subcell (I guess this is what's done here) mesh first then apply atmosphere fixing?

wthrowe:
Because atmosphere fixing (sometimes) needs to calculate things like `v^i g_ij v^j`, which requires the fluid velocity and spatial metric to be on the same grid.  (SwapGrTags only actually swaps if the variables are on the wrong grid.)

nilsdeppe (approved):

PR #4238, opened 2022-09-20
Make AlgorithmArray patch work with macOS 13

geoffrey4444:
## Proposed changes

After upgrading to the macOS 13 public beta, I found that the charm patch for `AlgorithmArray.def.h` no longer worked; instead, I would get an error from the patch command that applying the patch failed.

The patch simply adds a pair of () on one line in `AlgorithmArray.def.h.`

I find that the following equivalent patch, differing only in updating the line numbers and in adding 3 lines of context on either side of the change, works for me on macOS 13.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupCharmModuleFunctions.cmake
+++ b/cmake/SetupCharmModuleFunctions.cmake
@@ -11,7 +11,11 @@ function(add_charm_module MODULE)
     add_custom_target(module_All)
   endif()
 
-  set(_VERSION_SUFFIX "_v${CHARM_VERSION}")
+  if(APPLE)
+    set(_VERSION_SUFFIX "_macOS_v${CHARM_VERSION}")
+  else()
+    set(_VERSION_SUFFIX "_v${CHARM_VERSION}")
+  endif()
 
   add_custom_command(
     OUTPUT ${MODULE}.decl.h ${MODULE}.def.h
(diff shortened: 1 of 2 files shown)


nilsdeppe (changes requested):
Please add a different patch file (might need to generalize CMake to do the check?) because the current patch took a while to get working. This change will break some build somewhere, either older versions of macOS, or some Linux builds. I don't remember if CMake has the ability to discover and check different OS patches, but that could be added.

geoffrey4444:
> Please add a different patch file (might need to generalize CMake to do the check?) because the current patch took a while to get working. This change will break some build somewhere, either older versions of macOS, or some Linux builds. I don't remember if CMake has the ability to discover and check different OS patches, but that could be added.

I'll look into having cmake select a different patch file on macOS

geoffrey4444:
> Please add a different patch file (might need to generalize CMake to do the check?) because the current patch took a while to get working. This change will break some build somewhere, either older versions of macOS, or some Linux builds. I don't remember if CMake has the ability to discover and check different OS patches, but that could be added.

@nilsdeppe I just force-pushed a change so that my changes are in a separate patch file, as you requested. That file is only used `if(APPLE)`...otherwise, it uses the existing patch file, which is now unchanged.

nilsdeppe (approved):

PR #4239, opened 2022-09-20
Use angular ordering in Sphere intrp target

knelli2:
## Proposed changes

Previously, only the `KerrHorizon` target was able to have a different ordering for it's spherical harmonic points, making it the only target that could be used with CCE. Now you can also choose the angular ordering of the `Sphere` target, so it can also be used with CCE.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Targets/AngularOrdering.cpp
+++ b/src/ParallelAlgorithms/Interpolation/Targets/AngularOrdering.cpp
@@ -1,0 +1,37 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ParallelAlgorithms/Interpolation/Targets/AngularOrdering.hpp"
+
+#include <string>
+
+#include "Options/Options.hpp"
+#include "Options/ParseOptions.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+
+namespace intrp {
+std::ostream& operator<<(std::ostream& os, const AngularOrdering ordering) {
+  switch (ordering) {
+    case AngularOrdering::Strahlkorper:
+      return os << "Strahlkorper";
+    case AngularOrdering::Cce:
+      return os << "Cce";
+    default:
+      ERROR("Unknown AngularOrdering type");
+  }
+}
+}  // namespace intrp
+
+template <>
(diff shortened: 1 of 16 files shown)


Sizheng-Ma:
I've checked that the `Sphere` target leads to the exact CCE waveform that one would expect. And GH+CCE doesn't crash anymore.

src/ParallelAlgorithms/Interpolation/Targets/AngularOrdering.cpp
```diff
@@ -0,0 +1,42 @@
...
+
+#include "ParallelAlgorithms/Interpolation/Targets/AngularOrdering.hpp"
+
+#include <string>
+
+#include "Options/Options.hpp"
+#include "Options/ParseOptions.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+
+namespace Frame {
```
nilsdeppe: why do you need these forward decls here?

src/ParallelAlgorithms/Interpolation/Targets/AngularOrdering.hpp
```diff
@@ -0,0 +1,39 @@
...
+
+namespace intrp {
+/*!
+ * \brief Label for the ordering of spherical harmonic points on a sphere
+ *
+ * \details `%Strahlkorper` refers to points on a sphere ordered by SPHEREPACK
+ * because `Strahlkorper`s hold YlmSpherePacks internally. `%CCE` refers to
+ * points on a sphere ordered by Libsharp because CCE uses Libsharp internally.
+ */
+enum class AngularOrdering { Strahlkorper, CCE };
```
nilsdeppe: CCE -> Cce

nilsdeppe (commented):
Please squash immediately :)

src/ParallelAlgorithms/Interpolation/Targets/AngularOrdering.cpp
knelli2: Oops! Relic of a previous version.

nilsdeppe (approved):

markscheel (approved):

PR #4240, opened 2022-09-21
Make receive_data an inline entry method

knelli2:
## Proposed changes

Add the `[inline]` keyword to the `receive_data` entry method on all parallel components. This will speed up communication when the sender and receiver are on the same charm processing element (PE, usually the same core).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2:
Adding the `in progress` label because I want to test out these changes on a few more machines before I'm confident that this doesn't break anything or slow things down.

I've already tested this on wheeler and see improvements to BBH and ScalarWave over the current develop.

knelli2:
I've tested this out with BBH on Wheeler, Frontera, and Ocean against the latest develop. It's hard to give exact numbers since I didn't do a detailed comparison. For all these machines and both inline enabled vs disabled, I did two tests.
1. I used our "usual" number of nodes for BBH which amounts to ~10 elements per core for all the machines (I found this as the optimal number earlier this year when doing different testing).
2. I increased the number of elements per core by 4-6 times to roughly 40-60 elements per core (depending on the machine).

For test 1. I saw no improvement on any machine for inline enabled vs disabled. This is not unexpected because inlining is useful when more chares are on the same processing element (core). With only ~10 elements per core, we still have to go through the pupping framework quite often so I didn't expect much improvement. Glad to see there was no decrease in efficiency.

For test 2. I did see noticeable speed up on each machine. The inline test was around 10% faster than the non-inline test. This is expected because more elements per core means that more messages are sent to the same core so it uses inlining rather than the pupping framework.

All-in-all I'd say this works well.

nilsdeppe (dismissed):

knelli2:
@nilsdeppe The changes from #4238 are unnecessary now and are taken care of by the patches for inlining so I undid those changes.

nilsdeppe (approved):

PR #4243, opened 2022-09-22
Multi-linear tabulated interpolation

ermost:
## Proposed changes

Adds multi-linear tabulated interpolation. The primary use case is the tabulated equation of state class, but the code should be general enough to be useful in other contexts as well.

### DIFF
--- a/src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
+++ b/src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
@@ -1,0 +1,391 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <algorithm>
+#include <array>
+#include <cmath>
+#include <memory>
+
+#include "DataStructures/Index.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Requires.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace intrp {
+
+/*!
+ * \brief Performs linear interpolation in arbitrary dimensions.
+ *  The class is non-owning and expects a C-ordered array, (n, x, y, z).
+ *  The variable index, n, varies fastest in memory.
+ *  Note that this class is intentionally non-pupable.
+ *
+ *  \tparam Dimension dimensionality of the table
+ *  \tparam NumberOfVariables number of variables stored in the table
+ *  \tparam UniformSpacing indicated whether the table has uniform _spacing or
+ * not. This is useful for performance reasons in finding the appropriate table
+ * index.
+ */
+
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+class MultiLinearSpanInterpolation {
+
+ public:
+  template <size_t ThisDimension>
+  struct Weight {
+    std::array<double, two_to_the(ThisDimension)> weights;
+    Index<two_to_the(ThisDimension)> index;
+  };
+
+  size_t extents(const size_t which_dimension) const {
+    return number_of_points_[which_dimension];
+  }
+
+  void extrapolate_above_data(const size_t which_dimension, const bool value) {
+    allow_extrapolation_abov_data_[which_dimension] = value;
+  }
+
+  void extrapolate_below_data(const size_t which_dimension, const bool value) {
+    allow_extrapolation_below_data_[which_dimension] = value;
+  }
+
+  /// Compute interpolation weights for 1D tables
+  Weight<1> get_weights(const double x1) const;
+
+  /// Compute interpolation weights for 2D tables
+  Weight<2> get_weights(const double x1, const double x2) const;
+
+  /// Compute interpolation weights for 3D tables
+  Weight<3> get_weights(const double x1, const double x2,
+                        const double x3) const;
+
+  double interpolate(const Weight<Dimension>& weights,
+                     const size_t which_variable = 0) const {
+    double result = 0;
+
+    for (size_t nn = 0; nn < weights.weights.size(); ++nn) {
+      result += weights.weights[nn] *
+                y_[which_variable + NumberOfVariables * weights.index[nn]];
+    }
+
+    return result;
+  }
+
+  template <size_t... variables_to_interpolate>
+  std::array<double, sizeof...(variables_to_interpolate)> interpolate(
+      const Weight<Dimension>& weights) const {
+    static_assert(sizeof...(variables_to_interpolate) <= NumberOfVariables,
+                  "You are trying to interpolate more variables than this "
+                  "container holds.");
+
+    return std::array<double, sizeof...(variables_to_interpolate)>{
+        interpolate(weights, variables_to_interpolate)...};
+  }
+
+  template <size_t NumberOfVariablesToInterpolate, typename... T,
+            Requires<(std::is_floating_point_v<typename std::remove_cv_t<T>> and
+                      ...)> = nullptr>
+  std::array<double, NumberOfVariablesToInterpolate> interpolate(
+      std::array<size_t, NumberOfVariablesToInterpolate>&
+          variables_to_interpolate,
+      const T&... target_points) const {
+    static_assert(
+        sizeof...(T) == Dimension,
+        "You need to provide the correct number of interpolation points");
+    static_assert(NumberOfVariablesToInterpolate <= NumberOfVariables,
+                  "You are trying to interpolate more variables than this "
+                  "container holds.");
+
+    auto weights = get_weights(target_points...);
+
+    std::array<double, NumberOfVariablesToInterpolate> interpolatedValues;
+
+    for (size_t nn = 0; nn < NumberOfVariablesToInterpolate; ++nn) {
+      interpolatedValues[nn] =
+          interpolate(weights, variables_to_interpolate[nn]);
+    }
+
+    return interpolatedValues;
+  }
+
+  template <size_t NumberOfVariablesToInterpolate, size_t... I>
+  std::array<double, NumberOfVariablesToInterpolate> interpolate(
+      std::array<size_t, NumberOfVariablesToInterpolate>&
+          variables_to_interpolate,
+      std::array<double, Dimension>& target_points,
+      std::index_sequence<I...>) const {
+    return interpolate(variables_to_interpolate, target_points[I]...);
+  }
+
+  template <size_t NumberOfVariablesToInterpolate>
+  std::array<double, NumberOfVariablesToInterpolate> interpolate(
+      std::array<size_t, NumberOfVariablesToInterpolate>&
+          variables_to_interpolate,
+      std::array<double, Dimension>& target_points) const {
+    return interpolate(variables_to_interpolate, target_points,
+                       std::make_index_sequence<Dimension>{});
+  }
+
+  MultiLinearSpanInterpolation(
+      std::array<gsl::span<const double>, Dimension> x_,
+      gsl::span<const double> y_, Index<Dimension> number_of_points__);
+
+  double lower_bound(const size_t which_dimension) const {
+    return x_[which_dimension][0];
+  }
+
+  double upper_bound(const size_t which_dimension) const {
+    return x_[which_dimension][number_of_points_[which_dimension] - 1];
+  }
+
+ private:
+  /// Allow extrapolation above table bounds.
+  std::array<bool, Dimension> allow_extrapolation_below_data_;
+  /// Allow extrapolation below table bounds.
+  std::array<bool, Dimension> allow_extrapolation_abov_data_;
+  /// Inverse pacing of the table. Only used for uniform grids
+  std::array<double, Dimension> inverse_spacing_;
+  /// Spacing of the table. Only used for uniform grids
+  std::array<double, Dimension> spacing_;
+  /// Number of points per dimension
+  Index<Dimension> number_of_points_;
+
+  using DataPointer = gsl::span<const double>;
+  /// X values of the table. Only used if allocated
+  std::array<DataPointer, Dimension> x_;
+  /// Y values of the table. Only used if allocated
+  DataPointer y_;
+
+  /// Determine relative index bracketing for non-uniform _spacing
+  size_t find_index_general(const size_t which_dimension,
+                            const double& target_points) const;
+
+  /// Determine relative index bracketing for non-uniform _spacing
+  size_t find_index_uniform(const size_t which_dimension,
+                            const double& target_points) const;
+
+  size_t find_index(const size_t which_dimension,
+                    const double& target_points) const {
+    if constexpr (UniformSpacing) {
+      return find_index_uniform(which_dimension, target_points);
+    } else {
+      return find_index_general(which_dimension, target_points);
+    }
+  }
+};
+
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+size_t MultiLinearSpanInterpolation<
+    Dimension, NumberOfVariables,
+    UniformSpacing>::find_index_general(const size_t which_dimension,
+                                        const double& target_points) const {
+  size_t lower_index_bound = 0;
+  size_t upper_index_bound = number_of_points_[which_dimension] - 1;
+
+  ASSERT((target_points > x_[which_dimension][lower_index_bound]) or
+             allow_extrapolation_below_data_[which_dimension],
+         "Interpolation exceeds lower table bounds");
+  ASSERT((target_points < x_[which_dimension][upper_index_bound]) or
+             allow_extrapolation_abov_data_[which_dimension],
+         "Interpolation exceeds upper table bounds");
+
+  while (upper_index_bound > 1 + lower_index_bound) {
+    size_t current_index =
+        lower_index_bound + (upper_index_bound - lower_index_bound) / 2;
+    if (target_points < x_[which_dimension][current_index]) {
+      upper_index_bound = current_index;
+    } else {
+      lower_index_bound = current_index;
+    }
+  }
+  return lower_index_bound;
+}
+
+/// Determine relative index bracketing for non-uniform _spacing
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+size_t MultiLinearSpanInterpolation<
+    Dimension, NumberOfVariables,
+    UniformSpacing>::find_index_uniform(const size_t which_dimension,
+                                        const double& target_points) const {
+  // Compute coordinate relative to lowest table bound
+  const auto relative_coordinate = (target_points - x_[which_dimension][0]);
+
+  auto current_index = static_cast<size_t>(relative_coordinate *
+                                           inverse_spacing_[which_dimension]);
+
+  // We are exceeding the table bounds:
+  // Use linear extrapolation based of the lowest
+  // two points in the table
+  ASSERT(allow_extrapolation_below_data_[which_dimension] or
+             UNLIKELY(relative_coordinate > 0.),
+         "Interpolation exceeds lower table bounds.");
+
+  // We are exceeding the table bounds:
+  // Use linear extrapolation based of the highest
+  // two points in the table
+
+  ASSERT(allow_extrapolation_abov_data_[which_dimension] or
+             UNLIKELY(current_index + 1 < number_of_points_[which_dimension]),
+         "Interpolation exceeds upper table bounds.");
+
+  // Enforce index ranges
+  current_index = std::min(number_of_points_[which_dimension] - 2,
+                           std::max(0ul, current_index));
+
+  return current_index;
+}
+
+/// Compute interpolation weights for 1D tables
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+auto MultiLinearSpanInterpolation<Dimension, NumberOfVariables,
+                                  UniformSpacing>::get_weights(const double x1)
+    const -> Weight<1> {
+  // Relative normalized coordinate in x-direction
+  double xx;
+
+  Weight<1> weights;
+
+  auto index = Index<1>(find_index(0, x1));
+
+  if constexpr (UniformSpacing) {
+    xx = (x1 - x_[0][index[0]]) * inverse_spacing_[0];
+  } else {
+    xx = (x1 - x_[0][index[0]]) / (x_[0][index[0] + 1] - x_[0][index[0]]);
+  }
+
+  // Note: first index varies fastest
+  weights.weights = std::array<double, 2>({
+      (1. - xx),  // 0
+      xx,         // 1
+  });
+
+  // Compute indices
+
+  weights.index[0] = index[0];
+  weights.index[1] = index[0] + 1;
+
+  return weights;
+}
+
+/// Compute interpolation weights for 2D tables
+
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+auto MultiLinearSpanInterpolation<Dimension, NumberOfVariables,
+                                  UniformSpacing>::get_weights(const double x1,
+                                                               const double x2)
+    const -> Weight<2> {
+  Weight<2> weights;
+
+  auto index = Index<2>(find_index(0, x1), find_index(1, x2));
+
+  // Relative normalized coordinate in x,y-direction
+  double xx, yy;
+  if constexpr (UniformSpacing) {
+    xx = (x1 - x_[0][index[0]]) * inverse_spacing_[0];
+    yy = (x2 - x_[1][index[1]]) * inverse_spacing_[1];
+  } else {
+    xx = (x1 - x_[0][index[0]]) / (x_[0][index[0] + 1] - x_[0][index[0]]);
+    yy = (x2 - x_[1][index[1]]) / (x_[1][index[1] + 1] - x_[1][index[1]]);
+  }
+
+  // Note: first index varies fastest
+  weights.weights = std::array<double, 4>({
+      (1. - xx) * (1. - yy),  // 00
+      xx * (1. - yy),         // 10
+      (1. - xx) * yy,         // 01
+      xx * yy,                // 11
+  });
+
+  // Compute indices
+  //
+  for (size_t j = 0; j < 2; ++j) {
+    for (size_t i = 0; i < 2; ++i) {
+      auto tmp_index = Index<2>(index[0] + i, index[1] + j);
+      weights.index[i + 2 * j] = collapsed_index(tmp_index, number_of_points_);
+    }
+  }
+
+  return weights;
+}
+
+/// Compute interpolation weights for 3D tables
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+auto MultiLinearSpanInterpolation<Dimension, NumberOfVariables,
+                                  UniformSpacing>::get_weights(const double x1,
+                                                               const double x2,
+                                                               const double x3)
+    const -> Weight<3> {
+  // Relative normalized coordinate in x,y,z-direction
+  double xx, yy, zz;
+
+  Weight<3> weights;
+
+  auto index =
+      Index<3>(find_index(0, x1), find_index(1, x2), find_index(2, x3));
+
+  if constexpr (UniformSpacing) {
+    xx = (x1 - x_[0][index[0]]) * inverse_spacing_[0];
+    yy = (x2 - x_[1][index[1]]) * inverse_spacing_[1];
+    zz = (x3 - x_[2][index[2]]) * inverse_spacing_[2];
+  } else {
+    xx = (x1 - x_[0][index[0]]) / (x_[0][index[0] + 1] - x_[0][index[0]]);
+    yy = (x2 - x_[1][index[1]]) / (x_[1][index[1] + 1] - x_[1][index[1]]);
+    zz = (x3 - x_[2][index[2]]) / (x_[2][index[2] + 1] - x_[2][index[2]]);
+  }
+
+  // Note: first index varies fastest
+  weights.weights = std::array<double, 8>({
+      (1. - xx) * (1. - yy) * (1. - zz),  // 000
+      xx * (1. - yy) * (1. - zz),         // 100
+      (1. - xx) * yy * (1. - zz),         // 010
+      xx * yy * (1. - zz),                // 110
+      (1. - xx) * (1. - yy) * zz,         // 001
+      xx * (1. - yy) * zz,                // 101
+      (1. - xx) * yy * zz,                // 011
+      xx * yy * zz,                       // 111
+  });
+
+  // Compute indices
+  //
+  for (size_t k = 0; k < 2; ++k) {
+    for (size_t j = 0; j < 2; ++j) {
+      for (size_t i = 0; i < 2; ++i) {
+        auto tmp_index = Index<3>(index[0] + i, index[1] + j, index[2] + k);
+        weights.index[i + 2 * (j + 2 * k)] =
+            collapsed_index(tmp_index, number_of_points_);
+      }
+    }
+  }
+
+  return weights;
+}
+
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+MultiLinearSpanInterpolation<Dimension, NumberOfVariables, UniformSpacing>::
+    MultiLinearSpanInterpolation(
+        std::array<gsl::span<double const>, Dimension> x,
+        gsl::span<double const> y, Index<Dimension> number_of_points)
+    : number_of_points_(number_of_points), x_(x), y_(y) {
+  for (size_t i = 0; i < Dimension; ++i) {
+    spacing_[i] = x_[i][1] - x_[i][0];
+    inverse_spacing_[i] = 1. / spacing_[i];
+    allow_extrapolation_below_data_[i] = false;
+    allow_extrapolation_abov_data_[i] = false;
+  }
+}
+
+/// Multilinear span interpolation with uniform grid spacing
+template <size_t Dimension, size_t NumberOfVariables>
+using UniformMultiLinearSpanInterpolation =
+    MultiLinearSpanInterpolation<Dimension, NumberOfVariables, true>;
+
+/// Multilinear span interpolation with non-uniform grid spacing
+template <size_t Dimension, size_t NumberOfVariables>
+using GeneralMultiLinearSpanInterpolation =
+    MultiLinearSpanInterpolation<Dimension, NumberOfVariables, false>;
+
+}  // namespace intrp
--- a/tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
+++ b/tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
@@ -1,0 +1,148 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace {
+template <size_t Dim, size_t NumVar>
+void test() {
+  CAPTURE(Dim);
+  CAPTURE(NumVar);
+
+  MAKE_GENERATOR(gen);
+
+  std::uniform_real_distribution<> dist_func(-1., 1.);
+  std::uniform_int_distribution<size_t> dist_int(3, 5);
+
+  // Main idea: The table performs a multi-linear interpolation
+  // Each element in the table will be multi-linear function,
+  // so we can use this directly as a test.
+  // For simplicity, we will use the table to globally reproduce a
+  // single multi-linear function.
+
+  std::array<std::array<double, Dim>, NumVar> matrix;
+  std::array<double, NumVar> vector_b;
+
+  for (size_t i = 0; i < NumVar; ++i) {
+    vector_b[i] = dist_func(gen);
+    for (size_t j = 0; j < Dim; ++j) {
+      matrix[i][j] = dist_func(gen);
+    }
+  }
+
+  auto f = [&matrix, &vector_b](const std::array<double, Dim>& x) {
+    std::array<double, NumVar> res;
+
+    for (size_t i = 0; i < NumVar; ++i) {
+      res[i] = vector_b[i];
+      for (size_t j = 0; j < Dim; ++j) {
+        res[i] += matrix[i][j] * x[j];
+      }
+    }
+    return res;
+  };
+
+  // Create data data structures
+  std::array<std::unique_ptr<double[]>, Dim> X_data;
+  Index<Dim> num_x_points;
+
+  // Initialize discrete grid for the tabulated interpolation routine
+
+  size_t total_num_points = 1;
+  for (size_t n = 0; n < Dim; ++n) {
+    num_x_points[n] = dist_int(gen);
+    total_num_points *= num_x_points[n];
+
+    X_data[n] = std::move(std::make_unique<double[]>(num_x_points[n]));
+
+    for (size_t m = 0; m < num_x_points[n]; m++) {
+      X_data[n][m] = -1. + m * 2. / static_cast<double>(num_x_points[n] - 1);
+    }
+  }
+
+  // Correct for NumVars
+  total_num_points *= NumVar;
+
+  // Allocate storage for main table
+  auto dependent_variables = std::make_unique<double[]>(total_num_points);
+
+  std::array<gsl::span<const double>, Dim> independent_data_view;
+
+  // Fill dependent_variables
+
+  for (size_t ijk = 0; ijk < total_num_points / NumVar; ++ijk) {
+    Index<Dim> index;
+    auto tmp = std::array<double, Dim>{};
+
+    // Uncompress index
+    size_t myind= ijk;
+    for (size_t nn = 0; nn < Dim - 1; ++nn) {
+      index[nn] = myind % num_x_points[nn];
+      myind = (myind - index[nn])/num_x_points[nn];
+      tmp[nn] = X_data[nn][index[nn]];
+    }
+    index[Dim - 1] = myind;
+    tmp[Dim - 1] = X_data[Dim - 1][index[Dim - 1]];
+
+    for (size_t nv = 0; nv < NumVar; ++nv) {
+      std::array<double, NumVar> Fx = f(tmp);
+      dependent_variables[nv + NumVar * ijk] = Fx[nv];
+    }
+  }
+
+  std::array<double, Dim> x;
+  std::array<size_t, NumVar> which_var;
+
+  for (size_t which_Dim = 0; which_Dim < Dim; ++which_Dim) {
+    independent_data_view[which_Dim] = gsl::span<const double>{
+        X_data[which_Dim].get(), num_x_points[which_Dim]};
+    x[which_Dim] = dist_func(gen);
+  }
+
+  for (size_t nv = 0; nv < NumVar; ++nv) {
+    which_var[nv] = nv;
+  }
+
+  intrp::UniformMultiLinearSpanInterpolation<Dim, NumVar> uniform_intp(
+      independent_data_view, {dependent_variables.get(), total_num_points},
+      num_x_points);
+  intrp::GeneralMultiLinearSpanInterpolation<Dim, NumVar> general_intp(
+      independent_data_view, {dependent_variables.get(), total_num_points},
+      num_x_points);
+
+  for (size_t d = 0; d < Dim; ++d) {
+    uniform_intp.extrapolate_above_data(d, true);
+    general_intp.extrapolate_above_data(d, true);
+
+    uniform_intp.extrapolate_below_data(d, true);
+    general_intp.extrapolate_below_data(d, true);
+  }
+
+  auto y_expected = f(x);
+
+
+  auto y_interpolated = uniform_intp.interpolate(which_var, x);
+  auto y_interpolated_gen = general_intp.interpolate(which_var, x);
+
+  for (size_t nv = 0; nv < NumVar; ++nv) {
+    CHECK(std::abs(y_expected[nv] - y_interpolated[nv]) <
+          1.e-12 * std::abs(y_expected[nv]));
+    CHECK(std::abs(y_expected[nv] - y_interpolated_gen[nv]) <
+          1.e-12 * std::abs(y_expected[nv]));
+  }
+}
+}  // namespace
+
+SPECTRE_TEST_CASE("Unit.Numerical.Interpolation.MultiLinearSpanInterpolation",
+                  "[Unit][NumericalAlgorithms]") {
+    test<1,1>();
+    test<2,3>();
+    test<3,5>();
+}
--- a/src/NumericalAlgorithms/Interpolation/CMakeLists.txt
+++ b/src/NumericalAlgorithms/Interpolation/CMakeLists.txt
@@ -35,6 +35,7 @@ spectre_target_headers(
   LagrangePolynomial.hpp
   LinearLeastSquares.hpp
   LinearRegression.hpp
+  MultiLinearSpanInterpolation.hpp
   LinearSpanInterpolator.hpp
   PolynomialInterpolation.hpp
   PredictedZeroCrossing.hpp
--- a/tests/Unit/NumericalAlgorithms/Interpolation/CMakeLists.txt
+++ b/tests/Unit/NumericalAlgorithms/Interpolation/CMakeLists.txt
@@ -10,6 +10,7 @@ set(LIBRARY_SOURCES
   Test_CubicSpline.cpp
   Test_IrregularInterpolant.cpp
   Test_LagrangePolynomial.cpp
+  Test_MultiLinearSpanInterpolation.cpp
   Test_LinearLeastSquares.cpp
   Test_LinearRegression.cpp
   Test_PolynomialInterpolation.cpp


ermost:
Switched to 'Requires', thanks for the suggestion!

ermost:
Rebased and squashed

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Requires.hpp"
+
+namespace intrp {
+
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+class LinearSpanInterpolationND {
```
nilsdeppe: What do you think of naming this `MultiLinearSpanInterpolation`? The `ND` keeps confusing me with my initials 😅

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+#include <cmath>
+#include <memory>
+
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Requires.hpp"
+
+namespace intrp {
+
```
nilsdeppe: Could you add doxygen documentation:
```cpp
/*!
 * \brief Performs linear interpolation in each dimension.
 * 
 * \tparam Dimension the dimensionality
 * \tparam The number of variables ...
```

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+class LinearSpanInterpolationND {
+ public:
```
nilsdeppe: We prefer
```cpp
class Blah {
 public:
  /* all public functions and members */

 private:
  /* all private members */
};
```
unless there's a techincal reason why the ordering can't be that way.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+        return linearInterpolation1D_Uniform(
+            A, B, xA, xB, data.inverseSpacing[whichDimension], x1);
+      } else {
+        return linearInterpolation1D_General(A, B, xA, xB, x1);
+      }
+    }
+  };
+
+ public:
+  inline int getSize(size_t const whichDimension) {
```
nilsdeppe: no need to mark anything inside a class definition as `inline`, it is implicitly.

we use "west const" (`const size_t`)

we use `size_t` for sizes (for better or worse...)

mark member function as `const` when possible

we typically call this `extents(const size_t dimension)`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+  };
+
+  template <size_t N, typename Indices = std::make_index_sequence<Dimension>>
+  inline std::array<double, N> interpolate(
+      std::array<size_t, N> whichVariables,
+      std::array<double, Dimension>& xin) const {
+    return interpolate_helper(whichVariables, xin, Indices{});
+  };
+
+  LinearSpanInterpolationND() = delete;
```
nilsdeppe: All classes need to be serializable, and Charm++ requires everything to have a default constructor.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+
```
nilsdeppe: add a `pup(PUP::er& p)` function to serialize data members. I think all you need in the definition is
```cpp
  p | member_a_;
  p | member_b_;
  ...
```

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+  LinearSpanInterpolationND(gsl::span<double const> __y,
```
nilsdeppe: `__NAME` is reserved by the C++ standard. This is undefined behavior. We post-fix private member variables with `_`, e.g. `y_`. From the public interface it looks like you can make all member variables private, then have them named `y_`, `x_`, and `number_of_points_`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+                            std::array<gsl::span<double const>, Dimension> __x,
+                            std::array<size_t, Dimension> __numberOfPoints)
```
nilsdeppe: we typically use the `Index<Dim>` class for representing number of points/extents. This has a `product()` member function that calculates the total number of points, allows slicing, etc.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Requires.hpp"
+
+namespace intrp {
+
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+class LinearSpanInterpolationND {
+ public:
+  using DataPointer = gsl::span<double const>;
+  //! X values of the table. Only used if allocated
```
nilsdeppe: We use `///` and `/*!` style doxygen comments

I think these members should all be `private`, then rename to snake_case, and use trailing `_`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+      }
+    }
+  };
+
+ public:
+  inline int getSize(size_t const whichDimension) {
+    return numberOfPoints[whichDimension];
+  }
+
+  inline void extrapolateHigh(size_t whichDimension, bool value) {
```
nilsdeppe: `void extrapolate_high(const size_t dimension, const bool value)` (I'm not sure what `extrapolate_high/low` means. Is this `allow_extrapolation_above_data`? If so, maybe rename to something more explicit?

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+    extrapolate_high[whichDimension] = value;
+  }
+
+  inline void extrapolateLow(size_t whichDimension, bool value) {
+    extrapolate_low[whichDimension] = value;
+  }
+
+  inline double const& xmin(size_t whichDimension) {
```
nilsdeppe: west-const, `const size_t` (apply throughout, please)

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+    return x[whichDimension][0];
+  };
+
+  inline double const& xmax(size_t whichDimension) {
+    return x[whichDimension][numberOfPoints[whichDimension] - 1];
+  };
+
+  template <
+      size_t... whichVariables, typename... Xt,
+      Requires<(std::is_convertible_v<typename std::remove_cv_t<Xt>, double> &&
```
nilsdeppe: `&&` -> `and` (we prefer the C++ variants since they are easier to read)

do you want convertible or do you want to be exact `double`? That is, do we want to worry about people passing in some class that is convertible to a double? The only use-case I can think of is `float` being passed in, so maybe `std::is_floating_point_v<std::remove_cv_t<X>> and ...`?

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+  inline double const& xmin(size_t whichDimension) {
+    return x[whichDimension][0];
+  };
+
+  inline double const& xmax(size_t whichDimension) {
+    return x[whichDimension][numberOfPoints[whichDimension] - 1];
+  };
+
+  template <
+      size_t... whichVariables, typename... Xt,
```
nilsdeppe: `whichVariables` -> `WhichVariables` or maybe `VariablesToInterpolate`?

`Xt` -> `X` (or `T`, which is very standard)

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+      Requires<(std::is_convertible_v<typename std::remove_cv_t<Xt>, double> &&
+                ...)> = nullptr>
+  inline std::array<double, sizeof...(whichVariables)> interpolate(
```
nilsdeppe: function templates are always effectively `inline` since they produce a weak reference in terms of linking purposes

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+    for (int i = 0; i < N; ++i)
+      interpolatedValues[i] = RecursiveInterpolationKernelGeneral<
+          decltype(*this), Dimension - 1, CompressedIndexRecursionInitial_t,
+          Xt...>::interpolate(*this, indexArray, whichVariables[i], xin...);
+
+    return interpolatedValues;
+  };
+
+  template <size_t N, size_t... I>
+  inline std::array<double, N> interpolate_helper(
```
nilsdeppe: move into `private` block?

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+      std::array<size_t, N> whichVariables, std::array<double, Dimension>& xin,
+      std::index_sequence<I...>) const {
+    return interpolate(whichVariables, xin[I]...);
+  };
+
+  template <size_t N, typename Indices = std::make_index_sequence<Dimension>>
```
nilsdeppe: remove `Indices` and move it down to `interpolator_helper(which_variables, x, std::make_index_sequence<Dimension>{})`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+  inline std::array<double, N> interpolate(
+      std::array<size_t, N> whichVariables,
```
nilsdeppe: Both arrays can be `const std::array<?, ?>& blah` (fill in template parameters, not some weird C++ magic :P )

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+            CompressedIndexRecursionInitial_t, Xt...>::interpolate(*this,
+                                                                   indexArray,
+                                                                   xin...))...};
+  };
+
+  template <
+      size_t N, typename... Xt,
+      Requires<(std::is_convertible_v<typename std::remove_cv_t<Xt>, double> &&
+                ...)> = nullptr>
+  inline std::array<double, N> interpolate(std::array<size_t, N> whichVariables,
```
nilsdeppe: `const std::array<size_t, N>& which_variables`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+                                           Xt const&... xin) const {
```
nilsdeppe: `xin` -> `target_points` (same elsewhere)

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+    return std::array<double, sizeof...(whichVariables)>{
+        (RecursiveInterpolationKernel<
+            decltype(*this), Dimension - 1, whichVariables,
+            CompressedIndexRecursionInitial_t, Xt...>::interpolate(*this,
+                                                                   indexArray,
+                                                                   xin...))...};
+  };
+
+  template <
+      size_t N, typename... Xt,
```
nilsdeppe: Maybe rename `N` -> `NumberOfVarsToInterpolate`?

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+    while (upperIndexBound - lowerIndexBound > 1) {
+      int currentIndex =
+          lowerIndexBound + (upperIndexBound - lowerIndexBound) / 2;
+      if (xin < x[whichDimension][currentIndex])
+        upperIndexBound = currentIndex;
+      else
+        lowerIndexBound = currentIndex;
+    }
+    return lowerIndexBound;
+  };
```
nilsdeppe: No `;`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+    int upperIndexBound = numberOfPoints[whichDimension] - 1;
+
+    ASSERT((xin > x[whichDimension][lowerIndexBound]) or
+               extrapolate_low[whichDimension],
+           "Interpolation exceeds lower table bounds");
+    ASSERT((xin < x[whichDimension][upperIndexBound]) or
+               extrapolate_high[whichDimension],
+           "Interpolation exceeds upper table bounds");
+
+    while (upperIndexBound - lowerIndexBound > 1) {
```
nilsdeppe: Isn't all this just:
```cpp
return std::distance(x[which_dimension].begin(),
  std::lower_bound(x[which_dimension].begin(), x[which_dimension].end(), x_in));
```

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+    return lowerIndexBound;
+  };
+
+  //! Determine relative index bracketing for non-uniform spacing
+  inline size_t findIndexUniform(size_t const whichDimension,
+                                 double const& xin) const {
+    // Compute coordinate relative to lowest table bound
+    const auto relativeCoordinate = (xin - x[whichDimension][0]);
+
+    if (relativeCoordinate == 0) {
```
nilsdeppe: `0` -> `0.` (on next line too) and do `if (UNLIKELY(...)) { ... } else { ...}`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+      return 0;
+    }
+
+    // We are exceeding the table bounds:
+    // Use linear extrapolation based of the lowest
+    // two points in the table
+    if (relativeCoordinate < 0) {
+      ASSERT(extrapolate_low[whichDimension],
```
nilsdeppe: `ASSERT(relative_coordinate >= 0. or extrapolate_low[whichDimension], ...);`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+
+    auto currentIndex =
+        static_cast<int>(relativeCoordinate * inverseSpacing[whichDimension]);
+
+    // We are exceeding the table bounds:
+    // Use linear extrapolation based of the highest
+    // two points in the table
+
+    if (currentIndex > numberOfPoints[whichDimension] - 2) {
+      ASSERT(extrapolate_high[whichDimension],
```
nilsdeppe: `ASSERT(relative_coordinate <= x[which_dimension-2], ...);` (maybe I got the index offset wrong here), then move up to below other assert on lower bound

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+             "Interpolation exceeds upper table bounds.");
+
+      return numberOfPoints[whichDimension] - 2;
+    }
+
+    return currentIndex;
```
nilsdeppe: `return std::min(static_cast<size_t>(relativeCoordinate * inverseSpacing[whichDimension]), numberOfPoints[whichDimension] - 2);`

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+                          double const& xin) const {
+    if constexpr (UniformSpacing) {
+      return findIndexUniform(whichDimension, xin);
+    } else {
+      return findIndexGeneral(whichDimension, xin);
+    }
+  };
+
+  //! Recursive interpolaton requires computing compressed indices.
+  //  We assume the table layout to be in Fortran ordering.
```
nilsdeppe: `Fortran ordering (last index varies fastest)`

This should also be on the main documentation.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
```diff
@@ -0,0 +1,374 @@
...
+  };
+
+  // Unpack in dimension!
+  template <typename Data_t, size_t whichDimension, typename CompressedIndex_t,
+            typename X1, typename... Xt>
+  struct RecursiveInterpolationKernelGeneral {
+    static inline double const interpolate(
+        Data_t const& data, std::array<size_t, Dimension> const& indexArray,
+        size_t which_var, Xt const&... xin, X1 const& x1) {
+      auto const xA = data.x[whichDimension][indexArray[whichDimension]];
```
nilsdeppe: We've found that the recursion is (surprisingly) slow and instead opted for different definitions depending on the `Dimension` in other contexts (see, e.g. `collapsed_index` in src/DataStructures/Index.hpp). This is also what the LANL interpolation library does (I forget what it's called now)

nilsdeppe (changes requested):
I'll look at the tests once the interface is fixed.

Since all of our data other than EOS tables is in first-dim varies fastest, I'm wondering if it would be safer for us to just reorder the EOS tables and have everything always be first-dim fastest (C-ordering). I'm worried someone will use this class on non-EOS data and then get incorrect answers. What do you think?

Edit: another advantage of first-dim fastest is that we can reuse a lot of functions for computing indices

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: I didn't even think of that meaning of ND :)

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: Sure!

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: OK

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: Ah ok! I wasn't aware of this. fixed it now.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: Thanks for the background! I changed things accordingly.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: Ah ok, I added that back in.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: Neat, I switched to using the Index class in all parts of the code.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: OK!

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: I wasn't aware of is_floating_point_v, changed it now.

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: Changed it

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: OK!

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: I got rid of all of the recursion, and specialized to 1,2 and 3d for now.
As discussed, the performance turned out to be identical for use cases that are memory bound and have very large tables. for smaller problems that remain in cache, the specialized version should be much better. so we stick with that. (It's also more readable :) )

ermost:
I think I addressed all of the above and the unit tests are working to (locally).

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: OK

src/NumericalAlgorithms/Interpolation/LinearInterpolationND.hpp
ermost: I tried this, but it wasn't stable for edge cases. Let's maybe stick with the C style version.

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+
+namespace PUP {
+class er;
+}  // namespace PUP
+
+namespace intrp {
+
+/*!
+ * \brief Performs linear interpolation in arbitrary dimensions.
+ * The class is non-owning and expects a C-ordered array
```
nilsdeppe: Maybe add in parens `(in x,y,z this means x varies fastest in memory)`?

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+ *
+ *  \tparam Dimension dimensionality of the table
+ *  \tparam NumberOfVariables number of variables stored in the table
+ *  \tparam UniformSpacing indicated whether the table has uniform spacing or
+ * not. This is useful for performance reasons in finding the appropriate table
+ * index. Note that the variable index is contiguous in memory.
+ *
```
nilsdeppe: remove blank lines, and 36 too?

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+
+  using DataPointer = gsl::span<double const>;
+  /// X values of the table. Only used if allocated
+  std::array<DataPointer, Dimension> x;
+  /// Y values of the table. Only used if allocated
+  DataPointer y;
+
+ public:
+  template <size_t thisDimension>
+  struct weight_t {
```
nilsdeppe: `weight_t` -> `Weight` (we use Pascal/CamelCase for classes)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+    std::array<
+        double,
+        brigand::power<std::integral_constant<size_t, 2>,
```
nilsdeppe: We also have `two_to_the` in `Utilities/ConstantExpressions.hpp`. Then this would be `std::array<double, two_to_the(thisDimension)>`

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+  Index<Dimension> numberOfPoints;
+
+  using DataPointer = gsl::span<double const>;
+  /// X values of the table. Only used if allocated
+  std::array<DataPointer, Dimension> x;
+  /// Y values of the table. Only used if allocated
+  DataPointer y;
+
+ public:
+  template <size_t thisDimension>
```
nilsdeppe: `ThisDimension` (template parameters are Pascal/CamelCase)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+        double,
+        brigand::power<std::integral_constant<size_t, 2>,
+                       std::integral_constant<size_t, thisDimension>>::value>
+        weights;
+    Index<brigand::power<std::integral_constant<size_t, 2>,
+                         std::integral_constant<size_t, thisDimension>>::value>
+        index;
+  };
+
+  size_t extents(size_t const whichDimension) const {
```
nilsdeppe: apply throughout: `const size_t which_dimension` (we use west-const and snake_case for variables)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+    return numberOfPoints[whichDimension];
+  }
+
+  void extrapolateAboveData(size_t whichDimension, bool value) {
+    allow_extrapolation_above_data[whichDimension] = value;
+  }
+
+  void extrapolateBelowData(size_t whichDimension, bool value) {
```
nilsdeppe: `const bool value` (everywhere it's needed)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+  void pup(PUP::er& p) {
+    //    p | x;
+    //    p | y;
+    p | numberOfPoints;
+    p | inverseSpacing;
+    p | allow_extrapolation_below_data;
+    p | allow_extrapolation_above_data;
+  }
+
+  const double& xmin(size_t whichDimension) const {
```
nilsdeppe: maybe `lower_bound` and `upper_bound`? I don't feel strongly about this. General question about what you prefer.

`const size_t...` and below

Since the size of a pointer and `double` are the same we've always just returned `double` instead of `const double&`, which aids in eliminating a pointer chase.

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+ *  \tparam UniformSpacing indicated whether the table has uniform spacing or
+ * not. This is useful for performance reasons in finding the appropriate table
+ * index. Note that the variable index is contiguous in memory.
+ *
+ *
+ */
+
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+class MultiLinearSpanInterpolation {
+ private:
```
nilsdeppe: `public` before `private` (move to end of class)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+  /// Allow extrapolation above table bounds.
+  std::array<bool, Dimension> allow_extrapolation_below_data;
```
nilsdeppe: member variable have `_` suffix

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+  /// Allow extrapolation below table bounds.
+  std::array<bool, Dimension> allow_extrapolation_above_data;
+  /// Spacing of the table. Only used for uniform grids
+  std::array<double, Dimension> inverseSpacing, spacing;
```
nilsdeppe: `snake_case` a few times for member variables

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+
+  void extrapolateBelowData(size_t whichDimension, bool value) {
+    allow_extrapolation_below_data[whichDimension] = value;
+  }
+
+  /// Determine relative index bracketing for non-uniform spacing
+  size_t findIndexGeneral(const size_t whichDimension,
+                          const double& targetPoints) const;
+
+  //! Determine relative index bracketing for non-uniform spacing
```
nilsdeppe: `///` instead of `//!` (maybe just to a find-replace in the file just in case :) )

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+    size_t currentIndex =
+        lowerIndexBound + (upperIndexBound - lowerIndexBound) / 2;
+    if (targetPoints < x[whichDimension][currentIndex])
+      upperIndexBound = currentIndex;
+    else
+      lowerIndexBound = currentIndex;
+  }
+  return lowerIndexBound;
+
+  // if (targetPoints < x[whichDimension][0])
```
nilsdeppe: remove commented out code?

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+        auto tmp_index = Index<3>(index[0] + i, index[1] + j, index[2] + k);
+        weights.index[i + 2 * (j + 2 * k)] =
+            collapsed_index(tmp_index, numberOfPoints);
+      }
+
+  return weights;
+}
+
+template <size_t Dimension, size_t NumberOfVariables>
+using UniformLinearSpanInterpolation_t =
```
nilsdeppe: no trailing `_t` and below. We use that only for `::type` in a metafunction (i.e. eager metafunctions)

Document these two type aliases? Doesn't need to be long, just enough so they show up in the dox :)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+  // Enforce index ranges
+  currentIndex =
+      std::min(numberOfPoints[whichDimension] - 2, std::max(0ul, currentIndex));
+
+  return currentIndex;
+}
+
+/// Compute interpolation weights for 1D tables
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+typename MultiLinearSpanInterpolation<Dimension, NumberOfVariables,
```
nilsdeppe: I think if you use trailing return type syntax you can just do `Weight<1>`. I.e. `auto Mult...::get_weights() const -> Weight<1>` (I always forget if the `const` goes before or after the return type, sorry...)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+    Index<brigand::power<std::integral_constant<size_t, 2>,
+                         std::integral_constant<size_t, thisDimension>>::value>
+        index;
+  };
+
+  size_t extents(size_t const whichDimension) const {
+    return numberOfPoints[whichDimension];
+  }
+
+  void extrapolateAboveData(size_t whichDimension, bool value) {
```
nilsdeppe: Should we have users set these in the constructor rather than later? I'm not sure of the EOS use-case, but I personally feel more comfortable if people have to choose upon construction whether or not they want extrapolation. What do you think? Maybe a `const std::array<bool, Dimension>& extrapolate_in_dimension`?

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+    allow_extrapolation_above_data[whichDimension] = value;
+  }
+
+  void extrapolateBelowData(size_t whichDimension, bool value) {
+    allow_extrapolation_below_data[whichDimension] = value;
+  }
+
+  /// Determine relative index bracketing for non-uniform spacing
+  size_t findIndexGeneral(const size_t whichDimension,
```
nilsdeppe: Are these `find` functions intended to be part of the general interface? I assumed people would only use `get_weights` and `interpolate`

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+                  "container holds.");
+
+    return std::array<double, sizeof...(VariablesToInterpolate)>{
+        interpolate(weights, VariablesToInterpolate)...};
+  }
+
+  template <size_t NumberOfVariablesToInterpolate, typename... T,
+            Requires<(std::is_floating_point_v<typename std::remove_cv_t<T>> and
+                      ...)> = nullptr>
+  std::array<double, NumberOfVariablesToInterpolate> interpolate(
```
nilsdeppe, line 92: Question: what do you think of always forcing people to use the function that takes `get_weights()`? That way people are more likely to reuse weights :) So basically have calls be `interpolator.intperpolate<Vars...>(interpolator.get_weights(x0, x1, x2))`? It's more typing in some cases, but will hopefully encourage reuse...

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+        "You need to provide the correct number of interpolation points");
+    static_assert(NumberOfVariablesToInterpolate <= NumberOfVariables,
+                  "You are trying to interpolate more variables than this "
+                  "container holds.");
+
+    auto weights = getWeights(targetPoints...);
+
+    std::array<double, NumberOfVariablesToInterpolate> interpolatedValues;
+
+    for (size_t nn = 0; nn < NumberOfVariablesToInterpolate; ++nn)
```
nilsdeppe: we prefer braces around all loops, even single line ones. We've had bugs in SpEC where people indented "into" the for loop or if statement. Even if the code is technically what was meant, it's just harder for that to be clear

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+      interpolatedValues[nn] = interpolate(weights, VariablesToInterpolate[nn]);
+
+    return interpolatedValues;
+  }
+
+  template <size_t NumberOfVariablesToInterpolate, size_t... I>
+  std::array<double, NumberOfVariablesToInterpolate> interpolate(
+      Index<NumberOfVariablesToInterpolate>& VariablesToInterpolate,
```
nilsdeppe: So we generally only use `Index` for low-dimensional meshes (0-4). Here a `std::array<size_t, NumberOfVariablesToInterpolate>` seems more appropriate, but let me know if I'm missing something :)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+
+  template <size_t NumberOfVariablesToInterpolate>
+  std::array<double, NumberOfVariablesToInterpolate> interpolate(
+      Index<NumberOfVariablesToInterpolate>& VariablesToInterpolate,
+      std::array<double, Dimension>& targetPoints) const {
+    return interpolate(VariablesToInterpolate, targetPoints,
+                       std::make_index_sequence<Dimension>{});
+  }
+
+  MultiLinearSpanInterpolation(
```
nilsdeppe, line 133: can you define the constructor out-of-line? They tend to be rather bulky and the general recommendation is to always define them out of line.

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+    for (size_t i = 0; i < Dimension; ++i) {
+      spacing[i] = x[i][1] - x[i][0];
+      inverseSpacing[i] = 1. / spacing[i];
+      allow_extrapolation_below_data[i] = false;
+      allow_extrapolation_above_data[i] = false;
+    };
+  };
+
+  void pup(PUP::er& p) {
+    //    p | x;
```
nilsdeppe: So my guess is that serialization doesn't work correctly. Since this class gives a view on some data, my take on it is that it _shouldn't_ be serialized. I would remove the `pup` (since it's I think broken anyway) and then just document in the class Doxygen that this is an intentional design choice and that the class should be used locally in functions but not stored long-term in contexts where it would need serialization.

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+}
+
+/// Compute interpolation weights for 1D tables
+template <size_t Dimension, size_t NumberOfVariables, bool UniformSpacing>
+typename MultiLinearSpanInterpolation<Dimension, NumberOfVariables,
+                             UniformSpacing>::template weight_t<1>
+MultiLinearSpanInterpolation<Dimension, NumberOfVariables,
+                             UniformSpacing>::getWeights(const double x1)
+    const {
+  double xx;
```
nilsdeppe, line 248: can you give this a name that would help me understanding better what it is? I'm trying to think of something, but it's not trivial :P Maybe `distance_to_point` or `fractional_distance_to_point`?

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+  weights.weights = std::array<double, 4>({
+      (1. - xx) * (1. - yy),  // 00
+      xx * (1. - yy),         // 10
+      (1. - xx) * yy,         // 01
+      xx * yy,                // 11
+  });
+
+  // Compute indices
+  //
+  for (size_t j = 0; j < 2; ++j)
```
nilsdeppe: braces for `for` loop

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,397 @@
...
+    for (size_t i = 0; i < 2; ++i) {
+      auto tmp_index = Index<2>(index[0] + i, index[1] + j);
```
nilsdeppe, line 307: why not `const Index<2> tmp_index(index[0] + i, index[1] + j);`?

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace {
+template <size_t dim, size_t num_var>
```
nilsdeppe: `Dim` and `NumVar`

I would say `CAPTURE` them :) I've always found it useful!

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+void test() {
+  //  CAPTURE(Degree);
+
+  // pseudo random number generator
+  std::random_device rd;   // seed
```
nilsdeppe: we have a `MAKE_GENERATOR` macro that makes it easy to play back seeds, etc.

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+  // For simplicity, we will use the table to globally reproduce a
+  // single multi-linear function.
+
+
+  std::array< std::array<double, dim>, num_var> A;
+  std::array< double, num_var> b;
+
+  for (size_t i = 0; i < num_var; ++i) {
+    b[i] = dist_func(gen);
+    for (size_t j = 0; j < dim; ++j)
```
nilsdeppe: braces (and wherever else)

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+      A[i][j] = dist_func(gen);
+  }
+
+  auto f = [&](std::array<double, dim> const& x) {
```
nilsdeppe: we prefer to only capture what is used (avoids struct bloat and prevents bugs)

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+  std::uniform_int_distribution<size_t> dist_int(10, 20);
+
+  // Main idea: The table performs a multi-linear interpolation
+  // Each element in the table will be multi-linear function,
+  // so we can use this directly as a test.
+  // For simplicity, we will use the table to globally reproduce a
+  // single multi-linear function.
+
+
+  std::array< std::array<double, dim>, num_var> A;
```
nilsdeppe: we prefer longer descriptive names for variables. E.g. `table` or even `matrix_a` and `vector_b`. This at the very least just makes find-replace and searching easier in the file :)

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+  std::array< double, num_var> b;
+
+  for (size_t i = 0; i < num_var; ++i) {
+    b[i] = dist_func(gen);
+    for (size_t j = 0; j < dim; ++j)
+      A[i][j] = dist_func(gen);
+  }
+
+  auto f = [&](std::array<double, dim> const& x) {
+    std::array<double, num_var> res{0};
```
nilsdeppe: I'm not sure we can universally rely on `{0}` giving an entire array of `0`. I believe rules of aggregate initialization say know. We have a `make_array<Dim>(0.)` function.

We also use `0.` or `0.0` for doubles (just be consistent within each file)

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+    return res;
+  };
+
+  // Create data data structures
+  std::array<std::unique_ptr<double[]>, dim> X_data;
+  Index<dim> num_x_points;
+
+  // Initialize discrete grid for the tabulated interpolation routine
+
+  size_t num_total = 1;
```
nilsdeppe: total what? Points? Variables? :D

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+  for (size_t n = 0; n < dim; ++n) {
+    num_x_points[n] = dist_int(gen);
+    num_total *= num_x_points[n];
+
+    X_data[n] = std::move(std::make_unique<double[]>(num_x_points[n]));
+
+    for (size_t m = 0; m < num_x_points[n]; m++)
```
nilsdeppe: braces

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+      X_data[n][m] = -1. + m * 2. / static_cast<double>(num_x_points[n] - 1);
+  }
+
+  // Correct for num_vars
+  num_total *= num_var;
+
+  // Allocate storage for main table
+  auto Y = std::make_unique<double[]>(num_total);
```
nilsdeppe: `dependent_variables`?

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  std::array<gsl::span<double const>, dim> Xarray;
```
nilsdeppe: `independent_data_view`?

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+    for(size_t nn=0; nn< dim-1; ++nn){
+      index[nn] = myind % num_x_points[nn];
+      myind = (myind - index[nn])/num_x_points[nn];
+      tmp[nn] = X_data[nn][index[nn]];
+    }
+    index[dim-1] = myind;
+    tmp[dim-1] = X_data[dim-1][index[dim-1]];
+
+      for (size_t nv = 0; nv < num_var; ++nv) {
+    std::array<double, num_var> Fx = f(tmp); // FIMXE
```
nilsdeppe: I guess this needs fixing?

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,146 @@
...
+
+  // Allocate storage for main table
+  auto Y = std::make_unique<double[]>(num_total);
+
+  std::array<gsl::span<double const>, dim> Xarray;
+
+
+  // We are manually unscrambeling the compressed index here
+
+  for(size_t ijk=0; ijk < num_total/num_var; ++ijk){
```
nilsdeppe: Is this basically the opposite of `collapsed_index` in `Index.hpp`? That is, given some extents and a 1d index, get the 3d index?

nilsdeppe (changes requested):
I didn't go through the tests in great detail yet, just wanted to catch "bigger" things. Also didn't double check includes.

Please just squash everything into one commit. There are enough things that it'll just be easier to look at the full code after :)

We also prefer to let reviewers mark comments as "resolved". A lot of people keep track of comments they've addressed (or chosen not to) with :+1: and 👎 reactions to the review comments :)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
ermost: Hmm... I mean right now this is only checked with an ASSERT, and we have as default behavior to extrapolate. If we elevate this to the constructor, I feel something else should happen in this case...

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
ermost: I'll move them to private

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
ermost: I'm not sure. What happens if we only store one variable? 
Then the weights would not be useful... 

Is there a policy on using stuff like [[deprecated]] etc.?

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
ermost: It's a normalized relative coordinate. I added a comment.

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
ermost: I'm an "always auto" person :)

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
ermost: We are initializing it in the next line.

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
ermost: Ah no, that's a left over.

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
ermost: Exactly.

ermost: I cleaned up the comments to make this more clear

ermost:
I think I should have caught all review comments. I also made the test sample even smaller to make sure it doesn't timeout in debug mode.

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,387 @@
...
+  Weight<1> getWeights(const double x1) const;
+
+  /// Compute interpolation weights for 2D tables
+  Weight<2> getWeights(const double x1, const double x2) const;
+
+  /// Compute interpolation weights for 3D tables
+  Weight<3> getWeights(const double x1, const double x2, const double x3) const;
+
+  double interpolate(Weight<Dimension> const& weights,
+                     size_t which_variable = 0) const {
```
nilsdeppe: `const size_t`?

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,387 @@
...
+  struct Weight {
+    std::array<double, two_to_the(ThisDimension)> weights;
+    Index<two_to_the(ThisDimension)> index;
+  };
+
+  size_t extents(const size_t which_dimension) const {
+    return _number_of_points[which_dimension];
+  }
+
+  void extrapolateAboveData(size_t which_dimension, const bool value) {
```
nilsdeppe: `const size_t`?

also `snake_case` the function names: `extrapolate_above_data`, `extrapolate_below_data`, `get_weights`, `find_index`, `find_index_general`, `find_index_uniform`

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,387 @@
...
+    _allow_extrapolation_above_data[which_dimension] = value;
+  }
+
+  void extrapolateBelowData(size_t which_dimension, const bool value) {
```
nilsdeppe: `const size_t`?

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,387 @@
...
+  /// Compute interpolation weights for 1D tables
+  Weight<1> getWeights(const double x1) const;
+
+  /// Compute interpolation weights for 2D tables
+  Weight<2> getWeights(const double x1, const double x2) const;
+
+  /// Compute interpolation weights for 3D tables
+  Weight<3> getWeights(const double x1, const double x2, const double x3) const;
+
+  double interpolate(Weight<Dimension> const& weights,
```
nilsdeppe: `const Weight<Dimension>&` (west-const :) ) and a few places below too

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,387 @@
...
+                     size_t which_variable = 0) const {
+    double result = 0;
+
+    for (size_t nn = 0; nn < weights.weights.size(); ++nn) {
+      result += weights.weights[nn] *
+                _y[which_variable + NumberOfVariables * weights.index[nn]];
+    }
+
+    return result;
+  };
```
nilsdeppe: remove `;` (unneeded in this context)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,387 @@
...
+                          const double& target_points) const;
+
+  size_t findIndex(const size_t which_dimension,
+                   const double& target_points) const {
+    if constexpr (UniformSpacing) {
+      return findIndexUniform(which_dimension, target_points);
+    } else {
+      return findIndexGeneral(which_dimension, target_points);
+    }
+  };
```
nilsdeppe: remove `;`

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,387 @@
...
+    return _x[which_dimension][0];
+  };
+
+  double upper_bound(const size_t which_dimension) const {
+    return _x[which_dimension][_number_of_points[which_dimension] - 1];
+  };
+
+ private:
+  /// Allow extrapolation above table bounds.
+  std::array<bool, Dimension> _allow_extrapolation_below_data;
```
nilsdeppe: we use `_` after the variable name rather than before for member variables (apply to all below please)

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
nilsdeppe: Okay, sounds good! Let's try this and once we have some experience using the code we can always adjust :) 

We can use `[[deprecated]]` but we also use `-Werror` in CI so I'm not sure if that becomes problematic. I guess not because it would just prevent folks from committing new code that uses it.

tests/Unit/NumericalAlgorithms/Interpolation/Test_MultiLinearSpanInterpolation.cpp
```diff
@@ -0,0 +1,148 @@
...
+  std::array<double, NumVar> vector_b;
+
+  for (size_t i = 0; i < NumVar; ++i) {
+    vector_b[i] = dist_func(gen);
+    for (size_t j = 0; j < Dim; ++j) {
+      matrix[i][j] = dist_func(gen);
+    }
+  }
+
+  auto f = [&matrix, &vector_b](std::array<double, Dim> const& x) {
```
nilsdeppe: east-const

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
```diff
@@ -0,0 +1,387 @@
...
+        interpolate(weights, variables_to_interpolate)...};
+  }
+
+  template <size_t NumberOfvariables_to_interpolate, typename... T,
+            Requires<(std::is_floating_point_v<typename std::remove_cv_t<T>> and
+                      ...)> = nullptr>
+  std::array<double, NumberOfvariables_to_interpolate> interpolate(
+      std::array<size_t, NumberOfvariables_to_interpolate>&
+          variables_to_interpolate,
+      T const&... target_points) const {
```
nilsdeppe: `const T&...`

nilsdeppe (changes requested):
A few more minor stylistic things. Please rebase and squash immediately. I promise I'll get to the next review faster :D

nilsdeppe (changes requested):
Just some minor stylistic things :) Please rebase and squash immediately! I promise I'll get to the next review faster! :D

Edit: GitHub goofed, sorry for the accidental double comment!

src/NumericalAlgorithms/Interpolation/MultiLinearSpanInterpolation.hpp
ermost: OK!

ermost:
Thanks a lot for review! :)
I addressed all the comments.

ermost:
And yes, I'll try to become a west const person :)

nilsdeppe (approved):
Looks great! Thanks for implementing this :D

PR #4246, opened 2022-09-23
Remove Newton-Raphson from UniformCylindricalEndcap

markscheel:
## Proposed changes

Newton-Raphson is going away, so this PR removes it from UniformCylindricalEndcap.

Newton-Raphson was previously used only for corner cases where the root was barely bracketed.
Now those cases are handled differently.

Also does a few small tweaks of restrictions on map parameters that make the unit tests more robust
(I ran them 20k times without encountering an error)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/UniformCylindricalEndcap.cpp
+++ b/src/Domain/CoordinateMaps/UniformCylindricalEndcap.cpp
@@ -180,20 +180,20 @@ UniformCylindricalEndcap::UniformCylindricalEndcap(
       }()) {
   // Assumptions made in the map.  Some of these can be relaxed,
   // as long as the unit test is changed to test them.
-  ASSERT(z_plane_two >= z_plane_one + 0.03 * radius_one,
-         "z_plane_two must be >= z_plane_one + 0.03 * radius_one, not "
+  ASSERT(z_plane_two >= z_plane_one + 0.04 * radius_two,
+         "z_plane_two must be >= z_plane_one + 0.04 * radius_two, not "
              << z_plane_two << " " << z_plane_one << " "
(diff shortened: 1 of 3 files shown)


markscheel:
Uses toms748 so it will fail clang-tidy.

markscheel:
codecov failure is an error that should never happen (failure of toms748)

src/Domain/CoordinateMaps/UniformCylindricalEndcap.cpp
```diff
@@ -575,84 +576,122 @@ std::optional<std::array<double, 3>> UniformCylindricalEndcap::inverse(
...
+      // two bits.  If this is the case, then we set
+      // rhobar_min_increment to some small roundoff value with the
+      // same sign as trial_rhobar_min_increment.
+      // Note that rhobar is always between zero and one, so it is
+      // ok to use an absolute epsilon.
+      const double rhobar_min_increment =
+          abs(trial_rhobar_min_increment) >
+                  100.0 * std::numeric_limits<double>::epsilon()
+              ? trial_rhobar_min_increment
+              : boost::math::copysign(
```
nilsdeppe: https://en.cppreference.com/w/cpp/numeric/math/copysign ?
We also have `sgn` in `Utilities/Math.hpp` depending on exactly what you want

nilsdeppe (changes requested):
LGTM, squash the one thing right away please :)

wthrowe (commented):

markscheel:
changed to std::copysign and squashed.

nilsdeppe (approved):

wthrowe (approved):

PR #4247, opened 2022-09-23
Add UniformCylindricalSide.

markscheel:
## Proposed changes

Add a new map that maps a hollow cylinder to a portion of the region between two spheres.
This map works in concert with UniformCylindricalEndcap to produce a full "cylindered spherical shell"

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
+++ b/src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
@@ -1,0 +1,782 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/CoordinateMaps/UniformCylindricalSide.hpp"
+
+#include <cmath>
+#include <limits>
+#include <optional>
+#include <pup.h>
+#include <utility>
+
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "Parallel/PupStlCpp11.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/DereferenceWrapper.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+
+namespace domain::CoordinateMaps {
+
+namespace {
+// FunctionEvalTempVariables exists to remove code duplication between
+// function_to_zero and deriv_function_to_zero.
+struct FunctionEvalTempVariables {
+  FunctionEvalTempVariables(double lambda,
+                            const std::array<double, 3>& target_coords,
+                            const std::array<double, 3>& center_one,
+                            const std::array<double, 3>& center_two,
+                            double radius_one, double radius_two,
+                            double z_plane_minus_one, double z_plane_minus_two,
+                            double z_plane_plus_one, double z_plane_plus_two);
+  double one_plus_zbar_over_two;
+  double r_one_cos_theta_one;
+  double r_two_cos_theta_two;
+  double r_one_sin_theta_one;
+  double r_two_sin_theta_two;
+};
+
+FunctionEvalTempVariables::FunctionEvalTempVariables(
+    const double lambda, const std::array<double, 3>& target_coords,
+    const std::array<double, 3>& center_one,
+    const std::array<double, 3>& center_two, const double radius_one,
+    const double radius_two, const double z_plane_minus_one,
+    const double z_plane_minus_two, const double z_plane_plus_one,
+    const double z_plane_plus_two)
+    : one_plus_zbar_over_two(
+          (target_coords[2] + lambda * (z_plane_minus_one - z_plane_minus_two) -
+           z_plane_minus_one) /
+          ((1.0 - lambda) * (z_plane_plus_one - z_plane_minus_one) +
+           lambda * (z_plane_plus_two - z_plane_minus_two))),
+      r_one_cos_theta_one(z_plane_minus_one - center_one[2] +
+                          (z_plane_plus_one - z_plane_minus_one) *
+                              one_plus_zbar_over_two),
+      r_two_cos_theta_two(z_plane_minus_two - center_two[2] +
+                          (z_plane_plus_two - z_plane_minus_two) *
+                              one_plus_zbar_over_two),
+      r_one_sin_theta_one(
+          sqrt(square(radius_one) - square(r_one_cos_theta_one))),
+      r_two_sin_theta_two(
+          sqrt(square(radius_two) - square(r_two_cos_theta_two))) {}
+}  // namespace
+
+UniformCylindricalSide::UniformCylindricalSide(
+    const std::array<double, 3>& center_one,
+    const std::array<double, 3>& center_two, double radius_one,
+    double radius_two, double z_plane_plus_one, double z_plane_minus_one,
+    double z_plane_plus_two, double z_plane_minus_two)
+    : center_one_(center_one),
+      center_two_(center_two),
+      radius_one_([&radius_one]() {
+        // The equal_within_roundoff below has an implicit scale of 1,
+        // so the ASSERT may trigger in the case where we really
+        // want an entire domain that is very small.
+        ASSERT(not equal_within_roundoff(radius_one, 0.0),
+               "Cannot have zero radius_one");
+        ASSERT(radius_one > 0.0, "Cannot have negative radius_one");
+        return radius_one;
+                  }()), // LCOV_EXCL_LINE
+      radius_two_([&radius_two]() {
+        // The equal_within_roundoff below has an implicit scale of 1,
+        // so the ASSERT may trigger in the case where we really
+        // want an entire domain that is very small.
+        ASSERT(not equal_within_roundoff(radius_two, 0.0),
+               "Cannot have zero radius_two");
+        ASSERT(radius_two > 0.0, "Cannot have negative radius_two");
+        return radius_two;
+      }()), // LCOV_EXCL_LINE
+      z_plane_plus_one_(z_plane_plus_one),
+      z_plane_minus_one_(z_plane_minus_one),
+      z_plane_plus_two_(z_plane_plus_two),
+      z_plane_minus_two_(z_plane_minus_two) {
+  // Assumptions made in the map.  Some of these can be relaxed,
+  // as long as the unit test is changed to test them.
+  // The ASSERTS here match the ones in UniformCylindricalEndcap.
+  ASSERT(radius_one >= 0.08 * radius_two,
+         "Radius_one = " << radius_one << " must be >= 0.08 * radius_two ="
+                         << 0.08 * radius_two);
+  ASSERT(z_plane_plus_two == z_plane_plus_one or
+             z_plane_plus_two >= z_plane_plus_one + 0.03 * radius_two,
+         "z_plane_plus_two must be >= z_plane_plus_one "
+         "+ 0.03 * radius_two, or exactly equal to z_plane_plus_one, not "
+             << z_plane_plus_two << " " << z_plane_plus_one << " "
+             << z_plane_plus_one + 0.03 * radius_two);
+  ASSERT(z_plane_minus_two == z_plane_minus_one or
+             z_plane_minus_two <= z_plane_minus_one - 0.03 * radius_two,
+         "z_plane_minus_two must be >= z_plane_minus_one "
+         "- 0.03 * radius_two, or exactly equal to z_plane_minus_one, but "
+         "z_plane_minus_two = "
+             << z_plane_minus_two
+             << ", z_plane_minus_one = " << z_plane_minus_one
+             << ", and z_plane_minus_one - 0.03 * radius_two = "
+             << z_plane_minus_one - 0.03 * radius_two);
+  ASSERT(z_plane_minus_two != z_plane_minus_one or
+             z_plane_plus_two != z_plane_plus_one,
+         "Not tested if both the positive z-planes and the negative z-planes "
+         "are equal");
+
+  // The code below defines several variables that are used only in ASSERTs.
+  // We put that code in a #ifdef SPECTRE_DEBUG to avoid clang-tidy complaining
+  // about unused variables in release mode.
+#ifdef SPECTRE_DEBUG
+  const double cos_theta_min_one =
+      (z_plane_plus_one - center_one[2]) / radius_one;
+  const double cos_theta_max_one =
+      (z_plane_minus_one - center_one[2]) / radius_one;
+  const double cos_theta_min_two =
+      (z_plane_plus_two - center_two[2]) / radius_two;
+  const double cos_theta_max_two =
+      (z_plane_minus_two - center_two[2]) / radius_two;
+
+  ASSERT(cos_theta_min_one > cos(M_PI * 0.4),
+         "z_plane_plus_one is too close to the center of sphere_one: "
+         "cos_theta_min_one must be > "
+             << cos(M_PI * 0.4) << " but is " << cos_theta_min_one
+             << " instead.");
+  ASSERT(cos_theta_max_one < cos(M_PI * 0.6),
+         "z_plane_minus_one is too close to the center of sphere_one: "
+         "cos_theta_max_one must be < "
+             << cos(M_PI * 0.6) << " but is " << cos_theta_max_one
+             << " instead.");
+  ASSERT(cos_theta_min_one < cos(M_PI * 0.15),
+         "z_plane_plus_one is too far from the center of sphere_one: "
+         "cos_theta_min_one must be < "
+             << cos(M_PI * 0.15) << " but is " << cos_theta_min_one
+             << " instead.");
+  ASSERT(cos_theta_max_one > cos(M_PI * 0.85),
+         "z_plane_minus_one is too far from the center of sphere_one: "
+         "cos_theta_max_one must be > "
+             << cos(M_PI * 0.85) << " but is " << cos_theta_max_one
+             << " instead.");
+  ASSERT(cos_theta_min_two > (z_plane_plus_one_ == z_plane_plus_two_
+                                  ? cos(M_PI * 0.75)
+                                  : cos(M_PI * 0.4)),
+         "z_plane_plus_two is too close to the south pole: theta/pi="
+             << acos(cos_theta_min_two) / M_PI);
+  ASSERT(cos_theta_max_two < (z_plane_minus_one_ == z_plane_minus_two_
+                                  ? cos(M_PI * 0.25)
+                                  : cos(M_PI * 0.6)),
+         "z_plane_minus_two is too close to the north pole: theta/pi="
+             << acos(cos_theta_max_two) / M_PI);
+  ASSERT(cos_theta_min_two < cos(M_PI * 0.15),
+         "z_plane_plus_two is too close to the north pole: theta_min_two/pi="
+             << acos(cos_theta_min_two) / M_PI << " but it should be < 0.15");
+  ASSERT(cos_theta_max_two > cos(M_PI * 0.85),
+         "z_plane_minus_two is too close to the south pole: theta_max_two/pi="
+             << acos(cos_theta_max_two) / M_PI << " but it should be > 0.85");
+
+  const double dist_spheres = sqrt(square(center_one[0] - center_two[0]) +
+                                   square(center_one[1] - center_two[1]) +
+                                   square(center_one[2] - center_two[2]));
+
+  ASSERT(dist_spheres + radius_one <= 0.98 * radius_two,
+         "The map has been tested only for the case when "
+         "sphere_one is sufficiently contained inside sphere_two, without the "
+         "two spheres almost touching. Radius_one = "
+             << radius_one << ", radius_two = " << radius_two
+             << ", dist_spheres = " << dist_spheres
+             << ", (dist_spheres+radius_one)/radius_two="
+             << (dist_spheres + radius_one) / radius_two);
+
+  const double horizontal_dist_spheres =
+      sqrt(square(center_one[0] - center_two[0]) +
+           square(center_one[1] - center_two[1]));
+
+  if (z_plane_plus_two != z_plane_plus_one and
+      z_plane_minus_two != z_plane_minus_one) {
+    ASSERT(center_one[2] - radius_one <= center_two[2] + 0.2 * radius_two and
+               center_one[2] + radius_one >= center_two[2] - 0.2 * radius_two,
+           "The map has been tested only for the case when "
+           "sphere_one is not a tiny thing at the top or bottom of sphere_two. "
+           " Radius_one = "
+               << radius_one << ", radius_two = " << radius_two
+               << " center_one[2] = " << center_one[2]
+               << " center_two[2] = " << center_two[2]);
+  }
+
+  if (z_plane_plus_two == z_plane_plus_one) {
+    ASSERT(z_plane_minus_two <= z_plane_plus_two - 0.18 * radius_two,
+           "The map has been tested only if the sphere_two planes are far "
+           "enough apart. z_plane_plus_two="
+               << z_plane_plus_two << " z_plane_minus_two=" << z_plane_minus_two
+               << " radius_two=" << radius_two);
+  }
+
+  if (z_plane_minus_two == z_plane_minus_one) {
+    ASSERT(z_plane_plus_two >= z_plane_minus_two + 0.18 * radius_two,
+           "The map has been tested only if the sphere_two planes are far "
+           "enough apart. z_plane_plus_two="
+               << z_plane_plus_two << " z_plane_minus_two=" << z_plane_minus_two
+               << " radius_two=" << radius_two);
+  }
+
+  ASSERT(horizontal_dist_spheres <=
+             (z_plane_plus_two == z_plane_plus_one or
+                      z_plane_minus_two == z_plane_minus_one
+                  ? 0.0
+                  : std::min(radius_one,
+                             std::max(0.0, 0.95 * radius_two - radius_one))),
+         "The map has been tested only for the case when "
+         "sphere_one intersects the polar axis of sphere_two and is not "
+         "too far from the edge of sphere_two (or in the "
+         "case of equal plus or minus z_planes, when the two spheres have the "
+         "same x and y centers)."
+         " Radius_one = "
+             << radius_one << ", radius_two = " << radius_two
+             << ", dist_spheres = " << dist_spheres
+             << ", horizontal_dist_spheres = " << horizontal_dist_spheres);
+
+  const double theta_min_one = acos(cos_theta_min_one);
+  const double theta_max_one = acos(cos_theta_max_one);
+  const double theta_min_two = acos(cos_theta_min_two);
+  const double theta_max_two = acos(cos_theta_max_two);
+
+  // max_horizontal_dist_between_circles can be either sign;
+  // therefore alpha can be in either the first or second quadrant.
+  const double max_horizontal_dist_between_circles_plus =
+      horizontal_dist_spheres + radius_one_ * sin(theta_min_one) -
+      radius_two_ * sin(theta_min_two);
+  const double max_horizontal_dist_between_circles_minus =
+      horizontal_dist_spheres + radius_one_ * sin(theta_max_one) -
+      radius_two_ * sin(theta_max_two);
+
+  const double alpha_plus = atan2(z_plane_plus_two - z_plane_plus_one,
+                                  max_horizontal_dist_between_circles_plus);
+  ASSERT(alpha_plus > 1.1 * theta_min_one and alpha_plus > 1.1 * theta_min_two,
+         "Angle alpha_plus is too small: alpha_plus = "
+             << alpha_plus << ", theta_min_one = " << theta_min_one
+             << ", theta_min_two = " << theta_min_two
+             << ", max_horizontal_dist_between_circles_plus = "
+             << max_horizontal_dist_between_circles_plus
+             << ", horizontal_dist_spheres = " << horizontal_dist_spheres);
+
+  const double alpha_minus = atan2(z_plane_minus_one - z_plane_minus_two,
+                                   max_horizontal_dist_between_circles_minus);
+  ASSERT(alpha_minus > 1.1 * (M_PI - theta_max_one) and
+             alpha_minus > 1.1 * (M_PI - theta_max_two),
+         "Angle alpha_minus is too small: alpha_minus = "
+             << alpha_minus << ", theta_max_one = " << theta_max_one
+             << ", theta_max_two = " << theta_max_two
+             << ", max_horizontal_dist_between_circles_minus = "
+             << max_horizontal_dist_between_circles_minus
+             << ", horizontal_dist_spheres = " << horizontal_dist_spheres);
+#endif
+}
+
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, 3> UniformCylindricalSide::operator()(
+    const std::array<T, 3>& source_coords) const {
+  using ReturnType = tt::remove_cvref_wrap_t<T>;
+  const ReturnType& xbar = source_coords[0];
+  const ReturnType& ybar = source_coords[1];
+  const ReturnType& zbar = source_coords[2];
+  std::array<ReturnType, 3> target_coords{};
+  ReturnType& x = target_coords[0];
+  ReturnType& y = target_coords[1];
+  ReturnType& z = target_coords[2];
+
+  // Use y and z as temporary storage to avoid allocations,
+  // before setting them to their actual values.
+  z = sqrt(square(xbar) + square(ybar));  // z is rhobar = lambda+1 in the dox
+  // y is temporarily R1 sin(theta_1)(1-lambda) + R2 sin(theta_2) lambda
+  // Here R1 sin(theta_1) is calculated from R1 cos(theta_1) as
+  // expressed in the dox
+  y = sqrt(square(radius_two_) -
+           square(z_plane_minus_two_ - center_two_[2] +
+                  0.5 * (zbar + 1.0) *
+                      (z_plane_plus_two_ - z_plane_minus_two_))) *
+          (z - 1.0) +
+      sqrt(square(radius_one_) -
+           square(z_plane_minus_one_ - center_one_[2] +
+                  0.5 * (zbar + 1.0) *
+                      (z_plane_plus_one_ - z_plane_minus_one_))) *
+          (2.0 - z);
+  // Now compute actual x,y,z.
+  x = y * xbar / z + center_two_[0] * (z - 1.0) + center_one_[0] * (2.0 - z);
+  y = y * ybar / z + center_two_[1] * (z - 1.0) + center_one_[1] * (2.0 - z);
+  z = (z_plane_minus_two_ +
+       0.5 * (zbar + 1.0) * (z_plane_plus_two_ - z_plane_minus_two_)) *
+          (z - 1.0) +
+      (z_plane_minus_one_ +
+       0.5 * (zbar + 1.0) * (z_plane_plus_one_ - z_plane_minus_one_)) *
+          (2.0 - z);
+  return target_coords;
+}
+
+std::optional<std::array<double, 3>> UniformCylindricalSide::inverse(
+    const std::array<double, 3>& target_coords) const {
+  // First do some easy checks that target_coords is in the range
+  // of the map.  Need to accept points that are out of range by roundoff.
+
+  // check z <= z_plane_plus_two_
+  if (target_coords[2] > z_plane_plus_two_ and
+      not equal_within_roundoff(target_coords[2], z_plane_plus_two_)) {
+    return std::nullopt;
+  }
+
+  // check z >= z_plane_minus_two_
+  if (target_coords[2] < z_plane_minus_two_ and
+      not equal_within_roundoff(target_coords[2], z_plane_minus_two_)) {
+    return std::nullopt;
+  }
+
+  // check that point is in or on sphere_two
+  const double r_minus_center_two =
+      sqrt(square(target_coords[0] - center_two_[0]) +
+           square(target_coords[1] - center_two_[1]) +
+           square(target_coords[2] - center_two_[2]));
+  if (r_minus_center_two > radius_two_ and
+      not equal_within_roundoff(r_minus_center_two, radius_two_)) {
+    return std::nullopt;
+  }
+
+  // check that point is outside or on sphere_one_
+  const double r_minus_center_one =
+      sqrt(square(target_coords[0] - center_one_[0]) +
+           square(target_coords[1] - center_one_[1]) +
+           square(target_coords[2] - center_one_[2]));
+  if (r_minus_center_one < radius_one_ and
+      not equal_within_roundoff(r_minus_center_one, radius_one_)) {
+    return std::nullopt;
+  }
+
+  // Check if the point is inside the cone
+  auto point_inside_cone =
+      [this, &target_coords](const double z_one, const double z_two) {
+        const double lambda_tilde =
+            (target_coords[2] - z_one) / (z_two - z_one);
+        const double rho_tilde =
+            sqrt(square(target_coords[0] - center_one_[0] -
+                        lambda_tilde * (center_two_[0] - center_one_[0])) +
+                 square(target_coords[1] - center_one_[1] -
+                        lambda_tilde * (center_two_[1] - center_one_[1])));
+        const double circle_radius =
+            sqrt(square(radius_one_) -
+                 square(z_one - center_one_[2])) *
+                (1.0 - lambda_tilde) +
+            sqrt(square(radius_two_) -
+                 square(z_two - center_two_[2])) *
+                lambda_tilde;
+        return rho_tilde < circle_radius and
+               not equal_within_roundoff(
+                   rho_tilde, circle_radius,
+                   std::numeric_limits<double>::epsilon() * 100.0, radius_two_);
+      };
+  if ((target_coords[2] >= z_plane_plus_one_ and
+       z_plane_plus_one_ != z_plane_plus_two_ and
+       point_inside_cone(z_plane_plus_one_, z_plane_plus_two_)) or
+      (target_coords[2] <= z_plane_minus_one_ and
+       z_plane_minus_one_ != z_plane_minus_two_ and
+       point_inside_cone(z_plane_minus_one_, z_plane_minus_two_))) {
+    return std::nullopt;
+  }
+
+  // To find lambda we will do numerical root finding.
+  // function_to_zero is the function that is zero when lambda has the
+  // correct value.
+  const auto function_to_zero = [this, &target_coords](const double lambda) {
+    const FunctionEvalTempVariables tmp(lambda, target_coords, center_one_,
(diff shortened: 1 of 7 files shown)


markscheel:
Made a draft while I fix some codecov things.

markscheel:
clang-tidy failures are toms748

markscheel:
codecov doesn't display here anymore, dunno why.

markscheel:
Made a few changes to add a new use case that I didn't realize we would need.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -0,0 +1,606 @@
...
+/*!
+ * \ingroup CoordinateMapsGroup
+ *
+ * \brief Map from 3D unit right cylindrical shell to a volume that connects
+ *  portions of two spherical surfaces.
+ *
+ * \image html UniformCylSide.svg "A hollow cylinder maps to the shaded region."
+ *
+ * \details Consider two spheres with centers \f$C_1\f$ and \f$C_2\f$,
+ * and radii \f$R_1\f$ and \f$R_2\f$. Let sphere 1 be intersected by two
```
knelli2: Is it requried that sphere 1 be nested inside sphere 2?

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -0,0 +1,606 @@
...
+ *
+ * UniformCylindricalSide maps a 3D unit right cylindrical shell (with
+ * coordinates \f$(\bar{x},\bar{y},\bar{z})\f$ such that
+ * \f$-1\leq\bar{z}\leq 1\f$ and \f$1 \leq \bar{x}^2+\bar{y}^2 \leq
+ * 4\f$) to the shaded area in the figure above (with coordinates
+ * \f$(x,y,z)\f$).  The "inner surface" of the cylindrical shell
+ * \f$\bar{x}^2+\bar{y}^2=1\f$ is mapped to the portion of sphere 1
+ * that has \f$z^{-}_{\mathrm{P}1} \leq z \leq z^{+}_{\mathrm{P}1} \f$,
+ * and on this portion of the sphere the cosine of the angular coordinate
+ * \f$\cos\theta_1 =(z-C_1^2)/R_1\f$ is uniform in \f$\bar{z}\f$,
```
knelli2: I'm pretty sure $\theta$ here is the usual polar angle, but since we are dealing with both spheres and cylinders, I think it's best to be explicit whenever we define the angles so as to avoid confusion. I think changing it to `cosine of the polar angular coordinate`.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -0,0 +1,606 @@
...
+ * and the angular coordinate \f$\phi_1 = \atan((y-C_1^1)/(x-C_1^0))\f$
+ * is the same as \f$\phi = \atan(\bar{y}/\bar{x})\f$.
```
knelli2: `and the azimuthal angular coordinate...`

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -0,0 +1,606 @@
...
+ * UniformCylindricalSide maps a 3D unit right cylindrical shell (with
+ * coordinates \f$(\bar{x},\bar{y},\bar{z})\f$ such that
+ * \f$-1\leq\bar{z}\leq 1\f$ and \f$1 \leq \bar{x}^2+\bar{y}^2 \leq
+ * 4\f$) to the shaded area in the figure above (with coordinates
+ * \f$(x,y,z)\f$).  The "inner surface" of the cylindrical shell
+ * \f$\bar{x}^2+\bar{y}^2=1\f$ is mapped to the portion of sphere 1
+ * that has \f$z^{-}_{\mathrm{P}1} \leq z \leq z^{+}_{\mathrm{P}1} \f$,
+ * and on this portion of the sphere the cosine of the angular coordinate
+ * \f$\cos\theta_1 =(z-C_1^2)/R_1\f$ is uniform in \f$\bar{z}\f$,
+ * and the angular coordinate \f$\phi_1 = \atan((y-C_1^1)/(x-C_1^0))\f$
```
knelli2: Since you're using $x, y, z$ to represent the coordinates, I think it would look cleaner if $C_1^i$ used them as well. So $C_1^0 \rightarrow C_1^x$. (The $x, y, z$ here may need bars depending on what coordinates they are.)

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -0,0 +1,606 @@
...
+ * \f$z = z^{\pm}_{\mathrm{P}1}\f$,
+ * and let sphere 2 be intersected by two planes normal to the \f$z\f$ axis and
+ * located at \f$z = z^{\pm}_{\mathrm{P}2}\f$.  Here we assume that
+ * \f$z^{-}_{\mathrm{P}2} \leq z^{-}_{\mathrm{P}1}<
+ * z^{+}_{\mathrm{P}1} \leq z^{+}_{\mathrm{P}2}\f$.
+ *
+ * UniformCylindricalSide maps a 3D unit right cylindrical shell (with
+ * coordinates \f$(\bar{x},\bar{y},\bar{z})\f$ such that
+ * \f$-1\leq\bar{z}\leq 1\f$ and \f$1 \leq \bar{x}^2+\bar{y}^2 \leq
+ * 4\f$) to the shaded area in the figure above (with coordinates
```
knelli2: Is having the outer radius of the cylindrical shell have a radius of 2 just an arbitrary choice? I think it is but just want to make sure.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -0,0 +1,606 @@
...
+ * \f{align}
+ * x^0 &= C_1^0+\lambda(C_2^0-C_1^0) +
+ *        \cos\phi\left(R_1\sin\theta_1 +
+ *        \lambda(R_2\sin\theta_2-R_1\sin\theta_1)\right), \label{eq:x0} \\
+ * x^1 &= C_1^1+\lambda(C_2^1-C_1^1) +
+ *        \sin\phi\left(R_1\sin\theta_1 +
+ *        \lambda(R_2\sin\theta_2-R_1\sin\theta_1)\right), \label{eq:x1} \\
+ * x^2 &= C_1^2+\lambda(C_2^2-C_1^2) +
+ *        R_1\cos\theta_1 +
+ *        \lambda(R_2\cos\theta_2-R_1\cos\theta_1) \label{eq:x2}.
```
knelli2: You say above that these are equations for $x, y, z$ but then here they are $x^0, x^1, x^2$. I think consistency is best here. I'm leaning more towards using $x, y, z$ so we avoid confusion with $x^2$ on whether that's the $z$ coordinate or the square of the $x$ coordinate.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -0,0 +1,606 @@
...
+ * Finally, for \f$z^{+}_{\mathrm{P}2}\neq z^{+}_{\mathrm{P}1}\f$,
+ * consider the circle \f$S^{+}_1\f$
+ * defining the intersection of sphere 1
+ * and the plane \f$z = z^{+}_{\mathrm{P}1}\f$; this circle has radius
+ * \f$r_1 = R_1 \sin\theta_{1\mathrm{min}}\f$.  Similarly, the circle
+ * \f$S^{+}_2\f$ defining the intersection of sphere 2 and the plane \f$z =
+ * z^{+}_{\mathrm{P}2}\f$ has radius \f$r_2 = R_2
+ * \sin\theta_{2\mathrm{min}}\f$.  Now consider the cone that passes
+ * through these two circles.  A point in the range of the map must be outside
+ * or on this cone. The cone can be defined parametrically as
```
knelli2: By "outside" you mean further away from the $z$ axis, correct? Either way, maybe add a note explaining what "outside" means.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -0,0 +1,606 @@
...
+ * \f}
+ *
+ * ### inv_jacobian
+ *
+ * The inverse Jacobian is computed by numerically inverting the
+ * Jacobian.
+ *
+ * ### Restrictions on map parameters
+ *
+ * We demand that Sphere 1 is fully contained inside Sphere 2, and
```
knelli2, line 477: Ahh ok here's the condition on which sphere contains which. Still might be good to also mention this at the beginning.

knelli2 (changes requested):
Only a partial review. Went through the docs and checked for clarity and typos and checked most of the equations (except jacobian). I still need to check the code vs the docs.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -346,8 +352,8 @@ namespace domain::CoordinateMaps {
  * \f{align}
  *  \tilde{\lambda} &= \frac{z - z^{+}_{\mathrm{P}1}}
  *   {z^{+}_{\mathrm{P}2}-z^{+}_{\mathrm{P}1}}, \\
- *  \tilde{x} &= x - C_1^0 - \tilde{\lambda} (C_2^0-C_1^0),\\
- *  \tilde{y} &= y - C_1^1 - \tilde{\lambda} (C_2^1-C_1^1).\\
+ *  \tilde{x} &= x - C_1^x - \tilde{\lambda} (C_2^x-C_1^x),\\
+ *  \tilde{y} &= y - C_1^x - \tilde{\lambda} (C_2^y-C_1^y).\\
```
knelli2: In the second line (old 350, new 356), it should be `C_1^y` correct?

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -548,6 +515,85 @@ namespace domain::CoordinateMaps {
...
+ *
+ * We prohibit a tiny Sphere 1 near the edge of Sphere 2 by demanding that
+ * \f{align}
+ *  C^z_1 - R_1 &\leq C^z_2 + R_2/5,\\
+ *  C^z_1 + R_1 &\geq C^z_2 - R_2/5.
+ * \f}
+ * We also demand that the polar axis of Sphere 2 intersects Sphere 1
+ * somewhere:
+ * \f{align}
+ * \sqrt{(C^0_1-C^0_2)^2 + (C^1_1-C^1_2)^2} &\leq R_1.
```
knelli2: $0, 1, 2 \rightarrow x, y, z$ for this whole `\f{align}` block and the next one.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -548,6 +515,85 @@ namespace domain::CoordinateMaps {
...
+ * \f}
+ * and we demand that Sphere 1 is not too close to the edge of Sphere 2
+ * in the \f$x\f$ or \f$y\f$ directions:
+ * \f{align}
+ * \sqrt{(C^0_1-C^0_2)^2 + (C^1_1-C^1_2)^2} &\leq \mathrm{max}(0,0.95 R_2-R_1),
```
knelli2: If $0.95R_2 - R_1$ is negative, then this quantity has to be equal to zero, correct? Which would imply the $x, y$ coordinates of the two spheres are equal? Just want to make sure I understand.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
```diff
@@ -548,6 +515,85 @@ namespace domain::CoordinateMaps {
...
+ * appropriate parameter ranges.
+ *
+ * ### Restrictions for equal z planes
+ *
+ * If \f$z^+_{\mathrm{P}1} = z^+_{\mathrm{P}2}\f$ or
+ * \f$z^-_{\mathrm{P}1} = z^-_{\mathrm{P}2}\f$ we demand that
+ * \f$C_1^x=C_2^x\f$ and \f$C_1^y=C_2^y\f$, which simplifies the cases
+ * we need to test and agrees with our expected use cases.
+ * We also demand
+ * - \f$z^+_{\mathrm{P}2} \geq z^-_{\mathrm{P}2} + 0.18 R_2\f$
```
knelli2: Should this be in a `\f{align} \f}` block instead of a bullet point? Also, I think you are missing a condition here for $z_{\mathrm{P}1}^{\pm}$

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+      }()),
+      z_plane_plus_one_(z_plane_plus_one),
+      z_plane_minus_one_(z_plane_minus_one),
+      z_plane_plus_two_(z_plane_plus_two),
+      z_plane_minus_two_(z_plane_minus_two) {
+  // Assumptions made in the map.  Some of these can be relaxed,
+  // as long as the unit test is changed to test them.
+  // The ASSERTS here match the ones in UniformCylindricalEndcap.
+  ASSERT(radius_one >= 0.08 * radius_two,
+         "Radius_one too small " << radius_one << " " << 0.06 * radius_two);
```
knelli2: The `0.06` in the assert message should be `0.08`. Also, in the docs, you require $R_1 > 0.08R_2$, but here you have $R_1 >= 0.08R_2$.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+  ASSERT(z_plane_plus_two == z_plane_plus_one or
+             z_plane_plus_two >= z_plane_plus_one + 0.03 * radius_two,
```
knelli2, line 101: In the docs, you use $R_1$, but here you use $R_2$.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+         "z_plane_plus_two must be >= z_plane_plus_one "
+         "+ 0.03 * radius_two, or exactly equal to z_plane_plus_one, not "
+             << z_plane_plus_two << " " << z_plane_plus_one << " "
+             << z_plane_plus_one + 0.03 * radius_two);
+  ASSERT(z_plane_minus_two == z_plane_minus_one or
+             z_plane_minus_two <= z_plane_minus_one - 0.03 * radius_two,
```
knelli2, line 107: In the docs, you use $R_1$, but here you use $R_2$.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+      (z_plane_minus_one - center_one[2]) / radius_one;
+  const double cos_theta_min_two =
+      (z_plane_plus_two - center_two[2]) / radius_two;
+  const double cos_theta_max_two =
+      (z_plane_minus_two - center_two[2]) / radius_two;
+
+  ASSERT(
+      cos_theta_min_one > cos(M_PI * 0.4),
+      "z_plane_plus_one is too close to the center of sphere_one: cos(theta) = "
+          << cos_theta_min_one);
```
knelli2: For all of these asserts, could add a bit more to the error message? I like the first sentence, but then maybe something after it like
```
"cos_theta_min_one must be > cos(0.4 * pi), but is " << cos_theta_min_one << " instead."
```

Likewise for the other asserts. This map is very complicated so the more verbose the error message, the better IMO.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+             (z_plane_plus_two == z_plane_plus_one or
+                      z_plane_minus_two == z_plane_minus_one
+                  ? 0.0
+                  : std::min(radius_one,
+                             std::max(0.0, 0.95 * radius_two - radius_one))),
+         "The map has been tested only for the case when "
+         "sphere_one intersects the polar axis of sphere_two and is not "
+         "too far from the edge of sphere_two (or in the "
+         "case of equal plus or minus z_planes, when the two spheres have the "
+         "same x and y centers."
```
knelli2: End parenthesis.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+      horizontal_dist_spheres + radius_one_ * sin(theta_min_one) -
+      radius_two_ * sin(theta_min_two);
+  const double max_horizontal_dist_between_circles_minus =
+      horizontal_dist_spheres + radius_one_ * sin(theta_max_one) -
+      radius_two_ * sin(theta_max_two);
+
+  const double alpha_plus = atan2(z_plane_plus_two - z_plane_plus_one,
+                                  max_horizontal_dist_between_circles_plus);
+  ASSERT(alpha_plus > 1.1 * theta_min_one and alpha_plus > 1.1 * theta_min_two,
+         "Angle alpha_plus is too small: alpha = "
```
knelli2: Call it "alpha_plus" in the error message. Same for alpha_minus

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+  // therefore alpha can be in either the first or second quadrant.
+  const double max_horizontal_dist_between_circles_plus =
+      horizontal_dist_spheres + radius_one_ * sin(theta_min_one) -
+      radius_two_ * sin(theta_min_two);
+  const double max_horizontal_dist_between_circles_minus =
+      horizontal_dist_spheres + radius_one_ * sin(theta_max_one) -
+      radius_two_ * sin(theta_max_two);
+
+  const double alpha_plus = atan2(z_plane_plus_two - z_plane_plus_one,
+                                  max_horizontal_dist_between_circles_plus);
```
knelli2, line 246: Question, if the two plus z planes are equal, then alpha_plus is zero, right? But below, you compare to `theta_min_one` and `theta_min_two` based on equations `\label{eq:theta_1_min_res}` through `\label{eq:theta_2_max_res}` in the hpp file. What happens there?

markscheel:
Fixed things pointed out by @knelli2 and redid the restrictions when some of the z-planes are equal (which changes how parameters are chosen for the tests, and the checks in the constructor, but none of the actual code because the equations didn't change).  I did this as a new commit to make it easier for review.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+  std::array<ReturnType, 3> target_coords{};
+  ReturnType& x = target_coords[0];
+  ReturnType& y = target_coords[1];
+  ReturnType& z = target_coords[2];
+
+  // Use y and z as temporary storage to avoid allocations,
+  // before setting them to their actual values.
+  z = sqrt(square(xbar) + square(ybar));  // z is rhobar = lambda+1 in the dox
+  // y is temporarily R1 sin(theta_1)(1-lambda) + R2 sin(theta_2) lambda
+  y = sqrt(square(radius_two_) -
```
knelli2: Please add this comment for clarity on where this massive expression comes from
```
// where R1 sin(theta_1) is calculated from R1 cos(theta_1) as expressed in the dox
```

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+
+std::optional<std::array<double, 3>> UniformCylindricalSide::inverse(
+    const std::array<double, 3>& target_coords) const {
+  // First do some easy checks that target_coords is in the range
+  // of the map.  Need to accept points that are out of range by roundoff.
+
+  // check z <= z_plane_plus_two_
+  if (target_coords[2] > z_plane_plus_two_ and
+      not equal_within_roundoff(target_coords[2], z_plane_plus_two_)) {
+    return {};
```
knelli2: Just return `std::nullopt` for points that are invalid for the map. (same for the rest of this function) Also change the docs to say it just returns nullopt

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+  }
+
+  // check z >= z_plane_minus_two_
+  if (target_coords[2] < z_plane_minus_two_ and
+      not equal_within_roundoff(target_coords[2], z_plane_minus_two_)) {
+    return {};
+  }
+
+  // check that point is in or on sphere_two
+  const double r_minus_center_two_squared =
```
knelli2: Would it be better to take the `sqrt` immediately so you don't have to `square(radius_two_)` and then `sqrt(r_minus_center_two_squared)` later on? Same for the on or outside `sphere_one_` check.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+    // equal_within_roundoff too much.
+    return {};
+  }
+
+  // Check if the point is inside the cone
+  auto point_inside_cone =
+      [&target_coords](const double z_one, const double z_two,
+                       const double radius_one, const double radius_two,
+                       const std::array<double, 3>& center_one,
+                       const std::array<double, 3>& center_two) {
```
knelli2: Nitpick: You pass `radius_one_, radius_two_, center_one_, center_two_` for both of the calls to this function. Why not just capture `this` with `[this, &target_coords]` and use them directly? This change is not necessary, just helps with code readability.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,751 @@
...
+        if (function_at_new_lambda_min * function_at_lambda_min > 0.0) {
+          ERROR(
+              "Cannot find bracket after trying to adjust bracketing "
+              "lambda_min due "
+              "to roundoff : "
+              << lambda_min << " " << function_at_lambda_min << " "
+              << lambda_max << " " << function_at_lambda_max << " "
+              << new_lambda_min << " " << function_at_new_lambda_min << " "
+              << deriv_function_at_lambda_min << " "
+              << new_lambda_min - lambda_min << "\n");
```
knelli2: Can you label these in the error message? Otherwise it's just a sequence of seemingly random numbers. (Same below for the `max` branch, the error branch, and the root finder failure branch below)

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,517 @@
...
+    // of center_one[2].
+    const double max_radius_one_to_fit_between_plane_twos =
+        (z_plane_plus_two - z_plane_minus_two -
+         (z_upper_separation + z_lower_separation) * radius_two) /
+        (z_plane_frac_plus_one - z_plane_frac_minus_one);
+    CHECK(max_radius_one_to_fit_between_plane_twos > 0.0);
+
+    // We add an additional safety factor of 0.99 to
+    // max_radius_one_to_fit_inside_sphere_two so that radius_one
+    // doesn't get quite that large.
```
knelli2: If you don't add this safety factor, will the map fail?

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,517 @@
...
+                  center_two[2] - 0.98 * radius_two + radius_one});
+    CHECK(min_center_one_z <= max_center_one_z);
+    return min_center_one_z +
+           unit_dis(gen) * (max_center_one_z - min_center_one_z);
+  }();
+  CAPTURE(center_one_z);
+
+  // Now we can compute z_plane_plus_one and z_plane_minus_one If
+  // upper and lower planes are equal, put in exact value so there is
+  // no roundoff.
```
knelli2: I don't see where the replacement with the exact values is done if they are equal.

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,517 @@
...
+      [&z_plane_plus_one, &z_plane_plus_two, &theta_max_plus_one,
+       &theta_max_plus_two, &min_alpha_plus, &z_plane_minus_one,
+       &z_plane_minus_two, &theta_max_minus_one, &theta_max_minus_two,
+       &min_alpha_minus, &center_one_z, &center_two, &radius_one, &radius_two,
+       &unit_dis, &gen]() {
+        // Let rho be the horizontal (x-y) distance between the centers of
+        // the spheres.
+
+        // maximum rho so that sphere2 is inside of sphere 1, with a
+        // safety factor of 0.98
```
knelli2: Again, will the map fail if this safety factor isn't there?

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,517 @@
...
+      center_one, center_two, radius_one, radius_two, z_plane_plus_one,
+      z_plane_minus_one, z_plane_plus_two, z_plane_minus_two);
+  test_suite_for_map_on_cylinder(map, 1.0, 2.0, true, true);
+
+  // The following are tests that the inverse function correctly
+  // returns an invalid std::optional when called for a point that is
+  // outside the range of the map.
+
+  // Point with z > z_plane_plus_two.
+  CHECK_FALSE(map.inverse({{0.0, 0.0, z_plane_plus_two + 1.0}}));
```
knelli2: Explicitly call `.has_value()` here rather than implicitely converting an optional to a bool. Same for the rest fo these checks also

knelli2 (changes requested):
Phew!! Ok this is a complete review. It's a mix of questions and change requests. If there's anything that is unclear, just let me know

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
markscheel: Fixed.  Docs were wrong.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
markscheel: Done. But there is no extra condition for z^+-_P1 in addition to the ones already listed.  The reason that the extra condition for z^+-_P2 is needed is that previously, theta_2 min and max were required to be on opposite sides of the equator. I added a sentence about that in the dox.  But for sphere 1, theta min and theta max still need to be on opposite sides of the equator.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
markscheel: Uh oh. If the plus z planes are equal, I think alpha_plus should be pi, not zero. I may be computing alpha_plus incorrectly here, I need to check this.

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
markscheel: Oops. That comment should have been removed after my previous commit (nowadays equal z-planes have their own test and are not treated in this test).  Removed now.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
markscheel: For equal z-planes, the xy coords of the sphere centers are equal, so max_horizontal_dist_between_circles_plus is negative.  So atan2 should return pi, so I think this is ok.

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
markscheel: I reworded the comment.  This safety factor isn't new; its the same restriction that  `0.98 R_2 <= R_1 + |C_1-C_2|` but that places a limit on rho (because the radii and z-coords of the centers are now known).

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
markscheel: Yes, I added a comment about that.

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
markscheel: Yup. Fixed.

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
markscheel: The safety factor turns out to be unnecessary.  I removed it and ran the test 10,000 times and got no failures.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,749 @@
...
+             << ", theta_min_two = " << theta_min_two
+             << ", max_horizontal_dist_between_circles_plus = "
+             << max_horizontal_dist_between_circles_plus
+             << ", horizontal_dist_spheres = " << horizontal_dist_spheres);
+
+  const double alpha_minus = atan2(z_plane_minus_one - z_plane_minus_two,
+                                   max_horizontal_dist_between_circles_minus);
+  ASSERT(alpha_minus > 1.1 * (M_PI - theta_max_one) and
+             alpha_minus > 1.1 * (M_PI - theta_max_two),
+         "Angle alpha_minus is too small: alpha_plus = "
```
knelli2: `alpha_minus` not `alpha_plus` here.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
knelli2: Ahh ok I didn't realize `atan2` took into account the signs of each argument. Looks good then.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
knelli2: Ping for the labels on the "root is not bracketed" branch below.

knelli2 (changes requested):
Please squash all changes including these last couple. Looks good!

markscheel:
Made changes and squashed.

knelli2 (dismissed):
Ok I'm happy with everything! @geoffrey4444 can you please take a look at this?

src/Domain/CoordinateMaps/UniformCylindricalSide.hpp
geoffrey4444: It looks like now this is mentioned at the beginning

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+      radius_one_([&radius_one]() {
+        // The equal_within_roundoff below has an implicit scale of 1,
+        // so the ASSERT may trigger in the case where we really
+        // want an entire domain that is very small.
+        ASSERT(not equal_within_roundoff(radius_one, 0.0),
+               "Cannot have zero radius_one");
+        ASSERT(radius_one > 0.0, "Cannot have negative radius_one");
+        return radius_one;
+      }()),
+      radius_two_([&radius_two]() {
```
geoffrey4444, line 81: Is there an easy way to fix these codecov failures? I'm fine if not, just want to be sure you've thought about it. Or is codecov just wrong? Seems like you would have to test these lines if you test the constructor at all

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+      }()),
+      z_plane_plus_one_(z_plane_plus_one),
+      z_plane_minus_one_(z_plane_minus_one),
+      z_plane_plus_two_(z_plane_plus_two),
+      z_plane_minus_two_(z_plane_minus_two) {
+  // Assumptions made in the map.  Some of these can be relaxed,
+  // as long as the unit test is changed to test them.
+  // The ASSERTS here match the ones in UniformCylindricalEndcap.
+  ASSERT(radius_one >= 0.08 * radius_two,
+         "Radius_one too small " << radius_one << " " << 0.08 * radius_two);
```
geoffrey4444: Maybe help out the poor future users and change to say "Radius_one = 0.1234 must be >= 0.08 * radius_two == 0.5678" instead of just printing the two things to compare with a space

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+  ASSERT(z_plane_plus_two == z_plane_plus_one or
+             z_plane_plus_two >= z_plane_plus_one + 0.03 * radius_two,
+         "z_plane_plus_two must be >= z_plane_plus_one "
+         "+ 0.03 * radius_two, or exactly equal to z_plane_plus_one, not "
+             << z_plane_plus_two << " " << z_plane_plus_one << " "
+             << z_plane_plus_one + 0.03 * radius_two);
+  ASSERT(z_plane_minus_two == z_plane_minus_one or
+             z_plane_minus_two <= z_plane_minus_one - 0.03 * radius_two,
+         "z_plane_minus_two must be >= z_plane_minus_one "
```
geoffrey4444, line 108: same comment: help out the poor users and make the error clearly say what z1 is, what z2 is, and what z1 + 0.03 * radius 2 is

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+         "The map has been tested only for the case when "
+         "sphere_one intersects the polar axis of sphere_two and is not "
+         "too far from the edge of sphere_two (or in the "
+         "case of equal plus or minus z_planes, when the two spheres have the "
+         "same x and y centers)."
+         " Radius_one = "
+             << radius_one << ", radius_two = " << radius_two
+             << ", dist_spheres = " << dist_spheres
+             << ", horizontal_dist_spheres = " << horizontal_dist_spheres);
+
```
geoffrey4444, line 230: Just so I understand, how sure are you none of these conditions will break in the BBH map we want to make with this?

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+  // function that is zero when lambda has the correct value, and the
+  // derivative of that function with respect to lambda.
+  const auto function_and_deriv_to_zero = [this, &target_coords](
+                                              const double lambda) {
+    const double one_plus_zbar_over_two =
+        (target_coords[2] + lambda * (z_plane_minus_one_ - z_plane_minus_two_) -
+         z_plane_minus_one_) /
+        ((1.0 - lambda) * (z_plane_plus_one_ - z_plane_minus_one_) +
+         lambda * (z_plane_plus_two_ - z_plane_minus_two_));
+    const double r_one_cos_theta_one =
```
geoffrey4444: There's some code duplication between `function_to_zero` and `function_and_deriv_to_zero`...is there a reason for this code duplication? Would it add to the cost to call function_to_zero to get the function part of `function_to_zero` somehow?

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+            abs(trial_lambda_min_increment) >
+                    1000.0 * std::numeric_limits<double>::epsilon()
+                ? trial_lambda_min_increment
+                : std::copysign(1000.0 * std::numeric_limits<double>::epsilon(),
+                                trial_lambda_min_increment);
+        const double new_lambda_min = lambda_min + lambda_min_increment;
+        const double function_at_new_lambda_min =
+            function_to_zero(new_lambda_min);
+        if (function_at_new_lambda_min * function_at_lambda_min > 0.0) {
+          ERROR(
```
geoffrey4444, line 506: Apparently need an error test to test this situation? Is that hard to do?

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+            abs(trial_lambda_max_increment) >
+                    1000.0 * std::numeric_limits<double>::epsilon()
+                ? trial_lambda_max_increment
+                : std::copysign(1000.0 * std::numeric_limits<double>::epsilon(),
+                                trial_lambda_max_increment);
+        const double new_lambda_max = lambda_max + lambda_max_increment;
+        const double function_at_new_lambda_max =
+            function_to_zero(new_lambda_max);
+        if (function_at_new_lambda_max * function_at_lambda_max > 0.0) {
+          ERROR(
```
geoffrey4444, line 560: Can this be tested with an error test without too much more work?

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+        }
+        // Now the root is bracketed between lambda_max and new_lambda_max,
+        // so replace lambda_max and lambda_min and then fall through to the
+        // root finder.
+        lambda_min = lambda_max;
+        function_at_lambda_min = function_at_lambda_max;
+        lambda_max = new_lambda_max;
+        function_at_lambda_max = function_at_new_lambda_max;
+      } else {
+        ERROR(
```
geoffrey4444, line 583: Add error test?

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+    constexpr double rel_tol = 1.e-15;
+
+    try {
+      lambda =
+          // NOLINTNEXTLINE(clang-analyzer-core)
+          RootFinder::toms748(function_to_zero, lambda_min, lambda_max,
+                              function_at_lambda_min, function_at_lambda_max,
+                              abs_tol, rel_tol);
+    } catch (std::exception&) {
+      ERROR("Cannot find root after bracketing: lambda_min="
```
geoffrey4444, line 610: Error test?

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,752 @@
...
+             (get<2, 2>(jac) - 1.0) * (z_plane_plus_two_ - z_plane_minus_two_));
+  return jac;
+}
+
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, 3, Frame::NoFrame>
+UniformCylindricalSide::inv_jacobian(
+    const std::array<T, 3>& source_coords) const {
+  return determinant_and_inverse(jacobian(source_coords)).second;
+}
```
geoffrey4444, line 729: I'm just wondering: after all the checking on different ways the inverse map might not be invertible, how come we never have to worry about the Jacobin matrix being invertible? Can you explain it like this: determinant can't be zero unless one mapped coord not affected by any unmapped coords, something clearly pathological that won't happen in real life?

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
```diff
@@ -0,0 +1,518 @@
...
+void test_uniform_cylindrical_side() {
+  INFO("UniformCylindricalSide");
+
+  // Set up random number generator
+  MAKE_GENERATOR(gen);
+  std::uniform_real_distribution<> unit_dis(0.0, 1.0);
+  std::uniform_real_distribution<> interval_dis(-1.0, 1.0);
+  std::uniform_real_distribution<> angle_dis(0.0, 2.0 * M_PI);
+
+  // Choose some random center for sphere_two
```
geoffrey4444, line 169: Seems like there's some code duplication between here and the previous function. Any reason not to factor to anonymous namespace the parts of these functions (setting up the random values etc.) that are the same for both?

geoffrey4444 (changes requested):
Really sorry how long it took me to review, @markscheel ! I have minor comments but then after you take a look at them I'm happy to approve

geoffrey4444:
Also notice clang-tidy is failing

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
markscheel: I am not sure at all.  The numbers in some of the conditions will almost certainly need to be adjusted in the PR that combines the maps into the BBH domain.  I didn't want to go ahead with the domain PR before this one (and the other one #4267 ) were merged, because I don't want to submit a 6000-line PR.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
markscheel: Jacobian is taken care of.  Think about the 1d case which is easier.  If you have a map `X(x)`, then invertibility means the inverse map `x(X)` is a function, i.e. it is single-valued.  But the single-valued condition is equivalent to `x(X)` is monotonically increasing or decreasing, which is the same thing as the inverse Jacobian `dx/dX` being always the same sign, i.e. `dx/dX` is never zero.  And that is the same condition that the Jacobian `dX/dx` is finite.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
markscheel: This one I think is a codecov bug.  I think I can fix it by telling codecov not to look a this line.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
markscheel: I told codecov to ignore this error and the others too.  Those are all cases that should never happen unless there is a bug, so I have no idea how to cause them to happen intentionally.

src/Domain/CoordinateMaps/UniformCylindricalSide.cpp
markscheel: I added a new struct that computes the common things.

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalSide.cpp
markscheel: I could refactor these a bit, but there is very little code that is the same.  Just setting up center and radius of sphere two, and declaring the distributions.

In my first version, this was one function with 'if' statements, but that was really hard to understand.

And I fear that these functions might diverge even more once the maps are combined into a domain and I need to change the allowed ranges of parameters again (see your question further above).

markscheel:
The clang-tidy errors are toms748, which I cannot change.

markscheel:
Made all the requested changes, except did not refactor small amount of (currently) duplicate set-random-number code in the test.   For the error cases that codecov missed, told codecov not to look at them (those cases are all things that shouldn't happen and are difficult to test because I don't know how to make them happen).

nilsdeppe (commented):
LGTM, I have no further comments

geoffrey4444 (dismissed):
LGTM, except clang-tidy is failing. I don't need to see this again to merge!

markscheel:
squashed.

nilsdeppe (approved):

PR #4248, opened 2022-09-24
fix a mistake in the CCE tutorial

Sizheng-Ma:
## Proposed changes

There is a mistake in the CCE tutorial. The first column in the CCE output is supposed to be the time rather than the real part of `(l=0,m=0)`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2 (approved):
Tested this out and it fixes the plot that is produced.

nilsdeppe (approved):

PR #4250, opened 2022-09-26
Fix a typo in the comment of GaugeUpdateInertialTimeDerivatives

Sizheng-Ma:
## Proposed changes

A comment in `GaugeUpdateInertialTimeDerivatives` is wrong.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/GaugeTransformBoundaryData.cpp
+++ b/src/Evolution/Systems/Cce/GaugeTransformBoundaryData.cpp
@@ -587,8 +587,7 @@ void GaugeUpdateInertialTimeDerivatives::apply(
       l_max, 1, make_not_null(&eth_x_inertial), make_not_null(&eth_y_inertial),
       make_not_null(&eth_z_inertial), x_inertial, y_inertial, z_inertial);
 
-  // Interpolate evolution_gauge_u_at_scri and omega to the partially flat
-  // Bondi-like coordinates
+  // Interpolate evolution_gauge_u_at_scri and omega to the Cauchy coordinates
   auto& original_u_at_scri =
       get(get<::Tags::SpinWeighted<::Tags::TempScalar<6, ComplexDataVector>,
                                    std::integral_constant<int, 1>>>(


src/Evolution/Systems/Cce/GaugeTransformBoundaryData.cpp
```diff
@@ -587,8 +587,7 @@ void GaugeUpdateInertialTimeDerivatives::apply(
       l_max, 1, make_not_null(&eth_x_inertial), make_not_null(&eth_y_inertial),
       make_not_null(&eth_z_inertial), x_inertial, y_inertial, z_inertial);
 
-  // Interpolate evolution_gauge_u_at_scri and omega to the partially flat
-  // Bondi-like coordinates
+  // Interpolate evolution_gauge_u_at_scri and omega to the Cauchy coordinates
```
nilsdeppe, line 590: Is there a way I can convince myself this is correct? As in, where should I look to figure out what coordinates this is in? Not that I don't believe you, but I probably reviewed this last time and didn't realize the mistake :)

Sizheng-Ma: The most straightforward way is to look at the template of `Spectral::Swsh::Tags::SwshInterpolator` in the `.hpp` file, which corresponds to the **source** frame. For `GaugeUpdateInertialTimeDerivatives`, the template is `Tags::PartiallyFlatAngularCoords`, meaning that the interpolator transforms spin weighted quantities from the partially flat coordinates to the Cauchy coordinates. In contrast, functions of `GaugeAdjustedBoundaryValue` do the thing the other way around: they take Cauchy worldtube data and interpolate them to the partially flat coordinates (the coordinates that are used to evolve the characteristic evolution). So in those functions, the template of `SwshInterpolator` is `Tags::CauchyAngularCoords`.

Another rule of thumb is that `xxxxGaugeC`, `xxxxGaugeD`, and `xxxxGaugeOmega` are defined in the corresponding `xxxx` frame. Here `xxxx`  can be either `Cauchy` or `PartiallyFlat`. In this case, we are interpolating `Tags::PartiallyFlatGaugeOmega`, implying the partially flat coordinates are the source frame.

nilsdeppe: Okay, that makes sense. Thank you!

nilsdeppe (approved):

PR #4254, opened 2022-09-27
Have SpECTRE control system act more like SpECs control system

knelli2:
## Proposed changes

Also write the horizon centers to disk every time the control system horizon finder is called because this is a helpful diagnostic.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/ControlSystem/ApparentHorizons/Measurements.hpp
```diff
@@ -69,7 +72,8 @@ struct BothHorizons : tt::ConformsTo<protocols::Measurement> {
       using horizon_find_failure_callback =
           intrp::callbacks::ErrorOnFailedApparentHorizon;
       using post_horizon_find_callbacks =
-          tmpl::list<control_system::RunCallbacks<FindHorizon, ControlSystems>>;
+          tmpl::list<control_system::RunCallbacks<FindHorizon, ControlSystems>,
+                     ::ah::callbacks::ObserveCenters<InterpolationTarget>>;
```
knelli2, line 76: Hi.

src/ControlSystem/Tags.hpp
```diff
@@ -96,6 +96,18 @@ struct WriteDataToDisk {
       "Whether control system data should be saved during an evolution."};
   using group = ControlSystemGroup;
 };
+
+/// \ingroup OptionTagsGroup
+/// \ingroup ControlSystemGroup
+/// Option tag that determines how many measurements will occur per control
+/// system update.
+struct MeasurementsPerUpdate {
+  using type = int;
```
wthrowe, line 113: Set a lower bound of 1?

src/ControlSystem/Trigger.hpp
```diff
@@ -148,15 +161,53 @@ class Trigger : public DenseTrigger {
...
+    // to the min measure time. We already returned above if it is infinity
+    if (min_fot_expr_time == std::numeric_limits<double>::infinity()) {
+      return calculated_next_measurement;
+    }
+
+    double next_measurement_time;
+    // Just take the average of the two things we are comparing. They'll always
+    // be super close
+    const double scale =
+        0.5 * (calculated_next_measurement + min_fot_expr_time);
```
wthrowe: Pass zero as the scale.  The scale is to correct for error caused by intermediate results canceling in the calculation.

src/ControlSystem/Trigger.hpp
```diff
@@ -148,15 +161,53 @@ class Trigger : public DenseTrigger {
...
+
+    double min_fot_expr_time = std::numeric_limits<double>::infinity();
+    for (const auto& [name, fot] : functions_of_time) {
+      (void)name;
+      min_fot_expr_time = std::min(min_fot_expr_time, fot->time_bounds()[1]);
+    }
+
+    // If the expiration time is infinity, don't bother checking if it's equal
+    // to the min measure time. We already returned above if it is infinity
+    if (min_fot_expr_time == std::numeric_limits<double>::infinity()) {
```
wthrowe: Is the false branch tested?  Codecov thinks it isn't.

src/ControlSystem/Trigger.hpp
```diff
@@ -148,15 +161,53 @@ class Trigger : public DenseTrigger {
...
+        });
+    if (min_measure_time == std::numeric_limits<double>::infinity()) {
+      return min_measure_time;
+    }
+    const double calculated_next_measurement = time + min_measure_time;
+
+    double min_fot_expr_time = std::numeric_limits<double>::infinity();
+    for (const auto& [name, fot] : functions_of_time) {
+      (void)name;
+      min_fot_expr_time = std::min(min_fot_expr_time, fot->time_bounds()[1]);
```
wthrowe: Are you sure that the expiration times are in a consistent state on all elements when they run this function?  It looks like this can run any time after the measurement timescales update, which I don't think is sequenced with respect to the main FoTs.

tests/Unit/Helpers/ControlSystem/SystemHelpers.hpp
```diff
@@ -557,13 +566,28 @@ struct SystemHelper {
...
+        (void)name;
+        tmp_dt =
+            std::min(tmp_dt, min(measurement_timescale->func(local_time)[0]));
+      }
+
+      return tmp_dt;
+    };
+    double dt = calculate_dt(time);
+
+    // using obs_writer = typename Metavars::observer_component;
```
wthrowe: Remove.

wthrowe (changes requested):

src/ControlSystem/Trigger.hpp
knelli2: Good catch. I think my new implementation catches this, but let me know if you'd like me to change it. It isn't the prettiest solution.

src/ControlSystem/Trigger.hpp
```diff
@@ -143,6 +143,55 @@ class Trigger : public DenseTrigger {
...
+      // the functions of time when calculating the next measurement. The reason
+      // we have to do this complicated checking is because the functions of
+      // time are valid at the expiration times, so we can't just use a
+      // domain::functions_of_time_are_ready() call. That would just return true
+      // regardless of whether the functions of time have been updated. So we
+      // must check that the minimum expiration time is after the current time.
+      // If it is equal to the current time, then at least one of the functions
+      // of time hasn't been updated yet and we must wait until they are all
+      // updated so we set a callback.
+      if (must_check_fot_validity_) {
```
wthrowe: Why is this check done conditionally?  It looks like `next_measurement` always needs it to be true, and it doesn't look terribly expensive.

src/ControlSystem/Trigger.hpp
wthrowe: Thinking about this a bit more, I think this can fail in the other direction as well:  If some element that doesn't need to supply data is more than one measurement behind the elements actually used in a measurement, it could receive the FoT update before scheduling the trigger at the update time.

Is there some other way we could obtain the update time?  The expiration time was really not intended to be used as data.

wthrowe (changes requested):

src/ControlSystem/Trigger.hpp
knelli2: I believe these concerns are now moot because the functions of time aren't updated at the expiration time anymore.

knelli2 (commented):
After discussing with people on 10/18/22, it was decided that instead of updating the functions of time at the expiration time, they should be updated one measurement before the expiration time. The new fixups implement these changes.

src/ControlSystem/ExpirationTimes.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * \f$\tau_\mathrm{exp} = \alpha_\mathrm{update} \tau_\mathrm{damp}\f$ where
+ * \f$\alpha_\mathrm{update}\f$ is the update fraction supplied as input to the
+ * Controller and \f$\tau_\mathrm{damp}\f$ is/are the damping timescales
+ * supplied from the TimescaleTuner (\f$\tau_\mathrm{damp}\f$ is a DataVector
+ * with as many components as the corresponding function of time, thus
+ * \f$\tau_\mathrm{exp}\f$ will also be a DataVector of the same length).
+ *
+ * To protect against bad inputs, if the initial expiration time that is
+ * calculated is smaller than the initial time step, then the expiration time is
+ * simply set to the initial time step. However, the MeasurementTimescales have
```
wthrowe: You may need to revert d8ac31aa3f57aa9ae9ba18262e3b17ebb16fe557 to avoid a potential problem with self-start evaluation times.  It allows self-start to take initialization steps at times after the end of the first real step.  If you do, I have no objection to that happening.

src/ControlSystem/Trigger.hpp
wthrowe: I don't think that's relevant?  The numbers change a bit, but I don't see how it fixes the race.

src/ControlSystem/Trigger.hpp
```diff
@@ -148,15 +161,49 @@ class Trigger : public DenseTrigger {
...
+
+    double min_fot_expr_time = std::numeric_limits<double>::infinity();
+    for (const auto& [name, fot] : functions_of_time) {
+      (void)name;
+      min_fot_expr_time = std::min(min_fot_expr_time, fot->time_bounds()[1]);
+    }
+
+    // If the expiration time is infinity, don't bother checking if it's equal
+    // to the min measure time. We already returned above if it is infinity
+    if (min_fot_expr_time == std::numeric_limits<double>::infinity()) {
```
wthrowe: Is there a case where having a finite measurement time makes sense when the FoTs never expire?

wthrowe (changes requested):

src/ControlSystem/ExpirationTimes.hpp
knelli2: I've been using this for all the runs I've been doing since this was originally merged over a year ago, and it hasn't been an issue so far so I'll leave it for now. If it becomes an issue, we can deal with it then.

src/ControlSystem/Trigger.hpp
knelli2: I can't think of a regular use case. I could imagine some convoluted debugging scenario where we don't want to update the functions of time but still do the measurements? But that seems so specific that it should be done on a branch and not have the functionality on develop.

I'll remove this part.

src/ControlSystem/Trigger.hpp
knelli2: Hmmm you're right. We have the `control_system::Tags::MeasurementsPerUpdate` in the cache. What if we store which measurement we are on in the trigger, and then when we are at the update time (which is `current_measurement == measurements per update`), we store the expiration time in the trigger and use that expiration time. This way there's no race condition because nothing has been updated yet because we have to actually do the measurement the functions of time can be updated. There'd have to be a special case for t=0, but that's doable. What do you think @wthrowe?

wthrowe: Do we necessarily have to have done the measurement?  Someone has to have done the measurement, certainly, but a horizon-find might not wait for data from the wave zone.

What is the purpose of this roundoff check?  Can we just remove it?

knelli2: I don't think we can get rid of the roundoff check. We were having all these issues syncing up floating point numbers because the expiration time is calculated on the control system components while the next measurement time is calculated on the elements and those weren't lining up.

Given that, I just need an expiration time that's guaranteed to be valid and not subject to race conditions. This was the simplest way I could think of at the moment. If you have a better way, I'm open to suggestions, but I'd really like to get this in quickly so we can do BBH testing with it.

wthrowe: But why do they have to line up?

knelli2: Originally they had to because we updated at the expiration time. But now since we update before the expiration time, I guess they don't have to line up exactly. If the measurement time is less than the expiration time by roundoff (or equal to exactly), nothing happens because everything is valid. If the measurement time is greater than the expiration time by roundoff, then it will wait for the functions of time to update which is now guaranteed to happen because we updated the previous measurement.

I've tested a BBH where I got rid of the roundoff check and it's run fine so far up to 30M which is already a lot of updates.

knelli2:
@wthrowe All those fixups were getting kind of confusing so I squashed everything. Because we don't have to worry about roundoff for the expiration time in the Trigger anymore, there were basically no changes needed to the Trigger. One small thing about infinity but I factored that out into it's own commit.

I didn't change much else. The main things to look at are `ExpirationTimes.?pp`, `FunctionsOfTimeInitialize.hpp`, and `MeasurementTimescales.hpp` (and their corresponding tests). Basically everything else is the same as the last time you reviewed this PR.

src/ControlSystem/ExpirationTimes.hpp
```diff
@@ -0,0 +1,140 @@
...
+class Controller;
+class TimescaleTuner;
+/// \cond
+
+namespace control_system {
+/// \cond
+template <size_t DerivOrder>
+DataVector calculate_measurement_timescales(
+    const ::Controller<DerivOrder>& controller, const ::TimescaleTuner& tuner,
+    const int measurements_per_update);
```
wthrowe: Sorry if this was here before and I missed it, but if you're going to do it this way you should explicitly instantiate `calculate_measurement_timescales` so that this file doesn't depend on the details of implementations of functions elsewhere causing implicit instantiations.  Better would be to break the include loop.

wthrowe (changes requested):
Also rebase for a conflict.

knelli2:
@wthrowe Rebased to fix conflicts and just separated out the `calculate_measurement_timescales` function into a new file to avoid include loops.

wthrowe (dismissed):

src/ControlSystem/Trigger.hpp
```diff
@@ -143,20 +143,27 @@ class Trigger : public DenseTrigger {
...
-           tmpl::as_pack<ControlSystems>(
-               [&measurement_timescales, &time](auto... control_systems) {
-                 return std::min({min(
-                     measurement_timescales
-                         .at(tmpl::type_from<decltype(control_systems)>::name())
-                         ->func(time)[0])...});
-               });
+    const double min_measure_time = tmpl::as_pack<ControlSystems>(
+        [&measurement_timescales, &time](auto... control_systems) {
+          // LCOV_EXCL_START
```
nilsdeppe, line 154: Do you understand why you need to exclude all this from test coverage? It seems like this should be covered since it's the core part of this function

src/ControlSystem/CalculateMeasurementTimescales.hpp
```diff
@@ -0,0 +1,36 @@
...
+ * The update timescale is \f$\tau_\mathrm{update} = \alpha_\mathrm{update}
+ * \tau_\mathrm{damp}\f$ where \f$\tau_\mathrm{damp}\f$ is the damping timescale
+ * (from the TimescaleTuner) and \f$\alpha_\mathrm{update}\f$ is the update
+ * fraction (from the controller). For an Nth order control system, the averager
+ * requires at least N measurements in order to perform its finite
+ * differencing to calculate the derivatives of the control error. This implies
+ * that the largest the measurement timescale can be is \f$\tau_\mathrm{m} =
+ * \tau_\mathrm{update} / N\f$. To ensure that we have sufficient measurements,
+ * we calculate the measurement timescales as \f$\tau_\mathrm{m} =
+ * \tau_\mathrm{update} / (N+1)\f$
```
nilsdeppe: Could you add that the `N+1` factor is `measurements_per_update` and is set elsewhere? I kept looking for an `N+1` 🙃

src/ControlSystem/ExpirationTimes.hpp
```diff
@@ -0,0 +1,135 @@
...
+#include "DataStructures/DataVector.hpp"
+#include "Domain/Creators/DomainCreator.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace control_system {
+/*!
+ * \ingroup ControlSystemGroup
+ * \brief Calculate the next expiration time for the FunctionsOfTime.
+ *
+ * This is done by adding to the current time one old minimum measurement
```
nilsdeppe: Could you write a math equation here? The words are a bit, well, wordy 🙃

src/ControlSystem/ExpirationTimes.hpp
```diff
@@ -0,0 +1,135 @@
...
+ * update time and one old measurement after the update. This decision was made
+ * to minimize time spent waiting for the functions of time to be valid.
+ *
+ * Since functions of time are valid at their expiration time, we are actually
+ * able to do the next measurement if the expiration time is at that
+ * measurement. Thus we delay any potential waiting that may happen until the
+ * subsequent measurement (and by that time, most, if not all, functions of time
+ * should have been updated because an entire horizon find happened in the
+ * meantime). If the expiration time was earlier than the next measurement, we'd
+ * have to pause the Algorithm on the elements and wait until all the functions
```
nilsdeppe: maybe `elements` -> `DG elements`

src/ControlSystem/ExpirationTimes.hpp
```diff
@@ -0,0 +1,135 @@
...
+ * time. Based on how dense triggers are set up, which control_system::Trigger
+ * is a dense trigger, you calculate the next trigger (measurement) time at the
+ * current measurement time. However, at the function of time expiration time we
+ * need updated damping timescales from all control systems in order to
+ * calculate when the next measurement is going to be (and in turn, the next
+ * measurement expiration time). Thus, at the measurement that occurs at the
+ * function of time expiration time, our measurement timescales can't be valid
+ * and we must wait for updated ones. To achieve this, we set the measurement
+ * expiration time *before* the function of time expiration time, but *after*
+ * the previous measurement (the update measurement). The factor of one half is
```
nilsdeppe, line 80: Can you add an equation here? The 1/2 is in the code, but the comment doesn't really make that clear since the formula isn't explicitly given

src/ControlSystem/ExpirationTimes.hpp
```diff
@@ -0,0 +1,135 @@
...
+ * supplied from the TimescaleTuner (\f$\tau_\mathrm{damp}\f$ is a DataVector
+ * with as many components as the corresponding function of time, thus
+ * \f$\tau_\mathrm{exp}\f$ will also be a DataVector of the same length).
+ *
+ * If the control system isn't active then expiration time is
+ * `std::numeric_limits<double>::infinity()`.
+ *
+ * To protect against bad inputs, if the initial expiration time that is
+ * calculated is smaller than the initial time step, then the expiration time is
+ * simply set to the initial time step. However, the MeasurementTimescales have
```
nilsdeppe, line 109: Given this caveat and semi-silent error if both are invalid, would it be better to error on bad inputs? Or set the initial measurement time to 1/2 the initial time step?

src/ControlSystem/UpdateControlSystem.hpp
```diff
@@ -130,84 +118,96 @@ struct UpdateControlSystem {
...
-    // Begin step 4
     // Compute control error
     const DataVector Q =
         control_error(cache, time, function_of_time_name, data);
 
-    // Update the averager. We do this before we call controller.is_ready()
-    // because we still want the averager to be up to date even if we aren't
-    // updating at this time
+    // Update the averager. We do this for every measurement because we still
+    // want the averager to be up to date even if we aren't updating at this
```
nilsdeppe: [optional] `up-to-date`?

nilsdeppe (changes requested):
Overall LGTM, just some requests for more equations/dox :) Please rebase and squash immediately!

src/ControlSystem/Trigger.hpp
knelli2: No not really....I know for a fact the code is tested (the trigger won't work without it), but for some reason codecov says it isn't tested?? Maybe has something to do with the `tmpl::as_pack`? I just added the `LCOV*` to silence codecov's false warning.

nilsdeppe: Alright, sounds good :)

nilsdeppe (approved):

PR #4262, opened 2022-09-28
Updating Cce input file to have higher LMax and finer time sampling

keefemitman:
## Proposed changes

Update Cce input file to have higher LMax and finer time sampling after performing BBH convergence tests

### DIFF
--- a/tests/InputFiles/Cce/CharacteristicExtract.yaml
+++ b/tests/InputFiles/Cce/CharacteristicExtract.yaml
@@ -12,10 +12,10 @@ ResourceInfo:
   Singletons:
     CharacteristicEvolution:
       Proc: Auto
-      Exclusive: false
+      Exclusive: False
     H5WorldtubeBoundary:
       Proc: Auto
-      Exclusive: false
+      Exclusive: False
 
 Observers:
   VolumeFileName: "CharacteristicExtractVolume"
@@ -27,7 +27,7 @@ Cce:
       AdamsBashforthN:
         Order: 3
     StepChoosers:
-      - Constant: 1.0
+      - Constant: 0.5
       - Increase:
           Factor: 2
       - ErrorControl(SwshVars):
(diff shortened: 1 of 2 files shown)


tests/InputFiles/Cce/CharacteristicExtract.yaml
```diff
@@ -63,7 +63,7 @@ Cce:
   StartTime: 0.0
   EndTime: 1000.0
   BoundaryDataFilename: CceR0257.h5
-  H5IsBondiData: false
+  H5IsBondiData: False
```
knelli2, line 66: Why this change?

knelli2 (changes requested):
Can you also update the docs in `CCE.md` to reflect these changes? Also maybe give an explanation that these numbers were chosen based on BBH convergence tests?

tests/InputFiles/Cce/CharacteristicExtract.yaml
keefemitman: Just to be consistent with the other text in the file (not sure why it wasn't that way in the first place)

keefemitman:
> Can you also update the docs in `CCE.md` to reflect these changes? Also maybe give an explanation that these numbers were chosen based on BBH convergence tests?

Will do!

tests/InputFiles/Cce/CharacteristicExtract.yaml
knelli2: Ah ok. Change the ones at the top of the file in the `ResourceInfo` block as well just so the whole file is consistent.

keefemitman: Done!

knelli2 (commented):
Looks good. Can you squash these into one commit and keep the commit name 60 characters or shorter?

knelli2 (approved):

nilsdeppe (approved):

PR #4266, opened 2022-10-01
Fix bug causing events to be improperly rerun

wthrowe:
Introduced in 7ce002f346c9613309a743e87b83a8feb41d7ba0.  The flag that events had been run was cleared after running the events instead of after rescheduling, so if rescheduling failed the events could be rerun on a subsequent attempt.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/EventsAndDenseTriggers/EventsAndDenseTriggers.hpp
+++ b/src/Evolution/EventsAndDenseTriggers/EventsAndDenseTriggers.hpp
@@ -238,7 +238,9 @@ void EventsAndDenseTriggers::run_events(
                 std::numeric_limits<double>::signaling_NaN();
           });
     }
-    trigger_entry.is_triggered.reset();
+    // Mark this trigger as handled so we will not reprocess it if
+    // this method or is_ready is called again.
+    trigger_entry.is_triggered = false;
   }
 }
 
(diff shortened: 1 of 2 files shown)


nilsdeppe (approved):

knelli2:
I've tried this out and it does not fix the issue we've been having with deadlocks. I believe these changes are still necessary, but they don't completely solve the issue. Please hold off on merging this until we investigate further.

nilsdeppe:
We'd like to make a code release for Oct. and if there's currently a bug in develop that we can fix with an open PR I'd like to fix it... Fixing the global time stepping BBH evolutions could take a long time still so it seems like bad practice to have a bug in develop that we have a fix for but not merge the fix because there's a different bug...

knelli2 (approved):
@nilsdeppe I wasn't convinced before that this PR fixed all the issues on develop which is why I said to hold off on merging. After discussing with @wthrowe and @geoffrey4444, we've now determined that this PR is doing the correct thing and the issues with GTS BBH runs are unrelated to this PR. Good to merge!

nilsdeppe:
Okay, great! Thank you for discussing and figuring that out :)

PR #4267, opened 2022-10-04
Add UniformCylindricalFlatEndcap.

markscheel:
## Proposed changes

Another map necessary for a binary compact object domain decomposition using cylinders.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
+++ b/src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
@@ -1,0 +1,733 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp"
+
+#include <boost/math/special_functions/sign.hpp>
+#include <cmath>
+#include <limits>
+#include <optional>
+#include <pup.h>
+#include <sstream>
+#include <utility>
+
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/CoordinateMaps/CylindricalEndcapHelpers.hpp"
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "Parallel/PupStlCpp11.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/DereferenceWrapper.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+
+namespace domain::CoordinateMaps {
+
+namespace {
+
+// The function that the inverse map needs to root-find to find rhobar.
+double this_function_is_zero_for_correct_rhobar(
+    const double rhobar, const std::array<double, 3>& center_one,
+    const std::array<double, 3>& center_two, double radius_one,
+    const double radius_two, const double theta_max_one,
+    const std::array<double, 3>& target_coords) {
+  const double r_one_cos_theta_one = radius_one * cos(rhobar * theta_max_one);
+  const double lambda =
+      (target_coords[2] - center_one[2] - r_one_cos_theta_one) /
+      (center_two[2] - center_one[2] - r_one_cos_theta_one);
+  return square(target_coords[0] - center_one[0] -
+                lambda * (center_two[0] - center_one[0])) +
+         square(target_coords[1] - center_one[1] -
+                lambda * (center_two[1] - center_one[1])) -
+         square((1.0 - lambda) * radius_one * sin(rhobar * theta_max_one) +
+                lambda * radius_two * rhobar);
+}
+
+// min and max values of rhobar in the inverse function.
+std::array<double, 2> rhobar_min_max(
+    const std::array<double, 3>& center_one, const double radius_one,
+    const double theta_max_one, const std::array<double, 3>& target_coords) {
+  // Choose the minimum value of rhobar so that lambda >=0, where
+  // lambda is the quantity inside function_to_zero.  Note that the
+  // denominator of lambda inside that function is always positive.
+  const double rhobar_min =
+      target_coords[2] - center_one[2] >= radius_one
+          ? 0.0
+          : acos((target_coords[2] - center_one[2]) / radius_one) /
+                theta_max_one;
+  return {rhobar_min, 1.0};
+}
+
+// Returns whether the map is invertible if the target is on sphere_one.
+bool is_invertible_for_target_on_sphere_one(
+    const std::array<double, 3>& center_one,
+    const std::array<double, 3>& center_two, const double radius_one,
+    const double radius_two, const double theta_max_one,
+    const std::array<double, 3>& target_coords) {
+  const auto [rhobar_min, rhobar_max] =
+      rhobar_min_max(center_one, radius_one, theta_max_one, target_coords);
+  if (equal_within_roundoff(rhobar_min, rhobar_max)) {
+    // Special case where rhobar_min == rhobar_max to roundoff.  This
+    // case occurs when the target point (assumed to be on sphere_one)
+    // is at rhobar=1.  In this case there is nothing to test because
+    // there is only one possible value of rhobar, so there cannot be
+    // two roots for two different values of rhobar.
+    return true;
+  }
+  const size_t num_pts = 1000;
+  const double drhobar = (rhobar_max - rhobar_min) / (num_pts - 1.0);
+  // Because the target point is on sphere_one, the only root should
+  // be at rhobar = rhobar_min, where lambda=0.  The actual value of
+  // the function at rhobar_min may be either sign because of
+  // roundoff.  So ignore that point, and start at the next point.
+  // Keep track of the sign, and look for sign changes as we progress
+  // from point to point.
+  std::optional<double> sign;
+  for (size_t i = 1; i < num_pts; ++i) {
+    const double rhobar = rhobar_min + i * drhobar;
+    const double func = this_function_is_zero_for_correct_rhobar(
+        rhobar, center_one, center_two, radius_one, radius_two, theta_max_one,
+        target_coords);
+    if (not sign.has_value()) {
+      sign = func > 0.0 ? 1.0 : -1.0;
+    } else if (func * sign.value() <= 0.0) {
+      // There is another root, so function is not invertible.
+      //
+      // Note that the code should rarely get here if the map
+      // parameters are in range, so it is difficult to add an error
+      // test because I don't know how to make it get here on purpose;
+      // therefore we turn off codecov.
+      return false; // LCOV_EXCL_LINE
+    }
+  }
+  return true;  // No roots found.
+}
+}  // namespace
+
+bool is_uniform_cylindrical_flat_endcap_invertible_on_sphere_one(
+    const std::array<double, 3>& center_one,
+    const std::array<double, 3>& center_two, double radius_one,
+    double radius_two, double theta_max_one) {
+  if (center_one[1] == center_two[1] and center_one[0] == center_two[0]) {
+    return true;
+  }
+  // Choose the worst direction, which is the direction of x-y-plane
+  // projection of the difference between center_one and center_two.
+  const double phi =
+      atan2(center_one[1] - center_two[1], center_one[0] - center_two[0]);
+  const size_t num_pts = 100;
+  for (size_t i = 0; i < num_pts; ++i) {
+    const double theta_one = theta_max_one * (i / (num_pts - 1.0));
+    if (not is_invertible_for_target_on_sphere_one(
+            center_one, center_two, radius_one, radius_two, theta_max_one,
+            {{center_one[0] + radius_one * sin(theta_one) * cos(phi),
+              center_one[1] + radius_one * sin(theta_one) * sin(phi),
+              center_one[2] + radius_one * cos(theta_one)}})) {
+      // Note that the code should rarely get here if the map
+      // parameters are in range, so it is difficult to add an error
+      // test because I don't know how to make it get here on purpose;
+      // therefore we turn off codecov.
+      return false; // LCOV_EXCL_LINE
+    }
+  }
+  return true;
+}
+
+UniformCylindricalFlatEndcap::UniformCylindricalFlatEndcap(
+    const std::array<double, 3>& center_one,
+    const std::array<double, 3>& center_two, double radius_one,
+    double radius_two, double z_plane_one)
+    : center_one_(center_one),
+      center_two_(center_two),
+      radius_one_([&radius_one]() {
+        // The equal_within_roundoff below has an implicit scale of 1,
+        // so the ASSERT may trigger in the case where we really
+        // want an entire domain that is very small.
+        ASSERT(not equal_within_roundoff(radius_one, 0.0),
+               "Cannot have zero radius_one");
+        ASSERT(radius_one > 0.0, "Cannot have negative radius_one");
+        return radius_one;
+        // codecov doesn't understand lambdas in constructors for some
+        // reason, and says the next line is not covered even though it is.
+      }()), // LCOV_EXCL_LINE
+      radius_two_([&radius_two]() {
+        // The equal_within_roundoff below has an implicit scale of 1,
+        // so the ASSERT may trigger in the case where we really
+        // want an entire domain that is very small.
+        ASSERT(not equal_within_roundoff(radius_two, 0.0),
+               "Cannot have zero radius_two");
+        ASSERT(radius_two > 0.0, "Cannot have negative radius_two");
+        return radius_two;
+        // codecov doesn't understand lambdas in constructors for some
+        // reason, and says the next line is not covered even though it is.
+      }()), // LCOV_EXCL_LINE
+      z_plane_one_(z_plane_one),
+      theta_max_one_([&center_one, &radius_one, &z_plane_one]() {
+        const double cos_theta_max = (z_plane_one - center_one[2]) / radius_one;
+        ASSERT(abs(cos_theta_max) < 1.0,
+               "Plane one must intersect sphere_one, and at more than one "
+               "point. You probably specified a bad value of z_plane_one, "
+               "radius_one, or the z component of center_one. "
+               "Here z_plane_one="
+                   << z_plane_one << ", radius_one = " << radius_one
+                   << ", center_one=" << center_one
+                   << ", and cos_theta_max = " << cos_theta_max);
+        return acos(cos_theta_max);
+      }()) {
+  // The code below defines several variables that are used only in ASSERTs.
+  // We put that code in a #ifdef SPECTRE_DEBUG to avoid clang-tidy complaining
+  // about unused variables in release mode.
+#ifdef SPECTRE_DEBUG
+
+  // For some reason, codecov thinks that the following lambda never
+  // gets called, even though it is in all of the ASSERT messages below.
+  // LCOV_EXCL_START
+  const auto param_string = [this]() -> std::string {
+    std::ostringstream buffer;
+    buffer << "\nParameters to UniformCylindricalFlatEndcap:\nradius_one="
+           << radius_one_ << "\nradius_two=" << radius_two_
+           << "\ncenter_one=" << center_one_ << "\ncenter_two=" << center_two_
+           << "\nz_plane_one=" << z_plane_one_;
+    return buffer.str();
+  };
+  // LCOV_EXCL_STOP
+
+  // Assumptions made in the map.  The exact numbers for all of these
+  // can be changed, as long as the unit test is changed to test them.
+  ASSERT(center_two[2] >= center_one[2] + 1.05 * radius_one,
+         "center_two[2] must be >= center_one[2] + 1.05 * radius_one, not "
+             << center_two[2] << " " << center_one[2] + 1.05 * radius_one
+             << param_string());
+  ASSERT(center_two[2] <= center_one[2] + 5.0 * radius_one,
+         "center_two[2] must be <= center_one[2] + 5.0 * radius_one, not "
+             << center_two[2] << " " << center_one[2] + 5.0 * radius_one
+             << param_string());
+  ASSERT(radius_two >= 0.1 * radius_one * sin(theta_max_one_),
+         "radius_two is too small: " << radius_two << " "
+                                     << 0.1 * radius_one * sin(theta_max_one_)
+                                     << param_string());
+  ASSERT(theta_max_one_ < M_PI * 0.35,
+         "z_plane_one is too close to the center of sphere_one: theta/pi = "
+             << theta_max_one_ / M_PI << param_string());
+  ASSERT(theta_max_one_ > M_PI * 0.075,
+         "z_plane_one is too far from the center of sphere_one: theta/pi = "
+             << theta_max_one_ / M_PI << param_string());
+  ASSERT(
+      is_uniform_cylindrical_flat_endcap_invertible_on_sphere_one(
+          center_one_, center_two_, radius_one_, radius_two_, theta_max_one_),
+      "The map is not invertible at at least one point on sphere_one."
+      " center_one = "
+          << center_one_ << " center_two = " << center_two_
+          << " radius_one = " << radius_one_ << " radius_two = " << radius_two_
+          << " theta_max_one = " << theta_max_one_ << param_string());
+
+  const double horizontal_dist_spheres =
+      sqrt(square(center_one[0] - center_two[0]) +
+           square(center_one[1] - center_two[1]));
+
+  ASSERT(horizontal_dist_spheres <= radius_one_ * sin(theta_max_one_),
+         "Horizontal distance between centers is too large: "
+             << horizontal_dist_spheres << " "
+             << radius_one_ * sin(theta_max_one_) << param_string());
+
+  // max_horizontal_dist_between_circles can be either sign;
+  // therefore alpha can be in either the first or second quadrant.
+  const double max_horizontal_dist_between_circles =
+      horizontal_dist_spheres + radius_one_ * sin(theta_max_one_) - radius_two_;
+
+  const double alpha =
+      atan2(center_two[2] - z_plane_one, max_horizontal_dist_between_circles);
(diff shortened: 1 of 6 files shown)


markscheel:
Fixed a few codecov things.

The clang-tidy errors are apparently all toms748.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ *
+ * UniformCylindricalFlatEndcap maps a 3D unit right cylinder (with coordinates
+ * \f$(\bar{x},\bar{y},\bar{z})\f$ such that \f$-1\leq\bar{z}\leq 1\f$
+ * and \f$\bar{x}^2+\bar{y}^2 \leq 1\f$) to the shaded area
+ * in the figure above (with coordinates \f$(x,y,z)\f$).  The "bottom"
+ * of the cylinder \f$\bar{z}=-1\f$ is mapped to the portion of sphere
+ * 1 that has \f$z \geq z_{\mathrm{P}1}\f$, and on this portion of the
+ * sphere the angular coordinate \f$\theta_1 = \acos((z-C_1^2)/R_1)\f$
+ * is uniform in \f$\bar{\rho} = \sqrt{\bar{x}^2+\bar{y}^2}\f$ and the angular
+ * coordinate \f$\phi_1 = \atan((y-C_1^1)/(x-C_1^0))\f$ is the same as
```
knelli2: Same comment as the `UniformCylindricalSide` PR about using $x, y, z$ for the labels: $C_1^1 \rightarrow C_1^y$.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ *
+ * UniformCylindricalFlatEndcap provides the following functions:
+ *
+ * ### operator()
+ *
+ * `operator()` maps \f$(\bar{x},\bar{y},\bar{z})\f$ to \f$(x,y,z)\f$
+ * according to
+ *
+ * \f{align}
+ * x^0 &= C_1^0+\lambda(C_2^0-C_1^0) +
```
knelli2: Same here: $x^0, x^1, x^2 \rightarrow x, y, z$

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ * Note that by differentiating Eqs. (\f$\ref{eq:defQ}\f$) and
+ * (\f$\ref{eq:lambda_from_rho}\f$), one obtains
+ *
+ * \f{align}
+ * Q'(\bar{\rho}) =& -2 \frac{d\lambda}{d\bar{\rho}}\left[
+ *      \left(x^0-C_1^0-\lambda(C_2^0-C_1^0)\right)(C_2^0-C_1^0)+
+ *      \left(x^1-C_1^1-\lambda(C_2^1-C_1^1)\right)(C_2^1-C_1^1)
+ *      \right]\nonumber \\
+ *     &
+ *    -2 \left((1-\lambda)R_1\sin\theta_1 +
```
knelli2, line 204: I think this should be a `+2` instead of a `-2`.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ * \f$\bar{\rho}=0\f$ or \f$\bar{\rho}=1\f$.  When one of these cases is
+ * recognized, the root is returned as either \f$\bar{\rho}=0\f$ or
+ * \f$\bar{\rho}=1\f$.
+ *
+ * #### Quick rejection of points out of range of the map.
+ *
+ * It is expected that `inverse()` will often be passed points
+ * \f$(x,y,z)\f$ that are out of the range of the map; in this case
+ * `inverse()` returns a default-constructed
+ * `std::optional<std::array<double, 3>>`. To avoid the difficulty and
```
knelli2: `std::nullopt` here and below

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ * satisfy these conditions.
+ *
+ * Likewise, the inverse map can immediately reject any point with
+ * \f$z < z_{\mathrm{P}1}\f$.
+ *
+ * Finally, consider the circle \f$S_1\f$ defining the intersection of sphere 1
+ * and the plane \f$z = z_{\mathrm{P}1}\f$; this circle has radius
+ * \f$r_1 = R_1 \sin\theta_{1\mathrm{max}}\f$.
+ * Now consider the cone that passes through both \f$S_1\f$ and
+ * the circle \f$S_3\f$ bounding the upper disk.
```
knelli2: Doesn't $S_2$ bound the uppder disk with radius $R_2$ and center $C_2$? Or is $S_3$ a different circle?

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ * connects the same point on the circle \f$S_1\f$ with the
+ * corresponding point on the circle \f$S_2\f$ (the circle bounding
+ * the disk with center \f$C_2\f$ and radius \f$R_2\f$).
+ * Now consider the angle between \f$L\f$
+ * and \f$L'\f$, as measured from the interior of sphere 1, and Let
+ * \f$\alpha\f$ be the minimum value of this angle over the circle.
+ * \f$\alpha\f$ is shown in the figure above. If
+ * \f$\alpha < \theta_{1 \mathrm{max}}\f$, then the line segment \f$L'\f$
+ * intersects the mapped portion of sphere 1 twice, so the map is
+ * multiply valued. Therefore we demand that the map parameters are such that
```
knelli2: `multi-valued`

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ * have found empirically that if \f$Q(\bar{\rho})\f$ from
+ * Eq. (\f$\ref{eq:defQ}\f$) has only a single root between
+ * \f$\bar{\rho}_{\mathrm{min}}\f$ and \f$\bar{\rho}_{\mathrm{max}}\f$
+ * for all points \f$(x,y,z)\f$ on the surface of sphere 1 with
+ * \f$z\geq z_{\mathrm{P}1}\f$ and with \f$(x-C_1^0)/(y-C_1^1) =
+ * (C_1^0-C_2^0)/(C_1^1-C_2^1)\f$, then the map is single-valued
+ * everywhere.  We cannot numerically check every point in this
+ * one-parameter family of points, but we demand that this condition
+ * is satisfied for a reasonably large number of points in this
+ * family.  This check is not very expensive since it is done only
```
knelli2: Resonably large being 100? 1k? 10k? `std::numeic_limits<size_t>::max()`?

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ * once, in the constructor.
+ *
+ */
```
knelli2, line 458: Mention somewhere that this map can only be used in 3D.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ * ### Restrictions on map parameters
+ *
+ * We demand that \f$C^2_1 + 1.05 R_1 \leq C^2_2 \leq C^2_1 + 5 R_1\f$.
+ * It is possible to construct a valid
+ * map without this assumption, but the assumption simplifies the
+ * code, and the expected use cases obey this restriction.
+ *
+ * We also demand that the z plane in the above figure lies
+ * above the center of the sphere and is not too close to the center
+ * or edge of the sphere; specificially, we demand that
```
knelli2: `specifically`

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,692 @@
...
+  }
+  const size_t num_pts = 1000;
+  const double drhobar = (rhobar_max - rhobar_min) / (num_pts - 1.0);
+  // Because the target point is on sphere_one, the only root should
+  // be at rhobar = rhobar_min, where lambda=0.  The actual value of
+  // the function at rhobar_min may be either sign because of
+  // roundoff.  So ignore that point, and start at the next point.
+  // Keep track of the sign, and look for sign changes as we progress
+  // from point to point.  zero is a sentinel value.
+  double sign = 0.0;
```
knelli2: Maybe `std::optional<double>` would be better here? Just to avoid confusion with over-assigning meaning to `0.0`.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,692 @@
...
+        ASSERT(radius_two > 0.0, "Cannot have negative radius_two");
+        return radius_two;
+      }()),
+      z_plane_one_(z_plane_one),
+      theta_max_one_([&center_one, &radius_one, &z_plane_one]() {
+        const double cos_theta_max = (z_plane_one - center_one[2]) / radius_one;
+        ASSERT(abs(cos_theta_max) < 1.0,
+               "Plane one must intersect sphere_one, and at more than one "
+               "point. You probably specified a bad value of z_plane_one, "
+               "radius_one, or the z component of center_one.");
```
knelli2: List the values that were given?

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,692 @@
...
+        return acos(cos_theta_max);
+      }()) {
+  // Assumptions made in the map.  Some of these can be relaxed,
+  // as long as the unit test is changed to test them.
+  ASSERT(center_two[2] >= center_one[2] + 1.05 * radius_one,
+         "center_two[2] must be >= center_one[2] + 1.02 * radius_one, not "
+             << center_two[2] << " " << center_one[2] + 1.02 * radius_one);
```
knelli2: `1.02` -> `1.05`

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,692 @@
...
+  ASSERT(center_two[2] <= center_one[2] + 5.0 * radius_one,
+         "center_two[2] must be <= center_one[2] + 5.0 * radius_one, not "
+             << center_two[2] << " " << center_one[2] + 5.0 * radius_one);
+  ASSERT(radius_two >= 0.1 * radius_one * sin(theta_max_one_),
+         "radius_two is too small: " << radius_two << " "
+                                     << 0.1 * radius_one * sin(theta_max_one_));
```
knelli2: I don't know if this condition is listen in the docs?

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
```diff
@@ -0,0 +1,526 @@
...
+ * (1-\bar{z}) R_1 S(\bar{\rho},\theta_{1 \mathrm{max}}) +
+ * (1+\bar{z}) R_2
+ *     \right), \label{eq:x0alt} \\
+ * x^1 &= \frac{1}{2}\left((1-\bar{z})C_1^1 + (1+\bar{z})C_2^1\right) +
+ *        \frac{\bar{y}}{2}\left(
+ * (1-\bar{z})R_1 S(\bar{\rho},\theta_{1 \mathrm{max}}) +
+ * (1+\bar{z})R_2
+ *     \right), \label{eq:x1alt} \\
+ * x^2 &= \frac{1}{2}\left((1-\bar{z})C_1^2 + (1+\bar{z})C_2^2\right) +
+ * (1-\bar{z})R_1 \cos\theta_1, \label{eq:x2alt} \\
```
knelli2: Factor of 0.5 on the last term?

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,692 @@
...
+
+std::optional<std::array<double, 3>> UniformCylindricalFlatEndcap::inverse(
+    const std::array<double, 3>& target_coords) const {
+  // First do some easy checks that target_coords is in the range
+  // of the map.  Need to accept points that are out of range by roundoff.
+
+  // check z >= z_plane_one_
+  if (target_coords[2] < z_plane_one_ and
+      not equal_within_roundoff(target_coords[2], z_plane_one_)) {
+    return {};
```
knelli2: `return std::nullopt;`

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,692 @@
...
+          function_to_zero(new_rhobar_min);
+      if (function_at_new_rhobar_min * function_at_rhobar_min > 0.0) {
+        ERROR(
+            "Cannot find bracket after trying to adjust bracketing "
+            "rhobar_min due "
+            "to roundoff : "
+            << rhobar_min << " " << function_at_rhobar_min << " " << rhobar_max
+            << " " << function_at_rhobar_max << " " << new_rhobar_min << " "
+            << function_at_new_rhobar_min << " " << deriv_function_at_rhobar_min
+            << " " << new_rhobar_min - rhobar_min << "\n");
```
knelli2: Label the values in the error messages, here and below.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,692 @@
...
+  // memory allocations.
+  get<2, 2>(jac) = sqrt(square(xbar) + square(ybar));
+  get<2, 0>(jac) = 0.5 * radius_one_ *
+                   cylindrical_endcap_helpers::sin_ax_over_x(get<2, 2>(jac),
+                                                             theta_max_one_) *
+                   (1.0 - zbar);
+  get<1, 1>(jac) = get<2, 0>(jac) + 0.5 * radius_two_ * (1.0 + zbar);
+  get<2, 1>(jac) = theta_max_one_ * get<2, 0>(jac);
+  get<1, 2>(jac) =
+      0.5 * radius_two_ - 0.5 * radius_one_ *
```
knelli2: Factor out 0.5 so you only multiply it once? I don't actually know if it's more efficient or not.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,692 @@
...
+  p | center_two_;
+  p | radius_one_;
+  p | radius_two_;
+  p | z_plane_one_;
+  p | theta_max_one_;
+}
+
+bool operator==(const UniformCylindricalFlatEndcap& lhs,
+                const UniformCylindricalFlatEndcap& rhs) {
+  // don't need to compare theta_max_one_ or theta_max_two_
```
knelli2: We don't have a `theta_max_two_` in this class

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,167 @@
...
+  const CoordinateMaps::UniformCylindricalFlatEndcap map(
+      center_one, center_two, radius_one, radius_two, z_plane_one);
+  test_suite_for_map_on_cylinder(map, 0.0, 1.0, true);
+
+  // The following are tests that the inverse function correctly
+  // returns an invalid std::optional when called for a point that is
+  // outside the range of the map.
+
+  // Point with z less than z_plane_one.
+  CHECK_FALSE(map.inverse({{0.0,0.0,z_plane_one-1.0}}));
```
knelli2: `.has_value()` to be explicit

knelli2 (changes requested):
Full review.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
markscheel: Yeah it is S_2.  S_3 was a typo.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
markscheel: Added another '3D' in the \brief

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
markscheel: I still get -2.  This -2 comes from differentiating the last term in eq 10, which has a minus sign in front.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
markscheel: Yup, good catch.

markscheel:
Made changes.  In a separate commit, made another change to print more info in all of the ASSERTS in the constructor, which helps for debugging.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp
knelli2: You are correct. Differentiated it again and got -2 this time. Must have made a mistake the first time. Oops!

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,712 @@
...
+  p | radius_one_;
+  p | radius_two_;
+  p | z_plane_one_;
+  p | theta_max_one_;
+}
+
+bool operator==(const UniformCylindricalFlatEndcap& lhs,
+                const UniformCylindricalFlatEndcap& rhs) {
+  // don't need to compare theta_max_one_
+  // because is is uniquely determined from the other variables.
```
knelli2: `is is` -> `it is`

knelli2 (changes requested):
Looks good! Just one last change. Please squash everything together

knelli2:
Also @geoffrey4444 can you please take a look at this?

markscheel:
Made last small change and squashed.

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
knelli2: I still think this condition isn't covered in the docs.

However, upon closer inspection, I think this condition is similar to the $\alpha > 1.1 \theta_{1 \mathrm{max}}$ condition because it's a restriction on how small `radius_two` can be in relation to the size of the sphere. I would either add a comment here in the cpp file about it so we don't get confused later on why it's here, or just explicitly state the condition in the docs.

knelli2 (commented):
Ok everything else looks good. One more thing I overlooked when reviewing the fixups.

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,171 @@
...
+                   is_uniform_cylindrical_flat_endcap_invertible_on_sphere_one(
+                       center_one,
+                       {{center_one[0] + rho, center_one[1], z_center_two}},
+                       radius_one, radius_two, theta_max_one)
+               ? 1.0
+               : -1.0;
+  };
+  const double invertible_at_zero = is_invertible(0.0);
+  const double invertible_at_max = is_invertible(rho_max_so_far);
+  ASSERT(invertible_at_zero == 1.0, "Map should always be invertible at rho=0");
```
geoffrey4444: Why is this test an ASSERT instead of a CHECK? If you're running the test, don't you always want to check this?

tests/Unit/Domain/CoordinateMaps/Test_UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,171 @@
...
+                                     double theta_max_one) {
+  const auto is_invertible = [&](const double rho) {
+    // Direction shouldn't matter here, so just choose x-axis
+    return CoordinateMaps::
+                   is_uniform_cylindrical_flat_endcap_invertible_on_sphere_one(
+                       center_one,
+                       {{center_one[0] + rho, center_one[1], z_center_two}},
+                       radius_one, radius_two, theta_max_one)
+               ? 1.0
+               : -1.0;
```
geoffrey4444, line 33: Why are you using doubles instead of bools? I don't see how you could have roundoff error or anything with what's being done here, but it just seems kind of strange

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,712 @@
...
+  for (size_t i = 1; i < num_pts; ++i) {
+    const double rhobar = rhobar_min + i * drhobar;
+    const double func = this_function_is_zero_for_correct_rhobar(
+        rhobar, center_one, center_two, radius_one, radius_two, theta_max_one,
+        target_coords);
+    if (not sign.has_value()) {
+      sign = func > 0.0 ? 1.0 : -1.0;
+    } else if (func * sign.value() <= 0.0) {
+      // There is another root, so function is not invertible.
+      return false;
```
geoffrey4444: Do you understand why codecov is mad about this? Would it be hard to modify the unit test to handle this case? Same comment for all codecov failures

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,712 @@
...
+  const auto param_string = [this]() -> std::string {
+    std::ostringstream buffer;
+    buffer << "\nParameters to UniformCylindricalFlatEndcap:\nradius_one="
+           << radius_one_ << "\nradius_two=" << radius_two_
+           << "\ncenter_one=" << center_one_ << "\ncenter_two=" << center_two_
+           << "\nz_plane_one=" << z_plane_one_;
+    return buffer.str();
+  };
+
+  // Assumptions made in the map.  Some of these can be relaxed,
```
geoffrey4444: If any can't be relaxed, worth noting that in the comment here?

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,712 @@
...
+  const double r_minus_center_one_squared =
+      square(target_coords[0] - center_one_[0]) +
+      square(target_coords[1] - center_one_[1]) +
+      square(target_coords[2] - center_one_[2]);
+  if (r_minus_center_one_squared < square(radius_one_) and
+      not equal_within_roundoff(sqrt(r_minus_center_one_squared),
+                                radius_one_)) {
+    // sqrt above because we don't want to change the scale of
+    // equal_within_roundoff too much.
+    return {};
```
geoffrey4444: What's the difference between `return {}` and return `std::nullopt`, if any? If none, why not be consistent?

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -0,0 +1,712 @@
...
+        rhobar, center_one_, center_two_, radius_one_, radius_two_,
+        theta_max_one_, target_coords);
+  };
+
+  // To find rhobar we will do numerical root finding.
+  // function_and_deriv_to_zero returns a std::pair consisting of the
+  // function that is zero when rhobar has the correct value, and the
+  // derivative of that function with respect to rhobar.
+  const auto function_and_deriv_to_zero = [this, &target_coords](
+                                              const double rhobar) {
```
geoffrey4444: Why doesn't this function just call `this_function_is_zero_for_correct_rhobar`?

geoffrey4444 (changes requested):
Some minor comments for you to address and then I'm happy to approve. My apologies for taking so long to take a loot @markscheel !

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
markscheel: Oops. Those should have been std::nullopt.  Changed.

(there is no difference.  `return {}` and `return std::nullopt` both construct an invalid std::optional).

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
markscheel: I just removed this function because it is no longer needed as-is.  Only a function that computes the deriv is needed, which I added instead of this function.

markscheel:
I think I have addressed everything asked for.

geoffrey4444:
LGTM! I don't need to see this again before merging, but I'm still curious about the one place where you use 1.0 and -1.0 instead of a bool

geoffrey4444 (dismissed):

nilsdeppe (commented):
I have no additional comments

markscheel:
Squashed.

I added a comment to the code about 1.0 and -1.0 vs bool.   That function is used in a root finder that expects doubles, so it cannot be a bool.  (The root finder finds the value of rho where the function value changes from 1.0 to -1.0, i.e. where the function fails to be invertible).

nilsdeppe (dismissed):

knelli2:
@markscheel Needs a rebase now that #4247 is merged.

markscheel:
Rebased.

knelli2 (approved):

nilsdeppe (approved):

PR #4269, opened 2022-10-04
Fix uninitialized timescale tuners from options

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/ControlSystem/Tags/Test_MeasurementTimescales.cpp
+++ b/tests/Unit/ControlSystem/Tags/Test_MeasurementTimescales.cpp
@@ -18,13 +18,15 @@
 #include "ControlSystem/Tags/MeasurementTimescales.hpp"
 #include "ControlSystem/TimescaleTuner.hpp"
 #include "DataStructures/DataVector.hpp"
+#include "Domain/Creators/DomainCreator.hpp"
+#include "Helpers/ControlSystem/SystemHelpers.hpp"
 #include "Helpers/ControlSystem/TestStructs.hpp"
 #include "Utilities/GetOutput.hpp"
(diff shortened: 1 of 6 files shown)


tests/Unit/ControlSystem/Tags/Test_MeasurementTimescales.cpp
```diff
@@ -18,6 +18,8 @@
 #include "ControlSystem/Tags/MeasurementTimescales.hpp"
```
nilsdeppe, line 18: Sorry, I'm not familiar with these tests. Where would the test fail now if the the changes in this PR weren't made? It looks like most things are `static_assert` but the code changes make a runtime change to data. Is there a way to verify the runtime change occurred?

Edit: the reason I'm asking is because when we find a bug, I think it's important to add a test that would've caught the bug :)

nilsdeppe (changes requested):

tests/Unit/ControlSystem/Tags/Test_MeasurementTimescales.cpp
knelli2: I edited some tests so it should catch this in the future if anything changes. Basically now anywhere that uses timescales has them both initialized and uninitialized. Then the uninitialized ones will automatically be initialized with these new changes as they should have been before.

nilsdeppe: Thank you!

nilsdeppe (approved):

nilsdeppe:
CI failure is timeout in elliptic yaml tests

PR #4274, opened 2022-10-06
Have VectorImpl use static allocation for small sizes

knelli2:
## Proposed changes

Note to reviewers: The test is mostly whitespace changes, so hiding whitespace changes may be useful.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/DevGuide/ImplementingVectors.md
+++ b/docs/DevGuide/ImplementingVectors.md
@@ -32,18 +32,22 @@ high-performance arithmetic library
 vectors gracefully make use of the math functions defined for the Blaze types,
 but can be customized for the specific needs in SpECTRE computations.
 
-The pair of template parameters for `VectorImpl` are the type of the stored data
-(e.g. `double` for `DataVector`), and the result type for mathematical
-operations. The result type is used by Blaze to ensure that only compatible
-vector types are used together in mathematical expressions. For example, a
-vector representing `double` data on a grid (`DataVector`) cannot be added to a
-vector representing spectral coefficients (`ModalVector`). This avoids subtle
-bugs that arise when vector types are unintentionally mixed.  In nearly all
-cases the result type will be the vector type that is being defined, so, for
-instance, `DataVector` is a derived class of `VectorImpl<double,
-DataVector>`. This template pattern is known as the
-["Curiously Recurring Template Pattern"]
-(https://en.wikipedia.org/wiki/Curiously_recurring_template_pattern) (CRTP).
+The trio of template parameters for `VectorImpl` are the type of the stored data
+(e.g. `double` for `DataVector`), the result type for mathematical operations,
+and the static size. The result type is used by Blaze to ensure that only
+compatible vector types are used together in mathematical expressions. For
+example, a vector representing `double` data on a grid (`DataVector`) cannot be
+added to a vector representing spectral coefficients (`ModalVector`). This
+avoids subtle bugs that arise when vector types are unintentionally mixed.
+In nearly all cases the result type will be the vector type that is being
+defined, so, for instance, `DataVector` is a derived class of
+`VectorImpl<double, DataVector, 5>`. This template pattern is known as the
+["Curiously Recurring Template Pattern"](https://en.wikipedia.org/wiki/
+Curiously_recurring_template_pattern) (CRTP).
+The static size is used as an optimization for small vector sizes. If your
+vector is small, rather than doing heap allocations, it will use stack
+allocations in order to be efficient. The default static size is set by a global
+constexpr bool `default_vector_impl_static_size`.
 
 For the Blaze system to use the CRTP inheritance appropriately, it requires the
 specification of separate type traits in the `blaze` namespace.
@@ -66,14 +70,14 @@ will be implicitly inherited from `VectorImpl`. The assignment and constructors
 may be inherited calling the following alias code in the vector class
 definition:
 ```
-using VectorImpl<T,VectorType>::operator=;
-using VectorImpl<T,VectorType>::VectorImpl;
+using VectorImpl<T,VectorType,StaticSize>::operator=;
+using VectorImpl<T,VectorType,StaticSize>::VectorImpl;
 ```
 
 Only the mathematical operations supported on the base Blaze types are supported
 by default. Those operations are determined by the storage type `T` and by the
-Blaze library. See [blaze-wiki/Vector_Operations]
-(https://bitbucket.org/blaze-lib/blaze/wiki/Vector%20Operations).
+Blaze library. See [blaze-wiki/Vector_Operations](https://bitbucket.org/
+blaze-lib/blaze/wiki/Vector%20Operations).
 
 # Allowed operator specification {#blaze_definitions}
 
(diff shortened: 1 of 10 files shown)


src/DataStructures/VectorImpl.hpp
```diff
@@ -119,12 +128,12 @@ struct MarkAsVectorImpl {};
...
  * \note
- * - If created with size 0, then  `data()` will return `nullptr`
  * - If either `SPECTRE_DEBUG` or `SPECTRE_NAN_INIT` are defined, then the
  *   `VectorImpl` is default initialized to `signaling_NaN()`. Otherwise, the
  *   vector is filled with uninitialized memory for performance.
  */
-template <typename T, typename VectorType>
+template <typename T, typename VectorType,
+          size_t StaticSize = default_vector_impl_static_size>
 class VectorImpl
```
wthrowe, line 137: The TensorExpression code, at least, assumes `VectorImpl` only has two parameters.  For example, this won't match custom-sized vectors:
```c++
template <typename ValueType, typename VectorType>
struct get_binop_datatype_impl<VectorImpl<ValueType, VectorType>, ValueType> {
```

src/DataStructures/VectorImpl.hpp
```diff
@@ -297,43 +297,56 @@ class VectorImpl
...
 };
 
-template <typename T, typename VectorType>
-VectorImpl<T, VectorType>::VectorImpl(const VectorImpl<T, VectorType>& rhs)
-    : BaseType{rhs},
-      owned_data_(
-          rhs.size() > 0
-              ? cpp20::make_unique_for_overwrite<value_type[]>(rhs.size())
-              : nullptr) {
+/// \cond HIDDEN_SYMBOLS
```
wthrowe, line 333: Why hide these?

wthrowe (changes requested):
I would like more justification for this change.  What problem is it trying to solve?

src/DataStructures/VectorImpl.hpp
knelli2: I was getting errors when building the docs that Doxygen wasn't able to find some symbols. The declarations are already hidden above, so I just hid the definitions as well.

knelli2:
Added a new commit that addresses the concerns about TensorExpressons (big thanks to @macedo22). 

As for justification, the main reason that prompted me to add this was the functions of time. Every time we call `func`, we do a heap allocation for the `DataVector`(s) and we compute the time dependent part of the maps pretty often so we were doing a bunch of heap allocations. However, for almost all functions of time (except for shape) we know the number of components, the number won't change throughout the evolution, and each one has 4 components or less. With these changes, now we do stack allocations for the `DataVector`s in the functions of time.

Since `DataVector` is technically a `NodalVector`, it shouldn't really be used for the functions of time. So in a follow-up PR I was going to add a new vector type that we use for the functions of time.

knelli2:
Test failure seems unrelated

tests/Unit/DataStructures/Tensor/Expressions/CustomStaticSizeVector.hpp
```diff
@@ -0,0 +1,33 @@
+// Distributed under the MIT License.
```
wthrowe: Move this file into `tests/Unit/Helpers/`.

wthrowe (changes requested):
OK, that's a reasonable use case.

A few more files that I think need updating:
- docs/DevGuide/ImplementingVectors.md (probably just the last section)
- src/DataStructures/DynamicBuffer.hpp
- src/DataStructures/Variables.hpp

knelli2:
@wthrowe Ok I think I got everything. Let me know if I missed anything

src/DataStructures/VectorImpl.hpp
```diff
@@ -83,6 +83,10 @@ constexpr bool is_assignable_v = is_assignable<LhsDataType, RhsDataType>::value;
 /// `VectorImpl`
 struct MarkAsVectorImpl {};
 
+/// \ingroup DataStructuresGroup
+/// Default static size for vector impl
+constexpr size_t default_vector_impl_static_size = 5;
```
nilsdeppe: Should we keep this at `0` for now? I'm not sure it will benefit us except maybe in 1d, and if we know we need it, we can use the `HybridVector` you said you'll add. It'll also help keep down memory usage. @wthrowe @kidder thoughts on this?

docs/DevGuide/ImplementingVectors.md
```diff
@@ -324,25 +328,37 @@ memory. Non-owning vectors do not manage memory, nor can they change size. The
...
+   member variable called `static_owned_data_`. Since it is on the stack, this
+   doesn't require any memory management by the user.
+
+In either case, the base `blaze::CustomVector` must also be told about the
+pointer, which is always accomplished by calling the protected function
 `VectorImpl.reset_pointer_vector(const size_t set_size)`, which sets the
-`blaze::CustomVector` internal pointer to the pointer obtained by
-`std::unique_pointer.get()`.
+`blaze::CustomVector` internal pointer to either the pointer obtained by
+`std::unique_pointer.get()` or the pointer obtained by `std::array.get()`
```
wthrowe: `array.data()`

docs/DevGuide/ImplementingVectors.md
```diff
@@ -324,25 +328,37 @@ memory. Non-owning vectors do not manage memory, nor can they change size. The
...
+depending on the size of the vector.
 
 When a SpECTRE vector is constructed as non-owning by the `VectorImpl(ValueType*
 start, size_t set_size)` constructor, or becomes non-owning by the
-`set_data_ref` function, the internal `std::unique_ptr` named `owned_data_` no
-longer points to the data represented by the vector and can be thought of as
+`set_data_ref` function, neither the internal `std::unique_ptr` named
+`owned_data_` nor the internal `std::array` names `static_owned_data_`
```
wthrowe: "named"

src/DataStructures/DynamicBuffer.hpp
```diff
@@ -28,14 +28,19 @@ class er;
...
 class DynamicBuffer {
  public:
+  // Rather than having to template DynamicBuffer on the static size, just use
+  // the default static size since this only determines if something is derived
+  // from VectorImpl and not the specific size itself
   static constexpr bool is_data_vector_type =
-      std::is_base_of_v<VectorImpl<tt::get_fundamental_type_t<T>, T>, T>;
+      std::is_base_of_v<VectorImpl<tt::get_fundamental_type_t<T>, T,
+                                   default_vector_impl_static_size>,
+                        T>;
```
wthrowe: I don't understand.  Isn't this checking that the static size is the default?

src/DataStructures/VectorImpl.hpp
wthrowe: I like this idea.

wthrowe (changes requested):

src/DataStructures/DynamicBuffer.hpp
knelli2: Yes. So currently, `DynamicBuffer` is only used in `YlmSpherepack`. And `DynamicBuffer` is only instantiated for `double, DataVector`. No other types. So we want to check if it's using the default because that's what `DataVector` uses. If in the future we want to extend `DynamicBuffer` to other vector types that don't use the default static size, then we can do that.

knelli2:
Pushed some fixups setting the default static size to 0 and addressing other comments

src/DataStructures/DynamicBuffer.hpp
wthrowe: If it's only supposed to work for DataVector just check if it's a DataVector instead of doing stuff with VectorImpl inheritance.

tests/Unit/Helpers/DataStructures/VectorImplTestHelper.hpp
```diff
@@ -44,15 +44,20 @@ void vector_test_construct_and_assign(
...
-                default_vector_impl_static_size <= 20);
-  UniformCustomDistribution<size_t> static_sdist{
-      2, default_vector_impl_static_size};
-  UniformCustomDistribution<size_t> sdist{default_vector_impl_static_size + 1,
-                                          20};
-
-  std::vector<size_t> sizes{static_sdist(gen), default_vector_impl_static_size,
-                            sdist(gen)};
+  std::vector<size_t> sizes;
+  if constexpr (default_vector_impl_static_size >= 2) {
```
wthrowe: Instead of checking the default size, check the size for `VectorType` and call the test functions somewhere with a vector with non-zero static size.

wthrowe (changes requested):

nilsdeppe (commented):
I have no additional comments

knelli2:
Edited the existing fixups. Added a new test with a custom vector with non-zero static size

wthrowe:
OK, I'm happy now.

knelli2:
Squashed and rebased.

wthrowe (approved):

nilsdeppe (approved):

nilsdeppe:
CI failure is fixed in #4289

PR #4277, opened 2022-10-07
Fix unsigned-integer bug in FastFlow.

markscheel:
Previously it did the wrong thing when
current_iter_ < divergence_iter_ because of
unsigned-integer underflow.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ApparentHorizons/FastFlow.cpp
+++ b/src/ApparentHorizons/FastFlow.cpp
@@ -234,7 +234,7 @@ FastFlow::iterate_horizon_finder(
 
   // Treat the case in which residual_mesh_norm is increasing
   if (residual_mesh_norm > divergence_tol_ * min_residual_mesh_norm_ and
-      iter_at_min_residual_mesh_norm_ <= current_iter_ - divergence_iter_) {
+      iter_at_min_residual_mesh_norm_ + divergence_iter_ <= current_iter_) {
     // clang-tidy: std::move of trivially-copyable type
     return std::make_pair(Status::DivergenceError,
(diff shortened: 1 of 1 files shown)


markscheel:
Just to have it written down in the PR somewhere:

The effect of the bug is that FastFlow might exit because of divergence immediately in the first few iterations, if the relationship between the two residuals is not as expected.  Whereas what should happen is that FastFlow exits because of divergence only if it has gone a certain number of iterations (`divergence_iter_`) without seeing the residuals decrease in the expected way.

geoffrey4444 (approved):

geoffrey4444 (approved):
I verified that with this fix the run continues with my original control system settings, which cap the damping timescale at 20 instead of 2.67.

knelli2 (approved):
My run that was previously failing has also gone well past the point where the AH finder was diverging last time. I think this has fixed it.

nilsdeppe (approved):
LGTM, is it relatively easy to add a test for this? Maybe in a followup PR?

markscheel:
It's not straightforward as-is to add tests for when convergence fails, because that requires one to set up a metric and an initial Strahlkorper in such a way that the AH doesn't converge, and even worse, set it up so it diverges in a particular way.

But it should be possible to add such a test by refactoring FastFlow so that the convergence criteria are in a separate public member function (which is public only for the sake of the unit tests) and then for the test to give that function fake residuals that make it behave in a certain way.

nilsdeppe:
Okay, thanks for explaining!

PR #4278, opened 2022-10-09
Limit step size to foft expiration times in BBH runs with global timestepping

geoffrey4444:
## Proposed changes

When using global time stepping in a binary-black-hole evolution, insert an action just after dense triggers run that ensures that the step never exceeds the expiration time of the functions of time. This is necessary to evolve binary black holes with substep time steppers such as DormandPrince5 and RungeKutta4 with global timestepping.

### Upgrade instructions

After this PR, evolutions of binary black holes that use global timestepping must also use adaptivity. The action that avoids stepping beyond the expiration times of the functions of time sometimes shrinks the step size. Without some adaptive condition to enlarge the step size again (such as a CFL-based condition), the step size of a binary-black-hole evolution will steadily shrink.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
+++ b/src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
@@ -1,0 +1,175 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+/// \file
+/// Defines action AdvanceTime
+
+#pragma once
+
+#include <cmath>
+#include <iomanip>
+#include <limits>
+#include <optional>
+#include <tuple>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Time/Tags.hpp"
+#include "Time/Time.hpp"
+#include "Time/TimeStepId.hpp"
+#include "Time/TimeSteppers/TimeStepper.hpp"
+#include "Time/Utilities.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/FractionUtilities.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+#include "Evolution/EventsAndDenseTriggers/Tags.hpp"
+#include "Parallel/Printf.hpp"
+#include "Time/Tags.hpp"
+
+/// \cond
+namespace Parallel {
+template <typename Metavariables>
+class GlobalCache;
+}  // namespace Parallel
+namespace Tags {
+template <typename Tag>
+struct Next;
+}  // namespace Tags
+// IWYU pragma: no_forward_declare db::DataBox
+/// \endcond
+
+namespace Actions {
+/// \ingroup ActionsGroup
+/// \ingroup TimeGroup
+/// \brief Make sure step size does not exceed FunctionOfTime expiration times
+///
+/// \details For time steppers with substeps, functions of time might expire
+/// at a time earlier than one of the substeps; however, the functions of time
+/// are both needed at each substep and updated only after taking a full step.
+/// This can result in quiescence, i.e., in waiting to complete the next
+/// substep until the functions of time are updated, but the functions of time
+/// will only be updated after the next step is taken. To avoid this situation,
+/// at the start of each full step, this action checks whether the current
+/// step size exceeds the time remaining before the first function of time
+/// expires. If it does, then the step size is adjusted to be earlier than
+/// the expiration time of the next function of time to expire. Note that
+/// if the time stepper does not use substeps, this action does nothing.
+///
+/// Uses:
+/// - DataBox:
+///   - Tags::Time
+///   - Tags::TimeStep
+///   - Tags::TimeStepId
+///   - Tags::TimeStepper<>
+///   - domain::Tags::FunctionsOfTime
+///
+/// DataBox changes:
+/// - Adds: nothing
+/// - Removes: nothing
+/// - Modifies:
+///   - Tags::TimeStep
+struct LimitTimeStepToExpirationTimes {
+  template <typename DbTags, typename... InboxTags, typename Metavariables,
+            typename ArrayIndex, typename ActionList,
+            typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTags>& box, tuples::TaggedTuple<InboxTags...>& /*inboxes*/,
+      Parallel::GlobalCache<Metavariables>& cache,
+      const ArrayIndex& /*array_index*/, ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {  // NOLINT const
+    // First, check whether the time stepper uses substeps
+    const TimeStepper& time_stepper = db::get<Tags::TimeStepper<>>(box);
+    if (time_stepper.number_of_substeps() > 0) {
+      // Next, check if the current time step is at the beginning of a full step
+      const auto& time_step_id = db::get<Tags::TimeStepId>(box);
+      if (time_step_id.substep() == 0) {
+        // Get the functions of time and, if not empty, loop over them,
+        // finding the minimum expiration time
+        const auto& functions_of_time =
+            get<domain::Tags::FunctionsOfTime>(cache);
+        if (not functions_of_time.empty()) {
+          double min_expiration_time{std::numeric_limits<double>::max()};
+          for (const auto& [name, f_of_t] : functions_of_time) {
+            if (f_of_t->time_bounds()[1] < min_expiration_time) {
+              min_expiration_time = f_of_t->time_bounds()[1];
+            }
+          }
+
+          ASSERT(db::get<Tags::Time>(box) <= min_expiration_time,
+                 "Time (" << std::setprecision(20) << db::get<Tags::Time>(box)
+                          << ") is greater than expiration time ("
+                          << min_expiration_time << ") by an amount of "
+                          << db::get<Tags::Time>(box) - min_expiration_time);
+          if (db::get<Tags::Time>(box) == min_expiration_time) {
+            return {Parallel::AlgorithmExecution::Continue, std::nullopt};
+          }
+
+          const auto& initial_time_step{db::get<Tags::TimeStep>(box)};
+          const double step_end =
+              (time_step_id.step_time() + initial_time_step).value();
+          if (time_step_id.step_time().value() +
+                  2.0 * initial_time_step.value() <=
+              min_expiration_time) {
+            // The expiration times are far in the future (more than
+            // two steps).  No need to adjust anything.
+            return {Parallel::AlgorithmExecution::Continue, std::nullopt};
+          }
+
+          const double start{initial_time_step.slab().start().value()};
+          Slab new_slab{};
+          if (step_end + slab_rounding_error(time_step_id.step_time()) >=
+              min_expiration_time) {
+            // The minimum expiration time is within the next step (or
+            // within roundoff).  Step to it.
+            new_slab = Slab(start, min_expiration_time);
+          } else {
+            // The minimum expiration time is not within the next
+            // step, but is within two steps.  Shrink the step size to
+            // avoid taking a tiny step after this one.
+
+            // new_slab = initial_time_step.slab().with_duration_to_end(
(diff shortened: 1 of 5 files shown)


src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
...
+#include "Time/TimeSteppers/TimeStepper.hpp"
+#include "Time/Utilities.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/FractionUtilities.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+#include "Evolution/EventsAndDenseTriggers/Tags.hpp"
+#include "Parallel/Printf.hpp"
+#include "Time/Tags.hpp"
```
wthrowe, line 31: I think these three are unused (or a duplicate, for the last one).  If not, merge them with above.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+/// \file
+/// Defines action AdvanceTime
```
wthrowe, line 5: Remove.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
...
+#include "Domain/Structure/ElementId.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Time/Tags.hpp"
+#include "Time/Time.hpp"
+#include "Time/TimeStepId.hpp"
+#include "Time/TimeSteppers/TimeStepper.hpp"
+#include "Time/Utilities.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/FractionUtilities.hpp"
```
wthrowe, line 25: Unused.  Check the includes in general.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
...
+            }
+          }
+
+          ASSERT(db::get<Tags::Time>(box) <= min_expiration_time,
+                 "Time (" << std::setprecision(20) << db::get<Tags::Time>(box)
+                          << ") is greater than expiration time ("
+                          << min_expiration_time << ") by an amount of "
+                          << db::get<Tags::Time>(box) - min_expiration_time);
+          if (db::get<Tags::Time>(box) == min_expiration_time) {
+            return {Parallel::AlgorithmExecution::Continue, std::nullopt};
```
wthrowe, line 108: We know this actually should be `Retry` with a FoT update callback.  At a minimum, put a comment here.

It would be good to make an attempt to figure out why that doesn't work, because if there is some fundamental issue we may need to find a different solution than this action when we get to faster control system scales.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
...
+                  2.0 * initial_time_step.value() <=
+              min_expiration_time) {
+            // The expiration times are far in the future (more than
+            // two steps).  No need to adjust anything.
+            return {Parallel::AlgorithmExecution::Continue, std::nullopt};
+          }
+
+          const double start{initial_time_step.slab().start().value()};
+          Slab new_slab{};
+          if (step_end + slab_rounding_error(time_step_id.step_time()) >=
```
wthrowe, line 124: I know I wrote this, but I've changed my mind and don't like this increasing the step size, even by a small amount.  Please drop the `slab_rounding_error` term and update the comment below.

The slab-resizing code doesn't handle roundoff very well at the moment, but I'm doing a bunch of cleanup there at the moment and hope I'll get around to fixing that.  If I do, increasing the size here could confuse other code.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
...
+          } else {
+            // The minimum expiration time is not within the next
+            // step, but is within two steps.  Shrink the step size to
+            // avoid taking a tiny step after this one.
+
+            // new_slab = initial_time_step.slab().with_duration_to_end(
+            //     0.51 * (min_expiration_time - start));
+            // Note: 0.51 instead of 0.5 to guarantee no roundoff affecting
+            // what happens on the next step.
+            new_slab = Slab(start, 0.51 * min_expiration_time + 0.49 * start);
```
wthrowe, line 138: Weighting the average towards `min_expiration_time` like this is good, but add a check to avoid increasing the step size from its current value.

tests/Unit/Time/Actions/Test_LimitTimeStepToExpirationTimes.cpp
```diff
@@ -0,0 +1,175 @@
...
+  const auto& new_time_step = db::get<Tags::TimeStep>(box);
+  const auto& new_next_time_step = db::get<Tags::Next<Tags::TimeStep>>(box);
+  const auto& new_time_step_id = db::get<Tags::TimeStepId>(box);
+  const auto& new_next_time_step_id =
+      db::get<Tags::Next<Tags::TimeStepId>>(box);
+  const double new_slab_start_value{new_time_step.slab().start().value()};
+  const double new_slab_end_value{new_time_step.slab().end().value()};
+
+  CHECK(new_slab_start_value == time_step.slab().start().value());
+  CHECK((min_expiration_time == new_slab_end_value));
```
wthrowe, line 132: Remove extra parens.  They suppress some of the catch output.

wthrowe (changes requested):
You're getting a lot of CI failures.  It's not immediately obvious whether they are compiler-dependent or RNG-dependent.

wthrowe:
Make sure you put your upgrade instructions between the two "&lt;!-- UPGRADE INSTRUCTIONS --&gt;" lines when you submit PRs.  I've fixed it for you in this one.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
...
+///   - Tags::TimeStep
+///   - Tags::TimeStepId
+///   - Tags::TimeStepper<>
+///   - domain::Tags::FunctionsOfTime
+///
+/// DataBox changes:
+/// - Adds: nothing
+/// - Removes: nothing
+/// - Modifies:
+///   - Tags::TimeStep
```
knelli2, line 74: This action modifies a lot more tags than just this.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
knelli2: Agreed.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
...
+            // The minimum expiration time is within the next step (or
+            // within roundoff).  Step to it.
+            new_slab = Slab(start, min_expiration_time);
+          } else {
+            // The minimum expiration time is not within the next
+            // step, but is within two steps.  Shrink the step size to
+            // avoid taking a tiny step after this one.
+
+            // new_slab = initial_time_step.slab().with_duration_to_end(
+            //     0.51 * (min_expiration_time - start));
```
knelli2, line 135: I think this can be removed?

tests/Unit/Time/Actions/Test_LimitTimeStepToExpirationTimes.cpp
```diff
@@ -0,0 +1,175 @@
...
+  CHECK(new_time_step_id.substep_time().slab().start().value() ==
+        new_slab_start_value);
+  CHECK(new_time_step_id.substep_time().slab().end().value() ==
+        new_slab_end_value);
+
+  CHECK(new_next_time_step_id == db::get<Tags::TimeStepper<>>(box).next_time_id(
+                                     new_time_step_id, new_time_step));
+  CHECK(new_next_time_step.fraction() == time_step.fraction());
+  CHECK(new_next_time_step.slab() == new_next_time_step_id.step_time().slab());
+}
```
knelli2, line 162: I'm wondering if you should add some tests for all the different `if` statements in your action. Codecov wasn't run, but I think it'd say some parts of the code aren't tested. Things like

- Timestepper doesn't have substeps
- We aren't at the beginning of a timestep
- We are more than 2 steps away from the expiration time
- We are less than 2 steps more than 1 step away
- We are less than one step away

knelli2 (changes requested):
I'm not a huge fan of having to use an adaptive step size in order to use this action. Once you decide you need to change the time step, can you store the old time step in a tag of some kind? And then after we update the functions of time, reset the time step to the old value? I think this would require that `Continue` mentioned below to be a `Retry`.

I won't require this change as of right now, but it will probably be necessary soon so I wanted to mention it

wthrowe:
"Adaptive" is probably a bit of a grand term for what is actually required, which is really just something resetting the time step.  This would be sufficient:
```yaml
EventsAndTriggers:
  - - Always
    - - ChangeSlabSize:
          DelayChange: 0
          StepChoosers:
            - Constant: 0.01
```

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
```diff
@@ -0,0 +1,175 @@
...
+}  // namespace Parallel
+namespace Tags {
+template <typename Tag>
+struct Next;
+}  // namespace Tags
+// IWYU pragma: no_forward_declare db::DataBox
+/// \endcond
+
+namespace Actions {
+/// \ingroup ActionsGroup
```
nilsdeppe, line 46: Can this become a `SlabChooser`? Given the requirement that one uses `Slabchoosers` anyway, and that there is no check enforcing that (as far as I can tell), it seems like the safest thing to do is to have this be done as a `SlabChooser`. @wthrowe any thoughts on this?

nilsdeppe (changes requested):

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
wthrowe: I think it can be done at the cost of complicating the action loop some more.  The annoying part here is that Adams-Bashforth and Runge-Kutta time steppers do dense output differently.  AB requires the step size to change before dense output, while RK doesn't care.  (The AB behavior might change to match RK at some point, but not in the near future.)  This computation has to happen after dense output, but isn't used for AB.  So I think it would work if we had two versions of the ChangeSlabSize action, one that only ran for AB and one that only ran for RK, and put them on either side of dense output.

ChangeSlabSize is currently done well before dense output, but I think it commutes with all the actions in between in GTS mode.  (Not true in LTS mode.)

wthrowe:
I haven't thought it through, but this may be affected by similar race conditions as #4254.

src/Time/Actions/LimitTimeStepToExpirationTimes.hpp
nilsdeppe: Okay, thank you! I guess for me it's not clear if having two `ChangeSlabsize` actions is more complicated than having to add this action anyway. I'm fine either way, with a slight preference to having it be a `SlabChooser`, but I think you should make the decision :)

nilsdeppe:
I think we will want to be able to do BNS simulations with RK GTS. I've tested the subcell solver a lot more with RK since for GTS those seem to be more efficient. Is there a path to getting something like what is done in this PR merged?

wthrowe:
No, nothing like this will be merged.

The better solution is to use the control system timescales and such to control the time step.  The important thing is that there is a step between each measurement and the first update it contributes to.  I think that that interval is some multiple of the measurement timescale, so an easy way to do it is limit the time step to that.  That is more restrictive than necessary, since taking larger steps is fine as long as we hit that interval, but I don't know if it is easy for the elements to determine when that measurement and update happen.

nilsdeppe:
Okay, thanks for the explanation!

geoffrey4444:
@wthrowe has addressed the goal of this PR in a different way, so I'm going to close it. Happy to reopen if anyone would like me to revisit

knelli2:
This was addressed in #5593.

PR #4014, opened 2022-06-03
Explicitly use interpolator source vars in intrp framework

knelli2:
## Proposed changes

When using the Interpolator, the source variables passed around must be the ones defined in the `interpolator_source_vars` type alias in the metavariables. However this wasn't super clear to me as I was looking through the interpolation framework (see #3994). This PR mostly just renames template parameters and function arguments to make it obvious what tensors/variables we are talking about in these functions. I also added a couple `static_assert`s to enforce this as well. Hopefully this will aid in other people's understanding when they look at the interpolation framework.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

The reason for all this is because internally, the Interpolator parallel component stores a tag in its DataBox that has
```cpp
Variables<typename Metavariables::interpolator_source_vars> vars_from_element;
```
explicitly in it. That is why all these functions have to have the source vars from the metavariables. If this tag didn't explicitly reach into the metavariables, we wouldn't need to guarantee that the tensors passed in are the ones from the metavariables.

PR #4280, opened 2022-10-10
Dump volume data from interpolator upon failure

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Actions/DumpInterpolatorVolumeData.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Actions/DumpInterpolatorVolumeData.hpp
@@ -1,0 +1,135 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <optional>
+#include <string>
+#include <type_traits>
+#include <vector>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/Tensor/TensorData.hpp"
+#include "Domain/Tags.hpp"
+#include "IO/Observer/Actions/ObserverRegistration.hpp"
+#include "IO/Observer/ObservationId.hpp"
+#include "IO/Observer/ObserverComponent.hpp"
+#include "IO/Observer/Tags.hpp"
+#include "IO/Observer/VolumeActions.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Info.hpp"
+#include "Parallel/Invoke.hpp"
+#include "Parallel/Local.hpp"
+#include "ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp"
+#include "ParallelAlgorithms/Interpolation/Tags.hpp"
+#include "Utilities/MakeString.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace intrp::Actions {
+namespace detail {
+// This requires there to be a type alias named `interpolator_source_vars` in
+// the metavariables
+template <typename TemporalIdTag, typename Metavariables>
+ElementVolumeData construct_element_volume_data(
+    const ElementId<Metavariables::volume_dim>& element_id,
+    const typename intrp::Tags::VolumeVarsInfo<Metavariables,
+                                               TemporalIdTag>::Info& info) {
+  std::vector<TensorComponent> components{};
+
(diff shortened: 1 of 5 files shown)


src/ParallelAlgorithms/Interpolation/Actions/DumpInterpolatorVolumeData.hpp
```diff
@@ -0,0 +1,138 @@
...
+        }
+      });
+
+  const auto& mesh = info.mesh;
+  const auto& index_extents = mesh.extents();
+  const auto& array_bases = mesh.basis();
+  const auto& array_quadratures = mesh.quadrature();
+  std::vector<size_t> extents(index_extents.begin(), index_extents.end());
+  std::vector<Spectral::Basis> bases(array_bases.begin(), array_bases.end());
+  std::vector<Spectral::Quadrature> quadratures(array_quadratures.begin(),
```
wthrowe, line 63: Const or move.

src/ParallelAlgorithms/Interpolation/Actions/DumpInterpolatorVolumeData.hpp
```diff
@@ -0,0 +1,138 @@
...
+            typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTagList>& box,
+      const tuples::TaggedTuple<InboxTags...>& /*inboxes*/,
+      Parallel::GlobalCache<Metavariables>& cache,
+      const ArrayIndex& /*array_index*/, const ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    static_assert(
+        std::is_same_v<SourceVars,
+                       typename Metavariables::interpolator_source_vars>,
```
wthrowe: Is there a need for the template parameter, rather than just using the value from the metavariables?

src/ParallelAlgorithms/Interpolation/Actions/DumpInterpolatorVolumeData.hpp
```diff
@@ -0,0 +1,138 @@
...
+        "The source vars passed to DumpInterpolatorSourceVars must be the same "
```
wthrowe: Match action name.

wthrowe (changes requested):

src/ParallelAlgorithms/Interpolation/Actions/DumpInterpolatorVolumeData.hpp
knelli2: I did it to avoid "reaching up" into the metavariables, which, from what I understand, we were trying to avoid. I did similar things in #4014.

wthrowe: But you're reaching up into the metavariables right here.  In #4014 it was necessary for the pack expansion in the function arguments.

knelli2: Ok, then are we saying it's ok to reach up into the metavariables now? Or should I not reach up, and pass them in with the template parameter?

wthrowe: My understanding (I think @nilsvu and maybe @nilsdeppe have thought about this more) is that the bad thing is that there are some things in the metavariables that have special meanings and some that are basically just local variables, and there's no easy way to tell what is what.

When a quantity is only used in a couple of places, it is usually best to just pass it in as a template parameter.

When something is used in many places and is required to be exactly the same in all of them, or is difficult to supply as a template parameter for some reason, I'm not sure if we have a clear decision about what is best.  One possibility is dedicated structs with protocols that can be pulled from the metavariables, as is done for `factory_creation` and `domain`.  Many places still do the "reaching up".

In this case, since it is already being accessed in several places, I would just use the value from the metavariables and we can fix all uses once we have decided what the correct fix is.

nilsvu: I agree with your assessment @wthrowe. For the interpolation framework we should really clean up the Metavariables and put them in a struct and document them.

knelli2: Ok I just reached up for now. We can clean things up later. I did add a comment about it.

wthrowe (dismissed):

knelli2:
Rebased and fixed a conflict.

nilsvu (approved):
Do we always want to dump all this data on failure, or should this be controlled by an input-file toggle? @wthrowe feel free to merge at will

wthrowe (approved):

knelli2:
@nilsvu If it becomes a problem, I can add an input file toggle.

PR #4284, opened 2022-10-11
Clear GR solutions a bit

nilsdeppe:
## Proposed changes

This is necessary for setting initial data in the input file.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/GrMhd/GhValenciaDivClean/CMakeLists.txt
+++ b/src/Evolution/Executables/GrMhd/GhValenciaDivClean/CMakeLists.txt
@@ -56,33 +56,9 @@ add_ghmhd_with_horizon_executable(
   "${LIBS_TO_LINK}"
   )
 
-add_ghmhd_executable(
-  BlastWave
-  GeneralizedHarmonic::Solutions::WrappedGr<grmhd::AnalyticData::BlastWave>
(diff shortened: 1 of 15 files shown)


src/Evolution/Executables/GrMhd/GhValenciaDivClean/CMakeLists.txt
```diff
@@ -56,54 +56,18 @@ add_ghmhd_with_horizon_executable(
   "${LIBS_TO_LINK}"
   )
```
knelli2, line 57: Question: If the default values for these tests lead to immediate black hole formation, why remove the executables rather than change the default values of the tests? Also, is immediate black hole formation a bad thing for all of these executables?

nilsdeppe: These are standard test problems for special relativistic MHD, not GRMHD, so just don't apply with dynamical gravity. There's nothing in the literature to compare to with these tests...

knelli2: Ok thanks for the explanation

knelli2 (approved):

kidder (approved):

PR #4290, opened 2022-10-12
Add damped harmonic class for runtime gauge condition

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.hpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.hpp
@@ -4,10 +4,12 @@
 #pragma once
 
 #include <cstddef>
+#include <limits>
 
 #include "DataStructures/DataBox/Tag.hpp"
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "Domain/Tags.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Gauges.hpp"
 #include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
 #include "PointwiseFunctions/GeneralRelativity/TagsDeclarations.hpp"
 #include "Time/Tags.hpp"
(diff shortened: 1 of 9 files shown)


src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.hpp
```diff
@@ -195,5 +197,89 @@ void damped_harmonic(
...
+ */
+class DampedHarmonic final : public GaugeCondition {
+ public:
+  /// The width of the Gaussian for the spatial decay of the damped harmonic
+  /// gauge.
+  struct SpatialDecayWidth {
+    using type = double;
+    static constexpr Options::String help{
+        "Spatial width of weight function used in the damped harmonic "
+        "gauge."};
```
knelli2: Since you use the names of the amplitudes $A_X$ and exponents $e_X$ below, maybe change this to
`Spatial width (sigma_r) of weight function (W) used ...` or something similar, just so symbols and names match up with what is in the equations.

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.hpp
```diff
@@ -195,5 +197,89 @@ void damped_harmonic(
...
+  // NOLINTNEXTLINE(google-runtime-references)
+  void pup(PUP::er& p) override;
+
+ private:
+  double spatial_decay_width_{std::numeric_limits<double>::signaling_NaN()};
+  std::array<double, 3> amplitudes_{
+      {std::numeric_limits<double>::signaling_NaN(),
+       std::numeric_limits<double>::signaling_NaN(),
+       std::numeric_limits<double>::signaling_NaN()}};
+  std::array<int, 3> exponents_{{2, 2, 2}};
```
knelli2: Since `int` doesn't have a `signaling_NaN`, is there a way to ensure that components have been set properly? Not default initializing `exponents_`? Just want to protect against a silent failure where the exponents somehow don't get initialized properly to their input values and are just 2.

nilsdeppe: Not that I trivially know of. I could set it to the maximum value. We do that for `size_t` and most things will break horribly in that case. Would that be work for you?

knelli2: Yeah that sounds good!

nilsdeppe:
Rebased and squashed. Thanks for the review :)

knelli2 (approved):

kidder (approved):

Issue #4293, opened 2022-10-13
Early clean termination from file existence


nilsdeppe:
It would be very useful to be able to terminate a simulation early, i.e. before whatever final time was specified. This is useful for being able to terminate a job before the final time but being able to avoid output file corruption.

A discussion with @wthrowe @markscheel @knelli2 @ermost and @nilsvu considered two designs:
1. Check for file on disk
2. Termination handler

The second is a bit tricky because clusters might do all sorts of fun things with signals. SpEC had a lot of trouble with this.

going with option 1.:
Have a Trigger that holds a thread_local `TimeStepId` and `proc_id_to_check_on`. Then, essentially:
```cpp
if (my_proc == proc_id_to_check_on and current_time_step_id < last_time_step_id
    and file_system::check_if_file_exists(file_name_)) {
  // Do broadcast to all array elements signalling  a graceful end. This would set a flag
  // in the DataBox that an action would need to check for
}
```
The `proc_id_to_check_on` needs to be chosen in a way that guarantees an element will be on that core. That means it needs to be aware of the `Parallel::ResourceInfo`. An easy choice is the smallest number core that has an element on it. Which cores don't have elements is given by the `procs_to_ignore()` of `ResourceInfo`.

The synchronization algorithm described by @wthrowe for stopping all elements at the same Slab (or maybe Time?)
```
Synchronization algorithm: Stop the element at the start of the step (slab?)
before sending data (if some synchronize flag is set).  If we receive (or already
have) any flux communications, restart/don't stop and take another step.  No
element will take a step unless its neighbor is ahead of it, so eventually the
evolution will stop at the time of the most-advanced element.  Being local,
though, it requires the information about the most-advanced element to propagate
across the domain, which takes time.
```

wthrowe:
Not sure if it can be quite this general, but we could try for an input file appearance of
```yaml
EventsAndTriggers:
  - - And:
        - OnlyOneElement
        - FileExists:
            File: stop.txt
    - - Synchronized:
          - Completion
          - ObserveFields:
              ...
```
With big warnings in the docs for the special events and triggers that bad things will happen if you use them wrong.

Could also use this trigger to print diagnostics on only one element.

nilsdeppe:
I like that idea. Keeps things more modular.

wthrowe:
More thoughts:

[...]

PR #4294, opened 2022-10-13
Add some resource info into the GlobalCache

knelli2:
## Proposed changes

Adds the `Parallel::Tags::ResourceInfo` tag to the global cache.

To do this cleanly, I had to add an entry method to the global cache that will set the resource info tags. This entry method should only be called once, similar to `set_parallel_components`.

If any proc info is needed for #4293, this is where it will be stored.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/GlobalCache.hpp
+++ b/src/Parallel/GlobalCache.hpp
@@ -23,6 +23,7 @@
 #include "Parallel/Local.hpp"
 #include "Parallel/ParallelComponentHelpers.hpp"
 #include "Parallel/PupStlCpp17.hpp"
+#include "Parallel/ResourceInfo.hpp"
 #include "Parallel/Serialize.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
@@ -43,6 +44,10 @@
 namespace Parallel {
 template <typename Metavariables>
 struct GlobalCache;
+namespace Tags {
+template <typename Metavariables>
+struct ResourceInfo;
+}  // namespace Tags
 }  // namespace Parallel
 namespace mem_monitor {
 template <typename Metavariables>
@@ -467,6 +472,13 @@ class GlobalCache : public CBase_GlobalCache<Metavariables> {
   /// framework. Trying to do either of these will result in an ERROR.
   void compute_size_for_memory_monitor(const double time);
 
+  /// Entry method that will set the value of the Parallel::Tags::ResourceInfo
+  /// tag to the value passed in (if the tag exists in the GlobalCache)
+  ///
+  /// This is only meant to be called once.
+  void set_resource_info(
+      const Parallel::ResourceInfo<Metavariables>& resource_info);
+
   /// Retrieve the proxy to the global cache
   proxy_type get_this_proxy();
 
@@ -549,6 +561,7 @@ class GlobalCache : public CBase_GlobalCache<Metavariables> {
   MutableGlobalCache<Metavariables>* mutable_global_cache_{nullptr};
   mutable_global_cache_proxy_type mutable_global_cache_proxy_{};
   bool parallel_components_have_been_set_{false};
+  bool resource_info_has_been_set_{false};
   std::optional<main_proxy_type> main_proxy_;
   // Defaults for testing framework
   int my_proc_{0};
@@ -659,6 +672,20 @@ void GlobalCache<Metavariables>::compute_size_for_memory_monitor(
         "GlobalCache::compute_size_for_memory_monitor can only be called if "
         "the MemoryMonitor is in the component list in the metavariables.\n");
   }
+}
+
+template <typename Metavariables>
+void GlobalCache<Metavariables>::set_resource_info(
+    const Parallel::ResourceInfo<Metavariables>& resource_info) {
+  ASSERT(not resource_info_has_been_set_,
+         "Can only set the resource info once");
+  if constexpr (tmpl::list_contains_v<
+                    tags_list, Parallel::Tags::ResourceInfo<Metavariables>>) {
+    tuples::get<Tags::ResourceInfo<Metavariables>>(const_global_cache_) =
+        resource_info;
+  }
+  resource_info_has_been_set_ = true;
+  (void)resource_info;
 }
 #if defined(__GNUC__) && !defined(__clang__)
 #pragma GCC diagnostic pop
@@ -781,6 +808,7 @@ void GlobalCache<Metavariables>::pup(PUP::er& p) {
   p | mutable_global_cache_proxy_;
   p | main_proxy_;
   p | parallel_components_have_been_set_;
+  p | resource_info_has_been_set_;
   p | my_proc_;
   p | my_node_;
   p | my_local_rank_;
--- a/src/Parallel/ResourceInfo.hpp
+++ b/src/Parallel/ResourceInfo.hpp
@@ -8,6 +8,7 @@
 #include <ios>
 #include <optional>
 #include <pup.h>
+#include <set>
 #include <sstream>
 #include <string>
 #include <unordered_set>
@@ -108,12 +109,27 @@ struct SingletonInfoHolder {
   bool is_exclusive() const { return exclusive_; }
 
  private:
+  template <typename ParallelComponent>
+  friend bool operator==(const SingletonInfoHolder<ParallelComponent>& lhs,
+                         const SingletonInfoHolder<ParallelComponent>& rhs);
   // We use size_t here because we want a non-negative integer, but we use int
   // in the option because we want to protect against negative numbers. And a
   // negative size_t is actually a really large value (it wraps around)
   std::optional<size_t> proc_{std::nullopt};
   bool exclusive_{false};
 };
+
+template <typename ParallelComponent>
+bool operator==(const SingletonInfoHolder<ParallelComponent>& lhs,
+                const SingletonInfoHolder<ParallelComponent>& rhs) {
+  return lhs.proc_ == rhs.proc_ and lhs.exclusive_ == rhs.exclusive_;
+}
+
+template <typename ParallelComponent>
+bool operator!=(const SingletonInfoHolder<ParallelComponent>& lhs,
+                const SingletonInfoHolder<ParallelComponent>& rhs) {
+  return not(lhs == rhs);
+}
 
 template <typename ParallelComponents>
 struct SingletonPack;
@@ -206,8 +222,24 @@ struct SingletonPack<tmpl::list<ParallelComponents...>> {
   }
 
  private:
+  template <typename... Components>
+  friend bool operator==(const SingletonPack<tmpl::list<Components...>>& lhs,
+                         const SingletonPack<tmpl::list<Components...>>& rhs);
+
   tuples::tagged_tuple_from_typelist<local_tags> procs_{};
 };
+
+template <typename... Components>
+bool operator==(const SingletonPack<tmpl::list<Components...>>& lhs,
+                const SingletonPack<tmpl::list<Components...>>& rhs) {
+  return lhs.procs_ == rhs.procs_;
+}
+
+template <typename... Components>
+bool operator!=(const SingletonPack<tmpl::list<Components...>>& lhs,
+                const SingletonPack<tmpl::list<Components...>>& rhs) {
+  return not(lhs == rhs);
+}
 
 namespace detail {
 template <typename Component>
@@ -426,6 +458,10 @@ struct ResourceInfo {
   /// `allocate_array` function of the array component
   const std::unordered_set<size_t>& procs_to_ignore() const;
 
+  /// Returns a `std::set<size_t>` that has all processors available to put
+  /// elements on, meaning processors that aren't ignored.
+  const std::set<size_t>& procs_available_for_elements() const;
+
   /// Returns the proc that the singleton `Component` should be placed on.
   template <typename Component>
   size_t proc_for() const;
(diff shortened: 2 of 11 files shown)


src/Parallel/ResourceInfo.hpp
```diff
@@ -426,6 +426,9 @@ struct ResourceInfo {
   /// `allocate_array` function of the array component
   const std::unordered_set<size_t>& procs_to_ignore() const;
 
+  /// Returns a `std::set<size_t>` that has all processors with elements on them
+  const std::set<size_t>& procs_with_elements() const;
```
wthrowe: `#include <set>`

src/Parallel/GlobalCache.hpp
```diff
@@ -660,6 +674,26 @@ void GlobalCache<Metavariables>::compute_size_for_memory_monitor(
...
+    tuples::get<Tags::ProcsToIgnore>(const_global_cache_) = procs_to_ignore;
+  }
+  if constexpr (tmpl::list_contains_v<tags_list, Tags::ProcsWithElements>) {
+    tuples::get<Tags::ProcsWithElements>(const_global_cache_) =
+        procs_with_elements;
+  }
+  if constexpr (tmpl::list_contains_v<tags_list, Tags::FirstProcWithElements>) {
+    tuples::get<Tags::FirstProcWithElements>(const_global_cache_) =
+        *std::min_element(procs_with_elements.begin(),
+                          procs_with_elements.end());
```
wthrowe: The smallest element has to be the first one, because `std::set` is ordered.

wthrowe (changes requested):

wthrowe:
Squash directly.

src/Parallel/Tags/ResourceInfo.hpp
```diff
@@ -63,5 +67,36 @@ struct SingletonInfo : db::SimpleTag {
     return resource_info.template get_singleton_info<ParallelComponent>();
   }
 };
+
```
kidder, line 70: why not just add a single tag for a `ResourceInfo` and store that in the global cache?

src/Parallel/ResourceInfo.hpp
```diff
@@ -765,6 +778,13 @@ void ResourceInfo<Metavariables>::build_singleton_map(
          "number of auto exclusive singletons to be allocated is "
              << alg::accumulate(auto_exclusive_singletons_on_each_node, 0_st));
 
+  // procs_to_ignore_ is now complete. Now construct procs_with_elements_
+  for (size_t i = 0; i < num_procs; i++) {
+    if (procs_to_ignore_.find(i) == procs_to_ignore_.end()) {
+      procs_with_elements_.insert(i);
```
kidder: this is just the list of procs not ignored. I do not know if it is possible to exclude procs when migrating elements and creating new elements, or to insist that a particular proc have an element.

kidder (changes requested):
I have not had time yet to parse the issue and discussion this PR is targeting.  I will do so before Monday's meeting in case I have additional concerns to raise then...

knelli2:
Took @kidder's suggestion and just made one `Parallel::Tags::ResourceInfo` tag rather than several individual ones. Required a little rearranging of other things.

As for whether we will be able to exclude procs when migrating/creating/destroying elements, I figured we'd have to make our own load balancer in order to account for this. We can address that once we address the charm bug that's preventing use from migrating elements currently.

src/Parallel/ResourceInfo.hpp
wthrowe: Yes, this assumes every non-ignored proc has at least one element.  I guess that won't have to be true with load-balancing.  I guess we could recalculate the lowest proc after each load-balance, assuming we synchronize for those.

I suppose we could just pick an element to be special after each AMR regrid, too, and not have to worry about what processor it is on.

Do we currently error if more procs are allocated to elements than there are elements?

src/Parallel/Tags/ResourceInfo.hpp
```diff
@@ -63,5 +67,29 @@ struct SingletonInfo : db::SimpleTag {
...
+/// options, the return of `create_from_options` is just a default constructed
+/// Parallel::ResourceInfo.
+template <typename Metavariables>
+struct ResourceInfo : db::SimpleTag {
+  using type = Parallel::ResourceInfo<Metavariables>;
+
+  static constexpr bool pass_metavariables = true;
+
+  template <typename Metavars>
+  using option_tags = tmpl::list<Parallel::OptionTags::ResourceInfo<Metavars>>;
```
wthrowe, line 89: This tag isn't used below.  Is it needed to force it's inclusion in the input file for when this is initialized?  Please add a comment explaining.

wthrowe (changes requested):

src/Parallel/Tags/ResourceInfo.hpp
kidder: - We do have the concept of `zeroth_element` (it is the Element in Block 0 that contains BlockLogicalCoordinate (-1,-1,-1) ) so it is guaranteed to exist but its `ElementId` (and proc) will change with AMR 
- We do not currently error if a proc is empty (and I'm not even sure how to check this state without a global reduction)

src/Parallel/ResourceInfo.hpp
knelli2: Yes that is correct that it assumes every non-ignored proc has elements. Maybe I can rename the function to `nonignored_procs()` to avoid confusion?

For just picking an element, we could use the `is_zeroth_element()` function or something similar. Won't necessarily be the same element before/after an AMR regrid, but it is guaranteed to exist.

I don't think we error if there are more procs than elements, but I'm not sure (I've never tried). I'm not sure what the Z-curve does in that case.

src/Parallel/Tags/ResourceInfo.hpp
knelli2: You just beat me to it 😉

src/Parallel/ResourceInfo.hpp
wthrowe: What is the motivation for this PR, then?  I thought it was to give us a way to run something on only one element, but it sounds like we can do that more simply with `is_zeroth_element`.

knelli2: Yeah it sounds like we can do it easier with `is_zeroth_element`. However, if you wanted to guarantee the element was on a specific proc, then you couldn't use `is_zeroth_element`.

Regardless, I think it might still be a good idea to have the `ResourceInfo` in the global cache. It is global info that all components can take advantage of if need be. And if we continue to add things to it in the future, the new info will already be readily available to all components to use.

wthrowe: OK, let's add it.  I'm generally wary of adding stuff without an intended use case, but this is pretty small.

knelli2:
Rebased on develop.

wthrowe (dismissed):

src/Parallel/ResourceInfo.hpp
nilsdeppe: One motivation for this PR that I'm aware of is we will try some manual load-balancing using the Z-curve (or other space filling curves as we add them) combined with `proxy.migrateMe(target_proc)` but in this case we want to avoid placing elements on cores we aren't supposed to. So for that we want to have access to which procs we aren't allowed to use stored in the cache. If my memory is correct, the conclusion was "If we need some of the ResourceInfo, we may as well just add it all to the cache since it seems likely we will need it at _some_ point."

src/Parallel/GlobalCache.hpp
```diff
@@ -467,6 +472,13 @@ class GlobalCache : public CBase_GlobalCache<Metavariables> {
   /// framework. Trying to do either of these will result in an ERROR.
   void compute_size_for_memory_monitor(const double time);
 
+  /// Entry method that will set the value of the Parallel::Tags::ResourceInfo
+  /// tag to the value passed in (if the tag exists in the GlobalCache)
+  ///
+  /// This is only meant to be called once.
```
nilsdeppe, line 478: If it should only be called once, should there be a member variable that we set once this is called to avoid multiple calls?

src/Parallel/GlobalCache.hpp
```diff
@@ -660,6 +672,17 @@ void GlobalCache<Metavariables>::compute_size_for_memory_monitor(
...
+
+template <typename Metavariables>
+void GlobalCache<Metavariables>::set_resource_info(
+    const Parallel::ResourceInfo<Metavariables>& resource_info) {
+  if constexpr (tmpl::list_contains_v<
+                    tags_list, Parallel::Tags::ResourceInfo<Metavariables>>) {
+    tuples::get<Tags::ResourceInfo<Metavariables>>(const_global_cache_) =
+        resource_info;
+  }
+  (void)resource_info;
```
nilsdeppe, line 688: [optional] you can also mark the function argument with `[[maybe_unused]]`

src/Parallel/ResourceInfo.hpp
```diff
@@ -242,10 +247,30 @@ bool operator!=(const SingletonPack<tmpl::list<Components...>>& lhs,
...
+// that haven't defined the initialization_tags type alias (like the testing
+// framework). And even though we don't call the entry method on the GlobalCache
+// related to ResourceInfo, all the type aliases inside ResourceInfo are still
+// constructed regardless. So rather than changing every mock component to have
+// an empty type alias, we just treat no type alias as not having the tag we are
+// looking for.
+CREATE_HAS_TYPE_ALIAS(initialization_tags)
+CREATE_HAS_TYPE_ALIAS_V(initialization_tags)
+
+template <typename Component, typename Tag, bool HasInitializationTags>
```
nilsdeppe: you can just default the last argument to `bool HasInitializationTags = has_initialization_tags_v<Component>` and then not specify it everywhere else

nilsdeppe (changes requested):
Overall LGTM, a few minor suggestions. Please squash immediately :)

src/Parallel/GlobalCache.hpp
```diff
@@ -549,6 +561,7 @@ class GlobalCache : public CBase_GlobalCache<Metavariables> {
   MutableGlobalCache<Metavariables>* mutable_global_cache_{nullptr};
   mutable_global_cache_proxy_type mutable_global_cache_proxy_{};
   bool parallel_components_have_been_set_{false};
+  bool resource_info_has_been_set_{false};
```
nilsdeppe, line 564: Add to `pup` function?

nilsdeppe (commented):
LGTM, just one question! Again, if anything does need to change, just squash immediately :)

src/Parallel/GlobalCache.hpp
knelli2: 🙃 yup!

nilsdeppe (dismissed):
:tada:

tests/Unit/Parallel/Test_GlobalCache.cpp
```diff
@@ -566,11 +565,11 @@ SPECTRE_TEST_CASE("Unit.Parallel.MutableGlobalCache.NullptrConstructError",
 // --------- registration stuff below -------
 
 PUPable_def(UseCkCallbackAsCallback)
-// clang-tidy: possibly throwing constructor static storage
-// clang-tidy: false positive: redundant declaration
-PUP::able::PUP_ID Triangle::my_PUP_ID = 0;   // NOLINT
-PUP::able::PUP_ID Square::my_PUP_ID = 0;     // NOLINT
-PUP::able::PUP_ID Arthropod::my_PUP_ID = 0;  // NOLINT
+    // clang-tidy: possibly throwing constructor static storage
```
kidder: this doesn't look right

src/Parallel/ResourceInfo.hpp
```diff
@@ -426,6 +496,12 @@ struct ResourceInfo {
   /// `allocate_array` function of the array component
   const std::unordered_set<size_t>& procs_to_ignore() const;
 
+  /// Returns a `std::set<size_t>` that has all processors with elements on them
+  const std::set<size_t>& procs_with_elements() const;
```
kidder: still think this needs a better name, maybe `procs_available_for_elements` as there is no guarantee a proc has an element.

src/Parallel/ResourceInfo.hpp
```diff
@@ -426,6 +496,12 @@ struct ResourceInfo {
...
+
+  /// Returns the first proc that has elements on it.
+  size_t first_proc_with_elements() const;
```
kidder: not sure if this function is useful at all, and since it is trivial to get from the previous function, I'd just get rid of it

kidder (changes requested):
you can squash any changes

tests/Unit/Parallel/Test_GlobalCache.cpp
knelli2: I agree, but that's what my clang-format did for some reason. Not sure why. I just manually kept it the same

src/Parallel/Tags/ResourceInfo.hpp
wthrowe, line 89: Repeating an old comment since the threading got messed up and it got buried:

This tag isn't used below.  Is it needed to force it's inclusion in the input file for when this is initialized?  Please add a comment explaining.

src/Parallel/GlobalCache.hpp
```diff
@@ -660,6 +673,19 @@ void GlobalCache<Metavariables>::compute_size_for_memory_monitor(
         "the MemoryMonitor is in the component list in the metavariables.\n");
   }
 }
+
+template <typename Metavariables>
+void GlobalCache<Metavariables>::set_resource_info(
+    const Parallel::ResourceInfo<Metavariables>& resource_info) {
+  ASSERT(!resource_info_has_been_set_, "Can only set the resource info once");
```
wthrowe: `not`

wthrowe (changes requested):

wthrowe (approved):

kidder (approved):

PR #4295, opened 2022-10-13
Add `ObserveLineSegment` callback

nikwit:
## Proposed changes

This adds the `ObserveLineSegment` post_interpolation_callback which observes data on a `LineSegment`. The test re-uses parts of `Test_ObserveTimeSeriesAndSurfaceData.cpp`, particularly the MockActionTesting part.

Note that this Action could technically be used on any `InterpolationTarget` since it is never specified in a post_interpolation_callback to be a `LineSegment`. During this week's spectre call the argument was made that more structural information such as quadrature and basis should be included for other InterpolationTargets and one post_interpolation_callback should be written for each one. I agree, that e.g. for a Strahlkorper it would be good to know the basis and quadrature of the inertial coordinates onto which it is interpolated but this will not allow you to fully re-construct the points in the domain since other parameters such as the radius are only given in the input file.

Also, I think that simply outputting data on the points of any interpolation target should be possible and simple. It is ensured that the information can be re-constructed by enforcing that the `InertialCoordinates` are always written to file as well. Some Interpolation targets such as `SpecifiedPoints` would require the exact same post_interpolation_callback since they also carry no structural information.  I would therefore argue that this action should be named something like `ObserveInterpolationTarget` and be used with any target.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Callbacks/ObserveLineSegment.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Callbacks/ObserveLineSegment.hpp
@@ -1,0 +1,115 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/TagName.hpp"
+#include "DataStructures/Tensor/TensorData.hpp"
+#include "IO/Observer/ObserverComponent.hpp"
+#include "IO/Observer/Tags.hpp"
+#include "IO/Observer/VolumeActions.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Invoke.hpp"
+#include "Parallel/Local.hpp"
+#include "Parallel/Reduction.hpp"
+#include "ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp"
+#include "ParallelAlgorithms/Interpolation/Protocols/PostInterpolationCallback.hpp"
+#include "Utilities/Functional.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/PrettyType.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace intrp {
+namespace callbacks {
+
+/* \brief post_interpolation_callback that outputs requested tensors
+ * interpolated onto a LineSegment. The data is written as volume data into the
+ * `Reductions` file with Quadrature `CellCentered` and Basis
+ * `FiniteDifference`.
+ *
+ * Uses:
+ * - Metavariables
+ *   - `temporal_id`
+ * - DataBox:
+ *   - `TensorsToObserve`
+ * - GlobalCache:
+ *   - `observers::Tags::ReductionFileName`
+ *
+ * Conforms to the intrp::protocols::PostInterpolationCallback protocol
+ *
+ * For requirements on InterpolationTargetTag, see
+ * intrp::protocols::InterpolationTargetTag
+ *
+ */
+template <typename TensorsToObserve, typename InterpolationTargetTag>
+struct ObserveLineSegment
+    : tt::ConformsTo<intrp::protocols::PostInterpolationCallback> {
+  static constexpr double fill_invalid_points_with =
+      std::numeric_limits<double>::quiet_NaN();
+
+  using const_global_cache_tags =
+      tmpl::list<observers::Tags::ReductionFileName>;
+
+  template <typename DbTags, typename Metavariables, typename TemporalId>
+  static void apply(const db::DataBox<DbTags>& box,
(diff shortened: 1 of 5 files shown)


tests/Unit/ParallelAlgorithms/Interpolation/Test_ObserveLineSegment.cpp
```diff
@@ -0,0 +1,518 @@
...
+             array_indices_with_queued_simple_actions) > 0) {
+    ActionTesting::invoke_random_queued_simple_action<
+        typename metavars::component_list>(
+        make_not_null(&runner), generator,
+        array_indices_with_queued_simple_actions);
+    array_indices_with_queued_simple_actions =
+        ActionTesting::array_indices_with_queued_simple_actions<
+            typename metavars::component_list>(make_not_null(&runner));
+  }
+  // There should be four more threaded actions, so invoke them and check
```
knelli2: Four or two?

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveLineSegment.hpp
```diff
@@ -0,0 +1,114 @@
...
+  static void apply(const db::DataBox<DbTags>& box,
+                    Parallel::GlobalCache<Metavariables>& cache,
+                    const TemporalId& temporal_id) {
+    static_assert(
+        tmpl::list_contains_v<TensorsToObserve,
+                              domain::Tags::Coordinates<
+                                  Metavariables::volume_dim, Frame::Inertial>>,
+        "The inertial coordinates must always be observed when observing "
+        "tensors on a line segment. This is so the data can be re-constructed "
+        "from the reductions file alone.");
```
knelli2: Maybe say `from the output file alone` to avoid explicitly saying reduction just to make it easier in the future when we rename things.

knelli2 (changes requested):
Squash directly.

I don't think this callback can be used generally as is because you hard code the basis and quadratures and that might be different for different targets. So I'd say keep the name for now and when we decide to make everything general we can revisit this an example of how to do so.

nikwit:
I squashed in your suggestions, thanks!

Writing volume data requires writing quadrature and basis which is sensible for DG elements and some `InterpolationTarget`s but not for others. For the `LineSegment` there is essentially no structure so I just use `FiniteDifference` and `CellCentered` (there are no cells). It's essentially just a filler since there is no real basis or quadrature to the points, so we need to write the coordinates anyway. I'm just arguing that this is a fine solution for all `InterpolationTarget`s though I agree that a specific callback for an `InterpolationTarget` would be better if it makes sense to implement one for that target (currently there are none).

knelli2 (dismissed):

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveLineSegment.hpp
```diff
@@ -0,0 +1,114 @@
...
+  static void apply(const db::DataBox<DbTags>& box,
+                    Parallel::GlobalCache<Metavariables>& cache,
+                    const TemporalId& temporal_id) {
+    static_assert(
+        tmpl::list_contains_v<TensorsToObserve,
+                              domain::Tags::Coordinates<
+                                  Metavariables::volume_dim, Frame::Inertial>>,
+        "The inertial coordinates must always be observed when observing "
+        "tensors on a line segment. This is so the data can be re-constructed "
+        "from the output file alone.");
```
markscheel: I had to read this message a few times to understand what it meant (I first thought it was talking about observing grid coords instead of inertial coords).  How about:

"When observing tensors on a line segment, please include the inertial coordinates in TensorsToObserve.  This is so that the output file contains the coordinates, so then the output file is all that is needed for subsequent visualization or analysis of the output data."

markscheel (commented):
Looks good. One suggested change to an error message.  Please squash immediately.

nikwit:
I changed it to your version of the error message @markscheel, thanks for your reviews!

tests/Unit/ParallelAlgorithms/Interpolation/Test_ObserveLineSegment.cpp
```diff
@@ -0,0 +1,518 @@
...
+#include "DataStructures/Variables.hpp"
+#include "Domain/Block.hpp"
+#include "Domain/CoordinateMaps/CoordinateMap.hpp"
+#include "Domain/Creators/Disk.hpp"
+#include "Domain/Creators/Interval.hpp"
+#include "Domain/Creators/RegisterDerivedWithCharm.hpp"
+#include "Domain/Creators/Shell.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/ElementMap.hpp"
+#include "Domain/LogicalCoordinates.hpp"
```
knelli2: `Domain/LogicalCoordinates.hpp` doesn't exist. I think you want `NumericalAlgorithms/Spectral/LogicalCoordinates.hpp`.

Edit: Also you get get rid of the IWYU pragmas if you want. They aren't necessary.

knelli2 (changes requested):

knelli2 (approved):

nilsdeppe (approved):

Issue #3561, opened 2021-09-29
Remove dependence of ParallelInterpolation on higher-level libs


nilsvu:
The `ParallelInterpolation` library currently depends on a bunch of libraries like `ApparentHorizon`, that are really _using_ the interpolation framework, not the other way round. My suggestion is to disentangle these dependencies one-by-one, so we end up with a general `ParallelInterpolation` framework that other parallel algorithms can use. These are the libraries that the `ParallelInterpolation` library currently depend on, but shouldn't (not all of them are listed in the `CMakeLists.txt`):

- [x] `ApparentHorizons` (The `ApparentHorizons` lib should move to `ParallelAlgorithms/` and depend on `ParallelInterpolation`. I suggest naming it `ApparentHorizonFinder`, because later we want to factor out Strahlkorper/Ylm-related things.) Edit: partially addressed in #5267.
- [x] `Cce` and `GeneralizedHarmonic` (The `SendGhWorldtubeData` and friends should move to the CCE lib)
- [ ] `GeneralRelativity` (The `KerrHorizon` interpolation target depends on this. I'm not sure what to do with this one, maybe just leave it for now.)
- [x] `DgSubcell`. Edit: Done in #5274.
- [x] `Time`. Edit: Done in #5275.

nilsvu:
@markscheel Do you agree with this? Would you be willing to do the `ApparentHorizons` lib?

markscheel:
Here is my suggestion for ApparentHorizons:
  - [x] Move Strahlkorper, StrahlkorperFunctions, ChangeCenterOfStrahlkorper into NumericalAlgorithms/SphericalHarmonics
  - [x] Similarly, move Tags and ComputeItems that depend only on Strahlkorper to NumericalAlgorithms/SphericalHarmonics
  - [x] Create ParallelAlgorithms/ApparentHorizonFinder, and put into it
    - FastFlow
    - ObjectLabel
    - ComputeHorizonVolumeQuantities
  - [x] What to do with StrahlkorperGr, and Tags/ComputeItems that depend on both Strahlkorper and GR quantities??
  - [x] What to do with StrahlkorperInDifferentFrame?  (depends on Domain and FunctionsOfTime)??

nilsvu:
Sounds good to me!

- What do you think about moving StrahlkorperGr and friends into the GeneralRelativity library? It computes GR quantities like expansion, Ricci scalar etc.
[...]

PR #4299, opened 2022-10-17
Remove CCE dependency from Interpolation

knelli2:
## Proposed changes

Addresses the CCE part of #3561.

Moves `SendGhWorldtubeData` to CCE.

### Upgrade instructions

`src/ParallelAlgorithms/Interpolation/Callbacks/SendGhWorldtubeData.hpp` has been moved to `src/Evolution/Systems/Cce/SendGhWorldtubeData.hpp`

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nikwit:
maybe it would be more logical to move the callback into `CCE` itself? By forward declaring it, it seems like it can be used without CCE but if the executable does not have CCE as a link target, you will end up with an undefined symbol, or am I not seeing it right?

knelli2:
You're right, but that's true of any forward declaration and we use forward declarations in a lot of places. I guess it's just a preference on whether we want to keep all the interpolation callbacks in the same place or not. I don't see an obvious reason to require them to be in the same place.

Furthermore, up until now, we haven't had any callbacks that truly depended on other libs. But in theory you could write a callback that does depend upon other libs. So in that case, you'd want to store it elsewhere. I think that's enough reason for me to move it to CCE. Does that sound good @nikwit?

nikwit:
I think the reason we use forward declaration is mostly to reduce compilation time so we only include the symbols that we need rather than the entire header. Correct me if I'm wrong but the forward declaration here does not do much in terms of dependencies and I suspect just deleting the link dependency from CMake would also compile and link fine. I agree we should put it into `CCE` to actually get rid of the dependency and so that the symbol can not exist without its definition.

knelli2:
@nikwit ok it has been moved!

nikwit (approved):
Looks good, but I am not sure if we shouldn't also move it into the CCE namespace? Whaat do you think?

knelli2:
We could, but currently this isn't actually used anywhere except a test so not super important. I think I'll leave it as is for now

nilsdeppe (approved):

PR #4163, opened 2022-08-15
Introduce electron fraction to GrMhd modules

ermost:
## Proposed changes

This PR introduces the electron fraction in the GrMhd routines of the codes. Currently the electron fraction is implemented as an independent advected scalar, that is not yet coupled with the equation of state (EoS).
This will require fully tabulated EoS support to be implemented first.

### Upgrade instructions

The electron fraction needed to be added to all tests, and interfaces involved GrMhd routines. This might affect existing codes calling any GrMhd function that is not included in the development branch.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4309, opened 2022-10-21
Add missing documentations on electron fraction

yoonso0-0:
## Proposed changes

As a follow up of #4163, this PR fills some missing docs with a couple of minor fixes.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/System.hpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/System.hpp
@@ -21,14 +21,18 @@
 /// \ingroup EvolutionSystemsGroup
 /// \brief Items related to general relativistic magnetohydrodynamics (GRMHD)
 namespace grmhd {
-/// The Valencia formulation of ideal GRMHD with divergence cleaning.
+/// The Valencia formulation of ideal GRMHD with divergence cleaning coupled
+/// with electron fraction.
 ///
 /// References:
-/// - [Numerical 3+1 General Relativistic Magnetohydrodynamics: A Local
-/// Characteristic Approach](http://iopscience.iop.org/article/10.1086/498238)
-/// - [GRHydro: a new open-source general-relativistic magnetohydrodynamics code
-/// for the Einstein toolkit]
-/// (http://iopscience.iop.org/article/10.1088/0264-9381/31/1/015005)
+/// - Numerical 3+1 General Relativistic Magnetohydrodynamics: A Local
+/// Characteristic Approach \cite Anton2006
+/// - GRHydro: a new open-source general-relativistic magnetohydrodynamics code
+/// for the Einstein toolkit \cite Moesta2014
+/// - Black hole-neutron star mergers with a hot nuclear equation of state:
+/// outflow and neutrino-cooled disk for a low-mass, high-spin case
+/// \cite Deaton2013
+///
 namespace ValenciaDivClean {
 
 struct System {
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/Tags.hpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/Tags.hpp
@@ -32,7 +32,8 @@ struct TildeD : db::SimpleTag {
   using type = Scalar<DataVector>;
 };
 
-/// The densitized electron_fraction \f${\tilde Ye}\f$
+/// The densitized electron number density times the baryon mass
+/// \f$\tilde{Y}_e = {\tilde D} Y_e\f$
 struct TildeYe : db::SimpleTag {
   using type = Scalar<DataVector>;
 };
--- a/docs/References.bib
+++ b/docs/References.bib
@@ -21,6 +21,21 @@
   SLACcitation   = "%%CITATION = GR-QC/0305023;%%"
 }
 
+@article{Anton2006,
+  author    = {Luis Antón and Olindo Zanotti and Juan A. Miralles
+               and José M. Martí and José M. Ibáñez and José A. Font
+               and José A. Pons},
+  doi       = {10.1086/498238},
+  journal   = {The Astrophysical Journal},
+  month     = {jan},
+  pages     = {296},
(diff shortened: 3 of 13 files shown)


src/Evolution/Systems/GrMhd/ValenciaDivClean/System.hpp
```diff
@@ -21,14 +21,18 @@
...
+/// with electron fraction.
 ///
 /// References:
 /// - [Numerical 3+1 General Relativistic Magnetohydrodynamics: A Local
 /// Characteristic Approach](http://iopscience.iop.org/article/10.1086/498238)
 /// - [GRHydro: a new open-source general-relativistic magnetohydrodynamics code
 /// for the Einstein toolkit]
 /// (http://iopscience.iop.org/article/10.1088/0264-9381/31/1/015005)
+///
+/// FIXME : add a reference on Y_e?
```
yoonso0-0: I need a help on this. Should we add any reference on Y_e or just leave it as now?

yoonso0-0:
For all the GRMHD analytic data / solutions, I thought it'd be better update those after we have a proper EOS call for initialization.

src/Evolution/Systems/GrMhd/ValenciaDivClean/System.hpp
ermost: I'm not sure what people's preferences are as there are many references on this, most of which then focus more of the neutrino transport part:

How about these? (They don't do M1 or any sophisticated transport, since we only need to reference the electron fraction equation, but any other paper by Francois etc. would also do)

https://articles.adsabs.harvard.edu/pdf/1996A%26A...311..532R
https://arxiv.org/pdf/1304.3384.pdf

ermost:
> For all the GRMHD analytic data / solutions, I thought it'd be better update those after we have a proper EOS call for initialization.

I agree. That's forthcoming :)

src/Evolution/Systems/GrMhd/ValenciaDivClean/Tags.hpp
```diff
@@ -32,7 +32,7 @@ struct TildeD : db::SimpleTag {
   using type = Scalar<DataVector>;
 };
 
-/// The densitized electron_fraction \f${\tilde Ye}\f$
+/// The densitized electron fraction times rest mass density \f${\tilde Y}_e\f$
```
ermost: How about saying `// The densitized electron number density (including the baryon mass factor) \f${\tilde Y}_e =  {\tilde D} Y_e\f$`
since rho*Y_e / m_B is the electron number density?

ermost (changes requested):

src/Evolution/Systems/GrMhd/ValenciaDivClean/System.hpp
yoonso0-0: The second one has the GR version of equation so I'm good with it. Once we have a paper with neutrino treatment in the future maybe we can replace with that.

ermost: Sounds good, then let's take that one for now!

src/Evolution/Systems/GrMhd/ValenciaDivClean/Tags.hpp
yoonso0-0: maybe `the densitized electron number density times the baryon mass`?

ermost: that works, too!

yoonso0-0:
Pushed a fixup :)

ermost:
Thanks! Looks good to me.

src/Evolution/Systems/GrMhd/ValenciaDivClean/System.hpp
```diff
@@ -21,14 +21,19 @@
...
 ///
 /// References:
 /// - [Numerical 3+1 General Relativistic Magnetohydrodynamics: A Local
 /// Characteristic Approach](http://iopscience.iop.org/article/10.1086/498238)
 /// - [GRHydro: a new open-source general-relativistic magnetohydrodynamics code
-/// for the Einstein toolkit]
-/// (http://iopscience.iop.org/article/10.1088/0264-9381/31/1/015005)
+/// for the Einstein toolkit](http://iopscience.iop.org/article/10.1088/
+/// 0264-9381/31/1/015005)
+/// - [Black hole-neutron star mergers with a hot nuclear equation of state:
```
nilsdeppe: use `\cite` here instead of adding links directly.

nilsdeppe (changes requested):
You can squash immediately.

yoonso0-0:
Rebased and squashed

nilsdeppe (dismissed):

nilsdeppe:
What changed on the last push?

@ermost would you mind letting us know if you're happy with this PR now when you get a chance? Thanks!

ermost:
Sure, all good to go from my side. Not sure what's the proper workflow is yet, I already commented above that this looked good to me :)

nilsdeppe:
Ah sorry! I just looked at the top right under "Reviewers" where your status is still "Requested Changes". Normally we try to do an "Approve" in a review :)

ermost (approved):

ermost:
AH! Ok, I didn't know that I needed to tick that box. Did so now 👍

nilsdeppe:
Awesome, thank you! :D I should've told you, sorry!

yoonso0-0:
@ermost Thank you! @nilsdeppe Sorry, CI failed from weird timeouts so I pushed the branch again. No changes made.

nilsdeppe:
Ah okay! If you see unrelated timeouts, just ignore them :) It's possible to restart individual builds now with GitHub Actions, but I'm not sure what the necessary permissions are. We typically just ignore and merge anyway if there are unrelated timeouts :)

nilsdeppe (approved):

PR #4312, opened 2022-10-21
Make electron fraction cutoff and min value as an input file option

yoonso0-0:
## Proposed changes

Since we have TCI criteria on the minimum value of the electron fraction Y_e, `FixConservatives` action should be fixing Y_e to a value bigger than the TCI threshold. Rather than using the hard-coded value `0.0`, this PR enables to specify the cutoff and minimum value of Y_e in input files.

### Upgrade instructions

In the `FixConservatives:` option group in the input file for GRMHD executables, add following two options
```
CutoffYe: <some double value>
MinimumValueOfYe: <some double value>
```
`CutoffYe` must be bigger than or equal to `MinimumValueOfYe`.

### DIFF
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/FixConservatives.cpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/FixConservatives.cpp
@@ -99,6 +99,7 @@ namespace grmhd::ValenciaDivClean {
 FixConservatives::FixConservatives(
     const double minimum_rest_mass_density_times_lorentz_factor,
     const double rest_mass_density_times_lorentz_factor_cutoff,
+    double minimum_electron_fraction, double electron_fraction_cutoff,
     const double safety_factor_for_magnetic_field,
     const double safety_factor_for_momentum_density,
     const Options::Context& context)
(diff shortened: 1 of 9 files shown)


ermost:
Thanks for adding that in!

The cutoff can indeed be helpful, i.e. if there is something funky in the table for the lowest Ye values.

The `MinimumValueOfYe` will be set by the equation of state table. So this will be set automatically from that code. When evolving polytopes we might want to set it from the input file, but we might need to discuss this. Because either it won't have any meaning, or there will be a given Ye distribution, which one might want to advect.

nilsdeppe:
Yea, I'm not sure we need to worry about this right now. I think `0.0` is a safe lower bound, and we will need to think carefully about how we actually set a different lower bound, including during reconstruction _and_ if going below the bound actually means we should go to FD. All the other conservative fixing code does what SpEC does and is easier to understand physically. Have you looked at what SpEC does?

ermost:
I just had a look at that. To me it seems that after reconstruction they impose physicality of the thermodynamic state, which in case of the Neos table routines just resets Ye to the table bounds if you exceed them. 
I do exactly the same in my own code. 

I currently also apply a positivity preserving limiter on tilde Ye, but would think that this is likely only triggered when also tilde  D is triggered. for that, the natural lower bound is probably tilde D_min * Ye_min (both consistent with the table bounds, or atmosphere if higher than the bound).

nilsdeppe:
Ah okay, well I'm glad it's consistent between FIL and SpEC!

I agree, we will want to use a positivity-preserving flux limiter on tilde Y_e as well, and also agree with your suggested bound :)

nilsdeppe:
CI failure is unrelated random.

PR #4320, opened 2022-10-27
Factor out detail namespace in ComputeHorizonVolumeQuantities

mar-celine:
## Proposed changes

This is the first in a series of PRs to add interpolation targets for ExcisionBoundaryA and ExcisionBoundaryB similar to how targets exist for AhA and AhB. A future PR will use the
factored out namespace in a new ComputeExcisionBoundaryVolumeQuantities hpp that will be used by the target.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/TaggedContainers.hpp
+++ b/src/DataStructures/TaggedContainers.hpp
@@ -1,0 +1,80 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <type_traits>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/TempBuffer.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Requires.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TypeTraits/IsA.hpp"
+
+/// @{
+/*!
+ * \ingroup DataStructuresGroup
+ * \brief Retrieves a desired tag from data structures containing tags.
+ *
+ * \details Given multiple containers retrieve a desired tag from the first
+ * container in which it is found. The containers are searched in the order
+ * in which they are supplied (i.e. `first_vars` is checked before `vars`).
+ * An error will be emitted if the tag cannot be found in any container.
+ */
+template <typename Tag, typename... TagsOrTagsList, typename... Ts,
+          template <class...> class U,
+          Requires<(not tt::is_a_v<gsl::not_null, U<TagsOrTagsList...>>)and not(
+              ... and tt::is_a_v<gsl::not_null, std::decay_t<Ts>>)> = nullptr>
+const typename Tag::type& get(const U<TagsOrTagsList...>& first_vars,
+                              const Ts&... vars) {
+  if constexpr (tt::is_a_v<db::DataBox, U<TagsOrTagsList...>>) {
+    if constexpr (not db::tag_is_retrievable_v<Tag, U<TagsOrTagsList...>> and
+                  sizeof...(Ts) != 0) {
+      return get<Tag>(vars...);
+    } else {
+      return get<Tag>(first_vars);
+    }
+  } else {
+    if constexpr (not tmpl::list_contains_v<
+                      tmpl::flatten<tmpl::list<TagsOrTagsList...>>, Tag> and
+                  sizeof...(Ts) != 0) {
+      return get<Tag>(vars...);
+    } else {
+      return get<Tag>(first_vars);
+    }
+  }
+}
+
+template <typename Tag, typename... TagsOrTagsList, typename... Ts,
+          template <class...> class U>
+gsl::not_null<typename Tag::type*> get(
+    const gsl::not_null<U<TagsOrTagsList...>*> first_vars,
+    const gsl::not_null<Ts>... vars) {
+  if constexpr (tt::is_a_v<db::DataBox, U<TagsOrTagsList...>>) {
+    if constexpr (not db::tag_is_retrievable_v<Tag, U<TagsOrTagsList...>> and
+                  sizeof...(Ts) != 0) {
+      if constexpr (sizeof...(Ts) == 1) {
+        return make_not_null(&get<Tag>(*vars...));
+      } else {
+        return get<Tag>(vars...);
+      }
+    } else {
+      return make_not_null(&get<Tag>(*first_vars));
+    }
+  } else {
+    if constexpr (not tmpl::list_contains_v<
+                      tmpl::flatten<tmpl::list<TagsOrTagsList...>>, Tag> and
+                  sizeof...(Ts) != 0) {
+      (void)first_vars;
+      if constexpr (sizeof...(Ts) == 1) {
+        return make_not_null(&get<Tag>(*vars...));
+      } else {
+        return get<Tag>(vars...);
+      }
+    } else {
+      return make_not_null(&get<Tag>(*first_vars));
+    }
+  }
+}
+/// @}
--- a/tests/Unit/DataStructures/Test_TaggedContainers.cpp
+++ b/tests/Unit/DataStructures/Test_TaggedContainers.cpp
@@ -1,0 +1,72 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <cstddef>
+#include <string>
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/TaggedContainers.hpp"
+#include "DataStructures/Tags.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
+#include "Helpers/DataStructures/DataBox/TestHelpers.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace {
+namespace Tags {
+struct MyDouble : db::SimpleTag {
+  using type = double;
+};
+struct MyScalar : db::SimpleTag {
+  using type = Scalar<DataVector>;
+};
+struct MyDouble2 : db::SimpleTag {
+  using type = double;
+};
+}  // namespace Tags
+
+// Test that desired tags are retrieved.
+void test_tagged_containers() {
+  const auto box = db::create<db::AddSimpleTags<Tags::MyDouble>>(1.2);
+  tuples::TaggedTuple<Tags::MyDouble2> tuple(4.4);
+  Variables<tmpl::list<Tags::MyScalar>> vars(10ul, 3.2);
+
+  CHECK(get<Tags::MyDouble>(box) == 1.2);
+  CHECK(get<Tags::MyDouble2>(tuple) == 4.4);
+  CHECK(get<Tags::MyScalar>(vars).get() == 3.2);
+
+  // Note: tests do not pass if DataVector only has a single component!
+  const auto box2 = db::create<db::AddSimpleTags<Tags::MyScalar>>(
+      Scalar<DataVector>{10ul, 1.2});
+
+  CHECK(get<Tags::MyScalar>(vars, box2).get() == 3.2);
+  CHECK(get<Tags::MyScalar>(box2, vars).get() == 1.2);
+
+  CHECK(get<Tags::MyScalar>(tuple, box2, vars).get() == 1.2);
+  CHECK(get<Tags::MyScalar>(box2, tuple, vars).get() == 1.2);
+  CHECK(get<Tags::MyScalar>(box2, vars, tuple).get() == 1.2);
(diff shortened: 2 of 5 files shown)


src/ApparentHorizons/CMakeLists.txt
```diff
@@ -21,6 +21,7 @@ spectre_target_headers(
   HEADERS
   ComputeHorizonVolumeQuantities.hpp
   ComputeHorizonVolumeQuantities.tpp
+  ComputeVolumeQuantities.hpp
```
knelli2: Could you rename this file? Naming it `ComputeVolumeQuantities` gives the impression that it has to do with the actual computation of volume quantities when in reality the function just gets a tag out of a Variables or a TempBuffer. Maybe `ComputeVolumeQuantitiesDetail.hpp` to signify it's a helper or even `GetTagFromTargetOrTemp.hpp`? Not sure which is better

knelli2 (changes requested):

src/ApparentHorizons/CMakeLists.txt
kidder: Better yet move the file to something like `DataStructures/TaggedContainers.hpp` as it has absolutely nothing to do with ApparentHorizons.  @nilsdeppe you discussed this with me as a new file for a similar type of function...

src/ApparentHorizons/ComputeVolumeQuantities.hpp
```diff
@@ -0,0 +1,35 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstdint>
```
kidder: clean up the includes; you have way too many of them that are not needed

src/ApparentHorizons/ComputeVolumeQuantities.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace detail {
```
kidder: remove the detail namespace; this can be in the global namespace

src/ApparentHorizons/ComputeVolumeQuantities.hpp
```diff
@@ -0,0 +1,35 @@
...
+
+template <typename Tag, typename DestTags, typename TempTags>
+typename Tag::type& get_from_target_or_temp(
```
kidder: just name the function `get` and document what it does

src/ApparentHorizons/ComputeVolumeQuantities.hpp
```diff
@@ -0,0 +1,35 @@
...
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace detail {
+
+template <typename Tag, typename DestTags, typename TempTags>
```
kidder: rename DestTags to TagsList

src/ApparentHorizons/ComputeVolumeQuantities.hpp
```diff
@@ -0,0 +1,35 @@
...
+typename Tag::type& get_from_target_or_temp(
+    const gsl::not_null<Variables<DestTags>*> target_vars,
```
kidder: rename target_vars -> vars

kidder (changes requested):

src/ApparentHorizons/CMakeLists.txt
nilsdeppe: Yes, this implementation is a basic version of what I have... It should take an arbitrary number of containers that have a tag list as their parameters, then peel off the first container until the right container is found. In the case of a `DataBox` it should use the tag retrieveable lookup. while for other cases it's list_contains. There should also be one version that takes all the containers as `const &` and returns a `const &`, and a second that takes all the containers as `gsl::not_null` and returns a `gsl::not_null`

nilsdeppe: For completeness, here's my solution:
```cpp
 template <typename Tag, typename... TagsOrTagsList, typename... Ts,
           template <class...> class U>
 const typename Tag::type& get(const U<TagsOrTagsList...>& first_vars,
                               const Ts&... vars) {
   if constexpr (tt::is_a_v<db::DataBox, U<TagsOrTagsList...>>) {
     if constexpr (not db::tag_is_retrievable_v<U<TagsOrTagsList...>, Tag> and
                   sizeof...(Ts) != 0) {
       return get<Tag>(vars...);
     } else {
       return get<Tag>(first_vars);
     }
   } else {
     if constexpr (not tmpl::list_contains_v<
                       tmpl::flatten<tmpl::list<TagsOrTagsList...>>, Tag> and
                   sizeof...(Ts) != 0) {
       return get<Tag>(vars...);
     } else {
       return get<Tag>(first_vars);
     }
   }
 }

 template <typename Tag, typename... TagsOrTagsList, typename... Ts,
           template <class...> class U>
 gsl::not_null<typename Tag::type*> get(
     const gsl::not_null<U<TagsOrTagsList...>*> first_vars,
     const gsl::not_null<Ts*>... vars) {
   if constexpr (tt::is_a_v<db::DataBox, U<TagsOrTagsList...>>) {
     if constexpr (not db::tag_is_retrievable_v<U<TagsOrTagsList...>, Tag> and
                   sizeof...(Ts) != 0) {
       return get<Tag>(vars...);
     } else {
       return make_not_null(&get<Tag>(*first_vars));
     }
   } else {
     if constexpr (not tmpl::list_contains_v<
                       tmpl::flatten<tmpl::list<TagsOrTagsList...>>, Tag> and
                   sizeof...(Ts) != 0) {
       return get<Tag>(vars...);
     } else {
       return make_not_null(&get<Tag>(*first_vars));
     }
   }
 }
```
This should work on any tagged container that either has a tags list or a set of tags as template parameter, and also works on a DataBox where there might be base tags. I've only tested this in my production code, not written an explicit test for it yet. I would put it in the file Larry suggested, `DataStructures/TaggedContainers.hpp`

kidder (changes requested):
please add a test for the functions in `src/DataStructures/TaggedContainers.hpp`

kidder (commented):
add document the functions

knelli2:
@gsb76 Could you squash everything? With adding the new `get` function, the fixups are a bit confusing and I think everything has changed enough that it's ok so just rebase and squash. I think you have two separate commits: 1) TaggedContainers and 2) refactor ComputeHorizonVolumeQuantities, but correct me if I'm wrong.

tests/Unit/DataStructures/Test_TaggedContainers.cpp
```diff
@@ -0,0 +1,48 @@
...
+  tuples::TaggedTuple<Tags::MyDouble2> tuple(4.4);
+  Variables<tmpl::list<Tags::MyScalar>> vars(10, 3.2);
+
+  auto retrieved_value1 = get<Tags::MyDouble>(box);
+  auto retrieved_value2 = get<Tags::MyDouble2>(tuple);
+  auto retrieved_value3 = get<Tags::MyScalar>(vars);
+  CHECK(retrieved_value1 == 1.2);
+  CHECK(retrieved_value2 == 4.4);
+  CHECK(retrieved_value3.get() == 3.2);
+}
```
nilsdeppe, line 66: Can you also add:
```cpp
const auto box2 = db::create<db::AddSimpleTags<Tags::MyScalar>>(Scalar<DataVector>{1.2});
CHECK(get<Tags::MyScalar>(vars, box2).get() == 3.2);
CHECK(get<Tags::MyScalar>(box2, vars).get() == 1.2);
CHECK(get<Tags::MyScalar>(tuple, box2, vars).get() == 1.2);
CHECK(get<Tags::MyScalar>(box2, tuple, vars).get() == 1.2);
CHECK(get<Tags::MyScalar>(box2, vars, tuple).get() == 1.2);

CHECK(*get<Tags::MyDouble>(make_not_null(&box) == 1.2);
CHECK(*get<Tags::MyDouble2>(make_not_null(&tuple) == 4.4);
CHECK(get<Tags::MyScalar>(make_not_null(&box)->get() == 3.2);

CHECK(get<Tags::MyScalar>(make_not_null(&vars), make_not_null(&box2)).get() == 3.2);
CHECK(get<Tags::MyScalar>(make_not_null(&box2), make_not_null(&vars)).get() == 1.2);
// Same thing as the 3-arg variant but with `not_null` here
```

src/DataStructures/TaggedContainers.hpp
```diff
@@ -0,0 +1,80 @@
...
+ * \details Given multiple containers retrieve a desired tag from the first
+ * container in which it is found. The containers are searched in the order
+ * in which they are supplied (i.e. `first_vars` is checked before `vars`).
+ * An error will be emitted if the tag cannot be found in any container.
+ */
+template <typename Tag, typename... TagsOrTagsList, typename... Ts,
+          template <class...> class U,
+          Requires<(not tt::is_a_v<gsl::not_null, U<TagsOrTagsList...>>)and not(
+              ... and tt::is_a_v<gsl::not_null, std::decay_t<Ts>>)> = nullptr>
+const typename Tag::type& get(const U<TagsOrTagsList...>& first_vars,
```
knelli2, line 29: This seems like if the tag isn't in either, you'll get a compile-time error along the lines of `Tag not in first_vars` (because the tag isn't retrievable out of `vars...` so it goes to the second branch of the `constexpr if` in both cases), which is sort of unhelpful because the tag isn't actually in either of these. I think there should be a `static_assert` at the beginning of this function to check that the tag is in at least one container, but I don't think that's trivial to do because we don't know the type of `Ts...`.

If that's impractical to implement, I think it'll be ok, but if it's easy to give an easy error message, I think we should do that.

tests/Unit/DataStructures/Test_TaggedContainers.cpp
```diff
@@ -0,0 +1,48 @@
...
+};
+}  // namespace Tags
+
+// Test that desired tags are retrieved.
+void test_tagged_containers() {
+  auto box = db::create<db::AddSimpleTags<Tags::MyDouble>>(1.2);
+  tuples::TaggedTuple<Tags::MyDouble2> tuple(4.4);
+  Variables<tmpl::list<Tags::MyScalar>> vars(10, 3.2);
+
+  auto retrieved_value1 = get<Tags::MyDouble>(box);
```
nilsdeppe: Refactor these into `CHECK(get<...>(...)` and then test with all combinations of 1, 2, 3 args that won't result in a compiler error to make sure you get what is expected. There are a lot of code paths in the hpp that need to be hit, with a bunch of special cases.

nilsdeppe (changes requested):
Just more tests 🙃 you can squash immediately!

tests/Unit/DataStructures/Test_TaggedContainers.cpp
knelli2: Can you also add some tests where you have the same tag in multiple containers with different values so you are sure it always gets it from the first container?

knelli2 (changes requested):

src/DataStructures/TaggedContainers.hpp
nilsdeppe: I don't think it's super easy to do, and the compiler error will point to this get call. If people run into confusion a lot I think we can do the work to add a check, but it's not clear to me it's worth the effort right now. I don't expect this function to be used widely anyway

tests/Unit/DataStructures/Test_TaggedContainers.cpp
nilsdeppe: That's the first 2 `CHECK` lines in the tests I requested

knelli2: Ah ok yeah I missed that

mar-celine:
@kidder @nilsdeppe all checks have passed! :tada:

knelli2 (approved):

nilsdeppe (approved):

kidder (approved):

PR #4326, opened 2022-10-30
Fix map sizes in GH+GRMHD FD code

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/GhValenciaDivClean/Tags.hpp
+++ b/src/Evolution/Systems/GrMhd/GhValenciaDivClean/Tags.hpp
@@ -6,6 +6,10 @@
 #include "DataStructures/DataBox/Tag.hpp"
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "PointwiseFunctions/Hydro/Tags.hpp"
+#include "Utilities/TMPL.hpp"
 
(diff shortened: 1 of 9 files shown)


src/Evolution/Systems/GrMhd/GhValenciaDivClean/Tags.hpp
```diff
@@ -70,8 +70,8 @@ using primitive_grmhd_reconstruction_tags =
 /// \brief Tags sent for spacetime evolution. All tags sent are the
 /// reconstruciton+spacetime tags
 using spacetime_reconstruction_tags =
-    tmpl::list<gr::Tags::SpacetimeMetric<3>, GeneralizedHarmonic::Tags::Phi<3>,
-               GeneralizedHarmonic::Tags::Pi<3>>;
+    tmpl::list<gr::Tags::SpacetimeMetric<3>, GeneralizedHarmonic::Tags::Pi<3>,
```
kidder, line 73: why is the order switched here?

nilsdeppe: This is now the correct order. I have (or will have, lost track) a PR that factors this order into a tmpl::list so different places don't need to by synced.

nilsdeppe: Derp, that's this PR. The list with Pi first is the order in the GH system and should be the one used...

nilsdeppe: Okay... I've refactored this changed into the first commit :D

kidder (approved):

PR #4332, opened 2022-10-31
Allow assigning a subset of tags to a Variables

nilsdeppe:
## Proposed changes

Allows assigning only some tags from another container to the vars. This is helpful in GH+GRMHD where I get both theGH and the GRMHD primitive tags from the analytic solution/data and need to assign them to separate variables.

Most of this PR is just adding tests...

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/Variables.hpp
+++ b/src/DataStructures/Variables.hpp
@@ -271,24 +271,33 @@ class Variables<tmpl::list<Tags...>> {
 
   /// @{
   /// \brief Assign a subset of the `Tensor`s from another Variables or a
-  /// tuples::TaggedTuple
+  /// tuples::TaggedTuple. Any tags that aren't in both containers are
+  /// ignored.
   ///
   /// \note There is no need for an rvalue overload because we need to copy into
   /// the contiguous array anyway
-  template <typename... SubsetOfTags,
-            Requires<tmpl2::flat_all<tmpl::list_contains_v<
-                tmpl::list<Tags...>, SubsetOfTags>...>::value> = nullptr>
+  template <typename... SubsetOfTags>
   void assign_subset(const Variables<tmpl::list<SubsetOfTags...>>& vars) {
-    EXPAND_PACK_LEFT_TO_RIGHT(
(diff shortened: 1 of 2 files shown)


src/DataStructures/Variables.hpp
```diff
@@ -292,6 +292,40 @@ class Variables<tmpl::list<Tags...>> {
...
 
+  /// @{
+  /// \brief Assign a subset of the `Tensor`s from another Variables or a
+  /// tuples::TaggedTuple. Any tags that aren't in both containers are
+  /// ignored.
+  ///
+  /// \note There is no need for an rvalue overload because we need to copy into
+  /// the contiguous array anyway
+  template <typename... SubsetOfTags>
+  void assign_overlapping_subset(
```
knelli2: This may be a naive question, but it seems like this new function is just a slightly more general `assign_subset` (it doesn't require all of `SubsetOfTags...` to be in `Tags...`). Why not just get rid of the restriction in `assign_subset` and change the functionality there?

nilsdeppe: I think the original reason for having the restriction was that you'd get errors if some of the tags didn't match rather than them silently being ignored. I don't really feel strongly either way.

knelli2: I think it'd be ok to modify `assign_subset` to remove the restriction of having the tags match, but I also understand it might be nice to keep the compile time check. Up to you. If you decide to leave it as it, add some docs to `assign_subset` saying that all tags in `SubsetOfTags` must be in `Tags`.

nilsdeppe: Sounds good!

@wthrowe @kidder would you prefer a new function or just extend the existing?

kidder: I don't have a preference

nilsdeppe:
@knelli2 I just did what you suggested and removed the restriction on the existing function. Thanks for reviewing :)

knelli2 (commented):
Looks good. Rebase on develop to fix CI first.

nilsdeppe:
Rebased... 🤞

kidder (approved):

PR #4337, opened 2022-11-01
Add runtime initial data to GH

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Utilities/TypeTraits/FastPointerCast.hpp
```diff
@@ -0,0 +1,37 @@
...
+struct can_static_cast<T, U,
+                       std::void_t<decltype(static_cast<U>(std::declval<T>()))>>
+    : std::true_type {};
+
+template <typename T, typename U>
+static constexpr bool can_static_cast_v = can_static_cast<T, U>::value;
+}  // namespace detail
+
+/*!
+ * \brief `t` which is of type `T*` to a `U`. If a `static_cast<U>(t)` is
```
knelli2: Add `Cast` to the beginning of the first sentence. Needs some verbage.

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
```diff
@@ -1,24 +1,13 @@
...
-    "${LIBS_TO_LINK}"
-    )
-endfunction(add_generalized_harmonic_executable_with_horizon)
-
 function(add_generalized_harmonic_executable_without_horizon
-    INITIAL_DATA_NAME DIM INITIAL_DATA BOUNDARY_CONDITIONS LIBS_TO_LINK)
+    DIM LIBS_TO_LINK)
   add_spectre_parallel_executable(
-    "EvolveGh${INITIAL_DATA_NAME}${DIM}D"
+    "EvolveGhNoBlackHole${DIM}D"
```
knelli2: For consistency, maybe keep the name of the executable as `EvolveGeneralizedHarmonicWithoutHorizon${DIM}D` (or shorten GeneralizedHarmonic to Gh) so it's similar to the file it comes from? Even though there isn't a huge conceptual leap between "no BH" -> "without horizon", I think it's best to be consistent.

I'm not a huge fan of the "with/without horizon" naming scheme, so maybe we could rename the files instead? But just making the names of the executables the same as the current files seems like the least amount of work.

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
```diff
@@ -46,48 +35,34 @@ set(LIBS_TO_LINK
   Utilities
   )
 
-add_generalized_harmonic_executable_with_horizon(
-  KerrSchild
-  GeneralizedHarmonic::Solutions::WrappedGr<gr::Solutions::KerrSchild>
-  GeneralizedHarmonic::Solutions::WrappedGr<gr::Solutions::KerrSchild>
+add_spectre_parallel_executable(
+  EvolveGhSingleBlackHole
```
knelli2, line 39: Same comment about names here and below.

src/Evolution/Systems/GeneralizedHarmonic/BoundaryConditions/DirichletAnalytic.cpp
```diff
@@ -26,6 +51,81 @@ DirichletAnalytic<Dim>::get_clone() const {
...
+          gr::Tags::SpacetimeMetric<Dim, Frame::Inertial, DataVector>>,
+      tmpl::append<
+          GeneralizedHarmonic::Solutions::all_solutions<Dim>,
+          tmpl::conditional_t<
+              Dim == 3,
+              tmpl::append<
+                  GeneralizedHarmonic::Solutions::RelativisticEuler::
+                      all_solutions,
+                  GeneralizedHarmonic::Solutions::grmhd::all_solutions,
+                  GeneralizedHarmonic::AnalyticData::grmhd::all_analytic_data>,
```
knelli2: This list of solutions has shown up twice now (first time in GhValenciaDivCleanBase with Dim==3). Maybe factor it out somewhere so it can be reused?

src/Evolution/Systems/GeneralizedHarmonic/BoundaryConditions/DirichletAnalytic.cpp
```diff
@@ -26,6 +51,81 @@ DirichletAnalytic<Dim>::get_clone() const {
...
+    const Scalar<DataVector>& interior_gamma2, const double time) const {
+  *gamma1 = interior_gamma1;
+  *gamma2 = interior_gamma2;
+  ASSERT(analytic_prescription_ != nullptr,
+         "The analytic prescription must be set.");
+  auto boundary_values = call_with_dynamic_type<
+      tuples::TaggedTuple<
+          GeneralizedHarmonic::Tags::Pi<Dim, Frame::Inertial>,
+          GeneralizedHarmonic::Tags::Phi<Dim, Frame::Inertial>,
+          gr::Tags::SpacetimeMetric<Dim, Frame::Inertial, DataVector>>,
```
knelli2: Isn't this `system<Dim>::variables_tag::tags_list`? Same for the other places in this file.

tests/InputFiles/GeneralizedHarmonic/GaugeWave1D.yaml
```diff
@@ -39,13 +44,12 @@ DomainCreator:
...
-      LowerBoundary:  DirichletAnalytic
-      UpperBoundary:  DirichletAnalytic
-
-AnalyticSolution:
-  GaugeWave:
-    Amplitude: 0.5
-    Wavelength: 6.283185307179586
+      LowerBoundary:
+        DirichletAnalytic:
+          AnalyticPrescription: *InitialData
```
knelli2, line 49: Not a change request, but this feels like it can be broken pretty easily. Some poor soul (me) copies the ID instead of using the YAML `& -> *` syntax, then changes the ID but forgets to update the BCs and bad things happen.

Do you think there'd be a way to have DirichletAnalytic retrieve the ID somehow on its own so users don't have to worry about duplicating the ID parameters (YAML or otherwise)? If not, or it isn't trivial, don't worry about it, but I can imagine some issues arising because of this.

knelli2 (changes requested):

tests/InputFiles/GeneralizedHarmonic/GaugeWave1D.yaml
nilsdeppe: This is intentional. There is generally no reason why you should apply the same BC as the initial condition. The current develop forces this on users and it's been a bit annoying, e.g. see the Minkowski BC Geoffrey wrote. Being able to specify them separately is both reasonable and necessary. I agree it adds an avenue for bugs, but I don't see a way to prevent that while adding the flexibility. The best I've got is have the example input files use the `&` and `*` syntax :)

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
nilsdeppe: I'm happy to rename files! What you're saying is you prefer the names `SingleBlackHole` and `NoBlackHole` but you want the `hpp`s to be renamed to e.g. `EvolveGeneralizedHarmonicSingleBlackHole.hpp`?

tests/InputFiles/GeneralizedHarmonic/GaugeWave1D.yaml
knelli2: Ok that makes sense. This is alright then.

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
knelli2: Yeah how about we rename the file and the executable to `EvolveGhNoBlackHole` and `EvolveGhSingleBlackHole`. This makes the naming more consistent with the BBH which is `EvolveGhBinaryBlackHole`.

nilsdeppe:
Rebased and made changes. Thanks for the review!

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
knelli2: Sorry, I wasn't clear in my last comment. Could you also change `EvolveGeneralizedHarmonicWith(out)Horizon(Fwd).hpp` -> `EvolveGhSingle[No]BlackHole(Fwd).hpp` so the hpp file and executable have matching names?

knelli2 (changes requested):

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
nilsdeppe: I left it with `EvolveGeneralizedHarmonicSingle[No]BlackHole.hpp` since this is consistent with all other evolution systems. At some point I think we should remove the system name from these headers since they're in a system directory, but that should be done standard across all evolution systems, I think. Is that all okay with you?

nilsdeppe:
Pushed another update, sorry for missing the comment!

knelli2 (dismissed):

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicNoBlackHole.hpp
```diff
@@ -26,16 +26,15 @@
 // could be an analytic solution, analytic data, or imported numerical data.
 // Third template parameter specifies the analytic solution used when imposing
 // dirichlet boundary conditions or against which to compute error norms.
-template <size_t VolumeDim, typename InitialData, typename BoundaryConditions>
+template <size_t VolumeDim, bool UseNumericalInitialData>
```
kidder, line 25: update the comment

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
```diff
@@ -55,12 +55,11 @@
 // could be an analytic solution, analytic data, or imported numerical data.
 // Second template parameter specifies the analytic solution used when imposing
 // dirichlet boundary conditions or against which to compute error norms.
-template <typename InitialData, typename BoundaryConditions>
-struct EvolutionMetavars<3, InitialData, BoundaryConditions>
+template <size_t VolumeDim, bool UseNumericalInitialData>
```
kidder, line 54: update the comment

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicNoBlackHoleFwd.hpp
```diff
@@ -0,0 +1,15 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+
+/// \cond
+namespace evolution {
+struct NumericInitialData;
```
kidder: don't need this?

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHoleFwd.hpp
```diff
@@ -0,0 +1,15 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+
+/// \cond
+namespace evolution {
+struct NumericInitialData;
```
kidder: don't need this

kidder (changes requested):

nilsdeppe:
Rebased, squashed. I just deleted the comments since they were already wrong before I started and the template parameter names are actually clear.

kidder (approved):

PR #4344, opened 2022-11-02
Add tags and fluxes to the ForceFree evolution system

yoonso0-0:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Evolution/Systems/ForceFree/Fluxes.hpp
```diff
@@ -0,0 +1,86 @@
...
+ *   = -\frac{1}{\sqrt{-g}} n_\mu e^{\mu ijk} = \frac{1}{\sqrt{\gamma}} e^{ijk}
+ * \f}
+ *
+ * where $\epsilon^{\mu\nu\rho\sigma}$ is the Levi-Civita tensor, \f$g\f$ is the
+ * determinant of spacetime metric, \f$\gamma\f$ is the determinant of spatial
+ * metric, \f$n^\mu\f$ is the normal to spatial hypersurface. Also,
+ * \f$e^{abcd}\f$ and \f$e^{ijk}\f$ are usual antisymmetric _symbols_ (which
+ * only have the value \f$\pm 1\f$) with 4 and 3 indices, respectively, with the
+ * convention \f$e^{0123} = e^{123} = +1\f$.
+ *
```
teukolsky, line 85: We use the (sensible) convention e_{0123}=+1!

ermost: Agree with @teukolsky! 

@yoonso0-0, just so that I follow the convention here: e is just [abcd], and epsilon is the tensor, right?
So that epsilon_{0123} =  -epsilon^{0123} = +1; but e_{0123} = e^{0123} = +1 ?

src/Evolution/Systems/ForceFree/Tags.hpp
```diff
@@ -0,0 +1,190 @@
...
+ * \brief Groups option tags related to the divergence cleaning of the GRFFE
+ * system.
+ */
+struct ConstraintDampingGroup {
+  static std::string name() { return "ConstraintDamping"; }
+  static constexpr Options::String help{
+      "Options related to constraint damping"};
+  using group = ForceFreeGroup;
+};
+
```
ermost, line 134: The force-free current includes a damping term, that acts as an effective resistivity.
It might make sense to add an option for that, too. Sometimes this needs to be adjusted.
`Eta` might be a good name; it has the same dimensions as the Kappa's

src/Evolution/Systems/ForceFree/Fluxes.hpp
teukolsky: Ah! I missed the point that e was the permutation symbol and epsilon the tensor. It's best not to treat the indices of e as things that can be raised or lowered, so as to avoid confusion. A good notation is e=[\alpha \beta \gamma \delta], with [0123]=+1. Then there's no confusion: \epsilon_{\alpha\beta\gamma\beta]=1/\sqrt{\det{-g}} [\alpha\beta\gamma\delta].

yoonso0-0: @ermost correct :)  @teukolsky sorry for the confusion! I'll update the docs.

src/Evolution/Systems/ForceFree/Tags.hpp
yoonso0-0: Thanks for pointing this out!

From your paper I see `Nu` was used and in Alic's paper `SigmaB` was used. I think we can use `Eta` when we describe our equations, but would prefer some descriptive name inside the code.

yoonso0-0: something like `RelaxationParameter`? I cannot think of a good name on this, but I can put this parameter in a separate option group named like `ForceFreeCurrent:`

ermost: How about `ParallelConductivity`? Physically that parameter is the conductivity parallel to magnetic field lines.

yoonso0-0: Sounds nice :+1:

yoonso0-0:
Rebased and pushed fixups

ermost (dismissed):
fixUps look good to me

ermost (dismissed):

src/Evolution/Systems/ForceFree/Fluxes.cpp
```diff
@@ -0,0 +1,118 @@
...
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
+#include "PointwiseFunctions/GeneralRelativity/IndexManipulation.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace ForceFree {
+
+namespace detail {
+void fluxes_impl(
```
wthrowe, line 20: Change the detail namespace to an anonymous namespace.  Alternatively, combine this function with the one that calls it.

src/Evolution/Systems/ForceFree/Fluxes.cpp
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  for (LeviCivitaIterator<3> it; it; ++it) {
+    const auto& i = it[0];
+    const auto& j = it[1];
+    const auto& k = it[2];
+
+    tilde_e_flux->get(j, i) += -get(lapse) * it.sign() *
+                               tilde_b_one_form.get(k) /
+                               get(sqrt_det_spatial_metric);
```
wthrowe: Move `it.sign()` to immediately after the `-`.  This reduces the number of DataVector math operations by one.  General rule is to put all non-DataVector factors first.

Optional: the `one_form` variables are only used in the combination `lapse * one_form / sqrt_det_spatial_metric`, so those factors could be included in the precomputed value.

wthrowe (changes requested):

src/Evolution/Systems/ForceFree/Fluxes.cpp
yoonso0-0: @nilsdeppe I think I asked about this to you before, on separating `_impl` function inside a detail namespace. If I remember correctly this is for reducing memory allocations when computing the DG time derivative.

src/Evolution/Systems/ForceFree/Fluxes.cpp
yoonso0-0: I'll change the temp variable from `one_form` to the product you wrote

src/Evolution/Systems/ForceFree/Fluxes.cpp
yoonso0-0: @wthrowe GRMHD is also using this `_impl` splitting

nilsdeppe: Yes, in the GRMHD system it's done because we need to compute the fluxes and sources on different grids with FD

src/Evolution/Systems/ForceFree/Fluxes.cpp
yoonso0-0: I'll add the common factor as another temp variable

yoonso0-0: Pushed a fixup for this.

yoonso0-0:
Rebased and pushed fixup. I interchanged `Psi` and `Phi` part in the docs to match the code arguments order

yoonso0-0:
@wthrowe I pushed another fixup taking care of what's addressed in #4381

wthrowe:
Will this system use subcell?  I'm not familiar enough with it to know what typical results look like.

nilsdeppe:
Yes

wthrowe:
OK, sounds good then.  Go ahead and squash.

wthrowe (approved):

nilsdeppe (approved):

yoonso0-0:
Thanks for the review!

PR #4135, opened 2022-08-01
Switch away from Boost Newton-Raphson implementation

wthrowe:
The boost implementation [does not fall back to bisection correctly](https://github.com/boostorg/math/issues/808).  (I believe it only works properly for monotonic functions.)  This replaces it with an `rtsafe`-like one.  Since we can't use NR code directly, I've taken the decision criteria from there, but otherwise redone the implementation.  The biggest difference from `rtsafe` is this version defers constructing a bisection bracket until necessary, which for good initial guesses is likely never.  (The passed bracket is checked for validity in Debug builds.)

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4349, opened 2022-11-03
Remove newton_raphson root finder

wthrowe:
This supersedes #4135.  Pinging people who were involved in that discussion: @kidder @nilsdeppe @teukolsky.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/NumericalAlgorithms/RootFinding/NewtonRaphson.hpp
```diff
@@ -1,120 +1,18 @@
...
- * \note The parameter `digits` specifies the precision of the result in its
- * desired number of base-10 digits.
- *
- * \throws `convergence_error` if the requested precision is not met after
- * `max_iterations` iterations.
+ * For the nonlinear solver, see NonlinearSolver::newton_raphson.  We
+ * do not provide a Newton-Raphson root finder.  The Boost
+ * implementation is buggy and can return the wrong answer, and it is
+ * not clear that performance would be better than RootFinder::toms748
+ * in practice.
```
nilsdeppe, line 15: Maybe add a sentence that if the derivatives are cheap to evaluate NR is generally better, but for practical cases where the function and the derivative are the dominant cost of the root find, TOMS748 is comparable? There's some discussion on the Boost website that makes it look like NR is better, but in the caveats they explicitly state this is just because of overhead, not root finding speed per-se.

nilsdeppe (dismissed):
LGTM, one suggestion on a dox comment. Feel free to ignore if you disagree (but please tell me in that case so we don't wait to merge :) ).

src/NumericalAlgorithms/RootFinding/NewtonRaphson.hpp
wthrowe: > There's some discussion on the Boost website

Well, it's easy to be fast if you don't need to get the right answer.  (I've added some text.)

wthrowe (commented):
I was able to reproduce some of the CI failures with a local gcc build, so I've made some adjustments to fix those.  Hopefully that will also resolve the clang failures, which I was not able to reproduce.

src/NumericalAlgorithms/RootFinding/NewtonRaphson.hpp
nilsdeppe: Yes, and then advertise it 🙃 I think the new paragraph is very helpful! Thanks for adding that!

nilsdeppe (approved):
@kidder @teukolsky any requested changes or are you happy with this change?

teukolsky:
> @kidder @teukolsky any requested changes or are you happy with this change?

I'm happy

kidder (approved):

kidder:
ignoring clang-tidy errors in boost and failing mac CI build

PR #4350, opened 2022-11-04
Core-collapse Supernova Profile Readin

pajkosmi:
## Proposed changes

This change reads in initial data from a supernova progenitor file and allows the star to collapse with a fixed metric after providing user-defined polytropic EOS quantities.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -0,0 +1,457 @@
...
+ * \brief Read a massive star supernova progenitor from file.
+ */
+class ProgenitorProfile {
+ public:
+  ProgenitorProfile() = default;
+
+  ProgenitorProfile(const std::string& filename);
+  // interpolate function
+  std::array<double, 6> interpolate(double target_radius,
+                                    double target_cos_theta,  // not needed?
```
nilsdeppe: I guess decide what to do with the `target_cos_theta` parameter?

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -0,0 +1,457 @@
...
+  double maximum_radius_{std::numeric_limits<double>::signaling_NaN()};
+  double max_density_ratio_for_linear_interpolation_ = 1.0e2;
+  double polytropic_index_{std::numeric_limits<double>::signaling_NaN()};
+  size_t num_radial_points_{std::numeric_limits<size_t>::max()};
+  size_t num_angular_points_{std::numeric_limits<size_t>::max()};
+  size_t num_grid_points_{std::numeric_limits<size_t>::max()};
+
+  // Begin constants to translate units from cgs to c=G=M_Sun=(k_Boltzmann=1)=1
+  // Divide cm/s by speed_of_light_cgs_ to get into c=G=M_Sun=1 units
+  double speed_of_light_cgs_ = 2.99792458e10;
```
nilsdeppe: You can mark these `constexpr` (they are compile-time constants)

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -0,0 +1,457 @@
...
+  // converting to c=G=M_Sun=1 units; see
+  // https://iopscience.iop.org/article/10.3847/2041-8213/aaa401.
+  // G factor from
+  // https://journals.aps.org/rmp/abstract/10.1103/RevModPhys.93.025010.
+  // c2g_dens_~1.61887e-18 [cm^3/g]
+  double newtons_grav_constant_ = 6.67430e-8;
+  double c2g_dens_ = square(g_times_m_sun_) * newtons_grav_constant_ /
+                     pow<6>(speed_of_light_cgs_);
+
+  // Multiply Kelvin by c2g_temp_ to get into c=G=M_Sun=k_Boltzmann=1 units.
```
nilsdeppe, line 87: Do we actually get to set 4 constants to 1? I thought we only get to set 3 of them to one, but for EOS's you choose M_sun=1.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -0,0 +1,457 @@
...
+ * X^2 dr^2 + r^2 d\theta^2 + r^2 \sin^2\theta d\phi^2$, following
+ * \cite Oconnor2015.
+ *
+ * Units:
+ * The input file is assumed to be in cgs units and
+ * will be converted to dimensionless units upon
+ * readin, setting $c = G = M_\odot = 1$.
+ *
+ * Rotation:
+ * Once the stationary progenitor is constructed, an artificial
```
nilsdeppe, line 132: do you need a blank line here (` *`) to get a line break in the dox?

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -0,0 +1,457 @@
...
+ * \brief Evolve a stellar collapse (progenitor) profile through collapse.
+ *
+ * Read the 1D core-collapse supernova (CCSN) profile from file, assuming
+ * the following format of the data:
+ * * 1st row - number of radial points in data set, followed by 6 columns of 0s
+ * * Subsequent rows contain these column headers:
+ *   * radius, rest mass density, specific internal energy, radial velocity,
+ *     electron fraction, $X$, $\Phi$.
+ *
+ * These columns assume a metric of the form $ds^2 = -e^{2\Phi} dt^2 +
```
nilsdeppe: I'm not sure if you've looked at the rendered dox, but `\f$` is normally what we need to do to get latex to render, but maybe that's been changed.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -0,0 +1,457 @@
...
+  /// The remaining hydrodynamic primitive variables (e.g., pressure)
+  /// will be calculated based on this $K$ for $P=K\rho^{\Gamma}$.
+  struct PolytropicConstant {
+    using type = double;
+    static constexpr Options::String help = {
+        "The polytropic constant of the fluid."};
+    static type lower_bound() { return 0.; }
+  };
+
+  /// Adiabatic index of the system at readin.
```
nilsdeppe, line 206: add a `///` empty line after this and mark this line with `/// \brief Ad....`

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -0,0 +1,457 @@
...
+  /// Central angular velocity artificially assigned at readin.
+  struct CentralAngularVelocity {
+    using type = double;
+    static constexpr Options::String help = {
+        "Central angular velocity of progenitor"};
+    static type upper_bound() { return 1; }
+    static type lower_bound() { return -1; }
+  };
+
+  /// Differential rotation parameter for artificially assigned
```
nilsdeppe, line 240: Is there any guidance you can give either here or maybe in the general documentation on what are interesting values? Or maybe point to some papers that give interesting simulations/values?

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -0,0 +1,909 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
```
nilsdeppe, line 3: include the `CcsnCollapse.hpp` header first

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -0,0 +1,909 @@
...
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace grmhd::AnalyticData {
+namespace detail {
+ProgenitorProfile::ProgenitorProfile(const std::string& filename) {
+  std::ifstream prog_file(filename);
+  std::string header_line;
```
nilsdeppe, line 41: It looks like you forgot to commit your tests

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -0,0 +1,909 @@
...
+  interpolate_vars_if_necessary(vars);
+  auto spatial_velocity = make_with_value<tnsr::I<DataType, 3>>(x, 0.0);
+  // x and y components will have a radial velocity (from file),
+  // and toroidal velocity (from rotation law) component.
+
+  using std::cos;
+  using std::sin;
+  // Omega = omega0/(1 + (r/A)^2)
+  // vphi = distance_from_rotation_axis * Omega
+  auto rotational_velocity =
```
nilsdeppe: `auto` is probably not what you want. With expression templates `auto` will be the expression class of the RHS, not the resulting value. E.g. if you did `auto c = a + b;` then the type would be `blaze::AddExpression<Vector, Vector>` and then every time you use `c` it will do the math, rather than doing the math only once. The flip side is you also want to minimize (or really eliminate) allocations. So what  we often do is something like `get<2>(spatial_velocity) = /*math for rotational velocity*/;`, then use `get<2>(spatial_velocity)` where you need the rotational velocity for the x, y components, and then overwrite `2` at the end. What components/allocations you can reuse depends on the exact calculation, and sometimes then you end up with say `get<2>(spatial_velocity) *= cool_math;`, but this ultimately makes a difference in heavily multithreaded environments (i.e. everywhere nowadays...) since it eliminates system calls and so a serial bottleneck.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -0,0 +1,909 @@
...
+tuples::TaggedTuple<hydro::Tags::LorentzFactor<DataType>>
+CcsnCollapse::variables(
+    const gsl::not_null<IntermediateVariables<DataType>*> vars,
+    const tnsr::I<DataType, 3>& x,
+    tmpl::list<hydro::Tags::LorentzFactor<DataType>> /*meta*/) const {
+  interpolate_vars_if_necessary(vars);
+
+  auto lorentz_factor = make_with_value<Scalar<DataType>>(get<0>(x), 0.0);
+  // Compute spatial proper velocity, then Lorentz factor.
+  const auto spatial_metric =
```
nilsdeppe, line 623: just FYI, here there are allocations too, but we have figured out a good way to fix this (yet) with analytic solutions/data, and it's a low priority since it's _only_ in the analytic prescriptions. So leave as-is :)

nilsdeppe (changes requested):
Overall looks good! I think you forgot to commit the tests.

Just in case you haven't had a chance to look, we have https://spectre-code.org/contributing_to_spectre.html (everything you've done so far is perfect!). The next steps I would suggest are:
1. squash all the current commits into one
2. add new fixup commits addressing commits. Please let the review click `Resolve`. I like to use the :+1: reaction to track which reviewer comments I've addressed.
3. push fixups (and tests), then ping me and I'll look at the fixups and test.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
pajkosmi: I have decided to keep it, in case we map in multiD progenitors in the future.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
pajkosmi: We do set Msun to 1, independent of EOS. This helps save us having to redimensionalize a different way for differently formulated EOSs (e.g., tabulated vs hybrid).  Note made in code comments.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
pajkosmi: Appropriate references have been added.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
pajkosmi: Yes, without the extra line it appears as a single paragraph (at least in Google Chrome).

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
pajkosmi: The `$` rendered for me, but I switched to `\f$` for consistency.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
pajkosmi: FYI I have addressed the changes, but need to figure out why the CCSN test fails on some clang/gcc builds.  I will squash, rebase and let you know once I track down the crash.

nilsdeppe:
My best guess for  the failure is somewhere you do `size_t - number` and that goes below zero and then it wraps and you try to do a gigantic allocation. Not sure why only on Clang.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -576,23 +573,34 @@ CcsnCollapse::variables(
...
   // vphi = distance_from_rotation_axis * Omega
-  auto rotational_velocity =
+  // Temporarily store rotational velocity in z velocity component;
+  // will overwrite at the end.
+  get<2>(spatial_velocity) =
       vars->radius * vars->sin_theta * central_angular_velocity_ /
       (1.0 + square(vars->radius * inv_diff_rot_parameter_));
 
+  // speed of light check
+  if (UNLIKELY(max(get<2>(spatial_velocity)) > 1.0)) {
```
nilsdeppe: Should this be an `ASSERT` or an `ERROR`? The reason I ask is `ERROR` is meant to deal with things like input file problems, while `ASSERT` is more meant to deal with code bugs. `ASSERT` checks disappear in release mode and increase performance, while `ERROR`s don't and so add a performance penalty. I _think_ this seems like it should be an `ASSERT` but up to you!

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
nilsdeppe: Sounds good on both fronts!

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
nilsdeppe: Huh, neat! Thanks for checking, letting me know, and for making the consistency change :D

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
nilsdeppe: ping on the extra line

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
nilsdeppe: ping here

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -0,0 +1,472 @@
...
+  /// simplified CCSN models.
+  struct AdiabaticIndex {
+    using type = double;
+    static constexpr Options::String help = {
+        "The adiabatic index that will trigger collapse."};
+    static type lower_bound() { return 1.0; }
+  };
+
+  /// Central angular velocity artificially assigned at readin.
+  /// Currently limited by the angular velocity of an object rotating in a
```
nilsdeppe, line 222: `///` here too

nilsdeppe (commented):
LGTM, just a matter of figuring out the test failure

pajkosmi:
@nilsdeppe I fleshed out the test coverage and addressed your comments.  The only failing test was `Unit.Evolution.Systems.GeneralizedHarmonic.DuDt` which I didn't modify.

tests/Unit/PointwiseFunctions/AnalyticData/GrMhd/Test_CcsnCollapse.cpp
```diff
@@ -0,0 +1,250 @@
...
+          ->get_clone();
+
+  const auto deserialized_option_solution =
+      serialize_and_deserialize(option_solution);
+  const auto& ccsn_progenitor =
+      dynamic_cast<const grmhd::AnalyticData::CcsnCollapse&>(
+          *deserialized_option_solution);
+
+#ifdef SPECTRE_DEBUG
+  const std::unique_ptr<evolution::initial_data::InitialData>
```
nilsdeppe: I'm missing something in terms of the code inside this `ifdef`. I don't see a `CHECK` or `CHECK_THROWS_WITH`. Is one of those supposed to be there? Oh! It's used below on 188. Could you move the create, etc. down there so it's clearer how it's used?

tests/Unit/PointwiseFunctions/AnalyticData/GrMhd/Test_CcsnCollapse.cpp
```diff
@@ -0,0 +1,250 @@
...
+  // Relevant metric variables
+  const auto& spatial_metric =
+      get<gr::Tags::SpatialMetric<3, Frame::Inertial, DataVector>>(vars);
+  const auto& sqrt_det_spatial_metric_analytic =
+      get<gr::Tags::SqrtDetSpatialMetric<DataVector>>(vars);
+  const auto& det_spatial_metric_numeric =
+      determinant_and_inverse(spatial_metric).first;
+
+  // Metric must always be greater or equal to 1.0 (i.e. flat space)
+  CHECK(abs(min(get(sqrt_det_spatial_metric_analytic)) - 1.0) >= 1.0e-15);
```
nilsdeppe: These "approximate" things can be check with:
```cpp
CHECK(min(get(sqrt_det_spatial_metric_analytic)) approx(1.0));
```
same with the 2 below. We also have `CHECK_ITERABLE_APPROX` that can be applied over `DataVector`s, etc. You can also use custom `approx` variants that have larger scales (e.g. 1.0e-9 instead of 1.0e-14)

nilsdeppe (changes requested):
LGTM, couple minor things. Please squash into one commit and have the commit message be something like:
```
Add analytic data for CCSN profile read from disk

# more info here if you have it
```
First line should be <= 60 chars, 2nd line blank, then additional details if they need to be there. We prefer verby commits, e.g. `Add, Remove, Fix`, etc.

nilsdeppe (approved):

PR #4353, opened 2022-11-04
Pin macOS python version on CI

knelli2:
## Proposed changes

Pin version to 3.10

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -693,6 +693,10 @@ ${{ env.CACHE_KEY_SUFFIX }}"
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3.0.1
+      - name: Force Python 3.10
+        uses: actions/setup-python@v4
+        with:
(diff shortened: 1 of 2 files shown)


nilsvu:
Oh sorry I haven't seen this PR before opening #4356. Let's see how that one does :)

knelli2:
@nilsvu No worries! In my testing I wound up also having to use `actions/setup-python@v4`, but I found that I also had to explicitly specify the python executable in cmake to be 3.10 like I did here, otherwise it would still find 3.11

knelli2:
Ok got everything to pass now. @nilsvu If you're happy with this solution, shall we just use this?

nilsvu:
I think I found the root cause of the problem. See #4356.

knelli2:
Closing because this issue is fixed by #4356.

nilsvu:
Thanks for figuring this out together 👍

PR #4357, opened 2022-11-07
Gh mesh velocity damping

geoffrey4444:
## Proposed changes

Adds previously unpublished constraint damping terms to the generalized harmonic system. These terms are included in SpEC. This PR only adds the constraint-damping mesh-velocity terms; other mesh-velocity corrections, universally applicable to any system, are already implemented elsewhere in spectre.

### Upgrade instructions

The generalized-harmonic time derivative and characteristic speeds now take additional parameters because of the newly introduced dependence on the mesh velocity.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/BoundaryCorrections/UpwindPenalty.cpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/BoundaryCorrections/UpwindPenalty.cpp
@@ -83,7 +83,8 @@ double UpwindPenalty<Dim>::dg_package_data(
   }
 
   if (normal_dot_mesh_velocity.has_value()) {
-    get<0>(*packaged_char_speeds) -= get(*normal_dot_mesh_velocity);
+    get<0>(*packaged_char_speeds) -=
+        get(*normal_dot_mesh_velocity) * (1. + get(constraint_gamma1));
     get<1>(*packaged_char_speeds) -= get(*normal_dot_mesh_velocity);
     get<2>(*packaged_char_speeds) -= get(*normal_dot_mesh_velocity);
     get<3>(*packaged_char_speeds) -= get(*normal_dot_mesh_velocity);
--- a/src/Evolution/Systems/GeneralizedHarmonic/Characteristics.cpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/Characteristics.cpp
@@ -5,12 +5,14 @@
 
 #include <algorithm>  // IWYU pragma: keep
 #include <array>
+#include <optional>
 
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/EagerMath/DotProduct.hpp"
 #include "DataStructures/Tensor/EagerMath/Magnitude.hpp"  // IWYU pragma: keep
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "DataStructures/Variables.hpp"  // IWYU pragma: keep
+#include "Domain/TagsTimeDependent.hpp"
 #include "PointwiseFunctions/GeneralRelativity/IndexManipulation.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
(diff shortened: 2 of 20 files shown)


wthrowe:
PR upgrade instructions have to be between the two "UPGRADE INSTRUCTIONS" lines in order to be included in the release notes.  I've fixed it for you here.

src/Evolution/Systems/GeneralizedHarmonic/BoundaryCorrections/UpwindPenalty.cpp
```diff
@@ -65,22 +66,13 @@ double UpwindPenalty<Dim>::dg_package_data(
...
-    dot_product(make_not_null(&shift_dot_normal), shift, normal_covector);
-    get(shift_dot_normal) *= -1.0;
-    // the metric mode speed is the zero speed times (1 + gamma_1)
-    get<0>(*packaged_char_speeds) =
-        (1.0 + get(constraint_gamma1)) * get(shift_dot_normal);
-    // 2 = plus, 3 = minus
-    get<2>(*packaged_char_speeds) = get(lapse) + get(shift_dot_normal);
-    get<3>(*packaged_char_speeds) = -get(lapse) + get(shift_dot_normal);
-  }
+  const std::array<DataVector, 4> char_speeds =
```
nilsdeppe: Please use a `not_null` version here. This adds a _huge_ about of memory allocations and copies. The code was intentionally written without using the char speeds function because there's no way to have it with zero memory allocations without a significant rewrite.

src/Evolution/Systems/GeneralizedHarmonic/Characteristics.cpp
```diff
@@ -25,10 +27,15 @@ void characteristic_speeds(
...
     const tnsr::I<DataVector, Dim, Frame>& shift,
-    const tnsr::i<DataVector, Dim, Frame>& unit_normal_one_form) {
+    const tnsr::i<DataVector, Dim, Frame>& unit_normal_one_form,
+    const std::optional<tnsr::I<DataVector, Dim, Frame>>& mesh_velocity) {
   const auto shift_dot_normal = get(dot_product(shift, unit_normal_one_form));
   (*char_speeds)[0] =
       -(1. + get(gamma_1)) * shift_dot_normal;  // lambda(VSpacetimeMetric)
+  if (mesh_velocity.has_value()) {
+    (*char_speeds)[0] -=
+        get(gamma_1) * get(dot_product((*mesh_velocity), unit_normal_one_form));
```
nilsdeppe, line 37: if you're going to be using this function in the UpwindPenalty, then you need to get rid of memory allocations like the one generation by `dot_product` here and on line 32 with `shift_dot_normal`. Because this function gets called once per face per element memory allocations end up having a non-negligible impact on performance.

nilsdeppe (changes requested):

knelli2:
Can confirm @nilsdeppe's comments. I just tried this out on my own branch to see if it worked and it's painfully slow compared to before.

src/Evolution/Systems/GeneralizedHarmonic/BoundaryCorrections/UpwindPenalty.cpp
geoffrey4444: I didn't appreciate the performance implications; instead of making this change, I'm just going to revert to the way it was and subtract the damping correction when the other mesh velocity terms are subtracted

src/Evolution/Systems/GeneralizedHarmonic/Characteristics.cpp
geoffrey4444: Good point that I hadn't considered. It's much easier to just leave it the way it was and add the damping mesh velocity correction to upwind penalty separate from the free function, so that's the way I think I should go, unless anyone objects

geoffrey4444:
@nilsdeppe @knelli2 Thanks for the feedback! I just force-pushed an update that no longer calls characteristic_speeds() from UpwindPenalty. I didn't push as a fixup because I had to rebase anyway to avoid conflicts...I hope that's not too inconvenient for you to review

src/Evolution/Systems/GeneralizedHarmonic/Characteristics.cpp
nilsdeppe: That's good with me!

nilsdeppe (approved):
LGTM!

knelli2 (approved):
Ok just tried this out again and I don't notice a decrease in speed from runs without these changes. Looks good!

Note: I believe the super slow speed I mentioned in my previous comment was due to bad nodes on wheeler, not the code being slow. Regardless, less allocations is better.

nilsdeppe:
I'll merge tomorrow after we've put out the Nov. release

knelli2:
I should also note that this didn't change the large constraint violations I've been seeing in my BBH runs.

geoffrey4444:
I also have not seen any improvement in the constraints in my BBH test runs with the new terms, sadly. But for consistency with spec, I'm still glad to have them.

PR #4358, opened 2022-11-08
Add error-based slab-size adjustment

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/InputFiles/NewtonianEuler/RiemannProblem1D.yaml
+++ b/tests/InputFiles/NewtonianEuler/RiemannProblem1D.yaml
@@ -59,10 +59,39 @@ AnalyticSolution:
     PressureStarTol: 1e-9
 
 EventsAndTriggers:
-  - - Slabs:
-        Specified:
-          Values: [10]
+  - - TimeCompares:
+        Comparison: GreaterThanOrEqualTo
+        Value: 0.2
     - - Completion
+  - - Always
+    - - ChangeSlabSize:
+          # Smaller values for DelayChange make the error control more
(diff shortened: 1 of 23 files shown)


tests/InputFiles/NewtonianEuler/RiemannProblem1D.yaml
```diff
@@ -63,6 +63,39 @@ EventsAndTriggers:
         Specified:
           Values: [10]
     - - Completion
+  # - - TimeCompares:
```
nilsdeppe: Question: is it reasonable to specify a final time rather than number of steps, even it's quite a bit shorter than the final time of the actual problem? Just so that people only need to specify the final time they want. I guess the nice thing about specifying slabs is we know in the GTS case that no matter what adaptivity does we will have a specific number of steps... :shrug:

nilsdeppe (dismissed):
LGTM, just one question. If you decide to change anything, please squash, otherwise just feel free to merge :)

tests/InputFiles/NewtonianEuler/RiemannProblem1D.yaml
wthrowe: Turns out that at the low resolution of the test the whole evolution to 0.2 only takes 60 steps, which is probably fast enough to just run.  I left the LTS one with a fixed slab number because that's a reasonable condition if LTS is being used without varying the slab size.

nilsdeppe (approved):

PR #4371, opened 2022-11-14
Raise SIGTRAP in Blaze when in debug mode

nilsdeppe:
## Proposed changes

Since the option parser uses exceptions quite a bit, using 'catch throw' in GDB is not the best experience. Blaze has support for custom error handling, so we will now raise SIGTRAP before throwing. This is analogous to what we do with our own errors. It also disables error checks in Blaze in release mode, consistent with how we disable all error checks in SpECTRE in release mode.

@nikwit this should fix the annoyance you were facing while debugging.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupBlaze.cmake
+++ b/cmake/SetupBlaze.cmake
@@ -119,6 +119,31 @@ target_compile_definitions(Blaze
   BLAZE_USE_ALWAYS_INLINE=${_BLAZE_USE_ALWAYS_INLINE}
   )
 
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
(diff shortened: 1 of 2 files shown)


wthrowe (approved):
Tested locally in debug mode: program stops on a blaze error under gdb and I get an exception if not under gdb.

nilsdeppe:
CI failures are random unrelated.

nikwit:
Also works for me, thanks :)

nilsdeppe:
Great, glad to hear it! Thanks for reporting the issue and testing the fix! :D

PR #4373, opened 2022-11-14
Streamline ElementVolumeData

nilsvu:
## Proposed changes

Preparation for upcoming regridding PRs.

- No need to have a separate `ExtentsAndTensorVolumeData` struct anymore. Delete in favor of `ElementVolumeData`.
- Add a convenience constructor to `ElementVolumeData` that takes an ElementId and a Mesh. The ElementId is used to get the grid name (reducing the chance of non-standard grid names in H5 files, allowing code to rely on the format of the grid name). The mesh is used to get the extents, basis, and quadrature.
- Minor cleanups in code that calls `ContributeVolumeData`.

### Upgrade instructions

- `TensorData.*hpp` (with `TensorComponent` and `ElementVolumeData`) has moved to `IO/H5/`.
- The constructor of `ElementVolumeData` has changed. The order of arguments is now: element_name, components, extents, basis, quadrature. There's also a convenience constructor that takes an ElementId, components, and a mesh.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/DataStructures/Tensor/TensorData.hpp
```diff
@@ -45,49 +50,46 @@ bool operator!=(const TensorComponent& lhs, const TensorComponent& rhs);
...
+  /// Spectral::Quadrature in every dimension of the grid
   std::vector<Spectral::Quadrature> quadrature{};
-  std::string element_name{};
 };
 
 bool operator==(const ElementVolumeData& lhs, const ElementVolumeData& rhs);
 bool operator!=(const ElementVolumeData& lhs, const ElementVolumeData& rhs);
+
+template <size_t Dim>
+ElementVolumeData::ElementVolumeData(const ElementId<Dim>& element_id,
```
knelli2: Why not put this in the cpp file along with the other constructor?

knelli2 (changes requested):
Just one thing. You can squash immediately.

src/DataStructures/Tensor/TensorData.hpp
nilsvu: We can't have DataStructures depend on Domain (where ElementId is), so I just forward declare here. I guess we could consider moving this TensorData file around, but it's not obvious to me where exactly, so I think this is an ok solution for now.

nilsvu:
Rebased on #4280

src/DataStructures/Tensor/TensorData.hpp
knelli2: Ahh I see. This is ok then.

There doesn't seem to be anything super special about `ElementVolumeData` that it has to be in DataStructures. `TensorComponent` I think should stay. But maybe `ElementVolumeData` can be moved somewhere else so it can depend on Domain? However, that's beyond the scope of this PR though.

knelli2 (dismissed):

src/DataStructures/Tensor/TensorData.hpp
nilsdeppe: Yea, DataStructures can't depend on Domain. However, adding this dependency in a header file already creates that dependency, right?

nilsvu: The forward declaration is fine, dependency-wise. The calling code has to a include DomainStructure/ElementId.hpp, which is also fine I think.

nilsdeppe: That's still a dependency, though. You need to have `INTERFACE Domain` in CMake. It won't break because of cyclic dependencies at the shared lib level, but in general I think we've been avoiding that and the cyclic dependency detection CMake code I wrote counts those as "bad"

nilsvu: True. Technically I don't have to add the `INTERFACE Domain` in CMake to get things to compile (though it would be the correct thing to do). I don't see a great solution because I want to make sure that get_output(element_id) is used for grid names at a very low level so we can rely on it when reading volume files back in. Also I want to get extents/basis/quadrature from a Mesh because it's just annoying to deal with otherwise. All I can think of is put a factory method into `Observer`. It's not as low-level as I want it to be, but a compromise?

nilsvu: Actually we could move ElementVolumeData to H5 and depend H5 on DomainStructure (not Domain). Thoughts?

knelli2: If you can get that to work, I think it makes sense. Our volume H5 files have to know things about the domain structure regardless, so it follows that the library would depend on DomainStructure.

Edit: After a quick glance at dependencies of DomainStructure, I don't think it'll introduce any new cyclic dependencies.

nilsdeppe: If `Domain` depends on H5 then that won't work. I'm not sure that's the case, though. Generally, `DomainStructure` seems like a reasonable place to have the serialized volume data associated with an element.

nilsvu: My suggestion: Move `TensorData.*pp` to `H5` (that's where it should be anyway I think). Then depend H5 on DomainStructure. Gives us the option of working with things like ElementId more directly in H5 volume data later, and DomainStructure will most likely never need to depend on H5. Thoughts?

FYI, `Domain` depends on lots of things, including `H5` through `ReadSpecPiecewisePolynomial`. Whether or not that's the right thing to do is a question for another time :) `DomainStructure` is just a collection of low-level data structures with very few dependencies.

nilsdeppe: Yep, that sounds good to me!

Yea... The ReadSpec... should not be there in my opinion...

nilsvu:
@knelli2 @nilsdeppe moved TensorData.*pp to IO/H5/. Ready to review.

knelli2 (approved):

nilsdeppe (approved):

Issue #3937, opened 2022-04-25
Regridding


nilsvu:
We need to interpolate data from one domain to another, e.g. for initial data solved on another domain, or to transition from the inspiral domain to the ringdown domain. I believe this is mostly a matter of iterating over all elements of both domains in some reasonable way. The only unsolved question I have is:

- How do we keep track of the block maps in numerical (initial) data saved to disk?

We could require the domain is specified in the input file, but that may get out of sync when the code is updated. Maybe there's no way around that. Another option would be to store the domain in the H5 files, e.g. serialized with PUP. That also needs the capability to deserialize data that may have been written by an older version of the code (do we have that?). Either way, we need to be very careful and put in lots of checks to make sure the domain hasn't changed in some subtle ways, so the data that's loaded in isn't reinterpreted on a slightly different domain. Checking the inertial coordinates are the same as written in the file might be sufficient.

nilsvu:
@geoffrey4444 @sxs-collaboration/spectre-core-devs some notes for next Thursday

nilsdeppe:
Notes from the call on Thursday, Apr. 28, 2022:

Initial data interpolation:
- Loop over each element on each node then each element requests the
  interpolated data at its grid points. We will not be able to hold the entire
  initial data computational domain on a single node most likely.
- How do you do the interpolation? In particular, we have Block maps in the
  initial data that we somehow need to know.
  ND: Could we do 3d Lagrangian interpolation?
  Saul: Worried about losing accuracy
  Mark: How do we handle in evolution where functions of time, grid frame
        change?
  Need to store Black maps, functions of time, excision spheres. Can we do this
  with PUP? Probably need some sort of versioning?
  How to version in PUP? Answer: serialize an integer version number at the
      beginning of pup function, then during read in need to select accordingly.
  What if Charm++ changes the way they serialize code? Answer: Unlikely to
      change, but not much we can do about it. Ignore for now.
- How to encode the maps into HDF5 files:
  Add a read_write-type function to `CoordinateMap(Base)` that takes a
  `PUP::er` as its argument. Stores an integer for the source frame, target
  frame, dimension, and its a unique integer for the combination of maps it
  holds. This unique integer is stored in a typelist of all the maps that the
  domain creators use. The list would be something like
  ```
  tmpl::list<tmpl::pair<tmpl::list<Affine3D, Wedge<3>>, 0>,
             tmpl::pair<tmpl::list<Shape, Rotation>, 1>,  ...>;
[...]

PR #4378, opened 2022-11-15
Write domain into H5 volume files

nilsvu:
## Proposed changes

Reconstructing the domain from volume data files is needed for accurate interpolation onto another set of points (see #3937).

- [x] The first commit is #4376 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/H5/VolumeData.cpp
+++ b/src/IO/H5/VolumeData.cpp
@@ -8,6 +8,7 @@
 #include <boost/iterator/transform_iterator.hpp>
 #include <hdf5.h>
 #include <memory>
+#include <optional>
 #include <ostream>
 #include <string>
 #include <vector>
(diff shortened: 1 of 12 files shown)


src/IO/H5/VolumeData.cpp
```diff
@@ -348,6 +350,11 @@ void VolumeData::write_volume_data(
     h5::write_data(observation_group.id(), pole_connectivity,
                    {pole_connectivity.size()}, "pole_connectivity");
   }
+  // Write the serialized domain
+  if (serialized_domain.has_value()) {
+    h5::write_to_attribute(observation_group.id(), "domain",
```
nilsdeppe: attributes have to be pretty small, I think under 64KB. Are we sure we won't reach that for larger domains? E.g. if the functions of time are written we will probably exceed that with the Shape map. And I think it's the _total_ size of all attributes together needs to be under 64KB. Why not just write this into a dataset instead?

nilsdeppe (changes requested):
Overall LGTM, I just think we should write the domain as a dataset to avoid the size limits of attributes. Please squash immediately :)

nilsdeppe (dismissed):

nilsvu:
ready for review now! First commit is #4388.

nilsdeppe (approved):

PR #4381, opened 2022-11-15
Move functions into anonymous namespaces and similar

wthrowe:
This is based on gcc's `-Wmissing-declarations`, which is for catching global functions that are defined without a declaration in the header (or anywhere else, but declarations are rarely other places).  clang also accepts the flag, but it doesn't seem to work, at least in a simple test.

I've tried to split things up by type of change, and also to separate out anything I could imagine people might object to.

This does not enable the warning in our build system or silence the warning in two places where it is expected, but I can do that if people like the idea.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/EnableWarnings.cmake
+++ b/cmake/EnableWarnings.cmake
@@ -20,6 +20,7 @@ if(${ENABLE_WARNINGS})
 -Wformat-y2k;\
 -Wformat=2;\
 -Winvalid-pch;\
+-Wmissing-declarations;\
 -Wmissing-field-initializers;\
 -Wmissing-format-attribute;\
 -Wmissing-include-dirs;\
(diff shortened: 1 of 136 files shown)


nilsdeppe:
I haven't looked through the code, but this seems like a useful warning to enable :) It encourages good practice that follows our guidelines so I'm in favor of having the warning!

wthrowe:
Added a commit enabling it.  Easy enough to drop if it causes problems.

Looks like clang implemented the wrong warning when they copied the gcc flag: https://github.com/llvm/llvm-project/issues/16660

wthrowe:
I'll get the python stuff fixed tomorrow.

nilsdeppe (commented):
LGTM. I can take a look at the remaining changes for the python files when you've got them. Thanks for cleaning this up! :D

wthrowe:
Rebased on develop, put some stuff in ParallelInfo.cpp in an anonymous namespace (squashed into "Put helper functions in anonymous namespaces"), and added a commit adding headers for the python bindings.

nilsvu (approved):
❤️

nilsdeppe (approved):

PR #4391, opened 2022-11-18
Avoid roundoff error in boost odeint in QuatFoT

knelli2:
## Proposed changes

Internally, boost does a `t - t0 < numeric_limits::epsilon` which doesn't account for relative error, forcing us to deal with it.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
+++ b/src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
@@ -90,13 +90,24 @@ void QuaternionFunctionOfTime<MaxDeriv>::solve_quaternion_ode(
     const gsl::not_null<boost::math::quaternion<double>*>
         quaternion_to_integrate,
     const double t0, const double t) const {
+  // Boost is stupid and assumes the times it's integrating between are order
+  // unity. So if t or t0 is > 10, then you can't take t-t0 to machine precision
+  // which causes roundoff error within boost. It was found experimentally that
+  // an infinite loop *within boost* is possible if such a condition is met.
+  // Thus, we rescale the times and the RHS by this factor to ensure the times
+  // are of order unity to avoid this infinite loop. The resulting quaternion is
+  // left unchanged.
+  const double factor = std::max(1.0, std::max(t0, t));
+
   // lambda that stores the internals of the ode
   const auto quaternion_ode_system =
-      [this](const boost::math::quaternion<double>& state,
-             boost::math::quaternion<double>& dt_state, const double time) {
-        const boost::math::quaternion<double> omega =
-            datavector_to_quaternion(angle_f_of_t_.func_and_deriv(time)[1]);
-        dt_state = 0.5 * state * omega;
+      [this, &factor](const boost::math::quaternion<double>& state,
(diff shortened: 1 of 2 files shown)


src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -90,13 +90,24 @@ void QuaternionFunctionOfTime<MaxDeriv>::solve_quaternion_ode(
...
         quaternion_to_integrate,
     const double t0, const double t) const {
+  // Boost is stupid and assumes the times it's integrating between are order
+  // unity. So if t or t0 is > 10, then you can't take t-t0 to machine precision
+  // which causes roundoff error within boost. It was found experimentally that
+  // an infinite loop *within boost* is possible if such a condition is met.
+  // Thus, we rescale the times and the RHS by this factor to ensure the times
+  // are of order unity to avoid this infinite loop. The resulting quaternion is
+  // left unchanged.
+  const double factor = std::max(1.0, std::max(t0, t));
```
wthrowe, line 100: Optional: `std::max({1.0, t0, t})`

I assume because this is control system stuff we don't have to worry about either of those being negative.

wthrowe (approved):
I looked at the Boost code.  It's definitely doing sketchy things with roundoff, although I can't immediately spot how it leads to a loop.

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
nilsdeppe: Do the initializer-lists do a memory allocation in a case like this?

wthrowe: From [dcl.init.list]: "An object of type `std::initializer_list<E>` is constructed from an initializer list as if the implementation generated and materialized a prvalue of type 'array of N `const E`'...and the `std::initializer_list<E>` object is constructed to refer to that array."  From [initializer_list.syn]: "An object of type `initializer_list<E>` provides access to an array of objects of type `const E` [Note: A pair of pointers or a pointer plus a length would be obvious representations for `initializer_list`...]"

So `std::initializer_list` is basically a view into an array (with the usual "as-if" rule caveat), so it never allocates memory itself (although its contents are copied and become const, which will have whatever side-effects that has normally).

wthrowe:
If you have a simple way to reproduce the bug you should also file an issue with Boost.

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
nilsdeppe: Thanks!

knelli2: I think I'll leave it as is. And @wthrowe correct, control systems should only be used in forward-in-time evolutions, so they will be positive.

nilsdeppe (approved):

nilsdeppe:
CI failure is GH DuDt

PR #4394, opened 2022-11-18
Update sphere domain creator

knelli2:
## Proposed changes

Does two things to the sphere domain creator:

1. Adds time dependence
2. Uses a BulgedCube map for the inner cube. If the inner cube sphericity is exactly 0, reverts to using an Equiangular or Affine map to avoid the root find in the BulgedCube.

In my testing of scalar wave, I wanted to be able to run on a sphere that was rotating which is why I added the time dependence. Then I noticed my constraint violations were (understandably) larger around the central cube boundary so I wanted to see if smoothing out the boundary helped at all which is why I opted for a BulgedCube as the central block.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/Sphere.cpp
+++ b/src/Domain/Creators/Sphere.cpp
@@ -10,6 +10,7 @@
 #include "Domain/BoundaryConditions/None.hpp"
 #include "Domain/BoundaryConditions/Periodic.hpp"
 #include "Domain/CoordinateMaps/Affine.hpp"
+#include "Domain/CoordinateMaps/BulgedCube.hpp"
 #include "Domain/CoordinateMaps/CoordinateMap.hpp"
 #include "Domain/CoordinateMaps/CoordinateMap.tpp"
 #include "Domain/CoordinateMaps/Equiangular.hpp"
@@ -22,6 +23,7 @@
 #include "Domain/DomainHelpers.hpp"
 #include "Domain/Structure/BlockNeighbor.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/GetOutput.hpp"
 #include "Utilities/MakeArray.hpp"
 
 namespace Frame {
@@ -32,7 +34,7 @@ struct BlockLogical;
 namespace domain::creators {
 Sphere::Sphere(
     typename InnerRadius::type inner_radius,
-    typename OuterRadius::type outer_radius,
+    typename OuterRadius::type outer_radius, const double inner_cube_sphericity,
     typename InitialRefinement::type initial_refinement,
     typename InitialGridPoints::type initial_number_of_grid_points,
     typename UseEquiangularMap::type use_equiangular_map,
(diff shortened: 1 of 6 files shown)


nilsvu:
Could you briefly mention in the description why you are making this change?

knelli2:
@nilsvu Done! Hopefully that gives a decent picture of why I made the changes.

src/Domain/Creators/Sphere.hpp
```diff
@@ -64,6 +71,16 @@ class Sphere : public DomainCreator<3> {
     static constexpr Options::String help = {"Radius of the Sphere."};
   };
 
+  struct InnerCubeSphericity {
+    using type = double;
+    static constexpr Options::String help = {
+        "Sphericity of inner cube. A sphericity of 0 uses an actual cube as "
```
mar-celine: I would change "actual cube" to "product of 1D maps" or similar

src/Domain/Creators/Sphere.hpp
```diff
@@ -25,6 +25,7 @@ template <typename Map1, typename Map2, typename Map3>
 class ProductOf3Maps;
 template <size_t Dim>
 class Wedge;
+class BulgedCube;
```
mar-celine: Can be put after `Affine` (if want alphabetical)

mar-celine (commented):
Looks good, thanks for doing this!

Slightly tangential, I know that running a map with a root find makes the run take longer - are we taking advantage of the fact that the BulgedCube map is not time dependent? It's inverting the same map at each time step so I wanted to know if we have done everything we can to make this as efficient as it could be.

knelli2:
> Slightly tangential, I know that running a map with a root find makes the run take longer - are we taking advantage of the fact that the BulgedCube map is not time dependent? It's inverting the same map at each time step so I wanted to know if we have done everything we can to make this as efficient as it could be.

I'm not sure what the generic coordinate map class does. I would hope for all the time independent maps that everything related is only calculated once, but I really don't know.

tests/Unit/Domain/Creators/Python/Test_Sphere.py
```diff
@@ -9,6 +9,7 @@ class TestSphere(unittest.TestCase):
     def test_construction(self):
         sphere = Sphere(inner_radius=1.,
                         outer_radius=2.,
+                        inner_cube_sphericity=0.0,
```
mar-celine: Can be `0.` to be consistent with lines 10 and 11

mar-celine (commented):
minor comment, otherwise looks good!

tests/Unit/Domain/Creators/Python/Test_Sphere.py
knelli2: Updated!

nilsdeppe (dismissed):

mar-celine (dismissed):

src/Domain/Creators/Sphere.cpp
```diff
@@ -58,6 +62,10 @@ Sphere::Sphere(typename InnerRadius::type inner_radius,
     PARSE_ERROR(context,
                 "Cannot have periodic boundary conditions with a Sphere");
   }
+  if (time_dependence_ == nullptr) {
+    time_dependence_ =
+        std::make_unique<domain::creators::time_dependence::None<3>>();
```
nilsvu, line 77: include None.hpp

nilsvu (changes requested):
squash directly.

We really need to clean up the domain creator tests (#4408) ... I do some of that in #4406 but more drastic cleanups are clearly needed. Nothing for this PR.

nilsvu (approved):

PR #4398, opened 2022-11-21
Add Shape TimeDependence to Creators

mar-celine:
## Proposed changes

Adds to Domain/Creators/TimeDependence a Shape creator so that existing domains (e.g. Shell) can add as an additional option a Shape that takes in Kerr parameters and produces maps with the corresponding ylms to produce the Kerr shape. The linear falloff is hard-coded in this Creator.

Also included is an additional test to Test_SphereTransition that I thought was missing.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/TimeDependence/Shape.cpp
+++ b/src/Domain/Creators/TimeDependence/Shape.cpp
@@ -1,0 +1,149 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/Creators/TimeDependence/Shape.hpp"
+
+#include <array>
+#include <cstddef>
+#include <limits>
+#include <memory>
+#include <optional>
+#include <ostream>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/IndexType.hpp"
+#include "Domain/CoordinateMaps/CoordinateMap.tpp"
+#include "Domain/CoordinateMaps/MapInstantiationMacros.hpp"
+#include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp"
+#include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp"
+#include "Domain/Creators/TimeDependence/TimeDependence.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/PiecewisePolynomial.hpp"
+#include "Options/Options.hpp"
+#include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrHorizon.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/StdArrayHelpers.hpp"
+
+namespace domain {
+namespace creators::time_dependence {
+using SphereTransition =
+    domain::CoordinateMaps::ShapeMapTransitionFunctions::SphereTransition;
+
+Shape::Shape(const double initial_time, const size_t l_max, const double mass,
+             const std::array<double, 3> spin,
+             const std::array<double, 3> center, const double inner_radius,
+             const double outer_radius, const Options::Context& context)
+    : initial_time_(initial_time),
+      l_max_(l_max),
+      mass_(mass),
+      spin_(spin),
+      center_(center),
+      inner_radius_(inner_radius),
+      outer_radius_(outer_radius),
+      transition_func_(std::make_unique<SphereTransition>(
+          SphereTransition{inner_radius, outer_radius})) {
+  if (mass <= 0.0) {
+    PARSE_ERROR(context,
+                "Tried to create a Shape TimeDependence, but "
+                "the mass ("
+                    << mass << ") must be strictly positive.");
+  }
+  if (magnitude(spin) >= 1.0) {
+    PARSE_ERROR(context,
+                "Tried to create a Shape TimeDependence, but "
+                "the magnitude of the spin ("
+                    << spin << ") is greater than one.");
+  }
+  // There is no PARSE_ERROR for the `inner_radius` < `outer_radius` condition
+  // because the SphereTransition already checks for this condition.
+}
+
+std::unique_ptr<TimeDependence<3>> Shape::get_clone() const {
+  return std::make_unique<Shape>(initial_time_, l_max_, mass_, spin_, center_,
+                                 inner_radius_, outer_radius_);
+}
+
+std::vector<
+    std::unique_ptr<domain::CoordinateMapBase<Frame::Grid, Frame::Inertial, 3>>>
+Shape::block_maps_grid_to_inertial(
+    const size_t number_of_blocks) const {
+  ASSERT(number_of_blocks > 0,
+         "Must have at least one block on which to create a map.");
+
+  std::vector<std::unique_ptr<
+      domain::CoordinateMapBase<Frame::Grid, Frame::Inertial, 3>>>
+      result{number_of_blocks};
+  result[0] = std::make_unique<GridToInertialMap>(grid_to_inertial_map());
+  for (size_t i = 1; i < number_of_blocks; ++i) {
+    result[i] = result[0]->get_clone();
+  }
+  return result;
+}
+
+std::unordered_map<std::string,
+                   std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
+Shape::functions_of_time(
+    const std::unordered_map<std::string, double>& initial_expiration_times)
+    const {
+  std::unordered_map<std::string,
+                     std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
+      result{};
+
+  // Functions of time don't expire by default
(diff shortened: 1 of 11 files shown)


src/Domain/Creators/TimeDependence/Shape.cpp
```diff
@@ -0,0 +1,151 @@
...
+using SphereTransition =
+    domain::CoordinateMaps::ShapeMapTransitionFunctions::SphereTransition;
+
+Shape::Shape(const double initial_time, const size_t l_max, const double mass,
+             const std::array<double, 3> spin,
+             const std::array<double, 3> center, const double inner_radius,
+             const double outer_radius, const Options::Context& context)
+    : initial_time_(initial_time),
+      l_max_(l_max),
+      mass_(mass),
```
knelli2, line 43: Check for positive mass?

src/Domain/Creators/TimeDependence/Shape.hpp
```diff
@@ -0,0 +1,184 @@
...
+
+  GridToInertialMap grid_to_inertial_map() const;
+  double initial_time_{std::numeric_limits<double>::signaling_NaN()};
+  size_t l_max_{2};
+  double mass_{std::numeric_limits<double>::signaling_NaN()};
+  std::array<double, 3> spin_{
+      make_array<3>(std::numeric_limits<double>::signaling_NaN())};
+  std::array<double, 3> center_{
+      make_array<3>(std::numeric_limits<double>::signaling_NaN())};
+  inline static const std::string function_of_time_name_{"Shape"};
```
knelli2, line 181: This assumes we will only have 1 shape map in the domain. Is this an ok assumption? Meaning will we ever want to be able to add two shape time dependences at a time? My guess is no, but just thought I'd ask.

tests/Unit/Domain/Creators/TimeDependence/Test_Shape.cpp
```diff
@@ -0,0 +1,341 @@
...
+  TIME_DEPENDENCE_GENERATE_COORDS(make_not_null(&gen), 3,
+                                  inner_radius / sqrt(3.0),
+                                  outer_radius / sqrt(3.0));
+
+  for (const auto& block_map : block_maps) {
+    // We've checked equivalence above
+    // (CHECK(*block_map == expected_block_map);), but have sometimes been
+    // burned by incorrect operator== implementations so we check that the
+    // mappings behave as expected.
+    const double time_offset = dist(gen) + 1.2;
```
knelli2: Just define
```cpp
const double check_time = initial_time + dist(gen) + 1.2;
```
so you don't have to add it everywhere.

tests/Unit/Domain/Creators/TimeDependence/Test_Shape.cpp
```diff
@@ -0,0 +1,341 @@
...
+SPECTRE_TEST_CASE("Unit.Domain.Creators.TimeDependence.Shape",
+                  "[Domain][Unit]") {
+  constexpr double initial_time{1.3};
+  // l_max of 16 needed to pass tests using `approx` as tests
+  // compare ylm output with analytic solution.
+  constexpr size_t l_max{16};
+  constexpr double mass{1.0};
+  const std::array<double, 3> spin{{0.1, 0.4, -0.5}};
+  const std::array<double, 3> center{{-0.02, 0.013, 0.024}};
+  //  const std::array<double, 3> center{{-0.02, 0.013, 0.024}};
```
knelli2: Remove?

knelli2 (changes requested):
Looks good, just a couple changes.

src/Domain/Creators/TimeDependence/Shape.hpp
mar-celine: After discussion with @nilsdeppe I confirmed that Domains that require the equivalent of two shape maps are highly specialized, so there will be no reason to expect them to make use of the Shape TimeDependence functionality being added, so it's ok to keep this as just "Shape."

knelli2: Ok! Sounds good!

src/Domain/Creators/TimeDependence/Shape.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  // If we have control systems, overwrite the expiration time with the one
+  // supplied by the control system
+  if (initial_expiration_times.count(function_of_time_name_) == 1) {
+    expiration_time = initial_expiration_times.at(function_of_time_name_);
+  }
+
+  // This is Kerr r_+, the radius of the outer horizon.
+  const double horizon_boyer_lindquist =
+      mass_ * (1. + sqrt(1. - dot(spin_, spin_)));
```
nilsvu: Haven't looked at this carefully yet, but I think this always conforms to a Kerr horizon. We excise a bit inside the horizon, so (at least for initial data solves, but I think also for evolutions?) we need to conform the excision to an ellipsoid of _any_ constant Boyer-Lindquist radius r_BL, not necessarily r_BL=r_+. The r_BL is given by the `inner_radius` parameter of the excision. Here's the commit on my branch that fixes that: https://github.com/nilsvu/spectre/commit/494b47479c963942ffe12a412a8e0d712a7f0403
There's also a description in [my thesis](https://publishup.uni-potsdam.de/opus4-ubp/frontdoor/deliver/index/docId/56226/file/vu_diss.pdf) (Sec. 4.1.2) that might help. (It's not particularly complicated, I just haven't seen it written down anywhere. I should really move it into the spectre docs.)

tests/Unit/Domain/Creators/TimeDependence/Test_Shape.cpp
```diff
@@ -0,0 +1,345 @@
...
+    // phi) keep in mind how this interacts with r==0, consider an assert.
+    const auto output_point =
+        (*block_map)(grid_coords_double, check_time, functions_of_time);
+    test_r_theta_phi(grid_coords_double, output_point, inner_radius,
+                     outer_radius, mass, spin, center);
+  }
+}
+
+void test_equivalence() {
+  const Transition sphere_transition{1.0, 100.};
```
knelli2: Is this `sphere_transition` used in this equivalence test?

src/Domain/Creators/TimeDependence/Shape.cpp
knelli2: I think we want a generic constant BL radius for more complicated domains like BBH, but in a time dependence that will most likely be used only with the shell domain, I don't know if it's that important.

tests/Unit/Domain/Creators/TimeDependence/Test_Shape.cpp
mar-celine: It's not :sweat_smile: I'll remove it, thanks!

src/Domain/Creators/TimeDependence/Shape.cpp
mar-celine: Thanks for this! Should this change be implemented here? I know the fix relies on a `kerr_radius` function that is not on develop yet, should that also be added? Or maybe in a future PR?

knelli2: If it needs something that's not on develop yet, I'd say defer the change to a future PR, but up to @nilsvu.

src/Domain/Creators/TimeDependence/Shape.cpp
```diff
@@ -0,0 +1,157 @@
...
+
+  // If we have control systems, overwrite the expiration time with the one
+  // supplied by the control system
+  if (initial_expiration_times.count(function_of_time_name_) == 1) {
+    expiration_time = initial_expiration_times.at(function_of_time_name_);
+  }
+
+  // This is Kerr r_+, the radius of the outer horizon.
+  // const double horizon_boyer_lindquist =
+  //    mass_ * (1. + sqrt(1. - dot(spin_, spin_)));
```
nilsvu: delete comment

src/Domain/Creators/TimeDependence/Shape.cpp
```diff
@@ -0,0 +1,157 @@
...
+  return lhs.initial_time_ == rhs.initial_time_ and lhs.l_max_ == rhs.l_max_ and
+         lhs.mass_ == rhs.mass_ and lhs.spin_ == rhs.spin_ and
+         lhs.center_ == rhs.center_ and
+         lhs.inner_radius_ == rhs.inner_radius_ and
+         lhs.outer_radius_ == rhs.outer_radius_ and
+         (lhs.transition_func_ == nullptr) ==
+             (rhs.transition_func_ == nullptr) and
+         ((lhs.transition_func_ == nullptr and
+           rhs.transition_func_ == nullptr) or  // LCOV_EXCL_LINE
+          *lhs.transition_func_ == *rhs.transition_func_);
```
nilsvu: no need to check this I think, because you already check inner and outer radius

src/Domain/Creators/TimeDependence/Shape.hpp
```diff
@@ -0,0 +1,184 @@
...
+namespace Frame {
+struct Distorted;
+struct Grid;
+struct Inertial;
+}  // namespace Frame
+/// \endcond
+
+namespace domain::creators::time_dependence {
+/*!
+ * \brief A Shape conforming to a Kerr horizon, as given by
```
nilsvu: please update these docs so they reflect that the inner surface conforms to a surface of constant Boyer-Lindquist radius `r_BL` in Kerr-Schild coordinates. The value of `r_BL` is `inner_radius`.

nilsvu (changes requested):
please squash directly

nilsvu (approved):

PR #4403, opened 2022-11-21
Fix GenerateXdmf for #4378

nilsvu:
## Proposed changes

Broken due to interaction with #4378.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/GenerateXdmf.py
+++ b/src/Visualization/Python/GenerateXdmf.py
@@ -246,6 +246,10 @@ def generate_xdmf(file_prefix, output, subfile_name, start_time, stop_time,
                 components.remove('grid_names')
                 components.remove('bases')
                 components.remove('quadratures')
+                if 'domain' in components:
+                    components.remove('domain')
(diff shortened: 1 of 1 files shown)


nilsdeppe (approved):

src/Visualization/Python/GenerateXdmf.py
```diff
@@ -246,6 +246,10 @@ def data_item(data_type):
                 components.remove('grid_names')
```
nilsdeppe, line 246: Would something like a `list` of things to remove if they exist be useful/more maintainable?

nilsvu: 🤷‍♂️ I think we should eventually use the Py bindings here. Then we don't need to maintain this list in two places at all (`h5::VolumeData::list_tensor_components` already handles this).

nilsdeppe: Ah yea, agreed :)

PR #4405, opened 2022-11-22
Add analytic gauge condition to GH

nilsdeppe:
## Proposed changes

Allows using things like KS as the gauge.

After this I will have another 2-3 PRs to add tags and switch the exucatble, plus a couple helper things before the runtime gauge choices are completely done.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/AnalyticChristoffel.hpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/AnalyticChristoffel.hpp
@@ -1,0 +1,86 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <memory>
+#include <pup.h>
+#include <string>
+
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Gauges.hpp"
+#include "Options/Options.hpp"
+#include "PointwiseFunctions/InitialDataUtilities/InitialData.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+namespace gsl {
+template <class T>
+class not_null;
+}  // namespace gsl
+template <size_t Dim>
+class Mesh;
+/// \endcond
+
+namespace GeneralizedHarmonic::gauges {
+/*!
+ * \brief Imposes the analytic gauge condition,
+ * \f$H_a=\Gamma_a^{\mathrm{analytic}}\f$ from an analytic solution or analytic
+ * data.
+ *
+ * \warning Assumes \f$\partial_t \Gamma_a=0\f$ i.e. the solution is static or
+ * in harmonic gauge.
+ */
+class AnalyticChristoffel final : public GaugeCondition {
+ public:
+  /// \brief What analytic solution/data to prescribe.
+  struct AnalyticPrescription {
+    static constexpr Options::String help =
+        "What analytic solution/data to prescribe.";
+    using type = std::unique_ptr<evolution::initial_data::InitialData>;
+  };
+
+  using options = tmpl::list<AnalyticPrescription>;
+
+  static constexpr Options::String help{
+      "Apply the analytic gauge condition H_a = Gamma_a, where Gamma_a comes "
+      "from the AnalyticPrescription."};
+
+  AnalyticChristoffel() = default;
+  AnalyticChristoffel(const AnalyticChristoffel&);
+  AnalyticChristoffel& operator=(const AnalyticChristoffel&);
+  AnalyticChristoffel(AnalyticChristoffel&&) = default;
+  AnalyticChristoffel& operator=(AnalyticChristoffel&&) = default;
+  ~AnalyticChristoffel() override = default;
+
+  explicit AnalyticChristoffel(
+      std::unique_ptr<evolution::initial_data::InitialData>
+          analytic_prescription);
+
+  /// \cond
+  explicit AnalyticChristoffel(CkMigrateMessage* msg);
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(AnalyticChristoffel);  // NOLINT
+  /// \endcond
+
+  template <size_t SpatialDim>
+  void gauge_and_spacetime_derivative(
+      gsl::not_null<tnsr::a<DataVector, SpatialDim, Frame::Inertial>*> gauge_h,
+      gsl::not_null<tnsr::ab<DataVector, SpatialDim, Frame::Inertial>*>
+          d4_gauge_h,
+      const Mesh<SpatialDim>& mesh, const double time,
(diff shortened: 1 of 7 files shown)


src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Analytic.hpp
```diff
@@ -0,0 +1,81 @@
...
+/// \endcond
+
+namespace GeneralizedHarmonic::gauges {
+/*!
+ * \brief Imposes the analytic gauge condition,
+ * \f$H_a=\Gamma_a^{\mathrm{analytic}}\f$.
+ *
+ * \warning Assumes \f$\partial_t \Gamma_a=0\f$ i.e. the solution is static.
+ */
+class Analytic final : public GaugeCondition {
```
knelli2: How do you feel about naming this `AnalyticChristoffel`? Just `Analytic` doesn't really give any insight into what the gauge is because technically $H_a = 0, H_a = 2 \Gamma_a^{\mathrm{Analytic}}$ are all analytic as well, but aren't what we *mean* when we say analytic.

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Analytic.hpp
```diff
@@ -0,0 +1,81 @@
...
+  /// \brief What analytic solution/data to prescribe.
+  struct AnalyticPrescription {
+    static constexpr Options::String help =
+        "What analytic solution/data to prescribe.";
+    using type = std::unique_ptr<evolution::initial_data::InitialData>;
+  };
+
+  using options = tmpl::list<AnalyticPrescription>;
+
+  static constexpr Options::String help{"Apply an analytic gauge condition."};
```
knelli2: Could you change this to `Apply the analytic gauge condition H_a = Gamma_a, where Gamma_a comes from the AnalyticPrescription`.

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/RegisterDerived.cpp
```diff
@@ -3,21 +3,40 @@
...
+#include "Utilities/TMPL.hpp"
 
 namespace GeneralizedHarmonic::gauges {
 namespace {
 template <size_t Dim>
 void impl() {
-  Parallel::register_classes_with_charm(tmpl::list<>{});
+  // The analytic gauge condition also can hold all the different solutions, so
+  // register those too.
+  using solutions = tmpl::append<
```
knelli2: Isn't this the same typelist as in `GeneralizedHarmonic/AllSolutions.hpp`?

tests/Unit/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Test_Analytic.cpp
```diff
@@ -0,0 +1,199 @@
...
+        tmpl::pair<evolution::initial_data::InitialData,
+                   tmpl::list<GeneralizedHarmonic::Solutions::WrappedGr<
+                                  gr::Solutions::KerrSchild>,
+                              GeneralizedHarmonic::Solutions::WrappedGr<
+                                  gr::Solutions::GaugeWave<Dim>>>>>;
+  };
+};
+
+template <size_t Dim>
+void test(const Mesh<Dim>& mesh) {
```
knelli2: [optional] Rename to `test_gauge_wave` to match `test_ks`

tests/Unit/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Test_Analytic.cpp
```diff
@@ -0,0 +1,199 @@
...
+                                      make_not_null(&d4_gauge_h), mesh, time,
+                                      inertial_coords, inverse_jacobian);
+
+  CHECK_ITERABLE_APPROX(
+      gauge_h, (tnsr::a<DataVector, Dim, Frame::Inertial>(num_points, 0.0)));
+  CHECK_ITERABLE_APPROX(d4_gauge_h, (tnsr::ab<DataVector, Dim, Frame::Inertial>(
+                                        num_points, 0.0)));
+}
+
+void test_ks(const Mesh<3>& mesh) {
```
knelli2: Not a change request, but this is basically just copying the internals of the analytic gauge condition. Not really sure what the best way is to test it though. What'd be optimal is if you calculated $\Gamma_a$ by hand given some special choice of mass and spin that weren't zero and just compared the new function against what you got by hand rather than redoing the calculation. But that's a lot of work and is probably unnecessary.

knelli2 (changes requested):

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Analytic.hpp
nilsdeppe: Yea, sounds good to me. I have no attachment to the current name, just couldn't  think of anything better at the time :D

tests/Unit/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Test_Analytic.cpp
nilsdeppe: Yea, I can't think of anything better offhand because even testing the 1-index constraint requires recomputing the stuff inside the original class...

nilsdeppe:
Rebased and squashed. Hopefully I got all the requests :) Thanks for reviewing!

knelli2 (dismissed):

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/AnalyticChristoffel.hpp
```diff
@@ -0,0 +1,85 @@
...
+class Mesh;
+/// \endcond
+
+namespace GeneralizedHarmonic::gauges {
+/*!
+ * \brief Imposes the analytic gauge condition,
+ * \f$H_a=\Gamma_a^{\mathrm{analytic}}\f$ from an analytic solution or analytic
+ * data.
+ *
+ * \warning Assumes \f$\partial_t \Gamma_a=0\f$ i.e. the solution is static.
```
kidder: I'm confused; you test this on a gauge wave which is not static

nilsdeppe: The derivative being zero is fine (it's just a test). I'm confused on that H_a seems to vanish. That's not what I get analytically...

nilsdeppe: Updated to specify it can be either static or in harmonic gauge. I think technically any spatially and temporally constant function is okay, but H_a=0 is the only realistically used one that I know of.

kidder (approved):

PR #4406, opened 2022-11-22
Split boundary conditions out of Domain

nilsvu:
## Proposed changes

- Store boundary conditions in a separate tag in the global cache.
- Allows to serialize the domain into volume data files and read it back in with another executable.
- Also clean up the domain creator tests. Removing some tests that only duplicate src code and adding tests that check consistency of domain creators.

### Upgrade instructions

Boundary conditions have moved. They are not stored in the `Domain` anymore but separately in `domain::Tags::ExternalBoundaryConditions` in the global cache.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Block.cpp
+++ b/src/Domain/Block.cpp
@@ -11,7 +11,6 @@
 #include <typeinfo>
 #include <utility>
 
-#include "Domain/BoundaryConditions/BoundaryCondition.hpp"
 #include "Domain/Structure/BlockNeighbor.hpp"
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/DirectionMap.hpp"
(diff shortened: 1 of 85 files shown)


src/Domain/Block.cpp
```diff
@@ -11,7 +11,6 @@
 #include <typeinfo>
```
nilsdeppe, line 11: I don't really know how to review this commit. The commit message doesn't help me understand why the domain creators are no longer having the boundary conditions. Where are they going? This commit just does a lot and would be a lot easier to review if it's possible to split into smaller logical pieces...

nilsvu: The domain creators still have the boundary conditions! They have just moved from `create_domain()` to the new `external_boundary_conditions()` function, and then they are stored in `domain::Tags::ExternalBoundaryConditions`.

This commit is definitely annoying to review (and the refactoring has been pretty annoying too 😅). I can split it up into a few logical commits and squash them together later if that helps?

nilsdeppe: Ah okay! I see the pattern now :) Can't (yet) promise I will understand the test changes, but I can now understand the `src` changes, so leave it as-is for now :)

src/Domain/Creators/Interval.cpp
```diff
@@ -70,6 +73,12 @@ Interval::Interval(
     time_dependence_ =
         std::make_unique<domain::creators::time_dependence::None<1>>();
   }
+  if ((lower_boundary_condition_ == nullptr) !=
+      (upper_boundary_condition_ == nullptr)) {
+    PARSE_ERROR(context,
```
nilsdeppe, line 78: add error test here?

src/Domain/Creators/Shell.cpp
```diff
@@ -212,6 +192,33 @@ Domain<3> Shell::create_domain() const {
...
 
+std::vector<DirectionMap<
+    3, std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>>>
+Shell::external_boundary_conditions() const {
+  if (inner_boundary_condition_ == nullptr) {
+    return {};
+  }
+  // This assumes 6 wedges making up the shell. If you need to support the
+  // FourOnEquator or OneAlongMinusX configurations the below code needs to be
+  // updated. This would require adding more boundary condition options to the
```
nilsdeppe, line 203: I realize you just moved this code, but would it be easy to add an `ERROR` here?

src/Evolution/Systems/Burgers/FiniteDifference/BoundaryConditionGhostData.hpp
```diff
@@ -94,11 +95,6 @@ void BoundaryConditionGhostData::apply(
   const size_t ghost_zone_size{reconstructor.ghost_zone_size()};
 
   for (const auto& direction : element.external_boundaries()) {
-    ASSERT(external_boundary_condition.at(direction) != nullptr,
```
nilsdeppe, line 97: why remove these `ASSERT`s? They have been useful in the past

nilsdeppe (changes requested):

src/Evolution/Systems/Burgers/FiniteDifference/BoundaryConditionGhostData.hpp
nilsvu: This code path got deleted in Block.cpp (here: https://github.com/sxs-collaboration/spectre/pull/4406/files#diff-e5e91039819dea4f628e25bd658c1c0c70c5099bc30278d9af198a754d5aa1a4L59). It is tested for all domain creators (here: https://github.com/sxs-collaboration/spectre/pull/4406/files#diff-964647fb533b432c1126907763d7322aef942a141aadec2ed188ead9199ba346R100). So I believe this assert is not needed anymore.

nilsvu:
Rebased on #4394 and pushed a fixup. Thanks for reviewing this annoying refactor 😅

src/Evolution/Systems/Burgers/FiniteDifference/BoundaryConditionGhostData.hpp
nilsdeppe: Sounds good!

nilsdeppe:
LGTM, please squash! And you're welcome :D Thanks for doing the annoying refactor!

nilsdeppe (approved):
CI failure is unrelated timeout

PR #4407, opened 2022-11-22
Prepare for serializing domains into output files

nilsvu:
## Proposed changes

More steps towards #3937 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/FixedHashMap.hpp
+++ b/src/DataStructures/FixedHashMap.hpp
@@ -180,8 +180,15 @@ class FixedHashMap {
 
   // NOLINTNEXTLINE(google-runtime-references)
   void pup(PUP::er& p) {
-    p | data_;
-    p | size_;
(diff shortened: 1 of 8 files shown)


nilsdeppe (approved):
LGTM! Do you think we would want to be able to get the version from classes? E.g. make it a member variable and then have a `version()` function?

nilsvu:
I don't think the version has any meaning outside of the pup function. When _packing_ it is the current version of the code that's hardcoded in the code. When _unpacking_ it is read from the serialized data to understand how to deserialize the rest of the data, but once unpacking is done the object is identical to objects constructed in other ways.

nilsdeppe:
Yea, fair enough. We'll see if we find it useful for debugging, but can always add in the future :)

kidder (approved):

PR #4409, opened 2022-11-22
Add largest char speed for GH+GRMHD

nilsdeppe:
## Proposed changes

This assumes -1<gamma_1<=0 which in our normal simulations it always is. I think the condition where this check is sufficient is more relaxed than that, but values of gamma_1 outside that range probably shouldn't be used anyway. Or at least not far outside that range.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Test_Characteristics.cpp
+++ b/tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Test_Characteristics.cpp
@@ -1,0 +1,42 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <limits>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/Characteristics.hpp"
+#include "Evolution/Systems/GrMhd/GhValenciaDivClean/Characteristics.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "Helpers/PointwiseFunctions/GeneralRelativity/TestHelpers.hpp"
+
+namespace {
+void check_max_char_speed(const DataVector& used_for_size) {
+  MAKE_GENERATOR(gen);
+
+  const Scalar<DataVector> gamma_1{used_for_size.size(), 0.0};
+  const auto lapse = TestHelpers::gr::random_lapse(&gen, used_for_size);
+  const auto shift = TestHelpers::gr::random_shift<3>(&gen, used_for_size);
+  const auto spatial_metric =
+      TestHelpers::gr::random_spatial_metric<3>(&gen, used_for_size);
+
+  double max_char_speed = std::numeric_limits<double>::signaling_NaN();
+  grmhd::GhValenciaDivClean::Tags::ComputeLargestCharacteristicSpeed<
+      Frame::Inertial>::function(make_not_null(&max_char_speed), lapse, shift,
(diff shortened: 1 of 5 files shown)


tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Test_Characteristics.cpp
```diff
@@ -0,0 +1,91 @@
...
+  MAKE_GENERATOR(gen);
+
+  const Scalar<DataVector> gamma_1{used_for_size.size(), 0.0};
+
+  // Fraction of times the test can randomly fail
+  const double failure_tolerance = 1e-10;
+  // Minimum fraction of the claimed result that must be found in some
+  // random trial
+  const double check_minimum = 0.9;
+  // Minimum fraction of the claimed result that can be found in any
```
kidder: Maximum

tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Test_Characteristics.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+namespace {
+template <size_t Dim>
+void check_max_char_speed(const DataVector& used_for_size) {
+  CAPTURE(Dim);
+  MAKE_GENERATOR(gen);
+
+  const Scalar<DataVector> gamma_1{used_for_size.size(), 0.0};
+
+  // Fraction of times the test can randomly fail
```
kidder: I don't understand what you are doing here with trials

kidder (changes requested):

tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Test_Characteristics.cpp
nilsdeppe: I just took this from the GH one. Didn't look through it in any detail, just set gamma_1 to zero to compare :shrug: I guess the easier test is just to compare to the GH one with gamma_1=0. Is that okay with you? We can file an issue to look at the GH one, if you want.

yoonso0-0 (approved):

tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Test_Characteristics.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "Evolution/Systems/GeneralizedHarmonic/Characteristics.hpp"
+#include "Evolution/Systems/GrMhd/GhValenciaDivClean/Characteristics.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "Helpers/PointwiseFunctions/GeneralRelativity/TestHelpers.hpp"
+
+namespace {
+void check_max_char_speed(const DataVector& used_for_size) {
+  MAKE_GENERATOR(gen);
+
+  const Scalar<DataVector> gamma_1{used_for_size.size(), 0.0};
```
yoonso0-0, line 19: LGTM. Just a quick question: do we need to test with `gamma_1` being other than 0.0 here? Like some random value between -1 and 0, as the doc says.

yoonso0-0 (approved):

tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Test_Characteristics.cpp
nilsdeppe: That's just guaranteed analytically, so would only be testing that the GH implementation is correct...

kidder (approved):

PR #4421, opened 2022-11-24
Make UniformRotation time dependence just Rotation

knelli2:
## Proposed changes

Still about the z-axis though.

It's helpful to test non-uniform rotation on the simpler domains.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


geoffrey4444 (dismissed):
I have a hack I've been using that does this. Thanks for getting around to adding this for real!

src/Domain/Creators/TimeDependence/RotationAboutZAxis.hpp
```diff
@@ -45,7 +45,7 @@ struct Inertial;
 
 namespace domain::creators::time_dependence {
 /*!
- * \brief A uniform rotation about the \f$z\f$ axis:
+ * \brief A rotation about the \f$z\f$ axis:
```
nilsdeppe: How is the rotation no longer uniform? Even a global acceleration makes the rotation spatially uniform, which is how I always interpreted the "uniform" part. I think the shorter class name is fine either way since have a non-uniform rotation is kind of difficult/annoying.

nilsdeppe (dismissed):

src/Domain/Creators/TimeDependence/RotationAboutZAxis.hpp
knelli2: I always interpreted "uniform" for both rotation and translation to mean non-accelerating since `UniformTranslation` can only specify the velocity and before this change, `UniformRotation` could only specify angular velocity.

nilsdeppe: Oh, translation is definitely meant as spatially uniform, temporal changes are sort of irrelevant for much of the code. I think only the control system cares, everything else only cares about the velocity.

knelli2: Ahh ok makes sense. I can change the docs back if you want to say `A spatially uniform rotation`

nilsdeppe: Yea, I wouldn't mind the docs being explicit like that :) But I think the class name is better. Is that okay with you?

knelli2: Yeah I'll just change the docs and leave the class name.

Edit: Done! :)

nilsdeppe (approved):
Thanks :D

PR #4425, opened 2022-11-29
Allow specifying which tags to reconstruct and support BCs in GH+GRMHD

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/GhValenciaDivClean/FiniteDifference/ReconstructWork.tpp
+++ b/src/Evolution/Systems/GrMhd/GhValenciaDivClean/FiniteDifference/ReconstructWork.tpp
@@ -36,11 +36,13 @@
 #include "Utilities/TMPL.hpp"
 
 namespace grmhd::GhValenciaDivClean::fd {
-template <typename SpacetimeTagsToReconstruct, typename PrimsTags,
+template <typename SpacetimeTagsToReconstruct,
+          typename PrimTagsForReconstruction, typename PrimsTags,
           typename SpacetimeAndConsTags, typename TagsList,
           size_t ThermodynamicDim, typename HydroReconstructor,
           typename SpacetimeReconstructor,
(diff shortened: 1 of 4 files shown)


kidder (commented):
LGTM; can you improve your code coverage?

nilsdeppe:
I think so. Let me take a look and report back. Might not get to it until tomorrow or Monday, but I think it's fairly easy to do

src/Evolution/Systems/GrMhd/GhValenciaDivClean/FiniteDifference/ReconstructWork.tpp
```diff
@@ -344,8 +361,12 @@ void reconstruct_fd_neighbor_work(
         }
       });
 
-  spacetime_vars_for_grmhd(vars_on_face);
-  ValenciaDivClean::fd::compute_conservatives_for_reconstruction(vars_on_face,
-                                                                 eos);
+  if constexpr (tmpl::size<SpacetimeTagsToReconstruct>::value != 0) {
```
yoonso0-0, line 364: Q. Are we planning to use flexible set of metric related tags for reconstruction? Can you give an example of situations where we reconstruct different set of metric vars?

yoonso0-0 (dismissed):

src/Evolution/Systems/GrMhd/GhValenciaDivClean/FiniteDifference/ReconstructWork.tpp
nilsdeppe: One would be reconstructing lapse, shift, and spatial metric instead of the spacetime metric. Lapse, in particular, since that must be positive

nilsdeppe:
I've rebased and squashed in a change to `PrimReconstructor.hpp` that should be testing the new code (we'll see what codecov says 🙃 )

kidder (approved):

PR #4428, opened 2022-11-29
Add function to print diagnostic info related to the domain

knelli2:
## Proposed changes

This function signature was chosen so this can be used in the `allocate_array` function of the DgElementArray and after AMR.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/DiagnosticInfo.hpp
+++ b/src/Domain/DiagnosticInfo.hpp
@@ -1,0 +1,92 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <sstream>
+#include <string>
+#include <vector>
+
+#include "Domain/Domain.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Info.hpp"
+#include "Parallel/Printf.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/Numeric.hpp"
+#include "Utilities/StdHelpers.hpp"
+
+namespace domain {
+/// Returns a `std::string` with diagnostic information about how elements and
+/// grid points are distributed on the nodes and cores.
+template <size_t Dim, typename Metavariables>
+std::string diagnostic_info(const Domain<Dim>& domain,
+                            const Parallel::GlobalCache<Metavariables>& cache,
+                            const std::vector<size_t>& elements_per_core,
+                            const std::vector<size_t>& elements_per_node,
+                            const std::vector<size_t>& grid_points_per_core,
+                            const std::vector<size_t>& grid_points_per_node) {
+  const size_t total_number_of_elements =
+      alg::accumulate(elements_per_node, 0_st);
+  const size_t total_number_of_grid_points =
+      alg::accumulate(grid_points_per_node, 0_st);
+  const size_t total_number_of_blocks = domain.blocks().size();
+  const size_t number_of_cores = Parallel::number_of_procs<size_t>(cache);
+  const size_t number_of_nodes = Parallel::number_of_nodes<size_t>(cache);
+
+  // Sanity checks
+  ASSERT(total_number_of_elements == alg::accumulate(elements_per_core, 0_st),
+         "Number of elements determined from elements_per_core ("
+             << alg::accumulate(elements_per_core, 0_st)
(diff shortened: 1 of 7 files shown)


src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -58,24 +61,57 @@ struct DefaultElementsAllocator
...
         num_of_procs_to_use, initial_refinement_levels, procs_to_ignore};
+
+    // Will be used to print domain diagnostic info
+    std::vector<size_t> elements_per_core(number_of_procs, 0_st);
+    std::vector<size_t> elements_per_node(number_of_nodes, 0_st);
+    std::vector<size_t> grid_points_per_core(number_of_procs, 0_st);
+    std::vector<size_t> grid_points_per_node(number_of_nodes, 0_st);
+
     for (const auto& block : domain.blocks()) {
+      const auto& initial_extent = initial_extents[block.id()];
```
kidder: Can do this in one line with `const size_t grid_points_per_element = alg::accumulate(initial_extents[block.id()], 1_st, std::multiplies<size_t>());`

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -90,45 +93,81 @@ void DgElementArray<Metavariables, PhaseDepActionList>::allocate_array(
...
+  std::vector<size_t> elements_per_core(number_of_procs, 0_st);
+  std::vector<size_t> elements_per_node(number_of_nodes, 0_st);
+  std::vector<size_t> grid_points_per_core(number_of_procs, 0_st);
+  std::vector<size_t> grid_points_per_node(number_of_nodes, 0_st);
+
+  size_t which_proc = 0;
   for (const auto& block : domain.blocks()) {
-    const auto initial_ref_levs = initial_refinement_levels[block.id()];
+    const auto& initial_ref_levs = initial_refinement_levels[block.id()];
+    const auto& initial_extent = initial_extents[block.id()];
```
kidder: see above

kidder (changes requested):

kidder (dismissed):

src/Domain/DiagnosticInfo.hpp
```diff
@@ -0,0 +1,109 @@
...
+  ASSERT(number_of_nodes == grid_points_per_node.size(),
+         "Number of nodes determined from grid_points_per_node ("
+             << grid_points_per_node.size()
+             << ") does not match number of nodes determined from "
+                "global cache ("
+             << number_of_nodes << ").");
+
+  using ::operator<<;
+  std::stringstream ss{};
+  ss << "----- Domain Info -----\n";
```
nilsdeppe: why make all these separate lines? I think you can actually just do:
```cpp
return MakeString{} << all_your_stuff;
```
to avoid dealing with the stringstream manually. Otherwise no need for a new `ss <<` each line.

src/Domain/DiagnosticInfo.hpp
```diff
@@ -0,0 +1,109 @@
...
+
+/// Prints domain diagnostic information using the `diagnostic_info()` function.
+template <size_t Dim, typename Metavariables>
+void print_diagnostic_info(const Domain<Dim>& domain,
+                           const Parallel::GlobalCache<Metavariables>& cache,
+                           const std::vector<size_t>& elements_per_core,
+                           const std::vector<size_t>& elements_per_node,
+                           const std::vector<size_t>& grid_points_per_core,
+                           const std::vector<size_t>& grid_points_per_node) {
+  Parallel::printf(
```
nilsdeppe: I generally don't like things that print, especially something like this that just prints a string. It's generally better to just have the end user decide where to place the string. E.g. stream operators don't do any `std::cout`-type stuff internally. The change request is: just remove `print_diagnostic_info` and call `Parallel::printf(...)` wherever you need to print.

nilsdeppe (changes requested):

src/Domain/DiagnosticInfo.hpp
knelli2: Just removed the `ss <<` so the code formatting stayed nice

nilsdeppe (approved):

PR #4435, opened 2022-11-30
Add DG time derivative to ForceFree system

yoonso0-0:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ForceFree/TimeDerivativeTerms.cpp
+++ b/src/Evolution/Systems/ForceFree/TimeDerivativeTerms.cpp
@@ -1,0 +1,107 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/ForceFree/Fluxes.hpp"
+#include "Evolution/Systems/ForceFree/Sources.hpp"
+#include "Evolution/Systems/ForceFree/TimeDerivativeTerms.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Christoffel.hpp"
+#include "PointwiseFunctions/GeneralRelativity/IndexManipulation.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace ForceFree {
+
+void TimeDerivativeTerms::apply(
+    const gsl::not_null<tnsr::I<DataVector, 3, Frame::Inertial>*>
+        non_flux_terms_dt_tilde_e,
+    const gsl::not_null<tnsr::I<DataVector, 3, Frame::Inertial>*>
+        non_flux_terms_dt_tilde_b,
+    const gsl::not_null<Scalar<DataVector>*> non_flux_terms_dt_tilde_psi,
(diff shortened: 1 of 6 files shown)


src/Evolution/Systems/ForceFree/TimeDerivativeTerms.cpp
```diff
@@ -0,0 +1,107 @@
...
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/ForceFree/Fluxes.hpp"
+#include "Evolution/Systems/ForceFree/Sources.hpp"
+#include "Evolution/Systems/ForceFree/TimeDerivativeTerms.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Christoffel.hpp"
+#include "PointwiseFunctions/GeneralRelativity/IndexManipulation.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace ForceFree {
```
nilsdeppe, line 15: This namespace is still undocumented, makes it a bit tough to find this in the doxygen. I guess we usually document these in `System.hpp`, right?

nilsdeppe (approved):

src/Evolution/Systems/ForceFree/TimeDerivativeTerms.cpp
yoonso0-0: Yup!

src/Evolution/Systems/ForceFree/TimeDerivativeTerms.hpp
```diff
@@ -0,0 +1,119 @@
...
+}  // namespace gsl
+/// \endcond
+
+namespace ForceFree {
+
+/*!
+ * \brief Compute the time derivative of the conserved variables for the GRFFE
+ * equations with divergence cleaning.
+ *
+ */
```
teukolsky, line 26: Did I miss somewhere that this was documented? We usually give the equations explicitly and add a reference. Note that there are several FFE formulations, so the reference is important.

teukolsky: OK, I see this is all in the other PR.

nilsdeppe: The documentation has been spread between several structs in other systems (fluxes on the fluxes struct, sources on the sources struct, etc.) which has made it difficult to see the full evolution equations at once. Yoonsoo's seeing how documenting them on the system struct (see this PR: https://github.com/sxs-collaboration/spectre/pull/4440 ) works out. With subcell there's also a second `TimeDerivative` class and so I think moving the full documentation of the system to either the `System` class, or maybe actually the namespace (e.g. `GeneralizedHarmonic` or `ForceFree`) will make it easier to find and also easier to copy the equations into papers. Do you have any thoughts on this change/direction?

teukolsky: I don't feel strongly about this - perhaps it is better to have all the documentation in one place. I only commented because I was expecting to see documentation here based on the other systems.

nilsdeppe: Ah okay. Sounds good!

PR #4437, opened 2022-11-30
Dump Interpolator volume data on failure

knelli2:
## Proposed changes

I fixed the issues I was having with this before that prompted me to disable the feature before in #4404.

I added an overload to ContributeVolumeDataToWriter that allows it to be used from the Interpolator.

I also added a tag that allows you to toggle this feature just in case it causes issues in the future, so users can easily disable it.

### Upgrade instructions

If your executable uses the Interpolator parallel component, you'll need to add the following block to your input file:
```yaml
Interpolator:
  DumpVolumeDataOnFailure: false
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/Observer/VolumeActions.hpp
+++ b/src/IO/Observer/VolumeActions.hpp
@@ -6,6 +6,7 @@
 #include <cstddef>
 #include <iterator>
 #include <mutex>
+#include <optional>
 #include <unordered_map>
 
 #include "DataStructures/DataBox/DataBox.hpp"
@@ -162,6 +163,23 @@ void write_data(const std::string& h5_file_name,
 struct ContributeVolumeDataToWriter {
   template <typename ParallelComponent, typename DbTagsList,
             typename Metavariables, typename ArrayIndex>
+  static void apply(db::DataBox<DbTagsList>& box,
+                    Parallel::GlobalCache<Metavariables>& cache,
+                    const ArrayIndex& /*array_index*/,
+                    const gsl::not_null<Parallel::NodeLock*> node_lock,
+                    const observers::ObservationId& observation_id,
+                    ArrayComponentId observer_group_id,
+                    const std::string& subfile_name,
+                    std::unordered_map<observers::ArrayComponentId,
+                                       std::vector<ElementVolumeData>>&&
+                        received_volume_data) {
+    apply_impl<Tags::InterpolatorTensorData, ParallelComponent>(
+        box, cache, node_lock, observation_id, observer_group_id, subfile_name,
+        received_volume_data);
+  }
+
+  template <typename ParallelComponent, typename DbTagsList,
+            typename Metavariables, typename ArrayIndex>
   static void apply(
       db::DataBox<DbTagsList>& box, Parallel::GlobalCache<Metavariables>& cache,
       const ArrayIndex& /*array_index*/,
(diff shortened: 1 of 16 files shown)


src/IO/Observer/VolumeActions.hpp
```diff
@@ -283,6 +310,29 @@ struct ContributeVolumeDataToWriter {
...
+
+      std::vector<ElementVolumeData> volume_data_to_write;
+
+      if constexpr (std::is_same_v<tmpl::at_c<VolumeDataAtObsId, 1>,
+                                   ElementVolumeData>) {
+        volume_data_to_write.reserve(volume_data.size());
+        for (const auto& id_and_element : volume_data) {
+          volume_data_to_write.push_back(id_and_element.second);
+        }
+      } else {
```
nilsdeppe, line 323: `if constexpr (tt::is_maplike_v<...>) {`? Because it looks like you're assuming the data structure is map-like. Having said that, I think the actual requirement is that `tt::is_a_v<std::pair, ...>`

src/IO/Observer/VolumeActions.hpp
```diff
@@ -283,6 +310,29 @@ struct ContributeVolumeDataToWriter {
     if (perform_write) {
       ASSERT(not volume_data.empty(),
              "Failed to populate volume_data before trying to write it.");
+
+      std::vector<ElementVolumeData> volume_data_to_write;
+
+      if constexpr (std::is_same_v<tmpl::at_c<VolumeDataAtObsId, 1>,
+                                   ElementVolumeData>) {
+        volume_data_to_write.reserve(volume_data.size());
+        for (const auto& id_and_element : volume_data) {
```
nilsdeppe: `const auto& [id, element]`

src/IO/Observer/VolumeActions.hpp
```diff
@@ -283,6 +310,29 @@ struct ContributeVolumeDataToWriter {
...
+          volume_data_to_write.push_back(id_and_element.second);
+        }
+      } else {
+        size_t total_size = 0;
+        for (const auto& vol_data_at_array_id : volume_data) {
+          total_size += vol_data_at_array_id.second.size();
+        }
+        volume_data_to_write.reserve(total_size);
+
+        for (const auto& id_and_element : volume_data) {
```
nilsdeppe: `const auto& [id, element]`

src/ParallelAlgorithms/Interpolation/Actions/DumpInterpolatorVolumeData.hpp
```diff
@@ -80,6 +80,10 @@ struct DumpInterpolatorVolumeData {
       Parallel::GlobalCache<Metavariables>& cache,
       const ArrayIndex& array_index, const ActionList /*meta*/,
       const ParallelComponent* const /*meta*/) {
+    if (not db::get<intrp::Tags::DumpVolumeDataOnFailure>(box)) {
+      return {Parallel::AlgorithmExecution::Continue, std::nullopt};
```
nilsdeppe, line 87: Is this actually not tested? If so, could you add a test please?

nilsdeppe (changes requested):
You can squash

src/ParallelAlgorithms/Interpolation/Actions/DumpInterpolatorVolumeData.hpp
```diff
@@ -76,19 +78,22 @@ struct DumpInterpolatorVolumeData {
...
       const ParallelComponent* const /*meta*/) {
-    auto& observer_writer = Parallel::get_parallel_component<
-        observers::ObserverWriter<Metavariables>>(cache);
-    auto& my_proxy = Parallel::get_parallel_component<ParallelComponent>(cache);
-    const auto& file_prefix =
-        Parallel::get<observers::Tags::VolumeFileName>(cache);
-    const size_t my_node =
-        Parallel::my_node<size_t>(*Parallel::local_branch(my_proxy));
-    const std::string filename{file_prefix + std::to_string(my_node)};
+    if (not db::get<intrp::Tags::DumpVolumeDataOnFailure>(box)) {
```
nilsvu: Add `const_global_cache_tags = tmpl::list<intrp::Tags::DumpVolumeDataOnFailure>` to this action. Then you don't have to worry about this elsewhere.

src/ParallelAlgorithms/Interpolation/Actions/InitializeInterpolator.hpp
```diff
@@ -52,6 +54,7 @@ namespace Actions {
 /// - Modifies: nothing
 template <typename VolumeVarsInfos, typename InterpolatedVarsHolders>
 struct InitializeInterpolator {
+  using initialization_tags = tmpl::list<Tags::DumpVolumeDataOnFailure>;
```
nilsvu: delete this line (see comment above)

src/IO/Observer/VolumeActions.hpp
knelli2: No here I'm trying to check if the "value" of the `std::unordered_map` represented by `VolumeDataAtObsId` is an `ElementVolumeData` or a `std::vector<ElementVolumeData>`. `tmpl::at_c<VolumeDataAtObsId, 1>` gets this "value" and then I check if it's an `ElementVolumeData`. The unordered map is a pair regardless.

nilsvu (changes requested):
Sorry I forgot about this PR. Looks like a good solution 👍

knelli2:
Rebased and squashed everything in.

nilsdeppe (approved):

nilsvu (approved):

PR #4439, opened 2022-11-30
Fix DistributedObject for tracing

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/DistributedObject.hpp
+++ b/src/Parallel/DistributedObject.hpp
@@ -1009,7 +1009,7 @@ DistributedObject<ParallelComponent, tmpl::list<PhaseDepActionListsPack...>>::
     // profiling is fairly small (<2%), we still avoid it when we aren't
     // tracing.
 #ifdef SPECTRE_CHARM_PROJECTIONS
-    if constexpr (Parallel::is_array_proxy<cproxy_type>::value) {
+    if constexpr (Parallel::is_array<parallel_component>::value) {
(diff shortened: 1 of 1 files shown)


nilsdeppe (approved):

kidder:
how did it work before?

nilsdeppe:
If I remember correctly we've changed the metafunctions since this was originally added when we moved to singletons being an array

src/Parallel/DistributedObject.hpp
```diff
@@ -1009,7 +1009,7 @@ DistributedObject<ParallelComponent, tmpl::list<PhaseDepActionListsPack...>>::
     // profiling is fairly small (<2%), we still avoid it when we aren't
     // tracing.
 #ifdef SPECTRE_CHARM_PROJECTIONS
-    if constexpr (Parallel::is_array_proxy<cproxy_type>::value) {
+    if constexpr (Parallel::is_array<parallel_component>::value) {
       if (not this->thisProxy[array_index_]
                   .template invoke_iterable_action<
```
knelli2, line 1014: @kidder Not sure. I haven't used tracing/projections in a while. Might have been broken when we changed singletons to be single element charm arrays.  The error occurred when the parallel component was a singleton. The proxy type is an array proxy (because singletons are single element arrays), but our singletons don't have an entry method called `invoke_iterale_action`. Thus this bug was hit.

PR #4449, opened 2022-12-02
Add SetPiFromGauge mutator to GH system

nilsdeppe:
## Proposed changes

This will replace the current DH gauge initialization action. This sets Pi in the initial data so we are in whatever evolution gauge is specified by H_a.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/SetPiFromGauge.cpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/SetPiFromGauge.cpp
@@ -1,0 +1,172 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/SetPiFromGauge.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataBox/Prefixes.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Dispatch.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Gauges.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Christoffel.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/DerivSpatialMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/ExtrinsicCurvature.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/GaugeSource.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/Pi.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/SpacetimeDerivativeOfSpacetimeMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/SpatialDerivOfLapse.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/SpatialDerivOfShift.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/TimeDerivOfSpatialMetric.hpp"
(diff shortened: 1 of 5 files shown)


src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/SetPiFromGauge.cpp
```diff
@@ -0,0 +1,174 @@
...
+        grid_to_inertial_map,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time,
+    const tnsr::I<DataVector, Dim, Frame::ElementLogical>& logical_coordinates,
+    const tnsr::aa<DataVector, Dim, Frame::Inertial>& spacetime_metric,
+    const tnsr::iaa<DataVector, Dim, Frame::Inertial>& phi,
+    const gauges::GaugeCondition& gauge_condition) {
+  // Since we are called during initialization, Tags::Time isn't valid and so we
+  // can't retrieve coordinates, Jacobians, etc.
```
knelli2: If `Tags::Time` is invalid, how are we able to use the initial time here? I thought the initial time is valid during initialization, just not during self start?

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/SetPiFromGauge.hpp
```diff
@@ -0,0 +1,69 @@
...
+ * This is necessary to ensure the initial data is in the desired evolution
+ * gauge.
+ */
+template <size_t Dim>
+struct SetPiFromGauge {
+ public:
+  using return_tags =
+      tmpl::list<GeneralizedHarmonic::Tags::Pi<Dim, Frame::Inertial>>;
+  using argument_tags =
+      tmpl::list<::Tags::Time, domain::Tags::Mesh<Dim>,
```
knelli2, line 40: Follow-up to other comment, if `Tags::Time` is invalid, should this be `Tags::InitialTime`?

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/SetPiFromGauge.cpp
nilsdeppe: `InitialTime` got remove, but I forgot to remove the comment

nilsdeppe:
Rebased, renamed `initial_time` to `time`, removed comment, squashed. There are a few places that compute the coordinates explicitly still (`SetVariables` comes to mind), but I think that's a universal cleanup of initialization

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/SetPiFromGauge.hpp
```diff
@@ -0,0 +1,69 @@
...
+                 domain::Tags::ElementMap<Dim, Frame::Grid>,
+                 domain::CoordinateMaps::Tags::CoordinateMap<Dim, Frame::Grid,
+                                                             Frame::Inertial>,
+                 domain::Tags::FunctionsOfTime,
+                 domain::Tags::Coordinates<Dim, Frame::ElementLogical>,
```
knelli2, line 45: I'm a little confused now. If things are valid during initialization, why not just only add `domain::Tags::InverseJacobian<Dim, Frame::ElementLogical, Frame::Inertial>` here so the compute tag does the work, and you don't have to calculate the coordinates explicitly yourself? If things aren't valid during initialization, could you explain why?

nilsdeppe: I don't know if they are valid now. They might be because of the initialization tag change Larry did, but there are several other places that compute it this way too, and I think we should change them all at once if we do change it.

Edit: and all this becomes a mess ones it interacts with subcell. I'd rather not have to fight with that yet again... Basically, this is the easiest implementation when considering subcell.

kidder: This depends upon whether you intend this function to be applicable for dg only, or for subcell as well.  If dg only, then you can just grab the appropriate compute tags.  If you want this to work for both dg and subcell, then you need to template on the types of any tags that are different for dg/subcell (currently the mesh and logical coordinates). (This is why I think SetVariables is templated on LogicalCoordinatesTag)

nilsdeppe: All of GH needs fairly extensive subcell support...

kidder (approved):

knelli2 (approved):
I'm ok with leaving this as is and then changing all the GH code at once later on.

PR #4451, opened 2022-12-03
Interpolator only recognizes targets that use the Interpolator

knelli2:
## Proposed changes

Previously, if two targets had the same temporal id type, but only one uses the Interpolator, the interpolator would hold onto the volume data even after the target that used the interpolator was finished. This is because the target that didn't use the interpolator never said it was finished with the volume data at that time. Therefore, volume data would never be purged from the interpolator and build up.

Now the interpolator only checks if the target that actually uses the Interpolator still needs the volume data. Added a test that checks for this.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceiveVolumeData.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceiveVolumeData.hpp
@@ -5,6 +5,7 @@
 
 #include <algorithm>
 #include <deque>
+#include <type_traits>
 #include <unordered_map>
 #include <utility>
 
@@ -22,6 +23,28 @@
 
 namespace intrp {
 namespace Actions {
+namespace detail {
+template <typename Metavariables, typename Tag, typename = std::void_t<>>
+struct get_interpolating_component_or_interpolator {
+  using type = Interpolator<Metavariables>;
+};
+template <typename Metavariables, typename Tag>
+struct get_interpolating_component_or_interpolator<
+    Metavariables, Tag,
+    std::void_t<
+        typename Tag::template interpolating_component<Metavariables>>> {
+  using type = typename Tag::template interpolating_component<Metavariables>;
+};
+template <typename Metavariables, typename Tag>
+using get_interpolating_component_or_interpolator_t =
+    typename get_interpolating_component_or_interpolator<Metavariables,
+                                                         Tag>::type;
+
+template <typename Metavariables, typename Tag>
+constexpr bool using_interpolator_component_v = std::is_same_v<
+    get_interpolating_component_or_interpolator_t<Metavariables, Tag>,
+    Interpolator<Metavariables>>;
+}  // namespace detail
 
 /// \ingroup ActionsGroup
 /// \brief Adds volume data from an `Element`.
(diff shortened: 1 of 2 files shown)


tests/Unit/ParallelAlgorithms/Interpolation/Test_InterpolatorReceiveAndDumpVolumeData.cpp
```diff
@@ -257,10 +257,14 @@ struct mock_interpolation_target {
 
   using replace_these_simple_actions =
       tmpl::list<intrp::Actions::InterpolationTargetReceiveVars<
-          typename Metavariables::InterpolationTargetA>>;
```
geoffrey4444, line 260: Please make the commit message into something that begins with a verb

tests/Unit/ParallelAlgorithms/Interpolation/Test_InterpolatorReceiveAndDumpVolumeData.cpp
```diff
@@ -313,12 +317,34 @@ struct MockMetavariables {
         intrp::callbacks::ObserveTimeSeriesOnSurface<tmpl::list<>,
                                                      InterpolationTargetA>;
   };
+  struct InterpolationTargetB
+      : tt::ConformsTo<intrp::protocols::InterpolationTargetTag> {
```
geoffrey4444: Maybe I'm being pedantic, but what if instead you avoided duplicating code by using a template parameter and if constexpr to decide whether or not to add the interpolating_component ... otherwise, aren't the two targets the same?

geoffrey4444 (changes requested):

tests/Unit/ParallelAlgorithms/Interpolation/Test_InterpolatorReceiveAndDumpVolumeData.cpp
knelli2: I intentionally want one target to have the type alias `interpolating_component` and one to not have the type alias. Couldn't do anything with `if constexpr` since they are type aliases, so I just made the second target inherit from the first one.

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceiveVolumeData.hpp
```diff
@@ -65,7 +88,14 @@ struct InterpolatorReceiveVolumeData {
     tmpl::for_each<typename Metavariables::interpolation_target_tags>(
         [&holders, &this_temporal_id_is_done, &temporal_id](auto tag_v) {
           using tag = typename decltype(tag_v)::type;
-          if constexpr (std::is_same_v<TemporalId, typename tag::temporal_id>) {
+          // First off, this interpolation target must actually be using the
+          // interpolator. If it's using some other component, no sense in
+          // holding the volume data at this temporal id for this target.
+          // Second, if this isn't the temporal id we're interested in, then
+          // don't do anything.
```
markscheel: Make this clearer? (in particular "don't do anything" is not completely clear).  Maybe something like
```
Here we decide whether this interpolation target is "done" (i.e. it does not
need to interpolate) at this temporal_id. If it is "done", then we don't 
need to store the volume data. Usually "done" means that it has already
done its interpolation at this temporal_id. But note that if an interpolation
target is not using the Interpolator at all, it is considered "done", because
we don't need any volume data for it and should not store it.
Similarly, interpolation targets whose temporal_id has a different type
than TemporalId are considered "done" because they too do not use
the Interpolator and don't need volume data to be saved.
```

markscheel (commented):
Thanks for finding this. Previously, targets that didn't use the interpolator had a different type for temporal_id.

One suggested change: clarifying the comment.  Please squash immediately.

markscheel (approved):

nilsdeppe (approved):

geoffrey4444 (approved):

PR #4456, opened 2022-12-04
clean up and rename Initialization tags

kidder:
## Proposed changes

- Rename initialization_tags to simple_tags_from_options (It makes it more clear what they are used for, i.e. adding simple tags to the DataBox for items initialized from input file options.)
- Get rid of array_allocation_tags in array components.  (In all cases, they end up being needed as simple tags in the DataBox; just make life simpler.)
- Decouple ResourceInfo from simple_tags_from_options (Makes life simpler.  Avoids using tag lists meant for DataBox initialization to initialize something in the GlobalCache.)
- Always add ResourceInfo to the GlobalCache and provide a reference item to fetch it from a Distributed Object's DataBox as well.

### Upgrade instructions

- Do not add ResourceInfo, AvoidGlobalProc0, or Singletons tags to tag lists. 
- If you have an input file, you need to specify a ResourceInfo option, and if the executable has Singletons, you must specify the Singletons sub-option.  To keep prior behavior use:
```
ResourceInfo:
  AvoidGlobalProc0: false
  Singletons: Auto
```
- If you have specified explicit options for a singleton, you will need to do so for all singletons.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/DevGuide/Parallelization.md
+++ b/docs/DevGuide/Parallelization.md
@@ -206,16 +206,16 @@ Each %Parallel Component struct must have the following type aliases:
    phase. If there are no iterable actions in a phase then a
    `PhaseAction` need not be specified for that phase. However, at
    least one `PhaseAction`, even if it is empty, must be specified.
-4. `using initialization_tags` which is a `tmpl::list` of all the tags
+4. `using simple_tags_from_options` which is a `tmpl::list` of all the tags
    that will be inserted into the initial `db::DataBox` of each component.
    These tags are db::SimpleTag%s that have have a `using option_tags`
    type alias and a static function `create_from_options` (see the
    example below).  This list can usually be constructed from the
    initialization actions of the component (i.e. the list of actions
    in the `PhaseAction` list for the `Initialization` phase) using the
-   helper function `Parallel::get_initialization_tags` (see the
+   helper function `Parallel::get_simple_tags_from_options` (see the
    examples of components below).  Each initialization action may
-   specify a type alias `using initialization_tags` which are a
+   specify a type alias `using simple_tags_from_options` which are a
    `tmpl::list` of tags that will be fetched from the db::DataBox by the
    action.
 5. `using const_global_cache_tags` is set to a `tmpl::list` of tags
@@ -257,7 +257,7 @@ signature of the `allocate_array` functions must be:
 \code
 static void allocate_array(
     Parallel::CProxy_GlobalCache<metavariables>& global_cache,
-    const tuples::tagged_tuple_from_typelist<initialization_tags>&
+    const tuples::tagged_tuple_from_typelist<simple_tags_from_options>&
     initialization_items,
     const std::unordered_set<size_t>& procs_to_ignore);
 \endcode
@@ -388,16 +388,16 @@ ResourceInfo:
 
 First is the `AvoidGlobalProc0` option. This will be available if any Array or
 Singleton component adds the `Parallel::Tags::AvoidGlobalProc0` tag to its
-`initialization_tags` type alias. This option will tell the program to not put
-*any* Array Elements or Singletons on the global zeroth charm-core. This core
-is sometimes used to write data to disk which is typically much slower than the
-program execution. The second is the `Singletons:` option. This will be
+`simple_tags_from_options` type alias. This option will tell the program to not
+put *any* Array Elements or Singletons on the global zeroth charm-core. This
+core is sometimes used to write data to disk which is typically much slower
+than the program execution. The second is the `Singletons:` option. This will be
 available for every Singleton that adds the `Parallel::Tags::SingletonInfo` tag
-to its `initialization_tags` type alias. `AhA` is the `pretty_type::name()` of
-a Singleton in the program and the user has a choice of which proc to place the
-Singleton on (`Auto` will let the program decide) and whether to exclude Array
-Elements or other Singletons from being put on this core. This is useful in
-case the Singleton does some expensive computation that shouldn't be slowed
+to its `simple_tags_from_options` type alias. `AhA` is the `pretty_type::name()`
+of a Singleton in the program and the user has a choice of which proc to place
+the Singleton on (`Auto` will let the program decide) and whether to exclude
+Array Elements or other Singletons from being put on this core. This is useful
+in case the Singleton does some expensive computation that shouldn't be slowed
 down by having lots of Array Elements on the same core. In the figure above,
 `AvoidGlobalProc0` is true, and `Sing. 2` requested to be exclusively on core
 `2`.
--- a/docs/DevGuide/Parallelization.md
+++ b/docs/DevGuide/Parallelization.md
@@ -223,10 +223,7 @@ Each %Parallel Component struct must have the following type aliases:
    component, or simple actions called on the parallel component.
    These tags correspond to const items that are stored in the
    Parallel::GlobalCache (of which there is one copy per Charm++
-   node).  The alias can be omitted if the list is empty.  (See
-   `array_allocation_tags` below for specifying tags needed for the
-   `allocate_array` function, but will not be added to the
-   Parallel::GlobalCache.)
+   node).  The alias can be omitted if the list is empty.
 6. `using mutable_global_cache_tags` is set to a `tmpl::list` of tags
    that correspond to mutable items that are stored in the
    Parallel::GlobalCache (of which there is one copy per Charm++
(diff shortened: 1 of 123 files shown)


src/Parallel/ResourceInfo.hpp
```diff
@@ -249,24 +249,24 @@ bool operator!=(const SingletonPack<tmpl::list<Components...>>& lhs,
...
+// are still constructed regardless. So rather than changing every mock
+// component to have an empty type alias, we just treat no type alias as not
+// having the tag we are looking for.
+CREATE_HAS_TYPE_ALIAS(simple_tags_from_options)
+CREATE_HAS_TYPE_ALIAS_V(simple_tags_from_options)
 
 template <typename Component, typename Tag,
-          bool HasInitializationTags = has_initialization_tags_v<Component>>
+          bool HasInitializationTags =
+              has_simple_tags_from_options_v<Component>>
```
knelli2: `HasInitializationTags` -> `HasSimpleTagsFromOptions`

This change is optional because this code will just be removed in a later commit so doesn't really matter if it's fixed or not.

docs/DevGuide/Parallelization.md
```diff
@@ -264,14 +261,16 @@ static void allocate_array(
 The `allocate_array` function is called by the Main parallel component
 when the execution starts and will typically insert elements into
 array parallel components. If the `allocate_array` function depends
-upon input options, the array component must specify a `using
-array_allocation_tags` type alias that is a `tmpl::list` of tags which
+upon input options that are not in the GlobalCache, the array component must
+include the appropriate tag in the `simple_tags_from_options` type alias that
+is a `tmpl::list` of tags which
```
knelli2: This sentence is very long and I even got confused for a second. Can you do 
```
`simple_tags_from_options` type alias. This type alias is a `tmpl::list` ...
```

src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -49,9 +49,7 @@ template <size_t Dim>
 struct DefaultElementsAllocator
     : tt::ConformsTo<Parallel::protocols::ArrayElementsAllocator> {
   template <typename ParallelComponent>
-  using array_allocation_tags =
-      tmpl::list<domain::Tags::InitialRefinementLevels<Dim>,
-                 domain::Tags::InitialExtents<Dim>>;
+  using array_allocation_tags = tmpl::list<>;
```
knelli2, line 51: Why keep this empty type alias? Is it needed to compile?

src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -151,8 +149,8 @@ struct DgElementArray {
   using simple_tags_from_options =
       tmpl::append<Parallel::get_simple_tags_from_options<
                        Parallel::get_initialization_actions_list<
-                           phase_dependent_action_list>,
-                       array_allocation_tags>,
+                           phase_dependent_action_list>>,
+                   array_allocation_tags,
```
knelli2: If this is type alias is empty (but still necessary), why include it here?

src/Parallel/GlobalCache.hpp
```diff
@@ -479,6 +476,11 @@ class GlobalCache : public CBase_GlobalCache<Metavariables> {
   void set_resource_info(
       const Parallel::ResourceInfo<Metavariables>& resource_info);
 
+  /// Retrieve the resource_info
```
knelli2, line 479: Could you add that you have to call this on the `Parallel::local_branch` of the GlobalCache? Not really sure what would happen if you call this on a proxy. I guess it would fail to build since the proxy doesn't have the function, but I think still good to be explicit.

tests/InputFiles/GeneralizedHarmonic/BinaryBlackHole.yaml
```diff
@@ -19,6 +19,15 @@ ResourceInfo:
...
       Exclusive: true
+    MemoryMonitor:
+      Proc: Auto
+      Exclusive: false
+    Rotation:
+      Proc: Auto
+      Exclusive: false
+    Expansion:
+      Proc: Auto
+      Exclusive: false
```
knelli2, line 30: Not a change request for you, but for me. After this PR is merged, I should make the `SingletonPack` use an `Options::Auto<SingletonInfoHolder>` instead of just a `SingletonInfoHolder` so these lines can just be

```yaml
    MemoryMonitor: Auto
    Rotation: Auto
    Expansion: Auto
```

which looks a lot cleaner imo.

tests/Unit/Parallel/Test_ResourceInfo.cpp
```diff
@@ -271,62 +208,31 @@ void test_single_core() {
...
     resource_info.build_singleton_map(cache);
 
-    const size_t proc_0 =
-        resource_info.template proc_for<FakeSingletonInfoOnly<metavars, 0>>();
-    const size_t proc_1 =
-        resource_info.template proc_for<FakeSingletonInfoOnly<metavars, 1>>();
-
-    // Only running on once proc
-    CHECK(proc_0 == 0);
-    CHECK(proc_1 == 0);
```
knelli2: Could you keep these checks in? I know the procs should be the same as the ones below that were gotten from the singleton info's, but that's what we're testing. That they are the same, just gotten different ways.

knelli2 (changes requested):
Thanks for doing this! I didn't like all the hoops I had to jump through when the ResourceInfo wasn't required in all executables. This is much cleaner.

src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
kidder: There is another ElementAllocator `src/ParallelAlgorithms/LinearSolver/Multigrid/ElementsAllocator.hpp` which has a non-empty `array_allocation_tags`

src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
kidder: I chose that all ElementAllocators should explicitly define the list even if it is empty (see below)

src/Parallel/GlobalCache.hpp
kidder: but that is true of any non-entry member function of any component, so seems a bit weird to make that explicit for this particular function

tests/InputFiles/GeneralizedHarmonic/BinaryBlackHole.yaml
kidder: Oh, yes I like this

src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
knelli2: That makes sense, except the protocol in `src/Parallel/Protocols/ArrayElementsAllocator.hpp` doesn't require the `array_allocation_tags` type alias at all. It actually doesn't require anything. I'm not suggesting you change anything, but @nilsvu why is the protocol empty?

src/Parallel/GlobalCache.hpp
knelli2: You're right, I hadn't thought of that. Ok as is.

src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
nilsvu: It's documented!

knelli2: Where? I'm not seeing it [here](https://spectre-code.org/structParallel_1_1protocols_1_1ArrayElementsAllocator.html).

nilsvu: I see this:

Bildschirm­foto 2022-12-06 um 11 43 15 AM

kidder:
@knelli2 pushed fixups including more updates to documentation

src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
knelli2: Ok yeah it's hard to check type aliases that are templated so this is alright.

knelli2 (commented):
Fixups look good. You can squash.

knelli2 (approved):

nilsdeppe (approved):

PR #4457, opened 2022-12-05
Parallelize ExtractDatFromH5.py

knelli2:
## Proposed changes

I noticed while trying to extract dat files from a run with CCE that if the files being extracted are fairly large (~1G or larger) this script runs pretty slow. Mostly due to writing large amounts of data to disk. Parallelizing the writing portion sped things up significantly.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/H5/Python/ExtractDatFromH5.py
+++ b/src/IO/H5/Python/ExtractDatFromH5.py
@@ -3,24 +3,44 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
-from spectre.IO import H5 as spectre_h5
+from spectre.Visualization.ReadH5 import available_subfiles
+import h5py
 import click
-import spectre.DataStructures
+import multiprocessing as mp
 import numpy as np
 import os
 import shutil
 
 
-def extract_dat_files(filename, out_dir, list=False, force=False):
+def save_dat_data(dat_path, h5_filename, out_dir):
+    dat_dir = os.path.join(out_dir, os.path.dirname(dat_path))
+    os.makedirs(dat_dir, exist_ok=True)
+    dat_filename = os.path.join(out_dir, dat_path)
+
+    with h5py.File(h5_filename, "r") as h5file:
+        dat_file = h5file.get(dat_path)
+        legend = dat_file.attrs["Legend"]
+        dat_data = np.array(dat_file)
+
+    header = "\n".join(f"[{i}] " + "{}"
+                       for i in range(len(legend))).format(*legend)
+
+    np.savetxt(dat_filename,
+               dat_data,
+               delimiter=' ',
+               fmt="% .15e",
+               header=header)
+
+
+def extract_dat_files(filename, out_dir, num_cores, list=False, force=False):
     """Extract dat files from an H5 file
 
     Extract all Dat files inside a SpECTRE HDF5 file. The resulting files will
     be put into the 'OUT_DIR'. The directory structure will be identical to the
     group structure inside the HDF5 file.
     """
-    h5file = spectre_h5.H5File(filename, "r")
-
-    all_dat_files = h5file.all_dat_files()
+    with h5py.File(filename, "r") as h5file:
+        all_dat_files = available_subfiles(h5file, extension=".dat")
 
     if list:
         print_str = "\n ".join(all_dat_files)
@@ -40,28 +60,19 @@ def extract_dat_files(filename, out_dir, list=False, force=False):
             raise ValueError(
                 f"Could not make directory '{out_dir}'. Already exists.")
 
-    for dat_path in all_dat_files:
-        split_path = dat_path.split("/")
-        dat_dir = out_dir + "/".join(split_path[:-1])
-        dat_filename = out_dir + dat_path
+    num_dat_files = len(all_dat_files)
 
-        os.makedirs(dat_dir, exist_ok=True)
-
-        dat_file = h5file.get_dat(dat_path[:-4])
-
-        legend = dat_file.get_legend()
-        header = "\n".join(f"[{i}] " + "{}"
-                           for i in range(len(legend))).format(*legend)
-
-        dat_data = np.array(dat_file.get_data())
-
-        np.savetxt(dat_filename,
-                   dat_data,
-                   delimiter=' ',
-                   fmt="% .15e",
-                   header=header)
-
-        h5file.close()
+    # Only use multiprocessing if we are using more than one core. Otherwise
+    # avoid the overhead
+    if num_cores > 1:
+        with mp.Pool(processes=num_cores) as pool:
+            pool.starmap(
+                save_dat_data,
+                zip(all_dat_files, [filename] * num_dat_files,
+                    [out_dir] * num_dat_files))
+    else:
+        for dat_filename in all_dat_files:
+            save_dat_data(dat_filename, filename, out_dir)
 
     print(f"Successfully extracted all Dat files into '{out_dir}'")
 
@@ -75,6 +86,11 @@ def extract_dat_files(filename, out_dir, list=False, force=False):
 @click.argument("out_dir",
                 type=click.Path(file_okay=False, dir_okay=True, writable=True),
                 required=False)
+@click.option('--num-cores',
+              '-j',
+              default=1,
+              show_default=True,
+              help="Number of cores to run on.")
 @click.option('--force',
               '-f',
               is_flag=True,
--- a/src/IO/H5/Python/ExtractDatFromH5.py
+++ b/src/IO/H5/Python/ExtractDatFromH5.py
@@ -24,7 +24,7 @@ def extract_dat_files(filename, out_dir, list=False, force=False):
 
     if list:
         print_str = "\n ".join(all_dat_files)
-        print("Dat files within '{}':\n {}".format(filename, print_str))
+        print(f"Dat files within '{filename}':\n {print_str}")
         return
 
     if out_dir is None:
@@ -38,8 +38,7 @@ def extract_dat_files(filename, out_dir, list=False, force=False):
             os.mkdir(out_dir)
         else:
             raise ValueError(
-                "Could not make directory '{}'. Already exists.".format(
-                    out_dir))
+                f"Could not make directory '{out_dir}'. Already exists.")
 
     for dat_path in all_dat_files:
         split_path = dat_path.split("/")
@@ -51,7 +50,7 @@ def extract_dat_files(filename, out_dir, list=False, force=False):
         dat_file = h5file.get_dat(dat_path[:-4])
 
         legend = dat_file.get_legend()
-        header = "\n".join("[{}] ".format(i) + "{}"
+        header = "\n".join(f"[{i}] " + "{}"
                            for i in range(len(legend))).format(*legend)
 
         dat_data = np.array(dat_file.get_data())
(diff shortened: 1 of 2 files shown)


src/IO/H5/Python/ExtractDatFromH5.py
```diff
@@ -32,10 +40,20 @@ def extract_dat_files(filename, **kwargs):
             raise ValueError(
                 f"Could not make directory '{out_dir}'. Already exists.")
 
+    # We gather all the data first because
+    # 1. The slowest part is actually writing, so that's what we want to
+    #    parallelize (if we are on multiple cores)
+    # 2. So we don't have to worry about accessing an h5file in parallel which
+    #    multiprocessing would have to do if we parsed and wrote in parallel
```
nilsvu: Are you sure this actually speeds up the script? The reason I'm asking is that it sounds like the bottleneck here is IO, not the CPU. If an operation is IO bound then parallelizing it to more CPUs is unlikely to speed it up, since the file system is busy with writing files anyway and can't do it any faster (It could actually slow down). I'm confused how the parallelization would help here. Maybe it's the a caching thing (wild guess): switching back and forth between reading the H5 file and writing a text file could be slower than reading everything out of the H5 file first, then writing. I don't know enough about file systems to know for sure. In that case you could just delete the multiprocessing part and get the same speedup. Maybe try that and see if that speeds up the script already?

Then again, loading all dat files into memory before writing them out sounds like it could exceed the RAM, at least in the slow case when the data files are O(GB). Not sure really what to do about this.

Can't we just plot data out of the H5 file? If all you want to do is plot our H5 data with gnuplot (or similar) then take a look here: https://groups.google.com/g/comp.graphics.apps.gnuplot/c/mIUjokGVsSU No need to do expensive extractions. Actually `h5totxt` sounds like exactly what this script does.

nilsvu (changes requested):

src/IO/H5/Python/ExtractDatFromH5.py
knelli2: So I tested this on both my local machine and on wheeler. On both machines I was extracting dat files from an H5 file that was about ~4G (they were different files but they were close enough in size). Here is the time it took on each to extract the files:

| Case | Local | Wheeler |
| -- | -- | -- |
| Develop | 4m 37s | 15m 30s |
| This PR on 1 core | 5m 6s | 15m 48s |
| This PR on 8 cores | 1m 14s | 1m 40s |

This PR on 1 core is exactly your suggestion of gathering data first, then writing it on 1 core (no multi-threading happens). So yes I am sure that these changes speed up the script a lot when trying to extract large files, presumably because there's just more cores writing data to different files in parallel.

You make a good point though about RAM usage. What I should probably do is add another option where you can specify which root groups of the H5 file you want to extract. That way you can just exclude the group that has most of the data (\*cough\* cce \*cough\*)

nilsvu: Alright I probably don't understand filesystems well enough. In that case just don't extract all data first before writing but dispatch to multiprocessing right away, then read + write. Reading from a single h5 file in parallel "should be fine" ™️.

Before doing that though: Why not just use `h5totxt` from https://github.com/NanoComp/h5utils. We don't have to reinvent the wheel. I tried it today and it works well :) Even with gnuplot.

nilsvu: If you want to try on wheeler:

```
module load /home/nilsvu/spack/share/spack/modules/linux-rhel7-haswell/h5utils-1.13.1-gcc-9.3.0-xinhocl
h5totxt --help
```

wthrowe: I've suspected for a while that the bottleneck for H5 IO is processing done in the hdf5 library rather than actual reading and writing.

nilsvu: @wthrowe that sounds plausible but it's not supported by Kyle's findings here I think. He parallelized only the `np.savetxt` part of the script.

wthrowe: You're right.  That is weird.

knelli2: So I've tried `h5totxt` and it does work, but it's a bit clunky. The txt files it writes aren't nicely formatted, it can only do one subfile at a time, not a whole group, and in `gnuplot` plotting multiple dat files from multiple h5 files becomes cumbersome. For those reasons, I'd like to keep this utility.

Now it's just a question of what do. I've shown that this parallelization does in fact speed up extraction on a local machine and on a cluster. Are you (@nilsvu) ok with adding this in? I also intend to add a feature where you can only dump certain groups and not others. This will help so we don't even have to worry about large groups. What do you think?

nilsvu: Sure, I'm fine with keeping this tool around. To avoid the RAM issue just parallelize the whole read + write routine (see above: https://github.com/sxs-collaboration/spectre/pull/4457#discussion_r1044949107)

knelli2: Ok was able to do this with `h5py`. Our python bindings for some reason can't open read-only h5 files in parallel.

nilsvu: Yep we should fix that! Can you open an issue with the error message?

src/IO/H5/Python/ExtractDatFromH5.py
```diff
@@ -3,24 +3,46 @@
...
 
 
-def extract_dat_files(filename, out_dir, list=False, force=False):
+def save_dat_data(dat_path, h5_filename, out_dir):
+    with h5py.File(h5_filename, "r") as h5file:
+        split_path = dat_path.split("/")
+        dat_dir = os.path.join(out_dir, "/".join(split_path[:-1]))
+        dat_filename = os.path.join(out_dir, dat_path)
+
+        os.makedirs(dat_dir, exist_ok=True)
```
nilsvu: Move this before the `with`. Also I suggest:

```py
dat_dir = os.path.dirname(dat_path)
```

Then you don't need `split_path`.

src/IO/H5/Python/ExtractDatFromH5.py
```diff
@@ -3,24 +3,46 @@
...
+
+        dat_file = h5file.get(dat_path)
+
+        legend = dat_file.attrs["Legend"]
+        header = "\n".join(f"[{i}] " + "{}"
+                           for i in range(len(legend))).format(*legend)
```
nilsvu: move this down, out of the `with`

src/IO/H5/Python/ExtractDatFromH5.py
```diff
@@ -3,28 +3,50 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
-from spectre.IO import H5 as spectre_h5
+from spectre.Visualization.PlotDatFile import available_subfiles
```
nilsvu: Import from `spectre.Visualization.ReadH5` now

src/IO/H5/Python/ExtractDatFromH5.py
```diff
@@ -76,6 +88,11 @@ def extract_dat_files(filename, out_dir, list=False, force=False):
 @click.argument("out_dir",
                 type=click.Path(file_okay=False, dir_okay=True, writable=True),
                 required=False)
+@click.option('--num-cores',
+              '-j',
+              default=1,
```
nilsvu, line 91: Set default to `None` so the script uses all available cores automatically? If you want to do that, set `show_default="all available cores"` or similar.
You also need to adjust the `if num_cores > 1` clause above, or just delete it because the overhead it saves is probably not worth the effort :)

tests/Unit/IO/H5/Python/Test_ExtractDatFromH5.py
```diff
@@ -44,7 +44,7 @@ def test_extract_dat_files(self):
         h5file.close()
 
         # All defaults, data should be same as expected
-        extract_dat_files(self.h5_filename, out_dir=self.created_dir)
+        extract_dat_files(self.h5_filename, self.created_dir, 1)
```
nilsvu: Always use named arguments in Python when it improves clarity. Here it does, so write `out_dir=` and `num_cores=`.

tests/Unit/IO/H5/Python/Test_ExtractDatFromH5.py
```diff
@@ -60,15 +60,22 @@ def test_extract_dat_files(self):
 
         # Test that we get an error if we try to run again and 'force' flag is
         # False.
-        self.failUnlessRaises(ValueError,
-                              extract_dat_files,
-                              self.h5_filename,
-                              out_dir=self.created_dir)
+        self.failUnlessRaises(ValueError, extract_dat_files, self.h5_filename,
+                              self.created_dir, 1)
```
nilsvu: Same comment about named arguments

tests/Unit/IO/H5/Python/Test_ExtractDatFromH5.py
```diff
@@ -60,15 +60,22 @@ def test_extract_dat_files(self):
...
-                          force=True)
+        extract_dat_files(self.h5_filename, self.created_dir, 1, force=True)
+
+        memory_data = np.loadtxt(memory_data_path)
+        timestep_data = np.loadtxt(timestep_data_path)
+
+        self.assertEqual(memory_data.all(), expected_memory_data.all())
+        self.assertEqual(timestep_data.all(), expected_timestep_data.all())
+
+        self.tearDown()
```
nilsvu: Don't call tearDown yourself (is that safe?). Just split into a separate function if you need to.

nilsvu (changes requested):

src/IO/H5/Python/ExtractDatFromH5.py
knelli2: I'm very hesitant to have the default all available cores because of "head-node etiquette" on clusters. Hopefully you aren't extracting so many big files that you use up a full head-node, but to just avoid this altogether I'd like to keep the default at 1. The user is then responsible for paralleling things.

nilsvu: ok 👍

tests/Unit/IO/H5/Python/Test_ExtractDatFromH5.py
```diff
@@ -76,8 +80,22 @@ def test_extract_dat_files(self):
...
+        # Parallelize. Use 2 cores (this is all we get on CI)
+        extract_dat_files(self.h5_filename,
+                          out_dir=self.created_dir,
+                          num_cores=2)
+
+        memory_data = np.loadtxt(memory_data_path)
+        timestep_data = np.loadtxt(timestep_data_path)
+
+        self.assertEqual(memory_data.all(), expected_memory_data.all())
+        self.assertEqual(timestep_data.all(), expected_timestep_data.all())
```
nilsvu: I don't think this does what you want. The `.all()` actually evaluates the boolean value of all elements in the array, which is true if all elements are nonzero. To compare the two arrays, `import numpy.testing as npt` and use `npt.assert_almost_equal`. Fix occurrence above too please.

nilsvu (changes requested):

nilsvu (approved):

PR #4462, opened 2022-12-06
Retrieve coordinate derivative of pressure from TOV star solution 

yoonso0-0:
## Proposed changes

Add `()` operator returning dp/dx^i of TOV solution. This quantity is needed to compute initial magnetic fields.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.cpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.cpp
@@ -461,6 +461,32 @@ void TovVariables<DataType, Region>::operator()(
 
 template <typename DataType, StarRegion Region>
 void TovVariables<DataType, Region>::operator()(
+    const gsl::not_null<tnsr::i<DataType, 3>*> deriv_pressure,
+    [[maybe_unused]] const gsl::not_null<Cache*> cache,
+    ::Tags::deriv<hydro::Tags::Pressure<DataType>, tmpl::size_t<3>,
+                  Frame::Inertial> /*meta*/) const {
+  // This function is not required for solving TOV equations themselves, but
+  // needed for computing magnetic fields in the MagnetizedTov initial data.
+
+  if constexpr (Region == StarRegion::Center or
+                Region == StarRegion::Exterior) {
+    get<0>(*deriv_pressure) = 0.0;
+    get<1>(*deriv_pressure) = 0.0;
+    get<2>(*deriv_pressure) = 0.0;
+  } else {
+    // cache evaluation only gets triggered when it's needed
+    const auto& dr_pressure =
+        get(cache->get_var(*this, Tags::DrPressure<DataType>{}));
+
+    // dp/dx^i = (dr/dx^i)(dp/dr) = (x^i/r)(dp/dr)
+    for (size_t d = 0; d < 3; ++d) {
+      (*deriv_pressure).get(d) = dr_pressure * coords.get(d) / radius;
+    }
+  }
+}
+
+template <typename DataType, StarRegion Region>
+void TovVariables<DataType, Region>::operator()(
     const gsl::not_null<Scalar<DataType>*> lorentz_factor,
     const gsl::not_null<Cache*> /*cache*/,
     hydro::Tags::LorentzFactor<DataType> /*meta*/) const {
--- a/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.hpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.hpp
@@ -99,7 +99,10 @@ using TovVariablesCache = cached_temp_buffer_from_typelist<tmpl::list<
     hydro::Tags::SpecificEnthalpy<DataType>,
     hydro::Tags::RestMassDensity<DataType>,
     hydro::Tags::ElectronFraction<DataType>, hydro::Tags::Pressure<DataType>,
-    Tags::DrPressure<DataType>, hydro::Tags::SpecificInternalEnergy<DataType>,
+    Tags::DrPressure<DataType>,
+    ::Tags::deriv<hydro::Tags::Pressure<DataType>, tmpl::size_t<3>,
+                  Frame::Inertial>,
+    hydro::Tags::SpecificInternalEnergy<DataType>,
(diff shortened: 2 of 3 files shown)


src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.hpp
```diff
@@ -66,6 +66,10 @@ struct DrPressure : db::SimpleTag {
   using type = Scalar<DataType>;
 };
 template <typename DataType>
+struct DcoordsPressure : db::SimpleTag {
```
nilsvu: No need to add this tag. Just use `::Tags::deriv<hydro::Tags::Pressure<DataType>, tmpl::size_t<3>, Frame::Inertial>`.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.cpp
```diff
@@ -459,6 +459,39 @@ void TovVariables<DataType, Region>::operator()(
   }
 }
 
+template <typename DataType, StarRegion Region>
+void TovVariables<DataType, Region>::operator()(
+    const gsl::not_null<tnsr::i<DataType, 3>*> dcoords_pressure,
```
nilsvu: `deriv_pressure` for consistency with the rest of the file

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.cpp
```diff
@@ -459,6 +459,39 @@ void TovVariables<DataType, Region>::operator()(
...
+    [[maybe_unused]] const gsl::not_null<Cache*> cache,
+    Tags::DcoordsPressure<DataType> /*meta*/) const {
+  const auto& dr_pressure =
+      get(cache->get_var(*this, Tags::DrPressure<DataType>{}));
```
nilsvu: move into the `else` branch so the cache evaluation only gets triggered when it's needed

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.cpp
```diff
@@ -459,6 +459,39 @@ void TovVariables<DataType, Region>::operator()(
...
+
+  // This function is not required for solving TOV equations themselves, but
+  // needed for computing magnetic fields in the MagnetizedTov initial data.
+  if constexpr (Region == StarRegion::Exterior) {
+    (*dcoords_pressure).get(0) = 0.0;
+    (*dcoords_pressure).get(1) = 0.0;
+    (*dcoords_pressure).get(2) = 0.0;
```
nilsvu: use compile-time version: `get<0>(*deriv_pressure)`

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.cpp
```diff
@@ -459,6 +459,39 @@ void TovVariables<DataType, Region>::operator()(
...
+  } else {
+    for (size_t i = 0; i < get_size(radius); ++i) {
+      const auto& ri = get_element(radius, i);
+      // dp/dx^i goes to zero at the center
+      if (UNLIKELY(ri < 1e-15)) {
```
nilsvu: You don't need to do this branching, just add an `if constexpr (Region == StarRegion::Center)` clause where you set the pressure gradient at the center to zero. Then remove the loop over grid points.

nilsvu (changes requested):

nilsvu:
Could you add a tiny little test? The smallest you can come up with :D

yoonso0-0:
> Could you add a tiny little test? The smallest you can come up with :D

@nilsvu I tried to find where the tests for derivatives (like `DrPressure` or `dt<Lapse>`) are written in for TOV, but couldn't find it myself. Can you point a good spot to put it in?

nilsvu:
The other variables are tested as part of `verify_grmhd_solution`. Is this one tested in your upcoming magnetized TOV star PR? That would be fine too I guess.

yoonso0-0:
@nilsvu I came to a conclusion that this is something that `Test_TovStar` should check, so added a few lines for that :)

tests/Unit/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Test_TovStar.cpp
```diff
@@ -42,6 +42,32 @@ void verify_solution(const TovStar& solution, const std::array<double, 3>& x) {
   const auto domain = brick.create_domain();
   verify_grmhd_solution(solution, domain.blocks()[0], mesh, 1.e-7, 1.234,
                         1.e-4);
+
+  // Check that analytically computed dp/dx^i actually equals numerical
+  // derivative of pressure
```
nilsvu, line 47: Looks good! I think you can just use `numerical_derivative` from TestHelpers.hpp here. Something like this should do it:

```cpp
  const auto dpdx_numerical = numerical_derivative(
      [&solution](const std::array<double, 3>& local_x) {
        return std::array<double, 1>{get(get<Pressure>(solution.variables(
            tnsr::I<double, 3>(local_x), 0., tmpl::list<Pressure>{})))};
      },
      x, d, 1e-4);
```

nilsvu (changes requested):
Please squash, including this

tests/Unit/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Test_TovStar.cpp
yoonso0-0: Ah, thank you!

yoonso0-0:
Squashed fixes.

nilsvu (dismissed):

nilsvu:
There's a gcc-9 error. Just silence it by adding `(void)point_index` here: https://github.com/sxs-collaboration/spectre/blob/e1f8435693ee697dd3e1e98deab70680ce95083a/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/TovStar.hpp#L454

nilsvu:
Actually, nvm, I fixed this in #4538, so just rebase :)

yoonso0-0:
Thank you!!

nilsvu (approved):

PR #4271, opened 2022-10-05
GH+CCE for a gauge wave

Sizheng-Ma:
## Proposed changes

This PR adds a GH+CCE executable for a gauge wave (without horizons). It lays the foundation for CCM. The system is **not** compatible with the local time stepping.

### Upgrade instructions

I add one more template to `SendGhWorldtubeData` so that it can distinguish between `LTS` and `GTS`. Please remember to update your `CceWorldtubeTarget` @knelli2 

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4109, opened 2022-07-12
Add BBH+CCE

knelli2:
## Proposed changes

Adds CCE to a new BBH executable. To run CCE, you have to add the `CceWorltubeTarget` Event to the `EventsAndDenseTriggers:` block of the input file.

~It should be noted that there is a known issue with this that around 200M, the CCE timestep gets too small for the evolution to continue. Currently we don't have enough diagnostics in CCE to debug this, so I will be adding those in later.~ This was solved in #4239.

### Upgrade instructions

To run CCE in the BBH exectuable, you have to add the `CceWorltubeTarget` Event to the `EventsAndDenseTriggers:` block of the input file along with other CCE options which can be found in the CharacteristicExtract example input file.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

~Depends on and includes #4107.~
~Depends on and includes #4239.~
Depends on and includes #4451

PR #4466, opened 2022-12-07
Decouple metavariables and LTS in step chooser simple tags

knelli2:
## Proposed changes

I encountered this issue while helping @Sizheng-Ma rebase #4271 on the latest develop. The changes in #4358 made it so that LTS step choosers were added to the list of all step choosers only if the `local_time_stepping` type alias in the metavariables was true. However, in #4271, the GH evolution uses GTS, but the CCE evolution uses LTS. This caused the LTS step choosers to not be added to the list of all step choosers for CCE and as a result, some step chooser simple tags were missing from the CCE DataBox.

My solution was instead of using `Metavariables::local_time_stepping` in the `all_step_choosers` type alias in `src/Time/StepChoosers/StepChooser.hpp` to determine if the LTS choosers should be added, pass in a second template bool indicating if we are using LTS. That way different evolution systems can specify if they are using LTS or GTS.

I didn't run into this issue in #4109 because the GH evolution uses LTS there and so the LTS step choosers were added to all step choosers regardless.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/Burgers/EvolveBurgers.hpp
+++ b/src/Evolution/Executables/Burgers/EvolveBurgers.hpp
@@ -303,7 +303,8 @@ struct EvolutionMetavars {
           tmpl::list<>>,
 
       Initialization::Actions::AddComputeTags<
-          StepChoosers::step_chooser_compute_tags<EvolutionMetavars>>,
+          StepChoosers::step_chooser_compute_tags<EvolutionMetavars,
+                                                  local_time_stepping>>,
       ::evolution::dg::Initialization::Mortars<volume_dim, system>,
       Initialization::Actions::Minmod<1>,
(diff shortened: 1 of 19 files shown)


knelli2:
@wthrowe Please let me know if you'd like a different solution. This was just the quickest fix @nilsdeppe and I came up with on the fly.

nilsvu (approved):
We should do this more 👍

nilsdeppe (approved):

wthrowe (approved):
Really, all these lists should be removed, but I guess that's a bigger project.

The ErrorControl tags belong in `UpdateU` and the other things that take time steps, the others should be using `ObservationBox`.

PR #4469, opened 2022-12-07
CylindricalBinaryCompactObject: use new maps.

markscheel:
Now uses UniformCylindricalEndcap, UniformCylindricalSide, and UniformCylindricalFlatEndcap instead of the old maps CylindricalEndcap, CylindricalSide, CylindricalFlatEndcap.

Also now has the option to add an extra outer cylindered-sphere shell, and to add an extra inner cylindered-sphere shell for either of the excised regions.

Note that instead of making a new CylindricalBinaryCompactObject, here we replace the old one; same with the test.  Eventually the old maps should be deprecated and removed.

Does not yet handle time-dependent maps in the way we want it to, but that can be the next PR.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/Creators/CylindricalBinaryCompactObject.hpp
```diff
@@ -93,11 +96,26 @@ namespace domain::creators {
...
- *   blocks. They all border the inner boundary "A".
+ *   blocks. They all border the inner boundary "A" if
+ *   `IncludeInnerSphereA` is false.
+ * - If `IncludeInnerSphereA` is true, then there are new blocks
+ *   InnerSphereEAFilledCylinder, InnerSphereMAFilledCylinder, and
+ *   InnerSphereEACylinder. These are in a Block group called "InnerSphereA",
+ *   and all of these border the inner excision boundary "A".
  * - The Block group called "InnerB" consists of all the EB, and MB
- *   blocks. They all border the inner boundary "B".
+ *   blocks. They all border the inner boundary "B"if
```
knelli2: Missing space between `"B"` and `if`

src/Domain/Creators/CylindricalBinaryCompactObject.hpp
```diff
@@ -79,8 +78,12 @@ namespace domain::creators {
...
- * blocks extend to the excision boundary and the "CA" and "CB" blocks
+ * has these extra shells inside "EA" only if the option `IncludeInnerSphereA`
+ * is true, it has the extra shells inside "EB" only if the option
+ * `IncludeInnerSphereB` is true, and it has the extra shells outside
+ * "CA" and "CB" only if `IncludeOuterSphere` is true.
+ * If the shells are absent, then the "EA" and "EB"
+ * blocks extend to the excision boundaries and the "CA" and "CB" blocks
  * extend to the outer boundary.
  *
  * The Blocks are named as follows:
```
knelli2, line 89: The line below this is missing a comma. Can't reach it with code comments though. I know this isn't related to your changes but I just noticed it while reading the docs.

knelli2 (changes requested):
Just some minor dox things. Squash directly. Other than that, looks great! Excited to try this out soon!

markscheel:
Made changes and squashed.
Also added a new fixup commit that removes some strange codecov errors.

knelli2 (commented):
Fixups look good, you can squash. Yeah codecov occasionally does really stupid things like saying a function call is not tested when it most definitely is. No clue why.

markscheel:
Squashed.

knelli2 (approved):

geoffrey4444 (approved):
I don't have any further code suggestions to add. 

I tested this domain (with #4526 and #4491) and was able to evolve a BBH with it. I noticed constraints in the added inner spheres were large and did not converge well with increasing refinement there. I suspect we want something like the Logarithmic spacing option in BinaryCompactObject, but that would be for a future PR.

nilsdeppe (approved):

PR #4472, opened 2022-12-07
Add `InterpolateVolumeData.py` to Python CLI

nilsvu:
### Upgrade instructions

Invoke `spectre interpolate-vol -h` to run the `InterpolateVolumeData` script.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/InterpolateVolumeData.py
+++ b/src/Visualization/Python/InterpolateVolumeData.py
@@ -79,10 +79,10 @@ def interpolate_h5_file(source_file_path,
         same file, but in a different subfile specified by `target_volume_data`.
     source_volume_data: str, optional
         the name of the .vol file inside the source file where the source data
-        can be found. Requires leading `/` but no `.vol` file extension.
+        can be found.
     target_volume_data: str, optional
         the name of the .vol file inside the target file where the target data
-        is written. Requires leading `/` but no `.vol` file extension.
+        is written.
     obs_start: float, optional
         disregards all observations with observation value strictly before
         `obs_start`
@@ -92,6 +92,13 @@ def interpolate_h5_file(source_file_path,
     obs_stride: float, optional
         will only take every `obs_stride` observation
     """
+
+    source_volume_data = source_volume_data.rstrip(".vol")
+    target_volume_data = target_volume_data.rstrip(".vol")
+    if not source_volume_data.startswith("/"):
+        source_volume_data = "/" + source_volume_data
+    if not target_volume_data.startswith("/"):
+        target_volume_data = "/" + target_volume_data
 
     if target_file_path == source_file_path:
         if source_volume_data == target_volume_data:
--- a/src/Visualization/Python/InterpolateVolumeData.py
+++ b/src/Visualization/Python/InterpolateVolumeData.py
@@ -8,10 +8,12 @@ from spectre import Spectral, Interpolation
 import numpy as np
 import sys
 from multiprocessing import Pool
-import argparse
+import click
 import re
 import glob
 import logging
+
+logger = logging.getLogger(__name__)
 
 basis_dict = {
     "Legendre": Spectral.Basis.Legendre,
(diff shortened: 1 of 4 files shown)


src/Visualization/Python/InterpolateVolumeData.py
```diff
@@ -187,169 +190,115 @@ def interpolate_h5_file(source_file_path,
...
+@click.command(help=interpolate_h5_file.__doc__)
+@click.option(
+    "--source-file-prefix",
+    required=True,
+    help=("The prefix for the .h5 source files. All files starting with the "
+          "prefix followed by a number will be interpolated."))
+@click.option("--source-subfile-name",
+              required=True,
+              help=("The name of the volume data subfile within the "
+                    "source files in which the data is contained"))
```
nikwit, line 202: maybe add that it should be preprended by a "/" and not have a suffix.

src/Visualization/Python/InterpolateVolumeData.py
```diff
@@ -187,169 +190,115 @@ def interpolate_h5_file(source_file_path,
...
+    help=(
+        "The prefix for the target files where the interpolated data is "
+        "written. When no target file is specified, the interpolated data is "
+        "written to the corresponding source file in a new volume data "
+        "subfile."))
+@click.option("--target-subfile-name",
+              required=True,
+              help=("The name of the volume data subfile within the target "
+                    "files where the data will be written."))
+@click.option("--tensor-component",
```
nikwit, line 215: components*

nilsvu:
@nikwit Do you have some time to review this?

src/Visualization/Python/InterpolateVolumeData.py
```diff
@@ -187,169 +190,115 @@ def interpolate_h5_file(source_file_path,
...
+              multiple=True,
+              help=("The names of the tensors that are to be interpolated. "
+                    "Accepts regular expression. "
+                    "If none are specified, all tensors are interpolated."))
+@click.option(
+    "--target-extents",
+    callback=(lambda ctx, param, value: list(map(int, value.split(',')))
+              if value else []),
+    required=True,
+    help=("The extents of the target grid. "
```
nikwit: It should be given as a comma-separated list without spaces e.g. ...

The error message with the old argparse style `3 4 5` is not very helpful because it thinks you are giving more command line options so we should be more explicit here.

tests/Unit/Visualization/Python/Test_InterpolateVolumeData.py
```diff
@@ -19,18 +22,19 @@ def setUp(self):
         with two tensor components each
         """
 
-        self.path = unit_test_build_path()
-        self.file_name = os.path.join(self.path, "interpolation_1.h5")
+        self.test_dir = os.path.join(unit_test_build_path(), "Visualization",
```
nikwit, line 25: I faintly remember there was some reason we did not use `tempfile` here, was it python2 support or portability? Can we use it now?

nikwit (changes requested):
I tried it on a sample file and it still seemed to work :) you can squash immediately

src/Visualization/Python/InterpolateVolumeData.py
nilsvu: Added a commit normalizing these paths

src/Visualization/Python/InterpolateVolumeData.py
nilsvu: This option is given multiple times, one for each component, so I think the singular is nicer?

tests/Unit/Visualization/Python/Test_InterpolateVolumeData.py
nilsvu: There's also `click.isolated_filesystem()`. I haven't looked into this yet, but could be nice

nikwit (dismissed):
Great. You can add the temporary file system if you like.

nilsvu:
Yea I'll look at the temporary filesystem stuff, but not in this PR

nilsvu:
Rebased on #4543

nilsdeppe (approved):

PR #4422, opened 2022-11-28
Tabulated equation of state

ermost:
This is a draft of the tabulated equation of state class.

This PR extends the EquationOfState class to three thermodyn. dimensions and adds example routines for the Nuclear EOS class.

TODO: Define speed of sound interface and add tests.

This PR depends on #4243  

What's not included: Actual read in of the HDF5 table. Need to couples this to the existing infrastructure.

PR #4473, opened 2022-12-07
Couple 3D equation of state to GrMhd

ermost:
This fills in calls for all `if constexpr ThemoDynamicDim == 3` cases.

This PR relies on #4422.


nilsdeppe:
@ermost this was merged as part of your other PR, right? If so, can we close it?

PR #4475, opened 2022-12-07
Add H5 extraction scripts to Python CLI

nilsvu:
## Proposed changes

Move `ExtractInputSourceYamlFromH5.py` to H5 and add it to the Python CLI, as well as `ExtractDatFromH5.py`. Both scripts now take input and output file/dir names as required arguments because CLIs can be very confusing if you don't explicitly specify where they write the data.

### Upgrade instructions

- Invoke `spectre extract-input -h` to run the `ExtractInputSourceYamlFromH5.py` script.
- Invoke `spectre extract-dat -h` to run the `ExtractDatFromH5.py` script.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/IO/H5/Python/ExtractDatFromH5.py
```diff
@@ -4,28 +4,36 @@
...
+    if list:
         print_str = "\n ".join(all_dat_files)
         print("Dat files within '{}':\n {}".format(filename, print_str))
         return
 
-    out_dir = kwargs["output_directory"]
-
+    if out_dir is None:
+        raise ValueError(
+            "An output directory is required unless listing file content.")
```
knelli2, line 32: Can we keep it so that if you don't specify an output directory, it defaults to `extracted_NameOfH5File`?

src/IO/H5/Python/CMakeLists.txt
```diff
@@ -12,6 +12,8 @@ spectre_python_add_module(
   File.cpp
   TensorData.cpp
   VolumeData.cpp
+  PYTHON_FILES
+  ExtractInputSourceYamlFromH5.py
```
knelli2: Should this first be put in the `PYTHON_EXECUTABLES` and in the next commit move it to `PYTHON_FILES`?

knelli2 (changes requested):

src/IO/H5/Python/ExtractDatFromH5.py
nilsvu: We could but I chose not to! I should have made that a separate commit and mentioned it in the description, sorry. I think output file/dir names should never(?) be defaults in a CLI, because then you run the script and don't know what happened / where to look for any output. It's super quick and easy to put the name of the output dir that you want as the last argument, and then you know exactly where output files are going. Added a note to the PR description. Sounds ok?

src/IO/H5/Python/CMakeLists.txt
nilsvu: Doesn't really matter! Only affects whether or not the script is symlinked to bin/. I'll remove `PYTHON_EXECUTABLES` altogether once all scripts are added to the CLI.

nilsvu:
Rebased on #4477

src/IO/H5/Python/ExtractDatFromH5.py
knelli2: Yeah that's ok. I still think it's alright for there to be a default, but not worth holding up the CLI PRs

knelli2 (dismissed):

nilsdeppe:
needs a rebase

nilsvu:
Rebased, and made the output file optional in ExtractInputSourceYamlFromH5.py, writing to stdout if no output file is specified

nilsdeppe (approved):

PR #4485, opened 2022-12-08
Fix bug in UpdateU

nikwit:
@wthrowe this came up when rebasing my branches. I am using UpdateU but haven't added `StepChoosers::ErrorControl` to my databox so it did not compile. You have added this if block in `AdvanceTime`. Is there a reason it is not in `UpdateU` or just an overlook?


wthrowe (approved):
This change is correct.  I just missed it.

PR #4486, opened 2022-12-09
Add action to initialize DataBox items

kidder:
This allows multiple initialization actions to be converted into a mutators that are called by a single action.  This reduces charm++ overhead and makes testing easier as initialization mutators do not need to use the action testing framework.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Actions/InitializeItems.hpp
+++ b/src/ParallelAlgorithms/Actions/InitializeItems.hpp
@@ -1,0 +1,75 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <optional>
+#include <utility>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+/// \cond
+namespace Parallel {
+template <typename Metavariables>
+class GlobalCache;
+}  // namespace Parallel
+/// \endcond
+
+namespace Initialization {
+namespace Actions {
+/*!
+ * \ingroup ActionsGroup
+ * \brief Mutate DataBox items by calling db::mutate_apply on each Mutator in
+ *  the order they are specified
+ *
+ * There's a specialization for `InitializeItems<tmpl::list<Mutators...>>` that
+ * can also be used if a `tmpl::list` is available.
+ *
+ * \details In addition to the requirements specified by db::mutate_apply, each
+ * Mutator must define the type aliases of this action.
+ */
+template <typename... Mutators>
+struct InitializeItems {
+  /// Tags for constant items added to the GlobalCache.  These items are
+  /// initialized from input file options.
+  using const_global_cache_tags = tmpl::remove_duplicates<tmpl::flatten<
+      tmpl::append<typename Mutators::const_global_cache_tags...>>>;
+  /// Tags for mutable items added to the MutableGlobalCache.  These items are
+  /// initialized from input file options.
+  using mutable_global_cache_tags = tmpl::remove_duplicates<tmpl::flatten<
+      tmpl::append<typename Mutators::mutable_global_cache_tags...>>>;
+  /// Tags for simple DataBox items that are initialized from input file options
+  using simple_tags_from_options = tmpl::remove_duplicates<tmpl::flatten<
+      tmpl::append<typename Mutators::simple_tags_from_options...>>>;
+  /// Tags for simple DataBox items that are default initialized.  They may be
+  /// mutated by the Mutators.
+  using simple_tags = tmpl::remove_duplicates<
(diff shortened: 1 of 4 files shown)


src/ParallelAlgorithms/Actions/InitializeItems.hpp
```diff
@@ -0,0 +1,75 @@
...
+  template <typename DbTagsList, typename... InboxTags, typename Metavariables,
+            typename ArrayIndex, typename ActionList,
+            typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTagsList>& box,
+      const tuples::TaggedTuple<InboxTags...>& /*inboxes*/,
+      const Parallel::GlobalCache<Metavariables>& /*cache*/,
+      const ArrayIndex& /*array_index*/, ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    EXPAND_PACK_LEFT_TO_RIGHT(db::mutate_apply<Mutators>(make_not_null(&box)));
```
nilsdeppe, line 64: I think there were complaints that mutators had to initialize tags or something. If default constructed tags are okay, we could do
```cpp
[&box]() {
if constexpr (can_call_mutate_apply_v<Mutator>) { // new type trait
  db::mutate_apply<Mutators>(make_not_null(&box));
}
}
```
Just to be clear, I'm not sure if we want this or that you should add it in this PR, though it is fairly easy to support with this PR.

nilsdeppe (approved):
Nice :D Thanks for doing this!

src/ParallelAlgorithms/Actions/InitializeItems.hpp
kidder: Yeah there was a case where the initialization action AddSimpleTags had to add a do-nothing mutator.  I'm not going to do anything about it in this PR, but I'll let you see how I plan to deal with it when I start converting initialization actions into mutators.

nilsdeppe: Okay, that sounds good to me!

src/ParallelAlgorithms/Actions/InitializeItems.hpp
```diff
@@ -0,0 +1,75 @@
...
+/*!
+ * \ingroup ActionsGroup
+ * \brief Mutate DataBox items by calling db::mutate_apply on each Mutator in
+ *  the order they are specified
+ *
+ * There's a specialization for `InitializeItems<tmpl::list<Mutators...>>` that
+ * can also be used if a `tmpl::list` is available.
+ *
+ * \details In addition to the requirements specified by db::mutate_apply, each
+ * Mutator must define the type aliases of this action.
```
knelli2, line 32: So currently, actions don't have to define a `*_global_cache_tags` alias, it'll be treated as empty if it doesn't exist. Just to be clear, we aren't having that option for Mutators. We are requiring every Mutator to have all these aliases even if they are empty? Also these aliases don't remove duplicates. Should we do that?

kidder: Yes, my plan is to require them.  Removing duplicates here is probably a good thing to do as well

knelli2: Ok sounds good. Then yeah can you remove duplicates in all these lists and add some tests for that?

knelli2 (changes requested):

src/ParallelAlgorithms/Actions/InitializeItems.hpp
nilsdeppe: I agree with Kyle :)

nilsdeppe (changes requested):

src/ParallelAlgorithms/Actions/InitializeItems.hpp
```diff
@@ -0,0 +1,75 @@
...
+ */
+template <typename... Mutators>
+struct InitializeItems {
```
nilsvu, line 35: Do you intend to remove the existing `MutateApply` action? That's a special case of this one (single mutator). Currently CCE uses it mostly, which should probably move to just calling functions.

kidder: Currently my plan is to convert initialization actions into mutators passed to this.  I expect  the actions `AddSimpleTags` and `AddComputeTags` will end up being removed by this process.  And any use of `MutateApply` in the list of initialization actions will be eliminated.

@nilsdeppe and I have long ago come to the conclusion that CCE needs to be refactored into fewer actions and more functions. This would lead to fewer things stored in the CCE DataBoxes, leading to improved compilation time and memory usage.    But currently this isn't high on my priority list.

kidder:
@knelli2 @nilsdeppe  rebased and squashed addition of remove_duplicates

knelli2 (approved):

nilsdeppe (approved):

PR #4460, opened 2022-12-06
Allow sphere target to hold multiple spheres of different radii

knelli2:
## Proposed changes

All spheres have the same center, l_max, and angular ordering, just different radii. This will be useful when we want to dump data onto multiple CCE worldtube radii.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4489, opened 2022-12-09
Add post interp callback to dump metric data in Bondi-Sachs form

knelli2:
## Proposed changes

This adds a new PostInterpolationCallback that will take metric data on concentric extraction spheres, convert it to the Bondi-Sachs form, and dump it to a separate file for each extraction radius. The format of the output data and the filenames are chosen so that it can be read in with the CCE executable.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments
~Depends on and includes #4460.~

### DIFF
--- a/src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
+++ b/src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
@@ -1,0 +1,294 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cmath>
+#include <cstddef>
+#include <iomanip>
+#include <iterator>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/TagName.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "DataStructures/Variables.hpp"
+#include "DataStructures/VariablesTag.hpp"
+#include "Evolution/Systems/Cce/BoundaryData.hpp"
+#include "Evolution/Systems/Cce/OptionTags.hpp"
+#include "Evolution/Systems/Cce/Tags.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
+#include "IO/Observer/ReductionActions.hpp"
+#include "NumericalAlgorithms/Spectral/SwshCoefficients.hpp"
+#include "NumericalAlgorithms/Spectral/SwshCollocation.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Invoke.hpp"
+#include "ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp"
+#include "ParallelAlgorithms/Interpolation/Protocols/PostInterpolationCallback.hpp"
+#include "ParallelAlgorithms/Interpolation/Targets/AngularOrdering.hpp"
+#include "ParallelAlgorithms/Interpolation/Targets/Sphere.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeString.hpp"
+#include "Utilities/PrettyType.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace intrp {
+namespace callbacks {
+/*!
+ * \brief Post interpolation callback that dumps metric data in Bondi-Sachs form
+ * on a number of extraction radii given by the `intrp::TargetPoints::Sphere`
+ * target.
+ *
+ * To use this callback, the target must be the `intrp::TargetPoints::Sphere`
+ * target in the inertial frame. This callback also expects that the GH source
+ * vars on each of the target spheres are:
+ *
+ * - `gr::Tags::SpacetimeMetric`
+ * - `GeneralizedHarmonic::Tags::Pi`
+ * - `GeneralizedHarmonic::Tags::Phi`
+ *
+ * This callback will write a new `H5` file for each extraction radius in the
+ * Sphere target. The name of this file will be a file prefix specified by the
+ * Cce::Tags::FilePrefix prepended onto `CceRXXXX.h5` where the XXXX is the
+ * zero-padded extraction radius rounded to the nearest integer. The quantities
+ * that will be written are
+ *
+ * - `Cce::Tags::BondiBeta`
+ * - `Cce::Tags::Dr<Cce::Tags::BondiJ>`
+ * - `Cce::Tags::Du<Cce::Tags::BondiR>`
+ * - `Cce::Tags::BondiH`
+ * - `Cce::Tags::BondiJ`
+ * - `Cce::Tags::BondiQ`
+ * - `Cce::Tags::BondiR`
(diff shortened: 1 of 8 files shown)


src/Evolution/Systems/Cce/OptionTags.hpp
```diff
@@ -53,6 +53,15 @@ struct CceEvolutionPrefix {
   using group = Evolution;
 };
 
+struct BondiSachsOutputFilePrefix {
+  using type = std::string;
+  static constexpr Options::String help{
+      "Filename prefix for dumbing Bondi-Sachs data on worltube radii. Files "
```
wthrowe: "dumping"

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -0,0 +1,301 @@
...
+
+  template <typename DbTags, typename Metavariables, typename TemporalId>
+  static void apply(const db::DataBox<DbTags>& box,
+                    Parallel::GlobalCache<Metavariables>& cache,
+                    const TemporalId& temporal_id) {
+    const auto& sphere =
+        Parallel::get<Tags::Sphere<InterpolationTargetTag>>(cache);
+    const auto& filename_prefix = Parallel::get<Cce::Tags::FilePrefix>(cache);
+
+    ASSERT(sphere.angular_ordering == intrp::AngularOrdering::Cce,
```
wthrowe: ERROR, as this is controlled from the input file.

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -0,0 +1,301 @@
...
+ * target.
+ *
+ * To use this callback, the target must be the `intrp::TargetPoints::Sphere`
+ * target in the inertial frame. This callback also expects that the GH source
+ * vars on each of the target spheres are:
+ *
+ * - `gr::Tags::SpacetimeMetric`
+ * - `GeneralizedHarmonic::Tags::Pi`
+ * - `GeneralizedHarmonic::Tags::Phi`
+ * - `::Tags::deriv<GeneralizedHarmonic::Tags::Phi>`
```
wthrowe: I don't see any uses of `deriv<Phi>`.  Why does it have to be there?

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -0,0 +1,301 @@
...
+        db::get<::Tags::Variables<gh_source_vars_from_interpolation>>(box);
+
+    const auto& all_spacetime_metric =
+        get<gr::Tags::SpacetimeMetric<3, ::Frame::Inertial>>(all_gh_vars);
+    const auto& all_pi =
+        get<GeneralizedHarmonic::Tags::Pi<3, ::Frame::Inertial>>(all_gh_vars);
+    const auto& all_phi =
+        get<GeneralizedHarmonic::Tags::Phi<3, ::Frame::Inertial>>(all_gh_vars);
+
+    Variables<gh_source_vars_for_cce> tmp_buffer;
```
wthrowe: You never allocate any memory in this buffer.  It's not saving you anything, and I don't think changing the `data()` value in a `Variables`-backed `Tensor` is a supported operation.

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -0,0 +1,301 @@
...
+          }
+        }
+      }
+
+      offset += num_points_single_sphere;
+
+      // Bondi data
+      Variables<cce_boundary_tags> bondi_boundary_data{
+          num_points_single_sphere};
+      ComplexModalVector goldberg_mode_buffer{square(l_max + 1)};
```
wthrowe: These two *are* allocations inside the loop, so moving them outside would help.

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -0,0 +1,301 @@
...
+
+      Cce::create_bondi_boundary_data(make_not_null(&bondi_boundary_data), phi,
+                                      pi, spacetime_metric, radius, l_max);
+
+      const std::vector<std::string> all_legend = build_legend(l_max, false);
+      const std::vector<std::string> real_legend = build_legend(l_max, true);
```
wthrowe: These two lines may cause the majority of the allocations in this function.  Move them out of the loop.

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -0,0 +1,301 @@
...
+            data_to_write_buffer.reserve(number_of_components(l_max, is_real));
+            data_to_write_buffer.emplace_back(
+                intrp::InterpolationTarget_detail::get_temporal_id_value(
+                    temporal_id));
+
+            // We loop over ell and m rather than just the total number of modes
+            // because we don't print negative m or the imaginary part of m=0
+            // for real quantities.
+            for (size_t ell = 0; ell <= l_max; ell++) {
+              for (int m = is_real ? 0 : -ell; m <= static_cast<int>(ell);
```
wthrowe: Cast the first `ell` too.

wthrowe (changes requested):

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
knelli2: Well I need a `Variables` of g, pi, phi that only has one sphere's worth of points in order to pass to `create_bondi_data` for each extraction radius. Any suggestions on how to do that then without a copy? I already have all the data in `all_gh_vars` for all extraction spheres, Just need to split it up a bit.

wthrowe: You're not passing this to anything, and it has zero points.

Avoiding the copies with non-owning DataVectors is good.  I just don't understand why you want an empty temporary buffer.

knelli2: Oh I see what you mean. I used a `Variables` just to easily get un-allocated tensors for g, pi, and phi so I could set their data refs when I loop over the radii. I guess I could have explicitly wrote out the unallocated tensors, but that was more lines so I didn't.

wthrowe: It's three lines instead of two.

knelli2: Ok I changed it.

wthrowe (dismissed):
Looks good.  Squash.

nilsdeppe (approved):
I don't love that we need a new file for each extraction radius, but I assume that's a current requirement of CCE, right? I'm fine leave as-is for now, but we should probably combine them into the reductions file at some point (or maybe one CCE file with subfiles)

nilsdeppe:
CI failure is timeout in binary compact object test.

knelli2:
@nilsdeppe Yes the reason I did it like this is because currently CCE is set up to read in metric data which is output from SpEC in a specific format (different files for each radius, named a certain way). I agree eventually it'd be nice to have CCE read in data from SpECTRE in our own format, but that's a whole change to CCE which is why I didn't bother doing it here. Just wanted to get waveforms working quickly.

nilsdeppe:
Okay, thanks for clarifying! I completely agree, getting it into one file is a much bigger change and very low priority :) Thanks!

PR #3913, opened 2022-04-05
Make element_logical_coordinates deterministic

kidder:
## Proposed changes

Changes element_logical_coordinates so that it will return the same output irregardless of the ordering of the input vector of ElementIds

For ease of review, I broke this into 4 commits which I will squash
- The first modifies one of the existing tests, but keeps the current behavior, the test case should always pass
- The second commit shuffles the input ElementIds which will randomly break the modified test.  The test case will randomly pass or fail
- The third commit adds a new test that will always fail (unless there is a very unlikely random shuffle of a large container that preserves the relative ordering of a large number of points on element boundaries).
- The fourth commit changes the behavior of element_logical_coordinates (which changes the expected values for one target point of the modified test); the test case should always pass.

See further discussion below...

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

This should eliminate random noise in iterations of the apparent horizon finder arising from a random Element being selected to interpolate to a point on the boundary of more than one Element.  

Previously the algorithm returned the first element in the list that shared the boundary point.  Now it selects an Element only if a boundary point is on the lower side of the element or on the boundary of its containing Block in all dimensions that the target point is on the boundary.  In other words, previously an Element was said to contain the point x if x was in the interval [x_l, x_u] where x_l and x_u are the lower and upper bounds of the BlockLogical coordinate frame.  Now, the Element will not contain the point if x = x_u unless x_u = 1.0 (which is on the boundary of the Block).  

Note that it is not the responsibility of this function to deal with choosing between elements on block boundaries (the input block_coord_holders should have associated such a point with the Block with the lowest BlockId

Note that with the previous behavior, if multiple Elements sharing a target boundary point were on different nodes (or maybe even cores) multiple Elements will provide data for the target point.  With the new behavior, each target point should only be claimed by a single Element.  This may allow simplifying some logic in the horizon finder.

Note the added test assumes all Elements are on the same refinement levels within a Block.  I will add a test for a randomly refined Block in a future commit or PR.

PR #4510, opened 2022-12-13
Fix a bug at roundoff error in block_logical_coords

nilsvu:
This bug appeared when interpolating between two BBH domains with bulged frustums. Points at inner frustum edges were not mapped to the source domain because the frustum inverse reports logical coordinates outside [-1, 1] by roundoff error.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/BlockLogicalCoordinates.cpp
+++ b/src/Domain/BlockLogicalCoordinates.cpp
@@ -11,6 +11,7 @@
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "Domain/Domain.hpp"  // IWYU pragma: keep
 #include "Domain/Structure/BlockId.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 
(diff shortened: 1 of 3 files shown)


kidder (changes requested):
please add a test that would fail/pass before/after this change

markscheel:
Should this bug be fixed in Frustrum instead?  (That is, should Frustrum be changed to never return logical coords outside of [-1,1]?  ).

I worry that this change ignores #3913 and similar places where we assume logical coords are 1.0 and/or -1.0 on the boundary.

Or maybe #3913 and other places need to account for roundoff differently.

nilsvu:
Many map inverses currently return values outside -1,1. Yea, maybe we should require all maps to ensure they return only values within their range. That's a lot of work though and needs to be tested thoroughly. Not all maps inverses return logical coords either, right? I haven't made up my mind on what's the best way to handle this.

kidder:
- So as @markscheel says checking exact vs checking within roundoff should probably be a choice of the user of this function.  
- The testing of maps needs to be done on the maps held by the Blocks, which are usually a composition of multiple maps, not individual maps.

nilsvu:
This is now ready to review for whoever wants to @kidder @markscheel @nilsdeppe @wthrowe

nilsdeppe (approved):
LGTM

wthrowe (approved):

kidder (approved):

PR #4518, opened 2022-12-15
Add ExcisionBoundaryA InterpolationTarget to EvolveGhSingleBlackHole

mar-celine:
## Proposed changes

This is in the series of PRs that is working towards observing the char speeds on the excision surface.
This PR adds the ExcisionBoundaryA InterpolationTarget. The target points are given by a Sphere in
Frame::Grid. As the Shape maps do not change the angular locations of the collocation points on the surface, we
can use this target to interpolate Frame::Inertial quantities onto the excision boundary.

Addresses #4591 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
+++ b/src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
@@ -1,0 +1,218 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.hpp"
+
+#pragma once
+
+#include <cstdint>
+#include <type_traits>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/TaggedContainers.hpp"
+#include "DataStructures/Tags/TempTensor.hpp"
+#include "DataStructures/TempBuffer.hpp"
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "DataStructures/Variables.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Christoffel.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/Christoffel.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/ExtrinsicCurvature.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/Ricci.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Lapse.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Shift.hpp"
+#include "PointwiseFunctions/GeneralRelativity/SpacetimeNormalVector.hpp"
+#include "PointwiseFunctions/GeneralRelativity/SpatialMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Transform.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace ah {
+
+/// Single frame case
+template <typename SrcTagList, typename DestTagList>
+void ComputeExcisionBoundaryVolumeQuantities::apply(
+    const gsl::not_null<Variables<DestTagList>*> target_vars,
+    const Variables<SrcTagList>& src_vars, const Mesh<3>& /*mesh*/) {
+  static_assert(
+      std::is_same_v<tmpl::list_difference<SrcTagList, allowed_src_tags>,
+                     tmpl::list<>>,
+      "Found a src tag that is not allowed");
+  static_assert(
+      std::is_same_v<tmpl::list_difference<required_src_tags, SrcTagList>,
+                     tmpl::list<>>,
+      "A required src tag is missing");
+
+  static_assert(
+      std::is_same_v<tmpl::list_difference<DestTagList,
+                                           allowed_dest_tags<Frame::Inertial>>,
+                     tmpl::list<>>,
+      "Found a dest tag that is not allowed");
+  static_assert(
+      std::is_same_v<tmpl::list_difference<required_dest_tags<Frame::Inertial>,
+                                           DestTagList>,
+                     tmpl::list<>>,
+      "A required dest tag is missing");
+
+  if (target_vars->number_of_grid_points() !=
+      src_vars.number_of_grid_points()) {
+    target_vars->initialize(src_vars.number_of_grid_points());
+  }
+
+  get<gr::Tags::SpacetimeMetric<3, Frame::Inertial>>(*target_vars) =
+    get<gr::Tags::SpacetimeMetric<3, Frame::Inertial>>(src_vars);
+
+  const auto& spacetime_metric =
+      get<gr::Tags::SpacetimeMetric<3, Frame::Inertial>>(src_vars);
+
+  using spatial_metric_tag = gr::Tags::SpatialMetric<3, Frame::Inertial>;
+  using inv_spatial_metric_tag =
+    gr::Tags::InverseSpatialMetric<3, Frame::Inertial>;
+  using lapse_tag = gr::Tags::Lapse<DataVector>;
+  using shift_tag = gr::Tags::Shift<3, Frame::Inertial>;
+
+  // All of the temporary tags, including some that may be repeated
+  // in the target_variables (for now).
+  using full_temp_tags_list =
+      tmpl::list<spatial_metric_tag, inv_spatial_metric_tag,
+                 lapse_tag, shift_tag>;
+
+  // temp tags without variables that are already in DestTagList.
+  using temp_tags_list =
+      tmpl::list_difference<full_temp_tags_list, DestTagList>;
+  TempBuffer<temp_tags_list> buffer(get<0, 0>(spacetime_metric).size());
+
+  // These may or may not be temporaries
+  auto& lapse = *(get<lapse_tag>(
+      target_vars, make_not_null(&buffer)));
+  auto& shift = *(get<shift_tag>(
+      target_vars, make_not_null(&buffer)));
+  auto& spatial_metric = *(get<spatial_metric_tag>(
+      target_vars, make_not_null(&buffer)));
+  auto& inv_spatial_metric = *(get<inv_spatial_metric_tag>(
+      target_vars, make_not_null(&buffer)));
+
+  // Actual computation starts here
+
+  gr::spatial_metric(make_not_null(&spatial_metric), spacetime_metric);
+  // put determinant of 3-metric temporarily into lapse to save memory.
+  determinant_and_inverse(make_not_null(&lapse),
+                          make_not_null(&inv_spatial_metric),
+                          spatial_metric);
+  gr::shift(make_not_null(&shift), spacetime_metric, inv_spatial_metric);
+  gr::lapse(make_not_null(&lapse), shift, spacetime_metric);
+}
+
+/// Dual frame case
+template <typename SrcTagList, typename DestTagList, typename TargetFrame>
+void ComputeExcisionBoundaryVolumeQuantities::apply(
+    const gsl::not_null<Variables<DestTagList>*> target_vars,
+    const Variables<SrcTagList>& src_vars, const Mesh<3>& /*mesh*/,
+    const Jacobian<DataVector, 3, TargetFrame, Frame::Inertial>& jacobian,
+    const InverseJacobian<DataVector, 3, Frame::ElementLogical, TargetFrame>&
+        /*inverse_jacobian*/) {
+
+  static_assert(
+      std::is_same_v<tmpl::list_difference<SrcTagList, allowed_src_tags>,
+                     tmpl::list<>>,
+      "Found a src tag that is not allowed");
+  static_assert(
+      std::is_same_v<tmpl::list_difference<required_src_tags, SrcTagList>,
+                     tmpl::list<>>,
+      "A required src tag is missing");
+
+  static_assert(
+      std::is_same_v<
+          tmpl::list_difference<DestTagList, allowed_dest_tags<TargetFrame>>,
+          tmpl::list<>>,
+      "Found a dest tag that is not allowed");
+  static_assert(
+      std::is_same_v<
+          tmpl::list_difference<required_dest_tags<TargetFrame>, DestTagList>,
+          tmpl::list<>>,
+      "A required dest tag is missing");
+
+  const auto& spacetime_metric =
+      get<gr::Tags::SpacetimeMetric<3, Frame::Inertial>>(src_vars);
+
+  if (target_vars->number_of_grid_points() !=
+      src_vars.number_of_grid_points()) {
+    target_vars->initialize(src_vars.number_of_grid_points());
+  }
+
+  using spatial_metric_tag = gr::Tags::SpatialMetric<3, TargetFrame>;
+  using inv_spatial_metric_tag = gr::Tags::InverseSpatialMetric<3, TargetFrame>;
+  using lapse_tag = gr::Tags::Lapse<DataVector>;
+  using inertial_shift_tag = gr::Tags::Shift<3, Frame::Inertial>;
+
+  // Additional temporary tags used for multiple frames
+  using inertial_spatial_metric_tag =
+      gr::Tags::SpatialMetric<3, Frame::Inertial>;
+  using inertial_inv_spatial_metric_tag =
+      gr::Tags::InverseSpatialMetric<3, Frame::Inertial>;
+
+  // All of the temporary tags, including some that may be repeated
+  // in the target_variables (for now).
+  using full_temp_tags_list =
+      tmpl::list<spatial_metric_tag, inv_spatial_metric_tag,
+                 lapse_tag, inertial_shift_tag,
+                 inertial_spatial_metric_tag,
+                 inertial_inv_spatial_metric_tag>;
+
+  // temp tags without variables that are already in DestTagList.
+  using temp_tags_list =
+      tmpl::list_difference<full_temp_tags_list, DestTagList>;
+  TempBuffer<temp_tags_list> buffer(get<0, 0>(spacetime_metric).size());
+
+  // These may or may not be temporaries, depending on if they are asked for
+  // in target_vars.
+  auto& lapse = *(get<lapse_tag>(
+      target_vars, make_not_null(&buffer)));
+  auto& inertial_shift = *(get<inertial_shift_tag>(
+      target_vars, make_not_null(&buffer)));
+  auto& inertial_spatial_metric = *(get<inertial_spatial_metric_tag>(
+      target_vars, make_not_null(&buffer)));
+  auto& inertial_inv_spatial_metric =
+      *(get<inertial_inv_spatial_metric_tag>(
+          target_vars, make_not_null(&buffer)));
+  auto& spatial_metric = *(get<spatial_metric_tag>(
+      target_vars, make_not_null(&buffer)));
+  auto& inv_spatial_metric = *(get<inv_spatial_metric_tag>(
+      target_vars, make_not_null(&buffer)));
+
+  // Actual computation starts here
+
+  gr::spatial_metric(make_not_null(&inertial_spatial_metric), spacetime_metric);
+  // put determinant of 3-metric temporarily into lapse to save memory.
+  determinant_and_inverse(make_not_null(&lapse),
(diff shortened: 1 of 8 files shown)


src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -62,19 +62,23 @@ struct ObserveSurfaceData
...
     // Spherepack-evenly-distributed over the inertial-frame sphere
     // (they are Spherepack-evenly-distributed over the HorizonFrame
     // sphere).
-    std::vector<TensorComponent> tensor_components{
-        {"InertialCoordinates_x"s, get<0>(inertial_strahlkorper_coords)},
-        {"InertialCoordinates_y"s, get<1>(inertial_strahlkorper_coords)},
-        {"InertialCoordinates_z"s, get<2>(inertial_strahlkorper_coords)}};
-
+    std::vector<TensorComponent> tensor_components;
+    if constexpr (WriteCoordinates) {
```
nilsdeppe:
```cpp
if constexpr (db::tag_is_retrievable_v<StrahlkorperTags::CartesianCoords<::Frame::Inertial>, db::DataBox<DbTags>>) {
  ...
}
```

nilsdeppe:
You can squash. Is it possible to observe the data and verify things are written as expected?

tests/InputFiles/GeneralizedHarmonic/KerrSchild.yaml
```diff
@@ -162,6 +165,11 @@ EventsAndTriggers:
           Interval: 5
           Offset: 2
     - - AhA
+  - - Slabs:
+        EvenlySpaced:
+          Interval: 5
+          Offset: 2
```
nilsvu, line 171: set offset to 0?

tests/InputFiles/GeneralizedHarmonic/KerrSchild.yaml
```diff
@@ -193,3 +201,10 @@ ApparentHorizons:
       DivergenceIter: 5
       MaxIts: 100
     Verbosity: Verbose
+
+InterpolationTargets:
+  ExcisionBoundaryA:
+    Lmax: 10
+    Center: [0., 0., 0.]
+    Radius: 2.0
```
nilsvu, line 209: set to excision boundary. Also use yaml references.

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -0,0 +1,221 @@
...
+                     tmpl::list<>>,
+      "A required dest tag is missing");
+
+  if (target_vars->number_of_grid_points() !=
+      src_vars.number_of_grid_points()) {
+    target_vars->initialize(src_vars.number_of_grid_points());
+  }
+
+  get<gr::Tags::SpacetimeMetric<3, Frame::Inertial>>(*target_vars) =
+    get<gr::Tags::SpacetimeMetric<3, Frame::Inertial>>(src_vars);
```
nilsvu, line 69: This whole class is extremely complicated when all it does is forward the spacetime metric. Are we going to put more logic here? Can you reuse the Apparent horizon class instead? Why are the two apply functions so different?

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
```diff
@@ -86,21 +91,57 @@ struct EvolutionMetavars
...
+        gr::Tags::ShiftCompute<volume_dim, Frame::Inertial, DataVector>,
+        gr::Tags::LapseCompute<volume_dim, Frame::Inertial, DataVector>>>;
+    using compute_target_points =
+        intrp::TargetPoints::Sphere<ExcisionBoundaryA, ::Frame::Grid>;
+    using post_interpolation_callback =
+        intrp::callbacks::ObserveSurfaceData<tags_to_observe, ExcisionBoundaryA,
+                                             ::Frame::Grid>;
+    // run_callbacks
+    template <typename metavariables>
+    using interpolating_component = typename metavariables::gh_dg_element_array;
```
nilsvu, line 117: Is this used at all? Can you delete it?

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
```diff
@@ -86,21 +91,57 @@ struct EvolutionMetavars
...
       : tt::ConformsTo<Options::protocols::FactoryCreation> {
     using factory_classes = Options::add_factory_classes<
         typename gh_base::factory_creation::factory_classes,
-        tmpl::pair<Event, tmpl::list<intrp::Events::Interpolate<
-                              3, AhA, interpolator_source_vars>>>>;
+        tmpl::pair<
+            Event,
+            tmpl::list<
+                intrp::Events::Interpolate<3, AhA, interpolator_source_vars>,
+                intrp::Events::InterpolateWithoutInterpComponent<
```
nilsvu, line 133: Why is this "without interp component" and the AH is with interp component?

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
```diff
@@ -110,11 +151,21 @@ struct EvolutionMetavars
...
 
+  // Add ElementInitInterpPoints to initialization_actions.
+  using helper_init =
+      tmpl::push_back<tmpl::pop_back<initialization_actions>,
+                      intrp::Actions::ElementInitInterpPoints<
+                          intrp::Tags::InterpPointInfo<EvolutionMetavars>>,
+                      tmpl::back<initialization_actions>>;
+
+  // the dg element array needs to be re-declared to capture the new type
+  // aliases for the action lists.
```
nilsvu: Is this comment correct? It wasn't there before and isn't related to the change if I understand correctly. Maybe a remnant of a rebase?

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
```diff
@@ -110,11 +151,21 @@ struct EvolutionMetavars
 
   using typename gh_base::step_actions;
 
+  // Add ElementInitInterpPoints to initialization_actions.
+  using helper_init =
```
nilsvu: call this `initialization_actions` and remove line 150. Instead, use `gh_base::initialization_actions` in this implementation.

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
```diff
@@ -110,11 +151,21 @@ struct EvolutionMetavars
...
+  // the dg element array needs to be re-declared to capture the new type
+  // aliases for the action lists.
   using gh_dg_element_array = DgElementArray<
       EvolutionMetavars,
       tmpl::flatten<tmpl::list<
-          Parallel::PhaseActions<Parallel::Phase::Initialization,
-                                 initialization_actions>,
+          Parallel::PhaseActions<
+              Parallel::Phase::Initialization,
+              tmpl::list<initialization_actions, helper_init>>,
```
nilsvu: this seems to add init actions twice. with the change I commented above you can just drop this.

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
```diff
@@ -154,14 +205,18 @@ struct EvolutionMetavars
...
-      observers::ObserverWriter<EvolutionMetavars>,
-      std::conditional_t<UseNumericalInitialData,
-                         importers::ElementDataReader<EvolutionMetavars>,
-                         tmpl::list<>>,
-      gh_dg_element_array, intrp::Interpolator<EvolutionMetavars>,
-      intrp::InterpolationTarget<EvolutionMetavars, AhA>>>;
+  using component_list =
+
+      tmpl::flatten<tmpl::list<
+
```
nilsvu: fix the formatting here please. all you did was add the excision boundary component right

nilsvu (changes requested):

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
mar-celine: Hi Nils, I don't exactly remember the difference, but because the apparent horizon find requires doing multiple iterations that requires computing quantities on a surface multiple times in a single horizon find, while for the excision boundary we just want to compute quantities once, we don't need the functionality that the additional interp component gives us. Maybe @markscheel has something to add?

knelli2: Yeah the reason is because of what @gsb76 said. The apparent horizon finds send data to the interpolator which holds it until all iterations of the horizon find are complete. But for the excision surface, we just interpolate to the surface once and don't need to keep volume data around, so we do the interpolation on the elements rather than send it to the interpolator. So the `InterpWithoutInterpComponent` does the interpolation on the elements and immediately sends to the targets.

Re: your above comment about the `interpolating_component` alias: It is necessary because that specifies what component we do the interpolation on in `InterpWithoutInterpComponent`

src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicSingleBlackHole.hpp
knelli2: See comment below

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
nilsdeppe: 1. Yes. We need to compute all ingredients for char speeds on a moving mesh.
2. No, because AH uses different variables.
3. One for static meshes and one for moving meshes.

nilsvu: there's some commented-out code, so some cleanup is needed here I think

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -0,0 +1,184 @@
...
+  using lapse_tag = ::Tags::TempScalar<0, DataVector>;
+  using shift_tag = ::Tags::TempI<1, 3, Frame::Inertial, DataVector>;
+  using spacetime_normal_vector_tag =
+      ::Tags::TempA<2, 3, Frame::Inertial, DataVector>;
+
+  // Additional temporary tags used for multiple frames
+  using inertial_metric_tag = gr::Tags::SpatialMetric<3, Frame::Inertial>;
+  using inertial_inv_metric_tag =
+      gr::Tags::InverseSpatialMetric<3, Frame::Inertial>;
+  //using inertial_ex_curvature_tag =
```
nilsdeppe: Remove commented code here

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -0,0 +1,184 @@
...
+  //    gr::Tags::ExtrinsicCurvature<3, Frame::Inertial>;
+  using logical_deriv_metric_tag = ::Tags::TempTensor<
```
nilsdeppe: We don't need derivatives on the excision boundary right now (we might at some point if we want to do high-order crossing predictions). Remove all derivative-related code.

nilsdeppe (commented):
Couple small comments for removing dead code. You can squash that immediately

nilsdeppe (commented):
LGTM. @nilsvu can you please look at the fixups?

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -0,0 +1,173 @@
...
+
+  gr::spatial_metric(make_not_null(&inertial_metric), psi);
+  // put determinant of 3-metric temporarily into lapse to save memory.
+  determinant_and_inverse(make_not_null(&lapse),
+                          make_not_null(&inertial_inv_metric), inertial_metric);
+
+  gr::shift(make_not_null(&shift), psi, inertial_inv_metric);
+  gr::lapse(make_not_null(&lapse), shift, psi);
+  gr::spacetime_normal_vector(make_not_null(&spacetime_normal_vector), lapse,
+                              shift);
```
nilsvu: why does the moving mesh case compute so many more variables than the static mesh? I understand this is just copied from the AH, but maybe add some comments? Edit: I think this code doesn't actually run in any test because the KerrSchild.yaml has a static mesh. So maybe replace this implementation with an `ERROR("not yet implemented")` for now until it runs in a test? up to you

nilsvu (dismissed):
LGTM please squash

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
mar-celine: I can remove the spacetime normal vector computation, this was needed for the computation of the extrinsic curvature but because it's not computed anymore I can remove it.

nilsdeppe (commented):
Ah yea, can you add a test for this? It looks like there's a `Test_ComputeHorizonVolumeQuantities.cpp` that you can use as a model

tests/Unit/ApparentHorizons/Test_ComputeExcisionBoundaryVolumeQuantities.cpp
```diff
@@ -0,0 +1,434 @@
...
+    auto& lower_metric_inertial =
+        get<::gr::Tags::SpatialMetric<3, Frame::Inertial>>(
+            inertial_metric_vars);
+    transform::to_different_frame(make_not_null(&lower_metric_inertial),
+                                  spatial_metric,
+                                  inv_jacobian_grid_to_inertial);
+
+    // Now differentiate the inertial-frame lapse, shift, spatial metric
+    // to get inertial derivatives.
+    const auto deriv_inertial_metric_vars = partial_derivatives<
```
nilsdeppe: Do you need the derivs? I didn't think we needed them in this case

tests/Unit/ApparentHorizons/Test_ComputeExcisionBoundaryVolumeQuantities.cpp
```diff
@@ -0,0 +1,434 @@
...
+    // Now fill src_vars
+    get<::gr::Tags::SpacetimeMetric<3, Frame::Inertial>>(src_vars) =
+        gr::spacetime_metric(lapse, shift_inertial, lower_metric_inertial);
+    get<::GeneralizedHarmonic::Tags::Phi<3, Frame::Inertial>>(src_vars) =
+        GeneralizedHarmonic::phi(lapse, inertial_d_lapse, shift_inertial,
+                                 inertial_d_shift, lower_metric_inertial,
+                                 inertial_d_spatial_metric);
+  }
+
+  if constexpr (tmpl::list_contains_v<
```
nilsdeppe: I don't think this should ever be `true`, right?

tests/Unit/ApparentHorizons/Test_ComputeExcisionBoundaryVolumeQuantities.cpp
```diff
@@ -0,0 +1,434 @@
...
+          determinant_and_inverse(
+              get<::gr::Tags::SpatialMetric<3, Frame::Inertial>>(
+                  inertial_metric_vars))
+              .second;
+      const auto& inertial_inverse_spatial_metric =
+          get<gr::Tags::InverseSpatialMetric<3, Frame::Inertial>>(dest_vars);
+      CHECK_ITERABLE_APPROX(expected_inertial_inverse_spatial_metric,
+                            inertial_inverse_spatial_metric);
+    }
+  }
```
nilsdeppe, line 317: Can you also test for lapse & shift please?

tests/Unit/ApparentHorizons/Test_ComputeExcisionBoundaryVolumeQuantities.cpp
```diff
@@ -0,0 +1,434 @@
...
+    }
+  }());
+
+  if constexpr (std::is_same_v<TargetFrame, Frame::Inertial>) {
+    // Src vars are always in inertial frame. Here TargetFrame is inertial,
+    // so no frame transformation is needed.
+
+    get<::gr::Tags::SpacetimeMetric<3, TargetFrame>>(src_vars) =
+        gr::spacetime_metric(lapse, shift, spatial_metric);
+    get<::GeneralizedHarmonic::Tags::Phi<3, TargetFrame>>(src_vars) =
```
nilsdeppe: do you need `Phi` and `Pi` here? I didn't think so

tests/Unit/ApparentHorizons/Test_ComputeExcisionBoundaryVolumeQuantities.cpp
```diff
@@ -0,0 +1,434 @@
...
+
+  // Set up analytic solution.
+  gr::Solutions::KerrSchild solution(1.0, {{0.1, 0.2, 0.3}},
+                                     {{0.03, 0.01, 0.02}});
+  const auto solution_vars_target_frame = solution.variables(
+      target_frame_coords, 0.0,
+      typename gr::Solutions::KerrSchild::tags<DataVector, TargetFrame>{});
+  const auto& lapse =
+      get<gr::Tags::Lapse<DataVector>>(solution_vars_target_frame);
+  const auto& dt_lapse =
```
nilsdeppe: remove derivs?

nilsdeppe (changes requested):

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -33,6 +33,8 @@
 #include "Utilities/Gsl.hpp"
 #include "Utilities/TMPL.hpp"
 
+#include <iostream>
```
nilsdeppe: remove `iostream` include when you squash :)

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -0,0 +1,206 @@
...
+                 inertial_inv_metric_tag>;
+
+  // temp tags without variables that are already in DestTagList.
+  using temp_tags_list =
+      tmpl::list_difference<full_temp_tags_list, DestTagList>;
+  TempBuffer<temp_tags_list> buffer(get<0, 0>(psi).size());
+
+  // These may or may not be temporaries, depending on if they are asked for
+  // in target_vars.
+  auto& lapse = *(get<lapse_tag>(
```
nilsdeppe, line 175: what do you think about naming these all `target_frame_lapse` etc? Totally optional!

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -0,0 +1,206 @@
...
+      target_vars, make_not_null(&buffer)));
+  auto& inertial_shift = *(get<inertial_shift_tag>(
+      target_vars, make_not_null(&buffer)));
+  auto& inertial_metric = *(get<inertial_metric_tag>(
```
nilsdeppe: Probably not your fault because you took what was in the other file, but can you name the `metric` and `inv_metric` `spatial_metric` and `inv_spatial_metric`? Similarly, `psi` -> `spacetime_metric`. Just do a find-replace in this file in the order `inv_metric` -> `inv_spatial_metric`, `inertial_metric` -> `inertial_spatial_metric`, `psi` -> `spacetime_metric`. Doing just `metric` will need to be manual because find-replace will probably hit sub-words

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -0,0 +1,206 @@
...
+  // Transform spatial metric
+  transform::to_different_frame(make_not_null(&metric), inertial_metric,
+                                jacobian);
+
+  // Invert transformed 3-metric.
+  // put determinant of 3-metric temporarily into lapse to save memory.
+  determinant_and_inverse(make_not_null(&lapse), make_not_null(&inv_metric),
+                          metric);
+
+  gr::shift(make_not_null(&inertial_shift), psi, inertial_inv_metric);
```
nilsdeppe: don't we want to transform the shift into the `TargetFrame` or are we delaying that change? I'm fine leaving it for a followup PR, but then please just add a comment here.

src/ApparentHorizons/ComputeExcisionBoundaryVolumeQuantities.tpp
```diff
@@ -0,0 +1,206 @@
...
+  gr::lapse(make_not_null(&lapse), inertial_shift, psi);
```
nilsdeppe: maybe add a comment `// lapse is inertial_lapse`? Or just name it `inertial_lapse`?

nilsdeppe (commented):
A few minor naming changes is all that's left, I think :) Please rebase on the latest develop, squash, rename, and then squash in the rename too :) No fixup needed for the renames!

nilsdeppe (approved):

PR #4520, opened 2022-12-15
Refine GenerateXdmf command-line interface

nilsvu:
## Proposed changes

Make GenerateXdmf more consistent with other CLIs by taking a file glob as input instead of a `--file-prefix`. This is useful to run over surface data files, which have no node suffix (before, we had to manually rename the surface file adding a '0' suffix to run GenerateXdmf over it). Also make the script a noop when the input is empty, plus a few more quality of life changes.

### Upgrade instructions

When invoking `spectre generate-xdmf` pass a file glob instead of `--file-prefix`, like this:

```
spectre generate-xdmf path/to/VolumeData*.h5 -d SubfileName -o OutputName
```

You can also pass a single file now, which is useful for surface data like AH finder output.

Omit the subfile name to print all available subfiles.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Visualization/Python/GenerateXdmf.py
```diff
@@ -8,9 +8,22 @@
 import h5py
 import logging
 import numpy as np
+import rich
 import sys
 
 
+def available_subfiles(h5file):
```
knelli2: This is a duplicate (except for the `".vol"` part) of the `available_subfiles` function in `PlotDatFile.py`. Maybe we should combine these two functions?

knelli2 (changes requested):

src/Visualization/Python/GenerateXdmf.py
nilsvu: Done in a separate commit 👍

knelli2 (dismissed):

src/Visualization/Python/GenerateXdmf.py
```diff
@@ -312,11 +312,12 @@ def data_item(data_type):
 
 
 @click.command(help=generate_xdmf.__doc__)
-@click.option(
-    '--file-prefix',
-    required=True,
-    help=("The common prefix of the H5 volume files to load, excluding "
-          "the node number integer(s)"))
+@click.argument('h5files',
```
nilsdeppe, line 334: What does the error message for this look like and how do I actually use this option? There's no `--h5files` type syntax, right? I don't love positional arguments, I think we should stick to flags.

src/Visualization/Python/GenerateXdmf.py
```diff
@@ -322,7 +328,7 @@ def data_item(data_type):
                 nargs=-1)
 @click.option('--output',
               '-o',
-              required=True,
+              type=click.Path(writable=True),
               help="Output file name, an xmf extension will be added")
```
nilsdeppe: Can you document that it will be written to stdout if this isn't specified? That seems like a nice thing to know.

Also, when do we ever want this written to stdout? To me this seems like an anti-feature because I'll forget the `-o` flag and then get my terminal flooded with XMDF nonsense...

nilsdeppe (changes requested):

src/Visualization/Python/GenerateXdmf.py
knelli2: It looks like if no h5 files are specified, nothing will happen (line 24 above). If a filename is specified but that file doesn't exist, it looks like you'll get an `h5py` error about not being able to open a file.

nilsvu: The [click docs](https://click.palletsprojects.com/en/8.1.x/parameters/) agree that most things should be options (flags), except for input files / URLs. Arbitrary number of input files or globs only work with (positional) arguments. The documentation for them is part of the main help string. Here's the help message (edited a bit in a fixup to make the argument name clearer):

```
Usage: spectre generate-xdmf [OPTIONS] [H5FILES]...

  Generate an XDMF file for ParaView and VisIt

  Read volume data from the 'H5FILES' and generate an XDMF file. The XDMF file
  points into the 'H5FILES' files so ParaView and VisIt can load the volume
  data. To process multiple files suffixed with the node number, specify a
  glob like 'VolumeData*.h5'.

  To load the XDMF file in ParaView you must choose the 'Xdmf Reader', NOT
  'Xdmf3 Reader'.

Options:
  -o, --output PATH        Output file name, an xmf extension will be added
  -d, --subfile-name TEXT  Name of the volume data subfile in the H5 files. A
                           '.vol' extension is added if needed. If
                           unspecified, list all '.vol' subfiles and exit.
  --stride INTEGER         View only every stride'th time step
  --start-time FLOAT       The earliest time at which to start visualizing.
                           The start-time value is included.
  --stop-time FLOAT        The time at which to stop visualizing. The stop-
                           time value is not included.
  --coordinates TEXT       The coordinates to use for visualization  [default:
                           InertialCoordinates]
  -h, --help               Show this message and exit.
```

nilsvu: @knelli2 if you specified a file that doesn't exist then you'll get a usage error before the code even gets to `h5py`, because of the `click.Path` checks in the argument`. Here's the error:

```
$ ./bin/spectre generate-xdmf file
Usage: spectre generate-xdmf [OPTIONS] [H5FILES]...
Try 'spectre generate-xdmf -h' for help.

Error: Invalid value for '[H5FILES]...': File 'file' does not exist.
```

src/Visualization/Python/GenerateXdmf.py
nilsvu: Amended the help string.

My thought was to "be a good CLI citizen" and make this work like `sed` and other CLIs that write to stdout if no output file is specified. People like to pipe CLI output through `grep` or `sed` or things like that, and want to choose when to write it to a file. I have no particularly strong opinion here though 🤷‍♂️

nilsdeppe (dismissed):
Okay, I'm happy :)

Re to stdout: let's try it :) We can also change if we get user feedback!

nilsvu:
@knelli2 could you check the fixup too please?

knelli2 (dismissed):
Looks good

nilsvu:
Rebased and squashed. Thanks for your reviews 👍

knelli2 (approved):

nilsvu:
Merging because all that happened was a squash and I don't want to bother people on holiday :)

Issue #4408, opened 2022-11-22
Clean up domain creator tests


nilsvu:
Lots of domain creator tests just re-implement the code they're supposed to test. Some suggestions from @nilsdeppe and I for cleaning things up:

- [ ] Regression tests: check inertial coords of block corners. This is easy to reason about and the test doesn't have to reconstruct the coord maps.
- [ ] Remove `test_domain_construction` from domain creator tests (but keep in Domain tests). It reimplements source code.
- [ ] Boundary conditions are consistent: neighboring blocks have the same b.c. in the same direction (accounting for orientation).
- [ ] Go over all external directions, check that boundary conditions are correct. E.g. in upper-z direction of a cylinder the b.c. is always the "upper z" b.c. This is easier to reason about than reconstructing the full block-vector of b.c., and we already check generically that all external directions have b.c. and vice versa.

PR #4526, opened 2022-12-16
CylindricalBinaryCompactObject: add ExcisionSpheres.

markscheel:
Adds ExcisionSpheres to CylindricalBinaryCompactObject.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/CylindricalBinaryCompactObject.cpp
+++ b/src/Domain/Creators/CylindricalBinaryCompactObject.cpp
@@ -30,6 +30,7 @@
 #include "Domain/FunctionsOfTime/QuaternionFunctionOfTime.hpp"
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/DirectionMap.hpp"
+#include "Domain/Structure/ExcisionSphere.hpp"
 #include "Domain/Structure/OrientationMap.hpp"
 #include "NumericalAlgorithms/RootFinding/QuadraticEquation.hpp"
 
@@ -39,6 +40,14 @@ std::array<double, 3> rotate_to_z_axis(const std::array<double, 3> input) {
       OrientationMap<3>{std::array<Direction<3>, 3>{Direction<3>::lower_zeta(),
                                                     Direction<3>::upper_eta(),
                                                     Direction<3>::upper_xi()}},
+      input);
+}
+std::array<double, 3> rotate_from_z_to_x_axis(
+    const std::array<double, 3> input) {
+  return discrete_rotation(
+      OrientationMap<3>{std::array<Direction<3>, 3>{Direction<3>::upper_zeta(),
+                                                    Direction<3>::upper_eta(),
+                                                    Direction<3>::lower_xi()}},
       input);
 }
 std::array<double, 3> flip_about_xy_plane(const std::array<double, 3> input) {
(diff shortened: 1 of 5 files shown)


src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -801,7 +810,62 @@ Domain<3> CylindricalBinaryCompactObject::create_domain() const {
...
 
-  Domain<3> domain{std::move(coordinate_maps)};
+  // Excision spheres
+  std::unordered_map<std::string, ExcisionSphere<3>> excision_spheres{};
+
+  std::unordered_map<size_t, Direction<3>> abutting_directions_A;
+  size_t first_inner_sphere_block = 46;
+  if (include_inner_sphere_A_) {
+    for (size_t i = 0; i < 10; ++i) {
+      abutting_directions_A.emplace(first_inner_sphere_block + i,
```
geoffrey4444, line 821: Possible to satisfy all the codecov complains here?

geoffrey4444 (changes requested):
Happy to approve, but I see codecov failures and wanted to make sure we don't care about those first. No other code changes to suggest

markscheel:
Rebased and added directives to silence codecov in a separate commit.

knelli2 (dismissed):

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -283,9 +284,63 @@ void test_connectivity_once(const bool with_sphere_e,
...
+      }
+      for (size_t i = 27; i < 32; ++i) {
+        abutting_directions_A.emplace(i, Direction<3>::lower_zeta());
+      }
+    }
+    std::unordered_map<size_t, Direction<3>> abutting_directions_B;
+    if(include_inner_sphere_B) {
+      const size_t first_inner_sphere_block = include_inner_sphere_A ? 60 : 46;
+      for (size_t i = first_inner_sphere_block;
+           i < first_inner_sphere_block + 10; ++i) {
```
nilsvu: This test just reimplements the code that we're testing, doesn't it? That's not super useful and hard to maintain. Can we add some consistency checks to `test_domain_creator` instead, e.g. that the block has an external boundary in that direction etc?

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -801,7 +810,70 @@ Domain<3> CylindricalBinaryCompactObject::create_domain() const {
...
+      // LCOV_EXCL_START
+      abutting_directions_B.emplace(first_inner_sphere_block + i,
+                                    Direction<3>::lower_zeta());
+      // LCOV_EXCL_STOP
+    }
+    for (size_t i = 0; i < 4; ++i) {
+      // LCOV_EXCL_START
+      abutting_directions_B.emplace(first_inner_sphere_block + 10 + i,
+                                    Direction<3>::lower_xi());
+      // LCOV_EXCL_STOP
```
nilsvu, line 860: Unless this is an lcov bug (is it?), please don't add lcov excl statements to code that could reasonably be called in a test. The codecov CI test is not required to merge PRs so we can decide to ignore it.

nilsvu (changes requested):

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
markscheel: I think this code *is* called in the test (unless I messed something up somewhere).  So I would consider this an LCOV bug.

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
markscheel: I realize that this test (like many other tests) are re-implementations of the same code in a different way, but I'm not sure what you want me to do. This is basically the same excision_spheres test as in Test_BinaryCompactObject (but the case here is a bit more complicated).  `test_domain_creator` is general, and the property of which blocks have which excision boundaries in which directions are domain-dependent, e.g. they are different for BinaryCompactObject and CylindricalBinaryCompactObject, and they are different depending on the options.

nilsvu: My suggestion is to replace this reimplementation with a consistency check in `test_domain_creator` that all `abutting_directions()` in all `DomainCreator::excision_spheres()` are also `Block::external_boundaries()`. That should be enough. Maybe there's a way to make sure that the excision sphere is fully covered with blocks, but I can't think of a nice way to do that right now.

nilsvu: And you're right about the "like many other tests": reimplementations are done all over the place in the domain creator tests, which make them very hard to maintain. I opened an issue for that here: #4408

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
nilsvu: Yea this is weird. It worries me a bit because I haven't seen such an obvious lcov bug before. Well I have no change request here 🤷‍♂️

geoffrey4444 (dismissed):

markscheel:
I added a new test to test_domain_creator.  This revealed a bug in Shell, which I fixed.  Removed the old test.

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -287,60 +286,6 @@ void test_connectivity_once(const bool with_sphere_e,
...
-    expected_excision_spheres.emplace(
-        "ObjectBExcisionSphere",
-        ExcisionSphere<3>{inner_radius_objectB, center_objectB,
-                          abutting_directions_B});
-    expected_excision_spheres.emplace(
-        "ObjectAExcisionSphere",
-        ExcisionSphere<3>{inner_radius_objectA, center_objectA,
-                          abutting_directions_A});
-
-    CHECK(domain.excision_spheres() == expected_excision_spheres);
```
nilsvu: I'd say keep the check that the number of excision spheres is 2, and also their name, radius, and center.

nilsvu (commented):
LGTM please squash (including this)

markscheel:
Made change and squashed.  Left the Shell commit and the new-test commit separate since they are logically distinct.

nilsvu (dismissed):
Thanks for adding the consistency test 👍

markscheel:
I rebased onto develop because there was a conflict.

nilsvu (approved):

knelli2 (approved):

PR #4529, opened 2022-12-17
Rename objects in BBH domains

knelli2:
## Proposed changes

This PR does 4 things that are all related to each other:

1. Swaps the meaning of `ObjectA` and `ObjectB` in the BinaryCompactObject domain creator to match what is in the CylindricalBinaryCompactObject domain creator and what SpEC uses. Namely that `ObjectA` is on the postiive x-axis and `ObjectB` is on the negative x-axis.
2. Adds some additional restrictions on `ObjectA` and `ObjectB` in the BinaryCompactObject domain creator. Namely that if both objects are excised, then `ObjectA` must now be the larger of the two objects (larger inner radii and x-coord smaller in magnitude).
3. Adds some protections in the control system to error during option parsing if excision spheres aren't added to the domain.
4. Renames `ObjectA` to `PrimaryRightObjectA` and `ObjectB` to `SecondaryLeftObjectB` throughout the code to make it clear that the larger object is called `A` and is on the positive x-axis and the smaller object is called `B` and is on the negative x-axis.

### Upgrade instructions

If you are using the `BinaryCompactObject` or `CylindricalBinaryCompactObject` domain creators, implement the follow name changes in your input files:
- `ObjectA` -> `PrimaryRightObjectA`
- `ObjectB` -> `SecondaryLeftObjectB`

Also, additional restrictions were added to the `BinaryCompactObject` domain creator so read the documentation to see what is allowed.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Expansion.hpp
+++ b/src/ControlSystem/ControlErrors/Expansion.hpp
@@ -42,8 +42,8 @@ namespace ControlErrors {
  *   \delta a &= a\left( \frac{\vec{X}\cdot\vec{C}}{||\vec{C}||^2} - 1 \right)
  * \\ \delta a &= a\left( \frac{X_0}{C_0} - 1 \right) \f}
  *
- * where \f$\vec{X} = \vec{x}_B - \vec{x}_A\f$ and \f$\vec{C} = \vec{c}_B -
- * \vec{c}_A\f$. Here, object B is located on the positive x-axis and object A
+ * where \f$\vec{X} = \vec{x}_A - \vec{x}_B\f$ and \f$\vec{C} = \vec{c}_A -
+ * \vec{c}_B\f$. Here, object A is located on the positive x-axis and object B
  * is located on the negative x-axis, \f$\vec{X}\f$ is the difference in
  * positions of the centers of the mapped objects, and
  * \f$\vec{C}\f$ is the difference of the centers of the excision spheres, all
@@ -88,11 +88,11 @@ struct Expansion : tt::ConformsTo<protocols::ControlError> {
     const double current_position_of_B = get<center_B>(measurements)[0];
 
     // A is to the left of B in grid frame. To get positive differences,
-    // take B - A
+    // take A - B
     const double expected_expansion_factor =
         current_expansion_factor *
-        (current_position_of_B - current_position_of_A) /
-        (grid_position_of_B - grid_position_of_A);
+        (current_position_of_A - current_position_of_B) /
+        (grid_position_of_A - grid_position_of_B);
 
     return {expected_expansion_factor - current_expansion_factor};
   }
--- a/src/ControlSystem/ControlErrors/Expansion.hpp
+++ b/src/ControlSystem/ControlErrors/Expansion.hpp
@@ -83,17 +83,23 @@ struct Expansion : tt::ConformsTo<protocols::ControlError> {
     using center_A = control_system::QueueTags::Center<::ah::ObjectLabel::A>;
     using center_B = control_system::QueueTags::Center<::ah::ObjectLabel::B>;
 
-    ASSERT(domain.excision_spheres().count("ObjectAExcisionSphere") == 1,
+    ASSERT(domain.excision_spheres().count(
+               "PrimaryRightObjectAExcisionSphere") == 1,
            "Excision sphere for ObjectA not in the domain but is needed to "
            "compute Expansion control error.");
-    ASSERT(domain.excision_spheres().count("ObjectBExcisionSphere") == 1,
+    ASSERT(domain.excision_spheres().count(
+               "SecondaryLeftObjectBExcisionSphere") == 1,
            "Excision sphere for ObjectB not in the domain but is needed to "
            "compute Expansion control error.");
 
     const double grid_position_of_A =
-        domain.excision_spheres().at("ObjectAExcisionSphere").center()[0];
+        domain.excision_spheres()
+            .at("PrimaryRightObjectAExcisionSphere")
+            .center()[0];
     const double grid_position_of_B =
-        domain.excision_spheres().at("ObjectBExcisionSphere").center()[0];
+        domain.excision_spheres()
+            .at("SecondaryLeftObjectBExcisionSphere")
+            .center()[0];
     const double current_position_of_A = get<center_A>(measurements)[0];
     const double current_position_of_B = get<center_B>(measurements)[0];
 
--- a/src/ControlSystem/ControlErrors/Expansion.hpp
+++ b/src/ControlSystem/ControlErrors/Expansion.hpp
@@ -14,6 +14,7 @@
 #include "DataStructures/DataVector.hpp"
 #include "Options/Options.hpp"
 #include "Parallel/GlobalCache.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/ProtocolHelpers.hpp"
 #include "Utilities/TMPL.hpp"
 #include "Utilities/TaggedTuple.hpp"
@@ -59,6 +60,8 @@ namespace ControlErrors {
  *   control_system::Systems::Expansion Expansion \endlink control system
  */
 struct Expansion : tt::ConformsTo<protocols::ControlError> {
+  static constexpr size_t expected_number_of_excisions = 2;
+
   using options = tmpl::list<>;
   static constexpr Options::String help{
       "Computes the control error for expansion control. This should not "
@@ -79,6 +82,13 @@ struct Expansion : tt::ConformsTo<protocols::ControlError> {
 
     using center_A = control_system::QueueTags::Center<::ah::ObjectLabel::A>;
     using center_B = control_system::QueueTags::Center<::ah::ObjectLabel::B>;
+
+    ASSERT(domain.excision_spheres().count("ObjectAExcisionSphere") == 1,
+           "Excision sphere for ObjectA not in the domain but is needed to "
+           "compute Expansion control error.");
+    ASSERT(domain.excision_spheres().count("ObjectBExcisionSphere") == 1,
+           "Excision sphere for ObjectB not in the domain but is needed to "
+           "compute Expansion control error.");
 
     const double grid_position_of_A =
         domain.excision_spheres().at("ObjectAExcisionSphere").center()[0];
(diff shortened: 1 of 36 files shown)


src/ControlSystem/ControlErrors/Expansion.hpp
```diff
@@ -80,19 +83,32 @@ struct Expansion : tt::ConformsTo<protocols::ControlError> {
...
+           "compute Expansion control error.");
+    ASSERT(domain.excision_spheres().count(
+               "SecondaryLeftObjectBExcisionSphere") == 1,
+           "Excision sphere for ObjectB not in the domain but is needed to "
+           "compute Expansion control error.");
+
     const double grid_position_of_A =
-        domain.excision_spheres().at("ObjectAExcisionSphere").center()[0];
+        domain.excision_spheres()
+            .at("PrimaryRightObjectAExcisionSphere")
```
nilsvu, line 97: 3 modifiers in the name seems excessive (primary/secondary, left/right, A/B). Will also be annoying to use when talking to people (which of the 3 do we actually use in conversations?). My suggestion is to go with the short but descriptive `Left` and `Right`. In their docs we write "this is object A/B in SpEC". Reasoning:

- Left and right are unambiguous in the grid frame and easy to understand.
- We shouldn't have to carry the legacy names A and B from SpEC only for comparisons. Explaining in the docs is enough. @markscheel argued on Slack that with "A/B" in the name you can shorten variable names to "_a" and "_b" suffixes, but I think we shouldn't do that because then we lose the clarity again. Just use "_left" and "_right" suffixes, that's much clearer.
- Currently we don't assume the left or right object is bigger anywhere in the code I think, so we don't need the "Primary/Secondary" modifier. When we add pre/post processing code that makes that assumption we can document it then, e.g. if we end up specifying mass ratio instead of component masses in ID generation code we can say "in the generated ID, the right BH will be the bigger one". I don't think we should introduce this assumption now when we don't have to. "Left" and "Right" is short and clear and easy to say when talking to people.

What do others think? I know naming things is always hard :D

nilsvu (changes requested):

src/ControlSystem/ControlErrors/Expansion.hpp
knelli2: > * Currently we don't assume the left or right object is bigger anywhere in the code I think,

So in @markscheel's cylindrical BBH domain, it does make this assumption. Object A (Right) must be the larger object. The regular BCO domain doesn't make this assumption. That's why my second commit here actually does restrict the BCO domain to have the Right object be the larger one (if both objects are excised). Whether or not we want this we can discuss, but I definitely think we should be consistent between the domains.

haraldp271: It seems that A/B is consistent with SpEC  (as long as 'A' refers to the object near the positive x-axis, which is more massive than the object 'B' near the negative x-axis), is short and concise, and is easily explained to newcomers.  As such I would favour simplicity and consistency with SpEC over other considerations.  

I agree with Nils Vu's concerns that multiple descriptors are just complicated and confusing (e.g. if we have 'PrimaryRight', is there also a 'PrimaryLeft'?).  Moreover, 'primary' isn't fully descriptive by itself ('more massive'?  or 'BH in an NS-BH binary'?), nor is 'Right' ('right with respect to what'?).   So why change that which isn't broken?

By the way, one more consideration: in SXS metadata, we refer to the two objects as "1" and "2".  (with "1" being on the positive x-axis, and being more massive).

nilsvu: I agree, if we don't find a great alternative to A/B then let's just stick with it, at least for now. The easiest course of action is to do nothing :)

I do think that A/B is very nondescriptive, and Left/Right would be clearer. @haraldp271 "right with respect to what?" Answer: To the right of the companion (along the grid-frame x-axis). I find that pretty intuitive :)

nilsdeppe: I'm not happy with A/B. While it's easily explained, it's difficult to remember. I have, yet again, forgotten which one is which. The argument I've heard for keeping A/B to match SpEC is that then people who know what A/B are in SpEC don't need to look up anything in the documentation. I think we should instead focus on what's easy to understand for newcomers to the code, and having to explain A/B and then they having to remember is not ideal. I'm not particularly attached to Primary/Secondary or Left/Right, though I do think Primary/Secondary is better since this describes the physics of the objects and I don't have to think about which one is the bigger hole when looking in dat files. In both cases, I think matching to SpEC is straightforward for anyone who actually remembers what SpEC does: A is to the right (I think?) and the bigger hole (I think?). The divide in opinions here seems to me to be that people who have spent a lot of time dealing with which hole is which in SpEC prefer sticking to SpEC, while people who are less initiated prefer something more descriptive. 

There are 2 conventions happening here, I think. One is the name, and the other is whether the hole on the +x-axis in the grid frame is the larger. I think it is only the second that we need to match to SpEC to keep comparisons between the codes easy and ultimately doesn't affect how easy it is to understand the code. I believe there are no objections to matching SpEC in terms of where the objects are located in the grid frame.

To Harald's point of "why fix which isn't broken" @nilsvu  and are are telling you that A/B is broken. I have been working with SpEC for 8.5 years and still forget which object is which in terms of A and B. In terms of what does Primary mean, I was under the impression that the primary being the more massive object is standard in astronomy, but I'm admittedly not an expert on this. Also, in NS-BH simulations, the BH will be the more massive one in terms of astrophysical simulations, no?

haraldp271: At the end of the day, this is a convention choice, so there will be no natural, best choice.  Let me bring up some other considerations:
- We also have a remnant to think about, which in SpEC is labeled 'C'.  We could use 'Primary', 'Secondary', 'Remnant'.   A, B, C always struck me as a nice sequence of labels, with 'C' also matching the starting letter of the term "common horizon".  

- LVK has standardized on using the labels "1" and "2".  The convention is that  "1" is the more massive object, and that is something LVK folks, and paper-readers just have to remember.     (In paper text, though, the adjectives primary and secondary are used).  
- LVK uses 'f' for the remnant
- In SXS catalogs and papers we also use labels 1 and 2. 

- I have been staring at SpEC BBH diagnostic plots over the last few days, where it is important to show a lot of information in constrained space.   I am pasting two panels of the 24 I am regularly looking at.  In the legends, 'A' and 'B' indicates the two black holes and lead to high information density.  Text like 'Primary' or 'Secondary' would make the legends more unwieldy. Integers (object 1 / 2) would clash with other integers (like numbering elements or spherical harmonic modes). 
image
image

- Perhaps this is just me, but with 'Primary' and 'Secondary', I always wonder whether astronomers perhaps sometimes for some objects defined this based on something they found useful 100 years ago.  E.g. the 'primary' is that component that emits the light that is visible in the telescope.  [I simply don't know, but astronomers have many conventions that aren't theoretically natural].
- (as unrelated side: While astrophysical expectations are that m_NS < m_BH, I can easily imagine that somebody in the future will want to simulate more exotic things, like a NS-BH binary with a small m_BH.  Or something with exotic compact objects like boson stars, which have other maximum masses than NS.  So, while unrelated, it would be good to avoid an implicit assumption that a BH will always be the most massive object).

haraldp271: > I do think that A/B is very nondescriptive, and Left/Right would be clearer. @haraldp271 "right with respect to what?" Answer: To the right of the companion (along the grid-frame x-axis). I find that pretty intuitive :)

Intuition is also related to the experience of the beholder ;).  E.g. here, you are helped in liking 'right' by a lot of coding on grids.   Whereas for me, the term 'right' brings to mind the notion of observer-dependence.  

I imagine a movie of two BHs orbiting, where BH_right will sometimes be to the left of BH_left.  ... or if this is a precessing system,  BH_right could be __above__ BH_left.

nilsdeppe: - Sure, it's a nice, nondescriptive sequence 🙃 To me that's more convincing of humans liking patters than a good naming convention 🙃 
- Yes, we should consider what LVK does, though I think in this case moving from A to 1 doesn't really clarify anything.
- Maybe we should use labels "primary" and "secondary" in papers since these are clearer. This would also be consistent with LVK paper convention then.
- In regards to the plots, I'm sure it's not at all confusing that `SphereC` is _not_ the sphere around `AhC`. I think using `SphereP`, `SphereS`, `SphereR` would be fine fine matching the AH names while being shorter. This is an important thing to try and keep consistent, though, I agree. My current problem is that I still need to memorize the convention of where A and B are on the grid, in addition to which is the bigger hole. At least P and S would allow me to forget about which is the bigger hole...
- At least the fields I'm familiar with (planetary and compact objects) seem to use primary for the more massive object. This is pretty standard.
- Sure, one can always come up with an exotic system that has different or uncommon settings. That's not a reason to make 99% of the simulations difficult to understand. Also, there's no ambiguity if the primary is always the more massive object. Then the BH is the secondary because it's less massive. Seems consistent to me.
- Intuition: I agree, and this is why you prefer `A` and `B` 🙃 As I said, 8.5 years later and I still forget which is which because it is silly information to have to store.

nilsvu, line 97: @nilsdeppe and I came up with a new suggestion:

- Since the domain creators and the control systems care about the two excisions being "left" or "right" we use that for option names in the domain creator (quite intuitive when writing input files and building the initial domain) and for control system things in the code. This deep into the code we can assume people know about the grid frame which makes "left" and "right" unambiguous.
- For output datasets we let the user choose subfile names for the two horizons. We should let the user choose subfile names anyway so they know where to look for the output data just by reading the input file. Of course this means we offload the problem of naming the horizons to the user, but they can choose what makes most sense for their simulation and anyone can use the input file to get the names. Probably "Primary" and "Secondary" are good names here for BBH simulations, but for a comparison run with SpEC the user could name them "A" and "B" (if they really want to). This also means that output files from different simulations can have different subfile names, which is always the case when we let users choose subfile names (or file names) in the input file. Since the names are all in the input file, I think this is better than hardcoded names that magically appear in the output files. For archiving runs / building catalogs we can consider normalizing the names later, using the input files. Note that input files are also written into the H5 files and can be extracted with `spectre extract-input`. The input files could look like this:

```yaml
DomainCreator:
  BinaryCompactObject:
    ObjectRight:  # Currently "ObjectA"
      XCoord: 8.  # Now obvious that this must be positive
      # ...

ApparentHorizons:
  Right:  # Currently "AhA"
    SubfileName: "Primary"  # New option
    FastFlow:
      # ...
```

PR #4532, opened 2022-12-19
Swap meaning of objects A and B in BinaryCompactObject domain creator

knelli2:
Now object A is the object on the positive x-axis and object B is on the negative x-axis.

Also all control errors have been updated with the new meaning along with the BinaryTrajectories helper struct.

## Proposed changes

The cylindrical BBH domain has the same convention as SpEC, that object A is on the positive x-axis, and B is on the negative x-axis. The BinaryCompactObject BBH domain has the opposite convention. This swaps the BCO BBH domain to be the same as the cylindrical BBH domain and SpEC. We should be consistent with all of our binary domains.

### Upgrade instructions

When using the BinaryCompactObject domain creator, make sure that ObjectA is on the right (on the positive x-axis) and ObjectB is on the left (on the negative x-axis).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Expansion.hpp
+++ b/src/ControlSystem/ControlErrors/Expansion.hpp
@@ -42,8 +42,8 @@ namespace ControlErrors {
  *   \delta a &= a\left( \frac{\vec{X}\cdot\vec{C}}{||\vec{C}||^2} - 1 \right)
  * \\ \delta a &= a\left( \frac{X_0}{C_0} - 1 \right) \f}
  *
- * where \f$\vec{X} = \vec{x}_B - \vec{x}_A\f$ and \f$\vec{C} = \vec{c}_B -
- * \vec{c}_A\f$. Here, object B is located on the positive x-axis and object A
+ * where \f$\vec{X} = \vec{x}_A - \vec{x}_B\f$ and \f$\vec{C} = \vec{c}_A -
+ * \vec{c}_B\f$. Here, object A is located on the positive x-axis and object B
  * is located on the negative x-axis, \f$\vec{X}\f$ is the difference in
  * positions of the centers of the mapped objects, and
  * \f$\vec{C}\f$ is the difference of the centers of the excision spheres, all
@@ -87,12 +87,12 @@ struct Expansion : tt::ConformsTo<protocols::ControlError> {
     const double current_position_of_A = get<center_A>(measurements)[0];
     const double current_position_of_B = get<center_B>(measurements)[0];
 
-    // A is to the left of B in grid frame. To get positive differences,
-    // take B - A
+    // A is to the right of B in grid frame. To get positive differences,
+    // take A - B
     const double expected_expansion_factor =
         current_expansion_factor *
-        (current_position_of_B - current_position_of_A) /
-        (grid_position_of_B - grid_position_of_A);
+        (current_position_of_A - current_position_of_B) /
+        (grid_position_of_A - grid_position_of_B);
 
     return {expected_expansion_factor - current_expansion_factor};
   }
--- a/src/ControlSystem/ControlErrors/Translation.hpp
+++ b/src/ControlSystem/ControlErrors/Translation.hpp
@@ -42,19 +42,19 @@ namespace ControlErrors {
  * \details Computes the error in how much the system has translated by using
  * Eq. (42) from \cite Ossokine2013zga. The equation is
  *
- * \f[ \left(0, \delta\vec{T}\right) = a\mathbf{q}\left(\mathbf{x}_B -
- * \mathbf{c}_B - \mathbf{\delta q}\wedge\mathbf{c}_B - \frac{\delta
- * a}{a}\mathbf{c}_B \right)\mathbf{q}^*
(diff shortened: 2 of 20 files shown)


src/ControlSystem/ControlErrors/Expansion.hpp
```diff
@@ -88,11 +88,11 @@ struct Expansion : tt::ConformsTo<protocols::ControlError> {
     const double current_position_of_B = get<center_B>(measurements)[0];
 
     // A is to the left of B in grid frame. To get positive differences,
```
markscheel: A should be to the right of B.

src/ControlSystem/ControlErrors/Translation.hpp
```diff
@@ -42,19 +42,19 @@ namespace ControlErrors {
...
- * \f[ \left(0, \delta\vec{T}\right) = a\mathbf{q}\left(\mathbf{x}_B -
- * \mathbf{c}_B - \mathbf{\delta q}\wedge\mathbf{c}_B - \frac{\delta
- * a}{a}\mathbf{c}_B \right)\mathbf{q}^*
+ * \f[ \left(0, \delta\vec{T}\right) = a\mathbf{q}\left(\mathbf{x}_A -
+ * \mathbf{c}_A - \mathbf{\delta q}\wedge\mathbf{c}_A - \frac{\delta
+ * a}{a}\mathbf{c}_A \right)\mathbf{q}^*
  * \f]
  *
- * where object B is located on the positive x-axis, bold face letters are
+ * where object A is located on the positive x-axis, bold face letters are
```
markscheel: do we want to say "in the grid frame" here?

knelli2: Yes I'll add that. It'll have to be changed once we move to the distorted frame, but so will everything else so that's ok.

knelli2:
Ok @markscheel, squashed in the changes

markscheel (dismissed):

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -210,15 +210,15 @@ BinaryCompactObject::BinaryCompactObject(
       }
     }
   };
-  add_object_region("ObjectA", "Shell");  // 6 blocks
-  add_object_region("ObjectA", "Cube");   // 6 blocks
   add_object_region("ObjectB", "Shell");  // 6 blocks
   add_object_region("ObjectB", "Cube");   // 6 blocks
+  add_object_region("ObjectA", "Shell");  // 6 blocks
+  add_object_region("ObjectA", "Cube");   // 6 blocks
```
nilsvu: can we keep this unchanged (more below)? the right-object blocks will come before left-object blocks, but I think that's good.

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -338,7 +338,8 @@ class BinaryCompactObject : public DomainCreator<3> {
     static std::string name() { return "InnerRadius"; }
     using type = Options::Auto<double>;
     static constexpr Options::String help = {
-        "Inner radius of the outer spherical shell. Set to 'Auto' to compute a "
+        "Inner radius of the outer spherical shell. Set to 'Auto' to compute "
+        "a "
```
nilsvu: drop this formatting change

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -383,7 +384,8 @@ class BinaryCompactObject : public DomainCreator<3> {
     static double lower_bound() { return 0.; }
     static double upper_bound() { return 1.; }
     // Suggest spherical frustums to encourage upgrading, but keep supporting
-    // cubical frustums until spherical frustums are sufficiently battle-tested
+    // cubical frustums until spherical frustums are sufficiently
+    // battle-tested
```
nilsvu: drop this formatting change (is your editor line width set to 80 chars correctly?)

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -106,15 +106,15 @@ BinaryCompactObject::BinaryCompactObject(
     number_of_blocks_++;
   }
 
-  if (object_A_.x_coord >= 0.0) {
+  if (object_A_.x_coord <= 0.0) {
```
nilsvu, line 109: This change needs to be carried over to some input files (the tests shouldn't pass currently)

nilsvu (changes requested):

nilsvu:
Thanks for making this consistency change @knelli2!

src/Domain/Creators/BinaryCompactObject.cpp
knelli2: Thanks @gsb76 for help with this!

knelli2:
@nilsvu Squashed in my changes because not rearranging the blocks simplified the changes in the BCO.cpp and Test_BCO.cpp files significantly.

nilsvu (approved):

markscheel (approved):

PR #4534, opened 2022-12-19
Add docs to ah::ObjectLabel

nilsvu:
## Proposed changes

Define that A should always be to the right of B in the grid frame. This is the SpEC convention. We make this assumption in the control systems (see also #4532). Discussion about potentially renaming A/B is in #4529.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ApparentHorizons/ObjectLabel.hpp
+++ b/src/ApparentHorizons/ObjectLabel.hpp
@@ -8,7 +8,12 @@
 
 namespace ah {
 /// Labels for the objects in a binary system.
-enum class ObjectLabel { A, B };
+enum class ObjectLabel {
+  /// The object along the positive x-axis in the grid frame
+  A,
+  /// The object along the negative x-axis in the grid frame
+  B
+};
 
 std::string name(const ObjectLabel x);
 


src/ApparentHorizons/ObjectLabel.hpp
```diff
@@ -8,7 +8,12 @@
 
 namespace ah {
 /// Labels for the objects in a binary system.
-enum class ObjectLabel { A, B };
+enum class ObjectLabel {
+  /// The object on the right side (along the grid-frame x-axis)
+  A,
+  /// The object on the left side (along the grid-frame x-axis)
```
knelli2: Maybe don't say right and left here because that's subjective to viewing angle. I know our convention is negative x on the left, but only saying positive/negative x-axis is unambiguous. How about just
```
The object along the positive/negative x-axis in the grid frame
```
?

knelli2 (changes requested):

src/ApparentHorizons/ObjectLabel.hpp
nilsvu: ok, done 👍

knelli2 (approved):

markscheel (approved):

nilsdeppe (approved):

PR #4111, opened 2022-07-12
Allow TOV executables to use generic equations of state

isaaclegred:
## Proposed changes

This PR changes the structure of the EoS creation from input files and the structure of the TOV initial data for GRMHD to allow for use of generic equations of state implemented in spectre rather than just a polytropic equation of state.  This framework can be extended to allow any executable to use a generic equation of state in the input file.  

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

Currently the EoS is specified in the input file as a subset of the initial data, therefore changing the EoS to be generic only required changing the initial data.  In addition, most likely, the EoS used in the simulation should be made independent of the EoS used to generate the initial data and should be specified separately.  For example, it will become valuable to use 1-D EoSs in the initial data generation  while using a 2-D or 3-D EoS in the evolution.  This can be or can not be added to this pull request.  (Also need to fix merge conflicts with recent changes)

PR #4535, opened 2022-12-19
Clean up comments in Tov.cpp

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
@@ -103,10 +103,6 @@ void lindblom_rhs(
       d_log_conformal_factor = -0.25 * d_mass_over_radius;
     }
   } else {
-    // This statement is triggered by one of the test examples,
-    // Test_Tov.cpp: void {anonymous}::test_rueter()
-    // Test_Tov.cpp: void {anonymous}::test_baumgarte_shapiro()
-    // not sure what to do about that
     if constexpr (CoordSystem == TovCoordinates::Schwarzschild) {
       ASSERT(mass_over_radius < .5,
              "Compactness of Star is greater than BH"
@@ -220,8 +216,8 @@ void TovSolution::integrate(
       // Transform observed radius to isotropic, so we use the isotropic radius
       // for all interpolations below
       observer.radius[i] /= square(observer.conformal_factor[i]);
-      // The interpolation is not safe otherwise
     }
+    // The interpolation is not safe otherwise
     observer.radius.back() = outer_radius_;
     conformal_factor_interpolant_ = intrp::BarycentricRational(
         observer.radius, observer.conformal_factor, 5);
@@ -229,8 +225,6 @@ void TovSolution::integrate(
 
   mass_over_radius_interpolant_ =
       intrp::CubicSpline(observer.radius, observer.mass_over_radius);
-  // log_enthalpy(radius) is almost linear so an interpolant of order 3
-  // maximizes precision
   log_enthalpy_interpolant_ =
       intrp::CubicSpline(observer.radius, observer.log_enthalpy);
 }


src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
```diff
@@ -220,17 +216,15 @@ void TovSolution::integrate(
...
-      // The interpolation is not safe otherwise
     }
+    // The interpolation is not safe otherwise
     observer.radius.back() = outer_radius_;
     conformal_factor_interpolant_ = intrp::BarycentricRational(
         observer.radius, observer.conformal_factor, 5);
   }
 
   mass_over_radius_interpolant_ =
       intrp::CubicSpline(observer.radius, observer.mass_over_radius);
```
nilsvu, line 227: @isaaclegred why was this changed to cubic spline? The answer is probably in #4111 but that's a very long thread.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
```diff
@@ -103,10 +103,6 @@ void lindblom_rhs(
       d_log_conformal_factor = -0.25 * d_mass_over_radius;
     }
   } else {
-    // This statement is triggered by one of the test examples,
-    // Test_Tov.cpp: void {anonymous}::test_rueter()
-    // Test_Tov.cpp: void {anonymous}::test_baumgarte_shapiro()
-    // not sure what to do about that
     if constexpr (CoordSystem == TovCoordinates::Schwarzschild) {
       ASSERT(mass_over_radius < .5,
```
nilsvu, line 107: @isaaclegred why did you add this assert? You changed the `test_baumgarte_shapiro` to use the polytropic scaling relations "to prevent nonsensical behavior" ([here](https://github.com/sxs-collaboration/spectre/pull/4111/files#diff-ef62474f8d2a8b459659b395a68f36b1e6b0c8692373bbe2f7b2a4ad599c818b)). As far as I understand the behavior is perfectly sensical, just in different units!

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
```diff
@@ -220,17 +216,15 @@ void TovSolution::integrate(
       // Transform observed radius to isotropic, so we use the isotropic radius
       // for all interpolations below
       observer.radius[i] /= square(observer.conformal_factor[i]);
-      // The interpolation is not safe otherwise
     }
+    // The interpolation is not safe otherwise
```
nilsvu, line 220: (comment on arbitrary line) @isaaclegred Why did you have to increase the tolerance in `test_tov_dp_dr`?: https://github.com/sxs-collaboration/spectre/pull/4111/files#diff-ef62474f8d2a8b459659b395a68f36b1e6b0c8692373bbe2f7b2a4ad599c818bR165

nilsvu:
@isaaclegred a couple of questions about your recent changes in #4111. That was altogether a very nice PR, don't get me wrong! I just have a couple of questions about the technical details in the TOV solver class.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
```diff
@@ -220,17 +216,15 @@ void TovSolution::integrate(
...
     observer.radius.back() = outer_radius_;
```
nilsvu, line 221: @isaaclegred what goes wrong with the interpolation if you don't do this?

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
nilsdeppe: I don't think so. If I remember correctly, the issue you run into is that, while you can solve the TOV equations for such a compact star, it's not in the wrong units (which are set by the EOS) but you have a TOV star that is a BH and so that's not physical. Admittedly, this was long enough ago that Isaac and I talked about it that I don't remember all the details, but it was essentially that, I think.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
nilsdeppe: Aren't you doing extrapolation if you don't do this? That's naively what I would be concerned about. You want the interpolation to be `[0, outer_radius]`

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
nilsdeppe: I think it was 1) stability and 2) smoothnees of the interpolation. I think either the r=0 or outer radius was going unstable.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
nilsdeppe: I don't remember this. Look at the link I wouldn't expect it to need to be changed. Maybe something that ended up left over from debugging?

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
isaaclegred: Yes, it had to do with stability when many steps are required to reach the boundary of the star.  I didn't know enough about barycentric rational interpolation to figure out exactly what was up, but after something like a radius of 100000000 M_sun came out of the TOV solver I concluded it was probably a pole in the interpolant.  This was a problem primarily for the phase-transition star, where many many steps are needed to get high precision.  You might be able to fix this by downsampling the dense output further?  I'm not sure.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
isaaclegred: Ok yeah perhaps this change should be reverted.  The difficulty is that this is not just a change of units, because the values of m(h)/r(h) change depending on this choice.  If m/r exceeds 1/2, then r(h) is no longer monotonic, which I do not understand.   I think perhaps the problem is that the global M, and R are invariant under this scaling, but the entire star doesn't just scale up, it is reconfigured.  I'm pretty sure that if you tried to construct TOV volume data with the EoS that was there before, it wouldn't make any sense.  I'm fine with removing this check from here, it probably shouldn't be the TOV solvers responsibility to check this, but I also think it's reasonable to call the TOV star that was in the `test_baumgarte_shapiro`   before nonsensical :).

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
isaaclegred: I think it is because the accuracy we got probably dropped in this case when we switched to cubic splines.  In practice this wasn't a big deal because the solution-to-grid discretization error is going to always be worse, but it may be a good motivation to try to find a way to bring back barycentric rational interpolation, if this turns out to be generically true.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
isaaclegred: Floating point error in the line above can make the last `observer_radius` value go above the `outer_radius` value, so if you want to interpolate to `outer_radius` you have to explicitly include it.

knelli2:
@nilsvu will look at this again

PR #3903, opened 2022-03-30
Add XCTS TOV solution, enabling simple head-on BNS initial data

nilsvu:
## Proposed changes

This PR allows solving single TOV stars, and also composing binaries with TOV stars using the `Xcts::AnalyticData::Binary` class. Currently, the matter terms in the XCTS equations are just fixed to the TOV matter profile and the gravity is solved for this source. In the future I'll make the matter terms dynamic, so they solve the Euler equations with some equilibrium conditions (e.g. quasi circular orbits). Until then, this PR is probably enough to seed experimental head-on BNS evolutions to get the GRMHD code ready for BNS mergers.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4539, opened 2022-12-21
Decrease initial step size in TOV integrator, switch conformal factor interpolant to CubicSpline

nilsvu:
## Proposed changes

Rebasing on #4111 led to strong Hamiltonian constraint violations near the origin in #3903. I tracked this down to the switch from BarycentricRational to CubicSpline for the M(r) and ln(h) interpolators in the TOV solution. It looks like the initial step size in the TOV integrator was too large, leading to significant differences between the two interpolations at the start of the integration interval. Here's a quick plot that shows the large step size near r=0 (dots are integration steps, line is the interpolation):

0c712d5b-bf32-4d92-80a7-2d928267ed37

Here's a plot of the adaptive step size in both r and ln(h):

fe635671-3da3-4bd9-9339-16f3c701e2be

Therefore, I reduced the initial step size to `relative_tolerance * (ln(h_R) - ln(h_C))`. That seems to sample the origin densely enough so that the Hamiltonian constraint violations vanish. Here are the updated plots:

7f14e704-887a-4a40-b12c-ee6a7dce700f

ac49ea31-f936-4ce1-a707-1c387499bdb1

Any other ideas would be welcome too.

With the dense sampling near the origin the BarycentricRational interpolation for the conformal factor (which wasn't switched in #4111) became unstable, so I switched it to CubicSpline too.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
@@ -223,8 +223,8 @@ void TovSolution::integrate(
       // The interpolation is not safe otherwise
     }
     observer.radius.back() = outer_radius_;
-    conformal_factor_interpolant_ = intrp::BarycentricRational(
-        observer.radius, observer.conformal_factor, 5);
+    conformal_factor_interpolant_ =
+        intrp::CubicSpline(observer.radius, observer.conformal_factor);
   }
 
   mass_over_radius_interpolant_ =
@@ -300,7 +300,9 @@ void TovSolution::pup(PUP::er& p) {  // NOLINT
   p | injection_energy_;
   p | mass_over_radius_interpolant_;
   p | log_enthalpy_interpolant_;
-  p | conformal_factor_interpolant_;
+  if (coordinate_system_ == TovCoordinates::Isotropic) {
+    p | conformal_factor_interpolant_;
+  }
 }
 
 #define DTYPE(data) BOOST_PP_TUPLE_ELEM(0, data)
--- a/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Python/Bindings.cpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Python/Bindings.cpp
@@ -5,6 +5,10 @@
 
 #include "PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Python/Tov.hpp"
 
+namespace py = pybind11;
+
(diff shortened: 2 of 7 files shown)


teukolsky:
I think "time step" should be "stepsize" in several places. The integration is in r.

nilsvu:
@teukolsky definitely, I'll change the wording. The integration is actually in `ln(h)` :)

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
```diff
@@ -177,11 +177,9 @@ void TovSolution::integrate(
           central_specific_internal_energy, central_pressure)));
   lindblom_rhs<CoordSystem>(make_not_null(&dvars), vars, central_log_enthalpy,
                             equation_of_state);
-  double initial_step =
-      -std::min(std::abs(1.0 / dvars[0]), std::abs(1.0 / dvars[1]));
-  if constexpr (CoordSystem == TovCoordinates::Isotropic) {
-    initial_step = -std::min(std::abs(initial_step), std::abs(1.0 / dvars[2]));
-  }
+  const double initial_step =
```
nilsdeppe: Can you please document why you made this change and the motivation? The previous definition is from Hairer and is what is canonically recommended as the initial step size. Yours is quite arbitrary as far as I can tell.

nilsdeppe (changes requested):
This seems fine, but I wonder if this could be fixed by adjusting the absolute tolerance. The quantities being integrated (u=r^2 & v=m/r~r) are both zero at the origin, I think. I don't know how quickly they grow above the absolute tolerance, but it might be worth adjusting the absolute tolerance to see if that's sufficient of a fix. The fact that decreasing the step size near the origin improved things to me says it's a tolerance issue

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/Tov.cpp
nilsvu: I dropped this change and instead tightened the absolute tolerance. I'm a bit worried about the tolerances here.. they seem rather fragile. An absolute tolerance of 1e-14 wasn't enough for the XCTS constraints. Also, loosening the relative tolerance to just 1e-12 fails the `test_tov_dp_dr`. And, when tightening the absolute tolerance further (to, say, 1e-30 or even 0), the `test_tov` times out in the Newtonian limit case.

nilsdeppe (approved):

PR #4547, opened 2023-01-02
Fix writing functions of time in observer, bind Domain items in Python

nilsvu:
## Proposed changes

- Functions of time were not written by evolution executables because they are stored in the _mutable_ global cache. This is now fixed, and I updated the test volume H5 file in the repo with output from the current EvolveScalarWave3D executable.
- Bind Domain and FunctionsOfTime in Python so they can be read from H5 volume files.
- Add an `inertial_coordinates` function in Python that uses the Domain and FunctionsOfTime to map logical to inertial coords. This will allow things like interpolation routines in Python, and plotting the Lagrange interpolation in DG elements.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/Python/InertialCoordinates.cpp
```diff
@@ -0,0 +1,93 @@
...
+        const tnsr::I<DataVector, Dim, Frame::ElementLogical> logical_coords(
+            element_logical_coords);
+        tnsr::I<DataVector, Dim, Frame::Inertial> inertial_coords{};
+
+        // Map logical to inertial coords
+        const auto& block = domain.blocks()[element_id.block_id()];
+        if (block.is_time_dependent()) {
+          if (not time.has_value()) {
+            throw std::runtime_error(
+                "The 'time' argument is required for time-dependent domains.");
```
knelli2: Do we have a way to test errors in python? The two `throw` statements don't seem to be tested.

tests/Unit/Domain/Python/Test_InertialCoordinates.py
```diff
@@ -0,0 +1,45 @@
...
+            inertial_coordinates_3d(element_logical_coords,
+                                    element_id=element_id,
+                                    domain=domain,
+                                    time=0.,
+                                    functions_of_time=functions_of_time))
+
+        # Domain is [0, 2 pi]^3
+        npt.assert_allclose(
+            inertial_coords,
+            np.array([[0., np.pi], [0., 2 * np.pi], [0., 2 * np.pi]]))
```
knelli2: The comment and the actual array don't match.

knelli2 (changes requested):

tests/Unit/Domain/Python/Test_InertialCoordinates.py
nilsvu: It's correct, extended it a bit to make it clearer :)

src/Domain/Python/InertialCoordinates.cpp
nilsvu: Done, also changed to `invalid_argument` error which maps to `ValueError` in Python

nilsvu:
Rebased on #4543

knelli2 (commented):
Looks good. You can squash.

nilsvu:
Thanks for reviewing!

knelli2 (approved):

nilsdeppe (approved):

PR #4550, opened 2023-01-05
Add preload of jemalloc library to spectre script

kidder:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SpectreSetupPythonPackage.cmake
+++ b/cmake/SpectreSetupPythonPackage.cmake
@@ -33,6 +33,10 @@ configure_or_symlink_py_file(
 )
 # Also link the main entry point to bin/
 set(PYTHON_SCRIPT_LOCATION "spectre")
+set(JEMALLOC_PRELOAD "")
+if(BUILD_PYTHON_BINDINGS AND "${JEMALLOC_LIB_TYPE}" STREQUAL SHARED)
+  set(JEMALLOC_PRELOAD "LD_PRELOAD=\${LD_PRELOAD}\${LD_PRELOAD:+:}${JEMALLOC_LIBRARIES}")
+endif()
 configure_file(
   "${CMAKE_SOURCE_DIR}/cmake/SpectrePythonExecutable.sh"
   "${CMAKE_BINARY_DIR}/tmp/spectre")
--- a/cmake/SpectrePythonExecutable.sh
+++ b/cmake/SpectrePythonExecutable.sh
@@ -3,5 +3,5 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
-PYTHONPATH="@PYTHONPATH@" @Python_EXECUTABLE@ \
+PYTHONPATH="@PYTHONPATH@" @JEMALLOC_PRELOAD@ @Python_EXECUTABLE@ \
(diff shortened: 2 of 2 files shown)


cmake/SpectreSetupPythonPackage.cmake
```diff
@@ -33,6 +33,10 @@ configure_or_symlink_py_file(
 )
 # Also link the main entry point to bin/
 set(PYTHON_SCRIPT_LOCATION "spectre")
+set(JEMALLOC_PRELOAD "")
+if(BUILD_PYTHON_BINDINGS AND "${JEMALLOC_LIB_TYPE}" STREQUAL SHARED)
+  set(JEMALLOC_PRELOAD "LD_PRELOAD=${JEMALLOC_LIBRARIES}")
```
nilsdeppe: Not for this PR, but my general education... @wthrowe would there be issues here if someone did `LD_PRELOAD=... ./bin/specter ...` where either our or the user's `LD_PRELOAD` is overwritten/ignored?

nilsdeppe (dismissed):

nilsvu (dismissed):

cmake/SpectreSetupPythonPackage.cmake
wthrowe: This will overwrite any preexisting value, so if anyone is relying on that it could cause problems.  Properly adding to these colon-separated values is annoying.  I think the most correct thing would be `LD_PRELOAD=${LD_PRELOAD}${LD_PRELOAD:+:}new_lib`.  That middle thing can be read as "Look at `LD_PRELOAD`, and, treating unset as empty (`:`), if it is not empty then substitute (`+`) the string `:`."

I'm not sure whether prepending or appending is correct here.  The above appends, prepending can be done by reversing the order of the three sections.

kidder: squashed a change to append if LD_PRELOAD is defined

cmake/SpectreSetupPythonPackage.cmake
```diff
@@ -33,6 +33,14 @@ configure_or_symlink_py_file(
 )
 # Also link the main entry point to bin/
 set(PYTHON_SCRIPT_LOCATION "spectre")
+set(JEMALLOC_PRELOAD "")
+if(BUILD_PYTHON_BINDINGS AND "${JEMALLOC_LIB_TYPE}" STREQUAL SHARED)
+  if(DEFINED ENV{LD_PRELOAD})
```
wthrowe: I'd be OK with the old version that didn't respect LD_PRELOAD at all, but if you are going to respect it you need to use the value in the script's environment, not cmake's environmment.

wthrowe (changes requested):

cmake/SpectreSetupPythonPackage.cmake
kidder: ah okay; squashed and pushed again...

nilsdeppe (commented):
@wthrowe just merge whenever you're happy! I don't think @nilsvu and I need to re-approve this PR... It's small enough :)

wthrowe (approved):

PR #4557, opened 2023-01-06
Bind block/element_logical_coords in Python

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Python/BlockLogicalCoordinates.cpp
+++ b/src/Domain/Python/BlockLogicalCoordinates.cpp
@@ -1,0 +1,74 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/Python/BlockLogicalCoordinates.hpp"
+
+#include <array>
+#include <cstddef>
+#include <memory>
+#include <optional>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/IdPair.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/BlockLogicalCoordinates.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/Structure/BlockId.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Utilities/GetOutput.hpp"
+#include "Utilities/MakeArray.hpp"
+
+namespace py = pybind11;
+
+namespace domain::py_bindings {
+
+namespace {
+template <size_t Dim>
+void bind_block_logical_coordinates_impl(py::module& m) {  // NOLINT
+  py::class_<
+      IdPair<domain::BlockId, tnsr::I<double, Dim, Frame::BlockLogical>>>(
+      m, ("BlockIdAndLogicalCoord" + get_output(Dim) + "D").c_str());
+  m.def(
+      "block_logical_coordinates",
+      [](const Domain<Dim>& domain,
+         const tnsr::I<DataVector, Dim>& inertial_coords,
+         const std::optional<double>& time = std::nullopt,
+         const std::optional<std::unordered_map<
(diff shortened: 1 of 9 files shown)


src/Domain/Python/BlockLogicalCoordinates.cpp
```diff
@@ -0,0 +1,75 @@
...
+        const tnsr::I<DataVector, Dim, Frame::Inertial> x{inertial_coords};
+        // Transform functions-of-time map to unique_ptrs because pybind11 can't
+        // handle unique_ptrs easily as function arguments (it's hard to
+        // transfer ownership of a Python object to C++)
+        std::unordered_map<
+            std::string,
+            std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
+            functions_of_time_ptrs{};
+        if (functions_of_time.has_value()) {
+          for (const auto& [name, fot] : *functions_of_time) {
```
nilsdeppe, line 53: can this call our `clone_unique_ptrs` function instead? I think that does the same thing. You might even be able to do:
```
block_log... (..., clone_unique_ptrs(functions_of_time.value_or(unordered_map...{}));
```

src/Domain/Python/BlockLogicalCoordinates.cpp
```diff
@@ -0,0 +1,75 @@
...
+namespace {
+template <size_t Dim>
+void bind_block_logical_coordinates_impl(py::module& m) {  // NOLINT
+  py::class_<
+      IdPair<domain::BlockId, tnsr::I<double, Dim, Frame::BlockLogical>>>(
+      m, ("BlockIdAndLogicalCoord" + get_output(Dim) + "D").c_str());
+  m.def(
+      "block_logical_coordinates",
+      [](const Domain<Dim>& domain,
+         const std::array<DataVector, Dim>& inertial_coords,
```
nilsdeppe: I guess when the tensor bindings are merged this could get updated?

src/Domain/Python/ElementLogicalCoordinates.cpp
```diff
@@ -0,0 +1,66 @@
...
+  m.def(
+      "element_logical_coordinates",
+      [](const std::vector<ElementId<Dim>>& element_ids,
+         const std::vector<std::optional<
+             IdPair<domain::BlockId,
+                    tnsr::I<double, Dim, typename Frame::BlockLogical>>>>&
+             block_coord_holders) {
+        const auto element_logical_coords =
+            element_logical_coordinates(element_ids, block_coord_holders);
+        // Transform result to Python types
```
nilsdeppe: is what's transformed here tensor->array? If not, it would be helpful to state what is actually transformed and why. It's not clear to me

nilsdeppe (changes requested):
Looks good! Squash any changes immediately please!

src/Domain/Python/BlockLogicalCoordinates.cpp
nilsvu: The argument type here is a map of const-refs, not unique_ptrs, so I think clone_unique_ptrs doesn't have a suitable overload. (const-refs can be passed easily from python, unique_ptrs are harder to pass because apparently it's hard to transfer ownership of an object from Python to C++)

src/Domain/Python/BlockLogicalCoordinates.cpp
nilsvu: done

src/Domain/Python/ElementLogicalCoordinates.cpp
nilsvu: With the Tensor bindings I removed this code and bound the return value struct instead

src/Domain/Python/BlockLogicalCoordinates.cpp
nilsdeppe: Oh right... Sorry!

nilsdeppe (dismissed):

nilsdeppe:
I don't understand the clang-tidy error. Could you add a nolint or explanation or something?

nilsvu:
clang-tidy was probably confused that we didn't do anything with the result of `py::class_<BlockId>`. Added a binding for `get_index` since why not :)

nilsdeppe (approved):

PR #4560, opened 2023-01-09
Bind tensor derivatives in Python

nilsvu:
### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4561, opened 2023-01-09
Add ApplyPointwise to Python CLI

nilsvu:
## Proposed changes

Pointwise transformations of volume data. Allows to run any Python function over H5 vol data files. Doesn't yet do derivatives (need #4560 and a bit more for that) or parallelization (need to think a bit more about the best way to parallelize this). Should be functional enough to generate derived datasets from volume data though.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/ApplyPointwise.py
+++ b/src/Visualization/Python/ApplyPointwise.py
@@ -1,0 +1,327 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import click
+import importlib
+import inspect
+import logging
+import numpy as np
+import re
+import spectre.IO.H5 as spectre_h5
+from pydoc import locate
+from typing import Union, Iterable, Dict, Optional, Sequence, List
+
+logger = logging.getLogger(__name__)
+
+
+def snake_case_to_camel_case(s: str) -> str:
+    """snake_case to CamelCase"""
+    return "".join([t.title() for t in s.split("_")])
+
+
+def parse_pybind11_signature(callable) -> inspect.Signature:
+    # Pybind11 functions don't currently expose a signature that's easily
+    # readable, see issue https://github.com/pybind/pybind11/issues/945.
+    # Therefore, we parse the docstring. Its first line is always the function
+    # signature.
+    match = re.match(callable.__name__ + r"\((.+)\) -> (.+)", callable.__doc__)
+    if not match:
+        raise ValueError(
+            f"Unable to extract signature for function '{callable.__name__}'. "
+            "Please make sure it is a pybind11 binding of a C++ function. "
+            "If it is, please file an issue and include the following first "
+            "line of its docstring:\n\n" + callable.__doc__.partition("\n")[0])
+    match_args, match_ret = match.groups()
+    parameters = []
+    for arg in match_args.split(","):
+        arg = arg.strip()
+        if " = " in arg:
+            arg, default_value = arg.split(" = ")
+        else:
+            default = inspect.Parameter.empty
+        name, annotation = arg.split(": ")
+        annotation = locate(annotation)
+        parameters.append(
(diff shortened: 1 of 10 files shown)


src/Visualization/Python/ApplyPointwise.py
```diff
@@ -0,0 +1,313 @@
...
+
+    Any pybind11 binding of a C++ function will also work, as long as it takes
+    exclusively tensors as arguments. The kernels can be loaded from any
+    available Python module, such as 'spectre.PointwiseFunctions'. You can also
+    execute a Python file that defines kernels with the '--exec' / '-e' option.
+
+    By default, the data for the input arguments are read from datasets in the
+    volume files with the same names, transformed to CamelCase. For example, the
+    input dataset names for the 'shift_magnitude' function above would be
+    'Shift(_x,_y,_z)' and 'SpatialMetric(_xx,_yy,_zz,_xy,_xz,_yz)'. The output
```
nilsdeppe: Maybe add a sentence "That is, the code uses the name `shift` from the function argument, changes it to CamelCase, then reads the `Shift(_x,_y,_z)` datasets into a `tnsr.I[DataVector, 3]` before passing it to the function." This would've made it easier for me to understand.

tests/Unit/Visualization/Python/Test_ApplyPointwise.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_snake_case_to_camel_case(self):
+        self.assertEqual(snake_case_to_camel_case("hello_world"), "HelloWorld")
+
+    def test_apply_pointwise(self):
+        open_h5_files = [spectre_h5.H5File(self.h5_filename, "a")]
+        open_volfiles = [
+            h5file.get_vol("/element_data") for h5file in open_h5_files
+        ]
+
+        kernel = Kernel(psi_squared)
```
nilsdeppe: Is there an easy way to test something like the shift magnitude where multiple components need to be retrieved and combined?

nilsdeppe (changes requested):
This looks great! Couple suggestions. Let me know whether or not you agree :) You can squash immediately if you like them!

tests/Unit/Visualization/Python/Test_ApplyPointwise.py
nilsvu: added another kernel, please see fixup

nilsdeppe (commented):
Awesome! Thank you! Please squash!

nilsdeppe (dismissed):

src/Visualization/Python/ApplyPointwise.py
```diff
@@ -0,0 +1,327 @@
...
+            # can slice the tensor data into elements and call those kernels
+            # that request it element-wise, passing them the mesh and Jacobian
+            # as well.
+            for kernel in kernels:
+                kernel_args = [
+                    all_tensor_data[input_name]
+                    for input_name in kernel.input_names
+                ]
+
+                transformed_tensor = kernel.callable(*kernel_args)
```
nilsdeppe, line 180: If I read this right, each kernel must return exactly one tensor. Would it be easy to (maybe in a followup pR) allow returning a tuple/list/dict of tensors (whatever is easiest)?

nilsdeppe (approved):

src/Visualization/Python/ApplyPointwise.py
nilsvu: Yea I meant to do that but didn't come up with a nice way of specifying the dataset names for the tensors. Maybe returning a dict is the easiest solution.

nilsdeppe: Yea, I would go for a dict :)

PR #4604, opened 2023-01-12
Add option to export FD coordinates in ExportCoords executable

knelli2:
## Proposed changes

Just allows you to export the coordinates of a domain with FD spacing or DG spacing. This will still actually work for curved/time dependent meshes because all we are doing are exporting coordinates. No evolution is actually happening.

Here is a 5x5x5 aligned lattice with a single refinement in the x direction on both a DG grid and a FD grid.

DG Grid:
image

FD Grid:
image

### Upgrade instructions

Add an option called `ActiveGrid:` to the `SpatialDiscretization:` block of your ExportCoordinates input file. Choose either `Dg` or `Subcell` for the active grid.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

A future optimization could be to add the ability to specify which blocks output FD and which output DG.

### DIFF
--- a/src/Executables/ExportCoordinates/ExportCoordinates.hpp
+++ b/src/Executables/ExportCoordinates/ExportCoordinates.hpp
@@ -19,6 +19,8 @@
 #include "Domain/Protocols/Metavariables.hpp"
 #include "Domain/Structure/ElementId.hpp"
 #include "Domain/Tags.hpp"
+#include "Evolution/DgSubcell/Mesh.hpp"
+#include "Evolution/DgSubcell/Tags/ActiveGrid.hpp"
 #include "Evolution/DiscontinuousGalerkin/DgElementArray.hpp"
 #include "Evolution/Initialization/DgDomain.hpp"
 #include "Evolution/Initialization/Evolution.hpp"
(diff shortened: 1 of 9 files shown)


src/Executables/ExportCoordinates/ExportCoordinates.hpp
```diff
@@ -195,6 +197,50 @@ struct FindGlobalMinimumGridSpacing {
...
+      "This will be applied to all elements in the domain. If want to export "
+      "the FD grid, the number of grid points in each element will be 2N-1, "
+      "rather than N."};
+  using group = SpatialDiscretization::OptionTags::SpatialDiscretizationGroup;
+};
+}  // namespace OptionTags
+
+namespace Tags {
+struct UseDgGrid : db::SimpleTag {
+  using type = bool;
```
kidder: Instead of a bool please use the ActiveGrid enum.  I would find this to be much clearer.

kidder (changes requested):

src/Executables/ExportCoordinates/ExportCoordinates.hpp
knelli2: I didn't know that existed, thanks for letting me know!

kidder (approved):

nilsdeppe (approved):

PR #4605, opened 2023-01-13
Add Composition coord map

nilsvu:
## Proposed changes

This makes Jacobians available in Python, which in turn will allow us to take derivatives and integrals over volume data (edit: deferred to next PR).

We can later consider replacing ElementMap with the Composition coordinate map that I'm adding here.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/Composition.hpp
+++ b/src/Domain/CoordinateMaps/Composition.hpp
@@ -1,0 +1,227 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <memory>
+#include <pup.h>
+#include <tuple>
+#include <type_traits>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/CoordinateMaps/CoordinateMap.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Parallel/CharmPupable.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace domain::CoordinateMaps {
+
+/*!
+ * \brief A composition of coordinate maps at runtime
+ *
+ * Composes a sequence of `domain::CoordinateMapBase` that step through the
+ * `Frames`. The result is another `domain::CoordinateMapBase`. This is
+ * different to `domain::CoordinateMap`, which does the composition at compile
+ * time. The use cases are different:
+ *
+ * - Use `domain::CoordinateMap` to compose maps at compile time to go from one
+ *   (named) frame to another using any number of coordinate transformation. The
+ *   coordinate transformations are concatenated to effectively form a single
+ *   map, and intermediate frames have no meaning. This has the performance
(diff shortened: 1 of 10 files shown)


src/Domain/CoordinateMaps/Composition.hpp
```diff
@@ -0,0 +1,212 @@
...
+ *   domain::element_to_block_logical_map), then deforms the logical cube to a
+ *   wedge using the map described above (Grid frame), and then rotates the grid
+ *   with a time-dependent rotation map (Inertial frame).
+ *
+ * \warning Think about performance implications before using this Composition
+ * class. In an evolution it's usually advantageous to keep at least some of the
+ * maps separate, e.g. to avoid reevaluating the static maps in every time step.
+ * You can also access individual components of the composition in this class.
+ */
+template <typename Frames, size_t Dim,
```
nilsdeppe, line 62: can you document the `Frames` template parameter?

Is it possible to write deduction guidelines for the constructor? That would be extremely handy for a class like this...

tests/Unit/Domain/CoordinateMaps/Test_Composition.cpp
```diff
@@ -0,0 +1,151 @@
...
+  CHECK_FALSE(map.is_identity());
+  CHECK_FALSE(map.inv_jacobian_is_time_dependent());
+  CHECK_FALSE(map.jacobian_is_time_dependent());
+
+  const auto xi =
+      make_with_random_values<tnsr::I<DataVector, 2, Frame::ElementLogical>>(
+          make_not_null(&generator), make_not_null(&logical_dist),
+          DataVector(5));
+  const auto x = map(xi);
+  const auto jacobian = map.jacobian(xi);
```
nilsdeppe, line 76: given the non-trivial math being done in the implementation, can you test a 3d map with a full jacobian? At least that jac * inv_jac == identity. Seems tough to get that property to hold by luck :)

nilsdeppe (changes requested):
you can squash immediately :)

src/Domain/CoordinateMaps/Composition.hpp
nilsvu: deduction seems to work 👍 see fixup

tests/Unit/Domain/CoordinateMaps/Test_Composition.cpp
```diff
@@ -39,13 +39,12 @@ void test_composition() {
...
   const ElementId<2> element_id{0, {{{1, 0}, {0, 0}}}};
-  const Composition<
-      tmpl::list<Frame::ElementLogical, Frame::BlockLogical, Frame::Inertial>,
-      2>
-      map{element_to_block_logical_map(element_id),
-          std::make_unique<
-              CoordinateMap<Frame::BlockLogical, Frame::Inertial, Affine2D>>(
-              Affine2D{{-1., 1., 0., 1.}, {-1., 1., 1., 3.}})};
+  // Testing template deduction here
+  const Composition map{
```
nilsdeppe, line 45: Nice! This is a lot simpler :+1:

nilsdeppe (commented):
LGTM, please squash!

nilsdeppe (approved):

PR #4607, opened 2023-01-13
Pin macOS version on CI

nilsvu:
Recently GitHub moved to macOS 12, which made all of our unit tests time out for some reason (maybe the switch to AppleClang 13 -> 14, though I'm building locally with AppleClang 14 just fine). Pin to macOS 11 until we can fix this.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -679,7 +679,7 @@ ${{ env.CACHE_KEY_SUFFIX }}"
   # Build all test executables and run unit tests on macOS
   unit_tests_macos:
     name: Unit tests on macOS
-    runs-on: macos-latest
+    runs-on: macos-11
     env:
       # We install some low-level dependencies with Homebrew. They get picked up
       # by `spack external find`.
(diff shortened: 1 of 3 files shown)


knelli2 (dismissed):
I was noticing this as well, thanks for doing this

nilsvu:
@knelli2 for some reason the BoundaryMessage test reports "subprocess aborted". Any idea why that might be or what's special about the BoundaryMessage test?

nilsvu:
The error message is empty for some reason and the test looks like it successfully ran, but it still reports "subprocess aborted" and fails. Searching for this I found a case where the failure was "malloc: pointer being freed was not allocated". That could be related maybe?

knelli2:
@nilsvu So the BoundaryMessage class does some pointer math and `reinterpret_cast`ing internally. I wonder if a pointer or something isn't being cleaned up properly in the test? I don't have access to a Mac so I can't test for myself, but am happy to debug this with you

nilsvu:
I opened #4610 for the BoundaryMessage test failure. I suggest we merge this PR now to revive CI and Kyle and I work on fixing BoundaryMessage in a follow up.

knelli2 (approved):
If we merge this, Mac CI will run again, but will still fail on the boundary message test. I guess that's better than no tests being run on Mac.

nilsvu:
Yea, currently the macOS CI test runs for 6 hours and times out, that's a lot of wasted CI time. Better have it fail the BoundaryMessage test until we fix it.

kidder (approved):

PR #4611, opened 2023-01-16
Add power monitor python script

nikwit:
## Proposed changes

Adds python functions that compute the modal coefficients of any tensor field from volume data. These can then be used to write power monitors that can be viewed in Paraview.


nilsvu:
Cool! From a first glance it looks like this could simplify significantly (down to just a kernel that computes the power monitors) by using the `ApplyPointwise` script that was merged today: #4561. My understanding is that you're writing a scalar for every direction, so you could return a tnsr.I from the kernel and ApplyPointwise handles all the IO (and soon also the parallelization) for you. What do you think?

On second thought we need to iterate element-wise and pass the Mesh here, which ApplyPointwise doesn't do yet. I could add that capability quickly (was going to wait after #4605 to compute derivatives but can do it now too). Or I refactor your script once that capability is in. Thoughts?

nikwit:
We can just keep this PR open until ypu have merged those changes if you like.

nilsvu:
#4646 is now open. It allows to do this:

```sh
spectre apply-pointwise H5FILES -d element_data \
  -k spectre.Visualization.power_monitors_volume \
  -i tensor_component=Shift_x
```

My suggestion is that you write this function with bindings (maybe in src/Visualization/):

```cpp
template <size_t Dim>
tnsr::I<DataVector, Dim> power_monitors_volume(
  const DataVector& tensor_component, const Mesh<Dim>& mesh) {
    // Expand the power monitors over the volume data
}
```

This gives you all the IO, the CLI, etc for free.

nikwit:
Well that is exactly what @guilara added in #4622, no?

nilsvu:
No we need to expand the modes over the volume like you did in this PR and return a tensor.

guilara:
Hi. The function that I added doesn't fill the volume. it returns an array of size N_x + N_y + N_z, not the product of these. What do you mean precisely @nilsvu by 'expanding the modes in the volume' ? Taking some tensor product of the different directions?

nilsvu:
@guilara and I talked about this today in a call and came to the conclusion that the main thing we need is a function to estimate the truncation error from power monitors. Then we can generate volume data by setting each component of a vector over the grid of an element to the constant truncation error in that dimension.

nikwit:
The volume data is very simulation dependent. For example one could return the magnitude of the highest coefficient or the magnitude of the second highest coefficient. It's impossible to know unless the user checks which modes are affected by the filter. Also, if your solution is decreasing at larger radii it is more useful to return the range that the coefficients cover e.g. log(coef[0]) - log(coefs[-1]) which will tell you how many orders of magnitude are spanned by the solution but it might be very unexpected to get the answer as a log.

This is why I left this function to be specified by the user as `coef_func` in the current implementation and added a description of finding a sensible function as docstring to `write_resolution_data`. Can the user maybe easily write the function themselves with a lambda in the CLI?

nilsvu:
What you're describing is the problem of estimating the truncation error. There's a method described in the Szilagy paper that uses the unfiltered modes. @guilara is implementing a simpler method that was implemented in dg-charm (the pre-spectre prototype). We may have to use unfiltered data for AMR, or cut away the filtered modes. Either way, we need a truncation error function implemented on the C++ side, and then we can use it in pybindings too. I added capability to #4646 to expand a list of doubles over the volume of each element. That should help visualize the truncation error one we have it.

nikwit:
ok, I'm not sure how to proceed here. Given that the trunction error functions will be pretty much 2 lines., should I just add a few sensible choices to the C++ side?

nilsvu:
I think the pybindings in #4725 cover what's missing to visualize a truncation error estimate in 3d. If that's all you need then I think you can close this PR. If there's anything else I didn't think of then let me know!

PR #4612, opened 2023-01-17
Replace CheckCxxLinkerFlag module with CMake one

wthrowe:
Our module is broken, at least in that it calls charmc.  This requires cmake 3.18.

### Upgrade instructions

The minimum version of CMake has been increased to 3.18.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (dismissed):

wthrowe:
Right, the containers need updating before CI will pass.  I think I've updated the definitions correctly in this PR.  Can someone do the rebuild?

nilsdeppe:
I can try build the container at 12:30pm Pacific. I'll report back if I run into issues, then we can coordinate push to DockerHub and merging this :)

cmake/SpectreTargetSources.cmake
```diff
@@ -26,37 +26,11 @@
...
-    list(APPEND _SOURCE_FILES ${SOURCE_FILE})
-  endforeach()
-  target_sources(${TARGET} ${_SOURCE_FILES})
+  # New behavior is available, so just forward to it by ensuring
+  # that we have the policy set to request the new behavior, but
+  # don't change the policy setting for the calling scope
+  cmake_policy(PUSH)
+  cmake_policy(SET CMP0076 NEW)
+  target_sources(${TARGET} ${ARGN})
+  cmake_policy(POP)
```
nilsvu: we don't need this anymore I think. The "new" policy (now always available and set to "new") is to convert relative to absolute paths. So just call `target_sources` and nothing else.

Please update the function docs.

nilsvu (changes requested):

cmake/SpectreTargetSources.cmake
wthrowe: When did the behavior change?  I don't see anything referencing versions between 3.13 and 3.18 in the CMake docs.

nilsvu: Changed in 3.13: https://cmake.org/cmake/help/latest/policy/CMP0076.html

nilsvu: This means the policy is set to "new" automatically for projects with a min cmake version >= 3.13

nilsdeppe:
I've got the dev and ci environments built. I can push them whenever we want to test this PR on that and then merge.

cmake/SpectreTargetSources.cmake
wthrowe: Ah, I see.  It's a nonlocal thing.  I'll update it.

nilsvu (approved):

nilsvu:
@nilsdeppe Do you want to just push the container, then rerun CI, then merge?

nilsvu:
Note that you could do this without disrupting other PRs by pushing an image that still has the old CMake installed (just copy the part of the dockerfile and build). Then the old CMake would get removed from the container next time we build it with the merged version of the dockerfile. Up to you.

nilsdeppe:
Okay, pushed and restart CI

PR #4614, opened 2023-01-17
Add stream operator for MortarData

kidder:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (approved):
I don't think we need to stream about the Jacobian values for Gauss points, but it might prove useful for debugging. I'd say delay that as long as possible, since Kyle's BoundaryMessage stuff will eventually leak into this code.

PR #4622, opened 2023-01-18
Add power monitors

### LINKED ISSUE #4585: Domain diagnostics: add power monitors
# Feature request:

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] Enable spectre to output power monitor spectra

guilara:
## Proposed changes

Add a function that return and array of power monitors in each dimension and the corresponding python bindings and unit tests.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
+++ b/src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
@@ -1,0 +1,72 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "NumericalAlgorithms/LinearOperators/PowerMonitors.hpp"
+
+#include <array>
+#include <cmath>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/ModalVector.hpp"
+#include "DataStructures/SliceIterator.hpp"
+#include "NumericalAlgorithms/LinearOperators/CoefficientTransforms.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace PowerMonitors {
+
+template <size_t Dim>
+void power_monitors(const gsl::not_null<std::array<DataVector, Dim>*> result,
+                const DataVector& input_data_vector, const Mesh<Dim>& mesh) {
+  // Get modal coefficients
+  const ModalVector modal_coefficients =
+      to_modal_coefficients(input_data_vector, mesh);
+
+  double slice_sum = 0.0;
+  size_t n_slice = 0;
+  size_t n_stripe = 0;
+  for (size_t sliced_dim = 0; sliced_dim < Dim; ++sliced_dim) {
+    n_slice = mesh.extents().slice_away(sliced_dim).product();
+    n_stripe = mesh.extents(sliced_dim);
+
+    gsl::at(*result, sliced_dim).destructive_resize(n_stripe);
+
+    for (size_t index = 0; index < n_stripe; ++index) {
+      slice_sum = 0.0;
+      for (SliceIterator si(mesh.extents(), sliced_dim, index); si; ++si) {
+        slice_sum += square(modal_coefficients[si.volume_offset()]);
+      }
+      slice_sum /= n_slice;
+      slice_sum = sqrt(slice_sum);
+
+      gsl::at(*result, sliced_dim)[index] = slice_sum;
+    }
+  }
+}
+
+template <size_t Dim>
+std::array<DataVector, Dim> power_monitors(
+    const DataVector& input_data_vector, const Mesh<Dim>& mesh) {
+  std::array<DataVector, Dim> result{};
+  power_monitors(make_not_null(&result), input_data_vector, mesh);
+  return result;
+}
+
+}  // namespace PowerMonitors
+
+#define DIM(data) BOOST_PP_TUPLE_ELEM(0, data)
+
+#define INSTANTIATE(_, data)                                                   \
+  template std::array<DataVector, DIM(data)>                                   \
+  PowerMonitors::power_monitors(const DataVector& input_data_vector,           \
+                                     const Mesh<DIM(data)>& mesh);             \
+  template void PowerMonitors::power_monitors(                                 \
+    const gsl::not_null<std::array<DataVector, DIM(data)>*> result,            \
+    const DataVector& input_data_vector, const Mesh<DIM(data)>& mesh);
+
+GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3))
+
+#undef INSTANTIATE
+#undef DIM
--- a/src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
+++ b/src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
@@ -1,0 +1,48 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+/// \endcond
+
+/*!
+ * \brief Items for assessing truncation error in spectral methods.
+ */
+namespace PowerMonitors {
+
+/// @{
+/*!
+ * \ingroup SpectralGroup
+ * \brief Returns array of power monitors in each spatial dimension.
+ *
+ * Computed following Sec. 5.1 of Ref. \cite Szilagyi2014fna.
+ * For example, in the x dimension (indexed by \f$ k_0 \f$), we compute
+ *
+ * \f{align*}{
+ *  P_{k_0}[\psi] = \sqrt{ \frac{1}{N_1 N_2}
+ *   \sum_{k_1,k_2} \left| C_{k_0,k_1,k_2} \right|^2} ,
+ * \f}
+ *
+ * where \f$ C_{k_0,k_1,k_2}\f$ are the modal coefficients
+ * of variable \f$ \psi \f$.
+ *
+ */
+template <size_t Dim>
+void power_monitors(gsl::not_null<std::array<DataVector, Dim>*> result,
+                    const DataVector& input_data_vector, const Mesh<Dim>& mesh);
+
+template <size_t Dim>
+std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
+                                           const Mesh<Dim>& mesh);
+/// @}
+
+}  // namespace PowerMonitors
--- a/src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
+++ b/src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
@@ -1,0 +1,33 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "NumericalAlgorithms/LinearOperators/Python/PowerMonitors.hpp"
+
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include "DataStructures/DataVector.hpp"
+#include "NumericalAlgorithms/LinearOperators/PowerMonitors.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+
+namespace py = pybind11;
+
+namespace PowerMonitors::py_bindings {
+
+namespace {
+template <size_t Dim>
+void bind_power_monitors_impl(py::module& m) {  // NOLINT
+  m.def("power_monitors",
+        py::overload_cast<const DataVector&, const Mesh<Dim>&> (
+          &power_monitors<Dim>),
+        py::arg("input_data_vector"), py::arg("mesh"));
+}
+}  // namespace
+
+void bind_power_monitors(py::module& m) {
+  bind_power_monitors_impl<1>(m);
+  bind_power_monitors_impl<2>(m);
+  bind_power_monitors_impl<3>(m);
+}
+
+}  // namespace PowerMonitors::py_bindings
--- a/tests/Unit/NumericalAlgorithms/LinearOperators/Python/Test_PowerMonitors.py
+++ b/tests/Unit/NumericalAlgorithms/LinearOperators/Python/Test_PowerMonitors.py
@@ -1,0 +1,41 @@
+#Distributed under the MIT License.
+#See LICENSE.txt for details.
+
+from spectre.NumericalAlgorithms.LinearOperators import power_monitors
+from spectre.Spectral import (Mesh2D, Basis, Quadrature)
+from spectre.DataStructures import DataVector
+
+import numpy as np
+import unittest
+
+
+class TestPowerMonitors(unittest.TestCase):
+    # Check the case for a constant function where the power monitors
+    # should be given by the first basis function
+    def test_power_monitors(self):
+        num_points_per_dimension = 4
+
+        extent = num_points_per_dimension
+        basis = Basis.Legendre
+        quadrature = Quadrature.GaussLobatto
+        mesh = Mesh2D(extent, basis, quadrature)
+
+        np_vec = np.ones(mesh.number_of_grid_points())
+        test_vec = DataVector(np_vec)
+
+        test_array = power_monitors(test_vec, mesh)
+        np_test_array = np.asarray(test_array)
+
+        check_vec_0 = np.zeros(num_points_per_dimension)
+        check_vec_0[0] = 1.0 / np.sqrt(num_points_per_dimension)
+
+        check_vec_1 = np.zeros(num_points_per_dimension)
+        check_vec_1[0] = 1.0 / np.sqrt(num_points_per_dimension)
+
+        np_check_array = np.array([check_vec_0, check_vec_1])
+
+        np.testing.assert_allclose(np_test_array, np_check_array, 1e-12, 1e-12)
+
+
+if __name__ == '__main__':
+    unittest.main()
--- a/tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
+++ b/tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
@@ -1,0 +1,101 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+// \file
+// Tests of power monitors.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <array>
+#include <cmath>
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Framework/TestCreation.hpp"
+#include "NumericalAlgorithms/LinearOperators/PowerMonitors.hpp"
+#include "NumericalAlgorithms/Spectral/LogicalCoordinates.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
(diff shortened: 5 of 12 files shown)


src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+namespace gsl {
+template <typename T>
+class not_null;
+}  // namespace gsl
```
nilsvu: Delete this forward declaration (it's unused).

docs/GroupDefs.hpp
```diff
@@ -588,6 +588,11 @@
  * \brief Classes and functions useful for performance optimizations.
  */
 
+/*!
+ * \defgroup PowerMonitorsGroup Power Monitors
+ * \brief Functions to assess truncation error in spectral methods.
+ */
```
nilsvu: I think we don't need this group, the documentation on the namespace is enough.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -0,0 +1,53 @@
...
+/// \endcond
+
+/*!
+ * \brief Items for assessing truncation error in spectral methods.
+ */
+namespace PowerMonitors {
+
+/// @{
+/*!
+ * \ingroup PowerMonitorsGroup
```
nilsvu: delete this line too when you delete the group

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -0,0 +1,53 @@
...
+ * \brief Returns array of power monitors in each spatial dimension.
+ *
+ * The are computed following Sec. 5.1 of Ref. \cite Szilagyi2014fna.
```
nilsvu: Sentence missing "power monitors", or did you mean to write "They"?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -0,0 +1,53 @@
...
+ *  P_{k_0}[\psi] = \sqrt{ \frac{1}{N_1 N_2}
+ *   \sum_{k_1,k_2} \left| C_{k_0,k_1,k_2} \right|^2} ,
+ * \f}
+ *
+ * where \f$ C_{k_0,k_1,k_2}\f$ are the modal coefficients
+ * of variable \f$ \psi \f$.
+ *
+ */
+template <size_t Dim>
+std::array<DataVector, Dim> power_monitor_array(const DataVector&,
```
nilsvu: Please name this function just `power_monitors`.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -0,0 +1,53 @@
...
+#include <string>
+
+#include "DataStructures/DataBox/Prefixes.hpp"
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Domain/Tags.hpp"
+#include "Evolution/Systems/ScalarWave/Tags.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
```
nilsvu, line 11: Please check the includes, you don't need all these. Only Gsl, Mesh, and cstddef. Also add `<array>`.

Also check the includes in the cpp file please.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -0,0 +1,53 @@
...
+ *   \sum_{k_1,k_2} \left| C_{k_0,k_1,k_2} \right|^2} ,
+ * \f}
+ *
+ * where \f$ C_{k_0,k_1,k_2}\f$ are the modal coefficients
+ * of variable \f$ \psi \f$.
+ *
+ */
+template <size_t Dim>
+std::array<DataVector, Dim> power_monitor_array(const DataVector&,
+                                                const Mesh<Dim>&);
```
nilsvu: Please add argument names to the function arguments (they must be the same as in the cpp file or one of the automatic checks will complain).

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -0,0 +1,78 @@
...
+    for (size_t index = 0; index < n_stripe; ++index) {
+      slice_sum = 0.0;
+      // We might want to move SliceIterator to the middle loop
+      for (SliceIterator si(mesh.extents(), sliced_dim, index); si; ++si) {
+        slice_sum += square(modal_coefficients[si.volume_offset()]);
+      }
+      slice_sum /= n_slice;
+      slice_sum = sqrt(slice_sum);
+
+      gsl::at(result, sliced_dim).data()[index] = slice_sum;
```
nilsvu: just `gsl::at(result, sliced_dim)[index]` (no need to access `.data()`)

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+      if (index >= gsl::at(result, sliced_dim).size()) {
+        Parallel::printf("Error in index of DataVector in array \n");
+      }
```
nilsvu: Delete this, it should never happen because of the `destructive_resize` above. If you want to write consistency checks like this use `ASSERT` from `Utilities/ErrorHandling/Assert.hpp`. Such asserts run only in Debug builds and get optimized away in Release builds.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
nilsvu: Please add a second overload of this function that "returns by reference". That means it returns `void` but takes a `gsl::not_null<std::array<DataVector, Dim>*>` as first argument. It writes directly into this output buffer so we can write into existing memory. Like this:

```cpp
void power_monitors(gsl::not_null<std::array<DataVector, Dim>*> result,
    const DataVector& input_data, const Mesh<Dim>& mesh);
```

In the existing "return by value" function you can just forward to the "return by reference" overload:

```cpp
std::array<DataVector, Dim> power_monitor_array(const DataVector& input_data, const Mesh<Dim>& mesh) {
  std::array<DataVector, Dim> result{};
  power_monitors(make_not_null(&result), input_data, mesh);
  return result;
}
```

You can find many examples of this pattern in the code, e.g. in `IndefiniteIntegral.*pp`.

In the Python bindings you'll have to disambiguate between the two overloads. Just search for `py::overload_cast` in the code.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -0,0 +1,78 @@
...
+  size_t n_stripe = 0;
+  for (size_t sliced_dim = 0; sliced_dim < Dim; ++sliced_dim) {
+    n_slice = mesh.extents().slice_away(sliced_dim).product();
+    n_stripe = mesh.extents(sliced_dim);
+
+    gsl::at(result, sliced_dim).destructive_resize(n_stripe);
+
+    for (size_t index = 0; index < n_stripe; ++index) {
+      slice_sum = 0.0;
+      // We might want to move SliceIterator to the middle loop
```
nilsvu: Is that possible or useful? Else just delete the comment.

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+#include "DataStructures/DataVector.hpp"
+#include "NumericalAlgorithms/LinearOperators/PowerMonitors.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+
+namespace py = pybind11;
+
+namespace PowerMonitors::py_bindings {
+
+void bind_power_monitor_array(py::module& m) {
```
nilsvu: It's easier to define a function templated on `Dim` like `bind_power_monitors_impl<Dim>` and just call it three times in `bind_power_monitors`. See `Spectral/Python/LogicalCoordinates.cpp` for an example.

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
```diff
@@ -0,0 +1,41 @@
...
+  m.def(
+      "power_monitor_array",
+      [](const DataVector& input_data_vector, const Mesh<1>& mesh) {
+        return PowerMonitors::power_monitor_array(input_data_vector, mesh);
+      },
```
nilsvu: You don't need to wrap this in a lambda function. Just write `&power_monitors<Dim>`. When you have the two overloads, write `py::overload_cast<const DataVector&, const Mesh<Dim>&>(&power_monitors<Dim>)`.

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
```diff
@@ -0,0 +1,41 @@
...
+      py::arg("DataVector"), py::arg("mesh"), py::return_value_policy::copy,
```
nilsvu: - Please name these arguments exactly like the arguments in C++.
- You can delete the `return_value_policy` I think (or did you need it for something?)
- You can also delete the docstring, because people can just look at the C++ function for documentation.

tests/Unit/NumericalAlgorithms/LinearOperators/Python/Test_PowerMonitors.py
```diff
@@ -0,0 +1,41 @@
...
+
+import numpy as np
+import unittest
+
+
+class TestPowerMonitors(unittest.TestCase):
+    # Check the case for a constant function where the power monitors
+    # should be given by the first basis function
+    def test_power_monitor_array(self):
+        elements_per_dimension = 4
```
nilsvu: Name this maybe `num_points`. The word "element" means a finite-element cell in our code, which has a grid of points.

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
```diff
@@ -0,0 +1,116 @@
...
+  // We test against this array
+  const std::array<DataVector, 2> check_power_monitor_array{
+      check_data_vector_x, check_data_vector_y};
+
+  CHECK_ITERABLE_APPROX(test_power_monitor_array, check_power_monitor_array);
+}
+
+void test_power_monitor_array() {
+  test_power_monitor_array_impl();
+  test_power_monitor_array_second_impl();
```
nilsvu: just move these into the `SPECTRE_TEST_CASE` below, no need for the extra function.

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
```diff
@@ -0,0 +1,116 @@
...
+  size_t number_of_points_per_dimension = 4;
+
+  const Mesh<2_st> mesh{number_of_points_per_dimension,
+                        Spectral::Basis::Legendre,
+                        Spectral::Quadrature::GaussLobatto};
+
+  const auto logical_coords = logical_coordinates(mesh);
+  DataVector u_nodal_expected(mesh.number_of_grid_points(), 0.0);
+
+  // Build a test function containing only one Legendre basis function
```
nilsvu, line 59: ... per dimension

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
```diff
@@ -0,0 +1,116 @@
...
+  size_t x_mode = 0;
+  size_t y_mode = 1;
+  std::vector<std::array<size_t, 2>> coeffs_to_include = {{x_mode, y_mode}};
+  auto coeff = coeffs_to_include[0];
+
+  DataVector basis_function =
+      basis_factor *
+      Spectral::compute_basis_function_value<Spectral::Basis::Legendre>(
+          coeff[0], get<0>(logical_coords));
+  for (size_t dim = 1; dim < 2; ++dim) {
```
nilsvu: Begin at dim = 0 and modify `u_nodal_expected` directly? That would simplify this code quite a bit. Also name just `u_nodal` because "expected" usually means the thing you compare to.

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+  const auto logical_coords = logical_coordinates(mesh);
+  DataVector u_nodal_expected(mesh.number_of_grid_points(), 0.0);
+
+  // Build a test function containing only one Legendre basis function
+  double basis_factor = 1.0;
+  size_t x_mode = 0;
+  size_t y_mode = 1;
+  std::vector<std::array<size_t, 2>> coeffs_to_include = {{x_mode, y_mode}};
+  auto coeff = coeffs_to_include[0];
```
nilsvu: Remove the outer std::vector?

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
```diff
@@ -0,0 +1,116 @@
...
+
+  const Mesh<2_st> mesh{number_of_points_per_dimension,
+                        Spectral::Basis::Legendre,
+                        Spectral::Quadrature::GaussLobatto};
+
+  const auto logical_coords = logical_coordinates(mesh);
+  DataVector u_nodal_expected(mesh.number_of_grid_points(), 0.0);
+
+  // Build a test function containing only one Legendre basis function
+  double basis_factor = 1.0;
```
nilsvu: delete?

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
```diff
@@ -0,0 +1,116 @@
...
+      DataVector{number_of_points_per_dimension, 0.0};
+  check_data_vector_x[x_mode] = 1.0 / sqrt(number_of_points_per_dimension);
+
+  // In the y direction
+  DataVector check_data_vector_y =
+      DataVector{number_of_points_per_dimension, 0.0};
+  check_data_vector_y[y_mode] = 1.0 / sqrt(number_of_points_per_dimension);
+
+  // We test against this array
+  const std::array<DataVector, 2> check_power_monitor_array{
```
nilsvu: (this would be the "expected")

nilsvu (changes requested):
The implementation and tests look great! Thanks for implementing this. I have a bunch of comments on formal stuff. Usually we ask people to make changes in separate commits prefixed with "fixup" and push them, so it's easier to see what changed. However, in this case please feel free to squash everything down to a single commit (all commits you have here and any changes), because it's quite short. You'll have to force-push to this branch.

docs/GroupDefs.hpp
nilsdeppe: If we have a `Spectral`-ish group we could park it there. Or something like `NumericalAlgorithmsGroup` if that exists. I agree, this shouldn't have its own group.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -0,0 +1,48 @@
...
+ *  P_{k_0}[\psi] = \sqrt{ \frac{1}{N_1 N_2}
+ *   \sum_{k_1,k_2} \left| C_{k_0,k_1,k_2} \right|^2} ,
+ * \f}
+ *
+ * where \f$ C_{k_0,k_1,k_2}\f$ are the modal coefficients
+ * of variable \f$ \psi \f$.
+ *
+ */
+template <size_t Dim>
+std::array<DataVector, Dim> power_monitors(const DataVector&, const Mesh<Dim>&);
```
nilsdeppe: Can we have a `not_null` implementation of this? I'm worried that with AMR usage this will end up allocating way too often (with GH it would be 150 allocations per AMR call, which is about 150x too many)

nilsvu: Yep here's my comment about that above: https://github.com/sxs-collaboration/spectre/pull/4622#discussion_r1073872567

guilara:
Hi @nilsvu and @nilsdeppe. Thank you for your comments and the thorough review. I've made the requested changes.

docs/GroupDefs.hpp
guilara: I'll put it in Spectral for now. Let me know if you change your mind.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -0,0 +1,72 @@
...
+#include "NumericalAlgorithms/LinearOperators/CoefficientTransforms.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace PowerMonitors {
+
+template <size_t Dim>
+void power_monitors(gsl::not_null<std::array<DataVector, Dim>*> result,
```
nilsvu: Add `const` before `gsl::not_null` (only here, not in the hpp). The reason is that the variable `result` is const in this implementation, it always holds the pointer that was passed in (although the contents of the pointer get modified).

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -0,0 +1,48 @@
...
+ * \f}
+ *
+ * where \f$ C_{k_0,k_1,k_2}\f$ are the modal coefficients
+ * of variable \f$ \psi \f$.
+ *
+ */
+template <size_t Dim>
+void power_monitors(gsl::not_null<std::array<DataVector, Dim>*> result,
+                    const DataVector& input_data_vector, const Mesh<Dim>& mesh);
+/// @}
```
nilsvu: Move this a few lines down so it encloses both overloads. Then the docs apply to both.

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
```diff
@@ -0,0 +1,34 @@
...
+namespace PowerMonitors::py_bindings {
+
+namespace {
+template <size_t Dim>
+void bind_power_monitors_impl(py::module& m) {  // NOLINT
+  m.def("power_monitors",
+        py::overload_cast<const DataVector&, const Mesh<Dim>&> (
+          &power_monitors<Dim>),
+        py::arg("input_data_vector"), py::arg("mesh"),
+        py::return_value_policy::copy);
```
nilsvu: Do you need this return value policy or can you just delete it?

nilsvu (changes requested):
Very nice! Only a couple of minor things.

For some reason GitHub shows a bunch of commits that aren't yours. Could you please rebase on the develop branch and make sure only your new commit sits on top of it? Also your previous commit names show up in the commit message at the moment, that's not particularly important but while you're rebasing why not clean that up.

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
guilara: I remember having trouble with something related to this when I was making the plots in python. I'll remove if for now.

nilsvu: Ok, we can add this back if we encounter problems

nilsvu (approved):
Nice job 👍 Thanks for your work on this!

PR #4629, opened 2023-01-20
Add some AMR helper functions

kidder:
## Proposed changes

These functions are useful for creating new elements and doing sanity checks when doing h-refinement

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Amr/Helpers.cpp
+++ b/src/Domain/Amr/Helpers.cpp
@@ -6,9 +6,11 @@
 #include <array>
 #include <boost/rational.hpp>
 #include <cstddef>
+#include <deque>
 
 #include "Domain/Amr/Flag.hpp"
 #include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/ElementId.hpp"
 #include "Domain/Structure/OrientationMap.hpp"
 #include "Domain/Structure/SegmentId.hpp"
@@ -145,6 +147,47 @@ std::vector<ElementId<VolumeDim>> ids_of_children(
   }
 }
 
+template <size_t VolumeDim>
+std::deque<ElementId<VolumeDim>> ids_of_joining_neighbors(
+    const Element<VolumeDim>& element,
+    const std::array<Flag, VolumeDim>& flags) {
+  using ::operator<<;
+  ASSERT(alg::count(flags, Flag::Join) > 0,
+         "Element " << element.id() << " is not joining given flags " << flags);
+  ASSERT(alg::count(flags, Flag::Split) == 0,
+         "Splitting and joining an Element is not supported");
+  std::deque<ElementId<VolumeDim>> result;
+  for (size_t d = 0; d < VolumeDim; ++d) {
+    if (gsl::at(flags, d) == Flag::Join) {
+      const Direction<VolumeDim> sibling_direction{
+          d, element.id().segment_id(d).side_of_sibling()};
+      const auto& neighbor_sibling_ids_in_this_dim =
+          element.neighbors().at(sibling_direction).ids();
+      for (auto sibling_id : neighbor_sibling_ids_in_this_dim) {
+        result.emplace_back(sibling_id);
+      }
+    }
+  }
+  return result;
+}
+
+template <size_t VolumeDim>
+bool is_child_that_creates_parent(const ElementId<VolumeDim>& element_id,
+                                  const std::array<Flag, VolumeDim>& flags) {
+  using ::operator<<;
+  ASSERT(alg::count(flags, Flag::Join) > 0,
+         "Element " << element_id << " is not joining given flags " << flags);
+  ASSERT(alg::count(flags, Flag::Split) == 0,
+         "Splitting and joining an Element is not supported");
+  for (size_t d = 0; d < VolumeDim; ++d) {
+    if (gsl::at(flags, d) == Flag::Join and
+        element_id.segment_id(d).index() % 2 == 1) {
+      return false;
+    }
+  }
+  return true;
+}
+
 #define DIM(data) BOOST_PP_TUPLE_ELEM(0, data)
 
 #define INSTANTIATE(_, data)                                                   \
@@ -163,7 +206,13 @@ std::vector<ElementId<VolumeDim>> ids_of_children(
       const std::array<amr::domain::Flag, DIM(data)>& flags);                  \
   template std::vector<ElementId<DIM(data)>> ids_of_children(                  \
       const ElementId<DIM(data)>& element_id,                                  \
-      const std::array<amr::domain::Flag, DIM(data)>& flags);
+      const std::array<amr::domain::Flag, DIM(data)>& flags);                  \
+  template std::deque<ElementId<DIM(data)>> ids_of_joining_neighbors(          \
+      const Element<DIM(data)>& element,                                       \
+      const std::array<Flag, DIM(data)>& flags);                               \
+  template bool is_child_that_creates_parent(                                  \
+      const ElementId<DIM(data)>& element_id,                                  \
+      const std::array<Flag, DIM(data)>& flags);
 
 GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3))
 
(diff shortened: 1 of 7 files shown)


src/Utilities/BoostHelpers.hpp
```diff
@@ -0,0 +1,36 @@
+// Distributed under the MIT License.
```
nilsvu: Maybe name this file `PupBoost.hpp` (like the `PupStlCpp17.hpp` that we have)? That could make it easier to find. I'm not a fan of the `Helpers` name because it doesn't tell you anything about what's in there except "it's probably not important".

src/Domain/Amr/Helpers.hpp
```diff
@@ -43,6 +44,14 @@ std::array<size_t, VolumeDim> desired_refinement_levels_of_neighbor(
     const std::array<Flag, VolumeDim>& neighbor_flags,
     const OrientationMap<VolumeDim>& orientation);
 
+/// \ingroup AmrGroup
+/// Fraction of the logical volume of a block covered by an element
+///
+/// \note The sum of this over all the elements of a block should be one
+template <size_t VolumeDim>
+boost::rational<size_t> fraction_of_block_volume(
```
nilsvu, line 57: have you considered using `std::ratio` instead of `boost::rational`? I haven't worked with either, I just found `std::ratio` in the C++ standard.

src/Domain/Amr/Helpers.hpp
```diff
@@ -58,4 +58,11 @@ boost::rational<size_t> fraction_of_block_volume(
 template <size_t VolumeDim>
 bool has_potential_sibling(const ElementId<VolumeDim>& element_id,
                            const Direction<VolumeDim>& direction);
+
+/// \ingroup AmrGroup
+/// \brief The ElementId of the parent of the Element with `element_id` with
+/// refinement `flags`
```
nilsvu: Could you make clearer here how the `flags` relate to the `element_id`? First I thought that `element_id` is the result of applying the `flags` to the `id_of_parent`, so it's a sort of inverse operation, but now I think I have figured out that `id_of_parent` is the result of joining elements with the `element_id`.

src/Domain/Amr/Helpers.hpp
```diff
@@ -65,4 +66,12 @@ bool has_potential_sibling(const ElementId<VolumeDim>& element_id,
 template <size_t VolumeDim>
 ElementId<VolumeDim> id_of_parent(const ElementId<VolumeDim>& element_id,
                                   const std::array<Flag, VolumeDim>& flags);
+
+/// \ingroup AmrGroup
+/// \brief The ElementIds of the children of the Element with `element_id` with
+/// refinement `flags`
```
nilsvu: Same comment about explaining this in a bit more detail

src/Domain/Amr/Helpers.cpp
```diff
@@ -95,6 +95,56 @@ ElementId<VolumeDim> id_of_parent(const ElementId<VolumeDim>& element_id,
...
+      gsl::at(child_segment_ids, d) =
+          gsl::at(flags, d) == Flag::Split
+              ? std::vector{segment_id.id_of_child(Side::Lower),
+                            segment_id.id_of_child(Side::Upper)}
+              : std::vector{segment_id};
+    }
+
+    std::vector<ElementId<VolumeDim>> result{};
+    for (const auto segment_id_xi : child_segment_ids[0]) {
+      for (const auto segment_id_eta : child_segment_ids[1]) {
```
nilsvu, line 132: FYI there's a similar implementation in `ParallelAlgorithms/LinearSolver/Multigrid/Hierarchy.cpp`. There I have an `_impl` function that takes a SegmentId, and template specializations for 1D, 2D, and 3D ElementIds that do an increasing number of loops. I find that easier to understand, but up to you.

src/Domain/Amr/Helpers.hpp
```diff
@@ -74,4 +78,25 @@ template <size_t VolumeDim>
...
     const std::array<Flag, VolumeDim>& flags);
+
+/// \ingroup AmrGroup
+/// \brief The ElementIds of the neighbors of `element` that will join with it
+/// given refinement `flags`
+///
+/// \note This function only returns the face neighbors of `element` that will
+/// join with it, and not the joining corner neighbors
+template <size_t VolumeDim>
+std::deque<ElementId<VolumeDim>> ids_of_joining_neighbors(
```
nilsvu, line 101: Why `std::deque`?

nilsvu (changes requested):

src/Domain/Amr/Helpers.hpp
kidder: std::ratio is a compile time rational

nilsvu: Oh right 👍

src/Domain/Amr/Helpers.hpp
nilsdeppe: Yea, based on this comment I have no idea what the output is from different inputs...

src/Domain/Amr/Helpers.cpp
nilsdeppe: I prefer `if constexpr`

src/Domain/Amr/Helpers.cpp
```diff
@@ -145,6 +147,47 @@ std::vector<ElementId<VolumeDim>> ids_of_children(
...
+template <size_t VolumeDim>
+std::deque<ElementId<VolumeDim>> ids_of_joining_neighbors(
+    const Element<VolumeDim>& element,
+    const std::array<Flag, VolumeDim>& flags) {
+  using ::operator<<;
+  ASSERT(alg::count(flags, Flag::Join) > 0,
+         "Element " << element.id() << " is not joining given flags " << flags);
+  ASSERT(alg::count(flags, Flag::Split) == 0,
+         "Splitting and joining an Element is not supported");
+  std::deque<ElementId<VolumeDim>> result;
```
nilsdeppe, line 159: Maybe do a double pass to be able to use `reserve` to avoid repeated allocations? Or possibly reserve the max size. I'm not sure how big the max size is and how long this would be kept around in terms of memory overhead.

nilsdeppe (changes requested):
I'm happy for you to squash

src/Domain/Amr/Helpers.hpp
wthrowe: What about `Utilities/Rational.hpp`, which we have because the boost one is really slow?

Edit: Although I guess we already use `boost::rational` in a number of places.

kidder: This is only used for a sanity check reduction, so speed shouldn't be an issue.

When I add the reduction action using this I'll try out using `Rational` instead of `boost::rational`

src/Domain/Amr/Helpers.cpp
kidder: originally I did have 3 functions, but I switched to this as I liked it better...

src/Domain/Amr/Helpers.cpp
kidder: there is no `reserve` for a `std::deque`.   In 3D the container returned by this function will have 1-4 elements.  
The deque is not kept around long.  It will probably be more clear once I add the action that uses this function, so I could drop this commit for now if you like

src/Domain/Amr/Helpers.hpp
kidder: the action that uses it pops from the front and pushes to the back...

kidder:
fixups pushed

kidder:
fixups v2 pushed

nilsdeppe (commented):
LGTM! Thanks for the clarifications!

nilsvu (dismissed):

nilsvu (approved):

src/Domain/Amr/Helpers.cpp
nilsdeppe: It's fine to keep. My concern is that, independent of the length of the deque, it might do a lot of allocations (e.g. if it grows to size 4 it might do as many as 4 allocations). But we can address that later, and maybe see if Boost has a static deque, or implement our own. We'll see :)

nilsdeppe (approved):

src/Domain/Amr/Helpers.cpp
wthrowe: > static deque

I've pretty much written this for my new time-stepper history implementation, although I haven't separated it out into a separate utility.  I'll see how easy it is to factor out, since it sounds like it might be useful elsewhere.

PR #4630, opened 2023-01-20
Make Jacobians available in Python

nilsvu:
## Proposed changes

This PR adds adds pybindings for the ElementLogical -> Inertial coordinate map in our domains. This makes Jacobians available in addition to coordinates. Therefore, it replaces the `inertial_coordinates` pybinding. The map is used in `iter_elements` to compute coordinates, Jacobians, etc for each element on demand. The next PR will use this to enable derivatives and integrals in ApplyPointwise.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
@nikwit do you have some time to review this?

src/IO/H5/Python/IterElements.py
```diff
@@ -4,16 +4,62 @@
...
+
+    @property
+    def dim(self):
+        return self.mesh.dim
+
+    @cached_property
+    def logical_coordinates(self):
+        return logical_coordinates(self.mesh)
+
+    @cached_property
```
nikwit, line 45: if you mutate the `self.time`, the cache will not be aware of this afaik, so it would always give you the value of the first evaluation, no?

nikwit (commented):
looks good, one major point

src/IO/H5/Python/IterElements.py
nilsvu: The time won't be mutated because a new `Element` is returned at every observation ID. Pushed a fixup that makes the Element class immutable.

src/IO/H5/Python/IterElements.py
```diff
@@ -38,6 +84,14 @@ def iter_elements(volfiles: Union[spectre_h5.H5Vol,
...
+        domain = deserialize_domain[dim](volfile.get_domain(obs_id))
+        if domain.is_time_dependent():
+            time = volfile.get_observation_value(obs_id)
+            functions_of_time = deserialize_functions_of_time(
+                volfile.get_functions_of_time(obs_id))
+        else:
+            time = None
+            functions_of_time = None
         all_grid_names = volfile.get_grid_names(obs_id)
         all_element_ids = list(map(ElementId[dim], all_grid_names))
```
nikwit: use list comprehension?

src/IO/H5/Python/IterElements.py
nikwit: ok, I guess that makes sense once we have h-refinement and the elements can change between time steps.

nikwit (dismissed):
looks good, please squash

src/IO/H5/Python/IterElements.py
nilsvu: Don't you like the `map`? I think it's a fine way of converting a list of strings into a list of ElementIds, no?

nikwit: From what I read `map` and `filter` are pretty much considered obsolete to create lists in favor of list comprehensions. But I do not care that much.

nilsvu: ok changed to list comprehension

nilsdeppe (approved):

PR #4633, opened 2023-01-21
Disable HDF5 file locking to prevent crashes

### LINKED ISSUE #4522: Simulations crash with H5 errors when the file is read in another process
# Feature request:

`h5py` [is apparently built with thread-safe hdf5](https://docs.h5py.org/en/stable/whatsnew/2.4.html?highlight=thread#thread-safety-improvements) which is why you can read data in parallel from the same read-only h5 file (it handles locks and stuff). We can't do this with out pybindings. We'll get an error. It would be nice if we could support this if we are using a thread-safe version of hdf5.

nilsvu:
## Proposed changes

Fixes #4522 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupHdf5.cmake
+++ b/cmake/SetupHdf5.cmake
@@ -34,6 +34,34 @@ if(HDF5_IS_PARALLEL)
   endif()
 endif()
 
+# Check if file locking API is available. The versions supporting this feature
+# are listed here:
+# https://github.com/HDFGroup/hdf5/blob/develop/doc/file-locking.md
+include(CheckCXXSourceCompiles)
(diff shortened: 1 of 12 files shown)


nilsdeppe (dismissed):

nilsvu:
Ugh looks like we have to update HDF5 in the container to avoid the new warning. Need 1.10.7+ or 1.12.1+.

nilsdeppe (approved):

nilsvu:
@wthrowe do you want to have a look over this since you pointed us to the HDF5 file locking env variable? Else, feel free to just merge.

wthrowe (approved):

PR #4635, opened 2023-01-21
Add Punctures analytic data and executable

nilsvu:
## Proposed changes

This allows to solve for any set of superposed punctures.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/PointwiseFunctions/AnalyticData/Punctures/MultiplePunctures.hpp
```diff
@@ -0,0 +1,191 @@
...
+  };
+  struct Momentum {
+    using type = std::array<double, 3>;
+    static constexpr Options::String help{
+        "The linear momentum P of the puncture"};
+  };
+  struct Spin {
+    using type = std::array<double, 3>;
+    static constexpr Options::String help{
+        "The angular momentum S of the puncture"};
```
nilsdeppe: What are the bounds? Is this the dimensionless spin?

src/PointwiseFunctions/AnalyticData/Punctures/MultiplePunctures.hpp
```diff
@@ -0,0 +1,191 @@
...
+  };
+  struct Mass {
+    using type = double;
+    static constexpr Options::String help{"The puncture mass M"};
+    static double lower_bound() { return 0.; }
+  };
+  struct Momentum {
+    using type = std::array<double, 3>;
+    static constexpr Options::String help{
+        "The linear momentum P of the puncture"};
```
nilsdeppe: In what units?

nilsdeppe (changes requested):
You can squash immediately

src/PointwiseFunctions/AnalyticData/Punctures/MultiplePunctures.hpp
nilsvu: These are the Bowen-York linear and angular momentum parameters in the puncture equation. They are not dimensionless. I extended the help string a bit. These are the standard parameters in the punctures ID literature (I reference the original Brandt/Brügmann paper in the dox, https://arxiv.org/abs/gr-qc/9703066).

src/PointwiseFunctions/AnalyticData/Punctures/MultiplePunctures.hpp
nilsvu: See comment above

src/PointwiseFunctions/AnalyticData/Punctures/MultiplePunctures.hpp
nilsdeppe: Isn't there a BH associated with the puncture that has a mass, spin and angular momentum? This thing has to have some form of units... What units are you using?

nilsvu: I don't know what other units to put here than P scaling linearly in mass and S scaling quadratically. We're in c=G=1 so there's only a mass scale m, and you can choose it. P/m is dimensionless and S/m^2 is dimensionless. The punctures literature usually chooses P and S including mass factors. You would set P^x=mv for example. So I can't think of anything to add here.

nilsdeppe: Okay, that clarifies thing. I think the important thing is to clearly define that `P^x=mv` ,etc. Can you use the dimensionless spin instead? That's what we use everywhere else and is just easier to work with. I don't know about the linear momentum, but specifying the velocity in the range `[0, 1]` naively seems more intuitive to me, but I don't have any experience with the linear momentum. Is this the coordinate momentum then? So it is bounded above by the coordinate speed of light? These are all very important details that are currently not documented and, from experience, will make it challenging for someone to use. Getting units right usually ends up being a non-negligible part of the battle.

nilsvu: Ok, using the dimensionless spin and momentum seem more useful to me too. The bare mass of each puncture is an unambiguous parameter, so it's easy to rescale spin and momentum when comparing to other codes. Pushed a fixup.

nilsdeppe (commented):
LGTM, thanks for making the change! Please go ahead and squash&rebase

nilsdeppe (approved):

Issue #4580, opened 2023-01-12
Improve messaging with pointer arithmetic


geoffrey4444:
# Feature request:

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] Enhance speed of BBH simulations by improving messaging, using pointer arithmetic

geoffrey4444:
Do this for global time stepping first. Then, (please?) @wthrowe will help as needed as @knelli2 tries local time stepping.

geoffrey4444:
Works for global time stepping + DG ... working on subcell + global time stepping. Then @knelli2 + @wthrowe will meet

geoffrey4444:
@knelli2 has a refactoring PR in... then need to work on communication for subcell (some done, not tested yet), @nilsdeppe will review

PR #4636, opened 2023-01-22
Have MortarData store DataVectors rather than std::vector

knelli2:
## Proposed changes

This is a step towards #4580. Because we will eventually be using charm messages to send the boundary data and we want to avoid copies when we are on the same node, the MortarData needs to be able to store non-owning vectors and std::vector doesn't support this. That's why we make this switch to using DataVectors.

While making this change, I also had to change everything that uses the MortarData. This included ComputeTimeDerivaive, ApplyBoundaryCorrections, and all the subcell code because it previously expected the neighbor data to be a std::vector.

Basically all of the changes are in the last commit.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Helpers/Evolution/DiscontinuousGalerkin/Actions/ComputeTimeDerivativeImpl.tpp
+++ b/tests/Unit/Helpers/Evolution/DiscontinuousGalerkin/Actions/ComputeTimeDerivativeImpl.tpp
@@ -1015,6 +1015,8 @@ void test_impl(const Spectral::Quadrature quadrature,
   CAPTURE(UseMovingMesh);
   CAPTURE(Dim);
   CAPTURE(system_type);
+  CAPTURE(HasPrims);
+  CAPTURE(PassVariables);
   CAPTURE(quadrature);
   CAPTURE(dg_formulation);
   using metavars = Metavariables<Dim, system_type, LocalTimeStepping,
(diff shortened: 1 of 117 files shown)


tests/Unit/Helpers/Evolution/DiscontinuousGalerkin/Actions/ComputeTimeDerivativeImpl.tpp
```diff
@@ -1750,17 +1752,26 @@ void test_impl(const Spectral::Quadrature quadrature,
...
-            boundary_data_on_mortar.data() + boundary_data_on_mortar.size()};
+
+        DataVector expected_data{};
+        if (Spectral::needs_projection(face_mesh, mortar_mesh, mortar_size)) {
+          expected_data = DataVector{
+              mortar_mesh.number_of_grid_points() *
+              Variables<mortar_tags_list>::number_of_independent_components};
+          Variables<mortar_tags_list> projected_packaged_data{
+              expected_data.data(), expected_data.size()};
+          const auto blah = ::dg::project_to_mortar(packaged_data, face_mesh,
```
nilsdeppe: maybe rename `blah` to something a bit clearer? I think you did this in `src` already

nilsdeppe (changes requested):
LGTM, one suggestion.  Thank you so much for doing this! Please squash immediately because I don't want to hold this up very long, given the number of files it changes :)

tests/Unit/Helpers/Evolution/DiscontinuousGalerkin/Actions/ComputeTimeDerivativeImpl.tpp
knelli2: Oops! Yeah forgot to change that. I just copied what I did in `InternalMortarDataImpl.hpp` (also copied the comment that went along with it for clarity)

knelli2:
Also rebased on develop

nilsdeppe (approved):
Thank you so much for doing this! :D Nice work!

PR #4637, opened 2023-01-24
Add Punctures ADM mass

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
These are a lot of lines for adding a single pointwise function along with a compute item and pybindings. I wonder if we can eliminate some of the boilerplate code.. it distracts from spending time on the actual implementation of the function. I don't have great ideas though..

src/PointwiseFunctions/Punctures/AdmIntegrals.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace Punctures {
+
+void adm_mass_integrand(const gsl::not_null<Scalar<DataVector>*> result,
+                        const Scalar<DataVector>& field,
+                        const Scalar<DataVector>& alpha,
+                        const Scalar<DataVector>& beta) {
+  get(*result) =
```
nilsdeppe: I would do:
```cpp
get(*result) =  pow<7>(get(alpha) * (get(field) + 1.) + 1.);
get(*result) = 0.5 / M_PI * get(beta) / get(*result);
```
I'm worried that `pow<7>` will expand into repeated nested expressions (which it should to avoid an allocation)

src/PointwiseFunctions/Punctures/CMakeLists.txt
```diff
@@ -0,0 +1,29 @@
+# Distributed under the MIT License.
```
nilsdeppe, line 1: This whole `AdmMass` thing sounds very `GeneralRelativity`. Why not put it in the Gr directory? If I were looking for this that's where I would look, not `Punctures`

nilsdeppe (changes requested):
You can squash

src/PointwiseFunctions/Punctures/AdmIntegrals.cpp
nilsvu: good call. I did this now:

```cpp
  get(*result) = get(alpha) * (get(field) + 1.) + 1.;
  get(*result) = pow<7>(get(*result));
  get(*result) = 0.5 / M_PI * get(beta) / get(*result);
```

src/PointwiseFunctions/Punctures/CMakeLists.txt
nilsvu: This form of the ADM mass is specific to the puncture system. The XCTS system has another form (split into a volume integral and an excision surface integral) and in GH there's yet another form.

nilsvu:
In the spirit of #4651 I moved the pybindings directly into `Bindings.cpp` so at least there are fewer files.

nilsdeppe (approved):

PR #3792, opened 2022-02-06
Frontera Submit Script Update

AlexCarpenter46:
## Proposed changes

Updates the frontera submit script to look and act like the other submit scripts. Also integrated the ability to automatically resubmit jobs from checkpoints into this submit script as well.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

Issue #2987, opened 2021-03-15
Behavior of submission scripts is not standardized


wthrowe:
The submission scripts for different machines create different directory structures, have different defaults, and possibly other very-user-visible behavior.  They should be standardized to reduce confusion.

nilsdeppe:
Some thoughts on this:
1. I think we should do something similar to SpEC with regard to levs. We should, eventually, standardize what LevN means in terms of target truncation error. We can, before that, do runs as `LevN_%06d`. I don't love SpEC's `_AB`-type naming. While not completely unintuitive, I find just numbers to be a bit easier to deal with. We can do something like 5 digits, which would give us, with a standard 24-hour queue, about 100k days, so almost 3 decades. I doubt we will have any simulations taking that long.
2. I have mixed feelings about the separate Ringdown directory. I think @markscheel said if he could go back he wouldn't have made it separate because it complicates scripts quite a bit. So unless there's an obvious major reason to stick to the `LevN_Ringdown` directory, I say we flatten that out. We've even been discussing just having the C++ executable figure out if it's doing a switch to ringdown. In that case I really see no reason to have a separate sub-directory.
3. I don't really want `LevN_M/Run` directories. This just causes a headache in terms of number of files (we've hit file system limits because of the number of files). I also don't see that it adds anything unless you allow the code to modify input files, which I think we decided strongly against.
4. Do we need a `LevN_M/bin` symlink? I guess this is a way to update the executables in different segments, which we will inevitably need some way to do. @markscheel thoughts on this?
5. While a bit annoying, I think the `EccK/Ev/LevN_M` structure is actually quite reasonable, at least I can't think of anything better.
[...]

PR #4638, opened 2023-01-24
Add resubmit capability to Wheeler submit script

knelli2:
## Proposed changes

This is very similar to the `OceanClang.sh` submit script.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/SubmitScripts/Wheeler.sh
+++ b/support/SubmitScripts/Wheeler.sh
@@ -65,6 +65,39 @@ echo
 
 cd ${RUN_DIR}
 
-mpirun -n ${SLURM_NTASKS} \
+checkpoints=0
+current_checkpoint=000000
+if [[ $checkpoints == 0 ]]; then
+  mpirun -n ${SLURM_NTASKS} \
        ${SPECTRE_EXECUTABLE} ++ppn ${CHARM_PPN} +setcpuaffinity \
        --input-file ${SPECTRE_INPUT_FILE}
+  sleep 10s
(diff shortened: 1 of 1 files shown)


wthrowe (changes requested):
Quoting @kidder in #3792:
> If we are going to standardize submit scripts, there needs to be a group discussion among core developers and other interested parties before proceeding with this...

See also my (rather empty) issue #2987.

My own interpretation of all that is that we don't have consensus that what Ocean (or any other machine) does is desirable, so modifying the behavior of other machines to match is frowned upon until some thought has been put into that question.  I encourage you to start that discussion, although if you want to have it here or on the above issue make sure to advertise it since most potentially interested parties don't read comments on random PRs.

knelli2:
@wthrowe Sorry, this was discussed at the meeting today (1/23/23) and the consensus was that we don't know what all the submit scripts should look like in general, but since Wheeler is used by a lot of us and we want this capability on Wheeler specifically, to just change it for Wheeler and not other machines.

wthrowe (approved):
OK.  (This would have been a good thing to include in the meeting minutes.)

I will object to any future changes along these lines, however.

knelli2:
@wthrowe Sorry I forgot to add that to the minutes. I made an edit at the end for future reference (https://github.com/sxs-collaboration/spectre-call-notes/issues/179)

nilsdeppe (approved):
I agree with @wthrowe, we shouldn't keep doing these hacks. I've added my thoughts on that issue.

PR #4641, opened 2023-01-24
Store block names and groups in Domain/Block, bind in Python

nilsvu:
## Proposed changes

Makes block names and groups accessible through the Domain. Also writes the names to H5 files (through serialization) so they are available in post-processing.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Block.cpp
+++ b/src/Domain/Block.cpp
@@ -23,10 +23,12 @@ Block<VolumeDim>::Block(
     std::unique_ptr<domain::CoordinateMapBase<
         Frame::BlockLogical, Frame::Inertial, VolumeDim>>&& stationary_map,
     const size_t id,
-    DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>> neighbors)
+    DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>> neighbors,
+    std::string name)
     : stationary_map_(std::move(stationary_map)),
       id_(id),
-      neighbors_(std::move(neighbors)) {
+      neighbors_(std::move(neighbors)),
+      name_(std::move(name)) {
   // Loop over Directions to search which Directions were not set to neighbors_,
   // set these Directions to external_boundaries_.
   for (const auto& direction : Direction<VolumeDim>::all_directions()) {
@@ -116,7 +118,7 @@ void Block<VolumeDim>::inject_time_dependent_map(
 
 template <size_t VolumeDim>
 void Block<VolumeDim>::pup(PUP::er& p) {
-  size_t version = 0;
+  size_t version = 1;
   p | version;
   // Remember to increment the version number when making changes to this
   // function. Retain support for unpacking data written by previous versions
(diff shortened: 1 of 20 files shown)


src/Domain/Domain.cpp
```diff
@@ -97,8 +113,7 @@ bool Domain<VolumeDim>::is_time_dependent() const {
 }
 
 template <size_t VolumeDim>
-bool operator==(const Domain<VolumeDim>& lhs,
-                const Domain<VolumeDim>& rhs) {
+bool operator==(const Domain<VolumeDim>& lhs, const Domain<VolumeDim>& rhs) {
   return lhs.blocks() == rhs.blocks() and
          lhs.excision_spheres() == rhs.excision_spheres();
```
nilsdeppe: why not compare `block_groups_`?

tests/Unit/Domain/Test_Domain.cpp
```diff
@@ -97,8 +102,13 @@ void test_1d_domains() {
             std::make_unique<CoordinateMap<Frame::BlockLogical, Frame::Inertial,
                                            CoordinateMaps::Affine>>(
                 make_coordinate_map<Frame::BlockLogical, Frame::Inertial>(
-                    CoordinateMaps::Affine{-1., 1., 2., 0.}))));
+                    CoordinateMaps::Affine{-1., 1., 2., 0.}))),
+        {}, {"Left", "Right"}, {{"All", {"Left", "Right"}}});
     CHECK_FALSE(domain_no_corners.is_time_dependent());
```
nilsdeppe, line 107: can you add a test that checks that if the block groups are different the domain is not equal? I find the current `==` behavior unintuitive

src/Domain/Python/Block.cpp
```diff
@@ -0,0 +1,35 @@
...
+
+namespace domain::py_bindings {
+
+namespace {
+template <size_t Dim>
+void bind_block_impl(py::module& m) {  // NOLINT
+  py::class_<Block<Dim>>(m, ("Block" + get_output(Dim) + "D").c_str())
+      .def("is_time_dependent", &Block<Dim>::is_time_dependent)
+      .def_property_readonly("id", &Block<Dim>::id)
+      .def_property_readonly("name", &Block<Dim>::name);
```
nilsdeppe: Add method to get maps? Up to you :)

nilsdeppe (changes requested):
You can squash immediately

src/Domain/Python/Block.cpp
nilsvu: Depends on #4630 for the `CoordinateMap` bindings. We can add the methods later.

nilsvu: With #4630 merged I just rebased and added this now.

src/Domain/Block.cpp
```diff
@@ -131,11 +133,14 @@ void Block<VolumeDim>::pup(PUP::er& p) {
     p | neighbors_;
     p | external_boundaries_;
   }
+  if (version >= 1) {
+    p | name_;
```
nilsdeppe, line 137: Maybe I'm just missing it, but is there an `==` that should have this for Block too?

nilsdeppe (dismissed):
LGTM, just one question. If anything needs to change, squash immediately :)

src/Domain/Block.cpp
nilsvu: ugh thanks for catching this too 🤦‍♂️

nilsdeppe (approved):

PR #4642, opened 2023-01-24
Add `this_machine` Python function

nilsvu:
## Proposed changes

This PR adds information about machines such as supercomputers. Machines are defined as YAML files in `support/Machines/` and selected in the CMake configuration. This means we can read the machine file in C++ if we want to do resubmissions from within executables at some point. For now I added a Python function that reads the machine file.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Python/Machines.py
+++ b/support/Python/Machines.py
@@ -1,0 +1,92 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+"""Support for host machines, such as supercomputers.
+
+Machines are defined as YAML files in 'support/Machines/'. To add support for a
+new machine, add a YAML file that defines a `Machine:` key with the attributes
+listed in the `Machine` class below.
+
+To select a machine, specify the `MACHINE` option when configuring the CMake
+build.
+"""
+
+import os
+from dataclasses import dataclass
+
+# functools.cache was added in Py 3.9. Fall back to 'lru_cache' in earlier
+# versions, which is pretty much the same but slightly slower.
+try:
+    from functools import cache
+except ImportError:
+    from functools import lru_cache
+    cache = lru_cache(maxsize=None)
+
+import yaml
+
+
+@dataclass(frozen=True)
+class Machine(yaml.YAMLObject):
+    """A machine we know how to run on, such as a particular supercomputer.
+
+    Many configuration options for job submission are hardcoded in the submit
+    script for the machine (such as the total number of available cores per
+    node). Here we provide parameters that we want to adjust when scheduling
+    jobs, but that have sensible defaults based on our experience running on the
+    machine. Parameters are set to 'None' when they have no sensible default for
+    the machine.
+
+    Attributes:
+      Name: A short name for the machine. Must match the YAML file name.
+      Description: A description of the machine. Give some basic context and
+        any information that may help people get started using the machine.
+        Provide links to wiki pages, signup pages, etc., for additional
+        information.
+      DefaultProcsPerNode: Default number of worker threads spawned per node.
+        It is often advised to leave one core per node or socket free for
+        communication, so this might be the number of cores or hyperthreads
+        per node minus one.
+      DefaultQueue: Default queue that jobs are submitted to. On Slurm systems
+        you can see the available queues with `sinfo`.
+      DefaultTimeLimit: Default wall time limit for submitted jobs. For
+        acceptable formats, see: https://slurm.schedmd.com/sbatch.html#OPT_time
+    """
+    yaml_tag = '!Machine'
+    yaml_loader = yaml.SafeLoader
+    # The YAML machine files can have these attributes:
+    Name: str
+    Description: str
+    DefaultProcsPerNode: int
+    DefaultQueue: str
+    DefaultTimeLimit: str
+
+
+# Parse YAML machine files as Machine objects
+yaml.SafeLoader.add_path_resolver('!Machine', ['Machine'], dict)
+
+
+class UnknownMachineError(Exception):
+    """Indicates we were unsuccessful in identifying the current machine"""
+    pass
+
+
+@cache
+def this_machine(machinefile_path=os.path.join(os.path.dirname(__file__),
+                                               'Machine.yaml')) -> Machine:
+    """Determine the machine we are running on.
+
+    Raises `UnknownMachineError` if no machine was selected. Specify the
+    `MACHINE` option in the CMake build configuration to select a machine, or
+    pass the `machinefile_path` argument to this function.
+
+    Arguments:
+      machinefile_path: Path to a YAML file that describes the current machine.
+        Defaults to the machine selected in the CMake build configuration.
+    """
+    if not os.path.exists(machinefile_path):
+        raise UnknownMachineError(
+            "No machine was selected. Specify the 'MACHINE' option when "
+            "configuring the build with CMake. If you are running on a new "
+            "machine, please add it to 'support/Machines/'. The machine file "
+            f"was expected at the following path:\n  {machinefile_path}")
+    with open(machinefile_path, 'r') as open_machinefile:
+        return yaml.safe_load(open_machinefile)['Machine']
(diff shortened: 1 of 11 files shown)


support/Python/Machines.py
```diff
@@ -0,0 +1,74 @@
...
+    Attributes:
+      Name: A short name for the machine. Must match the YAML file name.
+      Description: A description of the machine. Give some basic context and
+        any information that may help people get started using the machine.
+        Provide links to wiki pages, signup pages, etc., for additional
+        information.
+      DefaultProcsPerNode: Default number of worker threads spawned per node.
+        It is often advised to leave one core per node or socket free for
+        communication, so this might be the number of cores or hyperthreads
+        per node minus one.
```
knelli2, line 47: Should we have this taken care of somewhere else? By that I mean, the machine file just specifies how many total cores per node there are, and then somewhere else (maybe when creating submit scripts later on) we specify how many comm/worker cores we want.

The reason I say this is because on the larger machines we have access to (Frontera, Anvil, ...) where there are a ton of cores I found for the BBHs that I actually needed multiple comm cores per physical node, effectively splitting the physical node up into several virtual charm nodes. However, the number of virtual nodes we want to create could be problem dependent (maybe splitting the node in half, maybe 3 virtual nodes, etc...), so in that case we wouldn't want that specified here in the machine file.

Thoughts?

support/Python/Machines.py
```diff
@@ -0,0 +1,74 @@
...
+        acceptable formats, see: https://slurm.schedmd.com/sbatch.html#OPT_time
+    """
+    yaml_tag = '!Machine'
+    yaml_loader = yaml.SafeLoader
+    # The YAML machine files can have these attributes:
+    Name: str
+    Description: str
+    DefaultProcsPerNode: int = None
+    DefaultQueue: str = None
+    DefaultTimeLimit: str = None
```
knelli2: Why make these default to `None`? These seem like important things that should have to be specified.

knelli2 (changes requested):

support/Python/Machines.py
nilsvu: Sorry, I probably wasn't super clear about the purpose of these values. I added the following to the docstring:

> Many configuration options for job submission are hardcoded in the submit
> script for the machine (such as the total number of available cores per
> node). Here we provide parameters that we want to adjust when scheduling
> jobs, but that have sensible defaults based on our experience running on the
> machine. Parameters are set to 'None' when they have no sensible default for
> the machine.

Does that help? For the PPN we could delete the default and require users to always specify it. However, I think the default is useful because I don't want to remember every time when I schedule a run how many cores the machine has and what's the recommended number of worker threads on this machine. The default is based on experience. It can still be adjusted when scheduling a job if necessary, e.g. to do scaling tests. If the ideal PPN ends up being very problem-dependent then we can think about a way to set it based on the problem. For the majority of cases I don't want to think about this every time I schedule a run though.

support/Python/Machines.py
nilsvu: See added docs above. I think we should only provide defaults where they make sense.

support/Python/Machines.py
knelli2: Yeah that makes it clearer. I guess we'll need a utility to edit submit scripts (and update their parameters) once we decide on how we want submit scripts to look.

tests/support/Python/Test_Machines.py
```diff
@@ -0,0 +1,28 @@
...
+        yaml.safe_dump(
+            dict(Machine=dict(Name="TestMachine",
+                              Description="Just for testing")),
+            open(self.machinefile_path, 'w'))
+
+    def test_this_machine(self):
+        with self.assertRaises(UnknownMachineError):
+            this_machine('NonexistentMachinefile.yaml')
+        self.assertEqual(
+            this_machine(self.machinefile_path).Name, "TestMachine")
```
knelli2: Could you add tests for all the parameters you have inside `Machine`? Even the `None` ones, just so we confirm that everything is getting parsed correctly?

tests/support/Python/Test_Machines.py
```diff
@@ -0,0 +1,28 @@
...
+from spectre.Informer import unit_test_build_path
+
+
+class TestMachines(unittest.TestCase):
+    def setUp(self):
+        self.machinefile_path = os.path.join(unit_test_build_path(),
+                                             'TestMachine.yaml')
+        yaml.safe_dump(
+            dict(Machine=dict(Name="TestMachine",
+                              Description="Just for testing")),
```
knelli2: Maybe also add a test where you set one of the fields that would have been `None`?

support/Python/Machines.py
```diff
@@ -0,0 +1,83 @@
...
+        acceptable formats, see: https://slurm.schedmd.com/sbatch.html#OPT_time
+    """
+    yaml_tag = '!Machine'
+    yaml_loader = yaml.SafeLoader
+    # The YAML machine files can have these attributes:
+    Name: str
+    Description: str
+    DefaultProcsPerNode: Optional[int] = None
+    DefaultQueue: Optional[str] = None
+    DefaultTimeLimit: Optional[str] = None
```
knelli2: [Can't continue the previous thread, so making a new one]

Personally, I think these are all things we should have to specify in the `MACHINE.yaml` file (as defaults that could potentially be overridden). If you want a default-created submit script for a machine (like you mentioned above, not having to input number of worker threads every time you schedule a job), you'll need actual values for these, not `None`.

knelli2 (changes requested):
You can squash any changes

support/Python/Machines.py
nilsvu: I'm not sure I understand you correctly. These values can currently be specified in the machine file, and then they'll be used as default. But they don't **have to** be specified in case there's no obvious default (like maybe the machine has multiple queues and you always want the user to choose one or something).

support/Python/Machines.py
nilsvu: Yea I have that written already. My plan is to treat submit scripts as templates with placeholders that can be substituted. They essentially already are templates, only they are edited by hand. So you instead of the current `SPECTRE_BUILD_DIR=/path/to/build/dir` that's in many submit scripts we write `SPECTRE_BUILD_DIR={{ build_dir }}` and the parameter gets substituted. The syntax for placeholders follows Jinja (https://jinja.palletsprojects.com/en/3.1.x/templates/). The CLI is just this:

```sh
$ spectre schedule INPUT_FILE --num_nodes 4  # ...
```

nilsdeppe: Just for SpEC context: SpEC controls things like PPN from a central file (one that holds _all_ supercomputers, so one per system is much nicer). That so far seems to be working quite well. One thing we do need to make sure we can support is running on systems with mixed nodes like Caltech HPC. Ideally we deal with what types of nodes we get at startup, but I don't know CaltechHPC's exact launching config. E.g. if they let peolpe use a subset of a node. We could just restrict ourselves to having the submit script choose whatever queue is emptiest and submit on those nodes :shrug: Nothing to be solved for this PR, just something to think about. I notice multi-queues are mentioned below: I think we want to give good defaults still, ideally we decide dynamically what queue to choose. I think that'll make it easier for new users to just submit jobs and get a reasonable default.

nilsdeppe (commented):
I have no additional requests to @knelli2 but just some things to think about for the future directions

support/Python/Machines.py
knelli2: Ah, ok. I *do* think there are some obvious defaults for these for each machine :) (even the ones with multiple queues). But I don't think this is worth holding up this PR. Leave it as is for now.

nilsvu: You might be right that we'll always have sensible defaults for these. I removed the optional for now and added some tests. If we encounter a machine where we don't have a sensible default, we can make the field optional.

knelli2 (approved):

nilsdeppe (approved):

PR #4644, opened 2023-01-25
Jacobian diagnostic

### LINKED ISSUE #4586: Domain diagnostic: jacobian
# Feature request:

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] Add a diagnostic that compares the jacobian provided by the map to just numerically computing the jacobian.

geoffrey4444:
## Proposed changes

Implements a diagnostic that compares the analytic and numeric Jacobians. Resolves #4586.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/JacobianDiagnostic.cpp
+++ b/src/Domain/JacobianDiagnostic.cpp
@@ -1,0 +1,94 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/JacobianDiagnostic.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeWithValue.hpp"
+
+namespace domain {
+template <size_t Dim, typename Fr>
+void jacobian_diagnostic(
+    const gsl::not_null<
+        tnsr::i<DataVector, Dim, typename Frame::ElementLogical>*>
+        jacobian_diag,
+    const Jacobian<DataVector, Dim, typename Frame::ElementLogical, Fr>&
+        analytic_jacobian,
+    const TensorMetafunctions::prepend_spatial_index<
+        tnsr::I<DataVector, Dim, Fr>, Dim, UpLo::Lo,
+        typename Frame::ElementLogical>& numeric_jacobian_transpose) {
+  const size_t number_of_points = analytic_jacobian.begin()->size();
+  destructive_resize_components(jacobian_diag, number_of_points);
+
+  // i_hat = logical frame index
+  // i = mapped frame index
+  Scalar<DataVector> jacobian_component_sum{number_of_points};
+  for (size_t i_hat = 0; i_hat < Dim; ++i_hat) {
+    jacobian_diag->get(i_hat) = -abs(analytic_jacobian.get(0, i_hat));
+    for (size_t i = 1; i < Dim; ++i) {
+      jacobian_diag->get(i_hat) -= abs(analytic_jacobian.get(i, i_hat));
+    }
+    get(jacobian_component_sum) = abs(numeric_jacobian_transpose.get(i_hat, 0));
+    for (size_t i = 1; i < Dim; ++i) {
+      get(jacobian_component_sum) +=
+          abs(numeric_jacobian_transpose.get(i_hat, i));
+    }
+    jacobian_diag->get(i_hat) /= get(jacobian_component_sum);
+    jacobian_diag->get(i_hat) += 1.0;
+  }
+}
+
+template <size_t Dim, typename Fr>
+void jacobian_diagnostic(
+    const gsl::not_null<
+        tnsr::i<DataVector, Dim, typename Frame::ElementLogical>*>
+        jacobian_diag,
+    const ::Jacobian<DataVector, Dim, Frame::ElementLogical, Fr>&
+        analytic_jacobian,
+    const tnsr::I<DataVector, Dim, Fr>& mapped_coords,
+    const ::Mesh<Dim>& mesh) {
+  // Note: Jacobian has the source frame index second, but
+  // logical_partial_derivative prepends the logical (source frame) index.
+  // So this is actually the transpose of the numerical jacobian.
+  const auto numerical_jacobian_transpose =
+      logical_partial_derivative(mapped_coords, mesh);
+
+  jacobian_diagnostic(jacobian_diag, analytic_jacobian,
+                      numerical_jacobian_transpose);
+}
+}  // namespace domain
+
+#define DIM(data) BOOST_PP_TUPLE_ELEM(0, data)
+#define FRAME(data) BOOST_PP_TUPLE_ELEM(1, data)
+
+#define INSTANTIATE(_, data)                                                \
+  template void domain::jacobian_diagnostic(                                \
+      const gsl::not_null<                                                  \
+          tnsr::i<DataVector, DIM(data), typename Frame::ElementLogical>*>  \
+          jacobian_diag,                                                    \
+      const Jacobian<DataVector, DIM(data), Frame::ElementLogical,          \
+                     FRAME(data)>& analytic_jacobian,                       \
+      const TensorMetafunctions::prepend_spatial_index<                     \
+          tnsr::I<DataVector, DIM(data), FRAME(data)>, DIM(data), UpLo::Lo, \
+          typename Frame::ElementLogical>& numeric_jacobian_transpose);     \
+  template void domain::jacobian_diagnostic(                                \
+      const gsl::not_null<                                                  \
+          tnsr::i<DataVector, DIM(data), typename Frame::ElementLogical>*>  \
+          jacobian_diag,                                                    \
+      const ::Jacobian<DataVector, DIM(data), Frame::ElementLogical,        \
+                       FRAME(data)>& analytic_jacobian,                     \
+      const tnsr::I<DataVector, DIM(data), FRAME(data)>& mapped_coords,     \
+      const ::Mesh<DIM(data)>& mesh);
+
+GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3), (Frame::Grid, Frame::Inertial))
+
+#undef DIM
+#undef FRAME
+#undef INSTANTIATE
--- a/src/Domain/JacobianDiagnostic.hpp
+++ b/src/Domain/JacobianDiagnostic.hpp
@@ -1,0 +1,110 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <string>
+
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Domain/Tags.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+/// \endcond
+
+namespace domain {
+/*!
+ * \ingroup ComputationalDomainGroup
+ * \brief A diagonstic comparing the analytic and numerical Jacobians for a
+ * map.
+ *
+ * Specifically, returns
+ * \f[
+ * C_{\hat{i}} = 1 -
+ * \frac{\sum_i |\partial_{\hat{i}} x^i|}{\sum_i |D_{\hat{i}} x^i|}
+ * \f], where \f$x^{\hat{i}}\f$ are the logical
+ * coordinates, \f$x^i\f$ are the coordinates in the target frame,
+ * \f$\partial_{\hat{i}}x^i\f$ is the analytic Jacobian, and \f$D_{\hat{i}}
+ * x^i\f$ is the numerical Jacobian.
+ *
+ * \note This function accepts the transpose of the numeric Jacobian as a
+ * parameter, since the numeric Jacobian will typically be computed via
+ * logical_partial_derivative(), which prepends the logical (source frame)
+ * derivative index. Tensors of type Jacobian, in contrast, have the derivative
+ * index second.
+ */
+template <size_t Dim, typename Fr>
+void jacobian_diagnostic(
+    const gsl::not_null<
+        tnsr::i<DataVector, Dim, typename Frame::ElementLogical>*>
+        jacobian_diag,
+    const Jacobian<DataVector, Dim, typename Frame::ElementLogical, Fr>&
+        analytic_jacobian,
+    const TensorMetafunctions::prepend_spatial_index<
+        tnsr::I<DataVector, Dim, Fr>, Dim, UpLo::Lo,
+        typename Frame::ElementLogical>& numeric_jacobian_transpose);
+
+/*!
+ * \ingroup ComputationalDomainGroup
+ * \brief A diagonstic comparing the analytic and numerical Jacobians for a
+ * map.
+ *
+ * Specifically, returns
+ * \f[
+ * C_{\hat{i}} = 1 -
+ * \frac{\sum_i |\partial_{\hat{i}} x^i|}{\sum_i |D_{\hat{i}} x^i|}
+ * \f],
+ * where \f$x^{\hat{i}}\f$ are the logical coordinates, \f$x^i\f$ are the
+ * coordinates in the target frame, \f$\partial_{\hat{i}}x^i\f$ is the analytic
+ * Jacobian, and \f$D_{\hat{i}} x^i\f$ is the numerical Jacobian.
+ *
+ * \note This function accepts the analytic jacobian, mapped coordinates, and
+ * mesh as a parameter. The numeric jacobian is computed
+ * internally by differentiating the mapped coordinates with respect to the
+ * logical coordinates.
+ */
+template <size_t Dim, typename Fr>
+void jacobian_diagnostic(
+    const gsl::not_null<tnsr::i<DataVector, Dim, Frame::ElementLogical>*>
+        jacobian_diag,
+    const ::Jacobian<DataVector, Dim, Frame::ElementLogical, Fr>&
+        analytic_jacobian,
+    const tnsr::I<DataVector, Dim, Fr>& mapped_coords, const ::Mesh<Dim>& mesh);
+
+namespace Tags {
+/// \ingroup DataBoxTagsGroup
+/// \ingroup ComputationalDomainGroup
+/// \brief A diagonstic comparing the analytic and numerical Jacobians for a
+/// map. See `domain::jacobian_diagnostic` for details.
+template <size_t Dim>
+struct JacobianDiagnostic : db::SimpleTag {
+  using type = tnsr::i<DataVector, Dim, typename Frame::ElementLogical>;
+};
+
+/// \ingroup DataBoxTagsGroup
+/// \ingroup ComputationalDomainGroup
+/// \brief Computes the Jacobian diagnostic, which compares the analytic
+/// Jacobian (provided by some coordinate map) to a numerical Jacobian computed
+/// using numerical partial derivatives. The coordinates must be in the target
+/// frame of the map. See `domain::jacobian_diagnostic` for details of the
+/// calculation.
+template <size_t Dim, typename TargetFrame>
+struct JacobianDiagnosticCompute : JacobianDiagnostic<Dim>, db::ComputeTag {
+  using base = JacobianDiagnostic<Dim>;
+  using return_type = typename base::type;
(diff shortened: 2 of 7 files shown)


src/Domain/JacobianDiagnostic.hpp
```diff
@@ -0,0 +1,99 @@
...
+      const tnsr::I<DataVector, MapTag::dim, typename MapTag::target_frame>&
+          mapped_coords,
+      const ::Mesh<MapTag::dim>& mesh) {
+    const auto analytic_jacobian =
+        determinant_and_inverse(element_map.inv_jacobian(source_coords)).second;
+    // Note: Jacobian has the source frame index second, but
+    // logical_partial_derivative prepends the logical (source frame) index.
+    // So this is actually the transpose of the numerical jacobian.
+    const auto numerical_jacobian_transpose =
+        logical_partial_derivative(mapped_coords, mesh);
```
nilsvu: I'd say take the derivative in the `jacobian_diagnostic` function so calling the function is agnostic to the transpose stuff. Just pass the inertial coords and the mesh to `jacobian_diagnostic`.

src/Domain/JacobianDiagnostic.hpp
```diff
@@ -0,0 +1,99 @@
...
+  static constexpr auto function(
+      const gsl::not_null<return_type*> jacobian_diag,
+      const typename MapTag::type& element_map,
+      const tnsr::I<DataVector, MapTag::dim, typename MapTag::source_frame>&
+          source_coords,
+      const tnsr::I<DataVector, MapTag::dim, typename MapTag::target_frame>&
+          mapped_coords,
+      const ::Mesh<MapTag::dim>& mesh) {
+    const auto analytic_jacobian =
+        determinant_and_inverse(element_map.inv_jacobian(source_coords)).second;
```
nilsvu: I don't remember which map compositions we store in the DataBox in evolutions, but I know we have both the grid and the inertial inv jacobian in there. So it seems easiest to take the `domain::Tags::InverseJacobian<Dim, Frame::ElementLogical, TargetFrame` as an argument directly instead of the map.

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -0,0 +1,79 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/JacobianDiagnostic.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"     // IWYU pragma: keep
+#include "DataStructures/Tensor/Tensor.hpp"  // IWYU pragma: keep
```
nilsvu: delete iwyu pragmas

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -0,0 +1,79 @@
...
+      numerical_difference_sum.get(i_hat) +=
+          numeric_jacobian_transpose.get(i_hat, i);
+    }
+  }
+  for (size_t i_hat = 0; i_hat < Dim; ++i_hat) {
+    // If the numeric and analytic components are both sufficiently small,
+    // treat them as equal.
+    constexpr double eps_for_ratio = 1.e-10;
+    if (max(abs(jacobian_difference_sum.get(i_hat))) < eps_for_ratio and
+        max(abs(numerical_difference_sum.get(i_hat))) < eps_for_ratio) {
```
nilsvu: Use `equal_within_roundoff` instead of `eps_for_ratio`.

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -0,0 +1,79 @@
...
+      jacobian_difference_sum.get(i_hat) = 0.0;
+      continue;
+    }
+    jacobian_difference_sum.get(i_hat) /=
+        (numerical_difference_sum.get(i_hat) + eps_to_avoid_div_by_zero);
```
nilsvu: Use `equal_within_roundoff` to divide only if both numerator and denominator are nonzero.

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -0,0 +1,79 @@
...
+void jacobian_diagnostic(
+    const gsl::not_null<Scalar<DataVector>*> jacobian_diagnostic,
+    const Jacobian<DataVector, Dim, typename Frame::ElementLogical, Fr>&
+        analytic_jacobian,
+    const TensorMetafunctions::prepend_spatial_index<
+        tnsr::I<DataVector, Dim, Fr>, Dim, UpLo::Lo,
+        typename Frame::ElementLogical>& numeric_jacobian_transpose) {
+  auto jacobian_difference_sum =
+      make_with_value<tnsr::i<DataVector, Dim, typename Frame::ElementLogical>>(
+          get<0, 0>(analytic_jacobian), 0.0);
```
nilsvu: Minor suggestions: I was confused these are named `difference` but no difference is computed afaik. Maybe name them `analytic_jacobian_sum` and `numeric_jacobian_sum`? You can also simplify this code like this if you want:

```cpp
const size_t num_points = analytic_jacobian.begin()->size();
tnsr::i<DataVector, Dim, Frame::ElementLogical> analytic_jacobian_sum{num_points, 0.};
```

(Haven't tested this, but I think `Tensor` has a constructor that's forwarded to all components. You only need `make_with_value` if you're working generically with both DataVector and double)

nilsvu (changes requested):

src/Domain/JacobianDiagnostic.hpp
```diff
@@ -0,0 +1,99 @@
...
+ *
+ * \note This function accepts the transpose of the numeric Jacobian as a
+ * parameter, since the numeric Jacobian will typically be computed via
+ * logical_partial_derivative(), which prepends the logical (source frame)
+ * derivative index. Tensors of type Jacobian, in contrast, have the derivative
+ * index second.
+ */
+template <size_t Dim, typename Fr>
+void jacobian_diagnostic(
+    const gsl::not_null<Scalar<DataVector>*> jacobian_diagnostic,
```
nilsvu: Let's return a vector here to look at the logical directions independently

src/Domain/JacobianDiagnostic.hpp
geoffrey4444: I'd like to leave this as-is so that the free function only takes in tensors, making it easy to test

nilsvu: This won't work great with the python bindings because we have to compute the logical derivative outside the function. If the function only takes jacobian, coordinates, and mesh then it works with the python bindings out of the box. Suggestion: keep the current function that takes only tensors, name it `jacobian_diagnostic_impl`, and test it the way you currently do. Move the function that's currently in the compute item to a free function and name it `jacobian_diagnostic`, so we can call it easily from python. wdyt?

geoffrey4444:
> This won't work great with the python bindings because we have to compute the logical derivative outside the function. If the function only takes jacobian, coordinates, and mesh then it works with the python bindings out of the box. Suggestion: keep the current function that takes only tensors, name it `jacobian_diagnostic_impl`, and test it the way you currently do. Move the function that's currently in the compute item to a free function and name it `jacobian_diagnostic`, so we can call it easily from python. wdyt? 

I think that could work, but what if we name both `jacobian_diagnostic`...since they take different parameters, why not use the same name?

nilsvu:
sure that sounds fine

src/Domain/JacobianDiagnostic.hpp
geoffrey4444: As we discussed below, I'll do this except name both `jacobian_diagnostic`

src/Domain/JacobianDiagnostic.cpp
geoffrey4444: This check is no longer necessary, because instead I changed this so instead of summing the components, it does sqrt(sum(square))) of the components, so that you don't get zero unless all components of the jacobian are zero. So I no longer do this check at all.

src/Domain/JacobianDiagnostic.cpp
geoffrey4444: This check is no longer necessary, because instead I changed this so instead of summing the components, it does sqrt(sum(square))) of the components, so that you don't get zero unless all components of the jacobian are zero. So I no longer do this check at all.

geoffrey4444: I don't want to ignore a term if the denominator is zero; instead, I want to get a huge number (so that I can see in the diagnostic that something is really wrong). So I think I should leave the epsilon here.

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -0,0 +1,105 @@
...
+template <size_t Dim, typename Fr>
+void jacobian_diagnostic(
+    const gsl::not_null<
+        tnsr::i<DataVector, Dim, typename Frame::ElementLogical>*>
+        jacobian_diag,
+    const ::InverseJacobian<DataVector, Dim, Frame::ElementLogical, Fr>&
+        analytic_inverse_jacobian,
+    const tnsr::I<DataVector, Dim, Fr>& mapped_coords,
+    const ::Mesh<Dim>& mesh) {
+  const auto analytic_jacobian =
```
nilsdeppe: Why not just take the jacobian as an argument instead of the inverse jacobian?

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -0,0 +1,105 @@
...
+  auto analytic_jacobian_sum =
+      tnsr::i<DataVector, Dim, typename Frame::ElementLogical>{number_of_points,
+                                                               0.0};
+  auto numerical_jacobian_sum =
+      tnsr::i<DataVector, Dim, typename Frame::ElementLogical>{number_of_points,
+                                                               0.0};
+  constexpr double eps_to_avoid_div_by_zero = 1.e-20;
+
+  // i_hat = logical frame index
+  // i = mapped frame index
```
nilsdeppe, line 31: You can do this with 1 `Scalar<DataVector>` as temporary storage and just use `abs` instead of RMS.

nilsdeppe (changes requested):

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -15,48 +15,64 @@
...
         typename Frame::ElementLogical>& numeric_jacobian_transpose) {
-  auto jacobian_difference_sum =
-      make_with_value<tnsr::i<DataVector, Dim, typename Frame::ElementLogical>>(
-          get<0, 0>(analytic_jacobian), 0.0);
-  auto numerical_difference_sum =
-      make_with_value<tnsr::i<DataVector, Dim, typename Frame::ElementLogical>>(
-          get<0, 0>(analytic_jacobian), 0.0);
+  destructive_resize_components(jacobian_diag,
+                                analytic_jacobian.begin()->size());
+  const size_t number_of_points = analytic_jacobian.begin()->size();
```
nilsvu, line 27: (nitpick, optional) move up one line and use in `destructive_resize`

src/Domain/JacobianDiagnostic.hpp
```diff
@@ -69,31 +106,20 @@ struct JacobianDiagnostic : db::SimpleTag {
...
-    // So this is actually the transpose of the numerical jacobian.
-    const auto numerical_jacobian_transpose =
-        logical_partial_derivative(mapped_coords, mesh);
-
-    ::domain::jacobian_diagnostic(jacobian_diag, analytic_jacobian,
-                                  numerical_jacobian_transpose);
-  }
+  using argument_tags = tmpl::list<
+      ::domain::Tags::InverseJacobian<Dim, Frame::ElementLogical, TargetFrame>,
+      MappedCoordsTag, Mesh<Dim>>;
```
nilsvu: I think you can delete `MappedCoordsTag` and instead write `domain::Tags::Coordinates<Dim, TargetFrame>`.

Please update the docs of struct struct as well.

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -15,48 +15,64 @@
...
-      jacobian_difference_sum.get(i_hat) = 0.0;
-      continue;
-    }
-    jacobian_difference_sum.get(i_hat) /=
-        (numerical_difference_sum.get(i_hat) + eps_to_avoid_div_by_zero);
-    jacobian_difference_sum.get(i_hat) += 1.0;
+    analytic_jacobian_sum.get(i_hat) =
+        sqrt(analytic_jacobian_sum.get(i_hat)) * -1.0;
+    analytic_jacobian_sum.get(i_hat) /=
+        (sqrt(numerical_jacobian_sum.get(i_hat)) + eps_to_avoid_div_by_zero);
```
nilsvu: Maybe instead of the epsilon thing you just set the quantity to NaN (or `quiet_NaN`) if the denominator is zero? That should propagate fine to ParaView.

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -15,48 +15,64 @@
...
-      jacobian_difference_sum.get(i_hat) -= analytic_jacobian.get(i, i_hat);
-      numerical_difference_sum.get(i_hat) +=
-          numeric_jacobian_transpose.get(i_hat, i);
+      analytic_jacobian_sum.get(i_hat) +=
+          square(analytic_jacobian.get(i, i_hat));
+      numerical_jacobian_sum.get(i_hat) +=
+          square(numeric_jacobian_transpose.get(i_hat, i));
     }
   }
   for (size_t i_hat = 0; i_hat < Dim; ++i_hat) {
```
nilsvu: I think you can delete these two lines

src/Domain/JacobianDiagnostic.hpp
```diff
@@ -47,20 +47,57 @@ namespace domain {
...
+
 namespace Tags {
 /// \ingroup DataBoxTagsGroup
 /// \ingroup ComputationalDomainGroup
 /// \brief A diagonstic comparing the analytic and numerical Jacobians for a
 /// map. See `domain::jacobian_diagnostic` for details.
+template <size_t Dim>
 struct JacobianDiagnostic : db::SimpleTag {
-  using type = Scalar<DataVector>;
+  using type = tnsr::i<DataVector, Dim, typename Frame::ElementLogical>;
```
nilsvu, line 85: delete `typename`

tests/Unit/Domain/Python/Test_JacobianDiagnostic.py
```diff
@@ -6,14 +6,10 @@
...
 def jacobian_diagnostic(analytic_jacobian, numeric_jacobian_transpose):
-    analytic_sum = analytic_jacobian.sum(axis=0)
-    numeric_sum = numeric_jacobian_transpose.sum(axis=1)
-    for i, x in enumerate(analytic_sum):
-        if abs(x) < 1.e-10 and abs(numeric_sum[i]) < 1.e-10:
-            analytic_sum[i] = 0.0
-            numeric_sum[i] = 0.0
-    jac_diff_sum = 1.0 - analytic_sum / (numeric_sum + 1.e-20)
-    jac_diag = np.sqrt(np.dot(jac_diff_sum, jac_diff_sum))
+    analytic_sum = (analytic_jacobian * analytic_jacobian).sum(axis=0)
```
nilsvu: `analytic_jacobian**2`

nilsvu (changes requested):
Please squash

nilsvu:
@geoffrey4444 would you be willing to add a pybinding for this as well? Can also be a follow-up PR.

geoffrey4444:
> @geoffrey4444 would you be willing to add a pybinding for this as well? Can also be a follow-up PR.

I do want to add pybindings, but in a follow-up PR

src/Domain/JacobianDiagnostic.cpp
geoffrey4444: I prefer leaving the epsilon as-is

nilsdeppe: You can't get a zero denominator anymore. That was only because the different terms could have canceling signs, but with absolute/RMS the denominator has to be positive, otherwise the Jacobian isn't invertible and the map is definitely bad.

tests/Unit/Domain/Python/Test_JacobianDiagnostic.py
geoffrey4444: I'm no longer using RMS, but I'll keep in mind that **2 is ok to use in future python tests

src/Domain/JacobianDiagnostic.cpp
nilsvu: so, delete the eps?

geoffrey4444:
Thanks for the reviews @nilsvu and @nilsdeppe ! I responded stop your remaining comments and squashed...this is ready for you to have another look when you get a chance.

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -0,0 +1,94 @@
...
+        tnsr::I<DataVector, Dim, Fr>, Dim, UpLo::Lo,
+        typename Frame::ElementLogical>& numeric_jacobian_transpose) {
+  const size_t number_of_points = analytic_jacobian.begin()->size();
+  destructive_resize_components(jacobian_diag, number_of_points);
+  Scalar<DataVector> jacobian_component_sum{number_of_points, 0.0};
+
+  // i_hat = logical frame index
+  // i = mapped frame index
+  for (size_t i_hat = 0; i_hat < Dim; ++i_hat) {
+    for (size_t i = 0; i < Dim; ++i) {
```
nilsdeppe: I think since will almost certainly use this we want to optimize a bit further:
```
jacobian_diag->get(i_hat) = -abs(abs(analytic_jacobian.get(0, i_hat));
for (size_t i = 1; ...) { ... }
get(jacobian_component_sum) =  abs(numeric_jacobian_transpose.get(i_hat, 0));
for (size_t i = 1; ...) { ... }
```
and then we can also eliminate the `0.0` on line 28, on 37  and on 44.

nilsdeppe (commented):
A small pedantic performance consideration since I think we'll use this in AMR and so we will want this to be quite fast

Edit: you can squash immediately!

nilsvu (dismissed):
don't wait for my re-approval

src/Domain/JacobianDiagnostic.cpp
geoffrey4444: Done, squashed, and force-pushed...good suggestion!

src/Domain/JacobianDiagnostic.cpp
```diff
@@ -0,0 +1,93 @@
...
+  destructive_resize_components(jacobian_diag, number_of_points);
+
+  // i_hat = logical frame index
+  // i = mapped frame index
+  for (size_t i_hat = 0; i_hat < Dim; ++i_hat) {
+    jacobian_diag->get(i_hat) = -abs(analytic_jacobian.get(0, i_hat));
+    for (size_t i = 1; i < Dim; ++i) {
+      jacobian_diag->get(i_hat) -= abs(analytic_jacobian.get(i, i_hat));
+    }
+    Scalar<DataVector> jacobian_component_sum{number_of_points};
```
nilsdeppe: Uh oh, memory allocation in a loop! Can you go back to allocating outside the loop. Just move this line to line 30, nothing else needs to be changed!

geoffrey4444: 🤦‍♂️that's what I get for trying to get it in 5 minutes before lunch! "O, let's declare this when we need it!", I thought.

Anyway, thanks for catching, fixed!

p.s. I saw some CI failures, but they looked not related to my PR but to python

nilsdeppe (dismissed):

nilsdeppe:
You have missing linker dependencies. I think you need `Domain` to depend on `Spectral` as a `PUBLIC` instead of `INTERFACE` library. I.e. move `Spectral` from `INTERFACE` to `PUBLIC` in `src/Domain/CMakeLists.txt`

src/Domain/JacobianDiagnostic.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+#pragma once
+
+#include <cstddef>
+#include <string>
+
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Domain/Tags.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
```
nilsvu: <s>No it's not `Spectral`.</s> Move this include to the cpp, then depend on `PRIVATE LinearOperators`. Edit: Actually yes, depend on `PUBLIC Spectral` also and include Mesh.hpp here`.

Also delete the DeterminantAndInverse include.

src/Domain/JacobianDiagnostic.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+/// \endcond
+
+namespace domain {
+/*!
+ * \ingroup ComputationalDomainGroup
+ * \brief A diagonstic comparing the analytic and numerical Jacobians for a
```
nilsvu, line 22: Typo "diagnostic", but you can fix it some other time

nilsvu (approved):

PR #4646, opened 2023-01-25
Support derivatives and integrals in ApplyPointwise, rename to TransformVolumeData

### LINKED ISSUE #4587: Python bindings to enable domain diagnostics post processing
# Feature request:

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [x] Add ability to compute domain diagnostics in post processing (e.g., by calling C++ functions) with python bindings

nilsvu:
## Proposed changes

Apply python functions elementwise to all volume data in H5 files, including taking derivatives and volume integrals.

### Upgrade instructions

`spectre apply-pointwise` is now `spectre transform-volume-data` (or `spectre transform-vol`).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/Tensor/Python/__init__.py
+++ b/src/DataStructures/Tensor/Python/__init__.py
@@ -57,3 +57,10 @@ class JacobianMeta(TensorMeta):
 Scalar = {DataVector: ScalarDV, float: ScalarD}
 Jacobian = JacobianMeta(inverse=False)
 InverseJacobian = JacobianMeta(inverse=True)
+
+# Define a type annotation that means "any tensor". This should really be a
+# common superclass of all Tensor types, but we currently don't have that in C++
+# so we use a type alias to `typing.Any` as a workaround.
+from typing import Any
+
+Tensor = Any
--- a/src/Visualization/Python/ApplyPointwise.py
+++ b/src/Visualization/Python/ApplyPointwise.py
@@ -443,7 +443,6 @@ def parse_kernels(kernels, exec_files, map_input_names):
               "-k",
               "kernels",
               multiple=True,
-              required=True,
               help=("Python function(s) to apply to the volume data. "
                     "Specify as 'path.to.module.function_name', where the "
                     "module must be available to import. "
@@ -542,6 +541,8 @@ def apply_pointwise_command(h5files, subfile_name, kernels, exec_files,
     volfiles = [h5file.get_vol(subfile_name) for h5file in open_h5_files]
 
     # Load kernels
+    if not kernels:
+        raise click.UsageError("No '--kernel' / '-k' specified.")
     kernels = list(parse_kernels(kernels, exec_files, map_input_names))
 
     # Apply!
--- a/src/Visualization/Python/ApplyPointwise.py
+++ b/src/Visualization/Python/ApplyPointwise.py
@@ -16,6 +16,7 @@ from spectre.DataStructures import DataVector
 from spectre.DataStructures.Tensor import (Frame, InverseJacobian, Jacobian,
                                            Tensor, tnsr)
 from spectre.IO.H5.IterElements import Element, iter_elements
+from spectre.NumericalAlgorithms.LinearOperators import definite_integral
 from spectre.Spectral import Mesh
 
 # functools.cached_property was added in Py 3.8. Fall back to a plain
@@ -271,9 +272,10 @@ class Kernel:
         return output
 
 
-def apply_pointwise(volfiles: Union[spectre_h5.H5Vol,
-                                    Iterable[spectre_h5.H5Vol]],
-                    kernels: Sequence[Kernel]):
+def apply_pointwise(
+        volfiles: Union[spectre_h5.H5Vol, Iterable[spectre_h5.H5Vol]],
+        kernels: Sequence[Kernel],
+        integrate: bool = False) -> Union[None, Dict[str, Sequence[float]]]:
     """Apply pointwise functions to data in the 'volfiles'
 
     Arguments:
@@ -283,6 +285,16 @@ def apply_pointwise(volfiles: Union[spectre_h5.H5Vol,
         files will be transformed.
       kernels: List of pointwise transformations to apply to the volume data
         in the form of 'Kernel' objects.
+      integrate: Compute the volume integral over the kernels instead of
+        writing them back into the volume files. The integral is computed in
+        inertial coordinates for every tensor component of all kernels and over
+        all observations in the volume files. For example, if a kernel returns a
+        vector named 'Shift' then this function returns integrals named
+        'Shift(_x,_y,_z)'. In addition, the corresponding observation values are
+        returned (named 'Time'), and the inertial volume (named 'Volume').
+
+    Returns:
+      None, or the volume integrals if 'integrate' is True.
     """
     # Collect all tensor components that we need to apply the kernels
     all_tensors: Dict[str, TensorArg] = {}
@@ -302,14 +314,27 @@ def apply_pointwise(volfiles: Union[spectre_h5.H5Vol,
             else:
                 all_tensors[tensor_name] = tensor_arg
     logger.debug("Input datasets: " + str(list(all_tensors.keys())))
-    logger.info("Output datasets: " +
-                str([kernel.output_name for kernel in kernels]))
+    # Logging output datasets as INFO when they are written back into the volume
+    # files so the user can find them
+    logger.log(
+        logging.DEBUG if integrate else logging.INFO,
+        "Output datasets: " + str([kernel.output_name for kernel in kernels]))
+
+    # We collect integrals in this dict and return it
+    if integrate:
+        integrals: Dict[str, Sequence[float]] = {}
 
     if isinstance(volfiles, spectre_h5.H5Vol):
         volfiles = [volfiles]
     for volfile in volfiles:
         all_observation_ids = volfile.list_observation_ids()
         num_obs = len(all_observation_ids)
+        if integrate and "Time" not in integrals:
+            integrals["Time"] = [
+                volfile.get_observation_value(obs_id)
+                for obs_id in all_observation_ids
+            ]
+
         for i_obs, obs_id in enumerate(all_observation_ids):
             # Load tensor data for all kernels
             all_tensor_data = {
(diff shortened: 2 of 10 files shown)


nilsvu:
@geoffrey4444 FYI, here's the PR that will allow us to compute power monitors, Jacobian diagnostics, etc over volume data. It works like this:

```sh
spectre transform-vol H5FILES -d element_data -k spectre.domain.jacobian_diagnostic [--integrate]
```

The result of the function is written back into the H5 files and can be visualized in ParaView.

wthrowe:
Should this utility be renamed, since it does non-pointwise things now?

nilsvu:
@wthrowe If you're referring to derivatives: we do some derivatives in `src/PointwiseFunctions/` and I think that's ok. Derivatives are pointwise in the continuum limit, right :) If you're referring to integrals: that's more of a side product of this function, and you can read it as "apply this pointwise function, then also integrate it up", so seems fine to me too. Sounds ok?

src/DataStructures/Tensor/Python/__init__.py
```diff
@@ -55,3 +55,10 @@ def __init__(self, inverse: bool):
 Scalar = {DataVector: ScalarDV, float: ScalarD}
 Jacobian = JacobianMeta(inverse=False)
 InverseJacobian = JacobianMeta(inverse=True)
+
+# Define a type annotation that means "any tensor". This should really be a
```
nilsdeppe, line 61: We do sort of have this in C++: TensorAsExpression. I'm not sure that's what you want here, but might be good to add why note in the comment.

src/Visualization/Python/ApplyPointwise.py
```diff
@@ -463,9 +530,45 @@ def shift_magnitude(
...
+                                    integrate=integrate,
+                                    **kwargs)
         progress.update(task_id, completed=len(volfiles))
 
+    # Write integrals to output file or print to terminal
+    if integrate:
+        integral_values = np.stack(list(integrals.values())).T
+        integral_names = list(integrals.keys())
+        if output:
+            if output.endswith(".txt"):
```
nilsdeppe: or `.dat` or `.tsv`?

src/Visualization/Python/ApplyPointwise.py
```diff
@@ -244,6 +246,11 @@ def apply_pointwise(volfiles: Union[spectre_h5.H5Vol,
         files will be transformed.
       kernels: List of pointwise transformations to apply to the volume data
         in the form of 'Kernel' objects.
+      integrate: Compute the volume integral over the kernels instead of
+        writing them back into the volume files.
```
nilsdeppe: Can you explicitly state what needs to be returned? I don't understand what happens if I return `\gamma_{ij}`. Do I get `\int d\xi J \gamma_{ij}` or `d\xi \gamma_{ij}`? Or can I only return scalars? Do I need to return the tensors multiplied by the Jacobian determinant?

src/Visualization/Python/ApplyPointwise.py
```diff
@@ -62,11 +76,116 @@ def get_tensor_component_names(tensor_name: str,
...
+    - Mesh: Annotate the argument with a 'Mesh' type.
+    - Coordinates: Annotate the argument with a 'tnsr.I' type and name it
+        "logical_coords" / "logical_coordinates" / "xi" or "inertial_coords" /
+        "inertial_coordinates" / "x".
+    - Jacobians: Annotate the argument with a 'Jacobian' or 'InverseJacobian'.
+    - Any tensor dataset: Annotate the argument with the tensor type. By
+        default, the argument name is transformed to CamelCase to determine the
+        dataset name in the volume data file. Specify a mapping in
+        'map_input_names' to override the default.
+
```
nilsdeppe: Can you give an example of how one would for example compute the 1-index constraint in scalar wave? There's a lot to unpack here, and I think having an example would make that much easier to unpack.

How does this interact with C++ functions wrapped with pybind? Can you document that too? I'm not clear how I could call a function that takes `psi, mesh, inv_jac` and compute the 1-index constraint locally if I have a C++ bound function.

nilsdeppe (changes requested):
You can squash. Mostly dox because I don't fully understand how I should be writing my python functions

wthrowe (changes requested):
No.  Let's name this correctly before it has lots of uses and we're stuck with it.

Derivatives are not pointwise, not even in some limit.  From what I can find in the code, the only person who has added derivatives to `PointwiseFunctions` is you.  The fact that I missed your error the first time doesn't mean I have to keep letting you do it.

I agree that integrals are a different sort of thing entirely, so I drop any objections I had related to them.

(Haven't reviewed the code itself.)

src/DataStructures/Tensor/Python/__init__.py
nilsvu: I need a common superclass of all Tensor types. `TensorAsExpression` isn't a superclass of `Tensor`, and also it has template parameters (I don't really understand its inheritance pattern).

src/Visualization/Python/ApplyPointwise.py
nilsvu: I just removed this check. Now the output is an H5 file if the file extension is `.h5`, and a text file otherwise.

src/Visualization/Python/ApplyPointwise.py
nilsvu: When you return `\gamma_{ij}` from the kernel then the integral that's computed is `\int d\xi J \gamma_{ij}`. That's one integral per component in inertial coords. Basically you write a kernel that returns a tensor and then decide "let's integrate it over the volume". I added some more docs.

src/Visualization/Python/ApplyPointwise.py
nilsvu: Added some docs. To call a binding just pass something like `-k spectre.PointwiseFunctions.ScalarWave.one_index_constraint` to the CLI. Not sure how well that works for bindings that have multiple overloads, we'll see. Once a couple of suitable bindings are merged, like #4644 or #4637, I can add a test.

nilsvu:
@wthrowe ok let's talk about pointwise naming, maybe next Monday after the spectre call? Can we avoid holding up this PR with that?

wthrowe:
Three business days is not an unreasonably long code review.

src/Visualization/Python/ApplyPointwise.py
nilsdeppe: Okay, sounds good!

nilsvu:
No the time is fine but if we want to rename the script I'd like to do it in a separate PR. I understand the change in this PR is related to the name, but renaming is easier in a separate PR.

nilsvu:
Pushed three more commits extending support for pybind11 overloads and DataVector arguments (I'll add tests for the pybind11 overload parsing once we have some suitable bindings merged, like the jacobian diagnostic).

nilsdeppe (commented):
I have no requested changes on the new commits.

nilsvu:
Rebased and renamed to `TransformVolumeData`

nilsdeppe (dismissed):

nilsvu:
@nikwit would you have some time to review this too?

wthrowe (dismissed):
I approve the naming changes.  I'm not good enough with python to really review the rest.

src/Visualization/Python/ApplyPointwise.py
```diff
@@ -133,52 +279,60 @@ def apply_pointwise(volfiles: Union[spectre_h5.H5Vol,
...
+                # elements, and reassemble the result into a contiguous dataset
+                if kernel.elementwise:
+                    transformed_tensor_data = np.zeros(
+                        (kernel.output_type.size, total_num_points))
+                    for element in iter_elements(volfile, obs_id):
+                        transformed_tensor = kernel(all_tensor_data, element)
+                        for i, component in enumerate(transformed_tensor):
+                            transformed_tensor_data[
+                                i, element.data_slice] = component
+                    transformed_tensor = kernel.output_type(
```
nikwit: here and other places this will do a copy even though this could be avoided for a lot of types, no?

src/Visualization/Python/ApplyPointwise.py
```diff
@@ -63,49 +76,176 @@ def get_tensor_component_names(tensor_name: str,
...
+      argument.
+    """
+    if arg.annotation is inspect.Parameter.empty:
+        raise ValueError(
+            "Please annotate all arguments in the kernel function. "
+            f"Argument '{arg.name}' has no annotation.")
+
+    if arg.annotation in Mesh.values():
+        return ElementArg("mesh")
+    elif (arg.name in ["logical_coords", "logical_coordinates", "xi"]
```
nikwit: I think it would be better if you check what the name is and then assert that it has the right type

nikwit (changes requested):
Overall I like the idea of having this functionality and how most of it is structured. But I have some concerns about the using of typing and argument names for control flow in python. 

1.) I have never seen that the actual name of a function argument or variable in python (or any language) should affect the behavior of a program but here we are using it as keys to the observed quantities. Unless one reads the implementation (which takes a while to understand) it is a bit intransparent what is going on for a user. If we decide to keep it this way it should be pointed out most prominently in the documentation that this is happening.

2.) I don't like the mandatory typing in python. It is not a statically typed language and while typing hints can be a useful tool to make code a bit clearer I think they should be optional and not ever used for control flow ( they are called "hints" after all). I think it might even be possible to completely get rid of tensors on the python side and directly convert everything to appropriately sized `np.array`s. Then dynamic checking of the dims can be done. 

Overall, I would  prefer a simpler and more pythonic solution, such as 
`apply_pointwise(["Psi", "InertialCoordinates"], func_to_apply)` where `func_to_apply` takes `Psi` and `InertialCoordinates` as arguments. 
Is  there a reason why this is not possible?

src/Visualization/Python/ApplyPointwise.py
nilsvu: good call, I added `copy=False` so the Tensor constructor points into the Numpy array (this is supported through the buffer protocol, see bindings in Python/Tensor.cpp)

src/Visualization/Python/ApplyPointwise.py
nilsvu: Also removed the "xi" option because maybe that can be used to mean something other than logical coords (some people name the displacement field in elasticity "xi").

nilsvu:
@nikwit The argument _names_ are used as default dataset names so you don't have to type them all on the command line (e.g. names like "shift", "lapse" etc are standard enough in our bound functions that it would be silly to write `-i shift=Shift` every time you invoke this script). To make this transparent to the user I added logging output that prints the dataset names that are used, so there's no magic going on behind the scenes (or at least, you understand it).

The argument _types_ are used to cast the data into the type that the function expects (mainly pybind11 bindings of C++ functions). I think that's a good use of type hints, considering that we're passing Python objects to C++, which is a statically typed language, and pybind11 actually asserts the argument types are correct when a binding is called. Maybe this can be done entirely by pybind11 though, using #4680. I'll play around with that a bit to see if it's enough to make the type hints optional.

nilsvu:
@nikwit using #4680 I made return type annotations optional (please see last commit). You can now return tensors, DataVectors, or Numpy arrays from the kernels, and they get written to datasets in the volfiles. You can also return a dict from the kernel to specify the dataset names. We still need the argument type annotations so we can cast the dataset to the right tensor type. As I wrote above, I think that's a pretty good use of type annotations.

nikwit (dismissed):

nilsdeppe (approved):

PR #4653, opened 2023-01-27
Add distorted frame to block_logical_coordinates

markscheel:
block_logical_coordinates now understand the distorted frame.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/BlockLogicalCoordinates.cpp
+++ b/src/Domain/BlockLogicalCoordinates.cpp
@@ -60,14 +60,55 @@ std::vector<block_logical_coord_holder<Dim>> block_logical_coordinates(
           } else {
             continue;  // Not in this block
           }
-        } else {  // frame is different than ::Frame::Inertial
+        } else if constexpr (std::is_same_v<Frame, ::Frame::Distorted>) {
+          // Point is in the distorted frame, so we need to map to the grid
+          // frame and then the logical frame.
+          if (not block.has_distorted_frame()) {
+            // Note that block.has_distorted_frame() can be different for
+            // different Blocks.  However, the template parameter Frame is
+            // compile-time and is the same for all Blocks.
+            //
+            // Explanation of the logic here:
+            // 1. Recall that block_logical_coordinates loops through all the
+            //    Blocks, and skips all the Blocks except for the first Block
+            //    it finds that contains the point x.
+            // 2. If Frame is ::Frame::Distorted but
+            //    block.has_distorted_frame() is false, then this block
+            //    cannot contain the point x. Therefore, we should simply
+            //    skip this block.  If it turns out that no blocks contain
+            //    the point x, then we will get an error later.
+            //    (Note that our primary use case for ::Frame::Distorted is to
+            //    find an apparent horizon in the distorted frame. In that
+            //    case, only the Blocks near a horizon have a distorted frame
+            //    because only those Blocks have distortion maps. Thus,
+            //    the Blocks that are skipped here are those that are far
+            //    from horizons).
+            continue; // Not in this block
+          }
+          const auto moving_inv =
+              block.moving_mesh_grid_to_distorted_map().inverse(
+                  x_frame, time, functions_of_time);
+          if (not moving_inv.has_value()) {
+            continue; // Not in this block
+          }
+          // logical to grid map is time-independent.
+          const auto inv = block.moving_mesh_logical_to_grid_map().inverse(
+              moving_inv.value());
+          if (inv.has_value()) {
+            x_logical = inv.value();
+          } else {
+            continue;  // Not in this block
+          }
+        } else {
+          // frame is different than ::Frame::Inertial or ::Frame::Distorted.
           // Currently 'time' is unused in this branch.
           // To make the compiler happy, need to trick it to think that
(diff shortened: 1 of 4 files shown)


knelli2 (dismissed):

src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -60,14 +60,33 @@ std::vector<block_logical_coord_holder<Dim>> block_logical_coordinates(
           } else {
             continue;  // Not in this block
           }
-        } else {  // frame is different than ::Frame::Inertial
+        } else if constexpr (std::is_same_v<Frame, ::Frame::Distorted>) {
+          // Point is in the distorted frame, so we need to map to the grid
+          // frame and then the logical frame.
+          const auto moving_inv =
+              block.moving_mesh_grid_to_distorted_map().inverse(
```
nilsvu, line 89: Please add an assert that `block.has_distorted_frame()`. Also, I think the inertial coord case in the branch above needs to handle the distorted frame, right? Please test this case where you have a distorted frame and want to invert an inertial coord.

src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -151,7 +171,7 @@ std::vector<block_logical_coord_holder<Dim>> block_logical_coordinates(
       const functions_of_time_type& functions_of_time);
 
 GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3),
-                        (::Frame::Inertial, ::Frame::Grid))
+                        (::Frame::Inertial, ::Frame::Grid, ::Frame::Distorted))
```
nilsvu: (Nitpick) let's try to order frames in lists like this "grid, distorted, inertial", i.e. the order in which they are composed

tests/Unit/Domain/Test_BlockAndElementLogicalCoordinates.cpp
```diff
@@ -183,7 +183,8 @@ void fuzzy_test_block_and_element_logical_coordinates_unrefined(
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time = std::unordered_map<
             std::string,
-            std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>{}) {
+            std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>{},
+    const bool has_distorted_frame = false) {
```
nilsvu: You can get this from the blocks

nilsvu (changes requested):

src/Domain/BlockLogicalCoordinates.cpp
markscheel: The inertial coord case above is fine as is; it doesn't matter if there is a distorted frame or not.

However, the new distorted frame case does need to be changed, because the current case will fail if some blocks don't have distorted frames.

nilsvu: Update from the call today: Would be good to add some docs to Block about which maps are available.

markscheel: I found a better way to treat this case.  See the long comment in BlockLogicalCoordinates.cpp

markscheel:
CodeCov will likely complain about line 86.  There is currently no (meaningful) way of testing this line, because there is currently no way of creating a Domain where some (but not all) blocks have a distorted frame.   For the sake of codecov, I could make a new test where I make a full Domain without a distorted frame and test that block_logical_coords returns std::nullopt when called on a distorted-frame point; is doing so worth it?

src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -63,11 +63,33 @@ std::vector<block_logical_coord_holder<Dim>> block_logical_coordinates(
...
+            //    block.has_distorted_frame() is false, then this block
+            //    cannot contain the point x. Therefore, we should simply
+            //    skip this block.  If it turns out that no blocks contain
+            //    the point x, then we will get an error later.
+            //    (Note that our primary use case for ::Frame::Distorted is to
+            //    find an apparent horizon in the distorted frame. In that
+            //    case, only the Blocks near a horizon have a distorted frame
+            //    because only those Blocks have distortion maps. Thus,
+            //    the Blocks that are skipped here are those that are far
+            //    from horizons).
```
nilsvu, line 85: This means that you can't use `block_logical_coordinates` in its current form to resolve `Distorted` -> `Logical` over the full grid. Calling `block_logical_coordinates` with distorted-frame coordinates will only return blocks that have a distorted frame. That's probably fine, but please document this clearly in the dox for this function.

nilsvu (changes requested):
please squash

kidder:
I don't think you should go to extreme effort to satisfy Codecov.  If it can be tested when we do have a DomainCreator with some blocks without a distorted frame, then open an issue to address it then.

markscheel:
Add warning about distorted frame in the dox, and squashed.

nilsvu (approved):

knelli2 (approved):

PR #4654, opened 2023-01-27
Add ScalarWaveWorldtube library

nikwit:
This adds the worldtube library. I think it is best to add it directly to `CurvedScalarWave` since it only works with this system.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/CMakeLists.txt
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/CMakeLists.txt
@@ -1,0 +1,26 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+set(LIBRARY ScalarWaveWorldtube)
+
+add_spectre_library(${LIBRARY})
+
+spectre_target_sources(
+  ${LIBRARY}
+  PRIVATE
+  Tags.cpp
+  )
+
+spectre_target_headers(
+  ${LIBRARY}
+  INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/src
+  HEADERS
+  Tags.hpp
+  Worldtube.hpp
+  )
+
+target_link_libraries(
+  ${LIBRARY}
+  PUBLIC
+  Options
+  )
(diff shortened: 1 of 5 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -0,0 +1,14 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "Options/Options.hpp"
+
+namespace CurvedScalarWave::Worldtube {
```
nilsvu, line 8: Add dox to this namespace please. For more extensive docs you can add a `Worldtube.hpp` file that only documents the namespaces.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -0,0 +1,14 @@
...
+namespace OptionTags {
```
nilsvu, line 12: Add dox here too please so the tags show up in Doxygen. Can be just one line.

nilsvu (changes requested):

nikwit:
@nilsvu thanks. I added some initial docs and will add more specific docs once I merge the functionalities so I can reference them. I squashed directly since there was not much left of the original commit.

nilsvu (dismissed):
Nice description of the scheme :)

src/Evolution/Systems/CurvedScalarWave/Worldtube/CMakeLists.txt
```diff
@@ -0,0 +1,26 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+set(LIBRARY Worldtube)
```
nilsvu: Oh maybe name this lib something like `CswWorldtube` or `ScalarWaveWorldtube`, because you want another GH worldtube lib later?

nilsdeppe: ^ agreed, and to distinguish it from CCE more, since CCE also uses worldtubes!

nilsvu (approved):

PR #3142, opened 2021-04-30
Allow reference tags to list multiple arguments

nilsvu:
## Proposed changes

Remove `parent_tag`s and use `argument_tag`s instead. This removes the restriction that reference tags can only have a single argument. My use case for this is to store an elastic constitutive relation _per block_ in the global cache and reference the correct one in the element's DataBox, which needs both the global cache tag and the element ID.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `major new feature` if appropriate.

PR #4655, opened 2023-01-27
Add worldtube excision sphere tag

nikwit:
adds a tag that returns the radius of the worldtube

~~depends on #4654~~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
@@ -3,6 +3,16 @@
 
 #pragma once
 
+#include <cstddef>
+#include <memory>
+#include <string>
+#include <unordered_map>
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "Domain/Creators/DomainCreator.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/OptionTags.hpp"
+#include "Domain/Structure/ExcisionSphere.hpp"
 #include "Options/Options.hpp"
 
 namespace CurvedScalarWave::Worldtube {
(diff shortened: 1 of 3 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -3,12 +3,54 @@
...
+  static constexpr Options::String help{
+      "The name of the excision sphere as returned by the domain."};
+  using group = Worldtube;
+};
+
 }  // namespace OptionTags
+
+namespace Tags {
+template <size_t Dim>
+struct WorldtubeRadius : db::SimpleTag {
```
nilsvu: The excision sphere essentially _is_ your worldtube, right? So how about you name this `Worldtube` and make it hold the excision sphere. Then you can get the radius, center, etc from it whenever you want. Right now you can just copy the excision sphere from the domain, but we should get #3142 merged so you can make this a reference tag.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -3,12 +3,54 @@
...
+  using type = double;
+  using option_tags = tmpl::list<domain::OptionTags::DomainCreator<Dim>,
+                                 OptionTags::ExcisionSphere>;
+  static constexpr bool pass_metavariables = false;
+  static double create_from_options(
+      const std::unique_ptr<::DomainCreator<Dim>>& domain_creator,
```
nilsvu, line 55: I think the argument type should be `const DomainCreator<Dim>&`. This is easy to test with `TestHelpers::test_option_tag_factory_creation`. Please use that in your test.

nilsvu (changes requested):
Please add dox to things so they show up in Doxygen

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: no, I think it's a pointer. I don't understand, `TestHelpers::test_option_tag_factory_creation` tests option tags with the type of a base class pointer which I do not have here.

nilsvu: You're right. I don't remember if we have a quick way to test an option-created tag like this. The way you did seems fine.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -0,0 +1,70 @@
...
+ */
+namespace OptionTags {
+/*!
+ * \brief Options for the worldtube
+ */
+struct Worldtube {
+  static constexpr Options::String help = {"Options for the Worldtube"};
+};
+/*!
+ * \brief Name of the excision sphere desginated to act as a worldtube
```
nilsvu: Typo "designated"

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -0,0 +1,70 @@
...
+
+/*!
+ * \brief Tags related to the worldtube
+ */
+namespace Tags {
+/*!
+ * \brief The excision sphere corresponding to the worldtube
+ */
+template <size_t Dim>
+struct ExcisionSphere : db::SimpleTag {
```
nilsvu, line 49: Do you prefer this name over `Worldtube`? Up to you

nilsvu (dismissed):
You can fix the typo when rebasing

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: yes, I think so. The worldtube has a lot more properties than the `ExcisionSphere` which are mostly in the databox of the Singleton which I will add.

nilsvu (approved):

PR #4656, opened 2023-01-27
Add 3D BBH visualization script

AlexCarpenter46:
## Proposed changes

This script can be used with BBH runs to make quick visualizations. The script is formatted to be very similar to the generate-xdmf script. Requires SpatialRicciScalar and Lapse to be output in the volume data to function.

To run the script `spectre render-3d bbh volume_xmf_file -o Binary_Pic.png -a aha_xmf -b ahb_xmf -c Wide --show-time`

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

This PR isn't ready to be merged I don't think, I'm only starting this to get more discussion around how we want this to work and to get some help with how to make a test for this script.

### DIFF
--- a/src/Visualization/Python/Render3D/Bbh.py
+++ b/src/Visualization/Python/Render3D/Bbh.py
@@ -1,0 +1,291 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+import os
+
+import click
+import rich.columns
+
+logger = logging.getLogger(__name__)
+
+
+def _parse_step(ctx, param, value):
+    if value is None:
+        return None
+    if value.lower() == "first":
+        return 0
+    if value.lower() == "last":
+        return -1
+    return int(value)
+
+
+def ah_vis(ah_xmf: str, render_view: str):
+    """Helper function for visualizing the apparent horizons of the objects.
+
+    Arguments:
+    ah_xmf: Path to the xmf file of the object.
+    render_view: The current view in paraview to add the horizon to."""
+    import paraview.simple as pv
+
+    Ah_xmf = pv.XDMFReader(registrationName=ah_xmf, FileNames=[ah_xmf])
+    transform_1 = pv.Transform(registrationName="Transform1", Input=Ah_xmf)
+    transform_1.Transform = "Transform"
+    transform_1.Transform.Translate = [0.0, 0.0, 2.0]
+    transform_1_display = pv.Show(
+        transform_1, render_view, "UnstructuredGridRepresentation"
+    )
+    transform_1_display.SetScalarBarVisibility(render_view, False)
+    # Sets apparent horizon color to black
+    transform_1_display.AmbientColor = [0.0, 0.0, 0.0]
+    transform_1_display.DiffuseColor = [0.0, 0.0, 0.0]
+
+    render_view.Update()
+    pv.ColorBy(transform_1_display, None)
+
+
+def render_bbh(
+    volume_xmf: str,
+    output: str,
+    aha_xmf: str,
+    ahb_xmf: str,
+    time_step: int = 0,
+    animate: bool = False,
+    camera_angle: str = "Side",
+    color_map: str = "Rainbow Uniform",
+    show_grid: bool = False,
+    show_time: bool = False,
+):
+    """Generate Pictures from XMF files for BBH Visualizations
+
+    Generates pictures from BBH runs using the XMF files generated using
+    generate-xdmf. This script requires that the Lapse and SpatialRicciScalar
+    were output in the volume data.
+
+    Arguments:
+
+      volume_xmf: Path to the volume data xmf file.
+      output: Name of output file generated from paraview. Include extensions
+      such as '.png'
+      aha_xmf: Path to the apparent horizon xmf file for object A.
+      ahb_xmf: Path to the apparent horizon xmf file for object B.
+      camera_angle: Specified camera angle, defaults to Side if empty. Other
+      possible angles Top and Wide
+      color_map: Color map for the lapse, defaults to 'Rainbow Uniform'. Other
+      color maps include 'Inferno (matplotlib)', 'Viridis (matplotlib)', etc.
+      show_grid: Shows the grid lines of the domain.
+      show_time: Shows the simulation time.
+
+    To splice all the pictures into a video, try using FFmpeg"""
+    import paraview.simple as pv
+
+    version = pv.GetParaViewVersion()
+    if version < (5, 11) or version > (5, 11):
+        logger.warning(
+            "WARNING: Your Paraview version is not 5.11, "
+            "the script may not work correctly."
+        )
+
+    # Volume Data Visualization
+    volume_files_xmf = pv.XDMFReader(
+        registrationName=volume_xmf, FileNames=[volume_xmf]
+    )
+
+    # Check for Lapse and SpatialRicciScalar
+    variables = list(volume_files_xmf.PointData.keys())
+    assert (
+        "Lapse" in variables
+    ), "Lapse not found in volume data, the script will not work correctly."
+    assert "SpatialRicciScalar" in variables, (
+        "SpatialRicciScalar not found in volume data, the script will not work"
+        " correctly."
+    )
+
+    render_view = pv.GetActiveViewOrCreate("RenderView")
+
+    # Color the grid
+    color_transfer_function = pv.GetColorTransferFunction("Lapse")
+    color_transfer_function.Discretize = 0
+    color_transfer_function.ApplyPreset(color_map, True)
+    color_transfer_function.InvertTransferFunction()
+
+    # Slice volume data
+    slice = pv.Slice(registrationName="slice", Input=volume_files_xmf)
+    slice.SliceType = "Plane"
+    slice.HyperTreeGridSlicer = "Plane"
+    slice.SliceOffsetValues = [0.0]
+    slice.SliceType.Normal = [0.0, 0.0, 1.0]
+    slice.Triangulatetheslice = 0
+
+    # Warp grid by spatial ricci scalar
+    warp_by_scalar = pv.WarpByScalar(
+        registrationName="WarpByScalar", Input=slice
+    )
+    warp_by_scalar.Scalars = ["POINTS", "SpatialRicciScalar"]
+    warp_by_scalar.ScaleFactor = 2.5
+    warp_by_scalar.Normal = [0.0, 0.0, -1.0]
+    warp_by_scalar_display = pv.Show(
+        warp_by_scalar, render_view, "GeometryRepresentation"
+    )
+    warp_by_scalar_display.SetScalarBarVisibility(render_view, False)
+
+    # Apparent Horizon Visualization
+    if aha_xmf:
+        ah_vis(aha_xmf, render_view)
+    if ahb_xmf:
+        ah_vis(ahb_xmf, render_view)
+
+    if show_grid:
+        warp_by_scalar_display.Representation = "Surface With Edges"
+
+    pv.LoadPalette(paletteName="GradientBackground")
+    render_view.OrientationAxesVisibility = 0
+    pv.SetActiveSource(warp_by_scalar)
+    warp_by_scalar_display.Opacity = 0.8
+    pv.ColorBy(warp_by_scalar_display, ("POINTS", "Lapse"))
+    layout = pv.GetLayout()
+    layout.SetSize(1920, 1080)
+
+    # Camera placements
+    # Top down view
+    if camera_angle == "Top":
+        render_view.CameraPosition = [0.0, 0.0, 36.90869716569761]
+        render_view.CameraFocalPoint = [0.0, 0.0, 0.6894899550131899]
+        render_view.CameraViewUp = [0, 1, 0]
+        render_view.CameraParallelScale = 424.27024700303446
+    # Wide/Inbetween View
+    elif camera_angle == "Wide":
+        render_view.CameraPosition = [
+            -89.0,
+            -17.0,
+            25.0,
+        ]
+        render_view.CameraFocalPoint = [
+            -0.3921962951264054,
+            1.6346750682876983,
+            -0.34522248814953405,
+        ]
+        render_view.CameraViewUp = [
+            0.0,
+            0.0,
+            1.0,
+        ]
+    # Side View
+    else:
+        render_view.CameraPosition = [
+            -29.944619336722987,
+            -3.666072157343372,
+            2.895224044348878,
+        ]
+        render_view.CameraFocalPoint = [
+            -0.13267040638072278,
+            0.6356115665206243,
+            -0.37352608789235847,
+        ]
+        render_view.CameraViewUp = [0.0, 0.0, 1.0]
+        render_view.CameraParallelScale = 519.6152422706632
+
+    # Simulation time
+    if show_time:
+        time_filter = pv.AnnotateTimeFilter(
+            registrationName="annotate_time_filter", Input=slice
+        )
+        time_filter.Format = "Time: {time:0.2f}M"
+        annotate_time_filter_display = pv.Show(
+            time_filter, render_view, "TextSourceRepresentation"
+        )
+        annotate_time_filter_display.FontSize = 45
+
+    # Capture all frames
+    animation_scene = pv.GetAnimationScene()
+    animation_scene.PlayMode = "Snap To TimeSteps"
+
+    # Save animation/screenshot
+    if animate:
+        pv.SaveAnimation(output, render_view)
+    else:
+        render_view.ViewTime = volume_files_xmf.TimestepValues[time_step]
+        pv.Render()
+        pv.SaveScreenshot(output, render_view)
+
+
+@click.command(name="bbh", help=render_bbh.__doc__)
+@click.argument(
+    "volume_xmf",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+)
+@click.option(
+    "--output",
+    "-o",
+    type=click.Path(
+        exists=False, file_okay=True, dir_okay=False, writable=True
+    ),
+    required=True,
+    help="Output file. Include extension such as '.png'.",
+)
+@click.option(
+    "--aha-xmf",
+    "-a",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    help="Optional xmf file for AhA visualization",
+)
+@click.option(
+    "--ahb-xmf",
+    "-b",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    help="Optional xmf file for AhB visualization",
+)
+@click.option(
+    "--time-step",
+    "-t",
+    callback=_parse_step,
+    default="first",
+    show_default=True,
+    help=(
+        "Select a time step. Specify '-1' or 'last' to select the last time"
+        " step."
+    ),
+)
+@click.option(
+    "--animate", is_flag=True, help="Produce an animation of all time steps."
+)
+@click.option(
+    "--camera-angle",
+    "-c",
+    default="Side",
+    type=click.Choice(["Side", "Top", "Wide"]),
+    help=(
+        "Determines which camera angle to use: Default is the Side view.Top"
+        " view is right above the excisions at t = 0. Wide is further out and"
+        " inbetween Side and Top view"
+    ),
+)
+@click.option(
+    "--color-map",
+    "-m",
+    default="Rainbow Uniform",
+    help=(
+        'Determines how to color the domain, common color maps are "Inferno'
+        ' (matplotlib)", "Viridis (matplotlib). Defaults to Rainbow Uniform."'
+    ),
+)
+@click.option(
+    "--show-grid",
+    is_flag=True,
+    help="Show grid lines",
+)
+@click.option(
+    "--show-time",
+    is_flag=True,
+    help="Show simulation time",
+)
+def render_bbh_command(**kwargs):
+    _rich_traceback_guard = True
+    render_bbh(**kwargs)
+
+
+if __name__ == "__main__":
+    logging.basicConfig(level=logging.INFO)
+    render_bbh_command(help_option_names=["-h", "--help"])
--- a/src/Visualization/Python/Render3D/CMakeLists.txt
+++ b/src/Visualization/Python/Render3D/CMakeLists.txt
@@ -8,4 +8,5 @@ spectre_python_add_module(
   __init__.py
   Clip.py
   Domain.py
+  Bbh.py
 )
--- a/src/Visualization/Python/Render3D/__init__.py
+++ b/src/Visualization/Python/Render3D/__init__.py
@@ -9,6 +9,7 @@ from spectre.support.CliExceptions import RequiredChoiceError
 class Render3DCommands(click.MultiCommand):
     def list_commands(self, ctx):
         return [
+            "bbh",
             "clip",
             "domain",
         ]
@@ -24,6 +25,10 @@ class Render3DCommands(click.MultiCommand):
             )
 
             return render_domain_command
+        elif name == "bbh":
+            from spectre.Visualization.Render3D.Bbh import render_bbh_command
+
+            return render_bbh_command
         raise RequiredChoiceError(
             f"The command '{name}' is not implemented.",
             choices=self.list_commands(ctx),
--- a/src/Visualization/Python/Render3D/Clip.py
+++ b/src/Visualization/Python/Render3D/Clip.py
@@ -147,6 +147,10 @@ def render_clip_command(
     render_view.ResetCamera(True)
     render_view.CameraViewAngle /= zoom_factor
 
+    # Capture all frames
+    animation_scene = pv.GetAnimationScene()
+    animation_scene.PlayMode = "Snap To TimeSteps"
+
     if animate:
         pv.SaveAnimation(output, render_view)
     else:
--- a/src/Visualization/Python/Render3D/Domain.py
+++ b/src/Visualization/Python/Render3D/Domain.py
@@ -112,6 +112,10 @@ def render_domain(
     pv.ResetCamera()
     camera.Zoom(zoom_factor)
 
+    # Capture all frames
+    animation_scene = pv.GetAnimationScene()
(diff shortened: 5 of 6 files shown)


wthrowe:
Not a review of the code, but the upgrade instructions appear to be specific to one particular computer.

They also are missing the marker comments that will get them included in the release notes.  You should edit your initial post to add them:
```
### Upgrade instructions

<!--
If this PR makes changes that other people should be aware of when upgrading
their code, describe what they should do between the two UPGRADE INSTRUCTIONS
lines below.
-->
<!-- UPGRADE INSTRUCTIONS -->

<!-- UPGRADE INSTRUCTIONS -->
```

knelli2:
Thanks for adding this @AlexCarpenter46! It looks really cool :) I agree with @wthrowe's statement that this looks like it will only work for a specific machine (I think Wheeler?) and with your statement that we need some discussion about how we want this to work for the general case. For that reason, I'm adding the "in progress" label.

In the meantime, could you share some pictures of what this script produces? Colorful pictures are always fun to look at :)

AlexCarpenter46:
Yeah of course, I have a bunch of test runs on Wheeler in `panfs/ds09/sxs/acarpenter/Trace_Run/`. Here are some of the pictures of the first frame with the different camera angle options. Also I think @wthrowe is right, this might only work on wheeler and potentially stampede.
Binary_Pic 0000_angle_0
Binary_Pic 0000_angle_1
Binary_Pic 0000_angle_2

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+                   output_destination, camera_angle):
+    """Generate Pictures from XMF files for BBH Visualizations
+
+    Generate pictures from BBH runs using the XMF files generated using
+    generate-xdmf. For files to be read properly, the XMF files being pointed to
+    should be in the same directory as the Volume.h5 and Surfaces.h5 files.
+
+    To splice all the pictures into a video, look into FFmpeg"""
+
+    # create a new 'XDMF Reader'
```
nilsvu: Are these comments autogenerated by ParaViews recording feature? Either way, please remove comments that don't give any more information than the line of Python code (I don't need a comment explaining that the next line creates a new XDMF reader when the next line literally reads `reader = XDMFReader()`, or a comment saying "get animation scene" when the next line is `GetAnimationScene()`).

Please do keep or add comments that give some more context or explanations about the code, or that split the code into logical paragraphs.

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+    volulmeFilesxmf = XDMFReader(
+        registrationName=volume_xmf,
+        FileNames=[file_destination + '/' + volume_xmf])
```
nilsvu: Use `os.path.join`

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+    """Generate Pictures from XMF files for BBH Visualizations
+
+    Generate pictures from BBH runs using the XMF files generated using
+    generate-xdmf. For files to be read properly, the XMF files being pointed to
+    should be in the same directory as the Volume.h5 and Surfaces.h5 files.
+
+    To splice all the pictures into a video, look into FFmpeg"""
+
+    # create a new 'XDMF Reader'
+    volulmeFilesxmf = XDMFReader(
```
nilsvu: Please use snake_case for all variable names and function names, so this should be `volume_files_xmf`.

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+        FileNames=[file_destination + '/' + volume_xmf])
+    volulmeFilesxmf.PointArrayStatus = ['Lapse', 'SpatialRicciScalar']
+    # get animation scene
+    animationScene1 = GetAnimationScene()
+    # update animation scene based on data timesteps
+    animationScene1.UpdateAnimationUsingDataTimeSteps()
+    # get active view
+    renderView1 = GetActiveViewOrCreate('RenderView')
+    # show data in view
+    volulmeFilesxmfDisplay = Show(volulmeFilesxmf, renderView1,
```
nilsvu: Don't need to show this if you later hide it again

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+    # show data in view
+    warpByScalar1Display = Show(warpByScalar1, renderView1,
+                                'GeometryRepresentation')
+    # init the 'PiecewiseFunction' selected for 'ScaleTransferFunction'
+    warpByScalar1Display.ScaleTransferFunction.Points = [
+        0.4745234251022339, 0.0, 0.5, 0.0, 0.996696412563324, 1.0, 0.5, 0.0
+    ]
+    # init the 'PiecewiseFunction' selected for 'OpacityTransferFunction'
+    warpByScalar1Display.OpacityTransferFunction.Points = [
+        0.4745234251022339, 0.0, 0.5, 0.0, 0.996696412563324, 1.0, 0.5, 0.0
```
nilsvu: You use this transfer function a lot. Store it in a variable and explain it a bit.

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+    Hide(slice1, renderView1)
+    # show color bar/color legend
+    warpByScalar1Display.SetScalarBarVisibility(renderView1, True)
+    # update the view to ensure updated data information
+    renderView1.Update()
+    # Apply a preset using its name.
+    lapseLUT.ApplyPreset('Rainbow Uniform', True)
+    # invert the transfer function
+    lapseLUT.InvertTransferFunction()
+    if (aha_xmf != ''):
```
nilsvu: `if aha_xmf:`

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+        # create a new 'XDMF Reader'
+        ahAxmf = XDMFReader(registrationName=aha_xmf,
```
nilsvu: Factor this code into a separate function so you can call it twice, for the two horizons.

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+            -29.944619336722987, -3.666072157343372, 2.895224044348878
+        ]
+        renderView1.CameraFocalPoint = [
+            -0.13267040638072278, 0.6356115665206243, -0.37352608789235847
+        ]
+        renderView1.CameraViewUp = [0.0, 0.0, 1.0]
+        renderView1.CameraParallelScale = 519.6152422706632
+
+    # save animation
+    SaveAnimation(output_destination + '/Binary_Pic.png',
```
nilsvu: Use `os.path.join` here too

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+@click.option('--file-destination',
+              '-f',
+              required=True,
+              help="Path to directory containing the volume files xmf and"
+              "h5 files")
+@click.option('--output-destination',
+              '-o',
+              required=True,
+              help="Output directory where pictures will be placed")
+@click.option('--volume-xmf',
```
nilsvu: Make this option a `type=click.Path(exists=True, file_ok=True, dir_ok=False, readable=True`, and the aha, ahb options too. Then remove the `file_destination` option because I think you don't need it.

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+                  FrameRate=60)
+
+
+@click.command(help=generate_video.__doc__)
+@click.option('--file-destination',
+              '-f',
+              required=True,
+              help="Path to directory containing the volume files xmf and"
+              "h5 files")
+@click.option('--output-destination',
```
nilsvu: Make this also a `type=click.Path`. Name it maybe `--output-dir`?

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
...
+    '-a',
+    help="Optional xmf file for AhA for apparent horizon visualization")
+@click.option(
+    '--ahb-xmf',
+    '-b',
+    help="Optional xmf file for AhB for apparent horizon visualization")
+@click.option(
+    '--camera-angle',
+    default=0,
+    type=int,
```
nilsvu: Try `click.Choices`

support/Python/__main__.py
```diff
@@ -20,6 +20,7 @@ def list_commands(self, ctx):
             "clean-output",
             "extract-dat",
             "extract-input",
+            "generate-video",
```
nilsvu: How about `render-bbh`?

nilsvu (changes requested):
I think this is great, having ParaView rendering scripts in the repository would be very nice even though we have to figure out a way to test them. ParaView is large (couple of GB), but it should be fine to install it in a container and run the rendering tests only in one of the build configurations on CI. The first step towards this is adding a `vis` target to the Dockerfile (extending the `dev` target I think) and installing a minimal version of ParaView in it. Then we have to find ParaView in our CMake build system and only enable the rendering tests if we found it. These steps should be separate PRs.

src/Visualization/Python/GenerateVideo.py
AlexCarpenter46: Yeah, most of the comments are auto generated, I only added where I thought it might be helpful, will remove most of these

src/Visualization/Python/GenerateVideo.py
```diff
@@ -0,0 +1,312 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import click
+import glob
+import logging
+from paraview.simple import *
+
```
nilsdeppe: A lot of this will depend on the ParaView version. Can you check the version and add a warning message if the version isn't the one you used? We will inevitably need to have `if-else` checks for different versions.

src/Visualization/Python/GenerateVideo.py
AlexCarpenter46: This seems to be a default setting for Paraview that it decided to track in the script. However, I tried taking them out completely and nothing changed, so I think all of these can be safely taken out

src/Visualization/Python/CMakeLists.txt
```diff
@@ -13,5 +13,6 @@ spectre_python_add_module(
   InterpolateToMesh.py
   PlotDatFile.py
   ReadH5.py
+  RenderBBH.py
```
nilsvu: To be consistent with CamelCase naming please name this `RenderBbh.py`

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,178 @@
...
+
+import click
+import glob
+import logging
+import os
+import paraview.simple
+from paraview.simple import *
+
+#### disable automatic camera reset on 'Show'
+paraview.simple._DisableFirstRenderCameraReset()
```
nilsvu: Move into the render_bbh function

src/Visualization/Python/GenerateVideo.py
nilsvu: Reminder of this. Please be sure that all variable names are snake_case. Don't use capital letters in any variable names.

src/Visualization/Python/GenerateVideo.py
nilsvu: Reminder of this. The AHs are shown first, then hid.

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,178 @@
...
+    """Generate Pictures from XMF files for BBH Visualizations
+
+    Generate pictures from BBH runs using the XMF files generated using
+    generate-xdmf. For files to be read properly, the XMF files being pointed to
+    should be in the same directory as the Volume.h5 and Surfaces.h5 files.
+
+    To splice all the pictures into a video, look into FFmpeg"""
+    version = paraview.simple.GetParaViewVersion()
+    if (version < (5, 10) or version > (5, 10)):
+        print("WARNING: Your Paraview version is not 5.10, "
```
nilsvu: Use logging. write `logger = logging.getLogger(__name__)` at the very top of the script after imports. Then use `logger.warning` here.

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,178 @@
...
+    volume_files_xmf = XDMFReader(registrationName=volume_xmf,
+                                  FileNames=[volume_xmf])
+    volume_files_xmf.PointArrayStatus = ['Lapse', 'SpatialRicciScalar']
+    animation_scene1 = GetAnimationScene()
+    animation_scene1.UpdateAnimationUsingDataTimeSteps()
+    render_view1 = GetActiveViewOrCreate('RenderView')
+    lapse_LUT = GetColorTransferFunction('Lapse')
+    lapse_PWF = GetOpacityTransferFunction('Lapse')
+    render_view1.ResetCamera(False)
+    materialLibrary1 = GetMaterialLibrary()
```
nilsvu: What does this do?

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,178 @@
...
+    SaveAnimation(os.path.join(output_dir, 'Binary_Pic.png'),
+                  render_view1,
+                  ImageResolution=[1920, 1080],
+                  FrameRate=60)
+
+
+@click.command(help=render_bbh.__doc__)
+@click.option('--output-dir',
+              '-o',
+              type=click.Path(exists=True, file_okay=False, dir_okay=True),
```
nilsvu: Add `writable=True`

AlexCarpenter46:
Since it's been a little while, I decided to just rebase and everyone viewing now should treat this as a fresh PR. The test will fail because I'm not exactly sure what a good way to test this is. It'd need 3D volume data with the Lapse and SpatialRicciScalar output (as I'm requiring those things if you're trying to use this script). I've added some new options as well that do some cool things like set the stride, change the color map and so on. I could add an option for Psi4 data that would make something closer to my videos. Let me know if we want something like that!

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+
+    Ah_xmf = pv.XDMFReader(registrationName=ah_xmf, FileNames=[ah_xmf])
+    Ah_xmf.PointArrayStatus = ["RicciScalar"]
+    ricciScalarLUT = pv.GetColorTransferFunction("RicciScalar")
+    ricciScalarPWF = pv.GetOpacityTransferFunction("RicciScalar")
+    Ah_xmf_Display = pv.Show(
+        Ah_xmf, render_view, "UnstructuredGridRepresentation"
+    )
+    Ah_xmf_Display.SetScalarBarVisibility(render_view, False)
+    render_view.Update()
```
nilsvu: You can delete these lines and the `pv.Hide` below I think.

Actually I think you can delete a lot more of the lines in this function. They show and then hide again a bunch of things. Also the business with the transfer functions etc.

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import os
+
+import click
+import rich.columns
+
+
+def ah_vis(ah_xmf, render_view):
```
nilsvu: Add a docstring to explain briefly what this function does please, and document the arguments it takes.

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+    color_map,
+    show_grid,
+    stride,
+):
+    """Generate Pictures from XMF files for BBH Visualizations
+
+    Generate pictures from BBH runs using the XMF files generated using
+    generate-xdmf. For files to be read properly, the XMF files being pointed to
+    should be in the same directory as the VolumeData.h5 and SurfacesData.h5
+    files.
```
nilsvu: Delete this last sentence, it's not generally true.

FYI: GenerateXdmf can write absolute or relative paths. If it writes absolute paths, the files can be moved around on the same machine. If it writes relative paths, the files must be kept relative to each other, but can be transferred to other machines. Either way, they may not be in the same directory.

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+
+    To splice all the pictures into a video, try using FFmpeg"""
+    import paraview.simple as pv
+
+    version = pv.GetParaViewVersion()
+    if version < (5, 11) or version > (5, 11):
+        print(
```
nilsvu: use `logger.warning`

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+            "the script may not work correctly."
+        )
+
+    # Volume Data Visualization
+    volume_files_xmf = pv.XDMFReader(
+        registrationName=volume_xmf, FileNames=[volume_xmf]
+    )
+
+    # Check for Lapse and SpatialRicciScalar
+    variables = volume_files_xmf.PointData.keys()
```
nilsvu: wrap in `list(...)`. Then:

```py
assert "Lapse" in variables, "doesn't have lapse..."
assert "SpatialRicciScalar" in variables, "..."
```

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+    output_dir,
+    volume_xmf,
+    aha_xmf,
+    ahb_xmf,
+    camera_angle,
+    color_map,
+    show_grid,
+    stride,
+):
+    """Generate Pictures from XMF files for BBH Visualizations
```
nilsvu: Please add a section to the docs where you explain each argument. It can also be helpful to add type annotations (see other files like `Domain.py`) and defaults for the arguments.

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+    render_view.OrientationAxesVisibility = 0
+    pv.SetActiveSource(warp_by_scalar)
+    warp_by_scalar_display.Opacity = 0.8
+    pv.ColorBy(warp_by_scalar_display, ("POINTS", "Lapse"))
+    layout = pv.GetLayout()
+    layout.SetSize(1920, 1080)
+
+    # Camera placements
+    # Top down view
+    if camera_angle == "1":
```
nilsvu: I like this idea of good predefined camera angles. Use names instead of numbers though.

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+
+    # save animation
+    animation_scene = pv.GetAnimationScene()
+    animation_scene.Stride = stride
+    pv.SaveAnimation(
+        os.path.join(output_dir, "Binary_Pic.png"),
+        render_view,
+        ImageResolution=[1920, 1080],
+        FrameRate=60,
+    )
```
nilsvu: Just `pv.SaveAnimation(output, render_view)` is enough, so have the user pass in an `output` filename rather than a directory. Then the user decided the filename and format, and ParaView takes care of the rest.

src/Visualization/Python/RenderBBH.py
```diff
@@ -0,0 +1,250 @@
...
+
+@click.command(help=render_bbh.__doc__)
+@click.option(
+    "--output-dir",
+    "-o",
+    type=click.Path(exists=True, file_okay=False, dir_okay=True),
+    required=True,
+    help="Output directory where pictures will be placed",
+)
+@click.option(
```
nilsvu: make this a `click.argument`, because this is the main file being operated on.

support/Python/__main__.py
```diff
@@ -31,6 +31,7 @@ def list_commands(self, ctx):
             "interpolate-to-mesh",
             "interpolate-to-points",
             "plot",
+            "render-bbh",
```
nilsvu: move this into the `Render3D` subdirectory please and add to the subcommand there, which already has `domain` and `slice` subcommands.

nilsvu (changes requested):
You can squash everything directly

AlexCarpenter46:
I think this might be cool to have at the sxs workshop. Here are some pictures of different output and things you can do!
Screenshot 2024-07-26 at 4 34 40 PM
Screenshot 2024-07-26 at 4 35 19 PM
Screenshot 2024-07-26 at 4 36 19 PM

src/Visualization/Python/Render3D/CMakeLists.txt
```diff
@@ -8,4 +8,5 @@ spectre_python_add_module(
   __init__.py
   Clip.py
   Domain.py
+  RenderBbh.py
```
nilsvu: Just `Bbh.py` (see the other files in this directory). Also the name of the CLI command should be `spectre render-3d bbh`, not `spectre render-3d render-bbh` as it currently is.

src/Visualization/Python/Render3D/RenderBbh.py
```diff
@@ -0,0 +1,257 @@
...
+import logging
+import os
+
+import click
+import rich.columns
+
+logger = logging.getLogger(__name__)
+
+
+"""
```
nilsvu: Move into function like other docstrings

src/Visualization/Python/Render3D/RenderBbh.py
```diff
@@ -0,0 +1,257 @@
...
+    generate-xdmf. This script requires that the Lapse and SpatialRicciScalar
+    were output in the volume data.
+
+    Arguments:
+
+      volume_xmf: Path to the volume data xmf file.
+      output: Name of output file generated from paraview. Include extensions
+      such as '.png'
+      aha_xmf: Path to the apparent horizon xmf file for object A.
+      aha_xmf: Path to the apparent horizon xmf file for object B.
```
nilsvu: ahb

src/Visualization/Python/Render3D/RenderBbh.py
```diff
@@ -0,0 +1,257 @@
...
+      camera_angle: Specified camera angle, defaults to Side if empty.
```
nilsvu: Specify type of argument and possible options

src/Visualization/Python/Render3D/RenderBbh.py
```diff
@@ -0,0 +1,257 @@
...
+            registrationName="annotate_time_filter", Input=slice
+        )
+        time_filter.Format = "Time: {time:0.2f}M"
+        annotate_time_filter_display = pv.Show(
+            time_filter, render_view, "TextSourceRepresentation"
+        )
+        annotate_time_filter_display.FontSize = 45
+
+    # Save animation
+    pv.SaveAnimation(output, render_view)
```
nilsvu: Can you do the same selection between single time step and animation here as is done in the other Render3d scripts?

src/Visualization/Python/Render3D/RenderBbh.py
```diff
@@ -0,0 +1,257 @@
...
+
+
+@click.command(help=render_bbh.__doc__)
```
nilsvu: Add `name="bbh"`

src/Visualization/Python/Render3D/__init__.py
```diff
@@ -9,6 +9,7 @@
 class Render3DCommands(click.MultiCommand):
     def list_commands(self, ctx):
         return [
+            "render-bbh",
```
nilsvu: `"bbh"`

src/Visualization/Python/Render3D/__init__.py
```diff
@@ -24,6 +25,12 @@ def get_command(self, ctx, name):
             )
 
             return render_domain_command
+        elif name == "render-bbh":
```
nilsvu: `"bbh"`

tests/Unit/Visualization/Python/Render3D/Test_RenderBbh.py
```diff
@@ -0,0 +1,44 @@
...
+
+    def test_render_bbh(self):
+        runner = CliRunner()
+        result = runner.invoke(
+            render_bbh_command,
+            [self.vol_test_data, "-o", self.test_dir, "-c", "Top"],
+            catch_exceptions=False,
+        )
+        self.assertEqual(result.exit_code, 0, msg=result.output)
+        self.assertTrue(os.path.exists(self.output_file), msg=result.output)
```
nilsvu: Does this test work? There's no BBH data in the repo

nilsvu (changes requested):

nilsvu:
@AlexCarpenter46 yes I agree, would be very nice!

tests/Unit/Visualization/Python/Render3D/Test_RenderBbh.py
AlexCarpenter46: No, it doesn't work, it's sort of just a skeleton for if we did have bbh data in the rep with spatial ricci and lapse output

nilsvu: Ok just delete it then. We should add a test that generates fake BBH data and then runs the rendering script, but we can do that after SXScon (open an issue please).

nilsvu (dismissed):
LGTM please squash

src/Visualization/Python/Render3D/Bbh.py
```diff
@@ -0,0 +1,291 @@
...
+    slice.HyperTreeGridSlicer = "Plane"
+    slice.SliceOffsetValues = [0.0]
+    slice.SliceType.Normal = [0.0, 0.0, 1.0]
+    slice.Triangulatetheslice = 0
+
+    # Warp grid by spatial ricci scalar
+    warp_by_scalar = pv.WarpByScalar(
+        registrationName="WarpByScalar", Input=slice
+    )
+    warp_by_scalar.Scalars = ["POINTS", "SpatialRicciScalar"]
```
nilsvu, line 126: Mind adding `SpatialRicciScalar` to the output in `Inspiral.yaml`, so we can actually use this visualization script with default output? Or make `Lapse` and `SpatialRicciScalar` optional in this script. (if you add the output, please add a comment that it's used for visualization in Render3D/Bbh.py).

nilsvu (approved):

PR #4658, opened 2023-01-27
Add StaticDeque

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/StaticDeque.hpp
+++ b/src/DataStructures/StaticDeque.hpp
@@ -1,0 +1,549 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <algorithm>
+#include <cstddef>
+#include <iterator>
+#include <pup.h>
+#include <type_traits>
+
+#include "Utilities/PrintHelpers.hpp"
+#include "Utilities/Requires.hpp"
+#include "Utilities/StlBoilerplate.hpp"
+
+namespace static_deque_detail {
+// "The extent to which an implementation determines that a type
+// cannot be an input iterator is unspecified, except that as a
+// minimum integral types shall not qualify as input iterators."
+//   -- N4659 26.2.1.17
+template <typename U>
+constexpr static bool is_input_iterator = not std::is_integral_v<U>;
+}  // namespace static_deque_detail
+
+/// A class implementing the std::deque interface with static storage.
+///
+/// Differences from std::deque:
+/// * The size of the container cannot be increased above \p Capacity.
+/// * Operations moving the entire container are \f$O(n)\f$ instead of
+///   \f$O(1)\f$ in the size and invalidate all references and
+///   iterators.
+/// * Erasing elements from the front of the queue invalidates all
+///   iterators (but not references).
+///
+/// This last point is not a fundamental limitation, but could be
+/// corrected with a more complicated iterator implementation if the
+/// standard behavior is found to be useful.
+template <typename T, size_t Capacity>
+class StaticDeque
+    : public stl_boilerplate::RandomAccessSequence<StaticDeque<T, Capacity>,
+                                                   T> {
+ public:
+  // Aliases needed in the class definition.  The rest can just be
+  // inherited.
+  using size_type = typename StaticDeque::size_type;
+  using difference_type = typename StaticDeque::difference_type;
+  using iterator = typename StaticDeque::iterator;
+  using const_iterator = typename StaticDeque::const_iterator;
+
+  StaticDeque() = default;
+  explicit StaticDeque(size_type count);
+  StaticDeque(size_type n, const T& value);
+  template <
+      typename InputIterator,
+      Requires<static_deque_detail::is_input_iterator<InputIterator>> = nullptr>
+  StaticDeque(InputIterator first, InputIterator last);
+  StaticDeque(const StaticDeque& other);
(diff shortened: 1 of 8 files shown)


teukolsky:
Hi Will,
What is the proposed use of this?

wthrowe:
Yeah, probably should have mentioned that somewhere.

It's to store the past values for multistep time steppers.  The usage pattern makes this naturally a deque, and knowing what time steppers we have we know the queue will never have more than 8 elements.

I'm pretty sure libstdc++'s implementation of `std::deque` will keep allocating blocks at one end and deallocating them at the other.  (The libstdc++ code is a bit dense, so it is possible I'm not interpreting it correctly.)  For something as large as the GH variables, I think each block will hold one step, so there will be one allocation per step.

I think libc++'s implementation is actually smart enough to keep and reuse its storage blocks, so it would be fine for the main history.  It's not clear whether the initial allocation of space would be significant in the IMEX code, which creates a temporary `History<double>`.  We can't rely on using this implementation anyway, so it's academic.

This idea was also mentioned in passing in some unrelated recent PR, but I don't remember which or how useful it would actually have been there.

nilsdeppe:
Ah, I think @kidder used a `std::deque` in some AMR code

src/DataStructures/StaticDeque.hpp
```diff
@@ -0,0 +1,520 @@
...
+    push_back(value);
+  }
+}
+
+template <typename T, size_t Capacity>
+template <typename... Args>
+T& StaticDeque<T, Capacity>::emplace_front(Args&&... args) {
+  ASSERT(size() < max_size(), "Container is full.");
+  auto* const new_entry =
+      new (entry_address(Capacity - 1)) T(std::forward<Args>(args)...);
```
nilsdeppe, line 298: Just to make sure I understand:

You store a series of bytes that has the correct size for the capacity, you then use placement-new to add new objects into the byte string? This might be nice to add to the comment before `data_` in the class. Something like `New data is added using placement-new.`

src/DataStructures/StaticDeque.hpp
```diff
@@ -0,0 +1,520 @@
...
+  // This must be done last in case T's constructor throws an
+  // exception.
+  if (start_position_ == 0) {
+    start_position_ = Capacity - 1;
```
nilsdeppe, line 302: Another thing to make sure I understand, and that might be useful to state in a C++ comment (sorry if I missed the comment!): here rather than touching every element, you use a `start_position_` to keep track of where in the static buffer you start from, then pushing to the front can wrap to the end of the buffer, but pushing to the back will never wrap? Maybe add to the `data_` comment: `This means that for adding data to the front we need to check if our start position is 0, since then we start adding at the end and update start_position_ to the end of the byte stream.` ?

src/DataStructures/StaticDeque.hpp
```diff
@@ -0,0 +1,520 @@
...
+template <typename T, size_t Capacity>
+void StaticDeque<T, Capacity>::pop_front() {
+  ASSERT(not this->empty(), "Container is empty.");
+  if (start_position_ == Capacity - 1) {
+    start_position_ = 0;
+  } else {
+    ++start_position_;
+  }
+  --size_;
+  entry_address(Capacity - 1)->~T();
```
nilsdeppe, line 456: Since explicitly calling destructors looks fishy, maybe add a C++ comment briefly saying it's intentional and why? I worry that someone will think their bug is here in 5 years from now because of some rule they read online 🙃

src/DataStructures/StaticDeque.hpp
```diff
@@ -0,0 +1,520 @@
...
+  }
+}
+
+template <typename T, size_t Capacity>
+void StaticDeque<T, Capacity>::pup(PUP::er& p) {
+  size_type sz = size();
+  p | sz;
+  resize(sz);
+  for (auto& entry : *this) {
+    p | entry;
```
nilsdeppe, line 522: If we end up with this class holding a fundamental types (I currently don't see that we would get that a lot), then we would want this to use `PUP_array` that serializes a byte stream directly (might be more efficient). I don't think that's important now, but just want to mention it :)

nilsdeppe (changes requested):
This looks good, a couple requests for a bit more C++ comments that I think would make it clearer to readers in the future. Please squash them in directly :)

src/DataStructures/StaticDeque.hpp
wthrowe: For fundamentals (and I think any trivially copyable types), I'm pretty sure we can just serialize the byte stream and the other member variables and restore the exact state.  For more complicated things, we have to be careful not to try to pup the unused memory as objects of type T, so the current version just accesses the objects through the public interface and doesn't try to restore the details of the internal representation.  (The elements may be "rotated" in the internal circular buffer after deserializing.)  Switching to the first implementation when allowed would certainly be possible, but let's put it off for now.  I've added a comment summarizing this.

wthrowe (commented):
I have expanded the comment explaining the meaning of the member variables and the logical treatment of the storage, and added some short comments for the other things.

src/DataStructures/StaticDeque.hpp
nilsdeppe: I agree with you! Thanks for adding the nice comment!

nilsdeppe (approved):

PR #4665, opened 2023-01-30
StrahlkorperCoords src can be Frame::Distorted.

markscheel:
StrahlkorperCoordsInDifferentFrame now can use distorted coords as its input.

Formerly only Frame::Grid was allowed as input.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
+++ b/src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
@@ -8,6 +8,7 @@
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "DataStructures/Variables.hpp"
 #include "Domain/Block.hpp"
+#include "Domain/BlockLogicalCoordinates.hpp"
 #include "Domain/CoordinateMaps/CoordinateMap.hpp"
 #include "Domain/Domain.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
@@ -15,6 +16,7 @@
 #include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/YlmSpherepack.hpp"
 #include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 
@@ -27,6 +29,8 @@ void strahlkorper_coords_in_different_frame(
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time,
     const double time) {
+  static_assert(std::is_same_v<DestFrame, ::Frame::Inertial>,
+                "Destination frame must currently be Inertial frame");
   destructive_resize_components(
       dest_cartesian_coords, src_strahlkorper.ylm_spherepack().physical_size());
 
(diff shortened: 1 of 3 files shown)


src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
```diff
@@ -55,38 +55,72 @@ void strahlkorper_coords_in_different_frame(
...
+            grid_to_distorted_map.inverse(x_src, time, functions_of_time);
+        if (x_grid.has_value()) {
+          const auto x_logical = logical_to_grid_map.inverse(x_grid.value());
+          // x_logical might be an empty std::optional.
+          if (x_logical.has_value() and get<0>(x_logical.value()) <= 1.0 and
+              get<0>(x_logical.value()) >= -1.0 and
+              get<1>(x_logical.value()) <= 1.0 and
+              get<1>(x_logical.value()) >= -1.0 and
+              get<2>(x_logical.value()) <= 1.0 and
+              get<2>(x_logical.value()) >= -1.0) {
```
nilsvu: This looks like it could suffer from the same roundoff error bug we had in BlockLogicalCoordinates (#4510). Perhaps apply the same fix that we have in `BlockLogicalCoordinates.cpp`? Same in the code branch above.

src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
```diff
@@ -55,38 +55,72 @@ void strahlkorper_coords_in_different_frame(
...
+        }
+      }
+    } else {
+      static_assert(std::is_same_v<SrcFrame, ::Frame::Distorted>,
+                    "Src frame must be Distorted if it is not Grid");
+      const auto& grid_to_distorted_map =
+          block.moving_mesh_grid_to_distorted_map();
+      const auto& distorted_to_inertial_map =
+          block.moving_mesh_distorted_to_inertial_map();
+      const auto& logical_to_grid_map = block.moving_mesh_logical_to_grid_map();
```
nilsvu: FYI, we have a runtime `Composition` coordinate map now in `Domain/CoordinateMaps/Composition.hpp`. Since you're working on frame things at the moment, maybe it can be useful for you at some point. Please note the warning on usage/performance in the Composition docs.

nilsvu (changes requested):

src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
markscheel: Thanks!  Already using `Composition` in my next branch (not a PR yet)

src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
```diff
@@ -55,38 +55,85 @@ void strahlkorper_coords_in_different_frame(
...
+      // Fill only those dest_cartesian_coords that are in this Block.
+      // Determine which coords are in this Block by checking if the
+      // inverse grid-to-logical map yields logical coords between -1 and 1.
+      for (size_t s = 0; s < get<0>(src_cartesian_coords).size(); ++s) {
+        const tnsr::I<double, 3, SrcFrame> x_src{
+            {get<0>(src_cartesian_coords)[s], get<1>(src_cartesian_coords)[s],
+             get<2>(src_cartesian_coords)[s]}};
+        const auto x_logical = logical_to_grid_map.inverse(x_src);
+        const double one_plus_eps =
+            1.0 + std::numeric_limits<double>::epsilon() * 100.0;
```
nilsvu: Suggestion: to avoid potential bugs like this, call `block_logical_coordinates` before the loop over blocks. It will do the SrcFrame -> BlockLogical mapping and return a vector of optionals, one for each grid point. Then, in this loop over grid points, check `block_logical_coords[s].has_value() and block_logical_coords[s]->id == block.id()`.

nilsvu (changes requested):
You can squash

src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
markscheel: Even better: if I call `block_logical_coordinates`, then I think I don't need to loop over blocks at all here.  I just need to loop over points and call either `grid_to_inertial_map` or `grid_to_distorted_map` for the (guaranteed unique) Block that is returned by `block_logical_coordinates` for that point.  Any objection to me doing that?

nilsvu: sounds good

markscheel: Made this change and squashed (and rebased).

src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
```diff
@@ -53,41 +56,31 @@ void strahlkorper_coords_in_different_frame(
...
-        // Blocks, it might get filled twice, but that's ok.
-      }
+  // Block by Block.  Since each point in general has a different
+  // block (given by block_logical_coordinates), this means we must
+  // map point by point.
+  const auto block_logical_coords = block_logical_coordinates(
+      domain, src_cartesian_coords, time, functions_of_time);
+  for (size_t s = 0; s < block_logical_coords.size(); ++s) {
+    const auto& block =
+        domain.blocks()[block_logical_coords[s].value().id.get_index()];
```
nilsvu: Assert has_value first?

The current logic will break if the Strahlkorper extends into a block that has no distorted frame. Please make that clear in the function dox.

nilsvu (changes requested):

src/ApparentHorizons/StrahlkorperCoordsInDifferentFrame.cpp
markscheel: Added an assert and squashed.

nilsvu: Ping on the function dox (sorry, I may have added that bit in an edit to the comment)

markscheel: Updated dox and squashed.

nilsvu (approved):

PR #4666, opened 2023-01-30
Change `ExcisionSphere::center()` return type to `tnsr::I<double, 3, Frame::Grid>`

nikwit:
The original return type was `std::array<double, 3>` but that makes it ambiguous in which coordinates the center is given.

### DIFF
--- a/src/ControlSystem/ControlErrors/Rotation.hpp
+++ b/src/ControlSystem/ControlErrors/Rotation.hpp
@@ -12,6 +12,7 @@
 #include "ControlSystem/Tags.hpp"
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
 #include "Options/Options.hpp"
 #include "Parallel/GlobalCache.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
@@ -81,16 +82,19 @@ struct Rotation : tt::ConformsTo<protocols::ControlError> {
            "Excision sphere for ObjectB not in the domain but is needed to "
            "compute Rotation control error.");
 
-    const DataVector grid_position_of_A = array_to_datavector(
-        domain.excision_spheres().at("ObjectAExcisionSphere").center());
-    const DataVector grid_position_of_B = array_to_datavector(
-        domain.excision_spheres().at("ObjectBExcisionSphere").center());
+    const tnsr::I<double, 3, Frame::Grid>& grid_position_of_A =
+        domain.excision_spheres().at("ObjectAExcisionSphere").center();
+    const tnsr::I<double, 3, Frame::Grid>& grid_position_of_B =
+        domain.excision_spheres().at("ObjectBExcisionSphere").center();
     const DataVector& current_position_of_A = get<center_A>(measurements);
     const DataVector& current_position_of_B = get<center_B>(measurements);
 
     // A is to the right of B in grid frame. To get positive differences,
     // take A - B
-    const DataVector grid_diff = grid_position_of_A - grid_position_of_B;
+    const auto grid_diff_tnsr = tenex::evaluate<ti::I>(
+        grid_position_of_A(ti::I) - grid_position_of_B(ti::I));
+    const DataVector grid_diff{
+        {grid_diff_tnsr[0], grid_diff_tnsr[1], grid_diff_tnsr[2]}};
     const DataVector current_diff =
         current_position_of_A - current_position_of_B;
 
--- a/src/Domain/Creators/CylindricalBinaryCompactObject.cpp
+++ b/src/Domain/Creators/CylindricalBinaryCompactObject.cpp
@@ -842,8 +842,10 @@ Domain<3> CylindricalBinaryCompactObject::create_domain() const {
   }
   excision_spheres.emplace(
       "ObjectAExcisionSphere",
-      ExcisionSphere<3>{radius_A_, rotate_from_z_to_x_axis(center_A_),
-                        abutting_directions_A});
+      ExcisionSphere<3>{
+          radius_A_,
+          tnsr::I<double, 3, Frame::Grid>(rotate_from_z_to_x_axis(center_A_)),
+          abutting_directions_A});
 
   std::unordered_map<size_t, Direction<3>> abutting_directions_B;
   if (include_inner_sphere_B_) {
@@ -870,8 +872,10 @@ Domain<3> CylindricalBinaryCompactObject::create_domain() const {
   }
   excision_spheres.emplace(
       "ObjectBExcisionSphere",
-      ExcisionSphere<3>{radius_B_, rotate_from_z_to_x_axis(center_B_),
-                        abutting_directions_B});
+      ExcisionSphere<3>{
+          radius_B_,
+          tnsr::I<double, 3, Frame::Grid>(rotate_from_z_to_x_axis(center_B_)),
+          abutting_directions_B});
 
   Domain<3> domain{std::move(coordinate_maps), std::move(excision_spheres),
                    block_names_, block_groups_};
--- a/src/Domain/Structure/ExcisionSphere.hpp
+++ b/src/Domain/Structure/ExcisionSphere.hpp
@@ -6,14 +6,13 @@
 
 #pragma once
 
-#include <array>
 #include <cstddef>
 #include <iosfwd>
 #include <limits>
 #include <unordered_map>
 
+#include "DataStructures/Tensor/Tensor.hpp"
 #include "Domain/Structure/Direction.hpp"
-#include "Utilities/MakeArray.hpp"
 
 /// \cond
 namespace PUP {
@@ -39,7 +38,7 @@ class ExcisionSphere {
   /// \param abutting_directions the set of blocks that touch the excision
   /// sphere, along with the direction in which they touch it.
   ExcisionSphere(
-      double radius, std::array<double, VolumeDim> center,
+      double radius, tnsr::I<double, VolumeDim, Frame::Grid> center,
       std::unordered_map<size_t, Direction<VolumeDim>> abutting_directions);
 
   /// Default constructor needed for Charm++ serialization.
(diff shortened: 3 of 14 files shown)


src/ControlSystem/ControlErrors/Rotation.hpp
```diff
@@ -81,21 +85,23 @@ struct Rotation : tt::ConformsTo<protocols::ControlError> {
...
+    const auto grid_diff_tnsr = tenex::evaluate<ti::I>(
+        grid_position_of_A(ti::I) - grid_position_of_B(ti::I));
+    const DataVector grid_diff{
+        {grid_diff_tnsr[0], grid_diff_tnsr[1], grid_diff_tnsr[2]}};
+    const auto current_diff = current_position_of_A - current_position_of_B;
 
-    const double grid_dot_current = dot(grid_diff, current_diff);
-    const DataVector grid_cross_current = cross(grid_diff, current_diff);
+    const auto grid_dot_current = dot(grid_diff, current_diff);
+    const auto grid_cross_current = cross(grid_diff, current_diff);
```
nilsvu: Personally I think the explicit types that you removed here were helpful to understand this code, in particular because you're dealing with DataVector, Tensor, double, etc and it's not obvious from reading the code what the type is. Your IDE knows it, but GitHub doesn't. I'd say keep the explicit types. Same in the other files.

src/ControlSystem/ControlErrors/Rotation.hpp
```diff
@@ -12,6 +12,10 @@
 #include "ControlSystem/Tags.hpp"
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/CrossProduct.hpp"
+#include "DataStructures/Tensor/EagerMath/DotProduct.hpp"
```
nilsvu: you don't use these I think

src/Domain/Structure/ExcisionSphere.hpp
```diff
@@ -6,14 +6,13 @@
 
 #pragma once
 
-#include <array>
 #include <cstddef>
 #include <iosfwd>
 #include <limits>
 #include <unordered_map>
 
+#include "DataStructures//Tensor/Tensor.hpp"
```
nilsvu: fix extra slash

tests/Unit/ControlSystem/ControlErrors/Test_Shape.cpp
```diff
@@ -64,10 +64,11 @@ void test_shape_control_error() {
   MAKE_GENERATOR(generator);
   domain::FunctionsOfTime::register_derived_with_charm();
 
-  const std::array<double, 3> origin{{0.0, 0.0, 0.0}};
+  const tnsr::I<double, 3, Frame::Grid> origin{0.};
+  const std::array<double, 3> origin_array{{0., 0., 0.}};
   const double ah_radius = 1.5;
   const double initial_time = 0.0;
-  Strahlkorper fake_ah{10, 10, origin};
+  Strahlkorper fake_ah{10, 10, origin_array};
```
nilsvu: I think you can use `make_array` from `Utilities/MakeArray.hpp` here and in other places in this PR. `make_array<double, 3>(origin)` should do it.

nilsvu (changes requested):

nilsvu:
Please rebase as well to fix the conflicts

src/ControlSystem/ControlErrors/Rotation.hpp
knelli2: I second this. Please keep the types here (and in the `Translation` control error) explicit.

src/ControlSystem/ControlErrors/Rotation.hpp
knelli2: Correct. Once you convert from the `tnsr` to `DataVector` it uses the `DataVector` `dot`.

knelli2 (changes requested):

src/ControlSystem/ControlErrors/Rotation.hpp
nikwit: okay, it's one of Scott Meyer's rules to pretty much always prefer auto, also because it makes these refactors a lot easier. But I don't care that much.

src/ControlSystem/ControlErrors/Rotation.hpp
nikwit: oops, leftover from previous attempt

src/ControlSystem/ControlErrors/Rotation.hpp
```diff
@@ -12,6 +12,8 @@
 #include "ControlSystem/Tags.hpp"
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Expressions/Evaluate.hpp"
```
nilsvu: (nitpick, optional) You don't need this include, tensor expressions are included through Tensor.hpp

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -840,10 +840,14 @@ Domain<3> CylindricalBinaryCompactObject::create_domain() const {
...
   }
+  const auto center_A_x_axis = rotate_from_z_to_x_axis(center_A_);
   excision_spheres.emplace(
       "ObjectAExcisionSphere",
-      ExcisionSphere<3>{radius_A_, rotate_from_z_to_x_axis(center_A_),
-                        abutting_directions_A});
+      ExcisionSphere<3>{
+          radius_A_,
+          tnsr::I<double, 3, Frame::Grid>{
+              {center_A_x_axis[0], center_A_x_axis[1], center_A_x_axis[2]}},
```
nilsvu: Tensor has a constructor from std::array, so you can probably use that here and in other places.

src/ControlSystem/ControlErrors/Rotation.hpp
```diff
@@ -81,21 +83,24 @@ struct Rotation : tt::ConformsTo<protocols::ControlError> {
...
+        grid_position_of_A(ti::I) - grid_position_of_B(ti::I));
+    const DataVector grid_diff{
+        {grid_diff_tnsr[0], grid_diff_tnsr[1], grid_diff_tnsr[2]}};
     const DataVector current_diff =
         current_position_of_A - current_position_of_B;
 
-    const double grid_dot_current = dot(grid_diff, current_diff);
-    const DataVector grid_cross_current = cross(grid_diff, current_diff);
+    const auto grid_dot_current = dot(grid_diff, current_diff);
+    const auto grid_cross_current = cross(grid_diff, current_diff);
```
nilsvu: These two return types I find useful as well

nilsvu (changes requested):
Please squash, including this

src/ControlSystem/ControlErrors/Rotation.hpp
nilsvu: We've followed that rule for a long time and it's still the default, but we have started to prefer explicit return types where it makes the code easier to understand.

src/ControlSystem/ControlErrors/Rotation.hpp
```diff
@@ -12,6 +12,7 @@
 #include "ControlSystem/Tags.hpp"
 #include "DataStructures/DataBox/DataBox.hpp"
```
knelli2, line 13: [random line]

While you're fixing the build error, could you shorten the commit message so it's under 72 characters? Maybe just
```
Change return type of ExcisionSphere::center()
```

knelli2 (changes requested):

nilsvu (dismissed):

knelli2 (dismissed):

nilsvu:
@gsb76 do you want to have a quick look at this before I merge it?

nilsvu:
@nikwit please do a quick rebase to fix the conflict

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -291,11 +291,13 @@ void test_connectivity_once(const bool with_sphere_e,
     const auto& excision_sphere_a =
         domain.excision_spheres().at("ObjectAExcisionSphere");
     CHECK(excision_sphere_a.radius() == inner_radius_objectA);
-    CHECK(excision_sphere_a.center() == center_objectA);
     const auto& excision_sphere_b =
         domain.excision_spheres().at("ObjectBExcisionSphere");
     CHECK(excision_sphere_b.radius() == inner_radius_objectB);
-    CHECK(excision_sphere_b.center() == center_objectB);
+    for (size_t i = 0; i < 3; ++i) {
```
mar-celine, line 297: If it's possible to do this with `CHECK_ITERABLE_APPROX` I would do that, otherwise this is ok

tests/Unit/ControlSystem/ControlErrors/Test_Shape.cpp
```diff
@@ -64,10 +65,10 @@ void test_shape_control_error() {
   MAKE_GENERATOR(generator);
   domain::FunctionsOfTime::register_derived_with_charm();
 
-  const std::array<double, 3> origin{{0.0, 0.0, 0.0}};
+  const tnsr::I<double, 3, Frame::Grid> origin{0.};
```
mar-celine: For consistency have `0.` be `0.0` to match the formatting elsewhere in this file

mar-celine (commented):
Adding the Frame is :+1: , just minor comments if possible, I don't want to hold anything up

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
nikwit: I think this check should be exact because the floats get copied only.

nilsvu (dismissed):

nilsvu:
CI is complaining about long lines in BBH domain test

nilsvu (approved):

PR #4668, opened 2023-01-30
Add tags related to electric current density terms in ForceFree system

yoonso0-0:
- rename SpatialCurrentDensity to CurrentDensity
- Add two separate tags for TildeJ (J times lapse time sqrt det spatial metric) term. One is explicit and the other is implicit part.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ForceFree/Tags.hpp
+++ b/src/Evolution/Systems/ForceFree/Tags.hpp
@@ -41,13 +41,6 @@ struct ChargeDensity : db::SimpleTag {
 };
 
 /*!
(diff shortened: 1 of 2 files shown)


wthrowe:
What is the reason for this change?

How is this split defined?  I usually associate implicit and explicit with terms in the RHS, not with variables.  (This might be obvious if I was more familiar with the FFE equations.)

yoonso0-0:
@wthrowe I'll move to another PR to answer your question

PR #4670, opened 2023-01-30
Split electric current density to stiff and non-stiff parts and update Fluxes and Sources in ForceFree system.

yoonso0-0:
## Proposed changes

- Add two tags `DriftTildeJ` and `ParallelTildeJ`, which corresponds to the drift component (non-stiff part) and parallel component (stiff part) of electric current density term. Sum of these two is the total electric current (TildeJ); please see equations in system documentation (https://spectre-code.org/structForceFree_1_1System.html). TildeJ = DriftTildeJ + ParallelTildeJ is in the source term of the variable TildeE, and also in the flux term of another evolved variable TildeQ.
- `DriftTildeJ` is explicit, therefore can be handled as a ComputeTag.
- `ParallelTildeJ` is implicit, and needs to be computed and stored after implicit solve (for TildeE) is completed using the IMEX framework. The reason why we want to store this quantity is we need the total `(DriftTildeJ + ParallelTildeJ)` for computing the flux for TildeQ at the next time step.
- The tag `CurrentDensity` is not used for evolution, just a tag for observation purpose only.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ForceFree/Fluxes.cpp
+++ b/src/Evolution/Systems/ForceFree/Fluxes.cpp
@@ -37,10 +37,10 @@ void fluxes_impl(
     const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_b,
     const Scalar<DataVector>& tilde_psi, const Scalar<DataVector>& tilde_phi,
     const Scalar<DataVector>& tilde_q,
-    const tnsr::I<DataVector, 3, Frame::Inertial>& spatial_current_density,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& drift_tilde_j,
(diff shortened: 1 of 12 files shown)


yoonso0-0:
@wthrowe I hope what I described here helps you understand what I'm trying to do :) Please let me know if something's unclear or can be improved.

wthrowe:
Sorry, I still don't understand.  What does it mean for a variable to be implicit, particularly one that doesn't appear to be one of the evolved variables?  What are the actual implicit and explicit equations you are solving?

On a slightly different topic, why do you want to compute and store this, instead of just computing it where you use it?

PR #4671, opened 2023-01-31
Fix unused variable warning in H5/File.cpp

knelli2:
## Proposed changes

This happens because the variable is only used inside some header flags, so if the flag is false, then the variable never gets used.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/H5/File.cpp
+++ b/src/IO/H5/File.cpp
@@ -57,6 +57,8 @@ H5File<Access_t>::H5File(std::string file_name, bool append_to_file,
                // Ignore file locks when they are disabled on the file system
                true),
            "Failed to configure file locking.");
+#else
+  (void)use_file_locking;
 #endif
 
   file_id_ = file_exists


nilsvu:
This only happens when you're using an old HDF5 version, which is dangerous because you'll crash simulations by just reading output files. So I think we can leave the compiler warning as another hint that things are not quite right :) You should have gotten a CMake warning as well, didn't you?

Which system are you compiling on that doesn't have a recent HDF5 installed?

knelli2:
I was building on local. I did get the cmake warning. But if we aren't supporting the previous method of running with the older version of hdf5, then we should error instead of printing the warning. If we are supporting the previous method of running, then we need to allow for that as well in the c++ code.

nilsvu:
It's allowed, but prints a warning. I thought that's good :D

knelli2:
But I get a c++ error when I use the old method. If we still support the old method, we shouldn't get a c++ error.

nilsvu:
Ok, agreed. The HDF5 default on Ubuntu 20.04 is 1.10.4. We don't care about file locking much on a local machine, so it would be inconvenient to drop support for 1.10.4 just yet. This being a supported version we want to build without warnings. Therefore I agree with this change.

nilsvu (approved):

PR #4673, opened 2023-01-31
Use mpi-smp charm on CI

knelli2:
## Proposed changes

To be able to test that charm messages are working properly, we have to be able to run on multiple charm-nodes. This requires us to have a non-multicore build of charm on CI. I chose MPI because that's what we often use in our production environments.

To keep the size of the container down, I replaced the clang-multicore build of charm with a clang-mpi-smp build.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -321,7 +321,6 @@ jobs:
         build_type: [Debug, Release]
         use_pch: [ON]
         use_xsimd: [ON]
-        charm_name: ["Charm 7.0.0"]
         include:
           # Configure ccache sizes. The cache becomes ineffective if it can't
           # hold at least one full build. The sizes for the build configurations
@@ -380,11 +379,15 @@ jobs:
             BUILD_SHARED_LIBS: OFF
             use_xsimd: OFF
             MEMORY_ALLOCATOR: JEMALLOC
-          # Test compatibility with oldest supported CMake version
+          # Test compatibility with oldest supported CMake version and using an
+          # mpi version of charm
           - compiler: clang-13
             use_pch: ON
             build_type: Debug
             CMAKE_EXECUTABLE: /opt/local/cmake/3.18.2/bin/cmake
+            CHARM_ROOT: /work/charm_7_0_0/mpi-linux-x86_64-smp-clang
+            # MPI running tests is a bit slower than multicore
+            TEST_TIMEOUT_FACTOR: 3
 
     container:
       image: sxscollaboration/spectre:ci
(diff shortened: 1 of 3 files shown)


knelli2:
One of the CI builds is going to fail until the container is updated to have charm built with MPI.

wthrowe (dismissed):

.github/workflows/Tests.yaml
```diff
@@ -225,7 +225,7 @@ jobs:
           -D CMAKE_C_COMPILER=clang
           -D CMAKE_CXX_COMPILER=clang++
           -D CMAKE_Fortran_COMPILER=gfortran-9
-          -D CHARM_ROOT=/work/charm_7_0_0/multicore-linux-x86_64-clang
+          -D CHARM_ROOT=/work/charm_7_0_0/multicore-linux-x86_64-gcc
```
nilsvu: Use the env var `${CHARM_ROOT}` from the container here and in the other places. You can actually delete this line and the build will use `${CHARM_ROOT}` automatically.

.github/workflows/Tests.yaml
```diff
@@ -322,6 +322,7 @@ jobs:
         use_pch: [ON]
         use_xsimd: [ON]
         charm_name: ["Charm 7.0.0"]
+        charm_build: [/work/charm_7_0_0/multicore-linux-x86_64-gcc]
```
nilsvu: Delete this line, and also `charm_name` above. I believe this entry in the matrix is only needed if you want to _add_ builds to the matrix, not to modify an existing one. Instead, just set `CHARM_ROOT` for one of the builds and fall back to the default, like `BUILD_PYTHON_BINDINGS` below.

nilsvu (changes requested):

.github/workflows/Tests.yaml
nilsvu: Also, if you move this commit into a separate PR that we can merge first, people can already rebase on this and fewer PRs will get disrupted by the new Docker container.

knelli2: See #4677.

knelli2:
Squashed in changes. Now dependent on #4677.

knelli2:
Rebased on develop. No longer dependent on #4677

nilsvu (dismissed):
I can build and push the container on Monday, if no one has done it by then.

nilsdeppe:
LGTM. Just need to coordinate updating the container since it might cause existing PRs to fail CI (which is fine!)

nilsvu:
@knelli2 I pushed the new container and also added a commit that upgrades Blaze in the container. Please pull that commit to your local branch before force-pushing again (only in case you need to force push, we might just merge this PR as-is if CI passes).

knelli2:
Ok I'll wait until CI finishes to do anything. I'll check back later and if things are failing or something went wrong, I'll rebase on latest develop and force push

nilsvu:
Just to clarify: my commit that adds Blaze is _on this branch_, not on develop. So just `git pull` before rebasing / pushing.

knelli2:
Ahh yes ok, thanks for clarifying. I totally would've forgotten to do that lol

nilsvu:
@knelli2 the new Charm build wasn't found by CMake. Maybe the directory name just isn't correct?

nilsvu:
We encountered some issues with Blaze 3.8.2, so I pushed the container without the Blaze upgrade and restarted CI on this PR

nilsvu (approved):

PR #4674, opened 2023-01-31
Add new receive_data entry method to DistributedObject

knelli2:
## Proposed changes

This one takes a Charm++ message (pointer) and inserts it into the inbox.

Also, a corresponding Parallel::receive_data function was added. This can only be used on Array components at the moment.

You also need to properly register messages with charm or you'll get incorrect behavior so there's also code added to handle that automatically.

The parallel test for this can be run on one or two charm nodes.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -407,6 +407,13 @@ jobs:
         # We hash the content of the compiler rather than the location and mtime
         # to make sure the cache works across the different machines
         CCACHE_COMPILERCHECK: content
+        # These vars are to allow running charm with MPI as root inside the
+        # container which arises from using the "--privileged" flag just below.
+        OMPI_ALLOW_RUN_AS_ROOT: 1
+        OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
+      # See https://lists.cs.illinois.edu/lists/arc/charm/2018-10/msg00011.html
+      # for why we need this
+      options: --privileged
     steps:
       - name: Checkout repository
         uses: actions/checkout@v3.0.1
--- a/src/Evolution/DiscontinuousGalerkin/Messages/BoundaryMessage.hpp
+++ b/src/Evolution/DiscontinuousGalerkin/Messages/BoundaryMessage.hpp
@@ -11,6 +11,7 @@
 #include "Domain/Structure/ElementId.hpp"
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
 #include "Time/TimeStepId.hpp"
+#include "Utilities/GetOutput.hpp"
 #include "Utilities/PrettyType.hpp"
 
 #include "Evolution/DiscontinuousGalerkin/Messages/BoundaryMessage.decl.h"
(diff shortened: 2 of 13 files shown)


knelli2:
Although this isn't mutually exclusive with #4673, we can only test the multi-node functionality of this test if we have built charm with MPI. So the tests here should still pass with multicore charm, but we aren't testing the full Charm++ message capabilities. For that reason, I've adding the dependent label until #4673 is merged.

nilsdeppe (dismissed):

nilsdeppe:
The newly added test failed for a reason I do not understand

nilsdeppe:
Looks like you've gotten to the next CI failure. We probably want to set the 2 env vars listed in the error message in CMake _if_ the env has `SPECTRE_CONTAINER=0`. I don't know of a way to guard against singularity users, though. Another option, which I think is better, is to define the 2 vars in the CI yaml file

knelli2:
Just added the env vars to the CI file so this isn't in our container normally. If people really need to subvert the charm warnings (like I just did) in their container, then they should do so at their discretion.

.github/workflows/Tests.yaml
```diff
@@ -394,9 +394,6 @@ jobs:
       env:
         # We run unit tests with the following compiler flags:
         # - `-Werror`: Treat warnings as error.
-        # - `-march=x86-64`: Make sure we are building on a consistent
-        #   architecture so caching works. This is necessary because GitHub
-        #   may run the job on different hardware.
```
nilsvu: Are you sure this is incorrect? We do set `OVERRIDE_ARCH` below

knelli2: I guess technically the comment is correct. But, we don't set this as an `env` variable anymore. It's hard-coded into most of the cmake commands on CI (except for our architecture checks), so it doesn't seem like this is the correct place for the comment. Should I move it somewhere else?

src/Evolution/DiscontinuousGalerkin/Messages/BoundaryMessage.hpp
```diff
@@ -30,6 +31,11 @@ template <size_t Dim>
 struct BoundaryMessage : public CMessage_BoundaryMessage<Dim> {
   using base = CMessage_BoundaryMessage<Dim>;
 
+  // Needed for charm registration
+  static std::string name() {
+    return "BoundaryMessage<" + get_output(Dim) + ">";
```
nilsvu, line 36: can you just use pretty_type instead of defining your own name?

nilsdeppe: No, because then there's an implicit default. This _needs_ to be uniquely defined for Charm++, so much so that it needs to encode template parameters. `pretty_type` makes it too easy for there to be a subtle bug.

.github/workflows/Tests.yaml
nilsvu: I see. I think it would be good to keep the comment about caching and GitHub. I'd say just leave the comment here unless you can find a better place for it.

src/Evolution/DiscontinuousGalerkin/Messages/BoundaryMessage.hpp
knelli2: No I can't. That was my first instinct too. However, for charm registration in `CharmRegistration.hpp`, I specifically need the `< >` part of the name included which isn't when I use `pretty_type`.

.github/workflows/Tests.yaml
knelli2: Ok I'll remove that commit and just leave the comment in.

nilsdeppe (approved):

Issue #3545, opened 2021-09-23
Improvements for the BinaryCompactObject domain


nilsvu:
# Feature request:

From experiments with elliptic solves on our `BinaryCompactObject` domain I find that it currently needs about twice as many grid points as the SpEC initial data domain to reach the same accuracy (a bit worse actually). Here's some details, and suggestions for improvements that we can discuss:

- The **radial** distribution of grid points seems mostly fine, except the two inner cubes and the enveloping cube-to-sphere could be improved with a logarithmic or 1/r radial map.
- The **angular** direction has three problems:
  - [x] 1. Blocks vary a lot in angular size, since in +x and -x direction a single block covers the same angular size that is covered by two blocks in the y/z directions. This means to get a uniform domain we have to h-refine all blocks once in azimuthal direction, and the +/-x blocks also once in polar direction. This introduces many grid points that would be more effectively spent on p-refinement, and some unavoidable h-nonconforming boundaries. Edit: #3941 is a partial step towards this. Still needs to be activated for BBH domain. Edit2: #4934 implements this.
  - [x] 2. Equiangular maps are currently not implemented. Tests on the Cylinder showed that equiangular maps help significantly to increase angular resolution. Edit: done in #4676.
  - [x] 3. The transitions from spheres to cubes to enveloping cube to sphere introduce angular Jacobians that the code needs to resolve, and additional grid points. The two inner spheres-to-cubes seem fine in my tests since they do a lot of the work of resolving the physical features. But the enveloping cube-to-sphere introduces a layer of grid points that aren’t as effective. I think a big optimization would be to remove the enveloping cube altogether, and transition from the two inner cubes to a sphere right away. Edit: done in #3882.

Here's the slice through the current BBH domain for reference:

![Bildschirmfoto 2021-09-23 um 18 07 50](https://user-images.githubusercontent.com/746230/134543761-51d97363-67f0-4c48-bd6d-0c3bb6a3e0a4.png)

mar-celine:
Summary of a chat between me and @nilsleiffischer :

There is a path to address these issues that involves a new mapping for the Frustum blocks in the domain.
A: **angular ii**  can be addressed by adding equiangular mapping functionality to the Frustum maps. This is non-trivial because the domain requires that the inner face of a Frustum map (closest to the black holes) use the full [-1,1] equiangular map, while having the outer face use a half [0,1] or [-1,0] equiangular map, so as to be compatible with the half wedges in the wave zone.

[...]

PR #4676, opened 2023-01-31
Enable equiangular maps in BBH domain

nilsvu:
## Proposed changes

Another step towards #3545.

### Upgrade instructions

In input files using the `BinaryCompactObject` domain creator, add the `UseEquiangularMap: True` option.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/BinaryCompactObject.hpp
+++ b/src/Domain/Creators/BinaryCompactObject.hpp
@@ -597,7 +597,7 @@ class BinaryCompactObject : public DomainCreator<3> {
       const typename InitialRefinement::type& initial_refinement,
       const typename InitialGridPoints::type& initial_number_of_grid_points,
       bool use_equiangular_map = true, bool use_projective_map = true,
-      double frustum_sphericity = 0.0,
+      double frustum_sphericity = 1.0,
       const std::optional<double>& radius_enveloping_sphere = std::nullopt,
       CoordinateMaps::Distribution radial_distribution_outer_shell =
           CoordinateMaps::Distribution::Linear,
@@ -623,7 +623,7 @@ class BinaryCompactObject : public DomainCreator<3> {
       const typename InitialRefinement::type& initial_refinement,
       const typename InitialGridPoints::type& initial_number_of_grid_points,
       bool use_equiangular_map = true, bool use_projective_map = true,
-      double frustum_sphericity = 0.0,
+      double frustum_sphericity = 1.0,
       const std::optional<double>& radius_enveloping_sphere = std::nullopt,
       CoordinateMaps::Distribution radial_distribution_outer_shell =
           CoordinateMaps::Distribution::Linear,
--- a/src/Domain/Creators/BinaryCompactObject.hpp
+++ b/src/Domain/Creators/BinaryCompactObject.hpp
@@ -349,6 +349,13 @@ class BinaryCompactObject : public DomainCreator<3> {
         "domain. See main help text for details."};
   };
 
+  struct UseEquiangularMap {
+    using type = bool;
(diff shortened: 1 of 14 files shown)


src/Domain/DomainHelpers.cpp
```diff
@@ -717,7 +717,8 @@ std::vector<domain::CoordinateMaps::Frustum> frustum_coordinate_maps(
         use_equiangular_map,
         projective_scale_factor,
         false,
-        sphericity});
+        sphericity,
+        -1.});
```
mar-celine: Can have be `-1.0` and `1.0` on line 736 for consistency with the rest of the file. Also for clarity for the last argument (since the other arguments have variable names) can add a comment on line 702 (The frustums on the left are given a `-1.0` indicating a lower half equiangular map while the frustums on the right are given a `1.0` indicating an upper half equiangular map) (Ideally this should probably be an enum, but that was my own mistake!)

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -589,7 +596,8 @@ class BinaryCompactObject : public DomainCreator<3> {
       double outer_radius_domain,
       const typename InitialRefinement::type& initial_refinement,
       const typename InitialGridPoints::type& initial_number_of_grid_points,
-      bool use_projective_map = true, double frustum_sphericity = 0.0,
+      bool use_equiangular_map = true, bool use_projective_map = true,
+      double frustum_sphericity = 0.0,
```
mar-celine: Minor comment here and for line 626: Now that we have equiangular capability, and the frustum layer can be made spherical, should this default value for the `frustum_sphericity` be kept at `0.0`?  Maybe it can be changed to be `1.0` now unless there is a good reason to avoid using that value by default.

src/Utilities/CartesianProduct.hpp
```diff
@@ -0,0 +1,64 @@
+// Distributed under the MIT License.
```
mar-celine, line 1: This would be good as a separate commit because it is conceptually unrelated to the domain changes.

mar-celine (commented):
Looks good! Just some minor comments

src/Utilities/CartesianProduct.hpp
nilsvu: It is a separate commit :)

mar-celine (dismissed):
Looks good to me!

src/Utilities/CartesianProduct.hpp
```diff
@@ -0,0 +1,64 @@
...
+    // bind the first free variable in the lambda, and
+    // keep one free variable for each of the remaining ranges
+    detail::insert_tuples(
+        [&op, &it](InputIterator2... elems) { op(it, elems...); }, tail...);
+  }
+}
+
+}  // namespace detail
+
+// convert a tuple of ranges to the range of tuples representing the Cartesian
```
nilsdeppe: Why no Doxygen comment here?

src/Utilities/CartesianProduct.hpp
```diff
@@ -0,0 +1,64 @@
...
+namespace detail {
+
+// the lambda is fully bound with one element from each of the ranges
+template <class Op>
+void insert_tuples(Op op) {
+  // evaluating the lambda will insert the currently bound tuple
+  op();
+}
+
+// "peal off" the first range from the remaining tuple of ranges
```
nilsdeppe: `peal` -> `peel`, same below

nilsdeppe (changes requested):
You can squash and rebase immediately

nilsdeppe (approved):

PR #4677, opened 2023-02-01
Only use charm built with gcc on CI

knelli2:
On both local and Wheeler, I've been using charm built with gcc but SpECTRE built with clang and I've had no issues

## Proposed changes

Separated from #4673 so we don't interrupt CI when we update the docker container.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -225,7 +225,7 @@ jobs:
           -D CMAKE_C_COMPILER=clang
           -D CMAKE_CXX_COMPILER=clang++
           -D CMAKE_Fortran_COMPILER=gfortran-9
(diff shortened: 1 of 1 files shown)


wthrowe:
As an additional data point, I've also been using one charm build with different compilers without a problem on my dev machine.

nilsvu (approved):

nilsdeppe (approved):

nilsvu:
The two clang Release builds took a very long time because their cache was invalidated by this change. Not really sure what to do about that. I restarted them for now.

nilsvu:
They passed now, so I'm merging this and hoping for caching to kick in. The one clang release build took a scary 5.5h though, so we may have to do something about build times soon..

PR #4678, opened 2023-02-01
Add function to copy mutable creation items from DataBox into a TaggedTuple

kidder:
## Proposed changes

This allows data to be sent to newly created Elements from their parent or children during AMR

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/DataBox/DataBox.hpp
+++ b/src/DataStructures/DataBox/DataBox.hpp
@@ -22,6 +22,7 @@
 #include "DataStructures/DataBox/Subitems.hpp"
 #include "DataStructures/DataBox/TagName.hpp"
 #include "DataStructures/DataBox/TagTraits.hpp"
+#include "Parallel/Serialize.hpp"
 #include "Utilities/CleanupRoutine.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/ErrorHandling/StaticAssert.hpp"
@@ -256,6 +257,11 @@ class DataBox<tmpl::list<Tags...>> : private detail::Item<Tags>... {
   template <typename Tag>
   const auto& get() const;
 
+  /// \brief Copy the items with tags `TagsOfItemsToCopy` from the DataBox
+  /// into a TaggedTuple, should be called by the free function db::copy_items
+  template <typename... TagsOfItemsToCopy>
+  tuples::TaggedTuple<TagsOfItemsToCopy...> copy_items() const;
+
   /// Retrieve a mutable reference to the tag `Tag`, should be called
   /// by the free function db::get_mutable_reference
   template <typename Tag>
@@ -274,6 +280,10 @@ class DataBox<tmpl::list<Tags...>> : private detail::Item<Tags>... {
                     AddImmutableItemTagsList /*meta*/, Args&&... args);
 
  private:
+  template <typename CopiedItemsTagList, typename DbTagList>
+  // clang-tidy: redundant declaration
+  friend auto copy_items(const DataBox<DbTagList>& box);
+
   template <typename... MutateTags, typename TagList, typename Invokable,
             typename... Args>
   // clang-tidy: redundant declaration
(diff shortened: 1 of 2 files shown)


src/DataStructures/DataBox/DataBox.hpp
```diff
@@ -203,6 +204,12 @@ class DataBox<tmpl::list<Tags...>> : private detail::Item<Tags>... {
   using mutable_subitem_tags = tmpl::flatten<
       tmpl::transform<mutable_item_tags, db::Subitems<tmpl::_1>>>;
 
+  /// A list of all the immutable item tags used to create the DataBox
```
nilsvu: "mutable"?

src/DataStructures/DataBox/DataBox.hpp
```diff
@@ -244,26 +251,28 @@ class DataBox<tmpl::list<Tags...>> : private detail::Item<Tags>... {
   std::string print_types() const;
 
   /// Print the items
+
```
nilsvu: drop

nilsvu (changes requested):
This technically adds a cyclic dependency between Parallel and DataStructures. Should probably be resolved by moving PUP things out of Parallel. But I have another suggestion: I have needed a `get_items` function in some places in the elliptic code (search for "get_items", implemented as a lambda in some places). It gets multiple items from the DataBox and returns a tuple that can be assigned to structured bindings. How about we make that a DataBox function. Then I think you can build the TaggedTuple using get_items and serialize/deserialize where you need it.

src/DataStructures/DataBox/DataBox.hpp
kidder: duh 👍

kidder:
Yes, I agree we should probably make a Serialization library somewhere and put various things in it to break the dependency.
I don't think it matters whether copy_items returns a tuple or TaggedTuple, it will still need to call `deserialize(serialize(item))` in order to obtain a copy of a non-copyable type.

nilsvu:
Yes to the Serialization lib. We can place it in Utilities/Serialization/, as its own lib (because it's sort of an extension of Charm and has to live at the same low level as Utilities, below DataStructures).

I didn't suggest to switch TaggedTuple -> tuple. My suggestion is to make a `get_items` function with the same reference semantics as `get`. I think that would be very useful for things like this:

```cpp
const auto& [x, y] = db::get_items<Tags::X, Tags::Y>(box);
```

It replaces long lists of `db::get`. I thought it could also be useful for copying items, but maybe it's not enough (getting item types for deserialize can be annoying).

kidder:
Okay, so I'd prefer to refactor in a separate PR.   And yes, I see the usefulness of `get_items`, but I need to know the storage types in order to copy the items, so again, I'd prefer adding that in a future PR.

nilsvu (dismissed):
sounds good. please squash the changes above directly.

nilsvu (approved):

src/DataStructures/DataBox/DataBox.hpp
```diff
@@ -787,6 +807,47 @@ SPECTRE_ALWAYS_INLINE const auto& get(const DataBox<TagList>& box) {
...
+ * \ingroup DataBoxGroup
+ * \brief Copy the items from the DataBox into a TaggedTuple
+ *
+ * \return The objects corresponding to CopiedItemsTagList
+ *
+ * \note The tags in CopiedItemsTagList must be a subset of
+ * the mutable_item_creation_tags of the DataBox
+ */
+template <typename CopiedItemsTagList, typename DbTagList>
+SPECTRE_ALWAYS_INLINE auto copy_items(const DataBox<DbTagList>& box) {
```
nilsdeppe, line 845: do you only want a copy, or would a move one be useful too? If it's always a copy, that's fine, just asking :)

nilsdeppe (approved):
LGTM, one question

src/DataStructures/DataBox/DataBox.hpp
kidder: A move could also be useful and might be the right choice; I'll do that in a separate PR as moving an item will need to reset subitems and compute items that depend on the moved item...

PR #4680, opened 2023-02-01
Implicitly convert Numpy arrays to Tensors

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/Tensor/Python/Tensor.cpp
+++ b/src/DataStructures/Tensor/Python/Tensor.cpp
@@ -4,6 +4,7 @@
 #include "DataStructures/Tensor/Python/Tensor.hpp"
 
 #include <memory>
+#include <pybind11/numpy.h>
 #include <pybind11/operators.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
@@ -143,12 +144,16 @@ void bind_tensor_impl(py::module& m, const std::string& name) {  // NOLINT
                  throw std::runtime_error(
                      "Incompatible format: expected a double array.");
                }
-               if (info.ndim != 2) {
+               // 1D arrays are allowed to construct scalars. Higher-rank
+               // tensors require 2D arrays.
+               if (not(info.ndim == 2 or
+                       (info.ndim == 1 and TensorType::rank() == 0))) {
                  throw std::runtime_error(
                      "Tensor data is expected to be 2D with shape (size, "
-                     "num_points).");
+                     "num_points). Scalars can be 1D.");
                }
-               const auto size = static_cast<size_t>(info.shape[0]);
+               const size_t size =
+                   info.ndim == 1 ? 1 : static_cast<size_t>(info.shape[0]);
                if (size != TensorType::size()) {
                  throw std::runtime_error(
                      "This tensor type has " +
@@ -156,11 +161,15 @@ void bind_tensor_impl(py::module& m, const std::string& name) {  // NOLINT
                      " independent components, but data has first dimension " +
                      std::to_string(size) + ".");
                }
-               const auto num_points = static_cast<size_t>(info.shape[1]);
+               const auto num_points =
+                   static_cast<size_t>(info.shape[info.ndim == 1 ? 0 : 1]);
                auto data = static_cast<double*>(info.ptr);
                const std::array<size_t, 2> strides{
-                   {static_cast<size_t>(info.strides[0] / info.itemsize),
-                    static_cast<size_t>(info.strides[1] / info.itemsize)}};
+                   {info.ndim == 1
+                        ? 1
+                        : static_cast<size_t>(info.strides[0] / info.itemsize),
+                    static_cast<size_t>(info.strides[info.ndim == 1 ? 0 : 1] /
+                                        info.itemsize)}};
                if (copy) {
                  TensorType result{num_points};
                  for (size_t i = 0; i < size; ++i) {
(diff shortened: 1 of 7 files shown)


nilsvu:
@nikwit this seems to work well, so it's a start :)

nikwit:
Cool! Does that mean we can get rid of tensors and datavectors entirely on the python side?

nilsvu:
Bindings still return tensors and DataVectors, and I believe those can't implicitly convert to Numpy arrays. Edit: but we can treat them as such, because they should duck-type as Numpy arrays.

nikwit:
I still want to check if thats not possible somehow, that would be great!

nilsvu:
sure please try

nilsvu:
Maybe specializing `type_caster` has a chance of returning numpy arrays instead of tensors. I think that's how pybind11 implements Eigen types. Not sure if it's worth the effort to try, or if it's actually the right thing to do for tensors.

nikwit:
yes, it looks to be very similar to this: https://stackoverflow.com/questions/42645228/cast-numpy-array-to-from-custom-c-matrix-class-using-pybind11

I'm surprised it is this complicated, after all we only want it to call `np.asarray` whenever we return a tensor.

nilsvu:
Let's do the implicit conversion _from_ numpy arrays first, since it's a big improvement for convenience. Question: enable it for all Tensors, or only scalars and (co-)vectors? For higher-rank tensors (particularly with symmetries) the ordering of indices isn't obvious.

I'm not sure we want to return all tensors as numpy arrays. It's easy enough to convert in Python if you want to: `np.array(tensor)` does it. In many cases you can also duck-type the tensor or DataVector as an array. Tensor objects have useful features like access to component suffixes, symmetries, etc. And we'd have to get reference semantics right so the data isn't copied on the implicit conversion. Overall I see little benefit from implicitly converting tensors to Numpy arrays, but lots of benefit from implicitly converting arrays to tensors (scalars and vectors at least).

nilsdeppe:
I don't think we should go in this direction because the internal ordering of Tensor will not be intuitive for users. Scalar and Vector is fine, we allow this in C++ (array->Tensor, but explicitly), anything higher rank is just asking for trouble. If we end up realizing we should store things in a different order than we currently do, we would break a lot of Python code if they use numpy arrays for e.g. the metric.

nikwit:
I'm happy with the implicit conversion, definitely for scalars and vectors and also for higher ranks. If we decide to change the storage order in C++, then the only thing we should need to change is the buffer protocol so numpy knows how it is stored as far as I understand. On the python side there should not be any confusion about storage indices since we are always using arrays which have a clearly defined memory layout (and no symmetries).

@nilsvu do you want to update #4646 and get rid of the typing in python? If yes, in a follow up PR or directly?

nilsdeppe:
I guess I don't see how it would work for higher rank tensors, but I don't know enough about numpy. For a rank 2 dim 3 symmetric tensor `gamma[1, 2]` might be the underlying location, but then `gamma[2, 1]` would have to alias to `gamma[1, 2]` (or whatever the aliasing we do is). If that can be done in a way that is agnostic to Tensor storage order I guess general bindings can work, I just don't know if it's possible. How would I create such an object in a numpy arary? Can I create symmetric rank 2 numpy arrays? If not, then we simply cannot use numpy arrays for higher ranks because they are fundamentally different.

nilsvu:
I agree and edited the PR so only scalars and vectors can be implicitly converted. I think that's already a big win for convenience.

src/DataStructures/Tensor/Python/Tensor.cpp
```diff
@@ -143,24 +144,32 @@ void bind_tensor_impl(py::module& m, const std::string& name) {  // NOLINT
                  throw std::runtime_error(
                      "Incompatible format: expected a double array.");
                }
-               if (info.ndim != 2) {
+               // 1D arrays are allowed to construct scalars. Higher-rank
+               // tensors require 2D arrays.
+               if (not(info.ndim == 2 or
+                       (info.ndim == 1 and TensorType::rank() == 0))) {
                  throw std::runtime_error(
                      "Tensor data is expected to be 2D with shape (size, "
```
nikwit, line 152: the error message should tell you explicitly that you only allow scalars and vectors.

tests/Unit/DataStructures/Tensor/Python/Test_Tensor.py
```diff
@@ -35,11 +36,19 @@ def test_construct_from_list(self):
...
         for copy in [True, False]:
             coords = tnsr.I[DataVector, 3, Frame.Inertial](data, copy=copy)
             for i, (a, b) in enumerate(zip(coords, data)):
                 npt.assert_equal(a, b, f"Mismatch at index {i}")
             npt.assert_equal(np.array(coords), data)
+            # Construction of Scalar from 1D array
+            scalar = Scalar[DataVector](data_scalar, copy=copy)
+            npt.assert_equal(np.array(scalar), [data_scalar])
+        # Implicit conversion from Numpy array to Tensor
+        adm_mass_integrand(field=np.random.rand(4),
```
nikwit, line 60: can you also test Vectors and an error test for tensors?

nikwit (changes requested):
numpy does not support symmetric tensors afaik. My suggestion was to always have full numpy arrays on the python side without symmetries. When converted to C++, we could just check that it satisfies the symmetries of the tensor that the user wants to convert to and copy. But this is fine with me.

src/DataStructures/Tensor/Python/Tensor.cpp
nilsvu: You can, explicitly, construct higher-rank tensors from numpy arrays. The tensor gives you the index ordering (e.g. with `component_suffix`), so you know how to order the array. Only the implicit conversion is restricted to scalars and vectors. I edited the error message here a bit to include that scalars can be 1D numpy arrays.

tests/Unit/DataStructures/Tensor/Python/Test_Tensor.py
nilsvu: done

nilsdeppe (commented):
I have no additional comments

nilsvu:
@nikwit please check the fixup

nikwit (dismissed):

nilsdeppe (approved):

PR #4683, opened 2023-02-02
Add input file for head-on BNS initial data

nilsvu:
## Proposed changes

Generates this matter & gravity configuration:

Bildschirm­foto 2023-02-01 um 10 53 21 PM

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/InputFiles/Xcts/HeadOnBns.yaml
+++ b/tests/InputFiles/Xcts/HeadOnBns.yaml
@@ -1,0 +1,196 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+# Executable: SolveXcts
+# Check: parse
+# Timeout: 30
+
+ResourceInfo:
+  AvoidGlobalProc0: false
+  Singletons: Auto
+
+Background: &background
+  Binary:
+    XCoords: [&x_left -20., &x_right 20.]
+    ObjectLeft: &star_left
+      TovStar:
+        # Values taken from Sec. V.C and Fig. 8 in
+        # https://arxiv.org/abs/1708.07358.
+        # This central density corresponds to h = 1.2, or about 5e14 g/cm^3.
+        CentralDensity: 8.087415253997405e-4
+        EquationOfState:
+          PolytropicFluid:
+            PolytropicConstant: 123.6489
+            PolytropicExponent: 2.
+        Coordinates: Isotropic
+    ObjectRight: *star_left
+    AngularVelocity: 0.
+    Expansion: 0.
+    FalloffWidths: None
+
+InitialGuess: *background
+
+DomainCreator:
+  BinaryCompactObject:
+    ObjectA:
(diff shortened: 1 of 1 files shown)


tests/InputFiles/Xcts/HeadOnBns.yaml
```diff
@@ -0,0 +1,192 @@
...
+ResourceInfo:
+  AvoidGlobalProc0: false
+  Singletons: Auto
+
+Background: &background
+  Binary:
+    XCoords: [&x_left -20., &x_right 20.]
+    ObjectA: &star_right
+      TovStar:
+        CentralDensity: 0.0008087415253997405  # h = 1.2
```
nilsdeppe: Where does this and the K come from? In single stars K=100 is standard and gives a maximum mass of the NS that's ~2.2 M_sun. Also, can you use scientific notation here? I find it rather difficult to parse all the leading zeroes. 8.08e-4 is much easier to read. Also, having the central density in terms of g/cm^3 in the comment would be more helpful than the specific enthalpy.

nilsdeppe (changes requested):
You can squash

tests/InputFiles/Xcts/HeadOnBns.yaml
nilsvu: Added the reference where these numbers are from, and the other things. Thanks for your review!

nilsdeppe:
RIP, all CI failed...

tests/InputFiles/Xcts/HeadOnBns.yaml
```diff
@@ -0,0 +1,195 @@
...
+  AvoidGlobalProc0: false
+  Singletons: Auto
+
+Background: &background
+  Binary:
+    XCoords: [&x_left -20., &x_right 20.]
+    ObjectA: &star_right
+      TovStar:
+        # Values taken from Sec. V.C and Fig. 8 in
+        # https://arxiv.org/abs/1708.07358.
```
nilsdeppe, line 18: I think we should use the numbers from https://arxiv.org/pdf/2010.14518.pdf instead. It's more recent and also evolves the binaries, and allows for a simple thermal component.

nilsvu:
Conflict with #4684 that was merged earlier

tests/InputFiles/Xcts/HeadOnBns.yaml
nilsvu: You can replace this with an EOS of your choice when generating ID :) happy to put another one here as the example. The paper you link doesn't
list the central density though, or maybe I've missed it. Also, is the thermal component implemented?

nilsdeppe: Okay, let's just leave it as-is. Like you said, we can adjust later :)

nilsdeppe (approved):

PR #4686, opened 2023-02-03
Add object centers tags

knelli2:
## Proposed changes

The basic control system control errors (Expansion, Rotation, Translation) should be agnostic to where the reference centers of the objects come from. All they need to know is what the center actually is. Currently, all control errors hard-code getting the reference centers from the excision spheres. This PR adds some simple and base tags to abstract this away. I only added tags for reference centers that come from excision spheres right now, but adding tags for the centers of BNSs should be simple in the future.

I also simplified the control system initialization action to a mutator.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

Please forgive the names of these tags. I know this is still an open discussion/debate in #4529, but I chose these names just to be consistent with the names we currently have.


src/Domain/Tags.hpp
```diff
@@ -122,6 +122,48 @@ struct ElementMap : db::SimpleTag {
...
 };
 
+/// @{
+/// \ingroup DataBoxTagsGroup
+/// \ingroup ComputationalDomainGroup
+/// Base tags to retrieve the centers of objects in the domain.
+///
+/// They are separate tags because we may have one or two objects in the domain.
+struct GridCenterObjectA : db::BaseTag {};
+struct GridCenterObjectB : db::BaseTag {};
```
wthrowe: There is a lot of duplicate code in this commit.  Can the tags be templated on the objects so we don't need two of everything?

wthrowe (changes requested):

src/Domain/Tags.hpp
knelli2: I did consider this. We already have an object label in the apparent horizons library `::ah::ObjectLabel::{A,B}`. So I'd have to move that to the Domain library now to use it here. This was the conclusion reached in #3561. Does that sound reasonable?

nilsdeppe: I can't speak for Will, but this sounds like a good plan to me!

wthrowe: Yes, now that it's become useful let's do the move.

knelli2:
Ok added another commit that moves the ObjectLabel from the ApparenHorizons library to the Domain library.

wthrowe (dismissed):
Looks good once you fix the compilation error.

knelli2:
Oops! Missed that one. Fixed now.

wthrowe (dismissed):

src/Domain/Tags.hpp
```diff
@@ -122,6 +123,33 @@ struct ElementMap : db::SimpleTag {
   using type = ::ElementMap<VolumeDim, TargetFrame>;
 };
 
+/// \ingroup DataBoxTagsGroup
+/// \ingroup ComputationalDomainGroup
+/// Base tag to retrieve the centers of objects in the domain corresponding to
+/// the `ObjectLabel`.
+template <ObjectLabel Label>
+struct GridCenter : db::BaseTag {};
```
nilsvu: Please define these tags in a new file in Domain/Tags/. The Tags.hpp file has become quite cluttered and we've moved to a Tags/ subdirectory. Helps with recompilations too.

src/Domain/Tags.hpp
```diff
@@ -122,6 +123,33 @@ struct ElementMap : db::SimpleTag {
...
+/// \ingroup ComputationalDomainGroup
+/// The center of the excision sphere for the given object.
+///
+/// Even though this can easily be retrieved from the domain, we add it as its
+/// own tag so we can access it through the base tag. This way, other things
+/// (like the control system) can grab the center and be agnostic to what the
+/// object actually is.
+template <ObjectLabel Label>
+struct ExcisionGridCenter : GridCenter<Label>, db::SimpleTag {
+  using type = DataVector;
```
nilsvu: Can we store this as `tnsr<double, Dim, Grid>`? DataVector is really not the right data type for this.

nilsvu (changes requested):

src/Domain/Tags.hpp
knelli2: Yes we can, but the control system needs a `DataVector` in order to do math with so I'll end up copying it then. Not really sure which is preferred.

nilsvu: What sort of math? I'd rather use the right data type here and then make sure we can do the math with it that we need.

knelli2: So at the end of the day, I need a `DataVector` to update the functions of time. Currently, the AH centers come as `std::array` from the Strahlkorper which I convert to a `DataVector` before I insert it into the control system DataBox. So the most natural type for the grid centers was a `DataVector`. I have to be be able to add, subtract, dot, and cross these quantities.

So I guess I could make everything take a `tnsr` and do math with those and then copy it into a `DataVector` at the very end, but that requires a lot more changes and I think is beyond the scope of this PR.

nilsvu: Actually I think everything you need is already handled by #4666. There are more places that could switch to `tnsr`, but as you say, that's beyond the scope of this change.

knelli2: Yes #4666 will make the tag a `tnsr`, but I'll still have to copy it into a `DataVector` inside the control error. Once this is merged I'll consider changing how we store the AH center to also be a `tnsr` so we can do math with a "proper" data type for what we are doing.

nilsvu: Let's get #4666 merged quickly, it's only a rebase away. Then the tag you're adding here has the correct type, and you can consider changing types in other places later.

src/Domain/Tags.hpp
nilsvu: ping on this (move to own file). Maybe name `ObjectCenter` or `BinaryObjectCenter`? That the coords are in the grid frame is now implied by the data type. Same for the simple tag (maybe just `ExcisionCenter`?)

tests/Unit/Domain/Test_Tags.cpp
```diff
@@ -60,6 +68,93 @@ void test_simple_tags() {
...
+      "GridCenter");
+  TestHelpers::db::test_base_tag<Tags::GridCenter<ObjectLabel::B>>(
+      "GridCenter");
+  TestHelpers::db::test_simple_tag<Tags::ExcisionGridCenter<ObjectLabel::A>>(
+      "GridCenterObjectA");
+  TestHelpers::db::test_simple_tag<Tags::ExcisionGridCenter<ObjectLabel::B>>(
+      "GridCenterObjectB");
+
+  const auto domain_creator =
+      TestHelpers::test_option_tag<domain::OptionTags::DomainCreator<3>,
```
nilsvu: Just use the constructor, no need to go through option creation

tests/Unit/Domain/Test_Tags.cpp
```diff
@@ -60,6 +68,93 @@ void test_simple_tags() {
...
+          domain_creator);
+  const auto grid_center_B =
+      Tags::ExcisionGridCenter<ObjectLabel::B>::create_from_options(
+          domain_creator);
+
+  CHECK(grid_center_A == tnsr::I<double, 3, Frame::Grid>{{8.0, 0.0, 0.0}});
+  CHECK(grid_center_B == tnsr::I<double, 3, Frame::Grid>{{-5.5, 0.0, 0.0}});
+
+  const auto creator_no_excision =
+      TestHelpers::test_option_tag<domain::OptionTags::DomainCreator<3>,
```
nilsvu: just use the constructor

tests/Unit/Domain/Test_Tags.cpp
```diff
@@ -60,6 +68,93 @@ void test_simple_tags() {
...
+          "  InitialRefinement: [0,0,0]\n"
+          "  InitialGridPoints: [2,2,2]\n"
+          "  TimeDependence: None\n");
+
+  CHECK_THROWS_WITH(
+      Tags::ExcisionGridCenter<ObjectLabel::A>::create_from_options(
+          creator_no_excision),
+      Catch::Contains(" is not in the domains excision spheres but is needed "
+                      "to generate the ExcisionGridCenter"));
+  CHECK_THROWS_WITH(
```
nilsvu: delete, testing for one object is enough

tests/Unit/ControlSystem/Systems/Test_Rotation.cpp
```diff
@@ -101,9 +101,12 @@ void test_rotation_control_system(const bool newtonian) {
 
   // Setup runner and all components
   using MockRuntimeSystem = ActionTesting::MockRuntimeSystem<metavars>;
-  MockRuntimeSystem runner{{"DummyFileName", std::move(domain), 4},
-                           {std::move(initial_functions_of_time),
-                            std::move(initial_measurement_timescales)}};
+  MockRuntimeSystem runner{
+      {"DummyFileName", std::move(domain), 4,
+       tnsr::I<double, 3, Frame::Grid>{{0.5 * initial_separation, 0.0, 0.0}},
+       tnsr::I<double, 3, Frame::Grid>{{-0.5 * initial_separation, 0.0, 0.0}}},
```
nilsvu: get from domain? here and in the other tests

nilsvu (changes requested):

src/ControlSystem/Actions/Initialization.hpp
```diff
@@ -25,11 +27,13 @@ namespace Actions {
  * GlobalCache:
  * - Uses:
  *   - `control_system::Tags::MeasurementTimescales`
+ *   - `domain::Tags::ExcisionGridCenter<domain::ObjectLabel::A>`
+ *   - `domain::Tags::ExcisionGridCenter<domain::ObjectLabel::B>`
```
nilsvu: Update names

nilsvu (dismissed):

nilsvu (approved):

tests/Unit/Domain/Test_Tags.cpp
```diff
@@ -60,6 +69,48 @@ void test_simple_tags() {
...
+  using Object = domain::creators::BinaryCompactObject::Object;
+  using Refinement =
+      domain::creators::BinaryCompactObject::InitialRefinement::type;
+  using GridPoints =
+      domain::creators::BinaryCompactObject::InitialGridPoints::type;
+
+  const std::unique_ptr<DomainCreator<3>> domain_creator =
+      std::make_unique<domain::creators::BinaryCompactObject>(
+          Object{0.2, 5.0, 8.0, true, true}, Object{0.6, 4.0, -5.5, true, true},
+          100.0, 500.0, Refinement{1}, GridPoints{5});
```
nilsvu: just put `1_st` and `5_st` here, and delete the type aliases. that should fix the compiler error

nilsvu (changes requested):

tests/Unit/Domain/Test_Tags.cpp
knelli2: Ugh clang-10 tricking me into thinking I don't need `1_st`... 🙄

nilsvu (approved):

PR #4687, opened 2023-02-03
Add executable to dump EOS for RotNS

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/PointwiseFunctions/Hydro/Units.hpp
```diff
@@ -0,0 +1,41 @@
...
+/// `mass_cgs = mass_geometric * mass_unit`
+constexpr double mass_unit = 1.0 / 5.02765209e-34;
+/// `length_cgs = length_geometric * length_unit`
+constexpr double length_unit = 1.0 / 6.77140812e-6;
+/// `time_cgs = time_geometric * time_unit`
+constexpr double time_unit = 1.0 / 2.03001708e5;
+/// Baryon mass, used to go from number density to rest mass
+/// density. I.e. `rho_cgs = n_cgs * baryon_mass`, where `n_gcs` is the number
+/// density in CGS units.
+constexpr double baryon_mass = 1.659e-24;
```
ermost: Is it possible to use a flexible baryon mass? Not all EOSs use the same one, so that quantity could be misinterpreted. (I just recently had that issue confusing atomic mass unit for neutron mass...)

src/Executables/DumpEquationOfStateForRotNS/DumpEquationOfStateForRotNS.cpp
```diff
@@ -0,0 +1,181 @@
...
+        eos.pressure_from_density(rest_mass_density_geometric);
+    const Scalar<double> specific_internal_energy_geometric =
+        eos.specific_internal_energy_from_density(rest_mass_density_geometric);
+    const Scalar<double> total_energy_density_geometric{
+        get(rest_mass_density_geometric) *
+        (1.0 + get(specific_internal_energy_geometric))};
+
+    const double total_energy_density_cgs =
+        get(total_energy_density_geometric) *
+        hydro::units::cgs::rest_mass_density_unit;
```
ermost: Maybe add a comment that the table stores energy_density/c^2, since for energy_density in cgs you would need to multiply by that ( i.e. rest-mass part is rho c^2)

docs/GroupDefs.hpp
```diff
@@ -442,6 +442,11 @@
  * <td> \ref ParallelInfoExecutablePage "ParallelInfo" </td>
  * <td> Executable for checking number of nodes, cores, etc.</td>
  * </tr>
+ * <tr>
+ * <td> DumpEquationOfStateForRotNS </td>
```
nilsvu: Use "export" as a nicer word for "dump", and consistent with ExportCoordinates?

src/PointwiseFunctions/Hydro/Units.hpp
nilsdeppe: Hmm, do you think we should just not define this globally then? Since the other quantities are accepted physical constants they seem fine to have globally, but since the baryon mass is a choice people tweak, maybe that we should just leave to the individual EOS?

docs/GroupDefs.hpp
nilsdeppe: Sure!

src/PointwiseFunctions/Hydro/Units.hpp
ermost: Ideally, you would need this to be stored in two places. In the EOS, to know what baryon mass its compatible with. That would give you the correct number density (schematically n_baryon = rho/eos::baryon_mass;)
However, you ID code might assume that you are using a different baryon mass. e.g. Lorene assumes that you always have atomic mass units. So you might want to rescale things in the exporter, to make sure that you get the same
rho = n_export * exporter::baryon_mass density.

nilsdeppe: Okay, I will hard-code the mass into the exporter and file an issue that we need to figure out how to do baryon mass for the EOS. Is it that each EOS would define a member function `baryon_mass`? I guess for a polytrope this could depend on `K`? Or is this something we can hard-code to e.g. the value here? Then I guess for tables we would need to get it from compose, right?

Does that sound like a reasonable plan?

ermost: Sounds great! :)  Also let me add, the only time this matters as far as I'm aware is when using neutrino interactions.
So for a polytrope you probably wouldn't need this, or just define a baryon mass, as long as it's used consistently that should be fine. The value here would probably be perfect for everything that doesn't have its on baryon mass.

Yep, most tables specify it, or in the case of compose the internal energy comes just rescaled to the neutron mass. (That value is currently hardcoded in the table import PR, too).

nilsdeppe: Okay, great! Thank you :D

nilsdeppe:
I've made the changes, rebased, and opened #4694

ermost:
Thanks! :)  That looks good 👍

src/Executables/ExportEquationOfStateForRotNS/ExportEquationOfStateForRotNS.cpp
```diff
@@ -0,0 +1,190 @@
...
+#include "Options/ParseOptions.hpp"
+#include "Parallel/Printf.hpp"
+#include "PointwiseFunctions/Hydro/EquationsOfState/EquationOfState.hpp"
+#include "PointwiseFunctions/Hydro/Tags.hpp"
+#include "PointwiseFunctions/Hydro/Units.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/FileSystem.hpp"
+#include "Utilities/TMPL.hpp"
+
+// Charm looks for this function but since we build without a main function or
```
nilsvu, line 25: There is a main function here. The comment probably still applies because we link in Charm but don't use it?

src/Executables/ExportEquationOfStateForRotNS/ExportEquationOfStateForRotNS.cpp
```diff
@@ -0,0 +1,190 @@
...
+    const size_t number_of_log10_number_density_points_for_dump,
+    const std::string& output_file_name,
+    const double lower_bound_rest_mass_density_cgs,
+    const double upper_bound_rest_mass_density_cgs) {
+  using std::log10;
+  using std::pow;
+  // Baryon mass, used to go from number density to rest mass
+  // density. I.e. `rho_cgs = n_cgs * baryon_mass`, where `n_gcs` is the number
+  // density in CGS units. This is the baryon mass that RotNS uses. This
+  // might be different from the baryon mass that the EoS uses.
```
nilsvu, line 41: Link issue you created here?

src/Executables/ExportEquationOfStateForRotNS/ExportEquationOfStateForRotNS.cpp
```diff
@@ -0,0 +1,190 @@
...
+       ++log10_number_density_index) {
+    using std::pow;
+    const double number_density_cgs =
+        pow(10.0, log10_lower_bound_number_density_cgs +
+                      static_cast<double>(log10_number_density_index) *
+                          delta_log_number_density_cgs);
+    const double rest_mass_density_cgs =
+        number_density_cgs * baryon_mass_of_rotns_cgs;
+
+    // Note: we will want to add the baryon mass to our EOS interface.
```
nilsvu, line 72: Or link issue here

src/PointwiseFunctions/Hydro/Units.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * would be nice, but is a non-trivial amount of work.
+ */
+namespace units {
+/*!
+ * \brief Entities for converting between geometric units where
+ * \f$G=c=M_{\odot}=1\f$ and CGS units.
+ *
+ * Note: the baryon mass is not implemented here since it depends on what the
+ * specific place assumes. E.g. different EoS can assume different baryon
+ * masses.
```
nilsvu: Or here

tests/InputFiles/ExportEquationOfStateForRotNS/ExportEosForRotNS.yaml
```diff
@@ -0,0 +1,56 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+# Executable: ExportEquationOfStateForRotNS
+# Check: parse;execute
+# ExpectedOutput:
+#   EOS.GN
+
+EquationOfState:
+  Enthalpy(Enthalpy(Enthalpy(Spectral))):
```
nilsvu, line 14: What does this mean? :D

tests/InputFiles/ExportEquationOfStateForRotNS/ExportEosForRotNS.yaml
```diff
@@ -0,0 +1,56 @@
...
+    ReferenceDensity: 0.001822038675716006
+    MinimumDensity: 0.001822038675716006
+    MaximumDensity: 0.0034
+    TrigScaling: 0.0
+    PolynomialCoefficients:
+      [1.2730598378768818, 1.2730598378768818, 0.6365299189384409,
```
nilsvu, line 20: Could you add a few notes here where you got the numbers from that you put into this input file? Maybe link a paper or just say "these are some arbitrary numbers", just so we remember where they come from.

nilsvu (changes requested):

src/Executables/ExportEquationOfStateForRotNS/ExportEquationOfStateForRotNS.cpp
nilsdeppe: Yes, this is a charm function we need to define. I've added a bit of wording that hopefully makes this clearer.

nilsdeppe:
Rebased and squashed in links to the issue and paper that uses the EOS.

nilsvu (approved):
LGTM!

PR #3977, opened 2022-05-16
CCE preliminaries for GH+CCE executable

kidder:
## Proposed changes

This is a rebased version of #2323 that contains all of the commits except the one for the actual executable.
I added one commit that factored the metavariables for CharacteristicExtract into two structs, one of which contains information that will be common to the stand-alone CCE and the GH+CCE executables

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4690, opened 2023-02-03
Add action that does consistency/sanity checks on time stepper history

knelli2:
## Proposed changes

While working with CCM, I've been burned a couple times now by self-start not completing properly. There is nothing wrong with the self start code itself. However, if there is a deadlock in the `StepActions`, then because of quiescence detection, the self-start phase will finish (without error) and the algorithm continues to the evolution with incorrect time stepper history. This can cause all sorts of problems in the evolution and make things hard to debug.

This PR adds a new action that does some consistency/sanity checks on the history data so as to avoid this. Because quiescence can sometimes happen during self start (which we don't want), we can't put this action in the self-start phase itself because the action won't get run. We also don't want to put this in the Evolve phase because this check only needs to be done once. So I added a new phase `CheckTimeStepperHistory` that goes between self-start and evolve to do these checks.

I also added this new phase to most of our evolution executables.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/Burgers/EvolveBurgers.hpp
+++ b/src/Evolution/Executables/Burgers/EvolveBurgers.hpp
@@ -325,6 +325,9 @@ struct EvolutionMetavars {
               Parallel::Phase::InitializeTimeStepperHistory,
               SelfStart::self_start_procedure<step_actions, system>>,
 
+          Parallel::PhaseActions<Parallel::Phase::CheckTimeStepperHistory,
+                                 SelfStart::check_self_start_actions>,
+
           Parallel::PhaseActions<
               Parallel::Phase::Evolve,
               tmpl::list<Actions::RunEventsAndTriggers, Actions::ChangeSlabSize,
@@ -348,10 +351,11 @@ struct EvolutionMetavars {
       "The analytic solution is: Linear\n"
       "The numerical flux is:    LocalLaxFriedrichs\n"};
 
-  static constexpr std::array<Parallel::Phase, 5> default_phase_order{
+  static constexpr std::array<Parallel::Phase, 6> default_phase_order{
       {Parallel::Phase::Initialization,
        Parallel::Phase::InitializeTimeStepperHistory, Parallel::Phase::Register,
-       Parallel::Phase::Evolve, Parallel::Phase::Exit}};
+       Parallel::Phase::CheckTimeStepperHistory, Parallel::Phase::Evolve,
+       Parallel::Phase::Exit}};
 
   // NOLINTNEXTLINE(google-runtime-references)
   void pup(PUP::er& /*p*/) {}
--- a/src/Evolution/Executables/Cce/CharacteristicExtract.hpp
+++ b/src/Evolution/Executables/Cce/CharacteristicExtract.hpp
@@ -81,9 +81,10 @@ struct EvolutionMetavars : CharacteristicExtractDefaults<false> {
       "Perform Cauchy Characteristic Extraction using .h5 input data.\n"
       "Uses regularity-preserving formulation."};
 
-  static constexpr std::array<Parallel::Phase, 4> default_phase_order{
+  static constexpr std::array<Parallel::Phase, 5> default_phase_order{
       {Parallel::Phase::Initialization,
-       Parallel::Phase::InitializeTimeStepperHistory, Parallel::Phase::Evolve,
+       Parallel::Phase::InitializeTimeStepperHistory,
+       Parallel::Phase::CheckTimeStepperHistory, Parallel::Phase::Evolve,
        Parallel::Phase::Exit}};
 
   // NOLINTNEXTLINE(google-runtime-references)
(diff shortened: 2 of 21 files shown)


wthrowe:
Why is this not caught by our general deadlock detection?

knelli2:
The general deadlock detection only runs during the `Exit` phase. So during self-start we quiesce incorrectly, meaning that the parallel component didn't terminate properly. However, the algorithm doesn't know that this was an "improper" quiescence and since we use quiescence detection to switch phases, the algorithm goes to the next phase (`Evolve` usually). Here, it starts evolving because a TimeStepId and the history are in the DataBox and the time steppers can handle any number of history values. But the evolution is wrong because one of those two weren't set properly from the (silently) failed self-start phase. And the fact that the component didn't terminate properly is wiped away by the `Evolve` phase, thus leading to a silent error.

My change is effectively checking that we didn't quiesce incorrectly in the self-start phase. Maybe we could check for deadlock after every phase? However, my changes give a more informed error message than a general deadlock detection. Thoughts?

wthrowe:
Isn't this deadlock detection code?  It runs at the start of every phase.
```c++
template <typename ParallelComponent, typename... PhaseDepActionListsPack>
void DistributedObject<ParallelComponent,
                       tmpl::list<PhaseDepActionListsPack...>>::
    start_phase(const Parallel::Phase next_phase) {
  try {
    // terminate should be true since we exited a phase previously.
    if (not get_terminate() and not halt_algorithm_until_next_phase_) {
      ERROR(
          "An algorithm must always be set to terminate at the beginning of a "
          "phase. Since this is not the case the previous phase did not end "
          "correctly. The previous phase is: "
          << phase_ << " and the next phase is: " << next_phase
          << ", The termination flag is: " << get_terminate()
          << ", and the halt flag is: " << halt_algorithm_until_next_phase_);
    }
```

knelli2:
I found the offending commit: https://github.com/sxs-collaboration/spectre/pull/3977/commits/4784b2539b05225a95c92bdafe4582332e56fe71

The CCE action list was explicitly set to terminate cleanly while it is waiting for data by using `Pause` rather than `Retry` in the action that waits for data from CCE. In the evolution this is perfectly fine, but for self-start we want CCE to error if it didn't finish its algorithm (it *has* to receive all its data or else it can't self-start properly).

There are a few options here I think.
1. Add a template bool to `ReceiveWorldtubeData` called `TerminateUponWaiting`. If this bool is true, then we `Pause`. If it's false, then we `Retry`. We use the true version during `Evolve` but the false version during self-start.
2. Use this PR to catch things.
3. Do both 1 and 2.

nilsdeppe:
@wthrowe would you mind taking a look at this again?

wthrowe:
I think you could get the same effect by checking that `::SelfStart::is_self_starting(history.back()) or history.integration_order() == order_` in `AdamsBashforth::clean_history`.

knelli2:
I'd like this to be general though. I don't want to just restrict it to AB. These changes were really helpful in ensuring CCE was self-started properly and I imagine ensuring things self-started properly for other systems would be useful as well

wthrowe:
AB is the only one that uses the self-start code.  The others only need the t=0 RHS evaluation to start.

knelli2:
Yes that is true.

I guess my main question is: do you have an objection to these changes? These changes are completely general and will catch any issues with self-start for any time stepper we may use (even if most don't actually do self-start). This also happens before the Evolve phase so we catch things early if there are any issues. In my opinion, the blanket solution seems the simplest way to go because it's a catch-all and is not time-stepper dependent.

wthrowe:
You seem to have a strange definition of "simplest" if the 200+ lines adding a new phase are simpler than an ERROR.  That's actually my main objection here: that this is hugely complicated for the task you are trying to do.

knelli2:
The task I'm trying to do is guarantee we have self started properly for all executables for all different time steppers we could possibly use. ~200 lines doesn't seem like that much to achieve this.

wthrowe:
I'm not convinced your formula is correct for all time steppers.  It's just empirically how our current ones behave.

For example, I have not worked out the expected state for predictor-corrector methods, but I suspect the answer depends on how you choose the order of the predictor.

knelli2:
Ok that's fair, I know nothing about predictor corrector time steppers and this PR was based on only the time steppers we currently have on develop (at least the ones that I'm aware of).

However, my goal still stands: I'd like it to be impossible to get to the Evolve phase unless we have verified that self-start has succeeded and whatever history we need is there. For all executables. For all time steppers. This requires some kind of sanity/safety check before Evolve. Either at the end of self-start, or in a new phase itself, doesn't really matter. This was immensely helpful for CCM and it may potentially be helpful for predictor-corrector time steppers (I truly don't know, I'm just speculating).

So, how would you recommend going about this?

wthrowe:
The only code that can evaluate the correctness of history state is the time steppers, because they are what manage that state.  You can add a check there and call it from your thing.

And you can't put it at the end of the SelfStart phase because the problem you want to detect is that phase never reaching its end.

I still don't see why this needs a new phase and can't wait until the first call to the time stepper in the Evolve phase.  The only other thing that makes non-trivial use of the history is `RunEventsAndDenseTriggers`, which can be handled by doing the check in the dense output function as well as `update_u`/cleanup.

knelli2:
I'd like this in, but don't currently have the time to work on it

PR #4691, opened 2023-02-03
Add an AMR refinement criterion useful for testing

kidder:
## Proposed changes

The new criterion drives each Element in the Domain to a target resolution and refinement level.  Once it reaches the target, it will perform a given a decision that will cause the grid to oscillate between the target grid and a grid which differs by either one in resolution or refinement level.

This allows testing AMR actions in a controlled manner to make sure the mechanics of transferring data is correct.

In addition to testing, this criteria may be needed in order to drive the grid to desired initial refinement levels in order to bypass issues with Charm++

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Amr/Criteria/DriveToTarget.hpp
+++ b/src/ParallelAlgorithms/Amr/Criteria/DriveToTarget.hpp
@@ -1,0 +1,109 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <pup.h>
+
+#include "Domain/Amr/Flag.hpp"
+#include "Domain/Tags.hpp"
+#include "ParallelAlgorithms/Amr/Criteria/Criterion.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+template <size_t>
+class ElementId;
+template <size_t>
+class Mesh;
+/// \endcond
+
+namespace amr::Criteria {
+/*!
+ * \brief Refine the grid towards the target number of grid points and
+ * refinement levels in each dimension and then oscillate about the target.
+ *
+ * \details If the grid is at neither target in a given dimension, the
+ * flag chosen will be in the priority order Split,
+ * IncreaseResolution, DecreaseResolution, Join.
+ *
+ * \note To remain at the target, set the OscillationAtTarget Flags to
+ * DoNothing.
+ *
+ * \note This criterion is primarily for testing the mechanics of refinement.
+ */
+template <size_t Dim>
+class DriveToTarget : public Criterion {
+ public:
+  /// The target number of grid point in each dimension
+  struct TargetNumberOfGridPoints {
+    using type = std::array<size_t, Dim>;
+    static constexpr Options::String help = {
+        "The target number of grid points in each dimension."};
+  };
+
+  /// The target refinement level in each dimension
+  struct TargetRefinementLevels {
+    using type = std::array<size_t, Dim>;
+    static constexpr Options::String help = {
+        "The target refinement level in each dimension."};
+  };
+
+  /// The AMR flags chosen when the target number of grid points and refinement
+  /// levels are reached
+  struct OscillationAtTarget {
+    using type = std::array<domain::Flag, Dim>;
+    static constexpr Options::String help = {
+        "The flags returned when at the target."};
+  };
+
+  using options = tmpl::list<TargetNumberOfGridPoints, TargetRefinementLevels,
+                             OscillationAtTarget>;
+
+  static constexpr Options::String help = {
+      "Refine the grid towards the TargetNumberOfGridPoints and "
+      "TargetRefinementLevels, and then oscillate about them by applying "
+      "OscillationAtTarget."};
+
+  DriveToTarget() = default;
+
+  DriveToTarget(const std::array<size_t, Dim>& target_number_of_grid_points,
+                const std::array<size_t, Dim>& target_refinement_levels,
+                const std::array<domain::Flag, Dim>& flags_at_target);
+
+  /// \cond
+  explicit DriveToTarget(CkMigrateMessage* msg);
+  using PUP::able::register_constructor;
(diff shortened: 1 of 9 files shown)


wthrowe (dismissed):

src/ParallelAlgorithms/Amr/Criteria/DriveToTargetAndOscillate.hpp
```diff
@@ -0,0 +1,122 @@
...
+  struct TargetRefinementLevels {
+    using type = std::array<size_t, Dim>;
+    static constexpr Options::String help = {
+        "The target refinement levels in each dimension."};
+  };
+
+  /// The AMR flags chosen when the target resolution and refinement levels are
+  /// reached
+  struct DecisionAtTarget {
+    using type = std::array<domain::Flag, Dim>;
```
nilsvu: Maybe make this an optional so we can drive to the target resolution and stay there? Then you can move the "AndOscillate" from the class name to this option, maybe name "OscillateAtTarget".

kidder: you can stay at the target by choosing the flags to be `DoNothing`

nilsvu: Oh good. Then just move the "oscillate" name here?

kidder: okay, done and squashed

src/ParallelAlgorithms/Amr/Criteria/DriveToTarget.hpp
```diff
@@ -0,0 +1,124 @@
...
+ * \note This criterion is primarily for testing the mechanics of refinement.
+ */
+template <size_t Dim>
+class DriveToTarget : public Criterion {
+ public:
+  /// The target resolution
+  struct TargetResolution {
+    using type = std::array<size_t, Dim>;
+    static constexpr Options::String help = {
+        "The target resolution in each dimension."};
```
nilsvu: Note here that "resolution" means number of grid points?

(Not for this PR) We've been a bit inconsistent throughout the code with this naming. Sometimes we use "number of grid points", sometimes it's "extents" (which isn't particularly clear because there's a notion of "size" attached to it), and in AMR it's "resolution" which also isn't super clear because h-refinement is also a notion of resolution. Nothing to resolve in this PR, but we should probably make this more consistent. I suggest "num points" because everyone understands what that means.

src/ParallelAlgorithms/Amr/Criteria/DriveToTarget.hpp
```diff
@@ -0,0 +1,124 @@
...
+  /// \endcond
+
+  using compute_tags_for_observation_box = tmpl::list<>;
+
+  using argument_tags = tmpl::list<::domain::Tags::Mesh<Dim>>;
+
+  template <typename ArrayIndex, typename Metavariables>
+  auto operator()(const Mesh<Dim>& current_mesh,
+                  Parallel::GlobalCache<Metavariables>& /*cache*/,
+                  const ArrayIndex& array_index) const;
```
nilsvu: You rely on ArrayIndex to be an ElementId, so specialize to that and remove the template param I think.

Then you could move the implementation to the cpp (with a private impl function that doesn't depend on the unused metavars). Up to you.

tests/Unit/ParallelAlgorithms/Amr/Criteria/Test_DriveToTarget.cpp
```diff
@@ -0,0 +1,169 @@
...
+    const std::vector<
+        std::tuple<std::array<size_t, VolumeDim>, std::array<size_t, VolumeDim>,
+                   std::array<amr::domain::Flag, VolumeDim>>>& test_cases,
+    const std::string& option_string) {
+  Parallel::register_factory_classes_with_charm<Metavariables<VolumeDim>>();
+
+  const amr::Criteria::DriveToTarget criterion{target_extents, target_levels,
+                                               flags_at_target};
+  const auto criterion_from_option_string =
+      TestHelpers::test_creation<std::unique_ptr<amr::Criterion>,
```
nilsvu, line 78: I think we have a test helper that creates the metavars for you so you don't have to define them in this test. `test_factory_creation` or something like that.

tests/Unit/ParallelAlgorithms/Amr/Criteria/Test_DriveToTarget.cpp
```diff
@@ -0,0 +1,169 @@
...
+                   expected_flags);
+    test_criterion(*serialize_and_deserialize(criterion_from_option_string),
+                   extents, levels, expected_flags);
+  }
+}
+}  // namespace
+
+SPECTRE_TEST_CASE("Unit.Amr.Criteria.DriveToTarget",
+                  "[Unit][ParallelAlgorithms]") {
+  const std::array target_extents_1d{4_st};
```
nilsvu: It helps to wrap a test like this in scopes like

```cpp
{
  INFO("1D");
  // ...
}
```

Then you get useful output from catch, and you can also avoid all those `_1d` suffixes because variable names are scoped.

nilsvu (changes requested):

tests/Unit/ParallelAlgorithms/Amr/Criteria/Test_DriveToTarget.cpp
kidder: The metavariables are used in the base class so I left this as is since I need to define them

kidder:
pushed a fixup

nilsvu (dismissed):

kidder:
squashed

nilsvu (approved):

nilsdeppe:
Why is auto-merge not enabled? The label says it should've been...

nilsvu:
I was going to wait at least one business day before merging. Enabling auto merge would have merged this directly. Hence the label, to remind myself to merge it next business day.

PR #4692, opened 2023-02-03
Fix floating point bug in Test_Tov.cpp, ASSERT/ERROR print to full precision

### LINKED ISSUE #4657: Failure in Unit.PointwiseFunctions.AnalyticSolutions.Gr.Tov
Failure is on my dev machine and is consistent across runs.

```
  Node: 0 Proc: 0
  Line: 262 of /home/wthrowe/spectre/src/PointwiseFunctions/AnalyticSolutions/
  RelativisticEuler/Tov.cpp
  'get_element(r, i) >= 0.0 and get_element(r, i) <= outer_radius_' violated!
  Function: DataType RelativisticEuler::Solutions::TovSolution::
  mass_over_radius(const DataType &) const [DataType = double]
  Invalid radius: 3.54491 not in [0.0, 3.54491]
  
############ ASSERT FAILED ############
```

The failure is triggered by the call on line 101 of the test:
```c++
      const double interpolated_mass{
          tov_out_full.mass_over_radius(intermediate_radius) * intermediate_radius};
```

nilsdeppe:
## Proposed changes

- The `ASSERT` as it was basically said `1.0 == 1.0`, but it differed at roundoff. Printing to full precision in `ASSERT`/`ERROR` will help us debug these subtle issues more easily. I expect this will be especially useful to newer folks.
- Elide some floating point math that leads to slightly different solves in the TOV solver test.

Fixes #4657 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Utilities/ErrorHandling/Assert.hpp
+++ b/src/Utilities/ErrorHandling/Assert.hpp
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <iomanip>
 #include <sstream>
 #include <string>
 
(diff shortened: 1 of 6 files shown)


src/Utilities/ErrorHandling/Assert.hpp
```diff
@@ -33,29 +34,31 @@
...
-                               static_cast<const char*>(__PRETTY_FUNCTION__), \
-                               avoid_name_collisions_ASSERT.str());           \
-    }                                                                         \
+#define ASSERT(a, m)                                                           \
+  do {                                                                         \
+    if (!(a)) {                                                                \
+      disable_floating_point_exceptions();                                     \
+      std::ostringstream avoid_name_collisions_ASSERT;                         \
+      /* clang-tidy: macro arg in parentheses */                               \
+      avoid_name_collisions_ASSERT << std::setprecision(18) << std::scientific \
```
wthrowe, line 43: Completely optional, since it's unrelated to everything else in the PR:  At some point we switched ERROR to use MakeString.  Changing ASSERT to match would be good.

wthrowe (dismissed):

isaaclegred (dismissed):

src/Utilities/ErrorHandling/Assert.hpp
nilsdeppe: I'll make this change. I'll do it in a new commit but this PR. That way CI needs fewer complete rebuilds :D

nilsdeppe: Okay, never mind... I get annoying non-trivial compiler errors about `operator<<` not existing in `StdHelpers.hpp`. I think we've had this sort of thing before... I'll just leave it for now. This might even be why `ASSERT` is changed but not `ERROR`

wthrowe (approved):

PR #4693, opened 2023-02-03
Enable no-math-errno and reciprocal optimizations.

nilsdeppe:
## Proposed changes

- `no-math-errno` allows the compiler to switch things like `sqrt` to vector intrinsics.
- Enable reciprocal math. We generally don't care about the difference between `a / b` and `a * one_over_b`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupCxxFlags.cmake
+++ b/cmake/SetupCxxFlags.cmake
@@ -59,6 +59,16 @@ set_property(TARGET SpectreFlags
   INTERFACE_COMPILE_OPTIONS
   $<$<COMPILE_LANGUAGE:CXX>:-ftemplate-backtrace-limit=0>)
 
+# Disable cmath setting the error flag. This allows the compiler to more
+# aggressively vectorize code since it doesn't need to respect some global
+# state.
+set_property(TARGET SpectreFlags
+  APPEND PROPERTY
+  INTERFACE_COMPILE_OPTIONS
+  $<$<COMPILE_LANGUAGE:C>:-fno-math-errno>
+  $<$<COMPILE_LANGUAGE:CXX>:-fno-math-errno>
+  $<$<COMPILE_LANGUAGE:Fortran>:-fno-math-errno>)
+
 # By default, the LLVM optimizer assumes floating point exceptions are ignored.
 create_cxx_flag_target("-ffp-exception-behavior=maytrap" SpectreFpExceptions)
 target_link_libraries(
--- a/cmake/SetupCxxFlags.cmake
+++ b/cmake/SetupCxxFlags.cmake
@@ -69,6 +69,15 @@ set_property(TARGET SpectreFlags
   $<$<COMPILE_LANGUAGE:CXX>:-fno-math-errno>
   $<$<COMPILE_LANGUAGE:Fortran>:-fno-math-errno>)
 
+# Allow the compiler to transform divisions into multiplication by the
+# reciprocal.
+set_property(TARGET SpectreFlags
+  APPEND PROPERTY
+  INTERFACE_COMPILE_OPTIONS
+  $<$<COMPILE_LANGUAGE:C>:-freciprocal-math>
+  $<$<COMPILE_LANGUAGE:CXX>:-freciprocal-math>
+  $<$<COMPILE_LANGUAGE:Fortran>:-freciprocal-math>)
+
 # By default, the LLVM optimizer assumes floating point exceptions are ignored.
 create_cxx_flag_target("-ffp-exception-behavior=maytrap" SpectreFpExceptions)
 target_link_libraries(


cmake/SetupCxxFlags.cmake
```diff
@@ -81,6 +81,19 @@ set_property(TARGET SpectreFlags
...
+# Define for C as well since C99 has an fma function.
+set_property(TARGET SpectreFlags
+  APPEND PROPERTY
+  INTERFACE_COMPILE_DEFINITIONS
+  $<$<COMPILE_LANGUAGE:CXX>:FP_FAST_FMA>
+  $<$<COMPILE_LANGUAGE:C>:FP_FAST_FMA>
+  $<$<COMPILE_LANGUAGE:CXX>:FP_FAST_FMAF>
+  $<$<COMPILE_LANGUAGE:C>:FP_FAST_FMAF>
+  $<$<COMPILE_LANGUAGE:CXX>:FP_FAST_FMAL>
+  $<$<COMPILE_LANGUAGE:C>:FP_FAST_FMAL>)
```
wthrowe: I think these macros are output from the standard library reporting support for features, not input to the library requesting features.  They should be set by `-mfma` (probably included in `-march` if appropriate).  Testing with gcc on my machine doesn't show any effect from defining these, but only gives an fma instruction with `-mfma` (which also defines the macros).

wthrowe (changes requested):

cmake/SetupCxxFlags.cmake
nilsdeppe: In my tests GCC defined them but clang unfortunately didn't, even if I have `-O3 -mfma -march=native` on a machine that supports fma. I found this rather surprising and annoying...

wthrowe: You are correct that clang is missing the definitions, but defining them on the command line still doesn't enable fma instructions or anything.

nilsdeppe: Ah, I see what you're saying now. I misinterpreted the cppreference wording as "You can define these to enable FMA" instead of "If these are defined, you are guaranteed an FMA". I'll drop the commit. Thanks!

nilsdeppe:
Okay, dropped the last commit, nothing else changed! Thanks for the review!

cmake/SetupCxxFlags.cmake
```diff
@@ -69,6 +69,18 @@ set_property(TARGET SpectreFlags
...
+#
+# Note: this can be problematic in some edge cases where a norm can become
+# negative due to the different rounding properties of FMA so if we run into
+# problems we'll need to reassess.
+set_property(TARGET SpectreFlags
+  APPEND PROPERTY
+  INTERFACE_COMPILE_OPTIONS
+  $<$<COMPILE_LANGUAGE:C>:-ffp-contract=on>
+  $<$<COMPILE_LANGUAGE:CXX>:-ffp-contract=on>
+  $<$<COMPILE_LANGUAGE:Fortran>:-ffp-contract=on>)
```
wthrowe: I looked into this one in a bit more detail, and it looks like this option means different things in clang and gcc.  For clang "on" is the default, and for GCC it looks like "on" actually turns off converting things to FMA instructions.  So we should  probably leave this one out.

GCC 11.3.1:
>        -ffp-contract=style
>            -ffp-contract=off disables floating-point expression contraction.
>            -ffp-contract=fast enables floating-point expression contraction
>            such as forming of fused multiply-add operations if the target has
>            native support for them.  -ffp-contract=on enables floating-point
>            expression contraction if allowed by the language standard.  This
>            is currently not implemented and treated equal to
>            -ffp-contract=off.
> 
>            The default is -ffp-contract=fast.

Clang (whatever version is documented on their website):
> -ffp-contract=\<arg>
>
> Form fused FP ops (e.g. FMAs): fast (fuses across statements disregarding pragmas) | on (only fuses in the same statement unless dictated by pragmas) | off (never fuses) | fast-honor-pragmas (fuses across statements unless dictated by pragmas). Default is ‘fast’ for CUDA, ‘fast-honor-pragmas’ for HIP, and ‘on’ otherwise. <arg> must be ‘fast’, ‘on’, ‘off’ or ‘fast-honor-pragmas’.

wthrowe (commented):
I should have gone through the others in more detail when I first looked at this.  It looks like another one might also be problematic.

cmake/SetupCxxFlags.cmake
nilsdeppe: Sigh... If only things were more consistent. I agree, let's omit. If we need to change to `fast-honor-pragmas` for clang later we can!

nilsdeppe:
Okay, pushed an update again!

wthrowe (approved):
The rest looks good.  Got to love it when "on" is a synonym for "off".

nilsvu:
Could you update the PR description and title because I think you dropped some changes?

PR #4697, opened 2023-02-06
CylBinaryCompactObject: add equiangular map.

markscheel:
## Proposed changes

Adds an equiangular map option to CylindricalBinaryCompactObject.

### Upgrade instructions

When using the `CylindricalBinaryCompactObject` domain creator one must now specify `UseEquiangularMap`.  Setting this option to false gives the previous behavior.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/CylindricalBinaryCompactObject.cpp
+++ b/src/Domain/Creators/CylindricalBinaryCompactObject.cpp
@@ -57,13 +57,10 @@ std::array<double, 3> flip_about_xy_plane(const std::array<double, 3> input) {
 
 namespace domain::creators {
 CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
-    typename CenterA::type center_A, typename CenterB::type center_B,
-    typename RadiusA::type radius_A, typename RadiusB::type radius_B,
-    typename IncludeInnerSphereA::type include_inner_sphere_A,
-    typename IncludeInnerSphereB::type include_inner_sphere_B,
-    typename IncludeOuterSphere::type include_outer_sphere,
-    typename OuterRadius::type outer_radius,
-    typename UseEquiangularMap::type use_equiangular_map,
+    std::array<double, 3> center_A, std::array<double, 3> center_B,
+    double radius_A, double radius_B, bool include_inner_sphere_A,
+    bool include_inner_sphere_B, bool include_outer_sphere, double outer_radius,
+    bool use_equiangular_map,
     const typename InitialRefinement::type& initial_refinement,
     const typename InitialGridPoints::type& initial_grid_points,
     std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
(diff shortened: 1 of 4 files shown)


src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -63,6 +63,7 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
     typename IncludeInnerSphereB::type include_inner_sphere_B,
     typename IncludeOuterSphere::type include_outer_sphere,
     typename OuterRadius::type outer_radius,
+    typename UseEquiangularMap::type use_equiangular_map,
```
nilsvu: Just write `bool`, no need to hide the type. We've removed this unnecessary obscuration in the other domain creators too (except for complicated `std::variant` types). A separate commit is fine.

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -364,8 +368,11 @@ void test_connectivity() {
         CAPTURE(include_inner_sphere_A);
         for (const bool include_inner_sphere_B : {true, false}) {
           CAPTURE(include_inner_sphere_B);
-          test_connectivity_once(with_sphere_e, include_inner_sphere_A,
-                                 include_inner_sphere_B, include_outer_sphere);
+          for (const bool use_equiangular_map : {true, false}) {
```
nilsvu: Use the new `cartesian_product` you avoid the nested loops. Also CAPTURE the bool.

nilsvu (changes requested):
Are you relatively certain that the unit tests would catch bugs where coord maps join discontinuously because of missing equiangular maps? For example, do they fail when you only make some of the cylinders equiangular?

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
markscheel: I capture the bool now.

I don't see how to use `cartesian_product` except for doing things that would make the code unreadable and/or bug prone, like creating a `cartesian_product({{false}}, {{true, false}}, {{true, false}}, {{true, false}},{{true, false}})` and then looping over it and trying to name the different fields `include_inner_sphere_A` etc (and get the order right) and CAPTURE them.
Maybe `cartesian_product` needs better documentation or an example?

markscheel:
I tried making only some of the wedges equiangular, or some of the cylinders equiangular, and the test still passes. I don't understand why `TestHelpers::domain::creators::test_domain_creator` doesn't test this.  What happens for the other BBH domain? Is that untested as well?

Made other changes except for cartesian_product.

kidder:
> I tried making only some of the wedges equiangular, or some of the cylinders equiangular, and the test still passes. I don't understand why `TestHelpers::domain::creators::test_domain_creator` doesn't test this. What happens for the other BBH domain? Is that untested as well?
> 
> Made other changes except for cartesian_product.

try cherry-picking the commit from https://github.com/kidder/spectre/tree/check_block_faces and see if the test still passes

markscheel:
Rebased, which also did some of the requested changes here. So now there are only 3 relevant fixup commits.
I will try https://github.com/kidder/spectre/tree/check_block_faces

markscheel:
Thanks @kidder !  when I cherry-pick  https://github.com/kidder/spectre/tree/check_block_faces, the current PR passes the tests as-is, but if I intentionally introduce a bug by turning on `use_equiangular_map` for some blocks but not others, then the test fails.

kidder:
okay, I'll clean that up and make a PR

nilsvu (dismissed):
I thought that continuity of maps was tested in one of the generic tests already, but apparently it isn't. Thanks for checking. We don't need to hold up the PR with that. @kidder thanks for adding that test.

kidder:
the current test only checks the corners of the block faces, not the interior...

markscheel:
squashed

nilsvu (approved):

nilsvu (changes requested):
The test timed out in one of the builds. I've seen timeouts in this test before. Can we reduce the amount of work done in the big cartesian product of options?
Edit: or we randomly skip some of the iterations of the loop ("fuzzy cartesian product")

nilsvu (dismissed):

nilsvu:
@markscheel would you mind adding the random-skipping code also to the BinaryCompactObject domain test while you rebase/squash? It's been timing out as well.

geoffrey4444 (dismissed):

nilsvu (approved):

PR #4698, opened 2023-02-06
Add weights for FD integration

ffoucart:
## Proposed changes

Add weights for low-order integration on a finite difference grid 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/Spectral/FiniteDifference.cpp
+++ b/src/NumericalAlgorithms/Spectral/FiniteDifference.cpp
@@ -16,10 +16,12 @@ std::pair<DataVector, DataVector> compute_collocation_points_and_weights<
     Basis::FiniteDifference, Quadrature::CellCentered>(
     const size_t num_points) {
   DataVector x{num_points};
-  DataVector w{num_points, std::numeric_limits<double>::signaling_NaN()};
   // The finite difference grid cells cover the interval [-1, 1]
-  constexpr double lower_bound = -1.0, upper_bound = 1.0;
+  constexpr double lower_bound = -1.0;
+  constexpr double upper_bound = 1.0;
   const double delta_x = (upper_bound - lower_bound) / num_points;
+  // Weights for integration using midpoint method
+  DataVector w{num_points, delta_x};
   for (size_t i = 0; i < num_points; ++i) {
     x[i] = lower_bound + 0.5 * delta_x + i * delta_x;
   }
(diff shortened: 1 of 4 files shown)


src/NumericalAlgorithms/Spectral/FiniteDifference.cpp
```diff
@@ -16,10 +16,12 @@ std::pair<DataVector, DataVector> compute_collocation_points_and_weights<
     Basis::FiniteDifference, Quadrature::CellCentered>(
     const size_t num_points) {
   DataVector x{num_points};
-  DataVector w{num_points, std::numeric_limits<double>::signaling_NaN()};
   // The finite difference grid cells cover the interval [-1, 1]
-  constexpr double lower_bound = -1.0, upper_bound = 1.0;
+  constexpr double lower_bound = -1.0;
+  constexpr double upper_bound = 1.0;
   const double delta_x = (upper_bound - lower_bound) / num_points;
+  // Weights assuming finite volume (or 2nd order finite difference integral)
```
nilsdeppe: None of this should mention FV, this is all restricted to FD as the "basis". For FV we will have other code. Can you update the commit message too?

src/NumericalAlgorithms/Spectral/FiniteDifference.cpp
```diff
@@ -16,10 +16,12 @@ std::pair<DataVector, DataVector> compute_collocation_points_and_weights<
     Basis::FiniteDifference, Quadrature::CellCentered>(
```
nilsdeppe, line 16: Picking a line at random: can you add tests for the changes?

nilsdeppe (changes requested):

src/NumericalAlgorithms/Spectral/FiniteDifference.cpp
ffoucart: There are also issues with the existing tests... they apparently want the inverse weight function to be defined.

ffoucart:
There seems to be an error specific to the GrMhdTOVStar test... an integrand with 27 (3^3) points that the code attempts to integrate on a 125 (5^3) point mesh, at first look. Maybe some mixup between the spectral and FD grids?

src/ParallelAlgorithms/Events/ObserveNorms.hpp
```diff
@@ -350,11 +344,7 @@ operator()(const typename ObservationValueTag::type& observation_value,
                                                 Frame::Inertial>>(box));
   const size_t number_of_points = mesh.number_of_grid_points();
   const double local_volume = [&mesh, &det_jacobian]() {
-    if (mesh.basis(0) == Spectral::Basis::FiniteDifference) {
```
nilsdeppe, line 353: Ah, the issue is that in this case is that the integrand of whatever the code is trying to take a volume integral of, i.e. the error terms, can't be done on the FD grid because the compute tag that computes the integrand isn't aware that it needs to know whether the grid is FD or DG. However, as far as I can tell the TOV star is using the right `error_compute` (see line 343 of `src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp`). Can you reproduce the error locally and get into a debugger to figure out a backtrace?

ffoucart:
run tests

ffoucart:
Jenkins run tests

nilsvu:
@ffoucart we had some issues with CI (upgrading our vector arithmetic library in the container). Should be resolved now. I restarted CI for you.

src/NumericalAlgorithms/Spectral/Spectral.hpp
```diff
@@ -172,6 +173,8 @@ T compute_basis_function_value(size_t k, const T& x);
  * \brief Compute the inverse of the weight function \f$w(x)\f$ w.r.t. which
  * the basis functions are orthogonal. See the description of
  * `quadrature_weights(size_t)` for details.
+ * This is arbitrarily set to 1 for FiniteDifference basis, to integrate
```
nilsdeppe, line 176: Sorry, why `1`? Or maybe my question is, why is it arbitrary?

nilsdeppe (commented):
LGTM, one question/clarification. You can squash any additional documentation immediately :)

src/NumericalAlgorithms/Spectral/Spectral.hpp
ffoucart: The only thing that is used is the product of the weights with the inverse weight function. Neither one individually is particularly meaningful for FD. So we set one to the desired value, and the other to 1...

ffoucart: I added some comments in the `quadrature_weights(size_t)` function and rebased.

nilsdeppe: Okay, thanks!

nilsdeppe (approved):

PR #4699, opened 2023-02-06
Add AMR criterion to Punctures

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/Punctures/AmrCriteria/CMakeLists.txt
+++ b/src/Elliptic/Systems/Punctures/AmrCriteria/CMakeLists.txt
@@ -1,34 +1,36 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
-set(LIBRARY Punctures)
+set(LIBRARY PuncturesAmrCriteria)
 
 add_spectre_library(${LIBRARY})
 
 spectre_target_sources(
   ${LIBRARY}
   PRIVATE
-  Sources.cpp
+  RefineAtPunctures.cpp
   )
 
 spectre_target_headers(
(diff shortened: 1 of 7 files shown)


kidder (commented):
Not a detailed review, but you will need to specify one or more amr::Criteria as options to this criterion that tells you whether you want to refine, coarsen, or do nothing and then modify the flag to the correct one based on whether you want to do h- or p-refinement.  Otherwise this will just keep refining the grid each time AMR is called independent of what other criteria you are using in the list that contains RefineAtPunctures.   

Alternatively, I've thought about introducing AMR policies which would modify the AMR flags after all the criteria have been reduced to a single decision,  to do things like make the flags isotropic, set a min/max resolution, ....   In which case, this could be handled as a policy.

nilsvu:
I'm not sure I understand. I do want to do this refinement at every AMR step. Since this is an elliptic solve I want to do a fixed number of refinement levels: Solve, refine, solve, etc. So I don't need to limit the AMR to a max refinement. In this particular case I even need only this one very simple AMR criterion that h-refines at the punctures and p-refines everywhere else. Does that clarify things? Do I need to make changes to this class?

In the future I'd like to add more complex AMR criteria that complement each other, so I'd like to understand how you envision multiple AMR criteria working together. Perhaps you could add some docs on that to the `amr::Criterion` class? That's the place people will look when implementing AMR criteria.

kidder:
ah, okay.  now I understand what you are trying to do.  I'll take a closer look

kidder (approved):

PR #4701, opened 2023-02-07
Prevent an Element from splitting and joining at the same time

kidder:
With anisotropic refinment, it is possible to split an Element in one dimension while joining with another Element in another dimension.  This vastly complicates the algorithm, so I've chosen not to allow it by choosing to turn the join into doing nothing.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/ParallelAlgorithms/Amr/Actions/EvaluateRefinementCriteria.hpp
```diff
@@ -97,6 +100,17 @@ struct EvaluateRefinementCriteria {
       }
     }
 
+    // An element cannot join if it is splitting in another dimension
+    if (alg::any_of(overall_decision, [](amr::domain::Flag flag) {
+          return flag == amr::domain::Flag::Split;
+        })) {
+      for (size_t d = 0; d < volume_dim; ++d) {
+        if (gsl::at(overall_decision, d) == amr::domain::Flag::Join) {
+          gsl::at(overall_decision, d) = amr::domain::Flag::DoNothing;
```
nilsvu: Does this logic have to be repeated here? The AMR logic is complicated enough, so it would be quite helpful to confine it to as few places as possible. If you put this here so it runs even when the element has no neighbors, then I suggest to allow the update function to work for the no-neighbors case.

nilsvu: This overrides the priority of flags returned by the AMR criteria above: we select "do nothing" even though another criterion may have selected a lower-priority flag than "join" that's compatible with the split in the other dimension. Is that ok / expected behavior?

nilsvu (changes requested):

src/ParallelAlgorithms/Amr/Actions/EvaluateRefinementCriteria.hpp
kidder: there is no lower-priority flag than `Join` so this is okay.  I can add a comment if you'd like

kidder: Yes, as I think it is better to make this change before sending the flags to neighbors, as each time `update_amr_decision` changes a flag it triggers sending the updated flags to each neighbor...

nilsvu: maybe factor out the logic into a separate function, and add some comments explaining what you wrote here?

nilsvu: Sounds good. A comment would be helpful.

kidder: okay, pushed a fixup

nilsvu (dismissed):

kidder:
squashed

nilsvu (approved):

PR #4704, opened 2023-02-07
Rewrite History

wthrowe:
## Proposed changes

This includes only the rewrite of the History class and minimal changes to keep all consumers functioning.  Code taking advantage of the new features will be added in future PRs, since this one is already quite large.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DgSubcell/Actions/TciAndRollback.hpp
+++ b/src/Evolution/DgSubcell/Actions/TciAndRollback.hpp
@@ -211,17 +211,11 @@ struct TciAndRollback {
 
           // Project the time stepper history to the subcells, excluding the
           // most recent inadmissible history.
-          TimeSteppers::History<
-              typename db::add_tag_prefix<::Tags::dt, variables_tag>::type>
-              subcell_history{active_history_ptr->integration_order()};
-          const auto end_it = std::prev(active_history_ptr->derivatives_end());
-          for (auto it = active_history_ptr->derivatives_begin(); it != end_it;
-               ++it) {
-            subcell_history.insert(
-                it.time_step_id(),
-                fd::project(*it, dg_mesh, subcell_mesh.extents()));
-          }
-          *active_history_ptr = std::move(subcell_history);
+          active_history_ptr->undo_latest();
+          active_history_ptr->map_entries(
+              [&dg_mesh, &subcell_mesh](const auto entry) {
+                *entry = fd::project(*entry, dg_mesh, subcell_mesh.extents());
+              });
           *active_grid_ptr = ActiveGrid::Subcell;
           *did_rollback_ptr = true;
           // Project the neighbor data we were sent for reconstruction since
--- a/src/Time/Actions/SelfStartActions.hpp
+++ b/src/Time/Actions/SelfStartActions.hpp
@@ -356,6 +356,18 @@ struct CheckForOrderIncrease {
                      "integration order.");
               mutable_history->integration_order(
                   mutable_history->integration_order() + 1);
+              // AdamsBashforth does cleanup internally when taking a
+              // step, but since we are resetting to the start instead
+              // of taking a step we have to do it here.  If we
+              // skipped this the extra value would harmlessly fall
+              // off the end after a few steps, but that would
+              // unnecessarily increase the size of the data held by
+              // the History object.  A different way to handle this
+              // would be to call shrink_to_fit() on the history a few
+              // steps after the start of the evolution.
+              for (const auto& record : *mutable_history) {
+                mutable_history->discard_value(record.time_step_id);
+              }
             });
       });
     }
--- a/src/Time/History.hpp
+++ b/src/Time/History.hpp
@@ -3,400 +3,918 @@
 
 #pragma once
 
+#include <boost/container/static_vector.hpp>
 #include <cstddef>
-#include <deque>
-#include <iterator>
-#include <tuple>
+#include <optional>
+#include <pup.h>
 #include <utility>
 
 #include "DataStructures/DataBox/PrefixHelpers.hpp"
 #include "DataStructures/DataBox/Prefixes.hpp"
 #include "DataStructures/MathWrapper.hpp"
-#include "Time/Time.hpp"
+#include "DataStructures/StaticDeque.hpp"
+#include "Parallel/PupStlCpp17.hpp"
 #include "Time/TimeStepId.hpp"
-
-// IWYU pragma: no_include <unordered_set>  // for swap?
-
-/// \cond
-namespace PUP {
-class er;
-}  // namespace PUP
-/// \endcond
+#include "Utilities/ContainsAllocations.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/StlBoilerplate.hpp"
+#include "Utilities/TMPL.hpp"
 
 namespace TimeSteppers {
 
-/// \cond
-template <typename Vars>
-class DerivIterator;
-template <typename T>
-class HistoryIterator;
-/// \endcond
+/// Largest number of past steps supported by the time stepper
+/// `History`.  Corresponds to the `number_of_past_steps()` method of
+/// `TimeStepper`.  AdamsBashforth with order 8 has the largest
+/// requirement.
+constexpr size_t history_max_past_steps = 7;
+/// Largest number of substeps supported by the time stepper
+/// `History`.  Corresponds to the `number_of_substeps()` and
+/// `number_of_substeps_for_error()` methods of `TimeStepper`.
+/// DormandPrince5 with an error estimate and Rk5Owren have the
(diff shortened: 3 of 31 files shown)


src/Time/History.hpp
```diff
@@ -3,400 +3,910 @@
...
+    explicit MutableSubsteps(const gsl::not_null<History*> history)
+        : history_(history) {}
+
+    gsl::not_null<History*> history_;
+  };
+  /// \endcond
+
+  /// Immutable, type-erased access to the history.  This method
+  /// returns a class derived from `ConstUntypedHistory<UntypedVars>`.
+  /// Any modifications to this class invalidate the returned object.
```
nilsdeppe: I found the wording a bit unclear. I assume you mean `Any modifications to the History object invalidate the returned ConstUntypedAccess object`. Would you mind being a bit more verbose like that?

src/Time/History.hpp
```diff
@@ -3,400 +3,910 @@
...
+  ConstUntypedAccess untyped() const { return ConstUntypedAccess(*this); }
+
+  /// Mutable, type-erased access to the history.  This method returns
+  /// a class derived from `MutableUntypedHistory<UntypedVars>`.  Any
+  /// modifications to this class invalidate the returned object,
```
nilsdeppe: Same comment/question about verbosity

src/Time/History.hpp
```diff
@@ -3,400 +3,910 @@
...
-};
+  /// Insert a new entry into the history.  It will be inserted as a
+  /// step or substep as appropriate.
+  ///
+  /// The supplied `time_step_id` must be newer than the current
+  /// newest entry, and if the substep of `time_step_id` is nonzero,
+  /// the id must be consistent with the existing data for the current
+  /// step.
+  ///
+  /// If `no_value` is passed, the created record will not have its
```
nilsdeppe: Could you do:

if `no_value` (i.e. `History::no_value` variable) is passed...

src/Time/History.hpp
```diff
@@ -3,400 +3,910 @@
...
+  /// `value` field set.  This is primarily useful for histories for
+  /// the implicit portions of IMEX time steppers.
```
nilsdeppe: Would you mind adding a bit more to the end here? E.g. `, where when solving for the stiff source terms iteratively...` or whatever the use-case is?

src/Time/History.hpp
```diff
@@ -3,400 +3,910 @@
...
+  ///
+  /// The `value` and `derivative` data will be copied into cached
+  /// allocations, if any are available.
```
nilsdeppe: Can you explicitly state `That is, only a copy (not a memory allocation) will occur whenever possible.`? I realize this is mostly rephrasing, but I think it's good to be overly verbose with performance things in this sort of context :)

src/Time/History.hpp
```diff
@@ -3,400 +3,910 @@
...
-template <typename DerivVars>
-class DerivIterator {
-  using Base =
-      typename std::deque<std::tuple<TimeStepId, DerivVars>>::const_iterator;
+  /// Insert data at the start of the history.  This can be useful
+  /// when initializing a multistep integrator.
+  ///
+  /// This should only be used for initialization, and cannot be used
+  /// for substep data.  The supplied `time_step_id` must be older
+  /// than the current first entry.
```
nilsdeppe, line 537: Would you mind doing a little ASCII diagram of where things are inserted and how you define "older" here? Is this equivalent to `push_front` or `push_back` in the underlying container? This is more of a question to make a brief note of the internal ordering. I know that's not part of the interface, but I think would help in understanding the details of the code

src/Time/History.hpp
```diff
@@ -3,400 +3,910 @@
...
+  boost::container::static_vector<StepRecord<Vars>, history_max_substeps>
+      substep_values_{};
+
+  // If the algorithm undoes a substep, it may want to reset the
+  // variables to the previous value.  We don't want the time stepper
+  // to have to keep track of this when discarding values, so we hang
+  // onto the last value if it gets discarded.
+  std::optional<Vars> latest_value_if_discarded_{};
+
+  // Memory allocations available for reuse.  We could use
```
nilsdeppe: I haven't (yet) gotten to where these are used. So my question might be moot, but is there any reason to think that the vague guarantees of when `std::vector` itself reallocates could be a problem? As in, we might not have the underlying containers reallocate, but we might have the `std::vector` reallocate. If it is, I guess then we would want this to be a static vector, right? It might be useful to add a sentence or so about the std::vector allocations.

src/Evolution/DgSubcell/Actions/TciAndRollback.hpp
```diff
@@ -211,17 +211,11 @@ struct TciAndRollback {
...
-              subcell_history{active_history_ptr->integration_order()};
-          const auto end_it = std::prev(active_history_ptr->derivatives_end());
-          for (auto it = active_history_ptr->derivatives_begin(); it != end_it;
-               ++it) {
-            subcell_history.insert(
-                it.time_step_id(),
-                fd::project(*it, dg_mesh, subcell_mesh.extents()));
-          }
-          *active_history_ptr = std::move(subcell_history);
+          active_history_ptr->undo_latest();
```
nilsdeppe, line 214: This is way nicer :D Thank you!

src/Time/Actions/SelfStartActions.hpp
```diff
@@ -356,6 +356,18 @@ struct CheckForOrderIncrease {
                      "integration order.");
               mutable_history->integration_order(
                   mutable_history->integration_order() + 1);
+              // AdamsBashforth does cleanup internally when taking a
+              // step, but since we are resetting to the start instead
+              // of taking a step we have to it here.  If we skipped
```
nilsdeppe: `to it here` -> `to do it here`?

nilsdeppe (changes requested):
LGTM, just some suggestions for clarifying comments. Please rebase and squash immediately!

src/Time/History.hpp
wthrowe: Left out the specific reference to `ConstUntypedAccess` since it is intentionally not documented as there should never be a reason to work with one explicitly, but hopefully made the comment clearer.

src/Time/History.hpp
wthrowe: I just put the qualifier on the existing name.  Repeating it seems unnecessary.

src/Time/History.hpp
wthrowe: This is all in a bit of flux at the moment, unfortunately.  As of this PR, the value field is never used and all the processing of it is actually wasted effort (but much of it is still being done).  The next PR in this sequence will change that and setting it will always be required.

The intended use case is for IMEX, but unfortunately the details are along the lines of "when calling some implicit-integration-related functions that don't exist on develop yet and haven't had their names or interfaces fully nailed down".  (I almost left these functions out of this PR.)  I can add something about stiff source terms if you'd like, but I've now encountered two situations where the solve isn't iterative, so I want to avoid that part.  (These are (1) getting an approximate solution from a semi-implicit solve, which is roughly a normal solve but stopping after one iteration, and (2) the case where the implicit problem can actually be solved analytically, which happens in some regimes for FFE.)

Implementation-wise, integration is split up like `y1 = y0 + dt * explicit_stuff + dt * other_stuff` (where `other_stuff` is just the implicit solve, at least for now).  As of a not-yet-submitted PR, the `value` field is used to obtain `y0`, and the normal TimeStepper interface *sets* `y1` to the first two terms together.  The implicit solve then *adds* its adjustment to the result.  Obviously, adding another constant term would be bad (i.e., zeroth order), so nothing except the main integration needs the `value` field.

After thinking through all that, maybe a good explanation is something about histories that do not represent the primary integration, but rather some adjustment on top of it.  I've made an attempt at adding something.

src/Time/History.hpp
wthrowe: Added a comparison to `push_front`, since that is really what this is.  "Older" has its usual physics meaning of "from an earlier time" (unless time is running backwards).  I changed to "earlier", which can't be confused with any sort of C++ or container-related thing.  (I usually use "before" for this relation, but that's not great when discussing an ordered container.)

src/Time/History.hpp
wthrowe: Hmm.  I think that actually is a concern.  The size of the vectors is bounded above, so I was thinking that the allocation would grow to possibly a bit bigger than that size and then not change.  However, while `std::vector` is usually not allowed to reallocate when you remove elements, I think it is permitted if you actually empty the container, which is likely here.  I'll switch them over.

src/Time/History.hpp
nilsdeppe: I like what you've put :) I think that helps a lot!

src/Time/History.hpp
nilsdeppe: Great, thank you!

nilsdeppe (commented):
LGTM, please squash!

nilsdeppe (approved):

PR #4711, opened 2023-02-08
Add electric current density (Ohm's law) to ForceFree system

yoonso0-0:
## Proposed changes

* Rename `SpatialCurrentDensity` to `ElectricCurrentDensity`.
* Add free functions that evaluate J_drift (explicit) and J_parallel (implicit) part of electric current density J.
* Add a compute tag which evaluates the total J (`TildeJ`)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
+++ b/src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
@@ -1,0 +1,144 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/LeviCivitaIterator.hpp"
+#include "DataStructures/Tags/TempTensor.hpp"
+#include "DataStructures/Tensor/EagerMath/DotProduct.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
+#include "PointwiseFunctions/GeneralRelativity/IndexManipulation.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace ForceFree {
+
+namespace {
+
+template <bool IncludeDriftCurrent, bool IncludeParallelCurrent>
+void tilde_j_impl(
+    const gsl::not_null<tnsr::I<DataVector, 3, Frame::Inertial>*> tilde_j,
+    const Scalar<DataVector>& tilde_q,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_e,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_b,
+    const double parallel_conductivity, const Scalar<DataVector>& lapse,
+    const Scalar<DataVector>& sqrt_det_spatial_metric,
+    const tnsr::ii<DataVector, 3, Frame::Inertial>& spatial_metric) {
+  static_assert(IncludeDriftCurrent or IncludeParallelCurrent);
+
+  Variables<tmpl::list<::Tags::Tempi<0, 3>, ::Tags::Tempi<1, 3>,
+                       ::Tags::TempScalar<0>, ::Tags::TempScalar<1>,
+                       ::Tags::TempScalar<2>>>
+      buffer{get(lapse).size()};
+
+  // compute one-forms of TildeE and TildeB in advance to reduce the number of
+  // dot products using spatial metric (which are slower than dot products
+  // without using spatial metric)
+  auto& tilde_e_one_form = get<::Tags::Tempi<0, 3>>(buffer);
+  auto& tilde_b_one_form = get<::Tags::Tempi<1, 3>>(buffer);
+  raise_or_lower_index(make_not_null(&tilde_e_one_form), tilde_e,
+                       spatial_metric);
+  raise_or_lower_index(make_not_null(&tilde_b_one_form), tilde_b,
+                       spatial_metric);
+
+  // Compute \tilde{B}^2 = \tilde{B}^j \tilde{B}_j. We need this quantity for
+  // both drift (explicit, non-stiff) and parallel (implicit, stiff) components
+  // of J^i.
+  auto& tilde_b_squared = get<::Tags::TempScalar<0>>(buffer);
+  dot_product(make_not_null(&tilde_b_squared), tilde_b, tilde_b_one_form);
+
+  if constexpr (IncludeParallelCurrent) {
+    (void)tilde_q;                  // avoid compiler warnings
+    (void)sqrt_det_spatial_metric;  // avoid compiler warnings
+
+    auto& tilde_e_squared = get<::Tags::TempScalar<1>>(buffer);
+    auto& tilde_e_dot_tilde_b = get<::Tags::TempScalar<2>>(buffer);
+    dot_product(make_not_null(&tilde_e_squared), tilde_e, tilde_e_one_form);
+    dot_product(make_not_null(&tilde_e_dot_tilde_b), tilde_e, tilde_b_one_form);
+
+    for (size_t i = 0; i < 3; ++i) {
+      (*tilde_j).get(i) =
+          parallel_conductivity *
+          (get(tilde_e_dot_tilde_b) * tilde_b.get(i) +
+           max(get(tilde_e_squared) - get(tilde_b_squared), 0.0) *
+               tilde_e.get(i));
+    }
+  } else {
+    // tilde_j should be initialized to zero before the summation performed in
+    // the next `if constexpr` block
+    for (size_t i = 0; i < 3; ++i) {
+      (*tilde_j).get(i) = 0;
+    }
+  }
+
+  if constexpr (IncludeDriftCurrent) {
+    (void)parallel_conductivity;  // avoid compiler warnings
+
+    for (LeviCivitaIterator<3> it; it; ++it) {
+      const auto& i = it[0];
+      const auto& j = it[1];
+      const auto& k = it[2];
+      (*tilde_j).get(i) +=
+          it.sign() * get(tilde_q) * tilde_e_one_form.get(j) *
+          tilde_b_one_form.get(k) /
+          get(sqrt_det_spatial_metric);  // the extra 1/sqrt{gamma} factor comes
+                                         // from the spatial Levi-Civita tensor
+    }
+  }
+
+  // overall factor
+  for (size_t i = 0; i < 3; ++i) {
+    (*tilde_j).get(i) *= get(lapse) / get(tilde_b_squared);
+  }
+}
+
+}  // namespace
+
+void ComputeDriftTildeJ::apply(
+    const gsl::not_null<tnsr::I<DataVector, 3, Frame::Inertial>*> drift_tilde_j,
+    const Scalar<DataVector>& tilde_q,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_e,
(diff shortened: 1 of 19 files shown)


src/Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp
```diff
@@ -0,0 +1,141 @@
...
+template <typename T>
+class not_null;
+}  // namespace gsl
+/// \endcond
+
+namespace ForceFree {
+
+namespace detail {
+
+void tilde_j_impl(
```
wthrowe: Remove this declaration.  It is definitely unused, since it doesn't match the definition in the cpp.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp
```diff
@@ -0,0 +1,141 @@
...
+                 Tags::ParallelConductivity, gr::Tags::Lapse<>,
+                 gr::Tags::SqrtDetSpatialMetric<>, gr::Tags::SpatialMetric<3>>;
+  using return_type = tnsr::I<DataVector, 3>;
+
+  static void apply(
+      gsl::not_null<tnsr::I<DataVector, 3, Frame::Inertial>*> drift_tilde_j,
+      const Scalar<DataVector>& tilde_q,
+      const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_e,
+      const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_b,
+      const double parallel_conductivity, const Scalar<DataVector>& lapse,
```
wthrowe: Remove `const` on non-reference arguments in declarations that are not definitions.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp
```diff
@@ -0,0 +1,141 @@
...
+                 Tags::ParallelConductivity, gr::Tags::Lapse<>,
+                 gr::Tags::SqrtDetSpatialMetric<>, gr::Tags::SpatialMetric<3>>;
+  using return_type = tnsr::I<DataVector, 3>;
+
+  static void apply(
+      gsl::not_null<tnsr::I<DataVector, 3, Frame::Inertial>*> parallel_tilde_j,
+      const Scalar<DataVector>& tilde_q,
+      const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_e,
+      const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_b,
+      const double parallel_conductivity, const Scalar<DataVector>& lapse,
```
wthrowe: Again remove `const`.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
+#include "PointwiseFunctions/GeneralRelativity/IndexManipulation.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace ForceFree {
+
+namespace detail {
```
wthrowe: Change to anonymous namespace.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+template <bool drift_current, bool parallel_current>
```
wthrowe: Template arguments should be `CamelCase`.  For boolean arguments, verbs are often clearer.  I suggest `IncludeDriftCurrent` and `IncludeParallelCurrent`.  I was confused about why they were separate arguments, and I think that makes it clearer.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+  if constexpr (drift_current) {
+    (void)parallel_conductivity;  // avoid compiler warnings
+
+    auto& tilde_e_one_form = get<::Tags::Tempi<0, 3>>(buffer);
+    auto& tilde_b_one_form = get<::Tags::Tempi<1, 3>>(buffer);
+    raise_or_lower_index(make_not_null(&tilde_e_one_form), tilde_e,
+                         spatial_metric);
+    raise_or_lower_index(make_not_null(&tilde_b_one_form), tilde_b,
+                         spatial_metric);
```
wthrowe: You could use these one-forms in B^2 at the start and in the parallel-current section to reduce the number of dot products with metrics, which are slower than dot products without metrics.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
```diff
@@ -0,0 +1,163 @@
...
+
+    if constexpr (not parallel_current) {
+      // should be initialized to zero before summation if there was no previous
+      // assignment operations.
```
wthrowe: [optional] This might be clearer as an `else` block for the first `if constexpr`.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
wthrowe: `static_assert(drift_current or parallel_current);`  I don't think it makes sense to call this with both false.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
```diff
@@ -0,0 +1,163 @@
...
+      const Scalar<DataVector>& tilde_q,                                     \
+      const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_e,                \
+      const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_b,                \
+      const double parallel_conductivity, const Scalar<DataVector>& lapse,   \
+      const Scalar<DataVector>& sqrt_det_spatial_metric,                     \
+      const tnsr::ii<DataVector, 3, Frame::Inertial>& spatial_metric)
+
+GENERATE_INSTANTIATIONS(INSTANTIATION, true, true)
+GENERATE_INSTANTIATIONS(INSTANTIATION, false, true)
+GENERATE_INSTANTIATIONS(INSTANTIATION, true, false)
```
wthrowe: Remove this whole instantiation section.  I think it isn't doing anything because this isn't the correct syntax (second and later arguments must be in parentheses) and the function in the `INSTANTIATION` macro doesn't match the function you're trying to instantiate.  Explicitly instantiating is unnecessary here because the function is only used in this file.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp
```diff
@@ -0,0 +1,141 @@
...
+      const Scalar<DataVector>& sqrt_det_spatial_metric,
+      const tnsr::ii<DataVector, 3, Frame::Inertial>& spatial_metric);
+};
+
+namespace Tags {
```
wthrowe, line 90: Include the tags above in the `Tags` namespace.

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp
```diff
@@ -0,0 +1,141 @@
...
+ * \f{align*}
+ *  \tilde{J}^i_\mathrm{parallel}
+ *    & = \alpha \sqrt{\gamma} \eta \left[
+ *        \frac{(E_lB^l)B^i}{B^2} + \frac{\mathcal{R}(E^2-B^2)}{B^2} E^i
+ *      \right]
+ * \f}
+ *
+ * where \f$\alpha\f$ is lapse, \f$\gamma\f$ is the determinant of the spatial
+ * metric, \f$E^i\f$ is the electric field, \f$B^i\f$ is the magnetic field,
+ * \f$\eta\f$ is the parallel conductivity, and \f$\mathcal{R}(x,0) = \max
```
wthrowe: How does the `R(E^2-B^2)` in the formula above map onto the `R(x, 0)` defined here?

tests/Unit/Evolution/Systems/ForceFree/Fluxes.py
```diff
@@ -3,9 +3,7 @@
 
 import numpy as np
 
-
-def levi_civita_symbol(i, j, k):
-    return np.sign(j - i) * np.sign(k - i) * np.sign(k - j)
+from TestFunctions import levi_civita_symbol
```
wthrowe: This is changed back in the next commit.  Mistake in squashing, maybe?

src/Evolution/Systems/ForceFree/Sources.cpp
```diff
@@ -32,23 +34,21 @@ void sources_impl(
...
     const tnsr::ii<DataVector, 3, Frame::Inertial>& extrinsic_curvature) {
   // S(\tilde{E}^i)
   raise_or_lower_index(source_tilde_e, d_lapse, inv_spatial_metric);
   for (size_t i = 0; i < 3; ++i) {
     source_tilde_e->get(i) -=
         get(lapse) * trace_spatial_christoffel_second.get(i);
     source_tilde_e->get(i) *= get(tilde_psi);
-    source_tilde_e->get(i) -= get(lapse) * get(sqrt_det_spatial_metric) *
-                              spatial_current_density.get(i);
+    source_tilde_e->get(i) += -tilde_j_drift.get(i);
```
wthrowe: `-=`?  Don't know if it matters for efficiency, but seems clearer.

src/Evolution/Systems/ForceFree/Sources.hpp
```diff
@@ -103,14 +103,14 @@ struct Sources {
       const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_e,
       const tnsr::I<DataVector, 3, Frame::Inertial>& tilde_b,
       const Scalar<DataVector>& tilde_psi, const Scalar<DataVector>& tilde_phi,
-      const Scalar<DataVector>& tilde_q,
-      const tnsr::I<DataVector, 3, Frame::Inertial>& spatial_current_density,
-      const double kappa_psi, const double kappa_phi,
+      const Scalar<DataVector>& tilde_q, const double kappa_psi,
+      const double kappa_phi, const double parallel_conductivity,
```
wthrowe: No `const`s in non-definitions.  I may have missed some more of these.

wthrowe (changes requested):

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.cpp
yoonso0-0: Good point. Thanks

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp
yoonso0-0: Sorry, that extra argument was typo!

src/Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp
yoonso0-0: @wthrowe Those two structs above are not `ComputeTag`s (don't have related SimpleTag and `base` type alias). Will it be clear if I remove the `Compute` prefix from names?

yoonso0-0: I pushed a fixup taking care of all your reviews only except for this. Let me know!

wthrowe: No, this is fine as-is.  Not sure why I thought they were tags.

yoonso0-0: Thanks

wthrowe:
Looks good.  Squash.

yoonso0-0:
Rebased on develop

wthrowe (approved):

Issue #4582, opened 2023-01-12
Improve element distribution with improved z-curve with weighting


geoffrey4444:
# Feature request:

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] Improve distribution of elements to account not only for distance from neighboring elements (to minimize communication costs) but also weight by the element's minimum grid spacing (which affects the element's cost).
- [ ] Detail 2
- [ ] Detail 3

### Detailed discussion:
Note: @nilsvu has explored some load balancing ideas [in this commit](https://github.com/nilsvu/spectre/commit/f6c2a9f13591eadb509101efa99e1dc3582ed2a7). 

The current algorithm uses a z curve to minimize communication costs, but this leads to large amounts of idle time, especially with local adaptive time stepping...because elements with more points must take smaller time steps (because of the CFL condition) and cost more per step (since there are more points).

The first idea is to introduce a weighting, or improved cutoffs, to try to get a better initial distribution of elements that is not as poorly balanced. The goal here is not to get the optimal load balancing, but to see how much improvement we can get with a first-cut, minimal improvement to the existing z-curve strategy.

nilsdeppe:
Current state: Nearly 2x speedup. @macedo22 is working on unit tests and code cleanup. This is great :D

geoffrey4444:
Done with #4713

PR #4713, opened 2023-02-09
Add and use weighted element distributions

macedo22:
Addresses #4582

## Proposed changes

This PR:
- Adds a weighted element distribution that weights elemental costs and attempts to spread costs evenly across processors, which is opposed to our current unweighted distribution (`BlockZCurveProcDistribution`), which assigns an even number of elements to each processor
- Updates `Evolution/DiscontinuousGalerkin/DgElementArray.hpp` to use this new weighted distribution as opposed to its current usage of the unweighted distribution (`BlockZCurveProcDistribution`)

The cost of each element is computed as `(number of grid points) / sqrt(minimum grid spacing in Frame::Grid)`.

To test the speedup of these changes, `EvolveGhBinaryBlackHole` was run for equal mass, non-spinning black holes to time 100M with initial time step 0.0002 on 4 of ocean's orca-1 nodes with 20 tasks per node and using blocks with p-refinements 9 and 10 and h-refinements 1 and 2 (different blocks with different refinements, the exact yaml used can be provided). Before these changes, this simulation took ~15.75 hours and after these changes, this simulation took ~9 hours, yielding a 75% speedup improvement for this particular simulation to get to 100M.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/ElementDistribution.cpp
+++ b/src/Domain/ElementDistribution.cpp
@@ -3,77 +3,287 @@
 
 #include "Domain/ElementDistribution.hpp"
 
-#include <algorithm>
 #include <array>
+#include <cmath>
 #include <cstddef>
-#include <numeric>
+#include <functional>
+#include <optional>
+#include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/IndexType.hpp"
+#include "Domain/Block.hpp"
+#include "Domain/CreateInitialElement.hpp"
+#include "Domain/ElementMap.hpp"
+#include "Domain/MinimumGridSpacing.hpp"
+#include "Domain/Structure/CreateInitialMesh.hpp"
+#include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/ElementId.hpp"
+#include "Domain/Structure/InitialElementIds.hpp"
 #include "Domain/Structure/ZCurve.hpp"
+#include "NumericalAlgorithms/Spectral/LogicalCoordinates.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+#include "Utilities/Algorithm.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Literals.hpp"
+#include "Utilities/Numeric.hpp"
 
 namespace domain {
+namespace {
+// \brief Get the cost of an `Element` computed as
+// `(number of grid points) / sqrt(minimum grid spacing in Frame::Grid)`
+//
+// \details As grid points in an `Element` increase, we expect the
+// computational cost of an `Element` to scale proportionally (if the minimum
+// grid spacing is held constant). In addition, the minimum grid spacing
+// between two points in an `Element` informs the time step that we take, where
+// the smaller the minimum spacing, the smaller time step we must take, which
+// means we expect computational work to scale inversely with the minimum grid
+// spacing.
+//
+// The reason that we use the square root of the spacing as opposed to just the
+// spacing in the denominator of the cost is that it was found experimentally
+// that using the square root yielded faster BBH simulation runtimes when using
+// local time stepping.
+template <size_t Dim>
+double get_num_points_and_grid_spacing_cost(
+    const ElementId<Dim>& element_id, const Block<Dim>& block,
+    const std::vector<std::array<size_t, Dim>>& initial_refinement_levels,
+    const std::vector<std::array<size_t, Dim>>& initial_extents,
+    const Spectral::Quadrature quadrature) {
+  Mesh<Dim> mesh = ::domain::Initialization::create_initial_mesh(
+      initial_extents, element_id, quadrature);
+  Element<Dim> element = ::domain::Initialization::create_initial_element(
+      element_id, block, initial_refinement_levels);
+  ElementMap<Dim, Frame::Grid> element_map{
+      element_id, block.is_time_dependent()
+                      ? block.moving_mesh_logical_to_grid_map().get_clone()
+                      : block.stationary_map().get_to_grid_frame()};
+  const tnsr::I<DataVector, Dim, Frame::ElementLogical> logical_coords =
+      logical_coordinates(mesh);
+  const tnsr::I<DataVector, Dim, Frame::Grid> grid_coords =
+      element_map(logical_coords);
+  const double min_grid_spacing =
+      minimum_grid_spacing(mesh.extents(), grid_coords);
+
+  return mesh.number_of_grid_points() / sqrt(min_grid_spacing);
+}
+}  //  namespace
+
+template <size_t Dim>
+std::unordered_map<ElementId<Dim>, double> get_element_costs(
+    const std::vector<Block<Dim>>& blocks,
+    const std::vector<std::array<size_t, Dim>>& initial_refinement_levels,
+    const std::vector<std::array<size_t, Dim>>& initial_extents,
+    const ElementWeight element_weight,
+    const std::optional<Spectral::Quadrature>& quadrature) {
+  std::unordered_map<ElementId<Dim>, double> element_costs{};
+
+  for (size_t block_number = 0; block_number < blocks.size(); block_number++) {
+    const auto& block = blocks[block_number];
+    const auto initial_ref_levs = initial_refinement_levels[block_number];
+    const std::vector<ElementId<Dim>> element_ids =
+        initial_element_ids(block.id(), initial_ref_levs);
+    const size_t grid_points_per_element = alg::accumulate(
+        initial_extents[block_number], 1_st, std::multiplies<size_t>());
+
+    for (const auto& element_id : element_ids) {
+      if (element_weight == ElementWeight::Uniform) {
+        element_costs.insert({element_id, 1.0});
+      } else if (element_weight == ElementWeight::NumGridPoints) {
+        element_costs.insert({element_id, grid_points_per_element});
+      } else {
+        ASSERT(element_weight == ElementWeight::NumGridPointsAndGridSpacing,
+               "Unknown element_weight");
+        ASSERT(quadrature.has_value(),
+               "Since element_weight is "
+               "ElementWeight::NumGridPointsAndGridSpacing, quadrature must "
+               "have a value");
+
+        element_costs.insert(
+            {element_id, get_num_points_and_grid_spacing_cost(
+                             element_id, block, initial_refinement_levels,
+                             initial_extents, quadrature.value())});
+      }
+    }
+  }
+
+  return element_costs;
+}
+
 template <size_t Dim>
 BlockZCurveProcDistribution<Dim>::BlockZCurveProcDistribution(
-    size_t number_of_procs_with_elements,
-    const std::vector<std::array<size_t, Dim>>& refinements_by_block,
+    const std::unordered_map<ElementId<Dim>, double>& element_costs,
+    const size_t number_of_procs_with_elements,
+    const std::vector<Block<Dim>>& blocks,
+    const std::vector<std::array<size_t, Dim>>& initial_refinement_levels,
+    const std::vector<std::array<size_t, Dim>>& initial_extents,
     const std::unordered_set<size_t>& global_procs_to_ignore) {
+  const size_t num_blocks = blocks.size();
+
+  ASSERT(
+      number_of_procs_with_elements > 0,
+      "Must have a non-zero number of processors to distribute elements to.");
+  ASSERT(num_blocks > 0, "Must have a non-zero number of blocks.");
+  ASSERT(
+      initial_refinement_levels.size() == num_blocks,
+      "`initial_refinement_levels` is not the same size as number of blocks");
+  ASSERT(initial_extents.size() == num_blocks,
+         "`initial_extents` is not the same size as number of blocks");
+
+  size_t num_elements = 0;
+  std::vector<size_t> num_elements_by_block(num_blocks);
+  for (size_t i = 0; i < num_blocks; i++) {
+    const size_t num_elements_current_block = two_to_the(alg::accumulate(
+        initial_refinement_levels[i], 0_st, std::plus<size_t>()));
+    num_elements_by_block[i] = num_elements_current_block;
+    num_elements += num_elements_current_block;
+  }
+
+  ASSERT(element_costs.size() == num_elements,
+         "`element_costs` is not the same size as the total number of elements "
+         "computed from `initial_refinement_levels`");
+
   block_element_distribution_ =
-      std::vector<std::vector<std::pair<size_t, size_t>>>(
-          refinements_by_block.size());
-  auto add_number_of_elements_for_refinement =
-      [](size_t lhs, const std::array<size_t, Dim>& rhs) {
-        size_t value = 1;
-        for (size_t i = 0; i < Dim; ++i) {
-          value *= two_to_the(gsl::at(rhs, i));
-        }
-        return lhs + value;
-      };
-  const size_t number_of_elements =
-      std::accumulate(refinements_by_block.begin(), refinements_by_block.end(),
-                      0_st, add_number_of_elements_for_refinement);
-  ASSERT(not refinements_by_block.empty(),
-         "`refinements_by_block` must be non-empty.");
-  // currently, we just assign uniform weight to elements. In future, it will
-  // probably be better to take into account p-refinement, but then the z-curve
-  // method will also require weighting.
-  size_t remaining_elements_in_block =
-      add_number_of_elements_for_refinement(0_st, refinements_by_block[0]);
-  size_t current_block = 0;
-  // This variable will keep track of how many global procs we've skipped over
-  // so far. This bookkeeping is necessary so the element gets placed on the
-  // correct global proc. The loop variable `i` does not correspond to global
-  // proc number. It's just an index
+      std::vector<std::vector<std::pair<size_t, size_t>>>(num_blocks);
+
+  std::vector<std::vector<ElementId<Dim>>> initial_element_ids_by_block(
+      num_blocks);
+  for (size_t i = 0; i < num_blocks; i++) {
+    initial_element_ids_by_block[i].reserve(num_elements_by_block[i]);
+    initial_element_ids_by_block[i] =
+        initial_element_ids(blocks[i].id(), initial_refinement_levels[i]);
+    alg::sort(initial_element_ids_by_block[i],
+              [](const ElementId<Dim>& lhs, const ElementId<Dim>& rhs) {
+                return z_curve_index(lhs) < z_curve_index(rhs);
+              });
+  }
+
+  double total_cost = 0.0;
+  for (const auto& element_id_and_cost : element_costs) {
+    total_cost += element_id_and_cost.second;
+  }
+
+  size_t current_block_num = 0;
+  size_t element_num_of_block = 0;
+  double cost_remaining = total_cost;
   size_t number_of_ignored_procs_so_far = 0;
-  for (size_t i = 0; i < number_of_procs_with_elements; ++i) {
+  // distribute Elements to all but the final proc
+  for (size_t i = 0; i < number_of_procs_with_elements - 1; ++i) {
     size_t global_proc_number = i + number_of_ignored_procs_so_far;
     while (global_procs_to_ignore.find(global_proc_number) !=
            global_procs_to_ignore.end()) {
       ++number_of_ignored_procs_so_far;
       ++global_proc_number;
     }
-    size_t remaining_elements_on_proc =
-        (number_of_elements / number_of_procs_with_elements) +
-        (i < (number_of_elements % number_of_procs_with_elements) ? 1 : 0);
-    while (remaining_elements_on_proc > 0) {
-      block_element_distribution_.at(current_block)
-          .emplace_back(std::make_pair(global_proc_number,
-                                       std::min(remaining_elements_in_block,
-                                                remaining_elements_on_proc)));
-      if (remaining_elements_in_block <= remaining_elements_on_proc) {
-        remaining_elements_on_proc -= remaining_elements_in_block;
-        ++current_block;
-        if (current_block < refinements_by_block.size()) {
-          remaining_elements_in_block = add_number_of_elements_for_refinement(
-              0_st, gsl::at(refinements_by_block, current_block));
+
+    // The target cost per proc is updated as we distribute to each proc since
+    // the total cost on a proc will nearly never be exactly the target average.
+    // If we don't adjust the target cost, then we risk either not using all
+    // procs (from overshooting the average too much on multiple procs) or
+    // piling up cost on the last proc (from undershooting the average on
+    // multiple procs). Updating the target cost per proc keeps the total cost
+    // spread somewhat evenly to each proc.
+    double target_cost_per_proc =
+        cost_remaining / static_cast<double>(number_of_procs_with_elements - i);
+    double cost_spent_on_proc = 0.0;
+    size_t total_elements_distributed_to_proc = 0;
+    bool add_more_elements_to_proc = true;
+    // while we haven't yet distributed all blocks and we still have cost
+    // allowed on the proc
+    while (add_more_elements_to_proc and (current_block_num < num_blocks)) {
+      const size_t num_elements_current_block =
+          num_elements_by_block[current_block_num];
+      size_t num_elements_distributed_to_proc = 0;
+      // while we still have elements left on the block to distribute and we
+      // still have cost allowed on the proc
+      while (add_more_elements_to_proc and
+             (element_num_of_block < num_elements_current_block)) {
+        const ElementId<Dim>& element_id =
+            initial_element_ids_by_block[current_block_num]
+                                        [element_num_of_block];
+        const double element_cost = element_costs.at(element_id);
+
+        if (total_elements_distributed_to_proc == 0) {
+          // if we haven't yet assigned any elements to this proc, assign the
+          // current element to the current proc to ensure it gets at least one
+          // element
+          cost_remaining -= element_cost;
+          cost_spent_on_proc = element_cost;
+          num_elements_distributed_to_proc = 1;
+          total_elements_distributed_to_proc = 1;
+          element_num_of_block++;
+        } else {
+          const double current_cost_diff =
+              abs(target_cost_per_proc - cost_spent_on_proc);
+          const double next_cost_diff =
+              abs(target_cost_per_proc - (cost_spent_on_proc + element_cost));
+
+          if (current_cost_diff <= next_cost_diff) {
+            // if the current proc cost is closer to the target cost than if we
+            // were to add one more element, then we're done adding elements to
+            // this proc and don't add the current one
+            add_more_elements_to_proc = false;
+          } else {
+            // otherwise, the current proc cost is farther from the target then
+            // if we were to add one more element, so we add the current element
+            // to the current proc
+            cost_spent_on_proc += element_cost;
+            cost_remaining -= element_cost;
+            num_elements_distributed_to_proc++;
+            total_elements_distributed_to_proc++;
+            element_num_of_block++;
+          }
         }
-      } else {
-        remaining_elements_in_block -= remaining_elements_on_proc;
-        remaining_elements_on_proc = 0;
+      }
+
+      // add a proc and its element allowance for the current block
+      block_element_distribution_.at(current_block_num)
+          .emplace_back(std::make_pair(global_proc_number,
+                                       num_elements_distributed_to_proc));
+      if (element_num_of_block >= num_elements_current_block) {
+        // if we're done assigning elements from the current block, move on to
+        // the elements in the next block
+        ++current_block_num;
+        element_num_of_block = 0;
       }
     }
+  }
+
+  // distribute all remaining Elements on the final proc
+
+  size_t global_proc_number =
+      number_of_procs_with_elements - 1 + number_of_ignored_procs_so_far;
+  while (global_procs_to_ignore.find(global_proc_number) !=
+         global_procs_to_ignore.end()) {
+    ++global_proc_number;
+  }
+
+  // distribute remaining Elements of Block we left off on
+  if (current_block_num < num_blocks) {
+    block_element_distribution_.at(current_block_num)
+        .emplace_back(std::make_pair(
+            global_proc_number,
+            num_elements_by_block[current_block_num] - element_num_of_block));
+  }
+
+  // distribute any Blocks that still remain after the Block we left off on
+  current_block_num++;
+  while (current_block_num < num_blocks) {
+    const size_t num_elements_current_block =
+        num_elements_by_block[current_block_num];
+    block_element_distribution_.at(current_block_num)
+        .emplace_back(
+            std::make_pair(global_proc_number, num_elements_current_block));
+    current_block_num++;
   }
 }
 
@@ -94,10 +304,26 @@ size_t BlockZCurveProcDistribution<Dim>::get_proc_for_element(
       "Processor not successfully chosen. This indicates a flaw in the logic "
       "of BlockZCurveProcDistribution.");
 }
+
 #define GET_DIM(data) BOOST_PP_TUPLE_ELEM(0, data)
 
-#define INSTANTIATION(r, data) \
-  template class BlockZCurveProcDistribution<GET_DIM(data)>;
+#define INSTANTIATION(r, data)                                               \
+  template class BlockZCurveProcDistribution<GET_DIM(data)>;                 \
+  double get_num_points_and_grid_spacing_cost(                               \
+      const ElementId<GET_DIM(data)>& element_id,                            \
+      const Block<GET_DIM(data)>& block,                                     \
+      const std::vector<std::array<size_t, GET_DIM(data)>>&                  \
+          initial_refinement_levels,                                         \
+      const std::vector<std::array<size_t, GET_DIM(data)>>& initial_extents, \
+      Spectral::Quadrature quadrature);                                      \
+  template std::unordered_map<ElementId<GET_DIM(data)>, double>              \
+  get_element_costs(                                                         \
+      const std::vector<Block<GET_DIM(data)>>& blocks,                       \
+      const std::vector<std::array<size_t, GET_DIM(data)>>&                  \
+          initial_refinement_levels,                                         \
+      const std::vector<std::array<size_t, GET_DIM(data)>>& initial_extents, \
+      ElementWeight element_weight,                                          \
+      const std::optional<Spectral::Quadrature>& quadrature);
 
 GENERATE_INSTANTIATIONS(INSTANTIATION, (1, 2, 3))
 
--- a/src/Domain/ElementDistribution.cpp
+++ b/src/Domain/ElementDistribution.cpp
@@ -11,86 +11,10 @@
 #include <vector>
 
 #include "Domain/Structure/ElementId.hpp"
-#include "Utilities/Algorithm.hpp"
+#include "Domain/Structure/ZCurve.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 
 namespace domain {
-
-namespace {
-// This interleaves the bits of the element index.
-// A sketch of a 2D block with 4x2 elements, with bit indices and resulting
-// z-curve
-//
-//        x-->
-//        00  01  10  11
-// y  0 |  0   2   4   6
-// |    |
-// v  1 |  1   3   5   7
-template <size_t Dim>
-size_t z_curve_index(const ElementId<Dim>& element_id) {
-  // for the bit manipulation of the element index, we need to interleave the
-  // indices in each dimension in order according to how many bits are in the
-  // index representation. This variable stores the refinement level and
-  // dimension index in ascending order of refinement level, representing a
-  // permutation of the dimensions
-  std::array<std::pair<size_t, size_t>, Dim>
-      dimension_by_highest_refinement_level;
-  for (size_t i = 0; i < Dim; ++i) {
-    dimension_by_highest_refinement_level.at(i) =
-        std::make_pair(element_id.segment_id(i).refinement_level(), i);
-  }
-  alg::sort(dimension_by_highest_refinement_level,
-            [](const std::pair<size_t, size_t>& lhs,
-               const std::pair<size_t, size_t>& rhs) {
-              return lhs.first < rhs.first;
-            });
-
-  size_t element_order_index = 0;
-
-  // 'gap' the lowest refinement direction bits as:
-  // ... x1 x0 -> ... x1 0 0 x0,
-  // then bitwise or in 'gap'ed and shifted next-lowest refinement direction
-  // bits as:
-  // ... y2 y1 y0 -> ... y2 0 y1 x1 0 y0 x0
-  // then bitwise or in 'gap'ed and shifted highest-refinement direction bits
-  // as:
-  // ... z3 z2 z1 z0 -> z3 z2 y2 z1 y1 x1 z0 y0 x0
-  // note that we must skip refinement-level 0 dimensions as though they are
-  // not present
-  size_t leading_gap = 0;
-  for (size_t i = 0; i < Dim; ++i) {
-    const size_t id_to_gap_and_shift =
-        element_id
-            .segment_id(
-                gsl::at(dimension_by_highest_refinement_level, i).second)
-            .index();
-    size_t total_gap = leading_gap;
-    if (gsl::at(dimension_by_highest_refinement_level, i).first > 0) {
-      ++leading_gap;
-    }
-    for (size_t bit_index = 0;
-         bit_index < gsl::at(dimension_by_highest_refinement_level, i).first;
-         ++bit_index) {
-      // This operation will not overflow for our present use of `ElementId`s.
-      // This technique densely assigns an ElementID a unique size_t identifier
-      // determining the Morton curve order, and `ElementId` supports refinement
-      // levels such that a global index within a block will fit in a 64-bit
-      // unsigned integer.
-      element_order_index |=
-          ((id_to_gap_and_shift & two_to_the(bit_index)) << total_gap);
-      for (size_t j = 0; j < Dim; ++j) {
-        if (i != j and
-            bit_index + 1 <
-                gsl::at(dimension_by_highest_refinement_level, j).first) {
-          ++total_gap;
-        }
-      }
-    }
-  }
-  return element_order_index;
-}
-}  // namespace
-
 template <size_t Dim>
 BlockZCurveProcDistribution<Dim>::BlockZCurveProcDistribution(
     size_t number_of_procs_with_elements,
--- a/src/Domain/ElementDistribution.hpp
+++ b/src/Domain/ElementDistribution.hpp
@@ -5,56 +5,80 @@
 
 #include <array>
 #include <cstddef>
+#include <optional>
+#include <unordered_map>
 #include <unordered_set>
 #include <utility>
 #include <vector>
 
-#include "Domain/Structure/ElementId.hpp"
+template <size_t Dim>
+class Block;
+
+template <size_t Dim>
+class ElementId;
+
+namespace Spectral {
+enum class Quadrature;
+}  // namespace Spectral
 
 namespace domain {
+/// The weighting scheme for assigning computational costs to `Element`s for
+/// distributing balanced compuational costs per processor (see
+/// `BlockZCurveProcDistribution`)
+enum class ElementWeight {
+  /// A weighting scheme where each `Element` is assigned the same computational
+  /// cost
+  Uniform,
+  /// A weighting scheme where each `Element`'s computational cost is equal to
+  /// the number of grid points in that `Element`
+  NumGridPoints,
+  /// A weighting scheme where each `Element`'s computational cost is weighted
+  /// by both the number of grid points and minimum spacing between grid points
+  /// in that `Element` (see `get_num_points_and_grid_spacing_cost()` for
+  /// details)
+  NumGridPointsAndGridSpacing
+};
+
+/// \brief Get the cost of each `Element` in a list of `Block`s where
+/// `element_weight` specifies which weight distribution scheme to use
+///
+/// \details It is only necessary to pass in a value for `quadrature` if
+/// the value for `element_weight` is
+/// `ElementWeight::NumGridPointsAndGridSpacing`. Otherwise, the argument isn't
+/// needed and will have no effect if it does have a value.
+template <size_t Dim>
+std::unordered_map<ElementId<Dim>, double> get_element_costs(
+    const std::vector<Block<Dim>>& blocks,
+    const std::vector<std::array<size_t, Dim>>& initial_refinement_levels,
+    const std::vector<std::array<size_t, Dim>>& initial_extents,
+    ElementWeight element_weight,
+    const std::optional<Spectral::Quadrature>& quadrature);
 
 /*!
  * \brief Distribution strategy for assigning elements to CPUs using a
  * Morton ('Z-order') space-filling curve to determine placement within each
- * block.
+ * block, where `Element`s are distributed across CPUs
  *
- * \details The element distribution assigns a balanced number of elements to
- * each processor that is allowed to have elements (default all). Specify which
- * processors aren't allowed to have elements by passing in an unordered set of
- * `size_t`s corresponding to the processor number. This distribution is
- * computed by first greedily assigning to each available processor an allowance
- * of [total number of elements]/[number of processors available] elements from
- * one or more blocks, starting with the lowest number block that still has
- * elements to contribute to an allowance. Then, once those allowances are
- * determined, a separate Z-order curve is established for each block and the
- * elements are assigned to processors within each block by greedily filling
- * each available processors' allowance by contiguous intervals along the
- * Z-order curve. Some examples:
- * - If there are 8 blocks, 16 elements per block, 16 cores, and all cores are
- * allowed to have elements: each core gets an allowance of 128 / 16 = 8
- * elements, so each core gets half of a block, and the 8 elements for each core
- * within the block are chosen via Z-order curve for the respective blocks.
- * - If there are 3 blocks, 4 elements per block, 4 cores, and all cores are
- * allowed to have elements: each core gets an allowance of 12 / 4 = 3 elements.
- * Core 0 gets three elements from the first block, core 1 gets one element from
- * the first block and two elements from the second block, core 2 gets two
- * elements from the second block and one from the third, and core 3 gets the
- * remaining three elements from the third block. Each collection of elements
- * within the blocks are then assigned using intervals along the Z-order curve
- * for each block.
- * - Same as the previous example, 3 blocks, 4 elements per block, and 4 cores,
- * except now we require that physical cores 1 and 3 don't have any elements on
- * them. The new distribution would look like:
- *   - Elements on old core 0 -> new core 0
- *   - No elements on new core 1
- *   - Elements on old core 1 -> new core 2
- *   - No elements on new core 3
- *   - Elements on old core 2 -> new core 4
- *   - Elements on old core 3 -> new core 5
- *
- * \note In the third example, even though only 4 cores are used to place
- * elements, the simulation is required to be run on at least 6 cores (4 cores
- * for elements + 2 cores without elements)
+ * \details The element distribution attempts to assign a balanced total
+ * computational cost to each processor that is allowed to have `Element`s.
+ * First, each `Block`'s `Element`s are ordered by their Z-curve index (see more
+ * below). `Element`s are traversed in this order and assigned to CPUs in order,
+ * moving onto the next CPU once the target cost per CPU is met. The target cost
+ * per CPU is defined as the remaining cost to distribute divided by the
+ * remaining number of CPUs to distribute to. This is an important distinction
+ * from simply having one constant target cost per CPU defined as the total cost
+ * divided by the total number of CPUs with elements. Since the total cost of
+ * `Element`s on a processor will nearly never add up to be exactly the average
+ * cost per CPU, this means that we would either have to decide to overshoot or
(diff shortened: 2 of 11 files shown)


src/Domain/WeightedElementDistribution.hpp
```diff
@@ -0,0 +1,147 @@
...
+
+ private:
+  // in this nested data structure:
+  // - The block id is the first index
+  // - There is an arbitrary number of CPUs per block, each with an element
+  //   allowance
+  // - Each element allowance is represented by a pair of proc number, number of
+  //   elements in the allowance
+  //   const size_t number_of_procs_with_elements_;
+  //   const std::unordered_set<size_t>& global_procs_to_ignore_;
```
knelli2: Remove?

wthrowe:
Is this optimized for global time-stepping or local time-stepping?

src/Domain/ZCurve.cpp
```diff
@@ -80,11 +83,176 @@ size_t z_curve_index_from_element_id(const ElementId<Dim>& element_id) {
...
+    gsl::at(result_segment_indices,
+            gsl::at(segment_indices_by_highest_refinement_level, i).second) =
+        gsl::at(segment_indices_by_highest_refinement_level, i).first;
+  }
+
+  return result_segment_indices;
+}
+
+template <>
+std::vector<ElementId<1>> initial_element_ids_in_z_curve_order<1>(
```
knelli2: These 3 template specializations seem like they can be combined into one function template with a couple `for` loops. Then you can instantiate them below when you generate other instantiations.

src/Domain/ZCurve.hpp
```diff
@@ -26,4 +28,42 @@ namespace domain {
...
+/// index and block refinements
+///
+/// \details The `Segment` indices are computed by extracting the bits of the
+/// `Segment` indices from the Z-curve index. Here is a sketch of a 2D block
+/// with 4x2 elements, with bit indices and the resulting z-curve:
+///
+/// \code
+///        x-->
+///        00  01  10  11
+/// y  0 |  0   2   4   6
```
knelli2: This is the same example as `z_curve_index_from_element_id` above. Could you maybe elaborate a little more on what the example shows/demonstrates?

src/Domain/WeightedElementDistribution.cpp
```diff
@@ -0,0 +1,240 @@
...
+          initial_extents, element_id, quadrature);
+      Element<Dim> element = ::domain::Initialization::create_initial_element(
+          element_id, block, initial_refinement_levels);
+      ElementMap<Dim, Frame::Grid> element_map{
+          element_id, block.is_time_dependent()
+                          ? block.moving_mesh_logical_to_grid_map().get_clone()
+                          : block.stationary_map().get_to_grid_frame()};
+
+      tnsr::I<DataVector, Dim, Frame::ElementLogical> logical_coords{};
+      domain::Tags::LogicalCoordinates<Dim>::function(
```
knelli2: Can you not just call the free function here rather than the function in the tag?

src/Domain/WeightedElementDistribution.cpp
```diff
@@ -0,0 +1,240 @@
...
+          make_not_null(&logical_coords), mesh);
+
+      tnsr::I<DataVector, Dim, Frame::Grid> grid_coords{};
+      domain::Tags::MappedCoordinates<
```
knelli2: If you look at the function for this tag, all it does is
```cpp
*target_coords = element_map(source_coords);
```
I think you can just do that yourself rather than calling into the compute tag function. It will look simpler.

src/Domain/WeightedElementDistribution.cpp
```diff
@@ -0,0 +1,240 @@
...
+          domain::Tags::ElementMap<Dim, Frame::Grid>,
+          domain::Tags::Coordinates<Dim, Frame::ElementLogical>>::
+          function(make_not_null(&grid_coords), element_map, logical_coords);
+
+      double minimum_grid_spacing =
+          std::numeric_limits<double>::signaling_NaN();
+      domain::Tags::MinimumGridSpacingCompute<Dim, Frame::Grid>::function(
```
knelli2: Again, just call the function? Don't go through the compute tag.

knelli2 (changes requested):
This is great Alex! Thank you for working on this. Just a couple of suggestions

src/Domain/WeightedElementDistribution.hpp
macedo22: Good catch, thanks

src/Domain/ZCurve.cpp
macedo22: I almost did this instead but then didn't just because I thought the combined one looked ugly with `if constexpr` inside the for loop 😂 but I don't have a strong opinion either way and will update it to the way you're describing

nilsdeppe: "Always" prefer `if constexpr`. It's easier to read and reason about than specializations. The quotes is if the entire function body is different then specializations are better

src/Domain/ZCurve.hpp
macedo22: I'm not sure I'm understanding fully, but if you're asking what new/different thing I'm trying to demonstrate by borrowing the same example, I will say that I'm not trying to say anything different than what `z_curve_index_from_element_id ` says via the example. The shared example just shows the correspondence between `ElementId`s and Z-curve index, i.e. how they map to each other, and `z_curve_index_from_element_id ` and `segment_indices_from_z_curve_index` are both the implementation of the mapping, just in opposite/inverse directions.

If I'm misunderstanding what you mean, let me know! I also just basically copy-pasted the documentation that already existed, so if you feel that more should be said than what was already there, please let me know specifically what would be helpful for me to expand on, e.g. if it's not clear that the example is showing a mapping from elements to z-curve index.

src/Domain/WeightedElementDistribution.cpp
macedo22: Yes, the reason I didn't call the function was just because I felt that going through the tag gave it a bit more clarity what compute item I was computing so that someone reading could quickly associate the logic to the tag. I don't have a strong opinion on this, though

src/Domain/WeightedElementDistribution.cpp
macedo22: Yes, I can just call the function

src/Domain/WeightedElementDistribution.hpp
```diff
@@ -0,0 +1,147 @@
...
+  /// The `number_of_procs_with_elements` argument represents how many procs
+  /// will have elements. This is not necessarily equal to the total number of
+  /// procs because some global procs may be ignored by the sixth argument
+  /// `global_procs_to_ignore`.
+  WeightedBlockZCurveProcDistribution(
+      size_t number_of_procs_with_elements,
+      const std::vector<Block<Dim>>& blocks,
+      const std::vector<std::array<size_t, Dim>>& initial_refinement_levels,
+      const std::vector<std::array<size_t, Dim>>& initial_extents,
+      const Spectral::Quadrature quadrature,
```
wthrowe: No `const` on non-reference in non-definitions.  Again in the next function.

src/Domain/WeightedElementDistribution.hpp
```diff
@@ -0,0 +1,147 @@
...
+      const std::vector<std::array<size_t, Dim>>& initial_extents,
+      const Spectral::Quadrature quadrature);
+
+  /// Gets the suggested processor number for a particular `ElementId`,
+  /// determined by the Morton curve weighted element assignment described in
+  /// detail in the parent class documentation.
+  size_t get_proc_for_element(const ElementId<Dim>& element_id) const;
+
+  std::vector<std::vector<std::pair<size_t, size_t>>>
+  block_element_distribution() const {
```
wthrowe: Return by (const) reference?

src/Domain/WeightedElementDistribution.cpp
macedo22: It's funny you say this because I also almost just stuck in this line instead of calling the function and the reason I called the function (regardless of how simple) is again just for clarity via association with known compute items, but I don't have a strong opinion on this either and can change it

src/Domain/ZCurve.cpp
```diff
@@ -80,11 +83,176 @@ size_t z_curve_index_from_element_id(const ElementId<Dim>& element_id) {
...
+  for (size_t i = 0; i < Dim; ++i) {
+    dimension_by_highest_refinement_level.at(i) =
+        std::make_pair(gsl::at(initial_ref_levs, i), i);
+  }
+  // dimensions in order of ascending refinement
+  alg::sort(dimension_by_highest_refinement_level,
+            [](const std::pair<size_t, size_t>& lhs,
+               const std::pair<size_t, size_t>& rhs) {
+              return lhs.first < rhs.first;
+            });
```
wthrowe: This is basically what you get from a plain `pair < pair` (except it checks the second element in case of a tie), so you can leave out the comparator.  That will also guarantee a consistent order when `lhs.first == rhs.first`, which is otherwise unspecified.

tests/Unit/Domain/Test_WeightedElementDistribution.cpp
```diff
@@ -0,0 +1,587 @@
...
+         stringize(use_equiangular_map) + "\n" + time_dependence;
+}
+
+// Test the computation of the weighting done by
+// `domain::WeightedBlockZCurveProcDistribution::get_cost_by_element_by_block`
+void test_cost_function() {
+  const auto domain_creator1 = make_domain_creator<3>(
+      "AlignedLattice:\n"
+      "  BlockBounds: [[0.0, 1.0, 2.0], [0.0, 1.0], [0.0, 1.0]]\n" +
+      std::string{"  IsPeriodicIn: [false, false, false]\n"} +
```
wthrowe: This can just be part of the surrounding literal, right?  Again for the others.

tests/Unit/Domain/Test_WeightedElementDistribution.cpp
```diff
@@ -0,0 +1,587 @@
...
+      "  InitialGridPoints: [4, 4, 4]\n"
+      "  InitialLevels: [2, 1, 0]\n"
+      "  RefinedLevels: []\n"
+      "  RefinedGridPoints: []\n"
+      "  BlocksToExclude: []\n");
+  const auto* aligned_blocks_creator1 =
+      dynamic_cast<const domain::creators::AlignedLattice<3>*>(
+          domain_creator1.get());
```
wthrowe: Do you need the cast here?  I think you're only calling virtual functions from the base class.  Again for the others below.

tests/Unit/Domain/Test_WeightedElementDistribution.cpp
```diff
@@ -0,0 +1,587 @@
...
+          aligned_blocks_creator2->initial_extents(),
+          Spectral::Quadrature::GaussLobatto);
+
+  const auto costs3 = domain::WeightedBlockZCurveProcDistribution<3>::
+      get_cost_by_element_by_block(
+          blocks3, aligned_blocks_creator3->initial_refinement_levels(),
+          aligned_blocks_creator3->initial_extents(),
+          Spectral::Quadrature::GaussLobatto);
+
+  Approx custom_approx_e16 = Approx::custom().epsilon(1.0e-15).scale(1.0);
```
wthrowe: `e16` -> `e15`, if I'm interpreting the name correctly.  (This is actually quite a bit stricter than our default `approx`, by the way.)

tests/Unit/Domain/Test_WeightedElementDistribution.cpp
```diff
@@ -0,0 +1,587 @@
...
+  const auto domain = domain_creator.create_domain();
+  const auto& blocks = domain.blocks();
+  const auto initial_refinement_levels =
+      domain_creator.initial_refinement_levels();
+  const auto initial_extents = domain_creator.initial_extents();
+
+  const size_t num_blocks = blocks.size();
+  size_t num_elements = 0;
+  std::vector<size_t> num_elements_by_block(num_blocks);
+  std::fill(num_elements_by_block.begin(), num_elements_by_block.end(), 0);
```
wthrowe: `std::vector<size_t> num_elements_by_block(num_blocks, 0);` will initialize with zeros.

tests/Unit/Domain/Test_WeightedElementDistribution.cpp
```diff
@@ -0,0 +1,587 @@
...
+      global_procs_to_ignore);
+  const auto proc_map = element_distribution.block_element_distribution();
+
+  const size_t total_procs =
+      number_of_procs_with_elements + global_procs_to_ignore.size();
+  std::vector<size_t> num_elements_by_proc(total_procs);
+  std::fill(num_elements_by_proc.begin(), num_elements_by_proc.end(), 0);
+  std::vector<size_t> actual_num_elements_by_block_in_dist(num_blocks);
+  std::fill(actual_num_elements_by_block_in_dist.begin(),
+            actual_num_elements_by_block_in_dist.end(), 0);
```
wthrowe: Initializing constructors, as above.  Check all uses of `std::fill`.

tests/Unit/Domain/Test_ZCurve.cpp
```diff
@@ -377,11 +381,129 @@ void test_z_curve_index_from_element_id_3d() {
...
+      gsl::at(expected_segment_indices, i) = element_id.segment_id(i).index();
+    }
+
+    CHECK(result_segment_indices == expected_segment_indices);
+
+    z_curve_index_hit[result_z_curve_index] = true;
+  }
+
+  // Check that there is a 1:1 mapping of ElementId to Z-curve index
+  for (const size_t index_hit : z_curve_index_hit) {
```
wthrowe: `bool`

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -40,12 +43,12 @@ CREATE_HAS_STATIC_MEMBER_VARIABLE_V(use_z_order_distribution)
  * `PhaseDepActionList`.
  *
  * The element assignment to processors is performed by
- * `domain::BlockZCurveProcDistribution` (using a Morton space-filling curve),
- * unless `static constexpr bool use_z_order_distribution = false;` is specified
- * in the `Metavariables`, in which case elements are assigned to processors via
- * round-robin assignment. In both cases, an unordered set of `size_t`s can be
- * passed to the `allocate_array` function which represents physical processors
- * to avoid placing elements on.
+ * `domain::WeightedBlockZCurveProcDistribution` (using a Morton space-filling
```
wthrowe: Has this been tested for global time-stepping?  It's not obvious that the weighted version will be better than the unweighted one in that case, and I think GTS the most commonly used evolution mode.  Perhaps we should keep the old behavior be default and make the weighted version opt-in, or at least give an option to opt out.

src/Domain/ZCurve.hpp
knelli2: > The shared example just shows the correspondence between `ElementId`s and Z-curve index, i.e. how they map to each other, and `z_curve_index_from_element_id ` and `segment_indices_from_z_curve_index` are both the implementation of the mapping, just in opposite/inverse directions.

Ohhh! This is what I was missing. That they are basically just opposites of each other. Would you be able to add that to the docs of both functions so it's clear there's a correspondence between them? The function names do hint towards it but it's not completely obvious.

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -4,25 +4,28 @@
 #pragma once
```
wthrowe, line 4: [Random line] The documentation needs to indicate that this was developed on a local-time-stepping evolution.

If we have data on it, determining whether this works well for global time-stepping (and subcell) would be good, but that's a lot of extra work and is certainly not required for this PR.

wthrowe (changes requested):

src/Domain/WeightedElementDistribution.cpp
knelli2: The free function is called `logical_coordinates` and has its own docs (the tag docs actually aren't super helpful). I don't think using the tag makes it clearer in this case.

src/Domain/WeightedElementDistribution.cpp
knelli2: I totally get your reasoning, but this is a lot of templates and tags for something that doesn't actually use tags, so I think just using `element_map(source_coords)` is clear enough (same as `logical_coords` above)

nilsvu: yes please, use the free functions :)

src/Domain/ZCurve.hpp
macedo22: Yep, thanks for letting me know that's not obvious!

tests/Unit/Domain/Test_ZCurve.cpp
macedo22: good catch!

src/Domain/WeightedElementDistribution.hpp
```diff
@@ -0,0 +1,147 @@
...
+ * \brief Distribution strategy for assigning elements to CPUs using a
+ * Morton ('Z-order') space-filling curve to determine placement within each
+ * block, where elements are assigned weighted costs that are distributed
+ * across CPUs
+ *
+ * \details The element distribution attempts to assign a balanced total
+ * computational cost to each processor that is allowed to have elements. The
+ * cost assigned to a processor is defined as the sum of the costs of the
+ * `Element`s assigned to a processor, and the cost of an `Element` is defined
+ * as `(number of grid points) / sqrt(minimum grid spacing in Frame::Grid)`.
```
nilsvu: Would it be possible with reasonable effort to do the following two things?

- Allow the consumer of this function to pass in the list of all element IDs instead of blocks + refinement levels. This would allow us to periodically re-distribute the elements when we do AMR, as an alternative to Charm's dynamic load balancers (which we're having some trouble with). I expect this will be quite important, and it's probably easier to set up the code to work with element IDs rather than global refinement levels now than to refactor it later.
- Also allow to pass in the weight for each element, so we can experiment with different weightings, instead of hardcoding the min grid spacing. My understanding is that the distribution algorithm is agnostic to the cost function, so decoupling the two would be very helpful. The min grid spacing weighting function can just be a free function that computes the weight given an element ID, mesh, etc.

nilsvu (changes requested):

src/Domain/WeightedElementDistribution.hpp
macedo22: - Yes, the reason I didn't have it accept the list of `ElementId`s as a parameter for the constructor of the distribution is that internally, the weighted distribution needs the elements ordered by z-curve index, so that either requires the caller to correctly call `initial_element_ids_in_z_curve_order` and pass that correct element order into the weighted distribution constructor (which could just assume its input is in the right order, which doesn't seem super safe), *or* we assume the caller would pass in a list of elements in an arbitrary order and internally the weighted distribution would have to sort them by z-curve index to ensure the right ordering is used. In other words, constructing the element list in z-curve order internally guarantees that we use the elements in z-curve order. I think having the constructor or an `update` function take in the list of elements is fine so long as we have a means of ensuring the element ordering used for the distribution is correct (whether that requires we internally sort the input element list or store the elements in z-curve order somewhere and we *know* the input element list is already sorted). Let me know your thoughts on this!
- I agree this is good in principle and the only hangup I have is the same thing as above - that the caller would need to pass in the costs already ordered in z-curve order or the the weighted distribution would internally have to sort the input costs (in which case, costs would need to be paired up with ElementIds).

So based on this, perhaps it seems like one of these things should happen:
- store the ElementIds in z-curve order somewhere so callers can use this and not worry about ordering it correctly
- internally sort the input element lists and costs so callers don't have to worry about ordering it correctly
- anything else that ensures the elements are ordered correctly and we don't have to police each call site to provide the correct order

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
macedo22: I have not tested this for GTS, but I'm happy to make it so that it only uses the weighted case if we're using LTS, for now while we don't know its impact when using GTS. Would I just need to check the value of `local_time_stepping` in the Metavariables or is there more to it?

src/Domain/WeightedElementDistribution.hpp
kidder: Since, as far as I can tell,  ZCurve assumes all Elements within a Block have the same refinement level in a given dimension, I don't see the point of passing in ElementIds...

nilsvu: Oh is that a hard requirement? It would be very nice to be able to load-balance using the weighted z-curve after a few AMR iterations.

kidder: I haven't looked into space-filling curves, so I don't know how hard it is to implement one that doesn't have that assumption.  But the current `z_curve_index_from_element_id` implementation is clearly not invertible (i.e. multiple ElementIds have the same z_curve_index) if this assumption is broken.

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
wthrowe: Pretty much.  If there is an easy way to template this code so the choice can be made independently from the time stepping that might be easier for people to experiment with.

src/Domain/WeightedElementDistribution.hpp
wthrowe: You could pass a functor for computing the weight given an element, but I don't know if it would be easy for the functor to actually access anything interesting.

knelli2: After reading through what will be necessary to use this as a basic load balance after AMR, I think it might be best to defer these changes to a later PR. This one is already very long (out of necessity), and I think @geoffrey4444 and I would like to get this in as quick as possible so our BBH runs take less time. Does this seem reasonable @nilsvu @kidder?

p.s. I think the changes suggested here are good and we should definitely do some (all?) of them. Just eventually. Not in this PR.

kidder: Yes, I agree.

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
macedo22: Just want to make sure I understand correctly. My current understanding is that Metavariables for some evolution has a `static constexpr bool use_local_time_stepping`, like in `EvolveGhBinaryBlackHole.hpp`. I could just use an `if constexpr` in `DgElementArray.hpp` to check `Metavariables::use_local_time_stepping` and do the unweighted distribution if it's false and use the weighted if it's true. However, since this value is set as `true` in `EvolveGhBinaryBlackHole.hpp`, that sounds like BBH runs would always use the weighted distribution. Since you're saying that you want people to have the option to opt-in to the weighted distribution, should I be deciding  which distribution to use based on something else that the user specifies in the yaml? Or do I have a misunderstanding of how/when LTS/GTS is decided for an executable?

knelli2: I think Will means instead of using time stepping (and the `metavars::use_local_time_stepping` bool) as the determining factor for whether to do weighted vs unweighted distributions, just add a template bool to `DgElementArray` called `UseWeightedZCurve` and use that (maybe default it to false??). That way if, for some reason, we need to turn off weighted distributions (for any executable, not just BBH) we don't also have to turn off LTS.

macedo22: Ok I can add that tparam to `DgElementArray`. In that case, should I just make the default be `false` and only update `EvolveGhBinaryBlackHole` metavars to use `true`? Or are there other executable metavars that I should also tell to use the weighted distribution (pass `true` to `DgElementArray` to use the weighted distribution)?

wthrowe: If you you're up for a bunch of mindless file updating, setting it to `use_local_time_stepping` everywhere is probably a good start.  While you've only tested on BBH, I would be surprised if this was not a big improvement on any LTS evolution.

I'm envisioning that we may end up with multiple distribution algorithms and this option could eventually evolve into some enum or something.  Maybe a distribution weighted by the number of points but leaving out the spacing factor, for example.  Or something that can handle nonuniform refinement.

macedo22: Sure, I'll do that! I agree with the vision for a few different distributions, one of them being just based on the number of grid points

nilsvu: For elliptic solves I think the weight should be based only on the number of points, not on the spacing factor (that's what I did for my thesis because there is no CFL condition), so allowing for different distributions would be very helpful. Passing in a runtime(!) bool to disable the spacing factor would be enough for now and probably quite easy to add. Making the bool an enum with cases `Weight::Uniform`, `Weight::MinGridSpacing`, `Weight::NumPoints` would probably also be quite easy and relatively future-proof, so I suggest that. Does that good to everyone?

src/Domain/WeightedElementDistribution.hpp
nilsdeppe: A general Z curve does not assume uniform refinement, but our implementation currently does. It will not work with intra-block h-refinement. To me this makes it better to deal with the manual LB in a completely separate PR, because a lot will need to change for that.

src/Domain/WeightedElementDistribution.hpp
```diff
@@ -0,0 +1,147 @@
...
+ * First, each `Block`'s `Element`s are ordered by their Z-curve index (see more
+ * below). `Element`s are traversed in this order and assigned to CPUs in order,
+ * moving onto the next CPU once the target cost per CPU is met. The target cost
+ * per CPU is defined as the remaining cost to distribute divided by the
+ * remaining number of CPUs to distribute to. This is an important distinction
+ * from simply having one constant target cost per CPU defined as the total cost
+ * divided by the total number of CPUs with elements. Since elemental costs will
```
nilsdeppe: what is an elemental cost?

src/Domain/WeightedElementDistribution.hpp
nilsdeppe: Where does `sqrt(min grid spacing)` come from? Isn't the time step size given by the minimum grid spacing? Could you add a (brief!) explanation into the dox as to where this comes from?

src/Domain/WeightedElementDistribution.hpp
```diff
@@ -0,0 +1,147 @@
...
+class ElementId;
+
+namespace Spectral {
+enum class Quadrature;
+}  // namespace Spectral
+
+namespace domain {
+
+/*!
+ * \brief Distribution strategy for assigning elements to CPUs using a
```
nilsdeppe: I'm not sure I really have a change request here for this PR, but it's not very encouraging to me that none of this is directly applicable to DG-FD hydro simulations. Those have quite a bit of load imbalance from switching between different schemes. The challenge in that case is we don't know if we need to do FD until _after_ we've initialized the element. This would mean more of a load balance than an initial element distribution. I think that being able to actually balance those simulations is extremely high priority. We can merge this PR without that, but it won't help the majority of our current science applications at all, unfortunately. Nils Vu's requests for some way of specifying a per-element weight is necessary in that case.

nilsdeppe (changes requested):

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
knelli2: The elliptic DgElementArray still uses the original element distribution. That wasn't changed in the PR. I recommend we hold off on adding even more things to this PR besides the already long list of change requests (and the template bool below). A follow up PR can do run-time things, add `enum`s, and potentially edit the elliptic DgElementArray if needed.

nilsvu: No need to change the elliptic DgElementArray in this PR of course (I didn't request that), but supporting different weight functions is a design question for this PR. Doesn't have to be implemented in this PR, but we should have a plan for the implementation. We don't want to refactor this completely later to support a feature we surely need.

knelli2: Ah ok I understand now. Yeah an option-creatable `enum` seems good as an alternative to the template bool.

macedo22: @nilsvu So I understand what is being asked of for this PR (so I can push fixups), which function are you suggesting take a bool/enum (since I don't think I'm allowed to change the signature of `DgElementArray::allocate_array()`)? Or based on what Kyle said, am I being asked to create a new option that is parsed from the yaml files and grabbed from the cache or something? I understand we are wanting to be forward-thinking about the design to accommodate features we want to add, I'm just unclear what I'm being asked to do for this PR regarding the runtime bool/enum

src/Domain/WeightedElementDistribution.hpp
macedo22: Yep! By "where it comes from" are you wanting me to comment on why minimum grid spacing is being used or why the square root is there? Or both?

src/Domain/WeightedElementDistribution.hpp
macedo22: `elemental cost` just refers to the computed cost of an `Element`. Since I previously defined how the cost is computed, I'm guessing your question means that `elemental costs` as a phrase wasn't clearly referring to the cost of `Element`s or the phrasing didn't clearly refer to the total cost of elements on a processor. Would it be more clear if I said `Since the total cost of Elements on a processor` or am I misinterpreting your question?

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
nilsvu: Can you just make the constructor of the element distribution take a bool/enum? That would be enough I think. Option-parsing etc can be done later.

macedo22: I can do that, but then I think that arg for `WeightedElementDistribution` would only make sense for specifying `Weight::NumPointsOverSqrtMinGridSpacing` or `Weight::NumPoints`, while the `Weight::Uniform` one wouldn't make sense as an option for this constructor since we instead construct `ElementDistribution` to do that, meaning if we want to have logic that decides between weighted (`WeightedElementDistribution`) vs unweighted (`ElementDistribution`), I think we need that bool/enum  checked before we decide which distribution to construct between weighted and unweighted. Either that or we need to merge the two distribution classes, maybe. Let me know what you think and if this made sense. Basically, I can do what you're suggesting, I just think that doesn't cover Will's request to make weighted/unweighted toggleable

src/Domain/WeightedElementDistribution.hpp
nilsdeppe: Yea, you've got it! Yes, `Since the total cost of Elements on a...` would've cleared up my confusion!

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
nilsvu: I thought that the weighted distribution just replaces the unweighted, and reduces to the unweighted for the `Weight::Uniform` enum case. If deleting the unweighted distribution is too much work for this PR then don't worry about it, but I thought that was the plan?

nilsvu:
Notes from call on Feb 14, 2023:

Present: @kidder @nilsdeppe @wthrowe @macedo22 @nilsvu @knelli2 

- Add argument `const std::unordered_map<ElementId, double>& weights` to the constructor of the element distribution.
- Delete the unweighted element distribution. The weighted distribution replaces it entirely. The code goes into the existing `Domain/ElementDistribution.hpp`.
- Two (maybe three) cost functions can go into `Domain/ElementDistribution.hpp`:

    1. weight by num points
    2. weight by num points + grid spacing
    3. uniform weight

  These functions take whatever data they need and return a `std::unordered_map<ElementId, double>` of weights. These can also be just one function where the result depends on the passed-in arguments (whatever is easiest). More specialized weight functions can be defined in other places if needed to build the global `std::unordered_map<ElementId, double>` of weights.
- Assert in the element distribution that the map of weight has an entry for each element.

Work for later (but probably quite soon):

- The algorithm will be generalized later to support non-uniform h-refinement. At that point we can remove the arguments `initial_refinement` and `initial_extents` from the constructor of the element distribution and just get the list of element IDs from the map of weights.
- Anything with dynamic load balancing will be dealt with later. It will probably build the map of weights from a reduction.
- Hilbert curves have been found to improve the performance considerably over the Morton (z-order) curve. We probably want that too, but priority is lower than non-uniform h-refinement. Refs:
  - https://github.com/N-BodyShop/utility/blob/master/structures/SFC.h
  - https://github.com/AMReX-Codes/amrex/blob/50f8e4314be75003610378d159134960ad27e64b/Src/Base/AMReX_DistributionMapping.cpp#L915

macedo22:
I kept the first commit intact (there weren't any change requests for that commit), but I had to redo the commits after that because fixups wouldn't really cut it. However, I believe that I addressed all of the change requests from the first review.

tests/Unit/Domain/Test_ElementDistribution.cpp
```diff
@@ -3,306 +3,810 @@
...
+    test_weighted_element_distribution(element_weight, *lattice_1d, *lattice_2d,
+                                       *binary_compact_object_creator);
   }
+
+  test_proc_retrieval(element_weight, *lattice_1d, *lattice_2d,
+                      *binary_compact_object_creator);
 }
 }  // namespace
+
+SPECTRE_TEST_CASE("Unit.Domain.ElementDistribution", "[Domain][Unit]") {
```
nilsdeppe: I know you like to be very thorough with your tests, but this is timing out. Is it possible to reduce the runtime of the tests without affecting coverage? E.g. since you've factored out some of the z-curve code, can this test be simpler because you can assume your z-curve tests are working?

src/Domain/ElementDistribution.cpp
```diff
@@ -3,50 +3,167 @@
...
+
+    for (const auto& element_id : element_ids) {
+      if (element_weight == ElementWeight::Uniform) {
+        element_costs.insert({element_id, 1.0});
+      } else if (element_weight == ElementWeight::NumGridPoints) {
+        element_costs.insert({element_id, grid_points_per_element});
+      } else {
+        ASSERT(quadrature.has_value(),
+               "Since element_weight is "
+               "ElementWeight::NumGridPointsAndGridSpacing, quadrature must "
```
knelli2: Can you also
```cpp
ASSERT(element_weight == ElementWeight::NumGridPointsAndGridSpacing, "Unknown element_weight");
```
since this is the last branch.

tests/Unit/Domain/Test_ElementDistribution.cpp
macedo22: Yes, I tried trimming down the time with a fixup on another branch and it's still timing out on some CI checks, so I'm still iterating on it but will push a fixup when I get it to run without any timeouts

tests/Unit/Domain/Test_ElementDistribution.cpp
```diff
@@ -3,306 +3,780 @@
...
   }
-  return number_of_elements;
+
+  // The minimum grid spacing for the first and third domain are equal, but the
+  // number of grid points in an element in the first is 64 while the number of
+  // grid points in an element in the third is 24. Since elemental cost for
+  // either domain::Elementweight::NumGridPoints or
+  // domain::Elementweight::NumGridPointsAndGridSpacing should only scale by
+  // the # of grid points, the elemental cost of the second domain should be a
+  // factor of 24/64 = 3/8 the cost.
```
knelli2: You say second domain here but have `cost3` below. Also add `... 3/8 the cost of the first domain`.

tests/Unit/Domain/Test_ElementDistribution.cpp
```diff
@@ -3,306 +3,780 @@
...
+                  "  IsPeriodicIn: [false, false, false]\n"
+                  "  InitialGridPoints: [4, 3, 2]\n"
+                  "  InitialLevels: [2, 1, 0]\n"
+                  "  RefinedLevels: []\n"
+                  "  RefinedGridPoints: []\n"
+                  "  BlocksToExclude: []\n"});
+  const auto domain3 = domain_creator3->create_domain();
+  const auto& blocks3 = domain2.blocks();
+
+  const auto costs1 = domain::get_element_costs(
```
knelli2, line 76: To reduce the amount of code duplication and avoid the `1,2,3`, once you create the three domain creators, can you do something like
```cpp
for (const auto& domain_creator : {dc1, dc2, dc3}) {
  const auto cost = ...;
  auto cost_it = cost.begin();
  while (...) {
    CHECK(...);
    cost_it++;
  }

  if (weight == grid_points) {
    ...
  } else {
    ...
  }
}
```

knelli2 (changes requested):

src/Domain/ZCurve.cpp
```diff
@@ -0,0 +1,91 @@
+// Distributed under the MIT License.
```
nilsvu: Move to `Domain/Structure/`

src/Domain/ZCurve.cpp
```diff
@@ -0,0 +1,91 @@
...
+#include "Domain/Structure/ElementId.hpp"
+#include "Utilities/Algorithm.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace domain {
+
+template <size_t Dim>
+size_t z_curve_index_from_element_id(const ElementId<Dim>& element_id) {
```
nilsvu: why the rename? I think `z_curve_index` is enough, the `from_element_id` is redundant with the argument.

Edit: Oh I think you did the rename because you added the inverse. I think just leave this as `z_curve_index` nevertheless. The inverse can have the `from_z_curve_index` because the argument is just a size_t, so needs to be unambiguated.

src/Domain/ZCurve.hpp
```diff
@@ -23,7 +25,52 @@ namespace domain {
...
+/// index
+///
+/// \details For details on `ElementId`'s Z-curve indices, see
+/// `domain::segment_indices_from_z_curve_index`
+///
+/// \param block_id the `Block` number
+/// \param initial_ref_levs the refinements levels of the `Block`
+/// \param grid_index the grid index of the `Block`
+template <size_t Dim>
+std::vector<ElementId<Dim>> initial_element_ids_in_z_curve_order(
```
nilsvu: Would this naive implementation also work:

```cpp
auto element_ids = initial_element_ids(ref_levs, grid_index);
alg::sort(element_ids, [](const ElementId<Dim>& lhs, const ElementId<Dim>& rhs) {
  return z_curve_index(lhs) < z_curve_index(rhs);
});
```

Maybe add the block_id to the comparison if you need to. I would think that `initial_element_ids` is already mostly in the right order, so sorting isn't too expensive, but I don't know for sure.

If this approach means you don't need the inverse then perhaps the simplified code is worth the extra cost of sorting? If this doesn't work, could you add comments explaining the design decisions to the documentation of this function and the inverse?

src/Domain/ElementDistribution.cpp
```diff
@@ -3,50 +3,167 @@
...
+#include "Utilities/Literals.hpp"
+#include "Utilities/Numeric.hpp"
 
 namespace domain {
+template <size_t Dim>
+double get_num_points_and_grid_spacing_cost(
+    const ElementId<Dim>& element_id, const Block<Dim>& block,
+    const std::vector<std::array<size_t, Dim>>& initial_refinement_levels,
+    const std::vector<std::array<size_t, Dim>>& initial_extents,
+    const Spectral::Quadrature quadrature, const size_t num_grid_points) {
```
nilsvu: you can delete the `num_grid_points` argument and just use `mesh.number_of_grid_points()`. I think the cost of multiplying three doubles is irrelevant here.

src/Domain/ElementDistribution.cpp
```diff
@@ -3,50 +3,167 @@
...
+                : block.stationary_map().get_to_grid_frame()};
+        const tnsr::I<DataVector, Dim, Frame::ElementLogical> logical_coords =
+            logical_coordinates(mesh);
+        const tnsr::I<DataVector, Dim, Frame::Grid> grid_coords =
+            element_map(logical_coords);
+        const double min_grid_spacing =
+            minimum_grid_spacing(mesh.extents(), grid_coords);
+
+        element_costs.insert(
+            {element_id, grid_points_per_element / sqrt(min_grid_spacing)});
```
nilsvu: Isn't all this just the implementation of `get_num_points_and_grid_spacing_cost` above? You probably intended to use that here?

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -45,7 +50,12 @@ CREATE_HAS_STATIC_MEMBER_VARIABLE_V(use_z_order_distribution)
  * in the `Metavariables`, in which case elements are assigned to processors via
  * round-robin assignment. In both cases, an unordered set of `size_t`s can be
  * passed to the `allocate_array` function which represents physical processors
- * to avoid placing elements on.
+ * to avoid placing elements on. If the space-filling curve is used, then if
+ * `static constexpr bool local_time_stepping = true;` is specified
```
nilsvu, line 58: isn't it the reverse? I think this should be `false`.

Also, why prefer `Uniform` over `NumGridPoints` for global time stepping?

src/ParallelAlgorithms/LinearSolver/Multigrid/ElementsAllocator.hpp
```diff
@@ -162,8 +166,14 @@ struct ElementsAllocator
       // processors
       const size_t num_of_procs_to_use =
           static_cast<size_t>(sys::number_of_procs()) - procs_to_ignore.size();
+      const std::unordered_map<ElementId<Dim>, double> element_costs =
+          domain::get_element_costs(
+              blocks, initial_refinement_levels, initial_extents,
+              domain::ElementWeight::Uniform, std::nullopt);
```
nilsvu: Make this `NumGridPoints` as well please

tests/Unit/Domain/Test_ElementDistribution.cpp
```diff
@@ -3,306 +3,780 @@
...
+};
+
+// Stringify a bool
+std::string stringize(const bool t) { return t ? "true" : "false"; }
+
+// Create an option string for a `domain::creators::BinaryCompactObject`
+std::string create_option_string(const bool excise_A, const bool excise_B,
+                                 const bool add_time_dependence,
+                                 const bool use_logarithmic_map_AB,
+                                 const bool use_equiangular_map,
```
nilsvu: Please remove everything from this test that isn't relevant for the element distribution algorithm. For instance, you don't need to create the domain creator from options, just use the constructor. Boundary conditions aren't relevant either, so just delete everything related to boundary conditions. Tests that do too many things can be quite hard to maintain. Please test only the logic in `ElementDistribution.*pp`, no more and no less.

nilsvu (changes requested):
Nice algorithm!

src/Domain/ElementDistribution.cpp
macedo22: whoops, yes that is what I intended, thank you!

src/Domain/ZCurve.hpp
macedo22: That should work! I only defined the inverse to avoid sorting, but if sorting is the preference, I can do that and yes, it would reduce the amount and simplify the code

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
macedo22: yes, good catch

src/Domain/ZCurve.hpp
nilsvu: 🤷‍♂️ I don't know, sorting is a lot easier and probably not prohibitively expensive, but I haven't thought much about this. Definitely keep the inverse code around on a branch if you decide to go the sorting route. Also maybe other reviewers have a more educated opinion here?

macedo22: The implementation is preserved on other branches, regardless of what we choose. I can wait to see what others say!

macedo22:
I'm guessing that there is some numerical error with summation on the failing MacOS test. I will iterate on increasing the epsilon for the allowed error for the `CHECK` that's failing to try and get it to pass.

src/Domain/ZCurve.cpp
macedo22: It's worth noting that I only factored this out because (a) at the time, I thought we would have both `ElementDistribution.cpp` and `WeightedElementDistribution.cpp` that both needed it and (b) I wanted to be able to call the z-curve function from a test (on develop, it's in an anonymous namespace in `ElementDistribution.cpp` and thus has no external linkage). Given this, do you still think it's worth factoring it out, or should I just scrap the first commit altogether?

nilsvu: IMO factoring out the z-curve is nice

src/Domain/ZCurve.hpp
wthrowe: I haven't thought about this particular problem, but the general advice I've seen with respect to sorting is that sorting is fast.  (Although sorting something that is nearly in order is not faster.)  In the asymptotic regime, which is probably actually relevant here, it's `O(n log n)`, and if you have a vector you were probably planning to do something at least `O(n)` anyway.  Of course if the comparator is expensive that could be an issue, but it doesn't look too bad here (although it probably adds another factor of `O(log n)`, which is fairly irrelevant).

The only question then is if the overall prefactor gets a lot worse.  Here, the inverse function being used in the current algorithm looks about as complex as the function used in the comparator, so it's probably not an issue.  If it is, there's an easy factor of ~2 to be gained by precomputing the indices of all the elements.

wthrowe (commented):
I have no additional comments, but I do want to add weight to @nilsvu's comment about simplifying the test.  There's a lot of unrelated stuff going on there.

nilsdeppe:
Can you please rebase and resolve conflicts to get CI to run?

macedo22:
Looks like the timeouts are back. I think the test cases being run are important to run and not over-the-top thorough (this is also after removing some test cases that *were* maybe not necessary), so I'm going to split it into a couple `SPECTRE_TEST_CASE`s instead of removing test cases. Once CI runs without timeouts, I'll re-request reviews, but let me know if there are any objections to this plan!

knelli2:
@macedo22 If you think the current test cases should be kept, but the tests are still timing out, you can just increase the test timeout to like 10 or 15 seconds. Unless somebody else wants more tests removed, I think it's ok to increase the timeout.

tests/Unit/Domain/Test_ElementDistribution.cpp
```diff
@@ -3,306 +3,597 @@
...
+  // 1D, single block
+  const auto lattice_1d = domain::creators::AlignedLattice<1>(
+      {{{{0.0, 1.0}}}}, {{4}}, {{6}}, {}, {}, {});
+
+  // 2D
+  const auto lattice_2d = domain::creators::AlignedLattice<2>(
+      {{{{0.0, 0.3}}, {{0.0, 0.8, 2.5, 4.9}}}}, {{2, 3}}, {{4, 5}}, {}, {}, {});
+
+  // 3D
+  const auto binary_compact_object_creator =
```
nilsdeppe: Just stop using this domain... It's insanely slow, I would guess that alone will get you the the speedup alone.

nilsdeppe (changes requested):
No, don't split into test cases. I'm also not sure this needs the timeouts increased.

The tests still look very complicated, e.g. the BBH domain is used. Why? That's overkill for what you want to test. A sphere or aligned lattice is more than enough. What do you actually _need_ to test? The bare minimum. What code did you _add_ and only test that. E.g. do you really need 9 tests of a uniform distribution? That seems like overkill. And then the same 9 tests for weighted distributions? Does the distribution effect which cores can or can't have elements? If not, then you don't need to test a tensor product of the parameters for them, you  just hold one fixed and vary the other. 

Edit: some context as to why we are so strongly against long tests unless they truly add a lot of value: timeout creep! Somebody sees "Oh, I can make the timeouts go away by increasing the allowed time" and this repeats, eventually we have a bunch of tests that take 2 minutes each. This seems unlikely, but SpEC never enforced any timeouts and there are "short" tests that run for 10 minutes(!!!!!!!!!). The net result is that nobody runs the tests anymore. It's really important that tests are minimally exhaustive. E.g. if I write a `square` function I only need to test that it works for a few different types, and maybe use `0.0` as a "special" value. I don't need to test for every floating point or integer value. The value added by testing more integers basically drops to zero after you test 0,1,2.

tests/Unit/Domain/Test_ElementDistribution.cpp
macedo22: Both the review comment and this comment are fair points. I used the BBH domain simply because I felt the 1D and 2D cases checked the box of testing simple domains and I thought it was good to include a more complex and realistic domain test case in addition to the simple ones, and I also had it in my head that BBH runs were one of the motivations for this change so I had it on the mind and wanted to make sure it worked. But you're right about the importance of testing the smallest things for unit tests and I also think you're right that the BBH is what is making it so slow, so I'll swap it out. And also a fair point about the distributions. I was thinking of covering Dim and proc cases to make sure the combinations worked, but in some cases they're not actually checking anything new from another one, despite the different case. So, tl;dr you're right, I'll simplify, and thanks for the input!

tests/Unit/Domain/Test_ElementDistribution.cpp
macedo22: I ended up not doing this because I found that there wasn't really a nice and clean way to do it because the `if` / `else` needs access to the 3 costs computed, which would be local to the `for` loop, so if I put the cost computation in the loop, I would need to compute them again for the `if` / `else` after. But please let me know if there is something I'm not seeing/understanding!

macedo22:
Thank you everyone for the reviews! Seven fixups have been added and I'll squash changes into the correct original commits after re-review:
- First 3 ([18569e0](https://github.com/sxs-collaboration/spectre/pull/4713/commits/18569e0a0f14ba412876331f72d737c4a710c668), [1563736](https://github.com/sxs-collaboration/spectre/pull/4713/commits/1563736c5d5e2f0d6693cde9e9530186f0c8d904), [e45e523](https://github.com/sxs-collaboration/spectre/pull/4713/commits/e45e52399279fc34d40d1a403f9b10741816a1ff)) are @nilsvu 's change requests
- 4th ([6ea906d](https://github.com/sxs-collaboration/spectre/pull/4713/commits/6ea906deef5b3c8779ea51e267bacd6465b75e75)) is @knelli2 's change requests
- 5th ([3f19d73](https://github.com/sxs-collaboration/spectre/pull/4713/commits/3f19d73886bc5db8b22d772a8c2ce3a1ec01ce68)) fixes a compile error that resulted from rebasing (`RandomAmr` seems to have been added and it doesn't have a `local_time_stepping` member, which is used by `DgElementArray`) 
- 6th ([946e387](https://github.com/sxs-collaboration/spectre/pull/4713/commits/946e387bff2d61c35dc8dda6eb444b4c1ded1446)) simplifies the `ElementDistribution` test (@nilsdeppe 's change request)
- 7th ([7abbc2e](https://github.com/sxs-collaboration/spectre/pull/4713/commits/7abbc2ed51a6b84a27286fe46679a8c1620a386a)) is an additional minor adjustment to the proc distribution logic for the last proc that I opted to make (see commit message for details)

src/Domain/ElementDistribution.cpp
```diff
@@ -154,7 +154,8 @@ BlockZCurveProcDistribution<Dim>::BlockZCurveProcDistribution(
   size_t element_num_of_block = 0;
   double cost_remaining = total_cost;
   size_t number_of_ignored_procs_so_far = 0;
-  for (size_t i = 0; i < number_of_procs_with_elements; ++i) {
+  // distribute Elements to all but the final proc
+  for (size_t i = 0; i < number_of_procs_with_elements - 1; ++i) {
     size_t global_proc_number = i + number_of_ignored_procs_so_far;
```
knelli2, line 181: Optional, but if you define `global_proc_number` outside this loop, then you don't need to recalculate it below. It'll already be at the last proc

knelli2 (commented):
One optional thing, but otherwise looks good!

src/Domain/ElementDistribution.cpp
macedo22: I can throw this in when I squash!

src/Domain/ElementDistribution.hpp
```diff
@@ -5,56 +5,102 @@
...
+/// the smaller the minimum spacing, the smaller time step we must take, which
+/// means we expect computational work to scale inversely with the minimum grid
+/// spacing.
+///
+/// The reason that we use the square root of the spacing as opposed to just the
+/// spacing in the denominator of the cost is that it was found experimentally
+/// that using the square root yielded faster BBH simulation runtimes when using
+/// local time stepping.
+template <size_t Dim>
+double get_num_points_and_grid_spacing_cost(
```
nilsdeppe: You don't call this outside of the cpp file. Why is it public instead of in an anonymous namespace in the `cpp`? If it's to document things, then I think you should put the dox on the enum value instead.

src/Domain/Structure/ZCurve.cpp
```diff
@@ -0,0 +1,107 @@
...
+    const size_t block_id, const std::array<size_t, Dim> initial_ref_levs,
+    const size_t grid_index) {
+  std::vector<ElementId<Dim>> element_ids =
+      initial_element_ids(block_id, initial_ref_levs, grid_index);
+  alg::sort(element_ids,
+            [](const ElementId<Dim>& lhs, const ElementId<Dim>& rhs) {
+              return z_curve_index(lhs) < z_curve_index(rhs);
+            });
+
+  return element_ids;
```
nilsvu: Just delete this function and sort in-place where you need to. Shallow wrappers like this can actually make the code harder to understand because you think they do something fancy but they really don't :)

nilsvu (changes requested):
LGTM! You can squash this in directly.

nilsdeppe (commented):
LGTM, one suggestion. You can squash everything right away :)

src/Domain/ElementDistribution.cpp
macedo22: Turns out I actually can't do this and have to recalculate it. This is because when we request a single proc, after the loop, the global proc number represents the last proc, but when we request more than one proc, it will represent the 2nd to last proc. For this reason, I just kept the recalculation

nilsdeppe (approved):

nilsvu (approved):

nilsvu:
@knelli2 and @wthrowe do you want to take another look at this one?

knelli2 (approved):
I'm happy with everything

wthrowe (approved):

PR #4719, opened 2023-02-10
Add abutting direction per element to `ExcisionSphere`

nikwit:
## Proposed changes

Adds a method to check whether an element abuts an excision sphere and returns an optional direction. This can be called by each element to e.g. identify the surfaces of the excision sphere in order to do face integrals.

~~@nilsvu can this be done with `ElementId` rather than `Element` by checking whether the segment is at the boundary of the given direction? I am worried that this assumes that the blocks have a certain orientation w.r.t. the excision sphere.~~

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Structure/ExcisionSphere.cpp
+++ b/src/Domain/Structure/ExcisionSphere.cpp
@@ -4,12 +4,15 @@
 #include "Domain/Structure/ExcisionSphere.hpp"
 
 #include <cstddef>
+#include <optional>
 #include <ostream>
 #include <pup.h>  // IWYU pragma: keep
 #include <pup_stl.h>
 #include <unordered_map>
 
 #include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Utilities/ConstantExpressions.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/StdHelpers.hpp"  // IWYU pragma: keep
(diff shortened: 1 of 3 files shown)


tests/Unit/Domain/Structure/Test_ExcisionSphere.cpp
```diff
@@ -72,12 +79,73 @@ void check_excision_sphere_3d() {
       {{0, Direction<3>::lower_xi()}, {2, Direction<3>::upper_zeta()}});
 }
 
+void test_abutting_direction_shell() {
+  const double inner_radius = 1.3;
+  for (const auto& ref_level : std::array<size_t, 3>{{0, 1, 2}}) {
```
nikwit, line 89: include `<array>`

kidder:
No it cannot be done with just the `ElementId` because you don't know which `SegmentId` to check

src/Domain/Structure/ExcisionSphere.cpp
```diff
@@ -25,6 +27,21 @@ ExcisionSphere<VolumeDim>::ExcisionSphere(
...
+ExcisionSphere<VolumeDim>::abutting_direction(
+    const Element<VolumeDim>& element) const {
+  const auto& element_id = element.id();
+  const size_t block_id = element_id.block_id();
+  if (abutting_directions_.count(block_id)) {
+    const auto& external_boundaries = element.external_boundaries();
+    if (external_boundaries.count(abutting_directions_.at(block_id))) {
+      return {abutting_directions_.at(block_id)};
+    }
+  }
```
nilsvu: Wouldn't this do it, using only the ElementID:

```cpp
const auto& direction = abutting_directions.at(block_id);
const auto& segment_id = element_id.segment_id(direction.dimension());
const size_t abutting_index = direction.side() == Side::Lower ? 0 : pow<segment_id.refinement_level()>(2);
if (segment_id.index() == abutting_index) {
  return direction;
}
```

Please double-check (and test) this.

kidder: ah, okay I missed the direction is known...
the false branch of abutting_index should be `two_to_the(segment_id.refinement_level()) - 1`

nilsvu (dismissed):

nilsvu (approved):

PR #3421, opened 2021-08-09
Specify `temporal_id` for interpolation in interpolation target tags

moxcodes:
## Proposed changes

Apologies to @wthrowe , whose commit I found I needed to revert to make this work.

To construct the coupled CCE-GH system without fairly radical rethinking of the communication pattern, we need to perform interpolation during self-start, and that can only be accomplished unambigously if we use a key for the interpolator maps that distinguishes between self-start stage -- so, a `double` that represents the current time won't work.

The control-flow of interpolation during self start will not use dense triggers.

To support the events and dense triggers needed for most use-cases, I've changed the interpolation framework to use a tag specified by the `temporal_id` type alias that must now be specified in the interpolation target tags. For almost all cases, that should be `::Tags::Time`, but for interpolation during self start can be `::Tags::TimeStepId`.

### Upgrade instructions

Interpolation target tags must now specify a `temporal_id`. In almost all cases it should be `using temporal_id = ::Tags::Time;`.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4720, opened 2023-02-10
Find apparent horizon in distorted frame

markscheel:
## Proposed changes

Two main changes:
 * compute_dest_vars_from_src_vars now recognizes distorted frame, so interpolation can be done in that frame.
 * Test_ApparentHorizonFinder now has a test where it finds the horizon in the distorted frame.

Other necessary changes are fixing a bug in UniformTranslation and adding instantiations of the distorted frame in many places.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.cpp
+++ b/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.cpp
@@ -14,11 +14,4 @@ double get_temporal_id_value(const LinkedMessageId<double>& id) {
 double get_temporal_id_value(const TimeStepId& time_id) {
   return time_id.substep_time().value();
 }
-double evaluate_temporal_id_for_expiration(const double time) { return time; }
-double evaluate_temporal_id_for_expiration(const LinkedMessageId<double>& id) {
-  return id.id;
-}
-double evaluate_temporal_id_for_expiration(const TimeStepId& time_id) {
-  return time_id.step_time().value();
-}
 }  // namespace intrp::InterpolationTarget_detail
--- a/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
+++ b/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
@@ -71,9 +71,6 @@ namespace InterpolationTarget_detail {
 double get_temporal_id_value(double time);
 double get_temporal_id_value(const LinkedMessageId<double>& id);
 double get_temporal_id_value(const TimeStepId& time_id);
-double evaluate_temporal_id_for_expiration(double time);
-double evaluate_temporal_id_for_expiration(const LinkedMessageId<double>& id);
-double evaluate_temporal_id_for_expiration(const TimeStepId& time_id);
 
 // apply_callback accomplishes the overload for the
 // two signatures of callback functions.
@@ -491,8 +488,7 @@ auto block_logical_coords(
       const auto& functions_of_time = get<domain::Tags::FunctionsOfTime>(cache);
       return ::block_logical_coordinates(
           domain, coords,
-          InterpolationTarget_detail::evaluate_temporal_id_for_expiration(
-              temporal_id),
+          InterpolationTarget_detail::get_temporal_id_value(temporal_id),
           functions_of_time);
     } else {
       // We error here because the maps are time-dependent, yet
@@ -657,8 +653,7 @@ void compute_dest_vars_from_source_vars(
         element_id, block.moving_mesh_logical_to_grid_map().get_clone()};
     const auto logical_coords = logical_coordinates(mesh);
     const auto time =
-        InterpolationTarget_detail::evaluate_temporal_id_for_expiration(
-            temporal_id);
+        InterpolationTarget_detail::get_temporal_id_value(temporal_id);
     const auto jac_logical_to_grid =
         map_logical_to_grid.jacobian(logical_coords);
     const auto invjac_logical_to_grid =
@@ -688,8 +683,7 @@ void compute_dest_vars_from_source_vars(
         block.moving_mesh_grid_to_distorted_map().get_clone()};
     const auto logical_coords = logical_coordinates(mesh);
     const auto time =
-        InterpolationTarget_detail::evaluate_temporal_id_for_expiration(
-            temporal_id);
+        InterpolationTarget_detail::get_temporal_id_value(temporal_id);
     const auto [inertial_coords, invjac_distorted_to_inertial,
                 jac_distorted_to_inertial,
                 distorted_to_inertial_mesh_velocity] =
--- a/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
+++ b/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
@@ -14,7 +14,9 @@
 #include "DataStructures/Tensor/Metafunctions.hpp"
 #include "DataStructures/VariablesTag.hpp"
 #include "Domain/BlockLogicalCoordinates.hpp"
+#include "Domain/CoordinateMaps/Composition.hpp"
 #include "Domain/Creators/Tags/Domain.hpp"
+#include "Domain/ElementToBlockLogicalMap.hpp"
 #include "Domain/TagsTimeDependent.hpp"
 #include "Parallel/GlobalCache.hpp"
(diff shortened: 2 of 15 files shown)


src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -670,6 +672,49 @@ void compute_dest_vars_from_source_vars(
...
+                jac_distorted_to_inertial,
+                distorted_to_inertial_mesh_velocity] =
+        block.moving_mesh_distorted_to_inertial_map()
+            .coords_frame_velocity_jacobians(
+                element_logical_to_distorted_map(
+                    logical_coordinates(mesh),
+                    InterpolationTarget_detail::
+                        evaluate_temporal_id_for_expiration(temporal_id),
+                    functions_of_time),
+                InterpolationTarget_detail::evaluate_temporal_id_for_expiration(
```
knelli2: Why is this not `InterpolationTarget_detail::get_temporal_id_value`? I know it's used above, but looking at the two functions, they are identical for all types except `TimeStepId`.  The difference is that `get_temporal_id` does `TimeStepId.substep_time().value()` while `evaluate_temporal_id_for_expiration` does `TimeStepId.step_time().value()`. They are different, but does that matter here?

knelli2 (changes requested):
You can squash directly

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -670,6 +672,49 @@ void compute_dest_vars_from_source_vars(
...
+        block.moving_mesh_grid_to_distorted_map().get_clone()};
+    const auto [inertial_coords, invjac_distorted_to_inertial,
+                jac_distorted_to_inertial,
+                distorted_to_inertial_mesh_velocity] =
+        block.moving_mesh_distorted_to_inertial_map()
+            .coords_frame_velocity_jacobians(
+                element_logical_to_distorted_map(
+                    logical_coordinates(mesh),
+                    InterpolationTarget_detail::
+                        evaluate_temporal_id_for_expiration(temporal_id),
```
nilsvu: you use these two things repeatedly below. Only create them once.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -670,6 +672,49 @@ void compute_dest_vars_from_source_vars(
...
+                                                vars_to_interpolate_to_target,
+                                            Frame::Distorted>) {
+    // Need to do frame transformations to Distorted frame.
+    const auto& functions_of_time = get<domain::Tags::FunctionsOfTime>(cache);
+    const auto& block = domain.blocks().at(element_id.block_id());
+    ASSERT(block.has_distorted_frame(),
+           "Cannot interpolate to distorted frame in a block that does not "
+           "have a distorted frame");
+    ElementMap<3, ::Frame::Grid> map_element_logical_to_grid{
+        element_id, block.moving_mesh_logical_to_grid_map().get_clone()};
```
nilsvu: do you need this at all?

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -670,6 +672,49 @@ void compute_dest_vars_from_source_vars(
...
+    domain::CoordinateMaps::Composition element_logical_to_distorted_map{
```
nilsvu: const

nilsvu (changes requested):

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
markscheel: That's a good question.  `evaluate_temporal_id_for_expiration` is used in the code for the other frames too, and I thought that was what @wthrowe wanted. @wthrowe should I be using `get_temporal_id_value` here (and for the other frames) instead?

wthrowe: Probably `_value`, but first, do you get any failures if you put `ASSERT(time_id.substep() == 0, "Interpolating on substep " << time_id);` in both `get_temporal_id_*` functions?

markscheel: @wthrowe all the tests pass if I put those `ASSERT`s into both functions.   Should `evaluate_temporal_id_for_expiration` be removed, and only `get_temporal_id_value` used everywhere?  (and should I keep the ASSERT?)

wthrowe: So at the moment, the functions always return the same value.  I'm not sure I want to guarantee that that will always be true, though.  If I remember right, this is only used during self start, which is only used for AdamsBashforth which doesn't have substeps.  Predictor-corrector methods will presumably go through a similar infrastructure with substeps, though.

I don't really remember what's going on here.  Why are there two of these?  The `_value` function is presumably to represent the time (or iteration or whatever) to use in maps and such.  What is the `_expiration` for?  My first thought was we needed something monotonic, but it isn't monotonic.  (Can these things get some documentation?)

knelli2: Both of these were added in #3421. It seems that `evaluate_temporal_id_for_expiration` was added because previously, `VerifyTemporalIdsAndSendPoints.hpp` used `pending_id.step_time().value()`. Whereas, `get_temporal_id_value` was added because `ObserveTimeSeriesOnSurface.hpp` used `temporal_id.substep_time().value()`. (take a look at the first commit of that PR)

So we needed one that got the `step_time` and the other that got the `substep_time`. However, I'm not sure if we need to use one or the other in each circumstance. I agree with @wthrowe that `_value` seems to just get the current time, but unclear what `_expiration` is actually needed for.

wthrowe: Looking more closely, I don't think any of the uses of `step_time()` here are correct.  (Uses of `step_time()` outside time-related code are usually wrong.)  So just get rid of that function and probably rename the other one to get rid of the now-unnecessary disambiguator.

markscheel: Ok, I will keep the one using `substep_time()` and remove the other one.  Should I leave the ASSERT in there?

knelli2: I say yes. It's just an `ASSERT` so won't affect productions stuff

wthrowe: No.  I intend to add valid cases where it will fail.

nilsvu (dismissed):

markscheel:
squashed.

knelli2 (dismissed):

nilsvu (dismissed):

nilsdeppe:
@wthrowe did you want to look at the part you were involved in discussing for this PR?

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include "DataStructures/LinkedMessageId.hpp"
 #include "Time/TimeStepId.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
```
wthrowe: Remove.

wthrowe (changes requested):

markscheel:
Removed include, rebased and fixed conflict.

wthrowe:
Er, try again?  It's still there.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.cpp
knelli2: Yup still there

knelli2 (changes requested):

markscheel:
Hopefully the include is gone now... sorry..

knelli2 (approved):
Yup! Gone now!

wthrowe (approved):

nilsvu (approved):

PR #4721, opened 2023-02-10
Fix bug in GHMHD time derivative

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp
+++ b/src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp
@@ -317,8 +317,14 @@ struct ComputeTimeDerivImpl<
         for (size_t i = 0; i < dt_var.size(); ++i) {
           if constexpr (not tmpl::list_contains_v<
                             tmpl::list<GrmhdSourceTags...>, GrmhdDtTags>) {
-            // Zero GRMHD tags that don't have sources.
-            dt_var[i] = 0.0;
+            // On the first iteration of the loop over `dim`, zero the GRMHD
+            // dt(u) for variables that do not have a source term . This is
+            // necessary to avoid `+=` to a `NaN` (debug mode) or random garbage
+            // (release mode). `add_cartesian_flux_divergence` does a `+=`
+            // internally.
+            if (dim == 0) {
+              dt_var[i] = 0.0;
+            }
           }
 
           evolution::dg::subcell::add_cartesian_flux_divergence(


kidder (dismissed):

src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp
```diff
@@ -318,7 +318,15 @@ struct ComputeTimeDerivImpl<
...
             // Zero GRMHD tags that don't have sources.
-            dt_var[i] = 0.0;
+            //
+            // Since dt(u) is independent of `dim`, and `dim` is really part
+            // of the `d_dim F^dim` computation, we only need to set a zero
+            // source term when `dim` is zero. This is necessary to avoid `+=`
+            // to a `NaN` (debug mode) or random garbage (release
+            // mode). `add_cartesian_flux_divergence` does a `+=` internally.
+            if (dim == 0) {
+              dt_var[i] = 0.0;
```
nilsvu, line 326: what was the impact of this bug?

src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp
```diff
@@ -318,7 +318,15 @@ struct ComputeTimeDerivImpl<
           if constexpr (not tmpl::list_contains_v<
                             tmpl::list<GrmhdSourceTags...>, GrmhdDtTags>) {
             // Zero GRMHD tags that don't have sources.
-            dt_var[i] = 0.0;
+            //
+            // Since dt(u) is independent of `dim`, and `dim` is really part
+            // of the `d_dim F^dim` computation, we only need to set a zero
```
nilsvu: where does this `d_dim F^dim` computation happen? I didn't find `F^dim` in this file, and don't see `dim` passed to a function in this loop, so I'm not sure where to look. I'm just trying to understand what happened here.

src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp
nilsdeppe: For `TildeD` only the `d_z F^z` term of `d_i F^i` was used. No papers have used this code, so there's no impact on science we've published.

src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp
kidder: dim is passed to `evolution::dg::subcell::add_cartesian_flux_divergence` on line 332 which does that computation, *adding* to `dt_var`. 

However, any variable that doesn't have a source term (the if constexpr on line 318) has been passed to this function in an uninitialized state.  Therefore it should be set to zero before being passed into the first call of `add_cartesian_flux_divergence`.  The bug had the effect that only the highest dim contributed to the flux divergence instead of all the dimensions...

nilsdeppe: `evolution::dg::subcell::add_cartesian_flux_divergence(`

`dim` is used on lines 306, 308, and 310, which are then captured by the lambda.

isaaclegred:
Are the test failures because the target values from before are wrong?

src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp
nilsvu: Ah I think what I was missing is "Zero GRMHD tags that don't have sources **in the first iteration of the loop over `dim`**" or something like that. Maybe add that to the comment on line 320?

knelli2:
@isaaclegred No the GCC test failures were form a bug in CI that has been fixed now.

src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp
nilsdeppe: Isn't that exactly what `we only need to set a zero source term when dim is zero.` means?

nilsvu: I needed while to understand that this refers to the first iteration of the loop, with subsequent iterations adding to the value 🤷‍♂️ I'm not blocking this, just tried to understand what was happening here

nilsdeppe: I've rebased and reworded the sentences. Hopefully it's clearer now :)

nilsvu (approved):
Thanks that helps :) I restarted the CI builds that timed out

PR #4722, opened 2023-02-10
Add `euclidan_surface_jacobian`

nikwit:
Adds a function that computes the euclidean surface element from the inverse jacobian.

~~depends on #4719~~

### DIFF
--- a/src/Domain/SurfaceJacobian.hpp
+++ b/src/Domain/SurfaceJacobian.hpp
@@ -1,0 +1,64 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Utilities/Gsl.hpp"
+
+/*!
+ * \brief Compute the Euclidean surface element from the inverse jacobian.
+ *
+ * \details The Euclidean surface element of a surface \f$\Sigma\f$ with
+ * constant logical coordinate \f$\xi^i\f$ is given by:
+ *
+ * \f{equation}
+ * J^\Sigma = J \sqrt{\delta^{jk} (J^{-1})^i_j (J^{-1})^i_k}
+ * \f}
+ *
+ * where \f$J^i_j = \partial x^i / \xi^j\f$ is the volume Jacobian with
+ * determinant \f$J\f$ and inverse \f$(J^{-1})^i_j = \partial \xi^i / \partial
+ * x^j\f$. The determinant of the inverse Jacobian can be passed as an argument
+ * as an overload, otherwise it will be calculated.
+ *
+ * \note The curved space surface element as well as time dependent maps are not
+ * implemented yet but can be added on demand.
+ *
+ * \param inverse_jacobian_face The inverse Jacobian from the ElementLogical
+ * frame to the Target frame sliced to the surface.
+ * \param inverse_jacobian_determinant_face The determinant of the inverse
+ * Jacobian sliced onto the surface. If not availabe, there exists an overload
+ * that will calculate it from the inverse Jacobian.
+ * \param direction The direction of the surface in the element.
+ * */
+template <size_t VolumeDim, typename TargetFrame>
+void euclidean_surface_jacobian(
+    const gsl::not_null<Scalar<DataVector>*> result,
+    const InverseJacobian<DataVector, VolumeDim, Frame::ElementLogical,
+                          TargetFrame>& inverse_jacobian_face,
+    const Scalar<DataVector>& inverse_jacobian_determinant_face,
+    const Direction<VolumeDim>& direction);
+
+template <size_t VolumeDim, typename TargetFrame>
+void euclidean_surface_jacobian(
+    const gsl::not_null<Scalar<DataVector>*> result,
+    const InverseJacobian<DataVector, VolumeDim, Frame::ElementLogical,
+                          TargetFrame>& inverse_jacobian_face,
+    const Direction<VolumeDim>& direction);
+
+template <size_t VolumeDim, typename TargetFrame>
+Scalar<DataVector> euclidean_surface_jacobian(
+    const InverseJacobian<DataVector, VolumeDim, Frame::ElementLogical,
+                          TargetFrame>& inverse_jacobian_face,
+    const Scalar<DataVector>& inverse_jacobian_determinant_face,
+    const Direction<VolumeDim>& direction);
+
+template <size_t VolumeDim, typename TargetFrame>
+Scalar<DataVector> euclidean_surface_jacobian(
+    const InverseJacobian<DataVector, VolumeDim, Frame::ElementLogical,
+                          TargetFrame>& inverse_jacobian_face,
+    const Direction<VolumeDim>& direction);
--- a/src/Domain/CMakeLists.txt
+++ b/src/Domain/CMakeLists.txt
@@ -23,6 +23,7 @@ spectre_target_sources(
   MinimumGridSpacing.cpp
   ObjectLabel.cpp
   SizeOfElement.cpp
+  SurfaceJacobian.cpp
   Tags.cpp
   TagsTimeDependent.cpp
 )
@@ -50,6 +51,7 @@ spectre_target_headers(
   ObjectLabel.hpp
   OptionTags.hpp
   SizeOfElement.hpp
+  SurfaceJacobian.hpp
   Tags.hpp
   TagsCharacteristicSpeeds.hpp
   TagsTimeDependent.hpp
--- a/src/Domain/SurfaceJacobian.cpp
+++ b/src/Domain/SurfaceJacobian.cpp
@@ -1,0 +1,101 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/SurfaceJacobian.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/Determinant.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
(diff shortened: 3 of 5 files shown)


src/Domain/SurfaceElement.hpp
```diff
@@ -0,0 +1,61 @@
...
+#include "Utilities/Gsl.hpp"
+
+/*!
+ * \brief Compute the Euclidean surface element from the inverse jacobian.
+ *
+ * \details The Euclidean surface element of a surface \f$\Sigma\f$ with
+ * constant logical coordinate \f$\xi^i\f$ is given by: \f{equation} J^\Sigma =
+ * J \sqrt{\delta^{jk} (J^{-1})^i_j (J^{-1})^i_k} \f} where \f$J^i_j = \partial
+ * x^i / \xi^j\f$ is the volume Jacobian with determinant \f$J\f$ and inverse
+ * \f$(J^{-1})^i_j = \partial \xi^i / \partial x^j\f$. The determinant of the
```
nilsvu: Please format equations like in `Domain/Tags/SurfaceJacobian.hpp`

src/Domain/SurfaceElement.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * inverse Jacobian can be passed as an argument as an overload, otherwise it
+ * will be calculated.
+ *
+ * \note The curved space surface element as well as time dependent maps are not
+ * implemented yet but can be added on demand.
```
nilsvu: Don't you need both curved space and time dependence for the worldtube integrals?

Also, you actually have to evaluate this at every time step, right? In that case it might be necessary for performance to split the computation into ElementLogical -> Grid and Grid -> Inertial, so you don't have to evaluate the static ElementLogical -> Grid at every time step. Up to you if you want to do this now or add notes to the docs that it's a possible performance optimization.

src/Domain/SurfaceElement.hpp
```diff
@@ -0,0 +1,61 @@
+// Distributed under the MIT License.
```
nilsvu: I suggest you name these things `SurfaceJacobian` to avoid name collisions with "element". There's already a tag in `Domain/Tags/SurfaceJacobian.hpp` with that name.

src/Domain/SurfaceElement.hpp
```diff
@@ -0,0 +1,61 @@
...
+ * will be calculated.
+ *
+ * \note The curved space surface element as well as time dependent maps are not
+ * implemented yet but can be added on demand.
+ *
+ * \param inverse_jacobian_face The inverse Jacobian from the ElementLogical
+ * frame to the Target frame sliced to the surface.
+ * \param inverse_jacobian_determinant_face The determinant of the inverse
+ * Jacobian sliced onto the surface. If not availabe, there exists an overload
+ * that will calculate it from the inverse Jacobian. \param direction The
```
nilsvu: \param in new line

src/Domain/SurfaceElement.cpp
```diff
@@ -0,0 +1,103 @@
...
+
+#include "Domain/SurfaceElement.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/Determinant.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Evolution/DiscontinuousGalerkin/ProjectToBoundary.hpp"
```
nilsvu: delete

src/Domain/SurfaceElement.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "Utilities/Gsl.hpp"
+
+template <size_t VolumeDim, typename TargetFrame>
+void euclidean_surface_element(
+    const gsl::not_null<Scalar<DataVector>*> result,
+    const InverseJacobian<DataVector, VolumeDim, Frame::ElementLogical,
+                          TargetFrame>& inverse_jacobian_face,
+    const Scalar<DataVector>& inverse_jacobian_determinant_face,
+    const Direction<VolumeDim>& direction) {
+  ASSERT(inverse_jacobian_face.get(0, 0).size() ==
```
nilsvu: include ErrorHandling/Assert

nilsvu: prefer `get<0, 0>(tensor)` when you know the indices at compile time. Same for `get(scalar)` over `scalar.get()`.

src/Domain/SurfaceElement.cpp
```diff
@@ -0,0 +1,103 @@
...
+#include "Domain/SurfaceElement.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/Determinant.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Evolution/DiscontinuousGalerkin/ProjectToBoundary.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
```
nilsvu: delete

src/Domain/SurfaceElement.hpp
```diff
@@ -0,0 +1,61 @@
...
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
```
nilsvu: delete

tests/Unit/Domain/Test_SurfaceElement.cpp
```diff
@@ -0,0 +1,129 @@
...
+  CHECK_ITERABLE_APPROX(result_1.get(), result_2.get());
+  CHECK_ITERABLE_APPROX(face_normal_mag_corrected, result_1.get());
+}
+
+// A test function sin^2(theta) * cos^2(phi) which evaluates to 4/3 * pi * R^2
+// when integrated over a sphere with radius R.
+DataVector test_function(const tnsr::I<DataVector, 3>& inertial_coords) {
+  const DataVector theta =
+      atan2(hypot(inertial_coords.get(0), inertial_coords.get(1)),
+            inertial_coords.get(2));
```
nilsvu: reminder to prefer `get<0>(coords)` over `coords.get(0)` when you know the indices at compile time

tests/Unit/Domain/Test_SurfaceElement.cpp
```diff
@@ -0,0 +1,129 @@
...
+  const DataVector phi = atan2(inertial_coords.get(1), inertial_coords.get(0));
+  return sin(theta) * sin(theta) * cos(phi) * cos(phi);
+}
+
+// Integrates the area and a test function over the excision surface of a shell
+void test_sphere_integral() {
+  for (const auto inner_radius : std::array<double, 3>{{1e-4, 1e-2, 1.4}}) {
+    for (const auto& ref_level : std::array<size_t, 3>{{0, 1, 2}}) {
```
nilsvu: this could be `for (const auto& [inner_radius, ref_level] : cartesian_product(make_array(1.e-4, 1.e-2, 1.4), make_array(0, 1, 2))` to flatten the loop

nilsvu (changes requested):
nice tests!

src/Domain/SurfaceElement.hpp
nikwit: No, I don't need those for now. The worldtube currently  represents the internal solution inside the worldtube in the grid frame and uses a euclidean coordinates.

I might need to add them later for non-circular orbits. The overloads should probably be the same ones as for `unnormalized_normal_vector`.

src/Domain/SurfaceElement.hpp
nikwit: oops, had the wrong formatter config

src/Domain/SurfaceElement.hpp
nilsvu: Would you mind adding an overload that takes a spatial metric? We'll need that sooner or later. Do you currently get by with the Euclidean surface integral because you're working with a perturbative solution in the worldtube? Would it be wrong to use the spatial metric in the integral?

nilsvu:
Fixup LGTM, please squash and rebase once the dependency PR is merged.

src/Domain/SurfaceElement.hpp
nikwit: Yes, the local solution is in Cartesian coordinates. I can add the proper curved space integral in a follow-up PR. Do you happen to know about a nice test  for that  like an analytical solution for a scalar quantity integrated over a sphere centered on a Schwarzschild/Kerr black hole?

nilsvu: Yea the irreducible mass (horizon area) of a Kerr horizon is a good quantity to compute [Eq. (1.70) in Baumgarte/Shapiro]. It's also computed for a Strahlkorper here: https://github.com/sxs-collaboration/spectre/blob/develop/tests/Unit/ApparentHorizons/Test_StrahlkorperGr.cpp#L928

src/Domain/SurfaceElement.hpp
nilsvu: Also name the functions `euclidean_surface_jacobian` please

src/Domain/SurfaceJacobian.hpp
```diff
@@ -0,0 +1,64 @@
...
+ *
+ * \param inverse_jacobian_face The inverse Jacobian from the ElementLogical
+ * frame to the Target frame sliced to the surface.
+ * \param inverse_jacobian_determinant_face The determinant of the inverse
+ * Jacobian sliced onto the surface. If not availabe, there exists an overload
+ * that will calculate it from the inverse Jacobian.
+ * \param direction The direction of the surface in the element.
+ * */
+template <size_t VolumeDim, typename TargetFrame>
+void euclidean_surface_element(
```
nilsvu: Also name the functions `euclidean_surface_jacobian` please

nilsvu (changes requested):
squash directly

src/Domain/SurfaceJacobian.hpp
nikwit: I think this is more consistent with `euclidean_area_element` in `ApparentHorizons/StrahlKorperGr.hpp` which calculates the same thing for Strahlkorpers.

So I guess just `euclidean_area_element` or do you prefer `euclidean_surface_jacobian`

nilsvu: Either works, just make the tag, the functions, and the file names consistent. I suggested "surface jacobian" just because that's the name of the existing tag.

nilsvu (approved):

PR #4663, opened 2023-01-30
Add script to plot power monitors

nilsvu:
## Proposed changes

Creates plots like this:

plot

Invoke the script like this:

```sh
$ spectre plot-power-monitors ElasticityVolume0.h5  -d VolumeData \
    -b InnerCube -b Wedges -y 'Displacement_*' -o plot.pdf
```

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4725, opened 2023-02-13
Add truncation error estimate

guilara:
## Proposed changes

Add truncation error estimate functions and pybindings. Change output of power monitors to return the log10 value of them.
#4585
### Upgrade instructions

Corrected script for new log factor #4663

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
+++ b/src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
@@ -12,6 +12,7 @@
 #include "NumericalAlgorithms/LinearOperators/CoefficientTransforms.hpp"
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
 #include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 
@@ -19,10 +20,9 @@ namespace PowerMonitors {
 
 template <size_t Dim>
 void power_monitors(const gsl::not_null<std::array<DataVector, Dim>*> result,
-                const DataVector& input_data_vector, const Mesh<Dim>& mesh) {
+                    const DataVector& u, const Mesh<Dim>& mesh) {
   // Get modal coefficients
-  const ModalVector modal_coefficients =
-      to_modal_coefficients(input_data_vector, mesh);
+  const ModalVector modal_coefficients = to_modal_coefficients(u, mesh);
 
   double slice_sum = 0.0;
   size_t n_slice = 0;
@@ -47,10 +47,71 @@ void power_monitors(const gsl::not_null<std::array<DataVector, Dim>*> result,
 }
 
 template <size_t Dim>
-std::array<DataVector, Dim> power_monitors(
-    const DataVector& input_data_vector, const Mesh<Dim>& mesh) {
+std::array<DataVector, Dim> power_monitors(const DataVector& u,
+                                           const Mesh<Dim>& mesh) {
   std::array<DataVector, Dim> result{};
-  power_monitors(make_not_null(&result), input_data_vector, mesh);
+  power_monitors(make_not_null(&result), u, mesh);
+  return result;
+}
+
+// The power_monitor argument should be made of type ModalVector
+// when pybindings for ModalVector are enabled
+double relative_truncation_error(const DataVector& power_monitor,
+                                 const size_t num_modes_to_use) {
+  ASSERT(
+      num_modes_to_use <= power_monitor.size(),
+      "Number of modes needs less or equal than the number of power monitors");
+  ASSERT(2_st <= num_modes_to_use,
+         "Number of modes needs to be larger or equal than 2.");
+  // Compute weighted average and total sum in the current dimension
+  double weighted_average = 0.0;
+  double weight_sum = 0.0;
+  double weight_value = 0.0;
+  const size_t last_index = num_modes_to_use - 1;
+  for (size_t index = 0; index <= last_index; ++index) {
+    // Compute current weight
+    weight_value =
+        exp(-square(index - static_cast<double>(last_index) + 0.5));
+    // Add weighted power monitor
+    const double mode = power_monitor[index];
+    ASSERT(not(mode == 0.0),
+           "A power monitor is zero bitwise. If this is one of "
+           "the highest modes, reduce the number of power monitors used to "
+           "estimate the relative truncation error.");
+    weighted_average += weight_value * log10(mode);
+    // Add term to weighted sum
+    weight_sum += weight_value;
+  }
+  weighted_average /= weight_sum;
+
+  // Maximum between the first two power monitors
+  double leading_term = std::max(power_monitor[0], power_monitor[1]);
+  ASSERT(not(leading_term == 0.0),
+         "The leading power monitor term is zero bitwise.");
+  leading_term = log10(leading_term);
+
+  // Compute relative truncation error
+  double result = leading_term - weighted_average;
+  return result;
+}
+
+template <size_t Dim>
+std::array<double, Dim> truncation_error(
+    const std::array<DataVector, Dim>& power_monitors_of_u,
+    const DataVector& u) {
+  std::array<double, Dim> result{};
+  // Use infinity norm to estimate the order of magnitude of the variable
+  const double umax = max(abs(u));
+  double relative_truncation_error_in_d = 0.0;
+  for (size_t d = 0; d < Dim; ++d) {
+    const auto& modes = gsl::at(power_monitors_of_u, d);
+    // Compute relative truncation error
+    relative_truncation_error_in_d =
+        relative_truncation_error(modes, modes.size());
+    // Compute absolute truncation error estimate
+    gsl::at(result, d) =
+        umax * pow(10.0, -1.0 * relative_truncation_error_in_d);
+  }
   return result;
 }
 
@@ -60,11 +121,14 @@ std::array<DataVector, Dim> power_monitors(
 
 #define INSTANTIATE(_, data)                                                   \
   template std::array<DataVector, DIM(data)>                                   \
-  PowerMonitors::power_monitors(const DataVector& input_data_vector,           \
+  PowerMonitors::power_monitors(const DataVector& u,                           \
                                      const Mesh<DIM(data)>& mesh);             \
   template void PowerMonitors::power_monitors(                                 \
     const gsl::not_null<std::array<DataVector, DIM(data)>*> result,            \
-    const DataVector& input_data_vector, const Mesh<DIM(data)>& mesh);
+    const DataVector& u, const Mesh<DIM(data)>& mesh);                         \
+  template std::array<double, DIM(data)> PowerMonitors::truncation_error(      \
+    const std::array<DataVector, DIM(data)>& power_monitors_of_u,              \
+    const DataVector& u);
 
 GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3))
 
--- a/src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
+++ b/src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
@@ -38,11 +38,69 @@ namespace PowerMonitors {
  */
 template <size_t Dim>
 void power_monitors(gsl::not_null<std::array<DataVector, Dim>*> result,
-                    const DataVector& input_data_vector, const Mesh<Dim>& mesh);
+                    const DataVector& u, const Mesh<Dim>& mesh);
 
 template <size_t Dim>
-std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
+std::array<DataVector, Dim> power_monitors(const DataVector& u,
                                            const Mesh<Dim>& mesh);
 /// @}
 
+/// @{
+/*!
+ * \ingroup SpectralGroup
+ * \brief Compute the negative log10 of the relative truncation error.
+ *
+ * Truncation error according to Eqs. (57) and (58) of Ref.
+ * \cite Szilagyi2014fna, i.e.,
+ *
+ * \f{align*}{
+ *  \mathcal{T}\left[P_k\right] = \log_{10} \max \left(P_0, P_1\right)
+ *   - \dfrac{\sum_{j=0}^{j_{\text{max}, k}} \log_{10} \left(P_j\right) w_j}
+ *   {\sum_{j=0}^{j_{\text{max}, k}} w_j} , \f}
+ *
+ * with weights
+ *
+ * \f{align*}{
+ *  w_j = \exp\left[ - \left(j - j_{\text{max}, k}
+ *          + \dfrac{1}{2}\right)^2 \right] .
+ * \f}
+ *
+ * where \f$ j_{\text{max}, k}  = N_k - 1 \f$ and  \f$ N_k \f$ is the number of
+ * modes or gridpoints in dimension k. Here the second term is a weighted
+ * average with larger weights toward the highest modes. This number should
+ * correspond to the number of digits resolved by the spectral expansion.
+ *
+ * \details The number of modes (`num_modes_to_use`) argument needs to be less
+ * or equal than the total number of power monitors (`power_monitor.size()`).
+ * In contrast with Ref. \cite Szilagyi2014fna, here we index the modes starting
+ * from zero.
+ *
+ */
+double relative_truncation_error(const DataVector& power_monitor,
+                                 const size_t num_modes_to_use);
+/// @}
+
+/// @{
+/*!
+ * \ingroup SpectralGroup
+ * \brief Returns an estimate of the absolute truncation error in each
+ * dimension.
+ *
+ * The estimate of the numerical error is given by
+ *
+ * \f{align*}{
+ *  \mathcal{E}\left[P_k\right] = u_\mathrm{max} \times 10^{- \mathcal{T}[P_k]},
+ * \f}
+ *
+ * where \f$ u_\mathrm{max} = \mathrm{max} |u|\f$ in the corresponding element
+ * and \f$ \mathcal{T}[P_k] \f$ is the relative error estimate
+ * computed from the power monitors \f$ P_k \f$.
+ *
+ * \warning This estimate is intended for visualization purposes only.
+ */
+template <size_t Dim>
+std::array<double, Dim> truncation_error(
+    const std::array<DataVector, Dim>& power_monitors_of_u,
+    const DataVector& u);
+/// @}
 }  // namespace PowerMonitors
--- a/src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
+++ b/src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
@@ -20,7 +20,12 @@ void bind_power_monitors_impl(py::module& m) {  // NOLINT
   m.def("power_monitors",
         py::overload_cast<const DataVector&, const Mesh<Dim>&> (
           &power_monitors<Dim>),
-        py::arg("input_data_vector"), py::arg("mesh"));
+        py::arg("data_vector"), py::arg("mesh"));
+  m.def(
+      "truncation_error",
+      py::overload_cast<const std::array<DataVector, Dim>&, const DataVector&>(
+          &truncation_error<Dim>),
+      py::arg("power_monitors_of_u"), py::arg("data_vector"));
 }
 }  // namespace
 
@@ -28,6 +33,10 @@ void bind_power_monitors(py::module& m) {
   bind_power_monitors_impl<1>(m);
   bind_power_monitors_impl<2>(m);
   bind_power_monitors_impl<3>(m);
+  m.def("relative_truncation_error",
+        py::overload_cast<const DataVector&, const size_t>(
+            &relative_truncation_error),
+        py::arg("power_monitor"), py::arg("num_modes_to_use"));
 }
 
 }  // namespace PowerMonitors::py_bindings
--- a/tests/Unit/NumericalAlgorithms/LinearOperators/Python/Test_PowerMonitors.py
+++ b/tests/Unit/NumericalAlgorithms/LinearOperators/Python/Test_PowerMonitors.py
@@ -6,8 +6,18 @@ import unittest
 import numpy as np
 
 from spectre.DataStructures import DataVector
-from spectre.NumericalAlgorithms.LinearOperators import power_monitors
-from spectre.Spectral import Basis, Mesh2D, Quadrature
+from spectre.NumericalAlgorithms.LinearOperators import (
+    power_monitors,
+    relative_truncation_error,
+    truncation_error,
+)
+from spectre.Spectral import (
+    Basis,
+    Mesh1D,
+    Mesh2D,
+    Quadrature,
+    logical_coordinates,
+)
 
 
 class TestPowerMonitors(unittest.TestCase):
@@ -21,8 +31,7 @@ class TestPowerMonitors(unittest.TestCase):
         quadrature = Quadrature.GaussLobatto
         mesh = Mesh2D(extent, basis, quadrature)
 
-        np_vec = np.ones(mesh.number_of_grid_points())
-        test_vec = DataVector(np_vec)
+        test_vec = np.ones(mesh.number_of_grid_points())
 
         test_array = power_monitors(test_vec, mesh)
         np_test_array = np.asarray(test_array)
@@ -36,6 +45,61 @@ class TestPowerMonitors(unittest.TestCase):
         np_check_array = np.array([check_vec_0, check_vec_1])
 
         np.testing.assert_allclose(np_test_array, np_check_array, 1e-12, 1e-12)
+
+    # Check that the truncation error for a straight line is consistent with the
+    # analytic expectation
+    def test_relative_truncation_error(self):
+        num_points_per_dimension = 2
+
+        extent = num_points_per_dimension
+        basis = Basis.Legendre
+        quadrature = Quadrature.GaussLobatto
+        mesh = Mesh1D(extent, basis, quadrature)
+        logical_coords = np.array(logical_coordinates(mesh))[0]
+
+        # Define the test function
+        slope, offset = 0.1, 1.0
+        test_vec = slope * logical_coords + offset
+
+        modes_all_dim = power_monitors(test_vec, mesh)
+        modes = modes_all_dim[0]
+
+        test_relative_truncation_error_exponent = relative_truncation_error(
+            modes, extent
+        )
+
+        test_relative_truncation_error = np.power(
+            10.0, -1.0 * test_relative_truncation_error_exponent
+        )
+
+        # For a linear function the slope and offset correspond to the power
+        # monitor values
+        # The weighted average of the highest modes is
+        avg = np.log10(np.abs(slope)) * np.exp(-0.25) + np.log10(
+            np.abs(offset)
+        ) * np.exp(-0.25)
+        avg = avg / (np.exp(-0.25) + np.exp(-0.25))
+        expected_relative_truncation_error = np.power(10.0, avg)
+
+        np.testing.assert_allclose(
+            test_relative_truncation_error,
+            expected_relative_truncation_error,
+            1e-12,
+            1e-12,
+        )
+
+        # Test absolute truncation_error
+        test_truncation_error = np.asarray(
+            truncation_error(modes_all_dim, test_vec)
+        )
+
+        expected_truncation_error = (
+            np.max(np.abs(test_vec)) * expected_relative_truncation_error
+        )
+
+        np.testing.assert_allclose(
+            test_truncation_error, expected_truncation_error, 1e-12, 1e-12
+        )
 
 
 if __name__ == "__main__":
--- a/tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
+++ b/tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
@@ -92,10 +92,73 @@ void test_power_monitors_second_impl() {
   CHECK_ITERABLE_APPROX(test_power_monitors, expected_power_monitors);
 }
 
+void test_relative_truncation_error_impl() {
+  // We recompute the truncation error for a function where we know the
+  // power monitors analytically
+  const size_t number_of_points_per_dimension = 8;
+  const Mesh<1_st> mesh{number_of_points_per_dimension,
+                        Spectral::Basis::Legendre,
+                        Spectral::Quadrature::GaussLobatto};
+  const auto logical_coords = logical_coordinates(mesh);
+
+  // Build a test function with no zero power monitors
+  const std::vector<int> coeffs = {0, 1, 2, 3, 4, 5, 6, 7};
+  DataVector u_nodal(mesh.number_of_grid_points(), 0.0);
+  double ampl = 0.0;
+  for (auto coeff : coeffs) {
+    ampl = pow(10.0, -coeff);
+    u_nodal +=
+        ampl *
+        Spectral::compute_basis_function_value<Spectral::Basis::Legendre>(
+            static_cast<size_t>(coeff), logical_coords.get(0_st));
+  }
+
+  // Compute the relative truncation error
+  const int last_coeff = 7;
+  double weight = 0.0;
+  double avg = 0.0;
+  double weight_sum = 0.0;
+  for (auto coeff : coeffs) {
+    ampl = pow(10.0, -coeff);
+    weight = exp(-square(coeff - last_coeff + 0.5));
+    avg += log10(ampl) * weight;
+    weight_sum += weight;
+  }
+  avg = avg / weight_sum;
+  // By construction the maximum of the magnitude of the first two modes is
+  // unity.
+  // We test the order of magnitude of the relative error
+  const double expected_relative_truncation_error = pow(10.0, avg);
+
+  const auto power_monitors =
+      PowerMonitors::power_monitors<1_st>(u_nodal, mesh);
+  const DataVector& power_monitor_x = gsl::at(power_monitors, 0_st);
+  // We use all of the modes as above
+  const double test_relative_truncation_error =
+      pow(10.0, -1.0 * PowerMonitors::relative_truncation_error(
+                           power_monitor_x, power_monitor_x.size()));
+
+  CHECK_ITERABLE_APPROX(expected_relative_truncation_error,
+                        test_relative_truncation_error);
+
+  // Test truncation error
+  const double test_truncation_error = gsl::at(
+      PowerMonitors::truncation_error<1_st>(power_monitors, u_nodal), 0_st);
+
+  // Compare with the result from the relative truncation error
+  const double expected_truncation_error_x =
+      max(abs(u_nodal)) *
+      pow(10.0, -1.0 * PowerMonitors::relative_truncation_error(
+                           power_monitor_x, power_monitor_x.size()));
+
+  CHECK_ITERABLE_APPROX(test_truncation_error, expected_truncation_error_x);
+}
+
 }  // namespace
 
 SPECTRE_TEST_CASE("Unit.Numerical.LinearOperators.PowerMonitors",
                   "[NumericalAlgorithms][LinearOperators][Unit]") {
   test_power_monitors_impl();
   test_power_monitors_second_impl();
+  test_relative_truncation_error_impl();
 }


src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -40,6 +74,7 @@ void power_monitors(const gsl::not_null<std::array<DataVector, Dim>*> result,
       }
       slice_sum /= n_slice;
       slice_sum = sqrt(slice_sum);
+      slice_sum = log10(std::max(slice_sum, log_floor));
```
nilsvu: I'd say let's not do this. You can easily take the log of the result of this function. Returning the log isn't more numerically precise or more intuitive or anything like that.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -19,6 +19,12 @@ class DataVector;
  */
 namespace PowerMonitors {
 
+namespace detail {
+
+double relative_truncation_error_impl(const DataVector& input_power_monitors,
```
nilsvu: You don't need to declare this function in the header. Instead, define it only in the cpp, and in an anonymous namespace (just delete the word `detail` so it's `namespace { ... }`).

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +51,80 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ *
+ * \f{align*}{
+ *  u_\mathrm{max} = \max_{ijk} \left\lvert u_{ijk} \right\rvert ,
+ * \f}
+ *
+ * where \f$ u_{ijk}\f$ are the nodal coefficients
+ * of variable \f$ u \f$.
+ *
+ */
+double maximum_of_variable(const DataVector& input_data_vector);
```
nilsdeppe: Do we really need 20 lines of code to wrap `max(abs(u))`? This function currently only gets used in one place.

nilsvu: I think this function is trivial enough that we don't need this separate function. Wrapper functions like this can actually make the code a bit harder to understand because you think it does something fancy when it really doesn't. So just inline the max(abs) where you need it.

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
```diff
@@ -22,12 +22,34 @@ void bind_power_monitors_impl(py::module& m) {  // NOLINT
           &power_monitors<Dim>),
         py::arg("input_data_vector"), py::arg("mesh"));
 }
+
+template <size_t Dim>
+void bind_relative_truncation_error_impl(py::module& m) {  // NOLINT
```
nilsvu: no need for these extra functions, you can just put these `m.def` calls in `bind_power_monitors_impl`.

nilsvu (changes requested):
You can squash these changes directly, I haven't reviewed the implementation yet

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +51,80 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ *   - \dfrac{\sum_{j} \log_{10} \left(P_j\right) w_j }{\sum_{j} w_j} ,
+ * \f}
+ *
+ * with weights
+ *
+ * \f{align*}{
+ *  w_j = \exp\left[ - (j - N_k + \dfrac{1}{2})^2 \right] .
+ * \f}
+ *
+ * where \f$ N_k \f$ is the number of power monitors. Here the second term
```
nilsvu: "number of power monitors" -> "number of modes or grid points in dimension k"

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +51,80 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+/// @{
+/*!
+ * \ingroup SpectralGroup
+ * \brief Relative truncation error.
+ *
+ * Truncation error according to Eqs. (57) and (58) of Ref.
+ * \cite Szilagyi2014fna, i.e.,
+ *
+ * \f{align*}{
+ *  \mathcal{T}\left[P_k\right] = \log_{10} \max \left(P_0, P_1\right)
```
nilsvu, line 57: Shall we remove the log10 here too? As mentioned in the comment above, it's easy enough to take a log of the return value if you want to, or plot it on a log scale. I think that's more intuitive than returning the log (you also never quite know what basis was chosen for the log and have to look it up in the docs). Also, your `error_estimate` function below exponentiates this again. So, just remove the log here.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +51,80 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ *  u_\mathrm{max}} ,
+ * \f}
+ *
+ * where \f$ \mathcal{T}[P_k] \f$ is the relative error estimate computed from
+ * the power monitors \f$ P_k \f$. We set \f$ \mathrm{atol} \f$ to a small
+ * number (\f$ 10^{-14} \f$) in order to avoid dividing by zero if \f$ u \f$
+ * vanishes. We set \f$ \mathrm{rtol} = 1 \f$.
+ */
+template <size_t Dim>
+void error_estimate(gsl::not_null<std::array<double, Dim>*> result,
```
nilsvu: maybe name this just `truncation_error`?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +51,80 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ * \ingroup SpectralGroup
+ * \brief Returns an estimate of the absolute truncation error.
+ *
+ * The estimate of the numerical error is given by
+ *
+ * \f{align*}{
+ *  \mathcal{E}\left[P_k\right] = \dfrac{u_\mathrm{max} \times 10^{
+ *   - \mathcal{T}[P_k]}}{
+ *  \mathrm{atol} + \mathrm{rtol} \times
+ *  u_\mathrm{max}} ,
```
nilsvu: Isn't this equation also a relative error, since there's a u_max in the denominator? Did you get this equation from a paper or dg-charm or something like that? If not, I suggest you just compute `T * max(abs(u))`, where `T` is the relative error from the other function (without the log).

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +51,80 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ * \f{align*}{
+ *  w_j = \exp\left[ - (j - N_k + \dfrac{1}{2})^2 \right] .
+ * \f}
+ *
+ * where \f$ N_k \f$ is the number of power monitors. Here the second term
+ * is a weighted average with larger weights toward the highest modes.
+ *
+ */
+template <size_t Dim>
+void relative_truncation_error(gsl::not_null<std::array<double, Dim>*> result,
```
nilsvu: you don't need a return-by-reference overload if you're just returning a std::array of doubles, so just delete it.

However, please add an overload that takes the pre-computed power monitors as a `std::array<DataVector, Dim>`.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -17,6 +17,39 @@
 
 namespace PowerMonitors {
 
+double detail::relative_truncation_error_impl(
+    const DataVector& input_power_monitors, const size_t upperBound) {
```
nilsvu: how about `upperBound` -> `num_modes`

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -17,6 +17,39 @@
...
+  // upperBound needs to be smaller than the number of power monitors in the
+  // current dimension, i.e.
+  // upperBound < gsl::at(input_power_monitors, sliced_dim).size()
```
nilsvu: `<=` I think? Also make this an assert: `ASSERT(num_modes <= ..., "message here")`

Also assert that num_modes and the size of the DataVector are both >= 2.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -17,6 +17,39 @@
...
+
+  // Compute weighted average and total sum in the current dimension
+  double weighted_average = 0.0;
+  double weight_sum = 0.0;
+  double weight_value = 0.0;
+  for (size_t index = 0; index < upperBound; ++index) {
+    // Compute current weight
+    weight_value = exp(-square(
+        index - static_cast<double>(upperBound) + 0.5)
+        );
```
nilsvu: Run clang-format over this file please. Easiest is to just run it like this when you have staged your changes in git:

```
$ git-clang-format
```

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -17,6 +17,39 @@
...
+    const DataVector& input_power_monitors, const size_t upperBound) {
+  // upperBound needs to be smaller than the number of power monitors in the
+  // current dimension, i.e.
+  // upperBound < gsl::at(input_power_monitors, sliced_dim).size()
+
+  // Compute weighted average and total sum in the current dimension
+  double weighted_average = 0.0;
+  double weight_sum = 0.0;
+  double weight_value = 0.0;
+  for (size_t index = 0; index < upperBound; ++index) {
```
nilsvu: (optional) just `index` -> `i` is fine

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
```diff
@@ -119,10 +119,143 @@ void test_power_monitors_second_impl() {
   CHECK_ITERABLE_APPROX(test_power_monitors, expected_power_monitors);
 }
 
+void test_relative_truncation_error_impl() {
```
nilsvu, line 95: can you compute the relative and absolute truncation error in the test function above, so you don't have to duplicate the setup? Same in the Python test.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -56,6 +89,65 @@ std::array<DataVector, Dim> power_monitors(
...
+    const gsl::not_null<std::array<double, Dim>*> result,
+    const DataVector& input_data_vector, const Mesh<Dim>& mesh) {
+  // Temporary: compute power monitors
+  const auto pm_array = power_monitors<Dim>(input_data_vector, mesh);
+
+  // Compute relative truncation error in each dimension
+  size_t number_of_points_per_dim = 0;
+  for (size_t sliced_dim = 0; sliced_dim < Dim; ++sliced_dim) {
+    // Number of power monitors per dimension
+    number_of_points_per_dim = gsl::at(pm_array, sliced_dim).size();
```
nilsvu: (optional) I don't think that moving the `size_t` out of the loop makes a big difference in performance (or any at all? not sure). I'd do this:

```cpp
const auto& modes =  gsl::at(pm_array, d);
detail::rel_trunc_err(modes, modes.size());
```

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -56,6 +89,65 @@ std::array<DataVector, Dim> power_monitors(
...
+template <size_t Dim>
+void relative_truncation_error(
+    const gsl::not_null<std::array<double, Dim>*> result,
+    const DataVector& input_data_vector, const Mesh<Dim>& mesh) {
+  // Temporary: compute power monitors
+  const auto pm_array = power_monitors<Dim>(input_data_vector, mesh);
+
+  // Compute relative truncation error in each dimension
+  size_t number_of_points_per_dim = 0;
+  for (size_t sliced_dim = 0; sliced_dim < Dim; ++sliced_dim) {
```
nilsvu: (optional) `sliced_dim` -> `d` is fine

nilsvu (changes requested):

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
guilara: True. I'll remove it.

guilara: I'll remove it.

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
guilara: Right. I'll remove them.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
guilara: Yes, I was following the documentation of dg-charm for this particular equation. I just wrote it as it was defining for now values for atol and rtol.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
guilara: I'll remove it and add the new function. If I wanted to do the same for error_estimate I would still need the input vector to compute the max abs. So should I leave it as it is?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
guilara: I wanted to use this function in the tests. Not sure how to call it if its not in the hpp file

nilsvu: You can also make it a proper public function if you prefer: no detail namespace or impl suffix, just a regular overload of `relative_truncation_error` that works in one dimension.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
nilsvu: You can add an overload for it that takes just the rel errors and the data

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
guilara: I have yet to remove the log10's from the power monitors function. I'll leave them here for now since I'm not sure about changing the formulas form the Szilagyi paper/dg-charm doc

nilsdeppe:
I don't understand the commit history here. Can you please clean up the commits so that you're only submitting things you've actually added/changed? It looks like there are a lot of commits from other PRs.

nilsvu:
@guilara One way to fix the commits is this:

```sh
git fetch upstream develop:develop
git rebase upstream/develop
git push --force
```

You can also do `git rebase -i upstream/develop` to have more control about cleaning up the history. And it's good to check the commits with `git log` before force-pushing to make sure you didn't lose any changes.

In this case, since it looks like your changes are all in the "Fix" commit, I suggest you use `git reset HEAD~2` to dissolve the first two commits and commit them again with a proper commit message.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
nilsdeppe: I think what's happening is there's one truncation error, and then you apply either only an absolute or relatively tolerance to it. I think the "right" way to define the truncation error (but we will need to experiment) is more akin to:
```
T=u - u_{non-pileup}
```
where `u_{non-pileup}` is the solution with all of the pileup modes set to zero. This gives you the absolute truncation error, and then you can use:
```
E=(u-u_{non_pileup})/(atol + rtol * max(abs(u), abs(u_{non-pileup})))
```
to get the "error measure" which, if bigger than 1 you refine and if smaller than 1 you're fine. This would put the entire truncation error business into the same framework as what is done with e.g. Dormand-Prince time stepping.

Not for this PR, but I think that's what we ultimately should try doing instead of what SpEC/dg-charm does.

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
```diff
@@ -92,10 +92,92 @@ void test_power_monitors_second_impl() {
...
+  for (size_t dim = 0; dim < 2; ++dim) {
+    u_nodal *=
+        Spectral::compute_basis_function_value<Spectral::Basis::Legendre>(
+            gsl::at(coeff, dim), logical_coords.get(dim));
+    u_nodal_one_less *=
+        Spectral::compute_basis_function_value<Spectral::Basis::Legendre>(
+            gsl::at(coeff, dim), logical_coords_one_less.get(dim));
+  }
+
+  //
```
nilsdeppe: Expand on this comment or remove?

nilsdeppe (commented):
Can you squash the test commit into the "adding" commit? Anything being added in one commit should be tested in the same commit

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +54,98 @@ std::array<DataVector, Dim> power_monitors(
...
+  // Compute weighted average and total sum in the current dimension
+  double weighted_average = 0.0;
+  double weight_sum = 0.0;
+  double weight_value = 0.0;
+  const double log_floor = 1.0e-34;
+  for (size_t index = 0; index < num_modes; ++index) {
+    // Compute current weight
+    weight_value = exp(-square(index - static_cast<double>(num_modes) + 0.5));
+    // Add weighted power monitor
+    // (Need to check log argument or add floor)
```
nilsvu: Use `if (not equal_within_roundoff(mode, 0.))` to avoid the `log_floor` and `max` stuff. Include it from `Utilities/EqualWithinRoundoff.hpp`.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +54,98 @@ std::array<DataVector, Dim> power_monitors(
...
+    // Add term to weighted sum
+    weight_sum += weight_value;
+  }
+  weighted_average /= weight_sum;
+
+  // Maximum between the first two power monitors
+  // (Need to check log argument or add floor)
+  double leading_term = std::max(gsl::at(input_power_monitors, 0_st),
+                                 gsl::at(input_power_monitors, 1_st));
+  leading_term = log10(std::max(leading_term, log_floor));
```
nilsvu: also use `equal_within_roundoff` here

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +54,98 @@ std::array<DataVector, Dim> power_monitors(
...
+}
+
+template <size_t Dim>
+std::array<double, Dim> relative_truncation_error(
+    const DataVector& input_data_vector, const Mesh<Dim>& mesh) {
+  ASSERT(2_st <= input_data_vector.size(),
+         "Number of data points needs to be larger or equal than 2.");
+  std::array<double, Dim> result{};
+  const auto pm_array = power_monitors<Dim>(input_data_vector, mesh);
+  result = relative_truncation_error(pm_array);
```
nilsvu: just `return` here

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +54,98 @@ std::array<DataVector, Dim> power_monitors(
...
+  return result;
+}
+
+template <size_t Dim>
+std::array<double, Dim> truncation_error(const DataVector& input_data_vector,
+                                       const Mesh<Dim>& mesh) {
+  // Get relative truncation error
+  auto trunc_error = relative_truncation_error<Dim>(input_data_vector, mesh);
```
nilsvu: `const` and put `relative` in the name, or just inline in the function call below

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +54,98 @@ std::array<DataVector, Dim> power_monitors(
...
+  // Calculate the error estimate
+  return truncation_error<Dim>(trunc_error, input_data_vector);
+}
+
+template <size_t Dim>
+std::array<double, Dim> truncation_error(
+    const std::array<double, Dim>& trunc_error,
```
nilsvu: name `relative_truncation_error`. It shouldn't be a problem that there's a function with that name too.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +54,98 @@ std::array<DataVector, Dim> power_monitors(
...
+    const DataVector& input_data_vector){
+  std::array<double, Dim> result{};
+  // Define tolerances
+  const double atol = 1.0e-14;
+  const double rtol = 1.0;
+
+  // Use infinity norm
+  auto umax = max(abs(input_data_vector));
```
nilsvu: `const double`

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +54,98 @@ std::array<DataVector, Dim> power_monitors(
...
+
+  double error = 0.0;
+  for (size_t sliced_dim = 0; sliced_dim < Dim; ++sliced_dim) {
+    double exponent = -1.0 * gsl::at(trunc_error, sliced_dim);
```
nilsvu: `const`, or just inline in line below

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
nilsvu: Ok. Please make clear in these docs that what is being computed is the negative log10 of the relative truncation error (right?), e.g. adjust the `\brief` at the top. Also add something like "this corresponds to the number of digits resolved by the spectral expansion".

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
nilsvu: Sounds good. I think for now just delete the denominator and just return `u_max * pow(10, -gsl::at(relative_truncation_error, sliced_dim))`. That the absolute error associated to Eq. (57) in the Szilagy paper. We can use that to do some first plots and write a first p-AMR criterion. Then we'll have to experiment with better criteria once AMR is working.

nilsvu (changes requested):

nilsvu:
Some comments from above are still unresolved (I clicked "resolved" for the ones that are done)

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
guilara: You were right, its exactly the same setting. I have merged the tests in one.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
guilara: I have now used the new function. However, I had to write an 'else' statement as I think I cannot drop the other terms. They should be set to some max resolution (which in log10 is not zero).

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
guilara: Corrected the description

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
guilara: Okay. I have now simplified this formula.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +55,94 @@ std::array<DataVector, Dim> power_monitors(
...
+         "Number of modes needs to be larger or equal than 2.");
+  // Compute weighted average and total sum in the current dimension
+  double weighted_average = 0.0;
+  double weight_sum = 0.0;
+  double weight_value = 0.0;
+  for (size_t i = 0; i < num_modes; ++i) {
+    // Compute current weight
+    weight_value = exp(-square(i - static_cast<double>(num_modes) + 0.5));
+    // Add weighted power monitor
+    const auto mode = gsl::at(input_power_monitors, i);
```
nilsvu: `const double` to make clear that your only copying a double here

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +55,94 @@ std::array<DataVector, Dim> power_monitors(
...
+    if (not equal_within_roundoff(mode, 0.0)) {
+      weighted_average += weight_value * log10(mode);
+    }
+    else {
+      // If zero within roundoff set a maximum resolution
+      weighted_average += - 34.0 * weight_value;
```
nilsvu: Why 34? Doesn't this term completely dominate the result if a mode is zero to numerical roundoff? Maybe we should ignore zero modes in the average or something like that. Do you have an idea?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +45,64 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ *
+ * with weights
+ *
+ * \f{align*}{
+ *  w_j = \exp\left[ - (j - N_k + \dfrac{1}{2})^2 \right] .
+ * \f}
+ *
+ * where \f$ N_k \f$ is the number of modes or gridpoints in dimension k. Here
+ * the second term is a weighted average with larger weights toward the highest
+ * modes. The negative of this number should correspond to the number of digits
```
nilsvu: Isn't this number always positive (or should be unless the modes are doing something weird)?

tests/Unit/NumericalAlgorithms/LinearOperators/Python/Test_PowerMonitors.py
```diff
@@ -36,6 +37,38 @@ def test_power_monitors(self):
...
+    def test_relative_truncation_error(self):
+        num_points_per_dimension = 3
+
+        extent = num_points_per_dimension
+        basis = Basis.Legendre
+        quadrature = Quadrature.GaussLobatto
+        mesh = Mesh2D(extent, basis, quadrature)
+
+        np_vec = np.ones(mesh.number_of_grid_points())
+        test_vec = DataVector(np_vec)
```
nilsvu: You don't need to do this, the Numpy array converts to a DataVector automatically

tests/Unit/NumericalAlgorithms/LinearOperators/Python/Test_PowerMonitors.py
```diff
@@ -36,6 +37,38 @@ def test_power_monitors(self):
...
+
+        test_relative_truncation_error_exponent = np.asarray(
+            relative_truncation_error(test_vec, mesh))
+
+        test_relative_truncation_error = np.power(
+            10.0 * np.ones(2), -1.0 * test_relative_truncation_error_exponent)
+
+        expected_truncation_error = 1.0e-16 * np.ones(2)
```
nilsvu: You can delete this and use just `0.` below, since you only check to a tolerance of 1e-12 anyway

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
nilsvu: Did this change maybe get lost in a rebase or something? I don't see it on GitHub :)

nilsvu (changes requested):

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
guilara: I can imagine the case when, for instance, even modes are zero (corresponding to -34) by some symmetry. Then this term might be large, but should average with the non zero ones giving some value in between --since at the end we divide by the total sum of the weights. 

But maybe instead of -34 I could use log10 of `std::numeric_limits<double>::epsilon()` ?

guilara: I've also been playing with the following variation which drops modes from the sum if they are zero within roundoff (and adjusting the weights when we drop modes on the higher end). 

I've noticed one issue with that implementation. (I put it here since I'm not yet convinced by it.) 

For the case of a test function that is equal to a constant (and for which modes of order >1 are zero) I cannot apply the Szilagyi formula anymore. Returning zero doesn't seem right since for this function one should easily get good resolution.
I am also worried that (for similar reasons) functions that are well resolved by few modes will seem to have large relative truncation errors that don't improve with the number of modes.

Then, I would still argue to keep a floor as its in the commit. 

```cpp
double relative_truncation_error(const DataVector& input_power_monitors,
                                 const size_t num_modes) {
  ASSERT(num_modes <= input_power_monitors.size(),
         "Number of modes needs less or equal than the number of input power "
         "monitors");
  ASSERT(2_st <= num_modes,
         "Number of modes needs to be larger or equal than 2.");

  // Maximum between the first two power monitors
  double leading_term = std::max(gsl::at(input_power_monitors, 0_st),
                                 gsl::at(input_power_monitors, 1_st));
  if (not equal_within_roundoff(leading_term, 0.0)) {
    leading_term = log10(leading_term);
  } else {
    // If leading term is zero within roundoff
    // cannot compute relative error using Eq. (57) of Ref.
    // \cite Szilagyi2014fna
    return 0.0;
  }

  // Compute weighted average and total sum in the current dimension
  double weighted_average = 0.0;
  double weight_sum = 0.0;
  double weight_value = 0.0;
  bool highest_non_zero_mode = false;

  int num_modes_in_average = num_modes;
  for (int i = num_modes_in_average - 1; 0 <= i; --i) {
    // Get power monitor
    const double mode = gsl::at(input_power_monitors, i);
    if (not equal_within_roundoff(mode, 0.0)) {
      // When we find the highest non zero mode this flag is set to true
      // and the number of modes considered in the average is fixed
      highest_non_zero_mode = true;
      // Compute current weight
      weight_value = exp(-square(i - num_modes_in_average + 0.5));
      weighted_average += weight_value * log10(mode);
    } else if (highest_non_zero_mode and num_modes_in_average > 2) {
      // Skip intermediate modes within roundoff error from zero
      continue;
    } else if (not highest_non_zero_mode and num_modes_in_average > 2) {
      // If the last mode is within roundoff error from zero remove from the
      // modes considered for the average
      // The new total number of modes is now
      --num_modes_in_average;
      // Reset values and restart
      i = num_modes_in_average - 1;
      weighted_average = 0.0;
      weight_sum = 0.0;
      continue;
    } else {
      // If all modes higher than the leading modes are within roundoff
      // no relative error can be computed
      return 0.0;
    }
    // Add term to weighted sum
    weight_sum += weight_value;
  }
  weighted_average /= weight_sum;

  // Compute relative truncation error
  return leading_term - weighted_average;
}
```

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
guilara: Yes, it is positive. I'll correct the text when I push the changes for the other comments.

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
guilara: Oh, I put the test for the absolute truncation error at the end of this test.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
nilsdeppe: Please drop the floor. The floor adds an absolute error tolerance to this function and has in practice not been needed for several thousand BBH simulations. I think we should first do what is working in SpEC, then we can experiment with different methods of estimating the truncation error later.

I'm not so worried about the highest modes being zero because I thought that's what the `num_modes` argument is used to fix. I think that's these "pile-up modes" that you're talking about, and `num_modes` passed into the truncation error estimate should ignore the pile-up modes. If you're worried about a mode being identically zero, I think adding an `ASSERT` for that is reasonable :)

I think symmetry cases like zero even modes are also going to be rare in practice. I can't think of any astrophysical system that has such a high degree of symmetry in polynomials.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -54,6 +55,94 @@ std::array<DataVector, Dim> power_monitors(
...
+    }
+    // Add term to weighted sum
+    weight_sum += weight_value;
+  }
+  weighted_average /= weight_sum;
+
+  // Maximum between the first two power monitors
+  double leading_term = std::max(gsl::at(input_power_monitors, 0_st),
```
nilsdeppe: no `gsl::at`, not needed for `DataVector`

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +45,64 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ * resolved by the spectral expansion.
+ *
+ */
+double relative_truncation_error(const DataVector& input_power_monitors,
+                                 const size_t num_modes);
+
+template <size_t Dim>
+std::array<double, Dim> relative_truncation_error(
```
nilsdeppe: Can you just do the 1d case right now? I'm worried about straying too far from what is working in SpEC and dg-charm.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +45,64 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ *
+ * \f{align*}{
+ *  \mathcal{E}\left[P_k\right] = u_\mathrm{max} \times 10^{- \mathcal{T}[P_k]},
+ * \f}
+ *
+ * where \f$ \mathcal{T}[P_k] \f$ is the relative error estimate computed from
+ * the power monitors \f$ P_k \f$.
+ */
+template <size_t Dim>
+std::array<double, Dim> truncation_error(const DataVector& input_data_vector,
```
nilsdeppe: can you please remove this overload? Again, I don't think we should stray from what we currently know works in SpEC & dg-charm. Doing so adds time to code review and adds more code we need to maintain that we don't know yet if we will need.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +45,64 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ * \f}
+ *
+ * where \f$ N_k \f$ is the number of modes or gridpoints in dimension k. Here
+ * the second term is a weighted average with larger weights toward the highest
+ * modes. The negative of this number should correspond to the number of digits
+ * resolved by the spectral expansion.
+ *
+ */
+double relative_truncation_error(const DataVector& input_power_monitors,
+                                 const size_t num_modes);
```
nilsdeppe: maybe `num_modes_to_use`? And document that it must be `<= input_power_monitors.size()`

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +45,64 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ *  w_j = \exp\left[ - (j - N_k + \dfrac{1}{2})^2 \right] .
+ * \f}
+ *
+ * where \f$ N_k \f$ is the number of modes or gridpoints in dimension k. Here
+ * the second term is a weighted average with larger weights toward the highest
+ * modes. The negative of this number should correspond to the number of digits
+ * resolved by the spectral expansion.
+ *
+ */
+double relative_truncation_error(const DataVector& input_power_monitors,
```
nilsdeppe: remove `input_` (all `const&` function arguments are by definition inputs) and remove `s` at end. This is for a single variable.

I'm wondering if the first argument should really be a `ModalVector` since I think it holds the model coefficients of the L2 norm of `u`. @nilsvu thoughts?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -45,4 +45,64 @@ std::array<DataVector, Dim> power_monitors(const DataVector& input_data_vector,
...
+ * the power monitors \f$ P_k \f$.
+ */
+template <size_t Dim>
+std::array<double, Dim> truncation_error(const DataVector& input_data_vector,
+                                       const Mesh<Dim>& mesh);
+
+template <size_t Dim>
+std::array<double, Dim> truncation_error(
+    const std::array<double, Dim>& relative_truncation_error,
+    const DataVector& input_data_vector);
```
nilsdeppe: remove `input_`

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
nilsdeppe: no `gsl::at` here for a `DataVector`, only for `std::array`

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -27,21 +28,21 @@ void power_monitors(const gsl::not_null<std::array<DataVector, Dim>*> result,
   double slice_sum = 0.0;
   size_t n_slice = 0;
   size_t n_stripe = 0;
-  for (size_t sliced_dim = 0; sliced_dim < Dim; ++sliced_dim) {
-    n_slice = mesh.extents().slice_away(sliced_dim).product();
-    n_stripe = mesh.extents(sliced_dim);
+  for (size_t d = 0; d < Dim; ++d) {
```
nilsdeppe: Why are you renaming these? I find `sliced_dim` to be a fairly clear variable name, same with `index`

nilsdeppe (changes requested):

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
nilsvu: my fault, I said `i` and `d` are fine for loop variables

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
nilsvu: Yes I agree it probably should be a `ModalVector`. I haven't added pybindings for `ModalVector` yet though (should be easy enough by copying most of what `DataVector` does). Open an issue to remember that we should do this, and/or add a comment to the code here?

guilara:
Hi. Did you have a chance to take a look at the last changes? The ones I implemented after our last discussion. I'm suspecting maybe you don't get a notification if I just update the code and write no further comments. Is that the case?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -48,9 +49,69 @@ void power_monitors(const gsl::not_null<std::array<DataVector, Dim>*> result,
...
+           "estimate the relative truncation error.");
+    weighted_average += weight_value * log10(mode);
+    // Add term to weighted sum
+    weight_sum += weight_value;
+  }
+  weighted_average /= weight_sum;
+
+  // Maximum between the first two power monitors
+  double leading_term = std::max(power_monitor[0], power_monitor[1]);
+  ASSERT(not equal_within_roundoff(leading_term, 0.0),
```
nilsdeppe: why is a solution that ranges from 1e-15 to 1e-30 not allowed? You've got an absolute tolerance in this `ASSERT`, same with that on line 78.  In my last comment I said _identically_ zero, anything that's not identically zero is technically fine, unless you start making assumptions about an absolute tolerance.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -48,9 +49,69 @@ void power_monitors(const gsl::not_null<std::array<DataVector, Dim>*> result,
...
+  leading_term = log10(leading_term);
+
+  // Compute relative truncation error
+  return leading_term - weighted_average;
+}
+
+template <size_t Dim>
+std::array<double, Dim> truncation_error(
+    const std::array<DataVector, Dim>& power_monitors,
+    const DataVector& data_vector) {
```
nilsdeppe: While I agree that this is technically a `DataVector`, can you give this a more descriptive name like `solution` or `variable` or something? Even `u` is better since `u` is often shorthand for the evolved vars. You can then name the first argument something like `power_monitors_of_u`

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
nilsdeppe: Sure, issue+code comment is fine, though it's not clear to me what makes changing it to a `ModalVector` in this PR difficult. Delaying it too long will make it a lot more work to refactor all the code that calls it...

nilsdeppe (changes requested):
Sorry, I don't get notifications when people push to their PRs (I got 100+ emails a day from GitHub when I disabled it...) Please just ping me in a comment on the PR or on Slack when you push changes and I'll look at them!

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PowerMonitors.cpp
nilsvu: Can you merge this into the `test_power_monitors_second_impl` function above to reduce the amount of duplicate setup code?

nilsvu (changes requested):
@guilara could you take a look at this PR? Would be nice to get it merged so we can move on implement AMR criteria.

guilara:
Sure, I'll take a look at it again @nilsvu

nilsdeppe (approved):
LGTM! Thanks for implementing this :)

nilsvu (approved):

PR #4726, opened 2023-02-13
Adds a mutator that calculates the coordinates of element faces abutting the worldtube

nikwit:
Adds a mutator that calculates the grid coordinates of all faces that abutting the worldtube. This is is for the worldtube singleton to call so it can interpolate its internal solution to the grid points and send it to each element.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFacesGridCoordinates.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFacesGridCoordinates.hpp
@@ -1,0 +1,100 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <optional>
+#include <unordered_map>
+#include <utility>
+#include <vector>
+
+#include "Domain/Block.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/ElementMap.hpp"
+#include "Domain/InterfaceLogicalCoordinates.hpp"
+#include "Domain/Structure/CreateInitialMesh.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Domain/Structure/ExcisionSphere.hpp"
+#include "Domain/Structure/InitialElementIds.hpp"
+#include "Domain/Tags.hpp"
+#include "Evolution/DiscontinuousGalerkin/Initialization/QuadratureTag.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace CurvedScalarWave::Worldtube::Initialization {
+
+/*!
+ * \brief Initializes a map of the grid coordinates centered on the worldtube of
+ * all faces that abut the worldtube with corresponding ElementIds.
+ *
+ * \details The worldtube singleton computes an internal solution in the grid
+ * frame and uses this map to evaluate compute it at the grid coordinates of
+ * each element face abutting the worldtube each time step. This data is sent to
+ * the corresponding element where it is used to apply pointwise boundary
+ * conditions. The `ElementFacesGridCoordinates` holds a map of all the element
+ * ids abutting the worldtube with the corresonding grid coordinates.
+ *
+ * \warning This currently assumes that initial domain remains the same and
+ * there is no AMR. To support this, the worldtube could send the
+ * coefficients of its internal solution to each element which can evaluate
+ * it on their current grid in the boundary conditions.
+ *
+ *  DataBox changes:
+ * - Adds: nothing
+ * - Removes: nothing
+ * - Modifies: Tags::ElementFacesCoordinatesMap<Dim>
+ */
+template <size_t Dim>
+struct InitializeElementFacesGridCoordinates {
+  using return_tags = tmpl::list<Tags::ElementFacesGridCoordinates<Dim>>;
+  using argument_tags =
+      tmpl::list<::domain::Tags::Domain<Dim>,
+                 ::domain::Tags::InitialExtents<Dim>,
+                 ::domain::Tags::InitialRefinementLevels<Dim>,
+                 evolution::dg::Tags::Quadrature, Tags::ExcisionSphere<Dim>>;
+  static void apply(
+      const gsl::not_null<std::unordered_map<
+          ElementId<Dim>, tnsr::I<DataVector, Dim, Frame::Grid>>*>
+          element_faces_grid_coords,
+      const Domain<Dim>& domain,
+      const std::vector<std::array<size_t, Dim>>& initial_extents,
+      const std::vector<std::array<size_t, Dim>>& initial_refinement,
+      const Spectral::Quadrature& quadrature,
+      const ::ExcisionSphere<Dim>& excision_sphere) {
+    const auto& blocks = domain.blocks();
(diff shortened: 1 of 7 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -66,5 +67,15 @@ struct ExcisionSphere : db::SimpleTag {
   }
 };
 
+/*!
+ * \brief A map that holds the grid coordinates of element faces abutting the
+ * worldtube with the corresponding ElementIds.
+ */
+template <size_t Dim>
+struct ElementFacesCoordinatesMap : db::SimpleTag {
```
nilsdeppe: Maybe name this `ElementFacesGridCoordinates`? `CoordinatesMap` to me is confusing with our actually coordinate map objects.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFaceCoordinatesMap.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace CurvedScalarWave::Worldtube::Initialization {
+
+/*!
+ * \brief Initializes a map of the grid coordinates of all faces that abut the
+ * worldtube with corresponding ElementIds.
```
nilsdeppe: Could you add what component this should be called on, why it needs initial extents and initial refinement, and also the associated assumptions this mutator makes? My understanding is this will get run on the Worldtube component, and assumes that no AMR happens because it explicitly uses the initial domain to compute the grid coordinates. I think that's totally fine, but should just be documented :)

nilsdeppe (changes requested):
Looks great! Just requesting some more documentation. Please rebase and squash immediately! Thanks for moving all this code into develop! It'll be awesome to have everything you've accomplished with the worldtube excision available :D

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFaceCoordinatesMap.hpp
nilsvu: Also, could you briefly outline why you need this information on the worldtube? It's always hard to review small pieces of code without the bigger context. To _receive_ data the worldtube doesn't need to know in advance which elements are sending data. To _send_ data (or impose boundary conditions) the worldtube could just remember from which elements it received data in that timestep and respond. I mention this because with AMR this sort of element registration will become a lot more complicated, so if you can avoid element registration altogether then it will be a lot easier for you to support AMR. So if you want to support AMR, decide if you want to plan for it now or refactor later (there's overhead attached to both).

nikwit: Thank you, I added some docs. 

@nilsvu you are right, to support AMR the worldtube would need to send the coefficients of the internal solution rather than evaluate it on each face which would likely also be more efficient. However, some differences do occur as you e.g. have to figure out how to rotate the coefficients into the inertial frame (which should be easy).

I definitely want to add this later but would like to pretty much do the same thing I did for the simulations in the paper on develop first so I can have a release/commit to repeat them if necessary.

nilsvu: Sounds good. Orientations are defined by the abutting_directions in the excision sphere, so available globally through the cache.

nilsdeppe (dismissed):

nilsdeppe:
LGTM, please squash!

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -72,7 +72,7 @@ struct ExcisionSphere : db::SimpleTag {
  * worldtube with the corresponding ElementIds.
  */
 template <size_t Dim>
-struct ElementFacesCoordinatesMap : db::SimpleTag {
+struct ElementFacesGridCoordinates : db::SimpleTag {
```
nilsvu, line 75: this rename is in the wrong commit, which is a bit confusing. Could you fix it?

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFacesGridCoordinates.hpp
```diff
@@ -0,0 +1,100 @@
...
+ * frame and evaluates it at the grid coordinates of each element face abutting
+ * the worldtube each time step. This data is sent to the corresponding element
+ * where it is used to apply pointwise boundary conditions. The
+ * `ElementFacesGridCoordinates` holds a map of all the element ids abutting the
+ * worldtube with the corresonding grid coordinates.
+ *
+ * \warning This currently assumes that initial domain remains the same and
+ * there is no AMR. To support this, the worldtube could send the
+ * coefficients of its internal solution to each element which can evaluate the
+ * it on their current grid in the boundary conditions.
```
nilsvu: delete "the"

I think the suggestion here could be: The elements could send their face grid coordinates alongside the data in each timestep, so the worldtube can respond with boundary conditions on these coordinates.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFacesGridCoordinates.hpp
```diff
@@ -0,0 +1,100 @@
...
+      const auto element_ids =
+          initial_element_ids(block_id, initial_refinement.at(block_id));
+      for (const auto element_id : element_ids) {
+        const auto direction = excision_sphere.abutting_direction(element_id);
+        if (direction.has_value()) {
+          const auto mesh = ::domain::Initialization::create_initial_mesh(
+              initial_extents, element_id, quadrature);
+          const auto face_mesh = mesh.slice_away(direction.value().dimension());
+          const auto& current_block = blocks.at(block_id);
+          ElementMap<Dim, Frame::Grid> element_map{
```
nilsvu: const

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFacesGridCoordinates.hpp
```diff
@@ -0,0 +1,100 @@
...
+              element_id,
+              current_block.is_time_dependent()
+                  ? current_block.moving_mesh_logical_to_grid_map().get_clone()
+                  : current_block.stationary_map().get_to_grid_frame()};
+          const auto face_logical_coords =
+              interface_logical_coordinates(face_mesh, direction.value());
+          auto faces_grid_coords = element_map(face_logical_coords);
+          for (size_t i = 0; i < 3; ++i) {
+            faces_grid_coords.get(i) -= worldtube_grid_coords.get(i);
```
nilsvu, line 91: Please add to the docs (of this function _and_ the tag) that this actually stores the grid coords _centered on the worldtube_. Seems important :D

nilsvu (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFacesGridCoordinates.hpp
nikwit: it seems a lot easier to just send the coefficients of its internal Taylor series and have each element evaluate it on its own face each time step.

nilsvu (approved):

PR #4727, opened 2023-02-13
Fix component naming in shape control system

knelli2:
## Proposed changes

The `m` returned from a SpherepackIterator is always positive (because of internal spherepack storage) so this was actually incorrect before because negatives `m`s were postiive.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Systems/Shape.hpp
+++ b/src/ControlSystem/Systems/Shape.hpp
@@ -69,8 +69,11 @@ struct Shape : tt::ConformsTo<protocols::ControlSystem> {
     const auto compact_index = iter.compact_index(i);
     if (compact_index.has_value()) {
       iter.set(compact_index.value());
-      return {"(l,m)=("s + get_output(iter.l()) + ","s + get_output(iter.m()) +
-              ")"s};
+      const int m =
+          iter.coefficient_array() == SpherepackIterator::CoefficientArray::a
+              ? static_cast<int>(iter.m())
+              : -static_cast<int>(iter.m());
+      return {"l"s + get_output(iter.l()) + "m"s + get_output(m)};
     } else {
       return std::nullopt;
     }
--- a/tests/Unit/ControlSystem/Systems/Test_Shape.cpp
+++ b/tests/Unit/ControlSystem/Systems/Test_Shape.cpp
@@ -460,7 +460,7 @@ void test_names() {
 
   CHECK(pretty_type::name<shape>() == "ShapeA");
 
-  const size_t l_max = 4;
+  const size_t l_max = 3;
   SpherepackIterator iter(l_max, l_max);
   const size_t size = iter.spherepack_array_size();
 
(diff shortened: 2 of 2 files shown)


src/ControlSystem/Systems/Shape.hpp
```diff
@@ -69,8 +69,9 @@ struct Shape : tt::ConformsTo<protocols::ControlSystem> {
     const auto compact_index = iter.compact_index(i);
     if (compact_index.has_value()) {
       iter.set(compact_index.value());
-      return {"(l,m)=("s + get_output(iter.l()) + ","s + get_output(iter.m()) +
-              ")"s};
+      const int m = i < num_components / 2 ? static_cast<int>(iter.m())
+                                           : -static_cast<int>(iter.m());
+      return {"(l,m)=("s + get_output(iter.l()) + ","s + get_output(m) + ")"s};
```
nilsvu: Maybe delete the "(l,m)=" and the parens, and just return "l,m"? Then I think you end up with names like `Shape_2,2` in the output which is intuitive enough I think.

src/ControlSystem/Systems/Shape.hpp
```diff
@@ -69,8 +69,9 @@ struct Shape : tt::ConformsTo<protocols::ControlSystem> {
     const auto compact_index = iter.compact_index(i);
     if (compact_index.has_value()) {
       iter.set(compact_index.value());
-      return {"(l,m)=("s + get_output(iter.l()) + ","s + get_output(iter.m()) +
-              ")"s};
+      const int m = i < num_components / 2 ? static_cast<int>(iter.m())
+                                           : -static_cast<int>(iter.m());
```
nilsvu: I think this should use `iter.coefficient_array()`, but the docs of `SpherepackIterator` should probably be improved to explain this.

nilsvu (changes requested):

src/ControlSystem/Systems/Shape.hpp
knelli2: With my current changes, the h5 file will look like:
```
/ControlSystem/Shape/(l,m)=(0,0).dat
/ControlSystem/Shape/(l,m)=(1,0).dat
/ControlSystem/Shape/(l,m)=(1,1).dat
/ControlSystem/Shape/(l,m)=(1,-1).dat
```

With your suggestion, it'll look like
```
/ControlSystem/Shape/0,0.dat
/ControlSystem/Shape/1,0.dat
/ControlSystem/Shape/1,1.dat
/ControlSystem/Shape/1,-1.dat
```

I think the second method looks a bit strange, but I don't have a strong preference

nilsvu: How about `Shape/l2m2.dat`?

knelli2: I like it! I'll do that

nilsvu (approved):

PR #4732, opened 2023-02-14
Added function to find contour in an element.

PunJustice:
## Proposed changes

Adds a function to look for a given level contour in an element along provided rays, intended for use in the initial data solver.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * \param target Target value for the contour level in the data.
+ * \param extents Extents for the data and coordinates.
+ * \param inertial_coords Coordinates in the element in Frame::Inertial.
+ * \param data Data to find the contour in.
+ * \param xi_points Values of \f$\xi\f$ to interpolate to, likely returned by
+ * `element_logical_coordinates`. \param eta_points Values of \f$\eta\f$ to
+ * interpolate to, likely returned by `element_logical_coordinates`.
+ */
+std::vector<std::optional<double>> FindInElement(
```
nilsvu: Please use snake_case for all function and variable names, and CamelCase for class names. See our code review guide for these sort of stylistic things: https://spectre-code.org/code_review_guide.html

Maybe name this function just `find_surface`?

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp
```diff
@@ -0,0 +1,36 @@
...
+ * inevitably be passed to a singleton to create a Strahlkorper. We assume the
+ * element is part of a wedge block, so that the \f$\zeta\f$ logical coordinate
+ * points in the radial direction.
+ *
+ * \param target Target value for the contour level in the data.
+ * \param extents Extents for the data and coordinates.
+ * \param inertial_coords Coordinates in the element in Frame::Inertial.
+ * \param data Data to find the contour in.
+ * \param xi_points Values of \f$\xi\f$ to interpolate to, likely returned by
+ * `element_logical_coordinates`. \param eta_points Values of \f$\eta\f$ to
```
nilsvu: `\param` on new line

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,109 @@
...
+
+namespace SurfaceFinder {
+
+std::vector<std::optional<double>> FindInElement(
+    const double target, Index<3>& extents,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& inertial_coords,
+    const Scalar<DataVector>& data, const DataVector& xhi_points,
+    const DataVector& eta_points) {
+  // Build the matrices for interpolation, as well as the result DataVector
+  const size_t xhi_number = xhi_points.size();
```
nilsvu: Maybe name `num_rays`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,109 @@
...
+  const size_t eta_number = eta_points.size();
```
nilsvu: The number of points in both dims must be the same. Assert that:

```cpp
const size_t num_rays = xi_points.size();
ASSERT(eta_points.size() == num_rays, "error message here")
```

And include `Utilities/ErrorHandling/Assert.hpp`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,109 @@
...
+  DataVector values_;
+  Index<1> extent_;
+};
+
+namespace SurfaceFinder {
+
+std::vector<std::optional<double>> FindInElement(
+    const double target, Index<3>& extents,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& inertial_coords,
+    const Scalar<DataVector>& data, const DataVector& xhi_points,
```
nilsvu: `xhi` -> `xi` I think everywhere in this PR

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,109 @@
...
+    const DataVector& eta_points) {
+  // Build the matrices for interpolation, as well as the result DataVector
+  const size_t xhi_number = xhi_points.size();
+  const size_t eta_number = eta_points.size();
+  const Matrix xhi_interp_matrix =
+      Spectral::interpolation_matrix<Spectral::Basis::Legendre,
+                                     Spectral::Quadrature::GaussLobatto>(
+          extents[0], xhi_points);
```
nilsvu: Instead of taking the `const Index<3>& extents` as an argument, take a `const Mesh<3>& mesh`. Then pass `mesh.slice_through(0)` to `interpolation_matrix`. That's calling the second overload of `interpolation_matrix` that takes a `Mesh<1>` and needs no `Basis` and `Quadrature` template params.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,109 @@
...
+  Index<1> extent_;
+};
+
+namespace SurfaceFinder {
+
+std::vector<std::optional<double>> FindInElement(
+    const double target, Index<3>& extents,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& inertial_coords,
+    const Scalar<DataVector>& data, const DataVector& xhi_points,
+    const DataVector& eta_points) {
```
nilsvu: You currently assume that these two DataVectors form a regular grid of `xi * eta` points. Unless I'm missing something that's not the case for the collocation points of a Strahlkorper. Please correct me if I'm wrong. Instead these two DataVectors should list all logical coordinates in flat storage. So instead of passing `{-1, 0, 1}` to both coords and assume the list of coords is `(-1,-1), (-1,0), (-1,1), (0,-1), (0,0), ...`, you should pass `{-1, -1, -1, 0, 0, ...}` to the first and `{-1, 0, 1, -1, 0, ...}` to the second coord (that's the xi and eta coord of all points, respectively).

To make this clearer, I think it's best to pass the logical coordinates of the rays as a `tnsr::I<DataVector, 2, Frame::ElementLogical>`. That preserves the frame information, and makes the data layout of the points clearer (tensors always hold data on a flattened grid of points). You can create the tensor like this: `tnsr::I<DataVector, 2, Frame::ElementLogical>{{{xi_points, eta_points}}}`.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,109 @@
...
+      Spectral::interpolation_matrix<Spectral::Basis::Legendre,
+                                     Spectral::Quadrature::GaussLobatto>(
+          extents[0], xhi_points);
+  const Matrix eta_interp_matrix =
+      Spectral::interpolation_matrix<Spectral::Basis::Legendre,
+                                     Spectral::Quadrature::GaussLobatto>(
+          extents[1], eta_points);
+  std::array<Matrix, 3> interp_array;
+  interp_array[0] = xhi_interp_matrix;
+  interp_array[1] = eta_interp_matrix;
```
nilsvu: assign the matrices directly to the array components to avoid the copy here. (alternatively use `std::move`, but assigning directly seems easiest)

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,109 @@
...
+  std::vector<std::optional<double>> result(xhi_number * eta_number);
+
+  // Interpolate the grid onto the rays, same with the radii. Subtract off the
+  // target contour.
+  const Scalar<DataVector> radii = magnitude(inertial_coords);
+  const Scalar<DataVector> interpolated_data{
+      apply_matrices(interp_array, data[0], extents)};
+  const Scalar<DataVector> interpolated_radii{
+      apply_matrices(interp_array, radii[0], extents)};
+  const Index<3> new_extents{
```
nilsvu: this doesn't exist anymore when making the change to xi_points and eta_points commented above. The rays don't form a regular grid. You can loop over the rays directly.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,109 @@
...
+  interp_array[1] = eta_interp_matrix;
+  std::vector<std::optional<double>> result(xhi_number * eta_number);
+
+  // Interpolate the grid onto the rays, same with the radii. Subtract off the
+  // target contour.
+  const Scalar<DataVector> radii = magnitude(inertial_coords);
+  const Scalar<DataVector> interpolated_data{
+      apply_matrices(interp_array, data[0], extents)};
+  const Scalar<DataVector> interpolated_radii{
+      apply_matrices(interp_array, radii[0], extents)};
```
nilsvu: Why do you need to interpolate the radii at all? I think from reading your code you're using it only to get the radius in inertial coords from the radius in logical coords. Instead, just return the radius in logical coordinates here. It can be mapped to inertial coords outside of this function using the coordinate maps. Then I think you also don't need the `inertial_coordinates` argument to this function anymore.

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,105 @@
...
+  const double inner_radius = 1.;
+  const double outer_radius = 5.;
+  const double sphericity = 0.5;
+  const size_t ref = 0;
+  const domain::creators::Sphere sphere_helper{
+      inner_radius, outer_radius, sphericity, ref, grid_points_r_angular, true};
+  const std::vector<std::array<size_t, dim>> ref_levels{
+      sphere_helper.initial_refinement_levels()};
+  const auto extents = sphere_helper.initial_extents();
+  Domain<dim> dom{sphere_helper.create_domain()};
```
nilsvu: Please don't abbreviate names too much. It's not so much work to type `domain` and makes the code a lot more readable :)

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,105 @@
...
+
+  std::vector<ElementId<dim>> ids = initial_element_ids(ref_levels);
+  const size_t N_elems = ids.size();
```
nilsvu: num_elements (always snake_case)

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,105 @@
...
+#include "NumericalAlgorithms/Spectral/LogicalCoordinates.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/YlmSpherepack.hpp"
+#include "ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp"
+#include "PointwiseFunctions/MathFunctions/Gaussian.hpp"
+
+SPECTRE_TEST_CASE("Unit.ParallelAlgorithms.SurfaceFinder.SurfaceFinder",
+                  "[Unit][ParallelAlgorithms]") {
```
nilsvu: Please add a few comments to this test to explain what's going on. Tests are often useful for people to understand how the code is intended to be used, so it's nice if comments talk you through what's happening :)

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,105 @@
...
+  std::unordered_map<ElementId<dim>, Scalar<DataVector>> radii_map;
+  std::unordered_map<ElementId<dim>, Scalar<DataVector>> data_map;
+  std::unordered_map<ElementId<dim>, Index<3>> extents_map;
+  std::unordered_map<ElementId<dim>,
+                     tnsr::I<DataVector, dim, Frame::ElementLogical>>
+      logical_coords_map(N_elems);
+  std::unordered_map<ElementId<dim>, tnsr::I<DataVector, dim, Frame::Inertial>>
+      inertial_coords_map(N_elems);
+
+  for (size_t i = 0; i < N_elems; i++) {
```
nilsvu: Either set up just a single element and find a contour on it, or set up a full sphere and construct a Strahlkorper from the full surface. Or both :) I think both would actually be very nice: single element as a unit test, and constructing a Strahlkorper as an integration test because that's how we want to use this code, so we want to make sure everything fits together.

nilsvu (changes requested):

src/NumericalAlgorithms/Spectral/Spectral.cpp
```diff
@@ -960,7 +960,8 @@ template Matrix Spectral::interpolation_matrix(const Mesh<1>&,
                                                const DataVector&);
 template Matrix Spectral::interpolation_matrix(const Mesh<1>&,
                                                const std::vector<double>&);
-
+template Matrix Spectral::interpolation_matrix(const Mesh<1>&,
```
nikwit, line 963: can you add a small test for this instantiation to make sure it works as you expect?

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp
```diff
@@ -0,0 +1,37 @@
...
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+
+/// \ingroup SurfacesGroup
+/// Contains functions that are used to find contour levels.
+namespace SurfaceFinder {
+/*!
+ * \ingroup SurfacesGroup
+
+ * \brief Function that interpolates data onto radial rays in the direction
+ * of the logical coordiantes \f$\xi\f$ and \f$\eta\f$ and tries to perform a
```
nikwit: coordinates*

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * multiple roots are along a ray.
+ *
+ * \param target target value for the contour level in the data.
+ * \param mesh mesh for the element.
+ * \param data data to find the contour in.
+ * \param logical_direction tensor containing the \f$\xi\f$ and \f$\eta\f$
+ * values of the rays extending into \f$\zeta\f$ direction.
+ * \param tolerance tolerance for toms748 rootfind.
+ */
+std::vector<std::optional<double>> find_surface(
```
nikwit: should this maybe be called `find_wedge_surface` if we assume it is a wedge?

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+};
+
+namespace SurfaceFinder {
+
+std::vector<std::optional<double>> find_surface(
+    const double target, const Mesh<3>& mesh, const Scalar<DataVector>& data,
+    const tnsr::I<DataVector, 2, Frame::ElementLogical>& logical_direction,
+    const double tolerance) {
+  const size_t num_rays = logical_direction[0].size();
+  const DataVector subtracted_data{data[0] -
```
nikwit: how about `const DataVector subtracted_data = get(data) - target;` ?

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+  std::array<Matrix, 3> interp_array;
+  std::vector<std::optional<double>> result(num_rays);
+  DataVector interpolated_data;
+
+  for (size_t i = 0; i < num_rays; i++) {
+    // Potential speed-up: Currently, the interpolation is done one ray at a
+    // time (i.e. the interpolation matrices are built to interpolate to one
+    // ray, and then applied to the data sequentially). Instead one could use
+    // something like irregular interpolant to do this all at once.
+    interp_array[0] = Spectral::interpolation_matrix(mesh.slice_through(0),
```
nikwit: you are computing this sliced mesh every iteration, can you compute it before the loop?

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+namespace SurfaceFinder {
+
+std::vector<std::optional<double>> find_surface(
+    const double target, const Mesh<3>& mesh, const Scalar<DataVector>& data,
+    const tnsr::I<DataVector, 2, Frame::ElementLogical>& logical_direction,
+    const double tolerance) {
+  const size_t num_rays = logical_direction[0].size();
+  const DataVector subtracted_data{data[0] -
+                                   DataVector{data[0].size(), target}};
+  std::array<Matrix, 3> interp_array;
```
nikwit: can you write explicitly that the third one is not initialized on purpose so it's skipped by `apply_matrices`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+  std::vector<std::optional<double>> result(num_rays);
+  DataVector interpolated_data;
+
+  for (size_t i = 0; i < num_rays; i++) {
+    // Potential speed-up: Currently, the interpolation is done one ray at a
+    // time (i.e. the interpolation matrices are built to interpolate to one
+    // ray, and then applied to the data sequentially). Instead one could use
+    // something like irregular interpolant to do this all at once.
+    interp_array[0] = Spectral::interpolation_matrix(mesh.slice_through(0),
+                                                     logical_direction[0][i]);
```
nikwit: use the `get` method for accessing tensor components.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+std::vector<std::optional<double>> find_surface(
+    const double target, const Mesh<3>& mesh, const Scalar<DataVector>& data,
+    const tnsr::I<DataVector, 2, Frame::ElementLogical>& logical_direction,
+    const double tolerance) {
+  const size_t num_rays = logical_direction[0].size();
+  const DataVector subtracted_data{data[0] -
+                                   DataVector{data[0].size(), target}};
+  std::array<Matrix, 3> interp_array;
+  std::vector<std::optional<double>> result(num_rays);
+  DataVector interpolated_data;
```
nikwit: it's probably best to initialize to the correct size right here. Good that you are re-using this allocation!

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+    // time (i.e. the interpolation matrices are built to interpolate to one
+    // ray, and then applied to the data sequentially). Instead one could use
+    // something like irregular interpolant to do this all at once.
+    interp_array[0] = Spectral::interpolation_matrix(mesh.slice_through(0),
+                                                     logical_direction[0][i]);
+    interp_array[1] = Spectral::interpolation_matrix(mesh.slice_through(1),
+                                                     logical_direction[1][i]);
+
+    // Interpolate data onto the ray.
+    interpolated_data =
```
nikwit: can you use the return by reference overload of `apply_matrices` ?

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+
+std::vector<std::optional<double>> find_surface(
+    const double target, const Mesh<3>& mesh, const Scalar<DataVector>& data,
+    const tnsr::I<DataVector, 2, Frame::ElementLogical>& logical_direction,
+    const double tolerance) {
+  const size_t num_rays = logical_direction[0].size();
+  const DataVector subtracted_data{data[0] -
+                                   DataVector{data[0].size(), target}};
+  std::array<Matrix, 3> interp_array;
+  std::vector<std::optional<double>> result(num_rays);
```
nikwit: can you explicitly initialize all of these to `std::nullopt` with `result(num_rays, std::nullopt)` ?

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+                                                     logical_direction[0][i]);
+    interp_array[1] = Spectral::interpolation_matrix(mesh.slice_through(1),
+                                                     logical_direction[1][i]);
+
+    // Interpolate data onto the ray.
+    interpolated_data =
+        apply_matrices(interp_array, subtracted_data, mesh.extents());
+
+    // Perform root-find.
+    if (interpolated_data[0] * interpolated_data[interpolated_data.size() - 1] <
```
nikwit: you want to see if you cross zero using the intermediate value theorem, right? how about `if( std::signbit(interpolated_data[0]) != std::signbit(interpolated_data[interpolated_data.size() - 1]))` 

its probably useful to define a variable for the size of interpolated_data as well to make this more readable.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+        0) {
+      RayHelper data_interpolator{interpolated_data};
```
nikwit: `const` RayHelper

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+
+// Wrapping for the interpolator for Toms748 rootfind.
+struct RayHelper {
+ public:
+  explicit RayHelper(const DataVector& ray)
+      : N_points_(ray.size()),
+        values_(ray),
+        extent_(Index<1>{std::array<size_t, 1>{N_points_}}) {}
```
nikwit: `extent_{{ray.size()}}` should work

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+  double operator()(double x) {
```
nikwit: `double operator()(const double x) const {`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "Domain/Domain.hpp"
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+
+// Wrapping for the interpolator for Toms748 rootfind.
+struct RayHelper {
+ public:
+  explicit RayHelper(const DataVector& ray)
```
nikwit: you don't need explicit I think

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+      : N_points_(ray.size()),
+        values_(ray),
+        extent_(Index<1>{std::array<size_t, 1>{N_points_}}) {}
+  double operator()(double x) {
+    Matrix int_mat =
```
nikwit: `const`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp
```diff
@@ -0,0 +1,37 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
```
nikwit, line 5: include `<vector>`, `<optional`>, `DataVector.hpp`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+                                       Spectral::Quadrature::GaussLobatto>(
+            N_points_, DataVector{1, x});
+    std::array<Matrix, 1> int_arr{int_mat};
+    DataVector result = apply_matrices(int_arr, values_, extent_);
+    return result[0];
+  }
+
+ private:
+  size_t N_points_;
+  DataVector values_;
```
nikwit: can you default initialize these?

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include <array>
+#include <cstddef>
+#include <limits>
+#include <string>
+#include <unordered_map>
```
nikwit: not needed

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include <array>
+#include <cstddef>
+#include <limits>
+#include <string>
```
nikwit: not needed

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include <array>
+#include <cstddef>
+#include <limits>
```
nikwit: not needed

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,86 @@
...
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp"
```
nikwit: can you put this right below the copyright?

nikwit (commented):
Nice. Here's the comments we discussed!

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "DataStructures/Matrix.hpp"
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "Domain/Domain.hpp"
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+#include "Utilities/Gsl.hpp"
+
+// Wrapping for the interpolator for Toms748 rootfind.
+struct RayHelper {
```
nilsvu: Move into the SurfaceFinder namespace below, and wrap in an additional "anonymous namespace" (just `namespace { ... }` without a name). That's what we do with everything that's not exposed in the header.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+    int_arr[0] =
+        Spectral::interpolation_matrix<Spectral::Basis::Legendre,
+                                       Spectral::Quadrature::GaussLobatto>(
+            N_points_, DataVector{1, x});
+    return apply_matrices(int_arr, values_, extent_)[0];
+  }
+
+ private:
+  size_t N_points_{};
+  DataVector values_{};
```
nilsvu: This copies the DataVector. Make this `const DataVector&` instead to avoid the copy.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "Domain/Domain.hpp"
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+#include "Utilities/Gsl.hpp"
+
+// Wrapping for the interpolator for Toms748 rootfind.
+struct RayHelper {
+ public:
+  explicit RayHelper(const DataVector& ray)
```
nilsvu: How about `RayInterpolant`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+      : N_points_(ray.size()), values_(ray), extent_{ray.size()} {}
+  double operator()(double x) const {
+    std::array<Matrix, 1> int_arr{};
```
nilsvu: Please avoid unnecessary abbreviations. How about `interpolation_matrices`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "NumericalAlgorithms/Spectral/Spectral.hpp"
+#include "Utilities/Gsl.hpp"
+
+// Wrapping for the interpolator for Toms748 rootfind.
+struct RayHelper {
+ public:
+  explicit RayHelper(const DataVector& ray)
+      : N_points_(ray.size()), values_(ray), extent_{ray.size()} {}
```
nilsvu: Take the Mesh<1> in addition to the DataVector as input, and store it instead of num_points and extent. Then use it in the function below to get the interpolation matrix (use the overload that takes a 1D mesh instead of Basis/Quadrature templates parameters).

You can actually delete the constructor altogether then, because you get an "aggregate constructor" automatically that initializes the two member variables.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+  double operator()(double x) const {
+    std::array<Matrix, 1> int_arr{};
+    int_arr[0] =
+        Spectral::interpolation_matrix<Spectral::Basis::Legendre,
+                                       Spectral::Quadrature::GaussLobatto>(
+            N_points_, DataVector{1, x});
+    return apply_matrices(int_arr, values_, extent_)[0];
+  }
+
+ private:
```
nilsvu: Make public to get the aggregate constructor for free. Then delete the trailing underscore from the member variables because they're public.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+  size_t N_points_{};
+  DataVector values_{};
+  Index<1> extent_{};
+};
+
+namespace SurfaceFinder {
+
+std::vector<std::optional<double>> find_wedge_surface(
+    const double target, const Mesh<3>& mesh, const Scalar<DataVector>& data,
+    const tnsr::I<DataVector, 2, Frame::ElementLogical>& logical_direction,
```
nilsvu: How about `angular_coords` ("logical" is already encoded in the tensor type)

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+    const double tolerance) {
+  const size_t num_rays = logical_direction[0].size();
+  const DataVector subtracted_data = get(data) - target;
+  // The third Matrix in this array is not initialised so it is skipped by
+  // apply_matrices()
+  std::array<Matrix, 3> interp_array;
```
nilsvu: Again avoid unnecessary abbreviations. `interpolation_matrices`

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+  std::vector<std::optional<double>> result(num_rays, std::nullopt);
+  const double ray_size = mesh.extents(2);
+  DataVector interpolated_data(ray_size);
+  const auto xi_index = mesh.slice_through(0);
```
nilsvu: xi_mesh and eta_mesh

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+    interp_array[0] =
+        Spectral::interpolation_matrix(xi_index, get<0>(logical_direction)[i]);
+    interp_array[1] =
+        Spectral::interpolation_matrix(eta_index, get<1>(logical_direction)[i]);
+
+    // Interpolate data onto the ray.
+    apply_matrices(make_not_null(&interpolated_data), interp_array,
+                   subtracted_data, mesh.extents());
+
+    // Perform root-find.
```
nilsvu: ...only if the element brackets a root.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+  DataVector values_{};
+  Index<1> extent_{};
+};
+
+namespace SurfaceFinder {
+
+std::vector<std::optional<double>> find_wedge_surface(
+    const double target, const Mesh<3>& mesh, const Scalar<DataVector>& data,
+    const tnsr::I<DataVector, 2, Frame::ElementLogical>& logical_direction,
+    const double tolerance) {
```
nilsvu: Take relative and absolute tolerance separately

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * \param target target value for the contour level in the data.
+ * \param mesh mesh for the element.
+ * \param data data to find the contour in.
+ * \param logical_direction tensor containing the \f$\xi\f$ and \f$\eta\f$
+ * values of the rays extending into \f$\zeta\f$ direction.
+ * \param tolerance tolerance for toms748 rootfind.
+ */
+std::vector<std::optional<double>> find_wedge_surface(
+    const double target, const Mesh<3>& mesh, const Scalar<DataVector>& data,
```
nilsvu: (Optional) Move the `data` argument to the front I think (that's just personal preference)

tests/Unit/NumericalAlgorithms/Spectral/Test_Spectral.cpp
```diff
@@ -557,6 +557,23 @@ void test_gauss_points_boundary_interpolation_and_lifting() {
...
+  const auto double_matrix =
+      Spectral::interpolation_matrix(mesh1d, interpolation_target);
+  const auto vector_matrix = Spectral::interpolation_matrix(
+      mesh1d, std::vector<double>{interpolation_target});
+  Approx local_approx = Approx::custom().epsilon(1e-12).scale(1.);
+  for (size_t i = 0; i < double_matrix.rows(); i++) {
+    for (size_t j = 0; j < double_matrix.columns(); j++) {
+      CHECK(double_matrix(i, j) == local_approx(vector_matrix(i, j)));
+    }
+  }
```
nilsvu: This can be just `CHECK_MATRIX_APPROX(double_matrix, vector_matrix)`

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  const MathFunctions::Gaussian<3, Frame::Inertial> offset_gaussian{
+      10., 5., std::array<double, 3>{0., 0., -1.}};
+  const auto data = offset_gaussian(inertial_coords);
+  const double target = 2.5;
+  const auto result =
+      SurfaceFinder::find_wedge_surface(target, mesh, data, ray_directions);
+
+  CHECK(result[0].has_value() == false);
+  CHECK(result[1].has_value() == true);
+  CHECK(result[0].has_value() == false);
```
nilsvu: The third should probably be `result[2]`.
Can you make this test a bit stronger by checking the value of the found root, if that's reasonably easy to compute? The found root depends on the resolution of course, so you can use a reasonably high resolution and compare with an appropriate tolerance.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp
```diff
@@ -0,0 +1,41 @@
...
+ * multiple roots are along a ray.
+ *
+ * \param target target value for the contour level in the data.
+ * \param mesh mesh for the element.
+ * \param data data to find the contour in.
+ * \param logical_direction tensor containing the \f$\xi\f$ and \f$\eta\f$
+ * values of the rays extending into \f$\zeta\f$ direction.
+ * \param tolerance tolerance for toms748 rootfind.
+ */
+std::vector<std::optional<double>> find_wedge_surface(
```
nilsvu: I like that the `wedge` in the name emphasizes that the function assumes the element is wedge shaped, but currently it sounds more like the function finds the surface of the wedge. How about `find_radial_surface`?

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  INFO(
+      "Take the collocation points from a Strahlkorper, find a particular "
+      "radius, and compare to the physical radius from the Strahlkorper.");
+
+  const double target = 4.0;
+  const auto data = magnitude(inertial_coords);
+
+  const Strahlkorper<Frame::Inertial> strahlkorper{
+      5, 5, 4.5, std::array<double, 3>{0., 0., 0.}};
+  const YlmSpherepack ylm = strahlkorper.ylm_spherepack();
```
nilsvu: const auto&

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  const auto [theta, phi] = ylm.theta_phi_points();
```
nilsvu: const auto&

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  const auto x = target * sin(theta) * cos(phi);
+  const auto y = target * sin(theta) * sin(phi);
+  const auto z = target * cos(phi);
```
nilsvu: Shouldn't this be cos(theta)? Why does this work?

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  const tnsr::I<DataVector, 3, Frame::Inertial> points{{{x, y, z}}};
+
+  const auto block_coords = block_logical_coordinates(domain, points);
+  auto element_coords = element_logical_coordinates(ids, block_coords);
```
nilsvu: const

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  tnsr::I<DataVector, 2, Frame::ElementLogical> ray_directions{
```
nilsvu: const

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  for (size_t i = 0; i < radius_zeta.size(); i++) {
+    CHECK(radius_zeta[i].has_value());
+    CHECK(radius_values[i] == custom_approx(target));
+  }
+}
+
+void test_strahlkorper_input(
+    const tnsr::I<DataVector, 3, Frame::Inertial>& inertial_coords,
+    const Mesh<3>& mesh, const Domain<3>& domain,
+    const std::vector<ElementId<3>>& ids,
```
nilsvu: Pass in only a single element ID because you use only one below

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  }
+}
+}  // namespace
+
+SPECTRE_TEST_CASE("Unit.ParallelAlgorithms.SurfaceFinder.SurfaceFinder",
+                  "[Unit][ParallelAlgorithms]") {
+  // Build wedge for tests.
+  static constexpr size_t dim = 3;
+  const std::array<size_t, 3> grid_points_r_angular{{7, 6, 6}};
+  const size_t init_refin = 0;
```
nilsvu: No unnecessary abbreviations (typing isn't / shouldn't be the hard part about coding :) your IDE surely has autocomplete).

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+    CHECK(radius_values[i] == custom_approx(target));
+  }
+}
+}  // namespace
+
+SPECTRE_TEST_CASE("Unit.ParallelAlgorithms.SurfaceFinder.SurfaceFinder",
+                  "[Unit][ParallelAlgorithms]") {
+  // Build wedge for tests.
+  static constexpr size_t dim = 3;
+  const std::array<size_t, 3> grid_points_r_angular{{7, 6, 6}};
```
nilsvu: first angular then r, no?

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  const size_t init_refin = 0;
+  const domain::creators::Sphere sphere{
+      1., 5., 0.9, init_refin, grid_points_r_angular, false};
```
nilsvu: The algorithm should work with sphericity=0, right? Would be good to use that here to make sure. You can just increase the number of grid points to meet the tolerance that you want.

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  const auto domain = sphere.create_domain();
+  const auto ref_levels = sphere.initial_refinement_levels();
+  const auto extents = sphere.initial_extents();
+  const auto ids = initial_element_ids(ref_levels);
```
nilsvu: You don't have to do this, you can just construct the single ElementId<3> that you want.

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,164 @@
...
+  const auto quadrature = Spectral::Quadrature::GaussLobatto;
+  const auto& block = domain.blocks()[ids[0].block_id()];
+  const auto mesh =
+      domain::Initialization::create_initial_mesh(extents, ids[0], quadrature);
+  const auto element =
+      domain::Initialization::create_initial_element(ids[0], block, ref_levels);
```
nilsvu: Don't need this I think

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,83 @@
...
+    interp_array[1] =
+        Spectral::interpolation_matrix(eta_index, get<1>(logical_direction)[i]);
+
+    // Interpolate data onto the ray.
+    apply_matrices(make_not_null(&interpolated_data), interp_array,
+                   subtracted_data, mesh.extents());
+
+    // Perform root-find.
+    if (std::signbit(interpolated_data[0]) !=
+        std::signbit(interpolated_data[ray_size - 1])) {
```
nilsvu: This doesn't work for Gauss quadrature because the outermost grid points are not at -1 and 1. Either disable for Gauss quadrature (emit an error if the radial mesh has Gauss quadrature) or just use the radial interpolant that you have already set up to evaluate the data at -1 and 1. For GaussLobatto quadrature you can still just take the outermost points directly.

tests/Unit/ParallelAlgorithms/SurfaceFinder/CMakeLists.txt
```diff
@@ -0,0 +1,25 @@
...
+  ""
+  )
+
+target_link_libraries(
+  ${LIBRARY}
+  PRIVATE
+  Domain
+  DomainCreators
+  DataStructures
+  SurfaceFinder
```
nilsvu: Alphabetize please

nilsvu (changes requested):

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
PunJustice: I think it shouldn't be disabled for Gauss quadrature, is it worth having a check for which quadrature it is or can I just use the interpolant always?

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
PunJustice: when adding const here, I get errors on line 125 citing
```
/Users/peterjamesnee/spectre/fork/tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp:125:30: error: no viable overloaded operator[] for type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>'
      {{get<0>(element_coords[id].element_logical_coords),
               ~~~~~~~~~~~~~~^~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/c++/v1/unordered_map:1404:18: note: candidate function not viable: 'this' argument has type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>', but method is not marked const
    mapped_type& operator[](const key_type& __k);
                 ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/c++/v1/unordered_map:1406:18: note: candidate function not viable: 'this' argument has type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>', but method is not marked const
    mapped_type& operator[](key_type&& __k);
                 ^
/Users/peterjamesnee/spectre/fork/tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp:126:30: error: no viable overloaded operator[] for type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>'
        get<1>(element_coords[id].element_logical_coords)}}};
               ~~~~~~~~~~~~~~^~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/c++/v1/unordered_map:1404:18: note: candidate function not viable: 'this' argument has type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>', but method is not marked const
    mapped_type& operator[](const key_type& __k);
                 ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/c++/v1/unordered_map:1406:18: note: candidate function not viable: 'this' argument has type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>', but method is not marked const
    mapped_type& operator[](key_type&& __k);
```

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
PunJustice: Yep my mistake, it still works since the `Strahlkorper` is spherical, so even if it is a wrong angle the root will be located at the same radius.

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
PunJustice: Adding const here results in the following error:

```
/Users/peterjamesnee/spectre/fork/tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp:125:30: error: no viable overloaded operator[] for type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>'
      {{get<0>(element_coords[id].element_logical_coords),
               ~~~~~~~~~~~~~~^~~
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/c++/v1/unordered_map:1404:18: note: candidate function not viable: 'this' argument has type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>', but method is not marked const
    mapped_type& operator[](const key_type& __k);
                 ^
/Applications/Xcode.app/Contents/Developer/Platforms/MacOSX.platform/Developer/SDKs/MacOSX12.3.sdk/usr/include/c++/v1/unordered_map:1406:18: note: candidate function not viable: 'this' argument has type 'const std::unordered_map<ElementId<3>, ElementLogicalCoordHolder<3>>', but method is not marked const
    mapped_type& operator[](key_type&& __k);
```

PunJustice: Nevermind, spoke to Niko about this!

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
nilsvu: It's an optimization 🤷‍♂️ if it's easy enough to do the switch between gauss and GaussLobatto then do it, else leave a comment.

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,170 @@
...
+      5, 5, 4.5, std::array<double, 3>{0., 0., 0.}};
+  const auto& ylm = strahlkorper.ylm_spherepack();
+  const auto& [theta, phi] = ylm.theta_phi_points();
+  const auto x = target * sin(theta) * cos(phi);
+  const auto y = target * sin(theta) * sin(phi);
+  const auto z = target * cos(theta);
+  const tnsr::I<DataVector, 3, Frame::Inertial> points{{{x, y, z}}};
+  const std::vector<ElementId<3>> id_vector{id};
+  const auto block_coords = block_logical_coordinates(domain, points);
+  auto element_coords = element_logical_coordinates(id_vector, block_coords);
```
nilsvu: const

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
```diff
@@ -0,0 +1,170 @@
...
+      {{get<0>(element_coords.at(id).element_logical_coords),
+        get<1>(element_coords.at(id).element_logical_coords)}}};
+
+  const auto radius_zeta =
+      SurfaceFinder::find_radial_surface(data, target, mesh, ray_directions);
+  const auto result = result_helper(ray_directions, radius_zeta);
+  const auto inertial_contour = element_map(result);
+  const auto radius_values = get(magnitude(inertial_contour));
+
+  Approx custom_approx = Approx::custom().epsilon(1.e-3).scale(1.0);
```
nilsvu: Can you tighten the tolerances in the tests if you use 12 points in all dimensions or 1 level of h refinement? With h refinement you just have to pick a suitable element that covers the surface, like `ElementId<3>{0, {{{1, 0}, {1, 0}, {1, 1}}}` or something like that. That's Block 0 (look in Sphere.cpp to understand which direction wedge that is), refinement level 1 in all directions, and the first element of two in angular directions and second element of two radially.

nilsvu (changes requested):
Looks good! You can squash the commits into one.

tests/Unit/ParallelAlgorithms/SurfaceFinder/Test_SurfaceFinder.cpp
PunJustice: I have tightened the tolerances by adding more points. To do so with h refinement would require rewriting one of the tests (the "bulging surface"), however if these still aren't tight enough I am happy to do so. :)

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,78 @@
...
+#include <array>
+#include <cstddef>
+#include <vector>
+
+#include "DataStructures/ApplyMatrices.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Index.hpp"
+#include "DataStructures/Matrix.hpp"
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "Domain/Domain.hpp"
```
nilsvu: You don't need this (anymore). Please delete this include and also remove `Domain` from the linked libraries in the CMakeLists.txt.

nilsvu (changes requested):
Looks great! One remaining thing to fix.

It's very reassuring that increasing resolution allowed you to tighten the tolerances by orders of magnitude.

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.cpp
```diff
@@ -0,0 +1,77 @@
...
+    const RayInterpolant data_interpolator{interpolated_data,
+                                           mesh.slice_through(2)};
+
+    // Perform root-find only if the element brackets a root.
+    if (std::signbit(data_interpolator(-1.)) !=
+        std::signbit(data_interpolator(1.))) {
+      result[i] = RootFinder::toms748(
+          data_interpolator, -1., 1., data_interpolator(-1.),
+          data_interpolator(1.), relative_tolerance, absolute_tolerance);
+    }
```
nilsvu: while you're fixing the thing above, just do the following as well (it's the rather "obvious" optimization we discussed above, let's just do it because it's so easy):

```suggestion
    const double lower_radial_bound = mesh.quadrature(2) == Spectral::Quadrature::GaussLobatto ? interpolated_data[0] : data_interpolator(-1.);
    const double upper_radial_bound = mesh.quadrature(2) == Spectral::Quadrature::GaussLobatto ? interpolated_data[mesh.extents(2) - 1] : data_interpolator(1.);
    if (std::signbit(lower_radial_bound)) !=
         std::signbit(upper_radial_bound)) {
       result[i] = RootFinder::toms748(
           data_interpolator, -1., 1., lower_radial_bound,
           upper_radial_bound, relative_tolerance, absolute_tolerance);
     }
```

nilsvu:
@ffoucart @nilsdeppe here's the surface finder algorithm we discussed some time ago, if you want to look at it before we merge it :)

src/ParallelAlgorithms/SurfaceFinder/SurfaceFinder.hpp
```diff
@@ -0,0 +1,43 @@
...
+ * \ingroup SurfacesGroup
+
+ * \brief Function that interpolates data onto radial rays in the direction
+ * of the logical coordinates \f$\xi\f$ and \f$\eta\f$ and tries to perform a
+ * root find of \f$\text{data}-\text{target}\f$. Returns the logical
+ * coordinates of the roots
+ * found.
+ *
+ * \details We assume the element is part of a wedge block, so that the
+ * \f$\zeta\f$ logical coordinate points in the radial direction. Will fail if
```
nilsdeppe, line 25: If I've understood the algorithm correctly, this restriction would be relatively straightforward to relax (replacing the hard-coded `0,1,2` with variables and passing in the direction). Could you add a sentence like `This can be generalized if necessary by passing in which logical direction to search for the surface.`?

nilsdeppe (commented):
I didn't do a detailed review, one small suggestion that's optional. I don't need to look at this again before it gets merged :)

nilsvu (dismissed):

ffoucart:
The main thing that seems different from the existing SpEC method is that you interpolate onto specific radial rays before finding the surface, while SpEC finds the surface along each radial ray of the grid, then projects the result r(theta,phi) onto spherical harmonics. We'll have to see if that's an issue at all or not; we could run into larger interpolation errors close to the surface this way.

nilsvu (approved):

nilsvu:
Thanks for your first contribution to SpECTRE @PunJustice! It's much appreciated and we invite you to add your name to the author list:

1. Edit the `Metadata.yaml` file in the repository.
2. Add an entry to the list in `Authors.Contributors` with your name, affiliation, etc. Note that the list is ordered alphabetically by last name.
3. Commit and open a PR with the change.

Once the pull request is merged, your name will appear on the [SpECTRE DOI on Zenodo](https://doi.org/10.5281/zenodo.4290404) with the next public release.

PR #4733, opened 2023-02-14
Add derivation of ConstraintPreserving BCs of CSW to docs and fix typo

nikwit:
@nilsdeppe can you check this against your derivation?

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/ConstraintPreservingSphericalRadiation.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/ConstraintPreservingSphericalRadiation.hpp
@@ -41,9 +41,8 @@ namespace CurvedScalarWave::BoundaryConditions {
  * are still likely to occur.
  *
  * The constraint-preserving part of the boundary conditions are set on the time
- * derivatives of the evolved fields according to \cite Holst2004wt . The
- * physical Bayliss-Turkel boundary conditions are additionally set onto the
- * time derivative of \f$\Pi\f$.
+ * derivatives of all evolved fields. The physical Bayliss-Turkel boundary
+ * conditions are additionally set onto the time derivative of \f$\Pi\f$.
  *
  * The constraints are defined as follows:
  *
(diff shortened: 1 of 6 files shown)


src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/ConstraintPreservingSphericalRadiation.hpp
```diff
@@ -41,35 +41,78 @@ namespace CurvedScalarWave::BoundaryConditions {
...
- *                    \lambda_\Psi n^i \mathcal{C}_i, \\
- * \partial_{t}\Pi&\to\partial_{t}\Pi-\left(\partial_t\Pi
- *                  - \partial_t\Pi^{\mathrm{BC}}\right)
- *                  +\gamma_2\lambda_\Psi n^i \mathcal{C}_i
- *                  =\partial_t\Pi^{\mathrm{BC}}
- *                  +\gamma_2\lambda_\Psi n^i \mathcal{C}_i, \\
- * \partial_{t}\Phi_i&\to\partial_{t}\Phi_i+
- *                     \lambda_0n^jP^k{}_i\mathcal{C}_{jk}
- *   = \partial_{t}\Phi_i+ \lambda_0n^j \mathcal{C}_{ji}.
+ * n^i \mathcal{C}_i &= n^i \partial_i Z^1 - \frac{1}{2}(U^{+} - U^-) + n^i
```
nilsdeppe: Flat SW uses `w^\Psi`, `w^0_i` `w^\pm` so let's be consistent with that. This is also the notation that the Rezzolla book uses, and is quite standard in the engineering literature. It's basically only SpEC that uses this notation.

src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/ConstraintPreservingSphericalRadiation.hpp
```diff
@@ -41,35 +41,78 @@ namespace CurvedScalarWave::BoundaryConditions {
...
+ * Z_i^2 \stackrel{!}{=} 0 \\
```
nilsdeppe: sorry, how should I interpret/understand `!=` compared to $\ne$?

src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/ConstraintPreservingSphericalRadiation.hpp
```diff
@@ -41,35 +41,78 @@ namespace CurvedScalarWave::BoundaryConditions {
...
+ * \partial_j Z^2_i
+ * \f}
+ *
+ * This condition is applied to the time derivative using the Bjorhus condition
+ * \cite Bjorhus1995 :
+ * \f{align*}{
+ * \partial_t u^\alpha + A^{i \alpha}_\beta \partial_i u^\beta &= F^\alpha \\
+ *  e^{\hat{\alpha}}_\alpha (\partial_t u^\alpha + A^{i \alpha}_\beta
+ * \partial_i u^\beta) &= e^{\hat{\alpha}}_\alpha F^\alpha  \\
+ * \partial_t u^{\hat{\alpha}} + e^{\hat{\alpha}}_\alpha A^{i
```
nilsdeppe: should the `\partial_t u^{hat{alpha}` already be a `d_t ...` here and on line 90?

nilsdeppe (changes requested):
Looks good, matches what I've done, whatever that's worth 😅 

You can squash immediately.

src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/ConstraintPreservingSphericalRadiation.hpp
nikwit: The exclamation mark is above, not in front of the equality sign (it just means equal). I use it here to indicate that we are demanding that there are is no incoming radiation, it does not come from any variable transformation.

nilsdeppe: Sorry, I wasn't confused on the location of the exclamation point, just its meaning. Could you add a sentence after this block that says something like `where we use \stackrel{!}{=} means we are demanding this be equal.`?

nikwit: I just changed it to a normal equal sign now and made the phrasing a bit clearer. I thought this was fairly standard notation in Maths but maybe that is just in the German literature^^  but if it is not that clear it would be more confusing than helpful I suppose

src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/ConstraintPreservingSphericalRadiation.hpp
```diff
@@ -41,35 +41,79 @@ namespace CurvedScalarWave::BoundaryConditions {
...
+ * (n^i \partial_i w^\Psi)_{BC}  &= \frac{1}{2}(w^{+} - w^-) - n^i w_i^0
+ * \f}
+ *
+ * and
+ *
+ * \f{align*}{
+ * 2 n^i \mathcal{C}_{ij} &= n^i \partial_i w^0_j + \frac{1}{2}n^i n_j
+ * (\partial_i w^+
+ * - \partial_i w^-) - \frac{1}{2}(\partial_j w^+ - \partial_j w^-) - n^i
+ * \partial_j w^0_i \stackrel{!}{=} 0 \\
```
nilsdeppe: There's another occurrence here, sorry!

nikwit: removed

nilsdeppe (approved):

PR #4735, opened 2023-02-14
Only include object centers tags if needed

knelli2:
## Proposed changes

Some control systems don't need centers to compute their control errors (like shape). This allows control errors to specify if they need center tags, and if so, they'll be added to the `const_global_cache_tags` type alias in the control system initialization action.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Translation.hpp
+++ b/src/ControlSystem/ControlErrors/Translation.hpp
@@ -72,6 +72,9 @@ namespace ControlErrors {
 struct Translation : tt::ConformsTo<protocols::ControlError> {
   static constexpr size_t expected_number_of_excisions = 2;
 
+  using object_centers =
+      domain::object_list<domain::ObjectLabel::A, domain::ObjectLabel::B>;
+
   using options = tmpl::list<>;
   static constexpr Options::String help{
       "Computes the control error for translation control. This should not "
--- a/src/ControlSystem/Protocols/ControlError.hpp
+++ b/src/ControlSystem/Protocols/ControlError.hpp
@@ -24,6 +24,11 @@ namespace control_system::protocols {
 ///   the example shown here:
 /// - a `static constexpr size_t expected_number_of_excisions` which specifies
 ///   the number of excisions necessary in order to compute the control error.
+/// - a type alias `object_centers` to a `domain::object_list` of
+///   `domain::ObjectLabel`s. These are the objects that will require the
+///   `domain::Tags::ObjectCenter`s tags to be in the GlobalCache for this
+///   control system to work.
+///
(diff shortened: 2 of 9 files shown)


src/ControlSystem/Protocols/ControlError.hpp
```diff
@@ -24,6 +24,11 @@ namespace control_system::protocols {
 ///   the example shown here:
 /// - a `static constexpr size_t expected_number_of_excisions` which specifies
 ///   the number of excisions necessary in order to compute the control error.
+/// - a type alias `object_centers` to a `domain::object_list` of
```
nilsvu, line 27: How about you just require a type alias `const_global_cache_tags` and collect those in the parallel component or in the init action? Then there's less magic involved, and this behaves like other places in the code where we collect global cache tags.

nilsvu (changes requested):

src/ControlSystem/Protocols/ControlError.hpp
knelli2: The control errors need to be agnostic to how the center tags were put into the cache. The control errors grab the base tag. My plan for this was in the initialization action in the future choose what actual simple tags (that are derived from the base tag) to use depending on measurements.

nilsdeppe (approved):
LGTM! I think having the object list is fine, probably better than `const_global_cache_tags`, especially since, as you say, we need to be agnostic to how they are added.

src/ControlSystem/ControlErrors/Translation.hpp
```diff
@@ -72,6 +72,9 @@ namespace ControlErrors {
 struct Translation : tt::ConformsTo<protocols::ControlError> {
   static constexpr size_t expected_number_of_excisions = 2;
 
+  using object_centers =
+      domain::object_list<domain::ObjectLabel::A, domain::ObjectLabel::B>;
```
nilsvu, line 76: this one only uses A I think?

nilsdeppe: I think translation follows the COM of the total system, so uses A&B to get the COM of the binary

knelli2: Part of the translation control error calculation is the rotation and expansion control error calculations. So even though in this function itself it only uses A, rotation and expansion need both. Hence translation needs both.

nilsvu: Right I missed this issue with the base tags. Edit: this comment is for the other thread.

nilsvu (approved):

PR #4743, opened 2023-02-16
Add curved area element

nikwit:
## Proposed changes
adds a function that computes the area element / surface jacobian for curved space
the test computes the area of a Kerr black hole in Kerr Schild coordinates

@nilsvu I decided to rename it to `area_element` and `euclidean_area_element` after all because otherwise it should really be `det_euclidean_surface_jacobian` which I don't really like.

### DIFF
--- a/src/Domain/AreaElement.hpp
+++ b/src/Domain/AreaElement.hpp
@@ -10,8 +10,9 @@
 #include "Domain/Structure/Direction.hpp"
 #include "Utilities/Gsl.hpp"
 
+/// @{
 /*!
- * \brief Compute the Euclidean are element from the inverse jacobian.
+ * \brief Compute the Euclidean area element from the inverse jacobian.
  *
  * \details The Euclidean area element of a surface \f$\Sigma\f$ with
  * constant logical coordinate \f$\xi^i\f$ is given by:
@@ -25,8 +26,8 @@
  * x^j\f$. The determinant of the inverse Jacobian can be passed as an argument
  * as an overload, otherwise it will be calculated.
  *
- * \note The curved space area element as well as time dependent maps are not
- * implemented yet but can be added on demand.
+ * \note Time dependent maps are not implemented yet but can be added on
+ * demand.
  *
  * \param inverse_jacobian_face The inverse Jacobian from the ElementLogical
  * frame to the Target frame sliced to the surface.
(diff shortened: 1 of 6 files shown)


src/Domain/AreaElement.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * \f{equation}
+ * J^\Sigma = J \sqrt{\delta^{jk} (J^{-1})^i_j (J^{-1})^i_k}
+ * \f}
+ *
+ * where \f$J^i_j = \partial x^i / \xi^j\f$ is the volume Jacobian with
+ * determinant \f$J\f$ and inverse \f$(J^{-1})^i_j = \partial \xi^i / \partial
+ * x^j\f$. The determinant of the inverse Jacobian can be passed as an argument
+ * as an overload, otherwise it will be calculated.
+ *
+ * \note The curved space area element as well as time dependent maps are not
```
PunJustice: Maybe I am confused, but isn't `area_element` precisely the curved space area element?

src/Domain/Tags/AreaElement.hpp
```diff
@@ -13,7 +13,8 @@ class DataVector;
 namespace domain::Tags {
 
 /*!
- * \brief The determinant of the induced Jacobian on a surface
+ * \brief The area element i.e. the determinant of the induced Jacobian on a
```
PunJustice: Nitpicking, but I think better is "The area element, i.e. the determinant of the induced Jacobian, on a surface."

PunJustice (commented):
Just small nitpicking things!

src/Domain/AreaElement.hpp
nikwit: yes, well spotted

nilsdeppe:
This needs a rebase. @nilsvu could you also please take a look?

src/Domain/AreaElement.hpp
```diff
@@ -11,9 +11,9 @@
 #include "Utilities/Gsl.hpp"
 
 /*!
- * \brief Compute the Euclidean surface element from the inverse jacobian.
+ * \brief Compute the Euclidean are element from the inverse jacobian.
```
nilsvu: typo `area`

src/Domain/Tags/AreaElement.hpp
```diff
@@ -29,7 +30,7 @@ namespace domain::Tags {
  * metric.
  */
 template <typename SourceFrame, typename TargetFrame>
-struct DetSurfaceJacobian : db::SimpleTag {
+struct AreaElement : db::SimpleTag {
```
nilsvu: tl;dr I suggest this:

- Keep this tag unchanged (no rename, no change in what it stores).
- Include the metric determinant in your `area_element` functions so they actually integrate to the area in curved space.

---

Details: I think in curved space we should make the distinction between the surface Jacobian determinant `J` and the area element `\sqrt{gamma} J`. The area element should actually integrate to the area, hence it should include the metric determinant. Arguably the surface Jacobian alone (without the metric determinant) isn't much use, but you probably don't want to go through the code that uses the `Tags::DetSurfaceJacobian` and make sure it uses the metric determinant properly. Therefore, just keep this tag unchanged. But in your `area_element` functions include the metric determinant, and check in the test that you get the horizon area of Kerr by just integrating over the area element.

src/Domain/AreaElement.hpp
```diff
@@ -63,3 +62,60 @@ Scalar<DataVector> euclidean_area_element(
                           TargetFrame>& inverse_jacobian_face,
     const Direction<VolumeDim>& direction);
 
+/*!
```
nilsvu, line 69: Use `/// @{` to group the dox

tests/Unit/Domain/Test_AreaElement.cpp
```diff
@@ -157,11 +154,81 @@ void test_sphere_integral() {
...
+    gr::Solutions::KerrSchild kerr_schild{mass, spin, {0., 0., 0.}};
+    double horizon_area = 0.;
+    for (const auto& element_id : element_ids) {
+      const auto& current_block = blocks.at(element_id.block_id());
+      const auto element_abutting_direction =
+          excision_sphere.abutting_direction(element_id);
+      if (element_abutting_direction.has_value()) {
+        const auto face_logical_coords = interface_logical_coordinates(
+            face_mesh, element_abutting_direction.value());
+        const ElementMap logical_to_grid_map(
```
nilsvu: You can also use `CoordinateMaps::Composition` to get the combined logical -> grid -> kerr deformed map. Then you don't need to concatenate Jacobians yourself etc. You can get the ElementLogical -> BlockLogical map from `ElementToBlockLogicalMap.hpp`, then compose with `block.stationary_map()` and the Kerr conforming map.

nilsvu (changes requested):

src/Domain/Tags/AreaElement.hpp
nikwit: I don't have a strong opinion but this is then inconsistent with `area_element` in `StrahlKorperGr.cpp` from what I can tell.

nilsvu: In the StrahlkorperGr test here https://github.com/sxs-collaboration/spectre/blob/develop/tests/Unit/ApparentHorizons/Test_StrahlkorperGr.cpp#L417 the integral over the area element is the area, without an additional metric determinant. It follows Appendix D in https://arxiv.org/abs/gr-qc/9606010, which is explained in a bit more detail in Appendix C in the Baumgarte/Shapiro book. That's what we should do for this `area_element` function as well. I think multiplying by the spatial metric determinant is enough.

nikwit: ah right, that makes sense. I changed it so

nilsvu:
Squash 👍

src/Domain/AreaElement.hpp
```diff
@@ -0,0 +1,130 @@
...
+ * as an overload, otherwise it will be calculated.
+ *
+ * \param inverse_jacobian_face The inverse Jacobian from the ElementLogical
+ * frame to the Target frame sliced to the surface.
+ * \param inverse_jacobian_determinant_face The determinant of the inverse
+ * Jacobian sliced onto the surface. If not availabe, there exists an overload
+ * that will calculate it from the inverse Jacobian.
+ * \param direction The direction of the surface in the element.
+ * \param inverse_spatial_metric The inverse spatial metric sliced to the
+ * surface.
```
nilsvu, line 94: Update dox with metric determinant

tests/Unit/Domain/Test_AreaElement.cpp
```diff
@@ -0,0 +1,230 @@
...
+        horizon_area += definite_integral(get(curved_area_element), face_mesh);
+      }
+    }
+    auto custom_approx = Approx::custom().epsilon(1e-8);
+    CHECK(horizon_area == custom_approx(8. * M_PI * mass * horizon_radius));
+  }
+}
+
+SPECTRE_TEST_CASE("Unit.Domain.AreaElement", "[Domain][Unit]") {
+  test_area_element_against_normal_vector();
```
nilsvu, line 218: Why did you remove the grid frame here?

tests/Unit/Domain/Test_AreaElement.cpp
```diff
@@ -0,0 +1,230 @@
...
+        const auto inertial_coords = logical_to_grid_map(face_logical_coords);
+        const auto inv_jacobian =
+            logical_to_grid_map.inv_jacobian(face_logical_coords);
+        const auto inv_jacobian_kerr_horizon =
+            grid_to_kerr_conforming_map.inv_jacobian(inertial_coords);
+        const auto total_inv_jacobian = tenex::evaluate<ti::I, ti::j>(
+            inv_jacobian(ti::I, ti::k) *
+            inv_jacobian_kerr_horizon(ti::K, ti::j));
+        const auto spacetime_vars = kerr_schild.variables(
+            grid_to_kerr_conforming_map(inertial_coords), 0.,
```
nilsvu: I think the different maps and frames are a bit confusing here (e.g. map is named grid but coords are named inertial, then there's another Inertial->Inertial map). Please clear this up a bit, or better yet use a Composition map that combines ElementLogical->BlockLogical->Wedge->HorizonConforming into one map (see my comment above https://github.com/sxs-collaboration/spectre/pull/4743#discussion_r1134797054).

nilsvu (changes requested):

tests/Unit/Domain/Test_AreaElement.cpp
nikwit: the `Minkowski` variables method is hard-coded on the inertial frame. I can template it if you think it's necessary.

nilsvu: no this is fine

tests/Unit/Domain/Test_AreaElement.cpp
nikwit: I am not sure how this is supposed to work. The problem is that the Kerr conforming map effectively goes from the intertial to the inertial frame, so I cannot put it directly into `Composition` because that will not be instantiated. 

So I have to somehow concatenate `block.stationary_map()` and the Kerr conforming map into a single `CoordinateMap` before that goes from `BlockLogical` ->`Inertial`, presumably using `make_coordinate_map`. But `block.stationary_map()` is of type `CoordinateMapBase` and `make_coordinate_map` needs the dynamic type at compile time as far as I can tell, so Im not sure how to do this.

nilsvu: Here's an example:

```cpp
        domain::CoordinateMaps::Composition coord_map{
            domain::element_to_block_logical_map(element_id),
            current_block.stationary_map().get_to_grid_frame(),
            std::make_unique<
                CoordinateMap<Frame::Grid, Frame::Inertial,
                              domain::CoordinateMaps::KerrHorizonConforming>>(
                kerr_schild_map)};
        const auto inertial_coords = coord_map(face_logical_coords);
        const auto inv_jacobian = coord_map.inv_jacobian(face_logical_coords);
```

I tested that this compiles.

nikwit: I see, so we get the map to the grid frame so all the frames are different. Do you think it would be a good idea to put the `Composition` implementation in the header file so it is possible to also have maps that map to the same frame in there? Currently this is not possible because the instantiation would not be there.

This also threw up a bug #4851 which needs to merged before this test can pass.

nilsvu:
👍 please add the sqrt(gamma) also to the math in the doc while you squash

nilsvu (approved):
LGTM, thanks for working this out @nikwit !

PR #4755, opened 2023-02-19
Speed up BBH domain tests

nilsvu:
## Proposed changes

These test have been timing out repeatedly. Instead of testing all possible combinations of options, this PR switches to fuzzy testing, as started in #4697.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Framework/TestHelpers.hpp
+++ b/tests/Unit/Framework/TestHelpers.hpp
@@ -397,6 +397,47 @@ void test_throw_exception(const ThrowingFunctor& func,
 
 /*!
  * \ingroup TestingFrameworkGroup
+ * \brief A random element between `start` and `end`
+ */
+template <typename Iter>
+Iter random_sample(Iter start, Iter end,
+                   const gsl::not_null<std::mt19937*> generator) {
+  const auto length = std::distance(start, end);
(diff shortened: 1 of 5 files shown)


tests/Unit/Framework/TestHelpers.hpp
```diff
@@ -395,6 +395,47 @@ void test_throw_exception(const ThrowingFunctor& func,
                         << __LINE__;                                    \
   }())
 
+/*!
+ * \ingroup TestingFrameworkGroup
+ * \brief A random element between `start` and `end`
+ */
+template <typename Iter>
+Iter random_sample(Iter start, Iter end,
```
nilsdeppe, line 403: This seems pretty generic, maybe move to `Utilities/RandomSample.hpp`?

nilsdeppe (approved):
LGTM, one suggestion. Squash if you agree, or just merge if you disagree!

tests/Unit/Framework/TestHelpers.hpp
nilsvu: I can't think of a use case outside of tests right now, so I suggest we keep it in `tests/` for now

PR #4758, opened 2023-02-20
Support exit codes

nilsvu:
## Proposed changes

Define exit code 2 as "program incomplete, continue from checkpoint". This can be used to resubmit the job (working on that).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Tutorials/CheckpointRestart.md
+++ b/docs/Tutorials/CheckpointRestart.md
@@ -21,6 +21,10 @@ Executables can checkpoint when:
      This reduces the disc space taken up by checkpoint files and stops using
      up the allocation's CPU-hours on work that would be redone anyway after the
      run is restarted.
+     The executable will return exit code 2 when it terminates from
+     `CheckpointAndExitAfterWallclock`, meaning it is incomplete and should
+     continue from the checkpoint. See `Parallel::ExitCode` for a definition of
+     all exit code.
    - using `VisitAndReturn(WriteCheckpoint)`. This is useful for writing more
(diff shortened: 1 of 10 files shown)


src/Parallel/Tags/ExitCode.hpp
```diff
@@ -0,0 +1,29 @@
...
+ * \brief Exit code of the executable
+ *
+ * The following exit codes are defined:
+ *
+ * - 0: Program is complete.
+ * - 1: Program aborted because of an error.
+ * - 2: Program is incomplete and should be continued from the last checkpoint.
+ *
+ * \par Design decisions
+ * - We implement the exit code as an `int` instead of an enum because its
```
nilsdeppe: Why not have the values be hard-coded in an `enum` and then convert them to an `int` where we need them to be an `int`? I worry that the current approach requires us to match something in this documentation block. E.g.
```cpp
enum class ExitCode : int {
  Complete = 0,
  Abort = 1,
  ContinueFromCheckpoint = 2
};
```
and then
```cpp
static_cast<int>(ExitCode::ContinueFromCheckpoint)
```
If you choose to make the `enum` not an `enum class` you can just do an implicit cast.

kidder: I would also prefer using enums in this manner...

nilsvu: ok done!

nilsdeppe (approved):
Nice :D

PR #4759, opened 2023-02-20
Add reduction action for AMR diagnostics

kidder:
## Proposed changes

This allows (among other things):
- Sanity check that the new Domain is fully covered by the new Elements
- Printing out statistics about the new Domain

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Executables/Examples/RandomAmr/InitializeDomain.hpp
+++ b/src/Executables/Examples/RandomAmr/InitializeDomain.hpp
@@ -1,0 +1,66 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <vector>
+
+#include "Domain/Block.hpp"
+#include "Domain/CreateInitialElement.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/Structure/CreateInitialMesh.hpp"
+#include "Domain/Structure/Element.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Domain/Tags.hpp"
+#include "Evolution/DiscontinuousGalerkin/Initialization/QuadratureTag.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace amr::Initialization {
+
+/// \ingroup InitializationGroup
+/// \brief Initialize items related to the structure of a Domain
+///
+/// \see InitializeItems
+///
+/// \note This only initializes the Element and the Mesh on each element of
+/// an array component.  This is all that is needed about the Domain in order to
+/// test the mechanics of adaptive mesh refinement.
+template <size_t Dim>
+struct Domain {
+  using const_global_cache_tags = tmpl::list<::domain::Tags::Domain<Dim>>;
+  using mutable_global_cache_tags = tmpl::list<>;
+  using simple_tags_from_options =
+      tmpl::list<::domain::Tags::InitialExtents<Dim>,
+                 ::domain::Tags::InitialRefinementLevels<Dim>,
+                 evolution::dg::Tags::Quadrature>;
+
+  using argument_tags =
+      tmpl::append<const_global_cache_tags, simple_tags_from_options,
+                   tmpl::list<::Parallel::Tags::ArrayIndex>>;
+
+  using return_tags =
+      tmpl::list<::domain::Tags::Mesh<Dim>, ::domain::Tags::Element<Dim>>;
+
+  using simple_tags = return_tags;
+  using compute_tags = tmpl::list<>;
+
+  /// Given the items fetched from a DataBox by the argument_tags, mutate
+  /// the items in the DataBox corresponding to return_tags
+  static void apply(
+      const gsl::not_null<Mesh<Dim>*> mesh,
+      const gsl::not_null<Element<Dim>*> element, const ::Domain<Dim>& domain,
+      const std::vector<std::array<size_t, Dim>>& initial_extents,
+      const std::vector<std::array<size_t, Dim>>& initial_refinement,
+      const Spectral::Quadrature& quadrature,
+      const ElementId<Dim>& element_id) {
+    const auto& my_block = domain.blocks()[element_id.block_id()];
+    *mesh = ::domain::Initialization::create_initial_mesh(
+        initial_extents, element_id, quadrature);
+    *element = ::domain::Initialization::create_initial_element(
+        element_id, my_block, initial_refinement);
+  }
+};
+}  // namespace amr::Initialization
--- a/src/Executables/Examples/RandomAmr/RandomAmr.hpp
+++ b/src/Executables/Examples/RandomAmr/RandomAmr.hpp
@@ -1,0 +1,92 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <deque>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "Domain/Creators/DomainCreator.hpp"
+#include "Domain/Creators/Factory1D.hpp"
+#include "Domain/Creators/Factory2D.hpp"
+#include "Domain/Creators/Factory3D.hpp"
+#include "Domain/Creators/RegisterDerivedWithCharm.hpp"
+#include "Evolution/DiscontinuousGalerkin/DgElementArray.hpp"
+#include "Executables/Examples/RandomAmr/InitializeDomain.hpp"
+#include "Options/Options.hpp"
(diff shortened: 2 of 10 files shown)


src/ParallelAlgorithms/Amr/Actions/Initialize.hpp
```diff
@@ -14,37 +13,49 @@
...
+/// \brief Initialize items related to adaptive mesh refinement
+///
+/// \details See the type aliases defined below for what items are added to the
+/// GlobalCache, MutableGlobalCache, and DataBox and how they are initialized
 template <size_t Dim>
 struct Initialize {
-  using return_tags =
-      tmpl::list<amr::Tags::Flags<Dim>, amr::Tags::NeighborFlags<Dim>>;
+  /// Tags for constant items added to the GlobalCache.  These items are
+  /// initialized from input file options.
```
nilsvu: We shouldn't (have to) add these explanations to every initialization mutator. How about adding a protocol that defines these aliases?

src/Executables/Examples/RandomAmr/InitializeDomain.hpp
```diff
@@ -0,0 +1,84 @@
...
+/// \details See the type aliases defined below for what items are added to the
+/// GlobalCache, MutableGlobalCache, and DataBox and how they are initialized
+///
+/// \note This only initializes the Element and the Mesh on each element of
+/// an array component.  This is all that is needed about the Domain in order to
+/// test the mechanics of of adaptive mesh refinement.
+template <size_t Dim>
+struct Domain {
+  /// Tags for constant items added to the GlobalCache.  These items are
+  /// initialized from input file options.
```
nilsvu: Same here. This verbosity / boilerplate code seems excessive. We can't document what each of these aliases mean in every initializer. It also distracts from the actual implementation in this file. A protocol would solve this I think.

src/Executables/Examples/RandomAmr/RandomAmr.hpp
```diff
@@ -0,0 +1,93 @@
...
+        tmpl::pair<DomainCreator<volume_dim>, domain_creators<volume_dim>>>;
+  };
+
+  using const_global_cache_tags = tmpl::list<>;
+
+  static constexpr auto default_phase_order =
+      std::array{Parallel::Phase::Initialization, Parallel::Phase::CheckDomain,
+                 Parallel::Phase::Exit};
+
+  /// The parallel components used in the executable
```
nilsvu: wrong comment I think

src/ParallelAlgorithms/Amr/Actions/UseAmrDiagnostics.hpp
```diff
@@ -0,0 +1,71 @@
...
+/// - That the fraction of Block volume (in the logical coordinate frame)
+/// covered by all Element%s is equal to the number of Block%s in the Domain
+///
+/// Prints the following (as integers):
+/// - The number of elements
+/// - The number of grid points
+/// - The average refinement level by logical dimension (i.e. not by the
+///   physical dimensions)
+/// - The average number of grid points by logical dimension
+struct UseAmrDiagnostics {
```
nilsvu: `CheckAmrDiagnostics`? Not sure what "use" means.

src/Executables/Examples/RandomAmr/RandomAmr.hpp
```diff
@@ -0,0 +1,93 @@
+// Distributed under the MIT License.
```
nilsvu, line 1: Should this executable go in `tests/`?

nilsvu (changes requested):

src/ParallelAlgorithms/Amr/Actions/Initialize.hpp
kidder: I have no objection to someone else doing that...

src/ParallelAlgorithms/Amr/Actions/UseAmrDiagnostics.hpp
kidder: That's what I had named it when it just checked the Block volume.  But then I didn't think that was best after I started doing additional things.  And I can easily imagine more things being done as we start to use AMR.  @nilsdeppe @wthrowe thoughts on the name?

src/Executables/Examples/RandomAmr/RandomAmr.hpp
kidder: It could, but I thought it could be a good example of a parallel executable that can easily be run on multiple nodes and cores, and used in a tutorial.   This is not the final state of the executable, it will do full h-refinement.  I just added it now to test the reduction action.

src/ParallelAlgorithms/Amr/Actions/Initialize.hpp
nilsvu: In this PR it would be enough to (1) delete these comments that we shouldn't (have to) duplicate in every initializer, and (2) link to `InitializeItems` in the docs (that's super important, else it's unclear what all these aliases even mean and why they're there). I opened #4761 for the rest.

Not sure what you mean with "no objections to someone else doing that". In my opinion we can't expect others to resolve issues like this. Happy to discuss in a call if you like.

src/ParallelAlgorithms/Amr/Actions/UseAmrDiagnostics.hpp
nilsdeppe: What about `RunAmrDiagnostics`?

src/Executables/Examples/RandomAmr/InitializeDomain.hpp
```diff
@@ -0,0 +1,67 @@
...
+namespace amr::Initialization {
+
+/// \ingroup InitializationGroup
+/// \brief Initialize items related to the structure of a Domain
+///
+/// \see InitializeItems
+///
+/// \note This only initializes the Element and the Mesh on each element of
+/// an array component.  This is all that is needed about the Domain in order to
+/// test the mechanics of of adaptive mesh refinement.
```
nilsvu: delete one "of"

src/Executables/Examples/RandomAmr/InitializeDomain.hpp
```diff
@@ -0,0 +1,67 @@
...
+template <size_t Dim>
+struct Domain {
+  using const_global_cache_tags = tmpl::list<::domain::Tags::Domain<Dim>>;
+  using mutable_global_cache_tags = tmpl::list<>;
+  using simple_tags_from_options =
+      tmpl::list<::domain::Tags::InitialExtents<Dim>,
+                 ::domain::Tags::InitialRefinementLevels<Dim>,
+                 evolution::dg::Tags::Quadrature>;
+
+  using default_initialized_simple_tags = tmpl::list<>;
```
nilsvu: I don't see this being used anywhere, and I don't think it is part of some generic interface (or is it?). This confusion is what I think we must fix (#4761). For the purpose of this PR just delete this line?

src/Executables/Examples/RandomAmr/RandomAmr.hpp
nilsvu: ok, I'm not sure if `Examples/` or `tests/` is the right place. Either seems fine I think. @wthrowe @nilsdeppe any opinions?

src/ParallelAlgorithms/Amr/Actions/Initialize.hpp
```diff
@@ -14,37 +13,31 @@
...
+/// \see InitializeItems
 template <size_t Dim>
 struct Initialize {
-  using return_tags =
-      tmpl::list<amr::Tags::Flags<Dim>, amr::Tags::NeighborFlags<Dim>>;
+  using const_global_cache_tags = tmpl::list<>;
+  using mutable_global_cache_tags = tmpl::list<>;
+  using simple_tags_from_options = tmpl::list<>;
+  using default_initialized_simple_tags =
+      tmpl::list<amr::Tags::NeighborFlags<Dim>>;
```
nilsvu: just inline this in `simple_tags` below? As noted above, I think this isn't part of a generic interface.

nilsvu (changes requested):
you can squash directly

src/ParallelAlgorithms/Amr/Actions/Initialize.hpp
kidder: where do you want it documented that this adds a default initialized tag into the DataBox?

nilsvu: that's part of the `simple_tags` interface, isn't it? That should be documented somewhere. I find the overlap between the `InitializeItems` interface and the `simple_tags` / `compute_tags` interface a bit confusing here, but nothing to fix in this PR.

nilsvu (approved):

PR #4760, opened 2023-02-20
Add 0th order puncture field to CurvedScalarWave

nikwit:
## Proposed changes

Adds the Detweiler-Whiting singular/puncture field expanded up to zeroth order in geodesic distance from a scalar charge.

Update: I vectorized the calculation of the puncture field which makes the calculation about 7x faster than before on a single core on my local machine.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
@@ -1,0 +1,441 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp"
+
+#include "DataStructures/DataBox/Prefixes.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/DynamicBuffer.hpp"
+#include "DataStructures/Tags/TempTensor.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace CurvedScalarWave::Worldtube {
+
+void puncture_field(
+    gsl::not_null<Variables<tmpl::list<
+        Tags::Psi, ::Tags::dt<Tags::Psi>,
+        ::Tags::deriv<Tags::Psi, tmpl::size_t<3>, Frame::Inertial>>>*>
+        result,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& coords, const double time,
+    const double orbital_radius, const double bh_mass, const size_t order) {
+  if (order == 0) {
+    puncture_field_0(result, coords, time, orbital_radius, bh_mass);
+  } else {
+    ERROR(
+        "Puncture order is only implemented up to order 0 but you requested "
+        "order "
+        << order);
+  }
+}
+
+void puncture_field_0(
+    gsl::not_null<Variables<tmpl::list<
+        Tags::Psi, ::Tags::dt<Tags::Psi>,
+        ::Tags::deriv<Tags::Psi, tmpl::size_t<3>, Frame::Inertial>>>*>
+        result,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& coords, const double time,
+    const double orbital_radius, const double BH_mass) {
+  const size_t grid_size = get<0>(coords).size();
+  result->initialize(grid_size);
+  const double r0 = orbital_radius;
+  const double M = BH_mass;
+  const double w = 1. / (r0 * sqrt(r0));
+  const double t = time;
+
+  const double charge_pos_x = r0 * cos(w * time);
+  const double charge_pos_y = r0 * sin(w * time);
+  const double charge_pos_z = 0.;
+
+  const DataVector& x = get<0>(coords);
+  const DataVector& y = get<1>(coords);
+  const DataVector& z = get<2>(coords);
+
+  const DataVector Dx = x - charge_pos_x;
+  const DataVector Dy = y - charge_pos_y;
+  const DataVector Dz = z - charge_pos_z;
+
+  // we use a dynamic buffer even though the size is known at compile
+  // time because TempBuffer only accepts 256 arguments and takes much
+  // longer to compile. The performance loss was measured to be about 10%.
+  DynamicBuffer<DataVector> temps(66, grid_size);
+
+  const double d_0 = 1.0 / r0;
+  const double d_1 = 3.0 * M;
+  const double d_2 = d_1 - r0;
+  const double d_3 = M * sqrt(M);
+  const double d_4 = r0 * r0;
+  const double d_5 = sqrt(-d_2 * d_4);
+  const double d_6 = d_3 * d_5;
+  const double d_7 = d_6 * sqrt(-d_0 * d_2);
+  const double d_8 = 4.0 * d_7;
+  const double d_9 = t * w;
+  const double d_10 = cos(d_9);
+  const double d_11 = sin(d_9);
+  const double d_12 = d_10 * d_11;
+  const double d_13 = d_11 * d_11;
+  const double d_14 = d_10 * d_10;
+  const double d_15 = d_4 * d_4 * sqrt(d_4);
+  const double d_16 = r0 * r0 * r0;
+  const double d_17 = M * M;
+  const double d_18 = r0 * r0 * r0 * r0;
+  const double d_19 = M * M * M;
+  const double d_20 = 6.0 * d_19;
+  const double d_21 = 1.0 / d_4;
+  const double d_22 = d_15 * d_21;
+  const double d_23 = 6.0 * M;
+  const double d_24 = 9.0 * d_17;
+  const double d_25 = d_22 - d_23 * d_4 + d_24 * r0;
+  const double d_26 = 1.0 / d_25;
+  const double d_27 = d_21 * d_26;
+  const double d_28 = -d_2;
+  const double d_29 = sqrt(d_28 * d_4);
+  const double d_30 = d_29 * d_3 * r0 * sqrt(d_0 * d_28);
+  const double d_31 = 4.0 * d_30;
+  const double d_32 = 1.0 / (d_2 * d_2);
+  const double d_33 = d_32 * r0 * 1.0 / d_22;
+  const double d_34 = 1.0 / d_15;
+  const double d_35 = d_16 * d_34;
+  const double d_36 = 1.0 / d_29;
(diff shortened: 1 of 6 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
```diff
@@ -0,0 +1,322 @@
...
+    gsl::not_null<Variables<tmpl::list<
+        Tags::Psi, ::Tags::dt<Tags::Psi>,
+        ::Tags::deriv<Tags::Psi, tmpl::size_t<3>, Frame::Inertial>>>*>
+        evolved_vars,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& coords, const double time,
+    const double orbital_radius, const double BH_mass) {
+  evolved_vars->initialize(get<0>(coords).size());
+  const double r0 = orbital_radius;
+  const double M = BH_mass;
+  const double w = pow(r0, -1.5);
```
nilsdeppe: not a full review, just a quick note. I realize most of this code probably comes from Mathematica, but assuming you'll be calling this quite frequently during the evolution you should go through and do optimizations like `pow(x, 2)` -> `pow<2>(x)`, or `pow(r0, -1.5)->1.0 / (r0 * sqrt(r0))` (pow is insanely expensive, you can do about 180 mult/add for one pow)

Edit: and also things like `10` -> `10.`

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * distance and transformed to Kerr-Schild coordinates.
+ *
+ * The function given here assumes that the particle has scalar charge \f$q=1\f$
+ * and is on a fixed geodesic orbit  with
+ * \f$z(t) = (R \cos{\omega t}, R \sin{\omega t}, 0)\f$ around a Schwarzschild
+ * black hole, where \f$R\f$ is the orbital radius and
+ * \f$ \omega = R^{-3/2} \f$. It returns the singular field at the requested
+ * coordinates as well as its time and spatial derivative.
+ *
+ * \note The expressions were computed with Mathematica and optimized by
```
nilsdeppe, line 41: Are there plans to make the notebook(s) public as part of the paper? We can add them into the spectre repo, not necessarily in this PR. Probably somewhere in `docs` or in the same directory as this file is :shrug:

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
```diff
@@ -0,0 +1,441 @@
...
+  const DataVector Dy = y - charge_pos_y;
+  const DataVector Dz = z - charge_pos_z;
+
+  // we use a dynamic buffer even though the size is known at compile
+  // time because TempBuffer only accepts 256 arguments and takes much
+  // longer to compile. The performance loss was measured to be about 10%.
+  DynamicBuffer<DataVector> temps(66, grid_size);
+
+  const double d_0 = 1.0 / r0;
+  const double d_1 = 3 * M;
```
nilsdeppe: is it easy to regex-replace [space][number][space with [space][number].0[space]? If not, I wouldn't worry about it... This is auto-generated code anyway

tests/Unit/Evolution/Systems/CurvedScalarWave/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ set(LIBRARY_SOURCES
   Test_PsiSquared.cpp
   Test_Tags.cpp
   Test_TimeDerivative.cpp
+  Worldtube/Test_PunctureField.cpp
```
nilsdeppe, line 20: Should/could this get squash into the previous commit?

nilsdeppe (changes requested):
LGTM, couple questions. Squash anything you decide to change :)

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp
wthrowe: Mathematica notebooks are kind of awful to have in a code repo.  We used to have some in SpEC and removed them because (quoting the commit message) "These notebook files are huge and interfere with git grep because they are quasi-binary."  To be more specific, several MB with much of that base64 encoded data.

nilsvu: Will the notebook get published with the paper? Then we can reference the supplementary files of the paper here.

nikwit: Let me discuss that with the Southampton folks. Alternatively I can just put it in a separate public github repo and link to it here. Burden of maintenance would fall on me but it seems unlikely to change.

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
nikwit: it's not that easy unfortunately because the spaces only appear after I `clang-format`. Could probably be done with a bit of work but might not be worth the effort - should be the same binary after all.

tests/Unit/Evolution/Systems/CurvedScalarWave/CMakeLists.txt
nikwit: whoops, meant to add the test in a separate commit. I squashed it all together now.

nikwit:
thank you, I squashed it in :)

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
nikwit: nevermind, fixed it now

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp
```diff
@@ -0,0 +1,60 @@
...
+ * applying common subexpression elimination with sympy. The memory allocations
+ * of temporaries were optimized manually.
+ */
+void compute_puncture_field(
```
nilsvu: (nitpick) delete `compute` from the function names, it's redundant (we don't prefix every function that computes something with `compute`).

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp
```diff
@@ -0,0 +1,60 @@
...
+    gsl::not_null<Variables<tmpl::list<
+        Tags::Psi, ::Tags::dt<Tags::Psi>,
+        ::Tags::deriv<Tags::Psi, tmpl::size_t<3>, Frame::Inertial>>>*>
+        evolved_vars,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& coords, const double time,
+    const double orbital_radius, const double bh_mass, size_t order);
+
+void compute_puncture_field_0(
```
nilsvu: add to dox. Use `/// @{` if you want to copy the docs from the other function.

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
```diff
@@ -0,0 +1,442 @@
...
+        "order "
+        << order);
+  }
+}
+
+void compute_puncture_field_0(
+    gsl::not_null<Variables<tmpl::list<
+        Tags::Psi, ::Tags::dt<Tags::Psi>,
+        ::Tags::deriv<Tags::Psi, tmpl::size_t<3>, Frame::Inertial>>>*>
+        evolved_vars,
```
nilsvu: Question @nilsdeppe: do you have any preference these days on passing Variables vs individual Tensors to functions like this? I often found it easier to work exclusively with Tensors in pointwise functions because then the function is agnostic to where the Tensors are actually stored (could be in separate Variables or in one Variables together with other things, or on a Variables with prefixes tags). Now we have nonowning variables so we could create suitable Variables when calling this function. But overall my experience is that avoiding Variables and Tags in pointwise functions makes everything easier. Thoughts?

nilsvu (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
nilsdeppe: A quick reply now, I can elaborate more if you want. I've started adding Variables-based overloads to GRMHD and GH for things like time derivatives. The motivation there being that 1. there are fewer arguments 2. we can now use structured bindings to easily extract out of a Variables 3. non-owning Variables makes it easier to deal with and 4. the `DataStructures/TaggedContainer.hpp` code makes working with multiple tagged containers simultaneously very easy and so overcomes another hurdle we had with using Variables more.

Basically, I think there are cases in generic code where manipulating a Variables (or set of them) is significantly easier than trying to get the right Tensor from the right tagged container, and so I'm generally in favor of doing so when it make the code a lot easier to read and write. E.g. there were places in GHGRMHD where we would take a bunch of takes from 3 or so tagged containers, put references of them in a TuppleTuple, just so we could pull them out later. This was extremely parameter pack and template dense code that made it hard to read, but the recent additions I listed above make it so that isn't needed anymore.

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp
```diff
@@ -0,0 +1,65 @@
...
+        ::Tags::deriv<Tags::Psi, tmpl::size_t<3>, Frame::Inertial>>>*>
+        result,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& coords, const double time,
+    const double orbital_radius, const double bh_mass, size_t order);
+
+/*!
+ * \brief Computes the puncture/singular field \f$\Psi^\mathcal{P}\f$ of a
+ * scalar charge in circular orbit around a Schwarzschild black hole as
+ * described in \cite Detweiler2003 expanded to order 0 in geodesic distance.
+ */
```
nilsvu, line 57: Add `\see puncture_field` to reference the main function if you get the chance (not holding up this PR)

nilsvu (approved):

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
nilsvu: Sounds reasonable, I'll give the new tools a try. Maybe generic Variables would also be a way to add new evolution variables like recently the electron fraction without refactoring half the code base.

nilsdeppe: That was one of the main motivations for the added functionality.

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp
nilsdeppe: Sounds good, let's not block this PR on making any notebooks public.

nilsdeppe (approved):

PR #4769, opened 2023-02-23
Allow the control system to use NS centers of mass as measurements

ffoucart:
Include test of the center of mass calculation.
The new measurement can be used by the Rotation, Expansion and Translation systems.

## Proposed changes

This PR adds the BothBNSCenters measurements and associated submeasurement, 
to be used to calculate the center of mass of neutron stars in BNS simulations and
drive the BNS control system. In this PR, the measurement is made compatible with
the Expansion, Translation, and Rotation components of the control system.

### Upgrade instructions

Any executable using the Expansion, Translation, or Rotation control system now
needs to template those systems on the Measurement used to get the location
of the compact objects (this is done for tested executables in this PR).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Measurements/BNSCenterOfMass.hpp
+++ b/src/ControlSystem/Measurements/BNSCenterOfMass.hpp
@@ -1,0 +1,179 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <string>
+
+#include "ControlSystem/Component.hpp"
+#include "ControlSystem/Protocols/Measurement.hpp"
+#include "ControlSystem/Protocols/Submeasurement.hpp"
+#include "ControlSystem/RunCallbacks.hpp"
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/LinkedMessageId.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Domain/Structure/ObjectLabel.hpp"
+#include "NumericalAlgorithms/LinearOperators/DefiniteIntegral.hpp"
+#include "Parallel/Reduction.hpp"
+#include "Time/Tags.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/StdArrayHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+template <size_t VolumeDim>
+class ElementId;
+template <size_t Dim>
+class Mesh;
+namespace Parallel {
+template <typename Metavariables>
+class GlobalCache;
+}  // namespace Parallel
+namespace domain::Tags {
+template <size_t Dim>
+struct Mesh;
+template <size_t Dim, typename Frame>
+struct Coordinates;
+}  // namespace domain::Tags
+namespace Events::Tags {
+template <typename InputFrame, typename OutputFrame>
+struct ObserverDetInvJacobian;
+}
+namespace grmhd::ValenciaDivClean {
+struct TildeD;
+}
+/// \endcond
+
+namespace control_system::measurements {
+
+template <typename ControlSystems>
+struct PostReductionSendBNSStarCentersToControlSystem;
+
+namespace Tags {
+/// \ingroup DataBoxTagsGroup
+/// \ingroup ControlSystemGroup
+/// DataBox tag for location of neutron star center
+/// (or more accurately, center of mass of the matter
+/// in the x>0 (label A) or x<0 (label B) region, in grid
+/// coordinates.
+template <::domain::ObjectLabel Center>
+struct NeutronStarCenter : db::SimpleTag {
+  using type = std::array<double, 3>;
+};
+}  // namespace Tags
+
+/// Factored Center of Mass calculation (for easier testing)
+/// This function computes the integral of tildeD (assumed to be the
+/// conservative baryon density in the inertial frame), as well as
+/// its first moment in the grid frame. The integrals are limited to
+/// \f$x>0\f$ (label A) or \f$x<0\f$ (label B).
+/// inv_det_jacobian is the inverse determinant of the jacobian of the
+///                  map between logical and inertial coordinates
+/// xGrid contains the coordinates in the grid frame
+void center_of_mass_integral_on_element(
+    const gsl::not_null<double*> mass_a, const gsl::not_null<double*> mass_b,
+    const gsl::not_null<std::array<double, 3>*> first_moment_A,
+    const gsl::not_null<std::array<double, 3>*> first_moment_B,
+    const Mesh<3>& mesh, const Scalar<DataVector>& inv_det_jacobian,
+    const Scalar<DataVector>& tilde_d,
+    const tnsr::I<DataVector, 3, Frame::Grid>& x_grid);
+
+/// Measurement providing the location of the center of mass of the
+/// matter in the \f$x>0\f$ and \f$x<0\f$ regions (assumed to correspond to the
+/// center of mass of the two neutron stars in a BNS merger).
+struct BothNSCenters : tt::ConformsTo<protocols::Measurement> {
+  struct FindTwoCenters : tt::ConformsTo<protocols::Submeasurement> {
+    /// Unused tag needed to conform to the submeasurement protocol.
+    template <typename ControlSystems>
+    using interpolation_target_tag = void;
+
+    /// Tags for the arguments to the apply function.
+    using argument_tags =
+        tmpl::list<domain::Tags::Mesh<3>,
+                   Events::Tags::ObserverDetInvJacobian<Frame::ElementLogical,
+                                                        Frame::Inertial>,
+                   grmhd::ValenciaDivClean::TildeD,
+                   domain::Tags::Coordinates<3, Frame::Grid>>;
+
+    /// Calculate integrals needed for CoM computation on each element,
+    /// then reduce the data.
+    template <typename Metavariables, typename ParallelComponent,
+              typename ControlSystems>
+    static void apply(const Mesh<3>& mesh,
+                      const Scalar<DataVector>& inv_det_jacobian,
+                      const Scalar<DataVector>& tilde_d,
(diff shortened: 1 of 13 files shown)


ffoucart:
Rebased due to update to CI test names.

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -0,0 +1,201 @@
...
+
+/// Factored Center of Mass calculation (for easier testing)
+/// This function computes the integral of tildeD (assumed to be the
+/// conservative baryon density in the inertial frame), as well as
+/// its first moment in the grid frame. The integrals are limited to
+/// \f$x>0\f$ (label A) or \f$x<0\f$ (label B).
+/// inv_det_jacobian is the inverse determinant of the jacobian of the
+///                  map between logical and inertial coordinates
+/// xGrid contains the coordinates in the grid frame
+static inline void center_of_mass_integral_on_element(
```
wthrowe: Don't declare functions `static`.  In a cpp file, use an anonymous namespace.  In an hpp, just leave it off.

In this case, also move the definition to a cpp file and remove the `inline`.

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -0,0 +1,201 @@
...
+    const gsl::not_null<double*> MassA, const gsl::not_null<double*> MassB,
+    const gsl::not_null<std::array<double, 3>*> FirstMomentA,
+    const gsl::not_null<std::array<double, 3>*> FirstMomentB,
+    const Mesh<3>& mesh, const Scalar<DataVector>& inv_det_jacobian,
+    const Scalar<DataVector>& tildeD,
+    const tnsr::I<DataVector, 3, Frame::Grid> xGrid) {
+  // Get Jacobian and mask for left/right stars
+  const DataVector det_jacobian = 1. / get(inv_det_jacobian);
+  const DataVector PositiveX =
+      (xGrid.get(0) + abs(xGrid.get(0))) / (2.0 * abs(xGrid.get(0)) + 1.e-16);
```
wthrowe: I believe `max(get<0>(xGrid), 0.0)` works.  (And prefer `get<N>(...)` to `.get(N)` if `N` is a constant.)

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -0,0 +1,201 @@
...
+/// Factored Center of Mass calculation (for easier testing)
+/// This function computes the integral of tildeD (assumed to be the
+/// conservative baryon density in the inertial frame), as well as
+/// its first moment in the grid frame. The integrals are limited to
+/// \f$x>0\f$ (label A) or \f$x<0\f$ (label B).
+/// inv_det_jacobian is the inverse determinant of the jacobian of the
+///                  map between logical and inertial coordinates
+/// xGrid contains the coordinates in the grid frame
+static inline void center_of_mass_integral_on_element(
+    const gsl::not_null<double*> MassA, const gsl::not_null<double*> MassB,
```
wthrowe: Snake-case all variables.  (`mass_a`, etc.)

src/ControlSystem/Systems/Translation.hpp
```diff
@@ -87,16 +88,22 @@ struct Translation : tt::ConformsTo<protocols::ControlSystem> {
 
   struct process_measurement {
```
wthrowe, line 89: Maybe not right now, but how hard would factoring out all these "just forward the centers" `process_measurement` structs be?  They all look pretty similar.

tests/Unit/ControlSystem/Test_Measurements.cpp
```diff
@@ -1,15 +1,26 @@
...
 
-#include "Framework/TestingFramework.hpp"
-
+#include "ControlSystem/Measurements/BNSCenterOfMass.hpp"
 #include "ControlSystem/Measurements/BothHorizons.hpp"
 #include "ControlSystem/Measurements/SingleHorizon.hpp"
+#include "DataStructures/DataVector.hpp"
 #include "Domain/Structure/ObjectLabel.hpp"
+#include "Framework/ActionTesting.hpp"
+#include "Framework/TestingFramework.hpp"
```
wthrowe: Keep this one at the top.  Special case for some reason I don't remember anymore.

wthrowe (changes requested):

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
ffoucart: I moved it to a cpp file, but this required changing the status of the library (before, this library did not have any cpp file). The reason I do not hide it in a detail/anonymous namespace is because this function is called by the test. I'm happy to have it either in the cpp or hpp, as people prefer.

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
ffoucart: Done (using step_function instead, but the concept is the same)

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
ffoucart: I think that's done as well.

src/ControlSystem/Systems/Translation.hpp
ffoucart: I think we could by templating on the system (Rotation, Expansion, Translation); probably a conversation to have with @knelli2 as well.

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -0,0 +1,182 @@
...
+    std::array<double, 3> center_a = first_moment_a;
+    std::array<double, 3> center_b = first_moment_b;
+    for (size_t d = 0; d < 3; d++) {
+      center_a[d] = center_a[d] / mass_a;
+      center_b[d] = center_b[d] / mass_b;
+    }
+    const auto center_databox = db::create<
+        db::AddSimpleTags<Tags::NeutronStarCenter<::domain::ObjectLabel::A>,
+                          Tags::NeutronStarCenter<::domain::ObjectLabel::B>>>(
+        center_a, center_b);
```
knelli2, line 172: If you `#include Utilities/StdArrayHelpers.hpp` we have an overload of the `opterator/()` that works with std::arrays. Then you can just put here `first_moment_a / mass_a, first_moment_b / mass_b`

src/ControlSystem/Systems/Translation.hpp
knelli2: Yeah we can probably add it to the `Measurements/` directory and template it on the control system like Francois said. At least for the simple (rot, exp, trans) control systems that is

tests/Unit/ControlSystem/Test_Measurements.cpp
knelli2: Still needs to be moved to the top.

tests/Unit/ControlSystem/Test_Measurements.cpp
```diff
@@ -23,3 +34,136 @@ static_assert(
...
+  MockRuntimeSystem runner{{}};
+  ActionTesting::emplace_singleton_component<control_system_component>(
+      make_not_null(&runner), ActionTesting::NodeId{0},
+      ActionTesting::LocalCoreId{0});
+  ActionTesting::set_phase(make_not_null(&runner), Parallel::Phase::Testing);
+  auto& cache = ActionTesting::cache<control_system_component>(runner, 0);
+
+  LinkedMessageId<double> measurement_id;
+  measurement_id.id = 1.0;
+  measurement_id.previous = 0.0;
```
knelli2: This can be initialized as `LinkedMessageId<double> measurement_id{1.0, 0.0};`

knelli2 (changes requested):
Looks good! Just a couple more small things.

We will want to add this to the SystemHelpers and run full tests with the new measurements. However, this is out of the scope of this PR so @ffoucart and I should talk about it once this is merged.

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
ffoucart: Done

tests/Unit/ControlSystem/Test_Measurements.cpp
ffoucart: Done

tests/Unit/ControlSystem/Test_Measurements.cpp
ffoucart: Not quite sure what happened here; I had moved it but apparently reverted the change later (and it's I think the third time this happens for this specific change... maybe applying clang-format to that file reorganized the includes?)

knelli2 (commented):
Ok I'm happy with the changes. Please wait for @wthrowe to confirm that you can squash in the changes

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -0,0 +1,179 @@
...
+          std::move(reduction_data), my_proxy, reduction_target_proxy);
+    }
+  };
+  /// List of submeasurements used by this measurement -- only FindTwoCenters
+  /// here.
+  using submeasurements = tmpl::list<FindTwoCenters>;
+};
+
+/// Action called after reduction of the center of mass data.
+/// MassA, MassB, FirstMomentA, FirstMomentB will contain the reduced data
```
wthrowe: Update the comment to match the new names.

src/ControlSystem/Systems/Translation.hpp
wthrowe: Right.  Anything that cares about more than the centers will have to implement its own thing.

wthrowe (commented):
Looks good.  Go ahead and squash, including the change to the comment.

wthrowe (approved):

knelli2 (approved):

PR #4772, opened 2023-02-25
Merge Shell into Sphere domain creator

### LINKED ISSUE #4724: Combine Shell and Sphere domain creators
Let's combine the Shell and Sphere domain creators, so it's easier to share features between them (radial partitioning, radial grid point distributions, etc). We can excise the interior of the Sphere like we do in the BBH domain creator.

nilsvu:
## Proposed changes

Combine the Shell and Sphere domain creators so they share features like anisotropic refinement, radial partitioning, radial grid point distributions, block names + groups, time-dependent maps, excluding wedges, equatorial compression, etc.

Closes #4724.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/Creators/Sphere.hpp
```diff
@@ -144,45 +257,50 @@ class Sphere : public DomainCreator<3> {
...
-         const typename InitialGridPoints::type& initial_number_of_grid_points,
-         bool use_equiangular_map, std::vector<double> radial_partitioning = {},
-         std::vector<domain::CoordinateMaps::Distribution> radial_distribution =
-             {domain::CoordinateMaps::Distribution::Linear},
-         std::unique_ptr<domain::creators::time_dependence::TimeDependence<3>>
-             time_dependence = nullptr,
-         std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
-             boundary_condition = nullptr,
-         const Options::Context& context = {});
+      "A 3D cubed sphere. Six wedges surround an inner cube or a spherical "
```
mar-celine: I think it is possible to specify options so that the inner cube is not cubical, or that the spherical excision is not spherical, so maybe we can say something like:
```
A 3D cubed sphere. Six Blocks of the Sphere surround an inner region,
which is either excised or filled in by a seventh Block. This inner
region can be cubical, spherical, or in between, depending on the
value of the `sphericity` parameter. Additional spherical shells, each
composed of six wedges, can be added with the 'RadialPartitioning'
option. The domain can be further modified by the application of an
EquatorialCompression map, which preserves the spherical shape of the
outer boundary, but will distort the shape of the inner excised region
if it is not also spherical.
```

but if we are disabling this option, you can disregard the "`but will distort...`"

src/Domain/Creators/Sphere.hpp
```diff
@@ -222,19 +340,24 @@ class Sphere : public DomainCreator<3> {
  private:
   double inner_radius_{};
   double outer_radius_{};
-  double inner_cube_sphericity_{};
+  std::variant<Excision, InnerCube> interior_{};
+  bool filled_ = false;
```
mar-celine: I would suggest changing `filled_` to `filled_interior_` (or `fill_interior_` to match names like `use_equiangular_map_`) so that it's clear the `bool` refers to a Domain configuration and not some other kind of information, (for example, whether some container is filled)

src/Domain/Creators/Sphere.hpp
```diff
@@ -38,15 +39,77 @@ class CoordinateMap;
...
+  double sphericity = std::numeric_limits<double>::signaling_NaN();
+};
+
+}  // namespace detail
+
+/*!
+ * \brief Create a 3D Domain in the shape of a sphere consisting of six wedges
+ * around a central cube or around a spherical excision.
+ *
+ * \image html WedgeOrientations.png "The orientation of each wedge in Shell."
```
mar-celine: `Shell` -> `Sphere`

src/Domain/Creators/Sphere.cpp
```diff
@@ -95,42 +112,60 @@ Sphere::Sphere(
...
-                "shell because it changes in sphericity. Add entries to "
-                "'RadialPartitioning' to add outer shells for which you can "
-                "select different radial distributions.");
+                "shell filled with a cube because it changes in sphericity. "
+                "Add entries to 'RadialPartitioning' to add outer shells for "
+                "which you can select different radial distributions.");
   }
 
+  // Determine number of blocks
+  const size_t num_blocks_per_shell =
```
mar-celine, line 133: I think we are using `shell` implicitly to mean "A spherical layer of Blocks that extends a single Block in the radial direction." and we should make that clear somewhere (to distinguish this use of shell from just a generic region between two spheres) (I will make another comment somewhere about where this should be mentioned)

Maybe it would be better if we also change `num_blocks_per_shell` and `num_shells` to `num_layers` because here in the current code we have the possibility that the "shell" is not a spherical shell, but we can avoid that by using `num_blocks_per_layer` and `num_layers`

src/Domain/Creators/Sphere.cpp
```diff
@@ -147,83 +182,142 @@ Sphere::Sphere(
...
+          context,
+          "Must specify either both inner and outer boundary conditions "
+          "or neither.");
+    }
+  }
+  if (outer_boundary_condition != nullptr and
+      which_wedges_ != ShellWedges::All) {
+    PARSE_ERROR(
+        context,
+        "Can only apply boundary conditions when using the full shell. "
```
mar-celine: -> `when the outer boundary of the domain is a sphere.`

src/Domain/Creators/Sphere.cpp
```diff
@@ -147,83 +182,142 @@ Sphere::Sphere(
...
-                              inner_radius_ / sqrt(3.0)),
-                  Equiangular(-1.0, 1.0, -1.0 * inner_radius_ / sqrt(3.0),
-                              inner_radius_ / sqrt(3.0)),
-                  Equiangular(-1.0, 1.0, -1.0 * inner_radius_ / sqrt(3.0),
-                              inner_radius_ / sqrt(3.0))}));
+  auto coord_maps = domain::make_vector_coordinate_map_base<Frame::BlockLogical,
+                                                            Frame::Inertial, 3>(
+      sph_wedge_coordinate_maps(
+          inner_radius_, outer_radius_,
+          filled_ ? std::get<InnerCube>(interior_).sphericity : 1.0, 1.0,
```
mar-celine: Does this option mean that if an excised region is used, then the boundary of the excised region must be spherical? I think in practice this will be all we need and you can ignore my comment on 284 of the hpp.
I think enabling non-spherical excisions might simplify the code but this will likely conflict with how the ExcisionSphere object assumes a spherical excision surface so probably not.

src/Domain/Creators/Sphere.hpp
```diff
@@ -38,15 +39,77 @@ class CoordinateMap;
...
+    static std::string name() { return "ExciseWithBoundaryCondition"; }
+    using type = std::unique_ptr<BoundaryConditionsBase>;
+    static constexpr Options::String help = {
+        "The boundary condition to impose on the excision surface."};
+  };
+  struct NoBoundaryCondition {
+    static std::string name() { return "Excise"; }
+    using type = bool;
+    static constexpr Options::String help = {
+        "Specify 'True' to excise the interior of the Sphere. Instead of "
```
mar-celine: going along with your other changes, should this be `Sphere` -> `sphere`?

src/Domain/Creators/Sphere.hpp
```diff
@@ -116,26 +185,70 @@ class Sphere : public DomainCreator<3> {
...
+        Options::Auto<EquatorialCompressionOptions, Options::AutoLabel::None>;
+    static constexpr Options::String help = {
+        "Apply an equatorial compression map to focus resolution on the "
+        "equator or on the poles."};
   };
 
   struct RadialPartitioning {
     using type = std::vector<double>;
     static constexpr Options::String help = {
         "Radial coordinates of the boundaries splitting the spherical shell "
```
mar-celine, line 250: Can change
```
splitting the spherical shell between InnerRadius and OuterRadius.
```

to

```
splitting the spherical region between InnerRadius and OuterRadius into multiple layers of Blocks.
```
(or multiple shells).

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -55,24 +55,34 @@
...
+  if (std::holds_alternative<creators::Sphere::InnerCube>(interior)) {
+    return std::get<creators::Sphere::InnerCube>(interior);
+  } else {
+    return creators::Sphere::Excision{
+        with_boundary_conditions ? create_boundary_condition(false) : nullptr};
+  }
+}
+
+// Calculate block logical coordinates residing on corners of inner cube or on
+// radial block faces of wedges. The radial direction in 3D wedges is zeta.
```
mar-celine: `zeta` -> `the positive zeta direction`

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -55,24 +55,34 @@
...
+                       const bool with_boundary_conditions) {
+  if (std::holds_alternative<creators::Sphere::InnerCube>(interior)) {
+    return std::get<creators::Sphere::InnerCube>(interior);
+  } else {
+    return creators::Sphere::Excision{
+        with_boundary_conditions ? create_boundary_condition(false) : nullptr};
+  }
+}
+
+// Calculate block logical coordinates residing on corners of inner cube or on
```
mar-celine: `block logical coordinates` -> `block logical coordinates of points`
`corners of inner cube` -> `inner cube corners` (or `corners of the inner cube`)

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -55,24 +55,34 @@
...
+// radial block faces of wedges. The radial direction in 3D wedges is zeta.
+// These coordinates will be used to ensure they lie on concentric spheres
```
mar-celine: `they` -> `the points`

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -116,11 +126,11 @@ tnsr::I<double, 3, Frame::BlockLogical> logical_coords(
       {{xi_logical_coord, eta_logical_coord, zeta_logical_coord}}};
 }
 
-template <typename... FuncsOfTime>
 void test_sphere_construction(
     const creators::Sphere& sphere, const double inner_radius,
-    const double outer_radius,
+    const double outer_radius, const bool filled,
```
mar-celine: same comment as above, `filled` -> `fill_interior`

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -133,24 +143,41 @@ void test_sphere_construction(
...
+  } else if (which_wedges == ShellWedges::FourOnEquator) {
+    CHECK(num_external_boundaries ==
+          ((filled ? 2 : 4) + 4 * (1 + num_shells * 2)));
+  } else if (which_wedges == ShellWedges::OneAlongMinusX) {
+    CHECK(num_external_boundaries == ((filled ? 5 : 1) + 1 + num_shells * 4));
+  }
+
   MAKE_GENERATOR(generator);
 
   // verify if adjacent to inner block
```
mar-celine: `block` -> `cube` for consistency, can add a comment saying `test_sphere_construction` is not used on Spheres with inner blocks with `sphericity != 0.0`? (Assuming only Spheres with cubical inner blocks are tested)

tests/Unit/Domain/Test_BlockAndElementLogicalCoordinates.cpp
```diff
@@ -703,7 +703,8 @@ void test_element_ids_are_uniquely_determined() {
 }
 
 void test_block_logical_coordinates_with_roundoff_error() {
-  const auto shell = domain::creators::Shell(1., 3., 0, {{3, 3}});
+  const auto shell = domain::creators::Sphere(
```
mar-celine, line 706: `shell` -> `sphere` and on line 708

tests/Unit/Domain/Test_BlockAndElementLogicalCoordinates.cpp
```diff
@@ -714,7 +715,7 @@ void test_block_logical_coordinates_with_roundoff_error() {
   std::vector<std::array<double, 3>> points{};
   std::vector<size_t> expected_block_ids{};
   // Block piercing z has ID 0, block piercing x has ID 4. Phi=0, Theta=Pi/4 is
-  // on their shared boundary. See also WedgeOrientations.png in the Shell docs.
+  // on their shared boundary. See also WedgeOrientations.png in the Wedge docs.
```
mar-celine: `Wedge` -> `Sphere` (the Wedge docs don't have this image, but you added the png to the Sphere documentation :relaxed:  )

also on line 758

src/Domain/Creators/Sphere.hpp
```diff
@@ -38,15 +39,77 @@ class CoordinateMap;
...
+    static double upper_bound() { return 1.0; }
+  };
+  using options = tmpl::list<Sphericity>;
+  double sphericity = std::numeric_limits<double>::signaling_NaN();
+};
+
+}  // namespace detail
+
+/*!
+ * \brief Create a 3D Domain in the shape of a sphere consisting of six wedges
```
mar-celine: `consisting of spherical layers of Blocks (each layer collectively referred to as a shell) around a central cube`

mar-celine (commented):
Looking pretty good! Just some minor clean up left of renaming, and clearing up some usage of the term "shell"/"layers of blocks"

one question, would this be a good time to rename Sphere to CubedSphere?
that is, if we might eventually have a spherical domain with spherical basis functions that would be more appropriately called Sphere than this one, but otherwise I'm okay with this new domain remaining Sphere.

src/Domain/Creators/Sphere.cpp
nilsvu: yes, I can't think of a case where we want the excision to be non-spherical.

tests/Unit/Domain/Creators/Test_Sphere.cpp
nilsvu: there's a test case with nonzero sphericity as well

tests/Unit/Domain/Test_BlockAndElementLogicalCoordinates.cpp
nilsvu: the interior is excised here so I think keeping the variable name `shell` is fine

src/Domain/Creators/Sphere.hpp
nilsvu: I just copied the option the option help string here

src/Domain/Creators/Sphere.hpp
nilsvu: See comment above "shell" naming above

src/Domain/Creators/Sphere.cpp
nilsvu: I think the name "shell" is pretty intuitive (better than "layer", which is used in the cylinder domain to denote partitioning in z-direction). You're right that "shell" should be defined, so I do that in the docs now. It says each spherical shell is composed if 6 wedges.

src/Domain/Creators/Sphere.hpp
nilsvu: Adjusted the help string. Added the bit about the equatorial compression to the corresponding option. I think we shouldn't explain all options (again) in the main help string, only when it makes sense to give broader context.

nilsvu:
pushed a fixup. To your question: I think keeping the name `Sphere` is good for now, it's very intuitive. We can consider renaming to `CubedSphere` or something like that if and when we need to (maybe we manage to build in spherical harmonic shells into this domain creator, then we don't have to rename anything).

src/Domain/Creators/Sphere.hpp
```diff
@@ -38,15 +39,77 @@ class CoordinateMap;
...
-///
-/// The inner cube is a BulgedCube except if the inner cube sphericity is
-/// exactly 0. Then an Equiangular or Affine map is used (depending on if it's
-/// equiangular or not) to avoid a root find in the BulgedCube map.
+namespace domain::creators {
+
+namespace detail {
+
+/// Options for excising the interior of the sphere
+struct Excision {
```
knelli2: There's a very similar setup in the BCO domain to deal with excisions. Should we try to combine them into a separate file? This PR is already pretty long so maybe not, but up to you.

src/Domain/Creators/Sphere.hpp
```diff
@@ -38,15 +39,77 @@ class CoordinateMap;
...
+struct InnerCube {
+  static constexpr Options::String help = {
+      "Fill the interior of the sphere with a cube."};
+  struct Sphericity {
+    static std::string name() { return "FillWithSphericity"; }
+    using type = double;
+    static constexpr Options::String help = {
+        "Sphericity of the inner cube. A sphericity of 0 uses a product "
+        "of "
+        "1D maps as the map in the center. A sphericity > 0 uses a "
```
knelli2: Formatting?

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -439,10 +466,13 @@ void test_sphere_factory_time_dependent() {
       "Sphere:\n"
       "  InnerRadius: 1\n"
       "  OuterRadius: 3\n"
-      "  InnerCubeSphericity: 0.0\n"
+      "  Interior:\n"
+      "    FillWithSphericity: 0.0\n"
```
knelli2: Maybe add an option construction test with an excised interior now that the shell tests will be removed?

src/Domain/Creators/Sphere.cpp
```diff
@@ -227,7 +227,8 @@ Sphere::Sphere(
       which_wedges_ != ShellWedges::All) {
     PARSE_ERROR(
         context,
-        "Can only apply boundary conditions when using the full shell. "
+        "Can only apply boundary conditions when the outer boundary of the "
+        "domain is a full sphere. "
```
knelli2, line 223: Formatting?

src/Domain/Creators/Sphere.hpp
```diff
@@ -218,7 +223,9 @@ class Sphere : public DomainCreator<3> {
         Options::Auto<EquatorialCompressionOptions, Options::AutoLabel::None>;
     static constexpr Options::String help = {
         "Apply an equatorial compression map to focus resolution on the "
-        "equator or on the poles."};
+        "equator or on the poles. The equatorial compression is an angular "
+        "redistribution of grid points and will preserve the spherical shape "
+        "of the inner and and outer boundaries."};
```
knelli2: One too many `and`s

knelli2 (changes requested):

knelli2:
Conflicts are probably with #4781

src/Domain/Creators/Sphere.hpp
nilsvu: Maybe we can combine them at some point, not sure. Better keep them separate for now I think.

src/Domain/Creators/Sphere.hpp
nilsvu: this is already fixed in the fixup

src/Domain/Creators/Sphere.cpp
nilsvu: I sometimes leave line breaks like this to keep git diffs small when editing these messages. Reflowing the full paragraph creates big diffs and can be annoying, and line breaks avoid that. Just a minor thing, but it helps sometimes.

tests/Unit/Domain/Creators/Test_Sphere.cpp
nilsvu: I prepended a commit to prepare testing time dependencies, and added two commits to test option-creation.

nilsvu:
Rebased, pushed fixups, and added a few commits for better tests.

knelli2 (commented):
Ok looks good. Wait for @gsb76 to squash

mar-celine:
@nilsvu you can squash!

nilsvu:
@sxs-collaboration/spectre-core-devs could one of you sign off on this please?

src/Domain/Creators/Sphere.hpp
```diff
@@ -38,15 +39,82 @@ class CoordinateMap;
...
+    static constexpr Options::String help = {
+        "The boundary condition to impose on the excision surface."};
+  };
+  struct NoBoundaryCondition {
+    static std::string name() { return "Excise"; }
+    using type = bool;
+    static constexpr Options::String help = {
+        "Specify 'True' to excise the interior of the sphere. Instead of "
+        "specifying 'False', replace with the 'FillWithSphericity' option."};
+  };
```
wthrowe: Is this an option with only one valid value?  Why do you need it?

wthrowe (dismissed):

src/Domain/Creators/Sphere.hpp
nilsvu: If boundary conditions are disabled (e.g. in ExportCoordinates), then there are two possibilities: (1) Fill interior with cube of specified sphericity, (2) just excise (no further options). It looks like this in the input file:

```yaml
Interior:
  FillWithSphericity: 0.
# or
Interior:
  Excise: True
```

This is the cleanest approach I've found for the "edge case" when boundary conditions are disabled. It mirrors the case when boundary conditions are enabled, which looks like this:

```yaml
Interior:
  FillWithSphericity: 0.
# or
Interior:
  ExciseWithBoundaryCondition: # ...
```

wthrowe: Maybe use `Auto` with "Excise" as the empty option, or add a custom parse function so the excision class can be created from just the string "Excise"?

nilsvu: I guess I could do the `Auto` approach (the custom parser doesn't work I think because it should be enabled only in the case without boundary conditions). Is it really worth the effort though? Boundary conditions are only disabled in edge cases, and mirroring the case with boundary conditions isn't too bad either I think

wthrowe: I won't approve it as-is, but I'll step back and allow one of the other core-devs to do so if they are happy with it.

nilsvu: I looked at this again and can't find a nice alternative to the current solution. Parsing as `Auto` gets quite complicated as well because it has to go through another constructor etc, and I think adding all that code isn't worth the increase in code complexity for this edge case with no boundary conditions.

nilsdeppe: My concern with all this is that after spending 5 minutes reading through all the help strings I still am not completely sure I know what's going on or what the input file will look like. Is there a way to have the `variant`-based selection give a name:
```
Interior:
  Cube:
    Sphericity: 0.8
```
vs
```
Interior:
  Excise:
    BoundaryCondition:
```
vs
```
Interior:
  Excise:
```

While I agree it's an edge case, if I'm confused, I'm worried others will be too. Since we special case things on whether or not there are options, isn't making this a `basic_option` hindering us from resolving this more cleanly? As in, can we actually template `Interior` on a bool of whether or not we need the boundary condition? Eg
```cpp
template <bool Bc>
  struct Interior {
    struct ExciseInnerCube {};
    using type = tmpl::conditional_t<Bc, std::variant<Excision, InnerCube>, Options::Auto<double, ExciseInnerCube>>;
    static std::string name() { return Bc ? "Interior" : "InnerCubeSphericity"; }
    static constexpr Options::String help = {. // Need swap docs?
        "Sphericity of inner cube. A sphericity of 0 uses a product of 1D maps "
        "as the map in the center. A sphericity > 0 uses a BulgedCube. A "
        "sphericity of exactly 1 is not allowed. See BulgedCube docs for why."};
  };
```
If the documentation is annoying to swap out, we could just specialize on the two cases.

Hopefully this gives a reasonable solution to @wthrowe's concerns that's also not very tedious to implement.

nilsdeppe (commented):
You can squash right away. If there's concerns, let's just jump in a hangout and talk about it to see what is a good solution :)

nilsvu:
I found another solution that doesn't require adding a new constructor (which would be needed with the Options::Auto approach). Please see the fixup @wthrowe @nilsdeppe.

nilsdeppe (commented):
LGTM :)

nilsvu:
@wthrowe please check the fixup when you get the chance

wthrowe:
I stepped down as a reviewer on this PR.

nilsvu:
Well the fixup addresses your comment above.. 🤷‍♂️ I'll just squash then

nilsdeppe (approved):

PR #4774, opened 2023-02-25
Add Worldtube ExpansionOrder tag

nikwit:
## Proposed changes
Adds an option tag that sets the expansion order of the internal worldtube solution.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
@@ -33,6 +33,7 @@ namespace OptionTags {
 struct Worldtube {
   static constexpr Options::String help = {"Options for the Worldtube"};
 };
+
 /*!
  * \brief Name of the excision sphere designated to act as a worldtube
  */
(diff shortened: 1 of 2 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -38,6 +39,17 @@ struct ExcisionSphere {
...
 
+/*!
+ * \brief The internal expansion order of the worldtube solution.
+ */
+struct ExpansionOrder {
+  using type = size_t;
+  static constexpr Options::String help{
+      "The internal expansion order of the worldtube solution. Currently only "
+      "order 0 is implemented"};
+  static size_t upper_bound() { return 0; }
```
nilsvu, line 55: Do you have short-term plans to extend this to higher order? Else you can just hardcode `expansion_order = 0` where you need to instead of going through the trouble of option-creating it.

nikwit: Yes, I will also add orders 1 and 2. But to make reviews easier, I will just add the simplest order 0 scheme first and then add the others on top.

nilsvu (approved):

nikwit:
CI timed out, can you re-trigger?

PR #4771, opened 2023-02-24
Add `CenteredFaceCoordinates` compute tag

nikwit:
## Proposed changes
Adds a convenient compute tag that computes the grid coordinates of each element face abutting the worldtube

PR #4775, opened 2023-02-25
Add `InertialParticlePosition` compute tag

nikwit:
## Proposed changes
Adds a compute tag that calculates the inertial position of the scalar charge particle assumed to be on a circular orbit. For generic orbits, we should probably inject block maps into `ExcisionSphere`.

~for convenience this depends on #4771~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
@@ -7,11 +7,13 @@
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
 
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/ExcisionSphere.hpp"
 #include "Evolution/DiscontinuousGalerkin/ProjectToBoundary.hpp"
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Time/Tags.hpp"
 #include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/Gsl.hpp"
 
@@ -45,5 +47,23 @@ void CenteredFaceCoordinatesCompute<Dim>::function(
   }
 }
 
+template <size_t Dim>
+void InertialParticlePositionCompute<Dim>::function(
+    gsl::not_null<tnsr::I<double, Dim, Frame::Inertial>*> inertial_position,
+    const ::ExcisionSphere<Dim>& excision_sphere, const double time) {
+  const auto& grid_position = excision_sphere.center();
+  const double orbital_radius = get(magnitude(grid_position));
+
+  // assume circular orbit around black hole with mass 1
+  const double angular_velocity = 1. / (sqrt(orbital_radius) * orbital_radius);
+  const double angle = angular_velocity * time;
+  inertial_position->get(0) =
+      cos(angle) * grid_position.get(0) - sin(angle) * grid_position.get(1);
+  inertial_position->get(1) =
+      sin(angle) * grid_position.get(0) + cos(angle) * grid_position.get(1);
+  inertial_position->get(2) = grid_position.get(2);
+}
+
 template struct CenteredFaceCoordinatesCompute<3>;
+template struct InertialParticlePositionCompute<3>;
 }  // namespace CurvedScalarWave::Worldtube::Tags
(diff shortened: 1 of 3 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -68,6 +74,56 @@ struct ExcisionSphere : db::SimpleTag {
   }
 };
 
+/*!
+ * \brief The position of the scalar charge particle orbiting the worldtube
```
nilsvu: orbiting the black hole?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -68,6 +74,56 @@ struct ExcisionSphere : db::SimpleTag {
   }
 };
 
+/*!
```
nilsvu, line 78: Wrap both tags in `/// @{` and `/// @}` so the docs apply to both

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -68,6 +74,56 @@ struct ExcisionSphere : db::SimpleTag {
...
+ * \brief The position of the scalar charge particle orbiting the worldtube
+ * given in inertial coordinates. We currently assume a circular orbit in the
+ * xy-plane with radius \f$R\f$ and angular velocity \f$\omega = R^{-3/2}\f$,
+ * where grid and inertial coordinates are equal at t = 0.
+ */
```
nilsvu, line 87: Please add a brief note why this can't be done with maps at the moment. Something like coordinate maps are currently stored only in blocks, and the worldtube is an excision region outside any block. We could inject the time-dependent grid -> inertial map into the worldtube to fix this.

I think we can currently compute the center of a Strahlkorper and transform a Strahlkorper to another frame, but that seems too complicated.

IMO you could just add an optional time-dependent grid -> inertial map to ExcisionSphere, similar to how it's done in Block.

Edit: To be clear, since my understanding is your scheme relies on these assumptions of a circular orbit etc, I think it's also absolutely fine to keep the current implementation.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -1,5 +1,69 @@
...
+}
+
+template <size_t Dim>
+void InertialParticlePositionCompute<Dim>::function(
+    gsl::not_null<tnsr::I<double, Dim, Frame::Inertial>*> inertial_position,
+    const ::ExcisionSphere<Dim>& excision_sphere, const double time) {
+  const auto& grid_position = excision_sphere.center();
+  const double orbital_radius = get(magnitude(grid_position));
+
+  // assume circular orbit around black hole with mass 1M
```
nilsvu: mass 1

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -1,5 +1,69 @@
...
+  const double angular_velocity = 1. / (sqrt(orbital_radius) * orbital_radius);
```
nilsvu, line 58: You could also get the angular velocity from the rotation function of time if you want. Currently you have to make sure that the angular velocity specified in the input file is the same as this assumption. Maybe add an error somewhere if it isn't, e.g. in an init action?

nilsvu (changes requested):

nilsvu:
Squash directly while you rebase

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: yes, I was thinking we should inject the time dependent map into the excision sphere. Once Shell and Sphere are merged together it can just be the same map that would go into the block that could fill the excision sphere for all domain creators.

I will probably add this once we start looking into scalar back-reaction.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
nikwit: Yes, I will add a check of this, perhaps even in an option tag.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: added another commit that does the same for `CenteredFaceCoordinates`

nilsvu (approved):

PR #4776, opened 2023-02-27
Generalize FaceCenteredCoordinates to Frame and Centered

nikwit:
Generalizes the `FaceCenteredCoordinates` compute tag to Grid and Inertial frames as well as centered and un-centered.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
@@ -3,30 +3,40 @@
 
 #include <cstddef>
 #include <optional>
+#include <type_traits>
 
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
 
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "DataStructures/Tensor/Slice.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/ExcisionSphere.hpp"
-#include "Evolution/DiscontinuousGalerkin/ProjectToBoundary.hpp"
+#include "Domain/Structure/IndexToSliceAt.hpp"
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
 #include "Time/Tags.hpp"
 #include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/Gsl.hpp"
 
 namespace CurvedScalarWave::Worldtube::Tags {
-template <size_t Dim>
-void CenteredFaceCoordinatesCompute<Dim>::function(
-    const gsl::not_null<std::optional<tnsr::I<DataVector, Dim, Frame::Grid>>*>
-        result,
+
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsuggest-attribute=noreturn"
+#endif  // defined(__GNUC__) && !defined(__clang__)
+template <size_t Dim, typename Frame, bool Centered>
+void FaceCoordinatesCompute<Dim, Frame, Centered>::function(
+    const gsl::not_null<std::optional<tnsr::I<DataVector, Dim, Frame>>*> result,
     const ::ExcisionSphere<Dim>& excision_sphere, const Element<Dim>& element,
-    const tnsr::I<DataVector, Dim, Frame::Grid>& grid_coords,
-    const Mesh<Dim>& mesh) {
+    const tnsr::I<DataVector, Dim, Frame>& coords, const Mesh<Dim>& mesh) {
   const auto direction = excision_sphere.abutting_direction(element.id());
   if (direction.has_value()) {
+    ASSERT(
+        mesh.quadrature(direction.value().dimension()) ==
+            Spectral::Quadrature::GaussLobatto,
+        "Expected GaussLobatto quadrature. Other quadratures are disabled "
+        "because interpolating the coordinates incurs an unnecessary error.");
     const size_t grid_size =
         mesh.slice_away(direction->dimension()).number_of_grid_points();
     if (result->has_value()) {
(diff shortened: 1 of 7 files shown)


tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelper.hpp
```diff
@@ -0,0 +1,110 @@
...
+#include "Framework/TestHelpers.hpp"
+#include "Helpers/Domain/BoundaryConditions/BoundaryCondition.hpp"
+#include "Helpers/Domain/Creators/TestHelpers.hpp"
+#include "Helpers/Domain/DomainTestHelpers.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace CurvedScalarWave::Worldtube {
+
+struct TestHelper {
```
nilsvu: make this a namespace? Usually we do something like `TestHelpers::CurvedScalarWave::Worldtube`

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelper.hpp
```diff
@@ -0,0 +1,110 @@
...
+        "      DecayTimescaleOuterBoundaryVelocity: 1.0\n"
+        "    RotationMap:\n"
+        "      InitialAngularVelocity: [0.0, 0.0," +
+        std::to_string(angular_velocity) +
+        "]\n"
+        "    SizeMap:\n"
+        "      InitialValues: [0.0, 0.0]\n"
+        "      InitialVelocities: [0.0, 0.0]\n"
+        "      InitialAccelerations: [0.0, 0.0]";
+    return TestHelpers::test_option_tag<domain::OptionTags::DomainCreator<3>,
```
nilsvu: Just call the constructor instead of going through option parsing?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -18,13 +18,41 @@
...
+      destructive_resize_components(make_not_null(&(result->value())),
+                                    grid_size);
+    } else {
+      result->emplace(grid_size);
+    }
+    evolution::dg::project_tensor_to_boundary(make_not_null(&(result->value())),
+                                              coords, mesh, direction.value());
+    if constexpr (Centered) {
+      for (size_t i = 0; i < Dim; ++i) {
+        result->value().get(i) -= excision_sphere.center().get(i);
```
nilsvu, line 56: This only works for grid coords, so add an `ERROR` in an `if constexpr (Frame is not Grid)` branch.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -18,13 +18,41 @@
...
+    const size_t grid_size =
+        mesh.slice_away(direction->dimension()).number_of_grid_points();
+    if (result->has_value()) {
+      destructive_resize_components(make_not_null(&(result->value())),
+                                    grid_size);
+    } else {
+      result->emplace(grid_size);
+    }
+    evolution::dg::project_tensor_to_boundary(make_not_null(&(result->value())),
+                                              coords, mesh, direction.value());
```
nilsvu: I forgot about this in the previous PR: this carries an interpolation error on Legendre-Gauss meshes. If you want to support LG meshes it would be better to use `interface_logical_coordinates` (from `Domain/InterfaceLogicalCoordinates.hpp`) and use coordinate maps to go grid/inertial coords. If you don't need LG meshes it would be safer to assert the quadrature and use `data_on_slice` from `DataStructures/Tensor/Slice.hpp` (and `index_to_slice_at` from `Domain/Structure/`).

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -36,12 +64,11 @@ void CenteredFaceCoordinatesCompute<Dim>::function(
       result->emplace(grid_size);
     }
     evolution::dg::project_tensor_to_boundary(make_not_null(&(result->value())),
-                                              grid_coords, mesh,
+                                              inertial_coords, mesh,
                                               direction.value());
     for (size_t i = 0; i < Dim; ++i) {
-      result->value().get(i) -= excision_sphere.center().get(i);
+      result->value().get(i) -= particle_position.get(i);
```
nilsvu, line 94: This is missing an `if constexpr (Centered)`, or an error branch that makes sure that this function isn't called for `Centered = false`.

nilsvu (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
nikwit: I don't really need to support Gauss points. But from what I can tell, exactly the same thing is done in `InternalMortarDataImpl.hpp` for both quadratures at each internal boundary every time step so this should not cause any additional (dominant) error. `project_tensor_to_boundary` essentially just mimics `data_on_slice` for GaussLobatto points.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelper.hpp
nikwit: I found this a lot easier to read because you can immediately see exactly what goes into the constructor without looking up all the dozens of arguments at each position if that is fine with you

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
nikwit: gcc complains that this function should be marked no_return for some instantiations. I guess it is true that the struct should not have this overload for some template parameters. I was trying to avoid solving this with `Requires` on the struct itself but can do that if you think SFINAE is the right thing to do here...

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelper.hpp
nilsvu: That C++ doesn't support named arguments is a complaint you have to take up with the C++ committee :D Maybe C++20 will improve things a little with designated initializers.

I guess option creation is fine if you prefer it.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
nilsvu: We have no choice but to project (interpolate) tensor data to the face for Gauss points, but coordinates and Jacobians can be evaluated directly on the face because we know them analytically. Add a comment at least, or disable Gauss points as mentioned above if you don't need them.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
nilsvu: Old GCC versions are quite bad at that, yea. We usually disable the warning with a pragma diagnostic push/pop for gcc, you'll find examples if you search the code.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelpers.hpp
```diff
@@ -0,0 +1,105 @@
...
+#include "Domain/Protocols/Metavariables.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "Helpers/Domain/BoundaryConditions/BoundaryCondition.hpp"
+#include "Helpers/Domain/Creators/TestHelpers.hpp"
+#include "Helpers/Domain/DomainTestHelpers.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace TestHelpers::CurvedScalarWave::Worldtube {
+struct metavariables {
```
nilsvu: CamelCase

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelpers.hpp
```diff
@@ -0,0 +1,105 @@
...
+      "    RadialDistribution: Linear\n"
+      "    BoundaryCondition:\n"
+      "      None\n"
+      "  InitialRefinement:\n"
+      "    ObjectAShell: [0, 0, 0]\n"
+      "    ObjectBShell: [0, 0, 0]\n"
+      "    ObjectACube: [0, 0, 0]\n"
+      "    ObjectBCube: [0, 0, 0]\n"
+      "    EnvelopingCube: [0, 0, 0]\n"
+      "    OuterShell: [0, 0, 0]\n"
```
nilsvu: Just `InitialRefinement: 0`

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_Tags.cpp
```diff
@@ -33,7 +33,7 @@
 #include "Helpers/Domain/DomainTestHelpers.hpp"
 #include "NumericalAlgorithms/Spectral/LogicalCoordinates.hpp"
 #include "ParallelAlgorithms/Initialization/MutateAssign.hpp"
-#include "TestHelper.hpp"
+#include "TestHelpers.hpp"
```
nilsvu: Full relative path please

nilsvu (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
nikwit: ok, I have disabled Gauss now

nilsvu (dismissed):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -116,31 +117,46 @@ struct InertialParticlePositionCompute : InertialParticlePosition<Dim>,
...
 /*!
- * \brief An optional that holds the grid coordinates, centered on the
- * worldtube, of an element face abutting the worldtube excision sphere. If the
- * element does not abut the worldtube, this holds std::nullopt. This tag should
- * be in the databox of element chares.
+ * \brief An optional that holds the coordinates of an element face abutting the
+ * worldtube excision sphere. If the element does not abut the worldtube, this
+ * holds std::nullopt. This tag should be in the databox of element chares. The
+ * available frames are Grid and Inertial. The Centered template tag can be
+ * turned on to centere the coordinates around the position of the scalar
```
nilsvu: typo "center"

nilsvu (changes requested):
To fix the macOS error I think just set the test helper lib to static, like in the commit that I sent you.

nikwit:
rebased and squashed, thank you for the review!

nilsvu (approved):

PR #4780, opened 2023-02-27
Add PunctureField compute tag

nikwit:
Adds a tag that computes the puncture field if the element abuts the worldtube.

~depends on #4776~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
@@ -11,9 +11,11 @@
 #include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
 #include "DataStructures/Tensor/Slice.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
 #include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/ExcisionSphere.hpp"
 #include "Domain/Structure/IndexToSliceAt.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp"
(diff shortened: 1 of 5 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -101,6 +103,25 @@ void FaceCoordinatesCompute<Dim, Frame, Centered>::function(
...
+        inertial_face_coords,
+    const ::ExcisionSphere<Dim>& excision_sphere, const double time,
+    const size_t expansion_order) {
+  if (inertial_face_coords.has_value()) {
+    if (not result->has_value()) {
+      result->emplace(get<0>(inertial_face_coords.value()).size());
+    }
+    puncture_field(
+        make_not_null(&(result->value())), inertial_face_coords.value(), time,
+        get(magnitude(excision_sphere.center())), 1., expansion_order);
```
nilsvu, line 119: Do you always assume the BH has mass 1 and also that the scalar charge is 1? Both should probably be quite prominently documented somewhere (maybe they already are). Or maybe put options for them in the input file and in places where you (have to) assume both are 1 you assert? If you want to relax this assumption you'd have to go through all of your code and find the places where you hardcoded literal ones, that can be quite annoying and hard to debug. Up to you.

nilsvu (approved):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
nikwit: Yes, we had always assumed so until now, I never tried anything else. The charge is definitely 1, that is hard-coded in the puncture field. For the mass, I will probably add an OptionTag once I made sure it works for different masses.

PR #4781, opened 2023-02-27
Use shape map in AH finder test

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/ApparentHorizons/Test_ApparentHorizonFinder.cpp
+++ b/tests/Unit/ApparentHorizons/Test_ApparentHorizonFinder.cpp
@@ -28,11 +28,14 @@
 #include "Domain/Creators/Tags/Domain.hpp"
 #include "Domain/Creators/Tags/FunctionsOfTime.hpp"
 #include "Domain/Creators/TimeDependence/RegisterDerivedWithCharm.hpp"
+#include "Domain/Creators/TimeDependence/Shape.hpp"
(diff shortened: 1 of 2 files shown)


tests/Unit/ApparentHorizons/Test_ApparentHorizonFinder.cpp
```diff
@@ -327,16 +330,30 @@ void test_apparent_horizon(const gsl::not_null<size_t*> test_horizon_called,
     std::vector<domain::CoordinateMaps::Distribution> radial_distribution{
         domain::CoordinateMaps::Distribution::Linear};
 
+    std::unique_ptr<domain::creators::time_dependence::TimeDependence<3>>
+        time_dependence;
+
+    if constexpr (std::is_same_v<Frame, ::Frame::Distorted>) {
```
markscheel: Maybe do 2 tests: 1 with UniformTranslation and 1 with Shape?   The Shape map has an Identity map for DistortedToInertial, so DistortedToInertial isn't really tested when the map is Shape.  Also, the Schwarzschild test has mass=1,spin=0, so the Shape map is an identity for that test.

knelli2: Ok tested both cases now

markscheel (approved):

nilsdeppe (approved):

PR #4782, opened 2023-02-27
Add db::creation_tag and db::tag_depends_on metafunctions

wthrowe:
Also includes two cleanup commits related to DataBox to avoid recompilations for a second PR touching DataBox.hpp.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/DataBox/DataBox.hpp
+++ b/src/DataStructures/DataBox/DataBox.hpp
@@ -290,6 +290,11 @@ class DataBox<tmpl::list<Tags...>> : private detail::Item<Tags>... {
   template <typename Tag>
   auto& get_mutable_reference();
 
+  /// Check whether a tags depends on another tag.  Should be called
+  /// through the metafunction db::tag_depends_on.
+  template <typename Consumer, typename Provider>
+  constexpr static bool tag_depends_on();
+
   // NOLINTNEXTLINE(google-runtime-references)
   void pup(PUP::er& p) {
     // We do not send subitems for both simple items and compute items since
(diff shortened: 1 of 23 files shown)


src/DataStructures/DataBox/DataBox.hpp
```diff
@@ -65,6 +65,38 @@ constexpr bool tag_is_retrievable_v =
...
+};
+}  // namespace detail
+
+/*!
+ * \ingroup DataBoxGroup
+ * \brief The tag added to \p Box referred to by \p Tag.  This
+ * resolves base tags and converts subitems to full items.
+ */
+template <typename Tag, typename Box>
+using real_tag =
```
kidder: If I understand what this is doing, in the type aliases in DataBox I have been calling the tags returned by this metafunction `creation_tags`.  If so can we use that nomenclature?

wthrowe:
Not sure what the best way to handle the build failures is.  It seems that in this context `DEBUG_STATIC_ASSERT` is required to be followed by a semicolon in Debug mode and forbidden from being followed by a semicolon in Release mode.

src/DataStructures/DataBox/DataBox.hpp
```diff
@@ -934,6 +988,26 @@ SPECTRE_ALWAYS_INLINE auto& get_mutable_reference(
...
+ * \brief Check whether the tag \p Consumer depends on the tag \p
+ * Provider in a given \p Box.
+ *
+ * This is equivalent to the question of whether changing \p Provider
+ * (through `db::mutate`, updating one of its dependencies, etc.)
+ * could change the value of `db::get<Consumer>`.  Tags depend on
+ * themselves, and an item and its subitems all depend on one another.
+ */
+template <typename Provider, typename Consumer, typename Box>
+using tag_depends_on =
```
kidder, line 1004: Not a change request, but I will probably get this backwards when I use/read it without looking up the documentation, as I was expecting the first template argument to be the Consumer tag that depends upon the Provider.

kidder:
> Not sure what the best way to handle the build failures is. It seems that in this context `DEBUG_STATIC_ASSERT` is required to be followed by a semicolon in Debug mode and forbidden from being followed by a semicolon in Release mode.

Add a similar code block as we do for the false branch of ASSERT?

src/DataStructures/DataBox/DataBox.hpp
wthrowe: Yes, those appear to be the same thing.  I will adopt that name.

src/DataStructures/DataBox/DataBox.hpp
wthrowe: I copied the order from `tt::is_a<Template, Type>`, but `std::is_base_of<Base, Derived>` is the opposite, so yeah, not obvious which is better.

wthrowe:
Right.  Of course.  Added a commit for that (at the beginning).

Did the rename without a fixup, because it was a find-replace rename.  Waiting to see if anyone else wants to weigh in on argument ordering.

kidder:
okay sounds good

src/DataStructures/DataBox/DataBox.hpp
nilsdeppe: The choice on `is_a` is the same as `is_base_of` in my opinion. It is `<generic, specific>`. That is, the generic template in `is_a` plays a similar role to the generic base class in `is_base_of`. That's why the ordering is that way :shrug:

wthrowe: The English is backward, though: "X is base of Y" and "Y is a X".

nilsdeppe: That's fair. I have no preference for what you want to do in this PR :) Just wanted to give context as to why I made `is_a` the way it is!

kidder: If neither of you have a preference, I would prefer to reverse the order of the template parameters

nilsdeppe: That's fine by me :)

kidder (approved):

PR #4784, opened 2023-02-27
Use CHECK_ITERABLE_APPROX in TE Product test

macedo22:
## Proposed changes

This PR updates the `TensorExpressions` products test to use `CHECK_ITERABLE_APPROX` instead of `CHECK` for any results that used floating point math.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/DataStructures/Tensor/Expressions/Test_Product.cpp
+++ b/tests/Unit/DataStructures/Tensor/Expressions/Test_Product.cpp
@@ -105,8 +105,8 @@ void test_outer_product_double(const gsl::not_null<Generator*> generator,
 
   for (size_t i = 0; i < dim; i++) {
     for (size_t j = 0; j < dim; j++) {
-      CHECK(Lij_from_R_Sij.get(i, j) == 5.6 * S.get(i, j));
-      CHECK(Lij_from_Sij_R.get(i, j) == S.get(i, j) * -8.1);
+      CHECK_ITERABLE_APPROX(Lij_from_R_Sij.get(i, j), 5.6 * S.get(i, j));
+      CHECK_ITERABLE_APPROX(Lij_from_Sij_R.get(i, j), S.get(i, j) * -8.1);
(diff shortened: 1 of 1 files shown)


macedo22:
I pushed these changes because they were requested but I will say that I don't think that any of the changes made should actually be approximate checks, and I think they should be passing with exact `CHECK`s because all of the equations in question are just an an outer product of two numbers, so all of these `CHECK`s that were changed to `CHECK_ITERABLE_APPROX` are just checking `A * B == A * B`, which I think should not require approximation, but maybe there is some low-level detail I don't know about!

Something interesting to note is that all of Guillermo's test failures were for the `std::complex<double>` test cases - an outer product of two `Tensor<std::complex<double>, ...>`s or an outer product of two `Tensor<ComplexDataVector, ...>`. I haven't yet been able to reproduce the errors (even using the same seed), so I can't yet dig into how far off the failing test values were. Even if `A * B` can be different than `B * A`, none of @guilara 's failing tests had reordered operands in the TE syntax tree, so not yet sure what's going on.

wthrowe:
Trouble with complex numbers is much less surprising than doubles would have been.  It's much harder to do efficient and accurate complex multiplication because of the potential roundoff trouble from the additions in both parts of `(a + b i) (c + d i) = (a b - b d) + (b c + a d) i`.  If there's vectorization or something involved I would expect a simple `double * double` to be unaffected, but one version might be more careful to minimize errors in the complex-number case.

macedo22:
I see, so it seems that there *are* reasonable things that could be happening here. @wthrowe Given what you've said, is it worth spending time to look into it and understand?

Edit: In light of your comment and comments on Slack, seems like this is reasonable, so I will also fix the remaining floating point math CHECKs in the file, push, and convert to a PR.

guilara:
Thank you @macedo22, I'll check if they now pass in my machine.

guilara:
Good news. These tests seem to pass now. Thanks!

wthrowe (approved):
Yes, I wouldn't worry about this.

PR #4789, opened 2023-02-28
Add CMake flag to enable debug checks in release

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupCxxFlags.cmake
+++ b/cmake/SetupCxxFlags.cmake
@@ -5,6 +5,18 @@ option(DEBUG_SYMBOLS "Add -g to CMAKE_CXX_FLAGS if ON, -g0 if OFF." ON)
 
 option(OVERRIDE_ARCH "The architecture to use. Default is native." OFF)
 
+option(ENABLE_SPECTRE_DEBUG "Enable ASSERTs and other SPECTRE_DEBUG options"
+  OFF)
+
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(ENABLE_SPECTRE_DEBUG ON)
+endif()
+
+if(${ENABLE_SPECTRE_DEBUG})
+  set_property(TARGET SpectreFlags
+    APPEND PROPERTY INTERFACE_COMPILE_DEFINITIONS SPECTRE_DEBUG)
+endif()
+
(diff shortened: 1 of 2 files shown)


cmake/SetupCxxFlags.cmake
```diff
@@ -5,6 +5,18 @@ option(DEBUG_SYMBOLS "Add -g to CMAKE_CXX_FLAGS if ON, -g0 if OFF." ON)
...
+option(ENABLE_SPECTRE_DEBUG "Enable ASSERTs and other SPECTRE_DEBUG options"
+  OFF)
+
+if (CMAKE_BUILD_TYPE STREQUAL "Debug")
+  set(ENABLE_SPECTRE_DEBUG ON)
+endif()
+
+if(${ENABLE_SPECTRE_DEBUG})
+  set_property(TARGET SpectreFlags
+    APPEND PROPERTY INTERFACE_COMPILE_DEFINITIONS SPECTRE_DEBUG)
```
nilsvu, line 17: Where was this set before?

nilsdeppe: This file, about 10 lines down

nilsvu: Delete it there then?

nilsdeppe: Sure, I didn't want to risk breaking anything but it presumably is safe. I expect all our code inherits from SpectreFlags :shrug: Let me know what you prefer :)

nilsvu: Either is fine with me, just note that currently the compiler def is added in debug builds even if the new option is explicitly turned off. Either make that clear in the docs and the option help string, or make sure tge compiler def is added if and only if the option is turned on.

nilsdeppe: Sorry, I'm confused, you can't disable the macro in debug mode, even if you set the option to `OFF`. The if on line 11 always enables this in debug mode

nilsvu: Exactly, so if I pass `ENABLE_SPECTRE_DEBUG=OFF` to CMake in Debug mode it will be ignored. Just make that clear in the docs.

nilsdeppe:
Okay, rebased and expanded the markdown documentation to clarify this only affects release builds.

nilsvu (approved):

Issue #4764, opened 2023-02-22
Add shape map to BBH domains


nilsvu:
Task list:

- [x] Add distorted frames to BBH domains, not changing any maps
- [ ] Add shape map -- PRs for single and binary domains

Both the cubical and cylindrical BBH domains should be updated in parallel.

geoffrey4444:
3/23/23: @nilsvu has volunteered to review this and dependent PRs, and @gsb76 will help as needed with reviews

PR #4792, opened 2023-02-28
Add distorted frame to BBH domains

knelli2:
## Proposed changes

Also adds the size map to the CylBBH domain so it has the same time dependent maps as the regular BBH domains.

Part of #4764.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/CylindricalBinaryCompactObject.cpp
+++ b/src/Domain/Creators/CylindricalBinaryCompactObject.cpp
@@ -19,6 +19,7 @@
 #include "Domain/CoordinateMaps/TimeDependent/ProductMaps.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/ProductMaps.tpp"
 #include "Domain/CoordinateMaps/TimeDependent/Rotation.hpp"
+#include "Domain/CoordinateMaps/TimeDependent/SphericalCompression.hpp"
 #include "Domain/CoordinateMaps/UniformCylindricalEndcap.hpp"
 #include "Domain/CoordinateMaps/UniformCylindricalFlatEndcap.hpp"
 #include "Domain/CoordinateMaps/UniformCylindricalSide.hpp"
@@ -365,6 +366,8 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
 CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
     double initial_time, ExpansionMapOptions expansion_map_options,
     std::array<double, 3> initial_angular_velocity,
+    std::array<double, 3> initial_size_map_values_A,
+    std::array<double, 3> initial_size_map_values_B,
     std::array<double, 3> center_A, std::array<double, 3> center_B,
     double radius_A, double radius_B, bool include_inner_sphere_A,
     bool include_inner_sphere_B, bool include_outer_sphere, double outer_radius,
@@ -397,6 +400,8 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
   initial_time_ = initial_time;
   expansion_map_options_ = expansion_map_options;
   initial_angular_velocity_ = initial_angular_velocity;
+  initial_size_map_values_[0] = initial_size_map_values_A;
+  initial_size_map_values_[1] = initial_size_map_values_B;
 }
 
 Domain<3> CylindricalBinaryCompactObject::create_domain() const {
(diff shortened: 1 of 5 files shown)


nilsvu:
The cyl BBH domain test timed out. Can we reduce its complexity somewhat like we have done recently with the Interval creator (just merged) and the Sphere (in #4772)?

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -382,10 +385,18 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
           use_equiangular_map, initial_refinement, initial_grid_points,
           std::move(inner_boundary_condition),
           std::move(outer_boundary_condition), context) {
+  if (not(include_inner_sphere_A and include_inner_sphere_B)) {
+    PARSE_ERROR(context,
```
nilsvu, line 394: Could you add a comment here why this limitation exists? If this is "not implemented yet" then mention that in the assert.

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -416,22 +417,32 @@ std::string create_option_string(
...
-         "  IncludeInnerSphereA: False\n"
-         "  IncludeInnerSphereB: False\n"
+         "  UseEquiangularMap: false\n" +
+         (add_time_dependence ? std::string("  IncludeInnerSphereA: True\n"
+                                            "  IncludeInnerSphereB: True\n")
+                              : std::string("  IncludeInnerSphereA: False\n"
+                                            "  IncludeInnerSphereB: False\n")) +
          "  IncludeOuterSphere: False\n"
          "  InitialRefinement:" +
          (with_additional_outer_radial_refinement
```
nilsvu, line 397: I think the true and false branches of this conditional and the one below are reversed. You didn't introduce this here, but would you mind fixing it?

knelli2:
@nilsvu Do you have any ideas in mind? I'm not sure which parts would be unnecessary.

nilsvu:
What comes to mind is:

- Move parse error tests out of the cartesian product into `test_parse_errors` so they don't get called unnecessarily.
- Merge `test_binary_factory`, `test_bbh_time_dependent_factory`, and `test_connectivity` into one function so there's only one cartesian product. Will be easy with the tools added in #4772.

src/Domain/Creators/CylindricalBinaryCompactObject.hpp
```diff
@@ -140,44 +146,79 @@ namespace domain::creators {
...
+      domain::CoordinateMap<
+          Frame::BlockLogical, Frame::Inertial,
+          CoordinateMaps::ProductOf2Maps<CoordinateMaps::Wedge<2>,
+                                         CoordinateMaps::Interval>,
+          CoordinateMaps::UniformCylindricalSide,
+          CoordinateMaps::DiscreteRotation<3>>,
+      domain::CoordinateMap<
+          Frame::Grid, Frame::Inertial,
+          domain::CoordinateMaps::TimeDependent::CubicScale<3>,
+          domain::CoordinateMaps::TimeDependent::Rotation<3>>,
```
nilsvu: Use the aliases you defined above here? Same in the cubical BBH domain.

src/Domain/Creators/CylindricalBinaryCompactObject.hpp
```diff
@@ -337,13 +377,31 @@ class CylindricalBinaryCompactObject : public DomainCreator<3> {
...
+  struct SizeMap {
+    static std::string name() { return "SizeMap" + get_output(Object); }
+    static constexpr Options::String help = {
+        "Options for a time-dependent size map about the specified object."};
+    using group = TimeDependentMaps;
+  };
+
+  template <domain::ObjectLabel Object>
+  struct SizeMapInitialData {
+    static std::string name() { return "InitialData"; }
```
nilsvu: This is "InitialValues" in the cubical BBH domain. Make the two consistent. I think "InitialValues" is better than "InitialData" (neither name tells me what sort of values to put here, but that's out of scope for this PR). I understand that the cubical BBH domain has 3 separate tags for the value and two derivatives, but "InitialData" too strongly implies PDE volume initial data (maybe I'm biased because I've been working on that for a long time).

nilsvu (changes requested):

src/Domain/Creators/CylindricalBinaryCompactObject.hpp
knelli2: I was matching the name of the Expansion map options here. I can change both to be "InitialValues"

nilsvu: Oh I see. Yes I think that would be better.

knelli2:
I'll refactor the test more once #4772 is merged

nilsvu (approved):

nilsvu:
@markscheel do you want to take a quick look before I merge this?

markscheel (commented):
Looks good!   Eventually the size/shape (and cutX) maps will need to extend outside of the inner shells (and then the inner shells don't need to exist) but that is complicated and should be done later.

markscheel (approved):

PR #4794, opened 2023-03-01
Add and use stringize function

knelli2:
## Proposed changes

Turns a bool into "true" or "false" string

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Utilities/CMakeLists.txt
+++ b/src/Utilities/CMakeLists.txt
@@ -15,6 +15,7 @@ spectre_target_sources(
   OptimizerHacks.cpp
   PrettyType.cpp
   Rational.cpp
+  Stringize.cpp
   WrapText.cpp
   )
 
@@ -68,6 +69,7 @@ spectre_target_headers(
   StdHelpers.hpp
   StlBoilerplate.hpp
   StlStreamDeclarations.hpp
+  Stringize.hpp
   TMPL.hpp
   TaggedTuple.hpp
   TmplDebugging.hpp
--- a/src/Utilities/Stringize.cpp
+++ b/src/Utilities/Stringize.cpp
@@ -1,0 +1,13 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Utilities/Stringize.hpp"
+
+#include <ios>
+#include <string>
+
+#include "Utilities/MakeString.hpp"
+
+std::string stringize(const bool t) {
+  return MakeString{} << std::boolalpha << t;
+}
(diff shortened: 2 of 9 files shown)


src/Utilities/GetOutput.hpp
```diff
@@ -5,14 +5,19 @@
...
  * \ingroup UtilitiesGroup
  * \brief Get the streamed output of `t` as a `std::string`
  */
 template <typename T>
 std::string get_output(const T& t) {
-  std::ostringstream os;
-  os << t;
-  return os.str();
+  if constexpr (std::is_same_v<T, bool>) {
+    return t ? "true" : "false";
```
wthrowe: The thing you're looking for is called `std::boolalpha`.

wthrowe (changes requested):
I don't think this is a good idea.  `get_output` is used in many places for testing that objects print correctly, or for finding the result of printing the object for some other reason.  Making it return something other than the plain result of streaming breaks that.

src/Utilities/GetOutput.hpp
knelli2: Should I use `std::boolalpha` here then in the stream part, or just make a new `stringize` function in Utilities?

nilsdeppe:
I agree with @wthrowe This could lead to confusion. I'm not sure what the right answer is because I also agree that being able to control the output would be nice. Possible `get_output` should take a stream modifier object? I haven't looked into that, so it might just not be possible

wthrowe:
Sorry for the confusion.  The two parts of my comment were unrelated.  I don't think we should do this, but if we do, or add something similar, we should use the standard library functionality.

As to what to do, I think `(MakeString{} << std::boolalpha << flag)` would give the right result, and is probably simple enough to not need a wrapper function.

knelli2:
Ok @wthrowe I left `get_output` alone and just added a global `stringize` function. I think it's worth having this because now all places that need to convert a bool to a string use the standard library functionality. Whereas previously none of them did

wthrowe:
I think "stringize" is a really generic name for something that only works on bool.  I suggest one of (a) rename it to make it's purpose clear, (b) allow it to take any type so it works on `vector<bool>` and similar (although then people will probably confuse it with `get_output`), or (c) inline it everywhere.

I think this is mostly used for creating option strings for tests, so a function to convert simple types to something the YAML parser will understand might be a general solution.  The yaml-cpp library probably has something to do that, but if we only care about a few things like bools and vectors writing our own is probably easier.

nilsvu:
Yea generating YAML output from types would be the way to go I think

kidder:
I agree with this, I was going to suggest something like `Options::to_yaml`

wthrowe:
As long as we don't want to do anything complicated so we care about indenting and stuff it looks like yaml-cpp makes that pretty easy: https://github.com/jbeder/yaml-cpp/wiki/How-To-Emit-YAML

PR #4802, opened 2023-03-03
Add integrands for BNS observers

ffoucart:
## Proposed changes

Add a few db::ComputeTag that can be used as integrand to diagnose BNS simulations.

NOTE: Because the TildeD tag is currently hard-coded to be a Scalar<DataVector> in Frame::Inertial, it could make sense to avoid using templates. On the other hand, we might want to use TildeD in a different frame in the future, or test calculations on doubles instead of DataVectors, in which case having the templates would be useful. I'm happy to remove the templates if that's the preferred choice.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/Hydro/MassWeightedFluidItems.cpp
+++ b/src/PointwiseFunctions/Hydro/MassWeightedFluidItems.cpp
@@ -1,0 +1,83 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "MassWeightedFluidItems.hpp"
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/DotProduct.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace hydro {
+
+template <typename DataType, size_t Dim, typename Frame>
+void u_lower_t(const gsl::not_null<Scalar<DataType>*> result,
+               const Scalar<DataType>& lorentz_factor,
+               const tnsr::i<DataType, Dim, Frame>& spatial_velocity_one_form,
+               const Scalar<DataType>& lapse,
+               const tnsr::I<DataType, Dim, Frame>& shift) {
+  dot_product(result, spatial_velocity_one_form, shift);
+  result->get() = get(lorentz_factor) * (get(lapse) * (-1.0) + result->get());
+}
+
+template <typename DataType>
+void mass_weighted_internal_energy(
+    const gsl::not_null<Scalar<DataType>*> result,
+    const Scalar<DataType>& tilde_d,
+    const Scalar<DataType>& specific_internal_energy) {
+  result->get() = get(tilde_d) * get(specific_internal_energy);
+}
+
+template <typename DataType>
+void mass_weighted_kinetic_energy(const gsl::not_null<Scalar<DataType>*> result,
+                                  const Scalar<DataType>& tilde_d,
+                                  const Scalar<DataType>& lorentz_factor) {
+  result->get() = get(tilde_d) * (get(lorentz_factor) - 1.0);
+}
+
+template <typename DataType, size_t Dim, typename Fr>
+void tilde_d_unbound_ut_criterion(
+    const gsl::not_null<Scalar<DataType>*> result,
+    const Scalar<DataType>& tilde_d, const Scalar<DataType>& lorentz_factor,
+    const tnsr::i<DataType, Dim, Fr>& spatial_velocity_one_form,
+    const Scalar<DataType>& lapse, const tnsr::I<DataType, Dim, Fr>& shift) {
+  u_lower_t(result, lorentz_factor, spatial_velocity_one_form, lapse, shift);
+  result->get() = get(tilde_d) * step_function(-1.0 - result->get());
+}
+
+#define DIM(data) BOOST_PP_TUPLE_ELEM(0, data)
+
+#define INSTANTIATE(_, data)                                         \
+  template void u_lower_t(                                           \
+      const gsl::not_null<Scalar<DataVector>*> result,               \
+      const Scalar<DataVector>& lorentz_factor,                      \
+      const tnsr::i<DataVector, DIM(data), Frame::Inertial>&         \
+          spatial_velocity_one_form,                                 \
+      const Scalar<DataVector>& lapse,                               \
+      const tnsr::I<DataVector, DIM(data), Frame::Inertial>& shift); \
+  template void tilde_d_unbound_ut_criterion(                        \
+      const gsl::not_null<Scalar<DataVector>*> result,               \
+      const Scalar<DataVector>& tilde_d,                             \
+      const Scalar<DataVector>& lorentz_factor,                      \
+      const tnsr::i<DataVector, DIM(data), Frame::Inertial>&         \
+          spatial_velocity_one_form,                                 \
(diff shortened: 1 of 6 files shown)


src/PointwiseFunctions/Hydro/MassWeightedFluidItems.hpp
```diff
@@ -0,0 +1,123 @@
...
+template <typename DataType>
+struct MassWeightedInternalEnergy : db::SimpleTag {
+  using type = Scalar<DataType>;
+};
+template <typename DataType>
+struct MassWeightedKineticEnergy : db::SimpleTag {
+  using type = Scalar<DataType>;
+};
+template <typename DataType>
+struct TildeDUnboundUt : db::SimpleTag {
```
wthrowe: It's not clear to me what the "Ut" suffix means here.

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.hpp
```diff
@@ -0,0 +1,123 @@
...
+
+  using return_type = Scalar<DataType>;
+
+  using base = MassWeightedInternalEnergy<DataType>;
+
+  static void function(const gsl::not_null<Scalar<DataType>*> result,
+                       const Scalar<DataType>& tilde_d,
+                       const Scalar<DataType>& specific_internal_energy) {
+    mass_weighted_internal_energy(result, tilde_d, specific_internal_energy);
+  }
```
wthrowe: You should be able to do just `static constexpr function = mass_weighted_internal_energy<DataType>;` or something like that.  It gets harder if there are overloads, but that's not relevant here.  Same for the others.

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+template <typename DataType, size_t Dim, typename Frame>
+void u_lower_t(const gsl::not_null<Scalar<DataType>*> result,
+               const Scalar<DataType>& lorentz_factor,
+               const tnsr::i<DataType, Dim, Frame>& spatial_velocity_one_form,
+               const Scalar<DataType>& lapse,
+               const tnsr::I<DataType, Dim, Frame>& shift) {
+  result->get() = get(lorentz_factor) *
+                  (get(lapse) * (-1.0) +
+                   get(dot_product(spatial_velocity_one_form, shift)));
```
wthrowe: You can avoid an allocation here by using the return-by-reference `dot_product`.  Something like (untested)
```c++
dot_product(result, spatial_velocity_one_form, shift);
get(*result) = get(lorentz_factor) * (get(*result) - get(lapse));
```

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.cpp
```diff
@@ -0,0 +1,90 @@
...
+    const Scalar<DataType>& tilde_d,
+    const Scalar<DataType>& specific_internal_energy) {
+  result->get() = get(tilde_d) * get(specific_internal_energy);
+}
+
+template <typename DataType>
+void mass_weighted_kinetic_energy(const gsl::not_null<Scalar<DataType>*> result,
+                                  const Scalar<DataType>& tilde_d,
+                                  const Scalar<DataType>& lorentz_factor) {
+  result->get() = get(tilde_d) * (get(lorentz_factor) - 1.0);
```
wthrowe, line 37: Don't know if it matters here or if you have the appropriate values available, but `lorentz_factor - 1` can be written with better numerical behavior at small velocities as `square(lorentz_factor) * square(velocity) / (lorentz_factor + 1)`.  I guess the kinetic energy will be dominated by regions where the velocity is large, so the extra math may not be worth it.

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.cpp
```diff
@@ -0,0 +1,90 @@
...
+}
+
+template <typename DataType, size_t Dim, typename Fr>
+void tilde_d_unbound_ut(
+    const gsl::not_null<Scalar<DataType>*> result,
+    const Scalar<DataType>& tilde_d, const Scalar<DataType>& lorentz_factor,
+    const tnsr::i<DataType, Dim, Fr>& spatial_velocity_one_form,
+    const Scalar<DataType>& lapse, const tnsr::I<DataType, Dim, Fr>& shift) {
+  Scalar<DataType> u_t{};
```
wthrowe: As above, you can use `result` to store the temporary and avoid an allocation.

wthrowe (changes requested):
I'm fine with things being templated if you want.

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.hpp
ffoucart: I'll add comments and maybe change the name; effectively it relates to what "unbound" means, in this case u_t<-1 (as opposed to hu_t<h_min, or more complex methods)

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.cpp
ffoucart: I'd agree there are improvements to be made; the same is true for the Lorentz factor itself, which is currently calculated as
1/sqrt(1-v^2)
instead of
sqrt(1+u^2)
(with u^2 = gamma^{ij} u_i u_j and u the 4-velocity)

ffoucart:
(force-push uses DataVector and Frame::Inertial in the explicit instantiation of functions, instead of a loop over a single allowed value...)

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.hpp
ffoucart: The simple version didn't work, but I got something a bit more complex from another file that seems to be ok and accomplishes the same thing (at least on wheeler).

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.cpp
ffoucart: The lorentz_factor ~ 1 case should indeed be unimportant here, so I'll leave it as is for now.

wthrowe:
Looks good to me.  Squash.

wthrowe (dismissed):

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.hpp
```diff
@@ -0,0 +1,124 @@
...
+namespace gsl {
+template <typename>
+struct not_null;
+}  // namespace gsl
+/// \endcond
+
+namespace hydro {
+
+namespace Tags {
+template <typename DataType>
```
nilsdeppe, line 28: could you please document these simple tags?

src/PointwiseFunctions/Hydro/MassWeightedFluidItems.hpp
```diff
@@ -0,0 +1,124 @@
...
+/// Contains TildeD restricted to regions marked as
+/// unbound, using the u_t < -1 criterion.
+template <typename DataType>
+struct TildeDUnboundUtCriterion : db::SimpleTag {
+  using type = Scalar<DataType>;
+};
+}  // namespace Tags
+
+template <typename DataType>
+void mass_weighted_internal_energy(
```
nilsdeppe, line 52: can you please document these functions?

nilsdeppe (changes requested):
you can squash immediately

ffoucart:
Added comments. Ignore the two intermediate commits, I accidentally pushed a different observer that I am working on.

nilsdeppe (approved):

PR #4798, opened 2023-03-02
Adds infrastructure and actions for elements to send data to the worldtube

nikwit:
## Proposed changes

Adds an action for elements to  integrate and send data  to the worldtube `SendToWorldtube` and an action for the worldtube to receive that data and add up the different pieces of the integral `ReceiveElementData`.

~depends on #4780~

PR #4804, opened 2023-03-03
Add actions to send data from worldtube to abutting elements

nikwit:
## Proposed changes
Adds the final piece of the 0th order worldtube scheme:

-  an action `SendToElements` in which the worldtube sends the regular field it calculated to the abutting elements 
-  an action `ReceiveWorldtubeData` in which the elements process the sent data into boundary conditions

~depends on #4798~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp
@@ -32,4 +32,21 @@ struct SphericalHarmonicsInbox
       std::map<temporal_id,
                std::unordered_map<ElementId<Dim>, Variables<tags_list>>>;
 };
+
+/*!
+ * \brief Inbox of the element chares that contains the regular field $\Psi^R$
+ * as well as its time and spatial derivative evaluated at the grid points of
+ * abutting element faces.
+ */
+template <size_t Dim>
+struct RegularFieldInbox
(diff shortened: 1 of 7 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp
```diff
@@ -32,4 +32,23 @@ struct SphericalHarmonicsInbox
...
+    : Parallel::InboxInserters::Map<RegularFieldInbox<Dim>> {
+  using tags_to_send =
+      tmpl::list<CurvedScalarWave::Tags::Psi,
+                 ::Tags::dt<CurvedScalarWave::Tags::Psi>,
+                 ::Tags::deriv<CurvedScalarWave::Tags::Psi, tmpl::size_t<Dim>,
+                               Frame::Grid>>;
+  using temporal_id = TimeStepId;
+  using type =
+      std::map<temporal_id,
+               std::unordered_map<ElementId<Dim>, Variables<tags_to_send>>>;
```
nilsvu: You don't need the map over element IDs here because you send from a singleton.

nilsvu (changes requested):

nilsvu (dismissed):

nikwit:
thanks for your review, I squashed it in :)

nilsdeppe (approved):

PR #4810, opened 2023-03-06
Split PyTensor instantiations into multiple files

### LINKED ISSUE #4741: Building PyTensor is slow
We currently instantiate 240 tensor types in `DataStructures/Tensor/Python/Tensor.cpp`, which takes quite a long time (couple of minutes, depending on the machine). Not sure how to speed that up though (suggestions welcome).

nilsvu:
## Proposed changes

Allows compiling them in parallel and reduces memory usage.

Fixes #4741 and (hopefully) issues on CI.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/Tensor/Python/InstantiateScalar.cpp
+++ b/src/DataStructures/Tensor/Python/InstantiateScalar.cpp
@@ -1,0 +1,13 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "DataStructures/Tensor/Python/Tensor.tpp"
+
+namespace py_bindings {
+
+void bind_scalar(pybind11::module& m) {
+  bind_tensor_impl<Scalar<DataVector>, TensorKind::Scalar>(m, "Scalar");
+  bind_tensor_impl<Scalar<double>, TensorKind::Scalar>(m, "Scalar");
+}
+
+}  // namespace py_bindings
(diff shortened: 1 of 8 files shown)


src/DataStructures/Tensor/Python/InstantiateScalar.cpp
```diff
@@ -0,0 +1,13 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "DataStructures/Tensor/Python/Tensor.tpp"
```
knelli2, line 4: Do you also need the `.hpp` file here as well?

nilsvu: It's included in the tpp 🤷‍♂️ this file just does instantiations so I think this is fine

nilsvu: Oh this is the scalar file.. wasn't sure here either, but doesn't matter much I think

knelli2: No it doesn't really matter. I thought we might have a guideline for this in our docs, but I couldn't find one. So this seems ok.

wthrowe: I think when the tpp+hpp case first came up none of us could figure out what the right answer was, so we just left it out of the guidelines and let people do what they felt like.

nilsdeppe (approved):
LGTM

Issue #4593, opened 2023-01-12
Char speed control: add DistortedFrame support


geoffrey4444:
# Feature request:

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [x] Find places in spectre that assume grid or inertial frame, and add support for the distorted frame.

### Detailed discussion:
Note: the distorted frame includes the shape and size maps, but not translation, rotation, or expansion maps.

PR #4813, opened 2023-03-07
Have the control system work in the Distorted frame

knelli2:
## Proposed changes

Instead of the Grid frame. We'll need this for shape and size control.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments
~Depends on and includes #4792.~

### DIFF
--- a/src/ApparentHorizons/ObserveCenters.hpp
+++ b/src/ApparentHorizons/ObserveCenters.hpp
@@ -26,6 +26,7 @@
 /// \cond
 namespace Frame {
 struct Grid;
+struct Distorted;
 struct Inertial;
 }  // namespace Frame
 /// \endcond
@@ -34,8 +35,8 @@ namespace ah {
 namespace callbacks {
 /*!
  * \brief Writes the center of an apparent horizon to disk in both the
- * Frame::Grid frame and Frame::Inertial frame. Intended to be used in the
- * `post_horizon_find_callbacks` list of an InterpolationTargetTag.
+ * `Frame` template parameter frame and Frame::Inertial frame. Intended to be
+ * used in the `post_horizon_find_callbacks` list of an InterpolationTargetTag.
  *
  * The centers will be written to a subfile with the name
  * `/ApparentHorizons/TargetName_Centers.dat` where `TargetName` is the
@@ -50,12 +51,17 @@ namespace callbacks {
  * - InertialCenter_y
  * - InertialCenter_z
  *
- * \note Requires StrahlkorperTags::Strahlkorper<Frame::Grid>
+ * The `Frame` template parameter must be either `::Frame::Grid` or
+ * `::Frame::Distorted`. Even though the template parameter can be
+ * `::Frame::Distorted`, we still write `GridCenter_?` because the centers of
+ * the objects are the same in the Grid and Distorted frames.
+ *
+ * \note Requires StrahlkorperTags::Strahlkorper<Frame>
  * and StrahlkorperTags::CartesianCoords<Frame::Inertial> and
- * StrahlkorperTags::EuclideanAreaElement<Frame::Grid> to be in the DataBox
+ * StrahlkorperTags::EuclideanAreaElement<Frame> to be in the DataBox
  * of the InterpolationTarget.
  */
-template <typename InterpolationTargetTag>
+template <typename InterpolationTargetTag, typename Frame>
 struct ObserveCenters {
   // Note that we don't add a const_global_cache_tags type alias here with
   // ah::Tags::ObserveCenters because we want to use the base tag and so must be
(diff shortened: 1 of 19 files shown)


knelli2:
@ffoucart Just an FYI, with this PR, you'll have to have a distorted frame in your BNS domain because the control system works in the distorted frame. The Grid -> Distorted map will just be identity for BNS but it has to exist. I've chosen to do it this way because if we ever do BHNS, we will *have* to have the distorted frame (for the horizon), so it's easiest to just always require the control system to work in the distorted frame.

ffoucart:
Thanks. We haven't even started putting the control system in the input files, so... we should be able to take that into account when we get there :-)

src/ApparentHorizons/ObserveCenters.hpp
```diff
@@ -50,12 +51,17 @@ namespace callbacks {
...
+ * `::Frame::Distorted`, we still write `GridCenter_?` because the centers of
+ * the objects are the same in the Grid and Distorted frames.
+ *
+ * \note Requires StrahlkorperTags::Strahlkorper<Frame>
  * and StrahlkorperTags::CartesianCoords<Frame::Inertial> and
  * StrahlkorperTags::EuclideanAreaElement<Frame::Grid> to be in the DataBox
  * of the InterpolationTarget.
  */
-template <typename InterpolationTargetTag>
+template <typename InterpolationTargetTag, typename Frame>
```
nilsvu, line 64: (optional) To avoid collision with the `Frame` namespace I've found it helpful to name template parameters like this more descriptively, like `ObservationFrame` in this case maybe.

tests/Unit/ApparentHorizons/Test_ObserveCenters.cpp
```diff
@@ -62,9 +62,9 @@ SPECTRE_TEST_CASE("Unit.ApparentHorizons.ObserveCenters",
   std::vector<std::array<double, 3>> grid_centers{};
   std::vector<std::array<double, 3>> inertial_centers{};
 
-  db::DataBox<tmpl::list<StrahlkorperTags::Strahlkorper<Frame::Grid>,
-                         StrahlkorperTags::CartesianCoords<Frame::Inertial>,
-                         StrahlkorperTags::EuclideanAreaElement<Frame::Grid>>>
+  db::DataBox<tmpl::list<StrahlkorperTags::Strahlkorper<Frame>,
+                         StrahlkorperTags::CartesianCoords<::Frame::Inertial>,
```
nilsvu, line 66: (this is what you avoid with my previous comment)

src/Domain/Creators/Tags/ObjectCenter.cpp
```diff
@@ -14,31 +14,56 @@
...
   const std::string name = "ExcisionSphere"s + get_output(Label);
   if (domain.excision_spheres().count(name) != 1) {
     ERROR(name << " is not in the domains excision spheres but is needed to "
                   "generate the ExcisionCenter<"
-               << Label << ">.");
+               << Label << "," << pretty_type::name<Frame>() << ">.");
   }
 
-  return domain.excision_spheres().at(name).center();
+  tnsr::I<double, 3, Frame> result{};
```
nilsvu: (optional) you can short-circuit here with an `if constexpr (Frame is grid) { return center; } else { return as distorted frame }`

src/Domain/Creators/Tags/ObjectCenter.hpp
```diff
@@ -13,34 +13,31 @@
...
 /// \ingroup ComputationalDomainGroup
-/// The grid frame center of the excision sphere for the given object.
+/// The center of the excision sphere for the given object in the given frame.
 ///
 /// Even though this can easily be retrieved from the domain, we add it as its
 /// own tag so we can access it through the base tag. This way, other things
 /// (like the control system) can grab the center and be agnostic to what the
 /// object actually is.
-template <ObjectLabel Label>
+template <ObjectLabel Label, typename Frame>
```
nilsvu: I think you can just drop this commit and work with the grid coords in the places where you retrieve this tag. Otherwise you should add the frame parameter also to `ObjectCenter` and that just complicates everything unnecessarily I think. As you said, per definition the grid and distorted frame excision centers are the same. Dropping this commit also removes quite a bit of unnecessary copying of tensors in the next commit I think.

nilsvu (changes requested):

knelli2:
@nilsvu Dropped the suggested commit. Didn't do the optional change though about `Frame`.

tests/Unit/ControlSystem/ControlErrors/Test_Shape.cpp
```diff
@@ -65,7 +65,7 @@ void test_shape_control_error() {
   MAKE_GENERATOR(generator);
   domain::FunctionsOfTime::register_derived_with_charm();
 
-  const tnsr::I<double, 3, Frame::Grid> origin{0.0};
+  const tnsr::I<double, 3, Frame::Distorted> origin{0.0};
```
nilsvu: I think you can just drop this change and the origin_grid below

tests/Unit/ControlSystem/Systems/Test_Shape.cpp
```diff
@@ -105,9 +105,13 @@ void test_shape_control(
   SpherepackIterator iter{l_max, l_max};
   const double ah_radius =
       ah_coefs_function_of_time.func(initial_time)[0][iter.set(0, 0)()];
-  const tnsr::I<double, 3, Frame::Grid>& center = excision_sphere.center();
+  const tnsr::I<double, 3, Frame::Grid>& grid_center = excision_sphere.center();
```
nilsvu, line 108: I think you can drop all this too

nilsvu (changes requested):

nilsvu:
@markscheel could you have a glance over this PR?

src/ApparentHorizons/ObserveCenters.hpp
```diff
@@ -50,12 +51,17 @@ namespace callbacks {
...
  *
- * \note Requires StrahlkorperTags::Strahlkorper<Frame::Grid>
+ * The `Frame` template parameter must be either `::Frame::Grid` or
+ * `::Frame::Distorted`. Even though the template parameter can be
+ * `::Frame::Distorted`, we still write `GridCenter_?` because the centers of
+ * the objects are the same in the Grid and Distorted frames.
+ *
+ * \note Requires StrahlkorperTags::Strahlkorper<Frame>
  * and StrahlkorperTags::CartesianCoords<Frame::Inertial> and
  * StrahlkorperTags::EuclideanAreaElement<Frame::Grid> to be in the DataBox
```
markscheel: `EuclideanAreaElement<Frame>`

markscheel (commented):
Looks good! Only one small dox change; ok to squash.

nilsvu (approved):

knelli2:
Fixes #4593

PR #4815, opened 2023-03-07
Observe hydro quantities in XCTS solver

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/Xcts/HydroQuantities.cpp
+++ b/src/Elliptic/Systems/Xcts/HydroQuantities.cpp
@@ -1,0 +1,23 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Elliptic/Systems/Xcts/HydroQuantities.hpp"
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "PointwiseFunctions/GeneralRelativity/IndexManipulation.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace Xcts::Tags {
+
+void LowerSpatialFourVelocityCompute::function(
+    const gsl::not_null<tnsr::i<DataVector, 3>*> result,
+    const tnsr::I<DataVector, 3>& spatial_velocity,
+    const tnsr::ii<DataVector, 3>& spatial_metric,
+    const Scalar<DataVector>& lorentz_factor) {
+  raise_or_lower_index(result, spatial_velocity, spatial_metric);
+  for (size_t d = 0; d < 3; ++d) {
+    result->get(d) *= get(lorentz_factor);
+  }
+}
+
+}  // namespace Xcts::Tags
--- a/src/Elliptic/Executables/Xcts/SolveXcts.hpp
+++ b/src/Elliptic/Executables/Xcts/SolveXcts.hpp
@@ -17,6 +17,7 @@
 #include "Elliptic/SubdomainPreconditioners/RegisterDerived.hpp"
 #include "Elliptic/Systems/Xcts/BoundaryConditions/Factory.hpp"
 #include "Elliptic/Systems/Xcts/FirstOrderSystem.hpp"
+#include "Elliptic/Systems/Xcts/HydroQuantities.hpp"
 #include "Elliptic/Triggers/Factory.hpp"
 #include "IO/Observer/Actions/RegisterEvents.hpp"
 #include "IO/Observer/Helpers.hpp"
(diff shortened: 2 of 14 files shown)


src/Elliptic/Systems/Xcts/HydroQuantities.cpp
```diff
@@ -0,0 +1,23 @@
...
+
+#include "Elliptic/Systems/Xcts/HydroQuantities.hpp"
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "PointwiseFunctions/GeneralRelativity/IndexManipulation.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace Xcts::Tags {
+
+void LowerSpatialFourVelocityCompute::function(
```
nilsdeppe, line 12: The SpECTRE GRMHD code actually uses `Wv^i` since we always already have `W` and `v^i`. The evolution code has no support for `u_i`. The reason is that everything always works in `W` and `v^i` so it's easier to just use `Wv^i` than `u_i=Wgamma_{ij}v^j`. I don't know why SpEC uses `u_i`, since `Wv^i` also guarantees a causal spatial velocity if interpolated. In summary: if you're observing this for the SpECTRE evolution code, you should do `Wv^i` since that's what the code uses. There's even a tag for it: https://github.com/sxs-collaboration/spectre/blob/9c14163b07470050cee18ea1a4e64bceb56886ec/src/PointwiseFunctions/Hydro/Tags.hpp#L212

nilsdeppe (changes requested):
LGTM, one question. If it's easier, we can discuss on Slack or a hangout :) Squash immediately!

src/Elliptic/Systems/Xcts/HydroQuantities.cpp
nilsvu: Where is `LorentzFactorTimesSpatialVelocity` added to the DataBox? I can see how the ValenciaDivClean system specifies primitive vars which are retrieved from the analytic solution / data, but it's not transparent to me how `W v^i` gets into the DataBox (it's not included in the ValenciaDivClean system. it's added to `SubcellOptions::recons_prim_tags` so maybe that has something to do with it?).

nilsdeppe: It's only used in the reconstruction, it's not added to the DataBox. What I'm saying is that tag exists and since we already use `Wv^i` it might be less confusing to just stick to `Wv^i` and avoid adding `u_i`, if possible :)

nilsvu: @ffoucart any preference for interpolating `W v^i` vs `W gamma_ij v^j`?

nilsvu: (@nilsdeppe this quantity is only written out of the ID solver, then interpolated to the evolution grid, then split into `W` and `v^i` and stored in the DataBox. So I think it doesn't necessarily have to be the same as an evolution / reconstruction quantity. I have no preference here, I think we should just pick whichever is best for interpolation / storage. Maybe we'll have to experiment with that a bit as well)

ffoucart: No particular preference between those two on my side, no.

nilsdeppe: Okay, probably the only thing that would be good is to make both SpEC and SpECTRE dump the same so it's easier to deal with. I have no preference overall.

nilsdeppe (approved):

src/Elliptic/Systems/Xcts/HydroQuantities.cpp
nilsvu: Ok I believe spec outputs u_i with https://github.com/sxs-collaboration/spec/pull/2221, so should we go with this for now? I also have no preference so we can switch to W v^i later anytime or try it if we run into issues with u_i. It's as easy as adding a compute item for it to the XCTS observations.

nilsdeppe: Yes, just do `u_i` for now. No need to try to figure if one is slightly better than the other :) Probably doesn't matter since they only differ by the spatial metric, which should be relatively smooth

nilsvu: This PR is ready to merge then I think

PR #4817, opened 2023-03-08
Add `GridtoInertialInverseJacobian` compute tag

nikwit:
## Proposed changes

Adds a compute tag that computes the grid to inertial inverse jacobian. If the mesh is moving it is be a data reference into `CoordinatesMeshVelocityAndJacobians`. Otherwise it throws an error if it is requested.

### DIFF
--- a/src/Domain/TagsTimeDependent.hpp
+++ b/src/Domain/TagsTimeDependent.hpp
@@ -103,6 +103,30 @@ struct InertialFromGridCoordinatesCompute
                                    CoordinatesMeshVelocityAndJacobians<Dim>>;
 };
 
+/// Computes the Grid to Inertial inverse Jacobian from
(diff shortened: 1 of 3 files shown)


src/Domain/TagsTimeDependent.hpp
```diff
@@ -103,6 +103,29 @@ struct InertialFromGridCoordinatesCompute
                                    CoordinatesMeshVelocityAndJacobians<Dim>>;
 };
 
+/// Computes the Grid to Inertial inverse Jacobian from
+/// `CoordinatesVelocityAndJacobians`
```
nilsdeppe: Can you document here that this'll throw an error if the mesh is static?

nilsdeppe (changes requested):
you can squash

nilsdeppe (approved):
Waiting until the March release is done before merging.

PR #4823, opened 2023-03-10
Write control system damping timescales to disk

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/WriteData.hpp
+++ b/src/ControlSystem/WriteData.hpp
@@ -33,6 +33,7 @@ namespace control_system {
  * - d2tLambda
  * - ControlError
  * - dtControlError
+ * - DampingTimescale
  *
  * where "Lambda" is the function of time value.
  *
@@ -62,7 +63,7 @@ void write_components_to_disk(
     const double time, Parallel::GlobalCache<Metavariables>& cache,
     const std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>&
         function_of_time,
-    const std::array<DataVector, 2>& q_and_dtq) {
+    const std::array<DataVector, 2>& q_and_dtq, const DataVector& timescales) {
   auto& observer_writer_proxy = Parallel::get_parallel_component<
       observers::ObserverWriter<Metavariables>>(cache);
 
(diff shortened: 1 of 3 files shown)


src/ControlSystem/WriteData.hpp
```diff
@@ -33,6 +33,7 @@ namespace control_system {
  * - d2tLambda
  * - ControlError
  * - dtControlError
+ * - Timescale
  *
  * where "Lambda" is the function of time value.
```
nilsdeppe, line 38: Why don't we just call it "FunctionOfTime" and "dtFunctionOfTime" in the output? Then you don't need to document what it is.

knelli2: #4827 I'll do this in a follow up PR

nilsdeppe: Sounds good!

src/ControlSystem/WriteData.hpp
```diff
@@ -33,6 +33,7 @@ namespace control_system {
  * - d2tLambda
  * - ControlError
  * - dtControlError
+ * - Timescale
```
nilsdeppe: I don't know enough about the control systems, so just asking to make sure: is there only one "Timescale"? By this I mean could someone who actually knows more about the control system be confused that this means "Timescale X" or "Timescale Y"? If so, it might be worth adding a bit more documentation. I'll leave it up to you to decide since I don't know enough here :)

nilsdeppe (dismissed):

src/ControlSystem/WriteData.hpp
knelli2: Technically there's only one timescale that matters and that's the damping timescale. However, we often talk about the update timescale, measurement timescale, expiration time, etc. which are all based off the damping timescale. So that could have been kind of confusing. I changed the name to be `DampingTimescale`.

And there's one damping timescale for each component of a function of time, but since we write one file for each component, that's already taken care of.

nilsdeppe: Okay, sounds good. Thanks for clarifying!

nilsdeppe (approved):

PR #4824, opened 2023-03-10
Refactor CylBBH test

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
+++ b/tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
@@ -6,11 +6,13 @@
 #include <array>
 #include <cmath>
 #include <cstddef>
+#include <iomanip>
 #include <iterator>
 #include <limits>
 #include <memory>
 #include <pup.h>
 #include <random>
+#include <sstream>
 #include <string>
 #include <tuple>
 #include <unordered_map>
@@ -36,36 +38,14 @@
 #include "Helpers/Domain/Creators/TestHelpers.hpp"
 #include "Helpers/Domain/DomainTestHelpers.hpp"
 #include "Utilities/CartesianProduct.hpp"
+#include "Utilities/GetOutput.hpp"
 #include "Utilities/MakeArray.hpp"
 #include "Utilities/ProtocolHelpers.hpp"
 #include "Utilities/TMPL.hpp"
 
-namespace domain::FunctionsOfTime {
-class FunctionOfTime;
-}  // namespace domain::FunctionsOfTime
-
 namespace {
-using BoundaryCondVector = std::vector<DirectionMap<
-    3, std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>>>;
 using ExpirationTimeMap = std::unordered_map<std::string, double>;
-
-template <size_t Dim, bool EnableTimeDependentMaps, bool WithBoundaryConditions>
-struct Metavariables {
-  struct domain : tt::ConformsTo<::domain::protocols::Metavariables> {
-    static constexpr bool enable_time_dependent_maps = EnableTimeDependentMaps;
-  };
-  using system = tmpl::conditional_t<WithBoundaryConditions,
-                                     TestHelpers::domain::BoundaryConditions::
-                                         SystemWithBoundaryConditions<Dim>,
-                                     TestHelpers::domain::BoundaryConditions::
-                                         SystemWithoutBoundaryConditions<Dim>>;
-  struct factory_creation
-      : tt::ConformsTo<Options::protocols::FactoryCreation> {
-    using factory_classes = tmpl::map<tmpl::pair<
-        DomainCreator<3>,
-        tmpl::list<::domain::creators::CylindricalBinaryCompactObject>>>;
-  };
-};
+using CylBCO = ::domain::creators::CylindricalBinaryCompactObject;
 
 std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
 create_inner_boundary_condition() {
@@ -97,47 +77,11 @@ create_outer_boundary_condition() {
       Direction<3>::upper_eta(), 314);
 }
 
-void test_connectivity_once(const bool with_sphere_e,
-                            const bool include_inner_sphere_A,
-                            const bool include_inner_sphere_B,
-                            const bool include_outer_sphere,
-                            const bool use_equiangular_map,
-                            const bool with_boundary_conditions) {
-  // Misc.:
-  constexpr size_t refinement = 1;
-  constexpr size_t grid_points = 3;
-  const double outer_radius = include_outer_sphere ? 100.0 : 30.0;
-
-  const double mass_ratio = with_sphere_e ? 4 : 1.2;
-  const double separation = 9.0;
-  const double y_offset = 0.05;
-  // Set centers so that the Newtonian COM is at the origin,
-  // except offset slightly in the y direction.
-  const std::array<double, 3> center_objectA = {separation / (1.0 + mass_ratio),
-                                                y_offset, 0.0};
-  const std::array<double, 3> center_objectB = {
-      -separation * mass_ratio / (1.0 + mass_ratio), y_offset, 0.0};
-
-  // Set radius = 2m for each object.
-  // Formula comes from assuming m_A+m_B=1.
-  const double inner_radius_objectA = 2.0 * mass_ratio / (1.0 + mass_ratio);
-  const double inner_radius_objectB = 2.0 / (1.0 + mass_ratio);
-
-  const domain::creators::CylindricalBinaryCompactObject binary_compact_object{
-      center_objectA,
-      center_objectB,
-      inner_radius_objectA,
-      inner_radius_objectB,
-      include_inner_sphere_A,
-      include_inner_sphere_B,
-      include_outer_sphere,
-      outer_radius,
-      use_equiangular_map,
-      refinement,
-      grid_points,
-      with_boundary_conditions ? create_inner_boundary_condition() : nullptr,
-      with_boundary_conditions ? create_outer_boundary_condition() : nullptr};
-
+std::pair<std::vector<std::string>,
+          std::unordered_map<std::string, std::unordered_set<std::string>>>
+block_names_and_groups(const bool include_inner_sphere_A,
+                       const bool include_inner_sphere_B,
+                       const bool include_outer_sphere) {
   std::vector<std::string> block_names{
       "CAFilledCylinderCenter", "CAFilledCylinderEast",
       "CAFilledCylinderNorth",  "CAFilledCylinderWest",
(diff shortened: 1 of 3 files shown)


tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -612,19 +502,162 @@ void test_parse_errors() {
...
+          Options::Context{false, {}, 1, 1}),
+      Catch::Matchers::Contains(
+          "Must specify either both inner and outer boundary "
+          "conditions or neither."));
+}
+
+void test_cylindrical_bbh() {
+  MAKE_GENERATOR(gen);
+
+  const std::vector<double> times_to_check{{1., 10.}};
```
wthrowe: Haven't reviewed this PR at all, but I was looking at this test because it keeps timing out and determined that this "10" case is over 2/3 of the runtime because of all the integration in the FoTs.  Do we need to include this case, or can we change it to a smaller number?

knelli2: I can probably lower it down to like 2.3 or something. I don't think there's anything special about 10

wthrowe: On my machine that reduces the test runtime by about a factor of 2, so that sounds good.

knelli2: Was that with this PR, or just develop?

wthrowe: That was just on develop.  I assumed it would be similar here since this PR doesn't appear to have affected the total runtime much.

knelli2: Ok made the change

wthrowe:
I would prefer for someone familiar with this code to review the actual refactor.  @nilsvu or @markscheel maybe?

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -333,60 +286,83 @@ std::string create_option_string(
...
+         "\n  IncludeInnerSphereB: " + stringize(add_time_dependence) +
+         "\n  IncludeOuterSphere: " + stringize(include_outer_sphere) +
+         "\n  InitialRefinement:" +
+         initial_structure(with_additional_outer_radial_refinement, 1) +
+         "\n  InitialGridPoints:" +
+         initial_structure(with_additional_grid_points, 3) + "\n" +
+         time_dependence + boundary_conditions;
+}
+
+void test_construction(const std::unique_ptr<DomainCreator<3>>& creator,
```
nilsvu: Take a `CylindricalBinaryCompactObject` explicitly here instead of the unique_ptr

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -600,16 +538,126 @@ void test_parse_errors() {
...
+  constexpr size_t refinement = 1;
+  constexpr size_t grid_points = 3;
+
+  const double separation = 9.0;
+  const double y_offset = 0.05;
+
+  // When we add sphere_e support we will make the following
+  // loop go over {true, false}
+  const bool with_sphere_e = false;
+  for (auto& [include_outer_sphere, include_inner_sphere_A,
```
nilsvu: probably better to take as `auto` (no reference) if you modify some of these below

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -600,16 +538,126 @@ void test_parse_errors() {
...
+    CAPTURE(include_outer_sphere);
+    CAPTURE(use_equiangular_map);
+    CAPTURE(with_boundary_conditions);
+    CAPTURE(with_additional_outer_radial_refinement);
+    CAPTURE(with_additional_grid_points);
+    CAPTURE(with_time_dependence);
+    bool with_control_systems = false;
+    if (with_time_dependence) {
+      with_control_systems = gsl::at(
+          random_sample<1>(make_array(true, false), make_not_null(&gen)), 0);
```
nilsvu: include in the cartesian product and override with `false` if not time dependent?

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -600,16 +538,126 @@ void test_parse_errors() {
...
+        separation / (1.0 + mass_ratio), y_offset, 0.0};
+    const std::array<double, 3> center_objectB = {
+        -separation * mass_ratio / (1.0 + mass_ratio), y_offset, 0.0};
+
+    // Set radius = 2m for each object.
+    // Formula comes from assuming m_A+m_B=1.
+    const double inner_radius_objectA = 2.0 * mass_ratio / (1.0 + mass_ratio);
+    const double inner_radius_objectB = 2.0 / (1.0 + mass_ratio);
+
+    std::unique_ptr<DomainCreator<3>> cyl_binary_compact_object{};
```
nilsvu: With #4772 now merged you can rebase and use `TestHelpers::domain::creators::test_creation` like in `Test_Sphere.cpp`:

```cpp
const CylBinCompObj cyl_bin_comp_obj{...};
test_construction(cyl_bin_comp_obj, ...);
TestHelpers::domain::creators::test_creation(option_string(...), cyl_bin_comp_obj);
```

This tests that option parsing works by comparing to the manually constructed object.

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -600,16 +538,126 @@ void test_parse_errors() {
...
+                    with_time_dependence,
+                    with_additional_outer_radial_refinement,
+                    with_additional_grid_points, include_outer_sphere,
+                    with_boundary_conditions, use_equiangular_map,
+                    center_objectA, center_objectB, inner_radius_objectA,
+                    inner_radius_objectB, outer_radius));
+      }
+
+      const auto& [initial_expiration_times, expected_functions_of_time] =
+          construct_expected_functions_of_time(with_control_systems);
```
nilsvu: Does this actually test something useful? It seems to me that it mostly reimplements the FoT function of the domain creator. I think it would be best to delete this test of the functions of time (which also gets rid of the `if (with_time_dependence)` clause) and instead check in `test_construction` that a few specific coordinates move around as you would expect. There's a super simple example in `Test_Interval.cpp`. We can defer this to a follow-up PR as well.

nilsvu (changes requested):

nilsvu:
Thanks for refactoring this @knelli2 !

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -612,19 +502,162 @@ void test_parse_errors() {
...
+    test_construction(cyl_binary_compact_object, with_boundary_conditions,
+                      include_inner_sphere_A, include_inner_sphere_B,
+                      include_outer_sphere, inner_radius_objectA,
+                      inner_radius_objectB, center_objectA, center_objectB,
+                      times_to_check);
+  }
 }
 }  // namespace
 
 // [[TimeOut, 60]]
```
nilsvu: Can you try reducing the timeout now? Maybe try 30 seconds or 45.

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
knelli2: It actually took quite a bit of changes to use this, so I put it in a fixup

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
knelli2: For the BCO domains where we manually enter the time dependent maps, I think something like this is useful, but yes please lets defer this until later

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -597,65 +620,114 @@ void test_cylindrical_bbh() {
...
+    const auto make_initial_structure =
+        // https://www.open-std.org/jtc1/sc22/wg21/docs/papers/2017/p0588r1.html
+        // ```
+        // If a lambda-expression [...] captures a structured binding
+        // (explicitly or implicitly), the program is ill-formed.
+        // ```
+        // There is a bug in the standard where structured bindings cannot by
+        // captured in a lambda. This is why we pass them in as arguments here
+        [](const size_t initial_value, const bool with_inner_A,
+           const bool with_inner_B, const bool with_outer_sphere) {
```
nilsvu: Move this lambda to a separate function, that's more readable anyway.

---

FYI: This is an annoying behavior of the standard, I agree. You can work around it by capturing the variables by name and by value like this:

```cpp
const auto lambda = [include_inner_sphere_A = include_inner_sphere_A, ...]() { ... };
```

You can even use the same names I believe.

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -597,65 +620,114 @@ void test_cylindrical_bbh() {
...
       for (const double time : times_to_check) {
         CAPTURE(time);
         TestHelpers::domain::creators::test_functions_of_time(
-            *cyl_binary_compact_object, expected_functions_of_time,
+            cyl_binary_compact_object, expected_functions_of_time,
             initial_expiration_times);
       }
-
     } else {
+      Parallel::printf("%s %s %s\n", stringize(include_inner_sphere_A),
```
nilsvu: delete

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -361,7 +348,16 @@ void test_construction(const std::unique_ptr<DomainCreator<3>>& creator,
   }
 }
 
-auto construct_expected_functions_of_time(const bool with_control_systems) {
+auto construct_creator_with_time_dependent_quantities(
```
nilsvu: (optional because this is just a test) maybe name `construct_time_dependent`? Also it would be better to return the domain creator and std::move it if you have to (I think that's why you did the return-by-reference).

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
nilsvu: ok

knelli2:
@nilsvu Made the changes and squashed everything in

nilsvu (approved):

nilsvu:
@markscheel do you want to look at this or should I just merge?

nilsvu:
Merging to unblock the PRs that depend on this. We can adjust the test later if anything comes up.

PR #4826, opened 2023-03-10
Add time stepping to worldtube

nikwit:
## Proposed changes
Adds the necessary time stepping actions to the worldtube singleton chare. 

~depends on #4825~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
@@ -5,9 +5,12 @@
 
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "DataStructures/DataBox/Tag.hpp"
+#include "Evolution/Initialization/Evolution.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeElementFacesGridCoordinates.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeEvolvedVariables.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ReceiveElementData.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/SendToElements.hpp"
 #include "IO/Observer/Actions/RegisterSingleton.hpp"
 #include "Options/Options.hpp"
 #include "Parallel/Algorithms/AlgorithmSingleton.hpp"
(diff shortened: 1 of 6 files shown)


tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_InitializeEvolvedVariables.cpp
```diff
@@ -0,0 +1,41 @@
...
+      TimeSteppers::History<variables_tag::type>{},
+      std::make_unique<TimeSteppers::AdamsBashforth>(4));
+
+  db::mutate_apply<Initialization::InitializeEvolvedVariables>(
+      make_not_null(&box));
+  CHECK(db::get<variables_tag>(box) == variables_tag::type(size_t(1), 0.));
+  CHECK(db::get<dt_variables_tag>(box) ==
+        dt_variables_tag::type(size_t(1), 0.));
+  CHECK(db::get<::Tags::HistoryEvolvedVariables<variables_tag>>(box) ==
+        TimeSteppers::History<variables_tag::type>(3));
```
wthrowe: This should be a 1, not a 3.  Congratulations, you found a bug in the `History` equality operator, which I will fix shortly.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
```diff
@@ -36,19 +44,35 @@ struct WorldtubeSingleton {
...
+    static constexpr size_t volume_dim = Dim;
+    static constexpr bool has_primitive_and_conservative_vars = false;
+    using variables_tag = ::Tags::Variables<tmpl::list<Tags::Psi0>>;
+  };
+  using step_actions =
+      tmpl::list<Actions::ReceiveElementData,
+                 Actions::SendToElements<Metavariables>,
+                 ::Actions::RecordTimeStepperData<
+                     typename worldtube_system::variables_tag>,
+                 ::Actions::UpdateU<typename worldtube_system::variables_tag>>;
```
wthrowe, line 67: I'm not sure if this is because this is a work in progress or something, but it looks like after updating `Psi0` with the time stepper you immediately overwrite it in `ReceiveElementData`.  What is going on here?

wthrowe (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
nikwit: `ReceiveElementData` now overwrites the `PsiMonopole`, not `Psi0`. But yes, at the moment the time stepper is a bit redundant because at expansion orders 0 and 1 it is just needed to change the slab size which I am adding in a follow up PR. The time derivative is never changed from 0, so nothing happens.

Once I add expansion order 2, the time stepper is needed to solve an ODE in the Singleton.

wthrowe: Ah, missed that that got changed in the rename.

Still a bit confused, though.  Why do you need to evolve an ODE to adjust the slab size?  Why do you care about the slab size if you're not evolving anything?

nikwit: At expansion order n = 0 or 1, I am not really evolving anything in the worldtube singleton. The elements abutting it send data to it and the singleton  sends it right back - the time stepper does not evolve anything. However, the `TimeStepId` is used as the key for the inboxes, so the singleton knows that it got the data from each neighboring element at a certain time step. When the global time step is changed, the worldtube needs to know about this and adjust its time step accordingly so the `TimeStepId` it receives from the elements is the same it is currently on. I suppose at these expansion orders, the worldtube singleton could probably get by without the time stepper but it is good to check the TimeStepId anyway.

Once I add expansion order 2, the time stepper in the worldtube will actually evolve an ODE. This ODE needs to be evolved at the same global time step as the elements so that the boundary information is exchanged at the correct times (same slab boundaries). At that point, the worldtube singleton will definitely need the correct global time step.

wthrowe: OK.  That makes sense.  Thanks for explaining.

wthrowe (approved):

nilsdeppe (approved):

PR #4830, opened 2023-03-13
Add `ChangeSlabSize` action to worldtube singleton

nikwit:
## Proposed changes

Adds capability for the worldtube to change the global slab size with the elements.

The worldtube singleton checks the `TimeStepId` of the inbox key  to see if a slab change has occurred and changes it accordingly.

~depends on #4826 and  #4825~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp
@@ -4,15 +4,25 @@
 #pragma once
 
 #include <cstddef>
+#include <map>
 #include <unordered_map>
 
+#include "DataStructures/DataBox/Prefixes.hpp"
 #include "DataStructures/Variables.hpp"
 #include "Domain/Structure/ElementId.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
 #include "Parallel/InboxInserters.hpp"
 #include "Time/TimeStepId.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+namespace Frame {
+struct Grid;
+}  // namespace Frame
+/// \endcond
 
 namespace CurvedScalarWave::Worldtube::Tags {
-
 /*!
  * \brief Inbox of the worldtube singleton chare which receives quantities
  * projected onto spherical harmonics.
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp
@@ -1,0 +1,88 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <memory>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/Prefixes.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "ParallelAlgorithms/EventsAndTriggers/Tags.hpp"
+#include "Time/Actions/ChangeSlabSize.hpp"
+#include "Time/Tags.hpp"
+#include "Time/TimeStepId.hpp"
+#include "Time/TimeSteppers/TimeStepper.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace CurvedScalarWave::Worldtube::Actions {
+
+/*!
+ * \brief Waits for the data from all neighboring elements and changes the slab
+ * size if a change in the global time step is detected.
+ * \details We check the slab size of the time step id sent by the elements. If
+ * this is different from the slab size currently used by the worldtube
+ * singleton, we assume a global slab size change has occurred in the elements
+ * and adjust the worldtube slab size accordingly.
+ */
+struct ChangeSlabSize {
+  static constexpr size_t Dim = 3;
+  using inbox_tags = tmpl::list<
+      ::CurvedScalarWave::Worldtube::Tags::SphericalHarmonicsInbox<Dim>>;
+
+  template <typename DbTagsList, typename... InboxTags, typename Metavariables,
+            typename ArrayIndex, typename ActionList,
+            typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTagsList>& box, tuples::TaggedTuple<InboxTags...>& inboxes,
+      const Parallel::GlobalCache<Metavariables>& /*cache*/,
+      const ArrayIndex& /*array_index*/, ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    const auto& time_step_id = db::get<::Tags::TimeStepId>(box);
+    const auto& inbox =
+        tuples::get<Tags::SphericalHarmonicsInbox<Dim>>(inboxes);
+    if (inbox.empty()) {
+      return {Parallel::AlgorithmExecution::Retry, std::nullopt};
+    }
+    ASSERT(inbox.size() == 1,
+           "Received data from two different time step ids.");
+    const auto inbox_time_step_id = inbox.begin()->first;
+    const auto& inbox_slab = inbox_time_step_id.step_time().slab();
+    // we received data from a time step id with a different slab size,
+    // indicating a change in the global time step.
+    if (inbox_time_step_id.step_time().slab() !=
+        time_step_id.step_time().slab()) {
+      ASSERT(inbox_slab.start() == time_step_id.step_time().slab().start(),
+             "The new slab should start at the same time as the old one.");
+      // time_step_id comparison does NOT compare slabs.
+      ASSERT(
+          inbox_time_step_id == time_step_id,
+          "Detected change in global time step id but the new time step id is "
+          "different.");
+      const auto new_step = inbox_slab.duration();
+      const auto new_next_time_step_id =
+          db::get<::Tags::TimeStepper<>>(box).next_time_id(inbox_time_step_id,
+                                                           new_step);
+      db::mutate<::Tags::Next<::Tags::TimeStepId>, ::Tags::TimeStep,
+                 ::Tags::Next<::Tags::TimeStep>, ::Tags::TimeStepId>(
+          make_not_null(&box),
+          [&new_next_time_step_id, &new_step, &inbox_time_step_id](
+              const gsl::not_null<TimeStepId*> next_time_step_id,
+              const gsl::not_null<TimeDelta*> time_step,
+              const gsl::not_null<TimeDelta*> next_time_step,
+              const gsl::not_null<TimeStepId*> local_time_step_id) {
+            *next_time_step_id = new_next_time_step_id;
+            *time_step = new_step;
+            *next_time_step = new_step;
+            *local_time_step_id = inbox_time_step_id;
+          });
+    }
+    return {Parallel::AlgorithmExecution::Continue, std::nullopt};
+  }
+};
+}  // namespace CurvedScalarWave::Worldtube::Actions
--- a/tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
+++ b/tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
@@ -1,0 +1,209 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <array>
+#include <cstddef>
+#include <memory>
+#include <random>
+#include <unordered_map>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Domain/Structure/InitialElementIds.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp"
+#include "Framework/ActionTesting.hpp"
+#include "Framework/TestCreation.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "Parallel/ParallelComponentHelpers.hpp"
+#include "Parallel/Phase.hpp"
+#include "Parallel/PhaseDependentActionList.hpp"
+#include "Time/Tags.hpp"
+#include "Time/Time.hpp"
+#include "Time/TimeStepId.hpp"
+#include "Time/TimeSteppers/Rk3HesthavenSsp.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeVector.hpp"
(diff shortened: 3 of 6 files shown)


wthrowe (changes requested):
This method of changing the slab size won't work.  The thing that immediately jumps out at me is that checking that  the ChangeSlabSize Event runs does not check whether a slab change will occur.  I haven't looked at the rest in detail.

Could you send the Slab _after_ it has been changed on the element, probably unconditionally?  I don't know what your communication pattern looks like, so I can't say whether that would have any unpleasant effects there, but it would simplify all of this logic.

nikwit:
Can you explain why this will not work? It should mimic what the elements do and worked when I tested it with the executable thats not yet on develop with different time steppers.

Sending it after the slab size has been changed on the element is harder. Id need to have a bool that indicates it has been changed on the element and somehow make sure the worldtube knows to expect the message and not send boundary conditions that should not be used.

wthrowe:
The trigger does not have to fire on the same step that the change occurs on.  There can be a delay in the change, and having one is encouraged.  (There are probably also LTS issues, but that's not important right now.)

I don't think there is any way for a different component to predict when the DG slab size will change.  There is a similar problem with the step size in LTS, where we have to send the step size along with the "flux data".

I don't know enough about your intended algorithm and communication pattern to make a suggestion beyond considering always sending the slab.

nikwit:
@wthrowe I added a simple algorithm that just uses the keys of the usual inbox to check if there is a global slab size change which seems to work for a bunch of scenarios I tested. Do you think this is better?

Also, is there a way I can check that the starting value of the new slab and old slab are the same?

wthrowe:
> Also, is there a way I can check that the starting value of the new slab and old slab are the same?

`slab1.start() == slab2.start()`

This looks like a much better way to go, except that I don't understand why the current implementation works.  `TimeStepId`s that only differ in the slab length should compare equal, so the new slab-changing code should never be running.  Changing it to explicitly compare the slabs is easy, but the fact that this works means there's something I don't understand going on here.

nikwit:
Yes, the  problem was indeed that the slab-changing code was running one time step after the slab change had occurred in the elements since onlye then the `step_time()` of the time step ids were different. This also explains why the `slab.start()` comparison did not work for me earlier.

I fixed this in a new commit by comparing the slabs directly. If this looks right to you, I'll add a working unit test.

nikwit:
reworked it in a separate action with unit test now

src/Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp
```diff
@@ -6,8 +6,11 @@
 #include <cstddef>
 #include <unordered_map>
 
+#include "DataStructures/DataBox/Prefixes.hpp"
 #include "DataStructures/Variables.hpp"
 #include "Domain/Structure/ElementId.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
```
wthrowe, line 14: Since you're fixing things, might as well also add `<map>`, `"Utilities/TMPL.hpp"`, and a forward declaration of `Frame::Grid`.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp
```diff
@@ -0,0 +1,90 @@
...
+      const Parallel::GlobalCache<Metavariables>& /*cache*/,
+      const ArrayIndex& /*array_index*/, ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    const auto& time_step_id = db::get<::Tags::TimeStepId>(box);
+    auto& inbox = tuples::get<Tags::SphericalHarmonicsInbox<Dim>>(inboxes);
+    if (inbox.empty()) {
+      return {Parallel::AlgorithmExecution::Retry, std::nullopt};
+    }
+    ASSERT(inbox.size() == 1,
+           "Received data from two different time step ids.");
```
wthrowe, line 53: Does your communication pattern couple things tightly enough that it shouldn't be possible to receive two messages without going through the action loop in between?  For example, in (GTS) DG your neighbor can send boundary data twice between your steps.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp
```diff
@@ -0,0 +1,90 @@
...
+            *next_time_step_id = new_next_time_step_id;
+            *time_step = new_step;
+            *next_time_step = new_step;
+            *local_time_step_id = inbox_time_step_id;
+          });
+    }
+    const size_t expected_number_of_senders =
+        db::get<Tags::ElementFacesGridCoordinates<Dim>>(box).size();
+    if (inbox.at(time_step_id).size() < expected_number_of_senders) {
+      return {Parallel::AlgorithmExecution::Retry, std::nullopt};
```
wthrowe: Generally messages should be waited for either by the action that uses them or, occasionally, by a dedicated action.  It seems odd for the `ChangeSlabSize` action to wait for messages that it is not using.  Can this be moved to the action that uses the data?

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
```diff
@@ -0,0 +1,213 @@
...
+      tmpl::list<MockWorldtubeSingleton<MockMetavariables<Dim>>>;
+  using const_global_cache_tags = tmpl::list<>;
+};
+
+SPECTRE_TEST_CASE("Unit.CurvedScalarWave.Worldtube.ChangeSlabSize", "[Unit]") {
+  static constexpr size_t Dim = 3;
+  using worldtube_chare = MockWorldtubeSingleton<MockMetavariables<Dim>>;
+  Parallel::register_classes_with_charm<TimeSteppers::Rk3HesthavenSsp>();
+
+  const auto time_stepper = std::make_unique<TimeSteppers::Rk3HesthavenSsp>();
```
wthrowe: No need for a pointer here, just use a plain `const TimeSteppers::Rk3HesthavenSsp`.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
```diff
@@ -0,0 +1,213 @@
...
+  using chare_type = ActionTesting::MockSingletonChare;
+  using array_index = int;
+  using phase_dependent_action_list = tmpl::list<
+      Parallel::PhaseActions<
+          Parallel::Phase::Initialization,
+          tmpl::list<ActionTesting::InitializeDataBox<
+              db::AddSimpleTags<
+                  ::Tags::TimeStepId, ::Tags::Next<::Tags::TimeStepId>,
+                  ::Tags::TimeStep, ::Tags::Next<::Tags::TimeStep>,
+                  ::Tags::Time, ::Tags::TimeStepper<TimeStepper>,
```
wthrowe: The action does not use `::Tags::Time`.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
```diff
@@ -0,0 +1,213 @@
...
+      Variables<tmpl::list<CurvedScalarWave::Tags::Psi,
+                           ::Tags::dt<CurvedScalarWave::Tags::Psi>>>;
+  {
+    ActionTesting::MockRuntimeSystem<MockMetavariables<Dim>> runner{{}};
+    create_new_runner(make_not_null(&runner));
+    auto& worldtube_inbox =
+        ActionTesting::get_inbox_tag<worldtube_chare,
+                                     Tags::SphericalHarmonicsInbox<Dim>>(
+            make_not_null(&runner), 0);
+    worldtube_inbox[time_step_id_1][element_ids.at(0)] = inbox_variables_type{};
```
wthrowe, line 132: Call the action before sending any messages and check the result.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
```diff
@@ -0,0 +1,213 @@
...
+    // ChangeSlabSize but only 1/2 elements have sent
+    CHECK_FALSE(ActionTesting::next_action_if_ready<worldtube_chare>(
+        make_not_null(&runner), 0));
```
wthrowe, line 135: Check the result of calling the action here.  I think this is when the change is expected (or in this case not expected since there's nothing to do).

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
```diff
@@ -0,0 +1,213 @@
...
+        ActionTesting::get_inbox_tag<worldtube_chare,
+                                     Tags::SphericalHarmonicsInbox<Dim>>(
+            make_not_null(&runner), 0);
+    // different slab start -> invalid
+    const double slab_2_start = 1.7;
+    const Slab slab_2(slab_2_start, slab_1_end);
+    const auto start_time_2 = slab_2.start();
+    const TimeStepId time_step_id_2(true, 1, start_time_2);
+    worldtube_inbox[time_step_id_2][element_ids.at(0)] = inbox_variables_type{};
+#ifdef SPECTRE_DEBUG
```
wthrowe: Move this up to contain the entire block.  Since all the checks are in in the `SPECTRE_DEBUG` section there's no reason to do the setup in release mode.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
```diff
@@ -0,0 +1,213 @@
...
+        ActionTesting::emplace_component_and_initialize<worldtube_chare>(
+            runner, 0,
+            {time_step_id_1, time_stepper->next_time_id(time_step_id_1, step_1),
+             step_1, step_1, start_time_1.value(),
+             std::make_unique<TimeSteppers::Rk3HesthavenSsp>(),
+             element_faces_grid_coords});
+        ActionTesting::set_phase(runner, Parallel::Phase::Testing);
+      };
+  const auto check_time_tags =
+      [](ActionTesting::MockRuntimeSystem<MockMetavariables<Dim>>& runner,
```
wthrowe: `const`

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ReceiveElementData.hpp
```diff
@@ -66,10 +66,6 @@ struct ReceiveElementData {
         db::get<Tags::ElementFacesGridCoordinates<Dim>>(box).size();
     const auto& time_step_id = db::get<::Tags::TimeStepId>(box);
     auto& inbox = tuples::get<Tags::SphericalHarmonicsInbox<Dim>>(inboxes);
-    if (inbox.count(time_step_id) == 0 or
-        inbox.at(time_step_id).size() < expected_number_of_senders) {
-      return {Parallel::AlgorithmExecution::Retry, std::nullopt};
-    }
```
wthrowe: Related to an earlier comment: I think these checks are correct and the others should be omitted.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp
```diff
@@ -0,0 +1,90 @@
...
+  template <typename DbTagsList, typename... InboxTags, typename Metavariables,
+            typename ArrayIndex, typename ActionList,
+            typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTagsList>& box, tuples::TaggedTuple<InboxTags...>& inboxes,
+      const Parallel::GlobalCache<Metavariables>& /*cache*/,
+      const ArrayIndex& /*array_index*/, ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    const auto& time_step_id = db::get<::Tags::TimeStepId>(box);
+    auto& inbox = tuples::get<Tags::SphericalHarmonicsInbox<Dim>>(inboxes);
```
wthrowe: `const`

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp
```diff
@@ -0,0 +1,90 @@
...
+           "Received data from two different time step ids.");
+    const auto inbox_time_step_id = inbox.begin()->first;
+    const auto& inbox_slab = inbox_time_step_id.step_time().slab();
+    // we received data from a time step id with a different slab size,
+    // indicating a change in the global time step
+    if (inbox_time_step_id.step_time().slab() !=
+        time_step_id.step_time().slab()) {
+      ASSERT(inbox_time_step_id.slab_number() == time_step_id.slab_number(),
+             "Detected change in global time step but the new slab number is "
+             "different.");
```
wthrowe: Check the full `inbox_time_step_id == time_step_id` (probably with a comment noting that this ignores the slab).

wthrowe (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp
nikwit: yes, at the moment it's coupled this tightly.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ChangeSlabSize.hpp
```diff
@@ -53,14 +54,16 @@ struct ChangeSlabSize {
...
-      ASSERT(inbox_time_step_id.slab_number() == time_step_id.slab_number(),
-             "Detected change in global time step but the new slab number is "
-             "different.");
       ASSERT(inbox_slab.start() == time_step_id.step_time().slab().start(),
              "The new slab should start at the same time as the old one.");
+      // time_step_id comparison does NOT compare slabs.
+      ASSERT(
+          inbox_time_step_id == time_step_id,
+          "Detected change in global time step id but the new slab number is "
+          "different.");
```
wthrowe, line 66: Update message.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
```diff
@@ -50,7 +50,7 @@ struct MockWorldtubeSingleton {
               db::AddSimpleTags<
                   ::Tags::TimeStepId, ::Tags::Next<::Tags::TimeStepId>,
                   ::Tags::TimeStep, ::Tags::Next<::Tags::TimeStep>,
-                  ::Tags::Time, ::Tags::TimeStepper<TimeStepper>,
+                  ::Tags::TimeStepper<TimeStepper>,
                   Tags::ElementFacesGridCoordinates<Dim>>,
```
wthrowe: `ElementFacesGridCoordinates` is no longer used.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/Test_SendToWorldtube.cpp
```diff
@@ -196,7 +196,6 @@ SPECTRE_TEST_CASE("Unit.CurvedScalarWave.Worldtube.SendToWorldtube", "[Unit]") {
         {element_faces_grid_coords, dummy_time_step_id});
 
     ActionTesting::set_phase(make_not_null(&runner), Parallel::Phase::Testing);
-
```
wthrowe: You're probably going to just drop the relevant commit, but make sure you don't end up with any changes to this file.

wthrowe (changes requested):
Squash, including these last couple changes.

nikwit:
squashed it in, thank you for the review

wthrowe:
You seem to have missed the last couple of changes I requested.

nikwit:
Yes, I think I somehow also managed to delete the previous fixup in some rebase error. This should be fine now.

wthrowe (approved):

PR #4831, opened 2023-03-13
Add observer at extremum

ffoucart:
## Proposed changes

Add an observer to find the extremum of a scalar, as well as the value of other tensors at that extremum

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
+++ b/src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
@@ -1,0 +1,404 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <optional>
+#include <pup.h>
+#include <string>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/ObservationBox.hpp"
+#include "DataStructures/DataBox/TagName.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Domain/Tags.hpp"
+#include "IO/Observer/GetSectionObservationKey.hpp"
+#include "IO/Observer/Helpers.hpp"
+#include "IO/Observer/ObservationId.hpp"
+#include "IO/Observer/ObserverComponent.hpp"
+#include "IO/Observer/ReductionActions.hpp"
+#include "IO/Observer/TypeOfObservation.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Options/Options.hpp"
+#include "Parallel/ArrayIndex.hpp"
+#include "Parallel/CharmPupable.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Invoke.hpp"
+#include "Parallel/Local.hpp"
+#include "Parallel/Reduction.hpp"
+#include "ParallelAlgorithms/Events/Tags.hpp"
+#include "ParallelAlgorithms/EventsAndTriggers/Event.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/Functional.hpp"
+#include "Utilities/Numeric.hpp"
+#include "Utilities/OptionalHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+
+namespace Events {
+/// @{
+/*!
+ * \brief Find the extremum of a Scalar<DataVector> over
+ * all elements, as well as the value of other functions
+ * at the location of that extremum.
+ *
+ *
+ * Here is an example of an input file:
+ *
+ * \snippet Test_ObserveAtExtremum.cpp input_file_examples
+ *
+ * \par Array sections
+ * This event supports sections (see `Parallel::Section`). Set the
+ * `ArraySectionIdTag` template parameter to split up observations into subsets
+ * of elements. The `observers::Tags::ObservationKey<ArraySectionIdTag>` must be
+ * available in the DataBox. It identifies the section and is used as a suffix
+ * for the path in the output file.
+ */
+template <typename ObservationValueTag, typename ObservableTensorTagsList,
+          typename NonTensorComputeTagsList = tmpl::list<>,
+          typename ArraySectionIdTag = void>
+class ObserveAtExtremum;
+
+template <typename ObservationValueTag, typename... ObservableTensorTags,
+          typename... NonTensorComputeTags, typename ArraySectionIdTag>
+class ObserveAtExtremum<ObservationValueTag,
+                        tmpl::list<ObservableTensorTags...>,
+                        tmpl::list<NonTensorComputeTags...>, ArraySectionIdTag>
+    : public Event {
+ private:
+  /// Reduction data will contain the time, and either the maximum
+  /// or the minimum of a function
+  template <typename MinMaxFunctional>
+  using ReductionData = Parallel::ReductionData<
+      // Observation value
+      Parallel::ReductionDatum<double, funcl::AssertEqual<>>,
+      // Maximum of first component of a vector
+      Parallel::ReductionDatum<std::vector<double>, MinMaxFunctional>>;
+
+  /// Information about the scalar whose extremum we search for,
+  /// the type of extremum, and the other tensors to observer at
+  /// that extremum
+  struct ObserveTensors {
+    static constexpr Options::String help = {
+        "The tensor to extremize, and other tensors to observe."};
+
+    struct Name {
+      using type = std::string;
+      static constexpr Options::String help = {
+          "The name of the scalar to extremize."};
+    };
+
+    struct ExtremumType {
+      using type = std::string;
+      static constexpr Options::String help = {
+          "The type of extremum -- either Min or Max."};
+    };
+
+    struct AdditionalData {
+      using type = std::vector<std::string>;
+      static constexpr Options::String help = {
+          "List of other tensors to observe at the extremum"};
+    };
+
+    using options = tmpl::list<Name, ExtremumType, AdditionalData>;
+
+    ObserveTensors() = default;
+
+    ObserveTensors(std::string in_scalar, std::string in_extremum_type,
+                   std::vector<std::string> in_additional_data,
+                   const Options::Context& context = {});
+
+    std::string scalar_name;
+    std::string extremum_type;
+    std::vector<std::string> additional_data{};
+  };
+
+ public:
+  /// The name of the subfile inside the HDF5 file
+  struct SubfileName {
+    using type = std::string;
+    static constexpr Options::String help = {
+        "The name of the subfile inside the HDF5 file without an extension and "
+        "without a preceding '/'."};
+  };
+  /// The scalar to extremize, and other tensors to observe at extremum
+  struct TensorsToObserve {
+    using type = ObserveTensors;
+    static constexpr Options::String help = {
+        "Struct specifying the scalar to extremize, the type of extremum "
+        "and other tensors to observe at that extremum."};
+  };
+
+  explicit ObserveAtExtremum(CkMigrateMessage* msg);
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(ObserveAtExtremum);  // NOLINT
+
+  using options = tmpl::list<SubfileName, TensorsToObserve>;
+
+  static constexpr Options::String help =
+      "Observe extremum of a scalar in the DataBox.\n"
+      "\n"
+      "Writes reduction quantities:\n"
+      " * ObservationValueTag (e.g. Time or IterationId)\n"
+      " * Extremum value of the desired scalar\n"
+      " * Additional data at extremum\n";
+
+  ObserveAtExtremum() = default;
+
+  ObserveAtExtremum(std::string subfile_name,
+                    ObserveTensors observe_tensors);
+
+  using observed_reduction_data_tags = observers::make_reduction_data_tags<
+      tmpl::list<ReductionData<funcl::Max<>>, ReductionData<funcl::Min<>>>>;
+
+  using compute_tags_for_observation_box =
+      tmpl::list<ObservableTensorTags..., NonTensorComputeTags...>;
+
+  using argument_tags = tmpl::list<ObservationValueTag, ::Tags::ObservationBox>;
+
+  template <typename ComputeTagsList, typename DataBoxType,
+            typename Metavariables, size_t VolumeDim,
+            typename ParallelComponent>
+  void operator()(const typename ObservationValueTag::type& observation_value,
+                  const ObservationBox<ComputeTagsList, DataBoxType>& box,
+                  Parallel::GlobalCache<Metavariables>& cache,
+                  const ElementId<VolumeDim>& array_index,
+                  const ParallelComponent* const /*meta*/) const;
+
+  using observation_registration_tags = tmpl::list<::Tags::DataBox>;
+
+  template <typename DbTagsList>
+  std::optional<
+      std::pair<observers::TypeOfObservation, observers::ObservationKey>>
+  get_observation_type_and_key_for_registration(
+      const db::DataBox<DbTagsList>& box) const {
+    const std::optional<std::string> section_observation_key =
+        observers::get_section_observation_key<ArraySectionIdTag>(box);
+    if (not section_observation_key.has_value()) {
+      return std::nullopt;
+    }
+    return {{observers::TypeOfObservation::Reduction,
+             observers::ObservationKey(
+                 subfile_path_ + section_observation_key.value() + ".dat")}};
+  }
+
+  using is_ready_argument_tags = tmpl::list<>;
+
+  template <typename Metavariables, typename ArrayIndex, typename Component>
+  bool is_ready(Parallel::GlobalCache<Metavariables>& /*cache*/,
+                const ArrayIndex& /*array_index*/,
+                const Component* const /*meta*/) const {
+    return true;
+  }
+
+  bool needs_evolved_variables() const override { return true; }
+
+  // NOLINTNEXTLINE(google-runtime-references)
+  void pup(PUP::er& p) override;
+
+ private:
+  std::string subfile_path_;
+  std::string scalar_name_;
+  std::string extremum_type_;
+  std::vector<std::string> additional_tensor_names_{};
+};
+/// @}
+
+/// \cond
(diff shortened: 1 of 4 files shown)


ffoucart:
Currently, this only calculate the Max, but adding an option for the Min is fairly easy if we agree on the overall structure.

ffoucart:
@nilsvu @nilsdeppe : Here is an outline of the observer calculating the Max of a scalar (and carrying information about the value of other quantities at that Max). It would be good for me to know whether the overall structure seems to make sense. The test works, but probably does not directly test the reduction operation (not sure -- what does the code do when reducing a single vector?). The legend might also need to be modified, depending on what the output conventions are.

ffoucart:
Second version also has minimum (kind of... the code returns minus the minimum, to limit complexity of the functional called by the reduction operation in this first draft; not sure how the code would react to being give a vector of size 0 otherwise...)

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+#include "Utilities/Functional.hpp"
+#include "Utilities/Numeric.hpp"
+#include "Utilities/OptionalHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace detail {
+
+// Functional called by the reduction opearation: given 2 vectors,
+// return the vector with the largest first component
+struct MaxOfFirstElement {
```
wthrowe: Move this into a more descriptive namespace, although it is very close to `funcl::Max` so actually you can probably just use that instead.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+      static constexpr Options::String help = {
+          "List of other tensors to observe at the extremum"};
+    };
+
+    using options = tmpl::list<Name, ExtremumType, AdditionalData>;
+
+    ObserveTensors() = default;
+
+    ObserveTensors(const std::string in_scalar,
+                   const std::string in_extremum_type,
```
wthrowe: Remove `const`s from these two.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+                   const std::vector<std::string>& in_additional_data,
+                   const Options::Context& context = {});
+
+    std::string scalar_name, extremum_type;
```
wthrowe: I don't know that we have an actual rule for this, but we generally put one member variable per line.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+    using type = std::string;
+    static constexpr Options::String help = {
+        "The name of the subfile inside the HDF5 file without an extension and "
+        "without a preceding '/'."};
+  };
+  /// The scalar to extremize, and other tensors to observe at extremum
+  struct TensorsToObserve {
+    using type = ObserveTensors;
+    static constexpr Options::String help = {
+        "List specifying The scalar to extremize, the type of extremum "
```
wthrowe: Remove "List specifying"?  I think that's the intent.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+                  tmpl::list<NonTensorComputeTags...>,
+                  ArraySectionIdTag>::ObserveAtExtremum(CkMigrateMessage* msg)
+    : Event(msg) {}
+
+template <typename ObservationValueTag, typename... ObservableTensorTags,
+          typename... NonTensorComputeTags, typename ArraySectionIdTag>
+ObserveAtExtremum<ObservationValueTag, tmpl::list<ObservableTensorTags...>,
+                  tmpl::list<NonTensorComputeTags...>, ArraySectionIdTag>::
+    ObserveAtExtremum(const std::string& subfile_name,
+                      const ObserveTensors& observe_tensors)
```
wthrowe: Take by non-const value.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+    : subfile_path_("/" + subfile_name),
+      scalar_name_(observe_tensors.scalar_name),
+      extremum_type_(observe_tensors.extremum_type) {
```
wthrowe: `std::move` these two.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+  additional_tensor_names_.reserve(observe_tensors.additional_data.size());
+  for (const auto& additional_tensor_name : observe_tensors.additional_data) {
+    additional_tensor_names_.push_back(additional_tensor_name);
+  }
```
wthrowe: Move `observe_tensors.additional_data` into `additional_tensor_names_` instead of copying the elements.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+}
+
+template <typename ObservationValueTag, typename... ObservableTensorTags,
+          typename... NonTensorComputeTags, typename ArraySectionIdTag>
+ObserveAtExtremum<ObservationValueTag, tmpl::list<ObservableTensorTags...>,
+                  tmpl::list<NonTensorComputeTags...>, ArraySectionIdTag>::
+    ObserveTensors::ObserveTensors(
+        const std::string in_scalar, const std::string in_extremum_type,
```
wthrowe: Remove `const`s from these two.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+        const std::vector<std::string>& in_additional_data,
```
wthrowe: Change to non-const non-reference.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+      observers::get_section_observation_key<ArraySectionIdTag>(box);
+  if (not section_observation_key.has_value()) {
+    return;
+  }
+
+  using tensor_tags = tmpl::list<ObservableTensorTags...>;
+
+  // List that will contain the local extremum, and the value
+  // of other tensors at that extremum
+  std::list<double> data_to_reduce_list{};
```
wthrowe: Don't use `std::list`.  Just use a `std::vector`.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+  // List containing a description of the data to be reduced.
+  std::list<std::string> legend_list{db::tag_name<ObservationValueTag>()};
```
wthrowe: Again, no list.

tests/Unit/ParallelAlgorithms/Events/Test_ObserveAtExtremum.cpp
```diff
@@ -0,0 +1,288 @@
...
+          ObserveAtExtremumEvent<TestSectionIdTag>{
+              "reduction0", {"Var0", "Min", {"Var0", "Var1"}}}),
+      "Min", Spectral::Basis::Legendre, Spectral::Quadrature::GaussLobatto,
+      "Section0");
+
+  INFO("create/serialize");
+  Parallel::register_factory_classes_with_charm<Metavariables<3, void>>();
+  const auto factory_event = TestHelpers::test_creation<std::unique_ptr<Event>,
+                                                        Metavariables<3, void>>(
+      // [input_file_examples]
```
wthrowe, line 267: You don't use this example anywhere.  Either add it to the docs or remove the label comments.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,412 @@
...
+template <typename ObservationValueTag, typename... ObservableTensorTags,
+          typename... NonTensorComputeTags, typename ArraySectionIdTag>
+class ObserveAtExtremum<ObservationValueTag,
+                        tmpl::list<ObservableTensorTags...>,
+                        tmpl::list<NonTensorComputeTags...>, ArraySectionIdTag>
+    : public Event {
+ private:
+  /// Reduction data will contain the time, and either the maximum
+  /// or the minimum of a function
+  using ReductionData = Parallel::ReductionData<
```
wthrowe, line 76: [optional] If you want to avoid the "negative of min" thing, I think you could template this on the functor (`funcl::Min` or `funcl::Max`) and choose which to send at runtime.

wthrowe (changes requested):

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
ffoucart: Is this what funcl::Max would do when given a std::vector? I want the decision of which vector to keep to be solely based on the first element, carrying the other element along (I'm not sure what the behavior of funcl::Max is in that case)

wthrowe: `funcl::Max` wraps `std::max` (the two-argument version), so it relies on the comparison operators for `std::vector`.  Those do a lexicographical comparison, which is the same as just comparing the first element unless there's a tie.

ffoucart: Ok, great, should make this easier.

tests/Unit/ParallelAlgorithms/Events/Test_ObserveAtExtremum.cpp
ffoucart: I though this was used by the surrounding code to create an event (and I did get some error from those input files at some point). Isn't this the same as the input file examples in Test_ObserveNorms?

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
ffoucart: I tried to do this, let me kwow if that's what you had in mind.

ffoucart:
Except for the second to last comment, everything should be updated according to your comments, @wthrowe

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
wthrowe: That is exactly what I was thinking.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -160,11 +144,11 @@ class ObserveAtExtremum<ObservationValueTag,
...
 
-  ObserveAtExtremum(const std::string& subfile_name,
-                    const ObserveTensors& observe_tensors);
+  ObserveAtExtremum(std::string subfile_name,
+                    ObserveTensors observe_tensors);
 
-  using observed_reduction_data_tags =
-      observers::make_reduction_data_tags<tmpl::list<ReductionData>>;
+  // using observed_reduction_data_tags =
+  //     observers::make_reduction_data_tags<tmpl::list<ReductionData>>;
```
wthrowe: I'm not entirely sure what this alias is used for, but I think it's for registering the possible reductions with charm or our wrappers or something.  The argument should probably be `tmpl::list<ReductionData<funcl::Max<>>, ReductionData<funcl::Min<>>>`.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -354,40 +332,54 @@ operator()(const typename ObservationValueTag::type& observation_value,
...
       }
   });
 
   // Transforms from list to vector for reduction operation
-  std::vector<double> data_to_reduce(data_to_reduce_list.begin(),
-                                     data_to_reduce_list.end());
-  std::vector<std::string> legend(legend_list.begin(), legend_list.end());
+  // std::vector<double> data_to_reduce(data_to_reduce_list.begin(),
+  //                                  data_to_reduce_list.end());
+  // std::vector<std::string> legend(legend_list.begin(), legend_list.end());
```
wthrowe: The change is what I wanted.  You can remove this.

tests/Unit/ParallelAlgorithms/Events/Test_ObserveAtExtremum.cpp
wthrowe: The `test_creation` function call is good, but as-is you're not using the `// [input_file_examples]` comments.  Those are used in documentation to put example code snippets in the docs.  For example, in `ObserveNorms.hpp` in the class documentation there is
```
 * Here is an example of an input file:
 *
 * \snippet Test_ObserveNorms.cpp input_file_examples
```
to match the `// [input_file_examples]` comments in the test.  You can either add a similar "snippet" line to your docs or remove the comments here.

wthrowe (changes requested):
Go ahead and squash, including these last few requests.

tests/Unit/ParallelAlgorithms/Events/Test_ObserveAtExtremum.cpp
ffoucart: Ah, got it; easy enough to add that to the documentation then.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
ffoucart: Ok, that makes sense from what I had read; I guess that part will only be tested once we actually plug this observer in a simulation doing a real reduction.

wthrowe (dismissed):

tests/Unit/ParallelAlgorithms/Events/Test_ObserveAtExtremum.cpp
```diff
@@ -0,0 +1,288 @@
...
+  auto serialized_event = serialize_and_deserialize(factory_event);
+  test<void>(std::move(serialized_event), "Max", Spectral::Basis::Legendre,
+             Spectral::Quadrature::GaussLobatto, std::nullopt);
+
+  test<void>(std::make_unique<ObserveAtExtremumEvent<void>>(
+                 ObserveAtExtremumEvent<void>{
+                     "reduction0", {"Var0", "Max", {"Var0", "Var1"}}}),
+             "Max", Spectral::Basis::FiniteDifference,
+             Spectral::Quadrature::CellCentered, std::nullopt);
+}
```
nilsdeppe, line 317: [picking random line] Can you add a test that creates two `ReductionData` like the ones used in the event and then do `reduction_data1.combine(std::move(reduction_data2));` and check that the resulting values are what you expect? Basically this is to make sure the part that we can't easily test with the mock runtime system (reductions) is at least somewhat tested.

ffoucart: Let me know if the last commit covers what you wanted.

nilsdeppe: Yep, that's perfect, thanks! Please go ahead and squash :)

nilsdeppe (approved):

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,404 @@
...
+                                   std::vector<std::string> in_additional_data,
+                                   const Options::Context& context)
+    : scalar_name(std::move(in_scalar)),
+      extremum_type(std::move(in_extremum_type)),
+      additional_data(std::move(in_additional_data)) {
+  if (((scalar_name != db::tag_name<ObservableTensorTags>()) and ...)) {
+    PARSE_ERROR(
+        context, "Tensor '"
+                     << scalar_name << "' is not known. Known tensors are: "
+                     << ((db::tag_name<ObservableTensorTags>() + ",") + ...));
```
nilsvu, line 245: (optional) I think it would be good to check if the tensor is a scalar already here at option parsing, so you don't have to wait for the first observation to get the error (which may not happen until some time into the evolution). If you agree, here's a suggestion:

I think the easiest way to accomplish this is to loop over the `ObservableTensorTags` with a `tmpl::for_each` until you hit the selected tensor, then check `tag::type::rank() == 0`. To deal with optionals you can use `if constexpr (tt::is_a_v<std::optional, tag::type>)`. If it makes things easier I think it would be fine to disallow optionals for the selected scalar (since optionals are used for errors to the analytic solution but you probably want to observe extrema in the density etc), but you can also just get the optional tensor type from `tag::type::value_type`.

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,404 @@
...
+            (extremum_type_ == "Min" and
+             (components[0][i] < components[0][index_of_extremum]))) {
+          index_of_extremum = i;
+        }
+      }
+      data_to_reduce.push_back(components[0][index_of_extremum]);
+      if (extremum_type_ == "Max") {
+        legend.push_back("Max of " + scalar_name_);
+      } else {
+        legend.push_back("Min of " + scalar_name_);
```
nilsvu, line 319: (optional) For consistency with `ObserveNorms` I suggest to name this `Max(scalar_name)` and `Min(scalar_name)`

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -0,0 +1,404 @@
...
+        const auto [component_names, components] = tensor.get_vector_of_data();
+        for (size_t j = 0; j < components.size(); j++) {
+          data_to_reduce.push_back(components[j][index_of_extremum]);
+          if (components.size() > 1) {
+            legend.push_back("Value of " + tensor_name + "_" +
+                             component_names[j] + " at " + extremum_type_ +
+                             " of " + scalar_name_);
+          } else {
+            legend.push_back("Value of " + tensor_name + " at " +
+                             extremum_type_ + " of " + scalar_name_);
```
nilsvu, line 347: (optional) Also here I suggest something like `At{scalar_name}Min({tensor_name})`.

ffoucart:
@nilsvu : I'm good with those proposed changes -- if you think it's all optional, do you mind if I do this in a separate PR, to make it easier for Patrick to test the coupling of the observer to an evolution in the mean time?

ffoucart:
(I do have a PR ready to go with those changes now; or at least one that works with clang on wheeler)

PR #4834, opened 2023-03-14
Add `CheckInputFile` tag

nikwit:
## Proposed changes
Adds an option creatable dummy tag that checks whether the input file specifies a circular orbit with the correct angular velocity and a background Kerr-Schild spacetime that is centered, non-spinning with mass 1.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
@@ -16,14 +16,18 @@
 #include "Domain/Creators/DomainCreator.hpp"
 #include "Domain/Creators/OptionTags.hpp"
 #include "Domain/Domain.hpp"
+#include "Domain/FunctionsOfTime/QuaternionFunctionOfTime.hpp"
 #include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/ElementId.hpp"
 #include "Domain/Structure/ExcisionSphere.hpp"
 #include "Domain/Tags.hpp"
+#include "Evolution/Systems/CurvedScalarWave/BackgroundSpacetime.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
 #include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
 #include "Options/Options.hpp"
+#include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp"
 #include "Time/Tags.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/Gsl.hpp"
 
 namespace CurvedScalarWave::Worldtube {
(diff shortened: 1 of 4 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -65,6 +67,51 @@ struct ExpansionOrder {
...
+    const double orbital_radius = get<0>(excision_sphere.center());
+    const auto& functions_of_time = domain_creator->functions_of_time();
+    if (not functions_of_time.count("Rotation")) {
+      ERROR("Expected functions of time to contain 'Rotation'.");
+    }
+    // dynamic cast to access `angle_func_and_deriv` method
+    const auto* rotation_function_of_time =
+        dynamic_cast<domain::FunctionsOfTime::QuaternionFunctionOfTime<3>*>(
+            &*functions_of_time.at("Rotation"));
+    ASSERT(rotation_function_of_time,
```
knelli2: Two things:

1. Why an ASSERT and not an error here? Things below will fail (in release mode) if this isn't a QuatFoT.
2. `rotation_function_of_time != nullptr` to be explicit

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -65,6 +67,51 @@ struct ExpansionOrder {
...
+           "Failed dynamic cast to QuaternionFunctionOfTime.");
+    const auto angular_velocity =
+        rotation_function_of_time->angle_func_and_deriv(0.).at(1);
+    if (equal_within_roundoff(orbital_radius, 0.)) {
+      ERROR("The orbital radius was set to 0.");
+    }
+    if (not(equal_within_roundoff(angular_velocity.at(0), 0.) and
+            equal_within_roundoff(angular_velocity.at(1), 0.) and
+            equal_within_roundoff(angular_velocity.at(2),
+                                  pow(orbital_radius, -1.5)))) {
```
knelli2: `equal_within_roundoff` can take iterables and `angular_velocity` is a DataVector so I think this can just be

```cpp
if (not equal_within_roundoff(angular_velocity, DataVector{0.0, 0.0, pow(orbital_radius, -1.5)})) {
```

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -65,6 +67,51 @@ struct ExpansionOrder {
...
+      ERROR(
+          "Only circular orbits are implemented at them moment so the "
+          "angular velocity should be [0., 0., orbital_radius^(-3/2)] = "
+          << "[0., 0., " << pow(orbital_radius, -1.5) << "]");
```
knelli2, line 128: I know this isn't actually used in a computation loop or anything, but try to avoid `pow`. (and above)

knelli2 (changes requested):
You can squash directly

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: I'd argue that in situations like this where the performance difference is virtually unmeasurable readability is more important.

nikwit:
thanks for your review, I squashed it in

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -65,6 +67,51 @@ struct ExpansionOrder {
...
+    const double orbital_radius = get<0>(excision_sphere.center());
+    const auto& functions_of_time = domain_creator->functions_of_time();
+    if (not functions_of_time.count("Rotation")) {
+      ERROR("Expected functions of time to contain 'Rotation'.");
+    }
+    // dynamic cast to access `angle_func_and_deriv` method
+    const auto* rotation_function_of_time =
+        dynamic_cast<domain::FunctionsOfTime::QuaternionFunctionOfTime<3>*>(
+            &*functions_of_time.at("Rotation"));
+    if (rotation_function_of_time != nullptr) {
```
knelli2: Oops, my comment above was for the ASSERT statement. I think now with the ERROR, this should be

```cpp
if (rotation_function_of_time == nullptr) { ERROR(); }
```

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
knelli2: Ok. Because it won't impact performance, I'm indifferent.

knelli2 (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -65,6 +67,51 @@ struct ExpansionOrder {
...
+      ERROR("Failed dynamic cast to QuaternionFunctionOfTime.");
+    }
+    const auto angular_velocity =
+        rotation_function_of_time->angle_func_and_deriv(0.).at(1);
+    if (equal_within_roundoff(orbital_radius, 0.)) {
+      ERROR("The orbital radius was set to 0.");
+    }
+    if (not equal_within_roundoff(
+            angular_velocity,
+            DataVector{0.0, 0.0, pow(orbital_radius, -1.5)})) {
```
wthrowe, line 124: This appears to have dimensional issues.  Angular velocity does not have units of length^(-3/2).

wthrowe (changes requested):
Not a full review.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
knelli2: Technically, I think there is a $\sqrt{\rm mass}$ missing from the numerator (because $G = 1$). However, if we assume $M = 1$, then I think this is ok. Not sure what @nikwit assumes though.

wthrowe: G=c=1 is fine, but I believe mass is an input file parameter.

wthrowe: Actually, it would be good to check all the properties of the background, like the spin.  Is the background accessible from here?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: yes, I just didn't pay attention, my bad

nikwit:
Added a fixup that checks the background spacetime of the input file

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
knelli2: Ok now I believe the units work out because it's required that $M = 1$

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_Tags.cpp
```diff
@@ -370,6 +371,58 @@ void test_puncture_field() {
...
+  }
+  {
+    const gr::Solutions::KerrSchild kerr_schild_spinning(
+        1., make_array(0.1, 0., 0.), make_array(0., 0., 0.));
+    CHECK_THROWS_WITH(
+        (Tags::CheckInputFile<3, gr::Solutions::KerrSchild>::
+             create_from_options(bbh_correct, "ExcisionSphere",
+                                 kerr_schild_spinning)),
+        Catch::Matchers::Contains("Black hole spin is not implemented yet but "
+                                  "you requested non-zero spin."));
```
knelli2, line 422: Can you also add an error test for unit mass? I don't believe you have one here.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -65,6 +69,68 @@ struct ExpansionOrder {
...
+      ERROR(
+          "Only circular orbits are implemented at them moment so the "
```
knelli2: `at them moment` -> `at the moment` (I missed this before)

knelli2 (changes requested):
Ok fixup looks good. Squash these in directly

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
wthrowe: Yep.  Looks good now.

wthrowe (commented):
I have now done a full review and have no additional comments.

nikwit:
I squashed in your suggestions, thank you for the reviews :)

knelli2 (approved):

wthrowe (approved):

PR #4836, opened 2023-03-14
Add Peter James Nee as a contributor.

PunJustice:
"Peter James" is technically my first name (and how I appear on papers), if this is too annoying I suppose it can be shortened.


wthrowe (approved):
I think you're the authority on how your name should be listed.

nilsvu (approved):

PR #4837, opened 2023-03-14
Adaptive order derivatives in GRMHD

nilsdeppe:
## Proposed changes

This is the last bit of code I have for high & adaptive order FD derivatives. It'll still need to get added to GH+GRMHD, but that doesn't need to happen immediately.

The only major thing that doesn't work yet is high-order FD with DG-FD. I'll get that working next since that I think is just sending the right data in the right places.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DgSubcell/SubcellOptions.cpp
+++ b/src/Evolution/DgSubcell/SubcellOptions.cpp
@@ -28,8 +28,7 @@ SubcellOptions::SubcellOptions(
     bool always_use_subcells, fd::ReconstructionMethod recons_method,
     bool use_halo,
     std::optional<std::vector<std::string>> only_dg_block_and_group_names,
-    std::optional<size_t> finite_difference_derivative_order,
-    const Options::Context& context)
+    ::fd::DerivativeOrder finite_difference_derivative_order)
     : initial_data_rdmp_delta0_(initial_data_rdmp_delta0),
       initial_data_rdmp_epsilon_(initial_data_rdmp_epsilon),
       rdmp_delta0_(rdmp_delta0),
@@ -43,14 +42,6 @@ SubcellOptions::SubcellOptions(
       finite_difference_derivative_order_(finite_difference_derivative_order) {
   if (not only_dg_block_and_group_names_.has_value()) {
     only_dg_block_ids_ = std::vector<size_t>{};
-  }
-  if (finite_difference_derivative_order_.has_value() and
-      not alg::found(std::array<size_t, 5>{{2, 4, 6, 8, 10}},
-                     finite_difference_derivative_order_.value())) {
-    PARSE_ERROR(context,
-                "The finite difference order must be Auto or one of "
-                "2,4,6,8,10, but got "
-                    << finite_difference_derivative_order_.value());
   }
 }
 
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Tag.hpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Tag.hpp
@@ -40,7 +40,8 @@ struct Reconstructor : db::SimpleTag {
   static type create_from_options(
       const type& reconstructor,
       const ::evolution::dg::subcell::SubcellOptions& subcell_options) {
-    if (not subcell_options.finite_difference_derivative_order().has_value() and
+    if (static_cast<int>(subcell_options.finite_difference_derivative_order()) <
+            0 and
         not reconstructor->supports_adaptive_order()) {
       ERROR_NO_TRACE(
           "Cannot use adaptive finite difference derivative order with "
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Tag.hpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Tag.hpp
@@ -47,6 +47,21 @@ struct Reconstructor : db::SimpleTag {
           "Cannot use adaptive finite difference derivative order with "
           "specified reconstructor.");
     }
+    if ((static_cast<int>(
+             subcell_options.finite_difference_derivative_order()) /
+             2 -
+         1) > static_cast<int>(reconstructor->ghost_zone_size())) {
+      ERROR_NO_TRACE(
+          "The derivative order chosen ("
+          << subcell_options.finite_difference_derivative_order()
+          << ") requires more ghost zones ("
+          << (static_cast<int>(
+                  subcell_options.finite_difference_derivative_order()) /
+                  2 -
+              1)
+          << ") than the reconstructor sends, "
+          << reconstructor->ghost_zone_size());
+    }
     return reconstructor->get_clone();
   }
 };
--- a/tests/Unit/Evolution/DgSubcell/Test_CellCenteredFlux.cpp
+++ b/tests/Unit/Evolution/DgSubcell/Test_CellCenteredFlux.cpp
@@ -55,7 +55,7 @@ struct Fluxes {
 };
 
 template <size_t Dim, bool ComputeOnlyOnRollback>
-void test(const size_t derivative_order, const bool did_rollback) {
+void test(const fd::DerivativeOrder derivative_order, const bool did_rollback) {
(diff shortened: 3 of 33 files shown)


src/Evolution/DgSubcell/SubcellOptions.cpp
```diff
@@ -44,14 +43,6 @@ SubcellOptions::SubcellOptions(
...
   }
-  if (finite_difference_derivative_order_.has_value() and
-      not alg::found(std::array<size_t, 5>{{2, 4, 6, 8, 10}},
-                     finite_difference_derivative_order_.value())) {
-    PARSE_ERROR(context,
-                "The finite difference order must be Auto or one of "
-                "2,4,6,8,10, but got "
-                    << finite_difference_derivative_order_.value());
-  }
 }
```
pajkosmi, line 47: Should there still be a check that `finite_difference_derivative_order` has a value?  Or by removing std::optional, is that already implied?

src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Tag.hpp
```diff
@@ -40,12 +40,28 @@ struct Reconstructor : db::SimpleTag {
...
+            0 and
         not reconstructor->supports_adaptive_order()) {
       ERROR_NO_TRACE(
           "Cannot use adaptive finite difference derivative order with "
           "specified reconstructor.");
     }
+    if ((static_cast<int>(
+             subcell_options.finite_difference_derivative_order()) /
+             2 -
+         1) > static_cast<int>(reconstructor->ghost_zone_size())) {
```
pajkosmi, line 53: Remind me why the `if` condition isn't `(static_cast<int>(subcell_options.finite_difference_derivative_order()) - 1) / 2 > static_cast<int>(reconstructor->ghost_zone_size()`?  The logic in the original commit would only check for 1 ghost zone, for a 5th order WENOZ stencil when it should be 2.  Are the even order finite difference stencils different?

src/Evolution/DgSubcell/SubcellOptions.cpp
nilsdeppe: This is an excellent question! Since it's no longer a `std::optional` it can't have a `None` value. Since it's an `enum` the parsing is actually done where the enum is defined: https://github.com/sxs-collaboration/spectre/blob/84d6a6b9e1d1fbb3f501e515a990927cc3ab2554/src/NumericalAlgorithms/FiniteDifference/DerivativeOrder.cpp#L38-L66

tests/Unit/Evolution/DgSubcell/Test_CellCenteredFlux.cpp
```diff
@@ -105,7 +104,10 @@ void test(const size_t derivative_order, const bool did_rollback) {
 
 SPECTRE_TEST_CASE("Unit.Evolution.Subcell.CellCenteredFlux",
                   "[Evolution][Unit]") {
-  for (const size_t derivative_order : {0_st, 2_st, 4_st}) {
+  using DO = fd::DerivativeOrder;
+  for (const fd::DerivativeOrder derivative_order :
```
pajkosmi: Could use `for (const DO derivative_order : ...)`.

tests/Unit/Evolution/Systems/GrMhd/ValenciaDivClean/Subcell/Test_TimeDerivative.cpp
```diff
@@ -368,32 +369,61 @@ std::array<double, 5> test(const size_t num_dg_pts,
...
     "Unit.Evolution.Systems.ValenciaDivClean.Subcell.TimeDerivative",
     "[Unit][Evolution]") {
-  using sos = std::optional<size_t>;
   std::optional<std::array<double, 5>> previous_error_5{};
   std::optional<std::array<double, 5>> previous_error_6{};
-  for (const sos fd_do : {sos{}, sos{2}, sos{4}, sos{6}, sos{8}, sos{10}}) {
+  std::array<double, 5> second_order_error_5{};
+  std::array<double, 5> second_order_error_6{};
+  using DO = ::fd::DerivativeOrder;
+  for (const ::fd::DerivativeOrder fd_do :
```
pajkosmi: Same as above.  `for (const DO fd_do : ...)`.

tests/Unit/Evolution/Systems/GrMhd/ValenciaDivClean/Subcell/Test_TimeDerivative.cpp
```diff
@@ -368,32 +369,61 @@ std::array<double, 5> test(const size_t num_dg_pts,
...
-      if (previous_error_6.has_value() and fd_do.value_or(2) > 2) {
+      if (previous_error_6.has_value()) {
         CHECK(gsl::at(six_pts_data, i) < gsl::at(previous_error_6.value(), i));
       }
     }
     previous_error_5 = five_pts_data;
     previous_error_6 = six_pts_data;
   }
+
+  // Check the adaptive correction order works.
```
pajkosmi, line 405: Some code duplication below.  Would structured bindings help?
```
{
  for (const auto& recon_order : make_array(DO::OneHigherThanRecons, DO::OneHigherThanReconsButFiveToFour)) {
    CAPTURE(recon_order);
    const auto five_pts_data = test(5, recon_order);
    const auto six_pts_data = test(6, recon_order);
    for (size_t i = 0; i < five_pts_data.size(); ++i) {
      CAPTURE(i);
      CHECK(gsl::at(six_pts_data, i) < gsl::at(five_pts_data, i));
      CHECK(gsl::at(five_pts_data, i) < gsl::at(second_order_error_5, i));
      CHECK(gsl::at(six_pts_data, i) < gsl::at(second_order_error_6, i));
    }
  }
}
```

pajkosmi (commented):
Only small changes that would streamline the code, along with some questions for my clarification.

src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Tag.hpp
nilsdeppe: Ah! Excellent question. For reconstruction we have:
```
o o | x x x
  ^   ^ ^ ^
``` 
where every point that has an `^` needs reconstruction, `o` is neighbor and `x` is ourselves. We need to do reconstruction in one neighbor element to get the neighbor's face solution on the boundary. For FD derivatives we just need:
```
o o | x x x
      ^ ^ ^
```
so we don't need to take the numerical derivative in our neighbor's cell, unlike reconstruction. This is why the FD derivatives are one stencil more compact that the reconstruction.

Does that answer your question?

pajkosmi: Yes it does; I was conflating the reconstruction and FD order.  So the FD order is for calculating spatial derivatives of quantities (e.g., the metric variables) to be used in (e.g.,) source terms, whereas reconstruction stencil/order (e.g., WENOZ) is specified through the reconstructions scheme?

nilsdeppe: Almost/mostly! This controls the order of the `d_i F^i` FD derivative. Currently the metric terms always use the highest derivative possible based on the FD reconstruction stencil size. In practice that means for 2nd-order reconstruction you do 4th order FD derivatives for the metric, 5th order recons gives 6th-order metric, etc.

Compared to FV, this would be controlling the order of the Gaussian quadrature used to compute `int n_i F^i`. In FD you change the FD derivative order.

Does that make sense?

nilsdeppe:
I've rebased and pushed fixup commits. Thanks for the review!

pajkosmi:
> I've rebased and pushed fixup commits. Thanks for the review!

@nilsdeppe looks good; you can squash it.

nilsdeppe:
Rebased and squashed! Thanks!

pajkosmi:
I approve this for merge, @sxs-collaboration/spectre-core-devs

kidder (approved):

PR #4838, opened 2023-03-15
Increase test timeout for gcc-10 Debug build

nilsvu:
It times out consistently. While we should speed up the failing tests as well, this configuration seems to run slower than newer gcc or clang builds, so we increase the timeout a bit.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (dismissed):

nilsdeppe:
Doesn't pass CMake :(

nilsvu:
So CMake has trouble with decimal numbers? Well I just set the timeout to an integer then 🤷‍♂️

nilsdeppe (approved):

PR #4839, opened 2023-03-15
Allow specifying single RadialDist in Sphere

knelli2:
## Proposed changes

When I have a lot of partitions, it got annoying to specify the distribution for each when they were all the same. This allows you to just specify one distribution and it'll be used for all the partitions.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/Sphere.cpp
+++ b/src/Domain/Creators/Sphere.cpp
@@ -49,6 +49,24 @@ Excision::Excision(
     : boundary_condition(std::move(local_boundary_condition)) {}
 }  // namespace detail
 
+namespace {
+struct DistributionVisitor {
+  size_t num_shells;
+
+  std::vector<domain::CoordinateMaps::Distribution> operator()(
+      const domain::CoordinateMaps::Distribution distribution) const {
+    return std::vector<domain::CoordinateMaps::Distribution>(num_shells,
+                                                             distribution);
+  }
+
+  std::vector<domain::CoordinateMaps::Distribution> operator()(
+      const std::vector<domain::CoordinateMaps::Distribution>& distributions)
+      const {
+    return distributions;
+  }
+};
+}  // namespace
+
 Sphere::Sphere(
     double inner_radius, double outer_radius,
     std::variant<Excision, InnerCube> interior,
@@ -57,7 +75,7 @@ Sphere::Sphere(
     bool use_equiangular_map,
     std::optional<EquatorialCompressionOptions> equatorial_compression,
     std::vector<double> radial_partitioning,
-    std::vector<domain::CoordinateMaps::Distribution> radial_distribution,
+    const typename RadialDistribution::type& radial_distribution,
     ShellWedges which_wedges,
     std::unique_ptr<domain::creators::time_dependence::TimeDependence<3>>
         time_dependence,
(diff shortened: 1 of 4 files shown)


src/Domain/Creators/Sphere.cpp
```diff
@@ -47,6 +47,22 @@ Excision::Excision(
     std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
         local_boundary_condition)
     : boundary_condition(std::move(local_boundary_condition)) {}
+
+struct DistributionVisitor {
```
wthrowe, line 53: Anonymous namespace.

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -492,36 +493,41 @@ void test_sphere() {
...
            creators::Sphere::EquatorialCompressionOptions{0.5, 2}}};
-  const std::array<std::vector<double>, 2> radial_partitioning{
-      {{}, {0.5 * (inner_radius + outer_radius)}}};
-  const std::array<std::vector<domain::CoordinateMaps::Distribution>, 2>
+  const double outer_minus_inner = outer_radius - inner_radius;
+  const std::array<std::vector<double>, 3> radial_partitioning{
+      {{},
+       {0.5 * (inner_radius + outer_radius)},
+       {inner_radius + 0.333 * outer_minus_inner,
+        inner_radius + 0.666 * outer_minus_inner}}};
```
wthrowe: If these are supposed to be thirds make them actually thirds, otherwise make them not look like thirds.

wthrowe (changes requested):

src/Domain/Creators/Sphere.cpp
knelli2: An anonymous namespace inside a detail namespace??

wthrowe: I don't care if you leave it in the detail namespace or move it out.

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -492,36 +493,41 @@ void test_sphere() {
...
            cartesian_product(
-               make_array(0_st, 1_st, 2_st), make_array(false, true),
-               equatorial_compressions, make_array(0.0, 1.0),
+               make_array(false, true), equatorial_compressions,
+               make_array(0_st, 1_st, 2_st),
                make_array(ShellWedges::All, ShellWedges::FourOnEquator,
                           ShellWedges::OneAlongMinusX),
                make_array(true, false), make_array(true, false)),
            make_not_null(&gen))) {
-    const auto& interior = interiors[interior_index];
```
nilsvu: Don't couple the interior selection to the radial distribution selection. I don't know why `array_index` was a double here, but just make it three size_ts.

tests/Unit/Domain/Structure/Test_ExcisionSphere.cpp
```diff
@@ -101,8 +101,8 @@ void test_abutting_direction_shell() {
         true,
         std::nullopt,
         {2.},
-        {domain::CoordinateMaps::Distribution::Linear,
-         domain::CoordinateMaps::Distribution::Linear}};
```
nilsvu, line 105: (nitpick) you can remove the outer braces

nilsvu (changes requested):

tests/Unit/Domain/Structure/Test_ExcisionSphere.cpp
knelli2: Decided to leave this as is. Personally I like seeing the outer braces because then I know this function doesn't actually take a vector as an argument; it takes something else (even if that something is constructible from just the vector)

nilsvu: Ok

nilsvu (approved):

wthrowe (approved):

PR #4840, opened 2023-03-16
Allow the use of simple cartesian grid around compact objects

ffoucart:
## Proposed changes

Add an option to the BinaryCompactObject domain generator to replace the inner 24 blocks covering the two compact objects and their immediate neighborhoods with 2 cartesian cubes.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/BinaryCompactObject.cpp
+++ b/src/Domain/Creators/BinaryCompactObject.cpp
@@ -60,8 +60,8 @@ bool BinaryCompactObject::Object::is_excised() const {
 
 // Time-independent constructor
 BinaryCompactObject::BinaryCompactObject(
-    Object object_A, Object object_B, const double envelope_radius,
-    const double outer_radius,
+    typename ObjectA::type object_A, typename ObjectB::type object_B,
+    const double envelope_radius, const double outer_radius,
     const typename InitialRefinement::type& initial_refinement,
     const typename InitialGridPoints::type& initial_number_of_grid_points,
     const bool use_equiangular_map, const bool use_projective_map,
@@ -77,9 +77,24 @@ BinaryCompactObject::BinaryCompactObject(
       use_projective_map_(use_projective_map),
       radial_distribution_outer_shell_(radial_distribution_outer_shell),
       outer_boundary_condition_(std::move(outer_boundary_condition)) {
+  // Get useful information about the type of grid used around each compact
+  // object
+  x_coord_a_ =
+      std::visit([](const auto& arg) { return arg.x_coord; }, object_A_);
+  x_coord_b_ =
+      std::visit([](const auto& arg) { return arg.x_coord; }, object_B_);
+  is_excised_a_ =
+      std::visit([](const auto& arg) { return arg.is_excised(); }, object_A_);
+  is_excised_b_ =
+      std::visit([](const auto& arg) { return arg.is_excised(); }, object_B_);
+  use_single_block_a_ =
+      std::holds_alternative<CartesianCubeAtXCoord>(object_A_);
+  use_single_block_b_ =
+      std::holds_alternative<CartesianCubeAtXCoord>(object_B_);
+
   // Determination of parameters for domain construction:
-  translation_ = 0.5 * (object_B_.x_coord + object_A_.x_coord);
-  length_inner_cube_ = abs(object_A_.x_coord - object_B_.x_coord);
+  translation_ = 0.5 * (x_coord_a_ + x_coord_b_);
+  length_inner_cube_ = abs(x_coord_a_ - x_coord_b_);
   length_outer_cube_ = 2.0 * envelope_radius_ / sqrt(3.0);
   if (use_projective_map_) {
     projective_scale_factor_ = length_inner_cube_ / length_outer_cube_;
@@ -92,19 +107,27 @@ BinaryCompactObject::BinaryCompactObject(
   // The envelope and outer shell have another 10 blocks each.
   number_of_blocks_ = 44;
   // For each object whose interior is not excised, add 1 block
-  if (not object_A_.is_excised()) {
+  if ((not use_single_block_a_) and (not is_excised_a_)) {
     number_of_blocks_++;
   }
-  if (not object_B_.is_excised()) {
+  if ((not use_single_block_b_) and (not is_excised_b_)) {
     number_of_blocks_++;
   }
 
-  if (object_A_.x_coord <= 0.0) {
+  // For each of the object replaced by a single block, remove (12-1)=11
+  if (use_single_block_a_) {
+    number_of_blocks_ -= 11;
+  }
+  if (use_single_block_b_) {
+    number_of_blocks_ -= 11;
+  }
+
+  if (x_coord_a_ <= 0.0) {
     PARSE_ERROR(
         context,
         "The x-coordinate of ObjectA's center is expected to be positive.");
   }
-  if (object_B_.x_coord >= 0.0) {
+  if (x_coord_b_ >= 0.0) {
     PARSE_ERROR(
         context,
         "The x-coordinate of ObjectB's center is expected to be negative.");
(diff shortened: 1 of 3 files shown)


ffoucart:
Rough outline for discussion at this point; I have not even fixed any call to the creator of this domain yet...

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -336,6 +336,13 @@ class BinaryCompactObject : public DomainCreator<3> {
     static constexpr Options::String help = {"Radius of the entire domain."};
   };
 
+  struct CoverCompactObjectsWithCubes {
```
knelli2: (Not a review)

I don't love this name because we technically cover the black holes with cubes as well. It's just with the BHs there's also some inner structure as well (and the cubes are DG not FD). However, I'm struggling to think of a better name:

- `CubesHaveStructure`
- `UseCartesianCubesAroundObjects`
- `CoverObjectsWithStrucutre`

Actually, I think the second name I suggested, `UseCartesianCubesAroundObjects`, is pretty clear. If it's true, it does what it says. If it's false, then the cubes use DG and have some inner structure. What do you think?

ffoucart: I'm alright with the alternative name. One thing to note though is that, as far as I understand the code, we won't require the use of FD in the cubes. We just need cubes to allow it for now, but can still use DG in smooth region.

nilsvu: Ideally we would replace the options that currently define objects A and B with a simple `CartesianCubeAtXCoord: 10.` or something like that. I think that can be done by changing the type of the `ObjectA` and `ObjectB` options from `Object` to `std::variant<Object, CartesianObject>` and implementing a new `CartesianObject` struct that parses however you want (I think it just needs a single x-coord option). Then you don't have to specify unused options like inner and outer radii for the objects. The result is something like this:

```yaml
DomainCreator:
  BinaryCompactObject:
    # BHNS domain
    ObjectA:
      CartesianCubeAtXCoord: 20.
    ObjectB:
      InnerRadius: 4.
      OuterRadius: 9.
      XCoord: -20.
      Interior:
        ExciseWithBoundaryCondition: # ...
      UseLogarithmicMap: False
  # ...
```

knelli2: This is true. Maybe `UseSingleCubeAroundObjects`? That way we don't say anything about cartesian because I believe that will be taken care of in the `SubcellOptions`. There you specify which blocks are allowed to do FD and which are restricted to just DG.

Edit: Oops, was writing this at the same time as @nilsvu. Yes what he said is probably the best solution

Edit 2: I think my point about using "cartesian" in the name still stands. Maybe `SingleCubeAtXCoord` instead?

ffoucart: I think what @nilsvu proposes should work and entirely avoids that extra option (which would also limit modifications of other input files...). But I just learned about std::variant this morning while looking through the rest of the code, so I had not considered it before :-)

nilsvu: @knelli2 what's your objection to "Cartesian" in the name? I think it makes clear that the grid is not curved, which is the point of the exercise. Other name suggestions are welcome of course. I don't see the connection to SubcellOptions.
Edit: Yes in SubcellOptions you can clamp some blocks to DG, but that isn't mutually exclusive with the grid being curved or not. Only currently you must clamp curved-mesh blocks to DG.

knelli2: Not sure why, but I always associate "cartesian" with "FD" (maybe because for vacuum we always use curved meshes with spectral methods). However, that's obviously wrong because we can do DG on a cartesian grid. So I don't have an objection, just my brain getting things mixed up 🧠 .

ffoucart:
New version attempts to use std::variant for the domain structure of ObjectA and ObjectB.
It is generally less intrusive in terms of input file changes (no existing input file should change), but makes the domain creation file itself more complex (to deal with the variant...)

ffoucart:
@nilsvu @knelli2 : The last commit makes the changes discussed on slack to option names. The domain has also now been tested with the ScalarWave executable, and seems to have the expected grid structure.

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -297,15 +302,31 @@ class BinaryCompactObject : public DomainCreator<3> {
...
+    static constexpr Options::String help = {
+        "Options to set a single cube at a location on the x-axis"};
+    struct XCoord {
+      static std::string name() { return "CartesianCubeAtXCoord";}
+      using type = double;
+      static constexpr Options::String help = {"x-coordinate of center."};
+    };
+    using options = tmpl::list<XCoord>;
+    CartesianCubeAtXCoord() = default;
+    CartesianCubeAtXCoord(double local_x_coord) : x_coord(local_x_coord) {}
```
nilsvu: (Optional) I think you can just delete these since they're autogenerated

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -547,6 +568,7 @@ class BinaryCompactObject : public DomainCreator<3> {
       "\n"
       "The two objects are enveloped by a sphere centered at the origin, "
       "and by an outer shell that can transition to large outer radii."
+      "Alternatively, each object can be a single cube."
```
nilsvu: Delete this line (the text above refers to the structure enveloping both objects)

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -77,9 +77,22 @@ BinaryCompactObject::BinaryCompactObject(
...
+  // object
+  const double x_coord_a =
+      std::visit([](auto& arg) { return arg.x_coord; }, object_A_);
+  const double x_coord_b =
+      std::visit([](auto& arg) { return arg.x_coord; }, object_B_);
+  const bool is_excised_a =
+      std::visit([](auto& arg) { return arg.is_excised(); }, object_A_);
+  const bool is_excised_b =
+      std::visit([](auto& arg) { return arg.is_excised(); }, object_B_);
+  const bool use_single_block_a = (object_A_.index() == 1);
```
nilsvu: Better use `std::holds_alternative` than `index` I think, just for readability

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -117,52 +138,63 @@ BinaryCompactObject::BinaryCompactObject(
...
+    }
   }
-  if (object_B_.is_excised() and
-      ((*object_B_.inner_boundary_condition == nullptr) !=
-       (outer_boundary_condition_ == nullptr))) {
-    PARSE_ERROR(context,
-                "Must specify either both inner and outer boundary conditions "
-                "or neither.");
+  if (!use_single_block_b) {
+    const Object& object_b = std::get<0>(object_B_);
```
nilsvu: Prefer `std::get<Object>` over the index (for readability)

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -207,18 +239,30 @@ BinaryCompactObject::BinaryCompactObject(
...
     }
   };
-  add_object_region("ObjectA", "Shell");  // 6 blocks
-  add_object_region("ObjectA", "Cube");   // 6 blocks
-  add_object_region("ObjectB", "Shell");  // 6 blocks
-  add_object_region("ObjectB", "Cube");   // 6 blocks
-  add_outer_region("Envelope");           // 10 blocks
-  add_outer_region("OuterShell");         // 10 blocks
-  if (not object_A_.is_excised()) {
+  if (not use_single_block_a) {
```
nilsvu: Swap `if/else` cases to avoid the negation (negations are surprisingly hard to read / understand)

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -207,18 +239,30 @@ BinaryCompactObject::BinaryCompactObject(
...
+    add_object_region("ObjectA", "Shell");  // 6 blocks
+    add_object_region("ObjectA", "Cube");   // 6 blocks
+  } else {
+    block_names_.emplace_back("ObjectASingleBlock0");
```
nilsvu: Just "ObjectA"

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -207,18 +239,30 @@ BinaryCompactObject::BinaryCompactObject(
...
+    block_groups_["ObjectASingleBlock"].insert("ObjectASingleBlock0");
```
nilsvu: Delete this line (it's a single block, so no need to group it)

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -285,76 +331,117 @@ BinaryCompactObject::BinaryCompactObject(
 }
 
 Domain<3> BinaryCompactObject::create_domain() const {
-  const double inner_sphericity_A = object_A_.is_excised() ? 1.0 : 0.0;
-  const double inner_sphericity_B = object_B_.is_excised() ? 1.0 : 0.0;
+  // Get useful information about the type of grid used around each compact
+  // object
```
nilsvu: Probably easier to store these in private member variables than to  recompute them

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -483,14 +579,16 @@ Domain<3> BinaryCompactObject::create_domain() const {
...
-                              object_A_.outer_radius,
-                              {{object_A_.x_coord, 0.0, 0.0}}};
-    CompressionMap size_B_map{size_map_function_of_time_names_[1],
-                              object_B_.inner_radius,
-                              object_B_.outer_radius,
-                              {{object_B_.x_coord, 0.0, 0.0}}};
+    CompressionMap size_A_map{
+        size_map_function_of_time_names_[0],
+        use_single_block_a ? 1.0 : std::get<0>(object_A_).inner_radius,
+        use_single_block_a ? 2.0 : std::get<0>(object_A_).outer_radius,
```
nilsvu: Could you add a comment why you chose values 1 and 2 here?

nilsvu (changes requested):
Please squash your existing commits to one (but pushing new changes as a fixup commit would be helpful for review)

src/Domain/Creators/BinaryCompactObject.cpp
knelli2: Do this throughout all files as well. Also when you do need to negate something, prefer `not` over `!`

src/Domain/Creators/BinaryCompactObject.cpp
knelli2: We don't even use the size map unless there is an excision. So I think for clarity it'd be better to default construct the maps, and then if an object is excised, actually construct the map with the radii.

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -109,6 +109,11 @@ namespace creators {
  *   hemispheres. The cutting plane always intersects the x-axis at the origin.
  * - The x-coordinate locations of the two objects should be chosen such that
  *   the center of mass is located at x=0.
+ * - Alternatively, one can replace the interior, shell and cube blocks of each
+ *   object with a single cube per object, Object A/B SingleBlock. This is less
```
knelli2:
```
 Alternatively, one can replace the inner shell and cube blocks of each object with a single cartesian cube.
```

knelli2 (changes requested):

src/Domain/Creators/BinaryCompactObject.cpp
ffoucart: Yes, this was a placeholder that I forgot to address, but the correct answer is probably to avoid creating the map in the absence of excision.

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -77,9 +77,22 @@ BinaryCompactObject::BinaryCompactObject(
       use_projective_map_(use_projective_map),
       radial_distribution_outer_shell_(radial_distribution_outer_shell),
       outer_boundary_condition_(std::move(outer_boundary_condition)) {
+  // Get useful information about the type of grid used around each compact
+  // object
+  x_coord_a_ = std::visit([](auto& arg) { return arg.x_coord; }, object_A_);
```
nilsvu: `const auto& arg` here and in the other std::visits

nilsvu (dismissed):
Please squash this in once @knelli2 has had a chance to look.

I think you aren't testing a time-dependent BNS domain currently. We can improve the tests later if you prefer, because @knelli2 is also working on some refactorings here. I think ideally we combine the BNS and BBH test cases and choose how to fill the two cubes randomly. This is how Test_Sphere.cpp works. We should also combine the time-dependent and time-independent test cases like in Test_Sphere.cpp, and check that in the time dependent case a few points that are easy to understand map to the correct coordinates, like the object centers or the corners of the cubes.

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -109,11 +109,8 @@ namespace creators {
  *   hemispheres. The cutting plane always intersects the x-axis at the origin.
  * - The x-coordinate locations of the two objects should be chosen such that
  *   the center of mass is located at x=0.
- * - Alternatively, one can replace the interior, shell and cube blocks of each
- *   object with a single cube per object, Object A/B SingleBlock. This is less
```
knelli2: You can leave in `This is less efficient, ...` to the end of the paragraph. I had only meant for you to replace the first sentence.

ffoucart:
I am not testing the time dependence yet, no. I was leaving that off for now because we can't do moving grids anyways.

knelli2 (changes requested):
Ok you can sqaush in everything

ffoucart:
Squashed, including the last two requested changes.

ffoucart:
I also changed all
`const Object& object = std::get<Object>(...)`
to 
`const auto& object = std::get<Object>(...)`
as clang-tidy dislikes the first version.

knelli2 (approved):

nilsvu (approved):

PR #4816, opened 2023-03-08
Allow con2prim after tci fail with DG clamp

pajkosmi:
## Proposed changes

If a cell is DG clamped, allow con2prim transformation to occur with hydro variables, even if cell is troubled. Also the check for # neighbors > 1 (found in PrepareNeighborData) now only occurs if an element uses subcell.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

Currently a draft PR to check code coverage & CI.

PR #4842, opened 2023-03-16
No fd element with dg block neighbor at initialize

pajkosmi:
## Proposed changes

1. Prevent elements from _initializing_ as FD when bordering a DG neighbor.  Likewise, make sure this check supersedes the `always_use_subcell` parameter.
2. Fix a bug that had block id as bool instead of size_t in `TciAndRollback.hpp`.
3. Rename `subcell_allowed_in_block` to `subcell_allowed_in_element`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

@nilsvu this may alleviate the artifacts you have been seeing with the head on NSNS evolution.

### DIFF
--- a/src/Evolution/DgSubcell/Actions/Initialize.hpp
+++ b/src/Evolution/DgSubcell/Actions/Initialize.hpp
@@ -94,9 +94,9 @@ struct Initialize {
   using const_global_cache_tags = tmpl::list<Tags::SubcellOptions<Dim>>;
 
   using simple_tags = tmpl::list<
-      Tags::ActiveGrid, Tags::DidRollback,
-      Tags::TciGridHistory, Tags::NeighborDataForReconstruction<Dim>,
-      Tags::TciDecision, Tags::NeighborTciDecisions<Dim>, Tags::DataForRdmpTci,
+      Tags::ActiveGrid, Tags::DidRollback, Tags::TciGridHistory,
+      Tags::NeighborDataForReconstruction<Dim>, Tags::TciDecision,
+      Tags::NeighborTciDecisions<Dim>, Tags::DataForRdmpTci,
       fd::Tags::InverseJacobianLogicalToGrid<Dim>,
       fd::Tags::DetInverseJacobianLogicalToGrid,
       subcell::Tags::CellCenteredFlux<typename System::flux_variables, Dim>>;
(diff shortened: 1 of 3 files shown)


src/Evolution/DgSubcell/Actions/Initialize.hpp
```diff
@@ -125,9 +125,23 @@ struct Initialize {
...
 
-    const bool subcell_allowed_in_block = not std::binary_search(
-        subcell_options.only_dg_block_ids().begin(),
-        subcell_options.only_dg_block_ids().end(), element.id().block_id());
+    // Loop over block neighbors and if neighbor id is inside of
+    // subcell_options.only_dg_block_ids(), then bordering DG-only block
+    const bool bordering_dg_block = alg::any_of(
+        element.neighbors(),
+        [&subcell_options](const auto& direction_and_neighbor) {
+          const int first_block_id =
```
knelli2: I believe `block_id()`s are stored as `size_t`

src/Evolution/DgSubcell/Actions/TciAndRollback.hpp
```diff
@@ -135,7 +135,7 @@ struct TciAndRollback {
     const bool bordering_dg_block = alg::any_of(
         element.neighbors(),
         [&subcell_options](const auto& direction_and_neighbor) {
-          const bool first_block_id =
+          const int first_block_id =
```
knelli2: Same as above comment. (Also change commit message and PR description)

knelli2 (changes requested):
You can squash directly.

src/Evolution/DgSubcell/Actions/TciAndRollback.hpp
nilsvu: @nilsdeppe This should have been caught by clang-tidy in #4816 but it didn't trigger... looking at the CI check output (https://github.com/sxs-collaboration/spectre/actions/runs/4389805990/jobs/7687734830?pr=4816) it looks like clang-tidy didn't run over this hpp file (maybe because it isn't included in any cpp file? not sure, I haven't looked into clang-tidy for a while). Any idea how we can prevent this sort of bug in the future?

pajkosmi:
@knelli2 changes have been addressed.  Also I had to convert a bool to size_t in `Test_TciAndRollback.cpp`.

knelli2 (approved):
Looks good!

For future reference, when I (or other reviewers) say to squash, we typically mean just squash the fixup commit into the relevant commit; not all commits into one. Your original 3 commits were perfectly fine. Don't worry about changing this PR, but just just wanted to let you know for future ones. 😄

src/Evolution/DgSubcell/Actions/TciAndRollback.hpp
nilsdeppe: I don't know exactly when clang-tidy _should_ check code in an hpp, but I thought it's whenever a cpp file uses that hpp. We run clang-tidy on cpp files only, which is the recommended thing, but a lot of checks that it used and I think should be catching haven't been caught in hpp files. E.g. `const int` declarations in hpp files aren't being flagged at all anymore. At least the local clang-tidy inside my IDE warns about these, so I don't think it's our configuration file. 
From the log it looks like clang-tidy didn't run over any of the `Test_*` files, which if it had I assume it would've noticed this. I don't know why it didn't run over any `Test_*` files, though.

nilsdeppe (approved):

PR #4843, opened 2023-03-17
Add 1st order puncture field

nikwit:
## Proposed changes

Adds the singular puncture field expanded up to 1st order in Kerr-Schild coordinates.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
@@ -25,9 +25,12 @@ void puncture_field(
     const double orbital_radius, const double bh_mass, const size_t order) {
   if (order == 0) {
     puncture_field_0(result, coords, time, orbital_radius, bh_mass);
+  } else if (order == 1) {
+    puncture_field_1(result, coords, time, orbital_radius, bh_mass);
   } else {
     ERROR(
-        "Puncture order is only implemented up to order 0 but you requested "
+        "The puncture field is only implemented up to expansion order 1 but "
+        "you requested "
         "order "
         << order);
   }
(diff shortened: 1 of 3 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
```diff
@@ -440,4 +443,1109 @@ void puncture_field_0(
...
+                           Frame::Inertial>>(*result)) =
+      (1.0 / 6.0) * M * d_16 * d_82 * dv_184 * dv_216 * dv_333 +
+      (1.0 / 2.0) * M * d_26 * d_31 * d_37 * dv_164 * dv_180 * dv_341 * dv_72 *
+          dv_89 * r0 -
+      dv_175 * (-d_129 * dv_58 + dv_339) - dv_281 * dv_333 -
+      dv_289 * (-d_129 * d_48 * dv_66 + 4.0 * d_22 * d_41 * dv_276 * dv_58 -
+                dv_287 * dv_336 - dv_337) -
+      dv_291 * dv_333 -
+      dv_328 *
+          (12 * M * d_107 * d_27 * d_34 * dv_194 * dv_195 * dv_198 * dv_200 *
```
nilsdeppe: Unclear if it's worth it, but here it looks like you could reduce the math a bit with some factoring of:
```
12 * M * d_107 * d_27 * d_34 * dv_194 * dv_200 * r0 *(...)
```
Just let me know if you don't care and then I'll just merge as-is :)

nilsdeppe (dismissed):

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
nikwit: oh wow, well spotted. weird that the sympy CSE did not pick that up. I squashed it in.

nilsdeppe (approved):

PR #4844, opened 2023-03-17
Clean up the time dependent options for BCO domains

knelli2:
## Proposed changes

Both BCO domains now (since #4792) share the exact same time dependent maps and options so we shouldn't be editing things twice.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #4824.~

### DIFF
--- a/src/Domain/Creators/BinaryCompactObject.cpp
+++ b/src/Domain/Creators/BinaryCompactObject.cpp
@@ -288,8 +288,8 @@ BinaryCompactObject::BinaryCompactObject(
 // Time-dependent constructor, with additional options for specifying
 // the time-dependent maps
 BinaryCompactObject::BinaryCompactObject(
-    double initial_time, double expansion_map_outer_boundary,
-    double initial_expansion, double initial_expansion_velocity,
+    double initial_time, double initial_expansion,
+    double initial_expansion_velocity,
     double asymptotic_velocity_outer_boundary,
     double decay_timescale_outer_boundary_velocity,
     std::array<double, 3> initial_angular_velocity,
@@ -312,7 +312,6 @@ BinaryCompactObject::BinaryCompactObject(
                           std::move(outer_boundary_condition), context) {
   enable_time_dependence_ = true;
   initial_time_ = initial_time;
-  expansion_map_outer_boundary_ = expansion_map_outer_boundary;
   initial_expansion_ = initial_expansion;
   initial_expansion_velocity_ = initial_expansion_velocity;
   asymptotic_velocity_outer_boundary_ = asymptotic_velocity_outer_boundary;
@@ -563,7 +562,7 @@ Domain<3> BinaryCompactObject::create_domain() const {
         distorted_to_inertial_block_maps{number_of_blocks_};
 
     CubicScaleMap expansion_map{
-        expansion_map_outer_boundary_, expansion_function_of_time_name_,
+        outer_radius_, expansion_function_of_time_name_,
         expansion_function_of_time_name_ + "OuterBoundary"s};
     RotationMap3D rotation_map{rotation_function_of_time_name_};
 
--- a/src/Domain/Creators/BinaryCompactObject.cpp
+++ b/src/Domain/Creators/BinaryCompactObject.cpp
@@ -37,6 +37,7 @@
 #include "Domain/CoordinateMaps/TimeDependent/Rotation.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/SphericalCompression.hpp"
 #include "Domain/CoordinateMaps/Wedge.hpp"
+#include "Domain/Creators/BinaryCompactObjectHelpers.hpp"
 #include "Domain/Creators/DomainCreator.hpp"  // IWYU pragma: keep
 #include "Domain/Creators/ExpandOverBlocks.hpp"
 #include "Domain/Domain.hpp"
@@ -58,7 +59,6 @@ bool BinaryCompactObject::Object::is_excised() const {
   return inner_boundary_condition.has_value();
 }
 
-// Time-independent constructor
 BinaryCompactObject::BinaryCompactObject(
     typename ObjectA::type object_A, typename ObjectB::type object_B,
     const double envelope_radius, const double outer_radius,
@@ -285,23 +285,14 @@ BinaryCompactObject::BinaryCompactObject(
   }
 }
 
-// Time-dependent constructor, with additional options for specifying
-// the time-dependent maps
 BinaryCompactObject::BinaryCompactObject(
-    double initial_time, double initial_expansion,
-    double initial_expansion_velocity,
-    double asymptotic_velocity_outer_boundary,
-    double decay_timescale_outer_boundary_velocity,
-    std::array<double, 3> initial_angular_velocity,
-    std::array<double, 2> initial_size_map_values,
-    std::array<double, 2> initial_size_map_velocities,
-    std::array<double, 2> initial_size_map_accelerations,
+    bco::TimeDependentMapOptions time_dependent_options,
     typename ObjectA::type object_A, typename ObjectB::type object_B,
     double envelope_radius, double outer_radius,
     const typename InitialRefinement::type& initial_refinement,
     const typename InitialGridPoints::type& initial_number_of_grid_points,
-    const bool use_equiangular_map, bool use_projective_map,
-    CoordinateMaps::Distribution radial_distribution_outer_shell,
+    const bool use_equiangular_map, const bool use_projective_map,
+    const CoordinateMaps::Distribution radial_distribution_outer_shell,
     std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
         outer_boundary_condition,
     const Options::Context& context)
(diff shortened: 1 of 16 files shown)


src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -0,0 +1,117 @@
...
+  // their initial expiration time to infinity (i.e. not expiring)
+  std::unordered_map<std::string, double> expiration_times{
+      {expansion_name, std::numeric_limits<double>::infinity()},
+      {rotation_name, std::numeric_limits<double>::infinity()},
+      {gsl::at(size_names, 0), std::numeric_limits<double>::infinity()},
+      {gsl::at(size_names, 1), std::numeric_limits<double>::infinity()}};
+
+  // If we have control systems, overwrite these expiration times with the ones
+  // supplied by the control system
+  for (auto& [name, expr_time] : initial_expiration_times) {
```
nilsvu: const

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -20,4 +33,156 @@ struct enable_time_dependent_maps<Metavariables,
...
+      static constexpr Options::String help = {
+          "The timescale for how fast the outer boundary velocity approaches "
+          "its asymptotic value."};
+    };
+    using options = tmpl::list<InitialValues, AsymptoticVelocityOuterBoundary,
+                               DecayTimescaleOuterBoundaryVelocity>;
+    ExpansionMapOptions() = default;
+    ExpansionMapOptions(std::array<double, 2> local_initial_values,
+                        double local_outer_boundary_velocity,
+                        double local_outer_boundary_decay_time);
```
nilsvu: You can delete these constructors because they're autogenerated I think

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -20,4 +33,156 @@ struct enable_time_dependent_maps<Metavariables,
...
+  inline static const std::string expansion_name{"Expansion"};
+  inline static const std::string expansion_outer_boundary_name{
+      "ExpansionOuterBoundary"};
+  inline static const std::string rotation_name{"Rotation"};
+  inline static const std::array<std::string, 2> size_names{{"SizeA", "SizeB"}};
+  // This is public because the BCO domains need to check if they have time
+  // dependent options when creating maps. A default constructed
+  // TimeDependentOptions is not time-dependent because what values would we use
+  // to create the maps?
+  bool is_time_dependent{false};
```
nilsvu: tl;dr: Delete this variable and hold a `std::optional<TimeDepOptions>` in the domain creators instead.

---

Since this variable is only used to return an empty FoT map and is a bit awkward here (relying on default-construction behavior), it's better to make the `is_time_dependent` decision in the calling code and delete this member variable. The calling code has to set this variable anyway (currently by default-construction, which is a bit intransparent), so it might as well return an empty map directly. I think it's easiest to hold the time dep options as an optional, because looking at your code in the domain creators you use it in `if (is_time_dep) { use time dep options }` clauses, which convert easily to an optional.

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -162,6 +163,14 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
   z_cutting_plane_ = cut_spheres_offset_factor_ *
                      ((1.0 - xi) * center_B_[2] + xi * center_A_[2]);
 
+  if (time_dependent_options_.is_time_dependent and
```
nilsvu: I think this will never trigger because `is_time_dependent` is always false at this point.

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -390,18 +396,14 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
...
                 "Currently, one or both objects is missing the inner spheres.");
   }
-  is_time_dependent_ = true;
-  initial_time_ = initial_time;
-  expansion_map_options_ = expansion_map_options;
-  initial_angular_velocity_ = initial_angular_velocity;
-  initial_size_map_values_[0] = initial_size_map_values_A;
-  initial_size_map_values_[1] = initial_size_map_values_B;
+
+  time_dependent_options_ = time_dependent_options;
```
nilsvu, line 398: std::move this, or are you sure it's trivially copyable?

src/Domain/Creators/CylindricalBinaryCompactObject.hpp
```diff
@@ -434,10 +354,7 @@ class CylindricalBinaryCompactObject : public DomainCreator<3> {
       const Options::Context& context = {});
 
   CylindricalBinaryCompactObject(
-      double initial_time, ExpansionMapOptions expansion_map_options,
-      std::array<double, 3> initial_angular_velocity,
-      std::array<double, 3> initial_size_map_values_A,
-      std::array<double, 3> initial_size_map_values_B,
+      bco::TimeDependentMapOptions time_dependent_options,
```
nilsvu, line 357: (optional) I think you can combine the two constructors now if the TimeDepOptions is a std::optional, defaults to std::nullopt, and you move it down before or after the boundary conditions. That simplifies the test as well because you only have a single constructor and don't need the `with_time_dependence` if clause.

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -480,7 +479,7 @@ Domain<3> BinaryCompactObject::create_domain() const {
         distorted_to_inertial_block_maps{number_of_blocks_};
 
     CubicScaleMap expansion_map{
-        expansion_map_outer_boundary_, expansion_function_of_time_name_,
+        outer_radius_, expansion_function_of_time_name_,
```
nilsvu: @geoffrey4444 could you confirm that setting the expansion map outer boundary always to the domain outer boundary is what you want?

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -574,16 +460,9 @@ class BinaryCompactObject : public DomainCreator<3> {
...
   BinaryCompactObject(
-      double initial_time, double initial_expansion,
-      double initial_expansion_velocity,
-      double asymptotic_velocity_outer_boundary,
-      double decay_timescale_outer_boundary_velocity,
-      std::array<double, 3> initial_angular_velocity,
-      std::array<double, 2> initial_size_map_values,
-      std::array<double, 2> initial_size_map_velocities,
-      std::array<double, 2> initial_size_map_accelerations, Object object_A,
+      bco::TimeDependentMapOptions time_dependent_options, Object object_A,
```
nilsvu: Same comment about combining the constructors

tests/InputFiles/ExportCoordinates/InputTimeDependent3D.yaml
```diff
@@ -47,16 +47,15 @@ DomainCreator:
     TimeDependentMaps:
       InitialTime: 0.0
       ExpansionMap:
-        InitialExpansion: 1.0
-        InitialExpansionVelocity: -4.6148457646200002e-05
+        InitialValues: [1.0, 0.0]
```
nilsvu: Keep the ~1e-5 velocity?

nilsvu (changes requested):

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
knelli2: Clang-tidy yelled at me when I originally had `std::move` and said it's trivially copyable. I was surprised by this

src/Domain/Creators/CylindricalBinaryCompactObject.hpp
knelli2: So I tried to do this originally (because I agree it'd be a lot simpler) but TL;DR it won't work because of options.

----

The reason it won't work is because we have two different sets of "optional" options, meaning boundary conditions and time dependent maps. These are compile-time options to the `using options =` list. So you have 4 different combos of things that all technically need their own constructor:

1. With BC, with TimeDep
2. With BC, without TimeDep
3. Without BC, with TimeDep
4. Without BC, without TimeDep

So if I have a general constructor like
```cpp
CylBCO(type common_opts,
       TimeDepOpts time_dep_opts = std::nullopt,
       BCs boundary_condtions = nullptr)
```

this constructor works fine for 1, 3, and 4, but it doesn't work for 2 because 2 expects a constructor like
```cpp
CylBCO(type common_opts,
       BCs boundary_condtions = nullptr,
       TimeDepOpts time_dep_opts = std::nullopt)
```

so you basically have to have a second constructor at that point.

nilsvu: Ok that's annoying but don't worry about it here then

knelli2:
Squash everything in because there weren't that many changes

nilsvu (dismissed):

nilsdeppe:
@knelli2 this needs a rebase to resolve conflicts

knelli2:
Ok fixed conflicts

nilsdeppe (approved):

PR #4846, opened 2023-03-18
Allow SphereTransition to be everywhere

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
@@ -12,6 +12,7 @@
 #include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/MakeWithValue.hpp"
 
 namespace domain::CoordinateMaps::ShapeMapTransitionFunctions {
 
@@ -49,9 +50,7 @@ std::optional<double> SphereTransition::original_radius_over_radius(
   }
   const double original_radius = (mag + distorted_radius * b_) / denom;
 
-  return original_radius >= r_min_ and original_radius <= r_max_
-             ? std::optional<double>{original_radius / mag}
-             : std::nullopt;
+  return std::optional<double>{original_radius / mag};
 }
 
 double SphereTransition::map_over_radius(
@@ -75,24 +74,46 @@ std::array<DataVector, 3> SphereTransition::gradient(
 template <typename T>
 T SphereTransition::call_impl(const std::array<T, 3>& source_coords) const {
   const T mag = magnitude(source_coords);
-  check_magnitudes(mag);
-  return a_ + b_ / mag;
+  T result = a_ + b_ / mag;
+  for (size_t i = 0; i < get_size(mag); ++i) {
+    if (get_element(mag, i) + eps_ < r_min_) {
+      get_element(result, i) = 1.0;
+    } else if (get_element(mag, i) - eps_ > r_max_) {
+      get_element(result, i) = 0.0;
+    }
+  }
+  return result;
 }
 
 template <typename T>
 T SphereTransition::map_over_radius_impl(
     const std::array<T, 3>& source_coords) const {
   const T mag = magnitude(source_coords);
-  check_magnitudes(mag);
-  return a_ / mag + b_ / square(mag);
+  T result = a_ / mag + b_ / square(mag);
+  for (size_t i = 0; i < get_size(mag); ++i) {
+    if (get_element(mag, i) + eps_ < r_min_) {
+      get_element(result, i) = 1.0 / get_element(mag, i);
+    } else if (get_element(mag, i) - eps_ > r_max_) {
+      get_element(result, i) = 0.0;
+    }
+  }
+  return result;
 }
 
 template <typename T>
 std::array<T, 3> SphereTransition::gradient_impl(
     const std::array<T, 3>& source_coords) const {
   const T mag = magnitude(source_coords);
-  check_magnitudes(mag);
-  return -b_ * source_coords / cube(mag);
+  std::array<T, 3> result = -b_ * source_coords / cube(mag);
+  for (size_t i = 0; i < get_size(mag); ++i) {
+    if (get_element(mag, i) + eps_ < r_min_ or
+        get_element(mag, i) - eps_ > r_max_) {
+      for (size_t j = 0; j < result.size(); j++) {
+        get_element(gsl::at(result, j), i) = 0.0;
+      }
+    }
+  }
+  return result;
 }
 
 bool SphereTransition::operator==(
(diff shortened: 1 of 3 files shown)


nikwit (approved):
This looks good.  Two questions:

1. The shape map is now not differentiable at `r_min` and `r_max` if I understand correctly. I am not sure if this is a problem. Alternatively, we could construct a simple cubic map that falls off from 1 to 0 by including a differentiability condition at `r_min` and `r_max`,. 

3. Will you still only include this map in the blocks that need it or in all blocks like is done in SpEC?

knelli2:
1. Yes you are correct. I also don't know if this will be a problem. If it is we can add in the cubic transition
2. For the BBH, yes I plan to only include this map in blocks that use it. The main motivation for this change was the shape time dependence when used in the SingleBH exe, because TimeDependences are applied to all blocks, but I only wanted the shape map to affect certain blocks.

nilsvu:
I think it is a problem that the Jacobian is discontinuous at r_min and r_max. It would be fine if the discontinuity is resolved by a block boundary, meaning that it takes a different value on either side of the block boundary, even if the coordinates of the points are the same (this happens on LGL grids). However, using the coordinates alone to decide which value the Jacobian takes (as is done in this PR) means that the discontinuity is on an element grid (either in the interior or exterior element, depending on the choice of plus or minus epsilon in this PR).
I think we either have to make the map smooth, or somehow pass along which side of the boundary we are, or evaluate on Gauss grids and interpolate to Gauss-Lobatto. To make the map smooth I think we have a sigmoid function implemented somewhere that I added for the Schwarz solver, that could do the job.
Edit: I meant `smoothstep` in Utilities/Math.hpp. At first order its first derivative is continuous at the edges (Jacobian), but the second isn't (Hessian). The differentiability increases with the order. So the function is not actually smooth at the edges, and I have to think if that's a problem or not.
Edit2: I believe the first-order smoothstep is fine here as long as r_min and r_max are at block boundaries, because only the Jacobian is evaluated on the grid. Then the non-smoothness of the map is resolved by block boundaries. If you go this route, please add a few notes on this discussion to the docs.

nikwit:
There are also options for completely smooth transition functions like the Gaussian error function.

nilsvu:
erf(x) does not transition in a finite interval, neither do the other smooth sigmoids that I know. I think transitioning in a finite interval would be good though.

nikwit:
SpEC should have the same issue and I think they use a linear fall-off function which is not differentiable. @markscheel do you know how/if that is addressed?

nilsvu:
As long as the map is used only in specific blocks or on Gauss quadrature grids this issue does not appear, so it might not matter in SpEC.

nilsdeppe:
In SpEC the shape is the identity in subdomains outside the `SphereA` and `SphereB` subdomains. It switches at what we call a block boundary in spectre. Gaussians were tried and didn't work well in SpEC, that's why there's the abrupt transition. One point of having blocks was so we could isolate the maps to them and so isolate any non-smoothness to places where it's guaranteed to be fine. This must be respected by all maps, it's a fundamental design decision of spectre that was made 5+ years ago

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
```diff
@@ -75,24 +74,46 @@ std::array<DataVector, 3> SphereTransition::gradient(
...
 template <typename T>
 std::array<T, 3> SphereTransition::gradient_impl(
     const std::array<T, 3>& source_coords) const {
   const T mag = magnitude(source_coords);
-  check_magnitudes(mag);
-  return -b_ * source_coords / cube(mag);
+  std::array<T, 3> result = -b_ * source_coords / cube(mag);
+  for (size_t i = 0; i < get_size(mag); ++i) {
+    if (get_element(mag, i) + eps_ < r_min_ or
+        get_element(mag, i) - eps_ > r_max_) {
```
nilsdeppe, line 110: this won't work if r_max_ is greater than 100.0 You are using an absolute tolerance, and so 100-eps_ is identically 100. Same issue with `r_min_` if it's big enough.. You can do something like `r_max_ * (1.0 + eps_)` to make it a relative scale, then it's safe. You can then probably also tighten `eps` to like `2*epsilon()`. Having said all that, this still will break smoothness inside of a block, which is an issue.

nilsvu:
I only see two options:

1. Try a first-order smoothstep (that's the cubic polynomial 3x^2-2x^3)
2. Allow the Shape TimeDependence to take a list of blocks (names / groups) it is limited to, instead of applying to all blocks. This needs an input file option.

knelli2:
I prefer option 2

nilsdeppe:
Sounds good to me!

knelli2:
Hmm actually, I'm not sure the best way to move forward with option 2. The TimeDependence doesn't know anything about the domain. All it gets is the number of blocks and we return a vector of coord maps that correspond to the vector of blocks. I'd have to make an assumption about the domain I'm using, even with block names/groups as inputs.

What really needs to happen is that the functions for constructing the maps need to take the block names and block groups in addition to the number of blocks. Then I'd be able to specify block names in the time dependence block of the input file and properly restrict the maps to certain blocks.

markscheel:
I thought the plan was to use TimeDependence only for simple tests where the maps are the same in every Block, and that anything more complicated will set the appropriate maps in the DomainCreator.   But I am fine with generalizing TimeDependence (say by passing in block names as Kyle suggests), if it makes things easier.

Note that Size, Shape, CutX, and Skew will all have Block-dependent maps that are non-smooth at Block boundaries. Same for the Expansion map once we upgrade it (the smooth CubicScale map fails for eccentric binaries in SpEC, and we have already replaced it in SpEC with a non-smooth simpler map). As Nils says above, non-smooth maps give an enormous simplification over smooth maps, since 1) you don't need to worry about overlaps between maps around different objects [like black hole A and black hole B], 2) you can use much simpler functions (like linear) that are easy to invert, and 3) if you can use simple functions (like linear) then the gradients are smaller (and sometimes they are zero).

Note also that there will eventually be multiple SphereTransition functions.  The current one is easy, which is why we are using it now, but the SphereTransition functions we eventually want to use will be different in different Blocks, and will be different for rectangular vs cylindrical BBH domains (these SphereTransition functions are already worked out for cylindrical BBH domains because SpEC uses them, but they shouldn't be hard to work out for the rectangular BBH case).

nilsvu:
That's option 3 that I forgot above:

3. Build shape map into Sphere domain creator.

Maybe the easiest?

nilsdeppe:
Kyle and I talked about this and we think the easiest is for the domain creator constructor to call a member function of the time dependence to set which block IDs go with what names & groups.

nilsvu:
Ok, please keep in mind that for rotating star initial data we'll need a filled Sphere domain with a shape map that tapers off towards the interior as well as towards the exterior. We can either put an extra spherical partition somewhere in the interior of the star and use the spherical transition function in reverse, or add a transition function that tapers off to the cubical faces. Nothing to add in this PR, just keep this in mind please so we won't have to refactor it all again.

nilsdeppe:
Okay, let's talk about this after lunch today if you're around? We can drag Mark into the conversation too. That conversation should focus on how to use the shape map when you include `r=0` in the domain. I have no idea what SpEC does for ID, they must have this problem too, I think...

nilsvu:
After lunch works. SpEC ID uses overlapping blocks, so it's possible that the shape map doesn't (need to) taper off at all.

knelli2:
Discussion with @markscheel @nilsvu @nilsdeppe on 3/30/23:

It was decided that it would just be easier to just hard code time dependent maps into the Sphere domain itself rather than change the TimeDependences to be knowledgeable about the blocks. For now, we will just add the Shape map and can add other maps in the future. In order to prepare for future maps to be added, though, the options for these hard coded time dependent maps will be stored in a new class I will create `SphereTimeDependentOptions` (name pending).

However, it's still useful to use TimeDependences with the Sphere domain for testing purposes. So the Sphere will need to be able to handle using either the hard coded options or a time dependence. The solution is variants. The new input file would look something like

```yaml
# Either have SphereTimeDependentOptions
Sphere:
  ...
  TimeDependentOptions: # This is a std::variant<TimeDependences, SphereTimeDependentOptions>
    Shape:
      LMax: 8
    ... # More options for more maps as we add them

# or TimeDependences
Sphere:
  ...
  TimeDependentOptions: # This is a std::variant<TimeDependences, SphereTimeDependentOptions>
    Rotation:
      ... # Options for only 1 time dependence (the rotation)
```

I'll get started on this probably next week sometime. I'm closing this PR and I'll just open a new one when I've added the discussed changes.

PR #4847, opened 2023-03-18
Factor out time dep maps for BCO domains

knelli2:
## Proposed changes

They both use the same maps, so put this into the `TimeDepndentMapOptions` helper for BCOs.

~I did have to make the member variable that stores the time dependent options mutable because I had to set some parameters. Let me know if there's a better way you see to do this.~

Working towards #4764.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #4824 and #4844.~

### DIFF
--- a/src/Domain/Creators/BinaryCompactObject.hpp
+++ b/src/Domain/Creators/BinaryCompactObject.hpp
@@ -44,14 +44,6 @@ class Wedge;
 template <size_t VolumeDim>
 class DiscreteRotation;
 class Frustum;
-namespace TimeDependent {
-template <size_t VolumeDim>
-class CubicScale;
-template <size_t VolumeDim>
-class Rotation;
-template <bool InteriorMap>
-class SphericalCompression;
-}  // namespace TimeDependent
 }  // namespace CoordinateMaps
 
 template <typename SourceFrame, typename TargetFrame, typename... Maps>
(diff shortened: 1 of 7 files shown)


src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -555,7 +518,10 @@ class BinaryCompactObject : public DomainCreator<3> {
   bool is_excised_b_ = false;
   bool use_single_block_a_ = false;
   bool use_single_block_b_ = false;
-  std::optional<bco::TimeDependentMapOptions> time_dependent_options_{};
+  // Mutable because we have to set some parameters in the time_dep_options in
+  // order to construct the maps
+  // NOLINTNEXTLINE
+  mutable std::optional<bco::TimeDependentMapOptions> time_dependent_options_{};
```
knelli2: Any suggestions for an alternative to this would be appreciated.

nilsvu: Construct the maps in the TimeDepOptions constructor so you can copy them later, or construct the maps on-demand since copying them has little benefit over constructing.

knelli2:
@nilsvu Ok I'm happy with this now.

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -151,6 +211,57 @@ struct TimeDependentMapOptions {
...
+   *
+   * - Expansion: `CubicScale<3>`
+   * - Rotation: `Rotation<3>`
+   * - SizeA/B: `SphericalCompression`
+   *
+   * If the inner/outer radii for an object are `std::nullopt`, this means that
+   * a Size map is not constructed for that object. An identity map will be used
+   * instead.
+   */
+  void build_maps(
```
nilsvu, line 227: If I understand correctly this function constructs the maps, and the other functions copy and return them. That's a bit of an awkward interface. Just have a private function that constructs each map with the needed arguments (like x_coord), then use those to build the compositions (passing args like x_coord along). There's no benefit in constructing-then-copying maps vs constructing on demand (that I can see).

nilsvu (dismissed):
If it helps rebasing you can squash the three commits into one

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
knelli2: The reason I chose this interface was because I didn't want to have to pass the same arguments (xcoord, ...) to the three different functions that build the maps. Visually in the domain creators that would just look cluttered, so I opted to just pass the arguments into a single function and build all the maps at once. Then when you need the composition, the function you call has 0 or 1 arguments (the bool).

This also makes adding maps in the future even easier because there's only one place you change arguments rather than like 6.

nilsvu: I'm not convinced this changing-state approach is better, but since I won't be able to continue this review until Thursday I'm happy to step back and have another person take over this review. If none can be found I can resume the review later this week of course.

nilsdeppe: I don't really understand the exact concern with changing state. Doing so in the constructor of the domain creator seems like a quite reasonable thing to do since setting the state of everything is the job of the constructor. I agree with Kyle that the shorter function calls and not needing to either compute more in `create_domain()` or store the inputs to the `build_maps` function as members of the domain creator is really nice. Given the current discussion here, I'm in favor of the current design Kyle has as opposed to the suggested change.

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -102,4 +110,95 @@ TimeDependentMapOptions::create_functions_of_time(
...
+    return std::make_unique<
+        CompressionMapForComposition<Frame::Grid, Frame::Distorted>>(
+        gsl::at(size_maps_, index));
+  }
+}
+
+template <domain::ObjectLabel Object>
+TimeDependentMapOptions::MapType<Frame::Grid, Frame::Inertial>
+TimeDependentMapOptions::everything_grid_to_inertial_map(
+    const bool include_distorted) const {
```
nilsdeppe: maybe name `include_distorted_frame` just to be extra clear?

nilsdeppe (commented):
LGTM, one small suggestion

knelli2:
@nilsdeppe @nilsvu Rebased on develop

nilsdeppe (approved):

PR #4850, opened 2023-03-20
Check type of tensor observed at extremum in constructor

ffoucart:
## Proposed changes

Check that the tensor whose extremum we observe in ObserveAtExtremum is a scalar and not optional, as suggested by @nilsvu in #4831 
Also modifies legend formatting according to the comments in the same PR.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
+++ b/src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
@@ -244,6 +244,23 @@ ObserveAtExtremum<ObservationValueTag, tmpl::list<ObservableTensorTags...>,
                      << scalar_name << "' is not known. Known tensors are: "
                      << ((db::tag_name<ObservableTensorTags>() + ",") + ...));
   }
+
+  tmpl::for_each<tmpl::list<ObservableTensorTags...>>(
+      [this, &context](auto tag_v) {
+        using tag = tmpl::type_from<decltype(tag_v)>;
+        const std::string tensor_name = db::tag_name<tag>();
+        if (tensor_name == scalar_name) {
+          if constexpr (tt::is_a_v<std::optional, typename tag::type>) {
+            if (tag::type::value_type::rank() != 0) {
+              PARSE_ERROR(context,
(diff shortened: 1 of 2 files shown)


ffoucart:
Note: I have checked that entering the wrong type of tensor as input returns the proper error message (if we should add that check to the standard test, let me know)

src/ParallelAlgorithms/Events/ObserveAtExtremum.hpp
```diff
@@ -244,6 +244,21 @@ ObserveAtExtremum<ObservationValueTag, tmpl::list<ObservableTensorTags...>,
...
   }
+
+  tmpl::for_each<tmpl::list<ObservableTensorTags...>>(
+      [this, &context](auto tag_v) {
+        using tag = tmpl::type_from<decltype(tag_v)>;
+        const std::string tensor_name = db::tag_name<tag>();
+        if (tensor_name == scalar_name) {
+          if constexpr (tt::is_a_v<std::optional, typename tag::type>) {
+            PARSE_ERROR(context,
+                        "Cannot observe extremum of optional at this point");
```
nilsvu: I think a simple check of `tag::type::value_type::rank()` would do the job here.

ffoucart: That seems to work, thanks!

nilsvu (dismissed):
Please squash 👍 FYI, it's helpful to prefix fixup commits with the word `fixup` so CI automatically flags them and waits for them to get squashed before merging.

Edit: in this case I think having the two separate commits is fine too, doesn't matter much. But the fixup thing is useful for other PRs.

ffoucart:
Sounds good, i'll try to use `fixup' more (I tend to dislike the fact that it leads to test failures, at least in SpEC... hate seeing that red cross symbol on the PR ;-) )

nilsvu:
I see what you mean :) unfortunately that's the only way to prevent merging a PR with fixup commits that I'm aware of.

nilsvu (approved):

PR #4864, opened 2023-03-22
Correct options to generate-xdmf in tutorials

ffoucart:
## Proposed changes

Modify example use of generate-xdmf to match current syntax (please check; this syntax worked for me, while the old one did not)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Tutorials/BeginnersTutorial.md
+++ b/docs/Tutorials/BeginnersTutorial.md
@@ -146,12 +146,11 @@ directory where you have the `H5` files, run
 
 ```
 spectre generate-xdmf \
-  --file-prefix ExportTimeDependentCoordinates3DVolume \
-  --subfile-name element_data --output BBH_Coords
+  --subfile-name element_data --output BBH_Coords \
+  ExportTimeDependentCoordinates3DVolume*h5
 ```
 
-The `--file-prefix` argument must be whatever is before the number `0` in the
-name of your `Volume` file. We output volume data per node so we append the
+We output volume data per node so we append the
 node number to each volume file we have. Since you're most likely running on a
 laptop, you'll only be running on one node so you should only get one output
 file for the volume. The `--subfile-name` argument is the group name inside the
--- a/docs/Tutorials/Visualization.md
+++ b/docs/Tutorials/Visualization.md
@@ -50,9 +50,8 @@ files. An XDMF file must be created from the volume data in order to do
 visualization using ParaView. To this end we provide the tool
 `generate-xdmf` in the Python command-line interface. Run it in your build
 directory as `spectre generate-xdmf`. It
-takes three required arguments which are passed to `--file-prefix`,
-`--subfile-name`, and `--output`. The argument passed to `--file-prefix` is the
-name of the H5 volume data, leaving out the node number and extension. The
+takes two required arguments which are passed to `--subfile-name`, and
+`--output`. The
 argument passed to `--subfile-name` is the name of the subfile inside the H5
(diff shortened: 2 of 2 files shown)


docs/Tutorials/BeginnersTutorial.md
```diff
@@ -146,12 +146,12 @@ directory where you have the `H5` files, run
...
-  --file-prefix ExportTimeDependentCoordinates3DVolume \
-  --subfile-name element_data --output BBH_Coords
+  --subfile-name element_data --output BBH_Coords \
+  ExportTimeDependentCoordinates3DVolume*h5
 ```
 
-The `--file-prefix` argument must be whatever is before the number `0` in the
-name of your `Volume` file. We output volume data per node so we append the
+The last argument must be a regular expression covering all h5 files to be
+observed (and no others). We output volume data per node so we append the
```
nilsvu: It's any number of positional arguments, they can be glob patterns (not regex), and they don't have to come last in the list of arguments.

Docs like this are hard to keep up-to-date, so maybe just delete the sentence? Your call.

docs/Tutorials/Visualization.md
```diff
@@ -50,9 +50,10 @@ files. An XDMF file must be created from the volume data in order to do
 visualization using ParaView. To this end we provide the tool
 `generate-xdmf` in the Python command-line interface. Run it in your build
 directory as `spectre generate-xdmf`. It
-takes three required arguments which are passed to `--file-prefix`,
-`--subfile-name`, and `--output`. The argument passed to `--file-prefix` is the
-name of the H5 volume data, leaving out the node number and extension. The
+takes two required arguments which are passed to `--subfile-name`, and
+`--output`, and a final argument which should be a regular expression.
```
nilsvu: See comment above.

Maybe just refer to `generate-xdmf --help` instead?

nilsvu (changes requested):

docs/Tutorials/BeginnersTutorial.md
ffoucart: Good for me; I was trying to stay as close as possible to the original.

docs/Tutorials/Visualization.md
```diff
@@ -50,9 +50,9 @@ files. An XDMF file must be created from the volume data in order to do
 visualization using ParaView. To this end we provide the tool
 `generate-xdmf` in the Python command-line interface. Run it in your build
 directory as `spectre generate-xdmf`. It
-takes three required arguments which are passed to `--file-prefix`,
-`--subfile-name`, and `--output`. The argument passed to `--file-prefix` is the
-name of the H5 volume data, leaving out the node number and extension. The
+takes two required arguments which are passed to `--subfile-name`, and
+`--output` (for more detail on the options, use `generate-xdmf --help`).
```
nilsvu: This is actually written below already, so just delete the part in parentheses (sorry haven't noticed it before)

nilsvu (changes requested):
Squash 👍

docs/Tutorials/Visualization.md
ffoucart: Second push removes that sentence.

nilsvu (approved):
Thanks for fixing the docs. I think it would be better to delete more parts of these docs that are redundant with the help string, and just refer to `--help` or even better print the help string to the docs. It's hard to keep docs up to date that don't live right next to the code. Nothing for you to fix in this PR though.

PR #4865, opened 2023-03-22
Add real spherical harmonics implementation

nikwit:
## Proposed changes

Adds a simple implementation to calculate real spherical harmonics using boost. I will only need to use it up to l=2. Alternatively, they could be calculated with the spin-weighted spherical harmonics, see the test.

### DIFF
--- a/src/NumericalAlgorithms/SphericalHarmonics/RealSphericalHarmonics.cpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/RealSphericalHarmonics.cpp
@@ -1,0 +1,42 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "NumericalAlgorithms/SphericalHarmonics/RealSphericalHarmonics.hpp"
+
+#include <boost/math/special_functions/spherical_harmonic.hpp>
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/Gsl.hpp"
+
+void real_spherical_harmonic(gsl::not_null<DataVector*> spherical_harmonic,
+                             const DataVector& theta, const DataVector& phi,
+                             size_t l, int m) {
+  ASSERT(size_t(abs(m)) <= l, "m needs to be smaller than l");
+  const int sign = (m % 2 == 0) ? 1 : -1;
+  const double prefactor = sign * M_SQRT2;
+  if (m < 0) {
+    for (size_t i = 0; i < theta.size(); ++i) {
+      (*spherical_harmonic)[i] = prefactor * boost::math::spherical_harmonic_i(
+                                                 l, abs(m), theta[i], phi[i]);
+    }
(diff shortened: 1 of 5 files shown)


src/NumericalAlgorithms/SphericalHarmonics/RealSphericalHarmonics.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * Y_{lm}(\theta, \phi) =
+ * \begin{cases}
+ * \sqrt{2} (-1)^m \mathcal{I}[Y_l^{|m|}] & \text{if } m < 0 \\
+ *  Y_l^{0}] & \text{if } m = 0 \\
+ * \sqrt{2} (-1)^m \mathcal{R}[Y_l^{m}] & \text{if } m > 0
+ * \end{cases}
+ * \end{equation}
+ *
+ * where \f$Y_l^m\f$ are the complex spherical harmonics and \f$\mathcal{R}\f$
+ * denotes the real part and \f$\mathcal{R}\f$ denotes the imaginary part.
```
nilsdeppe: `mathcal{R}` -> `mathcal{I}`?

src/NumericalAlgorithms/SphericalHarmonics/RealSphericalHarmonics.cpp
```diff
@@ -0,0 +1,40 @@
...
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/Gsl.hpp"
+
+void real_spherical_harmonic(gsl::not_null<DataVector*> spherical_harmonic,
+                             const DataVector& theta, const DataVector& phi,
+                             size_t l, int m) {
+  ASSERT(size_t(abs(m)) <= l, "m needs to be smaller than l");
+  for (size_t i = 0; i < theta.size(); ++i) {
+    if (m < 0) {
+      const int sign = (m % 2 == 0) ? 1 : -1;
```
nilsdeppe: [optional] you could move this check outside the loop, including multiplying by `M_SQRT2`.

src/NumericalAlgorithms/SphericalHarmonics/RealSphericalHarmonics.cpp
```diff
@@ -0,0 +1,40 @@
...
+      spherical_harmonic->at(i) =
```
nilsdeppe: I actually didn't know `DataVector` had an `at` function. Maybe switch to `(*spherical_harmonic)[i]` to be consistent with other parts of the code? At least `std::vector` the difference between `[]` and `at` is that `at` _always_ incurs the overhead of a bounds check. For `DataVector`, `[]` only bounds checks in debug mode

Same for `theta` and `phi`

nilsdeppe (changes requested):
Couple small things, please squash right away

nikwit:
Thanks! I also made a separate loop for each of the conditional branches to avoid checking m for each Datavector entry.

nilsdeppe (approved):
Nice! :D

PR #4867, opened 2023-03-22
Add ability to filter by block

knelli2:
## Proposed changes

Also change the `DisableForDebugging` option to just `Enable`.

### Upgrade instructions

If you are using filtering, change the `DisableForDebugging:` option to `Enable:` and also add a new option `BlocksToFilter:`. This new option is either `All` or a vector of block names.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


tests/InputFiles/CurvedScalarWave/PlaneWaveMinkowski1D.yaml
```diff
@@ -60,7 +60,8 @@ Filtering:
   ExpFilter0:
     Alpha: 12
     HalfPower: 32
-    DisableForDebugging: true
+    Enable: false
+    BlocksToFilter: All
```
nilsvu, line 64: Maybe just `Blocks`?

src/NumericalAlgorithms/LinearOperators/FilterAction.hpp
```diff
@@ -98,7 +105,22 @@ class Filter<FilterType, tmpl::list<TagsToFilter...>> {
...
+    const size_t block_id =
+        db::get<domain::Tags::Element<volume_dim>>(box).id().block_id();
+    const std::string& block_name =
+        Parallel::get<domain::Tags::Domain<volume_dim>>(cache)
+            .blocks()[block_id]
+            .name();
+
+    // Check if filtering is turned off. If it's on, check that this block is
+    // allowed to do filtering. A `nullopt` from blocks_to_filter() means that
+    // all blocks are allowed to do filtering
```
nilsvu: Also allow for block groups by checking if the specified name is in `domain.block_groups()` and then if the block name is in that group.

It would also be good to throw a parse error somewhere if the specified block names or groups are invalid, maybe in the option tag.

nilsvu (changes requested):

tests/InputFiles/CurvedScalarWave/PlaneWaveMinkowski1D.yaml
nilsdeppe: I prefer the current more verbose naming

src/NumericalAlgorithms/LinearOperators/ExponentialFilter.hpp
```diff
@@ -112,6 +129,7 @@ class Exponential {
   double alpha_{36.0};
   unsigned half_power_{16};
   bool enable_{true};
+  std::optional<std::vector<std::string>> blocks_to_filter_{};
```
nilsdeppe: would storing this as a `std::opt<unordered_set>` make lookup easier? Doing the `any_of` is a linear search, which is less ideal than a hash, I think. No big deal either way, I think, though.

tests/InputFiles/CurvedScalarWave/PlaneWaveMinkowski1D.yaml
nilsvu: Ok

knelli2:
To get the block groups working, I actually had to make quite a bit of changes so I put them in a fixup. Basically since the filter tag now needed to take the domain creator so I could check the names, the tag had to be moved into `Domain/Creators/Tags` to avoid cyclic dependencies (like other tags that use the domain creator)

src/NumericalAlgorithms/LinearOperators/FilterAction.hpp
```diff
@@ -98,7 +108,46 @@ class Filter<FilterType, tmpl::list<TagsToFilter...>> {
...
     const FilterType& filter_helper =
         Parallel::get<::Filters::Tags::Filter<FilterType>>(cache);
-    if (UNLIKELY(filter_helper.disable_for_debugging())) {
+    const size_t block_id =
+        db::get<domain::Tags::Element<volume_dim>>(box).id().block_id();
+    const auto& domain = Parallel::get<domain::Tags::Domain<volume_dim>>(cache);
+    const auto& block_groups = domain.block_groups();
+    const std::string& block_name = domain.blocks()[block_id].name();
+
+    const auto get_group_name = [&block_groups, &block_name]() {
```
nilsdeppe: move into `if ((not disable) and filter_helper.blocks_to_filter().has_value()) {` since it's only used there?

src/NumericalAlgorithms/LinearOperators/FilterAction.hpp
```diff
@@ -98,7 +108,46 @@ class Filter<FilterType, tmpl::list<TagsToFilter...>> {
...
+          current_block_group = group_name;
+        }
+      }
+      return current_block_group;
+    };
+
+    // Technically this whole next block could be done on a single line, but
+    // then it would be very dense and hard to understand. This way is easier to
+    // read and understand
+    bool disable = not filter_helper.enable();
```
nilsdeppe: `disable` -> `enable` would be a bit easier to understand, I think

nilsdeppe (changes requested):
LGTM, you can squash

knelli2:
Ok squashed

nilsdeppe (dismissed):

nilsdeppe:
This needs a rebase and update of an input file, it looks like

nilsdeppe (dismissed):
LGTM :)

src/Domain/Creators/Tags/Filter.hpp
```diff
@@ -0,0 +1,96 @@
+// Distributed under the MIT License.
```
nilsvu: We can't put all tags that depend on domain creators in here. I understand why the filtering tags can't go in NumericalAlgorithms/, so how about Evolution/?

knelli2: Ok I put it in Evolution

nilsdeppe (dismissed):

src/Evolution/Tags/Filter.hpp
```diff
@@ -0,0 +1,96 @@
...
+    // If this is nullopt, then we use all blocks
+    if (blocks_to_filter.has_value()) {
+      const Domain<Metavariables::volume_dim> domain =
+          domain_creator->create_domain();
+      const auto& blocks = domain.blocks();
+      const auto& block_groups = domain.block_groups();
+      std::unordered_set<std::string> block_names{};
+      for (const auto& block : blocks) {
+        block_names.emplace(block.name());
+      }
```
nilsvu: You can get the block names and groups from the domain creator directly, without creating the domain. Also doesn't need you to collect the names from the blocks.

src/NumericalAlgorithms/LinearOperators/FilterAction.hpp
```diff
@@ -98,9 +108,50 @@ class Filter<FilterType, tmpl::list<TagsToFilter...>> {
...
+    // Only do this check if filtering is enabled. A `nullopt` means all blocks
+    // are allowed to do filtering
+    if (enable and filter_helper.blocks_to_filter().has_value()) {
+      const auto& blocks_to_filter = filter_helper.blocks_to_filter().value();
+
+      const auto get_group_name = [&block_groups, &block_name]() {
+        std::string current_block_group{};
+        for (const auto& [group_name, block_names] : block_groups) {
+          if (block_names.count(block_name) == 1) {
+            current_block_group = group_name;
```
nilsvu: A block can be in any number of groups, so this check doesn't work I think. You could iterate over the blocks_to_filter and check if (1) the name is the current element's block name -> enable and return, or (2) the name is a key in block_groups and the current element's block name is in the block group -> enable and return. Would be good to catch this in the test. Maybe just use an existing domain creator?

nilsvu (changes requested):

nilsvu (dismissed):

nilsdeppe:
I'm also fine with the changes. Please rebase & squash

nilsdeppe (approved):

PR #4869, opened 2023-03-22
Don't print traceback for incorrect CLI command

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Python/__main__.py
+++ b/support/Python/__main__.py
@@ -74,7 +74,11 @@ class Cli(click.MultiCommand):
             from spectre.Visualization.TransformVolumeData import (
                 transform_volume_data_command)
             return transform_volume_data_command
-        raise NotImplementedError(f"The command '{name}' is not implemented.")
+
+        available_commands = " " + "\n ".join(self.list_commands(ctx))
+        raise click.UsageError(
+            f"The command '{name}' is not implemented. "
+            f"Available commands are:\n{available_commands}")
 
 
 def print_machine(ctx, param, value):


knelli2:
@nilsvu I'm not that well-versed in python, so is there a better way to do this?

nilsvu:
Yea set `_rich_traceback_guard = True` before raising the exception. Also, try raising a `click.UsageError` instead of NotImplemented because I think that prints helpful instructions so you you may not have to print the available commands.

knelli2:
Apparently, `click.UsageError` already doesn't print a backtrace so I just went with that

nilsvu (approved):

knelli2:
@nilsvu Do you think it'd be a good idea to add a new GitHub label called `Python`? I feel like it'd be nice to have to categorize some PRs (like this one)

nilsvu:
I don't see much practical benefit, do you? 🤷‍♂️

knelli2:
Yeah it serves a similar purpose as our other labels. It's just nice to have to sort things. I think it's most similar to the `documentation` label. It's just nice to see when reviewing

nilsvu:
Just FYI, all labels that we currently use have some meaning in code review. `documentation` means "let's not nitpick too much on wording and incomplete bits and such and rather merge this sooner".

wthrowe:
It does?  I always interpreted it as an indicator that you were going to mostly be reviewing prose instead of code, so people who like doing the former would see it.

knelli2:
I had the same thought at @wthrowe. Which is how I envisioned the `Python` label to go. It would say "If you review this PR, it'll be mostly python code"

PR #4874, opened 2023-03-22
Enforce 3-con in GH numeric initial data

geoffrey4444:
## Proposed changes

Currently, when reading numeric initial data for generalized-harmonic evolutions (e.g. BBHs), the 3-index constraint is set to zero if the data provided are ADM quantities but not if the evolution variables are provided directly. This PR makes the latter case also set the 3-index constraint to zero at t=0.

### Upgrade instructions

After this change, all numeric initial data for generalized-harmonic evolutions enforce that the three-index constraint vanishes at the initial time.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/Actions/NumericInitialData.hpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/Actions/NumericInitialData.hpp
@@ -69,15 +69,16 @@ struct Adm : tuples::tagged_tuple_from_typelist<
 // - Generalized harmonic variables
 using gh_vars =
     tmpl::list<gr::Tags::SpacetimeMetric<3, Frame::Inertial, DataVector>,
-               Tags::Pi<3, Frame::Inertial>, Tags::Phi<3, Frame::Inertial>>;
+               Tags::Pi<3, Frame::Inertial>>;
 struct GeneralizedHarmonic
     : tuples::tagged_tuple_from_typelist<
           db::wrap_tags_in<OptionTags::VarName, gh_vars>> {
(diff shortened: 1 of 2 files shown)


src/Evolution/Systems/GeneralizedHarmonic/Actions/NumericInitialData.hpp
```diff
@@ -236,8 +242,8 @@ struct SetNumericInitialData {
                     numeric_initial_data));
             *pi = std::move(
                 get<Tags::Pi<3, Frame::Inertial>>(numeric_initial_data));
-            *phi = std::move(
-                get<Tags::Phi<3, Frame::Inertial>>(numeric_initial_data));
+            // Set Phi to the numerical spatial derivative of spacetime_metric
+            partial_derivative(phi, *spacetime_metric, mesh, inv_jacobian);
```
nilsvu, line 247: This overrides phi entirely so we shouldn't load it from the file. Is there a use case where we would want to load phi from file instead? Otherwise just delete it in line 72, and adjust related docs.

nilsvu (changes requested):

geoffrey4444:
Thanks for the quick review! I made the change you requested and squashed, since the change is so small

nilsvu (dismissed):

nilsvu:
CI complains, you'll have to delete Phi from the unit test too

geoffrey4444:
Rats, only had a couple of minutes yesterday before teaching, didn't think through needing to update the test. Sorry! Just pushed a fix

nilsvu (approved):

PR #4875, opened 2023-03-23
Allow different boundary conditions in Brick

nilsdeppe:
This is useful for testing 1d GRMHD problems where y & z are periodic.

### Upgrade instructions

Change input files from:
```
    BoundaryCondition: Periodic
```
to
```
    BoundaryConditionInX: Periodic
    BoundaryConditionInY: Periodic
    BoundaryConditionInZ: Periodic
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/Brick.cpp
+++ b/src/Domain/Creators/Brick.cpp
@@ -44,7 +44,9 @@ Brick::Brick(
       initial_number_of_grid_points_in_xyz_(                 // NOLINT
           std::move(initial_number_of_grid_points_in_xyz)),  // NOLINT
       time_dependence_(std::move(time_dependence)),
-      boundary_condition_(nullptr) {
+      boundary_condition_in_x_(nullptr),
+      boundary_condition_in_y_(nullptr),
+      boundary_condition_in_z_(nullptr) {
   if (time_dependence_ == nullptr) {
     time_dependence_ =
         std::make_unique<domain::creators::time_dependence::None<3>>();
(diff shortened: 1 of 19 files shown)


src/Domain/Creators/Brick.cpp
```diff
@@ -134,8 +154,13 @@ Brick::external_boundary_conditions() const {
...
   }
-  for (const auto& direction : Direction<3>::all_directions()) {
-    boundary_conditions[0][direction] = boundary_condition_->get_clone();
+  for (const auto& side : {Side::Lower, Side::Upper}) {
+    boundary_conditions[0][Direction<3>{0, side}] =
+        boundary_condition_in_x_->get_clone();
+    boundary_conditions[0][Direction<3>{1, side}] =
+        boundary_condition_in_y_->get_clone();
+    boundary_conditions[0][Direction<3>{2, side}] =
+        boundary_condition_in_z_->get_clone();
```
nilsvu: This doesn't look like periodic boundary conditions in only a subset of the directions are handled correctly, or maybe I am missing something. Could you add a test for that? Line 155 returns empty boundary conditions if just the x-direction is periodic.

nilsvu (changes requested):

src/Domain/Creators/Brick.cpp
nilsdeppe: Yea, I think you're right. I've added tested and checks to resolve this. Thanks!

nilsvu (dismissed):

nilsdeppe:
I've rebased and squashed. Thanks so much for the review!

nilsvu (approved):

PR #4877, opened 2023-03-23
Rewrite AdamsBashforth LTS to reduce allocations

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/TimeSteppers/AdamsBashforth.cpp
+++ b/src/Time/TimeSteppers/AdamsBashforth.cpp
@@ -20,6 +20,7 @@
 #include "Time/SelfStart.hpp"
 #include "Time/Time.hpp"
 #include "Time/TimeStepId.hpp"
+#include "Utilities/Algorithm.hpp"
 #include "Utilities/CachedFunction.hpp"
 #include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
@@ -109,6 +110,8 @@ OrderVector<double> constant_coefficients(const size_t order) {
   }
 }
 
+// This includes the overall factor of step size.
+//
 // Only T=double is used, but this can be used with T=Rational to
 // generate coefficient tables.
 template <typename T>
@@ -117,7 +120,7 @@ OrderVector<T> variable_coefficients(const OrderVector<T>& control_times) {
   // the last time in the list to t=0.
 
   // The coefficients are, for each j,
-  // 1/step \int_0^{step} dt ell_j(t; -control_times),
+  // \int_0^{step} dt ell_j(t; -control_times),
   // where the step size is step=-control_times.back().
 
   const size_t order = control_times.size();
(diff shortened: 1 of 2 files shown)


wthrowe:
I have no clue what clang-tidy wants me to do.  It seems to be requesting that I combine an if and else block, but the else is for a different if.

src/Time/TimeSteppers/AdamsBashforth.cpp
```diff
@@ -374,6 +374,148 @@ void AdamsBashforth::boundary_dense_output_impl(
...
+    return remote_time_;
+  }
+
+  SmallStepIterator& operator++() {
+    ASSERT(local_time_ != local_end_ and remote_time_ != remote_end_,
+           "Overran iterator");
+    auto local_candidate = std::next(local_time_);
+    auto remote_candidate = std::next(remote_time_);
+
+    if (local_candidate == local_end_ and remote_candidate == remote_end_) {
```
knelli2: Clang-tidy is complaining that there are multiple branches with the same code in each branch. It wants you to combine them (untested)

```cpp
if (local_candidate == local_end_ and remote_candidate == remote_end_) {
  local_time_ = std::move(local_candidate);
  remote_time_ = std::move(remote_candidate);
} else if (local_candidate == local_end_ or before_(*remote_candidate, *local_candidate)) {
  remote_time_ = std::move(remote_candidate);
} else if (remote_candidate == remote_end_ or before_(*local_candidate, *remote_candidate)) {
  local_time_ = std::move(local_candidate);
} else {
  local_time_ = std::move(local_candidate);
  remote_time_ = std::move(remote_candidate);
}
```

However, this might still complain because the first and last branch are identical still...

wthrowe: You've introduced a bug by reordering the cases.  If `local_candidate != local_end_` and `remote_candidate == remote_end_` you're likely hitting UB.

knelli2: Not surprised as I have no clue what's going on here. If you're unable to find a way to combine branches I'd say just `NOLINT` this whole section

wthrowe: All the variables are iterators.  (OK, technically not since `before_` is a variable, too.)

nilsdeppe: Yea, I agree that it looks like clang-tidy wants you to create bugs in your code 🙃 I think what it's worried about is that the branches have the same content and need to be kept in sync. I think reasonable ways to go about silencing it:
1. `NOLINT`
2. a lambda for shared code (though seems a bit silly to have a 3 line lambda definition for a 1 line move)
3. you might be able to actually combine 437 & 439 without breaking logic

nilsdeppe (dismissed):

src/Time/TimeSteppers/AdamsBashforth.cpp
wthrowe: 1. That's what I'm going with, assuming I can figure out which lines to mark.
2. I assume it would then complain that I'm calling the same lambda in two branches.
3. I could.  If it solved the whole problem I might have done that, but since there are two other things to NOLINT I'm going to leave it because I think it's more readable as-is.

nilsdeppe: Sounds good! :)

nilsdeppe (approved):

Issue #4592, opened 2023-01-12
Char speed control: interface AH, excision surface, control system


geoffrey4444:
- [x] Mark: function to get the control error $Q$ (on a branch, how how to test?) -- in progress (branch SizeControlSignal)
- [x] Kyle: write control error class that calls the control error function
- [x] Kyle: template shape control on measurement so use `BothHorizons` measurement for BBH and single horizon measurement for single BH
- [x] Kyle: make measurement class for size (how to test?)
- [x] Kyle: add to control system, executables
- [x] Send the excision surface and apparent horizon, with needed quantities interpolated onto them, to the control system. Connect to existing actuator logic, to adjust the size map.
 
### Detailed discussion:
@markscheel can start on this now, using fake excision surface data, until #4591 is ready.

3/22/23 @knelli2 once these steps are done, "it should just work"

knelli2:
Idea for test of Mark's size control signal function:
- Put Schwarzschild solution in so we know what to expect upon output.

geoffrey4444:
Update: size control was receiving char speeds in wrong frame (fixed in Kyle's branch)
Mark rewrote his branch to add another few arguments (like radius of excision boundary in different frames)
Ping Kyle when he gets back next week

knelli2:
Closing because interface is all working. Now it's just implementation details. Will make a new issue.

PR #4878, opened 2023-03-23
Template Shape control on measurement

knelli2:
For a single BH, we just want the SingleHorizon measurement. However, for BBH, shape control should use the same measurements as the other slow control systems (because shape is one of the slow control systems) which is the BothHorizons measurement.

## Proposed changes

Part of #4592.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Systems/Shape.hpp
+++ b/src/ControlSystem/Systems/Shape.hpp
@@ -11,6 +11,7 @@
 
 #include "ControlSystem/Component.hpp"
 #include "ControlSystem/ControlErrors/Shape.hpp"
+#include "ControlSystem/Measurements/BothHorizons.hpp"
 #include "ControlSystem/Measurements/SingleHorizon.hpp"
 #include "ControlSystem/Protocols/ControlError.hpp"
 #include "ControlSystem/Protocols/ControlSystem.hpp"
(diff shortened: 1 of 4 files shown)


nilsdeppe (dismissed):

src/ControlSystem/Systems/Shape.hpp
```diff
@@ -108,13 +115,37 @@ struct Shape : tt::ConformsTo<protocols::ControlSystem> {
...
+        const LinkedMessageId<double>& measurement_id) {
+      // The measurement event will call this for both horizons, but we only
+      // need one of the horizons. So if it is called for the wrong horizon,
+      // just do nothing.
+      if constexpr (MeasureHorizon == Horizon) {
+        auto& control_sys_proxy = Parallel::get_parallel_component<
+            ControlComponent<Metavariables, Shape>>(cache);
+
+        Parallel::simple_action<::Actions::UpdateMessageQueue<
+            QueueTags::Strahlkorper<Frame::Grid>, MeasurementQueue,
```
wthrowe: `Frame::Distorted`?  That's whet everything else in these functions uses, at least.

wthrowe (changes requested):

src/ControlSystem/Systems/Shape.hpp
knelli2: Good catch!

wthrowe (approved):

nilsdeppe (approved):

PR #4884, opened 2023-03-28
Add compute tag for time deriv of strahlkorper

knelli2:
## Proposed changes

This is needed for size control. Part of #4592.

Since the previous horizon finds aren't evenly spaced, we need a FD stencil that can handle non-uniformly spaced times. The control system Averager has this internally. So this PR also factors that code out into its own function first, so it can be used in both places.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ApparentHorizons/Tags.hpp
+++ b/src/ApparentHorizons/Tags.hpp
@@ -10,6 +10,7 @@
 
 #include "ApparentHorizons/StrahlkorperGr.hpp"
 #include "ApparentHorizons/TagsDeclarations.hpp"  // IWYU pragma: keep
+#include "ApparentHorizons/TimeDerivStrahlkorper.hpp"
 #include "DataStructures/DataBox/Tag.hpp"
 #include "DataStructures/DataBox/TagName.hpp"
 #include "DataStructures/DataVector.hpp"
@@ -53,6 +54,28 @@ struct ObserveCentersBase : db::BaseTag {};
 struct ObserveCenters : ObserveCentersBase, db::SimpleTag {
   using type = bool;
 };
+
+/// @{
+/// Tag to compute the time derivative of the coefficients of a Strahlkorper
+/// from a number of previous Strahlkorpers.
+template <typename Frame>
+struct TimeDerivStrahlkorper : db::SimpleTag {
+  using type = ::Strahlkorper<Frame>;
+};
+
+template <typename Frame>
+struct TimeDerivStrahlkorperCompute : db::ComputeTag,
+                                      TimeDerivStrahlkorper<Frame> {
+  using base = TimeDerivStrahlkorper<Frame>;
+  using return_type = typename base::type;
+  static constexpr auto function = static_cast<void (*)(
+      gsl::not_null<::Strahlkorper<Frame>*>,
+      const std::deque<std::pair<double, ::Strahlkorper<Frame>>>&)>(
+      &ah::time_deriv_of_strahlkorper<Frame>);
+
+  using argument_tags = tmpl::list<PreviousStrahlkorpers<Frame>>;
+};
+/// @}
 }  // namespace ah::Tags
 
 /// \ingroup SurfacesGroup
--- a/src/ApparentHorizons/TimeDerivStrahlkorper.cpp
+++ b/src/ApparentHorizons/TimeDerivStrahlkorper.cpp
@@ -1,0 +1,86 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ApparentHorizons/TimeDerivStrahlkorper.hpp"
+
+#include <deque>
+#include <utility>
+
+#include "DataStructures/DataVector.hpp"
+#include "NumericalAlgorithms/FiniteDifference/NonUniform1D.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace Frame {
+struct Grid;
+struct Distorted;
+struct Inertial;
+}  // namespace Frame
+
+namespace ah {
+namespace {
+template <size_t NumTimes>
+static DataVector compute_coefs(
+    const std::deque<double>& times,
+    const std::deque<const DataVector*>& coefficients) {
+  const auto weights = fd::non_uniform_1d_weights<NumTimes>(times);
+
+  DataVector new_coefficients{coefficients.front()->size(), 0.0};
+
+  for (size_t i = 0; i < NumTimes; i++) {
+    new_coefficients += *coefficients[i] * gsl::at(gsl::at(weights, 1), i);
+  }
+
+  return new_coefficients;
+}
+}  // namespace
+
+template <typename Frame>
+void time_deriv_of_strahlkorper(
+    gsl::not_null<Strahlkorper<Frame>*> time_deriv,
+    const std::deque<std::pair<double, ::Strahlkorper<Frame>>>&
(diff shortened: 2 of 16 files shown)


wthrowe:
This seems to do a lot more than add a tag.

knelli2:
Well it adds a compute tag. And the code I needed for part of the compute tag was in the averager (the FD coefficients). So I factored it out. And by the end of it all, it wasn't that many line changes overall so I just kept it in one PR.

Edit: updated the title and description

src/NumericalAlgorithms/FiniteDifference/NonUniform1D.hpp
```diff
@@ -0,0 +1,33 @@
...
+/*!
+ * \ingroup FiniteDifferenceGroup
+ * \brief Returns the weights for a 1D non-uniform finite difference stencil.
+ *
+ * These weights are for the Lagrange interpolation polynomial and its
+ * derivatives evaluated at `times[0]`.
+ *
+ * The reason that `times` was chosen to be monotonically decreasing is because
+ * the intended use of this function is with data that is stored in a
+ * `std::dequeue` where the zeroth element is the "most recent" in time and then
```
wthrowe: "deque"

src/NumericalAlgorithms/FiniteDifference/NonUniform1D.hpp
```diff
@@ -0,0 +1,33 @@
...
+
+namespace fd {
+/*!
+ * \ingroup FiniteDifferenceGroup
+ * \brief Returns the weights for a 1D non-uniform finite difference stencil.
+ *
+ * These weights are for the Lagrange interpolation polynomial and its
+ * derivatives evaluated at `times[0]`.
+ *
+ * The reason that `times` was chosen to be monotonically decreasing is because
```
wthrowe, line 19: Not a change request, just want to make sure you're aware that the math doesn't care about anything except that the values are distinct.  Probably still worth asserting the ordering as a check against the caller going haywire.

src/NumericalAlgorithms/FiniteDifference/NonUniform1D.hpp
```diff
@@ -0,0 +1,33 @@
...
+ * the intended use of this function is with data that is stored in a
+ * `std::dequeue` where the zeroth element is the "most recent" in time and then
+ * later elements are further in the "past".
+ *
+ * \param times A monotonically *decreasing* sequence of times
+ * \return A 2D array of all finite difference weights that correspond to the
+ * input times.
```
wthrowe: Which dimension is which?

src/NumericalAlgorithms/FiniteDifference/NonUniform1D.hpp
```diff
@@ -0,0 +1,33 @@
...
+#include <cstddef>
+#include <deque>
+
+namespace fd {
+/*!
+ * \ingroup FiniteDifferenceGroup
+ * \brief Returns the weights for a 1D non-uniform finite difference stencil.
+ *
+ * These weights are for the Lagrange interpolation polynomial and its
+ * derivatives evaluated at `times[0]`.
```
wthrowe: What happens if `times.size() != StencilSize`?

src/ApparentHorizons/Tags.hpp
```diff
@@ -53,6 +56,68 @@ struct ObserveCentersBase : db::BaseTag {};
...
+struct TimeDerivStrahlkorper : db::SimpleTag {
+  using type = ::Strahlkorper<Frame>;
+};
+
+template <typename Frame>
+struct TimeDerivStrahlkorperCompute : db::ComputeTag,
+                                      TimeDerivStrahlkorper<Frame> {
+  using base = TimeDerivStrahlkorper<Frame>;
+  using return_type = typename base::type;
+  static void function(
```
wthrowe: Make this a free function.  It seems like it could be useful without a box.

src/ApparentHorizons/Tags.hpp
```diff
@@ -53,6 +56,68 @@ struct ObserveCentersBase : db::BaseTag {};
 struct ObserveCenters : ObserveCentersBase, db::SimpleTag {
   using type = bool;
 };
+
+template <typename Frame>
+struct TimeDerivStrahlkorper : db::SimpleTag {
```
wthrowe, line 62: [optional] We have a `Tags::dt<...>` representing time derivatives.

tests/Unit/ApparentHorizons/Test_Tags.cpp
```diff
@@ -205,6 +205,43 @@ struct SomeTag : db::SimpleTag {
...
+        db::AddComputeTags<
+            ah::Tags::TimeDerivStrahlkorperCompute<Frame::Inertial>>>(
+        previous_strahlkorpers);
+
+    const DataVector& time_deriv_strahlkorper_coefs =
+        db::get<ah::Tags::TimeDerivStrahlkorper<Frame::Inertial>>(box)
+            .coefficients();
+
+    // Since we made all the Strahlkorpers the same, the time deriv should be
+    // zero.
```
wthrowe: Test a case where the function actually does something.  One point with two times would be fine.

src/ParallelAlgorithms/Interpolation/Callbacks/FindApparentHorizon.hpp
```diff
@@ -64,13 +64,15 @@ namespace callbacks {
 /// type alias called `post_horizon_find_callbacks`, which is a list of
 /// structs, each of which has a function
 ///
-/// \snippet ApparentHorizons/Test_ApparentHorizonFinder.cpp post_horizon_find_callback_example
+/// \snippet ApparentHorizons/Test_ApparentHorizonFinder.cpp
+/// post_horizon_find_callback_example
```
wthrowe: Broke the docs.  Again below.

wthrowe (changes requested):

src/NumericalAlgorithms/FiniteDifference/NonUniform1D.hpp
knelli2: An error. There's an ASSERT

knelli2:
Added a new commit where I add the free function, so the commit that adds the tag is a lot shorter.

src/NumericalAlgorithms/FiniteDifference/NonUniform1D.hpp
wthrowe: Document that.

src/ApparentHorizons/TimeDerivStrahlkorper.cpp
```diff
@@ -0,0 +1,99 @@
...
+namespace Frame {
+struct Grid;
+struct Distorted;
+struct Inertial;
+}  // namespace Frame
+
+namespace ah {
+namespace detail {
+template <size_t NumTimes>
+static DataVector compute_coefs(
```
wthrowe, line 25: Anonymous namespace.

src/ApparentHorizons/TimeDerivStrahlkorper.cpp
```diff
@@ -0,0 +1,99 @@
...
+
+  time_deriv->coefficients() = std::move(new_coefficients);
+}
+
+namespace detail {
+#define TIMES(data) BOOST_PP_TUPLE_ELEM(0, data)
+
+#define INSTANTIATE(_, data)                      \
+  template DataVector compute_coefs<TIMES(data)>( \
+      const std::deque<double>&, const std::deque<const DataVector*>&);
```
wthrowe: You don't need to explicitly instantiate this, since it is only used in this file.

wthrowe (changes requested):

wthrowe (approved):

PR #4885, opened 2023-03-28
Transform time derivative of regular field to grid frame

nikwit:
## Proposed changes
This transforms the time derivative of the regular field to the grid frame before sending it to the worldtube. The advective term needed for this is stored in an extra tag which is used to transform the worldtube solution back to the inertial frame before applying boundary conditions to the elements.

For the 0th and 1st order this does not matter but for 2nd order evolution this is required.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
@@ -19,6 +19,7 @@
 #include "Domain/Structure/ExcisionSphere.hpp"
 #include "Domain/Structure/IndexToSliceAt.hpp"
 #include "Domain/Tags.hpp"
+#include "Domain/TagsTimeDependent.hpp"
 #include "Evolution/Systems/CurvedScalarWave/System.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp"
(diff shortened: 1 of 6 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
```diff
@@ -134,6 +134,31 @@ struct SendToWorldtube {
           get(dot_product(face_shift, face_phi)) -
           get(get<::Tags::dt<CurvedScalarWave::Tags::Psi>>(
               puncture_field.value()));
+
+      // is an optional so we can't put the tag into variables
+      auto& mesh_velocity_on_face =
```
nilsdeppe, line 145: Reading this I worry that you are overwriting `shift` with the velocity. Can you comment that this is okay and why, just for whoever looks at it in the future? Maybe even adding equations would be useful?

Also, should there be an `if (mesh_velocity.has_value())` to guard against errors on a static mesh?

nilsdeppe (changes requested):
LGTM, just one suggestion. Please squash right away

src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
nikwit: Thank you. I added a check for the mesh velocity and a comment that the shift allocation is free at this point in the action.

nilsdeppe: Thanks!

nilsdeppe (approved):

PR #4891, opened 2023-03-30
Register BNS observers with Factory

ffoucart:
## Proposed changes

Modifications needed to use BNS observers within spectre evolutions

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

A few modifications where needed in order to make it possible to use the new observers and the integrals of conserved quantities in BNS evolution.

(1) Adding the new observers to the list of observers in GhValenciaDivCleanBase.hpp
(2) Adding the tags that we would like to integrate for diagnostics purposes (e.g. MassWeightedInternalEnergyCompute) to the observe_field list in GhValenciaDivCleanBase.hpp (should we instead have a variable elsewhere in the code with the list of integrands we want to use?)
(3) Because SpatialVelocityOneForm is not at this point a db::ComputeTag and it is not stored in memory, I changed the input of an integrand from "SpatialVelocityOneForm" to "SpatialVelocity" and "SpatialMetric" and modified the relevant tests.
(4) Create an integrand_list variable containing the tags to be used in the VolumeIntegral observer (we can't input all tags, because the current version of VolumeIntegral does not handle optionals... and there are a lot of tensors there whose integral is pretty meaningless anyways)

### DIFF
--- a/src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
+++ b/src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
@@ -128,6 +128,8 @@
 #include "ParallelAlgorithms/Actions/MutateApply.hpp"
 #include "ParallelAlgorithms/Actions/TerminatePhase.hpp"
 #include "ParallelAlgorithms/Events/Factory.hpp"
+#include "ParallelAlgorithms/Events/ObserveAtExtremum.hpp"
+#include "ParallelAlgorithms/Events/ObserveVolumeIntegrals.hpp"
 #include "ParallelAlgorithms/EventsAndTriggers/Actions/RunEventsAndTriggers.hpp"
 #include "ParallelAlgorithms/EventsAndTriggers/Actions/RunEventsOnFailure.hpp"
 #include "ParallelAlgorithms/EventsAndTriggers/Completion.hpp"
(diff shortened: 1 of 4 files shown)


ffoucart:
See "Further Comments" in the original message for some open questions about this PR.

src/ParallelAlgorithms/Events/Factory.hpp
```diff
@@ -15,11 +15,11 @@
 namespace dg::Events {
 template <size_t VolumeDim, typename TimeTag, typename Fields,
           typename NonTensorComputeTagsList, typename ArraySectionIdTag = void>
-using field_observations = tmpl::flatten<
-    tmpl::list<ObserveFields<VolumeDim, TimeTag, Fields,
```
knelli2: This appears to only be a formatting change. Can you remove this or put it in a separate commit?

knelli2 (changes requested):
You can squash this in directly.

I really don't see what could make this time out for just one of our CI builds. Has it been timing out on the CI on your fork as well? When you squash, rebase on latest develop and lets see if it happens a 3rd time.

Edit: Somebody else should review to answer the open questions in the description. I just glanced at things to try and figure out why CI timed out and just noticed one thing.

src/ParallelAlgorithms/Events/Factory.hpp
ffoucart: Yes, just a remnant of the first method I tried to implement this... I squashed the change, as it doesn't do anything.

ffoucart:
((Latest version is rebased against develop 04/04))

src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
```diff
@@ -427,6 +434,13 @@ struct GhValenciaDivCleanTemplateBase<
               volume_dim, Frame::Inertial>,
           ::Events::Tags::ObserverCoordinatesCompute<volume_dim,
                                                      Frame::Inertial>>>;
+  using integrand_fields = tmpl::push_back<tmpl::append<
```
nilsvu: delete the empty `push_back`

nilsvu (changes requested):
small thing, squash it in

ffoucart:
Fixed and squashed.

nilsvu (approved):

PR #4896, opened 2023-03-31
Add `TimeDerivative` action to worldtube

nikwit:
## Proposed changes
For the second order worldtube evolution, we need to evolve the expanded scalar wave equation inside the worldtube. This PR adds the required ODE.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/TimeDerivative.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/TimeDerivative.hpp
@@ -1,0 +1,108 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Variables.hpp"
+#include "Domain/Tags.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace CurvedScalarWave::Worldtube {
+
+/// @{
+/*!
+ * \brief Calculates the time derivative of `Psi0`, the constant coefficient of
+ * the expansion of `Psi`.
+ *
+ * \details The derivation comes from expanding the scalar wave equation to
+ * second order and reads
+ *
+ * \f{equation}
+ * g^{00}_0 \ddot{\Psi}^R_0(t_s) + 2 g_0^{0i}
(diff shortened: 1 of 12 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/TimeDerivative.hpp
```diff
@@ -0,0 +1,129 @@
...
+ *
+ * \f{equation}
+ * g^{00}_0 \ddot{\Psi}^R_0(t_s) + 2 g_0^{0i}
+ * \dot{\Psi}^N_i(t_s) + 2 g_0^{ij} \Psi^N_{\langle ij \rangle}(t_s)
+ * + \frac{2 \delta_{ij}  g_0^{ij}}{R^2}
+ * \left(\Psi^N_0(t_s) - \Psi^R_0(t_s) \right) -
+ * \Gamma_0^0\dot{\Psi}R_0(t_s) -  \Gamma_0^i \Psi_i^N(t_s) = 0.
+ *\f}
+ *
+ * Here, \f$ \Gamma^\mu_0 \f$ and \f$ g^{\mu \nu}_0 \f$ are the trace of the
```
nilsdeppe, line 38: Is this the numerical from the DG domain or what is obtained analytically? Can you note what the `_0` here denotes? While reading the docs I was curious why `_0`

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/TimeDerivative.hpp
```diff
@@ -0,0 +1,129 @@
...
+      variables_tag,
+      Stf::Tags::StfTensor<Tags::PsiWorldtube, 0, Dim, Frame::Grid>,
+      Stf::Tags::StfTensor<Tags::PsiWorldtube, 1, Dim, Frame::Grid>,
+      Stf::Tags::StfTensor<Tags::PsiWorldtube, 2, Dim, Frame::Grid>,
+      Stf::Tags::StfTensor<::Tags::dt<Tags::PsiWorldtube>, 1, Dim, Frame::Grid>,
+      gr::Tags::InverseSpacetimeMetric<Dim, Frame::Grid, double>,
+      gr::Tags::TraceSpacetimeChristoffelSecondKind<Dim, Frame::Grid, double>,
+      Tags::ExcisionSphere<Dim>>;
+
+  static void apply(
```
nilsdeppe, line 61: Move definition to `.cpp` file?

nilsdeppe (changes requested):
LGTM, couple small questions/suggestions. Please squash immediately :)

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/TimeDerivative.hpp
nikwit: It's the 0th order of the Taylor expansion of these tensors i.e. simply  evaluated at the position of the particle. The first order would be the first derivative and so on. I don't really want to write the full expansion here because that would require a lot more definitions but will give the right equation number from the paper once we submit to the arxiv.

nikwit:
Thank you, I squashed in your suggestions :)

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/TimeDerivative.hpp
nilsdeppe: Okay, thanks!

nilsdeppe (approved):

PR #4897, opened 2023-03-31
Add `spectre status` CLI endpoint 

nilsvu:
## Proposed changes

Gives an overview of simulations running on the machine. Try it like this (e.g. on Wheeler):

```sh
$ ./bin/spectre status
```

Here's a recent sample output with the `-a / --allusers` flag (any resemblance of names to real persons is purely coincidental, and @pajkosmi is surely not setting Wheeler on fire):

Bildschirm­foto 2023-04-01 um 21 56 18

### Upgrade instructions

Add `pandas` and `humanize` to your Python environment. An easy way to update your Py env is this:

```
# in your Py env
pip install -r support/Python/requirements.txt
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/ReadH5.py
+++ b/src/Visualization/Python/ReadH5.py
@@ -1,10 +1,13 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
+from typing import List, Union
+
 import h5py
+import numpy as np
 
 
-def available_subfiles(h5file, extension):
+def available_subfiles(h5file: h5py.File, extension: str) -> List[str]:
     """List all subfiles with the given 'extension' in the 'h5file'.
 
     Parameters
@@ -24,3 +27,31 @@ def available_subfiles(h5file, extension):
 
     h5file.visit(visitor)
     return subfiles
+
+
+def to_dataframe(open_subfile: Union[h5py.Dataset, "spectre.IO.H5.H5Dat"]):
+    """Convert a '.dat' subfile to a Pandas DataFrame
+
+    This function isn't particularly complex, but it allows to convert a
+    subfile to a DataFrame in a single statement like this:
+
+        to_dataframe(open_h5_file["Norms.dat"])
+
+    Without this function, you would have to store the subfile in an extra
+    variable to access its "Legend" attribute.
+
+    Arguments:
+      open_subfile: An open h5py subfile representing a SpECTRE dat file,
+        or a spectre.IO.H5.H5Dat subfile, typically from a reductions file.
+
+    Returns: Pandas DataFrame with column names read from the "Legend"
+      attribute of the dat file.
+    """
+    import pandas as pd
+    try:
+        # SpECTRE H5 dat subfile
+        return pd.DataFrame(np.asarray(open_subfile.get_data()),
+                            columns=open_subfile.get_legend())
+    except AttributeError:
+        # h5py subfile
+        return pd.DataFrame(open_subfile, columns=open_subfile.attrs["Legend"])
--- a/support/Python/requirements.txt
+++ b/support/Python/requirements.txt
@@ -17,10 +17,14 @@
 # autocompletion
 click
 h5py >= 3.5.0
+# Humanize: For human-readable output like "2 hours ago"
+humanize
 # - We constrain the Numpy version because v1.22+ segfaults in Pypp tests, see
 #   issue: https://github.com/sxs-collaboration/spectre/issues/3844
 numpy < 1.22.0
 matplotlib
+# Pandas: To work with columns of data
+pandas
 pyyaml
 # Rich: to format CLI output and tracebacks
 rich >= 10.11.0
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -700,7 +700,8 @@ ${{ env.CACHE_KEY_SUFFIX }}"
       # to create the cache.
       SPECTRE_SPACK_DEPS: >-
         blaze boost brigand catch2 charmpp gsl hdf5 libsharp libxsmm openblas
-        py-h5py py-numpy py-matplotlib py-pybind11 py-scipy py-pyyaml yaml-cpp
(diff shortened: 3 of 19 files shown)


support/Python/requirements.txt
```diff
@@ -17,10 +17,14 @@
...
 click
 h5py >= 3.5.0
+# Humanize: For human-readable output like "2 hours ago"
+humanize
 # - We constrain the Numpy version because v1.22+ segfaults in Pypp tests, see
 #   issue: https://github.com/sxs-collaboration/spectre/issues/3844
 numpy < 1.22.0
 matplotlib
+# Pandas: To work with columns of data
+pandas
```
knelli2, line 27: I think we have to add these new pydeps to the docker container for CI to pass.

knelli2 (commented):
Not a full review.

This is really awesome!! I've tried it out on wheeler and it works really well! I have some nitpicks about the user interface which I think would be easier to talk about in person rather than starting super long github comments.

support/Python/requirements.txt
```diff
@@ -17,10 +17,14 @@
 # autocompletion
 click
 h5py >= 3.5.0
+# Humanize: For human-readable output like "2 hours ago"
```
nilsdeppe, line 20: I don't find "2 hours ago" very "human-readable". I care about the exact time, not some rounding that somebody else thinks is "good"

knelli2: It looks like this is only used for printing how long ago a job ended/started, not any scientific quantities. I think this is a reasonable use for this package and it looks readable.

nilsdeppe: I know what it's used for, I still want the exact time. I think adding a dependency to make output less clear isn't a good way to go...

nilsvu: I found something like "2 hours ago" more useful than showing a timestamp, but I don't feel super strongly about it.

nilsdeppe: My experience with GitHub switching to "X ago" was that it made it really annoying to figure out when anything was done. "2 days ago" leaves a huge window open, while "Apr. 1, 10am" is still easy for me to figure out it was 2 days ago, but I now also know immediately when the event actually occurred. My biggest concern is that with a terminal-based interface I probably (I say probably because I don't know what fancy voodoo magic terminals can do nowadays) can't hover my cursor over the "X ago" to get a detailed number (the fact that I find the hovering annoying is beside the point, I just want an easy way to get the exact number). I guess my take on it is that in my experience doing a bit of estimating to get "X ago" is easier than going the other way around

nilsvu: > to make output less clear

It's less precise but arguably more clear, in the sense that I can glance over it quickly and understand how long ago the job finished. That's why you wouldn't print, say, seconds since 1970 because you have no idea what time that is, even though it's very precise. For comparison, here's the formatted date vs the "humanized" output:

Bildschirm­foto 2023-04-03 um 13 33 32

Bildschirm­foto 2023-04-03 um 13 33 51

I like the second more, but wouldn't object if reviewers prefer the first.

nilsdeppe: I think "X hours ago" is okay-ish. But "X days ago" is bad, that's not helpful. "last week" is terrible, borderline useless IMO. I find the date is easier to deal with, on the hours timescale it's not really a big deal. I think "seconds since 1970" is an exaggerated comparison and not really relevant to the discussion (nor is it more precise than what you printed with DATE TIME).

One thing we could do is have a `~/.spectre` file that we read to allow people to customize these sorts of things. (Probably better would be `~/.config/spectre/*`

nilsvu: I'll just do timestamps for now. We can figure out nicer formatting later.

nilsvu: What I did now is this:

Bildschirm­foto 2023-04-03 um 13 59 19

For the date part it shows "today", "yesterday", or a date like "Mar 20". If the year is not the current year it will also append the year like "Mar 20 2022". Sound ok?

nilsdeppe: I think this is great! Thanks!

src/Visualization/Python/ReadH5.py
```diff
@@ -24,3 +24,25 @@ def visitor(name):
...
+    This function isn't particularly complex, but it allows to convert a
+    subfile to a DataFrame in a single statement like this:
+
+        to_dataframe(open_h5_file["Norms.dat"])
+
+    Without this function, you would have to store the subfile in an extra
+    variable to access its "Legend" attribute.
+
+    Arguments:
+      open_subfile: An open h5py subfile representing a SpECTRE dat file,
```
nilsdeppe, line 44: For my own knowledge, would this work with data read by our H5 interface? Or how hard would it be to add that?

nilsdeppe (dismissed):
LGTM!

knelli2 (dismissed):
I've tried this on wheeler and it works great!

There aren't really tests for the main part of this code which is fine for now. We should probably consider adding SLURM to the CI container so we can set up fake runs to check we get the proper output. But that's way beyond this PR so we can talk about adding it later.

src/Visualization/Python/ReadH5.py
nilsvu: Pushed a fixup that adds support for our H5 dat files. Note that I haven't used our H5 dat interface in Python much at all because dat files are so easy to load with just plain h5py and pandas. But can't hurt to support it I guess.

knelli2 (commented):
Looks good

nilsdeppe (commented):
LGTM! Please squash and rebase!

nilsdeppe (dismissed):

knelli2 (approved):

nilsdeppe (approved):

wthrowe:
Can you add installing the new dependencies to the upgrade instructions?

nilsvu:
done 👍

PR #4898, opened 2023-04-01
Add 2nd order puncture field

nikwit:
## Proposed changes
Adds the 2nd order puncture field


src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
```diff
@@ -1545,4 +1546,2893 @@ void puncture_field_1(
                  dv_212)));
 }
 
+// NOLINTNEXTLINE(google-readability-function-size, readability-function-size)
+void puncture_field_2(
+    gsl::not_null<Variables<tmpl::list<
```
nilsdeppe: `const gsl::not_null`

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
```diff
@@ -27,11 +27,12 @@ void puncture_field(
     puncture_field_0(result, coords, time, orbital_radius, bh_mass);
```
nilsdeppe, line 27: Given the shear number of lines at the different orders, I think it would be good to split them into different `.cpp` files to allow parallel compilation. I'm a bit worried that this will be a bit expensive to compile at once. But you have experience with this, so use your judgement!

nilsdeppe (changes requested):
LGTM, some suggestions. Please squash directly whatever changes you decide to make :)

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.cpp
nikwit: I have moved them into separate cpp files. The files don't take particularly long to compile but it makes it easier to find the function starts^^

nilsdeppe (approved):

PR #4899, opened 2023-04-01
Add `ObserveLineSegment` to worldtube

nikwit:
## Proposed changes
adds two InterpolationTargets to the worldube which are used to observed the co-moving x- and z-axis.

### DIFF
--- a/src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
+++ b/src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
@@ -75,11 +75,16 @@
 #include "ParallelAlgorithms/EventsAndTriggers/Trigger.hpp"
 #include "ParallelAlgorithms/Interpolation/Actions/ElementInitInterpPoints.hpp"
 #include "ParallelAlgorithms/Interpolation/Actions/InitializeInterpolationTarget.hpp"
+#include "ParallelAlgorithms/Interpolation/Actions/InterpolatorRegisterElement.hpp"
+#include "ParallelAlgorithms/Interpolation/Callbacks/ObserveLineSegment.hpp"
 #include "ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp"
 #include "ParallelAlgorithms/Interpolation/Events/InterpolateWithoutInterpComponent.hpp"
 #include "ParallelAlgorithms/Interpolation/InterpolationTarget.hpp"
+#include "ParallelAlgorithms/Interpolation/PointInfoTag.hpp"
 #include "ParallelAlgorithms/Interpolation/Protocols/InterpolationTargetTag.hpp"
+#include "ParallelAlgorithms/Interpolation/Protocols/PostInterpolationCallback.hpp"
 #include "ParallelAlgorithms/Interpolation/Tags.hpp"
+#include "ParallelAlgorithms/Interpolation/Targets/LineSegment.hpp"
 #include "ParallelAlgorithms/Interpolation/Targets/Sphere.hpp"
 #include "PointwiseFunctions/AnalyticData/AnalyticData.hpp"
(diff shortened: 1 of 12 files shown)


src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
```diff
@@ -158,6 +163,28 @@ struct EvolutionMetavars {
       tmpl::list<::Events::Tags::ObserverMeshCompute<volume_dim>,
                  deriv_compute>;
 
+  template <size_t Number>
+  struct Axis : tt::ConformsTo<intrp::protocols::InterpolationTargetTag> {
+    static std::string name() { return "Axis" + std::to_string(Number); }
```
knelli2: It looks like you're observing Psi on these axes. Maybe change this name to be `PsiAlongNAxis` where N is either "X" or "Z". That way it's clear in the input file what's being observed.

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
```diff
@@ -26,6 +26,12 @@ ResourceInfo:
     WorldtubeSingleton:
       Proc: Auto
       Exclusive: false
+    Axis1:
+      Proc: Auto
```
knelli2: If all of these are "Auto" and "false, you can just set the overall singleton block to "Singletons: Auto" to simplify things

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
```diff
@@ -94,6 +100,16 @@ EventsAndTriggers:
...
+  - - Slabs:
+        EvenlySpaced:
+          Interval: 1000
+          Offset: 0
+    - - Axis1
+  - - Slabs:
+        EvenlySpaced:
+          Interval: 1000
+          Offset: 0
+    - - Axis2
```
knelli2: Add this event right after Axis1 and that trigger since these are the same triggers

src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
```diff
@@ -158,6 +163,28 @@ struct EvolutionMetavars {
       tmpl::list<::Events::Tags::ObserverMeshCompute<volume_dim>,
                  deriv_compute>;
 
+  template <size_t Number>
```
knelli2, line 166: If you're observing along the x and z axes, use 0 and 2 respectively rather than 1 and 2 just so the index of the coordinates matches up (even though you're not actually indexing anything)

knelli2 (changes requested):
You can squash all these in

src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
nikwit: I changed it to `PsiAlongAxis` but since the axis can be specified in the input file to be any `LineSegment`, I'd rather not give them specific names so it does not look to be hard-coded. This way we also don't need a default case if `Number` is larger than 2.

nilsdeppe (dismissed):

src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
knelli2: Oh wait, this "Number" is just an index so you can have multiple of these targets in the input file. I didn't realize this. In that case, the 0 and 2 don't make much sense actually (because they don't correspond to indices of coordinates). 0 and 1 would be better. so could you switch it again? Sorry about that...

knelli2 (changes requested):

nikwit:
no problem, I undid the change.

nilsdeppe (approved):

knelli2 (approved):
Ok looks good now

PR #4901, opened 2023-04-03
StrahlkorperCoordsInDifferentFrame: time-indep map

markscheel:
StrahlkorperCoordsInDifferentFrame now handles the case where a block is time-independent, but the input and output coords have different Frame Tags, e.g. one is Inertial and the other is Grid.  This can happen even though (for the time-independent case) Grid, Inertial, and Distorted are really the same frame.

Previously StrahlkorperCoordsInDifferentFrame threw a runtime error in that case.

This fixes an error that has not yet occurred, but would have occurred in (for example) the single-BH executable for a time-independent map once more frame-related things are added.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2 (approved):
Looks good!

nilsdeppe (approved):

PR #4773, opened 2023-02-25
Enable AMR for ScalarWave

kidder:
In case anyone wants to test a p-refinement criteria, this adds the code necessary to do p-refinement for a scalar wave.

The input file:  `PlaneWave1DObserveExample.yaml` demonstrates this by oscillating between two resolutions during the evolution.

The `amr::Projector` mutator will need to be factored/generalized/rewritten to handle other evolution systems, local time-stepping, ...

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #4906, opened 2023-04-04
Add action to initialize a newly created child element during adaptive mesh refinement

kidder:
## Proposed changes

- This adds the action that will be called to initialize the data on each new element that was created by splitting an existing element.
- At the moment this only initializes the Element and Mesh, which is sufficient to test the infrastructure of AMR.
- A future PR will add the functionality for initializing the data needed for an evolution or elliptic system, most likely by calling a system dependent mutator (see the AdjustDomain action in #4773 for what I currently have in mind).  All the data needed by such a mutator is already being passed into the action in the `parent_items`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Domain/Amr/Test_NeighborsOfChild.cpp
+++ b/tests/Unit/Domain/Amr/Test_NeighborsOfChild.cpp
@@ -1,0 +1,186 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <array>
+#include <cstddef>
+#include <random>
+#include <string>
+#include <unordered_map>
+#include <vector>
+
+#include "Domain/Amr/Flag.hpp"
+#include "Domain/Amr/Helpers.hpp"
+#include "Domain/Amr/NeighborsOfChild.hpp"
+#include "Domain/Amr/NewNeighborIds.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/DirectionMap.hpp"
+#include "Domain/Structure/Element.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Domain/Structure/Neighbors.hpp"
+#include "Domain/Structure/SegmentId.hpp"
+#include "Domain/Structure/Side.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "Helpers/Domain/Amr/NeighborFlagHelpers.hpp"
+#include "Helpers/Domain/Structure/NeighborHelpers.hpp"
+#include "Utilities/CartesianProduct.hpp"
+#include "Utilities/ConstantExpressions.hpp"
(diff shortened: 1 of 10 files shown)


tests/Unit/Helpers/Domain/Structure/NeighborHelpers.cpp
```diff
@@ -464,7 +464,9 @@ std::vector<Neighbors<Dim>> valid_neighbors(
   const OrientationMap<Dim> aligned{};
   const size_t block_id = element_id.block_id();
   const size_t neighbor_block_id =
-      (face_type == FaceType::Block ? block_id + 1 : block_id);
+      (face_type == FaceType::Block
+           ? block_id + 2 * normal_dim + (side == Side::Lower ? 1 : 2)
+           : block_id);
```
wthrowe, line 469: There are valid domains where the neighbor blocks in two directions are the same.  Can the AMR code handle that?

tests/Unit/Domain/Amr/Test_NeighborsOfChild.cpp
```diff
@@ -0,0 +1,187 @@
...
+      TestHelpers::amr::valid_neighbor_flags(
+          element.id(), element_flags,
+          element.neighbors().at(Direction<Dim>::upper_xi()));
+  for (const auto& lower_xi_neighbor_flags : valid_lower_xi_neighbor_flags) {
+    for (const auto& upper_xi_neighbor_flags : valid_upper_xi_neighbor_flags) {
+      auto joined_flags = lower_xi_neighbor_flags;
+      for (const auto& flags : upper_xi_neighbor_flags) {
+        joined_flags.emplace(flags);
+      }
+      //      if constexpr (Dim == 1) {
```
wthrowe: Remove.

wthrowe (changes requested):
Also, `Unit.Domain.Amr.NeighborsOfChild` timed out on one of the CI builds.

tests/Unit/Helpers/Domain/Structure/NeighborHelpers.cpp
kidder: Yes.  The reason I don't allow it in the test is that I would have to add an extra step when creating the valid_neighbor_flags that ensure that a given element only occurs once and be a valid choice in all directions it is a neighbor.  In the AMR code, these are conditions are guaranteed to be satisfied by the passed in `neighbor_flags`.  So it is a choice to not make an already complicated test more complicated.   I can test an explicit example to the test if you wish.

wthrowe: OK, thanks for explaining.  I think it's fine as-is, then.

kidder:
squashed in the requested change, increasing the TimeOut for the test, and reducing the number of cases tested

wthrowe (approved):

PR #4910, opened 2023-04-05
Add option to use hard coded time dep maps in Sphere

knelli2:
## Proposed changes

Adds the ability to choose to use a TimeDependence or hard-coded time dependent maps in the Sphere domain creator. To have no time dependent maps, still specify `None`. Currently, the hard coded maps are

- Inner-most shell: a size and shape map from grid to distorted frame, identity map from distorted to inertial frame
- All other shells: no distorted frame. Only an identity map from grid to inertial
- Inner cube: Cannot have an inner cube (at the moment) when using hard coded time dependent maps

In the future we can add more maps and change where we place them, but this is sufficient for me to test shape control on a single BH while maintaining the flexibility of having TimeDependences for others.

### Upgrade instructions

If you are using the `Sphere` domain creator, the name of the time dependent options has changed from
```yaml
Sphere:
  ...
  TimeDependence:
    ...
```
to
```yaml
Sphere:
  ...
  TimeDependentMaps:
    ...
```

To specify no time dependent maps, still use `None` as before like
```yaml
Sphere:
  ...
  TimeDependentMaps: None
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/Sphere.cpp
+++ b/src/Domain/Creators/Sphere.cpp
@@ -7,6 +7,7 @@
 #include <cmath>
 #include <cstddef>
 #include <memory>
+#include <optional>
 #include <unordered_map>
 #include <utility>
 #include <variant>
@@ -77,8 +78,7 @@ Sphere::Sphere(
     std::vector<double> radial_partitioning,
     const typename RadialDistribution::type& radial_distribution,
     ShellWedges which_wedges,
-    std::unique_ptr<domain::creators::time_dependence::TimeDependence<3>>
-        time_dependence,
+    std::optional<TimeDepOptionType> time_dependent_options,
     std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
         outer_boundary_condition,
     const Options::Context& context)
@@ -90,7 +90,7 @@ Sphere::Sphere(
       equatorial_compression_(equatorial_compression),
       radial_partitioning_(std::move(radial_partitioning)),
       which_wedges_(which_wedges),
-      time_dependence_(std::move(time_dependence)),
+      time_dependent_options_(std::move(time_dependent_options)),
       outer_boundary_condition_(std::move(outer_boundary_condition)) {
   if (inner_radius_ > outer_radius_) {
     PARSE_ERROR(context,
@@ -149,11 +149,11 @@ Sphere::Sphere(
   }
 
   // Determine number of blocks
-  const size_t num_blocks_per_shell =
+  num_blocks_per_shell_ =
       which_wedges_ == ShellWedges::All
           ? 6
           : which_wedges_ == ShellWedges::FourOnEquator ? 4 : 1;
-  num_blocks_ = num_blocks_per_shell * num_shells_ + (fill_interior_ ? 1 : 0);
+  num_blocks_ = num_blocks_per_shell_ * num_shells_ + (fill_interior_ ? 1 : 0);
 
   // Create block names and groups
   static std::array<std::string, 6> wedge_directions{
@@ -244,9 +244,27 @@ Sphere::Sphere(
         "domain creator.");
   }
 
-  if (time_dependence_ == nullptr) {
-    time_dependence_ =
-        std::make_unique<domain::creators::time_dependence::None<3>>();
+  if (time_dependent_options_.has_value()) {
+    use_hard_coded_maps_ =
+        std::holds_alternative<sphere::TimeDependentMapOptions>(
+            time_dependent_options_.value());
+
+    if (use_hard_coded_maps_) {
+      if (fill_interior_) {
+        PARSE_ERROR(context,
+                    "Currently cannot use hard-coded time dependent maps with "
+                    "an inner cube. Use a TimeDependence instead.");
+      }
+
+      // Build the maps. We only apply the maps in the inner-most shell. The
+      // inner radius is what's passed in, but the outer radius is the outer
+      // radius of the inner-most shell so we have to see how many shells we
+      // have
+      std::get<sphere::TimeDependentMapOptions>(time_dependent_options_.value())
+          .build_maps(std::array{0.0, 0.0, 0.0}, inner_radius_,
+                      radial_partitioning_.empty() ? outer_radius_
+                                                   : radial_partitioning_[0]);
+    }
   }
 }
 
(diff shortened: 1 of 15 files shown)


src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -0,0 +1,185 @@
...
+
+  template <typename SourceFrame, typename TargetFrame>
+  using IdentityForComposition =
+      domain::CoordinateMap<SourceFrame, TargetFrame, IdentityMap>;
+  using GridToDistortedComposition =
+      domain::CoordinateMap<Frame::Grid, Frame::Distorted, ShapeMap,
+                            CompressionMap>;
+  using GridToInertialComposition =
+      domain::CoordinateMap<Frame::Grid, Frame::Inertial, ShapeMap,
+                            CompressionMap, IdentityMap>;
```
wthrowe: What is the reason for composing with an identity map?  Please add a comment explaining.

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -0,0 +1,122 @@
...
+#include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp"
+#include "Domain/FunctionsOfTime/FixedSpeedCubic.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/PiecewisePolynomial.hpp"
+#include "Domain/FunctionsOfTime/QuaternionFunctionOfTime.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/YlmSpherepack.hpp"
+
+namespace domain::creators::sphere {
+TimeDependentMapOptions::TimeDependentMapOptions(
+    double initial_time, std::array<double, 3> initial_size_values,
```
wthrowe: Const these.  The third argument is already done.

src/Domain/Creators/Sphere.cpp
```diff
@@ -7,6 +7,7 @@
 #include <cmath>
```
wthrowe, line 7: Typo in the commit message: "Mpas".

src/Domain/Creators/Sphere.cpp
```diff
@@ -316,20 +334,56 @@ Domain<3> Sphere::create_domain() const {
     }
   }
 
-  Domain<3> domain{std::move(coord_maps),       corners,      {},
+  Domain<3> domain{std::move(coord_maps), /*       corners,      {},*/
```
wthrowe: Remove?

src/Domain/Creators/Sphere.cpp
```diff
@@ -377,14 +431,24 @@ Sphere::external_boundary_conditions() const {
...
+    return std::visit(
+        [&initial_expiration_times](const auto& arg) {
+          using T = std::decay_t<decltype(arg)>;
+          if constexpr (std::is_same_v<T, sphere::TimeDependentMapOptions>) {
+            return arg.create_functions_of_time(initial_expiration_times);
+          } else {
+            return arg->functions_of_time(initial_expiration_times);
+          }
+        },
+        time_dependent_options_.value());
```
wthrowe: [optional] If you're just going to branch on the type anyway, it's I think it's cleaner to skip the visit and just do the access directly:
```c++
if (std::holds_alternative<sphere::TimeDependentMapOptions>(
        time_dependent_options_.value())) {
  return std::get<sphere::TimeDependentMapOptions>(
      time_dependent_options_.value())
      .create_functions_of_time(initial_expiration_times);
} else {
  return std::get<
      std::unique_ptr<domain::creators::time_dependence::TimeDependence<3>>>(
          time_dependent_options_.value())
          ->functions_of_time(initial_expiration_times);
}
```
(Type aliases and local reference variables might make this more readable.)

src/Domain/Creators/Sphere.hpp
```diff
@@ -125,10 +128,40 @@ namespace domain::creators {
...
+ * `domain::creators::time_dependence::RotationAboutZAxis`) to add time
+ * dependent maps. This method will add the same maps to all blocks in the
+ * domain. This method can be used with an inner cube or with an excision
+ * surface.
+ *
+ * ##### Hard-coded time dependent maps
+ * The Sphere domain creator also has the option to use some hard coded time
+ * dependent maps that may be useful in certain scenarios. This method adds the
+ * maps in `domain::creators::sphere::TimeDependentMapOptions` to the domain.
+ * Currently, the first (inner-most) shell has maps between `Frame::Grid`,
```
wthrowe, line 153: Add a comment somewhere in `TimeDependentMapOptions` reminding people to update the documentation here when they change that class.

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -474,13 +523,24 @@ void test_parse_errors() {
...
           "None boundary condition is not supported. If you would like "
           "an outflow-type boundary condition, you must use that."));
+  CHECK_THROWS_WITH(
+      creators::Sphere(inner_radius, outer_radius, inner_cube, refinement,
+                       initial_extents, use_equiangular_map,
+                       equatorial_compression, radial_partitioning,
+                       radial_distribution, which_wedges,
+                       domain::creators::sphere::TimeDependentMapOptions{
+                           1.0, {0.0, 0.0, 0.0}, 5},
+                       nullptr, Options::Context{false, {}, 1, 1}),
```
wthrowe: Omit the `Context` argument.

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -537,6 +600,13 @@ void test_sphere() {
     CAPTURE(which_wedges);
     CAPTURE(time_dependent);
     CAPTURE(with_boundary_conditions);
+    // If we aren't time dependent, just set the hard coded option to false to
+    // avoid ambiguity. If we are time dependent but we're filling the interior,
+    // then we can't use hard coded options
+    if ((not time_dependent) or (time_dependent and fill_interior)) {
```
wthrowe: [optional] The second `time_dependent` is redundant.  If it were false we would never get there.

wthrowe (changes requested):

src/Domain/Creators/SphereTimeDependentMaps.hpp
knelli2: I was under the assumption that since I (possibly) have a distorted -> inertial map which is just the identity, that I also need to include it in the overall grid -> inertial map so that they match. Is this not the case?

nilsvu: you do not!

knelli2:
Squashed in

nilsvu (dismissed):

knelli2:
We should hold off on merging this PR until #4920 is merged because that's what I'll be using.

wthrowe (dismissed):
Marking dependent on #4920.

knelli2:
Rebased now that #4920 and #4931 are merged.

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -474,17 +525,28 @@ void test_parse_errors() {
...
+                           1.0, {0.0, 0.0, 0.0}, 5},
+                       nullptr),
+      Catch::Matchers::Contains(
+          "Currently cannot use hard-coded time dependent maps with an inner "
+          "cube. Use a TimeDependence instead."));
 }
 
 void test_sphere() {
-  MAKE_GENERATOR(gen);
+  MAKE_GENERATOR(gen, 3206883144);
```
nilsvu: delete seed

nilsvu (changes requested):

tests/Unit/Domain/Creators/Test_Sphere.cpp
knelli2: Oops!

wthrowe (approved):

nilsvu (approved):

PR #4913, opened 2023-04-06
Add ability to specify colors/styles in `status` command

knelli2:
Currently this can only change the style of 'RUNNING', 'FAILED', etc. but can be extended in the future

## Proposed changes

Since everybody uses different colors on their terminal, the defaults may not show up properly. This allows the user to specify in the config file color/styles that will override the defaults so they can choose whatever colors they want (not all options have to be specified). The config file should look like

```yaml
status:
  state_styles:
    RUNNING: [bright_green]
    FAILED: [grey]
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/Status/Status.py
+++ b/tools/Status/Status.py
@@ -154,7 +154,7 @@ def _state_order(state):
         return None
 
 
-def _format(field: str, value: Any) -> str:
+def _format(field: str, value: Any, state_styles: dict) -> str:
     if field == "State":
         style = {
             "RUNNING": "[blue]",
@@ -163,6 +163,7 @@ def _format(field: str, value: Any) -> str:
             "FAILED": "[red]",
             "TIMEOUT": "[red]",
         }
+        style.update(state_styles)
         return style.get(value, "") + str(value)
     elif field in ["Start", "End"]:
         if pd.isnull(value):
@@ -174,7 +175,7 @@ def _format(field: str, value: Any) -> str:
 
 
 @rich.console.group()
-def render_status(show_paths, show_unidentified, **kwargs):
+def render_status(show_paths, show_unidentified, state_styles, **kwargs):
     job_data = fetch_job_data([
         "JobID",
         "User",
@@ -252,7 +253,8 @@ def render_status(show_paths, show_unidentified, **kwargs):
 
                 # Extract job status and format row for output
                 row_formatted = [
-                    _format(field, row[field]) for field in standard_fields
+                    _format(field, row[field], state_styles)
+                    for field in standard_fields
                 ]
                 with open(row["InputFile"], "r") as open_input_file:
                     try:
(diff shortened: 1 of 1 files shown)


nilsvu:
@knelli2 I added a general config file in #4916, so all you need to do here is add an option like `--state-styles`. The configuration from the file will be passed to it automatically. You can note in the help string that the option should be specified in the config file, then you don't need to worry about parsing a dictionary from the command line.

knelli2:
Should the default to this new option just be `None`?

nilsvu:
Empty dict I think. Edit: Actually it's up to you how to handle that case. Empty dict probably works well, but you can also handle `None`.

tools/Status/Status.py
```diff
@@ -340,6 +349,15 @@ def render_status(show_paths, show_unidentified, **kwargs):
               default=None,
               help=("On a new screen, refresh jobs every 'watch' number "
                     "of seconds. Exit out with Ctl+C."))
+@click.option("--state-styles",
+              "input_state_styles",
```
knelli2: @nilsvu When trying this on wheeler, I had to specify `input_state_styles` in the config file rather than just `state_styles`. Do you know why that is?

It's not a big deal, I can just swap around the names, but was curious why it chose the variable name and not the option name.

tools/Status/Status.py
```diff
@@ -154,16 +154,9 @@ def _state_order(state):
...
     if field == "State":
-        style = {
-            "RUNNING": "[blue]",
-            "COMPLETED": "[green]",
-            "PENDING": "[magenta]",
-            "FAILED": "[red]",
-            "TIMEOUT": "[red]",
-        }
-        return style.get(value, "") + str(value)
+        return state_styles.get(value, "") + str(value)
```
nilsvu: do this:

```py
style = {
  # ... (same as before)
}
style.update(state_styles)
```

Then you don't have to change any other code, only pass the `state_styles` from the options through.

tools/Status/Status.py
nilsvu: Click uses the variable name as key in the `default_map` that it passes to the command. Just delete the `input_state_styles` name and pass it through as `state_styles` (see comment above).

tools/Status/Status.py
```diff
@@ -340,6 +349,15 @@ def render_status(show_paths, show_unidentified, **kwargs):
               default=None,
               help=("On a new screen, refresh jobs every 'watch' number "
                     "of seconds. Exit out with Ctl+C."))
+@click.option("--state-styles",
+              "input_state_styles",
+              type=dict,
+              default=None,
```
nilsvu: If you make this an empty dict then it's easier to handle (`dict.update` does nothing).

nilsvu (changes requested):

knelli2:
Ok yup those changes made everything a lot simpler

tools/Status/Status.py
```diff
@@ -340,6 +343,14 @@ def render_status(show_paths, show_unidentified, **kwargs):
               default=None,
               help=("On a new screen, refresh jobs every 'watch' number "
                     "of seconds. Exit out with Ctl+C."))
+@click.option("--state-styles",
+              type=dict,
+              default={},
+              help=("Dictionary between sacct states and rich modifiers for "
+                    "how the state will be printed. Rather than always having "
+                    "to specify a dict on the command line, you can add this "
+                    "to the spectre config file. See `spectre -h` for its "
```
nilsvu: Could you give an example here?

nilsvu (approved):

PR #4916, opened 2023-04-06
Load config file in CLI

nilsvu:
## Proposed changes

The config file is expected in `~/.config/spectre.yaml` but can also be set with the `SPECTRE_CONFIG_FILE` environment variable. Suggestions welcome for the default place of the config file.

Options listed in the config file are used as defaults in the CLI subcommands. For example, this is a valid config file:

```yaml
status:
  starttime: now-2days
  state_styles:
    RUNNING: blink  # if this really is what you desire
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/support/Python/Test_Main.py
+++ b/tests/support/Python/Test_Main.py
@@ -1,0 +1,49 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import os
+import shutil
+import unittest
+
+import yaml
+from click.testing import CliRunner
+from spectre.__main__ import cli
+from spectre.Informer import unit_test_build_path, unit_test_src_path
+from spectre.tools.CleanOutput import MissingExpectedOutputError
+
+
+class TestMain(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = os.path.join(unit_test_build_path(),
+                                     "support/Python/Main")
+        os.makedirs(self.test_dir, exist_ok=True)
+
+    def tearDown(self):
+        shutil.rmtree(self.test_dir)
+
+    def test_config_file(self):
+        # Collect some test data
+        input_file = os.path.join(
+            unit_test_src_path(),
+            "../InputFiles/Poisson/ProductOfSinusoids1D.yaml")
+        # Create a config file
+        config_file_path = os.path.join(self.test_dir, "config.yaml")
+        with open(config_file_path, "w") as open_config_file:
+            yaml.safe_dump({"clean-output": {"force": True}}, open_config_file)
+        # Run. The 'clean-output' command should fail because there's no data to
+        # clean up in the test dir, but with the 'force' option from the config
+        # file it should pass.
+        runner = CliRunner()
+        with self.assertRaises(MissingExpectedOutputError):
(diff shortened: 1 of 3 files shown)


tests/support/Python/Test_Main.py
```diff
@@ -0,0 +1,44 @@
...
+            yaml.safe_dump({"clean-output": {"force": True}}, open_config_file)
+        # Run. The 'clean-output' command should fail because there's no data to
+        # clean up in the test dir, but with the 'force' option from the config
+        # file it should pass.
+        runner = CliRunner()
+        result = runner.invoke(
+            cli, ["clean-output", "-o", self.test_dir, input_file],
+            catch_exceptions=False,
+            env={"SPECTRE_CONFIG_FILE": config_file_path})
+        self.assertEqual(result.exit_code, 0, result.output)
```
knelli2, line 45: Can you also run this test before you write the config file and check that it fails? Just in case for some reason there are files in the directory (no clue how they'd get there, but in theory possible).

knelli2 (changes requested):

tests/support/Python/Test_Main.py
nilsvu: yep, done

knelli2 (changes requested):
Looks good, you can squash

knelli2 (dismissed):

nilsdeppe (dismissed):
I think the location is good. The impression I've gotten is that Linux is trying to move away from dot files so that the home dir isn't so cluttered (possibly also makes moving to a new computer easier? :shrug:)

My big question, and it doesn't need to be solved for this PR: How do I figure out all the things I can configure in the config file? There's something on this in the `help` string but I have to admit it's not enough for me to immediately know how to piece together a config file to my liking from nothing

knelli2:
No yeah currently there's no way to know what goes in the config file unless you just know. We'll have to either mention this in a docs/tutorial of some kind and/or add a short description to `spectre -h`. Maybe once we have more important things in the config file besides just terminal colors we can add this

nilsvu:
Changed the help string to this:

```
  -c, --config-file FILE  Configuration file in YAML format. Can provide
                          defaults for command-line options and additional
                          configuration. To specify options for subcommands,
                          list them in a section with the same name as the
                          subcommand. All options that are listed in the help
                          string for a subcommand are supported. Unless
                          otherwise specified in the help string, use the name
                          of the option with dashes replaced by underscores.
                          Example:
                          
                          status:
                            starttime: now-2days
                            state_styles:
                              RUNNING: blink
                          plot-dat:
                            stylesheet: path/to/stylesheet.mplstyle
                          
                          The path of the config file can also be specified by
                          setting the 'SPECTRE_CONFIG_FILE' environment
                          variable.  [default: ~/.config/spectre.yaml]
```

nilsvu:
So yes there is a way to figure out what to write in the config file by looking at the existing help strings.

knelli2 (approved):

nilsdeppe (approved):
Awesome, thank you!

PR #4925, opened 2023-04-07
Add worldtube observation action

nikwit:
## Proposed changes
Adds an action that observes the coefficients of the  Taylor series of the regular field held by the worldtube singleton.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
@@ -1,0 +1,141 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <string>
+#include <tuple>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/Prefixes.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "IO/Observer/Helpers.hpp"
+#include "IO/Observer/ObservationId.hpp"
+#include "IO/Observer/ObserverComponent.hpp"
+#include "IO/Observer/ReductionActions.hpp"
+#include "IO/Observer/TypeOfObservation.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Reduction.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace CurvedScalarWave::Worldtube::Actions {
+
+/*!
+ * \brief When Tags::ObserveCoefficientsTrigger is triggered, write the
+ * coefficients of the Taylor expansion of the regular field to file.
+ */
+struct ObserveWorldtubeSolution {
+  using reduction_data = Parallel::ReductionData<
+      Parallel::ReductionDatum<double, funcl::AssertEqual<>>,
+      Parallel::ReductionDatum<std::vector<double>, funcl::AssertEqual<>>>;
+
+  using observed_reduction_data_tags =
+      observers::make_reduction_data_tags<tmpl::list<reduction_data>>;
+  static constexpr size_t Dim = 3;
+  template <typename DbTagsList, typename... InboxTags, typename Metavariables,
+            typename ArrayIndex, typename ActionList,
+            typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTagsList>& box,
+      const tuples::TaggedTuple<InboxTags...>& /*inboxes*/,
+      Parallel::GlobalCache<Metavariables>& cache,
+      const ArrayIndex& /*array_index*/, const ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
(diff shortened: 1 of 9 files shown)


tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
```diff
@@ -97,9 +97,9 @@ EventsAndTriggers:
           Offset: 0
     - - PsiAlongAxis1
       - PsiAlongAxis2
-  - - Times:
-        Specified:
-          Values: &FinalTime [2000.]
+  - - TimeCompares:
+        Comparison: GreaterThan
+        Value: 2000.
```
wthrowe, line 102: I am concerned if this didn't work, and I am even more concerned if the new version does.  (For the latter, because you removed the definition of `FinalTime`.)

What was the reason for this change?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -155,6 +169,20 @@ struct ExcisionSphere : db::SimpleTag {
   }
 };
 
+/*!
+ * \brief Triggers at which to write the coefficients of the worldtube's
+ * internal Taylor series to file.
+ */
+struct ObserveCoefficientsTrigger : db::SimpleTag {
```
wthrowe, line 176: Test the tag.

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
```diff
@@ -109,24 +109,15 @@ EventsAndTriggers:
           StepChoosers:
             - Cfl:
                 SafetyFactor: 0.35
-  - - TimeCompares:
```
wthrowe, line 112: This removal seems unrelated to the rest of the commit.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
```diff
@@ -0,0 +1,140 @@
...
+      if (expansion_order > 1) {
+        const auto& psi_quadrupole = db::get<
+            Stf::Tags::StfTensor<Tags::PsiWorldtube, 2, Dim, Frame::Grid>>(box);
+        const auto& dt_psi_quadrupole =
+            db::get<Stf::Tags::StfTensor<::Tags::dt<Tags::PsiWorldtube>, 2, Dim,
+                                         Frame::Grid>>(box);
+        const double wt_radius =
+            db::get<Tags::ExcisionSphere<Dim>>(box).radius();
+        const double trace_psi_2 =
+            3. * (get(psi_monopole) - get(psi_0).at(0)) / wt_radius / wt_radius;
```
wthrowe: `/ square(wt_radius)` and again two lines later.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
```diff
@@ -0,0 +1,140 @@
...
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTagsList>& box,
+      const tuples::TaggedTuple<InboxTags...>& /*inboxes*/,
+      Parallel::GlobalCache<Metavariables>& cache,
+      const ArrayIndex& /*array_index*/, const ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    if (db::get<Tags::ObserveCoefficientsTrigger>(box).is_triggered(box)) {
+      const size_t expansion_order = db::get<Tags::ExpansionOrder>(box);
+      const auto& psi_monopole = db::get<
+          Stf::Tags::StfTensor<Tags::PsiWorldtube, 0, Dim, Frame::Grid>>(box);
```
wthrowe, line 53: (arbitrary line) I don't know this system well, but there's a lot of math here that is not obviously specific to this observation.  Are there other places where it might be useful, so that breaking it out into a separate function for reuse would make sense?  Or is this all just organizing into the structures required by the observer code, in which case no reason to factor it out.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ObserveWorldtubeSolution.cpp
```diff
@@ -0,0 +1,242 @@
...
+      : tt::ConformsTo<Options::protocols::FactoryCreation> {
+    using factory_classes =
+        tmpl::map<tmpl::pair<Trigger, tmpl::list<Triggers::TimeCompares>>>;
+  };
+};
+
+template <typename Generator>
+void check_observe_worldtube_solution(
+    gsl::not_null<Generator*> generator,
+    gsl::not_null<std::uniform_real_distribution<>*> dist,
```
wthrowe: `const` these two.

wthrowe (changes requested):

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
nikwit: I put the removal of the `StepToTimes` in the wrong commit  but fixed it now. 

This always worked, the reason is just to simplify things. The system reaches a steady-state solution so we do not care about a point landing, it just needs to roughly complete at some time.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
nikwit: 95 % of it is just copying it over into the observer structure so there is not too much point in putting it into free functions.
I am planning on writing more extensive documentation that references the equations and sections of the paper once we publish (hopefully next week)

wthrowe:
OK.  That addresses everything.  Go ahead and squash.

wthrowe (approved):

PR #4931, opened 2023-04-11
Allow sphere transition inverse to handle epsilon

knelli2:
## Proposed changes

The inverse of the shape map uses the `original_radius_over_radius` function of the transition function. Previously the sphere transition didn't account for epsilon differences so mapping a point on a boundary of the shape map to/from the grid frame would result in a point just outside of the bounds of the shape map, causing the inverse to be `nullopt` even though it was perfectly fine. Now, the sphere transition checks for epsilon differences in `original_radius_over_radius`

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu (dismissed):

nilsvu:
Needs a rebase

knelli2:
Done!

nilsvu (approved):

PR #4934, opened 2023-04-11
Add OpeningAngle to BinaryCompactObject

mar-celine:
## Proposed changes

Addresses issue #3545 by introducing an OpeningAngle option to the BinaryCompactObject Domain that allows the user to set the angular sizes of the Wedges in the Domain, making the angular distribution of points more uniform.

### Upgrade instructions

Add the new `OuterShell.OpeningAngle` option to `BinaryCompactObject` domains in input files. Set it to `90` to keep the previous behavior.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

The old domain. One can also see how the grid becomes rounded-cubical moving outwards from the two inner cubes. This is left over from bulging out from a cubical shell of frustums (the old enveloping cube), The main concern however, was the fact that the two half wedges have the same angular size as a full wedge, despite having twice the number of gridpoints between them. The proposed solution was to fan out the half wedges and make the end cap full wedges smaller in angular size.

old_rect_domain

The new domain created from the same options, except with OpeningAngle set to 120.0. The enveloping cube is stretched into a rectangular prism first, which is then bulged out, which seems to give a better gridpoint distribution than before. The Wedges in the outer layer are also opened up to conform to the new bulged out prism, and the equiangular map is revamped to accomodate variable opening angle sizes.
new_rect_domain


nilsvu:
Looks great! Thanks for your work on this! I'll review asap.

tests/Unit/Domain/CoordinateMaps/Test_Wedge3D.cpp
```diff
@@ -240,38 +250,42 @@ void test_wedge3d_random_radii() {
...
-          1.0 /
-          sqrt(with_equiangular_map
-                   ? 2.0 + square(tan(with_equiangular_map ? half_opening_angle
-                                                           : M_PI_4))
-                   : 3.0);
+          1.0 / sqrt(1.0 + square(cap_xi_one) + square(cap_eta_one));
+      /*1.0 / sqrt(with_equiangular_map
+                     ? 1.0 + square(tan(0.5 * opening_angle_xi)) +
+                           square(tan(0.5 * opening_angle_eta))
+                     : 3.0);*/
```
knelli2: Remove?

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -102,19 +103,19 @@ Frustum::Frustum(const std::array<std::array<double, 2>, 4>& face_vertices,
...
-  // distance from the origin to the vertex along the other two axes are the
-  // same as the first, assuming it is a vertex of the origin-centered cube.
-  radius_ = std::max({abs(upper_x_upper_base), abs(upper_y_upper_base),
-                      abs(upper_x_lower_base), abs(upper_y_lower_base),
-                      abs(lower_x_upper_base), abs(lower_y_upper_base),
-                      abs(lower_x_lower_base), abs(lower_y_lower_base),
-                      abs(upper_bound), abs(lower_bound)}) *
-            sqrt(3.0);
+  // the Frustum that is furthest away from the origin. For the rectangular
+  // BinaryCompactObject Domain, this vertex is assumed to lie on a rectangular
```
knelli2, line 110: Is this true for domains other than the BCO domain?

src/Domain/CoordinateMaps/Frustum.hpp
```diff
@@ -324,6 +324,9 @@ class Frustum {
   double sphericity_{std::numeric_limits<double>::signaling_NaN()};
   double radius_{std::numeric_limits<double>::signaling_NaN()};
   double phi_{std::numeric_limits<double>::signaling_NaN()};
+  double half_opening_angle_{std::numeric_limits<double>::signaling_NaN()};
+  double one_over_tan_half_opening_angle_{
+      std::numeric_limits<double>::signaling_NaN()};
 };
 
 bool operator!=(const Frustum& lhs, const Frustum& rhs);
```
knelli2, line 357: [Random line] Should there be a test for the changes to the Frustum?

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -316,6 +316,14 @@ class BinaryCompactObject : public DomainCreator<3> {
     static constexpr Options::String help = {"Radius of the entire domain."};
   };
 
+  struct OpeningAngle {
+    using group = OuterShell;
+    static std::string name() { return "OpeningAngle"; }
+    using type = double;
+    static constexpr Options::String help = {
+        "The opening angle of the two half wedges in degrees."};
```
knelli2: Maybe be a bit more descriptive? I don't think a user necessarily knows what the half wedges are. Maybe something like

```
Surrounding the two objects (which are encompassed by two cubes) are a series of wedges that abut the
 faces of these cubes. The wedges on the ends of these cubes are full wedges while the wedges around 
the rest of the faces are half wedges. This option is the opening angle of these half wedges in degrees.
```

If I said something incorrect here you can just fix it when you add it.

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -428,20 +434,20 @@ Domain<3> BinaryCompactObject::create_domain() const {
   // a sphere enveloping both objects. While the two objects can be offset from
   // the origin to account for their center of mass, the enveloping frustums are
   // centered at the origin.
+  const double sphericity = 1.0;
```
knelli2: `constexpr`

tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
```diff
@@ -105,17 +105,18 @@ void test_connectivity() {
 
   // Outer shell:
   constexpr double outer_radius = 32.4;
+  constexpr double opening_angle = 120.0;
```
knelli2: Can you instead make this part of the random sample so we can test different values? Maybe `make_array(75.0, 90.0, 120.0)`?

tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
```diff
@@ -451,6 +454,7 @@ void test_bns_domain_with_cubes() {
 
   // Outer shell:
   constexpr double outer_radius = 32.4;
+  constexpr double opening_angle = 90.0;
```
knelli2, line 456: Same comment as above. You'll probably need to edit the `create_option_string` as well to take arbitrary values.

knelli2 (changes requested):
Just some minor changes. Love the new image though!

src/Domain/CoordinateMaps/Frustum.cpp
mar-celine: This is a good question - there is only one other Domain that uses Frustums, the FrustalCloak Domain, but this Domain has not been updated or used in a while. I expect the BinaryCompactObject Domain to be the only domain the typical user uses that would lead them to encounter a Frustum map, which is why I added this comment.

I could add, `For information on how other Domains use the Frustum map, please see that Domain's documentation.` or something to that effect.

src/Domain/CoordinateMaps/Frustum.hpp
mar-celine: I will add a random `opening_angle` to the test suite for map to check that the map is valid for various opening angles with the correct inverse/jacobian. Additional tests for "does this map do what we want" will be taken care of by the BCO domain's "check connectivity" test that checks that the Frustums with opening angles conform to the Wedges with opening angles.

src/Domain/Creators/BinaryCompactObject.hpp
mar-celine: I don't want to make the help string too long and since this option is placed in `group = OuterShell` and the outer shell is in the main BCO documentation I can change `half wedges` to `half wedges of the outer shell` and the user should have enough information to find out what this is referring to with this string. But I will also add documentation on the half wedges in that place!

knelli2: Ok yeah that sounds good.

src/Domain/CoordinateMaps/Frustum.cpp
knelli2: Yeah I think that comment would be helpful.

knelli2 (commented):
Looks good. You can squash

mar-celine:
I've squashed! @nilsvu It's ready for you to review :relaxed:

src/Domain/CoordinateMaps/Wedge.cpp
```diff
@@ -464,8 +524,9 @@ void Wedge<Dim>::pup(PUP::er& p) {
     p | scaled_frustum_zero_;
     p | sphere_zero_;
     p | scaled_frustum_rate_;
-    p | sphere_rate_;
-    p | half_opening_angle_;
+  } if (version >= 1){
```
nilsvu: newline (maybe just run `git-clang-format`). Also you deleted `p | sphere_rate` (by accident?). I hope this made a test fail, else please add a test that would have failed.

You have to take into account here that data written by version 0 will have serialized `half_opening_angle`. The current code won't read that in correctly (it will crash). This should work (haven't tested it):

```cpp
if (version >= 0) {
  // ...
  p | sphere_rate_;
  if (version == 0) {
    double half_opening_angle;
    p | half_opening_angle;
    opening_angles_[0] = 2. * half_opening_angle;
    opening_angles_[1] = 2. * half_opening_angle;
    opening_angles_distribution_ = opening_angles_;
  }
}
if (version >= 1) {
  p | opening_angles_;
  p | opening_angles_distribution_;
}
```

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -506,6 +511,9 @@ void Frustum::pup(PUP::er& p) {
     p | sphericity_;
     p | radius_;
     p | phi_;
+  } if (version >= 1) {
```
nilsvu: clang-format

src/Domain/CoordinateMaps/Frustum.hpp
```diff
@@ -266,7 +266,7 @@ class Frustum {
           bool with_equiangular_map = false,
           double projective_scale_factor = 1.0,
           bool auto_projective_scale_factor = false, double sphericity = 0.0,
-          double transition_phi = 0.0);
+          double transition_phi = 0.0, double opening_angle = M_PI_2);
```
nilsvu, line 294: Add some docs please

src/Domain/DomainHelpers.hpp
```diff
@@ -162,6 +162,9 @@ size_t which_wedge_index(const ShellWedges& which_wedges);
  * \param radial_distribution Select the radial distribution of grid points in
  * the spherical shells.
  * \param which_wedges Select a subset of wedges.
+ * \param opening_angle sets the opening angle of the two half wedges that
```
nilsvu: Add a sentence that this only has an effect if `use_half_wedges` is turned on. Also maybe name something like `half_wedges_opening_angle` to clarify that this number applies to the half wedges? Default to `M_PI_4` in that case (and multiply by 2 to get the opening angle of the full wedge). This is an attempt to clarify what "opening angle" means for a full envelope of wedges, other suggestions welcome.

src/Domain/DomainHelpers.hpp
```diff
@@ -187,7 +190,8 @@ std::vector<domain::CoordinateMaps::Frustum> frustum_coordinate_maps(
     double length_inner_cube, double length_outer_cube,
     bool use_equiangular_map,
     const std::array<double, 3>& origin_preimage = {{0.0, 0.0, 0.0}},
-    double projective_scale_factor = 1.0, double sphericity = 0.0);
+    double projective_scale_factor = 1.0, double sphericity = 0.0,
+    double opening_angle = M_PI_2);
```
nilsvu, line 200: same comment as above (and please add to docs)

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -316,6 +318,15 @@ class BinaryCompactObject : public DomainCreator<3> {
     static constexpr Options::String help = {"Radius of the entire domain."};
   };
 
+  struct OpeningAngle {
```
nilsvu, line 321: Same comment here about clarifying what "opening angle" means for a full envelope of wedges. Maybe `HalfWedgesOpeningAngle` if you want to retain the ability to choose any value for the angles. Values of `45` and `60` degrees would correspond to the previous behavior and the new equal angular size behavior, respectively (note that in the help string). Alternatively, just a boolean like `EqualizeOpeningAngles` or something like that could also work. WDYT?

nilsvu (changes requested):

src/Domain/DomainHelpers.hpp
mar-celine: I want to keep `opening_angle` as the angle for the full wedge (or both halves combined) because this makes it easier to compare with the endcap full wedge which is also given an `opening_angle`, but I'll had that this parameter has no effect if `use_half_wedges` is turned off

src/Domain/Creators/BinaryCompactObject.hpp
mar-celine: (as mentioned above I think it's easier to be able to compare two half wedges with the end cap wedge using the full opening angle vs the half)

I'm not sure if we want to equalize by default yet, since we are sacrificing equiangular-ity in the endcap wedges (but so far runs have not shown poor behavior in those regions resulting from the change).

I have a guess that a value between 120.0 and 90.0 might be ideal to strike a balance between angular size and equiangular-ity, or it could also be the case that we don't want to equalize angular size in angle but rather in surface area / volume covered by each wedge/half wedge - it is not clear to me that a choice of 120.0 results in the half wedge having the same surface area or volume as an endcap wedge (and this still might not be what gives ideal behavior)

In any case, I think it's too soon to default yet :sweat_smile:

src/Domain/CoordinateMaps/Wedge.cpp
```diff
@@ -465,7 +525,23 @@ void Wedge<Dim>::pup(PUP::er& p) {
     p | sphere_zero_;
     p | scaled_frustum_rate_;
     p | sphere_rate_;
-    p | half_opening_angle_;
+    if (version == 0) {
+      double half_opening_angle;
+      p | half_opening_angle;
+      if constexpr (Dim == 2) {
+        opening_angles_[0] = 2. * half_opening_angle;
```
nilsvu: Move out of the if

src/Domain/CoordinateMaps/Wedge.cpp
```diff
@@ -465,7 +525,23 @@ void Wedge<Dim>::pup(PUP::er& p) {
...
+        opening_angles_distribution_[0] = opening_angles_[0];
```
nilsvu: `opening_angles_distribution_ = opening_angles_` outside the if should do it

src/Domain/CoordinateMaps/Frustum.cpp
nilsvu: Ping

src/Domain/CoordinateMaps/Wedge.cpp
```diff
@@ -465,7 +525,23 @@ void Wedge<Dim>::pup(PUP::er& p) {
     p | sphere_zero_;
     p | scaled_frustum_rate_;
     p | sphere_rate_;
-    p | half_opening_angle_;
+    if (version == 0) {
+      double half_opening_angle;
```
nilsvu: Clang-tidy wants you to initialize this. Either set to NaN or NOLINT this line.

src/Domain/DomainHelpers.hpp
nilsvu: Ok

nilsvu (changes requested):
Couple tiny things, you can squash everything. Please add some upgrade instructions to the PR description.

nilsvu (approved):

nilsvu (commented):
I think the relevant part for the upgrade instructions is that people have to add an `OpeningAngle` option to their input files and should get an idea what that does, and set it to `90` to keep the current behavior. Please update when you get the chance.

nilsvu:
I updated the upgrade instructions.

PR #4935, opened 2023-04-11
Add ability to watch status command

knelli2:
## Proposed changes

This will keep a screen up and refresh it every input number of seconds, similar to the `watch` command on linux

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/Status/Status.py
+++ b/tools/Status/Status.py
@@ -16,6 +16,9 @@ import humanize
 import pandas as pd
 import rich.console
 import rich.table
+import rich.rule
+import rich.live
+import time
 import yaml
 from spectre.Visualization.ReadInputFile import get_executable
 
@@ -170,40 +173,8 @@ def _format(field: str, value: Any) -> str:
         return str(value)
 
 
-@click.command()
-@click.option("-u",
-              "--uid",
-              "--user",
-              "user",
-              show_default="you",
-              help=("User name or user ID. "
-                    "See documentation for 'sacct -u' for details."))
-@click.option("-a",
-              "--allusers",
-              is_flag=True,
-              help=("Show jobs for all users. "
-                    "See documentation for 'sacct -a' for details."))
-@click.option("-p",
-              "--show-paths",
-              is_flag=True,
-              help=("Show job working directory and input file paths."))
-@click.option("-U",
-              "--show-unidentified",
-              is_flag=True,
-              help=("Also show jobs that were not identified "
-                    "as SpECTRE executables."))
-@click.option("-s",
-              "--state",
-              help=("Show only jobs with this state, "
-                    "e.g. running (r) or completed (cd). "
-                    "See documentation for 'sacct -s' for details."))
-@click.option("-S",
-              "--starttime",
-              show_default="start of today",
-              help=("Show jobs eligible after this time, e.g. 'now-2days'. "
-                    "See documentation for 'sacct -S' for details."))
-def status_command(show_paths, show_unidentified, **kwargs):
-    """Gives an overview of simulations running on this machine."""
+@rich.console.group()
+def render_status(show_paths, show_unidentified, **kwargs):
     job_data = fetch_job_data([
         "JobID",
         "User",
@@ -240,9 +211,6 @@ def status_command(show_paths, show_unidentified, **kwargs):
     # Add metadata so jobs can be grouped by state
     job_data["StateOrder"] = job_data["State"].apply(_state_order)
 
-    # Start printing things
-    console = rich.console.Console()
-
     # We'll print these columns
     standard_fields = [
         "State",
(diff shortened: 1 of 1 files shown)


tools/Status/Status.py
```diff
@@ -215,7 +186,8 @@ def status_command(show_paths, show_unidentified, **kwargs):
         "State",
         "WorkDir",
         "Comment",
-    ], **kwargs)
+    ], kwargs["Fields"], kwargs["User"], kwargs["AllUsers"], kwargs["State"],
```
nilsvu: This gives `KeyError`s for me. Also, I don't understand this change. The `**kwargs` are all remaining arguments to the function that aren't named. Just keep forwarding `**kwargs` here. (see also comment on the `watch` argument below)

tools/Status/ExecutableStatus/__init__.py
```diff
@@ -18,18 +18,25 @@
 ]
 
 
-def match_executable_status(executable_name: str) -> ExecutableStatus:
+def match_executable_status(
+        executable_name: str,
+        suppress_warningg: bool = False) -> ExecutableStatus:
```
nilsvu: I think you don't need this change. Instead, you can just disable logging entirely when the `--watch` option is passed to the `status_command`:

```py
# before the `rich.live.Live`
logging.disable()
```

tools/Status/Status.py
```diff
@@ -308,25 +279,87 @@ def status_command(show_paths, show_unidentified, **kwargs):
...
+              show_default="start of today",
+              help=("Show jobs eligible after this time, e.g. 'now-2days'. "
+                    "See documentation for 'sacct -S' for details."))
+@click.option("-w",
+              "--watch",
+              type=float,
+              default=None,
+              help=("On a new screen, refresh jobs every 'watch' number "
+                    "of seconds. Exit out with Ctl+C."))
+def status_command(show_paths, show_unidentified, **kwargs):
```
nilsvu: 1. Add the named argument `refresh_rate` here, so it doesn't get forwarded with `**kwargs`. Also add the name `refresh_rate` to the `click.option` above (so it's named `refresh_rate`, not `watch`):

```py
@click.option("-w",
              "--watch",
              "refresh_rate",
              type=float, # ...
```

2. Remove the named arguments `show_paths` and `show_unidentified` here so they just get forwarded with the `**kwargs` (you don't need them in this function).

tools/Status/Status.py
```diff
@@ -308,25 +279,87 @@ def status_command(show_paths, show_unidentified, **kwargs):
...
+
+    if not refresh_rate:
+        console.print(render_screen(show_paths, show_unidentified, **kwargs))
+        return
+
+    try:
+        with rich.live.Live(render_screen(show_paths, show_unidentified,
+                                          **kwargs),
+                            console=console,
+                            screen=True) as live:
```
nilsvu, line 358: Add `auto_refresh=False` here and add `refresh=True` in the `update` call below. Otherwise you get weird interaction between the default refresh rate (4fps) and the one you specified manually.

tools/Status/Status.py
```diff
@@ -170,40 +173,8 @@ def _format(field: str, value: Any) -> str:
...
-                    "See documentation for 'sacct -s' for details."))
-@click.option("-S",
-              "--starttime",
-              show_default="start of today",
-              help=("Show jobs eligible after this time, e.g. 'now-2days'. "
-                    "See documentation for 'sacct -S' for details."))
-def status_command(show_paths, show_unidentified, **kwargs):
-    """Gives an overview of simulations running on this machine."""
+@rich.console.group()
+def render_screen(show_paths, show_unidentified, **kwargs):
```
nilsvu: (optional) `render_status`?

nilsvu (changes requested):
Cool! Once I fixed the `kwargs` thing this worked well!

knelli2:
Ok squashed in. Tested it out on wheeler and seems to work well

tools/Status/Status.py
```diff
@@ -268,8 +236,9 @@ def status_command(show_paths, show_unidentified, **kwargs):
         if first_section:
             first_section = False
         else:
-            console.print("")
-        console.rule(f"[bold]{executable_name}", align="left")
+            yield ""
+        yield rich.rule.Rule(f"[bold]{executable_name}", align="left")
+        yield ""
```
nilsvu, line 241: delete extra line?

tools/Status/Status.py
```diff
@@ -308,25 +277,87 @@ def status_command(show_paths, show_unidentified, **kwargs):
...
+            yield table
 
     # Output jobs that couldn't be parsed
     unidentified_jobs = job_data[job_data["ExecutableName"].isnull()]
     if len(unidentified_jobs) > 0 and show_unidentified:
-        console.print("")
-        console.rule("[bold]Unidentified Jobs", align="left")
+        yield ""
+        yield rich.rule.Rule("[bold]Unidentified Jobs", align="left")
+        yield ""
```
nilsvu, line 294: also delete this extra line?

nilsvu (changes requested):

tools/Status/Status.py
knelli2: I specifically added in this extra line (and the one below) so there was a space between the executable name and the list of jobs:

image

I thought it looked nice, but don't have a preference either way

nilsvu: ok no strong preference either, so we can go with this

nilsvu (approved):

PR #4936, opened 2023-04-12
Fix status CLI for jobs running in same work dir

nilsvu:
## Proposed changes

This happens when a job is restarted and overwrites output in the work dir. The output belongs to the new job now, and the old job's data is lost. We shouldn't associate the new data with the old job.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2 (approved):
Tried on Wheeler, and looks good! Jobs that were run in the same directory that have finished (completed or failed) now don't print the time, speed, etc. Only the current running job prints. And the `workdir` prints nicely for jobs that were in the same directory.

nilsdeppe (approved):

PR #4937, opened 2023-04-13
Clenshaw method

isaaclegred:
## Proposed changes

This PR adds the Clenshaw method for computing three term recurrences.  It implements this in the `Enthalpy` EoS for evaluating a series of trigonometric terms in evaluating thermodynamic quantities.  Nonetheless, this code will probably be useful elsewhere, so it is implemented more generically.  

### Upgrade instructions

No changes to the interface should be necessary.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

Benchmarking results indicate that avoiding the evaluation of trig functions leads to about an order of magnitude (or more) speed up per term.  This can be significant when the same expanded function has to be called many times, like with the EoS in primitive recovery.

### DIFF
--- a/src/NumericalAlgorithms/Spectral/CMakeLists.txt
+++ b/src/NumericalAlgorithms/Spectral/CMakeLists.txt
@@ -30,6 +30,7 @@ spectre_target_headers(
   ${LIBRARY}
   INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/src
   HEADERS
+  Clenshaw.hpp
   ComplexDataView.hpp
   Filtering.hpp
   LogicalCoordinates.hpp
--- a/src/NumericalAlgorithms/Spectral/Clenshaw.hpp
+++ b/src/NumericalAlgorithms/Spectral/Clenshaw.hpp
@@ -1,0 +1,47 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cmath>
+#include <numeric>
+#include <utility>
+#include <vector>
+
+#include "DataStructures/DataVector.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/MakeWithValue.hpp"
+
+namespace Spectral {
+/// Clenshaw's algorithm for evaluating linear combinations of
+/// basis functions obeying a given three term recurrence relation.
+/// See numerical recipes 3rd edition 5.4.2 for details.
+/// `alpha` and `beta` define the recursion
+/// \f$ f_n(x) = \alpha f_{n-1}(x) + \beta f_{n-2}(x) \f$
+/// `f0_of_x` and `f1_of_x` are \f$ f_0(x)\f$ and \f$f_1(x)\f$ respectively
+template <typename CoefficientDataType, typename DataType>
+DataType evaluate_clenshaw(const std::vector<CoefficientDataType>& coefficients,
+                           const DataType& alpha, const DataType& beta,
+                           const DataType& f0_of_x, const DataType& f1_of_x) {
+  if (coefficients.empty()) {
+    return make_with_value<DataType>(f0_of_x, 0.0);
+  }
+
+  DataType y_upper{make_with_value<DataType>(f0_of_x, 0.0)};  // y_{N+2} = 0
+  DataType new_y_lower;
+  // Do the loop to calculate y_1 and y_2 (the accumulate function leaves
+  // y_lower_final = y_1, and y_upper = y_2)
+  const DataType y_lower_final =
+      std::accumulate(coefficients.rbegin(), coefficients.rend() - 1,
+                      make_with_value<DataType>(f0_of_x, 0.0),
+                      [&y_upper, &alpha, &beta, &new_y_lower](
+                          const auto& y_lower, const auto& c_k) {
+                        new_y_lower = c_k + alpha * y_lower + beta * y_upper;
+                        y_upper = y_lower;
+                        return new_y_lower;
+                      });
+  return f0_of_x * coefficients[0] + f1_of_x * y_lower_final +
+         beta * f0_of_x * y_upper;
+}
+
+}  // namespace Spectral
--- a/src/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.cpp
+++ b/src/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.cpp
@@ -10,11 +10,29 @@
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "NumericalAlgorithms/Spectral/Clenshaw.hpp"
 #include "PointwiseFunctions/Hydro/EquationsOfState/EquationOfState.hpp"
 #include "PointwiseFunctions/Hydro/EquationsOfState/Factory.hpp"
 #include "PointwiseFunctions/Hydro/SpecificEnthalpy.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/MakeWithValue.hpp"
+
+namespace {
+
+double evaluate_cosine_series(const std::vector<double>& coefficients,
+                              const double& cosx) {
+  // cos(nx) = cos((n-1)*x)*2*cosx - 1* cos((n-2)*x)
+  return Spectral::evaluate_clenshaw(coefficients, 2.0 * cosx, -1.0, cosx,
+                                     2.0 * cosx * cosx - 1.0);
+}
+double evaluate_sine_series(const std::vector<double>& coefficients, double x,
+                            const double& cosx) {
+  // sin(nx) = sin((n-1)*x) * 2*cos(x) - 1 * sin((n-2)x)
+  auto sinx = sin(x);
+  return Spectral::evaluate_clenshaw(coefficients, 2.0 * cosx, -1.0, sinx,
+                                     2.0 * cosx * sinx);
+}
+}  // namespace
 
 namespace EquationsOfState {
 template <typename LowDensityEoS>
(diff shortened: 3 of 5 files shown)


src/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.cpp
teukolsky, line 1: Hi Isaac,
I don't have time to check carefully, but are you storing the coefficients by computing N trig functions? A proper implementation of Clenshaw should need only 1 trig function, cos(theta), for the recurrence, and then maybe some more when you compute (5.4.23) in Num. Rec. We can talk about this after APS.

isaaclegred: Hi Saul, There was only one trig function evaluation total(or 2 in the case of sine).  The slowdown was in the body of the loop (5.4.22) which by itself should not be a slow operation, but seems to be slowed down by assignments (perhaps cache masses).  I fear that this may be compiler/hardware dependent.   I'm happy to talk about it after APS.

wthrowe: I think @teukolsky has it right here.  It's been a long time since I looked at Clenshaw recurrence and I don't remember the details of how it should work, but the old implementation was definitely evaluating trig functions in a loop.

isaaclegred: Oh I don't know about an existing implementation, I wrote my own version that was specialized to the problem here.  The version I wrote computed all of the trig functions outside of the Clenshaw method itself, so I'm quite confident it wasn't calling trig functions in the loop.

wthrowe: OK.  I was looking at the code you deleted.  If that's not what you're comparing to then never mind.

isaaclegred: Oh sorry yeah the "old" (currently implemented in develop) approach was doing a trig call per iteration (i.e. per fourier term).  I tried to speed this up with just pure Clenshaw (out of the book), which is what is in the first commit.  For reasons I don't fully understand this was still slow, though this may be hardware dependent.  In the second commit I use the recurrence to pre-compute the Coefficients of the polynomial in `cos(theta)` given by the relation between `cos(n * theta)` and Chebyshev polynomials in `cos(theta)`, this is essentially guaranteed to be fast, but I do not know if it is stable (it seems to be for low n which is the primary use case).

isaaclegred: Hi @teukolsky, I think I managed to improve the performance to the level we would hope for by using a standard library algorithm (`std::accumulate`) I think the reason for the slowdown was not necessarily anything that was happening in the loop so much as doing the loop itself  (profiling showed that the vast majority of the time was spent on the final instruction in the loop, regardless of what the instruction actually was).   I think the standard library algorithm I think gives the compiler more strict assumptions that let it make better optimizations.

src/NumericalAlgorithms/Spectral/CMakeLists.txt
```diff
@@ -31,6 +31,7 @@ spectre_target_headers(
   INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/src
   HEADERS
   ComplexDataView.hpp
+  Clenshaw.hpp
```
wthrowe, line 33: Alphabetize.

src/NumericalAlgorithms/Spectral/Clenshaw.hpp
```diff
@@ -0,0 +1,44 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+#pragma once
```
wthrowe, line 4: Blank line before `#pragma once`.

tests/Unit/NumericalAlgorithms/Spectral/Test_Clenshaw.cpp
```diff
@@ -0,0 +1,133 @@
...
+  return std::accumulate(
+      coefficients.begin(), coefficients.end(),
+      make_with_value<DataType>(x, 0.0),
+      [&x, &f_of_k_and_x, &counter](auto& sum, const auto& coefficient) {
+        counter++;
+        return sum + coefficient * f_of_k_and_x(counter - 1, x);
+      });
+}
+// Check that for small arguments the clenshaw expansion and the direct cosine
+// evaluation agree exactly.
```
wthrowe: The don't agree _exactly_ because of floating point.  I'd just remove the word "exactly".  I don't think further explanation is needed, since anyone looking at an algorithm like this will understand that sort of thing.

tests/Unit/NumericalAlgorithms/Spectral/Test_Clenshaw.cpp
```diff
@@ -0,0 +1,133 @@
...
+  const DataType cos_0x = make_with_value<DataType>(x, 1.0);
+  const DataType cos_x = cos(x);
+  const DataType true_value = evaluate_with_function_basis(
+      coefficients, x, [](const size_t& k, const DataType& z) {
+        return cos(static_cast<double>(k) * z);
+      });
+  const DataType clenshaw_value =
+      Spectral::evaluate_clenshaw(coefficients, alpha, beta, cos_0x, cos_x);
+  CHECK_ITERABLE_APPROX(clenshaw_value, true_value);
+  // CHECK(cos_0x == 1.0);
```
wthrowe: Remove commented line.

tests/Unit/NumericalAlgorithms/Spectral/Test_Clenshaw.cpp
```diff
@@ -0,0 +1,133 @@
...
+  const DataType cos_0x = make_with_value<DataType>(x, 1.0);
+  const DataType cos_x = cos(x);
+  const DataType true_value = evaluate_with_function_basis(
+      coefficients, x, [](const size_t& k, const DataType& z) {
+        return cos(static_cast<double>(k) * z);
+      });
+  const DataType clenshaw_value =
+      Spectral::evaluate_clenshaw(coefficients, alpha, beta, cos_0x, cos_x);
+  CHECK_ITERABLE_CUSTOM_APPROX(clenshaw_value, true_value, custom_approx);
+  // CHECK(cos_0x == 1.0);
```
wthrowe: Remove commented line.

tests/Unit/NumericalAlgorithms/Spectral/Test_Clenshaw.cpp
```diff
@@ -0,0 +1,133 @@
...
+  const DataType sin_0x = make_with_value<DataType>(x, 0.0);
+  const DataType sin_x = sin(x);
+  const DataType true_value = evaluate_with_function_basis(
+      coefficients, x, [](const size_t& k, const DataType& z) {
+        return sin(static_cast<double>(k) * z);
+      });
+  const DataType clenshaw_value =
+      Spectral::evaluate_clenshaw(coefficients, alpha, beta, sin_0x, sin_x);
+  CHECK_ITERABLE_APPROX(clenshaw_value, true_value);
+  // CHECK(cos_0x == 1.0);
```
wthrowe: Again.

tests/Unit/NumericalAlgorithms/Spectral/Test_Clenshaw.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+}  // namespace
+SPECTRE_TEST_CASE("Unit.Numerical.Spectral.Clenshaw.CosineExact",
+                  "[NumericalAlgorithms][Spectral][Unit]") {
+  CHECK(1.0 == cos(0.0));
```
wthrowe: Seems unnecessary.  Again below.

tests/Unit/NumericalAlgorithms/Spectral/Test_Clenshaw.cpp
```diff
@@ -0,0 +1,133 @@
...
+  test_cosine_series({3200.0, 2100.0, 5065.0, 200000.0, 2211.0}, M_PI * 2.0);
+
+  test_cosine_series(
+      {1.0, 2.0, 3.6, 9.5, M_E},
+      DataVector{1.0, M_PI, 1.04, 9.0211, 5.1, -92.2, -M_PI / 2.0});
+}
+SPECTRE_TEST_CASE("Unit.Numerical.Spectral.Clenshaw.CosineApproximate",
+                  "[NumericalAlgorithms][Spectral][Unit]") {
+  Approx coarse_approx{1.0e-12};
+  CHECK(1.0 == coarse_approx(cos(M_PI * 100000000.0)));
```
wthrowe, line 105: But this one can stay.  It shouldn't vary significantly between machines, but it looks to be explaining the tolerance chosen.

tests/Unit/NumericalAlgorithms/Spectral/Test_Clenshaw.cpp
```diff
@@ -0,0 +1,133 @@
...
+  test_cosine_series({1.0, 2.0, 3.6}, M_PI);
+  test_cosine_series({1.0, 2.0, 3.6, -21.1, M_E}, 2.1214353);
+  test_cosine_series({0.0, -1.5, 0.0001, -1.1, M_E}, 32.1222);
+  test_cosine_series({3200.0, 2100.0, 5065.0, 200000.0, 2211.0}, M_PI * 2.0);
+
+  test_cosine_series(
+      {1.0, 2.0, 3.6, 9.5, M_E},
+      DataVector{1.0, M_PI, 1.04, 9.0211, 5.1, -92.2, -M_PI / 2.0});
+}
+SPECTRE_TEST_CASE("Unit.Numerical.Spectral.Clenshaw.CosineApproximate",
```
wthrowe: Combine all the test cases into one, either as separate functions that are called or all in the same function.  We've found that the overhead from starting individual test cases is non-negligible on some clusters, so the current preferred style is to minimize the number of them.

src/NumericalAlgorithms/Spectral/Clenshaw.hpp
```diff
@@ -0,0 +1,44 @@
...
+#include "DataStructures/DataVector.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/MakeWithValue.hpp"
+
+namespace Spectral {
+/// Clenshaw's algorithm for evaluating linear combinations of
+/// basis functions obeying a given three term recurrence relation.
+/// See numerical recipes 3rd edition 5.4.2 for details.
+/// `alpha` and `beta` define the recursion
+/// \f$ f_n(x) = alpha f_{n-1}(x) + beta f_{n-2}(x) \f$
```
wthrowe: `\alpha` and `\beta`

src/NumericalAlgorithms/Spectral/Clenshaw.hpp
```diff
@@ -0,0 +1,44 @@
...
+/// `f0_of_x` and `f1_of_x` are \f$ f_0(x)\f$ and \f$f_1(x)\f$ respectively
+template <typename CoefficientDataType, typename DataType>
+DataType evaluate_clenshaw(const std::vector<CoefficientDataType>& coefficients,
+                           DataType alpha, DataType beta, DataType f0_of_x,
+                           DataType f1_of_x) {
```
wthrowe: Const-reference for all of these unless I'm missing them being modified by the function.

src/NumericalAlgorithms/Spectral/Clenshaw.hpp
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  DataType y_upper{make_with_value<DataType>(f0_of_x, 0.0)};  // y_{N+2} = 0
+  // Do the loop to calculate y_1 and y_2 (the accumulate function leaves
+  // y_lower_final= y_1, and y_upper=y_2)
+  const DataType y_lower_final = std::accumulate(
+      coefficients.rbegin(), coefficients.rend() - 1,
+      make_with_value<DataType>(f0_of_x, 0.0),
+      [&y_upper, &alpha, &beta](const auto& y_lower, const auto& c_k) {
+        DataType new_y_lower = c_k + alpha * y_lower + beta * y_upper;
```
wthrowe: I'm not going to try to minimize allocations unless we really care about performance on DataVectors, but moving the declaration of `new_y_lower` out of the loop so it isn't reallocated every time is an easy win.  (Keep using it as a temporary here, just declare it above and capture it in the lambda.)

wthrowe (changes requested):
Try to keep the first line of your commit message short, 72 characters max.

isaaclegred:
@wthrowe looks like I'll probably have to rebase to get the python formatting to work.  I'll do this along with reformatting the commit message when I squash.

src/NumericalAlgorithms/Spectral/Clenshaw.hpp
wthrowe: Sorry, I meant in `\f$ f_n(x) = alpha f_{n-1}(x) + beta f_{n-2}(x) \f$`.  The ones on the line before can be whichever you prefer.

tests/Unit/NumericalAlgorithms/Spectral/Test_Clenshaw.cpp
```diff
@@ -87,13 +85,11 @@ void test_sine_series(const std::vector<double>& coefficients,
   const DataType clenshaw_value =
       Spectral::evaluate_clenshaw(coefficients, alpha, beta, sin_0x, sin_x);
   CHECK_ITERABLE_APPROX(clenshaw_value, true_value);
-  // CHECK(cos_0x == 1.0);
 }
 
 }  // namespace
-SPECTRE_TEST_CASE("Unit.Numerical.Spectral.Clenshaw.CosineExact",
+SPECTRE_TEST_CASE("Unit.Numerical.Spectral.Clenshaw.Generic",
```
wthrowe: [optional] Unless it is disambiguating from another test or conveys some meaning, you can leave off the `.Generic`.

wthrowe (changes requested):
Squash and rebase, including the doc fix and the optional suggestion if you decide to make it.  (And yes, the python thing will be fixed by a rebase.)

wthrowe (approved):

wthrowe:
clang-tidy failure is the usual boost thing.

PR #4938, opened 2023-04-14
Add versioning to coord maps and functions of time, add docs

nilsvu:
## Proposed changes

The domain will be serialized and written to output files so it can be used for interpolations, as are `domain::FunctionOfTime` classes. To be able to read in domains written by older versions of the code the `pup` functions now support lightweight versioning (some already did). A `version` integer is written alongside the data when serializing, and read back in when deserializing.

### Upgrade instructions

This change breaks deserialization of domains written into volume data files by older versions of the code (and is intended to prevent such breaking changes in the future). For tools that need the deserialized domain (e.g `interpolate-to-coords` or `plot-power-monitors`) use an older version of the code, or regenerate your volume data files with the upgraded code.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
Clang-tidy failure is the unrelated TOMS748 issue

src/Domain/Domain.hpp
```diff
@@ -40,9 +40,61 @@ class CoordinateMapBase;
...
+ *   } else {
+ *     p | some_data_;
+ *   }
+ *   // ...
+ * }
+ * ```
+ *
+ * Make sure that all data types you serialize in the `pup` function also
+ * support versioning. Also make sure not to keep all factory-creatable classes
+ * registered that were written by old versions of the code.
```
wthrowe, line 97: Should this sentence be the opposite?

wthrowe (changes requested):
I'm surprised that nothing gives a warning for checking that an unsigned integer is non-negative, but since nothing does I'm fine with the checks.

src/Domain/Domain.hpp
nilsvu: Yep! Good catch.

wthrowe (approved):

PR #4940, opened 2023-04-20
Promote input file metadata to YAML

### LINKED ISSUE #3731: Parse input file headers as YAML
# Feature request:

Many of our test input files have headers like this:

```yaml

# Executable: EvolveScalarWave3D
# Check: parse;execute_check_output
# Timeout: 10
# ExpectedOutput:
#   ScalarWaveReductions.h5
# OutputFileChecks:
#   - Label: "convergence"
#     Subfile: "/ErrorNorms.dat"
#     FileGlob: "ScalarWaveReductions.h5"
#     AbsoluteTolerance: 1e-3

DomainCreator:
  # ...
```

I propose to upgrade these headers to proper YAML syntax. YAML files can contain multiple documents separated by a `---` line, so I propose we allow input files to contain 2 documents, of which the first is discarded by the option parser if it is present:

```yaml

Executable: EvolveScalarWave3D
Check: [parse, execute_check_output]
Timeout: 10
ExpectedOutput:
  - ScalarWaveReductions.h5
OutputFileChecks:
  - Label: "convergence"
    Subfile: "/ErrorNorms.dat"
    FileGlob: "ScalarWaveReductions.h5"
    AbsoluteTolerance: 1e-3

---

DomainCreator:
  # ...
```

We can then load the header as proper YAML to configure input file tests, and we can also use the header for metadata like this:

```yaml
Executable: EvolveScalarWave3D
Version: 2022.01.03
```

Input files published with papers can include such metadata to help reproducibility.

nilsvu:
## Proposed changes

Upgrades the YAML-like comments at the start of our input files to actual YAML:

Old:

```yaml
# Executable: EvolveScalarWave1D
# Check: parse;execute

DomainCreator:
  # ...
```

New:

```yaml
Executable: EvolveScalarWave1D
Testing:
  Check: parse;execute

---

DomainCreator:
  # ...
```

Metadata provide information for tools, whereas options provide information to the executable. Tools like `CheckOutputFiles` use the metadata for tests. Metadata can also provide information on how to run the input file, such as the name and version of the executable, and a description that may refer to published papers for details on the configuration (particularly useful when publishing input files as supplementary data to papers). Example:

```yaml
# Input file published with a paper:

Description: |
  Briefly describe the configuration and link to papers for details.
Executable: EvolveScalarWave1D
Version: 2023.04.07

---

DomainCreator:
  # ...
```

Closes #3731.

### Upgrade instructions

Every input file must now start with a metadata section terminated by `---`, like this:

```yaml
# Metadata here

---

# Options start here
```

The metadata section may also be empty:

```yaml
---
---

# Options start here
```

See [the option parsing documentation](https://spectre-code.org/dev_guide_option_parsing.html) for details.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Options/ParseOptions.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+#include "Informer/InfoFromBuild.hpp"
+#include "Utilities/FileSystem.hpp"
+
+namespace Options::detail {
+namespace {
+void check_metadata(const YAML::Node& metadata) {
+  // Validate executable name
+  if (const auto& exec_name = metadata["Executable"]) {
+    if (file_system::get_file_name(exec_name.as<std::string>()) !=
```
wthrowe, line 19: When would this be a path?

tests/InputFiles/Burgers/Step.yaml
```diff
@@ -1,10 +1,12 @@
...
 # See LICENSE.txt for details.
 
-# Executable: EvolveBurgers
-# CommandLineArgs: +balancer RandCentLB +p2
-# Timeout: 5
-# Check: parse;execute
+Executable: EvolveBurgers
+CommandLineArgs: +balancer RandCentLB +p2
+Timeout: 5
+Check: parse;execute
```
wthrowe: Can we somehow separate the test-related options from the production-related options, maybe in a "Testing" group or something?  Having a "CommandLineArgs" option in the example that is completely ignored in production is confusing.

wthrowe (changes requested):
Can you elaborate in the docs and upgrade instructions on when you need "---" and when "---\n---"?  I assume this is clear if you know the details of the YAML spec, but I'd rather not require our users to learn all the finer points there.

src/Options/ParseOptions.cpp
nilsvu: We may use the "Executable" field to tell tools which executable to run, in which case we sometimes want to specify an absolute path.

tests/InputFiles/Burgers/Step.yaml
nilsvu: Yes good idea!

nilsvu:
@wthrowe I added this paragraph (squashed in):

> You only need the leading `---` marker if the metadata section is empty. This is YAML's "document start marker" (see the [YAML spec](https://yaml.org/spec/1.2)). Any metadata fields at the beginning of the file also imply the start of a document, so you don't need the first `---` marker.

src/Options/ParseOptions.cpp
wthrowe: OK.  I can imagine that being useful for batch runs or something.

wthrowe (approved):
The documentation addition clears up my confusion there.

I'm happy with this, but this is really user-visible so we should give other people some time to look if they want to.

tests/InputFiles/CurvedScalarWave/PlaneWaveMinkowski3D.yaml
```diff
@@ -1,18 +1,21 @@
...
+  Timeout: 15
+ExpectedOutput:
+  - PlaneWaveMinkowski3DVolume0.h5
+  - PlaneWaveMinkowski3DReductions.h5
+OutputFileChecks:
+  - Label: constraint violations
+    Subfile: /Norms.dat
+    FileGlob: PlaneWaveMinkowski3DReductions.h5
+    SkipColumns: [0, 1, 2]
+    AbsoluteTolerance: 0.02
```
knelli2, line 16: I've been going back and forth on how I feel about having testing stuff at the top of these input files. Currently, we use the "test" input files also as the "example" input files that we point users to. I think it would be nice to keep these input files where they are and have them be used for testing only. These input files would have the `Testing:` block in the metadata like it does here. Then, separately, we have a directory `$SPECTRE_HOME/ExampleInputFiles/` where we hold input files that don't have the `Testing:` block (or `ExpectedOutput:` or `OutputFileChecks`). 

The downside to this is that these input files wouldn't be checked automatically (or we add special CMake magic to only run a parse test on all of them). I'm not sure which one is better because from a user perspective, having the example input files in the root of the repo (I think) makes the most logical sense rather than in the `test` directory. However, it may be more work for us to maintain two copies of some input files.

Another pro for having the `ExampleInputFiles/` directory is we can set some reasonable defaults for all inputs like domain, time stepping, etc. We could even add EventsAndTriggers we'd like people to run (because they already just copy the input files anyways, that way they are copying something we want them to use).

Either way, I think the `ExpectedOutput:` and `OutputFileChecks:` blocks should be nested inside the `Testing:` block because they are only used for testing.

nilsvu: Quick reply to this part:

> Either way, I think the ExpectedOutput: and OutputFileChecks: blocks should be nested inside the Testing: block because they are only used for testing.

These two blocks are used by the `CleanOutput` and `CheckOutputFiles` tools, respectively. At least `CleanOutput` is used outside of tests, e.g. to clean up a failed run before rerunning it. My draft of a scheduling script also invokes`CleanOutput` before running if passed a "cleanup" flag. Since these are separate tools, I chose to keep their options outside the `Testing` block. The `Testing` block is what's read by CMake.

knelli2: Ah ok, that seems reasonable. If the `CheckOutputFiles` block is actually only used for testing, I think it should be moved, but I don't know what your plans are for the scheduler.

cmake/AddSpectreExecutable.cmake
```diff
@@ -33,6 +33,7 @@ function(add_spectre_executable TARGET_NAME)
     RULE_LAUNCH_LINK "${CMAKE_BINARY_DIR}/tmp/WrapExecutableLinker.sh"
     LINK_DEPENDS "${CMAKE_BINARY_DIR}/tmp/WrapExecutableLinker.sh"
     )
+  target_link_options(${TARGET_NAME} PRIVATE "-DEXECUTABLE_NAME=${TARGET_NAME}")
```
nilsdeppe, line 36: I'm not sure what to do about the fact that someone can rename the executable on disk. I don't have a change request here, just something worth being aware of. Maybe we should call it something like "EXECUTABLE_LABEL"? I'm not sure, just wanted to bring it up :)

nilsdeppe (commented):
I'm happy with this. One question/thought about the executable name, but nothing that should hold up this PR

cmake/AddSpectreExecutable.cmake
nilsvu: Yea I thought about that too, and don't know what to do about it. We may just have to skip the check if the metadata specifies an absolute path.

nilsdeppe: I wonder if what we really "want" are 2 different things:
1. control the executable used for testing
2. check that the underlying executable, whatever it may be, is actually the right one for the input file. That is, we label the executables, but that label doesn't have to be the name. Now, for testing we could just have it be the same.
This could mean executables check that the input file "ExecutableLabel" is the same as the internal label, and then for tests, we just use that as the name in `BUILD/bin`.

Again, not sure, just some musings.

nilsvu: Right, I'll revisit this with the scheduling tool.

nilsdeppe: Yea, sounds good. Like I said, I don't know the answer and there's no reason to delay this PR for it. Thanks!

tests/InputFiles/CurvedScalarWave/PlaneWaveMinkowski3D.yaml
nilsvu: `CheckOutputFiles` is mostly used for testing, but it's an independent tool so you might use it outside of tests to run checks over you output data. E.g., check that the shift is zero to some tolerance in a stable TOV evolution or something like that. As I wrote above, the `Testing` block is what's read by CMake to set up the input file tests.

nilsvu:
@knelli2 @wthrowe @nilsdeppe can you take another look / merge?

knelli2 (approved):
I think this looks good.

Not for this PR, but I still think we should separate out test input files and example input files so we can have reasonable production parameters in the examples and then short/low-res parameters in the test input files.

PR #4943, opened 2023-04-21
Weyl Type D1

acifajkya:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/GeneralRelativity/Tags.hpp
+++ b/src/PointwiseFunctions/GeneralRelativity/Tags.hpp
@@ -256,6 +256,15 @@ struct WeylMagnetic : db::SimpleTag {
 };
 
 /*!
+ * \brief Computes a quantity measuring how far from type D spacetime is,
+ * using measure D1 [Eq. (8)] of \cite Bhagwat2017tkm.
+ */
+template <typename DataType, size_t Dim, typename Frame>
+struct WeylTypeD1 : db::SimpleTag {
+  using type = tnsr::ii<DataType, 3, Frame>;
+};
+
+/*!
  * \brief Computes the scalar \f$E_{ij} E^{ij}\f$ from the electric part of the
  * Weyl tensor \f$E_{ij}\f$ and the inverse spatial metric \f$\gamma^{ij}\f$,
  * i.e. \f$E_{ij} E^{ij} = \gamma^{ik}\gamma^{jl}E_{ij}E_{kl}\f$.
@@ -273,6 +282,17 @@ template <typename DataType>
 struct WeylMagneticScalar : db::SimpleTag {
   using type = Scalar<DataType>;
 };
+
+/*!
+ * \brief Computes the scalar \f$D_{ij} D^{ij}\f$ (Eq. (8) of
+ * \cite Bhagwat2017tkm) from \f$D_{ij}\f$ and the inverse spatial metric
+ * \f$\gamma^{ij}\f$, i.e. \f$D = \gamma^{ik}\gamma^{jl}D_{ij}D_{kl}\f$.
+ */
+template <typename DataType>
+struct WeylTypeD1Scalar : db::SimpleTag {
+  using type = Scalar<DataType>;
+};
+
 }  // namespace Tags
 
 /// GR Tags commonly needed for the evolution of hydro systems
--- a/src/PointwiseFunctions/GeneralRelativity/WeylTypeD1.cpp
+++ b/src/PointwiseFunctions/GeneralRelativity/WeylTypeD1.cpp
@@ -1,0 +1,170 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "PointwiseFunctions/GeneralRelativity/WeylTypeD1.hpp"
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/VectorImpl.hpp"
+#include "PointwiseFunctions/GeneralRelativity/WeylElectric.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeWithValue.hpp"
+
+namespace {
+
+template <typename DataType, size_t SpatialDim, typename Frame>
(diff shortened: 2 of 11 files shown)


src/PointwiseFunctions/GeneralRelativity/Tags.hpp
```diff
@@ -255,6 +264,17 @@ template <typename DataType>
 struct WeylMagneticScalar : db::SimpleTag {
   using type = Scalar<DataType>;
 };
+
+/*!
+ * \brief Computes the scalar \f$D_{ij} D^{ij}\f$ (from equation 8 of
+ * \cite Bhagwat2017tkm) from \f$D_{ij}\f$ and the inverse spatial metric
+ * \f$\gamma^{ij}\f$, i.e. \f$D_{ij} = \gamma^{ik}\gamma^{jl}E_{ij}D_{kl}\f$.
```
wthrowe: There's some sort of index confusion in the last equation here.

src/PointwiseFunctions/GeneralRelativity/WeylTypeD1.cpp
```diff
@@ -0,0 +1,172 @@
...
+tnsr::Ij<DataType, SpatialDim, Frame> up_down_from_lower_rank_2_tensor(
+    const tnsr::ii<DataType, SpatialDim, Frame>& lower_rank_2_tensor,
+    const tnsr::II<DataType, SpatialDim, Frame>& inverse_spatial_metric) {
+  auto result = make_with_value<tnsr::Ij<DataType, SpatialDim, Frame>>(
+      get<0, 0>(inverse_spatial_metric), 0.0);
+  for (size_t i = 0; i < SpatialDim; ++i) {
+    for (size_t j = 0; j < SpatialDim; ++j) {
+      for (size_t k = 0; k < SpatialDim; ++k) {
+        result.get(j, k) +=
+            lower_rank_2_tensor.get(i, k) * inverse_spatial_metric.get(i, j);
```
wthrowe, line 26: [optional] A lot of things in this file would look a lot cleaner using tensor expressions.  For example, these nested loops could be (untested)
```c++
tenex::evaluate<ti::J, ti::k>(
    make_not_null(&result),
    lower_rank_2_tensor(ti::i, ti::k) * inverse_spatial_metric(ti::I, ti::J));
```
(Or leave off the result argument for return-by-value.)  With that change you wouldn't even need to split this out as a helper function.

It can also handle more complicated expressions like the one for `weyl_type_D1` below.

src/PointwiseFunctions/GeneralRelativity/WeylTypeD1.hpp
```diff
@@ -0,0 +1,118 @@
...
+/*!
+ * \ingroup GeneralRelativityGroup
+ * \brief Computes a quantity measuring how far from type D spacetime is.
+ *
+ * \details Computes a quantity measuring how far from type D spacetime is,
+ * using measure D1 [Eq. (8) of \cite Bhagwat2017tkm]:
+ *
+ * \f{align}{
+ * \frac{a}{12} \gamma_{ij} - \frac{b}{a} E_{ij} - 4
+ E_{i}^{k} E_{jk} = 0 \f}
```
wthrowe: Formatting.

src/PointwiseFunctions/GeneralRelativity/WeylTypeD1.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+/// @{
+/*!
+ * \ingroup GeneralRelativityGroup
+ * \brief Computes the scalar \f$D_{ij} D^{ij}\f$ , a measure of a spacetime's
+ * devitation from type D.
+ *
+ * \details Computes the scalar \f$D_{ij} D^{ij}\f$ from \f$D_{ij}\f$ [Eq. (8)
+ * of \cite Bhagwat2017tkm] and the inverse spatial metric \f$\gamma^{ij}\f$,
+ * i.e. \f$D_{ij} = \gamma^{ik}\gamma^{jl}E_{ij}D_{kl}\f$.
```
wthrowe: Index confusion again.

src/PointwiseFunctions/GeneralRelativity/WeylTypeD1.cpp
```diff
@@ -0,0 +1,172 @@
...
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/VectorImpl.hpp"
+#include "PointwiseFunctions/GeneralRelativity/WeylElectric.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeWithValue.hpp"
+
+//
```
wthrowe: Remove empty comment.

src/PointwiseFunctions/GeneralRelativity/WeylTypeD1.cpp
```diff
@@ -0,0 +1,172 @@
...
+    const gsl::not_null<tnsr::ii<DataType, SpatialDim, Frame>*> weyl_type_D1,
+    const tnsr::ii<DataType, SpatialDim, Frame>& weyl_electric,
+    const tnsr::ii<DataType, SpatialDim, Frame>& spatial_metric,
+    const tnsr::II<DataType, SpatialDim, Frame>& inverse_spatial_metric) {
+  auto weyl_electric_up_down =
+      up_down_from_lower_rank_2_tensor(weyl_electric, inverse_spatial_metric);
+
+  auto upper_weyl_electric =
+      make_with_value<tnsr::II<DataType, SpatialDim, Frame>>(
+          get<0, 0>(inverse_spatial_metric), 0.0);
```
wthrowe, line 47: [optional] This function does a lot of memory allocations.  Since it looks like this is mostly for observing, that's not a big deal, but you will get better performance if you use a `Variables` with temporary tags, as is done for `Psi4`, for example.

wthrowe (changes requested):
I haven't compared with the reference, but I did check that the code and documentation match, except for the case that I commented on.

wthrowe:
There are still some errors from the template-parameter-order change.  When you fix those you can squash all the fixups in.

wthrowe (approved):

PR #4949, opened 2023-04-25
Add SpEC initial data classes

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
Discussing with @markscheel we came to the conclusion that trying to link with SpEC for testing this on CI is probably not worth the effort. Distributing the SpEC exporter in the public SpECTRE CI container may be problematic because SpEC is private (at the moment). We could distribute the SpEC exporter in another private container, or host the binary+header somewhere, but that's not trivial either. In practice, we will have SpEC installed on clusters and link to it. So, reluctantly, I suggest we merge this without tests.

Edit: This would be another use case for integration tests that we run on clusters.

cmake/SetupSpec.cmake
```diff
@@ -0,0 +1,42 @@
...
+
+# Optionally link SpEC libraries. Pass `SPEC_ROOT` to the CMake build
+# configuration to set up the following targets:
+#
+# - SpEC::Exporter: Functionality to load SpEC volume data and interpolate to
+#   arbitrary points. At the time of writing (Apr 20, 2023) you need to compile
+#   the exporter manually in SpEC's `Support/ApplyObservers/Exporter/` directory
+#   (see the `Makefile` in that directory).
+
+option(SPEC_ROOT "Path to the git directory of SpEC." OFF)
```
nilsdeppe, line 10: Can you document this in our build system docs where we describe "all" the CMake flags? Most of the notes in this file like MPI version & building the exporter manually should be somewhere in documentation. Maybe installation notes for the optional dependency or even a very. brief tutorial page? Not sure

src/PointwiseFunctions/AnalyticData/GeneralRelativity/InterpolateFromSpec.hpp
```diff
@@ -0,0 +1,97 @@
...
+  //            std::to_string(dataset_names.size()) +
+  //            " variables to interpolate, but requested " +
+  //            std::to_string(tmpl::size<Tags>::value) + " tags.");
+  // Transform coordinates into SpEC's expected format. SpEC expects grid
+  // coordinates. We assume that the grid coordinates coincide with inertial
+  // coordinates for the initial data.
+  const size_t num_points = get_size(get<0>(x));
+  std::vector<std::vector<double>> spec_grid_coords{num_points};
+  for (size_t i = 0; i < num_points; ++i) {
+    spec_grid_coords[i] = std::vector<double>{get_element(get<0>(x), i),
```
nilsdeppe, line 59: you can remove the `<double>` if you want here

src/PointwiseFunctions/AnalyticData/GeneralRelativity/InterpolateFromSpec.hpp
```diff
@@ -0,0 +1,97 @@
...
+                                              get_element(get<1>(x), i),
+                                              get_element(get<2>(x), i)};
+  }
+  // Allocate memory for the interpolation and point into it
+  tuples::tagged_tuple_from_typelist<Tags> interpolation_buffer{};
+  std::vector<std::vector<double*>> buffer_pointers{tmpl::size<Tags>::value};
```
nilsdeppe: std::vector constructors are silly, are you trying to create a 1-element vector or a vector of a given size? If given size, then `{}` -> `()` is safer

src/PointwiseFunctions/AnalyticData/GeneralRelativity/InterpolateFromSpec.hpp
```diff
@@ -0,0 +1,97 @@
...
+    using tag = tmpl::type_from<std::decay_t<decltype(tag_v)>>;
+    auto& tensor = get<tag>(interpolation_buffer);
+    destructive_resize_components(make_not_null(&tensor), num_points);
+    const size_t num_components = tensor.size();
+    buffer_pointers[var_i].resize(num_components);
+    // The SpEC exporter supports tensors up to symmetric rank 2, which are
+    // ordered xx, xy, xz, yy, yz, zz. Because this is also the order in
+    // which we store components in the Tensor class, we don't have to do
+    // anything special here.
+    // WARNING: If the Tensor storage order changes for some reason, this
```
nilsdeppe, line 78: Would it be possible to add an assert based on the `component_suffix` function?

src/PointwiseFunctions/AnalyticData/GeneralRelativity/SpecInitialData.hpp
```diff
@@ -0,0 +1,112 @@
...
+  /// These quantities are supported for interpolation from SpEC
+  template <typename DataType>
+  using interpolated_tags =
+      tmpl::list<gr::Tags::SpatialMetric<3, Frame::Inertial, DataType>,
+                 gr::Tags::ExtrinsicCurvature<3, Frame::Inertial, DataType>,
+                 gr::Tags::Lapse<DataType>,
+                 gr::Tags::Shift<3, Frame::Inertial, DataType>>;
+
+  /// These are the names in SpEC datasets corresponding to the quantities above
+  static const inline std::vector<std::string> vars_to_interpolate_{
```
nilsdeppe, line 100: Make these an input file option? For doing BNS initial data we will need additional vars

nilsdeppe (changes requested):
overall LGTM, some small thoughts

src/PointwiseFunctions/AnalyticData/GeneralRelativity/SpecInitialData.hpp
nilsvu: BNS ID is in a separate class, so I think this is ok.

nilsvu:
Updated with the requested changes. I think this is ready for review/merge. We still have to sort out some thread-safety issues in SpEC, but since we can only test things "in production" anyway it's easier to merge and fix issues when they come up.

nilsvu:
@nilsdeppe this is ready for another look

nilsdeppe (approved):

PR #4957, opened 2023-04-27
Add debug prints to control system

knelli2:
## Proposed changes

I found my self adding the same prints every time I needed to debug the control system so I just made them part of develop.

### Upgrade instructions

Add
```yaml
ControlSystems:
  ...
  Verbosity: Silent
```
to you input file with control systems. To turn on debug prints for the control system, change the verbosity to `Verbose` or `Debug`. Note that `Debug` will print things for every element of the domain.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Event.hpp
+++ b/src/ControlSystem/Event.hpp
@@ -11,7 +11,10 @@
 #include "DataStructures/LinkedMessageId.hpp"
 #include "Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp"
 #include "Domain/FunctionsOfTime/Tags.hpp"
+#include "IO/Logging/Verbosity.hpp"
 #include "ParallelAlgorithms/EventsAndTriggers/Event.hpp"
+#include "Utilities/GetOutput.hpp"
+#include "Utilities/PrettyType.hpp"
 #include "Utilities/ProtocolHelpers.hpp"
 #include "Utilities/Serialization/CharmPupable.hpp"
 #include "Utilities/TMPL.hpp"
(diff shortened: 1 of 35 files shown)


src/Utilities/PrettyType.hpp
```diff
@@ -739,4 +739,30 @@ std::string name(const T& /*unused*/) {
...
+template <typename List>
+std::string list_of_names() {
+  static_assert(tt::is_a_v<tmpl::list, List>);
+  std::stringstream ss{};
+  tmpl::for_each<List>([&ss](auto v) {
+    using type = tmpl::type_from<decltype(v)>;
+    static_assert(not tt::is_a_v<tmpl::list, type>,
+                  "The tmpl::list provided to pretty_type::list_of_names must "
+                  "be flattened.");
+    if constexpr (tmpl::index_of<List, type>::value != 0) {
```
wthrowe: This breaks if the first type appears multiple times.  There are compile-time ways to do this, but I'd just capture a bool.

src/ControlSystem/Event.hpp
```diff
@@ -97,6 +101,13 @@ class Event : public ::Event {
           using Submeasurement = tmpl::type_from<decltype(submeasurement)>;
           db::apply<Submeasurement>(box, measurement_id, cache, array_index,
                                     component, ControlSystems{});
+          if (Parallel::get<Tags::Verbosity>(cache) >= ::Verbosity::Debug) {
+            Parallel::printf(
+                "%s, time = %s: Running control system events for measurement "
+                "'%s'.\n",
+                array_index, measurement_id,
```
wthrowe: I think in tests we often use `ArrayIndex = int`, and that that will crash `printf` because it wants `%d` as a format specifier.  Maybe we don't care, but if we do wrapping in `get_output` should be safe.

wthrowe (changes requested):

knelli2:
Ok squashed (and rebased)

wthrowe (approved):

nilsvu (approved):

PR #4962, opened 2023-04-28
Add integer pow function

nikwit:
## Proposed changes
Adds a function that computes the power of a DataVector or double with integer powers not known at compile time.

### DIFF
--- a/src/DataStructures/Blaze/IntegerPow.hpp
+++ b/src/DataStructures/Blaze/IntegerPow.hpp
@@ -1,0 +1,101 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <blaze/math/DenseVector.h>
+#include <blaze/math/constraints/SIMDPack.h>
+#include <blaze/math/simd/BasicTypes.h>
+#include <blaze/system/Inline.h>
+#include <blaze/system/Vectorization.h>
+
+#include "Utilities/Math.hpp"
+
+namespace blaze {
+
+namespace detail {
+template <typename T>
+BLAZE_ALWAYS_INLINE SIMDdouble integer_pow_impl(const SIMDf64<T>& x,
+                                                const int e) {
+  ASSERT(e >= 0, "Negative powers are not implemented");
+  int ecount = e;
+  int bitcount = 1;
+  while (ecount >>= 1) {
+    ++bitcount;
+  }
+  SIMDdouble result = blaze::set(1.0);
+  while (bitcount) {
+    result *= result;
+    if ((e >> --bitcount) & 0x1) {
+      result *= x;
+    }
+  }
+  return result;
+}
+}  // namespace detail
+// This vectorized implementation of the integer pow function is necessary
+// because blaze does not offer its own version of a integer pow function.
+template <typename T>
+BLAZE_ALWAYS_INLINE SIMDdouble integer_pow(const SIMDf64<T>& b, const int e) {
+  switch (e) {
+    case 0:
+      return blaze::set(1.0);
+    case 1:
+      return b;
+    case 2:
+      return b * b;
+    case 3:
+      return b * b * b;
+    case 4: {
+      const SIMDdouble b2 = b * b;
+      return b2 * b2;
+    }
+    case 5: {
+      const SIMDdouble b2 = b * b;
+      return b2 * b2 * b;
+    }
+    case 6: {
+      const SIMDdouble b2 = b * b;
+      return b2 * b2 * b2;
+    }
+    case 7: {
+      const SIMDdouble b2 = b * b;
+      return b2 * b2 * b2 * b;
+    }
+    case 8: {
+      const SIMDdouble b2 = b * b;
+      const SIMDdouble b4 = b2 * b2;
+      return b4 * b4;
+    }
+    default:
+      return detail::integer_pow_impl(b, e);
+  }
+}
+
+struct IntegerPow {
+  int exponent;
+  explicit inline IntegerPow(const int e) : exponent(e) {}
+
+  BLAZE_ALWAYS_INLINE double operator()(const double a) const {
+    return ::integer_pow(a, exponent);
+  }
+
+  template <typename T>
+  BLAZE_ALWAYS_INLINE decltype(auto) load(const T& a) const {
+    BLAZE_CONSTRAINT_MUST_BE_SIMD_PACK(T);
+    return integer_pow(a, exponent);
+  }
+};
+}  // namespace blaze
+
+template <typename VT, bool TF>
+BLAZE_ALWAYS_INLINE decltype(auto) integer_pow(
+    const blaze::DenseVector<VT, TF>& vec, const int e) {
+  return map(*vec, blaze::IntegerPow{e});
+}
+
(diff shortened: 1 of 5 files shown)


src/Utilities/Math.hpp
```diff
@@ -162,3 +164,31 @@ template <typename T>
...
   return sgn_detail::sgn(val, std::is_signed<T>{});
 }
+
+/// \ingroup UtilitiesGroup
+/// \brief Raises type T to the integer power n.
+///
+/// \details The power is computed by squaring T to the closest power of 2 lower
+/// than n and then multiplying the remainder on. There are implementations with
+/// fewer multiplications but those require at least one allocation which is
+/// avoided here.
```
wthrowe: SpEC has the algorithm you want.

wthrowe (changes requested):

src/Utilities/Math.hpp
nikwit: The SpEC algorithm allocates temporaries. Are we sure we want that? We have often said that we can do a *a lot* of extra calculations before they become as expensive as an allocation in our parallel environment.

wthrowe: Where do you see temporaries?
```c++
  template<class T> T exp_by_squaring(const T b, const int e) {
    int ecount = e;
    int bitcount = 1;
    while(ecount >>= 1) {
      ++bitcount;
    }
    T ret = 1;
    while(bitcount) {
      ret *= ret;
      if((e >> --bitcount) & 0x1) {
        ret *= b;
      }
    }
    return ret;
  }
```

nikwit: ah sorry, I was looking at the wrong function

src/Utilities/Math.hpp
```diff
@@ -162,3 +164,23 @@ template <typename T>
...
+/// \brief Raises type T to the integer power n.
+template <typename T>
+T integer_pow(const T& x, const int e) {
+  ASSERT(e >= 0, "Negative powers are not implemented");
+  int ecount = e;
+  int bitcount = 1;
+  while (ecount >>= 1) {
+    ++bitcount;
+  }
+  auto result = make_with_value<T>(x, 1.);
```
nilsdeppe: This is bad, it's doing a memory allocation... I think the better thing to do is to plug into Blaze directly similar to how `src/DataStructures/Blaze/StepFunction.hpp` works. Then provide an overload for double/float. There was also some concern of temporaries in SpEC: that only matters if you are operating on the low-level type, and I would expect the switch-case to be faster than this more complicated implementation. Basically, the way I think we should do this is add the low-level support rather than trying to figure out some way around it at the high level.

nilsdeppe (changes requested):

src/Utilities/Math.hpp
nikwit: Are my new commits roughly what you meant? Im not sure what you mean by low level in this context.

src/Utilities/Math.hpp
```diff
@@ -162,3 +164,22 @@ template <typename T>
 constexpr T sgn(const T& val) {
   return sgn_detail::sgn(val, std::is_signed<T>{});
 }
+
+/// \ingroup UtilitiesGroup
+/// \brief Raises a double to the integer power n.
+BLAZE_ALWAYS_INLINE double integer_pow(const double x, const int e) {
```
wthrowe: `BLAZE_ALWAYS_INLINE` isn't defined here in some setups.  `SPECTRE_ALWAYS_INLINE` would be the one for this file, although a plain `inline` is probably more appropriate in this case.

wthrowe (changes requested):
The `double` version is fine, apart from the compilation issue.  I've never looked into the low-level blaze stuff, so I'll leave that commit to @nilsdeppe.

src/DataStructures/Blaze/IntegerPow.hpp
```diff
@@ -0,0 +1,75 @@
...
+#include "Utilities/Math.hpp"
+
+namespace blaze {
+// This vectorized implementation of the integer pow function is necessary
+// because blaze does not offer its own version of a integer pow function.
+template <typename T>
+BLAZE_ALWAYS_INLINE SIMDdouble integer_pow(const SIMDf64<T>& v, const int e)
+#if BLAZE_AVX512F_MODE || BLAZE_MIC_MODE
+{
+  return _mm512_set_pd(::integer_pow((*v).eval().value[7], e),
```
nilsdeppe: This sort of approach is generally risky (can't always avoid it) because this will actually not vectorize the code. It'll extract each element and then re-pack them. That's incredibly slow because it requires a ton of register loads and flushes. What I had in mind was:
```cpp
namespace detail {
BLAZE_ALWAYS_INLINE SIMDdouble integer_pow_impl(const SIMDf64<T>& x, const int e) {
  ASSERT(e >= 0, "Negative powers are not implemented");
  int ecount = e;
  int bitcount = 1;
  while (ecount >>= 1) {
    ++bitcount;
  }
  double result = 1.;
  while (bitcount) {
    result *= result;
    if ((e >> --bitcount) & 0x1) {
      result *= x;
    }
  }
  return result;
}
}  // namespace detail

BLAZE_ALWAYS_INLINE SIMDdouble integer_pow(const SIMDf64<T>& x, const int e) {
switch(e) {
    case 0:
      return blaze::set(1.0);
    case 1:
      return b;
    case 2:
      return b*b;
    case 3:
      return b*b*b;
    case 4:
      {
        const SIMDdouble b2 = b*b;
        return b2*b2;
      }
    case 5:
      {
        const SIMDdouble b2 = b*b;
        return b2*b2*b;
      }
    case 6:
      {
        const SIMDdouble b2 = b*b;
        return b2*b2*b2;
      }
    case 7:
      {
        const SIMDdouble b2 = b*b;
        return b2*b2*b2*b;
      }
    case 8:
      {
        const SIMDdouble b2 = b*b;
        const SIMDdouble b4 = b2*b2;
        return b4*b4;
      }
    default:
      return detail::integer_pow_impl(b, e);
    }
  }
```

src/DataStructures/Blaze/IntegerPow.hpp
```diff
@@ -0,0 +1,75 @@
...
+}
+#endif
+
+struct IntegerPow {
+  int exponent;
+  explicit inline IntegerPow(const int e) : exponent(e) {}
+
+  template <typename T>
+  BLAZE_ALWAYS_INLINE decltype(auto) operator()(const T& a) const {
+    return ::integer_pow(a, exponent);
```
nilsdeppe, line 80: I've forgotten the blaze stuff specifically for this, but this is always going to be a `double`? That is, `std::is_same_v<T, double>` is always true? If so, maybe just hard-code the `double`? If not, I would do:
```cpp
using ::integer_pow;
    return ::integer_pow(a, exponent);
```
then it'll always pick the right one.

nilsdeppe:
LGTM, please squash and rebase!

nikwit:
done, thanks for the reviews

nilsdeppe (approved):
@wthrowe could you please take a look when you've got a chance? Thanks!

wthrowe (approved):
Ignoring timeouts.

PR #4964, opened 2023-04-28
Speed up `DampedHarmonic`

nikwit:
## Proposed changes

Profiling showed that calculating the gauge function took  >60% of the `GH::TimeDerivative` calculation when using `DampedHarmonic`. This is because:

- it used a lot of `pow` calls which can be optimized for integers
- it repeatedly called the function `GeneralizedHarmonic::gauges::DampedHarmonicGauge_detail::spacetime_deriv_of_power_log_factor_metric_lapse` which does several allocations  and computes the same expensive temporaries every call.

This optimization optimizes the `pow` calls and removes the repeated allocations and computations of the same variables. On a single core, `GH::TimeDerivative` now takes 60 % of the original runtime with `DampedHarmonicGauge`. The `DampedHarmonic` calculation now takes 20 % of the  `GH::TimeDerivative` instead of 60 %.  There are certainly more optimizations that can be done here.

~depends on #4962~

@markscheel as discussed, it would be good to see if the implementation is the same in SpEC.


isaaclegred:
You can get O(several times) speedup in `integer_pow` by doing casework with the small-integer exponents.  I.e. do like a switch on exponent. E.g. 
```
Run on (24 X 3300 MHz CPU s)
2023-04-28 12:58:57
-------------------------------------------------------------
Benchmark                      Time           CPU Iterations
-------------------------------------------------------------
bench_ipow                   169 ns        169 ns    3947106
bench_ipow_with_cases         58 ns         58 ns   12126659
```

nikwit:
thanks @isaaclegred .This PR replaces a `pow(x, n)` where n was transformed to `double` to a direct integer calculation. Even with my suboptimal implementations it did not even show up on a profile anymore so in this specific case it does not matter much.

isaaclegred:
Oh yeah of course, 
just to be clear the results above are comparing the current version of integer_pow with a potential version using cases.  Neither does `pow` calls.  That distinction looks more like 
```
Run on (24 X 3300 MHz CPU s)
2023-04-28 13:44:30
-------------------------------------------------------------
Benchmark                      Time           CPU Iterations
-------------------------------------------------------------
bench_ipow                 24539 ns      24563 ns      24601
bench_pow                 254076 ns     254324 ns       2753
bench_ipow_with_cases       3640 ns       3642 ns     191855
```
As you've noted, the first order of magnitude is the one that makes the real difference :)
(this is squaring a DataVector with 10000 elements)

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
```diff
@@ -257,39 +279,66 @@ void damped_harmonic_impl(
     get<0>(d4_RW_L2) += get(weight) * d0_roll_on;
   }
 
+  time_deriv_of_lapse<DataVector, SpatialDim, Frame>(
+      make_not_null(&dt_lapse), lapse, shift, spacetime_unit_normal, phi, pi);
+  spatial_deriv_of_lapse<DataVector, SpatialDim, Frame>(
+      make_not_null(&d3_lapse), lapse, spacetime_unit_normal, phi);
```
nilsdeppe: You can reduce the work here by doing `-lapse * half_phi_two_normals`

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
```diff
@@ -257,39 +279,66 @@ void damped_harmonic_impl(
     get<0>(d4_RW_L2) += get(weight) * d0_roll_on;
   }
 
+  time_deriv_of_lapse<DataVector, SpatialDim, Frame>(
+      make_not_null(&dt_lapse), lapse, shift, spacetime_unit_normal, phi, pi);
```
nilsdeppe: Here you can use `lapse * (lapse * half_pi_two_normals - shift^i half_phi_two_normals_i)` to eliminate all the normal contractions work that otherwise is done

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
```diff
@@ -257,39 +279,66 @@ void damped_harmonic_impl(
...
+      result->get(a) = get(prefac_log) * d_logfac.get(a);
+    }
+  };
   // \partial_a \mu_{S} = \partial_a(A_S R_S W
   //                               \log(\sqrt{g}/N)^{c_{S}})
   // \partial_a \mu_1 = \partial_a(A_L1 R_L1 W
   //                               \log(\sqrt{g}/N)^{1+c_{L1}})
   // \partial_a \mu_2 = \partial_a(A_L2 R_L2 W
   //                               \log(1/N)^{1+c_{L2}})
-  DampedHarmonicGauge_detail::spacetime_deriv_of_power_log_factor_metric_lapse<
```
nilsdeppe, line 266: Can these detail functions be deleted now?

nilsdeppe (changes requested):
LGTM, a few suggestions you can squash right away :)

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
nikwit: I elided `d3_lapse` all together because it is only used once.

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
nikwit: yes, I removed them in a new commit.

nikwit:
thanks for the review, I pushed the updated changes now.

nilsdeppe (approved):
Nice! LGTM! Thanks for doing this :D

PR #4967, opened 2023-04-28
Add StrahlkorperInDifferentFrameAligned

markscheel:
## Proposed changes

    This is a simplified version of StrahlkorperInDifferentFrame,
    that assumes:
     * The map preserves the center of the Strahlkorper.
     * For each point on the Strahlkorper, the map preserves the angles
       between that point and the center.

Edit: this *does* know about the distorted frame now.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
+++ b/src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
@@ -20,10 +20,76 @@
 #include "NumericalAlgorithms/RootFinding/RootBracketing.hpp"
 #include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
-#include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
+
+namespace {
+// Transforms cartesian coordinates of a strahlkorper
+// from one frame to another, by looping over the blocks and
+// calling the correct map functions.
+template <typename SrcFrame, typename DestFrame>
+void coords_to_different_frame(
+    const gsl::not_null<tnsr::I<DataVector, 3, DestFrame>*>
+        dest_cartesian_coords,
+    const tnsr::I<DataVector, 3, SrcFrame>& src_cartesian_coords,
+    const Domain<3>& domain,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time,
+    const double time) {
+  // If we ever want a source in a different frame than Grid, we
+  // need to add 'if constexpr's for that case.
+  static_assert(std::is_same_v<SrcFrame, ::Frame::Grid>,
+                "Source frame must currently be Grid frame");
+  // We wish to map src_cartesian_coords to the destination frame.
+  // Each Block will have a different map, so the mapping must be done
+  // Block by Block.
+  // Note that if a point is on the boundary of two or more
+  // Blocks, it might get filled twice, but that's ok.
+  tnsr::I<double, 3, DestFrame> x_dest{};
+  tnsr::I<double, 3, SrcFrame> x_src{};
+  for (size_t s = 0; s < get<0>(src_cartesian_coords).size(); ++s) {
+    get<0>(x_src) = get<0>(src_cartesian_coords)[s];
+    get<1>(x_src) = get<1>(src_cartesian_coords)[s];
+    get<2>(x_src) = get<2>(src_cartesian_coords)[s];
+    for (const auto& block : domain.blocks()) {
+      const auto& logical_to_grid_map = block.moving_mesh_logical_to_grid_map();
+      const auto x_logical = logical_to_grid_map.inverse(x_src);
+      // x_logical might be an empty std::optional.
+      if (x_logical.has_value() and get<0>(x_logical.value()) <= 1.0 and
+          get<0>(x_logical.value()) >= -1.0 and
+          get<1>(x_logical.value()) <= 1.0 and
+          get<1>(x_logical.value()) >= -1.0 and
+          get<2>(x_logical.value()) <= 1.0 and
+          get<2>(x_logical.value()) >= -1.0) {
+        // If we get here (that is, if the logical coords are between -1 and
+        // +1), then the point is in this block. So we map it.
+        if constexpr (std::is_same_v<DestFrame, ::Frame::Distorted>) {
+          if (not block.has_distorted_frame()) {
+            throw std::runtime_error(
+                "Strahlkorper lies outside of distorted-frame region");
+          }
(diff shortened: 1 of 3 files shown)


src/ApparentHorizons/StrahlkorperInDifferentFrameHelpers.hpp
```diff
@@ -0,0 +1,35 @@
...
+}  // namespace domain::FunctionsOfTime
+namespace gsl {
+template <typename T>
+class not_null;
+}  // namespace gsl
+/// \endcond
+
+namespace strahlkorper_in_different_frame_helpers {
+template <typename SrcFrame, typename DestFrame>
+void coords_to_different_frame(
```
knelli2: Can you add a small amount of docs? Just saying that this loops over the blocks of the domain and transforms the coords from the source frame to the destination frame.

src/ApparentHorizons/StrahlkorperInDifferentFrameAligned.hpp
```diff
@@ -0,0 +1,53 @@
...
+/// The destination Strahlkorper is given the same l_max() and m_max()
+/// as the source Strahlkorper.
+///
+/// Note that because the Blocks inside the Domain allow access to
+/// maps only between a selected subset of frames, we cannot use
+/// strahlkorper_in_different_frame to map between arbitrary frames;
+/// allowing strahlkorper_in_different_frame to work on more frames
+/// requires adding member functions to Block.
+template <typename SrcFrame, typename DestFrame>
+void strahlkorper_in_different_frame_aligned(
```
knelli2: Can you place this in the same file as the original function? These are very related and another file just adds cmake complications.

src/ApparentHorizons/StrahlkorperInDifferentFrameAligned.hpp
```diff
@@ -0,0 +1,53 @@
...
+/// invariant. If those assumptions are not satisfied, then
+/// strahlkorper_in_different_frame_aligned will give the wrong answer
+/// and you must use strahlkorper_in_different_frame.  Those
+/// assumptions are satisfied for the shape and size maps but not for
+/// general maps.
+///
+/// The destination Strahlkorper is given the same l_max() and m_max()
+/// as the source Strahlkorper.
+///
+/// Note that because the Blocks inside the Domain allow access to
```
knelli2: Use `\note`

knelli2 (changes requested):
Just a couple things. Related to one of the comments, if you put both functions in the same file, I think you can just put the helper function in an anonymous namespace at the top of the file and that way you don't add any new files.

markscheel:
I put everything in one file, used \note, added some docs, and squashed.

markscheel:
Clang-tidy errors appear to all be from our good friend toms748.

knelli2 (dismissed):

markscheel:
There was a question of whether I can test this function using the Shape TimeDependence instead of the SphericalCompression TimeDependence.  Currently the answer is no, because the Shape TimeDependence has no way of setting a spherically-symmetric map (other than the identity).  I want a spherically-symmetric map so that I can analytically compute the mapped Strahlkorper independently, for the test.

knelli2:
The map coefficients are stored in the functions of time. Could you just make your own functions of time rather than using the ones from the domain creator?

markscheel:
I found a way to use TimeDependence::Shape for the test.  So here's what I did:

* I changed the test to use TimeDependence::Shape
* I closed #4968 and #4969
* I incorporated the distorted-frame stuff in #4969 into this PR (which is not that big of change now that I didn't need #4968 anymore).

src/ApparentHorizons/StrahlkorperInDifferentFrame.hpp
```diff
@@ -26,7 +26,7 @@ class not_null;
 /// The destination Strahlkorper has the same l_max() and m_max() as the
 /// source Strahlkorper.
 ///
-/// Note that because the Blocks inside the Domain allow access to
+/// \note Because the Blocks inside the Domain allow access to
 /// maps only between a selected subset of frames, we cannot use
 /// strahlkorper_in_different_frame to map between arbitrary frames;
 /// allowing strahlkorper_in_different_frame to work on more frames
```
knelli2, line 32: Can you add to this note that this function *only* works for SrcFrame = Grid and DestFrame = Inertial.

src/ApparentHorizons/StrahlkorperInDifferentFrame.hpp
```diff
@@ -39,3 +39,25 @@ void strahlkorper_in_different_frame(
...
+/// with the following two assumptions: First, the map leaves the
+/// center of the Strahlkorper invariant. Second, for every point on
+/// the Strahlkorper, the map leaves the angles about the center
+/// (i.e. the quantity
+/// \f$(x^i-c^i)/\sqrt{(x^0-c^0)^2+(x^1-c^1)^2+(x^2-c^2)^2}\f$)
+/// invariant. If those assumptions are not satisfied, then
+/// strahlkorper_in_different_frame_aligned will give the wrong answer
+/// and you must use strahlkorper_in_different_frame.  Those
+/// assumptions are satisfied for the shape and size maps but not for
+/// general maps.
```
knelli2, line 64: And then add a note here that this only works for SrcFrame = Grid and DestFrame = Distorted or Inertial. Just to make it clear that one of these functions can be used in the distorted frame (currently) but the other one can't.

knelli2 (changes requested):

knelli2 (commented):
Looks good. Please squash

markscheel:
Squashed (and rebased onto current develop)

knelli2 (dismissed):
@sxs-collaboration/spectre-core-devs Could one of you look at this? It's needed for size control.

src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
```diff
@@ -263,10 +293,61 @@ void strahlkorper_in_different_frame(
...
+      temp_buffer(src_strahlkorper.ylm_spherepack().physical_size());
+  auto& src_theta_phi =
+      get<::Tags::Tempi<0, 2, ::Frame::Spherical<SrcFrame>>>(temp_buffer);
+  auto& r_hat = get<::Tags::Tempi<1, 3, SrcFrame>>(temp_buffer);
+  auto& src_cartesian_coords = get<Tags::TempI<2, 3, SrcFrame>>(temp_buffer);
+  auto& dest_cartesian_coords = get<Tags::TempI<3, 3, DestFrame>>(temp_buffer);
+  auto& radius = get<::Tags::TempScalar<4>>(temp_buffer);
+
+  StrahlkorperTags::ThetaPhiCompute<SrcFrame>::function(
+      make_not_null(&src_theta_phi), src_strahlkorper);
```
wthrowe: These tag functions are free functions (`StrahlkorperFunctions::theta_phi` here).  Call those instead of going through the tags.

wthrowe (changes requested):

src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
markscheel: Done!

wthrowe:
Looks good.  Squash.

markscheel:
Squashed.

knelli2 (approved):

wthrowe (approved):
clang-tidy failures are just the usual TOMS stuff.

PR #4968, opened 2023-04-29
Add distorted frame to TimeDependence::SphericalCompression

markscheel:
Adds distorted frame to TimeDependence::SphericalCompression

This will be used for tests.
We treat extra arguments the same way they
are treated in UniformTranslation:
providing two sets of functionvstime
arguments means that the first set is
the grid_to_distorted map and the second
set is the distorted_to_inertial map.

Had to refactor Test_SphericalCompression to reduce duplicating code for that test.
Also, replaced an ERROR_TEST in Test_SphericalCompression with a CHECK_THROWS_WITH

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


markscheel:
TimeDependence::SphericalCompression will eventually be eliminated, so I am closing this PR.

PR #4970, opened 2023-04-29
Speed up `GH::TimeDerivative` calculation for harmonic gauge

nikwit:
## Proposed changes
When using harmonic gauge, some terms of the field equations can be set to zero and we can avoid calculating the second order Christoffel symbols and the sqrt of the determinant of the spatial metric alltogether.

This PR adds a check in `GH::TimeDerivative` whether the harmonic gauge is used and only computes these terms if it is not. The uncomputed variables are filled with NaNs in debug mode.

Benchmarking showed this leads to a ~15% speed-up of `GH::TimeDerivative`  on a single core on my laptop.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
@@ -11,6 +11,7 @@
 #include "Evolution/Systems/GeneralizedHarmonic/DuDtTempTags.hpp"
 #include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Dispatch.hpp"
 #include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Gauges.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/Harmonic.hpp"
 #include "Evolution/Systems/GeneralizedHarmonic/System.hpp"
 #include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
 #include "PointwiseFunctions/GeneralRelativity/Christoffel.hpp"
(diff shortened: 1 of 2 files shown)


src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
```diff
@@ -251,21 +251,35 @@ void TimeDerivative<Dim>::apply(
...
+        *inverse_spatial_metric, *da_spacetime_metric, *half_pi_two_normals,
+        *half_phi_two_normals, spacetime_metric, pi, phi, mesh, time,
+        inertial_coords, inverse_jacobian, gauge_condition);
+    // Compute source function last so that we don't need to recompute any of
+    // the other temporary tags.
+    for (size_t nu = 0; nu < Dim + 1; ++nu) {
+      gauge_constraint->get(nu) += gauge_function->get(nu);
+    }
+  } else {
+    get(*sqrt_det_spatial_metric) =
```
nilsdeppe, line 273: Wrap in `#ifdef SPECTRE_DEBUG`? The rest of the code doesn't set `NaN`s in release

nilsdeppe (changes requested):
LGTM, one small suggestion you can squash immediately

src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
nilsdeppe: Hmm, maybe my browser isn't refreshing right something went wrong during a push because I don't see the `ifdef`s

nikwit: sorry, I must have messed something up.

src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.hpp
```diff
@@ -97,6 +97,11 @@ namespace gh {
  * terms added to this implementation of the generalized harmonic system.
  * Mesh-velocity corrections that are applicable to all systems are made in
  * `evolution::dg::Actions::detail::volume_terms()`.
+ *
+ * \warning When using harmonic gauge,
+ * gr::Tags::SqrtDetSpatialMetric<DataVector> and
+ * gr::Tags::SpacetimeChristoffelSecondKind<Dim, Frame::Inertial, DataVector>
+ * are not computed and instead filled with signaling NaNs.
```
knelli2: "...in Debug mode"

knelli2 (changes requested):
Also update the PR description please.

knelli2 (approved):

nilsdeppe (approved):

PR #4977, opened 2023-05-02
Factory-create numeric initial data

nilsvu:
## Proposed changes

Allows to select numeric initial data in the input file, instead of compiling a separate executable for numeric initial data. To keep the size of commits somewhat manageable, this PR only converts the numeric ID importing but doesn't change the behavior of setting initial data from analytic solutions / data.

One change in functionality: The GHMHD exec imports GH and hydro numeric initial data in a single interpolation step now, instead of doing GH first and hydro second.

Reviewers: Select "Hide whitespace".

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Evolution/Executables/GeneralizedHarmonic/GeneralizedHarmonicBase.hpp
```diff
@@ -289,7 +289,8 @@ struct ObserverTags {
                                      non_tensor_compute_tags>;
 };
 
-template <size_t volume_dim, bool LocalTimeStepping>
+template <size_t volume_dim, bool LocalTimeStepping,
+          bool UseNumericalInitialData>
```
nilsvu, line 278: Note: this template parameter will be deleted again in the next PR that actually factory-creates the initial data in the GH executables.

src/Evolution/Systems/GeneralizedHarmonic/Actions/NumericInitialData.hpp
```diff
@@ -34,6 +29,29 @@
 
 namespace GeneralizedHarmonic {
 
+/*!
+ * \brief Compute initial GH variables from ADM variables
+ *
+ * - The spacetime metric is assembled from the spatial metric, lapse, and
+ *   shift. See `gr::spacetime_metric` for details.
+ * - Phi is computed from numeric derivatives of the spatial metric, lapse, and
```
nilsdeppe: `numeric` -> `numerical`

This is concerning. It should be computed directly as the numerical derivative of `g_{ab}`, otherwise it won't satisfy the 3-con.

src/Evolution/Systems/GeneralizedHarmonic/Actions/NumericInitialData.cpp
```diff
@@ -0,0 +1,51 @@
...
+    const gsl::not_null<tnsr::aa<DataVector, 3>*> spacetime_metric,
+    const gsl::not_null<tnsr::aa<DataVector, 3>*> pi,
+    const gsl::not_null<tnsr::iaa<DataVector, 3>*> phi,
+    const tnsr::ii<DataVector, 3>& spatial_metric,
+    const Scalar<DataVector>& lapse, const tnsr::I<DataVector, 3>& shift,
+    const tnsr::ii<DataVector, 3>& extrinsic_curvature, const Mesh<3>& mesh,
+    const InverseJacobian<DataVector, 3, Frame::ElementLogical,
+                          Frame::Inertial>& inv_jacobian) {
+  // Take numerical derivatives
+  const auto deriv_spatial_metric =
```
nilsdeppe: Change this to compute the derivative of `g_{ab}`. This won't satisfy the 3-con as-is.

src/Evolution/Systems/GrMhd/ValenciaDivClean/Actions/NumericInitialData.hpp
```diff
@@ -23,103 +24,269 @@
...
+      spatial_velocity->get(d) /= get(*lorentz_factor);
+    }
+    // Specific internal energy, pressure, and enthalpy from EOS
+    destructive_resize_components(specific_internal_energy, num_points);
+    destructive_resize_components(pressure, num_points);
+    destructive_resize_components(specific_enthalpy, num_points);
+    for (size_t i = 0; i < num_points; ++i) {
+      double& local_rest_mass_density = get(*rest_mass_density)[i];
+      // Apply the EOS only where the density is above the cutoff, because
+      // the EOS may be ill-defined at zero density. Also, the specific
```
nilsdeppe: It's not just the EOS, the fluid model itself breaks down in the zero-density limit. From stat mech, the fluid model assumes a large (Avogadro's number) of particles, which is not true in vacuum. 

Would it be possible to apply an actual atmosphere set in the input file? I think we currently just do that after the initial data is set, so that's fine, but I don't think you should call the treatment here "atmosphere" because it's fundamentally different from the atmosphere treatment.

nilsdeppe (changes requested):

src/Evolution/Systems/GeneralizedHarmonic/Actions/NumericInitialData.hpp
nilsvu: Changed to the numerical derivative of the spacetime metric.

src/Evolution/Systems/GrMhd/ValenciaDivClean/Actions/NumericInitialData.hpp
nilsvu: Yep, and I don't call this "atmosphere" :) The help string for the `DensityCutoff` is this:

> Where the density is below this cutoff the fluid variables are set to vacuum (zero density, pressure, energy and velocity, unit Lorentz factor and enthalpy). During the evolution, atmosphere treatment will typically kick in and fix the value of the fluid variables in these regions. Therefore, it makes sense to set this density cutoff to the same value as the atmosphere density cutoff.

Sound ok? I reworded the EOS comment a bit and added some docs.

src/Evolution/Systems/GeneralizedHarmonic/Actions/NumericInitialData.cpp
```diff
@@ -28,24 +28,23 @@ void initial_gh_variables_from_adm(
...
+  gr::spacetime_metric(spacetime_metric, lapse, shift, spatial_metric);
+
+  // Compute Phi from numerical derivative of the spacetime metric so it
+  // satisfies the 3-index constraint
+  partial_derivative(phi, *spacetime_metric, mesh, inv_jacobian);
+
+  // Compute Pi by choosing dt_lapse = 0 and dt_shift = 0 (for now).
+  // The mutator `SetPiFromGauge` should be combined with this.
   const auto deriv_spatial_metric =
       partial_derivative(spatial_metric, mesh, inv_jacobian);
```
nilsdeppe: You could, just in case, set the d_spatial_metric to the spatial components of `Phi`, but those numbers should be identical, so I don't care. Up to you!

src/Evolution/Systems/GrMhd/ValenciaDivClean/Actions/NumericInitialData.hpp
nilsdeppe: Yep, looks good!

nilsdeppe (commented):
LGTM, please squash! One optional thought, squash directly regardless of what you decide to do.

nilsdeppe (dismissed):

nilsdeppe:
This now needs a rebase, probably because template parameters

nilsdeppe (approved):

Issue #4761, opened 2023-02-21
Add protocol for initialization mutators


nilsvu:
All those type aliases used by `InitializeItems` are a prime example for a compile-time interface that must be documented properly. It's super confusing to have a struct in some `Actions` namespace/directory that defines all those type aliases, and no way to understand why they're there, what they do, how they're used, etc. We can't duplicate this information in code comments in every initialization mutator (which is what's currently done). A quick fix is to link `InitializeItems` in the docs of every initialize. A more maintainable solution is adding protocol. That probably should have been part of https://github.com/sxs-collaboration/spectre/pull/4486.

nilsvu:
Possible solutions discussed today:

- The type aliases actually implement two different protocols: (1) argument_tags/return_tags/apply define the mutator, and (2) simple_tags/compute_tags/... are collected by InitializeItems and then collected from all actions somewhere in the parallel infrastructure.
- Therefore, add a protocol for both and make initialization mutators conform to both.
- Once the protocols exists, the type aliases can maybe default to empty lists. This won't catch typos in the alias names, but at least the protocol defines the alias names and is linked right from the implementation. Having to write all these type aliases everywhere is excessively verbose. Alternatively, simplify the interface by removing the need for so many type aliases.
- We could also group the aliases in nested structs to make clear they implement two different protocols.

wthrowe:
As I expressed a few times during the meeting, I am unhappy with the idea of adding default values.  We've generally been moving away from defaults.

nilsvu:
I don't like defaults either, but I also find 7 required type aliases in every action excessive (currently only initialization mutators require them and actions use empty lists as default). These are 7 lists of tags that all do something slightly different. I think this is a big mental load on everyone who just wants to write a small action. Do you have another suggestion @wthrowe ?

Edit: As written above, I think with the protocol documenting the interface, empty lists as defaults are fine.

wthrowe:
Could you please be a bit more careful in this discussion with using the word "action"?  You're referring to initialization mutators as actions and also talking about real actions, and it's making you difficult to follow.

Anyway, my objection has nothing to do with documentation.

My problem with defaults for type aliases is there is no way to check that you've defined the correct thing.  There is no "using override" like there is for functions.  If you get it wrong, you may get an error in the same initialization mutator, you may get an error in some distant part of the code, or you may even get no error until someone later changes some unrelated thing.  These also aren't things that tend to get unit tested well.

[...]

PR #4983, opened 2023-05-03
Add some DataBox protocols

kidder:
## Proposed changes

work towards #4761 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/DataBox/Protocols/Mutator.hpp
+++ b/src/DataStructures/DataBox/Protocols/Mutator.hpp
@@ -1,0 +1,37 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <type_traits>
+
+#include "Utilities/ProtocolHelpers.hpp"
+
+namespace db::protocols {
+
+/// \brief A DataBox mutator
+///
+/// A class conforming to this protocol can be used as the template argument for
+/// a call to db::mutate_apply(const gsl::not_null<DataBox<BoxTags>*> box,
+/// Args&&... args).  The conforming class must provide the following:
+///
+/// - `return_tags`: A type list of tags corresponding to mutable items in the
+///   DataBox passed to `db::mutate_apply` that may be modified.
+/// - `argument_tags`: A type list of tags corresponding to items in the DataBox
+///   passed to `db::mutate_apply` that may not be modified.
+/// - `apply`: A static function whose return value is returned by
+///   `db::mutate_apply`, and that takes as arguments:
+///      - A `const gsl::not_null<Tag::type*>` for each `Tag` in `return_tags`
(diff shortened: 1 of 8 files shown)


kidder:
@sxs-collaboration/spectre-core-devs I would appreciate some feedback before I proceed further...

src/DataStructures/DataBox/Protocols/Mutator.hpp
```diff
@@ -0,0 +1,37 @@
...
+///      - A `const db::const_item_type<Tag, BoxTags>` for each `Tag` in
+///        `argument_tags`
+///      - The additional arguments passed to `db::mutate_apply`
+struct Mutator {
+  template <typename ConformingType>
+  struct test {
+    using argument_tags = typename ConformingType::argument_tags;
+    using return_tags = typename ConformingType::return_tags;
+    static_assert(
+        std::is_same_v<void, std::void_t<decltype(&ConformingType::apply)>>);
```
nilsvu, line 34: The apply function might be a template, in which case I think this won't work. Maybe do a quick test and delete this assert if it doesn't work? Also I think `db::mutate_apply` also works with `operator()`. Please add that to the docs.

src/DataStructures/DataBox/Protocols/Mutator.hpp
```diff
@@ -0,0 +1,37 @@
...
+/// - `return_tags`: A type list of tags corresponding to mutable items in the
+///   DataBox passed to `db::mutate_apply` that may be modified.
+/// - `argument_tags`: A type list of tags corresponding to items in the DataBox
+///   passed to `db::mutate_apply` that may not be modified.
+/// - `apply`: A static function whose return value is returned by
+///   `db::mutate_apply`, and that takes as arguments:
+///      - A `const gsl::not_null<Tag::type*>` for each `Tag` in `return_tags`
+///      - A `const db::const_item_type<Tag, BoxTags>` for each `Tag` in
+///        `argument_tags`
+///      - The additional arguments passed to `db::mutate_apply`
```
nilsvu, line 27: Add an example snippet please

nilsvu (changes requested):
I think the mutator protocol looks great, that's exactly what we need more of. It has a clear name and documents a compile-time interface that people would otherwise have a lot of trouble understanding.

Example: Just yesterday Mike and I tried to figure out what the interface for "variable fixers" is. It's not documented (at least we didn't find any docs). Only when we read the source code of the FixVariables action did we realize that a `VariableFixer` just needs to be (1) a mutator and (2) option-creatable. It would be so much clearer if those two were protocols.

My suggestion: Can we limit this PR to the `Mutator` protocol and merge that first because it's very clear, then move on to the others?

PR #4984, opened 2023-05-04
Add second way to specify shape coefs in sphere creator

knelli2:
## Proposed changes

To better test shape control, I want to start with different coefficients other than zero. This allows us to start with the coefficients of a kerr horizon.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -20,18 +22,32 @@
...
+  ERROR("FromH5File NOT IMPLEMENTED YET.");
+}
+
 TimeDependentMapOptions::TimeDependentMapOptions(
-    const double initial_time, const std::array<double, 3> initial_size_values,
-    const size_t initial_l_max)
+    const double initial_time,
+    const std::optional<std::array<double, 3>>& initial_size_values,
+    const size_t initial_l_max,
+    const std::optional<typename ShapeMapInitialValues::optional_type>&
```
nilsvu: `::type::value_type`

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -26,6 +28,54 @@ struct Inertial;
...
+      "horizon of certain Boyer Lindquist radius (the excision radius) "
+      "centered on the origin."};
+
+  double mass{std::numeric_limits<double>::infinity()};
+  std::array<double, 3> spin{std::numeric_limits<double>::infinity(),
+                             std::numeric_limits<double>::infinity(),
+                             std::numeric_limits<double>::infinity()};
+};
+
+struct FromH5File {
```
nilsvu: Delete this for now

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -95,16 +147,35 @@ struct TimeDependentMapOptions {
     using group = ShapeMapOptions;
   };
 
-  using options = tmpl::list<InitialTime, SizeMapInitialValues, ShapeMapLMax>;
+  struct ShapeMapInitialValues {
+    static std::string name() { return "InitialValues"; }
+    // Variants require two templates, but only one is implemented right now.
```
nilsvu: I don't think you need two or more types in a variant. Try with one, and delete the `FromH5File` stub.

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -95,16 +147,35 @@ struct TimeDependentMapOptions {
...
+    // Rather than doing Auto<KSFBL> and then having to edit the input file
+    // syntax again later on, just add in a not implemented option for now. This
+    // will be added in the future.
+    using optional_type =
```
nilsvu: Inline below. You can retrieve the optional type from `std::optional` as `value_type`.

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -50,25 +66,68 @@ TimeDependentMapOptions::create_functions_of_time(
...
+  DataVector size_deriv{1, 0.0};
+  DataVector size_2nd_deriv{1, 0.0};
+
+  if (initial_size_values_.has_value()) {
+    size_func[0] = gsl::at(initial_size_values_.value(), 0);
+    size_deriv[0] = gsl::at(initial_size_values_.value(), 1);
+    size_2nd_deriv[0] = gsl::at(initial_size_values_.value(), 2);
+  }
+
+  // CompressionMap FunctionOfTime
```
nilsvu: We don't use the SpherialCompression map here anymore, so please reword this comment

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -50,25 +66,68 @@ TimeDependentMapOptions::create_functions_of_time(
...
+      const DataVector radial_distortion =
+          1.0 - get(gr::Solutions::kerr_schild_radius_from_boyer_lindquist(
+                    inner_radius, ylm.theta_phi_points(), mass_and_spin.mass,
+                    mass_and_spin.spin)) /
+                    inner_radius;
+      shape_func = ylm.phys_to_spec(radial_distortion);
+      // Transform from SPHEREPACK to actual Ylm for size func
+      size_func_from_shape = shape_func[0] * sqrt(0.5 * M_PI);
+      // Set l=0 for shape map to 0 because size is going to be used
+      shape_func[0] = 0.0;
```
nilsvu, line 82: Can you just delete these lines and construct the Shape map with `size_function_of_time_name = std::nullopt` if no size values were given in the input file (also don't add the size FoT)? I think that would be less code and clearer (no spherepack conversions in this file etc).

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -26,6 +28,54 @@ struct Inertial;
...
+    static constexpr Options::String help = {
+        "The dim'less spin of the Kerr BH."};
+  };
+
+  using options = tmpl::list<Mass, Spin>;
+
+  static constexpr Options::String help = {
+      "Creates Ylm coefficients for the shape map that conform to a Kerr "
+      "horizon of certain Boyer Lindquist radius (the excision radius) "
+      "centered on the origin."};
```
nilsvu: This isn't necessarily the horizon. I suggest this:

> Conform to an ellipsoid of constant Boyer-Lindquist radius in Kerr-Schild coordinates. This Boyer-Lindquist radius is chosen as the value of the 'InnerRadius'. To conform to the outer Kerr horizon, choose an 'InnerRadius' of r_+ = M + sqrt(M^2-a^2).

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -26,6 +28,54 @@ struct Inertial;
...
+
+  double mass{std::numeric_limits<double>::infinity()};
```
nilsvu: `signaling_NaN` or just default-construct.

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -95,16 +147,35 @@ struct TimeDependentMapOptions {
...
+        std::variant<KerrSchildFromBoyerLindquist, FromH5File>;
+    using type = Options::Auto<optional_type, Spherical>;
+    static constexpr Options::String help = {
+        "Initial Ylm coefficients for the shape map. Specify 'Spherical' for "
+        "all coefficients to be initialized to zero."};
+    using group = ShapeMapOptions;
+  };
+
+  using options = tmpl::list<InitialTime, SizeMapInitialValues, ShapeMapLMax,
+                             ShapeMapInitialValues>;
```
nilsvu, line 148: Can we share the logic in this file with the BCO domains? Would be nice, because they need almost the same logic. Doesn't have to happen in this PR.

nilsvu (changes requested):

src/Domain/Creators/SphereTimeDependentMaps.hpp
knelli2: So no, variants don't need two types. But our option parsing code currently does require Alternatives to have two or more. If I remove this restriction, then I'm able to compile.

nilsvu: Actually I think the input file syntax for `T` and `std::variant<T>` is the same, so if the option parsing imposes that restriction you can delete the variant for now.

src/Domain/Creators/SphereTimeDependentMaps.cpp
knelli2: Sometimes I want there to be a size function of time where the coefficient is figured out automatically. This is needed for shape control where I do need a size function of time, but I have no clue what to set the initial value to if I use the KerrSchild coefficients. This will just figure it out for me.

src/Domain/Creators/SphereTimeDependentMaps.hpp
knelli2: We might be able to. Not sure what to do about the maps though because currently this assumes size/shape to distorted and then identity to inertial. The combinations of maps are more complicated in the BCO domain. I'd have to think about it. So yeah, definitely for a follow up PR if at all

knelli2:
Ok squashed and added a couple extra commits for the options/Auto stuff

src/Options/Options.hpp
```diff
@@ -164,7 +164,7 @@ struct create_from_yaml {
 /// \snippet Test_Options.cpp alternatives
 template <typename... AlternativeLists>
 struct Alternatives {
-  static_assert(sizeof...(AlternativeLists) >= 2,
+  static_assert(sizeof...(AlternativeLists) >= 1,
```
nilsvu, line 167: Update the error message. Also, @wthrowe are you on board with this change? The context is that a `std::variant` is well-defined for a single type. We could also omit the `std::variant` in the option type until we implement the second type.

src/Domain/Creators/SphereTimeDependentMaps.cpp
nilsvu: ok

nilsvu (dismissed):

nilsvu (approved):

src/Options/Options.hpp
wthrowe: Yes, he discussed it with me.  I don't think there are any technical problems with the change.  I suspect the requirement was there because I just wasn't thinking of TMP-generated option lists, or possibly because it will look weird in the help output.  The latter is probably still true, but we can fix that later if we think it's worth the effort.  (There are worse issues with help text.)

PR #4987, opened 2023-05-04
Allow control sys Info and state to be copyable

knelli2:
## Proposed changes

Also handle a NaN in the time deriv of a strahlkorper, and make the `ZeroCrossingPredictor` copyable. This is all needed for size control.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size/Info.cpp
+++ b/src/ControlSystem/ControlErrors/Size/Info.cpp
@@ -10,6 +10,28 @@
 #include "Utilities/Serialization/CharmPupable.hpp"
 
 namespace control_system::size {
+Info::Info(std::unique_ptr<State> in_state, double in_damping_time,
+           double in_target_char_speed, double in_target_drift_velocity,
+           double in_suggested_time_scale,
+           bool in_discontinuous_change_has_occurred)
+    : state(std::move(in_state)),
+      damping_time(in_damping_time),
+      target_char_speed(in_target_char_speed),
+      target_drift_velocity(in_target_drift_velocity),
+      suggested_time_scale(in_suggested_time_scale),
+      discontinuous_change_has_occurred(in_discontinuous_change_has_occurred) {}
+
+// NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init,-warnings-as-errors)
+Info::Info(const Info& rhs) {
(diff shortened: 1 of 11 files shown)


src/ControlSystem/ControlErrors/Size/Info.cpp
```diff
@@ -10,6 +10,42 @@
...
+  set_all_but_state(rhs);
+  state = rhs.state->get_clone();
+  return *this;
+}
+
+// NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init,-warnings-as-errors)
+Info::Info(Info&& rhs) {
+  set_all_but_state(rhs);
+  state = std::move(rhs.state);
+  rhs.state = nullptr;
```
wthrowe: How is this different from the default implementation?

src/ControlSystem/ControlErrors/Size/Info.hpp
```diff
@@ -16,6 +16,17 @@ namespace control_system::size {
 
 /// Holds information that is saved between calls of SizeControl.
 struct Info {
+  Info() = default;
+  Info(const Info& rhs);
+  Info& operator=(const Info& rhs);
+  Info(Info&& rhs);
+  Info& operator=(Info&& rhs);
+
+  Info(const std::unique_ptr<State>& in_state, double in_damping_time,
```
wthrowe: Why take the pointer by reference instead of value?  That would match the current interface.

wthrowe (changes requested):

src/ControlSystem/ControlErrors/Size/Info.cpp
knelli2: I'm not sure. I know I needed the copy constructor/assignment because of the `unique_ptr`, so I just did everything. I guess I can remove the `move` constructor and assignment because that's trivial even with a `unique_ptr`. Does that sound correct?

wthrowe: Remove and replace with `= default`, yes.  Defaulting things is less error-prone, and in some cases (although I think not this one) can allow some compiler optimizations.

knelli2:
Squashed and rebased

wthrowe (approved):

Issue #4694, opened 2023-02-03
Add baryon_mass function to EoS


nilsdeppe:
# Feature request:

Different EoS assume different baryon masses. when convert from number density to rest/baryon mass density this conversion factor is important. We need to add a function to all EoS that returns the assume baryon mass in different units. E.g. we might want `baryon_mass_cgs()`, `baryon_mass_geometric()`, and `baryon_mass_mev()`. We can extend the `Units.hpp` file being added in #4687 to make the conversions easier so that someone only needs to implement one of the 3 functions with a magic number.

PR #4992, opened 2023-05-08
Add baryon mass to eos

isaaclegred:
## Proposed changes

This PR adds a `baryon_mass` function to each EoS, addressing  #4694.  It also factors certain code into `Hydro/Units.hpp` to facilitate standardizing certain definitions of constants/units.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

The upshot of the technical parts of this PR is that  changing the value of G (even within the reported uncertainty) changes all microscopic-hydro things to factors of order 10^-5 in geometric units, so we want to standardize this value and track changes between versions to make comparisons to previous code versions (and other codes) as easy as possible.  On top of that we'd prefer to have standardized values for constants in a standard location, again to guarantee everything is self consistent and in rare cases to check consistency to other codes when G based conversions are the limiting factor.

### DIFF
--- a/src/Executables/ExportEquationOfStateForRotNS/ExportEquationOfStateForRotNS.cpp
+++ b/src/Executables/ExportEquationOfStateForRotNS/ExportEquationOfStateForRotNS.cpp
@@ -41,7 +41,9 @@ void dump_barotropic_eos(
   // might be different from the baryon mass that the EoS uses.
   //
   // https://github.com/sxs-collaboration/spectre/issues/4694
-  const double baryon_mass_of_rotns_cgs = 1.659e-24;
+  const double baryon_mass_of_rotns_cgs =
+      hydro::units::geometric::default_baryon_mass *
+      hydro::units::cgs::mass_unit;
   const double log10_lower_bound_number_density_cgs =
       log10(lower_bound_rest_mass_density_cgs / baryon_mass_of_rotns_cgs);
   const double log10_upper_bound_number_density_cgs =
@@ -66,15 +68,13 @@ void dump_barotropic_eos(
         pow(10.0, log10_lower_bound_number_density_cgs +
                       static_cast<double>(log10_number_density_index) *
                           delta_log_number_density_cgs);
-    const double rest_mass_density_cgs =
-        number_density_cgs * baryon_mass_of_rotns_cgs;
 
     // Note: we will want to add the baryon mass to our EOS interface.
     //
     // https://github.com/sxs-collaboration/spectre/issues/4694
-    const double baryon_mass_of_eos_cgs = baryon_mass_of_rotns_cgs;
-    const Scalar<double> rest_mass_density_geometric{rest_mass_density_cgs /
-                                                     baryon_mass_of_eos_cgs};
+    const Scalar<double> rest_mass_density_geometric{
+        number_density_cgs * cube(hydro::units::cgs::length_unit) *
+        eos.baryon_mass()};
     const Scalar<double> pressure_geometric =
         eos.pressure_from_density(rest_mass_density_geometric);
     const Scalar<double> specific_internal_energy_geometric =
@@ -83,11 +83,11 @@ void dump_barotropic_eos(
         get(rest_mass_density_geometric) *
         (1.0 + get(specific_internal_energy_geometric))};
 
-    // Note: the energy density is divided by c^2, so the rest-mass part is rho
-    // c^2
+    // Note: the energy density is divided by c^2
     const double total_energy_density_cgs =
         get(total_energy_density_geometric) *
-        hydro::units::cgs::rest_mass_density_unit;
+        hydro::units::cgs::rest_mass_density_unit *
+        square(hydro::units::cgs::speed_of_light);
 
     // should be dyne cm^(-3)
     const double pressure_cgs =
--- a/src/PointwiseFunctions/Hydro/Units.hpp
+++ b/src/PointwiseFunctions/Hydro/Units.hpp
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include <cmath>
+
 namespace hydro {
 /*!
  * \brief Functions, constants, and classes for converting between different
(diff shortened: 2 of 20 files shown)


src/PointwiseFunctions/Hydro/Units.hpp
```diff
@@ -78,5 +80,36 @@ constexpr double pressure_unit =
...
+/// estimates (see, e.g.
+/// https://journals.aps.org/prc/abstract/10.1103/PhysRevC.102.044321)
+constexpr double default_saturation_rest_mass_density = 4.5e-4;
+/// The default baryon mass in geometric units.
+/// Accepted value of atomic mass unit as expressed in
+/// solar masses
+/// Limited by precision of knowledge of solar mass.
+constexpr double default_baryon_mass = 8.34819e-58;
+}  // namespace nuclear
 }  // namespace units
```
teukolsky, line 179: It's not actually the solar mass that has limited precision: G M_sun is known to something like 10 digits. It's G that has limited precision, which makes the conversion to kg problematic. The only reason this might matter is when it comes to comparing different codes, or different versions of spectre where someone "updates" the units with improved values. The best way to handle this would be to define G as a constant in one place and use it explicitly where needed for conversions. Then updates would be clear. I'm happy not to do this in this PR, but then perhaps you should file this comment as an Issue.

isaaclegred: I agree; my main point with this comment is to point out that the baryon mass is actually known to much higher precision than is represented here.  I'd be happy to give numbers to highest precision in the units they're known best, and then report the uncertainty in G as the limiting factor in the conversion.

teukolsky: Maybe I'm being overly cautious, but this is an issue that has surfaced in other applications. The accepted value of G is likely to change by more than the reported uncertainty. The reason is that there are 2 sets of experiments, each with a different mean that lies several sigma away from the other set. When this finally gets sorted out, the accepted value could undergo a several sigma jump. So the best thing to do is to actually record the value of G used in one place in the code, and then explicitly use that value in the rare cases where we need M_sun in kg. Note that M_sun in kg is only needed with microphysics. All GR applications should use G M_sun.

isaaclegred:
Hi @nilsdeppe, @ermost, @pajkosmi, @ffoucart, @yoonso0-0, (feel free to ping anyone else working on hydro, maybe @kidder?).  In the process of explicitly adding baryon masses to EoSs,  I've moved some things around in an attempt to preempt getting lots of inconsistently defined constants appearing in different places in the code.  Please feel free to leave thoughts,  I think the only place things have changed in a way that could break things is in the tabulated eos, where some constants have changed to enforce consistency (this is currently breaking a test, but there are several ways to fix this including reverting standardizing the constants here if that's the right thing to do).

ermost:
Thanks for looking into this @isaaclegred! I think moving all of those constants into a single place is the right thing to do. If this introduces a minor rescaling of the conversion in the tables, that should be ok, as long as we remain consistent throughout the code. (I.e. we should just fix the test case)

Regarding the use of  namespaces for unit conversion. Do people have thoughts on using a templated class with constexpr member vars? (e.g.,  https://github.com/wokast/PyCactus/blob/master/PostCactus/postcactus/unitconv.py)

isaaclegred:
Thanks for input @ermost! Yeah, I think a templated class with `constexpr` values sounds fine to me; I can try to implement this.  Are you thinking something like `units<CGS>::length_unit` where `CGS` is an element of an enum (along with, for example, `SI`, and perhaps the geometric values)?

nilsdeppe:
This is good with me. Doing it via classes is also fine, but if we want something fancy we should check out https://github.com/aurora-opensource/au for an extremely good units library. This grew out of some very early discussions I had with folks at Uber ATG before it was bought by Aurora, so there's some bias, but I think they did a good job.

Probably in the case of something fancy, we should merge this first and then have a bigger discussion of the fancy thing.

ermost:
This looks really nice! Thanks for sharing @nilsdeppe. I will have a look :)

I'm completely on board with getting this merged asap.

I guess what I had in mind with templated inheritance is to make sure that all unit systems remain consistent. E.g., cgs has density and pressure units, while nuclear units don't have that defined yet. (In terms of variables having the same names)

Would a good solution for now just be to a) either ignore it, or b) to just manually define those things as well for the MeV units for the moment?

nilsdeppe:
You're welcome! :) 

I think going with b) sounds good so that the numbers are in the code. It's always tedious to have to look up the numbers and try to figure it out, so I think the sooner we have the conversion factors in, even if it's not automated, the better! @isaaclegred do you think it would be reasonable for you to add the MeV values to this PR? I'm happy with either the class template approach or a separate namespace. With class template maybe an enum for the unit system would be good? @ermost is that what you had in mind?

ermost:
That all sounds good to me! Either namespaces or classes with enum would work well.
What I had originally in mind was something in between, to split off derived units from base units, to keep things like pressure conversion uniform, e.g., 

```
template<typename BaseUnits>
class DerivedUnits{

  public:

  static constexpr double pressure_unit = BaseUnits::mass_unit/(BaseUnits::length_unit*BaseUnits::time_unit*BaseUnits::time_unit);
  static constexpr double rest_mass_density_unit =  BaseUnits::mass_unit / (BaseUnits::length_unit * BaseUnits::length_unit * BaseUnits::length_unit);
};

class Cgs : public DerivedUnits<Cgs>{

  public:
    static constexpr double G_Newton = 6.67430e-8;
    static constexpr double G_Newton_times_m_sun = 1.32712440042e26;
    static constexpr double speed_of_light = 29979245800.0;

    static constexpr double length_unit = G_Newton_times_m_sun / speed_of_light/speed_of_light;
    static constexpr double mass_unit = G_Newton_times_m_sun / G_Newton;
    static constexpr double time_unit = length_unit / speed_of_light;
};
```

isaaclegred:
Hi @ermost  and @nilsdeppe, thanks for your comments.   Sorry for the delayed response;  I've been traveling.  I have added the conversion factors for MeV units (i.e. `length_unit`(for fm ), `mass_unit` (for MeV/^2), `time_unit` (currently the light crossing time of a fm)), and shuffled things around a bit to clearly delineate nuclear units from the quantities given in geometric units.    

I think we may want to think about the possibility of defining constants and implementing unit conversion separately.  Many unit systems may depend on a small number of shared constants, defined in a particular system, for example, it doesn't make sense to define the G_cgs,  G_si, and G_nuclear separately, because then we have to remember to update all the unit systems if we chose to use a different value of G (and they are all related to each other via exact conversion factors).  Either way, I agree we should think about a units module.

src/Executables/ExportEquationOfStateForRotNS/ExportEquationOfStateForRotNS.cpp
```diff
@@ -41,7 +41,10 @@ void dump_barotropic_eos(
   // might be different from the baryon mass that the EoS uses.
   //
   // https://github.com/sxs-collaboration/spectre/issues/4694
-  const double baryon_mass_of_rotns_cgs = 1.659e-24;
+  const double baryon_mass_of_rotns_cgs =
+      hydro::units::geometric::default_baryon_mass *
+      hydro::units::cgs::mass_unit;
+  ;
```
nilsdeppe: remove `;`?

nilsdeppe (commented):
This is good with me! I'm happy for you to squash and rebase so we can merge.

@ermost would you mind taking a look again?

ermost:
Sorry for the delay in getting back to this. Thank you for adding the nuclear units. 
The current solution looks really good! I particularly like the extensive documentation in Units.hpp.

From my side this is good to go. Agree with previous discussion to think about a units module in a separate PR; no need to hold up the baryon mass functionality.

isaaclegred:
Ok I rebased and squashed everything.  Just pinging @pajkosmi again because I think this may touch some code that he is currently developing, I don't think it should lead to any major problems though.

isaaclegred:
Hi @nilsdeppe and @ermost, I think if there are no further comments from your side (and if this passes CI) then this should be ready to go (just rebased on develop).

nilsdeppe (approved):

ermost:
sounds great! thanks for all your work on this @isaaclegred!

PR #4994, opened 2023-05-09
Account for inverse roundoff in StrahlkorperInDiffFrame

knelli2:
## Proposed changes

Hit this in size control as I was testing #4967. Some coords were NaN because this function couldn't find the inverse because the inverse was outside [-1,1] by roundoff. This is handled by using `block_logical_coordinates` instead which already accounts for this.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
+++ b/src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
@@ -14,6 +14,7 @@
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "DataStructures/Variables.hpp"
 #include "Domain/Block.hpp"
+#include "Domain/BlockLogicalCoordinates.hpp"
 #include "Domain/CoordinateMaps/CoordinateMap.hpp"
 #include "Domain/Domain.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
@@ -27,7 +28,7 @@
 
 namespace {
 // Transforms cartesian coordinates of a strahlkorper
-// from one frame to another, by looping over the blocks and
+// from one frame to another, by calling block_logical_coordinates and
 // calling the correct map functions.
 template <typename SrcFrame, typename DestFrame>
 void coords_to_different_frame(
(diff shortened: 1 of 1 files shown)


src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
```diff
@@ -56,14 +57,27 @@ void coords_to_different_frame(
...
-          get<1>(x_logical.value()) <= 1.0 and
-          get<1>(x_logical.value()) >= -1.0 and
-          get<2>(x_logical.value()) <= 1.0 and
-          get<2>(x_logical.value()) >= -1.0) {
+      auto x_logical = logical_to_grid_map.inverse(x_src);
+      if (not x_logical.has_value()) {
+        continue;
+      }
+
+      // Inverses may by outside [-1,1] by roundoff. Fix that. This is handles
```
markscheel: [very minor] handles->handled

markscheel (commented):
Very minor comment change; looks good to me, please squash.

knelli2:
@markscheel Done!

src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
```diff
@@ -56,14 +57,27 @@ void coords_to_different_frame(
...
       const auto& logical_to_grid_map = block.moving_mesh_logical_to_grid_map();
-      const auto x_logical = logical_to_grid_map.inverse(x_src);
-      // x_logical might be an empty std::optional.
-      if (x_logical.has_value() and get<0>(x_logical.value()) <= 1.0 and
-          get<0>(x_logical.value()) >= -1.0 and
-          get<1>(x_logical.value()) <= 1.0 and
-          get<1>(x_logical.value()) >= -1.0 and
-          get<2>(x_logical.value()) <= 1.0 and
-          get<2>(x_logical.value()) >= -1.0) {
+      auto x_logical = logical_to_grid_map.inverse(x_src);
```
nilsvu: Can you call `block_logical_coordinates(domain, src_cartesian_coords, time, functions_of_time)` at the beginning of the function and use the result for the computation? That's what's being done in `StrahlkorperCoordsInDifferentFrame.cpp:63`.

nilsvu: What's actually the difference between the two?

knelli2: No because have no notion of elements here. This is just a bunch of random points on a sphere. block_logical_coords returns vectors of elements, but we need to check point-by-point which block the point is in so we get the correct maps

nilsvu: No, block_logical_coordinates returns a vector of size num_points, and gives you the block ID and the logical coords within that block for each input coordinate.

nilsvu (approved):

src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
```diff
@@ -43,50 +45,44 @@ void coords_to_different_frame(
   // need to add 'if constexpr's for that case.
   static_assert(std::is_same_v<SrcFrame, ::Frame::Grid>,
                 "Source frame must currently be Grid frame");
-  // We wish to map src_cartesian_coords to the destination frame.
-  // Each Block will have a different map, so the mapping must be done
-  // Block by Block.
-  // Note that if a point is on the boundary of two or more
-  // Blocks, it might get filled twice, but that's ok.
+  const auto block_logical_coords = block_logical_coordinates(
```
nilsvu, line 47: I think the function below needs this treatment as well (line 192).

nilsvu (changes requested):

nilsvu (approved):
Looks good! @markscheel please take a look at this as well.

markscheel (approved):

PR #4996, opened 2023-05-12
Add scheduling CLI

### LINKED ISSUE #2987: Behavior of submission scripts is not standardized

nilsvu:
## Proposed changes

Tool for scheduling runs, submitting jobs, setting up directories for segments and checkpoints, restarting from checkpoints, resubmitting jobs. Also allows to schedule ranges of runs, e.g. for convergence tests.

Try it like this:

```sh
spectre schedule INPUT_FILE -o ./Run
```

To run a convergence test put something like `InitialGridPoints: {{ num_points }}` in your input file and try this:

```sh
spectre schedule INPUT_FILE -p num_points=6...8 -o './Run_p{{ num_points - 1 }}'
```

Closes #2987.

### Upgrade instructions

- Install `Jinja2` in your Python environment (it's probably already installed because other packages need it).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Python/CMakeLists.txt
+++ b/support/Python/CMakeLists.txt
@@ -9,6 +9,7 @@ spectre_python_add_module(
   PYTHON_FILES
   DirectoryStructure.py
   Machines.py
+  Schedule.py
 )
 
 if(MACHINE)
@@ -19,7 +20,7 @@ if(MACHINE)
     )
   configure_file(
     ${CMAKE_SOURCE_DIR}/support/SubmitScripts/${MACHINE}.sh
-    ${SPECTRE_PYTHON_PREFIX}/support/Submit.sh
+    ${SPECTRE_PYTHON_PREFIX}/support/SubmitTemplate.sh
     @ONLY
     )
 endif()
--- a/support/Python/CMakeLists.txt
+++ b/support/Python/CMakeLists.txt
@@ -9,6 +9,7 @@ spectre_python_add_module(
   PYTHON_FILES
   DirectoryStructure.py
   Machines.py
+  Resubmit.py
   Schedule.py
 )
 
--- a/support/Python/CMakeLists.txt
+++ b/support/Python/CMakeLists.txt
@@ -20,6 +20,11 @@ if(MACHINE)
     ${SPECTRE_PYTHON_PREFIX}/support/Machine.yaml
     )
   configure_file(
+    ${CMAKE_SOURCE_DIR}/support/SubmitScripts/SubmitTemplateBase.sh
+    ${SPECTRE_PYTHON_PREFIX}/support/SubmitTemplateBase.sh
+    @ONLY
+    )
+  configure_file(
     ${CMAKE_SOURCE_DIR}/support/SubmitScripts/${MACHINE}.sh
     ${SPECTRE_PYTHON_PREFIX}/support/SubmitTemplate.sh
     @ONLY
--- a/support/Python/Schedule.py
+++ b/support/Python/Schedule.py
@@ -1,0 +1,943 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+import re
+import shutil
+import subprocess
+from pathlib import Path
+from typing import Optional, Sequence, Union
+
+import click
+import jinja2
+import jinja2.meta
+import yaml
+
+from spectre.support.DirectoryStructure import (
+    Checkpoint,
+    Segment,
+    list_checkpoints,
+    list_segments,
+)
+from spectre.tools.ValidateInputFile import validate_input_file
+from spectre.Visualization.ReadInputFile import find_phase_change
+
+logger = logging.getLogger(__name__)
+
+# CMake configures the submit script templates for the current machine to this
+# path
+default_submit_script_template = Path(__file__).parent / "SubmitTemplate.sh"
+
+
+def _resolve_executable(executable: Union[str, Path]) -> Path:
+    """Look up the 'executable' in the PATH
+
+    Raises 'ValueError' if the executable is not found.
+    """
+    logger.debug(f"Resolving executable: {executable}")
+    which_exec = shutil.which(executable)
+    if not which_exec:
+        raise ValueError(f"Executable not found in PATH: {executable}")
+    logger.debug(f"Found executable in PATH: {which_exec}")
+    return Path(which_exec).resolve()
+
+
+def _write_or_overwrite(
+    text: str, path: Path, error_hint: Optional[str] = None, force: bool = False
+):
+    """Write the 'text' to a file at 'path'
+
+    Raise an 'OSError' if the file already exists, unless called with 'force'
+    or if the existing file content is identical to the 'text'.
+    The 'hint' is appended to the error message.
+    """
+    if path.exists():
+        if path.read_text() == text:
+            return
+        if not force:
+            raise OSError(
+                f"File already exists at '{path}'. Retry with "
+                "'force' ('--force' / '-f') to overwrite."
+                + (("\n" + error_hint) if error_hint else "")
+            )
+    logger.debug(f"Write file: {path}")
+    path.write_text(text)
+
+
+def _copy_to_dir(src_file: Path, dest_dir: Path, force: bool = False) -> Path:
+    """Copy the 'src_file' to the 'dest_dir', keeping the file name the same
+
+    Returns the path to the new file.
+    """
+    assert src_file.is_file()
+    assert dest_dir.is_dir()
+    if src_file.resolve().parent == dest_dir.resolve():
+        return src_file
+    dest = (dest_dir / src_file.name).resolve()
+    if dest.exists() and not force:
+        raise OSError(
+            f"File already exists at '{dest}'. Retry with "
+            "'force' ('--force' / '-f') to overwrite."
+        )
+    logging.debug(f"Copy file: {src_file} -> {dest}")
+    shutil.copy(src_file, dest)
+    return dest
+
+
+def _copy_submit_script_template(
+    submit_script_template: Path,
+    dest_dir: Path,
+    template_env: jinja2.Environment,
+    force: bool = False,
+) -> Path:
+    dest = _copy_to_dir(submit_script_template, dest_dir, force=force)
+    # Also copy all referenced templates (the "SubmitBase.sh" parent template)
+    syntax_tree = template_env.parse(dest.read_text())
+    for referenced_template in jinja2.meta.find_referenced_templates(
+        syntax_tree
+    ):
+        referenced_template_src = (
+            submit_script_template.resolve().parent / referenced_template
+        )
+        _copy_to_dir(referenced_template_src, dest_dir, force=force)
+    return dest
+
+
+# Write `pathlib.Path` objects to YAML as plain strings
+def _path_representer(dumper: yaml.Dumper, path: Path) -> yaml.nodes.ScalarNode:
+    return dumper.represent_scalar("tag:yaml.org,2002:str", str(path))
+
+
+def schedule(
+    input_file_template: Union[str, Path],
+    scheduler: Optional[Union[str, Sequence]],
+    executable: Optional[Union[str, Path]] = None,
+    run_dir: Optional[Union[str, Path]] = None,
+    segments_dir: Optional[Union[str, Path]] = None,
+    copy_executable: Optional[bool] = None,
+    job_name: Optional[str] = None,
+    submit_script_template: Optional[Union[str, Path]] = None,
+    from_checkpoint: Optional[Union[str, Path, Checkpoint]] = None,
+    input_file_name: Optional[str] = None,
+    submit_script_name: str = "Submit.sh",
+    out_file_name: str = "spectre.out",
+    context_file_name: str = "SubmitContext.yaml",
+    submit: Optional[bool] = None,
+    clean_output: bool = False,
+    force: bool = False,
+    **kwargs,
+) -> Optional[subprocess.CompletedProcess]:
+    """Schedule executable runs with an input file
+
+    Configures the input file, submit script, etc. to the 'run_dir', and then
+    invokes the 'scheduler' to submit the run (typically "sbatch"). You can also
+    bypass the scheduler and run the executable directly by setting the
+    'scheduler' to 'None'.
+
+    # Selecting the executable
+
+    Specify either a path to the executable, or just its name if it's in the
+    'PATH'. If unspecified, the 'Executable' listed in the input file metadata
+    is used.
+
+    By default, the executable and submit scripts will be copied to the segments
+    directory to support resubmissions (see below). See the 'copy_executable'
+    argument docs for details on controlling this behavior.
+
+    # Segments and run directories
+
+    You can set either the 'run_dir' or the 'segments_dir' to specify where the
+    executable will run (but not both). If you specify a 'run_dir', the
+    executable will run in it directly. If you specify a 'segments_dir', a new
+    segment will be created and used as the 'run_dir'. Segments are named with
+    incrementing integers and continue the run from the previous segment. For
+    example, the following is a typical 'segments_dir':
+
+    \b
+    ```sh
+    # Copy of the executable
+    MyExecutable
+    # Copy of the submit script template (base and machine-specific)
+    SubmitTemplateBase.sh
+    SubmitTemplate.sh
+    # One segment per day
+    Segment_0000/
+        InputFile.yaml
+        Submit.sh
+        Output.h5
+        # Occasional checkpoints, and a checkpoint before termination
+        Checkpoints/
+            Checkpoint_0000/
+            Checkpoint_0001/...
+    # Next segment continues from last checkpoint of previous segment
+    Segment_0001/...
+    ```
+
+    You can omit the 'run_dir' if the current working directory already contains
+    the input file.
+
+    # Placeholders
+
+    The input file, submit script, 'run_dir', 'segments_dir', and 'job_name' can
+    have placeholders like '{{ num_nodes }}'. They must conform to the
+    [Jinja template format](https://jinja.palletsprojects.com/en/3.0.x/templates/).
+    The placeholders are resolved in the following stages.
+    The following parameters are available as placeholders:
+
+    1. 'job_name' (if specified):
+
+    \b
+        - All arguments to this function, including all additional '**kwargs'.
+          For example, the additional '**kwargs' can include parameters
+          controlling resolution in the input file.
+        - 'executable_name': Just the name of the executable (basename of the
+          'executable' path).
+
+    2. 'run_dir' and 'segments_dir':
+
+    \b
+        - All parameters from the previous stage.
+        - 'job_name': Either the resolved 'job_name' from the previous stage, or
+          the 'executable_name' if unspecified.
+
+    3. Input file & submit script:
+
+    \b
+        - All parameters from the previous stages.
+        - 'run_dir': Absolute path to the 'run_dir'.
+        - 'segments_dir': Absolute path to the 'segments_dir', or 'None' if no
+          segments directory is available.
+        - 'input_file': Relative path to the configured input file (in the
+          'run_dir').
+        - 'out_file': Absolute path to the log file (in the 'run_dir').
+        - 'spectre_cli': Absolute path to the SpECTRE CLI.
+        - Typical additional parameters used in submit scripts are 'queue' and
+          'time_limit'.
+
+    The parameters used to render the submit script are stored in a context file
+    (named 'context_file_name') in the 'run_dir' to support resubmissions. The
+    context file is used by 'spectre.support.resubmit' to schedule the next
+    segment using the same parameters.
+
+    # Scheduling multiple runs
+
+    You can pass ranges of parameters to the '**kwargs' of this function to
+    schedule multiple runs using the same input file template. For example, you
+    can do an h-convergence test by using a placeholder for the refinement level
+    in your input file:
+
+    \b
+    ```yaml
+    # In the domain creator:
+    InitialRefinement: {{ lev }}
+    ```
+
+    When a parameter in '**kwargs' is an iterable, the 'schedule' function will
+    recurse for every element in the iterable. For example, you can schedule
+    multiple runs for a convergence test like this:
+
+    \b
+    ```py
+    schedule(
+        run_dir="Lev{{ lev }}",
+        # ...
+        lev=range(1, 3))
+    ```
+
+    \f
+    Arguments:
+      input_file_template: Path to an input file. It will be copied to the
+        'run_dir'. It can be a Jinja template (see above).
+      scheduler: 'None' to run the executable directly, or a scheduler such as
+        "sbatch" to submit the run to a queue.
+      executable: Path or name of the executable to run. If unspecified, use the
+        'Executable' set in the input file metadata.
+      run_dir: The directory to which input file, submit script, etc. are
+        copied, and relative to which the executable will run.
+        Can be a Jinja template (see above).
+      segments_dir: The directory in which a new segment is created as the
+        'run_dir'. Mutually exclusive with 'run_dir'.
+        Can be a Jinja template (see above).
+      copy_executable: Copy the executable to the run or segments directory.
+        By default (when set to 'None'):
+          - If '--run-dir' / '-o' is set, don't copy.
+          - If '--segments-dir' / '-O' is set, copy to segments directory to
+            support resubmission.
+        When set to 'True':
+          - If '--run-dir' / '-o' is set, copy to the run directory.
+          - If '--segments-dir' / '-O' is set, copy to segments directory to
+            support resubmission. Still don't copy to individual segments.
+        When set to 'False': Never copy.
+      job_name: Optional. A string describing the job.
+        Can be a Jinja template (see above). (Default: executable name)
+      submit_script_template: Optional only when 'scheduler' is 'None'. Path to
+        a submit script. It will be copied to the 'run_dir'.
+        Can be a Jinja template (see above).
+      from_checkpoint: Optional. Path to a checkpoint directory.
+      input_file_name: Optional. Filename of the input file in the 'run_dir'.
+        (Default: basename of the 'input_file_template')
+      submit_script_name: Optional. Filename of the submit script. (Default:
+        "Submit.sh")
+      out_file_name: Optional. Name of the log file. (Default:
+        "spectre.out")
+      context_file_name: Optional. Name of the file that stores the context
+        for resubmissions in the `run_dir`. Used by `spectre.support.resubmit`.
+        (Default: "SubmitContext.yaml")
+      submit: Optional. If 'True', automatically submit jobs using the
+        'scheduler'. If 'False', skip the job submission. If 'None', prompt for
+        confirmation before submitting.
+      clean_output: Optional. When 'True', use
+        'spectre.tools.CleanOutput.clean_output' to clean up existing output
+        files in the 'run_dir' before scheduling the run. (Default: 'False')
+      force: Optional. When 'True', overwrite input file and submit script
+        in the 'run_dir' instead of raising an error when they already exist.
+
+    Returns: The 'subprocess.CompletedProcess' representing either the process
+      that scheduled the run, or the process that ran the executable if
+      'scheduler' is 'None'. Returns 'None' if no or multiple runs were
+      scheduled.
+    """
+    # Defaults
+    input_file_template = Path(input_file_template)
+    if not input_file_name:
+        input_file_name = input_file_template.resolve().name
+    if isinstance(from_checkpoint, Checkpoint):
+        from_checkpoint = from_checkpoint.path
+    if from_checkpoint:
+        from_checkpoint = Path(from_checkpoint).resolve()
+
+    # Snapshot function arguments for template substitutions
+    all_args = locals().copy()
+    del all_args["kwargs"]
+
+    # Recursively schedule ranges of runs
+    for key, value in kwargs.items():
+        # Check if the parameter is an iterable
+        if isinstance(value, str):
+            # Strings are iterable, but we don't want to treat them as such
+            continue
+        try:
+            iter(value)
+        except TypeError:
+            continue
+        # Recurse for each value of the iterable
+        for value_i in value:
+            logger.info(f"Recurse for {key}={value_i}")
+            kwargs_i = kwargs.copy()
+            kwargs_i.update({key: value_i})
+            try:
+                schedule(**all_args, **kwargs_i)
+            except:
+                logger.exception(f"Recursion for {key}={value_i} failed.")
+        return
+
+    # Set up template environment with basic configuration
+    template_env = jinja2.Environment(
+        undefined=jinja2.StrictUndefined,
+        trim_blocks=True,
+        keep_trailing_newline=True,
+    )
+
+    # Start collecting parameters for template substitutions. We filter 'None'
+    # values so they properly trigger undefined-variable errors when templates
+    # need them.
+    context = {
+        key: value
+        for key, value in dict(**all_args, **kwargs).items()
+        if value is not None
+    }
+
+    # Read input file template
+    input_file_contents = input_file_template.read_text()
+
+    # Resolve executable
+    if not executable:
+        metadata = next(yaml.safe_load_all(input_file_contents))
+        try:
+            executable = metadata["Executable"]
+        except KeyError as err:
+            raise ValueError(
+                "Specify an 'executable' ('--executable' / '-e') "
+                "or list one in the input file metadata "
+                "as 'Executable:'."
+            ) from err
+    executable = _resolve_executable(executable)
+    # Only set executable name because the path may change if we copy it later
+    context.update(executable_name=executable.name)
+
+    # Resolve number of cores, nodes, etc.
+    num_procs = kwargs.get("num_procs")
+    num_nodes = kwargs.get("num_nodes")
+    if num_procs:
+        assert (
+            num_nodes is None or num_nodes == 1
+        ), "Specify either 'num_procs' or 'num_nodes', not both."
+        num_nodes = 1
+    # Set the context variables only if defined, so they don't print as "None"
+    # https://jinja.palletsprojects.com/en/3.0.x/templates/#jinja-filters.default
+    if num_procs:
+        context.update(num_procs=num_procs)
+    if num_nodes:
+        context.update(num_nodes=num_nodes)
+
+    # Resolve job_name
+    if job_name:
+        job_name = template_env.from_string(job_name).render(context).strip()
+    else:
+        job_name = executable.name
+    context.update(job_name=job_name)
+
+    # Resolve run_dir and segments_dir
+    if run_dir and segments_dir:
+        raise ValueError(
+            "Specify either 'run_dir' ('--run-dir' / '-o') "
+            "or 'segments_dir' ('--segments-dir' / '-O'), not both."
+        )
+    elif not run_dir and not segments_dir:
+        # Neither run_dir nor segments_dir were specified. Set run_dir to the
+        # current working directory.
+        if input_file_template.resolve().parent == Path.cwd():
+            run_dir = Path.cwd()
+        else:
+            raise ValueError(
+                "Specify a 'run_dir' ('--run-dir' / '-o') "
+                "or a 'segments_dir' ('--segments-dir' / '-O'), "
+                "or place the input file into the current directory."
+            )
+    # At this point either run_dir or segments_dir are set, but not both.
+    if run_dir:
+        # Run directly in the run_dir. If the run_dir looks like a segment, set
+        # the segments_dir so resubmitting works.
+        run_dir = Path(
+            template_env.from_string(str(run_dir)).render(context).strip()
+        )
+        if Segment.match(run_dir):
+            segments_dir = run_dir.resolve().parent
+            all_segments = list_segments(segments_dir)
+    else:
+        # Run in next segment in the segments_dir
+        segments_dir = Path(
+            template_env.from_string(str(segments_dir)).render(context).strip()
+        )
+        all_segments = list_segments(segments_dir)
+        if all_segments:
+            run_dir = all_segments[-1].next.path
+        else:
+            run_dir = Segment.first(segments_dir).path
+    if segments_dir and all_segments:
+        # Make sure we're continuing the last checkpoint of the last segment.
+        # This requirement can be relaxed in the future if needed.
+        assert from_checkpoint, (
+            f"Found existing segments in directory '{segments_dir}'. Use"
+            " '--from-last-checkpoint' to continue from the last"
+            " checkpoint in this directory."
+        )
+        last_segment = all_segments[-1]
+        assert (
+            from_checkpoint.parent == last_segment.checkpoints_dir.resolve()
+        ), (
+            "You're not continuing from the last segment"
+            f" ({last_segment.path}). It is technically possible to continue"
+            " from a different checkpoint, but probably wrong. This assert"
+            " safeguards against inconsistent usage. If you want to continue"
+            f" from the checkpoint you specified ({from_checkpoint}), choose a"
+            " different directory to run in. Otherwise, use"
+            " '--from-last-checkpoint SEGMENTS_DIR' to continue from the"
+            " latest segment."
+        )
+        last_segment_checkpoints = last_segment.checkpoints
+        assert last_segment_checkpoints, (
+            f"The last segment '{last_segment.path}' has no checkpoints to"
+            " continue from. It is technically possible to continue from a"
+            " different checkpoint, but probably wrong. This assert safeguards"
+            " against inconsistent usage. If you want to continue from the"
+            f" checkpoint you specified ({from_checkpoint}), choose a different"
+            " directory to run in. Otherwise, remove the incomplete segment"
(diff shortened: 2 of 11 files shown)


nilsvu:
Notes from a discussion with @wthrowe @knelli2: We decided to look at template inheritance https://jinja.palletsprojects.com/en/3.1.x/templates/#template-inheritance to solve the issue with generating submit scripts.

wthrowe:
Following up on discussion elsewhere: The restart option can be added without duplicating the rest of the command with `${SPECTRE_CHECKPOINT:+ +restart "${SPECTRE_CHECKPOINT}"}`.  (The `:+` expansion gives the stuff to its right, but only if the variable on the left is defined and nonempty.)

nilsvu:
I implemented the submit script generation, take a look when you get the chance

src/Parallel/Main.hpp
```diff
@@ -140,15 +140,15 @@ class Main : public CBase_Main<Metavariables> {
...
 
  private:
-  // Return the dir name for the next Charm++ checkpoint as well as the pieces
-  // from which the name is built up: the basename and the padding. This is a
-  // "detail" function so that these pieces can be defined in one place only.
-  std::tuple<std::string, std::string, size_t> checkpoint_dir_basename_pad()
+  // Return the dir name for the Charm++ checkpoints as well as the prefix for
+  // checkpoint names and their padding. This is a "detail" function so that
+  // these pieces can be defined in one place only.
+  std::tuple<std::string, std::string, size_t> checkpoints_dir_prefix_pad()
```
knelli2: Once you make this a full fledged PR, can you factor these changes into another commit? This is just changing out checkpoint directory structure.

support/SubmitScripts/SubmitBase.sh
```diff
@@ -0,0 +1,79 @@
...
+echo "Nodelist: ${SLURM_JOB_NODELIST}"
+echo "Tasks: ${SLURM_NTASKS}"
+echo "CPUs per Task: ${SLURM_CPUS_PER_TASK}"
+echo "Charm ppn: ${CHARM_PPN}"
+echo "PATH: ${PATH}"
+echo "Executable: ${SPECTRE_EXECUTABLE}"
+echo "CLI: ${SPECTRE_CLI}"
+echo
+
+/usr/bin/modulecmd bash list
```
knelli2: I think this is specific to Wheeler.

support/SubmitScripts/SubmitBase.sh
```diff
@@ -0,0 +1,79 @@
...
+if [ $exit_code -eq 2 ]; then
+  sleep 10s
+  # Forward all parameters that were used to generate the submit script, so the
+  # resubmission uses the same configuration.
+  ${SPECTRE_CLI} schedule ${SPECTRE_INPUT_FILE} \
+    --submit-script-template "{{ submit_script_template }}" \
+    --executable ${SPECTRE_EXECUTABLE} \
+    --num-nodes ${SLURM_NNODES} \
+    --queue ${SLURM_JOB_PARTITION} \
+    --time-limit $(squeue -j ${SLURM_JOB_ID} -h --Format TimeLimit) \
```
knelli2: Is there a slurm env variable that will give you the current time limit?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
```
knelli2, line 2: Just a general comment: Would we be able to break up this file either into smaller files for the various classes/functions and maybe even separate commits? This is a lot for one commit and it's some dense code to go through.

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+    input_file_contents = input_file_template.read_text()
+
+    # Resolve executable
+    if not executable:
+        metadata = next(yaml.safe_load_all(input_file_contents))
+        try:
+            executable = metadata["Executable"]
+        except KeyError as err:
+            raise ValueError("Specify an 'executable' ('--executable' / '-e') "
+                             "or list one in the input file.") from err
```
knelli2: `"in the input file metadata as 'Executable:'"`

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+    elif not run_dir and not segments_dir:
+        # Neither run_dir nor segments_dir were specified. Set run_dir to the
+        # current working directory, but only if the input file is already there
+        # and we're running the executable directly (no scheduling).
+        if (scheduler is None
+                and input_file_template.resolve().parent == Path.cwd()):
+            run_dir = Path.cwd()
+        else:
+            raise ValueError("Specify a 'run_dir' ('--run-dir' / '-o') "
+                             "or a 'segments_dir' ('--segments-dir' / '-O').")
```
knelli2: `"or place an input file into the current directory if you aren't using a scheduler"`.

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+                    "You're not continuing from the previous segment's "
+                    "last checkpoint. This is technically possible, "
+                    "but probably wrong. This assert safeguards against "
+                    "inconsistent usage.")
+    context.update(run_dir=run_dir.resolve())
+    if segments_dir:
+        context.update(segments_dir=segments_dir.resolve())
+
+    # Resolve outfile
+    out_file = run_dir / out_file_name
```
knelli2, line 474: What is this syntax? Is it specific to `Path`?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+    input_file_path = run_dir / input_file_name
+    context.update(input_file=input_file_name)
+    rendered_input_file = template_env.from_string(input_file_contents).render(
+        context)
+    # Write the input file if it's not already there
+    if not (input_file_path.resolve() == input_file_template.resolve()
+            and rendered_input_file == input_file_contents):
+        if input_file_path.exists() and not force:
+            raise OSError(
+                f"File already exists at '{input_file_path}'. Delete it, "
```
knelli2: `"Input file already exists"`?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+        process.wait()
+        return process
+
+    # Copy executable
+    if segments_dir and executable.parent != segments_dir.resolve():
+        original_executable = executable
+        executable = (segments_dir / executable.name).resolve()
+        if not force and executable.exists():
+            raise OSError(
+                f"File already exists at '{executable}'. Delete it, "
```
knelli2: `"Executable already exists"`?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+
+    # Configure submit script
+    assert submit_script_template, (
+        "Please specify the 'submit_script_template'.")
+    submit_script_template = Path(submit_script_template).resolve()
+    submit_script_template_env = template_env.overlay(
+        loader=jinja2.FileSystemLoader(submit_script_template.parent))
+    # Copy submit script templates to segments directory
+    if segments_dir and submit_script_template.parent != segments_dir.resolve(
+    ):
```
knelli2: Is this really what yapf decided was the best formatting???

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+        original_submit_script_template = submit_script_template
+        submit_script_template = (segments_dir /
+                                  submit_script_template.name).resolve()
+        if not force and submit_script_template.exists():
+            raise OSError(
+                f"File already exists at '{submit_script_template}'. "
```
knelli2: `"Submit script template"`?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+                "Delete it, retry with 'force' ('--force' / '-f'), or choose "
+                "another 'segments_dir' ('--segments-dir' / '-O').")
+        logging.debug(
+            "Copy submit script template: "
+            f"{original_submit_script_template} -> {submit_script_template}")
+        shutil.copy(original_submit_script_template, submit_script_template)
+        submit_script_template_env = template_env.overlay(
+            loader=jinja2.FileSystemLoader(segments_dir))
+        ast = submit_script_template_env.parse(
+            submit_script_template.read_text())
```
knelli2: What's `ast`?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+        for referenced_template in jinja2.meta.find_referenced_templates(ast):
+            referenced_template_src = (original_submit_script_template.parent /
+                                       referenced_template)
+            referenced_template_dst = segments_dir / referenced_template
+            if not force and referenced_template_dst.exists():
+                raise OSError(
+                    f"File already exists at '{referenced_template_dst}'. "
```
knelli2: `"Submit script template"`? (I don't know how this one differs from the previous one)

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+    context.update(submit_script_template=submit_script_template)
+    rendered_submit_script = submit_script_template_env.get_template(
+        submit_script_template.name).render(context)
+    submit_script_path = run_dir / submit_script_name
+    # Write the submit script if it's not already there
+    if not (submit_script_path.resolve() == submit_script_template
+            and rendered_submit_script == submit_script_template.read_text()):
+        if not force and submit_script_path.exists():
+            raise OSError(
+                f"File already exists at '{submit_script_path}'. Delete it, "
```
knelli2: `"Submit script"`

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+    if submit or (submit is None and click.confirm(
+            f"Submit '{submit_script_path}'?", default=True)):
+        if isinstance(scheduler, str):
+            scheduler = [scheduler]
+        submit_process = subprocess.run(list(scheduler) + [submit_script_name],
+                                        cwd=run_dir,
+                                        capture_output=True,
+                                        text=True)
+        matched_submit_msg = re.match(r"Submitted batch job (\d+)",
+                                      submit_process.stdout)
```
knelli2: Is this the guaranteed message on all clusters?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+def _parse_param(value):
+    """Parse an additional command-line parameter for template substitutions
+
+    The following values are supported:
+
+    - Integers or floats
+    - List of values: "[1, 2, 3]"
+    - Exclusive range: "0..3" or "0..<3" (the latter is clearer, but "<" is a
+      special character in the shell)
+    - Inclusive range: "0...3"
```
knelli2, line 644: Is this standard notation? In bash `{0..5}` is inclusive.

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+from typing import List, Optional, Sequence, Union
+
+import click
+import jinja2
+import jinja2.meta
+import yaml
+
+logger = logging.getLogger(__name__)
+
+default_submit_script_template = Path(__file__).parent / 'Submit.sh'
```
knelli2: Where is this? I don't know what the `.parent` method of a Path does, but this looks like it's looking for a file named `Submit.sh` in the `support/` directory? Oh also, wouldn't this behavior change if you're in PY_DEV_MODE or not because of linking?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+@click.option(
+    '--submit-script-template',
+    default=default_submit_script_template,
+    show_default=True,
+    # No `type=click.Path` because this can be a Jinja template
+    help=("Path to a submit script. "
+          "It will be copied to the 'run_dir'. It can be a [Jinja template]("
+          "https://jinja.palletsprojects.com/en/3.0.x/templates/) "
+          "(see main help text for possible placeholders)."))
+@click.option('--queue', help=("Name of the queue."))
```
knelli2: I forget, is the plan to get these defaults (queue, time limit, nodes/cores) from the Machine file? Or are we getting rid of that?

support/Python/Schedule.py
```diff
@@ -0,0 +1,779 @@
...
+                              dir_okay=True,
+                              readable=True,
+                              path_type=Path),
+              help=("Restart from the last checkpoint in this directory."))
+@click.option('--job-name',
+              show_default="executable name",
+              help=("A short name for the job "
+                    "(see main help text for possible placeholders)."))
+@click.option('--num-procs',
+              '-j',
```
knelli2: Maybe `-c`? I know this is make syntax, but `-c` is srun syntax. Not sure which is better.

knelli2 (changes requested):
Haven't tested this out yet, but did a first pass of review. I'll probably have more things as I test it out.

support/SubmitScripts/SubmitBase.sh
nilsvu: Any reason why this isn't `module list`?

support/SubmitScripts/SubmitBase.sh
nilsvu: Unfortunately no :(

support/Python/Schedule.py
nilsvu: Yep, https://docs.python.org/3/library/pathlib.html

support/Python/Schedule.py
nilsvu: Yep :D Gotta switch to black at some point..

support/Python/Schedule.py
nilsvu: "abstract syntax tree" https://jinja.palletsprojects.com/en/3.1.x/api/#the-meta-api

support/Python/Schedule.py
nilsvu: Haven't seen Slurm output anything else, but we can change this if we run into issues

support/Python/Schedule.py
nilsvu: I got this from [Swift](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/basicoperators#Range-Operators) where `1..<3` is a half-open (exclusive) range and `1...3` is a closed (inclusive) range. Ranges are often a source of confusion and ambiguity, and I think the Swift syntax is quite clear.

support/Python/Schedule.py
nilsvu: Yep it's next to the `Schedule.py` file in `support/`. That's were CMake puts the machine-specific submit script template. It works with `PY_DEV_MODE` because the `.parent` is purely lexical (doesn't resolve symlinks). I'm open to better ideas how to copy around submit scripts!

support/Python/Schedule.py
nilsvu: At the moment I got rid of any use of the machine files in this script. Instead, there's only the submit script template (one per cluster), which sets these defaults. As it stands now we could delete the machine files.

support/Python/Schedule.py
nilsvu: I don't know either. I don't think there's a consensus among other tools. I got used to `-j` because of make. We could allow both, if that helps.

support/SubmitScripts/SubmitBase.sh
knelli2: Because for some reason, `module list` didn't work on the unlimited queue on Wheeler

support/Python/Schedule.py
knelli2: maybe name this `syntax_tree` then? Just to at least have some descriptor of what it is (even if I don't know what it actually is)

support/Python/Schedule.py
knelli2: If this follows Swift syntax, maybe add links to Swift in the relevant places so people can go look at that as well.

support/Python/Schedule.py
knelli2: Yeah do both

support/Python/Schedule.py
knelli2: Ok this seems fine then

support/Python/Schedule.py
knelli2: Ok I'll test it out and see how it works

support/Python/Schedule.py
nilsvu: Moved to generic function

support/Python/Schedule.py
nilsvu: Moved to generic function

support/Python/Schedule.py
nilsvu: Moved to generic function

nilsvu:
Thanks @knelli2, I implemented your suggestions and squashed because I moved commits / files / functions around. Sorry that that makes the smaller changes harder to find!

support/Python/Schedule.py
```diff
@@ -0,0 +1,702 @@
...
+
+    The input file, submit script, 'run_dir', 'segments_dir', and 'job_name' can
+    have placeholders like '{{ num_nodes }}'. They must conform to the
+    [Jinja template format](https://jinja.palletsprojects.com/en/3.0.x/templates/).
+    The placeholders are resolved in the following stages.
+    The following parameters are available as placeholders:
+
+    1. 'job_name' (if specified):
+
+        - All arguments to this function, including all additional '**kwargs'.
```
knelli2: These indents are rendered weird.

```
- All arguments to this function, including all additional '--params'.
      For example, the additional '--params' can include parameters
      controlling resolution in the input file.     - 'executable_name': Just
      the name of the executable (basename of the       'executable' path).
```

Same for the rest below.

src/Parallel/Main.hpp
```diff
@@ -941,39 +942,37 @@ void Main<Metavariables>::post_deadlock_analysis_termination() {
 
 template <typename Metavariables>
 std::tuple<std::string, std::string, size_t>
-Main<Metavariables>::checkpoint_dir_basename_pad() const {
-  const std::string basename = "SpectreCheckpoint";
-  constexpr size_t pad = 6;
+Main<Metavariables>::checkpoints_dir_prefix_pad() const {
+  const std::string checkpoints_dir = "Checkpoints";
+  const std::string prefix = "Checkpoint";
+  constexpr size_t pad = 3;
```
wthrowe: Leave it at 6.  Characters are cheap.

support/Python/DirectoryStructure.py
```diff
@@ -0,0 +1,116 @@
...
+        Segment000/
+            InputFile.yaml
+            Submit.sh
+            Output.h5
+            Checkpoints/
+        Segment001/...
+    ```
+
+    With SpEC we never needed more than 26*26=676 segments (AA-ZZ), so three
+    digits for segments is enough.
```
wthrowe: Is this true?  I vaguely think I've seen three-character SpEC segment IDs.

support/Python/Schedule.py
```diff
@@ -0,0 +1,704 @@
...
+    or if the existing file content is identical to the 'text'.
+    The 'hint' is appended to the error message.
+    """
+    if path.exists():
+        if path.read_text() == text:
+            return
+        if not force:
+            raise OSError(f"File already exists at '{path}'. Delete it or "
+                          "retry with 'force' ('--force' / '-f')." +
+                          (("\n" + error_hint) if error_hint else ""))
```
wthrowe: [optional] If we were confident that deleting the file was the correct thing to do, we would just do it here, so telling the user to do so may be giving bad instructions.  I'd just leave it at "File already exists.  Retry with --force to overwrite.".

support/Python/Schedule.py
```diff
@@ -0,0 +1,704 @@
...
+    Returns the path to the new file.
+    """
+    assert src_file.is_file()
+    assert dest_dir.is_dir()
+    if src_file.resolve().parent == dest_dir.resolve():
+        return src_file
+    dest = (dest_dir / src_file.name).resolve()
+    if dest.exists() and not force:
+        raise OSError(f"File already exists at '{dest}'. Delete it or "
+                      "retry with 'force' ('--force' / '-f').")
```
wthrowe: Same comment as above.

support/Python/Schedule.py
wthrowe: Does it give a useful return status?  That should be more robust than text parsing.

wthrowe (changes requested):
Why does wheeler have special instructions in the initial PR comment?  Is that still relevant, or have things become more uniform?

wthrowe:
Forgot to give my overall comment: I'm very happy with how the submit scripts are coming out with the templates.  This fixes the various problems with inconsistencies and duplication that we currently have, so this resolves the initial issue in #2987, although that spawned a lot of extra discussion that may not be considered resolved.

src/Parallel/Main.hpp
nilsvu: Yea they're cheap but become hard to read/count at some point. Try this: just glancing at `0000` you immediately know that it's 4 zeros. But glancing at `00000` you don't, you have to actually count. So I can make it 4 characters if you think we need that many, but anything beyond that I'd like to avoid. (also keep in mind that the checkpoint/segment names are actually not set in stone, I want to make sure that none of the code really relies on a hard-coded directory structure but just takes as input the files that it needs).

wthrowe: OK, but why do you want to count the digits?  I don't see a reason that should be a design goal.

If you don't want to discuss the counter format in this PR then don't change it.

nilsvu: I think readable directory names are a design goal. No need for excessive digits in my opinion. More important for segments than for checkpoints, so if you prefer I can keep the checkpoint format unchanged in this PR.

support/Python/DirectoryStructure.py
nilsvu: I haven't seen those, but that doesn't say much.

Three digits give us over 3 years in daily segments. That seems enough for the vast majority of simulations. As mentioned above the number of digits is not set in stone, so if we want to do a 5-year simulation or do hourly segments for some reason or something like that we can always increase the number of digits for that case.

nilsvu:
Pushed a fixup @wthrowe

src/Visualization/Python/ReadInputFile.py
```diff
@@ -44,3 +41,25 @@ def find_event(event_name: str, input_file: dict) -> dict:
...
+      input_file: The input file read in as a dictionary.
+
+    Returns: The phase change as a dictionary, or None if the phase change
+      wasn't found.
+    """
+    if not "PhaseChangeAndTriggers" in input_file:
+        return None
+    for trigger_and_phase_changes in input_file["PhaseChangeAndTriggers"]:
+        for phase_change in trigger_and_phase_changes["PhaseChanges"]:
+            if isinstance(phase_change, str):
```
knelli2: The phase change `CheckpointAndExitAfterWallclock` is a string though and we want that returned if it's there

knelli2 (changes requested):
Couple things. You can ignore the formatting one for now, but the phase change comment is because it's not behaving correctly. I have `CheckpointAndExitAfterWallclock` in my input file when trying to schedule just a stand alone run dir, and it doesn't error.

src/Visualization/Python/ReadInputFile.py
nilsvu: Oh I just used it wrong, the empty dict that's returned here evaluates to False in an if condition, so I changed it to `is not None`

knelli2: Ok now it works!

support/Python/Schedule.py
```diff
@@ -411,13 +420,43 @@ def schedule(input_file_template: Union[str, Path],
...
+        raise RuntimeError(f"Invalid input file '{input_file_path}':\n\n" +
+                           validate_process.stderr) from err
+    # - If the input file may request resubmissions, make sure we have a
+    #   segments directory
+    wallclock_exit_phase_change = find_phase_change(
+        "CheckpointAndExitAfterWallclock", input_file)
+    if wallclock_exit_phase_change is not None and not segments_dir:
+        raise ValueError(
+            "Found 'CheckpointAndExitAfterWallclock' in the input file but "
+            "no 'segments_dir' ('--segments-dir' / '-o') is set. "
```
knelli2: `-o` is for run dir. `-O` is for segments dir

nilsvu: oops

support/Python/Schedule.py
```diff
@@ -226,6 +230,15 @@ def schedule(input_file_template: Union[str, Path],
...
         Can be a Jinja template (see above).
+      copy_executable: By default, the executable is copied to the segments
+        directory to support resubmissions. It is not copied to individual
+        segments because we can't change the executable in-between segments
+        at the moment anyway.
+        If 'copy_executable' is set to True, the executable is also copied to
+        the run directory if no segments directory is set (it still won't be
+        copied to individual segments in a segments directory). This behavior
+        supports reproducibility.
+        If 'copy_executable' is set to False, the executable is never copied.
```
wthrowe: I'm confused.  Is the default true or false?  Is this saying the default depends on whether there is a segments directory?  (Which would be fine, but it should make that clear.)  Similarly confused by the short description below.

support/Python/DirectoryStructure.py
wthrowe: I've addressed the design comment in the other thread to try to keep the discussion more organized.

The point was trying to make here wasn't about your proposed design, but that I don't think "With SpEC we never needed more than 26*26=676 segments" is true, so you shouldn't claim it no matter what we decide for SpECTRE.

src/Parallel/Main.hpp
wthrowe: I hadn't actually compared them, so I tried creating a list of `CheckpointNNN` and `CheckpointNNNNNN` names from 0 to 19 (I got tired of typing at that point).  I actually found the six-digit version to be significantly easier to read, because the string of zeros acts as a separator between "Checkpoint" and the number.  It makes it easier to visually locate the first nonzero digit.  Presumably the same would hold for other prefixes.

[Continuing from a thread below, but consolidating because the topic is the same]

> Three digits give us over 3 years in daily segments. That seems enough for the vast majority of simulations. As mentioned above the number of digits is not set in stone, so if we want to do a 5-year simulation or do hourly segments for some reason or something like that we can always increase the number of digits for that case.

You assume that segments will always be based on a timer.  This has not always been true for SpEC, and while it is a goal for SpECTRE I don't think we should assume that we will always achieve it immediately.  SpEC segments could be extremely short (minutes, maybe less) because they were based on some event in the simulation that forced a new segment.

Anyway, @markscheel, you're probably the most experienced SpEC-runner here.  Do you want to weigh in?

docs/Tutorials/CheckpointRestart.md
```diff
@@ -31,9 +31,9 @@ Executables can checkpoint when:
 
 To restart an executable from a checkpoint file, run a command like this:
 ```
-./MySpectreExecutable +restart SpectreCheckpoint000123
+./MySpectreExecutable +restart Checkpoints/Checkpoint123
 ```
-where the `000123` should be the number of the checkpoint to restart from.
+where the `123` should be the number of the checkpoint to restart from.
```
wthrowe: Whatever we decide in the discussion on formatting, this documentation should make it clear that these are zero-padded, either by stating it or by using an example with a leading zero.

wthrowe (changes requested):

support/Python/Schedule.py
nilsvu: Pushed another fixup to clarify

src/Parallel/Main.hpp
nilsvu: I see what you mean with the zero-padding visually separating the number from the prefix. How about 4 digits for both segments and checkpoints then? Still not excessive, hence ok to type without having to pay too much attention to the number of keystrokes (which I find kind of hard for 6 digits), but surely enough and visually pleasing.

support/Python/Schedule.py
```diff
@@ -0,0 +1,754 @@
...
+@click.option('--copy-executable/--no-copy-executable',
+              default=None,
+              help=("Copy the executable to the run or segments directory. "
+                    "Default (when no flag is specified) is to copy only to "
+                    "the segments directory to support resubmission. "
+                    "With '--copy-executable' the executable is also copied "
+                    "to the run directory (only has an effect if "
+                    "'--run-dir' / '-o' is set). "
+                    "With '--no-copy-executable' the executable is never "
+                    "copied."))
```
knelli2: This is still a bit confusing. May I suggest

```
Copy the executable to the run or segments directory.
When no flag is specified:
 - If '--run-dir' / '-o' is set, don't copy.
 - If '-segments-dir' / '-O' is set, copy to segments directory to
   support resubmission.
If '--copy-executable' is specified:
 - If '--run-dir' / '-o' is set, copy to the run directory.
 - If '-segments-dir' / '-O' is set, copy to segments directory to
   support resubmission. (Note: the executable is not copied to
   individual segments)
If '--no-copy-executable' is specified: Never copy.
```

I know it's a bit verbose, but I think this is very clear on what the behavior is and is easy for users to follow.

nilsvu: Sounds good, done.

nilsvu:
@wthrowe @knelli2 lots of fixups for you to look at whenever you get the chance :)

support/Python/CMakeLists.txt
```diff
@@ -25,7 +25,7 @@ if(MACHINE)
     )
   configure_file(
     ${CMAKE_SOURCE_DIR}/support/SubmitScripts/${MACHINE}.sh
-    ${SPECTRE_PYTHON_PREFIX}/support/Submit.sh
+    ${SPECTRE_PYTHON_PREFIX}/support/SubmitTemplate.sh
```
knelli2, line 29: Also need to change this towards the top of `Status.py`:
```py
default_submit_script_template = Path(__file__).parent / "SubmitTemplate.sh"
```
and a couple places in comments/docs I think.

I also still think `SubmitBase.sh` should be `SubmitTemplateBase.sh`. Even though now it's clear neither should be used as a submission script, I have no clue what the difference between them is. I think having `NameBase.sh` and `Name.sh` is clear that `Name.sh` depends on `NameBase.sh` (and then the `Template` part still signals that it shouldn't actually be used to submit).

support/Python/Schedule.py
```diff
@@ -349,7 +351,10 @@ def schedule(
     context.update(executable_name=executable.name)
 
     # Resolve CLI
-    spectre_cli = shutil.which("spectre")
+    # This is the path of the `spectre` CLI where this script is installed.
+    # We may have to make this more robust for resubmissions if we run into
+    # problems or unexpected behavior.
+    spectre_cli = Path(__file__).parent.parent.parent.parent / "spectre"
```
knelli2: This seems very fragile (also build directories don't have to be placed inside the SPECTRE_HOME). What was wrong with the previous method?

support/SubmitScripts/SubmitBase.sh
```diff
@@ -0,0 +1,81 @@
...
+    --from-last-checkpoint ./Checkpoints \
```
knelli2: This must be

```sh
    --from-last-checkpoint ${RUN_DIR} \
```

The current way fails.

support/Python/Schedule.py
```diff
@@ -415,16 +415,21 @@ def schedule(
...
+                Path(from_checkpoint).resolve()
+                == last_checkpoint.path.resolve()
+            )
+        except (AssertionError, IndexError) as err:
+            raise AssertionError(
+                "You're not continuing from the previous segment's last"
+                " checkpoint. This is technically possible, but probably"
+                " wrong. This assert safeguards against inconsistent usage."
+                " Either run in a different directory, or use"
+                " '--from-checkpoint' to continue from the last checkpoint."
```
knelli2: Hmm this error message is trying to do two things at once and it makes things confusing. There are two ways that I run into this:

1. I'm trying to continue from a checkpoint that is not the last checkpoint in a `Checkpoints` directory.
2. I try to continue from a segment that is not the most recent segment in a segments directory.

I still think we shouldn't allow either at the moment, but we need to distinguish between the two. What about this:

```py
        assert from_checkpoint, "No checkpoint specified, but trying to restart from a checkpoint"
        last_segment = all_segments[-1]
        if Path(from_checkpoint).parent.parent.resolve() != last_segment.path.resolve():
            raise AssertionError(
                "You're not continuing from the latest segment "
                f"{last_segment.path.resolve()}. All segments:\n"
                f"{[seg.path.resolve() for seg in all_segments]}"
            )
        last_checkpoint = all_segments[-1].checkpoints[-1]
        if Path(from_checkpoint).resolve() != last_checkpoint.path.resolve():
            raise AssertionError(
                "You're not continuing from the latest checkpoint "
                f"{last_checkpoint.path.resolve()}. All checkpoints:\n"
                f"{[check.path for check in all_segments[-1].checkpoints]}"
            )
```

Feel free to make formatting nicer/more verbose messages, but I'd say keep the two messages separate.

knelli2 (changes requested):
Couple more comments about some of the current fixups.

support/Python/CMakeLists.txt
nilsvu: ok

support/Python/Schedule.py
nilsvu: This isn't as fragile as you think because this is in the build directory (not in `SPECTRE_HOME`). So it only breaks if we move the location of the python package inside the build dir.

This method ensures that we use the same CLI for resubmissions. The previous method would use whichever CLI is in the `PATH`, which we can't control.

support/SubmitScripts/SubmitBase.sh
nilsvu: Worked for me, did this fail when you tried? The code should resolve the relative path correctly (we're in the `RUN_DIR` here).

support/Python/Schedule.py
nilsvu: Added a test for this too

support/Python/Schedule.py
nilsvu: I split this error message into three, but kept the "last segment" vs "last checkpoint" together. The error prints the expected path of the checkpoint, which includes both the segment and the checkpoint, so I see no point in splitting it into separate errors messages.

support/Python/Schedule.py
knelli2: This will work even in `PY_DEV_MODE`?

support/SubmitScripts/SubmitBase.sh
knelli2: It had previously failed, but now it works. Not sure what changed.

support/Python/Schedule.py
knelli2: I still think the third error message you have isn't clear.

```
You're not continuing from the previous segment's last checkpoint
(Segment001/Checkpoints/Checkpoint000).
```

I agree that this is a true statement and that it tells you what you should be doing (i.e. continuing from this specific checkpoint which is the last checkpoint in the latest segment). However, it's confusing because it doesn't tell me what I did wrong. Did I specify a segment other than the latest one? Or did I specify the correct segment, but not the last checkpoint in the latest segment?

And to me, that's confusing because I'm going to go back to the (incorrect) command I used and look for the offending option. If I specified an incorrect segment, that's what I'll be looking for. If it's an incorrect checkpoint, then I'll be looking for that. But I have to know what I'm looking for in the first place and your error message doesn't tell me that.

Also, I didn't find this line very helpful.

```
Either run in a different directory, or use '--from[-last]-checkpoint'
to continue from the last checkpoint.
```

Again, what I should be doing is dependent on whether I specified the incorrect segment or checkpoint. If I specified the correct segment but an incorrect checkpoint, the suggestion is simple:

```
Use '--from-last-checkpoint {last_segment}' or '--from-checkpoint {last_checkpoint}' to
continue from the last checkpoint.
```

However, if I specified an incorrect segment it's a bit trickier because of potentially corrupted segments. Did I actually want to continue from the segment I specified? If so, then I need to remove all segments after that one. Or did I just accidentally specify the wrong segment and I meant to continue from the latest one? For both of these I think a message like this would be helpful:

```
You are not continuing from the latest segment {last_segment}. If you want to continue
from the last checkpoint in the segment you specified ({specified_segment}), you must
remove all the segments after that one [list segments to remove?]. Otherwise, specify
'--from-last-checkpoint {last_segment}' to continue from the latest segment.
```

So overall I think this should be the order of the `asserts`:

1. assert from_checkpoint, "Found segments. Use '--from-last-segment {last_segment}' to continue from the last checkpoint in {last_segment}"
2. assert we are in last segment. If not, use above message.
3. assert current segment (which is the latest segment) has checkpoints, "Oh no, no checkpoints to continue from"
4. assert continuing from last checkpoint in this segment (the latest segment). if not use above message.

I know this is a lot of logic and similar error messages whereas yours is less logic and combined messages. But I think the differences in the message that are tailored to the specific errors are very clear, will help users a lot, and make the experience of using the scheduler a lot smoother. What do you think?

support/Python/Schedule.py
nilsvu: Yep, symlinks aren't resolved

support/Python/Schedule.py
nilsvu: See fixup

nilsvu:
@knelli2 please check the latest fixups when you get the chance, then I can squash them all together for @wthrowe to look at (so he doesn't have to read through a million fixups)

nilsvu:
Pushed another fixup for resubmissions ✌️

knelli2:
Ok you can squash now. I imagine there will be bugs we'll need to iron out, but should be ok for now

support/Python/Schedule.py
knelli2: Sorry, but I still think these error messages will be confusing for users. However, I've held up this PR long enough and the messages are already decent-ish so I think it's better to just get this merged and I can do a follow-up PR.

nilsvu:
Squashed it all down. @wthrowe there were so many fixups that I lost track of what you've seen and what you haven't. I can probably dig up the git reflog if you prefer to see a single squashed-together fixup on top of your review above.

wthrowe:
The test is failing.

support/Python/Schedule.py
```diff
@@ -0,0 +1,1064 @@
...
+    # Segments and run directories
+
+    You can set either the 'run_dir' or the 'segments_dir' to specify where the
+    executable will run (but not both). If you specify a 'run_dir', the
+    executable will run in it directly. If you specify a 'segments_dir', a new
+    segment will be created and used as the 'run_dir'. Segments are named with
+    incrementing integers and continue the run from the previous segment. For
+    example, the following is a typical 'segments_dir':
+
+    \b
```
wthrowe, line 156: This is probably some python thing that I don't understand, but everything I can find on the internet says this is adding a backspace character.  Why?

support/Python/Schedule.py
```diff
@@ -0,0 +1,1064 @@
...
+      run_dir: The directory to which input file, submit script, etc. are
+        copied, and relative to which the executable will run.
+        Can be a Jinja template (see above).
+      segments_dir: The directory in which a new segment is created as the
+        'run_dir'. Mutually exclusive with 'run_dir'.
+        Can be a Jinja template (see above).
+      copy_executable: Copy the executable to the run or segments directory.
+        By default (when set to 'None'):
+          - If '--run-dir' / '-o' is set, don't copy.
+          - If '-segments-dir' / '-O' is set, copy to segments directory to
```
wthrowe: Typo in number of dashes.  Again 4 lines down.

support/Python/Schedule.py
```diff
@@ -0,0 +1,1064 @@
...
+        confirmation before submitting.
+      clean_output: Optional. When 'True', use
+        'spectre.tools.CleanOutput.clean_output' to clean up existing output
+        files in the 'run_dir' before scheduling the run. (Default: 'False')
+      force: Optional. When 'True', overwrite input file and submit script
+        in the 'run_dir' instead of raising an error when they already exist.
+
+    Returns: The 'subprocess.CompletedProcess' representing either the process
+      that scheduled the run, or the process that ran the executable if
+      'scheduler' is 'None'. Returns 'None' if no run was scheduled.
```
wthrowe: What if multiple runs are scheduled?

nilsvu:
Rebased!

support/Python/Schedule.py
```diff
@@ -0,0 +1,1064 @@
...
+    ),
+)
+@click.option(
+    "--copy-executable/--no-copy-executable",
+    default=None,
+    help=(
+        "Copy the executable to the run or segments directory. "
+        "(1) When no flag is specified: "
+        "If '--run-dir' / '-o' is set, don't copy. "
+        "If '-segments-dir' / '-O' is set, copy to segments "
```
wthrowe: Dashes again.

wthrowe (changes requested):
Mostly looked at the documentation, since I'm not good with python.

support/Python/Schedule.py
nilsvu: The `\b` escape marker prevents [click](https://click.palletsprojects.com/en/8.1.x/documentation/#preventing-rewrapping) from rewrapping the following paragraph: https://click.palletsprojects.com/en/8.1.x/documentation/#preventing-rewrapping

support/Python/Schedule.py
nilsvu: Changed to `Returns 'None' if no or multiple runs were scheduled.`

nilsvu:
Squashed in those fixes and split the changes into a few more commits

wthrowe (approved):

knelli2 (approved):
Looks good! Excited to start using this

nilsvu:
@wthrowe I think this is ready to go

Issue #3322, opened 2021-06-25
`MinimumGridSpacingCompute` is slow and does not appear to be memoized properly.


nikwit:
# Bug reports:

When profiling my curved scalar wave executable I find that over 20% of execution time is spent in `minimum_grid_spacing`. 
I can reproduce this behavior with the `EvolveScalarWavePlaneWave3D` from current develop. 

Firstly, why is this function so incredible slow? I faintly remember this coming up a few months ago but don't remember the details. Is it possible that the `goto` statement in there prevents compiler optimizations?

Secondly, e.g. for `EvolveScalarWavePlaneWave3D` the inertial coordinates should be time-independent, right? Why does the databox not cache the result properly given that the input arguments (the mesh and the coordinates) should in theory not be changing?

### Environment:

Add as an attachment `$SPECTRE_BUILD_DIR/LibraryVersions.txt` or
add its contents here.

# Feature request:

### Component:

- [ ] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] Detail 1
- [ ] Detail 2
- [ ] Detail 3

fmahebert:
When I previously looked into this a bit, the most promising thing I found was that the `operator==` for `std::array` was slower than explicitly checking by hand `if constexpr (Dim == 1) { return lhs[0] == rhs[0] } else if constexpr ...` and so on.

But I never got a satisfactory understanding of why the function remained so expensive. There is quite a bit of looping done, comparing lots of pairs of points... could speed that up by only comparing points near the boundary (where the points cluster), but that's making an assumption that _could_ fail. And to allow for time-dependent maps, you need to redo the work each time step. But even considering all this, it's still surprisingly expensive...

wthrowe:
> When I previously looked into this a bit, the most promising thing I found was that the `operator==` for `std::array` was slower than explicitly checking by hand `if constexpr (Dim == 1) { return lhs[0] == rhs[0] } else if constexpr ...` and so on.

Interesting.  libstdc++ implements this operation using `std::equal`.  I don't see any reason that should be slow (maybe the compiler can't figure out how to unroll the loop?), but writing a comparison that has no loops is straightforward: (untested)
```c++
template <typename T, size_t N, size_t... Is>
bool array_equal_impl(const std::array<T, N>& a, const std::array<T, N>& b,
                      std::index_sequence<Is...> /*meta*/) noexcept {
[...]

PR #5005, opened 2023-05-16
Add moving mesh terms to GhGrhd system

ffoucart:
## Proposed changes

This PR adds terms needed to capture moving meshes in the GhValenciaDivClean system to the TimeDerivative.hpp file of GhValenciaDivClean/Subcell.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DgSubcell/Tags/Jacobians.hpp
+++ b/src/Evolution/DgSubcell/Tags/Jacobians.hpp
@@ -8,7 +8,10 @@
 #include <string>
 
 #include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/Tensor/EagerMath/Determinant.hpp"
 #include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Domain/ElementMap.hpp"
+#include "Utilities/SetNumberOfGridPoints.hpp"
 
 namespace evolution::dg::subcell::fd::Tags {
 /// \brief The inverse Jacobian from the element logical frame to the grid frame
@@ -22,8 +25,8 @@ namespace evolution::dg::subcell::fd::Tags {
 template <size_t Dim>
 struct InverseJacobianLogicalToGrid : db::SimpleTag {
   static std::string name() { return "InverseJacobian(Logical,Grid)"; }
-  using type = std::optional<
-      ::InverseJacobian<DataVector, Dim, Frame::ElementLogical, Frame::Grid>>;
+  using type =
+      ::InverseJacobian<DataVector, Dim, Frame::ElementLogical, Frame::Grid>;
 };
 
 /// \brief The determinant of the inverse Jacobian from the element logical
@@ -33,6 +36,123 @@ struct InverseJacobianLogicalToGrid : db::SimpleTag {
 /// and reduce the memory footprint.
 struct DetInverseJacobianLogicalToGrid : db::SimpleTag {
   static std::string name() { return "Det(InverseJacobian(Logical,Grid))"; }
-  using type = std::optional<Scalar<DataVector>>;
+  using type = Scalar<DataVector>;
 };
+
+/// \brief The inverse Jacobian from the element logical frame to the inertial
+/// frame at the cell centers.
+///
+/// Specifically, \f$\partial x^{\bar{i}} / \partial x^i\f$, where \f$\bar{i}\f$
+/// denotes the element logical frame and \f$i\f$ denotes the inertial frame.
+///
+/// \note stored as a `std::optional` so we can reset it when switching to DG
+/// and reduce the memory footprint (is it useful for a ComputeTag though?).
+template <size_t Dim>
+struct InverseJacobianLogicalToInertial : db::SimpleTag {
+  static std::string name() { return "InverseJacobian(Logical,Inertial)"; }
+  using type = ::InverseJacobian < DataVector, Dim, Frame::ElementLogical,
+        Frame::Inertial >;
+};
+
+/// Compute item for the inverse jacobian matrix from logical to
+/// grid coordinates
+template <typename MapTagLogicalToGrid, size_t Dim>
+struct InverseJacobianLogicalToGridCompute : InverseJacobianLogicalToGrid<Dim>,
+                                             db::ComputeTag {
+  static constexpr size_t dim = Dim;
+  using base = InverseJacobianLogicalToGrid<Dim>;
+  using return_type = typename base::type;
+  using argument_tags = tmpl::list<
+      MapTagLogicalToGrid,
+      evolution::dg::subcell::Tags::Coordinates<dim, Frame::ElementLogical>>;
+  static void function(
+      const gsl::not_null<return_type*> inverse_jacobian_logical_to_grid,
+      const ElementMap<Dim, Frame::Grid>& logical_to_grid_map,
+      const tnsr::I<DataVector, dim, Frame::ElementLogical>& logical_coords) {
+    *inverse_jacobian_logical_to_grid =
+        logical_to_grid_map.inv_jacobian(logical_coords);
+  }
+};
+
+/// Compute item for the determinant of the inverse jacobian matrix
+/// from logical to grid coordinates
+template <size_t Dim>
+struct DetInverseJacobianLogicalToGridCompute
+    : DetInverseJacobianLogicalToGrid,
+      db::ComputeTag {
+  static constexpr size_t dim = Dim;
+  using base = DetInverseJacobianLogicalToGrid;
+  using return_type = typename base::type;
+  using argument_tags = tmpl::list<InverseJacobianLogicalToGrid<Dim>>;
+  static void function(
+      const gsl::not_null<return_type*> det_inverse_jacobian_logical_to_grid,
+      const ::InverseJacobian<DataVector, Dim, Frame::ElementLogical,
+                              Frame::Grid>
+          inverse_jacobian_logical_to_grid) {
+    *det_inverse_jacobian_logical_to_grid =
+        determinant(inverse_jacobian_logical_to_grid);
+  }
+};
+
+/// Compute item for the inverse jacobian matrix from logical to
+/// inertial coordinates
+template <typename MapTagGridToInertial, size_t Dim>
+struct InverseJacobianLogicalToInertialCompute
+    : InverseJacobianLogicalToInertial<Dim>,
+      db::ComputeTag {
+  static constexpr size_t dim = Dim;
+  using base = InverseJacobianLogicalToInertial<Dim>;
+  using return_type = typename base::type;
+  using argument_tags =
+      tmpl::list<MapTagGridToInertial,
+                 evolution::dg::subcell::Tags::Coordinates<dim, Frame::Grid>,
+                 ::Tags::Time, ::domain::Tags::FunctionsOfTime,
+                 InverseJacobianLogicalToGrid<Dim>>;
+  static void function(
+      const gsl::not_null<return_type*> inverse_jacobian_logical_to_inertial,
+      const ::domain::CoordinateMapBase<Frame::Grid, Frame::Inertial, dim>&
+          grid_to_inertial_map,
+      const tnsr::I<DataVector, dim, Frame::Grid>& grid_coords,
+      const double time,
+      const std::unordered_map<
+          std::string,
+          std::unique_ptr<::domain::FunctionsOfTime::FunctionOfTime>>&
+          functions_of_time,
+      const ::InverseJacobian<DataVector, Dim, Frame::ElementLogical,
+                              Frame::Grid>& inverse_jacobian_logical_to_grid) {
+    if (grid_to_inertial_map.is_identity()) {
+      // Optimization for time-independent maps; we just point to the
+      // logical-to-grid inverse jacobian.
+      const size_t num_pts = inverse_jacobian_logical_to_grid[0].size();
+      for (size_t storage_index = 0;
+           storage_index < inverse_jacobian_logical_to_grid.size();
+           ++storage_index) {
+        make_const_view(
+            make_not_null(&std::as_const(
+                (*inverse_jacobian_logical_to_inertial)[storage_index])),
+            inverse_jacobian_logical_to_grid[storage_index], 0, num_pts);
+      }
+    } else {
+      set_number_of_grid_points(inverse_jacobian_logical_to_inertial,
+                                get<0>(grid_coords).size());
+      // Get grid to inertial inverse jacobian
+      const auto& inverse_jacobian_grid_to_inertial =
+          grid_to_inertial_map.inv_jacobian(grid_coords, time,
+                                            functions_of_time);
+      for (size_t i = 0; i < Dim; i++) {
+        for (size_t j = 0; j < Dim; j++) {
+          auto& inv_jacobian_component =
+              inverse_jacobian_logical_to_inertial->get(i, j);
+          inv_jacobian_component = 0.;
+          for (size_t k = 0; k < Dim; k++) {
+            inv_jacobian_component +=
+                inverse_jacobian_logical_to_grid.get(i, k) *
+                inverse_jacobian_grid_to_inertial.get(k, j);
+          }
+        }
+      }
+    }
+  }
+};
+
 }  // namespace evolution::dg::subcell::fd::Tags
--- a/tests/Unit/Evolution/DgSubcell/Test_JacobianCompute.cpp
+++ b/tests/Unit/Evolution/DgSubcell/Test_JacobianCompute.cpp
@@ -1,0 +1,104 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <array>
+#include <cstddef>
+#include <memory>
+#include <optional>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "Domain/Block.hpp"
+#include "Domain/CoordinateMaps/CoordinateMap.hpp"
+#include "Domain/CoordinateMaps/CoordinateMap.tpp"
+#include "Domain/CoordinateMaps/Identity.hpp"
+#include "Domain/CreateInitialElement.hpp"
+#include "Domain/Creators/Tags/FunctionsOfTime.hpp"
+#include "Domain/Structure/Element.hpp"
+#include "Domain/Tags.hpp"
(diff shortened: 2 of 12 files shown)


ffoucart:
Current version allows for the evolution of a single TOV star on a mesh with constant velocity translation. It is still lacking a proper calculation of the inverse jacobian. Specifically, L468-474 of 
`src/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/TimeDerivative.hpp`
clearly need to be fixed.

ffoucart:
The code also cannot handle non-diagonal jacobian matrices for the coordinate maps.

ffoucart:
Added the calculation of the inverse jacobian (though this is largely untested beyond a single star on a constant velocity grid... for which the jacobian between grid and inertial coordinates is the identity matrix)

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -35,4 +35,76 @@ struct DetInverseJacobianLogicalToGrid : db::SimpleTag {
...
+/// \brief The inverse Jacobian from the element logical frame to the inertial
+/// frame at the cell centers.
+///
+/// Specifically, \f$\partial x^{\bar{i}} / \partial x^i\f$, where \f$\bar{i}\f$
+/// denotes the element logical frame and \f$i\f$ denotes the inertial frame.
+///
+/// \note stored as a `std::optional` so we can reset it when switching to DG
+/// and reduce the memory footprint (is it useful for a ComputeTag though?).
+template <size_t Dim>
+struct InverseJacobianLogicalToInertial : db::SimpleTag {
```
nilsdeppe, line 51: Could you add tests for the two new tags to `tests/Unit/Evolution/DgSubcell/Test_Tags.cpp` please?

nilsdeppe (changes requested):
You can squash the changes into the second commit immediately

ffoucart:
@nilsdeppe : I think this new version should work; presumably the cancelled test is part of the test time outs we are seeing on builds at the moment?

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -35,4 +35,76 @@ struct DetInverseJacobianLogicalToGrid : db::SimpleTag {
...
+struct InverseJacobianLogicalToInertialCompute
+    : InverseJacobianLogicalToInertial<Dim>,
+      db::ComputeTag {
+  static constexpr size_t dim = Dim;
+  using base = InverseJacobianLogicalToInertial<Dim>;
+  using return_type = typename base::type;
+  using argument_tags =
+      tmpl::list<MapTagGridToInertial,
+                 evolution::dg::subcell::Tags::Coordinates<dim, Frame::Grid>,
+                 ::Tags::Time, ::domain::Tags::FunctionsOfTime,
```
nilsvu, line 109: I think even on static meshes this gets reevaluated at every time step, because the value of `::Tags::Time` changes. Could be pretty bad for performance. See also #3322. Maybe just add a comment that this get reevaluated every time step even on static meshes.

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -35,4 +35,76 @@ struct DetInverseJacobianLogicalToGrid : db::SimpleTag {
...
+      const tnsr::I<DataVector, dim, Frame::Grid>& grid_coords,
+      const double time,
+      const std::unordered_map<
+          std::string,
+          std::unique_ptr<::domain::FunctionsOfTime::FunctionOfTime>>&
+          functions_of_time,
+      std::optional<::InverseJacobian<DataVector, Dim, Frame::ElementLogical,
+                                      Frame::Grid>>
+          inverse_jacobian_logical_to_grid) {
+    if (inverse_jacobian_logical_to_grid.has_value()) {
```
nilsvu: I ran into the issue that the logical->grid Jacobian is null at t=0 before taking the first time step, so the initial observation errored with "bad optional access".

src/Evolution/DgSubcell/Tags/Jacobians.hpp
ffoucart: Right, we might want a time-independent and a time-dependent version in the future, if we care about performance on static meshes.

src/Evolution/DgSubcell/Tags/Jacobians.hpp
ffoucart: Maybe a discussion for the code meeting later; I'm not sure what the solution would be here.

nilsdeppe: This is because it's currently only being computed in https://github.com/sxs-collaboration/spectre/blob/b1451b1e87807b37726484b69a656dd93c4cc6ab/src/Evolution/DgSubcell/Actions/TakeTimeStep.hpp#L81 We could add a compute tag for it. I think that should fix the issue and guarantee that the value is always valid (and remove the `std::optional`).

ffoucart: Do you mean a compute tag for the logical-to-grid inverse jacobian?
[And then removing std::optional for both the logical-to-grid and logical-to-inertial inverse jacobians?]

nilsdeppe: Yea, that's right! And then the l2i can just set the DataVector pointers to the l2g one when the mesh is static. That makes the computation super cheap :D

ffoucart:
@nilsdeppe Could you look at the last commit and see if that is roughly what you had in mind?
[I'm sure some test / executable aren't fully updated yet, but to get an idea of whether this is going in the right direction...]

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -76,33 +117,25 @@ struct InverseJacobianLogicalToInertialCompute
...
-        for (size_t j = 0; j < Dim; j++) {
-          auto& inv_jacobian_component =
-              inverse_jacobian_logical_to_inertial->value().get(i, j);
-          inv_jacobian_component = 0.;
-          for (size_t k = 0; k < Dim; k++) {
-            inv_jacobian_component +=
-                inverse_jacobian_logical_to_grid.value().get(i, k) *
-                inverse_jacobian_grid_to_inertial.get(k, j);
-          }
+    *inverse_jacobian_logical_to_inertial =
```
nilsdeppe: Here I would do:
```cpp
if (UNLIKELY(not inverse_jacobian_logical_to_inertial->has_value())) {
    *inverse_jacobian_logical_to_inertial =
            ::InverseJacobian<DataVector, Dim, Frame::ElementLogical,
                              Frame::Inertial>(get<0>(grid_coords).size());
} else {
  destructive_resize_components(make_not_null(&inverse_jacobian_logical_to_inertial->value()), get<0>(grid_coords).size());
}
```
to avoid zeroing and allocating memory except when strictly necessary

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -76,33 +117,25 @@ struct InverseJacobianLogicalToInertialCompute
...
-      for (size_t i = 0; i < Dim; i++) {
-        for (size_t j = 0; j < Dim; j++) {
-          auto& inv_jacobian_component =
-              inverse_jacobian_logical_to_inertial->value().get(i, j);
-          inv_jacobian_component = 0.;
-          for (size_t k = 0; k < Dim; k++) {
-            inv_jacobian_component +=
-                inverse_jacobian_logical_to_grid.value().get(i, k) *
-                inverse_jacobian_grid_to_inertial.get(k, j);
-          }
```
nilsdeppe: I would also do:
```cpp
if (grid_to_inertial_map.is_identity()) {
  // Optimize time-independent case by only setting pointers.
  if (UNLIKELY(not inverse_jacobian_logical_to_inertial->has_value())) {
    *inverse_jacobian_logical_to_inertial =
            ::InverseJacobian<DataVector, Dim, Frame::ElementLogical,
                              Frame::Inertial>(get<0>(grid_coords).size());
  }
  for (size_t storage_index = 0; storage_index < inverse_jacobian_logical_to_grid.size(); ++storage_index) {
    make_const_view(make_not_null(&std::as_const(inverse_jacobian_logical_to_inertial->value()[storage_index])),
       inverse_jacobian_logical_to_grid[storage_index], 0, inverse_jacobian_logical_to_grid[storage_index].size());
  }
} else {
  // All the code you currently have
}
```
This means that when the maps are time-dependent we don't compute anything, so the tag will be cheap to evaluate

nilsdeppe (commented):
Sorry for the delay! Couple suggestions but yes, this is exactly what I had in mind, thanks! You can squash things right away :)

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -33,6 +35,124 @@ struct InverseJacobianLogicalToGrid : db::SimpleTag {
...
+      const std::unordered_map<
+          std::string,
+          std::unique_ptr<::domain::FunctionsOfTime::FunctionOfTime>>&
+          functions_of_time,
+      const ::InverseJacobian<DataVector, Dim, Frame::ElementLogical,
+                              Frame::Grid>
+          inverse_jacobian_logical_to_grid) {
+    destructive_resize_components(inverse_jacobian_logical_to_inertial,
+                                  get<0>(grid_coords).size());
+    if (grid_to_inertial_map.is_identity()) {
```
ffoucart, line 123: @nilsdeppe Can you check whether the if...else... statement here does what you want? I had to modify things a bit because your code snippets were assuming that we were using std::optional, while we decided earlier to no longer use optionals for the inverse jacobians.

nilsdeppe: Yep, this is perfect!

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -33,6 +35,124 @@ struct InverseJacobianLogicalToGrid : db::SimpleTag {
...
+      const tnsr::I<DataVector, dim, Frame::ElementLogical>& logical_coords) {
+    *inverse_jacobian_logical_to_grid =
+        logical_to_grid_map.inv_jacobian(logical_coords);
+  }
+};
+
+/// Compute item for the determinant of the inverse jacobian matrix
+/// from logical to grid coordinates
+template <size_t Dim>
+struct DetInverseJacobianLogicalToGridCompute
```
nilsdeppe, line 80: Will you need the det for logical to inertial as well? I think the FD code needs it, I don't remember whether it was face-centered or cell-centered that it needed (probably both...). If you do need it, would you mind adding it now with the same `if-else` structure as the `InverseJacobianLogicalToInertialCompute`?

tests/Unit/Evolution/DgSubcell/Test_Tags.cpp
```diff
@@ -107,6 +110,11 @@ void test(const bool moving_mesh) {
                                                     Frame::Inertial>>>(
       "InertialCoordinates");
 
+  TestHelpers::db::test_compute_tag<
+      subcell::fd::Tags::InverseJacobianLogicalToInertialCompute<
```
nilsdeppe, line 114: Could you add tests for the few other new tags please?

nilsdeppe (commented):
LGTM, couple small things you can squash immediately

ffoucart:
Hmm, broke something in the test when rebasing... I had only tested that the moving TOV star was still running fine...

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -33,6 +35,122 @@ struct InverseJacobianLogicalToGrid : db::SimpleTag {
...
+      const ::domain::CoordinateMapBase<Frame::Grid, Frame::Inertial, dim>&
+          grid_to_inertial_map,
+      const tnsr::I<DataVector, dim, Frame::Grid>& grid_coords,
+      const double time,
+      const std::unordered_map<
+          std::string,
+          std::unique_ptr<::domain::FunctionsOfTime::FunctionOfTime>>&
+          functions_of_time,
+      const ::InverseJacobian<DataVector, Dim, Frame::ElementLogical,
+                              Frame::Grid>
```
nilsdeppe: `Frame::Grid>` -> `Frame::Grid>&`

The code below is ending up pointing to o temporary copy...

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -33,6 +35,122 @@ struct InverseJacobianLogicalToGrid : db::SimpleTag {
...
+          inverse_jacobian_logical_to_grid) {
+    destructive_resize_components(inverse_jacobian_logical_to_inertial,
+                                  get<0>(grid_coords).size());
+    if (grid_to_inertial_map.is_identity()) {
+      // Optimization for time-independent maps; we just point to the
+      // logical-to-grid inverse jacobian.
+      for (size_t storage_index = 0;
```
nilsdeppe, line 127: Add line before:
```
const size_t num_pts = inverse_jacobian_logical_to_grid[0].size();
```

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -33,6 +35,122 @@ struct InverseJacobianLogicalToGrid : db::SimpleTag {
...
+           storage_index < inverse_jacobian_logical_to_grid.size();
+           ++storage_index) {
+        auto& inv_jacobian_component =
```
nilsdeppe: Replace by:
```cpp
make_const_view(
    make_not_null(&std::as_const(
        (*inverse_jacobian_logical_to_inertial)[storage_index])),
    inverse_jacobian_logical_to_grid[storage_index], 0, num_pts);
```

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -33,6 +35,122 @@ struct InverseJacobianLogicalToGrid : db::SimpleTag {
...
+                              Frame::Grid>
+          inverse_jacobian_logical_to_grid) {
+    *det_inverse_jacobian_logical_to_grid =
+        determinant(inverse_jacobian_logical_to_grid);
+  }
+};
+
+/// Compute item for the inverse jacobian matrix from logical to
+/// inertial coordinates
+template <typename MapTagGridToInertial, size_t Dim>
```
nilsdeppe, line 99: Can you also add the test you have in the file `tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/Test_JacobianCompute.cpp` to the PR? I think we need to test our compute tags more for functionality. I started doing this for subcell and it helped find a lot of bugs...

nilsdeppe (changes requested):
You can squash immediately

src/Evolution/DgSubcell/Tags/Jacobians.hpp
ffoucart: Yeah... that'd do it.

ffoucart:
Added the new test and squashed. I don't understand the file check failure though.

nilsdeppe:
You can ignore the formatting error :)

ffoucart:
Ok, sounds good. The tests were otherwise all passing... but I just rebased against develop to see if it fixed the formatting. Will ignore it either way, and move on to rebase the second PR against this one.

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -33,6 +35,123 @@ struct InverseJacobianLogicalToGrid : db::SimpleTag {
...
+          grid_to_inertial_map,
+      const tnsr::I<DataVector, dim, Frame::Grid>& grid_coords,
+      const double time,
+      const std::unordered_map<
+          std::string,
+          std::unique_ptr<::domain::FunctionsOfTime::FunctionOfTime>>&
+          functions_of_time,
+      const ::InverseJacobian<DataVector, Dim, Frame::ElementLogical,
+                              Frame::Grid>& inverse_jacobian_logical_to_grid) {
+    destructive_resize_components(inverse_jacobian_logical_to_inertial,
```
nilsdeppe: This can go in the `else` and will silently conflict with https://github.com/sxs-collaboration/spectre/pull/5139 (just so we are aware)

tests/Unit/Evolution/DgSubcell/Test_JacobianCompute.cpp
```diff
@@ -0,0 +1,106 @@
...
+  // Check that the two jacobians in frames connected by an identity map are
+  // identical
+  for (size_t storage_index = 0; storage_index < inv_jac_inertial.size();
+       ++storage_index) {
+    CHECK_ITERABLE_APPROX(inv_jac_inertial[storage_index],
+                          inv_jac_grid[storage_index]);
+  }
+}
+
+// [[Timeout, 10]]
```
nilsdeppe: remove? I'd be surprised if this test takes that long :shrug:

src/Evolution/DgSubcell/Tags/Jacobians.hpp
```diff
@@ -22,8 +24,8 @@ namespace evolution::dg::subcell::fd::Tags {
 template <size_t Dim>
 struct InverseJacobianLogicalToGrid : db::SimpleTag {
   static std::string name() { return "InverseJacobian(Logical,Grid)"; }
-  using type = std::optional<
-      ::InverseJacobian<DataVector, Dim, Frame::ElementLogical, Frame::Grid>>;
+  using type = ::InverseJacobian < DataVector, Dim, Frame::ElementLogical,
```
nilsdeppe: Formatting: remove spaces around ` < ` and ` > `

nilsdeppe (commented):
LGTM, couple minor things and noting a silent conflict

ffoucart:
Last push addresses Nils' comment (except silent conflict). I also removed a
`#include <iostream>
`that I had accidentally left in.

nilsdeppe (dismissed):
LGTM, just need to coordinate with the other PR :)

ffoucart:
Hopefully merged with Will's PR...

nilsdeppe (approved):

ffoucart:
The tests that finished passed, but the unit tests took very long to run (and in two cases were cancelled due to running >6hrs). Is this likely to be a server issue or something that I should worry about?

nilsdeppe:
Nah, not your problem. I think Will's PR changed enough low-level things that we are hitting timeouts again :( We can merge this anyway

PR #5006, opened 2023-05-16
Enable Logarithmic Distribution in Envelope of BinaryCompactObject

### LINKED ISSUE #4695: Radial point distributions in frustum map
The BBH domain might benefit from a logarithmic or 1/r radial map in the bulged frustum. This can be added by applying the redistribution map to the radial logical direction before deforming it to the frustum. We can use the `domain::CoordinateMaps::Distribution` enum and possibly the `domain::CoordinateMaps::Interval` class.

mar-celine:
## Proposed changes

Addresses issue #3545 by allowing the user to specify a logarithmic distribution of points in the Blocks that use the Frustum map in the BinaryCompactObject Domain. Dependent on #4934 .

Fixes #4695 .

### Upgrade instructions

In input files that use the `BinaryCompactObject` domain creator, change the `Envelope.UseProjectiveMap` option to `Envelope.RadialDistribution`. Previous values of `True` and `False` correspond to `Projective` and `Linear` now.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/Frustum.cpp
+++ b/src/Domain/CoordinateMaps/Frustum.cpp
@@ -10,6 +10,7 @@
 #include "DataStructures/Tensor/EagerMath/Determinant.hpp"
 #include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/CoordinateMaps/Interval.hpp"
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/OrientationMap.hpp"
 #include "Domain/Structure/Side.hpp"
@@ -29,8 +30,8 @@ Frustum::Frustum(const std::array<std::array<double, 2>, 4>& face_vertices,
                  const double lower_bound, const double upper_bound,
                  OrientationMap<3> orientation_of_frustum,
                  const bool with_equiangular_map,
-                 const double projective_scale_factor,
-                 const bool auto_projective_scale_factor,
+                 const Distribution zeta_distribution,
+                 const std::optional<double> distribution_value,
                  const double sphericity, const double transition_phi,
                  const double opening_angle)
     // clang-tidy: trivially copyable
@@ -43,17 +44,15 @@ Frustum::Frustum(const std::array<std::array<double, 2>, 4>& face_vertices,
                                                              {{1.0, 1.0}}}} and
                    lower_bound == -1.0 and upper_bound == 1.0 and
                    orientation_of_frustum_ == OrientationMap<3>{} and
-                   not with_equiangular_map and projective_scale_factor == 1.0),
-      with_projective_map_(projective_scale_factor != 1.0),
+                   not with_equiangular_map and
+                   zeta_distribution == Distribution::Linear),
+      zeta_distribution_(zeta_distribution),
       sphericity_(sphericity) {
   ASSERT(sphericity_ >= 0.0 and sphericity_ <= 1.0,
          "The sphericity must be set between 0.0, corresponding to a flat "
          "surface, and 1.0, corresponding to a spherical surface, inclusive. "
          "It is currently set to "
              << sphericity << ".");
-  ASSERT(projective_scale_factor != 0.0,
-         "A projective scale factor of zero maps all coordinates to zero! Set "
-         "projective_scale_factor to unity to turn off projective scaling.");
   const double& lower_x_lower_base = face_vertices[0][0];
   const double& lower_y_lower_base = face_vertices[0][1];
   const double& upper_x_lower_base = face_vertices[1][0];
@@ -100,8 +99,6 @@ Frustum::Frustum(const std::array<std::array<double, 2>, 4>& face_vertices,
                               upper_y_lower_base + lower_y_lower_base);
   sigma_z_ = 0.5 * (upper_bound + lower_bound);
   delta_z_zeta_ = 0.5 * (upper_bound - lower_bound);
-  w_plus_ = projective_scale_factor + 1.0;
-  w_minus_ = projective_scale_factor - 1.0;
   phi_ = transition_phi;
   half_opening_angle_ = 0.5 * opening_angle;
   one_over_tan_half_opening_angle_ = 1.0 / tan(half_opening_angle_);
@@ -118,14 +115,23 @@ Frustum::Frustum(const std::array<std::array<double, 2>, 4>& face_vertices,
       square(std::max({abs(upper_y_upper_base), abs(upper_y_lower_base),
                        abs(lower_y_upper_base), abs(lower_y_lower_base)})) +
       square(std::max({abs(upper_bound), abs(lower_bound)})));
-  if (auto_projective_scale_factor) {
-    with_projective_map_ = true;
-    const double w_delta = sqrt(((upper_x_lower_base - lower_x_lower_base) *
+  inner_radius_ = sqrt(
+      square(std::max({abs(upper_x_lower_base), abs(lower_x_lower_base)})) +
+      square(std::max({abs(upper_y_lower_base), abs(lower_y_lower_base)})) +
+      square(lower_bound));
+
+  if (zeta_distribution_ == Distribution::Projective) {
+    double w_delta = distribution_value.has_value()
+                         ? distribution_value.value()
+                         : sqrt(((upper_x_lower_base - lower_x_lower_base) *
                                  (upper_y_lower_base - lower_y_lower_base)) /
                                 ((upper_x_upper_base - lower_x_upper_base) *
                                  (upper_y_upper_base - lower_y_upper_base)));
     w_plus_ = w_delta + 1.0;
     w_minus_ = w_delta - 1.0;
+  } else {
+    w_plus_ = 2.0;
+    w_minus_ = 0.0;
   }
 }
 
@@ -137,10 +143,25 @@ std::array<tt::remove_cvref_wrap_t<T>, 3> Frustum::operator()(
   const ReturnType& xi = source_coords[0];
   const ReturnType& eta = source_coords[1];
   const ReturnType& zeta = source_coords[2];
-  const ReturnType cap_zeta =
-      with_projective_map_
-          ? (w_minus_ + w_plus_ * zeta) / (w_plus_ + w_minus_ * zeta)
-          : zeta;
+  ReturnType cap_zeta;
+  if (zeta_distribution_ == Distribution::Projective) {
+    cap_zeta = (w_minus_ + w_plus_ * zeta) / (w_plus_ + w_minus_ * zeta);
+  } else if (zeta_distribution_ == Distribution::Linear) {
+    cap_zeta = zeta;
+  } else if (zeta_distribution_ == Distribution::Logarithmic) {
+    Interval log_mapping{
+        -1.0,
+        1.0,
+        -1.0,
+        1.0,
+        zeta_distribution_,
+        -(radius_ + inner_radius_) / (radius_ - inner_radius_)};
+    cap_zeta = log_mapping(std::array<T, 1>{{zeta}})[0];
+  } else {
+    ERROR(
+        "Only the distributions Linear, Projective, and Logarithmic are "
+        "supported.");
+  }
   const ReturnType cap_xi_zero = with_equiangular_map_ ? tan(M_PI_4 * xi) : xi;
   const double one_plus_phi_square = 1.0 + phi_ * phi_;
   const ReturnType cap_xi_upper =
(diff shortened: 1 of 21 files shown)


src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -219,6 +233,16 @@ std::optional<std::array<double, 3>> Frustum::inverse(
   if (zeta_distribution_ == Distribution::Projective) {
     logical_coords[2] = (-w_minus_ + w_plus_ * logical_coords[2]) /
                         (w_plus_ - w_minus_ * logical_coords[2]);
+  } else if (zeta_distribution_ == Distribution::Logarithmic) {
+    const double sigma_r = 0.5 * (inner_radius_ + radius_);
```
nilsvu: I suggest you give `domain::CoordinateMaps::Interval` a try. If you can use that to redistribute grid points on the logical axis before applying the frustum map then it would make the code here a lot simpler.

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -118,15 +114,21 @@ Frustum::Frustum(const std::array<std::array<double, 2>, 4>& face_vertices,
...
-  if (auto_projective_scale_factor) {
-    with_projective_map_ = true;
-    const double w_delta = sqrt(((upper_x_lower_base - lower_x_lower_base) *
-                                 (upper_y_lower_base - lower_y_lower_base)) /
-                                ((upper_x_upper_base - lower_x_upper_base) *
-                                 (upper_y_upper_base - lower_y_upper_base)));
+
+  double w_delta;
+  if (zeta_distribution_ == Distribution::Projective) {
+    w_delta = sqrt(((upper_x_lower_base - lower_x_lower_base) *
```
nilsvu: <s>move this to the `else` branch of the `if (distribution_value.has_value())` below.</s> done in a later commit

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -137,10 +139,12 @@ std::array<tt::remove_cvref_wrap_t<T>, 3> Frustum::operator()(
   const ReturnType& xi = source_coords[0];
   const ReturnType& eta = source_coords[1];
   const ReturnType& zeta = source_coords[2];
-  const ReturnType cap_zeta =
-      with_projective_map_
-          ? (w_minus_ + w_plus_ * zeta) / (w_plus_ + w_minus_ * zeta)
-          : zeta;
+  ReturnType cap_zeta;
```
nilsvu, line 146: (optional) a nice pattern to keep these variables `const` is to use a lambda:

```cpp
const ReturnType cap_zeta = [this, &zeta]() {
  if ... else ... return;
}();
```

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -298,10 +302,18 @@ tnsr::Ij<tt::remove_cvref_wrap_t<T>, 3, Frame::NoFrame> Frustum::jacobian(
...
+  ReturnType cap_zeta_deriv;
+  if (zeta_distribution_ == Distribution::Projective) {
+    cap_zeta = (w_minus_ + w_plus_ * zeta) / (w_plus_ + w_minus_ * zeta);
+    cap_zeta_deriv = (square(w_plus_) - square(w_minus_)) /
+                     square(w_plus_ + zeta * w_minus_);
```
nilsvu, line 342: Please run clang-format over this

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -435,10 +430,11 @@ Domain<3> BinaryCompactObject::create_domain() const {
...
   // centered at the origin.
   Maps maps_frustums = domain::make_vector_coordinate_map_base<
-      Frame::BlockLogical, Frame::Inertial, 3>(
-      frustum_coordinate_maps(length_inner_cube_, length_outer_cube_,
-                              use_equiangular_map_, {{-translation_, 0.0, 0.0}},
-                              projective_scale_factor_, 1.0, opening_angle_));
+      Frame::BlockLogical, Frame::Inertial, 3>(frustum_coordinate_maps(
+      length_inner_cube_, length_outer_cube_, use_equiangular_map_,
+      {{-translation_, 0.0, 0.0}},
+      domain::CoordinateMaps::Distribution::Projective,
```
nilsvu: <s>Always projective? There's still a `use_projective_map_` attribute. That should probably be deleted.</s> done in a later commit

src/Domain/DomainHelpers.hpp
```diff
@@ -196,8 +196,10 @@ std::vector<domain::CoordinateMaps::Frustum> frustum_coordinate_maps(
     double length_inner_cube, double length_outer_cube,
     bool use_equiangular_map,
     const std::array<double, 3>& origin_preimage = {{0.0, 0.0, 0.0}},
-    double projective_scale_factor = 1.0, double sphericity = 0.0,
-    double opening_angle = M_PI_2);
+    domain::CoordinateMaps::Distribution radial_distribution =
```
nilsvu, line 205: update the docs please

src/Domain/CoordinateMaps/Frustum.hpp
```diff
@@ -271,11 +272,10 @@ class Frustum {
    * \param orientation_of_frustum The orientation of the frustum in 3D space.
    * \param with_equiangular_map Determines whether to apply a tangent function
    * mapping to the logical coordinates (true) or not (false).
-   * \param projective_scale_factor The scale factor of the projective map
-   * applied to the zeta logical coordinate.
-   * \param auto_projective_scale_factor Determines whether to automatically
-   * calculate the projective scale factor based on the frustum geometry (true)
-   * or use the provided scale factor (false).
+   * \param zeta_distribution Whether to apply a linear, logarithmic, or
```
nilsvu, line 275: `radial_distribution`?

src/Domain/CoordinateMaps/Frustum.hpp
```diff
@@ -271,11 +272,10 @@ class Frustum {
...
+   * projective mapping to the logical zeta coordinate.
+   * \param distribution_value In the case of a projective map, the projective
```
nilsvu, line 277: keep the name `projective_scale_factor`?

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -538,7 +593,10 @@ bool operator==(const Frustum& lhs, const Frustum& rhs) {
          lhs.phi_ == rhs.phi_ and
          lhs.half_opening_angle_ == rhs.half_opening_angle_ and
          lhs.one_over_tan_half_opening_angle_ ==
-             rhs.one_over_tan_half_opening_angle_;
+             rhs.one_over_tan_half_opening_angle_ and
+         lhs.inner_radius_ == rhs.inner_radius_ and
+         lhs.radius_ == rhs.radius_ and
+         lhs.zeta_distribution_ == rhs.zeta_distribution_;
```
nilsvu: duplicate I think, also radius

Probably best to order these the same as the order of attributes in the class, if they aren't already

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -516,13 +567,17 @@ void Frustum::pup(PUP::er& p) {
     p | half_opening_angle_;
     p | one_over_tan_half_opening_angle_;
   }
+  if (version >= 2) {
+    p | inner_radius_;
+    p | zeta_distribution_;
```
nilsvu, line 581: In `version < 2` you have to unpack a bool where the `with_projective_map_` data was and set `zeta_distribution_` corresponding to that bool.

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -137,10 +143,25 @@ std::array<tt::remove_cvref_wrap_t<T>, 3> Frustum::operator()(
   const ReturnType& xi = source_coords[0];
   const ReturnType& eta = source_coords[1];
   const ReturnType& zeta = source_coords[2];
-  const ReturnType cap_zeta =
-      with_projective_map_
-          ? (w_minus_ + w_plus_ * zeta) / (w_plus_ + w_minus_ * zeta)
-          : zeta;
+  ReturnType cap_zeta;
+  if (zeta_distribution_ == Distribution::Projective) {
+    cap_zeta = (w_minus_ + w_plus_ * zeta) / (w_plus_ + w_minus_ * zeta);
```
nilsvu, line 148: Would it make sense to move this into the `Interval` map, so here in the frustum you don't have to switch between distributions at all? Then you can just do `cap_zeta = zeta_mapping(-1, 1, -1, 1, zeta_distribution, singularity_pos)`. So the `Interval` map handles the redistribution on the logical axis and the frustum just has to worry about the deformation of the logical cube.

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
```diff
@@ -54,7 +54,7 @@ DomainCreator:
         ExciseWithBoundaryCondition: Worldtube
       UseLogarithmicMap: false
     Envelope:
-      UseProjectiveMap: false
+      RadialDistribution: Logarithmic
```
nilsvu: Keep the old value of "Linear" or "Projective" I think (same in other input files)

nilsvu (changes requested):
The commits have quite a bit of overlap in the lines they change. Either sort that out please or just squash them together. The first commit with the added distribution is good, and if you add the projective rescaling to the `Interval` map to simplify the frustums then that would be good to keep in a separate commit as well.

src/Domain/CoordinateMaps/Frustum.hpp
mar-celine: I want to keep this as `zeta_distribution`, at least in the Frustum, because this Distribution map is applied to the zeta direction, which is not necessarily the radial direction (though they coincide in the BBH domain).

src/Domain/CoordinateMaps/Frustum.hpp
mar-celine: There will be another commit where the `distribution_value` is used to control the singularity point in the logarithmic map case, but this is not implemented yet.

src/Domain/CoordinateMaps/Frustum.cpp
mar-celine: The projective map is not currently written in such a way as to map from [A, B] to [a, b], so it can't be easily ported over to Interval right now. It only maps [-1,1] to [-1. 1] right now, but I can make this change and address this comment in a later PR (I do agree with the suggested change though)

src/Domain/CoordinateMaps/Frustum.cpp
nilsvu: ping

src/Domain/DomainHelpers.hpp
nilsvu: If you touch this file again, switch the docs to a list of `\param`s rather than the long paragraph that mentions each param.

src/Domain/CoordinateMaps/Frustum.hpp
nilsvu: ok

src/Domain/CoordinateMaps/Frustum.hpp
nilsvu: ok

src/Domain/CoordinateMaps/Frustum.cpp
nilsvu: radius is still duplicate

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -511,18 +562,28 @@ void Frustum::pup(PUP::er& p) {
     p | sphericity_;
     p | radius_;
     p | phi_;
+    if (version < 2) {
+      bool with_projective_map = false;
+      p | with_projective_map;
```
nilsvu: This won't work because the location of the bool in the data stream is after `is_identity_`, not here at the end. Move up to where you deleted `p | with_projective_map_`.

src/Domain/CoordinateMaps/Frustum.cpp
nilsvu: ok

src/Domain/CoordinateMaps/Frustum.cpp
mar-celine: clang-format does not change these lines when I try running it, is there some change I should make?

mar-celine:
@nilsvu I think now that the strong pup test is in this should be ready to be merged right?

nilsvu:
Could you do a quick rebase on develop again? #5135 just got merged, which makes the pup test a bit stronger and which should have caught the ordering issue.

nilsvu:
The domain versioning test is failing now. It could be #5145 or a bug in the pup function.

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -515,14 +572,38 @@ void Frustum::pup(PUP::er& p) {
...
+  if (version >= 2) {
+    p | inner_radius_;
+    p | zeta_distribution_;
+  } else {
+    // Only `Distribution::Logarithmic` uses `inner_radius`, which
+    // was not implemented until version 2, so setting inner_radius = NaN
+    // should not affect maps unpacked from old versions. `inner_radius` is
+    // computed from frustum vertices so it is not necessary to compare
+    // `inner_radius` values between maps in the equality operator. Note that
+    // this means that a Frustum will have `inner_radius` == 0.0 when unpacking
```
nilsvu: `== NaN`

tests/Unit/Helpers/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelpers.cpp
```diff
@@ -48,7 +48,7 @@ std::unique_ptr<DomainCreator<3>> worldtube_binary_compact_object(
       "    UseLogarithmicMap: false\n"
       "  Envelope:\n"
       "    Radius: 30.0\n"
-      "    UseProjectiveMap: false\n"
+      "    RadialDistribution: Projective\n"
```
nilsvu: linear

nilsvu (changes requested):

nilsvu (approved):

knelli2:
Fixes #4695.

PR #5009, opened 2023-05-16
Remove whitespace from registration names

### LINKED ISSUE #4944: Hash function changed somewhere around AppleClang 14 / M2 chips, breaks deserializing PUP::able objects from disk
The symptom of the problem is this:

- You serialize a domain to disk on a machine supported by CI (e.g. writing volume data to an H5 file).
- You try to deserialize the domain with AppleClang 14.0.2 on an M2 machine (I haven't tried other versions or configurations).
- You will get `Unrecognized PUP::able::PUP_ID` errors because the hashing function used by Charm++ to turn class name strings (from pretty_type in `register_classes_with_charm`) into IDs has changed.
- Also (but less importantly) you get different observation IDs when running the same simulation with the two different compiler versions, because they're hashes as well.

It's easy to reproduce: Just run any of the Python unit tests that deserialize the domain from the `VolTestData0.h5` in the repository.

nilsvu:
## Proposed changes

Fixes #4944.

This change breaks compatibility with data written by older versions of the code (and intends to prevent such compatibility issues in the future).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (dismissed):

wthrowe:
If we only care about GCC and Clang (and probably icc, but cppreference doesn't say), `typeid().name()` is probably quite stable, as the mangling procedure is specified as part of the [Itanium ABI](https://itanium-cxx-abi.github.io/cxx-abi/abi.html#typeid) (and libraries will break if it changes).  Not objecting to this method though.

nilsvu:
@wthrowe I'm undecided.. my reasoning is in the added docs here. The safest option is to implement our own mapping from types to registration names, but I don't think I can prioritize that right now.

nilsdeppe:
So if GCC/Clang change their mangling (not demangling), then every single Linux computer in the world will likely crash quite quickly... So that's basically not something they can do 🙃

nilsvu:
`typeid().name()` could return something other than the mangled name. It only happens to return that for gcc&clang.

nilsdeppe:
Yea, but what other compiler are we going to use?

nilsvu:
Dunno maybe ICC? Anyway I don't have a strong preference here, if you think the `typeid().name()` is more stable than `pretty_type::get_name()` then I switch it around 🤷‍♂️

nilsdeppe:
I believe ICC is effectively replaced by `dpc++` which actually uses clang. I believe ICC also does the same mangling and actually uses libstdc++ (ICC has to interoperate with GCC otherwise nobody could use it). So basically I think the only thing we can't support is MSVC.

I also don't have a strong preference immediately, I would guess the space-removed demangle and mangled names will be almost equally stable. We probably should just document that under the assumption of GCC/Clang/ICC using the mangled name would be stable and cross-compiler.

wthrowe:
Let's stick with `demangled - whitespace`.  It should be stable, since it's basically the source code.  Looking back at my SpEC work with this, there was a case where the mangling changed at some point (a bug fix, I think).  (Although who knows how boost would handle such a thing.)

nilsdeppe:
Alright, sounds good to me! Now it just needs to pass CI :D

nilsdeppe (approved):

PR #5017, opened 2023-05-19
Add AdamsMoultonPc time stepper

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Initialization/Evolution.hpp
+++ b/src/Evolution/Initialization/Evolution.hpp
@@ -175,8 +175,11 @@ struct TimeStepperHistory {
     // Will be overwritten before use
     dt_vars->initialize(mesh.number_of_grid_points());
 
+    // All steppers we have that need to start at low order require
+    // one additional point per order, so this is the order that
+    // requires no initial past steps.
     const size_t starting_order =
-        time_stepper.number_of_past_steps() == 0 ? time_stepper.order() : 1;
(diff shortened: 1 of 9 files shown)


knelli2:
Should you swap the order of the commits? As of right now, you add the new time stepper first but it isn't compatible with self start (based on your second commit message) so if the first commit is checked out, it seems like things would be broken.

nilsdeppe (commented):
I have no additional comments. @wthrowe just let me know what you think is best in terms of commit order and then we can merge :)

wthrowe:
I decided to squash them together.  Neither really makes sense without the other, and they don't modify the same files.

knelli2 (approved):
Have you tested this out on a full evolution yet?

wthrowe:
I tried it on a 1D scalar wave, but nothing more complicated.  The results look qualitatively correct.

In terms of performance (defined by counting RHS evaluations), at third order I see what I expect: with loose error tolerances (stability limited) it behaves comparably to an RK3 method (Rk3Owren), and with tight error tolerances (error limited) it's almost as good as Adams-Bashforth.  (In error-limited regimes, AB methods generally seem to outperform RK methods.)

At fifth order the step scaling doesn't seem to be working properly, but it doesn't seem to be working properly for AB5 either so that's probably an issue in either the error-control code or the common Adams-stepper logic.

nilsdeppe (approved):

PR #5019, opened 2023-05-19
Update RandomAmr to actually do random refinement

kidder:
Previously no refinement actions were being run; only the reduction done by CheckDomain was being tested.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Executables/Examples/RandomAmr/RandomAmr.hpp
+++ b/src/Executables/Examples/RandomAmr/RandomAmr.hpp
@@ -14,6 +14,7 @@
 #include "Domain/Creators/Factory2D.hpp"
 #include "Domain/Creators/Factory3D.hpp"
 #include "Domain/Creators/RegisterDerivedWithCharm.hpp"
+#include "Domain/Structure/ElementId.hpp"
 #include "Evolution/DiscontinuousGalerkin/DgElementArray.hpp"
 #include "Executables/Examples/RandomAmr/InitializeDomain.hpp"
 #include "Options/Options.hpp"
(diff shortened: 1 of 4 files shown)


tests/InputFiles/ExampleExecutables/RandomAmr1D.yaml
```diff
@@ -3,10 +3,18 @@
...
 Testing:
-  Check: parse;execute
+  Check: parse;execute_check_output
+ExpectedOutput:
+ - SpectreCheckpoint000000
+Timeout: 10
 
 ---
 
+Criteria:
```
knelli2, line 13: Not for this PR, but now that I see this in an input file, I think the yaml group should be called `AmrCriteria`. I think that would be clearer.

src/Executables/Examples/RandomAmr/RandomAmr.hpp
```diff
@@ -73,20 +96,44 @@ struct RandomAmrMetavars {
...
   using component_list =
-      tmpl::list<amr::Component<RandomAmrMetavars>, amr_element_component>;
+      tmpl::list<amr::Component<RandomAmrMetavars>, dg_element_array>;
 
   // NOLINTNEXTLINE(google-runtime-references)
   void pup(PUP::er& /*p*/) {}
 };
 
+template <typename Component>
+void register_amr_callbacks() {
```
nilsvu, line 115: Do you have plans to move this to a reusable place?

tests/InputFiles/ExampleExecutables/RandomAmr1D.yaml
nilsvu: I think just add an `Amr` option group and put this tag in there.

nilsvu (approved):

tests/InputFiles/ExampleExecutables/RandomAmr1D.yaml
kidder: okay, I'll do that in my next AMR PR

src/Executables/Examples/RandomAmr/RandomAmr.hpp
kidder: Yes, as I don't want to copy it into all the executables...  I'll do this in my next AMR PR as well

PR #5018, opened 2023-05-19
Remove template template from GHBase

knelli2:
## Proposed changes

Having the GH Base be a template template made a lot of things harder. This simplifies things significantly.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5024, opened 2023-05-20
Rename uses_partially_flat_cartesian_coordinates

Sizheng-Ma:
## Proposed changes

The evolution of the partially flat Cartesian coordinate system is only used by CCM (CCE never uses it). Thus it's reasonable to rename it to `evolve_ccm` to make its meaning clearer. This also makes future implementations easier.

This is dependent on PR #5018. Thanks to PR #5018, now the GH+CCE executable can be used to handle CCM as well, although the actual CCM actions are not ready yet.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Evolution/Executables/GeneralizedHarmonic/EvolveGeneralizedHarmonicCcm.hpp
```diff
@@ -73,13 +73,14 @@ class CProxy_GlobalCache;
 }  // namespace Parallel
 /// \endcond
```
knelli2: Instead of naming this file with `Ccm`, maybe rename to `EvolveGhAndCharacteristics` so that it's more general and can be both CCE and CCM.

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
```diff
@@ -12,16 +12,16 @@ function(add_generalized_harmonic_executable_without_horizon
...
 
-function(add_generalized_harmonic_cce_executable_without_horizon
-  EXECUTABLE_NAME INITIAL_DATA LIBS_TO_LINK)
+function(add_generalized_harmonic_ccm_executable_without_horizon
+  EXECUTABLE_NAME INITIAL_DATA EVOLVE_CCM LIBS_TO_LINK)
   add_spectre_parallel_executable(
-    "EvolveGhCce${EXECUTABLE_NAME}"
-    EvolveGeneralizedHarmonicCce
+    "EvolveGh${EXECUTABLE_NAME}"
+    EvolveGeneralizedHarmonicCcm
```
knelli2: These names are ok though because they are the actual executables and we want those to be Cce and Ccm

knelli2 (changes requested):

Sizheng-Ma:
Alright, done!

knelli2 (commented):
Looks good. Rebase and squash.

Sizheng-Ma:
> Looks good. Rebase and squash.

Done.

knelli2:
Can you remove the "fixup" comment from the second commit message?

Sizheng-Ma:
> Can you remove the "fixup" comment from the second commit message?

Okay, sounds good.

knelli2:
Sorry I wasn't clear. I meant remove the whole comment in the second line. Only have "Add ccm executable" in the commit message

Sizheng-Ma:
> Sorry I wasn't clear. I meant remove the whole comment in the second line. Only have "Add ccm executable" in the commit message

Oh, I see. Done.

knelli2 (dismissed):

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
```diff
@@ -86,14 +86,30 @@ add_spectre_parallel_executable(
   "${LIBS_TO_LINK};Cce;ControlSystem;GeneralRelativitySolutions;Importers"
 )
 
-add_generalized_harmonic_cce_executable_without_horizon(
-  GaugeWave
+add_generalized_harmonic_ccm_executable_without_horizon(
+  CceGaugeWave
```
nilsvu: I think this executable is not specific to a gauge wave since it chooses the initial data in the input file, right? In that case, remove the `GaugeWave` from the name.

nilsvu: I think there's no input file in the repo that compiles and tests a GH + CCE or CCM executable. Could you add one as an example, even if it only checks that the executable compiles and the input file can be parsed?

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhAndCharacteristics.hpp
```diff
@@ -73,13 +73,14 @@ class CProxy_GlobalCache;
 }  // namespace Parallel
 /// \endcond
 
-template <size_t VolumeDim, bool UseNumericalInitialData>
+template <size_t VolumeDim, bool UseNumericalInitialData, bool EvolveCcm>
 struct EvolutionMetavars
     : public GeneralizedHarmonicTemplateBase<VolumeDim,
                                              UseNumericalInitialData>,
-      public CharacteristicExtractDefaults<false> {
+      public CharacteristicExtractDefaults<EvolveCcm> {
```
nilsvu: Does this executable already do CCM or is something else needed, like adding the boundary conditions? I'm confused because nothing in this file really changed. If this can't do CCM yet then maybe wait with adding the executable in CMake?

src/Evolution/Systems/Cce/System.hpp
```diff
@@ -114,12 +114,12 @@
  */
 namespace Cce {
 
-template <bool EvolvePartiallyFlatCartesianCoordinates>
+template <bool EvolveCcm>
```
nilsvu, line 119: I'm a bit confused by this template parameter. The name suggests that in addition to CCE this would "evolve CCM" as well, but that doesn't make sense to me since the characteristic evolution is the same, right? Please add some documentation to this template parameter, and also refer to this documentation in other places where this template parameter occurs.

nilsvu (changes requested):

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
knelli2: These executables use too much memory on CI, thus we don't build them.

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhAndCharacteristics.hpp
knelli2: No it doesn't. Only some of the actions are added for CCM, but this just sets things up nicely for adding those new actions.

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
nilsvu: Ugh ok. Any ideas yet how to fix this?

src/Evolution/Systems/Cce/System.hpp
knelli2: Eventually the characteristic evolution will be able to do GH+CCM once we add those actions in. But we don't want to change any functionality of the stand alone CCE executable, so we add this template parameter to separate the two. Some docs explaining this could be nice

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
knelli2: @macedo22 found that option parsing took up a lot of compile-time memory and was working on something a while ago to alleviate that, but not sure how far she got. So as of right now, no.

src/Evolution/Systems/Cce/System.hpp
nilsvu: That makes a lot of sense in the context of the executable, but an `EvolveCcm` template parameter on the CCE system confuses me.

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
nilsvu: Keep in mind then that any changes could break the executable without us noticing.

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhAndCharacteristics.hpp
nilsvu: Ok sounds good

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
knelli2: Yes this has happened in the past. I'm aware of it and try my best to keep up with things

src/Evolution/Systems/Cce/System.hpp
knelli2: There's an additional evolved variable on the characteristic side for CCM

nilsvu: Yea from the name I understand it's another set of coordinates. I think just add some docs, then hopefully this will be clearer.

knelli2: You still need to add docs to this system. I'd just do

```
\note The template parameter `EvolveCcm` will add an extra evolved variable
to the characteristic system, namely `Cce::Tags::PartiallyFlatCartesianCoords`.
```

knelli2 (changes requested):

src/Evolution/Systems/Cce/System.hpp
Sizheng-Ma: > You still need to add docs to this system. I'd just do
> 
> ```
> \note The template parameter `EvolveCcm` will add an extra evolved variable
> to the characteristic system, namely `Cce::Tags::PartiallyFlatCartesianCoords`.
> ```

Alright, done! I also removed `Cce::ReducedWorldtubeDataManager` since it is outdated.

knelli2 (dismissed):
@nilsvu Could you take a final look?

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
```diff
@@ -79,6 +79,14 @@ add_spectre_parallel_executable(
   "EvolveGhCce"
   EvolveGhAndCharacteristic
   Evolution/Executables/GeneralizedHarmonic
-  "EvolutionMetavars<3>"
+  "EvolutionMetavars<3, false>"
+  "${LIBS_TO_LINK};Cce"
+)
+
+add_spectre_parallel_executable(
+  "EvolveCcm"
```
nilsvu: "EvolveGhCcm" for consistency?

nilsvu (changes requested):

src/Evolution/Executables/GeneralizedHarmonic/CMakeLists.txt
knelli2: Oh yeah probably a good idea. Missed that.

Sizheng-Ma: I'd prefer to rename `EvolveGhCce` to `EvolveCce` since CCE/CCM = GH + Characteristic. It's not reasonable to put GH together with either CCE or CCM. I've just added one more fixup commit.

knelli2: I don't like this. We have a standalone CCE executable that doesn't have anything to do with GH. I know it's called `CharacteristicExtract`, but between that and `EvolveCce` I have no clue what the difference is. `EvolveGhCce` and `EvolveGhCcm` make it clear that a GH evolution is happening here in addition to a characteristic evolution.

kidder: Also there are other formulations of the Einstein equations besides GH...

Sizheng-Ma: Alright. I've added `Gh` to both executables.

knelli2:
Looks good. Squash it in

Sizheng-Ma:
Alright, done!

nilsvu (approved):

PR #5027, opened 2023-05-22
Add size control diagnostics

knelli2:
## Proposed changes

First, rename everything that says "control signal" to "control error" because the "control signal" is what the Controller itself outputs and is what's used to update the Nth deriv of the function of time. Everything here is the "control error" which tells us how off we are from our desired value.

Then add a simple struct for gathering diagnostic information from the size control error calculation. This info will be printed out once I add the control error class.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size/Error.cpp
+++ b/src/ControlSystem/ControlErrors/Size/Error.cpp
@@ -1,7 +1,7 @@
 // Distributed under the MIT License.
 // See LICENSE.txt for details.
 
-#include "ControlSystem/ControlErrors/Size/Signal.hpp"
+#include "ControlSystem/ControlErrors/Size/Error.hpp"
 
 #include <algorithm>
 #include <cstddef>
@@ -27,7 +27,7 @@
 namespace control_system::size {
 
 template <typename Frame>
-double control_signal(
+double control_error(
     const gsl::not_null<Info*> info,
     const gsl::not_null<intrp::ZeroCrossingPredictor*> predictor_char_speed,
     const gsl::not_null<intrp::ZeroCrossingPredictor*>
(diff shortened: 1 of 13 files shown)


src/ControlSystem/ControlErrors/Size/Error.cpp
```diff
@@ -24,10 +24,12 @@
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 
+#include "Parallel/Printf.hpp"
+
```
wthrowe: Drop hunk.

src/ControlSystem/ControlErrors/Size/Error.hpp
```diff
@@ -24,6 +24,28 @@ class ZeroCrossingPredictor;
...
+  double dt_lambda_00;
+  double horizon_00;
+  double dt_horizon_00;
+  double control_error_delta_r;
+  double min_char_speed;
+  double min_comoving_char_speed;
+  double char_speed_crossing_time;
+  double comoving_char_speed_crossing_time;
+  double delta_r_crossing_time;
+  double control_error;
```
wthrowe: [optional] Stylistically, if this struct is the control error + diagnostic information, it might be clearer to have the control error at the start or end instead of in the middle of the diagnostics.

wthrowe (changes requested):

src/ControlSystem/ControlErrors/Size/Error.cpp
wthrowe: Ping.

knelli2: Oops, I removed it from the first commit, but it somehow moved to the second one

wthrowe (approved):

markscheel (approved):

PR #5028, opened 2023-05-22
Add a new mutator (re-)computing background metric variables on DG grid

yoonso0-0:
This replaces (and renames) the `evolution::initialization::GrTagsForHydro` action, and also can be used for both evolution and initialization phases. Many lines of the code are similar with the original `GrTagsForHydro` action.

Another PR for subcell will be submitted after testing a TOV star on a moving mesh.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Evolution/Actions/BackgroundGrVars.hpp
```diff
@@ -0,0 +1,184 @@
...
+ *  - Removes : nothing
+ *  - Modifies : nothing
+ *
+ */
+template <typename System, bool used_in_evolution_phase>
+struct BackgroundGrVars {
+  using simple_tags_from_options = tmpl::list<::Tags::Time>;
+
+  // Collect all the GR quantities used in the templated evolution system
+  using gr_variables_tag = ::Tags::Variables<tmpl::remove_duplicates<
```
yoonso0-0: This line (collection of _all_ the GR related tags) may be unnecessary, would be happy to get comments on this.

src/Evolution/Actions/BackgroundGrVars.hpp
```diff
@@ -0,0 +1,184 @@
...
+ private:
+  template <typename DbTagsList, typename T>
+  static void impl(const gsl::not_null<db::DataBox<DbTagsList>*> box,
+                   const T& solution_or_data) {
+    static constexpr size_t Dim = System::volume_dim;
+
+    using GrVars = typename gr_variables_tag::type;
+
+    if constexpr (used_in_evolution_phase) {
+      // Check if the mesh is actually moving i.e. block coordinate map is
```
kidder: what about the case where the solution is time-dependent, but the mesh is not moving?

yoonso0-0: That's a good point. Can we check if a solution is time-dependent?

kidder: hmm, I don't see a way quickly browsing.  For now just add a warning to the documentation of the action that it assumes the solution is time-independent, and create an issue to remind us to fix this at some point.  Probably doesn't make sense to do that until we can create initial data from the input file for all systems.

yoonso0-0: Okay. I also think this would not be a serious practical problem unless we want to use time-dependent GR solutions as a background, which I can't think of many for now?

src/Evolution/Actions/BackgroundGrVars.hpp
```diff
@@ -0,0 +1,184 @@
...
+ *  - Adds:
+ *    * system::spacetime_variables_tag
+ *    * system::flux_spacetime_variables_tag
+ *    * system::inverse_spatial_metric_tag
+ *  - Removes : nothing
+ *  - Modifies : nothing
+ *
+ */
+template <typename System, bool used_in_evolution_phase>
+struct BackgroundGrVars {
```
nilsdeppe: A few questions for our meeting later:

1. I think you can get rid of the `used_in_evolution_phase` and just check if the size of the variables are non-zero.
2. This seems to only be updating the active tags, but always using the GR coordinates. Naively I expect this to not work with subcell. I _think_ what works is if you forward declare the subcell tags and then do the `if constexpr (tag_is_retrievable) {` that'll work with just a forward decl. If not, we (I) need to figure out how to make that work, because it should. Then you can have an internal check basically for "Could I be doing subcell? If so, then check what the active grid is, get the right coordinates, etc. else just do the DG case"

So as-is, I expect the code to actually not work correctly because 1) the face tags aren't updated and 2) the cell-centered tags are always set to the DG grid. I think this code here is different from what you were sharing in our call, though.

yoonso0-0:
@nilsdeppe @kidder I changed the implementation of this from an action to a mutator.

src/Evolution/DiscontinuousGalerkin/BackgroundGrVars.hpp
```diff
@@ -0,0 +1,133 @@
...
+/*!
+ * \brief Allocate or assign background general relativity quantities needed
+ * for evolution systems run on a curved spacetime without solving Einstein
+ * equations (e.g. ValenciaDivclean, ForceFree).
+ *
+ * \warning This mutator assumes that the GR analytic data or solution
+ * specifying background spacetime metric is time-independent.
+ *
+ */
+template <typename System, typename Metavariables, bool using_runtime_id>
```
nilsdeppe: camel case `UsingRuntimeId`?

src/Evolution/DiscontinuousGalerkin/BackgroundGrVars.hpp
```diff
@@ -0,0 +1,133 @@
...
+      const gsl::not_null<GrVars*> background_gr_vars, const double time,
+      const Domain<Dim>& domain,
+      const tnsr::I<DataVector, Dim, Frame::Inertial>& inertial_coords,
+      const Mesh<Dim>& mesh, const Element<Dim>& element,
+      const T& solution_or_data) {
+    const size_t num_grid_pts = mesh.number_of_grid_points();
+
+    if (contains_allocations(*background_gr_vars)) {  // Evolution phase
+#ifdef SPECTRE_DEBUG
+      const size_t gr_vars_size = (*background_gr_vars).number_of_grid_points();
```
nilsdeppe: why not just inline this call since you only use the value once? Then you can get rid of the `ifdef`

src/Evolution/Executables/RelativisticEuler/Valencia/EvolveValencia.hpp
```diff
@@ -122,7 +123,7 @@ struct EvolutionMetavars {
       is_analytic_data_v<initial_data> xor is_analytic_solution_v<initial_data>,
       "initial_data must be either an analytic_data or an analytic_solution");
 
- using eos_base = typename EquationsOfState::get_eos_base<
+  using eos_base = typename EquationsOfState::get_eos_base<
```
nilsdeppe, line 127: remove indent?

nilsdeppe (changes requested):
You can squash immediately

src/Evolution/DiscontinuousGalerkin/BackgroundGrVars.hpp
```diff
@@ -0,0 +1,133 @@
...
+  template <typename T>
+  static void apply(
+      const gsl::not_null<GrVars*> background_gr_vars, const double time,
+      const Domain<Dim>& domain,
+      const tnsr::I<DataVector, Dim, Frame::Inertial>& inertial_coords,
+      const Mesh<Dim>& mesh, const Element<Dim>& element,
+      const T& solution_or_data) {
+    const size_t num_grid_pts = mesh.number_of_grid_points();
+
+    if (contains_allocations(*background_gr_vars)) {  // Evolution phase
```
wthrowe: Just a drive-by comment because I saw it in an email, but this seems like an extremely strange thing to branch on here.  I'm guessing you wanted `background_gr_vars->number_of_grid_points() != 0`.

src/Evolution/Executables/RelativisticEuler/Valencia/EvolveValencia.hpp
yoonso0-0: I believe this is an action of `clang-format` and actually corrects a missing whitespace

yoonso0-0:
Squashed and rebased on develop :)

nilsdeppe (approved):

nilsdeppe:
@kidder @wthrowe do either of you want to look to make sure you're happy too?

wthrowe:
I'm happy with how my comment was dealt with.  I don't think I need to do a full review.

kidder (approved):

Issue #5031, opened 2023-05-22
Worldtube tests are timing out


nilsvu:
Some worldtube tests repeatedly time out on CI:

- `Unit.Evolution.Systems.CurvedScalarWave.Worldtube.Tags`
- (couple more that I'll add when I see them again)

Probably their use of the BBH domain makes them quite slow. It would be good to remove the use of the BBH domain where the tested logic is independent of the domain.

nilsvu:
@nikwit the other test that's timing out is `Unit.Evolution.Systems.Worldtube.InitializeElementFacesGridCoordinates`

PR #5037, opened 2023-05-24
Speed up `InitializeElementFaces` test

nikwit:
removes the check for Gauss points and increases the time out to 10 secs (it takes 0.7 secs on my laptop in Debug)

addresses #5031


nilsvu (approved):
Cheers 🍻

PR #5045, opened 2023-05-25
Move YlmSpherepack into namespace

nilsvu:
This fixes segfaults I got when linking in SpEC libraries, probably because of collisions with SpEC's YlmSpherepack class. It's good to establish a namespace for spherical harmonic things anyway.

### Upgrade instructions

`YlmSpherepack` is `ylm::Spherepack` now.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2:
Is it possible to have SpEC in it's own namespace when using it in SpECTRE to avoid namespace collisions? I don't know much about linking like this or if it would even be possibly, but it seems more robust to not have to worry about changing our code when something is added in SpEC that could possibly conflict.

knelli2:
Is it possible to have SpEC in it's own namespace when using it in SpECTRE to avoid namespace collisions? I don't know much about linking like this or if it would even be possibly, but it seems more robust to not have to worry about changing our code when something is added in SpEC that could possibly conflict.

nilsvu:
Not really I think. You could include the SpEC headers within a namespace, but not sure how robust that is.

markscheel:
Looks good, but is `sh` a long enough name that it is understandable what it means?  Some abbreviations (like `gr`) are obvious, but maybe `spherical_harmonics` or `ylm` might be better.  If everyone else is happy with `sh` then I am ok with it too, but I want to make sure we aren't overlooking a better name.

nilsvu:
I'm not particularly happy about `sh` but I can't think of anything better. I think `spherical_harmonics` is too long for a top-level namespace. Other possible abbreviations are `sph` (can be confused with "smooth particle hydro"), `spherical`, `sphharm`. None a particularly great.

knelli2:
I think `sh` is fine

knelli2 (approved):

nilsdeppe (approved):

PR #5046, opened 2023-05-25
Account for the possibility of Subcell in BNSCenterOfMass

ffoucart:
## Proposed changes

The BNSCenterOfMass measurement did not account for the fact that mesh and coordinates might need to be retrieved on the subcell grid.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Measurements/BNSCenterOfMass.hpp
+++ b/src/ControlSystem/Measurements/BNSCenterOfMass.hpp
@@ -16,6 +16,7 @@
 #include "Domain/Structure/ObjectLabel.hpp"
 #include "NumericalAlgorithms/LinearOperators/DefiniteIntegral.hpp"
 #include "Parallel/Reduction.hpp"
+#include "ParallelAlgorithms/Events/Tags.hpp"
 #include "Time/Tags.hpp"
 #include "Utilities/ProtocolHelpers.hpp"
 #include "Utilities/StdArrayHelpers.hpp"
@@ -37,10 +38,6 @@ struct Mesh;
 template <size_t Dim, typename Frame>
 struct Coordinates;
 }  // namespace domain::Tags
-namespace Events::Tags {
-template <typename InputFrame, typename OutputFrame>
-struct ObserverDetInvJacobian;
-}
 namespace grmhd::ValenciaDivClean {
 struct TildeD;
 }
@@ -88,6 +85,8 @@ void center_of_mass_integral_on_element(
 /// Measurement providing the location of the center of mass of the
 /// matter in the \f$x>0\f$ and \f$x<0\f$ regions (assumed to correspond to the
 /// center of mass of the two neutron stars in a BNS merger).
+/// We use Events::Tags::ObserverXXX for tags that might need to be retrieved
+/// from either the Subcell or DG grid.
 struct BothNSCenters : tt::ConformsTo<protocols::Measurement> {
   struct FindTwoCenters : tt::ConformsTo<protocols::Submeasurement> {
     static std::string name() { return "BothNSCenters::FindTwoCenters"; }
@@ -97,11 +96,11 @@ struct BothNSCenters : tt::ConformsTo<protocols::Measurement> {
 
     /// Tags for the arguments to the apply function.
     using argument_tags =
-        tmpl::list<domain::Tags::Mesh<3>,
+        tmpl::list<Events::Tags::ObserverMesh<3>,
                    Events::Tags::ObserverDetInvJacobian<Frame::ElementLogical,
                                                         Frame::Inertial>,
                    grmhd::ValenciaDivClean::TildeD,
-                   domain::Tags::Coordinates<3, Frame::Distorted>>;
+                   Events::Tags::ObserverCoordinates<3, Frame::Distorted>>;
 
     /// Calculate integrals needed for CoM computation on each element,
     /// then reduce the data.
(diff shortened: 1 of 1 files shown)


ffoucart:
@knelli2 : I think this is what is needed for the BNS control system to make measurements on subcell grids. The same method worked for my observers at least...

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -14,6 +14,10 @@
 #include "DataStructures/LinkedMessageId.hpp"
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "Domain/Structure/ObjectLabel.hpp"
+#include "Evolution/DgSubcell/ActiveGrid.hpp"
+#include "Evolution/DgSubcell/Tags/ActiveGrid.hpp"
+#include "Evolution/DgSubcell/Tags/Coordinates.hpp"
+#include "Evolution/DgSubcell/Tags/Mesh.hpp"
```
nilsvu: This will depend the ControlSystem library, and hence also BBH simulations, on Subcell. I think we have to find a way around that. Maybe base tags or something like that, as @knelli2 suggested? @nilsdeppe any thoughts?

nilsvu (changes requested):

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -96,26 +103,39 @@ struct BothNSCenters : tt::ConformsTo<protocols::Measurement> {
...
-                   Events::Tags::ObserverDetInvJacobian<Frame::ElementLogical,
-                                                        Frame::Inertial>,
-                   grmhd::ValenciaDivClean::TildeD,
-                   domain::Tags::Coordinates<3, Frame::Distorted>>;
+    using argument_tags = tmpl::list<
+        domain::Tags::Mesh<3>, evolution::dg::subcell::Tags::Mesh<3>,
+        Events::Tags::ObserverDetInvJacobian<Frame::ElementLogical,
+                                             Frame::Inertial>,
+        grmhd::ValenciaDivClean::TildeD,
+        domain::Tags::Coordinates<3, Frame::Distorted>,
```
kidder: just use the Event::Tags::ObserverCoordinates (you are already using the DetInv one...)

kidder (changes requested):

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
kidder: and Event::Tags::ObserverMesh

knelli2:
So I think the three options we have are

1. Implement a consistent base/reference tag structure so devs don't have to constantly think about which mesh they are on when they want coordinates.
2. Switch to observer stuff like @kidder suggested.
3. Move this measurement out of the control system library potentially.

1 is the "right" way to do it, but that may be a lot of work. I'd be nice if all control system stuff was in the ControlSystem library and not spread out, so we'd only do 3 if we didn't have other options. I think 2 is the easiest solution as of right now.

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -88,6 +92,9 @@ void center_of_mass_integral_on_element(
 /// Measurement providing the location of the center of mass of the
 /// matter in the \f$x>0\f$ and \f$x<0\f$ regions (assumed to correspond to the
 /// center of mass of the two neutron stars in a BNS merger).
+/// Note that TildeD is on the currently active grid (Subcell or Df)
```
knelli2: `DG`

ffoucart:
That's good for me. I didn't know we had those Event tags for mesh and
coordinates. Is there a limitation to where they can be used?

nilsvu:
They can be used only in observations, not in timestepping code. They live in the observation box, which gets created from the databox when we run events and triggers.

ffoucart:
Is the control system measurement considered an observation in that respect?

knelli2:
@ffoucart The `apply` function that you are editing is called during an event so the box that you get the tags from is actually an observation box. So yes.

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -16,6 +16,7 @@
 #include "Domain/Structure/ObjectLabel.hpp"
 #include "NumericalAlgorithms/LinearOperators/DefiniteIntegral.hpp"
 #include "Parallel/Reduction.hpp"
+#include "ParallelAlgorithms/Events/Tags.hpp"
```
ffoucart, line 19: Do we want to instead declare all Events::Tags::ObserverXXX tags in the header?

knelli2: Up to you. We already have to depend on the `Events` library for `Event.hpp` in the control system so it doesn't really matter.

ffoucart: Ok, great; I'm happy with it as is.

ffoucart:
I think this modified version should account for the suggestions made yesterday evening.

knelli2 (approved):

kidder (approved):

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
nilsdeppe: Use the `EventsAndTriggers` tags and an observation box? That's the easiest way to decouple consumption from addition with DataBox

ffoucart:
@nilsvu : Was the change to using the Events tags answering your comment?

nilsvu (approved):

PR #5051, opened 2023-05-26
Use correct mesh and jacobian for integrals with subcell

ffoucart:
## Proposed changes

Modify ObserveVolumeIntegrals so that it uses the Dg or Subcell mesh as appropriate.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Events/ObserveVolumeIntegrals.hpp
+++ b/src/ParallelAlgorithms/Events/ObserveVolumeIntegrals.hpp
@@ -119,12 +119,12 @@ class ObserveVolumeIntegrals<
   using compute_tags_for_observation_box =
       tmpl::list<Tensors..., NonTensorComputeTags...>;
 
-  using argument_tags = tmpl::list<
-      ::Tags::ObservationBox, ObservationValueTag,
-      ::Events::Tags::ObserverMesh<VolumeDim>,
-      ::Events::Tags::ObserverDetInvJacobian<Frame::ElementLogical,
-                                           Frame::Inertial>,
-      Tensors...>;
+  using argument_tags =
+      tmpl::list<::Tags::ObservationBox, ObservationValueTag,
+                 ::Events::Tags::ObserverMesh<VolumeDim>,
+                 ::Events::Tags::ObserverDetInvJacobian<Frame::ElementLogical,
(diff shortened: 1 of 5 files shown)


knelli2 (approved):

knelli2 (changes requested):
I don't think all executables have the `Events::Tags::ObserverXXX` tags implemented. What if you didn't put these tags in the `argument_tags` list, but instead only had the observation box. Then inside the function, check if the observation box has the event tags; if it does, get that, if not, use the DG one?

ffoucart:
I guess it's that or updating all relevant executables... as this is an observer, ideally any executable that calls it could presumably use the more generic Events::Tag::XXX tags. I'm happy to go either way.

knelli2:
@ffoucart I have no preference. Whichever is less work for you.

Edit: Actually, if you change all the executables, that makes life easier for #5053.

src/ParallelAlgorithms/Events/ObserveVolumeIntegrals.hpp
```diff
@@ -120,8 +121,9 @@ class ObserveVolumeIntegrals<
 
   using argument_tags = tmpl::list<
       ::Tags::ObservationBox, ObservationValueTag,
-      domain::Tags::Mesh<VolumeDim>,
-      domain::Tags::DetInvJacobian<Frame::ElementLogical, Frame::Inertial>,
+      ::Events::Tags::ObserverMesh<VolumeDim>,
+      ::Events::Tags::ObserverDetInvJacobian<Frame::ElementLogical,
+                                           Frame::Inertial>,
```
nilsvu: Could you run clang-format over this? What I usually do is run git-clang-format when I have staged changes, or something like `git-clang-format HEAD~1` to format the last commit.

nilsvu (changes requested):

nilsvu:
There's a CI error. I think you have to add the ObserverDetInvJac tag to some executables.

ffoucart:
Yes; that's related to Kyle's comment, but I have not had the chance to fix it yet (yesterday was a holiday in the US)

nilsvu (approved):

PR #5052, opened 2023-05-26
Add RAII class for saving and restoring FPE state

wthrowe:
We still can't reliably recover from an FPE because throwing exceptions from a signal handler isn't well supported, but this should make FPE-handling preserved when other errors occur.  It will also prevent code from unintentionally enabling FPEs if it temporarily disables them when they were already disabled.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
+++ b/src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
@@ -7,6 +7,7 @@
 #include <cstddef>
 #include <optional>
 #include <stdexcept>
+#include <tuple>
 
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tags/TempTensor.hpp"
@@ -210,7 +211,9 @@ void strahlkorper_in_different_frame(
   };
 
   // We try to roughly bracket the root between r_min and r_max.
-  const auto [r_min, r_max] = [&dest_cartesian_coords, &center_dest]() {
+  double r_min{};
+  double r_max{};
+  std::tie(r_min, r_max) = [&dest_cartesian_coords, &center_dest]() {
     const DataVector dest_radius =
         sqrt(square(get<0>(dest_cartesian_coords) - center_dest[0]) +
              square(get<1>(dest_cartesian_coords) - center_dest[1]) +
--- a/src/Domain/CoordinateMaps/UniformCylindricalEndcap.cpp
+++ b/src/Domain/CoordinateMaps/UniformCylindricalEndcap.cpp
@@ -8,6 +8,7 @@
 #include <optional>
 #include <pup.h>
 #include <sstream>
+#include <tuple>
 #include <utility>
 
 #include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
@@ -43,7 +44,7 @@ double this_function_is_zero_for_correct_rhobar(
 }
 
 // min and max values of rhobar in the inverse function.
-std::array<double, 2> rhobar_min_max(
+std::tuple<double, double> rhobar_min_max(
     const std::array<double, 3>& center_one,
     const std::array<double, 3>& center_two, const double radius_one,
     const double radius_two, const double theta_max_one,
@@ -432,8 +433,9 @@ std::optional<std::array<double, 3>> UniformCylindricalEndcap::inverse(
         return std::make_pair(func_to_zero, deriv_of_func_to_zero);
       };
 
-  // Non-const because might be changed below.
-  auto [rhobar_min, rhobar_max] =
+  double rhobar_min{};
+  double rhobar_max{};
+  std::tie(rhobar_min, rhobar_max) =
       rhobar_min_max(center_one_, center_two_, radius_one_, radius_two_,
                      theta_max_one_, theta_max_two_, target_coords);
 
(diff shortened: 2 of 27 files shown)


wthrowe:
I think this is working now, but the gcc-11-release CI build keeps timing out, presumably because ccache is useless when you touch something used by ERROR and ASSERT.  Might still be a problem, though, since we can't populate the cache if we can't build.

tests/Unit/ErrorHandling/Test_FloatingPointExceptions.cpp
```diff
@@ -81,3 +87,111 @@ SPECTRE_TEST_CASE("Unit.ErrorHandling.FloatingPointExceptions.Disable",
...
+#ifdef __arm64__
+  CHECK(true);
+  return;
+  // Don't stick the rest of the function in an #else or something
+  // because it should still compile.
+#pragma GCC diagnostic ignored "-Wunreachable-code"
+#endif
+#endif
+
+  // Stack unwinding from out FPE handler doesn't work on all
```
kidder: out -> our?

tests/Unit/ErrorHandling/Test_FloatingPointExceptions.cpp
```diff
@@ -81,3 +87,111 @@ SPECTRE_TEST_CASE("Unit.ErrorHandling.FloatingPointExceptions.Disable",
...
+      {
+        const ScopedFpeState s3(true);
+        CHECK(exceptions_enabled());
+      }
+      CHECK(not exceptions_enabled());
+    }
+    CHECK(not exceptions_enabled());
+  }
+  CHECK(exceptions_enabled());
+}
```
kidder, line 218: can you add CHECK_THROWS that test the ASSERTs?

src/ApparentHorizons/StrahlkorperInDifferentFrame.cpp
```diff
@@ -210,7 +210,9 @@ void strahlkorper_in_different_frame(
   };
 
   // We try to roughly bracket the root between r_min and r_max.
-  const auto [r_min, r_max] = [&dest_cartesian_coords, &center_dest]() {
+  double r_min{};
+  double r_max{};
+  std::tie(r_min, r_max) = [&dest_cartesian_coords, &center_dest]() {
```
kidder, line 216: include <utility> for std::tie

src/Domain/CoordinateMaps/UniformCylindricalEndcap.cpp
```diff
@@ -43,7 +43,7 @@ double this_function_is_zero_for_correct_rhobar(
 }
 
 // min and max values of rhobar in the inverse function.
-std::array<double, 2> rhobar_min_max(
+std::tuple<double, double> rhobar_min_max(
```
kidder, line 47: include <tuple>

src/Domain/CoordinateMaps/UniformCylindricalFlatEndcap.cpp
```diff
@@ -44,7 +44,7 @@ double this_function_is_zero_for_correct_rhobar(
 }
 
 // min and max values of rhobar in the inverse function.
-std::array<double, 2> rhobar_min_max(
+std::tuple<double, double> rhobar_min_max(
```
kidder, line 48: include tuple

tests/Unit/Framework/SetupLocalPythonEnvironment.cpp
```diff
@@ -70,9 +70,9 @@ SetupLocalPythonEnvironment::SetupLocalPythonEnvironment(
 
     // On some python versions init_numpy() can throw an FPE, this occurred at
     // least with python 3.6, numpy 1.14.2.
-    disable_floating_point_exceptions();
+    ScopedFpeState disable_fpes(false);
     init_numpy();
-    enable_floating_point_exceptions();
+    disable_fpes.restore_exceptions();
```
kidder, line 75: why do you need to explicitly call restore here?

kidder (changes requested):
you can squash any changes

tests/Unit/ErrorHandling/Test_FloatingPointExceptions.cpp
wthrowe: I assume I intended "our", but it's not really related to anything particular to us, so I changed the phrase to "from a signal handler".

tests/Unit/Framework/SetupLocalPythonEnvironment.cpp
wthrowe: I don't really, but it felt like the other uses were intending to disable for the entire function while this was intending to do so only for the call to `init_numpy()`, so I added this so it would work as intended if more code was added below here.

tests/Unit/ErrorHandling/Test_FloatingPointExceptions.cpp
wthrowe: Added at the start because that part should work on ARM builds.

wthrowe (commented):
I also renamed the variables in macros to avoid potential name clashes.

tests/Unit/Framework/SetupLocalPythonEnvironment.cpp
kidder: okay

kidder (commented):
looks good, squash

kidder (approved):

kidder (changes requested):
The release builds are failing because these are checking that asserts are triggered.
I believe you need to surround them with a `#ifdef SPECTRE_DEBUG`

wthrowe:
Derp.  Also rebased for the db::mutate change.

kidder:
ignoring clang-tidy Boost failure, and build time out

PR #5058, opened 2023-05-30
Build Python docs with Sphinx

nilsvu:
## Proposed changes

Generate documentation for Pybindings and Python code, then link to it from our Doxygen documentation.

Looks something like this:

Bildschirmfoto 2023-05-30 um 10 45 34

Bildschirmfoto 2023-05-31 um 09 51 14

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


cmake/SpectreSetupPythonPackage.cmake
```diff
@@ -249,6 +243,17 @@ function(SPECTRE_PYTHON_ADD_MODULE MODULE_NAME)
       "${MODULE_LOCATION}/${PYTHON_FILE_JUST_NAME}"
       )
   endforeach(PYTHON_FILE ${ARG_PYTHON_FILES})
+
+  # Create empty __init__.py files if none exist
+  # We walk up the tree until we get to ${SPECTRE_PYTHON_PREFIX}
+  set(CURRENT_MODULE ${MODULE_LOCATION})
```
nilsdeppe, line 249: Are these written into the source tree or the build directory? I found when we switched from auto `__init__.py` to manual `__init__.py` in the source tree it was annoying to deal with switching branches. Should we have a bash script that adds these and checks for them in PRs?

nilsdeppe (changes requested):
LGTM, just one question/suggestion.

cmake/SpectreSetupPythonPackage.cmake
nilsvu: Build directory! So this shouldn't be a problem.

The branch switching back then was only annoying if you had `PY_DEV_MODE` enabled, because then the init.py files are symlinks.

knelli2: So it shouldn't be a problem even in `PY_DEV_MODE` now?

nilsvu: Not unless you have a branch where you add an init.py file in an intermediate directory in the Py package, like in PointwiseFunctions/.

nilsdeppe: I'm fine with this, but just to clarify your last statement: If I add new init files in the _src_ tree, then they may clash with the auto-generated ones in the _build_ dir?

nilsvu: Actually on second thought I think the issue with init.py files can't happen anymore at all because CMake doesn't overwrite the file anymore (there's an `EXISTS` check).

docs/_templates/autosummary/module.rst
```diff
@@ -0,0 +1,14 @@
+{% extends "!autosummary/module.rst" %}
+
+.. Distributed under the MIT License.
+   See LICENSE.txt for details.
+
+.. Include pybindings documentation. This won't do anything for modules that
+   have no pybindings but print a warning.
```
knelli2, line 7: There's a lot of warnings. I get them locally and they are on CI. Is there a way to silence these?

docs/index.rst
```diff
@@ -1,65 +1,43 @@
...
 
-   <a
-   href="https://github.com/sxs-collaboration/spectre/blob/develop/LICENSE.txt"><img
-   src="https://img.shields.io/badge/license-MIT-blue.svg"
-   alt="license"
-   data-canonical-src="https://img.shields.io/badge/license-MIT-blue.svg"
-   style="max-width:100%;"></a>
+The SpECTRE Python interface exposes some SpECTRE functionality in Python. For
+example, it helps with visualizing data, scheduling simulations, and working
+with simulation output. To get started quickly, compile the ``all-pybindings``
```
knelli2, line 9: "Or `cli`"

docs/conf.py
```diff
@@ -50,27 +54,32 @@
...
-# Tell sphinx what the pygments highlight language should be.
-highlight_language = "cpp"
-
 # -- Options for HTML output -------------------------------------------------
 
 # The theme to use for HTML and HTML Help pages.  See the documentation for
 # a list of builtin themes.
 #
-html_theme = "sphinx_rtd_theme"
+html_theme = "furo"
```
knelli2, line 62: Personally, I like the RTD theme better.

knelli2 (changes requested):
image

I think this is a good start! Here's a screenshot of what a Read-The-Docs theme looks like.

Could you also add Upgrade Instructions for how to build these docs?

docs/_templates/autosummary/module.rst
nilsvu: Can't silence these warnings because there's a hard-coded print in the sphinx code: https://github.com/sphinx-doc/sphinx/blob/master/sphinx/ext/autosummary/generate.py#L535

Lesson: always use `logging`

docs/conf.py
nilsvu: I chose this one because it's quite popular (even used by pip) and similar in style to our C++ docs. Can't argue with personal preference though :)

docs/_templates/autosummary/module.rst
knelli2: Rip...Guess we just need more pybindings then.

docs/conf.py
nilsvu: Also dark mode is a must ;)

docs/_templates/autosummary/module.rst
nilsvu: I could remove _all_ output, but that's probably not a good solution either

knelli2: Yeah no we should keep the output

knelli2 (dismissed):
Ok I'm happy with this as a first pass.

wthrowe (changes requested):
Haven't looked at the diff at all, but the documentation CI job is compiling a bunch of code.  I assume a local render would also do that.  Can we do anything about that?  Documentation generation is slow enough as it is.

knelli2:
@wthrowe Pybindings have to be built in order to generate the python docs. We could add another CI build that only does python docs?

wthrowe:
I'm not so much worried about CI as local generation.  If it takes me 15 minutes to render the docs that's kind of a problem.

nilsdeppe:
Possible avenues to think about: 

1. have a `doc-python` target that builds the python code and docs while `doc` and `check-doc` ignore the python docs? This would allow people to choose what to build locally, while in CI we of course are somewhat stuck with wanting complete docs.
2. I'm not sure if there's a way to generate stub files that would be enough for Sphinx. That might be an option.

I haven't looked into anything in detail, just wanted to give two possible solutions!

nilsvu:
Option 1. is how it currently works. The target is named `py-docs`. @wthrowe you don't have to build that target locally.

wthrowe:
OK, that's fine.

Next question: I can't download the artifact until all of CI is done, but based on the summary at the end it looks like most of the binding documentation is missing.  It only lists IO, NumericalAlgorithms, and some stuff in PointwiseFunctions.

nilsvu:
@wthrowe pushed a fixup. You can also download the artifact now. looks fine to me.

wthrowe (approved):
Output looks fine now.  Only skimmed the code changes.

nilsdeppe (approved):

PR #5060, opened 2023-05-30
Add SpEC ID to BBH and GHMHD executables

nilsvu:
## Proposed changes

This will compile and run, but won't work properly until SpEC thread-safety issues have been worked out.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
+++ b/src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
@@ -153,9 +153,18 @@
 #include "Utilities/ErrorHandling/FloatingPointExceptions.hpp"
 #include "Utilities/ErrorHandling/SegfaultHandler.hpp"
 #include "Utilities/Functional.hpp"
+#include "Utilities/NoSuchType.hpp"
 #include "Utilities/ProtocolHelpers.hpp"
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
 #include "Utilities/TMPL.hpp"
+
+// Check if SpEC is linked and therefore we can load SpEC initial data
+#ifdef HAS_SPEC_EXPORTER
+#include "PointwiseFunctions/AnalyticData/GeneralRelativity/SpecInitialData.hpp"
+using SpecInitialData = gr::AnalyticData::SpecInitialData;
+#else
+using SpecInitialData = NoSuchType;
+#endif
 
 /// \cond
 namespace Frame {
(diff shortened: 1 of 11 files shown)


src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
```diff
@@ -157,6 +157,14 @@
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
 #include "Utilities/TMPL.hpp"
 
+// Check if SpEC is linked and therefore we can load SpEC initial data
+#ifdef HAS_SPEC_EXPORTER
+#include "PointwiseFunctions/AnalyticData/GeneralRelativity/SpecInitialData.hpp"
+using SpecInitialData = gr::AnalyticData::SpecInitialData;
+#else
+using SpecInitialData = void;
```
nilsdeppe: What do you think about using `Utilitias/NoSuchType.hpp` instead of `void`? A bit clearer maybe?

src/Evolution/Initialization/InitialData.hpp
```diff
@@ -24,7 +24,7 @@ decltype(auto) initial_data(const SolutionOrData& solution_or_data,
 
 /// \cond
 template <typename SolutionOrData, typename Coordinates, typename... Tags,
-          Requires<is_analytic_data_v<SolutionOrData>> = nullptr>
+          Requires<not is_analytic_solution_v<SolutionOrData>> = nullptr>
```
nilsdeppe: Why not `is_analytic_data_v or is_numerical_data_v`? That would be easier to read IMO and probably safer?

src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
```diff
@@ -213,6 +213,14 @@
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
 #include "Utilities/TMPL.hpp"
 
+// Check if SpEC is linked and therefore we can load SpEC initial data
+#ifdef HAS_SPEC_EXPORTER
+#include "PointwiseFunctions/AnalyticData/GrMhd/SpecInitialData.hpp"
+using SpecInitialData = grmhd::AnalyticData::SpecInitialData;
+#else
+using SpecInitialData = void;
```
nilsdeppe: `void` -> `NoSuchType`?

nilsdeppe (changes requested):
You can squash anything you decide to do

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
nilsvu: I have a (very mild) preference for `void` because it's standard C++ and IMO quite clear

src/Evolution/Initialization/InitialData.hpp
nilsvu: done

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
nilsdeppe: I guess I find the opposite for the exact same reason. `void` can mean different things and in C++17 is (in some context) a valid type, while `NoSuchType` is meant to signify that some compile-time thing went wrong.  I think the confusing times is when you hit `void` in a `tmpl` context and then the question is "Is void good or bad?"

nilsvu: Fine with me, changed to `NoSuchType`

nilsdeppe (approved):

Issue #3266, opened 2021-06-11
Metavariables for phase control and registration need documentation


nilsvu:
# Bug reports:

### Expected behavior:

From reading the metavariables, e.g. for `EvolveScalarWave`, I understand which type aliases must be present and will be retrieved from somewhere further down in the code, and which are just there for convenience. Also, for the required type aliases I can find documentation to understand what they do. Ideally I find this documentation by following "breadcrumbs" in the code.

### Current behavior:

The metavars have type aliases related to phase control and registration that are not documented (or if docs exists, I don't find them from reading through the metavars. I have also tried searching in `*.md` files with no success.). Here's the ones I noticed: `phase_changes`, `initialize_phase_change_decision_data`, `phase_change_tags_and_combines_list`, `dg_registration_list`, `registration_list`.

### Suggested solution:

My suggestion is to follow the example of the `factory_creation` metavars for documenting these items.

nilsvu:
@moxcodes could you take this on when you find the time? I'm quite open for suggestions if you have ideas how to structure this sort of documentation. I like the way we have pioneered this for the `factory_creation` metavars though.

moxcodes:
Yep, it's a thing on my to-do list to write some documentation for phase control infrastructure generally (likely for the dev-guide), but I also think it would be valuable to follow the sub-struct model for grouping and documenting the metavariables options specifically.

nilsvu:
Ok great, thanks!

PR #5065, opened 2023-06-01
Allow adaptive mesh refinement for ExportCoordinates

kidder:
## Proposed changes

- Enables AMR when running ExportCoordinates
   - This only makes sense for a limited set of AMR criteria for testing
   - Will allow generating data on a dynamic grid to test post-processing scripts
- This adds the functionality to the AMR actions to:
   - (de)register (old) new array elements
   - Apply mutators to initialize items on each element

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/ElementRegistration.hpp
+++ b/src/Parallel/ElementRegistration.hpp
@@ -1,0 +1,91 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <type_traits>
+
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+namespace db {
+template <typename DbTagList>
+class DataBox;
+}  // namespace db
+namespace Parallel {
+template <typename Metavariables>
+class GlobalCache;
+}  // namespace Parallel
+/// \endcond
+
+namespace Parallel {
+namespace detail {
+template <typename Metavariables, typename ParallelComponent,
+          typename = std::void_t<>>
+struct has_registration_list : std::false_type {};
+
+template <typename Metavariables, typename ParallelComponent>
+struct has_registration_list<
+    Metavariables, ParallelComponent,
+    std::void_t<typename Metavariables::template registration_list<
+        ParallelComponent>::type>> : std::true_type {};
+
+template <typename Metavariables, typename ParallelComponent>
+constexpr bool has_registration_list_v =
+    has_registration_list<Metavariables, ParallelComponent>::value;
+
+template <typename Metavariables, typename ParallelComponent>
+using registration_list =
+    typename Metavariables::template registration_list<ParallelComponent>::type;
+}  // namespace detail
+
+/// @{
+/// \brief (De)register an array element for specified actions
+///
+/// \details Array elements are (de)registered with actions on components that
+/// need to know which elements are contributing data before the action can be
+/// executed.  If array elements are migrated (e.g. during load balancing),
+/// or are created/destroyed (e.g. during adaptive mesh refinement), these
+/// functions must be called in order to (un)register (old) new elements.
+/// The list of registration actions is obtained from
+/// `Metavariables::registration_list::type`.
+template <typename ParallelComponent, typename DbTagList,
+          typename Metavariables, typename ArrayIndex>
+void deregister_element(db::DataBox<DbTagList>& box,
+                        Parallel::GlobalCache<Metavariables>& cache,
+                        const ArrayIndex& array_index) {
+  if constexpr (detail::has_registration_list_v<Metavariables,
+                                                ParallelComponent>) {
+    using registration_list =
+        typename Metavariables::template registration_list<
+            ParallelComponent>::type;
+    tmpl::for_each<registration_list>(
+        [&box, &cache, &array_index](auto registration_v) {
+          using registration = typename decltype(registration_v)::type;
+          registration::template perform_deregistration<ParallelComponent>(
+              box, cache, array_index);
+        });
+  }
+}
+
+template <typename ParallelComponent, typename DbTagList,
+          typename Metavariables, typename ArrayIndex>
+void register_element(db::DataBox<DbTagList>& box,
+                      Parallel::GlobalCache<Metavariables>& cache,
+                      const ArrayIndex& array_index) {
+  if constexpr (detail::has_registration_list_v<Metavariables,
+                                                ParallelComponent>) {
+    using registration_list =
+        typename Metavariables::template registration_list<
+            ParallelComponent>::type;
+    tmpl::for_each<registration_list>(
+        [&box, &cache, &array_index](auto registration_v) {
+          using registration = typename decltype(registration_v)::type;
+          registration::template perform_registration<ParallelComponent>(
+              box, cache, array_index);
+        });
+  }
+}
+/// @}
+}  // namespace Parallel
--- a/src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
+++ b/src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
@@ -54,10 +54,7 @@ namespace amr::Actions {
 /// - Updates the Neighbors of the Element
 /// - Resets amr::Tags::Flag%s to amr::Flag::Undefined
 /// - Resets amr::Tags::NeighborFlags to an empty map
-///
-/// \warning At the moment, this action only updates the Mesh for a p-refined
-/// Element.  It does not update any data related to evolution or elliptic
-/// solves.
+/// - Mutates all return_tags of Metavariables::amr_mutators
 struct AdjustDomain {
   template <typename ParallelComponent, typename DbTagList,
             typename Metavariables>
@@ -101,34 +98,58 @@ struct AdjustDomain {
       }
 
     } else {
+      const auto old_mesh_and_element =
+          std::make_pair(db::get<::domain::Tags::Mesh<volume_dim>>(box),
+                         db::get<::domain::Tags::Element<volume_dim>>(box));
+      const auto& old_mesh = old_mesh_and_element.first;
+
+      // Determine new neighbors and update the Element
+      {  // avoid shadowing when mutating flags below
+        const auto& amr_flags_of_neighbors =
+            db::get<amr::Tags::NeighborFlags<volume_dim>>(box);
+        db::mutate<::domain::Tags::Element<volume_dim>>(
+            [&element_id, &amr_flags_of_neighbors](
+                const gsl::not_null<Element<volume_dim>*> element) {
+              auto new_neighbors = element->neighbors();
+              for (auto& [direction, neighbors] : new_neighbors) {
+                neighbors.set_ids_to(amr::new_neighbor_ids(
+                    element_id, direction, neighbors, amr_flags_of_neighbors));
+              }
+              *element = Element<volume_dim>(element_id, new_neighbors);
+            },
+            make_not_null(&box));
+      }
+
       // Check for p-refinement
       if (alg::any_of(my_amr_flags, [](amr::Flag flag) {
             return (flag == amr::Flag::IncreaseResolution or
                     flag == amr::Flag::DecreaseResolution);
           })) {
         db::mutate<::domain::Tags::Mesh<volume_dim>>(
-            [&my_amr_flags](const gsl::not_null<Mesh<volume_dim>*> mesh) {
-              *mesh = amr::projectors::mesh(*mesh, my_amr_flags);
+            [&old_mesh,
+             &my_amr_flags](const gsl::not_null<Mesh<volume_dim>*> mesh) {
+              *mesh = amr::projectors::mesh(old_mesh, my_amr_flags);
             },
             make_not_null(&box));
       }
 
-      // Need to reset AMR flags and determine new neighbors
-      db::mutate<::domain::Tags::Element<volume_dim>,
-                 amr::Tags::Flags<volume_dim>,
+      // Run the mutators on all elements, even if they did no h-refinement.
+      // This allows mutators to update mutable items that depend upon the
+      // neighbors of the element.
+      tmpl::for_each<typename Metavariables::amr_mutators>(
+          [&box, &old_mesh_and_element](auto mutator_v) {
+            using mutator = typename decltype(mutator_v)::type;
+            db::mutate_apply<mutator>(make_not_null(&box),
+                                      old_mesh_and_element);
+          });
+
+      // Reset the AMR flags
+      db::mutate<amr::Tags::Flags<volume_dim>,
                  amr::Tags::NeighborFlags<volume_dim>>(
-          [&element_id](
-              const gsl::not_null<Element<volume_dim>*> element,
-              const gsl::not_null<std::array<amr::Flag, volume_dim>*> amr_flags,
-              const gsl::not_null<std::unordered_map<
-                  ElementId<volume_dim>, std::array<amr::Flag, volume_dim>>*>
-                  amr_flags_of_neighbors) {
-            auto new_neighbors = element->neighbors();
-            for (auto& [direction, neighbors] : new_neighbors) {
-              neighbors.set_ids_to(amr::new_neighbor_ids(
-                  element_id, direction, neighbors, *amr_flags_of_neighbors));
-            }
-            *element = Element<volume_dim>(element_id, new_neighbors);
+          [](const gsl::not_null<std::array<amr::Flag, volume_dim>*> amr_flags,
+             const gsl::not_null<std::unordered_map<
+                 ElementId<volume_dim>, std::array<amr::Flag, volume_dim>>*>
+                 amr_flags_of_neighbors) {
             amr_flags_of_neighbors->clear();
             for (size_t d = 0; d < volume_dim; ++d) {
               (*amr_flags)[d] = amr::Flag::Undefined;
@@ -136,9 +157,6 @@ struct AdjustDomain {
           },
           make_not_null(&box));
     }
-
-    // In the near future, add the capability of updating all data needed for
-    // an evolution or elliptic system
   }
 };
 }  // namespace amr::Actions
--- a/src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
+++ b/src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
@@ -54,7 +54,7 @@ namespace amr::Actions {
 /// - Updates the Neighbors of the Element
 /// - Resets amr::Tags::Flag%s to amr::Flag::Undefined
 /// - Resets amr::Tags::NeighborFlags to an empty map
-/// - Mutates all return_tags of Metavariables::amr_mutators
+/// - Mutates all return_tags of Metavariables::amr::projectors
 struct AdjustDomain {
   template <typename ParallelComponent, typename DbTagList,
             typename Metavariables>
@@ -133,14 +133,14 @@ struct AdjustDomain {
             make_not_null(&box));
       }
 
-      // Run the mutators on all elements, even if they did no h-refinement.
-      // This allows mutators to update mutable items that depend upon the
+      // Run the projectors on all elements, even if they did no h-refinement.
+      // This allows projectors to update mutable items that depend upon the
       // neighbors of the element.
-      tmpl::for_each<typename Metavariables::amr_mutators>(
-          [&box, &old_mesh_and_element](auto mutator_v) {
-            using mutator = typename decltype(mutator_v)::type;
-            db::mutate_apply<mutator>(make_not_null(&box),
-                                      old_mesh_and_element);
+      tmpl::for_each<typename Metavariables::amr::projectors>(
+          [&box, &old_mesh_and_element](auto projector_v) {
+            using projector = typename decltype(projector_v)::type;
+            db::mutate_apply<projector>(make_not_null(&box),
+                                        old_mesh_and_element);
           });
 
       // Reset the AMR flags
--- a/src/ParallelAlgorithms/Amr/Actions/RegisterCallbacks.hpp
+++ b/src/ParallelAlgorithms/Amr/Actions/RegisterCallbacks.hpp
@@ -1,0 +1,38 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <deque>
+#include <vector>
+
+#include "Parallel/Callback.hpp"
+#include "ParallelAlgorithms/Amr/Actions/CollectDataFromChildren.hpp"
+#include "ParallelAlgorithms/Amr/Actions/Component.hpp"
+#include "ParallelAlgorithms/Amr/Actions/CreateChild.hpp"
+#include "ParallelAlgorithms/Amr/Actions/SendDataToChildren.hpp"
+#include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace amr {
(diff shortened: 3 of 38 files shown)


src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
```diff
@@ -101,44 +104,70 @@ struct AdjustDomain {
...
-              *mesh = amr::projectors::mesh(*mesh, my_amr_flags);
+            [&old_mesh,
+             &my_amr_flags](const gsl::not_null<Mesh<volume_dim>*> mesh) {
+              *mesh = amr::projectors::mesh(old_mesh, my_amr_flags);
+            });
+
+        // The current implementation assumes that:
+        // -- any mutated items does not need access to the old Element.
+        // -- no item besides the Element needs to be mutated if the Neighbors
+        //    of the Element have changed.
```
wthrowe: Doesn't the mortar information need to change?

src/ParallelAlgorithms/Amr/Projectors/TimeStepping.hpp
```diff
@@ -0,0 +1,121 @@
...
+      const gsl::not_null<double*> time,
+      const gsl::not_null<AdaptiveSteppingDiagnostics*>
+          adaptive_stepping_diagnostics,
+      const ElementId<Dim>& /*element_id*/,
+      const std::unordered_map<ElementId<Dim>, tuples::TaggedTuple<Tags...>>&
+          children_items) {
+    const auto slowest_child =
+        alg::min_element(children_items, [](const auto& a, const auto& b) {
+          return get<::Tags::TimeStep>(a.second) <
+                 get<::Tags::TimeStep>(b.second);
```
wthrowe: The time step is signed, so you'll need to handle the negative case to support backwards evolution.  Probably easiest to just take the absolute value, which already exists for that type.

wthrowe (changes requested):
Haven't commented on any of the design things we discussed elsewhere.

src/ParallelAlgorithms/Amr/Actions/RegisterCallbacks.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+namespace amr {
+template <typename Component>
+void register_callbacks() {
+  using ArrayIndex = typename Component::array_index;
+  register_classes_with_charm(
+      tmpl::list<
+          Parallel::SimpleActionCallback<
+              amr::Actions::CreateChild,
+              CProxy_AlgorithmSingleton<amr::Component<metavariables>, int>,
```
nilsvu: `metavariables` isn't defined here, not sure why this doesn't error.

src/Parallel/ElementRegistration.hpp
```diff
@@ -0,0 +1,80 @@
...
+namespace Parallel {
+namespace detail {
+template <typename Metavariables, typename ParallelComponent,
+          typename = std::void_t<>>
+struct has_registration_list : std::false_type {};
+
+template <typename Metavariables, typename ParallelComponent>
+struct has_registration_list<
+    Metavariables, ParallelComponent,
+    std::void_t<typename Metavariables::template registration_list<
```
nilsvu, line 31: note for future: we need to document this registration stuff in the metavariables as well. Also we could make the `registration_list` a `tmpl::map` to avoid the awkward templating on the component.

src/Parallel/ElementRegistration.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+template <typename Metavariables, typename ParallelComponent>
+using registration_list =
+    typename Metavariables::template registration_list<ParallelComponent>::type;
+}  // namespace detail
+
+template <typename ParallelComponent, typename DbTagList,
+          typename Metavariables, typename ArrayIndex>
+void deregister_element(db::DataBox<DbTagList>& box,
+                        Parallel::GlobalCache<Metavariables>& cache,
```
nilsvu, line 56: take these things as not-null pointers?

src/Parallel/ElementRegistration.hpp
```diff
@@ -0,0 +1,80 @@
...
+          using registration = typename decltype(registration_v)::type;
+          registration::template perform_deregistration<ParallelComponent>(
+              box, cache, array_index);
+        });
+  }
+}
+
+template <typename ParallelComponent, typename DbTagList,
+          typename Metavariables, typename ArrayIndex>
+void register_element(db::DataBox<DbTagList>& box,
```
nilsvu, line 74: Add some docs please

src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
```diff
@@ -54,10 +54,13 @@ namespace amr::Actions {
 /// - Updates the Neighbors of the Element
 /// - Resets amr::Tags::Flag%s to amr::Flag::Undefined
 /// - Resets amr::Tags::NeighborFlags to an empty map
+/// - Mutates all return_tags of Metavariables::amr_mutators if the Mesh changed
 ///
-/// \warning At the moment, this action only updates the Mesh for a p-refined
-/// Element.  It does not update any data related to evolution or elliptic
-/// solves.
+/// \warning The current implementation assumes that:
+/// - any mutated items does not need access to the old Element.
```
nilsvu: Not sure what "item" means in this context. Maybe use "mutator" or clarify the wording somehow?

Also please add information about the AMR mutators to the docs somewhere. Probably a small note here that refers to the AMR metavars docs that we talked about in the meeting could work. Also in the other actions that reach into the metavars.

src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
nilsvu: I think the idea is to pass meshes along with face data in DG communications and compute mortars on-the-fly in evolutions. However, for elliptic systems it would be nice to cache some things on mortars (Larry and I talked about this elsewhere). So I think it would be good to give every element the opportunity to update. Not sure how that works together with async AMR though.

nilsvu (changes requested):
Reviewed only the first couple of commits

src/ParallelAlgorithms/Amr/Actions/RegisterCallbacks.hpp
wthrowe: It doesn't error because of https://github.com/sxs-collaboration/spectre/blob/144c684133a565a0f28db40edcf8bedf3ec1ea6d/cmake/AddSpectreExecutable.cmake#L85

I've wondered off and on whether we should have given that a less generic name so this sort of thing didn't happen.

src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
wthrowe: What about for h-refinement?  I guess we can do the mortar splitting or merging when we first receive messages, but that sounds more complicated than reorganizing as soon as we know what we need to do.  I don't think we need the mesh information, but I'm not entirely sure of that.

nilsvu:
Notes from a discussion with @kidder:

- We can probably keep track of the neighbor meshes during AMR in an <s>`amr::Tags::NeighborMeshes`</s> `domain::Tags::NeighborMeshes` tag alongside `amr::Tags::NeighborFlags`. This would allow to update mortars etc in a simple AMR mutator so we don't have to change DG communication patterns. (also makes it easier to cache data on mortars like I do in the elliptic DG code, and to update things like Schwarz subdomains / multigrid in AMR mutators rather than requiring extra communications to exchange neighbor meshes). AMR operates on neighbors and meshes, so seems reasonable to keep track of those for AMR mutators.
- New parent/child elements should probably resume the algorithm where the destroyed elements left off, instead of starting at the beginning of the phase. (They inherit state from their parent/children). Can be done by adding phase bookmarks to DistributedObject constructor.

src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
nilsvu: @kidder and I discussed keeping track of the neighbor meshes during AMR, which would resolve the mortar issue (see notes below https://github.com/sxs-collaboration/spectre/pull/5065#issuecomment-1585785987).

kidder: Okay, I've changed this part of AdjustDomain to always pass the old mesh and element to the mutators, and removed the comment about how to do so in the future as it was confusing.   I chose to pass a `std::pair<Mesh,Element>` so that all three apply calls for the mutators take one additional argument to the db::mutate_apply call as this could allow a mutator to template the apply function on the additional argument in order to reduce code duplication.

src/ParallelAlgorithms/Amr/Projectors/TimeStepping.hpp
kidder: Okay, added a test for the negative case and fixed the implementation

src/ParallelAlgorithms/Amr/Actions/RegisterCallbacks.hpp
kidder: added another template parameter to properly get the metavariables into the function

src/Parallel/ElementRegistration.hpp
kidder: I agree, and will create an issue so we don't forget

src/Parallel/ElementRegistration.hpp
kidder: I'm keeping as-is as we use this pattern throughout the action framework when passing the DataBox, GlobalCache, and ArrayIndex around.

src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
kidder: I've added a commit with protocols.  I'll use them in the actions and metavariables once you are happy with the protocols.

kidder:
I've pushed some fixups. Please give feedback.
For this PR, I will use the protocols in the metavariables and actions once we're happy with the protocols
In my next AMR PR(s), I'll add the tracking of neighbor meshes and pass the phase bookmark so new elements can resume the algorithm at the same place as existing elements.

src/ParallelAlgorithms/Amr/Projectors/Protocols/Mutators.hpp
```diff
@@ -0,0 +1,31 @@
...
+/// A class conforming to this protocol is placed in the metavariables to
+/// provide the list of AMR mutators that will be applied by:
+/// - amr::Actions::InitializeChild and amr::Actions::InitializeParent in order
+///   to initialize data on newly created elements.
+/// - amr::Actions::AdjustDomain in order to update data on existing elements in
+///   case their Mesh or neighbors have changed.
+///
+/// Here is an example for a class conforming to this protocol:
+///
+/// \snippet Projectors/Test_Protocol.cpp amr_mutators
```
wthrowe: Test_Protocol -> Test_Protocol**s**

src/ParallelAlgorithms/Amr/Projectors/Protocols/Mutators.hpp
```diff
@@ -0,0 +1,31 @@
...
+
+#include <type_traits>
+
+#include "ParallelAlgorithms/Amr/Projectors/Protocols/Mutator.hpp"
+
+namespace amr::protocols {
+/// \brief Compile-time information for AMR mutators
+///
+/// A class conforming to this protocol is placed in the metavariables to
+/// provide the list of AMR mutators that will be applied by:
```
wthrowe: Would be good to get a link to the `Mutator` protocol somewhere in this protocol's documentation.

wthrowe (commented):
Fixups look good, other than two comments.

src/ParallelAlgorithms/Amr/Projectors/TimeStepping.hpp
```diff
@@ -0,0 +1,121 @@
+// Distributed under the MIT License.
```
nilsvu: Can we move this to `Time/` or `Evolution/`? As discussed in a call recently I think we have to be careful not to repeat the mistake we made with the parallel interpolation framework, which couples most of the code together. Libs should depend on AMR and implement their own projectors/mutators, not the other way round. I've had good success implementing AMR mutators right next to each initialization action, even in the same file. That worked well because usually you have to project all the same tags during AMR that you initialize at the beginning. When the AMR and init mutators live next to each other then they can be maintained together, possibly even sharing functions.

src/ParallelAlgorithms/Amr/Projectors/CoordinateMaps.hpp
```diff
@@ -0,0 +1,67 @@
+// Distributed under the MIT License.
```
nilsvu: Same comment about moving this to where the coord maps get initialized.

src/Parallel/ElementRegistration.hpp
```diff
@@ -40,6 +40,16 @@ using registration_list =
     typename Metavariables::template registration_list<ParallelComponent>::type;
 }  // namespace detail
 
+/// @{
+/// \brief (De)register an array element for specified actions
+///
+/// \details Array elements are (de)registered with actions on components that
+/// need to know which elements are contributing data before the actiion can be
```
nilsvu: typo `action`

src/Parallel/ElementRegistration.hpp
```diff
@@ -40,6 +40,16 @@ using registration_list =
...
+/// executed.  If array elements are migratrated (e.g. during load balancing),
```
nilsvu: typo `migrated`

src/Parallel/ElementRegistration.hpp
```diff
@@ -40,6 +40,16 @@ using registration_list =
...
+/// or are created/destroyed (e.g. during adaptive mesh refinement), these
+/// functions must be called in order to (un)resgister (old) new elements.
```
nilsvu: typo `register`

src/ParallelAlgorithms/Amr/Projectors/Protocols/Mutator.hpp
```diff
@@ -0,0 +1,47 @@
+// Distributed under the MIT License.
```
nilsvu: I think move this up to `Amr/Protocols/Mutator.hpp`

src/ParallelAlgorithms/Amr/Projectors/Protocols/Mutator.hpp
```diff
@@ -0,0 +1,47 @@
...
+#include <type_traits>
+
+#include "Utilities/ProtocolHelpers.hpp"
+
+namespace amr::protocols {
+
+/// \brief A DataBox mutator used in AMR actions
+///
+/// A class conforming to this protocol can be used as the template argument for
+/// a call to db::mutate_apply(const gsl::not_null<DataBox<BoxTags>*> box,
```
nilsvu: I think it would be clearer to focus these docs on helping the user implement a mutator, rather than being too technical. I think the fact that `db::mutate_apply` is invoked could be nice to mention, but doesn't help the user much because then they have to look up how mutate_apply works and how it fits together with this. I suggest:

- Start with explaining that mutators get called when refining elements.
- The explain the compile-time interface, highlighting that return tags are mutated, argument tags are additional args, and you need three apply functions that take exactly the return tags, arg tags, and one extra arg.
- Add one or more examples.

src/ParallelAlgorithms/Amr/Projectors/Protocols/Mutators.hpp
```diff
@@ -0,0 +1,31 @@
...
+/// - amr::Actions::InitializeChild and amr::Actions::InitializeParent in order
+///   to initialize data on newly created elements.
+/// - amr::Actions::AdjustDomain in order to update data on existing elements in
+///   case their Mesh or neighbors have changed.
+///
+/// Here is an example for a class conforming to this protocol:
+///
+/// \snippet Projectors/Test_Protocol.cpp amr_mutators
+struct Mutators {
```
nilsvu: How about we name this `AmrMetavariables` and in the metavars it's a section like this:

```cpp
struct amr : tt::ConformsTo<amr::protocols::AmrMetavariables> {
  using mutators = tmpl::list<...>;
};
```

Then this would be a place to put more metavars if needed and a way to have documented "sections" in the metavars.

nilsvu (changes requested):

kidder:
@nilsvu @wthrowe pushed another set of fixups starting with `fixup rebase time`

kidder:
@wthrowe @nilsvu please look at the latest fixups starting with `fixup rebase time`

src/ParallelAlgorithms/Amr/Projectors/Protocols/Mutators.hpp
```diff
@@ -11,16 +11,18 @@ namespace amr::protocols {
...
 ///   to initialize data on newly created elements.
 /// - amr::Actions::AdjustDomain in order to update data on existing elements in
 ///   case their Mesh or neighbors have changed.
 ///
+///
 /// Here is an example for a class conforming to this protocol:
 ///
-/// \snippet Projectors/Test_Protocol.cpp amr_mutators
-struct Mutators {
+/// \snippet Projectors/Test_Protocols.cpp amr_mutators
```
wthrowe: Doxygen still isn't finding this.

src/ParallelAlgorithms/Amr/Projectors/Protocols/Mutator.hpp
```diff
@@ -37,6 +41,9 @@ namespace amr::protocols {
 ///   amr::Actions::InitializeChild and amr::Actions::InitializeParent hold the
```
wthrowe: I missed this before, but you need a blank line before "The Mesh and Element..." two lines above here.

wthrowe (changes requested):
You can squash these once @nilsvu has looked at the current state.

nilsvu (commented):
Please feel free to squash. I'll do another review once you have squashed.

nilsvu (commented):
Would you mind factoring some or all commits from this PR into separate PRs? This PR does a lot of things, and we can get parts of it merged a lot faster I think.

Edit: We can merge the protocols first, then the domain/time mutators. We can merge the time stepping and registration stuff separately.

src/Evolution/Initialization/CoordinateMaps.hpp
```diff
@@ -0,0 +1,68 @@
...
+#include "Utilities/TMPL.hpp"
+
+namespace amr::projectors {
+/// \brief Initialize/update items related to coordinate maps after an AMR
+/// change
+template <size_t Dim>
+struct CoordinateMaps {
+  using return_tags = tmpl::list<::domain::Tags::ElementMap<Dim, Frame::Grid>,
+                                 ::domain::CoordinateMaps::Tags::CoordinateMap<
+                                     Dim, Frame::Grid, Frame::Inertial>>;
```
nilsvu: This initializes the tags added by `evolution::dg::Initialization::Domain` in `DgDomain.hpp`, so move this struct into the same file and name `ProjectDomain` or something like that. When I added support for AMR it helped a lot to have the AMR mutator right next to the initialization mutator. Same for the time stepping mutator.

nilsdeppe:
@nilsvu I can take over reviewing this PR. I have no problem reviewing it with all the commits. Given that it's been open for 2 months, I think we should focus on getting it merged.

nilsvu:
Well nothing happened on this PR for about a month, and not for lack of review, so implying that I slow it down by suggesting we merge the easy things first is a bit uncalled for I think..

nilsdeppe:
I wasn't implying that you slowed it down.

kidder:
@nilsvu after our discussions, I've put the AMR mutators in the same files as their corresponding initializers, renamed them to ProjectX where X is the name of the initializer, and put them in the same namespace as the initializer.

src/Parallel/ElementRegistration.hpp
nilsvu: https://github.com/sxs-collaboration/spectre/issues/3266

src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
```diff
@@ -101,44 +98,62 @@ struct AdjustDomain {
...
+             &my_amr_flags](const gsl::not_null<Mesh<volume_dim>*> mesh) {
+              *mesh = amr::projectors::mesh(old_mesh, my_amr_flags);
             },
             make_not_null(&box));
       }
 
-      // Need to reset AMR flags and determine new neighbors
-      db::mutate<::domain::Tags::Element<volume_dim>,
-                 amr::Tags::Flags<volume_dim>,
+      tmpl::for_each<typename Metavariables::amr_mutators>(
```
nilsvu: I think it would be helpful to keep/add the comment that currently only the (de)refined elements run the AMR mutators, and their neighbors don't get a chance to update. We may want to allow neighbors to run AMR mutators as well in the future.

src/ParallelAlgorithms/Amr/Protocols/AmrMetavariables.hpp
```diff
@@ -0,0 +1,33 @@
...
+///   case their Mesh or neighbors have changed.
+///
+///
+/// Here is an example for a class conforming to this protocol:
+///
+/// \snippet Amr/Test_Protocols.cpp amr_mutators
+struct AmrMetavariables {
+  template <typename ConformingType>
+  struct test {
+    using mutators = typename ConformingType::mutators;
```
nilsvu: Maybe `mutators` -> `projectors` everywhere? I'm just thinking how we want to call these things. "AMR projectors" seems a bit more descriptive to me than "AMR mutators" (they project data between elements at different resolution, rather than just generically "mutate" things). Then naming the classes `ProjectX` would be reasonable too.

src/ParallelAlgorithms/Amr/Protocols/Mutator.hpp
```diff
@@ -0,0 +1,55 @@
...
+///
+///   The Mesh and Element passed to amr::Actions::AdjustDomain are their
+///   values before the grid changes.  The tuples passed to
+///   amr::Actions::InitializeChild and amr::Actions::InitializeParent hold the
+///   items corresponding to the `DataBox<BoxTags>::mutable_item_creation_tags`
+///   of the parent (children) of the child (parent) being initialized.
+///
+/// For examples, see amr::projectors::TimeStepping and
+/// amr::projectors::CoordinateMaps
+struct Mutator {
```
nilsvu: Maybe `Projector` (see comment above)?

nilsvu (commented):
Looks good! Couple of suggestions.

src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
kidder: actually here the mutators are run on all elements (that are not newly created via h-refinement) regardless of whether their Mesh changed.  I'll add a comment here and mention this in the protocol

src/ParallelAlgorithms/Amr/Protocols/AmrMetavariables.hpp
kidder: sure

src/ParallelAlgorithms/Amr/Protocols/Mutator.hpp
kidder: sure

kidder:
@nilsvu rebased and squashed in your most recent suggested changes

nilsvu (approved):
LGTM! Thanks for making those changes.

nilsvu:
@wthrowe do you want so sign off in this?

wthrowe (approved):

PR #3963, opened 2022-05-07
Alter CCE interface managers for new changes to interpolation

kidder:
## Proposed changes

This is a rebased version of #3422, dropping the commit [Use evolution_less_equal for receive_local_time_stepping](https://github.com/sxs-collaboration/spectre/pull/3422/commits/d9eb48d055990b8e874804937dc8de6e8f411a0a), while also making sure the individual commits compile

This makes a number of changes to the interface for the GH system sending data to CCE based on the newer design considerations from the availability of dense interpolation:
- The interface managers no longer use the time derivatives of the GH variables, as they won't be as easily available from dense output
- The interface managers are no longer factory-creatable; the `GhLockstep` interface manager is the only available method during self-start, and the `GhLocalTimeStepping` interface manager is the only available method during the evolution
- Handle self-start separately using an action that manually calls the appropriate interpolation event, and take the `TimeStepId` as the relevant temporal id for those utilities intended for self start, and `double` for all other cases.

### Upgrade instructions

Any usage of the `GhInterfaceManagers` will likely need to be reworked to account for the new interfaces.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #3422, opened 2021-08-09
Alter CCE interface managers for new changes to interpolation

moxcodes:
## Proposed changes

This makes a number of changes to the interface for the GH system sending data to CCE based on the newer design considerations from the availability of dense interpolation:
- The interface managers no longer use the time derivatives of the GH variables, as they won't be as easily available from dense output
- The interface managers are no longer factory-creatable; the `GhLockstep` interface manager is the only available method during self-start, and the `GhLocalTimeStepping` interface manager is the only available method during the evolution
- Handle self-start separately using an action that manually calls the appropriate interpolation event, and take the `TimeStepId` as the relevant temporal id for those utilities intended for self start, and `double` for all other cases.

### Upgrade instructions

Any usage of the `GhInterfaceManagers` will likely need to be reworked to account for the new interfaces.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Dependencies

- [x] #3421

PR #5070, opened 2023-06-02
Somewhat fix ordering issues in interpolator

knelli2:
## Proposed changes

The reason this is "somewhat" is explained in the code comments. I couldn't figure out a permanent fix without some extensive changes (i.e. checking the temporal id type and then doing something custom if it's a LinkedMessageId). These changes have been enough for me to test size control without issues so far, but it's possible I just haven't hit the case where this change won't fix anything yet. I plan to deal with that later if needed.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/LinkedMessageId.hpp
+++ b/src/DataStructures/LinkedMessageId.hpp
@@ -72,3 +72,8 @@ struct LinkedMessageIdLessComparator {
     return linked_message_id.id < id;
   }
 };
+
+template <typename Id>
+bool operator<(const LinkedMessageId<Id>& lhs, const LinkedMessageId<Id>& rhs) {
+  return LinkedMessageIdLessComparator<Id>{}(lhs, rhs);
+}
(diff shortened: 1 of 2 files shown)


src/DataStructures/LinkedMessageId.hpp
```diff
@@ -72,3 +72,8 @@ struct LinkedMessageIdLessComparator {
     return linked_message_id.id < id;
   }
 };
+
+template <typename Id>
+bool operator<(const LinkedMessageId<Id>& lhs, const LinkedMessageId<Id>& rhs) {
+  return LinkedMessageIdLessComparator<Id>{}(lhs, rhs);
+}
```
wthrowe, line 79: Not new, so not necessary to change here, but why do we have `LinkedMessageIdLessComparator` instead of just some `operator<` functions?

wthrowe (approved):
As you say, not a true fix, but if it works around the problem for now that's an improvement.

src/DataStructures/LinkedMessageId.hpp
knelli2: No clue. @kidder? Git blame points to you from #3963.

kidder: I just rebased @moxcodes #3422 in that PR, so if `operator<` functions would be better that is fine with me...

knelli2: I can do it in a follow-up PR

PR #5071, opened 2023-06-02
Add more size error diagnostics

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size/Error.hpp
+++ b/src/ControlSystem/ControlErrors/Size/Error.hpp
@@ -35,6 +35,8 @@ struct ErrorDiagnostics {
   double dt_lambda_00;
   double horizon_00;
   double dt_horizon_00;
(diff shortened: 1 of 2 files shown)


src/ControlSystem/ControlErrors/Size/Error.hpp
```diff
@@ -35,6 +35,8 @@ struct ErrorDiagnostics {
   double dt_lambda_00;
   double horizon_00;
   double dt_horizon_00;
+  double min_delta_r;
```
kidder, line 38: Is there any documentation of these diagnostics?

knelli2: I plan to write these to disk in the actual control error class itself. There I'll have a full docs list of what everything is that I write

kidder (approved):

PR #5072, opened 2023-06-02
Add reset function to size Info

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size/Info.cpp
+++ b/src/ControlSystem/ControlErrors/Size/Info.cpp
@@ -3,16 +3,18 @@
 
 #include "ControlSystem/ControlErrors/Size/Info.hpp"
 
+#include <optional>
 #include <pup.h>
 #include <pup_stl.h>
 
 #include "ControlSystem/ControlErrors/Size/State.hpp"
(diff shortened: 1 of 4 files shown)


kidder (dismissed):

src/ControlSystem/ControlErrors/Size/Info.cpp
```diff
@@ -42,6 +42,14 @@ void Info::pup(PUP::er& p) {
   p | discontinuous_change_has_occurred;
 }
 
+void Info::reset() {
+  suggested_time_scale = 0.0;
```
nilsvu: What does the value of 0 mean? Should it be NaN?

knelli2: 0 is a sentinel value meaning that the control system simply hasn't suggested a timescale.

nilsvu: Sounds like something that should be documented..

knelli2: Ok I documented it and added a commit to make it a better sentinel (nullopt)

nilsvu (approved):

PR #5075, opened 2023-06-02
Allow control sys dense output during first time step

knelli2:
## Proposed changes

Size control needs a very small damping timescale compared to the other control systems initially. This may cause the measurement timescale (and thus the horizon finds/interpolations) to be smaller than a timestep initially. Previously the control system prevented this, but now it's ok. And since we can do dense output between steps without having to take the full step for linear-multistep methods, we don't have to worry about functions of time being expired.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ExpirationTimes.hpp
+++ b/src/ControlSystem/ExpirationTimes.hpp
@@ -103,24 +103,17 @@ double measurement_expiration_time(const double time,
  *
  * If the control system isn't active then expiration time is
  * `std::numeric_limits<double>::infinity()`.
- *
- * To protect against bad inputs, if the initial expiration time that is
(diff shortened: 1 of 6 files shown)


wthrowe (approved):
> And since we can do dense output between steps without having to take the full step for linear-multistep methods, we don't have to worry about functions of time being expired.

Don't assume that's going to continue to work.  It might get broken with the array to nodegroup conversion, and, even if not, you're going to need to support things other than LMMs eventually.

However, I don't think this change makes anything worse, and just makes the initial behavior match what happens later.  So no objection to it.

knelli2:
Yeah we'll need to think more about how to support substep methods.

PR #5078, opened 2023-06-06
Represent segments/checkpoints in Python, use in status CLI

nilsvu:
## Proposed changes

Preparation for #4996 to make that PR a bit smaller.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/support/Python/Test_DirectoryStructure.py
+++ b/tests/support/Python/Test_DirectoryStructure.py
@@ -1,0 +1,69 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+import shutil
+import unittest
+from pathlib import Path
+
+from spectre.Informer import unit_test_build_path
+from spectre.support.DirectoryStructure import (
+    Checkpoint,
+    Segment,
+    list_checkpoints,
+    list_segments,
+)
+
+
+class TestDirectoryStructure(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = Path(
+            unit_test_build_path(), "support/DirectoryStructure"
+        )
+        shutil.rmtree(self.test_dir, ignore_errors=True)
+        self.test_dir.mkdir(parents=True, exist_ok=True)
+
+    def tearDown(self):
+        shutil.rmtree(self.test_dir, ignore_errors=True)
+
+    def test_checkpoints(self):
+        checkpoint = Checkpoint.match(self.test_dir / "Checkpoint_0002")
+        self.assertEqual(
+            checkpoint, Checkpoint(path=self.test_dir / "Checkpoint_0002", id=2)
+        )
+
+        self.assertEqual(list_checkpoints(self.test_dir), [])
+        checkpoint.path.mkdir()
+        self.assertEqual(list_checkpoints(self.test_dir), [checkpoint])
+
+    def test_segments(self):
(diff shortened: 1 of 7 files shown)


nilsvu:
@wthrowe @knelli2 we can do this subset of #4996 first to make the other PR a bit smaller.

tools/Status/Status.py
```diff
@@ -224,6 +233,34 @@ def render_status(show_paths, show_unidentified, state_styles, **kwargs):
     if len(job_data) == 0:
         return
 
+    # Remove deleted jobs
+    if not show_deleted:
+        deleted_jobs = job_data[~job_data["WorkDir"].map(os.path.exists)]
+        job_data.drop(deleted_jobs.index, inplace=True)
+
+    # Keep only latest in a series of segments
+    job_data[["SegmentsDir", "SegmentId"]] = [
```
knelli2, line 242: Can we add a column that prints what segment we are on? I think that could be useful. Then we don't have to include it in the job name like SpEC does.

knelli2 (changes requested):
Should you include the changes to Main that actually change the checkpoint structure in this PR?

tools/Status/Status.py
nilsvu: I had that before but didn't find it particularly useful. I can add it if you want though. Not sure.

knelli2: For long runs I think it'd be useful. It's a very quick way to know how many times the simulation has restarted.

nilsvu:
Moved the changes to Main to #5079

tools/Status/Status.py
nilsvu: Right but is it useful to know the number of restarts? I don't know. Remember we have limited space in the terminal, so adding a column means things like the job name get truncated further.

knelli2: I feel like it should be an option. Everybody is going to want to see different things, so maybe (for a later PR) we should take the fields to print as arguments so users can customize them in their spectre.yaml file

nilsvu: Yea that would be nice

nilsvu:
@wthrowe could you take this?

tests/support/Python/Test_DirectoryStructure.py
```diff
@@ -0,0 +1,68 @@
...
+    list_segments,
+)
+
+
+class TestDirectoryStructure(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = Path(
+            unit_test_build_path(), "support/DirectoryStructure"
+        )
+        self.test_dir.mkdir(parents=True, exist_ok=True)
```
wthrowe, line 24: Delete any preexisting files.

wthrowe (changes requested):

tests/support/Python/Test_DirectoryStructure.py
nilsvu: ok fixed

wthrowe (approved):

nilsvu:
@wthrowe I think this is ready to go as well. @knelli2 is on vacation this week and we can make any changes to the `status` CLI in follow-ups if we want to.

wthrowe:
Agreed.  I don't see @knelli2's request as blocking.

PR #4000, opened 2022-05-27
Cleanup GH constraint tests & delete check output dirs

nilsdeppe:
## Proposed changes

@wthrowe does this deal with https://github.com/sxs-collaboration/spectre/issues/3512 ?

Fixes #3512 

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5080, opened 2023-06-07
Fix input file tests

### LINKED ISSUE #4911: Execute tests on CI can pass even if executable fails
This can happen if all the files are written that need to be checked, but the executable still fails. We should check error codes on CI.

nilsvu:
## Proposed changes

- The `execute_check_output` mode didn't report errors correctly (bug introduced in #4000). Fixes #4911.
- Nonzero exit codes were treated as test failure. Now the expected exit code can be specified.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


cmake/RunInputFileTest.sh
```diff
@@ -0,0 +1,36 @@
...
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+# Positional arguments to this script:
+# - $1: executable name
+# - $2: path to input file
+# - $3: directory name
+# - $4: additional command-line arguments forwarded to the executable
+# - $5: expected exit code
+# - $6: "true" to check output files or "false" to skip the check
```
knelli2: The code below is correct but these docs are wrong. 6 is the additional args, 4 is expected exit code, and 5 is check output.

knelli2 (changes requested):
Looks like one of our input file tests has been failing

kidder:
Given the number of times the `test_input_file` scripts have had bugs, it seems to me that this functionality should be tested somehow.  It is not obvious looking at the single line added in #4000 would break the functionality.   The lack of a test and lack of documentation that the exit code matters seems more to blame.

tests/InputFiles/ExampleExecutables/RandomAmr1D.yaml
```diff
@@ -3,10 +3,11 @@
 
 Executable: RandomAmr1D
 Testing:
-  Check: parse;execute_check_output
+  Check: parse;execute
+  Timeout: 10
+  ExpectedExitCode: 2
```
kidder, line 8: why is it 2?  How would I determine this?

kidder (changes requested):
there also needs to be some change to the dev guide documentation for writing an input file test and this documentation should be easy to find

tests/InputFiles/ExampleExecutables/RandomAmr1D.yaml
nilsvu: Added docs to `CheckpointAndExitAfterWallclock`

nilsvu:
I agree the bug was subtle. Added some docs on exit codes to `CheckpointAndExitAfterWallclock`. This was documented already in https://spectre-code.org/tutorial_checkpoint_restart.html but no-one reads dev guides because they're not connected to the code (that's also why they're always out of date). Do we have docs on input file tests? I agree it would be nice to refactor them (maybe in Python), add tests, etc, but I don't want to be drawn into a rabbit hole refactoring all that in this PR when all I'm doing is fixing a bug ("no good deed goes unpunished...").

kidder:
yeah, with a quick browse, I can't find any documentation about what options go into the input file for an input-file test.  If you know of some, it would be nice to link to them from the Writing Unit tests page as that is where I would start to look for anything related to tests.  And maybe from the Option Parsing doc page when discussing the metadata section.   If there isn't any documentation, then please make an issue that this needs to be done.

src/Parallel/PhaseControl/CheckpointAndExitAfterWallclock.hpp
```diff
@@ -77,6 +77,9 @@ struct CheckpointAndExitRequested {
  * \brief Phase control object that runs the WriteCheckpoint and Exit phases
  * after a specified amount of wallclock time has elapsed.
  *
+ * When the executable exits from here, it does so with
+ * `Parallel::ExitCode::ContinueFromCheckpoint`.
```
kidder, line 81: ah, I didn't remember we had defined our own enum for exit codes.  Maybe mention that in the script when it talks about exit codes

kidder (changes requested):

kidder:
and I agree that testing this framework is beyond the scope of this PR (maybe make an issue)

nilsvu:
Added some docs

kidder (approved):

knelli2 (approved):

PR #5081, opened 2023-06-07
Non diagonal jac subcell

ffoucart:
## Proposed changes

Takes into account the possibility that the jacobian of the map from logical to inertial coordinates is non-diagonal

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/Test_TimeDerivative.cpp
+++ b/tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/Test_TimeDerivative.cpp
@@ -22,6 +22,7 @@
 #include "Domain/CoordinateMaps/Identity.hpp"
 #include "Domain/CoordinateMaps/ProductMaps.hpp"
 #include "Domain/CoordinateMaps/ProductMaps.tpp"
+#include "Domain/CoordinateMaps/Rotation.hpp"
 #include "Domain/CreateInitialElement.hpp"
 #include "Domain/Creators/Tags/Domain.hpp"
 #include "Domain/Creators/Tags/ExternalBoundaryConditions.hpp"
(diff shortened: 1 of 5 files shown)


ffoucart:
Depends on #5005 
Tested so far on translation map only (which is diagonal...). The issues to discuss include
(a) Is it ok to multiply the boundary correction by the determinant of the jacobian as early as L755
(b) Should we directly calculate the inverse jacobian on faces rather than projecting from the DG grid
(c) The call to evolution::dg::subcell::add_cartesian_flux_divergence should, I think, work as coded... but the input of that function does not match its naming convention (i.e. we give the determinant of the inverse jacobian as the 'inverse jacobian'). Arguably, this was already true, as we previously gave a specific component of the inverse jacobian
(d) Does the code match Eq 34 of 2109.11645, or did I misunderstand something
@nilsdeppe

ffoucart:
Looking at the last commit only should be sufficient here, the other two are from the dependent PR.

tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/Subcell/Test_TimeDerivative.cpp
```diff
@@ -100,7 +101,8 @@ struct DummyEvolutionMetaVars {
   };
 };
 
-double test(const size_t num_dg_pts, std::optional<double> expansion_velocity) {
+double test(const size_t num_dg_pts, std::optional<double> expansion_velocity,
+            const bool TestNonDiagonalJacobian) {
```
nilsdeppe: `TestNonDiagonalJacobian` -> `test_non_diagonal_jacobian`

nilsdeppe (commented):
LGTM, I have no comments on the last 2 commits

nilsdeppe (approved):

PR #5083, opened 2023-06-07
Add ExtendConnectivity executable

nring21:
## Proposed changes

Adds a post-processing executable called `ExtendConnectivity`. This executable takes in a single H5 file for an evolution (1D, 2D, or 3D) and rewrites the connectivity for each time step, such that all elements within each block are interconnected.

Compatibility with subcell systems and extending the connectivity between different blocks are to be implemented later.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Executables/ExtendConnectivity/ExtendConnectivity.cpp
```diff
@@ -0,0 +1,86 @@
...
+ */
+int main(int argc, char** argv) {
+  boost::program_options::positional_options_description pos_desc;
+
+  boost::program_options::options_description desc("Options");
+  desc.add_options()(
+      "help,h,",
+      "show this help message\n\nNote: This does not work with subcell systems "
+      ", and the connectivity only extends *within* each block "
+      "and not between them. \n")(
```
wthrowe: Can you include a brief description in the help of what the executable does and why I would want to do it?  (Also you have a comma on the wrong line.)

wthrowe (commented):
Not a full review.

knelli2:
Question about design choice. What was the reason for making this an executable rather than another python CLI function? This seems like it'd lend itself very well to pybindings.

nilsdeppe:
Re executable: I think this is more historical. This was started over a year ago before the CLI was anywhere close to what it is now. I agree, using the Python CLI to wrap this would be better. I guess the steps for that would be:
1. make python bindings for the `extend_connectivity` function.
2. write a Python wrapper that calls that.
@nilsvu is there an example of this to demonstrate how involved it would be to do?

I'm not sure we should do that in this PR since `CombineH5` should also be refactored in the same way and the feature presented here has been requested quite regularly.

knelli2:
Yeah probably first step so we don't try to do too many things at once is just get this new function on develop. That'd be the first two commits of this PR. Then another PR for adding a pybinding for the function and either in that PR or yet another one, adding a CLI for the whole process.

For pybindings we already have a bunch for volume data in `src/IO/H5/Python/VolumeData.cpp` so just adding this new function to that should be enough.

nilsdeppe:
That is fine with me. @nring21 is that okay with you? Should we discuss this in a call?

nring21:
Yea, that sounds good to me! Discussing this in a call with you @nilsdeppe would definitely help, as I'm not super familiar with using the Python CLI to wrap functions

nilsvu:
To bind the connectivity functions you can look at `src/IO/H5/Python/VolumeData.cpp` and add your functions there, as @knelli2 said. Just mirror the C++ function to Python (same name & arguments).

Then to wrap the Python bindings in a CLI you can look at `src/IO/H5/Python/DeleteSubfiles.py` for a simple example. There's also `src/Visualization/Python/InterpolateToCoords.py` with some example how to use the H5 volume data Python bindings.

Or maybe is it possible to extend the connectivity right when writing the data to the H5 file from the simulation so we don't have to do any post-processing?

nring21:
Thank you so much! I'll take a look at them.

As for extending the connectivity when initially writing the data to the H5, I recall that an initial worry was longer runtimes but maybe if done in a clever way that's not necessarily a concern

src/IO/H5/VolumeData.cpp
```diff
@@ -159,6 +165,335 @@ void append_element_extents_and_connectivity(
...
+  for (size_t i = 0; i < block_number_for_each_element.size(); ++i) {
+    sorted_element_indices[block_number_for_each_element[i]].push_back(i);
+  }
+
+  return std::make_tuple(number_of_blocks,
+                         std::move(block_number_for_each_element),
+                         std::move(sorted_element_indices));
+}
+
+// Sort input by block
```
nilsdeppe: Can you expand this a bit more? I found it a bit surprising that there wasn't a `std::sort` call. Maybe describe how each input relates to the output?

knelli2:
A question I forgot to ask during our meeting: If I run this function/executable on a volume file, will our other post-processing routines work on the newly connected data?

src/IO/H5/VolumeData.cpp
```diff
@@ -159,6 +165,335 @@ void append_element_extents_and_connectivity(
...
+  size_t expected_number_of_grid_points = 0;
+
+  for (const auto& extents : block_extents) {
+    size_t element_grid_points = 1;
+    size_t element_connectivity_length = 1;
+    for (size_t j = 0; j < SpatialDim; j++) {
+      element_grid_points *= extents[j];
+      element_connectivity_length *= extents[j] - 1;
+    }
+    // Connectivity that already exists
```
nilsdeppe, line 270: I found this and the next comment to not give me a good idea of the next line. How does the fact that a connectivity exists (and how do I know it exists?) relate to `+= element_connectivity_length * 8`? Maybe first answer here on GitHub and then I can make a concrete suggestion for the code?

It seems like `// Used for reserving the length of block_logical_coords` could be on line 254 to describe the variable, though.

src/IO/H5/VolumeData.cpp
```diff
@@ -159,6 +165,335 @@ void append_element_extents_and_connectivity(
...
+    // Used for reserving the length of block_logical_coords
+    expected_number_of_grid_points += element_grid_points;
+  }
+
+  std::string grid_name_string = block_grid_names[0];
+  std::array<int, SpatialDim> h_ref_array = {};
+  size_t h_ref_previous_start_position = 0;
+  size_t additional_connectivity_length = 1;
+  for (size_t i = 0; i < SpatialDim; ++i) {
+    size_t h_ref_start_position =
```
nilsdeppe: `const` the next 3 variables?

src/IO/H5/VolumeData.cpp
```diff
@@ -159,6 +165,335 @@ void append_element_extents_and_connectivity(
...
+// Given a std::vector<double> where the elements are ordered in ascending order
+// a new std::vector<double> is generated where it is a list of the original
+// values in ascending order without duplicates
+// Example: [1,2,2,3] -> order() -> [1,2,3]
+std::vector<double> order_sorted_elements(
+    const std::vector<double>& sorted_elements) {
+  std::vector<double> ordered_elements;
+  ordered_elements.push_back(sorted_elements[0]);
+  for (size_t i = 1; i < sorted_elements.size(); ++i) {
+    if (sorted_elements[i] == ordered_elements.end()[-1]) {
```
nilsdeppe: What about `if (... != ...)` and then put the `else` part in there. The continue doesn't really do anything. Normally continue is used to skip a remaining bit of work in a loop, but there's no addition work inside this loop

src/IO/H5/VolumeData.cpp
```diff
@@ -367,6 +702,138 @@ void VolumeData::write_volume_data(
...
+        new_connectivity.push_back(block_and_grid_point_map[it]);
+      }
+    }
+
+    // Deletes the existing connectivity and replaces it with the new one
+    const std::string path = "ObservationId" + std::to_string(obs_id);
+    detail::OpenGroup observation_group(volume_data_group_.id(), path,
+                                        AccessType::ReadWrite);
+    const hid_t group_id = observation_group.id();
+    H5Ldelete(group_id, "connectivity", H5P_DEFAULT);
```
nilsdeppe: Can you move this into the `src/IO/H5/Helpers.cpp` file next to the function to write it? I realize it's trivial code right now, but if we decide to compress the connectivity in the H5 file, then the property list might not be `H5P_DEFAULT` anymore, and this could lead to difficult to find bugs. Then you can also use `CHECK_H5(H5L...)` to verify the operation succeeded.

nilsdeppe (commented):
Overall LGTM, just some small suggestions you can squash immediately

nilsdeppe:
> A question I forgot to ask during our meeting: If I run this function/executable on a volume file, will our other post-processing routines work on the newly connected data?

@knelli2 this only changes the data inside the connectivity (and by changes, it adds more data), which currently only ParaView cares what it actually is, so no other tools are affected.

src/IO/H5/VolumeData.cpp
nring21: I took a closer look at this code and I realize that I should actually fix a couple of things.

What is happening in this chunk of the function is that we are computing by hand "how much" connectivity should exist without element-to-element connections. Natively we only have connectivity for cells. Each cell should contribute `2^SpatialDim` values to the connectivity since that is how many grid points are necessary to define a cell in a particular spatial dimension (i.e. in 3D, 8 grid points are connected as a hexahedron so this adds 8 to the length of the connectivity list). We are thus guaranteed to know that this connectivity must exist, otherwise, we have nothing to visualize. We then later compute the additional connections necessary and put this all together to get the `expected_connectivity_length`.

As for the things I intend to change:
1) `element_connectivity_length` --> `number_of_cells_in_element` since this is more descriptive
2) `expected_connectivity_length += element_connectivity_length * 8;` --> `expected_connectivity_length += element_connectivity_length * pow(2, SpatialDim);` since this should technically change depending on `SpatialDim` (this never errored before since `expected_connectivity_length` only reserves a vector length)

If there's anything else to change let me know!

src/IO/H5/VolumeData.cpp
nring21: When moving the `H5Ldelete(...)` part into the file you specified, do I need to change `H5P_DEFAULT` to something else more general?

nilsdeppe: The code there uses our own `h5p_default()` function, so just use that. Some of the HDF5 macro stuff causes a lot of warnings, so we wrap the code and silence warnings inside wrappers.

src/IO/H5/VolumeData.cpp
nilsdeppe: That sounds perfect! Thank you for the clarifications!

nilsdeppe (dismissed):
LGTM! @knelli2 @wthrowe @nilsvu any other comments for now? The pybindings & CLI think we will do in a separate PR. Either Noah or another of the undergrads  will do that soon.

tests/Unit/Executables/CMakeLists.txt
```diff
@@ -30,3 +31,35 @@ add_dependencies(
...
+   --subfile-name VolumeData \
+   ${CMAKE_BINARY_DIR}/tmp/Test_ExtendConnectivity_Data0.h5 \
+   && rm ${CMAKE_BINARY_DIR}/tmp/Test_ExtendConnectivity_Data0.h5 \
+   && rm ${_CONNECTIVITY_OUTPUT}.xmf"
+  )
+set_standalone_test_properties("Unit.Executables.ExtendConnectivity")
+set_tests_properties(
+    "Unit.Executables.ExtendConnectivity"
+    PROPERTIES
+    LABELS "standalone;unit")
```
wthrowe: "standalone;unit" doesn't make sense.  Those are mutually exclusive types, along with "inputfiles" (despite us naming everything with a "Unit." prefix).  Drop this `set_test_properties` and the following `add_dependencies` for `unit-tests`.

wthrowe (changes requested):
The help text changes are good.

wthrowe:
Looks good.  Squash.

wthrowe (approved):

PR #5084, opened 2023-06-07
Add size control error

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #5072.~

### DIFF
--- a/src/ControlSystem/ControlErrors/Size.hpp
+++ b/src/ControlSystem/ControlErrors/Size.hpp
@@ -1,0 +1,253 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <memory>
+#include <pup.h>
+#include <string>
+
+#include "ControlSystem/ControlErrors/Size/Error.hpp"
+#include "ControlSystem/ControlErrors/Size/Info.hpp"
+#include "ControlSystem/Protocols/ControlError.hpp"
+#include "ControlSystem/Tags/QueueTags.hpp"
+#include "ControlSystem/Tags/SystemTags.hpp"
+#include "ControlSystem/TimescaleTuner.hpp"
+#include "DataStructures/DataBox/Prefixes.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Creators/Tags/Domain.hpp"
+#include "Domain/Structure/ObjectLabel.hpp"
+#include "IO/Observer/ReductionActions.hpp"
+#include "NumericalAlgorithms/Interpolation/ZeroCrossingPredictor.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
+#include "Options/String.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+/// \cond
+namespace domain::Tags {
+struct FunctionsOfTime;
+}  // namespace domain::Tags
+namespace Frame {
+struct Grid;
+struct Distorted;
+}  // namespace Frame
+/// \endcond
+
+namespace control_system::ControlErrors {
+/*!
+ * \brief Control error in the for the \f$l=0\f$ component of the
+ * `domain::CoordinateMaps::TimeDependent::Shape` map.
+ *
+ * \details The goal of this control error is
+ *
+ * 1. Keep the excision sphere inside the horizon
+ * 2. Maintain a fixed distance between the excision surface and the horizon
+ * surface.
+ * 3. Prevent the characteristic field \f$ u^-_{ab} \f$ associated with the
+ * characteristic speed \f$ v_- \f$ in `gh::characteristic_speeds` from coming
+ * into the domain.
+ *
+ * For a more detailed account of how this is accomplished, see
+ * `control_system::size::State` and `control_system::size::control_error` which
+ * this class calls.
+ *
+ * This class holds a `control_system::size::Info` and three different
+ * `intrp::ZeroCrossingPredictor`s internally which are needed to calculate the
+ * `control_system::size::control_error`. It also conforms to the
+ * `control_system::protocols::ControlError` protocol.
+ *
+ * In addition to calculating the control error, if the
+ * `control_system::Tags::WriteDataToDisk` tag inside the
+ * `Parallel::GlobalCache` is true, then a diagnostic file named
+ * `Diagnostics.dat` is also written to the same group that
+ * `control_system::write_components_to_disk` would write the standard control
+ * system output (`/ControlSystems/Size/`). The columns of this diagnostic file
+ * are as follows (with a small explanation if the name isn't clear):
+ *
+ * - %Time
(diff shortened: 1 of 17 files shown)


src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -0,0 +1,298 @@
...
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
+#include "Options/Options.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+#include "Parallel/Printf.hpp"
```
wthrowe: Remove.

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -0,0 +1,298 @@
...
+ * - MinDeltaR: The minimum of the `StrahlkorperGr::radial_distance` between the
+ *   horizon and the excision surfaces.
+ * - MinRelativeDeltaR: MinDeltaR divided by the `Strahlkorper::average_radius`
+ *   of the horizon
+ * - ControlErrorDeltaR: \f$ \dot{S}_{00} (\lambda_{00} -
+ * r_{\mathrm{excision}}^{\mathrm{grid}} / Y_{00}) / S_{00} - \dot{\lambda}_{00}
+ * \f$
+ * - TargetCharSpeed
+ * - MinCharSpeed
+ * - MinComovingCharSpeed: Eq. 98 in ArXiv:1211.6079
```
wthrowe: Add bibliography link.

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -0,0 +1,298 @@
...
+ * - CharSpeedCrossingTime: %Time at which the min char speed is predicted to
+ *   cross zero and become negative.
+ * - ComovingCharSpeedCrossingTime: %Time at which the min comoving char speed
+ * is predicted to cross zero and become negative.
```
wthrowe: [optional] Your indentation of the bullet continuation lines is inconsistent throughout this list.

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -0,0 +1,298 @@
...
+  /*!
+   * \brief Initializes the `intrp::ZeroCrossingPredictor`s with the number of
+   * times passed in (default 3).
+   *
+   * \details The internal `control_system::size::Info::state` is initialized to
+   * `control_system::size::States::Initial`. All zero crossing predictors are
+   * initialized with a minimum number of times 3. The input argument
+   * `max_times` represents the maximum number of times.
+   */
+  Size(const int max_times = 3) {
```
wthrowe: [optional] Move the non-templated functions into a cpp file.  (The class template argument can be explicitly instantiated.)

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -0,0 +1,298 @@
...
+
+  /*!
+   * \brief Check if the `control_system::size::control_error` has decided to
+   * switch states.
+   *
+   * \details If it has decided to switch states, reset the stored
+   * `control_system::size::Info` and return true. If it is staying in the same
+   * state, do nothing and return false.
+   */
+  bool discontinuous_change_has_occurred() {
```
wthrowe: While it is documented, it is surprising to me that this function modifies the internal state of the class.  It looks like it should just be querying the state.  What is the reason for this behavior?

If this should be happening, test it.

wthrowe (changes requested):

src/ControlSystem/ControlErrors/Size.hpp
knelli2: So if a discontinuous change has a occurred, 2 things need to happen

1. Stuff inside the internal `info_` needs to be reset
2. Stuff outside the control error (specifically the Averager) also needs to be reset.

I can add a separate function to the control error that resets things and make this function just query the state. Either way, it's actually hard to test this because we never publicly expose the internal `info_` because nothing outside of this class (except maybe a test) needs to know what's happening. I'm not sure I want to add a member function just for testing purposes. What do you think?

knelli2: Actually never-mind. I already expose the two things that are reset (suggested timescale and discontinuous change) so I can test the new reset function.

tests/Unit/ControlSystem/ControlErrors/Test_SizeError.cpp
```diff
@@ -161,6 +165,25 @@ void test_size_error(const double grid_excision_boundary_radius,
...
+  auto error = TestHelpers::test_creation<size_error>(
+      "MaxNumTimesForZeroCrossingPredictor: 4");
+
+  CHECK_FALSE(error.get_suggested_timescale().has_value());
+  CHECK_FALSE(error.discontinuous_change_has_occurred());
+
+  error.reset();
+
+  CHECK_FALSE(error.get_suggested_timescale().has_value());
+  CHECK_FALSE(error.discontinuous_change_has_occurred());
```
wthrowe: Is it possible to test a situation where the class actually does something?

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -0,0 +1,253 @@
...
+#include "DataStructures/DataBox/Prefixes.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Creators/Tags/Domain.hpp"
+#include "Domain/Structure/ObjectLabel.hpp"
+#include "IO/Observer/ReductionActions.hpp"
+#include "NumericalAlgorithms/Interpolation/ZeroCrossingPredictor.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
+#include "Options/Options.hpp"
```
wthrowe: Update for #5090.

wthrowe (changes requested):
That could have benefited from a fixup commit.

tests/Unit/ControlSystem/ControlErrors/Test_SizeError.cpp
knelli2: Technically yes, but it'd be a lot of work. I'd have to set up a cache and all the measurements with very specific values so that 1) I don't get an FPE and 2) the states actually suggest changing to a different state so you get a discontinuous change. All these combinations for switching states are tested in the `control_system::size::control_error` function, which this class calls, which is why I didn't copy that work here.

knelli2:
Squashed in the option string change. If you want I can copy over one of the cases from the control error function test that will make the state change

tests/Unit/ControlSystem/ControlErrors/Test_SizeError.cpp
wthrowe: If it's not simple you don't need to test a state change, since that's done elsewhere, but `operator()` is about half the code in the class.  You should at least call it, even if you can only do rough sanity checking on the result.

knelli2:
@wthrowe pushed a fixup with changes to the test to call the operator

wthrowe:
That looks good.  Go ahead and squash.

wthrowe (approved):

wthrowe:
Ignoring some timeouts in IO tests.

PR #5090, opened 2023-06-09
Factor some code from Options.hpp into their own files

kidder:
Factored out Options::String, Options::Context, and PARSE_ERROR. 
Reduced rebuilt targets from touching Options.hpp from 1854 to 1255

### Upgrade instructions

- If you use Options::String, include Options/String.hpp instead of Options/Options.hpp
- If you use Options::Context, include Options/Context.hpp instead of Options/Options.hpp
- If you use PARSE_ERROR, include Options/ParseError.hpp instead of Options/Options.hpp

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu (approved):
Merging quickly to avoid conflicts with these many files

PR #5093, opened 2023-06-10
Make db::mutate_apply and db::mutate consistent

nilsvu:
## Proposed changes

Make the order of arguments the same. Choosing the same order as `std::apply`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


wthrowe (approved):
Thank you.  That was always annoying.

Merging now because this is going to conflict with everything.

nilsdeppe:
The argument order is backwards to our standard... `gsl::not_null` should be first.

nilsvu:
I chose consistency with `std::apply` here. Also I think for `db::mutate_apply` this order was chosen because the lambda can be specified as a template parameter instead.

nilsdeppe:
Okay, that's reasonable. Thanks for the explanation!

PR #5096, opened 2023-06-13
Print backtraces on Blaze errors

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupPch.cmake
+++ b/cmake/SetupPch.cmake
@@ -47,6 +47,7 @@ target_link_libraries(
   Brigand
   Charmxx::charmxx
   Charmxx::pup
+  ErrorHandling
   HDF5::HDF5
   SpectreFlags
   SpectreKokkos
--- a/cmake/SpectreAddLibraries.cmake
+++ b/cmake/SpectreAddLibraries.cmake
@@ -3,7 +3,17 @@
 
 add_custom_target(libs)
 
+# Wraps CMake's `add_library` to handle PCH, compiler flags, etc.
+#
+# Arguments:
+# - NO_PCH (optional): Don't use PCH (precompiled headers). Use this option to
+#   exclude PCH from libraries that the PCH depend on.
+#
+# All additional arguments are forwarded to `add_library`.
 function(ADD_SPECTRE_LIBRARY LIBRARY_NAME)
+  cmake_parse_arguments(PARSE_ARGV 1 ARG "NO_PCH" "" "")
+  list(REMOVE_ITEM ARGN NO_PCH)
+
   add_library(${LIBRARY_NAME} ${ARGN})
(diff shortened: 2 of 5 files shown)


nilsdeppe (dismissed):

kidder:
What bug is this fixing?  develop builds fine for me...

nilsvu:
I came across this cyclic dependency when editing cmake files

kidder:
so how do I reproduce it?  I ask because I am changing what is included in the pre-compiled header

nilsvu:
This commit should do it: https://github.com/nilsvu/spectre/commit/03e122c40887a301a9d221295848a78c2fbb342e

nilsvu:
@kidder ok to merge this?

kidder:
I think the correct solution is to remove `#include <Utilities/ErrorHandling/Assert.hpp>` from the precompiled header

nilsvu:
Would be nice to use it to get backtraces from Blaze though (see commit above). I currently get unhelpful exception messages like `std::out_if_range` and can't easily jump into gdb to catch them. The commit above helped enormously with debugging.

wthrowe:
I don't think this is a bug on develop (except maybe for the Assert.hpp thing).

I like the idea of using ERROR in Blaze, but that's introducing the dependency, not triggering something that is already incorrect, so I think this change should be part of a future PR that does that.

nilsdeppe:
Blaze errors trigger a `SIGTRAP` in debug mode, nothing in release. Or, at least that's what should happen: https://github.com/sxs-collaboration/spectre/blob/23710349ae4b59f7dcffc35ca93ec14b520ad2d4/cmake/SetupBlaze.cmake#L124-L146

nilsvu:
@wthrowe Ok I pushed the Blaze error handling commit to this PR since that's the real issue.

@nilsdeppe Yep my understanding is that Blaze errors print as plain exceptions (no backtrace) but can be caught in gdb. I had issues with gdb/lldb both on Wheeler (loading symbols segfaults for the BBH exec) and on my M2 Mac (lldb triggers on some EXC_BAD_INSTRUCTION). Also it's a lot more convenient to get a printed backtrace on exceptions right away, in particular with simulations that may take a while to rerun in gdb. Without gdb/lldb it took me a very long time to find the Blaze instruction that threw the exception, but with the printed backtrace I found it immediately. It's still disabled in Release mode.

kidder (changes requested):
This commit does not address an issue with this PR, it just avoids a symptom.
The issue is that with this PR, the tests are finding our stream operators for std containers, instead of using the catch stream operator which will print something (I don't recall exactly what, maybe `-`) if something is not streamable.

knelli2:
@nilsvu will look into this. See `Utilities/BlazeExceptions.hpp`

PR #5099, opened 2023-06-15
Make transpose use SIMD

nilsdeppe:
## Proposed changes

This speeds up the transpose and partial derivatives by significantly reducing memory accesses during transposition. Partial derivatives runtime on an 8^3 grid is reduced by 33%, so a 1.5x speedup.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/CMakeLists.txt
+++ b/src/DataStructures/CMakeLists.txt
@@ -20,6 +20,7 @@ spectre_target_sources(
   LeviCivitaIterator.cpp
   SliceIterator.cpp
   StripeIterator.cpp
+  Transpose.cpp
   )
 
 spectre_target_headers(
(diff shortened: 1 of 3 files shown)


wthrowe:
I'll trust that this implementation is the right way to do it.  The test shows that it works.

I think you've managed to completely avoid ODR-using the AVX/SSE2 specializations of `transpose_block` when those features are not supported.  If so, you could just `#if` the entire functions out instead of replacing the body with an error.  That should also reduce the number of warnings somewhat.  (I'm not going to require this PR to be clang-tidy clean, but easy improvements are nice.)

nilsdeppe:
Agreed! I rebased and changed the macro locations so that the warning push/pop isn't needed. I also added a `NOLINTBEGIN`, which was added in clang-14 (I think CI is not that new), so at least in the future these will go away. I figure add a `NOLINTNEXTLINE` would be quite tedious both for reading and writing since half the lines would have that...

wthrowe (approved):

wthrowe:
Ignoring clang-tidy as discussed above.

PR #5106, opened 2023-06-20
Support segments in PlotPowerMonitors

nilsvu:
## Proposed changes

Remove assumption that all volfiles contain the same observation IDs.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/ReadH5.py
+++ b/src/Visualization/Python/ReadH5.py
@@ -1,10 +1,13 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
-from typing import List, Union
+import logging
+from typing import Iterable, List, Tuple, Union
 
 import h5py
 import numpy as np
+
+logger = logging.getLogger(__name__)
 
 
 def available_subfiles(h5file: h5py.File, extension: str) -> List[str]:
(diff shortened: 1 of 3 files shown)


src/Visualization/Python/ReadH5.py
```diff
@@ -69,3 +72,73 @@ def to_dataframe(
...
+    volfiles: Union["spectre.IO.H5.H5Vol", Iterable["spectre.IO.H5.H5Vol"]],
+    step: int = None,
+    time: float = None,
+) -> Tuple[int, float]:
+    """Select an observation in the 'volfiles'
+
+    Arguments:
+      volfiles: Open spectre H5 volume files. Can be a single volfile or a list,
+        but can also be an iterator that opens and closes the files on demand.
+        The volfiles are assumed to be ordered in time.
```
nilsdeppe: I'm not exactly understanding what you mean by ordered in time here. Couldn't I have a bunch of files from different nodes that are "at the same time"? Or some might not have a time at all that others have? So how do I order those? And which times exist and don't could change during the simulation (single segment), that is, file 0 might have time A and file 1 doesn't, but file 1 has time B and file 0 doesn't.

nilsdeppe (changes requested):
You can squash any changes immediately

src/Visualization/Python/ReadH5.py
nilsvu: Tried to clarify in a fixup. My intention is to avoid reading all h5files from all segments by assuming that segments are ordered in time.

nilsdeppe: Okay, this makes sense. Thanks! Please squash :)

nilsdeppe (approved):

PR #5111, opened 2023-06-21
Print message upon success of validate CLI

knelli2:
## Proposed changes

A success message is always helpful.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/ValidateInputFile.py
+++ b/tools/ValidateInputFile.py
@@ -106,7 +106,6 @@ def validate_input_file(
     found_hints = False
     path = []
     line_number = None
-    col = None
     msg = []
     for line in process.stderr.split("\n"):
         if str(input_file_path) in line:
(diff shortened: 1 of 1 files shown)


tools/ValidateInputFile.py
```diff
@@ -99,14 +99,14 @@ def validate_input_file(
     )
 
     if process.returncode == 0:
+        rich.print(f"Input file '{input_file_path}' parsed successfully.")
```
nilsvu: Print this at the end of the `validate_input_file_command`, not here. Never just print things in a Python function, that's very annoying if you call the function and can't turn off the printing (we just had this in #5058, remember?). Typically use `logging` so at least the verbosity can be controlled from the outside. But in this case the printing is part of the CLI, not the function, so just move it to the CLI command below.

nilsvu (changes requested):

nilsvu (approved):

PR #5115, opened 2023-06-22
Allow user to specify columns in status CLI

knelli2:
## Proposed changes

Now a user can choose from a preselected list of available columns which ones they want displayed. This can also be set in the config file so it's automatically taken care of.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/Status/Status.py
+++ b/tools/Status/Status.py
@@ -190,7 +190,8 @@ def _state_order(state):
 
 
 def _format(field: str, value: Any, state_styles: dict) -> str:
-    if pd.isnull(value):
+    # The SLURM field "NodeList" returns "None assigned" if a job is pending
+    if pd.isnull(value) or value == "None assigned":
         return "-"
     elif field == "State":
         style = {
@@ -205,6 +206,8 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
         return style.get(value, "") + str(value)
     elif field in ["Start", "End"]:
         return humanize.naturaldate(value) + " " + value.strftime("%X")
+    elif field == "Segment":
+        return str(int(value)).zfill(Segment.NUM_DIGITS)
     else:
         return str(value)
 
@@ -213,13 +216,16 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
 # is the human-readable name. These are sometimes the same
 AVAILABLE_COLUMNS = {
     "State": "State",
+    "Partition": "Queue",
     "End": "End",
     "User": "User",
     "JobID": "JobID",
     "JobName": "JobName",
+    "SegmentId": "Segment",
     "Elapsed": "Elapsed",
     "NCPUS": "Cores",
     "NNodes": "Nodes",
+    "NodeList": "NodeList",
     "WorkDir": "WorkDir",
     "Comment": "Comment",
 }
@@ -260,8 +266,11 @@ def render_status(
     columns,
     **kwargs,
 ):
+    columns_to_fetch = list(AVAILABLE_COLUMNS)
+    columns_to_fetch.remove("SegmentId")
+
     job_data = fetch_job_data(
-        AVAILABLE_COLUMNS.keys(),
+        columns_to_fetch,
         **kwargs,
     )
 
--- a/tools/Status/Status.py
+++ b/tools/Status/Status.py
@@ -190,7 +190,9 @@ def _state_order(state):
 
 
 def _format(field: str, value: Any, state_styles: dict) -> str:
-    if field == "State":
+    if pd.isnull(value):
+        return "-"
+    elif field == "State":
         style = {
             "RUNNING": "[blue]",
             "COMPLETED": "[green]",
@@ -202,12 +204,50 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
         style.update(state_styles)
         return style.get(value, "") + str(value)
     elif field in ["Start", "End"]:
-        if pd.isnull(value):
-            return "-"
-        else:
-            return humanize.naturaldate(value) + " " + value.strftime("%X")
+        return humanize.naturaldate(value) + " " + value.strftime("%X")
     else:
         return str(value)
+
+
+# Dictionary of available columns to print. Key is the SLURM name, and the value
+# is the human-readable name. These are sometimes the same
+AVAILABLE_COLUMNS = {
+    "State": "State",
+    "End": "End",
+    "User": "User",
+    "JobID": "JobID",
+    "JobName": "JobName",
+    "Elapsed": "Elapsed",
+    "NCPUS": "Cores",
+    "NNodes": "Nodes",
+    "WorkDir": "WorkDir",
+    "Comment": "Comment",
+}
+
+DEFAULT_COLUMNS = [
+    "State",
+    "End",
+    "User",
+    "JobID",
+    "JobName",
+    "Elapsed",
+    "Cores",
+    "Nodes",
+]
+
+
+def input_column_callback(ctx, param, value):
+    result = value.split(",") if value else []
+    result = [x.strip() for x in result]
+
+    for input_key in result:
+        if input_key not in AVAILABLE_COLUMNS.values():
+            raise click.BadParameter(
+                f"Input column '{input_key}' not in available columns (case"
+                f" sensitive) {list(AVAILABLE_COLUMNS.values())}"
+            )
+
+    return result
 
 
 @rich.console.group()
@@ -217,21 +257,11 @@ def render_status(
     show_deleted,
     show_all_segments,
     state_styles,
+    columns,
     **kwargs,
 ):
     job_data = fetch_job_data(
-        [
-            "JobID",
-            "User",
-            "JobName",
-            "NCPUS",
-            "NNodes",
-            "Elapsed",
-            "End",
-            "State",
-            "WorkDir",
-            "Comment",
-        ],
+        AVAILABLE_COLUMNS.keys(),
         **kwargs,
     )
 
@@ -266,6 +296,9 @@ def render_status(
                 & (job_data["SegmentId"] != latest_segment_id)
             ]
             job_data.drop(drop_segment_jobs.index, inplace=True)
+
+    # Rename columns from SLURM name to user name
+    job_data.rename(columns=AVAILABLE_COLUMNS, inplace=True)
 
     # List most recent jobs first
     job_data.sort_values("JobID", inplace=True, ascending=False)
@@ -307,24 +340,10 @@ def render_status(
     # Add metadata so jobs can be grouped by state
     job_data["StateOrder"] = job_data["State"].apply(_state_order)
 
-    # We'll print these columns
-    standard_fields = [
-        "State",
-        "End",
-        "JobID",
-        "JobName",
-        "Elapsed",
-        "NCPUS",
-        "NNodes",
-    ]
-    if kwargs["allusers"]:
-        standard_fields.insert(2, "User")
-    # Transform some column names for better readability
-    col_names = {
-        "NCPUS": "Cores",
-        "NNodes": "Nodes",
-    }
-    standard_columns = [col_names.get(col, col) for col in standard_fields]
+    # Remove the user column unless we're specifying all users
+    if not kwargs["allusers"]:
+        if "User" in columns:
+            columns.remove("User")
 
     # Group output by executable
     first_section = True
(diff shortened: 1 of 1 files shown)


tools/Status/Status.py
```diff
@@ -206,32 +215,83 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
             return "-"
         else:
             return humanize.naturaldate(value) + " " + value.strftime("%X")
+    elif field == "SegmentId":
+        if pd.isnull(value):
+            return "-"
+        else:
+            return f"Segment_{int(value):04}"
```
knelli2: I know this needs to be done better, but I couldn't figure out how to get the number of digits from the Segment class.

nilsvu: I'd say move the `pd.isnull` down to the catch-all `else` clause. The segment number can just render as an integer, no need to prefix `Segment_`.

tools/Status/Status.py
```diff
@@ -506,9 +555,39 @@ def render_status(
...
+    "--available-columns",
+    "-e",
+    is_flag=True,
+    help="Print a list of all available columns to use.",
+)
+def status_command(refresh_rate, available_columns, **kwargs):
     """Gives an overview of simulations running on this machine."""
 
+    if available_columns:
+        rich.print(f"Available columns: {list(_available_fields().keys())}")
```
nilsvu:
```suggestion
        rich.print(rich.columns.Columns(_available_fields().keys()))
```

No need to print "Available columns: ". The user knows what's going on because they explicitly specified `--available-columns`.

tools/Status/Status.py
```diff
@@ -206,32 +215,83 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
+    elif field == "SegmentId":
+        if pd.isnull(value):
+            return "-"
+        else:
+            return f"Segment_{int(value):04}"
     else:
         return str(value)
 
 
+def _available_fields():
```
nilsvu: Static constants like this are implemented in Python by convention as all-caps module-level variables:

```py
AVAILABLE_FIELDS = { ... }
```

Also:
- It would be helpful to name this constant and the two new CLI options consistently as "columns" (instead of "fields") so it's clear they belong together.
- Please add a comment about what the key and the value mean (one is the Slurm field and the other is the column name).
- I suggest you use this dictionary to do a `job_data.rename(columns=AVAILABLE_COLUMNS)`. Then you don't have to generate the columns names manually. For this you have to switch the key and value.

tools/Status/Status.py
```diff
@@ -506,9 +555,39 @@ def render_status(
         "See `spectre -h` for its path."
     ),
 )
-def status_command(refresh_rate, **kwargs):
+@click.option(
+    "--columns",
+    "-c",
+    "input_columns",
+    cls=PythonLiteralOption,
```
nilsvu: Just take the columns as a comma-separated list:

```py
@click.option(
     "--columns",
     "-c",
     callback=(
        lambda ctx, param, value: (
            value.split(",") if value else []
        )
    ),
)
```

nilsvu (changes requested):
Please squash directly

knelli2:
Added a fixup with the changes. I also split the changes for adding additional columns to a separate commit so now the original commit/fixup keeps the same available (and default) columns.

tools/Status/Status.py
```diff
@@ -202,36 +204,92 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
         style.update(state_styles)
         return style.get(value, "") + str(value)
     elif field in ["Start", "End"]:
-        if pd.isnull(value):
-            return "-"
-        else:
-            return humanize.naturaldate(value) + " " + value.strftime("%X")
+        return humanize.naturaldate(value) + " " + value.strftime("%X")
+    elif field == "Segment":
+        return str(int(value)).zfill(4)
```
nilsvu: Just delete so the number prints as an int (no need for the zfill)

tools/Status/Status.py
```diff
@@ -202,36 +204,92 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
     else:
         return str(value)
 
 
+# Dictionary of available columns to print. Key is the SLURM name, and the value
+# is the human-readable name. These are sometimes the same
+AVAILABLE_COLUMNS = {
+    "State": "State",
+    "Partition": "Partition",
```
nilsvu: Print as "Queue"? That's the name of the option in `spectre schedule` (maybe it should be "partition" for consistency with slurm though).

tools/Status/Status.py
```diff
@@ -202,36 +204,92 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
+    return columns
+
+
+# Returns a list of user columns to use
+def _default_columns_for_print():
+    keys = ""
+    for key in _default_columns().values():
+        keys += f"{key},"
+    keys = keys[:-1]
+    return keys
```
nilsvu: This is just `",".join(DEFAULT_COLUMNS)` (yes, Python is easier than C++)

tools/Status/Status.py
```diff
@@ -202,36 +204,92 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
+
+
 @rich.console.group()
 def render_status(
     show_paths,
     show_unidentified,
     show_deleted,
     show_all_segments,
     state_styles,
+    input_columns,
```
nilsvu: Just name this `standard_columns` then you can drop some of the changes below.

tools/Status/Status.py
```diff
@@ -202,36 +204,92 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
+# Returns a list of user columns to use
+def _default_columns_for_print():
+    keys = ""
+    for key in _default_columns().values():
+        keys += f"{key},"
+    keys = keys[:-1]
+    return keys
+
+
+def _verify_columns(input_columns: Sequence[str]):
```
nilsvu: I think just delete this function and move the assert to the callback below. Let the user reorder the columns if they want :)

tools/Status/Status.py
```diff
@@ -506,9 +553,35 @@ def render_status(
         "See `spectre -h` for its path."
     ),
 )
-def status_command(refresh_rate, **kwargs):
+@click.option(
+    "--columns",
+    "-c",
+    "input_columns",
```
nilsvu: `standard_columns` (`input` is not a very useful name)

tools/Status/Status.py
```diff
@@ -506,9 +553,35 @@ def render_status(
...
+    callback=input_column_callback,
+    default=_default_columns_for_print(),
+    show_default=True,
+    help=(
+        "List of columns that will be printed for all jobs (executable specific"
+        " columns will be added in addition to this list). This can also be"
+        " specified in the config file with the name 'input_columns'. Note that"
+        " if the '--allusers'  option is not specified, then the \"User\""
+        " column will be omitted. Specify the columns as a comma separated list"
+        " wrapped in single quotes; e.g. 'Col1, Col2, Col3'"
```
nilsvu: You don't need quotes if you remove the spaces (user should know that they need quotes if they put spaces on the command line)

tools/Status/Status.py
```diff
@@ -202,36 +204,92 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
+    "Elapsed": "Elapsed",
+    "NCPUS": "NumCores",
+    "NNodes": "NumNodes",
+    "NodeList": "Nodes",
+    "WorkDir": "WorkDir",
+    "Comment": "Comment",
+}
+
+
+def _default_columns():
```
nilsvu: Just make this a hardcoded list `DEFAULT_COLUMNS`

tools/Status/Status.py
```diff
@@ -190,7 +190,9 @@ def _state_order(state):
 
 
 def _format(field: str, value: Any, state_styles: dict) -> str:
-    if field == "State":
+    if pd.isnull(value) or value == "None assigned":
```
nilsvu, line 194: Which field can print as "None assigned"? Could you add a comment?

nilsvu (changes requested):

nilsvu:
You can squash everything together

tools/Status/Status.py
knelli2: But I want 4 digits to match the actual segment. I think it's easier to view if it's consistent with what's after the underscore of the segment. I'd like to leave this as is unless you have very strong objections.

tools/Status/Status.py
knelli2: I thought about using Queue. I don't have a strong preference other than consistency. If we want to keep the schedule option as Queue then I'll make this Queue.

tools/Status/Status.py
knelli2: Actually I think `input` here is more descriptive than `standard`. And also more accurate. These are columns that the user passes to the command as "inputs". There's no reason these passed in columns are "standard" by any metric (unless they use the defaults)

tools/Status/Status.py
knelli2: I'll add some docs explaining. I'll keep both possibilities though

tools/Status/Status.py
knelli2: The `NodeList` slurm filed returns "None assigned" if the job is pending. I'll add a comment

tools/Status/Status.py
nilsvu: Ok use `Segment.NUM_DIGITS` then

tools/Status/Status.py
nilsvu: you can't and shouldn't prefix every option or function argument with "input" just because it's a "user input".

I'm not attached to "standard". I chose that name to distinguish from the "extra" columns that each executable adds ("standard" = same for each exec). Probably just "columns" is fine too (but a bit misleading because more columns are added per exec).

tools/Status/Status.py
nilsvu: Ok. Use real col names I think.

tools/Status/Status.py
knelli2: How about "user_columns"

nilsvu: The intention is to distinguish from the exec-specific columns. I have no better idea than "standard" or "generic". Just plain "columns" is also fine.

tools/Status/Status.py
```diff
@@ -202,36 +204,64 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
+    "JobID": "JobID",
+    "JobName": "JobName",
+    "Elapsed": "Elapsed",
+    "NCPUS": "Cores",
+    "NNodes": "Nodes",
+    "WorkDir": "WorkDir",
+    "Comment": "Comment",
+}
+
+DEFAULT_COLUMNS = {
```
nilsvu: just a list of col names, no need to make this a dict

tools/Status/Status.py
```diff
@@ -202,36 +204,64 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
+}
+
+
+def input_column_callback(ctx, param, value):
+    result = value.split(",") if value else []
+    result = [x.strip() for x in result]
+
+    for input_key in result:
+        if input_key not in AVAILABLE_COLUMNS.values():
+            raise ValueError(
```
nilsvu: `ValueError` -> `click.BadParameter`

tools/Status/Status.py
```diff
@@ -213,13 +216,16 @@ def _format(field: str, value: Any, state_styles: dict) -> str:
...
     "End": "End",
     "User": "User",
     "JobID": "JobID",
     "JobName": "JobName",
+    "SegmentId": "Segment",
     "Elapsed": "Elapsed",
-    "NCPUS": "Cores",
-    "NNodes": "Nodes",
+    "NCPUS": "NumCores",
+    "NNodes": "NumNodes",
```
nilsvu: Note that there's limited horizontal space in the terminal and the "Num" prefix makes the col name a bit longer. maybe keep the shorter "Cores" and "Nodes" and make the one below "NodeList" like in Slurm? Cores and nodes are short integers so the col width is set by the col name, whereas "NodeList" is longer so a longer col name doesn't limit the col width.

nilsvu (changes requested):

tools/Status/Status.py
nilsvu: `columns_to_use` seems no better than just `columns` 😅

nilsvu: I feel like I'm getting too pedantic here though so if you prefer that name for some reason then that's fine with me.

knelli2: I prefer the name that gets this merged the fastest lol. I think `columns_to_use` is good enough for this.

nilsvu: Change it to `columns` then. It's public API people set in a config file and without a reason for making it different from the `--columns` option name it's unnecessarily confusing.

knelli2: done

nilsvu (approved):

PR #5126, opened 2023-06-26
Added the pybindings for Extend_Connectivity_Data

aasthabagree21:
## Proposed changes
Added the pybindings for Extend_Connectivity_Data function and generated a small test to ensure that the python fucntion is called correctly.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/H5/Python/VolumeData.cpp
+++ b/src/IO/H5/Python/VolumeData.cpp
@@ -49,7 +49,16 @@ void bind_h5vol(py::module& m) {
            py::arg("observation_id"))
       .def("get_data_by_element", &h5::VolumeData::get_data_by_element,
            py::arg("start_observation_value"), py::arg("end_observation_value"),
-           py::arg("components_to_retrieve") = std::nullopt);
+           py::arg("components_to_retrieve") = std::nullopt)
+      .def("extend_connectivity_data_1d",
+           &h5::VolumeData::extend_connectivity_data<1>,
+           py::arg("observation_ids"))
+      .def("extend_connectivity_data_2d",
+           &h5::VolumeData::extend_connectivity_data<2>,
+           py::arg("observation_ids"))
+      .def("extend_connectivity_data_3d",
+           &h5::VolumeData::extend_connectivity_data<3>,
+           py::arg("observation_ids"));
   m.def("offset_and_length_for_grid", &h5::offset_and_length_for_grid,
         py::arg("grid_name"), py::arg("all_grid_names"),
         py::arg("all_extents"));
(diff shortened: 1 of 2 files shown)


src/IO/H5/Python/VolumeData.cpp
```diff
@@ -49,7 +50,16 @@ void bind_h5vol(py::module& m) {
            py::arg("observation_id"))
       .def("get_data_by_element", &h5::VolumeData::get_data_by_element,
            py::arg("start_observation_value"), py::arg("end_observation_value"),
-           py::arg("components_to_retrieve") = std::nullopt);
+           py::arg("components_to_retrieve") = std::nullopt)
+      .def("extend_connectivity_data_1D",
```
nilsvu: use lower case: `_1d`, `_2d`, `_3d`

src/IO/H5/Python/VolumeData.cpp
```diff
@@ -6,6 +6,7 @@
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 #include <string>
+#include <iostream>
```
nilsvu: delete

src/IO/H5/Python/ExtendConnectivityData.py
```diff
@@ -0,0 +1,62 @@
+# Distributed under the MIT License.
```
nilsvu: Please add only the bindings in this PR, not the CLI

nilsvu (commented):
You can squash the fixups

src/IO/H5/Python/ExtendConnectivityData.py
aasthabagree21: Is there any simple way that I can delete this file from the PR?

knelli2: You'll have to do an interactive rebase and then "edit" the first commit. We can do a hangout if that's helpful

aasthabagree21: I think a hangout would be helpful. Thank you!

tests/Unit/IO/H5/Test_VolumeData.py
```diff
@@ -276,13 +284,17 @@ def test_offset_and_length_for_grid(self):
...
+                grid_name="[B0(L1I0,L1I0,L1I0)]",
                 all_grid_names=all_grid_names,
                 all_extents=all_extents,
             ),
             (0, 8),
         )
 
+    def test_extend_connectivity_data_3D(self):
+        obs_ids = self.vol_file.list_observation_ids()
+        self.vol_file.extend_connectivity_data_3d(obs_ids)
```
nilsvu, line 299: Can you add a small check that this did something? Like check the size of the connectivity dataset or something like that.

nilsvu (changes requested):

nilsvu (approved):

knelli2 (approved):

PR #5136, opened 2023-06-28
Infer output buffer tags in partial_derivatives

nilsvu:
## Proposed changes

This allows the compiler to infer the template parameters in the return-by-reference overloads of partial_derivatives. It also relaxes the restriction that the output buffer must have ::Tags::deriv prefixes. This is useful to compute derivatives directly into output buffers, e.g. for variables that are represented by another tag (like Phi = deriv(Psi)) or for variables augmented by more computations (like adding a Christoffel-symbol term for the covariant derivative) or for prefixed variables where the output buffer ignores the prefixes.

This is part of merging the code published in the thermal noise paper (https://iopscience.iop.org/article/10.1088/1361-6382/acad62/meta).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/DiscontinuousGalerkin/CMakeLists.txt
+++ b/src/Elliptic/DiscontinuousGalerkin/CMakeLists.txt
@@ -31,6 +31,7 @@ target_link_libraries(
   Domain
   DomainStructure
   ErrorHandling
+  LinearOperators
   Spectral
   Utilities
   INTERFACE
--- a/src/Elliptic/DiscontinuousGalerkin/Initialization.cpp
+++ b/src/Elliptic/DiscontinuousGalerkin/Initialization.cpp
@@ -24,7 +24,6 @@
 #include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/IndexToSliceAt.hpp"
 #include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
-#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.tpp"
 #include "NumericalAlgorithms/Spectral/LogicalCoordinates.hpp"
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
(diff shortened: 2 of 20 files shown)


nilsdeppe:
It's not at all clear to me why `Tags::deriv` should be optional. Wouldn't it be better to require the derivative tags to be derivatives? This change should be justified in the commit message beyond "I did it for this paper" since it's not clear to me _why_ you needed it for the paper.

nilsvu:
Dropped the second commit for now, leaving only the minor refactor of tensor derivatives. I'll revisit the `Tags::deriv` issue later.

nilsdeppe (dismissed):
Okay, I'm not inherently opposed to it, it just wasn't clear to me why it was needed since the changes weren't used other than to simplify a few calls.

nilsvu:
Yea I think I had issues with prefix tags (I wanted to compute derivatives of prefixed variables in a buffer, but to create the buffer I didn't have the prefixes available, or didn't want to bother dealing with the prefixes).

Also I think it would be reasonable to compute derivatives directly into an output buffer, e.g. to compute something like the covariant derivative of a variable you would compute the partial derivatives first, then add the Christoffel symbol term. You wouldn't necessarily want to use `Tags::deriv` for that.

nilsdeppe:
Yea, that's the use case I could imagine, where some derivative is already defined as some other quantity. E.g. `Phi_i = d_i Psi` or something. I assumed this was the use-case in the elliptic code, and if so I think it's totally fine to generalize! Same with the covariant derivative example, also a good use-case for generalizing!

nilsvu:
Ok, restored the second commit and added details to the commit message.

nilsdeppe:
Great! Looks like CI is unhappy, but I'm happy with the commit message!

nilsvu:
Had to add some missing `partial_derivative` instantiations

nilsdeppe (approved):

PR #5138, opened 2023-06-29
Have --force delete segment directories in schedule CLI

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Python/Schedule.py
+++ b/support/Python/Schedule.py
@@ -2,6 +2,7 @@
 # See LICENSE.txt for details.
 
 import logging
+import os
 import re
 import shutil
 import subprocess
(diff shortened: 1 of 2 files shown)


support/Python/Schedule.py
```diff
@@ -420,17 +421,37 @@ def schedule(
             template_env.from_string(str(segments_dir)).render(context).strip()
         )
         all_segments = list_segments(segments_dir)
+
+        # If we are forcing, we don't care about existing segment directories.
+        # We should delete them
```
nilsvu, line 426: I don't think we want to delete _everything_. There are different scenarios. For instance, if the user specified a checkpoint in a previous segment you don't want to delete the whole segments dir and then be like "sorry your checkpoint doesn't exist anymore".

nilsvu (changes requested):

support/Python/Schedule.py
knelli2: That's fair. What do we want to delete then? The reason I added this was actually because if there's an error, some directories and files are still created. Then when you fix the error and try to run `schedule` again, you get another error that the segment or other files already exist. I'd like to avoid that

nilsvu: Just make sure that `--force` does what you expect in the different scenarios. The best strategy in my experience is to look at where exceptions are thrown and replace those with `if force` clauses.

knelli2:
@knelli2 will look into this

PR #5139, opened 2023-07-05
Add set_number_of_grid_points utility

wthrowe:
## Proposed changes

This allows resizing existing objects in a similar way to the sizing in `make_with_value`.  It fixes one case where I was copying data just to get the correct structure, and will likely be convenient in other cases.

### Upgrade instructions

Replace `destructive_resize_components` with the new `set_number_of_grid_points`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FaceNormal.cpp
+++ b/src/Domain/FaceNormal.cpp
@@ -10,21 +10,19 @@
 #include "Domain/InterfaceLogicalCoordinates.hpp"
 #include "Domain/Structure/Direction.hpp"  // IWYU pragma: keep
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
-#include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 
 template <size_t VolumeDim, typename TargetFrame>
 void unnormalized_face_normal(
     const gsl::not_null<tnsr::i<DataVector, VolumeDim, TargetFrame>*> result,
-    const Mesh<VolumeDim - 1>& interface_mesh,
+    const Mesh<VolumeDim - 1>& /*interface_mesh*/,
     const InverseJacobian<DataVector, VolumeDim, Frame::ElementLogical,
                           TargetFrame>& inv_jacobian_on_interface,
     const Direction<VolumeDim>& direction) {
   const auto sliced_away_dim = direction.dimension();
   const double sign = direction.sign();
 
-  destructive_resize_components(result, interface_mesh.number_of_grid_points());
   for (size_t d = 0; d < VolumeDim; ++d) {
     result->get(d) = sign * inv_jacobian_on_interface.get(sliced_away_dim, d);
   }
--- a/src/Utilities/MakeWithValue.hpp
+++ b/src/Utilities/MakeWithValue.hpp
@@ -9,6 +9,7 @@
 #include <array>
 #include <complex>
 #include <type_traits>
+#include <vector>
 
 #include "Utilities/Algorithm.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
@@ -140,6 +141,24 @@ struct NumberOfPoints<std::array<T, Size>> {
         alg::all_of(input,
                     [&](const T& t) { return number_of_points(t) == points; }),
         "Inconsistent number of points in array entries.");
+    return points;
+  }
+};
+
+template <typename T>
+struct NumberOfPoints<std::vector<T>> {
+  static SPECTRE_ALWAYS_INLINE size_t apply(const std::vector<T>& input) {
+    // size_t is interpreted as the number of points in other
+    // contexts, but that doesn't make sense here.
+    static_assert(not std::is_same_v<T, size_t>,
+                  "Cannot get number_of_points from non-vector.");
+    ASSERT(not input.empty(),
+           "Cannot get number of points from empty std::vector.");
+    const size_t points = number_of_points(input[0]);
+    ASSERT(
+        alg::all_of(input,
+                    [&](const T& t) { return number_of_points(t) == points; }),
+        "Inconsistent number of points in vector entries.");
     return points;
   }
 };
--- a/src/Utilities/MakeWithValue.hpp
+++ b/src/Utilities/MakeWithValue.hpp
@@ -83,7 +83,7 @@ struct MakeWithValueImpl {
 /// \tparam ValueType The type of `value`. For most containers, this will be
 /// `double`.
 ///
-/// \see MakeWithValueImpls
+/// \see MakeWithValueImpls, set_number_of_grid_points
 template <typename R, typename T, typename ValueType>
 SPECTRE_ALWAYS_INLINE std::remove_const_t<R> make_with_value(
     const T& input, const ValueType& value) {
--- a/src/Utilities/MakeWithValue.hpp
+++ b/src/Utilities/MakeWithValue.hpp
@@ -10,6 +10,8 @@
 #include <complex>
 #include <type_traits>
 
+#include "Utilities/Algorithm.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/ForceInline.hpp"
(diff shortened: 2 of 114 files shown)


src/Utilities/GenericResize.hpp
```diff
@@ -0,0 +1,88 @@
...
+/// Change the number of grid points in an object.
+///
+/// Change the number of points stored in \p result to a given value
+/// or to match another object.  If \p pattern is a `size_t` it will
+/// be used as the number of points, otherwise it will be interpreted
+/// as data on a collection of grid points.
+///
+/// \see GenericResizeImpls, make_with_value
+template <typename T, typename U>
+void generic_resize(const gsl::not_null<T*> result, const U& pattern) {
```
nilsdeppe: I think GCC is unhappy because the arguments may be unused according to it...

nilsdeppe (commented):
LGTM, I think just a minor thing to get GCC to be happy. Thanks for doing this!

src/Utilities/GenericResize.hpp
```diff
@@ -0,0 +1,88 @@
...
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeWithValue.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+/// Implementations of \link generic_resize \endlink
+///
+/// Specializations may be trivial, meaning the object is similar to a
+/// `double` and doesn't actually represent grid data.  The only
+/// requirement for a trivial specializations is defining `is_trivial`
```
knelli2: `for trivial specializations`

src/Utilities/GenericResize.hpp
```diff
@@ -0,0 +1,88 @@
...
+  if constexpr (not GenericResizeImpls::GenericResizeImpl<T>::is_trivial) {
+    const size_t size = MakeWithValueImpls::number_of_points(pattern);
+    if (UNLIKELY(MakeWithValueImpls::number_of_points(*result) != size)) {
+      GenericResizeImpls::GenericResizeImpl<T>::apply(result, size);
+    }
+  }
+}
+
+template <>
+struct GenericResizeImpls::GenericResizeImpl<double> {
```
knelli2: Maybe also add a specialization for `std::vector`? I think that, along with the implementations in the next commit, covers all the common containers we'd normally use.

knelli2 (changes requested):
Just a couple small things. If you decide to add the `std::vector` impl, just squash.

src/Utilities/GenericResize.hpp
wthrowe: I added the same cases that work for `make_with_value`, which I think were chosen as types with arithmetic operations.

That said, I can't think of any reason to restrict these utilities to such types, and supporting `std::vector` sounds potentially useful, so I'll add it (and also for `make_with_value`).

wthrowe (commented):
Based mostly on conversations elsewhere on comparisons with the "destructive_resize" family of functions, I'm going to rename this.  Particularly with potential support for `std::vector`, I've decided I don't like the name "resize" because it acts on the elements of containers in some cases and sometimes on the object itself.  Instead, I'm going to go with something like `set_number_of_grid_points`, since that's pretty unambiguous.

The existing function `destructive_resize_components` is a special case of this function, and can be removed in favor of it (except for the unused three-argument form).

wthrowe:
Renamed.  Replaced `destructive_resize_components` since it is effectively the same.  Did a few other things required to make that work that you can see in the history.  Removed some silly branch-prediction hints that I happened to notice.

A few functions ended up with unused arguments that had previously only been used for unnecessary sizing.  I didn't investigate in depth because this PR already touches enough random code, but I found it odd.  In case anyone is interested, they were (specific overloads of):
```
src/Domain/FaceNormal.cpp unnormalized_face_normal
src/Evolution/Systems/Cce/BoundaryData.cpp du_j_worldtube_data
src/NumericalAlgorithms/LinearOperators/PartialDerivatives.cpp partial_derivative
```
I suspect other overloads of those functions probably still have similar effectively unused arguments, because I only removed sizing calls that were very obviously unneeded.

src/Utilities/MakeWithValue.hpp
```diff
@@ -81,7 +85,7 @@ struct MakeWithValueImpl {
 /// \tparam ValueType The type of `value`. For most containers, this will be
 /// `double`.
 ///
-/// \see MakeWithValueImpls
+/// \see MakeWithValueImpls, generic_resize
```
nilsvu: `set_number_of_grid_points`

src/Domain/FaceNormal.cpp
```diff
@@ -10,21 +10,19 @@
...
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
-#include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 
 template <size_t VolumeDim, typename TargetFrame>
 void unnormalized_face_normal(
     const gsl::not_null<tnsr::i<DataVector, VolumeDim, TargetFrame>*> result,
-    const Mesh<VolumeDim - 1>& interface_mesh,
+    const Mesh<VolumeDim - 1>& /*interface_mesh*/,
```
nilsvu, line 19: Could you add a note that this is left over from a refactor and can be deleted?

nilsvu (dismissed):
Amazing! Thanks for doing this!

The two small changes below can happen in a follow-up to avoid rebasing this PR.

nilsdeppe (dismissed):

nilsvu:
CI has a linking issue:

```
/work/build/bin/RunTests: symbol lookup error: /work/build/lib/libTest_Domain.so: undefined symbol: _ZN6domain19flat_logical_metricILm1EEEvN3gsl8not_nullIP6TensorI10DataVectorN7brigand4listIJNS5_17integral_constantIiLi1EEES8_EEENS6_IJN13Tensor_detail15TensorIndexTypeIXT_EL4UpLo1EN5Frame14ElementLogicalEL9IndexType0EEESG_EEEEEERKS3_IS4_NS6_IJNS7_IiLi2EEES8_EEENS6_IJNSB_IXT_ELSC_0ENSD_8InertialELSF_0EEESG_EEEE
```

knelli2 (commented):
Looks good except for the linking issue.

Question @wthrowe: In some places, you completely remove a call to `destructive_resize_components` instead of replacing it with `set_number_of_grid_points`. Why is that?

src/Domain/FaceNormal.cpp
wthrowe: Filed #5177.

src/Utilities/MakeWithValue.hpp
wthrowe: Fixed when I next push.

wthrowe:
Added some explicit instantiations to hopefully fix the linking error.  I don't get it locally, so I'l have to wait for CI.  Also rebased on develop.

> Question @wthrowe: In some places, you completely remove a call to `destructive_resize_components` instead of replacing it with `set_number_of_grid_points`. Why is that?

For most of them the next thing done with the variable was obviously something that would check the size and resize if necessary, like assigning from a blaze expression template.  For a few the next thing was an assignment from an rvalue, in which case the resize was actually harmful.  I suspect some of the remaining ones are also unnecessary, but I wasn't checking anything where it wasn't obvious.

nilsvu (approved):

knelli2 (approved):
The singular clang-tidy error doesn't seem important enough to hold up this PR.

nilsdeppe:
CI failures are timeouts and unrelated clang-tidy complaints.

Merging now to minimize conflicts of this PR.

nilsvu:
No, it also wasn't introduced in this PR (edit: @nilsdeppe was faster)

nilsvu:
Added upgrade instructions. @wthrowe feel free to edit.

wthrowe:
Removed upgrade instructions.  After discussion elsewhere I still don't agree that they are appropriate.

PR #5141, opened 2023-07-05
Add size StateHistory

knelli2:
## Proposed changes

This is useful for repopulating the Averager with previous control errors when a discontinuous change has occurred. Also add this to the size control error. It will be used in a later PR.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size/StateHistory.cpp
+++ b/src/ControlSystem/ControlErrors/Size/StateHistory.cpp
@@ -1,0 +1,61 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ControlSystem/ControlErrors/Size/StateHistory.hpp"
+
+#include <deque>
+#include <memory>
+#include <pup.h>
+#include <pup_stl.h>
+#include <unordered_map>
+#include <utility>
+
+#include "ControlSystem/ControlErrors/Size/AhSpeed.hpp"
+#include "ControlSystem/ControlErrors/Size/DeltaR.hpp"
+#include "ControlSystem/ControlErrors/Size/Info.hpp"
+#include "ControlSystem/ControlErrors/Size/Initial.hpp"
+#include "ControlSystem/ControlErrors/Size/State.hpp"
+#include "DataStructures/DataVector.hpp"
+
+namespace control_system::size {
+StateHistory::StateHistory() { initialize_stored_control_errors(); }
+
+StateHistory::StateHistory(const size_t num_times_to_store)
+    : num_times_to_store_(num_times_to_store) {
+  initialize_stored_control_errors();
+}
+
+void StateHistory::initialize_stored_control_errors() {
+  stored_control_errors_[States::Initial{}.number()];
+  stored_control_errors_[States::DeltaR{}.number()];
+  stored_control_errors_[States::AhSpeed{}.number()];
+}
+
+void StateHistory::store(double time, const Info& info,
+                         const ControlErrorArgs& control_error_args) {
+  const auto store_state = [this, &time, &info,
+                            &control_error_args](auto state) {
+    const double control_error = state.control_error(info, control_error_args);
+    std::deque<std::pair<double, double>>& history =
+        stored_control_errors_.at(state.number());
+    history.emplace_back(time, control_error);
+    while (history.size() > num_times_to_store_) {
+      history.pop_front();
+    }
+  };
+
+  store_state(States::Initial{});
+  store_state(States::DeltaR{});
+  store_state(States::AhSpeed{});
+}
+
+const std::deque<std::pair<double, double>>& StateHistory::state_history(
+    const size_t state_number) const {
+  return stored_control_errors_.at(state_number);
+}
+
+void StateHistory::pup(PUP::er& p) {
+  p | num_times_to_store_;
+  p | stored_control_errors_;
+}
+}  // namespace control_system::size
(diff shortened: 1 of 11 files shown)


src/ControlSystem/ControlErrors/Size/StateHistory.hpp
```diff
@@ -0,0 +1,61 @@
...
+   * \param info `control_system::size::Info`
+   * \param control_error_args `control_system::size::ControlErrorArgs`
+   */
+  void store(double time, const Info& info,
+             const ControlErrorArgs& control_error_args);
+
+  /*!
+   * \brief Return a copy of the stored control errors from all the states. (We
+   * copy rather than return a reference so that the user can choose if they
+   * want to keep the most recent entry, which is at the "back")
```
wthrowe: How does returning a copy help with that?  If the caller wants a version to modify they can copy from a reference.

If you do keep the behavior, this seems a bit long and detailed for a "brief".

src/ControlSystem/ControlErrors/Size/StateHistory.cpp
```diff
@@ -0,0 +1,71 @@
...
+StateHistory::StateHistory(const size_t num_times_to_store)
+    : num_times_to_store_(num_times_to_store) {
+  initialize_stored_control_errors();
+}
+
+void StateHistory::initialize_stored_control_errors() {
+  const auto initialize_state = [this](auto state) {
+    using StateType = std::decay_t<decltype(state)>;
+    std::unique_ptr<State> current_state = std::make_unique<StateType>();
+    stored_control_errors_[current_state->number()];
```
wthrowe: Just `state.number()`.  After that this lambda is only one line, so you can just inline it below.

src/ControlSystem/ControlErrors/Size/StateHistory.cpp
```diff
@@ -0,0 +1,71 @@
...
+  initialize_state(States::DeltaR{});
+  initialize_state(States::AhSpeed{});
+}
+
+void StateHistory::store(double time, const Info& info,
+                         const ControlErrorArgs& control_error_args) {
+  const auto store_state = [this, &time, &info,
+                            &control_error_args](auto state) {
+    using StateType = std::decay_t<decltype(state)>;
+    std::unique_ptr<State> current_state = std::make_unique<StateType>();
```
wthrowe: As above, just use `state` instead of doing an extra allocation.

src/ControlSystem/ControlErrors/Size/StateHistory.cpp
```diff
@@ -0,0 +1,71 @@
...
+    double control_error =
+        current_state->control_error(info, control_error_args);
+    size_t state_number = current_state->number();
+    stored_control_errors_.at(state_number)
+        .emplace_back(std::make_pair(time, control_error));
```
wthrowe: I think `make_pair` is unnecessary, and it undoes the advantage of `emplace_back` over `push_back`.

src/ControlSystem/ControlErrors/Size/StateHistory.cpp
```diff
@@ -0,0 +1,71 @@
...
+void StateHistory::store(double time, const Info& info,
+                         const ControlErrorArgs& control_error_args) {
+  const auto store_state = [this, &time, &info,
+                            &control_error_args](auto state) {
+    using StateType = std::decay_t<decltype(state)>;
+    std::unique_ptr<State> current_state = std::make_unique<StateType>();
+    double control_error =
+        current_state->control_error(info, control_error_args);
+    size_t state_number = current_state->number();
+    stored_control_errors_.at(state_number)
```
wthrowe: [optional] Store a reference to `stored_control_errors_.at(state_number)` so you don't do the map lookup four times.

src/ControlSystem/ControlErrors/Size/StateHistory.cpp
```diff
@@ -0,0 +1,71 @@
...
+    }
+  };
+
+  store_state(States::Initial{});
+  store_state(States::DeltaR{});
+  store_state(States::AhSpeed{});
+}
+
+std::deque<std::pair<double, double>> StateHistory::state_history(
+    size_t state_number) const {
```
wthrowe: `const` argument.

tests/Unit/ControlSystem/ControlErrors/Test_StateHistory.cpp
```diff
@@ -0,0 +1,90 @@
...
+void test_state_history(const size_t num_times_to_store) {
+  CAPTURE(num_times_to_store);
+  Info info{
+      std::make_unique<States::Initial>(), 1.0, 1.0, 1.0, std::nullopt, false};
+  ControlErrorArgs control_error_args{1.0, 1.0, 1.0, 1.0};
+
+  StateHistory state_history{num_times_to_store};
+
+  std::vector<double> times(num_times_to_store);
+  std::iota(times.begin(), times.end(), 0.0);
```
wthrowe: This vector seems to be a very complicated way to write `static_cast<double>`.  Is there some reason to have it?

wthrowe (changes requested):

wthrowe (approved):
Squash.

knelli2:
@wthrowe Squashed and rebased.

wthrowe (approved):

Issue #5145, opened 2023-07-06
Unit.Domain.Domain versioning test can fail because of floating-point optimizations


wthrowe:
```
/home/wthrowe/spectre/tests/Unit/Domain/Test_Domain.cpp:456: FAILED:
  CHECK( domain == expected_domain_creator.create_domain() )
with expansion:
  <uninformative text omitted>
```
This is in the `test_versioning()` function, checking that the domain stored in an H5 file matches an expected value.  On my machine with gcc 12, however, the compiler optimizes the constructor of the expected domain (in particular, by adding FMAs to the calculation of the `radius_` of the Frustum map) to give roundoff differences from the value stored in the file.  The test passes with `-mno-fma`.

We probably can't rely on bitwise-identical objects to be created in different builds if they do any significant work in their constructors, which I don't think is a problem in most cases.  I don't know what we should do about the test, though.

wthrowe:
Some example error messages from #6528:
```
-------------------------------------------------------------------------------
Unit.Domain.FunctionsOfTime.QuaternionFunctionOfTime
-------------------------------------------------------------------------------
/home/alexcarpenter/spectre/tests/Unit/Domain/FunctionsOfTime/Test_QuaternionFunctionOfTime.cpp:105
...............................................................................

/home/alexcarpenter/spectre/tests/Unit/Helpers/Utilities/Serialization/Versioning.hpp:118: FAILED:
  CHECK( to_compare(deserialize<Serialize>(serialization.data())) == to_compare(*std::get<const Serialize*>(expected)) )
with expansion:
  Quaternion:
  t=0: (1,0,0,0)
  t=0.6: (0.423039,0,0,0.906112)
  backlog=()
  Angle:
  t=0: (0,0,0) (0,0,3.78) (0,0,0)
  t=0.6: (0,0,2.268) (0,0,3.78) (0,0,0)
  backlog=()
  ==
  Quaternion:
  t=0: (1,0,0,0)
  t=0.6: (0.423039,0,0,0.906112)
  backlog=()
  Angle:
  t=0: (0,0,0) (0,0,3.78) (0,0,0)
  t=0.6: (0,0,2.268) (0,0,3.78) (0,0,0)
  backlog=()
with message:
  version := "version 5"

/home/alexcarpenter/spectre/tests/Unit/Helpers/Utilities/Serialization/Versioning.hpp:118: FAILED:
[...]

PR #5149, opened 2023-07-07
Silence gcc 12 and 13 warnings

wthrowe:
At least on my machine, there is still a runtime test failure as described in #5145.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Options/Auto.hpp
+++ b/src/Options/Auto.hpp
@@ -43,6 +43,18 @@ class Auto {
   Auto() = default;
   explicit Auto(T value) : value_(std::move(value)) {}
 
+  // These lines are just to work around a spurious warning.
+  Auto(Auto&&) = default;
+  Auto& operator=(Auto&&) = default;
+#if defined(__GNUC__) and not defined(__clang__) and __GNUC__ >= 12 and __GNUC__ < 14
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wmaybe-uninitialized"
+#endif
+  ~Auto() = default;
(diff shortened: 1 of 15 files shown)


src/Options/Auto.hpp
```diff
@@ -43,6 +43,18 @@ class Auto {
   Auto() = default;
   explicit Auto(T value) : value_(std::move(value)) {}
 
+  // These lines are just to work around a spurious warning.
+  Auto(Auto&&) = default;
+  Auto& operator=(Auto&&) = default;
+#if defined(__GNUC__) and not defined(__clang__) and __GNUC__ >= 12 and __GNUC__ < 13
```
knelli2: Isn't this just `__GNUC__ == 12`? (if so, same for the rest of this PR)

wthrowe: It is.  I often write these things as `>= first_broken and < first_optimistically_fixed` (optimistically assuming untested versions have been fixed).  In this case with only one version it's kind of silly, so I'll change it if you like.

knelli2: If you could, yeah so it's obvious it's just this one GCC version. If in the future, we find this is an issue with GCC 13 we can go back to this.

wthrowe:
Changed the conditional formatting, and decided while I was at this I'd do gcc 13 too (12.2.1 and 13.1.1, specifically).  Version 13 actually caught some minor bugs.

knelli2 (approved):
Looks good. Not sure what the python formatting errors are about. Is this rebased on latest develop?

kidder:
Ignoring python formatting issue (#5170)

PR #5150, opened 2023-07-07
Add standard trigger for separation between objects

knelli2:
## Proposed changes

Checks the distance between `domain::Tags::ObjectCenter<A>` and `B` in the inertial. If it's less than the input, then it's triggered.

This is a rough start to finding common horizons (or a horizon for BNS). It will certainly need to be modified/updated in the future (probably to a DenseTrigger even)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/Triggers.hpp
```diff
@@ -0,0 +1,98 @@
...
+    static constexpr Options::String help = {
+        "Separation of the two horizons to compare against."};
+  };
+
+  using options = tmpl::list<Value>;
+  static constexpr Options::String help{
+      "Trigger when the separation between the two horizons is less than a "
+      "certain distance."};
+
+  SeparationLessThan(double separation);
```
wthrowe: `explicit`

src/Domain/Triggers.cpp
```diff
@@ -0,0 +1,81 @@
...
+    const auto& grid_to_inertial_map = block.moving_mesh_grid_to_inertial_map();
+
+    inertial_object_center_a =
+        grid_to_inertial_map(object_center_a, time, functions_of_time);
+    inertial_object_center_b =
+        grid_to_inertial_map(object_center_b, time, functions_of_time);
+  }
+
+  // reuse allocation
+  tnsr::I<double, 3, Frame::Inertial>& position_difference =
```
wthrowe: There is no allocation.

src/Domain/Triggers.cpp
```diff
@@ -0,0 +1,81 @@
...
+      inertial_object_center_a;
+
+  position_difference = tenex::evaluate<ti::I>(inertial_object_center_a(ti::I) -
```
wthrowe: If there were an allocation, you would have to use the by-reference version to reuse it.  (But there isn't so this is fine.)

src/Domain/Triggers.cpp
```diff
@@ -0,0 +1,81 @@
...
+    const double time, const ::Domain<3>& domain, const Element<3>& element,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time,
+    const tnsr::I<double, 3, Frame::Grid>& object_center_a,
+    const tnsr::I<double, 3, Frame::Grid>& object_center_b) const {
+  const ElementId<3>& element_id = element.id();
+  const size_t block_id = element_id.block_id();
+
+  const Block<3>& block = domain.blocks()[block_id];
```
wthrowe: I'm confused about what's going on here.  It looks like you are using the maps from the current block to map the object centers.  Are these maps guaranteed to be the same for every element?

src/Domain/Triggers.hpp
```diff
@@ -0,0 +1,98 @@
...
+#include <pup.h>
+#include <string>
+#include <unordered_map>
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/Structure/ObjectLabel.hpp"
+#include "Options/String.hpp"
+#include "ParallelAlgorithms/EventsAndTriggers/Trigger.hpp"
+#include "Time/Utilities.hpp"
```
wthrowe: You don't use this.

src/Domain/Triggers.hpp
```diff
@@ -0,0 +1,98 @@
+// Distributed under the MIT License.
```
wthrowe: I think this should be in a less generically named file.

wthrowe (changes requested):

src/Domain/Triggers.cpp
knelli2: So I wasn't sure how to do this because we literally don't have elements at the centers of the horizons. So I can't check if the center is in an element/block to specifically use that map. But this trigger needs to be run on all elements. My solution was to make use of the fact that currently, all Grid->Distorted maps leave the centers of the objects unchanged, and all Distorted->Inertial maps are global. So if a block didn't have a distorted frame, I just call the Grid->Inertial map because I know they are (currently) defined everywhere. And if a block has a distorted frame, I just call the Distorted->Inertial map which (currently) should be the same as the Grid->Inertial map without a distorted frame.

This is all very hacky. But I couldn't think of a better way. Ideas are welcome

src/Domain/Triggers.hpp
nilsvu: Maybe `Evolution/Triggers/SeparationLessThan`?

src/Domain/Triggers.cpp
wthrowe: OK, it looks like that's true in SpEC, so it should be fine here, at least for now.

src/Domain/Triggers.hpp
wthrowe: Yes, let's get it out of Domain to avoid a new dependency there.

src/Domain/Triggers.cpp
knelli2: Yeah while we have Shape/Size/Rotaiton/Expansion/Translation, this will be true. Once we add Skew/Cut-X it won't be and we'll need to find a different way. Possibly we add some kind of thing that holds coordinate maps for inside the excision, except it doesn't actually have grid points/solutions.

But for now, I'll leave it as is.

wthrowe: Oh, you're right.  I misread the SpEC paper.  How does SpEC do this?

knelli2: I honestly don't know. @markscheel?

nilsdeppe: SpEC's maps are all global and then if-else branches depending on coordinate locations inside the calls

knelli2: Ah ok yeah that explains it. I'm thinking that we just have to imbue some object with coordinate maps that are valid inside the excision spheres. Maybe the `ExcisionSphere`s?

nilsvu: Yea I think adding coord maps to excision spheres would be reasonable. I think @nikwit needs something like that as well for worldtube things.

wthrowe: I agree, that makes sense.  Excision spheres are kind of like blocks in that they are covering a section of the grid-coordinates, just not a section with elements, so having some similar properties is reasonable.

knelli2: Ok I'll add maps to the `ExcisionSphere`s (probably in a separate PR).

wthrowe: Sounds good.  I don't think we need to hold this PR up for it.  Just fix it in the PR that adds the feature.  If you're not going to do it immediately file a bug so we don't forget.

knelli2:
Made (and squashed) all the changes other than the new excision sphere stuff which I'll do in a new PR. Added this trigger to a new `EvolutionTriggers` library.

src/Evolution/Triggers/SeparationLessThan.cpp
```diff
@@ -0,0 +1,79 @@
...
+  } else {
+    const auto& grid_to_inertial_map = block.moving_mesh_grid_to_inertial_map();
+
+    inertial_object_center_a =
+        grid_to_inertial_map(object_center_a, time, functions_of_time);
+    inertial_object_center_b =
+        grid_to_inertial_map(object_center_b, time, functions_of_time);
+  }
+
+  // reuse allocation
```
wthrowe: Remove comment.  Still not true.

wthrowe (changes requested):

wthrowe (approved):

src/Evolution/Triggers/SeparationLessThan.cpp
```diff
@@ -0,0 +1,78 @@
...
+    const tnsr::I<double, 3, Frame::Grid>& object_center_b) const {
+  const ElementId<3>& element_id = element.id();
+  const size_t block_id = element_id.block_id();
+
+  const Block<3>& block = domain.blocks()[block_id];
+
+  tnsr::I<double, 3, Frame::Inertial> inertial_object_center_a{};
+  tnsr::I<double, 3, Frame::Inertial> inertial_object_center_b{};
+
+  if (block.has_distorted_frame()) {
```
nilsvu, line 38: I think you don't need this because the grid-to-inertial map is available even when you have a distorted frame. Please double-check.

nilsvu (changes requested):

src/Evolution/Triggers/SeparationLessThan.cpp
knelli2: Yeah but the Shape map isn't global, so it isn't valid inside the excision sphere. Only expansion and rotation are. So in a block with a distorted frame, we can't call the shape map, so we have to call the distorted to inertial map. However, all of this is mute with #5184.

nilsvu: Ah ok, this should definitely be in a comment. Doesn't matter though if you'll refactor it with #5184 anyway.

nilsvu (approved):

PR #5153, opened 2023-07-07
Add user guide for TensorExpressions

macedo22:
## Proposed changes

This adds a user guide to our documentation on how to write and use `TensorExpression`s to evaluate tensor equations

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/DevGuide/TensorExpressions.md
+++ b/docs/DevGuide/TensorExpressions.md
@@ -1,0 +1,434 @@
+\cond NEVER
+Distributed under the MIT License.
+See LICENSE.txt for details.
+\endcond
+# Writing Tensor Equations with TensorExpressions {#writing_tensorexpressions}
+
+\tableofcontents
+
+SpECTRE's `TensorExpression`s interface allows you to write tensor equations in
+SpECTRE in C++ with syntax that resembles tensor index notation. To use it,
+simply add this include to the top of your file:
+```
+#include "DataStructures/Tensor/Tensor.hpp"
+```
+The following guide assumes a basic understanding of the `Tensor` class and
+\ref tnsr "tnsr" type aliases. **RHS** refers to the right hand side expression
+that we wish to compute and **LHS** refers to the resulting left hand side
+tensor that stores the result of computing the RHS expression.
+
+# Syntax {#te_syntax}
+`TensorExpression`s are arithmetic expressions of `Tensor`s that can be
+evaluated using `tenex::evaluate`. Terms used in the expression may be `Tensor`s
+or numbers (see [supported types](#te_data_type_support)).
+
+As a simple example of how `TensorExpression`s are used, if you would like to
+raise the index of some `Tensor` `R` with some inverse spacetime metric `Tensor`
+`g`, i.e. \f$R^c{}_b = R_{ab} g^{ac}\f$, you can compute this with
+`TensorExpression`s by doing:
+
+\snippet Expressions/Test_Examples.cpp te_example_evaluate_lhs_return
+
+where `R_up`, `R`, and `g` are rank 2 spacetime `Tensor`s and the `ti::*`
+variables are `TensorIndex`s representing generic tensor indices. Here is a
+breakdown of the different parts of this line:
+- the RHS expression to compute is the argument to
+\ref tenex::evaluate "evaluate": `R(ti::a, ti::b) * g(ti::A, ti::C)`
+- the result LHS `Tensor` is `R_up`
+- the LHS `Tensor`'s indices are the template arguments to
+\ref tenex::evaluate "evaluate": `ti::C, ti::b`
+
+The LHS \ref ::Symmetry "Symmetry" will be deduced from the RHS tensors'
+symmetries and order of operations.
+
+Alternatively, if you already have a LHS `Tensor` variable, you can pass it into
+the following \ref tenex::evaluate "evaluate" overload, where the LHS `Tensor`
+provided will be assigned to the result of the RHS expression:
+
+\snippet Expressions/Test_Examples.cpp te_example_evaluate_lhs_arg
+
+Note that to use this \ref tenex::evaluate "evaluate" overload, the LHS `Tensor`
+does not need to be previously sized unless the data type is a Blaze vector type
+(e.g. `DataVector`) *and* the RHS expression contains no `Tensor` terms
+(see [example](#te_assigning_to_a_number) where sizing is necessary). This
+overload is useful in a couple cases:
+- [Specifying the LHS symmetry](#te_specify_lhs_symmetry): One advantage of this
+overload is that it uses the \ref ::Symmetry "Symmetry" of the provided LHS
+tensor instead of deducing it from the RHS expression. This enables you to
+specify the LHS symmetry in cases where the previous
+\ref tenex::evaluate "evaluate" overload does not deduce the one you want. While
+the LHS index order in this example (`ti::C, ti::b`) could theoretically be
+deduced from the index types of `R_up`, we still require specifying them because
+this isn't the case for all equations and we would like to have a unified
+interface. See [this example](#te_specify_lhs_symmetry), which demonstrates a
+case where you might want to specify the LHS symmetry and where the LHS index
+order would not be deducible.
+- **[Using spatial and time indices on LHS spacetime indices](#te_spatial_time_index_lhs)**
+
+## Tensor indices {#te_tensor_indices}
+
+`TensorIndex`s represent generic tensor indices and are supplied as
+comma-separated lists in two places:
+- in parentheses for each tensor in the RHS expression
+- in the template parameters of \ref tenex::evaluate "evaluate" to specify the
+order of the LHS result tensor's indices.
+
+Each `TensorIndex` takes the form `ti::*` where `*` is a letter that encodes
+index properties:
+- Uppercase letters denote upper indices and lowercase letters denote lower
+indices
+- Letters `A/a - H/h` indicate spacetime indices, `I/i - N/n` indicate spatial
+indices, and `T/t` indicates a concrete time index. This is what is currently
+defined, but more spatial and spacetime indices (letters) can easily be added if
+needed. Note that there is no precedence or difference between the indices of
+some type, e.g. `ti::a`, `ti::b`, ... `ti::h` are equivalent
+
+The properties of each `TensorIndex` and the `Tensor`'s indices (typelist of
+\ref SpacetimeIndex "TensorIndexType"s) must be compatible:
+- valences (being upper or lower indices) must match
+- if a `Tensor`'s index is spacetime, you can use a spacetime `TensorIndex`,
+spatial `TensorIndex`, or concrete time `TensorIndex`
+- if a `Tensor`'s index is spatial, you must use a spatial `TensorIndex`
+
+To demonstrate correct and incorrect usage, let's say we have tensors
+\f$R_{ab}\f$ (two spacetime indices, e.g. type \ref tnsr "tnsr::ab") and
+\f$S_{ij}\f$ (two spatial indices, e.g. type \ref tnsr "tnsr::ij"):
+
+```
+R(ti::c, ti::d) // OK
+R(ti::c, ti::k) // OK, can use spatial TensorIndex on spacetime index
+R(ti::c, ti::t) // OK, can use time TensorIndex on spacetime index
+R(ti::c, ti::D) // ERROR, ti::D is upper but the 2nd index is lower
+S(ti::j, ti::k) // OK
+S(ti::a, ti::k) // ERROR, can't use spacetime TensorIndex on a spatial index
+S(ti::j, ti::t) // ERROR, can't use time TensorIndex on a spatial index
+```
+
+# Examples {#te_examples}
+## Basic operations {#te_basic_operations}
+
+In the following examples:
+- `R` is type \ref tnsr "tnsr::ab<DataVector, 3>"
+- `S` is type \ref tnsr "tnsr::ab<DataVector, 3>"
+- `T` is type \ref Scalar "Scalar<DataVector>"
+- `U` is type \ref tnsr "tnsr::Ab<DataVector, 3>"
+- `V` is type \ref tnsr "tnsr::aBC<DataVector, 3>"
+- `G` is type \ref tnsr "tnsr::a<DataVector, 3>"
+- `H` is type \ref tnsr "tnsr::A<DataVector, 3>"
+
+### Addition and subtraction {#te_addition_and_subtraction}
+\f$L_{ab} = R_{ab} + S_{ba}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_addition
+
+\f$L = 1 - T\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_subtraction
+
+### Contraction of a single tensor {#te_contraction}
+\f$L = U^{a}{}_{a}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_contraction_to_scalar
+
+\f$L^b = V_{a}{}^{ba}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_contraction_to_tensor
+
+### Inner and outer products {#te_products}
+\f$L = G_a H^{a}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_inner_product
+
+\f$L_{cb} = T G_a G_c U^{a}{}_{b}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_inner_and_outer_product
+
+### Division {#te_division}
+\f$L_a = \frac{G_a}{2}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_division_by_number
+
+\f$L_{ba} = \frac{R_{ab}}{T}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_division_by_tensor
+
+\f$L = \frac{5}{U^{a}{}_{a} + 1}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_division_by_tensor_expression
+
+### Square root {#te_square_root}
+\f$L = \sqrt{T}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_square_root_tensor
+
+\f$L = \sqrt{G_a H^a}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_square_root_inner_product
+
+## More features {#te_more_features}
+
+### Specifying the LHS symmetry {#te_specify_lhs_symmetry}
+When using the \ref tenex::evaluate "evaluate" overload that returns the LHS
+`Tensor`, the \ref ::Symmetry "Symmetry" of the LHS `Tensor` will be deduced
+from the RHS expression. However, in some cases the deduced LHS symmetry may
+not be what you want. To specify it yourself, you can pass your LHS `Tensor`
+(that has the desired \ref ::Symmetry "Symmetry") to the
+\ref tenex::evaluate "evaluate" overload that takes the LHS `Tensor` as the
+first argument.
+
+For example, if we have \f$L_{ab} = R_a R_b\f$, the indices of \f$L\f$ are
+symmetric. However, when we do:
+
+\snippet Expressions/Test_Examples.cpp te_example_deduced_lhs_symmetry_fail
+
+the type of `L` will be \ref tnsr "tnsr::ab" because it is not known at
+compile time that the two vectors in the product are the same. To override the
+deduced symmetry and make it a symmetric result, we can create a
+\ref tnsr "tnsr::aa" and pass it into the other overload:
+
+\snippet Expressions/Test_Examples.cpp te_example_deduced_lhs_symmetry_force
+
+### Assigning to a number {#te_assigning_to_a_number}
+You can assign a number (e.g. `double`) to a `Tensor` of any rank to fill all
+components with that value, e.g. \f$L_{ab} = -1\f$. How you do that is slightly
+different depending on the underlying data type of your `Tensor`:
+
+- When your `Tensor`'s data type is a number type (e.g. `double`,
+`std::complex<double>`):
+
+\snippet Expressions/Test_Examples.cpp te_example_assign_number_to_tensor_of_numbers
+
+- When your `Tensor`'s data type is a Blaze vector type (e.g. `DataVector`,
+`ComplexDataVector`), the `Tensor` must first be sized before calling
+\ref tenex::evaluate "evaluate" because there is no sizing information (from a
+`Tensor` component) in the RHS expression:
+
+\snippet Expressions/Test_Examples.cpp te_example_assign_number_to_tensor_of_vectors
+
+See [supported number types](#te_data_type_support) for the data types that the
+RHS number can be.
+
+### Using spatial and time indices on RHS spacetime indices {#te_spatial_time_index_rhs}
+If a `Tensor` has spacetime indices, you can use generic spatial indices and
+concrete time indices to refer to a subset of the components, as we see in
+literature.
+
+Lapse \f$\alpha\f$ computed from the spacetime metric \f$g_{ab}\f$ and shift
+\f$\beta^i\f$:
+
+\f$\alpha = \sqrt{\beta^i g_{it} - g_{tt}}\f$
+
+\snippet Expressions/Test_Examples.cpp te_example_rhs_spatial_and_time_indices
+
+### Using spatial and time indices on LHS spacetime indices {#te_spatial_time_index_lhs}
+Related to the previous example, you can also use generic spatial indices and
+concrete time indices for the spacetime indices of the LHS `Tensor` to assign
+subsets of the LHS `Tensor`'s components.
+
+Spacetime metric \f$g_{ab}\f$ computed from the lapse \f$\alpha\f$, shift
+\f$\beta^i\f$, and spatial metric \f$\gamma_{ij}\f$:
+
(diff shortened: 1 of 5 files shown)


docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+The following guide assumes a basic understanding of the `Tensor` class and
+\ref tnsr "tnsr" type aliases.
+
+# Syntax {#te_syntax}
+`TensorExpression`s are arithmetic expressions of `Tensor`s that can be
+evaluated using `tenex::evaluate`. Terms used in the expression may be `Tensor`s
+or numbers (see [supported types](#te_data_type_support)).
+
+As a simple example of how `TensorExpression`s are used, if you would like to
+raise the index of some `Tensor` `R` with some spacetime metric `Tensor` `g`,
```
AlexCarpenter46: with the inverse spacetime metric `Tensor` `g`

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+    make_not_null(&spacetime_metric), spatial_metric(ti::i, ti::j));
+```
+
+### Using the LHS Tensor in the RHS expression {#te_using_lhs_tensor_in_rhs}
+You can use the LHS `Tensor` in the RHS expression to emulate operations like
+`+=`, `*=`, etc. For example, say you would like to emulate the following:
+```
+// pseudocode
+L_ab = R_ab
+L_ab += + 2.0 * S_ba
```
AlexCarpenter46: L_ab += 2.0 * S_ba

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+```
+You can do the operation in the 2nd line above by calling
```
AlexCarpenter46: You can emulate the += operation by calling ...

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+- `ComplexDataVector`
+
+Support for more types can be added.
+
+## Operations {#te_operation_support}
+It's possible for terms in the expression to have different data types. The
+following table shows the data type that results from performing a binary
+operation (`+`, `-`, `*`, `/`) between two terms of given data types:
+
+<table>
```
AlexCarpenter46, line 352: It could be that the vscode preview is a little off, but I'm a bit confused about the labels on the table. If this is supposed to show the resulting type when doing operations on allowed types, aren't the outside operand types both RHS and the resulting type in the middle of the table the LHS? Here's a png of my preview, so it might just be that that's off. TE_Preview

AlexCarpenter46:
Just some general comments and wording, but I really like the comprehensiveness of this guide 😄

docs/DevGuide/TensorExpressions.md
macedo22: For me it renders differently when I open the documentation locally in a browser, so I would try that! This is what the table looks like for me in a browser (and it has horizontal scroll):

image

macedo22: and I'm actually just now realizing the LHS and RHS operand type row and column headers should not include the first row and column that they do and I'll fix that - what I mean is like "RHS operand type" header should only go as far left as the `double` column header, not also the column before it that includes the LHS operand types. So yah, will fix that

AlexCarpenter46: Ah okay cool, I think I was just confused on what LHS and RHS meant here, as I assumed that something like the resulting type ex) auto L = tenex::evaluate<ti::a, ti::b> would be the LHS and the RHS operations were things like what you're multiplying to produce the LHS tensor.

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+
+\tableofcontents
+
+SpECTRE's `TensorExpression`s interface allows you to write tensor equations in
+SpECTRE in C++ with syntax that resembles tensor index notation. To use it,
+simply add this include to the top of your file:
+```
+#include "DataStructures/Tensor/Tensor.hpp"
+```
+The following guide assumes a basic understanding of the `Tensor` class and
```
knelli2, line 15: Can you link this `Tensor` to the actual class? The link doesn't show up on web. Since you say in this sentence that this guide assumes you have some base knowledge of `Tensor` id' be good to have a link to it like you do to `tnsr`. (The other `Tensor` in this guide are fine as is)

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+```
+tenex::evaluate<ti::C, ti::b>(
+    make_not_null(&R_up), R(ti::a, ti::b) * g(ti::A, ti::C));
+```
+
+One advantage of this overload is that it uses the symmetry and index types
+(spatial/spacetime) of the provided LHS tensor instead of deducing them from the
+RHS expression, so this enables you to specify the LHS index structure in cases
+where the previous \ref tenex::evaluate "evaluate" overload does not deduce the
+one you want. Note that the LHS tensor does not need to be previously sized
```
knelli2: When would this occur? Is it worth providing an example?

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+- Uppercase letters denote upper indices and lowercase letters denote lower
+indices
+- Letters `A/a - H/h` indicate spacetime indices, `I/i - N/n` indicate spatial
+indices, and `T/t` indicates a concrete time index. This is what is currently
+defined, but more spatial and spacetime indices (letters) can easily be added if
+needed
+
+The properties of each `TensorIndex` and the `Tensor`'s indices (typelist of
+\ref SpacetimeIndex "TensorIndexType"s) must be compatible:
+- valences must match
```
knelli2: I totally should have known what the valence of a tensor was, but I had to look it up. Could you add a small note in parentheses like `valences (number of upper/lower indices) must match`

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+
+// ERROR: LHS and RHS indices don't match
+auto result1 = tenex::evaluate<ti::a, ti::c>(R(ti::a, ti::b) + S(ti:a::ti::b));
+// ERROR: Can't add Tensors with different indices
+auto result2 = tenex::evaluate<ti::a, ti::b>(R(ti::a, ti::b) + S(ti:a::ti:c));
+// ERROR: Repeated index in the RHS
+auto result3 =
+    tenex::evaluate<ti::a, ti::b, ti::c>(R(ti::a, ti::b) * S(ti:a::ti::c));
+// ERROR: Can't add Tensors with different Frame types
+auto result4 = tenex::evaluate<ti::a, ti::b>(R(ti::a, ti::b) + T(ti:a::ti::b));
```
knelli2: I believe the `S` and `T` tensors are missing some commas and have some semi-colons in the wrong place.

docs/DevGuide/TensorExpressions.md
knelli2: Yeah I think fixing the column headers would be clearer.

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+`TensorExpression`s by doing:
+```
+auto R_up = tenex::evaluate<ti::C, ti::b>(R(ti::a, ti::b) * g(ti::A, ti::C));
+```
+
+where `R_up`, `R`, and `g` are rank 2 spacetime `Tensor`s and the `ti::*`
+variables are `TensorIndex`s. Here, the argument to
+\ref tenex::evaluate "evaluate" is the RHS tensor expression to compute, `R_up`
+is the result LHS `Tensor`, and the template arguments to
+\ref tenex::evaluate "evaluate" are the LHS `Tensor`'s indices. The LHS symmetry
```
knelli2: Could you make this a bulleted list instead?

- The RHS tensor is the function argument to \ref tenex::evaluate "evaluate": `R(ti::a, ti::b) * g(ti::A, ti::C)`
- The LHS tensor is `R_up`
- The LHS tensor indices are the template areguments `<ti::C, ti::b>`

These are important terms you use throughout the guide and I think they should be stand out somehow.

docs/DevGuide/TensorExpressions.md
knelli2: This sentence is a bit long. Can you split it up at the `RHS expression, so this...` -> `RHS expression. This...`

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+provided will be assigned to the result of the RHS expression:
+
+```
+tenex::evaluate<ti::C, ti::b>(
+    make_not_null(&R_up), R(ti::a, ti::b) * g(ti::A, ti::C));
+```
+
+One advantage of this overload is that it uses the symmetry and index types
+(spatial/spacetime) of the provided LHS tensor instead of deducing them from the
+RHS expression, so this enables you to specify the LHS index structure in cases
```
knelli2: Why do you still need the `ti::C, ti::b` template parameters if the symmetries and index types of the LHS tensor are used?

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+where the previous \ref tenex::evaluate "evaluate" overload does not deduce the
+one you want. Note that the LHS tensor does not need to be previously sized
+unless the data type is a Blaze vector type (e.g. `DataVector`) *and* the RHS
+expression contains no `Tensor` terms (see [example](#te_assigning_to_a_number)
+where sizing is necessary). This overload is also used to
+[assign subsets of tensor components](#te_assigning_subsets_of_components).
+
+## Tensor indices (TensorIndexs) {#te_tensor_indices}
```
knelli2: I think you can get rid of `(TensorIndexs)`. The plural is pretty clear on its own

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+
+`TensorIndex`s represent generic tensor indices and are supplied as
+comma-separated lists in two places: (1) in parentheses for each tensor in the
+RHS expression and (2) in the template parameters of
```
knelli2: Can this be an actual numbered list like the bullets below? Draws the eyes to it more.

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+index properties:
+- Uppercase letters denote upper indices and lowercase letters denote lower
+indices
+- Letters `A/a - H/h` indicate spacetime indices, `I/i - N/n` indicate spatial
+indices, and `T/t` indicates a concrete time index. This is what is currently
+defined, but more spatial and spacetime indices (letters) can easily be added if
+needed
+
+The properties of each `TensorIndex` and the `Tensor`'s indices (typelist of
+\ref SpacetimeIndex "TensorIndexType"s) must be compatible:
```
knelli2, line 87: Is there anything special about certain `tensorIndex`s (other than `T/t`)? Like say I want to contract a `tnsr::ab`. Do I have to do `tenex::evaluate(R(ti::A, ti::a))`? Or can I do `tenex::evaluate(R(ti::H, ti::h))`? Similarly with result tensors: `auto L = tenex::evaluate(R(ti::a, ti::b) + S(ti::a, ti::b))` vs `auto L = tenex::evaluate(R(ti::e, ti::f) + S(ti::e, ti::f))`?

If we have to use `A/a` first, that should be noted. If not, say that it doesn't matter, but we still tend to use `A/a` and `I/i` first and then move down the alphabet.

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+```
+\f$L = \sqrt{G_a H^a}\f$
+```
+auto L = tenex::evaluate(sqrt(G(ti::a) * H(ti::A)));
+```
+
+## More features {#te_more_features}
+
+### Assigning to a number {#te_assigning_to_a_number}
+You can assign a number (e.g. `double`) to a `Tensor` of any rank to fill all
```
knelli2, line 192: `Tensor<double>` so it's clear. I know you say it below, but since a reader sees this first, I think it's best to be explicit here.

docs/DevGuide/TensorExpressions.md
knelli2: I think this example is a little confusing because of the swapped indices on $S_{ba}$.  Below you say that the LHS tensor must have the same index ordering as the RHS. In the "good" example you have `L(ti::a, ti::b) + 2.0 * S(ti::b, ti::a)` which have opposite ordering and in the "bad" example you have `L(ti::b, ti::a) + 2.0 * S(ti::b, ti::a)` which have the same ordering.

I think making S be $S_{ab}$ would be clearer.

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+Here are some examples of illegal math that the compiler will catch:
+
+```
+tnsr::ab<double, 3, Frame::Inertial> R{};
+tnsr::ab<double, 3, Frame::Inertial> S{};
+tnsr::ab<double, 3, Frame::Grid> T{};
+tnsr::AB<double, 2, Frame::Inertial> G{};
+
+// ERROR: LHS and RHS indices don't match
+auto result1 = tenex::evaluate<ti::a, ti::c>(R(ti::a, ti::b) + S(ti:a::ti::b));
```
knelli2: Could you add a blank line between the result line and the next "ERROR" comment?

knelli2 (dismissed):
Overall really nice guide! Thank you for adding this!

docs/DevGuide/TensorExpressions.md
macedo22: Yes, so I would love to. I tried many times to get it to link, but haven't been able to. I also looked elsewhere in spectre to find it successfully linked and I haven't found an instance yet, so I'm not sure how to get doxygen to link it. I think maybe the linking fails because the declaration of the class is not documented (not sure), and I can't get it to link to the defintiion, which is a partial template specialization and not sure how to tell doxygen what I want.

So if you happen to know how I could do this, that would be great, and then I think that needs to be done in a bunch of places throughout the site. Otherwise, I can keep trying things and poking around doxygen docs/issues

macedo22: e.g. I can try adding documentation tot he declaration of the `Tensor` class to see if that fixes it, unless you already know a solution

docs/DevGuide/TensorExpressions.md
macedo22: If would occur in something as simple as this:
```
tnsr::i<double, 3> R{};
auto L = tenex::evaluate<ti::a, ti::b>(R(ti::a) * R(ti::b));
```
because the compiler (result symmetry is computed at compile time) doesn't know that the two operands to `*` are the same `Tensor`. `L` would be deduced as `tnsr::ij`, not `tnsr::ii`.

I think it's useful to demonstrate this, and the only reason I didn't is because when I was writing this, I was trying to present meat of how to write an equation at the top of the file without writing lots of details and getting into the weeds until down below. Like "here is what it looks like, now let me tell you more below." But you're someone who experienced reading it so, do you think it would be best to include an example like the one I showed up at the top in the `Syntax` section where I'm introducing the two `evaluate` overloads, or do you think it's best to allude to it and link it to an example further down like I did with other things?

docs/DevGuide/TensorExpressions.md
macedo22: To specify the order of the LHS indices. You may want `L^c{}_b` or you may want `L_b{}^c`. If you think a note should be added, let me know

macedo22: To specify the order of the LHS indices. You may want `L^c{}_b` or you may want `L_b{}^c`. In this case, the index types of `b` and `c` are different, so yes at compile time, the order could be assigned for you, but this isn't the case if you have LHS indices of the same type. For example:

```
tnsr::ab<...> L{};
tnsr::ab<...> R{};
tnsr::ab<...> S{};

// this wouldn't compile, just what it would look like without the LHS index tparams
tenex::evaluate(make_not_null(&L), R(ti::a, ti::b) + S(ti::a, ti::b));
```
You might want the LHS to be `L_{ab}` or maybe you want it `L_ba`, so the `evaluate` tparams let you specify that order. And while the order could be inferred in cases where the index types are different, that handling isn't something I've implemented and I think the way it is now also means there is a unified interface / one way you do things and forces people to be explicit

macedo22: The point of the LHS tparams is to specify the order of the LHS indices. In this case, the index types of `b` and `c` are different, so yes at compile time, the order could in theory be deduced for you from the LHS tensor's index types, but this isn't the case if you have LHS indices of the same type. For example:

```
tnsr::ab<...> L{};
tnsr::ab<...> R{};
tnsr::ab<...> S{};

// this wouldn't compile right now, this is just what it would look like without the LHS index tparams
tenex::evaluate(make_not_null(&L), R(ti::a, ti::b) + S(ti::a, ti::b));
```
You might want the LHS to be `L_{ab}` or maybe you want it `L_ba`, so the `evaluate` tparams let you specify that order. And while the order could be inferred in cases where the index types are different (like the raising index example), doing that deduction isn't something I've implemented and I think the way it is now also means there is a unified interface / one way you do things and forces people to be explicit

So yes, the LHS index order of this example could be deduced fromt he input LHS `Tensor`, but I don't think we want to

docs/DevGuide/TensorExpressions.md
macedo22: No, `T/t` is the only special one. `a`, `b`, `c`, ..., `h` are equivalent and it doesn't matter what you use. I used a lot of `a` and `b` just because, but if you think the guide made it seem like you had to use them with a certain precedence, then I can either add a short line somewhere saying it doesn't matter or I can vary the letters I use throughout the guide. Let me know which you prefer!

docs/DevGuide/TensorExpressions.md
macedo22: Ah, so this example shows you that you can do this for `Tensor<DataType>` though (kinda) in that it shows you how to do it for a `Tensor<double>` and then for a `Tensor<DataVector>`. It was meant to be like "here's what the basic syntax looks like for assigning to a number (mainy to demonstrate that you have to use the evaluate that takes int he LHS tensor).... and here is what you have to do / the caveat if you have a blaze vector type." But it does seem this was not clear, so I think I will change the wording to make it sound more like what I just said here so that the point being demonstrated is more clear

docs/DevGuide/TensorExpressions.md
macedo22: This was a very useful comment to me because the constraint here is only that `L` must have the same ordering on both sides - other tensors' indices can be doing whatever they want, and I think that's precisely why I left `S` with the opposite order, to demonstrate that `L` is the only one that needs to match the LHS. I see now how the sentence makes it seem like the whole RHS needs to have some ordering, so I'll fix the wording to make it more clear that it's only for `L` and leave the ordering of `S`, if that works for you

docs/DevGuide/TensorExpressions.md
macedo22: Oh sorry, reread your comment and it answered my question - I'll just say explicitly that they are equal and there is no precedence

docs/DevGuide/TensorExpressions.md
macedo22: Because the example wasn't even meant to demonstrate `Tensor<double>` specifically - I could have used `Tensor<std::complex<double>>` and done a one-liner to demonstrate the syntax, so the `Tensor` without tparams was purposeful

docs/DevGuide/TensorExpressions.md
knelli2: I just tried moving the docs to the declaration and that fixed everything so it's all linked now. So I'd say just add a new commit before this one where you move the docs.

macedo22: Thanks! This is also useful to know for other missing links with the same problem and this should fix the linking to `Tensor` everywhere in spectre docs hopefullly

docs/DevGuide/TensorExpressions.md
knelli2: I'd say link to an example further down. I like your approach of using simple examples at the top to just demonstrate things and then having more detailed examples below.

docs/DevGuide/TensorExpressions.md
knelli2: Sounds good! Thanks for explaining! Could you add a note here then saying something like

```
Even though we use the LHS tensors index types rather than deducing them from the
RHS, for technical reasons, you must always specify the tparams to `evaluate` for non-Scalar
tensors.
```

docs/DevGuide/TensorExpressions.md
knelli2: Ah yeah I think explicitly splitting up the two cases would be good in your explanation.

docs/DevGuide/TensorExpressions.md
knelli2: As long as it's clear, that's good with me

docs/DevGuide/TensorExpressions.md
```diff
@@ -0,0 +1,411 @@
...
+`TensorExpression`s are arithmetic expressions of `Tensor`s that can be
+evaluated using `tenex::evaluate`. Terms used in the expression may be `Tensor`s
+or numbers (see [supported types](#te_data_type_support)).
+
+As a simple example of how `TensorExpression`s are used, if you would like to
+raise the index of some `Tensor` `R` with some spacetime metric `Tensor` `g`,
+i.e. \f$R^c{}_b = R_{ab} g^{ac}\f$, you can compute this with
+`TensorExpression`s by doing:
+```
+auto R_up = tenex::evaluate<ti::C, ti::b>(R(ti::a, ti::b) * g(ti::A, ti::C));
```
nilsvu: What do you think about making these code examples snippets from a compiled cpp file? Then they're guaranteed to work. It means going through and actually moving them to a cpp file, so I leave it up to you to decide if it's worth the effort.

macedo22: Sure, I can do this! Would it be reasonable to make a new test file specifically dedicated to this guide? That way there is one spot that all of the demonstration examples are?

macedo22: Because right now, the tests are separated by operation

nilsvu: Yea sounds good! The Brigand documentation does that too.

knelli2: We have example files for other things already that are just used to compile and be pretty docs things:

- `tests/Unit/Helpers/ParallelAlgorithms/Interpolation/Examples.hpp`
- `tests/Unit/Helpers/ControlSystem/Examples.hpp`

knelli2:
@macedo22  Since it's been a while, if you'd like to have these changes merged into spectre, please comment here or let the core-devs know. If you would no longer like to have these changes merged, or we don't hear from you by next Wed Aug 28 1pm PDT, we will close this PR. If you change your mind in the future, we can always re-open the PR.

macedo22:
@knelli2 I'd like this PR to be merged but the fixups aren't ready. This is at the top of my TO-DO list but I'm not sure when specifically I will have them in. If you prefer to close this to clear out clutter, then I can reopen it, but this is a PR that I would like merged once I have fixups in

knelli2:
Ok thanks! We'll keep it open then :)

macedo22:
I've gone ahead and rebased and added the requested changes!

tests/Unit/DataStructures/Tensor/Expressions/Test_Examples.cpp
```diff
@@ -0,0 +1,482 @@
...
+
+  // some examples below are forced to be more than one line against the wishes
+  // of clang-tidy because the doxygen \snippet function can't render
+  // single-lined code snippets
+
+  // addition
+  {
+    // [te_example_addition]
+    auto L =
+        tenex::evaluate<ti::a, ti::b>(R(ti::a, ti::b) + S(ti::b, ti::a));
```
macedo22, line 95: I personally don't know of a way to tell `clang-tidy` to ignore this line without `NOLINT`, etc. visually polluting the example snippet. If there is a way, let me know!

I purposely made examples like this into more than one line (ignored `clang-tidy`) because Doxygen's `\snippet` won't render single-lined code snippets (at least in `1.10.0` based on `mbot_install.sh`), so this is a temporary workaround. This bug may have been fixed in later Doxygen versions (see [issue](https://github.com/doxygen/doxygen/issues/10517) that seems like the behavior I saw, where no code is rendered).

macedo22: Oh, guess `clang-tidy` in the CI was fine with it? Guess this thread can be disregarded

wthrowe: Just for future reference: At some point (not super recent, but didn't exist when we started SpECTRE) clang-tidy got a region form that can be useful for this sort of thing:
```
// NOLINTBEGIN(...)
code
with
warnings
// NOLINTEND(...)
```

macedo22: Good to know! Thanks!

kidder:
@knelli2 @nilsvu please take a look so we can get this merged

nilsvu (approved):
LGTM

AlexCarpenter46 (approved):
Looks good to me, the guide reads really well and has very useful examples

PR #5155, opened 2023-07-07
Add grid anchors to domain creators

knelli2:
## Proposed changes

For the control system, we need to know the centers of the objects. For BBHs, excisions sufficed to find the centers. However, for BNSs this clearly won't work. But we still need to know what the centers of the objects are in order for the control systems to work. The information about where the centers are is known to the domain creator). Thus, we add a new function to the `DomainCreator` base class called `grid_anchors` which gets a set of `tnsr::I<double, 3, Frame::Grid>` indexed by a string. These grid anchors are special points of the domain in the grid frame that won't move and are potentially needed by other parts of the code (e.g. control systems). They aren't restricted to be objects. They can be any point within the domain.

This also uses the `grid_anchors` to initialize the `ObjectCenters` tag, rather than using excision spheres like it did before. And we clean up a control system tag.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/DomainCreator.hpp
+++ b/src/Domain/Creators/DomainCreator.hpp
@@ -14,10 +14,14 @@
 #include <unordered_set>
 #include <vector>
 
+#include "DataStructures/Tensor/Tensor.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
 #include "Domain/Structure/DirectionMap.hpp"
 
 /// \cond
+namespace Frame {
+struct Grid;
+}  // namespace Frame
 template <size_t>
 class Domain;
 namespace domain::BoundaryConditions {
@@ -46,6 +50,14 @@ class DomainCreator {
   virtual ~DomainCreator() = default;
 
   virtual Domain<VolumeDim> create_domain() const = 0;
+
+  /// A set of named coordinates in the grid frame, like the center of the
+  /// domain or the positions of specific objects in a domain
+  virtual std::unordered_map<std::string,
+                             tnsr::I<double, VolumeDim, Frame::Grid>>
+  grid_anchors() const {
+    return {};
+  }
 
   /// The set of external boundary condition for every block in the domain
   virtual std::vector<DirectionMap<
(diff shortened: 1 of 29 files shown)


src/Domain/Creators/DomainCreator.hpp
```diff
@@ -47,6 +51,14 @@ class DomainCreator {
 
   virtual Domain<VolumeDim> create_domain() const = 0;
 
+  /// A set of coordinates in the grid frame that are fixed and will remain
```
nilsvu: Suggestion: `A set of named coordinates in the grid frame, like the center of the domain or the positions of objects in a binary domain`

("grid frame" already implies "fixed in time")

tests/Unit/Helpers/Domain/Creators/TestHelpers.hpp
```diff
@@ -221,4 +223,20 @@ void test_functions_of_time(
                }
              });
 }
+
+// Only checks that grid anchors exist. Doesn't check what they actually are
+// (because they can be anywhere in the domain)
+template <size_t Dim>
+void test_grid_anchors(const DomainCreator<Dim>& creator,
```
nilsvu: Just delete this and check the grid anchors explicitly like you do for the two BCO domains (same as excision spheres). The other domains can either check nothing (since grid anchors aren't implemented yet) or check the grid anchors are `empty()`.

src/Domain/Creators/Tags/ObjectCenter.hpp
```diff
@@ -24,23 +24,11 @@ struct Grid;
 namespace domain::Tags {
 /// \ingroup DataBoxTagsGroup
 /// \ingroup ComputationalDomainGroup
-/// Base tag to retrieve the grid frame centers of objects in the domain
-/// corresponding to the `ObjectLabel`.
+/// The grid frame center of the given object.
```
nilsvu: Mention in the docs that this requires the domain creator to have a grid anchor named `CenterA/B`

nilsvu (changes requested):

nilsvu (approved):

PR #5161, opened 2023-07-09
Add prefetch function

nilsdeppe:
## Proposed changes

Wraps the compiler builtin prefetch function to make it more user-friendly.

Software prefetching reduced L1 dcache misses in Adams-Bashforth `update_u_common` from ~27% to ~3% on my machine. Since once we start using nodegroups we become completely dominated by cache misses and DRAM accesses, manual prefetching in pieces of the code that take up a lot of runtime will become increasingly important. The compiler builtin is rather ugly to use, so this gives it a more intuitive wrapper. From current profiling the places that using cache-blocking with prefetching will be critical are (in no particular order):
1. RHS (GH especially has like 70% cache misses)
2. time stepping work, both volume and boundary terms
3. boundary corrections
4. partial derivatives when applying the inverse Jacobian
5. transpose operations (SIMD sped this up a lot, but now there are still a lot of cache misses that prefetching could eliminate)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Utilities/System/Prefetch.hpp
+++ b/src/Utilities/System/Prefetch.hpp
@@ -1,0 +1,48 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <iosfwd>
+
+namespace sys {
+/// The cache location to prefetch data to.
+enum class PrefetchTo : int {
+  /// Prefetch into L1 data cache
+  ///
+  /// Typically the fastest CPU cache, about 32 KB in size.
+  L1Cache = 3,
+  /// Prefetch into L2 data cache
+  ///
+  /// Typically the second fastest CPU cache, about 128 or 256 KB in size.
+  L2Cache = 2,
+  /// Prefetch into L3 data cache
+  ///
+  /// Typically the slowest CPU cache and is shared among multiple cores with
+  /// sizes varying by factors of several.
+  L3Cache = 1,
+  /// Non-temporal is an element that is unlikely to be re-used. E.g., read
+  /// once, written but never read.
+  NonTemporal = 0,
+  /// Prefetch to L1 data cache for writing
+  WriteL1Cache = 7,
+  /// Prefetch to L2 data cache for writing
+  WriteL2Cache = 6
+};
+
+/// \brief Prefetch data into a specific level of data cache.
+template <PrefetchTo CacheLocation>
+#if defined(__GNUC__)
+__attribute__((always_inline)) inline
+#endif
+void prefetch(const void* address_to_prefetch) {
+  // The enum values are bit flags where the lowest two bits (right-most)
+  // control the cache level and the 3rd bit controls whether it is a write-only
+  // or read-write operation.
+  __builtin_prefetch(address_to_prefetch,
+                     (static_cast<int>(CacheLocation) >> 2) & 1,
+                     static_cast<int>(CacheLocation) & 0x3);
+}
+
+std::ostream& operator<<(std::ostream& os, PrefetchTo cache_location);
+}  // namespace sys
--- a/src/Utilities/System/CMakeLists.txt
+++ b/src/Utilities/System/CMakeLists.txt
@@ -10,6 +10,7 @@ spectre_target_sources(
   PRIVATE
   Abort.cpp
   ParallelInfo.cpp
+  Prefetch.cpp
   )
 
 spectre_target_headers(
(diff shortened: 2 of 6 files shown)


wthrowe (dismissed):

src/Utilities/System/Prefetch.hpp
```diff
@@ -0,0 +1,38 @@
...
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <iosfwd>
+
+namespace sys {
+/// The cache location to prefetch data to.
+enum class PrefetchTo : int {
+  /// Prefetch into L1 data cache
```
nilsvu, line 11: Could you add to these docs just some basic reminders about the hierarchy and size of these CPU caches? Maybe something like "Prefetch into L1 data cache (typically the fastest CPU cache, about X KB in size)." With a note in the enum dox that these depend on the architecture.

src/Utilities/System/Prefetch.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+/// \brief Prefetch data into a specific level of data cache.
+template <PrefetchTo CacheLocation>
+#if defined(__GNUC__)
+__attribute__((always_inline)) inline
+#endif
+void prefetch(const void* address_to_prefetch) {
+  __builtin_prefetch(address_to_prefetch,
+                     (static_cast<int>(CacheLocation) >> 2) & 1,
+                     static_cast<int>(CacheLocation) & 0x3);
```
nilsvu, line 44: Please add some docs or comments explaining how the `int` enum value is chosen and used. (Would be easier with the separate read/write template param.)

src/Utilities/System/Prefetch.hpp
```diff
@@ -0,0 +1,38 @@
...
+  /// Prefetch to L2 data cache for writing
+  WriteL2Cache = 6
+};
+
+/// \brief Prefetch data into a specific level of data cache.
+template <PrefetchTo CacheLocation>
+#if defined(__GNUC__)
+__attribute__((always_inline)) inline
+#endif
+void prefetch(const void* address_to_prefetch) {
```
nilsvu, line 38: Maybe two template parameters would be easier to understand, corresponding more directly to the read/write and locality arguments of `__builtin_prefetch`? Like `prefetch<Read, L1Cache>`.

nilsvu (changes requested):

src/Utilities/System/Prefetch.hpp
nilsdeppe: No, because then you could have invalid states that I don't know what they would do... The list isn't a tensor product of the cache level and write-only.

src/Utilities/System/Prefetch.hpp
nilsdeppe: I'm not sure what you want me to say here. This is manipulated to it givens generally the write numbers for compilers. I could say something like "Manipulate enum value to pass to compiler" but that's sort of redundant with the code. What are you looking for?

knelli2: I think what @nilsvu is looking for (and what I was wondering), is why are the enum values what they are? Why is L1Cache 3, why is L3Cache 1? A comment on that would be helpful. And then maybe also what the `>> 2` and `& 0x3` actually do? I know they are well defined operations, but they aren't common so a comment explaining what the operation returns would be helpful.

nilsdeppe: I still don't understand, sorry. The values are chosen so that the corresponding bit manipulations are the correct values passed to `__builtin_prefetch` to load something into the correct level of cache. I don't think it's good to re-document `__builtin_prefetch` since GCC (and presumably Clang) document it. I could use something like a switch or if-constexpr but that would, in the case of switch, be bad for performance since that's a runtime lookup, and if-constexpr would create a very long function.

I don't think we should document standard operators like bitwise and right shift in the code.

The point of the wrapper is to hide the details of `__builtin_prefetch`, so I'm not sure why we are focusing on the details of `__builtin_prefetch`. It's not the user-interface. The enum is supposed to be a helpful representation of bit flags. Would it be useful to write the values in the enum in binary format?

nilsdeppe: What about:
```
// The enum values are bit flags where the lowest two bits (right-most) control
// the cache level and the 3rd bit controls whether it is a write-only or read-write operation.
```

nilsvu: Sounds good 👍 That's enough for me

knelli2: Yeah that works for me as well

nilsdeppe:
CI failure is python formatting in files I didn't touch :shrug:

nilsvu (approved):

PR #5162, opened 2023-07-09
Minor cleanups to CombineH5

nilsdeppe:
## Proposed changes

- Fixes a compiler warning about not using an anonymous namespace
- Adds some status printing info, which I found useful because otherwise for large datasets you just sit there hoping something is happening. @aasthabagree21 will rework this to be part of the CLI soon.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Executables/CombineH5/CombineH5.cpp
+++ b/src/Executables/CombineH5/CombineH5.cpp
@@ -3,6 +3,7 @@
 
 #include <boost/program_options.hpp>
 #include <cstddef>
+#include <cstdlib>
 #include <iterator>
 #include <string>
 #include <vector>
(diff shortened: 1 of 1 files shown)


src/Executables/CombineH5/CombineH5.cpp
```diff
@@ -51,7 +55,9 @@ void combine_h5(const std::string& file_prefix, const std::string& subfile_name,
                 const std::string& output) {
   // Parses for and stores all input files to be looped over
   const std::vector<std::string>& file_names =
-      file_system::glob(file_prefix + "*.h5");
+      file_system::glob(file_prefix + "[0-9]*.h5");
+  Parallel::printf("Processing files:\n%s\n",
+                   std::string{MakeString{} << file_names}.c_str());
```
knelli2, line 60: I think this is ok because we'll be replacing it with a CLI soon where the user can determine how much they want printed.

knelli2 (approved):

src/Executables/CombineH5/CombineH5.cpp
nilsdeppe: Yea, @aasthabagree21 is going to clean this up a bit more for the CLI :)

PR #5163, opened 2023-07-10
Increase max spectral basis functions from 12 to 20

mar-celine:
## Proposed changes

Increases the spectral limit from 12 collocation points to 20.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


wthrowe:
I'm not objecting, but what's the use case?

knelli2:
This was mostly motivated by @nikwit trying to optimize the BBH domain and finding that we could decrease h-refinement and increase p-refinement and things would run faster. He increased p-ref above 12 and that seemed to work well. And 12 is just an arbitrary number so 20 is now just a higher arbitrary number so we can do some BBH runs. And all of @geoffrey4444's latest BBH runs for the past month or more have used p-ref higher than 12.

wthrowe:
OK, that's fine.  You need to follow the instructions seven lines above your change, though.

mar-celine:
Some test failures:

`Unit.Ader.Dg.Matrices`  by segfault - the `inverse_temporal_matrix` matrices are tested against hard coded matrices computed via a python implementation, but they are only hard coded up to NxN = 12x12 matrices. Someone will need to compute them up to 20x20.

`Unit.Evolution.Subcell.Tci.Rdmp` fails with an index out of range, for some reason `num_pts_1d=13` sets off an error. Here is some relevant code:
```
  // We lower the maximum number of 1d points in 3d in order to reduce total
  // test runtime.
  const size_t maximum_number_of_points_1d =
      Dim == 3 ? 7
               : Spectral::maximum_number_of_points<Spectral::Basis::Legendre>;
  for (size_t num_pts_1d = 2; num_pts_1d < maximum_number_of_points_1d;
       ++num_pts_1d) {
    test_rdmp_impl<Dim>(past_max_values, past_min_values, 1.0e-4, 1.0e-3,
                        num_pts_1d, 1.0, 1.0, false);

```

It is the 1d test that is failing.

`Unit.Evolution.Subcell.Tci.TwoMeshRdmp` is also failing, possibly for a similar reason.

`Unit.Numerical.LinearOperators.PartialDerivs` is also failing, with errors like
```
/__w/spectre/spectre/tests/Unit/NumericalAlgorithms/LinearOperators/Test_PartialDerivatives.cpp:410: FAILED:
  CHECK( du.data()[n] == approx(expected_du.data()[n]).epsilon(1.e-13) )
with expansion:
  0.00000000000011369 == Approx( 0.0 )
```

Should this tolerance be relaxed so that the tests pass?

Thanks!

wthrowe:
I'm guessing the subcell-related ones are for the reason I pointed out in my last comment.  You can't increase the DG limit without also increasing the FD limit.

nilsdeppe:
> I'm guessing the subcell-related ones are for the reason I pointed out in my last comment. You can't increase the DG limit without also increasing the FD limit.

This! They are related.

12 is less arbitrary than you might think. A reasonable argument for such an upper limit is that `1/2^12~2e-4` so the errors drop by nearly 4 orders of magnitude by 1 h-refinement. This is insanely rapid convergence and generally means you should reach machine precision extremely rapidly. `1/2^20 ~ 1e-8` which means one h-refinement should decrease the error by halfway to machine precision, which is overly aggressive. Normally you hit the convergent regime with an error somewhere around 1e-4 to 1e-6, then 1 h-refinement would take you to 1e-12 or 1e-14 in error(!!!!!). 16 would be a drop of `1e-5` per h-refinement, which is quite aggressive but doesn't seem unreasonably so. The point is if you need 20 points to resolve the solution then it is not locally well-approximated by a polynomial, and so h-refining will make the solution more local and so more likely to look like a polynomial.

knelli2:
@mar-celine is going to take another look at this. For the ADER tests, we'll just separately limit to 12.

mar-celine:
@knelli2 considering the only remaining test failures are timeouts, I hope this PR can start being looked at / reviewed again!

tests/Unit/Helpers/PointwiseFunctions/AnalyticSolutions/FirstOrderEllipticSolutionsTestHelpers.hpp
```diff
@@ -221,11 +221,13 @@ void verify_smooth_solution(
     const double tolerance_offset, const double tolerance_scaling,
     PackageFluxesArgs&& package_fluxes_args) {
   INFO("Verify smooth solution");
+  const size_t max_points =
+      (Spectral::maximum_number_of_points<Spectral::Basis::Legendre> >= 12)
```
nilsdeppe: I think this is just `std::min(Spectral::maximum_number_of_points<Spectral::Basis::Legendre>, 12)` (maybe `12_st`), similar elsewhere

tests/Unit/NumericalAlgorithms/LinearOperators/Test_CoefficientTransforms.cpp
```diff
@@ -89,7 +89,8 @@ void check_transforms(
 
   // Back to nodal coefficients, which should match what we set up initially
   const NodalVectorType u_nodal = to_nodal_coefficients(u_modal, mesh);
-  CHECK_ITERABLE_APPROX(u_nodal_expected, u_nodal);
+  Approx local_approx = Approx::custom().epsilon(1e-12).scale(1.0);
```
nilsdeppe, line 95: do we need the scale set to 1.0 or can we just do 1e-12?

tests/Unit/NumericalAlgorithms/LinearOperators/Test_PartialDerivatives.cpp
```diff
@@ -500,6 +501,9 @@ SPECTRE_TEST_CASE("Unit.Numerical.LinearOperators.LogicalDerivs",
   constexpr size_t max_points =
       Spectral::maximum_number_of_points<Spectral::Basis::Legendre> / 2;
   for (size_t n0 = min_points; n0 <= max_points; ++n0) {
+    if (n0 > 6 and n0 != max_points) {
+      continue;
```
nilsdeppe, line 506: Maybe add a comment `// To keep test time reasonable we don't check all possible values`?

tests/Unit/NumericalAlgorithms/Spectral/Test_Spectral.cpp
```diff
@@ -179,8 +182,9 @@ void test_linear_filter_impl() {
     dgemv_('N', n, n, 1.0, nodal_to_modal_matrix.data(),
            nodal_to_modal_matrix.spacing(), u_filtered.data(), 1, 0.0,
            u_filtered_spectral.data(), 1);
+    Approx local_approx = Approx::custom().epsilon(1e-13).scale(1.0);
```
nilsdeppe, line 185: same quesiton about a scale here

tests/Unit/NumericalAlgorithms/Spectral/Test_Spectral.cpp
```diff
@@ -67,11 +67,13 @@ void test_exact_differentiation_impl(const Function& max_poly_deg) {
       const auto analytic_derivative =
           unit_polynomial_derivative(p, collocation_pts);
       if (BasisType == Spectral::Basis::FiniteDifference and n > 20) {
-        Approx local_approx = Approx::custom().epsilon(1.0e-13).scale(1.0);
+        Approx local_approx = Approx::custom().epsilon(1.0e-11).scale(1.0);
         CHECK_ITERABLE_CUSTOM_APPROX(analytic_derivative, numeric_derivative,
                                      local_approx);
       } else {
-        CHECK_ITERABLE_APPROX(analytic_derivative, numeric_derivative);
+        Approx local_approx = Approx::custom().epsilon(1.0e-11).scale(1.0);
```
nilsdeppe, line 74: same question about the scale here

tests/Unit/NumericalAlgorithms/Spectral/Test_Spectral.cpp
```diff
@@ -242,7 +246,7 @@ template <Spectral::Basis BasisType, Spectral::Quadrature QuadratureType,
 void test_exact_extrapolation_impl(const Function& max_poly_deg) {
   const DataVector target_points{-1.66, 1., 1.5, 1.98, 2.};
   // Errors are larger when extrapolating outside of the original grid:
-  const double eps = 1.e-9;
+  const double eps = 1.e-5;
```
nilsdeppe: Oof, this is getting pretty darn large. Do you know for what number of grid points? We could keep the value(s) we test for  the same as before, which might help.

tests/Unit/Evolution/Ader/Test_Matrices.cpp
```diff
@@ -282,7 +282,12 @@ void test_predictor_inverse_temporal_matrix() {
       };
   for (size_t i = Spectral::minimum_number_of_points<
            Spectral::Basis::Legendre, Spectral::Quadrature::GaussLobatto>;
-       i <= Spectral::maximum_number_of_points<Spectral::Basis::Legendre>;
+       i <= 12;
```
nilsdeppe: maybe `std::min(Spectral::maximum_number_of_points<Spectral::Basis::Legendre>, 12_st)`?

tests/Unit/Evolution/Ader/Test_Matrices.cpp
```diff
@@ -282,7 +282,12 @@ void test_predictor_inverse_temporal_matrix() {
...
+
+       /*The maximum number of allowed collocation points is given by
```
nilsdeppe, line 289: Maybe: `We only derived the expected matrices up to 12 points, so we only test to that.`? Then it's clear _why_ we don't test more, not just that we don't.

nilsdeppe (changes requested):
Some small suggestions, please squash right away :)

teukolsky:
I'd like to echo Nils Deppe's comment from July 23. Using more than 12 points in a DG code is most likely a misuse of the code once we have AMR fully working. The fact that BBH runs may currently be faster with more than 12 points is just telling us we're running spectre as a spectral code: a small number of domains with a large number of basis functions. So I think it's fine to replace 12 with 20 for now, but once AMR is working, we should change it back to 12. We need to make it difficult for novice users to misuse the code.

tests/Unit/NumericalAlgorithms/LinearOperators/Test_CoefficientTransforms.cpp
mar-celine: I haven't checked how `scale` works but I saw that other tests that use `CHECK_ITERABLE_CUSTOM_APPROX` use `scale(1.0)` (and the tests will fail to pass without `scale` regardless of tolerance)

`scale` can be left out when doing just `CHECK` tests though, so I think it's specific to the `ITERABLE` case.

tests/Unit/NumericalAlgorithms/Spectral/Test_Spectral.cpp
mar-celine: Starting from the original `eps` being `1.e-9`, the number of points can be raised to 13 and this test will still pass.
14 pts requires `1.e-8`,
15 pts requires `1.e-8`,
16 pts requires `1.e-7`,
17 pts requires `1.e-7`,
18 pts requires `1.e-6`,
19 pts requires `1.e-6`,
20 pts requires `1.e-5` - so I will be restricted the test values to `12`, but let me know if I should do anything else!

tests/Unit/NumericalAlgorithms/LinearOperators/Test_CoefficientTransforms.cpp
nilsdeppe: Oh, sorry, I should've clarified. `scale` sets the the absolute tolerance to `epsilon * scale` (if I've understood Catch documentation correctly), while `epsilon` sets a relative tolerance. In any case, leave it as-is.

nilsdeppe (dismissed):

tests/Unit/NumericalAlgorithms/Spectral/Test_Spectral.cpp
nilsdeppe: I think this would be very useful to have in the test as a comment, eg
```cpp
 // 14 pts requires 1.e-8
 // 15 pts requires 1.e-8
 // 16 pts requires 1.e-7
 // 17 pts requires 1.e-7
 // 18 pts requires 1.e-6
 // 19 pts requires 1.e-6
 // 20 pts requires 1.e-5
```
Do you mind adding that quickly? That'll hopefully save someone in the future time trying to figure out what's going on if they increase the number of points

nilsdeppe (dismissed):

nilsdeppe:
The CI failures should get addressed. Seems like maybe random values uncovering something with either tolerances or a bug?

nilsdeppe (approved):

PR #4986, opened 2023-05-04
Pass grid to target velocity in Compute*Quantities

knelli2:
## Proposed changes

I will need this quantity for size control. This just passes it, but it's currently not used. That will be a future PR.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5165, opened 2023-07-10
Size unused tensor in IntrpTargetDetail

knelli2:
## Proposed changes

This is a hack, I know. But it segfaults if you don't do this. I hope to change this when we do the whole rewrite.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
+++ b/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
@@ -627,7 +627,7 @@ constexpr bool any_index_in_frame_v =
 
 /// Calls compute_vars_to_interpolate to compute
 /// InterpolationTargetTag::vars_to_interpolate_to_target from the source
-/// variables.  Does any frame tranformations needed.
+/// variables.  Does any frame transformations needed.
 template <typename InterpolationTargetTag, typename SourceTags,
           typename Metavariables, typename ElementId>
 void compute_dest_vars_from_source_vars(
@@ -646,7 +646,7 @@ void compute_dest_vars_from_source_vars(
     // For interpolation without an Interpolator ParallelComponent,
     // this is because the InterpWithoutInterpComponent event will be called
     // after the Action that keeps functions of time up to date.
-    // For interpolation with an Interpolator ParallelCompoent,
+    // For interpolation with an Interpolator ParallelComponent,
     // this is because the functions of time are made up to date before
     // calling SendPointsToInterpolator.
     if constexpr (any_index_in_frame_v<SourceTags, Frame::Inertial> and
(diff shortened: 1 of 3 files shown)


wthrowe:
Is this a compiler bug, a SpECTRE bug, ...?

knelli2:
This is a me bug that I introduced in #4986. It's runtime. Basically in my testing I never used an interpolation target that was in the grid frame so the unsized tensor was never used. But @geoffrey4444 just rebased from an old branch and ran into this bug so we need to size it properly. However, the quantity doesn't make sense in the grid frame (as it's a mesh velocity between the grid and the target frame, aka the grid frame) so I just set it to zero.

wthrowe:
I'm still not clear on what the bug is.  The argument was added in #4986, so I can't see how @geoffrey4444's old code could be expecting a different value since it wasn't passed at all before.  And if it's unused, how can it cause a segfault?

Can you add a test?  That would probably clear it up.

knelli2:
@wthrowe  So yes that argument was added and used in #4986. The only thing that actually uses it is `ComputeExcisionBoundaryVolumeQuantities`. It subtracts two things with a tensor expression where one of the things being subtracted is that new argument. The apply function in `ComputeExcisionBoundaryVolumeQuantities` is called by `compute_dest_vars_from_source_vars` which is edited in this PR. So previously, if the target frame was Grid, `compute_dest_vars_from_source_vars` would pass an empty tensor, then `ComputeExcisionBoundaryVolumeQuantities` would use that empty tensor in an operation with a fully sized tensor and cause a segfault. So @geoffrey4444's old code couldn't have had this issue because the argument didn't exist. And I never encountered it because I never did an observation where the target was in the Grid frame. But @geoffrey4444 does do that so when he rebased he hit the segfault, hence the PR.

I just checked and the apply function of `ComputeExcisionBoundaryVolumeQuantities` is tested, but it's tested in isolation so we pass properly sized tensors for that new argument. However, `compute_dest_vars_from_source_vars` isn't tested anywhere and is only used in production code, which is why all our tests pass.

wthrowe:
I haven't kept up with all the details of how this part of the code all ties together.  Is there a reason that it would be hard to write a test for this?

knelli2:
Ok well I was mistaken earlier. Technically `compute_dest_vars_from_source_vars` is tested via `Test_InterpolatorReceiveAndDumpVolumeData`, but it isn't tested with `ComputeExcisionBoundaryVolumeQuantities` in that test. And in the test for `ComputeExcisionBoundaryVolumeQuantities`, we don't use `compute_dest_vars_from_source_vars`. But it's the interplay between these two (in one specific frame), that causes the issue.

I'm not sure how "hard" it will be per-say, but it will definitely be tedious and consume a bit of my time. And with the impending rewrite of the interpolation framework, this is going to inevitably get changed so I'm not sure its the best use of my time to try and write a test now.

wthrowe:
You don't need to test it with the horizon stuff.  Can you just write an InterpolationTargetTag that checks that it is called with a non-empty tensor?

knelli2:
Oh yes I can do that. That sounds a lot easier than what I was thinking.

knelli2:
Ok @wthrowe I wrote a quick test that only checks the sizes of the mesh velocities and jacobians for transformations to both Grid and Distorted frame.

wthrowe (approved):

wthrowe:
Ignoring timeout.

PR #5166, opened 2023-07-11
Add Radial Function to Translation Map

AlexCarpenter46:
## Proposed changes

This pr adds a radial MathFunction to the Translation map that'll allow the outer boundary to remain fixed while the inner boundary translates with the compact objects. The constructor is overloaded to keep a uniform translation so that everywhere using the simple translation map doesn't break. There will be a followup pr to add translation map options into the EvolveGhBinaryBlackHole.hpp and the Binary yaml input files.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
@@ -3,17 +3,24 @@
 
 #include "Domain/CoordinateMaps/TimeDependent/Translation.hpp"
 
+#include <cmath>
 #include <ostream>
 #include <pup.h>
 #include <pup_stl.h>
 #include <utility>
 
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
 #include "DataStructures/Tensor/Identity.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "NumericalAlgorithms/RootFinding/RootBracketing.hpp"
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "PointwiseFunctions/MathFunctions/MathFunction.hpp"
+#include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/DereferenceWrapper.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/MakeArray.hpp"
 #include "Utilities/MakeWithValue.hpp"
 #include "Utilities/StdArrayHelpers.hpp"
 #include "Utilities/StdHelpers.hpp"
@@ -22,7 +29,29 @@ namespace domain::CoordinateMaps::TimeDependent {
 
 template <size_t Dim>
 Translation<Dim>::Translation(std::string function_of_time_name)
-    : f_of_t_name_(std::move(function_of_time_name)) {}
+    : f_of_t_name_(std::move(function_of_time_name)),
+      f_of_r_(nullptr),
+      center_(make_array<Dim, double>(0.0)) {}
+
+template <size_t Dim>
+Translation<Dim>::Translation(
+    std::string function_of_time_name,
+    std::unique_ptr<MathFunction<1, Frame::Inertial>> radial_function,
+    std::array<double, Dim>& center)
+    : f_of_t_name_(std::move(function_of_time_name)),
+      f_of_r_(std::move(radial_function)),
+      center_(center) {}
+
+template <size_t Dim>
+Translation<Dim>::Translation(const Translation<Dim>& Translation_Map)
+    : f_of_t_name_(Translation_Map.f_of_t_name_),
+      center_(Translation_Map.center_) {
+  if (Translation_Map.f_of_r_ == nullptr) {
+    f_of_r_ = nullptr;
+  } else {
+    f_of_r_ = Translation_Map.f_of_r_->get_clone();
+  }
+}
 
 template <size_t Dim>
 template <typename T>
@@ -31,23 +60,7 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::operator()(
     const std::unordered_map<
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time) const {
-  ASSERT(functions_of_time.count(f_of_t_name_) == 1,
-         "The function of time '" << f_of_t_name_
-                                  << "' is not one of the known functions of "
-                                     "time. The known functions of time are: "
-                                  << keys_of(functions_of_time));
-  std::array<tt::remove_cvref_wrap_t<T>, Dim> result{};
-  const DataVector function_of_time =
-      functions_of_time.at(f_of_t_name_)->func(time)[0];
-  ASSERT(function_of_time.size() == Dim,
-         "The dimension of the function of time ("
-             << function_of_time.size()
-             << ") does not match the dimension of the translation map (" << Dim
-             << ").");
-  for (size_t i = 0; i < Dim; i++) {
-    gsl::at(result, i) = gsl::at(source_coords, i) + function_of_time[i];
-  }
-  return result;
+  return coord_helper(source_coords, time, functions_of_time, 0);
 }
 
 template <size_t Dim>
@@ -62,6 +75,9 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
                                      "time. The known functions of time are: "
                                   << keys_of(functions_of_time));
   std::array<double, Dim> result{};
+  for (size_t i = 0; i < Dim; i++) {
+    gsl::at(result, i) = gsl::at(target_coords, i);
+  }
   const DataVector function_of_time =
       functions_of_time.at(f_of_t_name_)->func(time)[0];
   ASSERT(function_of_time.size() == Dim,
@@ -69,8 +85,23 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
              << function_of_time.size()
              << ") does not match the dimension of the translation map (" << Dim
              << ").");
+  double radial_function_value = 1.0;
+  if (f_of_r_ != nullptr) {
+    std::array<double, Dim> distance_to_center = target_coords;
+    double root_factor = 0.;
+    for (size_t i = 0; i < Dim; i++) {
+      // shifted center
+      gsl::at(distance_to_center, i) -= gsl::at(center_, i);
+      // root_factor needed for root_finder
+      root_factor +=
+          -2.0 * gsl::at(distance_to_center, i) * function_of_time[i];
+    }
+    const double root = root_finder(magnitude(distance_to_center),
+                                    sqrNorm(function_of_time), root_factor);
+    radial_function_value = (*f_of_r_)(root);
+  }
   for (size_t i = 0; i < Dim; i++) {
-    gsl::at(result, i) = gsl::at(target_coords, i) - function_of_time[i];
+    gsl::at(result, i) -= function_of_time[i] * radial_function_value;
   }
   return result;
 }
@@ -82,44 +113,161 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
     const std::unordered_map<
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time) const {
+  return coord_helper(source_coords, time, functions_of_time, 1);
+}
+
+template <size_t Dim>
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>
+Translation<Dim>::jacobian(
+    const std::array<T, Dim>& source_coords, const double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time) const {
+  if (f_of_r_ == nullptr) {
+    return identity<Dim>(dereference_wrapper(source_coords[0]));
+  } else {
+    std::array<tt::remove_cvref_wrap_t<T>, Dim> distance_to_center{};
+    for (size_t i = 0; i < Dim; i++) {
+      gsl::at(distance_to_center, i) =
+          gsl::at(source_coords, i) - gsl::at(center_, i);
+    }
+    auto radius = magnitude(distance_to_center);
+
+    const DataVector function_of_time =
+        functions_of_time.at(f_of_t_name_)->func(time)[0];
+
+    auto result = make_with_value<
+        tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>>(
+        dereference_wrapper(source_coords[0]), 0.0);
+    for (size_t i = 0; i < Dim; i++) {
+      for (size_t j = 0; j < Dim; j++) {
+        for (size_t k = 0; k < get_size(radius); k++) {
+          if (get_element(radius, k) > 1.e-13) {
+            result.get(i, j) = (*f_of_r_).first_deriv(get_element(radius, k)) *
+                               gsl::at(function_of_time, i) *
+                               gsl::at(distance_to_center, j) /
+                               get_element(radius, k);
+          } else {
+            result.get(i, j) = (*f_of_r_).second_deriv(get_element(radius, k)) *
+                               gsl::at(function_of_time, i) *
+                               gsl::at(distance_to_center, j);
+          }
+        }
+      }
+      result.get(i, i) += 1.0;
+    }
+    return result;
+  }
+}
+template <size_t Dim>
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>
+Translation<Dim>::inv_jacobian(
+    const std::array<T, Dim>& source_coords, const double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time) const {
+  if (f_of_r_ == nullptr) {
+    return identity<Dim>(dereference_wrapper(source_coords[0]));
+  } else {
+    return determinant_and_inverse(
+               jacobian(source_coords, time, functions_of_time))
+        .second;
+  }
+}
+
+template <size_t Dim>
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
+    const std::array<T, Dim>& source_coords, double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time,
+    const size_t function_or_deriv_index) const {
   ASSERT(functions_of_time.count(f_of_t_name_) == 1,
          "The function of time '" << f_of_t_name_
                                   << "' is not one of the known functions of "
                                      "time. The known functions of time are: "
                                   << keys_of(functions_of_time));
-  std::array<tt::remove_cvref_wrap_t<T>, Dim> result{};
-  const auto function_of_time_and_deriv =
-      functions_of_time.at(f_of_t_name_)->func_and_deriv(time);
-  const DataVector& velocity = function_of_time_and_deriv[1];
-  ASSERT(velocity.size() == Dim,
+  const auto func_or_deriv_of_time =
+      gsl::at(functions_of_time.at(f_of_t_name_)->func_and_deriv(time),
+              function_or_deriv_index);
+  ASSERT(func_or_deriv_of_time.size() == Dim,
          "The dimension of the function of time ("
-             << velocity.size()
+             << func_or_deriv_of_time.size()
              << ") does not match the dimension of the translation map (" << Dim
              << ").");
+  std::array<tt::remove_cvref_wrap_t<T>, Dim> result{};
+  // sizing the result and getting the radial function value
   for (size_t i = 0; i < Dim; i++) {
-    gsl::at(result, i) = make_with_value<tt::remove_cvref_wrap_t<T>>(
-        dereference_wrapper(gsl::at(source_coords, i)), velocity[i]);
+    gsl::at(result, i) = gsl::at(source_coords, i);
+  }
+  auto radial_function_value = make_with_value<tt::remove_cvref_wrap_t<T>>(
+      dereference_wrapper(source_coords[0]), 1.0);
+  if (f_of_r_ != nullptr) {
+    radial_function_value = (*f_of_r_)(magnitude(result - center_));
+  }
+  if (function_or_deriv_index == 1) {
+    for (size_t i = 0; i < Dim; i++) {
+      gsl::at(result, i) = 0.0;
+    }
+  }
+  for (size_t i = 0; i < Dim; i++) {
+    gsl::at(result, i) +=
+        gsl::at(func_or_deriv_of_time, i) * radial_function_value;
   }
   return result;
 }
 
 template <size_t Dim>
-template <typename T>
-tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>
-Translation<Dim>::jacobian(const std::array<T, Dim>& source_coords) const {
-  return identity<Dim>(dereference_wrapper(source_coords[0]));
-}
-
-template <size_t Dim>
-template <typename T>
-tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>
-Translation<Dim>::inv_jacobian(const std::array<T, Dim>& source_coords) const {
-  return identity<Dim>(dereference_wrapper(source_coords[0]));
+double Translation<Dim>::root_finder(const double target_radius,
+                                     const double squared_function_of_time,
+                                     const double root_factor) const {
+  const double center_offset = sqrt(squared_function_of_time);
+  // These bounds come from the SpEC implementation
+  double lower_bound = (target_radius - center_offset) * (1.0 - 1.e-9);
+  double upper_bound = (target_radius + center_offset) * (1.0 + 1.e-9);
+  if (lower_bound < 0.) {
+    lower_bound = 0.;
+  }
+  // If the upper and lower bounds are < 0., then there's no need to find a root
+  // the root will be 0.
+  if (upper_bound < 0.) {
+    return 0.;
+  }
+  double function_at_lower_bound = std::numeric_limits<double>::min();
+  double function_at_upper_bound = (*f_of_r_)(upper_bound);
+  // bracket function requires a function to return std::optionals
+  const auto radius_bracket_function =
+      [this, &target_radius, &squared_function_of_time,
+       &root_factor](const double x) -> std::optional<double> {
+    return square(target_radius) +
+           ((*f_of_r_)(x)) *
+               (root_factor + squared_function_of_time * (*f_of_r_)(x)) -
+           square(x);
+  };
+  RootFinder::bracket_possibly_undefined_function_in_interval(
+      &lower_bound, &upper_bound, &function_at_lower_bound,
+      &function_at_upper_bound, radius_bracket_function);
+  const double absolute_tol = std::numeric_limits<double>::epsilon() *
+                              std::max(target_radius, center_offset);
+  const double relative_tol = 2.0 * std::numeric_limits<double>::epsilon();
+  const auto wrapper = [this, &target_radius, &squared_function_of_time,
+                        &root_factor](const double x) {
+    return square(target_radius) +
+           ((*f_of_r_)(x)) *
+               (root_factor + squared_function_of_time * (*f_of_r_)(x)) -
+           square(x);
+  };
+  return RootFinder::toms748(wrapper, lower_bound, upper_bound,
+                             function_at_lower_bound, function_at_upper_bound,
+                             absolute_tol, relative_tol);
 }
 
 template <size_t Dim>
 void Translation<Dim>::pup(PUP::er& p) {
-  size_t version = 0;
+  size_t version = 1;
   p | version;
   // Remember to increment the version number when making changes to this
   // function. Retain support for unpacking data written by previous versions
@@ -127,11 +275,22 @@ void Translation<Dim>::pup(PUP::er& p) {
   if (version >= 0) {
     p | f_of_t_name_;
   }
+  if (version >= 1) {
+    p | f_of_r_;
+    p | center_;
+  } else if (p.isUnpacking()) {
+    f_of_r_ = nullptr;
+    center_ = make_array<Dim, double>(0.0);
+  }
 }
 
 template <size_t Dim>
 bool operator==(const Translation<Dim>& lhs, const Translation<Dim>& rhs) {
-  return lhs.f_of_t_name_ == rhs.f_of_t_name_;
+  return lhs.f_of_t_name_ == rhs.f_of_t_name_ and
+         (lhs.f_of_r_ == nullptr) == (rhs.f_of_r_ == nullptr) and
+         ((lhs.f_of_r_ == nullptr and rhs.f_of_r_ == nullptr) or
+          *lhs.f_of_r_ == *rhs.f_of_r_) and
+         lhs.center_ == rhs.center_;
 }
 
 // Explicit instantiations
@@ -148,31 +307,37 @@ GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3))
 
 #define DTYPE(data) BOOST_PP_TUPLE_ELEM(1, data)
 
-#define INSTANTIATE(_, data)                                           \
-  template std::array<tt::remove_cvref_wrap_t<DTYPE(data)>, DIM(data)> \
-  Translation<DIM(data)>::operator()(                                  \
-      const std::array<DTYPE(data), DIM(data)>& source_coords,         \
-      const double time,                                               \
-      const std::unordered_map<                                        \
-          std::string,                                                 \
-          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&   \
-          functions_of_time) const;                                    \
-  template std::array<tt::remove_cvref_wrap_t<DTYPE(data)>, DIM(data)> \
-  Translation<DIM(data)>::frame_velocity(                              \
-      const std::array<DTYPE(data), DIM(data)>& source_coords,         \
-      const double time,                                               \
-      const std::unordered_map<                                        \
-          std::string,                                                 \
-          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&   \
-          functions_of_time) const;                                    \
-  template tnsr::Ij<tt::remove_cvref_wrap_t<DTYPE(data)>, DIM(data),   \
-                    Frame::NoFrame>                                    \
-  Translation<DIM(data)>::jacobian(                                    \
-      const std::array<DTYPE(data), DIM(data)>& source_coords) const;  \
-  template tnsr::Ij<tt::remove_cvref_wrap_t<DTYPE(data)>, DIM(data),   \
-                    Frame::NoFrame>                                    \
-  Translation<DIM(data)>::inv_jacobian(                                \
-      const std::array<DTYPE(data), DIM(data)>& source_coords) const;
+#define INSTANTIATE(_, data)                                                \
+  template std::array<tt::remove_cvref_wrap_t<DTYPE(data)>, DIM(data)>      \
+  Translation<DIM(data)>::operator()(                                       \
+      const std::array<DTYPE(data), DIM(data)>& source_coords, double time, \
+      const std::unordered_map<                                             \
+          std::string,                                                      \
+          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&        \
+          functions_of_time) const;                                         \
+  template std::array<tt::remove_cvref_wrap_t<DTYPE(data)>, DIM(data)>      \
+  Translation<DIM(data)>::frame_velocity(                                   \
+      const std::array<DTYPE(data), DIM(data)>& source_coords, double time, \
+      const std::unordered_map<                                             \
+          std::string,                                                      \
+          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&        \
+          functions_of_time) const;                                         \
+  template tnsr::Ij<tt::remove_cvref_wrap_t<DTYPE(data)>, DIM(data),        \
+                    Frame::NoFrame>                                         \
+  Translation<DIM(data)>::jacobian(                                         \
+      const std::array<DTYPE(data), DIM(data)>& source_coords, double time, \
+      const std::unordered_map<                                             \
+          std::string,                                                      \
+          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&        \
+          functions_of_time) const;                                         \
+  template tnsr::Ij<tt::remove_cvref_wrap_t<DTYPE(data)>, DIM(data),        \
+                    Frame::NoFrame>                                         \
+  Translation<DIM(data)>::inv_jacobian(                                     \
+      const std::array<DTYPE(data), DIM(data)>& source_coords, double time, \
+      const std::unordered_map<                                             \
+          std::string,                                                      \
+          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&        \
+          functions_of_time) const;
 
 GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3),
                         (double, DataVector,
--- a/src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
@@ -11,28 +11,91 @@
 #include <unordered_map>
 
 #include "DataStructures/Tensor/TypeAliases.hpp"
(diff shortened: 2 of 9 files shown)


src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
 namespace TimeDependent {
 /*!
  * \ingroup CoordMapsTimeDependentGroup
- * \brief Translation map defined by \f$\vec{x} = \vec{\xi}+\vec{T}(t)\f$.
+ * \brief Translation map defined by \f$\vec{x} = \vec{\xi}+F(r)\vec{T}(t)\f$.
  *
- * The map adds a translation, \f$\vec{T}(t)\f$, to the coordinates
- * \f$\vec{\xi}\f$, where \f$\vec{T}(t)\f$ is a FunctionOfTime.
+ * \details The map adds a translation, \f$f(r)\vec{T}(t)\f$, to the coordinates
```
knelli2: `F(r)` not `f(r)` for consistency.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * \f$\vec{\xi}\f$, where \f$\vec{T}(t)\f$ is a FunctionOfTime and F(r) is a 1D
```
knelli2: `\f$F(r)\f$`

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * radial MathFunction. The radius of each point is found by subtracting the
+ * center of the map argument from the coordinates \f$\vec{\xi}\f$ or the
```
knelli2: `of the`

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * translated coordinate \f$\vec{\bar{\xi}}\f$. The Translation Map class is
+ * overloaded so that the user can choose if a radial depedence is needed for
+ * their problem.
```
knelli2: Please add
```
If a radial dependence is not specified, this sets \f$F(r) = 1\f$.
```

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ *
+ * ### Mapped Coordinates
+ *
+ * The translation map translates the coordinates \f$\vec{\xi}\f$ to the
+ * translated coordinates
+ * \f{equation}{
+ * $\vec{\bar{\xi}} = \vec{\xi} + F(r)\vec{T}(t)$
+ * \f}
+ *
+ * If you only supply a FunctionOfTime, the radial function will be set to 1.0
```
knelli2: `If you only supply a FunctionOfTime to the constructor of this class, ...`

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * causing a uniform translation for your coordinates. If a FunctionOfTime,
+ * MathFunction and map center is passed in, the radius will be found through
```
knelli2: Oxford comma! `FunctionOfTime, MathFunction, and ...`. Also `is` -> `are`.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * radial MathFunction based on your radius. When the radius is bigger
+ * than 1.e-13, the jacobian is given by:
+ * \f{equation}{
+ * J^{i}_{j} = dF(r) * $\vec{T}(t)$ * ($\vec{\xi} - \vec{c}$) / r
+ * \f}
+ * Where dF(r) is the first derivative of the MathFunction, \f$\vec{\xi}\f$ is
+ * the source coordinates, \f$\vec{c}\f$ is the center of your map, and r is the
+ * radius.
+ *
+ * At a radius smaller than 1e-13, we use l'hopital's to compute the jacobian as
```
knelli2: `L'Hopital's rule`

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -42,6 +108,15 @@ class Translation {
   Translation() = default;
   explicit Translation(std::string function_of_time_name);
 
+  explicit Translation(
+      std::string function_of_time_name,
+      std::unique_ptr<MathFunction<1, Frame::Inertial>> radial_function,
+      std::array<double, Dim>& center);
+
+  Translation(const Translation<Dim>& Translation_Map);
```
knelli2, line 113: clang-tidy is complaining that you don't have the rest of the constructors/assignment operators.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -55,7 +130,7 @@ class Translation {
   /// what should happen if the inverse were to succeed for some points in a
   /// DataVector but fail for other points.
   std::optional<std::array<double, Dim>> inverse(
-      const std::array<double, Dim>& target_coords, double time,
+      const std::array<double, Dim>& translated_coords, double time,
```
knelli2: Keep this as `target_coords` please. It matches all the other time dependent maps we have.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -22,12 +27,34 @@ namespace domain::CoordinateMaps::TimeDependent {
...
+      center_(Translation_Map.center_) {
+  if (Translation_Map.f_of_r_ == nullptr) {
+    f_of_r_ = nullptr;
+  } else {
+    f_of_r_ = Translation_Map.f_of_r_->get_clone();
+  }
+}
+
+template <size_t Dim>
+template <typename DataType>
```
knelli2: I know `T` isn't super descriptive, but let's keep it because all other time dependent maps use `T` and we want to keep everything consistent. (whole file)

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -36,23 +63,43 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::operator()(
                                   << "' is not one of the known functions of "
                                      "time. The known functions of time are: "
                                   << keys_of(functions_of_time));
-  std::array<tt::remove_cvref_wrap_t<T>, Dim> result{};
+  std::array<tt::remove_cvref_wrap_t<DataType>, Dim> result{};
```
knelli2: Actually, since you have to use `tt::remove_cvref_wrap_t<T>` a lot more often now, keep the original template `T`, but define `using DataType = tt::remove_cvref_wrap_t<T>;` to cut down on typing, and then you don't have to change this everywhere. You can define this type alias in every function that you need to.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * translated coordinate \f$\vec{\bar{\xi}}\f$. The Translation Map class is
+ * overloaded so that the user can choose if a radial depedence is needed for
+ * their problem.
+ *
+ * ### Mapped Coordinates
+ *
+ * The translation map translates the coordinates \f$\vec{\xi}\f$ to the
+ * translated coordinates
+ * \f{equation}{
+ * $\vec{\bar{\xi}} = \vec{\xi} + F(r)\vec{T}(t)$
```
knelli2: Remove these $. You use `\f$` to wrap math statements in regular text, outside of \f{equation} blocks. Same below in all the \f{equation} blocks.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * ### Inverse Translation
+ * The inverse translation translates the coordinate \f$\vec{\bar{\xi}}\f$ to
+ * the original coordinates
+ * \f{equation}{
+ * $\vec{\xi} = \vec{\bar{\xi}} - F(root)\vec{T}(t)$
+ * \f}
+ * However, we need to do a root find to see where F(r) and \f$\vec{T}(t)\f$
+ * cross zero. This is done by bracketing and finding the root of
+ * \f{equation}{
+ * \bar{r}^2 + F(x) * (Root_Factor + |$\vec{T}(t)$| * F(x)) - x^2
```
knelli2: `Root_Factor` doesn't render properly. Call it a variable, like `A`. Also, don't use `*` here. It's understood that it's multiplication.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * \f}
+ * This equation comes from the SpEC implementation where x is the root guess
+ * and the Root_Factor is given by
+ * \f{equation}{
+ * Root_Factor = -2.0 * |$\vec{\xi} - \vec{c}$| * $\vec{T}(t)$
```
knelli2: Is this a dot product or multiplication? If it's multiplication, you're missing the norm on $\vec{T}(t)$ and remove the `*`. If it's a dot product, remove the norm on $\vec{\xi} - \vec{c}$ and use `\cdot` instead of `*`.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -29,10 +30,75 @@ namespace CoordinateMaps {
...
+ * \f}
+ *
+ * ### Frame Velocity
+ *
+ * The frame velocity is given by
+ * \f{equation}{
+ * Frame_Velocity = $\vec{dT}(t)$ * F(r)
```
knelli2: Just call this $\vec{v}$, not `Frame_Velocity`. Also use $\frac{dT(t)}{dt}$ not just $dT$. The second could be considered a form. Same below, for both $T(t)$ and $F(r)$

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
knelli2: This function can be simplified a bit I think. You have some extra temporaries you define that can be avoided. I believe this will work:

```cpp
std::array<tt::remove_cvref_wrap_t<DataType>, Dim> result{};
    gsl::at(result, i) = gsl::at(source_coords, i);
  }
  const DataVector function_of_time =
      functions_of_time.at(f_of_t_name_)->func(time)[0];
  ASSERT(function_of_time.size() == Dim,
         "The dimension of the function of time ("
             << function_of_time.size()
             << ") does not match the dimension of the translation map (" << Dim
             << ").");
  // Section for handling overloaded constructor
  auto radial_function_value =
      make_with_value<tt::remove_cvref_wrap_t<DataType>>(gsl::at(result, 0),
                                                         1.0);
  if (f_of_r_ != nullptr) {
    // finding the radius
    radial_function_value = (*f_of_r_)(magnitude(result - center_));
  }
    gsl::at(result, i) += function_of_time[i] * radial_function_value;
  }
  return result;
```

Some explanation. The main things that helped simplify this were the operations in `StdArrayHelpers.hpp` and `make_with_value`. However, the `tt::remove_cvref_wrap_t<DataType>` made things a bit annoying, so I immediately put the `source_coords` into the correct data type `result` and just worked with that. This allowed the one line implementation of `radial_function_value = (*f_of_r_)(magnitude(result - center_));`. Also, if you initially `make_with_value` the `radial_function_value` with `1.0`, then you don't have to do a redundant assignment when `f_of_r_ == nullptr`.

Let me know if you have any questions about what happened here.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -71,11 +146,23 @@ class Translation {
...
   template <typename T>
   tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame> jacobian(
-      const std::array<T, Dim>& source_coords) const;
+      const std::array<T, Dim>& source_coords, const double time,
+      const std::unordered_map<
+          std::string,
+          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+          functions_of_time) const;
+
+  double root_finder(const double& translated_radius,
```
knelli2: Make this private.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -69,16 +116,38 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
              << function_of_time.size()
              << ") does not match the dimension of the translation map (" << Dim
              << ").");
+  double radial_function_value = 1.0;
+  if (f_of_r_ != nullptr) {
```
knelli2, line 89: This function can similarly be optimized a bit. Let me know if you want to discuss this one as well.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -96,42 +165,175 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
+  }
 }
 
 template <size_t Dim>
-template <typename T>
-tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>
-Translation<Dim>::inv_jacobian(const std::array<T, Dim>& source_coords) const {
-  return identity<Dim>(dereference_wrapper(source_coords[0]));
+double Translation<Dim>::root_finder(const double& translated_radius,
+                                     const double& magnitude_function_of_time,
```
knelli2: `magnitude_function_of_time` is misleading. This is presumably the magnitude squared because you do `sqrt` below.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -96,42 +165,175 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
+                                     const double& root_factor) const {
+  const double center_offset = sqrt(magnitude_function_of_time);
+  double lower_bound = (translated_radius - center_offset) * (1.0 - 1.e-9);
```
knelli2: Can you add a comment about why you chose `+/-1.e-9`? Even if it's "this is what spec did".

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -96,42 +165,175 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
+                        &root_factor](const double x) {
+    return square(translated_radius) +
+           ((*f_of_r_)(x)) *
+               (root_factor + magnitude_function_of_time * (*f_of_r_)(x)) -
+           square(x);
+  };
+  return RootFinder::toms748(wrapper, lower_bound, upper_bound,
+                             function_at_lower_bound, function_at_upper_bound,
+                             absolute_tol, relative_tol);
+  return 0.;
```
knelli2: Remove?

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -96,42 +165,175 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
              << velocity.size()
              << ") does not match the dimension of the translation map (" << Dim
              << ").");
+  // Section for handling overloaded constructor
+  tt::remove_cvref_wrap_t<DataType> radial_function_value(
```
knelli2: This is almost identical to the `operator()` I think except for the function of time gets the deriv instead of the function. Maybe factor this out into a free function that just takes either 0 or 1 for the function value vs derivative.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -96,42 +165,175 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
+    const DataVector function_of_time =
+        functions_of_time.at(f_of_t_name_)->func(time)[0];
+
+    auto result = make_with_value<
+        tnsr::Ij<tt::remove_cvref_wrap_t<DataType>, Dim, Frame::NoFrame>>(
+        dereference_wrapper(source_coords[0]), 0.0);
+    for (size_t i = 0; i < Dim; i++) {
+      for (size_t j = 0; j < Dim; j++) {
+        if constexpr (std::is_same_v<tt::remove_cvref_wrap_t<DataType>,
+                                     double>) {
```
knelli2: You should be able to use the same code for `double` and `DataVector` because the `first_deriv` function has both a `double` and `DataVector` overload.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -96,42 +165,175 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
+tnsr::Ij<tt::remove_cvref_wrap_t<DataType>, Dim, Frame::NoFrame>
+Translation<Dim>::jacobian(
+    const std::array<DataType, Dim>& source_coords, const double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time) const {
+  if (f_of_r_ == nullptr) {
+    return identity<Dim>(dereference_wrapper(source_coords[0]));
+  } else {
+    std::array<tt::remove_cvref_wrap_t<DataType>, Dim> distance_to_center{};
```
knelli2: This can also be simplified a bit. Especially the `Scalar`. I don't think that's necessary

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -96,42 +165,175 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
   // Remember to increment the version number when making changes to this
   // function. Retain support for unpacking data written by previous versions
   // whenever possible. See `Domain` docs for details.
   if (version >= 0) {
     p | f_of_t_name_;
   }
+  if (version >= 1) {
+    p | f_of_r_;
+    p | center_;
+  }
```
knelli2: You also need an `else if (p.isUnpacking())` clause here. This is because if we are reading a v0 object, we still need to set `f_of_r_` and `center_` to something because this is a v1 object. So you should pick defaults that match what a v0 object would expect (here, namely `nullptr` and 0).

src/PointwiseFunctions/MathFunctions/RegisterDerivedWithCharm.hpp
```diff
@@ -0,0 +1,7 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+void register_derived_with_charm();
+  // namespace domain::CoordinateMaps::MathFunctions
```
knelli2: You have the docs for a namespace, but no namspace. Please wrap this function in a namespace (probably `MathFunctions::`). Also add `_1d` to the name if you're only going to add 1D math functions.

src/PointwiseFunctions/MathFunctions/RegisterDerivedWithCharm.cpp
```diff
@@ -0,0 +1,12 @@
...
+// See LICENSE.txt for details.
+
+#include "PointwiseFunctions/MathFunctions/RegisterDerivedWithCharm.hpp"
+
+#include "PointwiseFunctions/MathFunctions/Gaussian.hpp"
+#include "PointwiseFunctions/MathFunctions/MathFunction.hpp"
+#include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
+
+void register_derived_with_charm() {
+    register_classes_with_charm<MathFunctions::Gaussian<1, Frame::Inertial>>();
```
knelli2: Add the rest of the 1D math functions to this so we can use them.

tests/Unit/Domain/CoordinateMaps/TimeDependent/CMakeLists.txt
```diff
@@ -28,6 +28,7 @@ target_link_libraries(
   DataStructures
   Domain
   DataStructuresHelpers
+  MathFunctions
```
knelli2, line 31: Add `MathFunctions` to the link libraries in the `src` directory as well.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -46,72 +54,114 @@ void test_translation() {
...
+  std::array<double, Dim> center{};
   fill_with_random_values(make_not_null(&point_xi), make_not_null(&gen),
                           make_not_null(&dist_double));
+  fill_with_random_values(make_not_null(&center), make_not_null(&gen),
+                          make_not_null(&dist_double));
+  const CoordinateMaps::TimeDependent::Translation<Dim> translation_map{
+      "translation"};
+  const CoordinateMaps::TimeDependent::Translation<Dim> radial_translation_map{
+      "translation",
+      std::make_unique<MathFunctions::Gaussian<1, Frame::Inertial>>(gaussian),
```
knelli2: You should be able to just pass the constructor arguments of Gaussian to `make_unique` and it'll construct the object for you rather than copying an existing object.

knelli2 (changes requested):
Looks good! Thanks for doing this! Let me know if you have any questions on things.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: Oh whoops, yeah that naming doesn't make much sense here

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: That makes sense! I meant to optimize these functions a bit, but forgot to

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: haha yeah this was from a test when the root finder wasn't working, will remove

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: They do, but since I have to compare the different radii to see if L'Hopital's applies or not, I think I have to check the DataType and parse through if it's a `DataVector` since the points might be above or below the threshold.

knelli2: There's actually a function for just this! In `src/Utilities/Containerhelpers.hpp` there is the `get_size` function. For a `double` this just returns 1 but for a `DataVector` it'll give you the size of the vector. Then in the same file there is a function `get_element` which takes the container, and the size you just got and gives you the element. For a `double`, it just returns the double. For a `DataVector`, it'll give you the ith element. So you can have the innermost loop be:

```cpp
for (size_t k=0 ; k<get_size(radius.get()); k++) {
  if (radius.get() > 1.e-13) {
    result.get(i, j) = (*f_of_r_).first_deriv(get_element(radius.get(), k)) *
      gsl::at(function_of_time, i) *
      gsl::at(distance_to_center, j) / get_element(radius.get(), k);
  } else {
    result.get(i, j) = (*f_of_r_).second_deriv(get_element(radius.get(), k)) *
      gsl::at(function_of_time, i) *
      gsl::at(distance_to_center, j);
  }
}
```

and it *should* handle it. I haven't tested this code yet.

AlexCarpenter46: oh wow, that's super cool, I'll give it a try see if it works!

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -15,87 +15,83 @@
...
  *
  * ### Mapped Coordinates
  *
  * The translation map translates the coordinates \f$\vec{\xi}\f$ to the
- * translated coordinates
- * \f{equation}{
- * $\vec{\bar{\xi}} = \vec{\xi} + F(r)\vec{T}(t)$
- * \f}
+ * target coordinates
+ * \f${equation}{
```
knelli2: So for this syntax (equations on separate lines), you don't include a `$` after the `\f`. It's just

```
\f{equation}{
...
\f}
```

If you want the equation in the middle of the sentence you do `\f$ ... \f$`. (And now you don't even need the `\f` when putting an equation in the middle of a sentence. You can just use $\LaTeX$ syntax, e.g. `$...$`)

Could you fix this for the rest of these blocks in the header file?

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -15,87 +15,83 @@
...
- * causing a uniform translation for your coordinates. If a FunctionOfTime,
- * MathFunction and map center is passed in, the radius will be found through
- * \f{equation}{
- * r = |$\vec{\xi} - \vec{c}$|
- * \f}
+ * If you only supply a FunctionOfTime to the constructor of this class, the
+ * radial function will be set to 1.0 causing a uniform translation for your
+ * coordinates. If a FunctionOfTime, MathFunction, and map center are passed in,
+ * the radius will be found through \f${equation}{ r = |\vec{\xi} - \vec{c}|
+ * \f$}
```
knelli2: If you want this equation between paragraphs, use the syntax in the other comment. If you want this in the middle of the sentence, remove `{equation}{` and the closing `}` at the end.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -15,87 +15,83 @@
...
- * \f}
+ * \f${equation}{
+ * \bar{r}^2 + F(x) (A + |\vec{T}(t)| F(x)) - x^2
+ * \f$}
  * This equation comes from the SpEC implementation where x is the root guess
- * and the Root_Factor is given by
- * \f{equation}{
- * Root_Factor = -2.0 * |$\vec{\xi} - \vec{c}$| * $\vec{T}(t)$
- * \f}
+ * and A is the Root Factor is given by
```
knelli2: `Root Factor is given by` -> `Root Factor given by`

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -15,87 +15,83 @@
...
+ * \f${equation}{
+ * A = -2.0 |\vec{\xi} - \vec{c}| \vec{T}(t)
```
knelli2: Is $\vec{T}(t)$ supposed to be $|\vec{T}(t)|$? Your code looks like it's supposed to be $A = -2.0\  (\vec{\xi} - \vec{c}) \cdot \vec{T}(t)$

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
knelli2: Ping on the `rule` part

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -174,6 +169,22 @@ class Translation {
...
+  // based on the option passed in, 0 for translated coordinates, and frame
+  // velocity for any other number.
+  template <typename T>
+  std::array<tt::remove_cvref_wrap_t<T>, Dim> coord_helper(
+      const std::array<T, Dim>& source_coords, double time,
+      const std::unordered_map<
+          std::string,
+          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+          functions_of_time,
+      const double& option) const;
```
knelli2: Instead of taking a double to decide function or derivative, take the index that you want to use as a `size_t` (no reference) (e.g. `0` for function and `1` for deriv). Also can you name it something more descriptive than `option`? Maybe `function_or_deriv_index`?

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -118,130 +86,74 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
-      magnitude_function_of_time += square(function_of_time[i]);
+      squared_function_of_time += square(function_of_time[i]);
       root_factor +=
           -2.0 * gsl::at(distance_to_center, i) * function_of_time[i];
     }
-    const double translated_radius = sqrt(radius_squared);
-    const double& root =
-        root_finder(translated_radius, magnitude_function_of_time, root_factor);
+    double root = root_finder(magnitude(distance_to_center),
+                              squared_function_of_time, root_factor);
```
knelli2: I think this part can be simplified down a bit more to (untested)

```cpp
double root_factor = 0.0;
for (size_t i=0; i<Dim; i++) {
  root_factor += -2.0 * gsl::at(result, i) * function_of_time[i];
}
const double root = root_finder(magnitude(result - center),
                                sqrNorm(function_of_time),
                                root_factor));
```

In general, but especially in the maps because they are called very often, we want to avoid doing unnecessary allocations and copies as this can negatively impact performance.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -174,6 +169,22 @@ class Translation {
...
+
+  double root_finder(const double& target_radius,
+                     const double& squared_function_of_time,
+                     const double& root_factor) const;
```
knelli2: These `double`s don't need to be references`. They can just be `double` here and then `const double` in the cpp.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -118,130 +86,74 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
 }
 
 template <size_t Dim>
-template <typename DataType>
-tnsr::Ij<tt::remove_cvref_wrap_t<DataType>, Dim, Frame::NoFrame>
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>
 Translation<Dim>::jacobian(
-    const std::array<DataType, Dim>& source_coords, const double time,
+    const std::array<T, Dim>& source_coords, double time,
```
knelli2: Keep the `const double` please

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -250,10 +162,10 @@ Translation<Dim>::jacobian(
   }
 }
 template <size_t Dim>
-template <typename DataType>
-tnsr::Ij<tt::remove_cvref_wrap_t<DataType>, Dim, Frame::NoFrame>
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>
 Translation<Dim>::inv_jacobian(
-    const std::array<DataType, Dim>& source_coords, const double time,
+    const std::array<T, Dim>& source_coords, double time,
```
knelli2: Keep the `const double` please

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -324,6 +287,9 @@ void Translation<Dim>::pup(PUP::er& p) {
   if (version >= 1) {
     p | f_of_r_;
     p | center_;
+  } else if (p.isUnpacking()) {
+    f_of_r_ = nullptr;
+    center_ = std::array<double, Dim>{0.};
```
knelli2:
```cpp
center_ = make_array<Dim, double>(0.0);
```

Only specifying one `0.` will actually only initialize the first element for Dim = 2 and 3. (Technically the rest are default initialized to `0.` as well, but better to be pedantic here)

Also, to use this
```cpp
#include "Utilities/MakeArray.hpp
```

src/PointwiseFunctions/MathFunctions/RegisterDerivedWithCharm.cpp
```diff
@@ -5,8 +5,14 @@
...
+#include "PointwiseFunctions/MathFunctions/PowX.hpp"
+#include "PointwiseFunctions/MathFunctions/Sinusoid.hpp"
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
 
+namespace MathFunctions {
 void register_derived_with_charm() {
-    register_classes_with_charm<MathFunctions::Gaussian<1, Frame::Inertial>>();
+  register_classes_with_charm<Gaussian<1, Frame::Inertial>>();
+  register_classes_with_charm<PowX<1, Frame::Inertial>>();
+  register_classes_with_charm<Sinusoid<1, Frame::Inertial>>();
```
knelli2: The `register_classes_with_charm` function can actually take a parameter pack so just do one function call with all the templates separated by `,`.

knelli2 (changes requested):
You can squash these and the previous changes together with the original commit. It'll make a final review easier.

Also about docs changes: Currently your docs don't render properly because of what I mentioned below. You can check the docs yourself before pushing by building `make doc`. Then inside `BUILD_DIR/docs/html/` are all the `.html` files for the webpage. I usually open `BUILD_DIR/docs/html/index.html` in my browser, which is the home page, and then navigate to the page I was editing and view things there.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
AlexCarpenter46: Oh yeah, this should be a dot product between the two vectors not the magnitudes being multiplied

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -11,28 +11,93 @@
...
+ * \f{equation}{
+ * r = |\vec{\xi} - \vec{c}|
+ * \f}
+ * where r is the radius and \f$\vec{c}\f$ is the center argument.
+ *
+ * ### Inverse Translation
+ * The inverse translation translates the coordinate \f$\vec{\bar{\xi}}\f$ to
+ * the original coordinates
+ * \f{equation}{
+ * \vec{\xi} = \vec{\bar{\xi}} - F(root)\vec{T}(t)
```
knelli2: `root` renders in $\LaTeX$ making it look weird. Either use a variable for it and say it's the root of `F`, or wrap it in `\mathrm{root}`.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -11,28 +11,93 @@
...
+ * \f}
+ * However, we need to do a root find to see where F(r) and \f$\vec{T}(t)\f$
```
knelli2: `\f$F(r)\f$`

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -11,28 +11,93 @@
...
+ * cross zero. This is done by bracketing and finding the root of
+ * \f{equation}{
+ * \bar{r}^2 + F(x) (A + |\vec{T}(t)| F(x)) - x^2
+ * \f}
+ * This equation comes from the SpEC implementation where x is the root guess
+ * and A is the Root Factor given by
```
knelli2: Both `x` and `A` inside of `\f$...\f$`. (Good rule of thumb, if it appears in an equation and is rendered by $\LaTeX$, it should look the same in paragraphs/sentences.)

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -11,28 +11,93 @@
...
+ * \f{equation}{
+ * A = -2.0 (\vec{\xi} - \vec{c}) \cdot \vec{T}(t)
+ * \f}
+ *
+ * ### Frame Velocity
+ *
+ * The frame velocity is given by
+ * \f{equation}{
+ * \vec{v} = \frac{\vec{dT}(t)}{dt} * F(r)
```
knelli2: No `*`

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -11,28 +11,93 @@
...
+ * \f}
+ * where \f$\frac{\vec{dT}(t)}{dt}\f$ is the first derivative of the
+ * FunctionOfTime.
+ *
+ * ### Jacobian
+ * The jacobian is computed through either first or second derivatives of the
+ * radial MathFunction based on your radius. When the radius is bigger
+ * than 1.e-13, the jacobian is given by:
+ * \f{equation}{
+ * J^{i}_{j} = \frac{dF(r)}{dr} * \vec{T}(t) * \frac{(\vec{\xi} - \vec{c})}{r}
```
knelli2: $\LaTeX$ is weird so for the `J^{i}_{j}` ( $J^{i}\_{j}$ ), it actually renders the `i` and `j` above each other, rather than how we normally write indices. To achieve that you have to do `{J^{i}}_{j}` ( ${J^{i}}_{j}$ ).

 Also no `*`. This notation is a bit confusing because you're mixing index notation with vector notation. I think it'd be clearer if you wrote this as 

${J^{i}}_{j} = \frac{dF(r)}{dr} T(t)^i \frac{(\xi_j - c_j)}{r}$

Same (on all changes) for the other jacobian below.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -82,56 +113,186 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
                                      "time. The known functions of time are: "
                                   << keys_of(functions_of_time));
-  std::array<tt::remove_cvref_wrap_t<T>, Dim> result{};
-  const auto function_of_time_and_deriv =
-      functions_of_time.at(f_of_t_name_)->func_and_deriv(time);
-  const DataVector& velocity = function_of_time_and_deriv[1];
-  ASSERT(velocity.size() == Dim,
+  const auto func_or_deriv_of_time =
+      functions_of_time.at(f_of_t_name_)
+          ->func_and_deriv(time)[function_or_deriv_index];
```
knelli2: `gsl::at()` since it's a runtime index

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -82,56 +113,186 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
          "The dimension of the function of time ("
-             << velocity.size()
+             << func_or_deriv_of_time.size()
              << ") does not match the dimension of the translation map (" << Dim
              << ").");
+  std::array<tt::remove_cvref_wrap_t<T>, Dim> result{};
   for (size_t i = 0; i < Dim; i++) {
-    gsl::at(result, i) = make_with_value<tt::remove_cvref_wrap_t<T>>(
-        dereference_wrapper(gsl::at(source_coords, i)), velocity[i]);
+    gsl::at(result, i) = gsl::at(source_coords, i);
```
knelli2, line 204: Use an `if (function_or_deriv_index == 0)` statement here. If it's true, set the result to `source_coords` like you do now. If it's false, set result to 0...

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -82,56 +113,186 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
...
+  // The option 0. specifies using the function_of_time and anything else
+  // specifies using the derivative of the function_of_time
+  if (function_or_deriv_index == 0) {
+    for (size_t i = 0; i < Dim; i++) {
+      gsl::at(result, i) += func_or_deriv_of_time[i] * radial_function_value;
+    }
+  } else {
+    for (size_t i = 0; i < Dim; i++) {
+      gsl::at(result, i) = func_or_deriv_of_time[i] * radial_function_value;
+    }
```
knelli2: ...then just always `+=` here. No need for the `if` check. (And remove the comment, I think that was for a previous version)

knelli2 (changes requested):
Just a few more things!

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
nilsvu: Note that you don't actually need the `\f`s. Standard LaTeX syntax works fine: `$x$`, `\begin{equation}`, etc

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: I changed a few things because if I don't set the result the source_coords then I can't use the magnitude function cause source_coords have the wrong typing. I tried to avoid any new allocations this way and this way it keeps the += at the end. I didn't know of a good way to properly size the std::array since it could be a datavector

knelli2 (approved):
Ok looks good! Great work!!

PR #5173, opened 2023-07-12
Stop multiplying by 0 in GH and copying data

nilsdeppe:
## Proposed changes

A few easy optimizations I noticed when profiling where we were doing memset and memcpy

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
```diff
@@ -56,7 +56,6 @@ void TimeDerivative<Dim>::apply(
         christoffel_first_kind_3_up,
     const gsl::not_null<Scalar<DataVector>*> lapse,
     const gsl::not_null<tnsr::I<DataVector, Dim>*> shift,
-    const gsl::not_null<tnsr::ii<DataVector, Dim>*> spatial_metric,
```
knelli2, line 59: Does it make more sense to just make the spatial metric a reference tag then? Do we ever set it to something other than the spatial part of the spacetime metric? If it's a reference tag then you can take it as an argument rather than `not_null`. Then you don't have to make a view every time this is used.

src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
```diff
@@ -300,17 +300,35 @@ void TimeDerivative<Dim>::apply(
   }
 
   // Equation for dt_pi
-  for (size_t mu = 0; mu < Dim + 1; ++mu) {
+
+  // WARNING: normal_dot_gauge_constraint is rescaled by gamma0!
+  get(*normal_dot_gauge_constraint) *= get(gamma0);
```
knelli2: Can you add a comment here explaining what you're doing? It looks like you're avoiding a multiplication of `normal_spacetime_one_form` for the relevant components because it's (1,0,0,0), but that's not immediately obvious.

tests/Unit/Evolution/Systems/GeneralizedHarmonic/BoundaryConditions/Bjorhus.py
```diff
@@ -14,30 +14,21 @@
```
knelli2: Can you format this file in a previous separate commit? I was shocked that removing a single arg form the c++ code could cause an almost 700 line change in a py file, until I realized it was largely unrelated formatting changes.

tests/Unit/Evolution/Systems/GeneralizedHarmonic/BoundaryConditions/Test_Bjorhus.cpp
```diff
@@ -295,9 +287,9 @@ SPECTRE_TEST_CASE("Unit.Evolution.Systems.GeneralizedHarmonic.BCBjorhus.Cls",
...
 
-  const DataVector used_for_size(3);
+  // const DataVector used_for_size(3);
 
-  test_with_random_values<1>(used_for_size);
-  test_with_random_values<2>(used_for_size);
-  test_with_random_values<3>(used_for_size);
+  // test_with_random_values<1>(used_for_size);
+  // test_with_random_values<2>(used_for_size);
+  // test_with_random_values<3>(used_for_size);
```
knelli2: Remove?

tests/Unit/PointwiseFunctions/GeneralRelativity/ComputeSpacetimeQuantities.py
```diff
@@ -5,22 +5,20 @@
```
knelli2: Lump this file into the formatting commit as well. Only one function was actually added here.

knelli2 (changes requested):
You can squash these directly.

tests/Unit/Evolution/Systems/GeneralizedHarmonic/BoundaryConditions/Test_Bjorhus.cpp
nilsdeppe: I should make sure they work :(

src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
nilsdeppe: I don't think a reference tag works because those point at an entire different object, whereas this points to a subset. I _think_ compute tags can set the `DataVector`s into the spacetime metric, but adding the argument tags in the places is not something I can reasonably prioritize right now. I do agree that it's worth trying to do what you suggested in the long term, though!

nilsdeppe:
Okay, rebased and made the changes. I squashed the re-enabling of the tests into the last commit

src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
```diff
@@ -301,6 +301,13 @@ void TimeDerivative<Dim>::apply(
 
   // Equation for dt_pi
 
+  // We first compute the n_a contributes but only for a=0 since n_i=0
```
knelli2: `contributions`

src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
```diff
@@ -301,6 +301,13 @@ void TimeDerivative<Dim>::apply(
...
+  // identically. We also use dt_Pi_{00} as temporary storage to avoid any extra
+  // allocations and multiple normal_dot_gauge_constraint=(n^a C_a) by gamma0
```
knelli2: `multiply`

knelli2 (commented):
Just some typos. You can squash it all together

knelli2 (dismissed):

nilsdeppe:
I'll see if I can figure out the python formatting... I think I was still using yapf instead of black. I'll ping you once I got that sorted out :)

nilsdeppe:
Okay, python formatting seems to be sorted out...

knelli2 (approved):

PR #5184, opened 2023-07-13
Add maps to ExcisionSpheres

### LINKED ISSUE #5180: Add coord maps to excision sphere
Helpful to compute coord separation between horizons, for worldtube things, etc. Excision spheres cover part of the grid similar to blocks but without elements, so adding coord maps is reasonable.

knelli2:
## Proposed changes

Discussion of why we want this in #5150. Fixes #5180.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


tests/Unit/Domain/Test_ExcisionSphere.cpp
```diff
@@ -149,6 +154,35 @@ void test_error() {
...
+                    Catch::Contains("Trying to access grid to inertial map "
+                                    "from ExcisionSphere, but it has "
+                                    "not been set."));
+
+  excision_sphere.inject_time_dependent_maps(time_dep_map->get_clone());
+  test_serialization(excision_sphere);
+  CHECK(excision_sphere.is_time_dependent());
+  CHECK(excision_sphere.moving_mesh_grid_to_inertial_map() == *time_dep_map);
+
+  const ExcisionSphere<3> copy_assigned_excision_sphere = excision_sphere;
```
wthrowe: That's not a copy assignment.  You need to assign to a preexisting object.

wthrowe (changes requested):

tests/Unit/Domain/Test_ExcisionSphere.cpp
knelli2: Oops 😅

tests/Unit/Domain/Test_ExcisionSphere.cpp
```diff
@@ -149,6 +154,35 @@ void test_error() {
...
+                                    "not been set."));
+
+  excision_sphere.inject_time_dependent_maps(time_dep_map->get_clone());
+  test_serialization(excision_sphere);
+  CHECK(excision_sphere.is_time_dependent());
+  CHECK(excision_sphere.moving_mesh_grid_to_inertial_map() == *time_dep_map);
+
+  ExcisionSphere<3> copied_excision_sphere{excision_sphere};
+  CHECK(copied_excision_sphere == excision_sphere);
+  copied_excision_sphere = excision_sphere;
```
wthrowe: Don't start with the correct answer.  The usual way to screw these up is to forget to assign a field, and this won't catch that.  Just default-construct one and then assign to it.

wthrowe (changes requested):

tests/Unit/Domain/Test_ExcisionSphere.cpp
knelli2: Third times a charm

src/Domain/ExcisionSphere.cpp
```diff
@@ -24,8 +24,9 @@ ExcisionSphere<VolumeDim>::ExcisionSphere(
     : radius_(radius),
       center_(center),
       abutting_directions_(std::move(abutting_directions)) {
-  ASSERT(radius_ > 0.0,
-         "The ExcisionSphere must have a radius greater than zero.");
+  if (radius_ <= 0.0) {
+    ERROR("The ExcisionSphere must have a radius greater than zero.");
```
nilsvu: Can you hit this by making bad decisions in an input file? That's when I use `ERROR`. For things that are programming errors use `ASSERT`.

src/Domain/ExcisionSphere.cpp
```diff
@@ -29,6 +32,56 @@ ExcisionSphere<VolumeDim>::ExcisionSphere(
...
+ExcisionSphere<VolumeDim>::ExcisionSphere(
+    const ExcisionSphere<VolumeDim>& rhs) {
+  radius_ = rhs.radius_;
+  center_ = rhs.center_;
+  abutting_directions_ = rhs.abutting_directions_;
+  if (rhs.grid_to_inertial_map_ == nullptr) {
+    grid_to_inertial_map_ = nullptr;
+  } else {
+    grid_to_inertial_map_ = rhs.grid_to_inertial_map_->get_clone();
+  }
```
nilsvu:
```cpp
  *this = rhs;
```

src/Domain/ExcisionSphere.cpp
```diff
@@ -29,6 +32,56 @@ ExcisionSphere<VolumeDim>::ExcisionSphere(
...
+}
+
+template <size_t VolumeDim>
+const domain::CoordinateMapBase<Frame::Grid, Frame::Inertial, VolumeDim>&
+ExcisionSphere<VolumeDim>::moving_mesh_grid_to_inertial_map() const {
+  // This is an ERROR rather than an ASSERT because we don't expect to be
+  // calling this as often as the coordinate maps in the Blocks, so always
+  // checking shouldn't be costly (and we wrap in an unlikely for good measure).
+  if (UNLIKELY(grid_to_inertial_map_ == nullptr)) {
+    ERROR(
```
nilsvu: same comment about user vs programmer error

src/Domain/ExcisionSphere.cpp
```diff
@@ -60,13 +113,29 @@ void ExcisionSphere<VolumeDim>::pup(PUP::er& p) {
...
 
 template <size_t VolumeDim>
 bool operator==(const ExcisionSphere<VolumeDim>& lhs,
                 const ExcisionSphere<VolumeDim>& rhs) {
-  return lhs.radius() == rhs.radius() and lhs.center() == rhs.center() and
-         lhs.abutting_directions() == rhs.abutting_directions();
+  // First check time dep maps. If both are time dep, safe to dereference
+  // unique_ptr
+  if ((lhs.is_time_dependent() and not rhs.is_time_dependent()) or
+      ((not lhs.is_time_dependent()) and rhs.is_time_dependent())) {
```
nilsvu:
```cpp
  if (lhs.is_time_dependent() != rhs.is_time_depentend()) {
```

src/Domain/ExcisionSphere.hpp
```diff
@@ -46,17 +53,37 @@ class ExcisionSphere {
...
+  /// should be able to take a point inside the excision sphere and map it to
+  /// the Inertial frame.
+  void inject_time_dependent_maps(
+      std::unique_ptr<
+          domain::CoordinateMapBase<Frame::Grid, Frame::Inertial, VolumeDim>>
+          moving_mesh_grid_to_inertial_map);
+
+  /// \brief The map going from the last time independent frame to the frame in
+  /// which the equations are solved. Only used when the coordinate map is
+  /// time-dependent.
```
nilsvu: "Only use when `is_time_dependent()` is true"

nilsvu (changes requested):

tests/Unit/Domain/Test_ExcisionSphere.cpp
nilsvu: Can you just use `test_copy_semantics`?

src/Domain/ExcisionSphere.cpp
knelli2: That's dependent on how the domain creator uses this. I don't see an issue with an extra check in the constructor. And this object doesn't make sense with a negative radius so why should we rely on external code to ensure a correct radius is passed in?

src/Domain/ExcisionSphere.cpp
knelli2: I've messed this up when coding things into the domain creator and then an ASSERT wasn't hit in release mode and I got a segfault/UB. If this was going to be called super often like in the RHS, I'd make it an ASSERT, but this will only be called on events so it shouldn't be slow. And for that reason, I'd always want the extra protection.

nilsvu: Just test things in Debug mode, that's what it's for :) And that's why the distinction between ERROR and ASSERT exist.

knelli2:
Rebased, squashed in the changes into the first 3 commits (because they were small and there weren't that many), and added a new commit that uses ExcisionSphere maps in the SeparationLessThan trigger

src/Evolution/Triggers/SeparationLessThan.cpp
```diff
@@ -13,52 +13,53 @@
...
-        grid_to_inertial_map(object_center_b, time, functions_of_time);
-  }
+    const tnsr::I<double, 3, Frame::Grid>& grid_object_center_a,
+    const tnsr::I<double, 3, Frame::Grid>& grid_object_center_b) const {
+  const std::unordered_map<std::string, ExcisionSphere<3>>& excision_spheres =
+      domain.excision_spheres();
+
+  const auto check_excision_sphere =
+      [&excision_spheres]([[maybe_unused]] const std::string& object) {
+        ASSERT(excision_spheres.count("ExcisionSphere" + object) == 1,
```
nilsvu: So.. this is ironic but these should be `ERROR`s I think 😅 Because they're caused by choices in the input file (choosing a domain that doesn't have two excisions, or that don't have time dependences). Ideally the check would happen at option-construction, but I don't see a great way of doing that ATM.

nilsvu (changes requested):

src/Evolution/Triggers/SeparationLessThan.cpp
knelli2: 🤦🏻

nilsvu (dismissed):

src/Evolution/Triggers/SeparationLessThan.cpp
```diff
@@ -13,52 +13,56 @@
...
-    inertial_object_center_b =
-        grid_to_inertial_map(object_center_b, time, functions_of_time);
-  }
+    const tnsr::I<double, 3, Frame::Grid>& grid_object_center_a,
+    const tnsr::I<double, 3, Frame::Grid>& grid_object_center_b) const {
+  const std::unordered_map<std::string, ExcisionSphere<3>>& excision_spheres =
+      domain.excision_spheres();
+
+  const auto check_excision_sphere =
+      [&excision_spheres]([[maybe_unused]] const std::string& object) {
```
wthrowe: Drop "maybe_unused".  It's always used.

wthrowe (changes requested):
Squash, including this.

knelli2:
Squashed and rebased

wthrowe (approved):

nilsvu (approved):

PR #5185, opened 2023-07-13
Add Shell0 block group to Sphere for single shell

knelli2:
## Proposed changes

No clue why this wasn't included before. It makes everything consistent.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/Sphere.cpp
+++ b/src/Domain/Creators/Sphere.cpp
@@ -170,9 +170,7 @@ Sphere::Sphere(
       const std::string wedge_name =
           shell_prefix + gsl::at(wedge_directions, direction);
       block_names_.emplace_back(wedge_name);
-      if (num_shells_ > 1) {
-        block_groups_[shell_prefix].insert(wedge_name);
-      }
+      block_groups_[shell_prefix].insert(wedge_name);
       block_groups_["Wedges"].insert(wedge_name);
     }
   }


src/Domain/Creators/Sphere.cpp
```diff
@@ -170,9 +170,7 @@ Sphere::Sphere(
       const std::string wedge_name =
           shell_prefix + gsl::at(wedge_directions, direction);
       block_names_.emplace_back(wedge_name);
-      if (num_shells_ > 1) {
```
nilsvu, line 173: FYI, this wasn't included because `Shell0` doesn't really make sense if there's only one shell. The group is identical to the `Wedges` group (which holds all shells). My weak preference is to leave this as-is, or name the group `Shell` (without the index).

knelli2: I actually prefer to have it `Shell0` because in my input file I had many shells:

```yaml
DomainCreator:
  Sphere:
    ...
    InitialRefinement:
      Shell0: [2, 2, 2]
      Shell1: [1, 1, 1]
      Shell2: [0, 0, 0]
    RadialPartitioning: [10, 20]
```

then I wanted to only have one shell so I did:

```yaml
DomainCreator:
  Sphere:
    ...
    InitialRefinement:
      Shell0: [2, 2, 2]
    RadialPartitioning: []
```

and was surprised that didn't just work. I don't really want to have to remember a special name for the single shell case.

nilsvu: ok fine with me

nilsvu (approved):

PR #5190, opened 2023-07-14
Make BBH time dependent options optional

knelli2:
## Proposed changes

This also now enforces a condition that in order to have a Shape map from the grid to distorted frame, you must have an excision. This was still technically true before because of how things were coded, but now an error will occur if you specify things incorrectly.

### Upgrade instructions

First, specifying options for shape/size maps has changed from

```yaml
DomainCreator:
  BinaryCompactObject:
    TimeDependentMaps:
      ...
      SizeMapA:
        InitialValues: [0.0, 0.0, 0.0]
      SizeMapB:
        InitialValues: [0.0, 0.0, 0.0]
      ShapeMapA:
        LMax: &LMax 10
      ShapeMapB:
        LMax: *LMax
```

to

```yaml
DomainCreator:
  BinaryCompactObject:
    TimeDependentMaps:
      ...
      ShapeMapA:
        LMax: &LMax 10
        SizeInitialValues: [0.0, 0.0, 0.0]
      ShapeMapB:
        LMax: *LMax
        SizeInitialValues: [0.0, 0.0, 0.0]
```

Then, to not have a time dependent map be included, specify `None` for its option like so:

```yaml
DomainCreator:
  BinaryCompactObject:
    TimeDependentMaps:
      InitialTime: 0.0
      ExpansionMap: None
      RotationMap:
        InitialAngularVelocity: [0.0, 0.0, 0.0]
      ShapeMapA:
        LMax: &LMax 10
        SizeInitialValues: [0.0, 0.0, 0.0]
      ShapeMapB: None
```

To disable time dependent maps all together do

```yaml
DomainCreator:
  BinaryCompactObject:
    TimeDependentMaps: None
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
```diff
@@ -529,21 +536,22 @@ void test_bns_domain_with_cubes() {
...
-          create_option_string(true, true, true, false, true, 0, 0, 0, 120.0,
-                               with_boundary_conditions));
+          create_option_string(true, true, true, with_time_dependence, false,
+                               true, 0, 0, 0, 120.0, with_boundary_conditions));
     } else {
       return TestHelpers::test_option_tag<domain::OptionTags::DomainCreator<3>,
                                           Metavariables<3, true, false>>(
-          create_option_string(true, true, true, false, true, 0, 0, 0, 120.0,
-                               with_boundary_conditions));
+          create_option_string(true, true, true, with_time_dependence, false,
```
nilsvu: Is `add_time_dependence` set to `true` somewhere?

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -42,16 +42,18 @@ create_grid_anchors(const std::array<double, 3>& center_a,
 
 TimeDependentMapOptions::TimeDependentMapOptions(
     double initial_time, ExpansionMapOptions expansion_map_options,
-    std::array<double, 3> initial_angular_velocity,
-    std::array<double, 3> initial_size_values_A,
-    std::array<double, 3> initial_size_values_B, const size_t initial_l_max_A,
-    const size_t initial_l_max_B, const Options::Context& context)
+    RotationMapOptions rotation_options,
```
nilsvu: You can take these as const-refs because you only access and copy their members

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -157,26 +159,27 @@ TimeDependentMapOptions::create_functions_of_time(
 
 void TimeDependentMapOptions::build_maps(
     const std::array<std::array<double, 3>, 2>& centers,
-    const std::array<std::optional<double>, 2>& object_inner_radii,
-    const std::array<std::optional<double>, 2>& object_outer_radii,
+    const std::optional<std::pair<double, double>>& object_A_inner_outer_radii,
+    const std::optional<std::pair<double, double>>& object_B_inner_outer_radii,
```
nilsvu, line 181: (optional) `std::pair<double, double>` -> `std::array<double, 2>`?

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -154,30 +154,47 @@ struct TimeDependentMapOptions {
     static constexpr Options::String help = {"Options for CubicScale map."};
   };
 
-  struct RotationMap {
+  struct RotationMapOptions {
```
nilsvu, line 228: If you want to be a little hacky and free up some names maybe move `using type = RotationMapOptions` in here and use in the list of `Options` directly? Then you can delete `struct RotationMap` below. Same with shape, size below. Trims the code a bit.

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -63,6 +65,84 @@ std::unordered_map<std::string, tnsr::I<double, 3, Frame::Grid>>
...
+      tmpl::bind<tmpl::remove, tmpl::pin<list_of_maps>, tmpl::_1>>;
+  using type = tmpl::push_back<
+      tmpl::remove_duplicates<tmpl::flatten<tmpl::transform<
+          sub_lists, produce_all_maps<tmpl::pin<SourceFrame>,
+                                      tmpl::pin<TargetFrame>, tmpl::_1>>>>,
+      domain::CoordinateMap<SourceFrame, TargetFrame, Maps...>>;
+};
+
+template <typename SourceFrame, typename TargetFrame, typename... Maps>
+using produce_all_maps_t = typename produce_all_maps<SourceFrame, TargetFrame,
```
nilsvu: Please test this

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -162,34 +181,82 @@ void TimeDependentMapOptions::build_maps(
...
   }
 }
 
+#if defined(__GNUC__) && !defined(__clang__)
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wsuggest-attribute=noreturn"
+#endif  // defined(__GNUC__) && !defined(__clang__)
+template <domain::ObjectLabel Object>
+bool TimeDependentMapOptions::has_distorted_frame_options() const {
+  ASSERT(Object == domain::ObjectLabel::A or Object == domain::ObjectLabel::B,
```
wthrowe: Can this be a static_assert?  I assume from the warning pragmas that this is compiled for other values, but is that just from the explicit instantiation or is it needed for something?

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -94,64 +101,76 @@ TimeDependentMapOptions::create_functions_of_time(
...
+        return std::array<DataVector, 4>{
+            {{gsl::at(lambda_options.value().initial_size_values, 0)},
+             {gsl::at(lambda_options.value().initial_size_values, 1)},
+             {gsl::at(lambda_options.value().initial_size_values, 2)},
+             {0.0}}};
+      };
+
+      size_t initial_l_max = 0;
+      initial_l_max = i == 0 ? shape_options_A_.value().l_max
+                             : shape_options_B_.value().l_max;
```
nilsvu: one line and const?

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -162,34 +181,82 @@ void TimeDependentMapOptions::build_maps(
...
                           ShapeMapTransitionFunction>
           transition_func = std::make_unique<
               domain::CoordinateMaps::ShapeMapTransitionFunctions::
                   SphereTransition>(inner_outer_radii.value().first,
                                     inner_outer_radii.value().second);
 
+      size_t initial_l_max = 0;
+      initial_l_max = i == 0 ? shape_options_A_.value().l_max
+                             : shape_options_B_.value().l_max;
+
```
nilsvu: one line and const

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -41,30 +41,37 @@ create_grid_anchors(const std::array<double, 3>& center_a,
...
+      rotation_options_(rotation_options),
+      shape_options_A_(shape_options_A),
+      shape_options_B_(shape_options_B) {
+  if (not(expansion_map_options_.has_value() or rotation_options_.has_value() or
+          shape_options_A_.has_value() or shape_options_B_.has_value())) {
+    PARSE_ERROR(context,
+                "Time dependent map options were specified, but all options "
+                "were 'None'. If you don't want time dependent maps, specify "
+                "'None' for the TimeDependentMapOptions. If you want time "
+                "dependent maps, specify options for at least one map.");
```
wthrowe, line 61: Is there a reason to forbid this?  It seems like it might be easier to experiment if you could toggle all the maps independently.

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -63,6 +65,84 @@ std::unordered_map<std::string, tnsr::I<double, 3, Frame::Grid>>
...
+ *  - Map1
+ *  - Map2
+ *  - Map3
+ *  - Map1, Map2
+ *  - Map1, Map3
+ *  - Map2, Map3
+ *  - Map1, Map2, Map3
+ */
+template <typename SourceFrame, typename TargetFrame, typename... Maps>
+struct produce_all_maps<SourceFrame, TargetFrame, tmpl::list<Maps...>> {
```
wthrowe: I gave you a better algorithm for this elsewhere.  Use that instead.

nilsvu (changes requested):

nilsvu:
Just some minor things

wthrowe (changes requested):

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
nilsvu: You can also make this runtime if that helps (no reason to be compile-time)

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
knelli2: If you set `TimeDependentMapOptions: None`, then time dependent maps are never actually injected into the domain, and the domain is static (`stationary_map` in the block is valida and `block.is_time_dependent()` is false). If you specify all maps as `None` instead (and I remove several errors that are currently in place and add in some identity maps), then you'd have a domain that is in practice time-independent, but the `stationary_map` in the block is nullptr and `block.is_time_dependent()` is true, causing the code to go through all the moving mesh stuff. Even though they'd be identities, I decided this wasn't ideal and to just not allow all individual maps to be `None`.

tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
knelli2: Yes, below in the `SPECTRE_TEST_CASE`. The previous argument was actually never used which is why I made this change.

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
knelli2: I technically can, however, I had this little intermediate commit to just standardize the constructor so all options were taken the same to make the change to `std::optional`s easier. If it's ok with you, I'd like to just leave it to avoid a complicated rebase with conflicts because I make this change and then immediately change it again the next commit.

nilsvu: Yea don't worry about the commit history here, an extra commit is fine too

knelli2:
Rebased and pushed a fixup.

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
knelli2: I still wasn't able to get yours to work, so until I do I just left mine.

wthrowe:
```c++
template <typename SourceFrame, typename TargetFrame, typename Maps>
using produce_all_maps_helper =
    tmpl::wrap<tmpl::push_front<Maps, SourceFrame, TargetFrame>,
               domain::CoordinateMap>;
template <typename SourceFrame, typename TargetFrame, typename... Maps>
using produce_all_maps_t = tmpl::transform<
    tmpl::remove<typename power_set<tmpl::list<Maps...>>::type, tmpl::list<>>,
    tmpl::bind<produce_all_maps_helper, tmpl::pin<SourceFrame>,
               tmpl::pin<TargetFrame>, tmpl::_1>>;
```
The parameter pack makes it a little tricky.

The "_t" name isn't appropriate anymore, of course.

knelli2: Ok yup this worked!! Pasting the `power_set` implementation you posted in slack here for completeness:

```cpp
template <typename List>
struct power_set {
  using rest = typename power_set<tmpl::pop_front<List>>::type;
  using type = tmpl::append<
      rest, tmpl::transform<rest, tmpl::lazy::push_front<
                                      tmpl::_1, tmpl::pin<tmpl::front<List>>>>>;
};

template <>
struct power_set<tmpl::list<>> {
  using type = tmpl::list<tmpl::list<>>;
};
```

nilsvu (approved):
Just the const thing is unresolved (see above). Please squash.

knelli2:
@wthrowe does the latest fixup look ok? @AlexCarpenter46 is going to be adding the translation map (from #5166) to the BBH and this should be merged first since it's a change to the BCO time dependent options.

wthrowe:
Looks fine, but what is "the const thing" you're still planning on changing?  I don't see any comments about constness above.

knelli2:
This: https://github.com/sxs-collaboration/spectre/pull/5190#discussion_r1264192134

wthrowe:
Ah, OK.  That's good.

knelli2:
@wthrowe @nilsvu Squashed, rebased, and made the const change thing. Should be good now

wthrowe (approved):

nilsvu (approved):

PR #5193, opened 2023-07-14
Validate CLI: fall back to printing full error message

### LINKED ISSUE #5192: Validate CLI doesn't print unknown error
I somehow managed to get an FPE *during* option parsing (don't ask me how, I've already fixed it). However, I tried to run `spectre validate` over this and got the following output.

```
knelli@wheeler:~$ spectre validate BinaryBlackHole.yaml 
InvalidInputFileError
```

And that was it. It'd be nice if the fallback was to just print whatever our `--check-options` originally output, even if it wasn't fancy and colorful.

nilsvu:
## Proposed changes

Closes #5192.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/ValidateInputFile.py
+++ b/tools/ValidateInputFile.py
@@ -134,7 +134,9 @@ def validate_input_file(
         input_file_path=input_file_path,
         line_number=line_number,
         yaml_path=path,
-        message="\n".join(msg).strip(),
+        message=(
+            "\n".join(msg).strip() if found_hints else process.stderr.strip()
+        ),
     )
     if print_context:
         rich.print(error.render_context())


knelli2 (approved):

tools/ValidateInputFile.py
```diff
@@ -134,7 +134,9 @@ def validate_input_file(
         input_file_path=input_file_path,
         line_number=line_number,
         yaml_path=path,
-        message="\n".join(msg).strip(),
+        message=(
+            "\n".join(msg).strip() if found_hints else process.stderr.strip()
+        ),
```
knelli2, line 139: Actually wait, will this print anything? The `render_context` in `InvalidInputFileError` is empty without a line number and yaml path (neither of which exist if `found_hints` is false)

knelli2 (changes requested):

tools/ValidateInputFile.py
nilsvu: The exception message will print. `render_context` is just extra stuff.

knelli2: Ok cool. I can't test this now because I fixed the issue in the code that was causing an FPE during option parsing, but hopefully next time I hit something this will show up.

knelli2 (approved):

tools/ValidateInputFile.py
nilsvu: This also happens when you put a wrong `Version` at the top of the file

PR #5198, opened 2023-07-17
Make PiecewisePolynomial threadsafe

### LINKED ISSUE #5181: Make PiecewisePolynomial thread-safe
Currently it isn't because we store a `std::vector` of times, and access that vector. Idea is to change to a `std::deque` with a stored `int` for the current size. Also, locking on updating would be good.

knelli2:
And also QuaternionFunctionOfTime

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.hpp
+++ b/src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.hpp
@@ -7,6 +7,7 @@
 #pragma once
 
 #include <array>
+#include <deque>
 #include <limits>
 #include <ostream>
 #include <pup.h>
@@ -58,15 +59,24 @@ struct StoredInfo<MaxDerivPlusOne, false> {
 void reset_expiration_time(const gsl::not_null<double*> prev_expiration_time,
                            const double next_expiration_time);
 
-/// Returns a StoredInfo corresponding to the closest element in the range of
-/// `StoredInfo.time`s that is less than `t`. The function throws an
-/// error if `t` is less than all `StoredInfo.time`s (unless `t` is just less
-/// than the earliest `StoredInfo.time` by roundoff, in which case it returns
-/// the earliest StoredInfo.)
+/*!
+ * \brief Returns a StoredInfo corresponding to the closest element in the range
+ * of `StoredInfo.time`s that is less than `t`.
+ *
+ * The function throws an error if `t` is less than all `StoredInfo.time`s
+ * (unless `t` is just less than the earliest `StoredInfo.time` by roundoff, in
+ * which case it returns the earliest StoredInfo.)
+ *
+ * This entire operation needs to be threadsafe so an additional parameter
+ * `all_info_size` is passed which is the size of `all_stored_infos` that's
+ * allowed to be used in this operation. This avoids threading undefined
+ * behavior.
+ */
 template <size_t MaxDerivPlusOne, bool StoreCoefs>
 const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
-    const double t, const std::vector<StoredInfo<MaxDerivPlusOne, StoreCoefs>>&
-                        all_stored_infos);
+    const double t,
+    const std::deque<StoredInfo<MaxDerivPlusOne, StoreCoefs>>& all_stored_infos,
+    const size_t all_info_size);
 
 template <size_t MaxDerivPlusOne, bool StoreCoefs>
 bool operator==(
--- a/src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
+++ b/src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
@@ -4,12 +4,15 @@
 #include "Domain/FunctionsOfTime/PiecewisePolynomial.hpp"
 
 #include <algorithm>
+#include <atomic>
+#include <deque>
 #include <iterator>
 #include <memory>
 #include <ostream>
 #include <pup.h>
 #include <pup_stl.h>
 #include <utility>  // IWYU pragma: keep
+#include <vector>
 
 #include "DataStructures/DataVector.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
@@ -17,6 +20,7 @@
 #include "Utilities/Gsl.hpp"
 #include "Utilities/Literals.hpp"
 #include "Utilities/MakeArray.hpp"
+#include "Utilities/Serialization/PupStlCpp11.hpp"
 
 namespace domain::FunctionsOfTime {
 template <size_t MaxDeriv>
@@ -24,7 +28,28 @@ PiecewisePolynomial<MaxDeriv>::PiecewisePolynomial(
     const double t, value_type initial_func_and_derivs,
     const double expiration_time)
     : deriv_info_at_update_times_{{t, std::move(initial_func_and_derivs)}},
-      expiration_time_(expiration_time) {}
+      expiration_time_(expiration_time) {
+  deriv_info_size_.store(1);
+}
+
+template <size_t MaxDeriv>
+PiecewisePolynomial<MaxDeriv>::PiecewisePolynomial(
+    const PiecewisePolynomial<MaxDeriv>& rhs) {
+  *this = rhs;
+}
+
+template <size_t MaxDeriv>
+// NOLINTNEXTLINE(bugprone-unhandled-self-assignment)
+PiecewisePolynomial<MaxDeriv>& PiecewisePolynomial<MaxDeriv>::operator=(
+    const PiecewisePolynomial<MaxDeriv>& rhs) {
+  if (this == &rhs) {
+    return *this;
+  }
+  deriv_info_at_update_times_ = rhs.deriv_info_at_update_times_;
+  expiration_time_ = rhs.expiration_time_;
+  deriv_info_size_.store(rhs.deriv_info_size_.load(std::memory_order_relaxed));
+  return *this;
+}
 
 template <size_t MaxDeriv>
 std::unique_ptr<FunctionOfTime> PiecewisePolynomial<MaxDeriv>::get_clone()
@@ -42,8 +67,9 @@ PiecewisePolynomial<MaxDeriv>::func_and_derivs(const double t) const {
           << ". The difference between times is " << t - expiration_time_
           << ".");
   }
-  const auto& deriv_info_at_t =
-      stored_info_from_upper_bound(t, deriv_info_at_update_times_);
+  const auto& deriv_info_at_t = stored_info_from_upper_bound(
+      t, deriv_info_at_update_times_,
+      deriv_info_size_.load(std::memory_order_relaxed));
   const double dt = t - deriv_info_at_t.time;
   const value_type& coefs = deriv_info_at_t.stored_quantities;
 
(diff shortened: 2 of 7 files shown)


src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.hpp
```diff
@@ -58,15 +59,28 @@ struct StoredInfo<MaxDerivPlusOne, false> {
...
+ *
+ * The function throws an error if `t` is less than all `StoredInfo.time`s
+ * (unless `t` is just less than the earliest `StoredInfo.time` by roundoff, in
+ * which case it returns the earliest StoredInfo.)
+ *
+ * This entire operation needs to be threadsafe so an additional parameter
+ * `all_info_size` is passed which is the size of `all_stored_infos` that's
+ * allowed to be used in this operator. This is because another thread may be
+ * adding a new element to `all_stored_infos` as this function is called and we
+ * need to guarantee this always returns the same result.
```
wthrowe: I don't think this is why the parameter is needed.  The correct result had better be independent of what any other thread is doing.  This is purely to avoid threading UB.

src/Domain/FunctionsOfTime/PiecewisePolynomial.hpp
```diff
@@ -108,9 +108,10 @@ class PiecewisePolynomial : public FunctionOfTime {
   // the values of that deriv order for all components.
   using value_type = std::array<DataVector, MaxDeriv + 1>;
 
-  std::vector<FunctionOfTimeHelpers::StoredInfo<MaxDeriv + 1>>
+  std::deque<FunctionOfTimeHelpers::StoredInfo<MaxDeriv + 1>>
       deriv_info_at_update_times_;
   double expiration_time_{std::numeric_limits<double>::lowest()};
+  size_t deriv_info_size_{};
```
wthrowe: Does this have to be atomic?

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.hpp
```diff
@@ -152,8 +152,9 @@ class QuaternionFunctionOfTime : public FunctionOfTime {
       std::ostream& os,
       const QuaternionFunctionOfTime<LocalMaxDeriv>& quaternion_f_of_t);
 
-  std::vector<FunctionOfTimeHelpers::StoredInfo<1, false>>
+  std::deque<FunctionOfTimeHelpers::StoredInfo<1, false>>
       stored_quaternions_and_times_;
+  size_t stored_quaternion_size_{};
```
wthrowe: Again, should this be atomic?

wthrowe (changes requested):

src/Domain/FunctionsOfTime/PiecewisePolynomial.hpp
knelli2: Hmmm yes probably. If I make it atomic, do I need to add in any locks or is that taken care of? Also, in the `stored_info_from_upper_bound` function, do I need to take a reference to the `size_t` then? Or is it alright to have it copy?

nilsdeppe: I believe so, yes. It also needs to be 64-byte aligned and padded to 64 bytes to make sure there's no false sharing going on.
```cpp
alignas(64) std::atomic<uint64_t> deriv_info_size_{};
char[64 - sizeof(std::atomic<uint64_t>)];
```
When reading the value we should use `deriv_info_size_.load(std::memory_order::relaxed)` and when updating we should use `deriv_info_size_.add(1, std::memory_order::memory_order_acq_rel)`

knelli2: What does
```cpp
char[64 - sizeof(std::atomic<uint64_t>)];
```
do specifically? It pads the atomic int? Does doing `alignas(64)` not guarantee it'll be 64 bytes?

nilsdeppe: the int is 64-bytes aligned, but if it doesn't end on a 64-byte boundary then you can get false sharing on the next cache line. That is, if the `atomic<int>` is 80 bytes then you want to align it at 64 bytes add pad it out to 128 bytes so the int is all in its own cache lines

knelli2:
Ok added a fixup making the size an atomic. I didn't know what to initialize the padding to because it's never used so I just didn't initialize it? Hopefully that's ok? I also didn't `pup` it because we never use it, it's just garbage. However, I could be overlooking a reason to initialize it to something like `0000...`. Thoughts?

src/Domain/FunctionsOfTime/PiecewisePolynomial.hpp
```diff
@@ -108,9 +109,11 @@ class PiecewisePolynomial : public FunctionOfTime {
   // the values of that deriv order for all components.
   using value_type = std::array<DataVector, MaxDeriv + 1>;
 
-  std::vector<FunctionOfTimeHelpers::StoredInfo<MaxDeriv + 1>>
+  std::deque<FunctionOfTimeHelpers::StoredInfo<MaxDeriv + 1>>
       deriv_info_at_update_times_;
   double expiration_time_{std::numeric_limits<double>::lowest()};
+  alignas(64) std::atomic_uint64_t deriv_info_size_{};
+  char unused_padding_[64 - sizeof(std::atomic_uint64_t)];
```
kidder: you need to make sure the size of `unused_padding_` is positive (otherwise you will get a lot of padding 😛 )

nilsdeppe:
> Ok added a fixup making the size an atomic. I didn't know what to initialize the padding to because it's never used so I just didn't initialize it? Hopefully that's ok? I also didn't `pup` it because we never use it, it's just garbage. However, I could be overlooking a reason to initialize it to something like `0000...`. Thoughts?

I don't think it needs to be initialized :shrug:

src/Domain/FunctionsOfTime/PiecewisePolynomial.hpp
nilsdeppe: Yea, I think `sizeof(std::atomic_uint64_t)` -> `sizeof(std::atomic_uint64_t) % 64` does it

knelli2:
Ok updated the fixup with `%` and a few other things because apple was unhappy (I actually did something wrong and it caught it so that was nice).

src/Domain/FunctionsOfTime/PiecewisePolynomial.hpp
```diff
@@ -108,9 +109,12 @@ class PiecewisePolynomial : public FunctionOfTime {
   // the values of that deriv order for all components.
   using value_type = std::array<DataVector, MaxDeriv + 1>;
 
-  std::vector<FunctionOfTimeHelpers::StoredInfo<MaxDeriv + 1>>
+  std::deque<FunctionOfTimeHelpers::StoredInfo<MaxDeriv + 1>>
       deriv_info_at_update_times_;
   double expiration_time_{std::numeric_limits<double>::lowest()};
+  alignas(64) std::atomic_uint64_t deriv_info_size_{};
+  // Pad memory to avoid false-sharing when accessing deriv_info_size_
+  char unused_padding_[64 - (sizeof(std::atomic_uint64_t) % 64)];
```
nilsdeppe: Can you do `= {}` to get rid of the nolints?

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -17,16 +19,39 @@
...
+// NOLINTNEXTLINE(cppcoreguidelines-pro-type-member-init)
+template <size_t MaxDeriv>
+QuaternionFunctionOfTime<MaxDeriv>::QuaternionFunctionOfTime(
+    const QuaternionFunctionOfTime<MaxDeriv>& rhs) {
+  *this = rhs;
+}
+
+template <size_t MaxDeriv>
+QuaternionFunctionOfTime<MaxDeriv>&
+// NOLINTNEXTLINE(bugprone-unhandled-self-assignment)
```
nilsdeppe, line 44: You should do what this suggests :)
```cpp
if (this == &rhs) {
  return *this;
}
```
at the start of the copy

nilsdeppe (changes requested):
LGTM, couple minor suggestions, I think you can squash. The PR is small enough :)

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
knelli2: So I did this but still got the clang-tidy error, so I left the NOLINT in. Not really sure why I still get the error because I do what [clang](https://clang.llvm.org/extra/clang-tidy/checks/bugprone/unhandled-self-assignment.html) says to do.

src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.hpp
```diff
@@ -58,15 +59,27 @@ struct StoredInfo<MaxDerivPlusOne, false> {
...
+ * (unless `t` is just less than the earliest `StoredInfo.time` by roundoff, in
+ * which case it returns the earliest StoredInfo.)
+ *
+ * This entire operation needs to be threadsafe so an additional parameter
+ * `all_info_size` is passed which is the size of `all_stored_infos` that's
+ * allowed to be used in this operation. This avoids threading undefined
+ * behavior.
+ *
+ * \note An external entity is responsible for passing in a valid
+ * `all_info_size`.
```
wthrowe: That's generally how function arguments work.  What are you trying to say here?

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -125,6 +154,11 @@ void PiecewisePolynomial<MaxDeriv>::update(
 
   func[MaxDeriv] = std::move(updated_max_deriv);
   deriv_info_at_update_times_.emplace_back(time_of_update, std::move(func));
+  // While it is technically possible for a different thread to call `update`
+  // during this call, there should never be two different threads trying to
+  // update this object at the same time in our algorithm. An update should come
+  // from only one place and nowhere else so this assignment should be ok
```
wthrowe: This comment should be on the previous statement.  The statement it is currently attached to would be fine in such a situation.

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -68,7 +108,9 @@ void QuaternionFunctionOfTime<MaxDeriv>::update_stored_info() {
           << angle_deriv_info.size() << " stored angles and "
           << stored_quaternions_and_times_.size() << " stored quaternions.");
 
-  const size_t last_index = stored_quaternions_and_times_.size();
+  // We copy into a size_t so this whole function uses the same index
+  const size_t last_index =
+      stored_quaternion_size_.load(std::memory_order_relaxed);
```
wthrowe: Not necessarily a change request, but why didn't you use a `size_t` to begin with?

wthrowe (changes requested):

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
knelli2: What do you mean by this? The datatype of `size_t` didn't change

wthrowe: You're specifically commenting on the result type, so I assume it is significant somehow?

I assumed this was some commentary on the choice of `uint64_t` for the atomic type instead of `size_t` (despite them being the same everywhere we support), which did seem odd but not something that needed changing.

knelli2: Oh yeah I added `size_t` because of the `uint64_t`. And I used `uint64_t` because that's what @nilsdeppe suggested. I'm guessing it's just to be explicit so we avoid different behavior on different machines because we do padding in the hpp?

nilsdeppe: Yea, in practice size_t and uint64_t are the same, but I've been a bit paranoid when it comes to atomics just to get some more control over what things translate into at the hardware level

knelli2: I just changed the wording of this comment to not use `size_t`

wthrowe (approved):

nilsdeppe (approved):

PR #5199, opened 2023-07-17
Add scalar stress energy tensor compute functions

guilara:
## Proposed changes

Add methods to compute the trace-reversed stress energy tensor of the scalar field and add it to the rhs of the metric equations

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ScalarTensor/StressEnergy.cpp
+++ b/src/Evolution/Systems/ScalarTensor/StressEnergy.cpp
@@ -3,4 +3,41 @@
 
 #include "Evolution/Systems/ScalarTensor/StressEnergy.hpp"
 
-namespace ScalarTensor {}  // namespace ScalarTensor
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace ScalarTensor {
+
+void add_stress_energy_term_to_dt_pi(
+    const gsl::not_null<tnsr::aa<DataVector, 3_st>*> dt_pi,
+    const tnsr::aa<DataVector, 3_st>& trace_reversed_stress_energy,
+    const Scalar<DataVector>& lapse) {
+  for (size_t a = 0; a < 4; ++a) {
+    for (size_t b = a; b < 4; ++b) {
+      dt_pi->get(a, b) -=
+          16.0 * M_PI * get(lapse) * trace_reversed_stress_energy.get(a, b);
+    }
+  }
+}
+
+void trace_reversed_stress_energy(
+    const gsl::not_null<tnsr::aa<DataVector, 3_st>*> stress_energy,
+    const Scalar<DataVector>& pi_scalar,
+    const tnsr::i<DataVector, 3_st>& phi_scalar,
+    const Scalar<DataVector>& lapse) {
+  get<0, 0>(*stress_energy) = square(get(lapse) * get(pi_scalar));
+  for (size_t i = 0; i < 3; ++i) {
+    stress_energy->get(0, i + 1) =
+        -get(lapse) * get(pi_scalar) * phi_scalar.get(i);
+  }
+  for (size_t i = 0; i < 3; ++i) {
+    for (size_t j = i; j < 3; ++j) {
+      stress_energy->get(i + 1, j + 1) = phi_scalar.get(i) * phi_scalar.get(j);
+    }
+  }
+}
+
+}  // namespace ScalarTensor
--- a/src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
+++ b/src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
@@ -3,4 +3,108 @@
 
 #pragma once
 
-namespace ScalarTensor {}  // namespace ScalarTensor
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "Evolution/Systems/ScalarTensor/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace ScalarTensor {
+
+/*!
+ * \brief Add in the trace-reversed stress-energy source term to the \f$\Pi\f$
+ * evolved variable of the ::gh system.
+ *
+ * \details The only stress energy source term in the Generalized Harmonic
+ * evolution equations is in the equation for \f$\Pi_{a b}\f$:
+ * \f[
+ * \partial_t \Pi_{ab} + \text{\{spatial derivative terms\}} =
+ * \text{\{GH source terms\}}
+ * - 16 \pi \alpha (T^{(\Psi)}_{ab} - \frac{1}{2} g_{a b} g^{cd}T^{(\Psi)}_{cd})
+ * ~.
+ * \f]
+ *
+ * (note that this function takes as argument the trace-reversed stress-energy
+ * tensor)
+ *
+ * This function adds that contribution to the existing value of `dt_pi`. The
+ * spacetime terms in the GH equation should be computed before passing the
+ * `dt_pi` to this function for updating.
+ *
+ * \param dt_pi Time derivative of the $\Pi_{ab}$ variable in the ::gh system.
+ * The vacuum part should be computed before with ::gh::TimeDerivative
+ * \param trace_reversed_stress_energy Trace-reversed stress energy tensor of
+ * the scalar $T^{(\Psi), \text{TR}}_{a b} \equiv T^{(\Psi)}_{ab} - \frac{1}{2}
+ *  g_{a b} g^{cd}T^{(\Psi)}_{cd} = \partial_a \Psi \partial_b \Psi $.
+ * \param lapse Lapse $\alpha$.
+ *
+ * \see `gh::TimeDerivative` for details about the spacetime
+ * part of the time derivative calculation.
+ */
+void add_stress_energy_term_to_dt_pi(
+    gsl::not_null<tnsr::aa<DataVector, 3_st>*> dt_pi,
+    const tnsr::aa<DataVector, 3_st>& trace_reversed_stress_energy,
+    const Scalar<DataVector>& lapse);
+
+/*!
+ * \brief Compute the trace-reversed stress-energy tensor of the scalar field.
+ *
+ * \details The trace-reversed stress energy tensor is needed to compute the
+ * backreaction of the scalar to the spacetime evolution and is given by
+ * \f{align*}{
+ * T^{(\Psi), \text{TR}}_{a b} &\equiv T^{(\Psi)}_{ab} - \frac{1}{2}
+ *  g_{a b} g^{cd}T^{(\Psi)}_{cd} \\
+ *  &= \partial_a \Psi \partial_b \Psi ~,
+ * \f}
+ *
+ * where \f$T^{(\Psi)}_{ab}\f$ is the standard stress-energy tensor of the
+ * scalar.
+ *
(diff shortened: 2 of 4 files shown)


src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
```diff
@@ -3,4 +3,93 @@
...
+ * tensor)
+ *
+ * This function adds that contribution to the existing value of `dt_pi`. The
+ * spacetime terms in the GH equation should be computed before passing the
+ * `dt_pi` to this function for updating.
+ *
+ * \see `gh::TimeDerivative` for details about the spacetime
+ * part of the time derivative calculation.
+ */
+void add_stress_energy_term_to_dt_pi(
```
nikwit, line 45: are you sure you need this function? can you not just add it in the action from which you presumably want to call it?

src/Evolution/Systems/ScalarTensor/StressEnergy.cpp
```diff
@@ -3,4 +3,41 @@
...
+
+void trace_reversed_stress_energy(
+    const gsl::not_null<tnsr::aa<DataVector, 3_st>*> stress_energy,
+    const Scalar<DataVector>& pi_scalar,
+    const tnsr::i<DataVector, 3_st>& phi_scalar,
+    const Scalar<DataVector>& lapse) {
+  get<0, 0>(*stress_energy) = square(lapse.get() * pi_scalar.get());
+  for (size_t i = 0; i < 3; ++i) {
+    stress_energy->get(0, i + 1) = -lapse.get();
+    stress_energy->get(0, i + 1) *= pi_scalar.get() * phi_scalar.get(i);
```
nikwit: can you just combine these two lines?

src/Evolution/Systems/ScalarTensor/StressEnergy.cpp
```diff
@@ -3,4 +3,41 @@
...
+    const Scalar<DataVector>& lapse) {
+  for (size_t a = 0; a < 4; ++a) {
+    for (size_t b = a; b < 4; ++b) {
+      dt_pi->get(a, b) -=
+          16.0 * M_PI * get(lapse) * trace_reversed_stress_energy.get(a, b);
+    }
+  }
+}
+
+void trace_reversed_stress_energy(
```
nikwit, line 26: you can just write `3` everywhere here instead of `3_st`, but it might be best to just template it on `Dim` in case we want to extend this.

src/Evolution/Systems/ScalarTensor/StressEnergy.cpp
```diff
@@ -3,4 +3,41 @@
...
+    const gsl::not_null<tnsr::aa<DataVector, 3_st>*> stress_energy,
+    const Scalar<DataVector>& pi_scalar,
+    const tnsr::i<DataVector, 3_st>& phi_scalar,
+    const Scalar<DataVector>& lapse) {
+  get<0, 0>(*stress_energy) = square(lapse.get() * pi_scalar.get());
```
nikwit: `get(lapse)` and `get(pi_scalar)` here and all scalars below

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
```diff
@@ -3,4 +3,93 @@
...
+ *
+ * where \f$T^{(\Psi)}_{ab}\f$ is the standard stress-energy tensor of the
+ * scalar.
+ *
+ * In terms of the evolved variables of the scalar,
+ * \f{align*}{
+    T^{(\Psi), \text{TR}}_{00} &= \alpha^2 \Pi^2 ~, \\
+    T^{(\Psi), \text{TR}}_{j 0} &= T^{(\Psi), \text{TR}}_{0j}
+                                 = - \alpha \Pi \Phi_j ~, \\
+    T^{(\Psi), \text{TR}}_{ij} &= \Phi_i \Phi_j ~,
```
nikwit, line 69: I find this naming a bit confusing because I initially thought that `Pi` is some contraction of `Pi_\alpha \beta` (and the same for `Phi_i`) but now realize this is an evolved Scalar from the CSW system. is this standard?

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
```diff
@@ -3,4 +3,93 @@
...
+ *
+ * where \f$\alpha\f$ is the lapse.
+ */
+void trace_reversed_stress_energy(
+    gsl::not_null<tnsr::aa<DataVector, 3_st>*> stress_energy,
+    const Scalar<DataVector>& pi_scalar,
+    const tnsr::i<DataVector, 3_st>& phi_scalar,
+    const Scalar<DataVector>& lapse);
+
+namespace Tags {
```
nikwit, line 87: I think this should be moved to `Tags.hpp` but not sure if we have clear policies on that, @nilsvu ?

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
```diff
@@ -3,4 +3,93 @@
...
+
+/*!
+ * \brief Compute tag for the trace reversed stress energy tensor.
+ *
+ * \details Compute using ScalarTensor::trace_reversed_stress_energy.
+ */
+struct TraceReversedStressEnergyCompute
```
nikwit, line 94: I would either template this on `Dim` or declare a `static constexpr size_t Dim = 3` inside the struct.

nikwit (changes requested):
some small commments

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
guilara: I was following the structure of GrMhdValencia. The idea is to call this function inside `TimeDerivative`, after computing the decoupled eqs for the metric and the scalar. The plan is to submit that PR soon.

src/Evolution/Systems/ScalarTensor/StressEnergy.cpp
guilara: Sure

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
guilara: Yes, I understand this might be somewhat confusing. I thought about adding a superindex ^(\Psi) or sth, but felt it was too much notation. I listed the evolved variables in `System` with their names. Maybe could write a warning to not assume contractions of them.

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
guilara: I've only tested the system in 3D, although both of them are templated in dim. Was trying to avoid generalizing for now. But templating this tag wouldn't hurt.

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
nikwit: Is this function going to become more complex at some point? Given that it is so simple it is probably  easier  and more readable to just write it out in `TimeDerivative` and add these docs to the action.

guilara: I don't think it is going to become much more complicated that this. What might happen in the two scalars system (that I want to add eventually) is that we call it twice.

src/Evolution/Systems/ScalarTensor/StressEnergy.cpp
guilara: Done

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
guilara: Added the `static constexpr` option

guilara:
Also added missing test for the compute tag

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
guilara: See #5212

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
```diff
@@ -3,4 +3,94 @@
...
+ * evolution equations is in the equation for \f$\Pi_{a b}\f$:
+ * \f[
+ * \partial_t \Pi_{ab} + \text{\{spatial derivative terms\}} =
+ * \text{\{GH source terms\}}
+ * - 16 \pi \alpha (T^{(\Psi)}_{ab} - \frac{1}{2} g_{a b} g^{cd}T^{(\Psi)}_{cd})
+ * ~.
+ * \f]
+ *
+ * (note that this function takes as argument the trace-reversed stress-energy
+ * tensor)
```
nilsvu, line 29: use `\param` to document all arguments

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
nilsvu: Agreed. No clear policy, but easier to find there.

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
```diff
@@ -3,4 +3,94 @@
...
+    const Scalar<DataVector>& lapse);
+
+namespace Tags {
+
+/*!
+ * \brief Compute tag for the trace reversed stress energy tensor.
+ *
+ * \details Compute using ScalarTensor::trace_reversed_stress_energy.
+ */
+struct TraceReversedStressEnergyCompute
```
nilsvu, line 94: Same comment about avoiding "compute-item programming" as in #5205. Depends on your use case.

nilsvu (changes requested):

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
guilara: Added

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
nikwit: yes, I think it would be more readable to just write out the function in the action. But if you prefer this, ok.

[optional] you can use tensor expressions here either way:
 
`tenex::evaluate(dt_pi, - 16. * M_PI * lapse() * trace_reversed_stress_energy(ti::a, ti::b));`

guilara: Ok, I would prefer to leave this function. It also keeps the parallel structure with the grmhd system.

guilara:
I am not sure why the gcc-11 test is failing. Any idea?

nikwit:
It just randomly timed out, unrelated to this PR.

nikwit (approved):

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
```diff
@@ -3,4 +3,106 @@
...
+ * tensor)
+ *
+ * This function adds that contribution to the existing value of `dt_pi`. The
+ * spacetime terms in the GH equation should be computed before passing the
+ * `dt_pi` to this function for updating.
+ *
+ * \param dt_pi Time derivative of the $\Pi_{ab}$ variable in the ::gh system.
+ * The vacuum part should be computed before with ::gh::TimeDerivative
+ * \param trace_reversed_stress_energy Trace-reversed stress energy tensor of
+ * the scalar $T_{ab}^{(\Psi)}$.
```
nilsvu: `_\mathrm{TR}` is missing I think? Maybe write out the definition as well here.

nilsvu (changes requested):

src/Evolution/Systems/ScalarTensor/StressEnergy.hpp
guilara: Oh, right. Fixed.

nilsvu (approved):

PR #5202, opened 2023-07-17
Update UpdateControlSystem to handle size control

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size/CMakeLists.txt
+++ b/src/ControlSystem/ControlErrors/Size/CMakeLists.txt
@@ -25,4 +25,5 @@ spectre_target_headers(
   Error.hpp
   Initial.hpp
   RegisterDerivedWithCharm.hpp
+  Update.hpp
   )
--- a/src/ControlSystem/ControlErrors/Size/CMakeLists.txt
+++ b/src/ControlSystem/ControlErrors/Size/CMakeLists.txt
@@ -1,10 +1,6 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
-set(LIBRARY SizeControlErrors)
-
-add_spectre_library(${LIBRARY})
-
 spectre_target_sources(
   ${LIBRARY}
   PRIVATE
@@ -29,16 +25,4 @@ spectre_target_headers(
   Error.hpp
   Initial.hpp
   RegisterDerivedWithCharm.hpp
-  )
-
-target_link_libraries(
-  ${LIBRARY}
-  PUBLIC
-  ApparentHorizons
-  Boost::boost
-  DataStructures
-  ErrorHandling
-  GSL::gsl
-  Interpolation
-  Parallel
   )
--- a/src/ControlSystem/ControlErrors/Size/Update.hpp
+++ b/src/ControlSystem/ControlErrors/Size/Update.hpp
@@ -1,0 +1,118 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <limits>
+#include <optional>
+#include <string>
+
+#include "ControlSystem/Averager.hpp"
+#include "ControlSystem/ControlErrors/Size.hpp"
+#include "ControlSystem/TimescaleTuner.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "Domain/Structure/ObjectLabel.hpp"
+#include "IO/Logging/Verbosity.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Printf.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace control_system::size {
+/*!
+ * \brief Updates the Averager with information from the \link
+ * control_system::ControlErrors::Size Size control error \endlink.
+ *
+ * \details After we calculate the control error, we check if a discontinuous
+ * change has occurred (the internal `control_system::size::State` changed). If
+ * no discontinuous change has occurred, then we do nothing. If one did occur,
+ * we get a history of control errors from the \link
+ * control_system::ControlErrors::Size Size control error \endlink and use that
+ * to repopulate the Averager history.
+ *
+ * \note See `control_system::Tags::Averager` for why the Averager is templated
+ * on `DerivOrder - 1`.
+ */
+template <size_t DerivOrder, ::domain::ObjectLabel Horizon,
+          typename Metavariables>
+void update_averager(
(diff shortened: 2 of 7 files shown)


wthrowe (dismissed):

src/ControlSystem/ControlErrors/Size/Update.hpp
```diff
@@ -0,0 +1,115 @@
...
+ * change has occurred (the internal `control_system::size::State` changed). If
+ * no discontinuous change has occurred, then we do nothing. If one did occur,
+ * we get a history of control errors from the \link
+ * control_system::ControlErrors::Size Size control error \endlink and use that
+ * to repopulate the Averager history.
+ */
+template <size_t DerivOrder, ::domain::ObjectLabel Horizon,
+          typename Metavariables>
+void update_averager(
+    const gsl::not_null<Averager<DerivOrder - 1>*> averager,
```
markscheel, line 39: Do we want a comment as to why Averager is DerivOrder-1 and not DerivOrder?

src/ControlSystem/UpdateControlSystem.hpp
```diff
@@ -61,18 +90,22 @@ namespace control_system {
...
  *    the current measurement back to 0.
- * 6. Update the damping timescales in the TimescaleTuner and compute the
- *    control signal using the control error and its derivatives.
- * 7. Calculate the new measurement timescale based off the updated damping
- *    timescales and the number of measurements per update.
+ * 6. Compute the control signal using the control error and its derivatives and
+ *    update the damping timescales in the TimescaleTuner. If this is \link
+ *    control_system::Systems::Size size \endlink control, there is an extra
+ *    step after we update the damping timescale. See
+ *    `control_system::size::update_tuner` for this step.
```
markscheel, line 85: [minor] There is no longer a 7.  Steps go from 6. to 8.

src/ControlSystem/UpdateControlSystem.hpp
```diff
@@ -131,6 +164,12 @@ struct UpdateControlSystem {
           function_of_time_name, time, current_measurement, Q);
     }
 
+    if constexpr (size::is_size_v<ControlSystem>) {
+      size::update_averager(
+          make_not_null(&averager), make_not_null(&control_error), cache, time,
+          tuner.curent_timescale(), function_of_time_name, current_measurement);
```
markscheel: curent_timescale -> current_timescale

src/ControlSystem/ControlErrors/Size/Update.hpp
```diff
@@ -0,0 +1,115 @@
...
+        control_error,
+    const Parallel::GlobalCache<Metavariables>& cache, const double time,
+    const DataVector& current_timescale,
+    const std::string& function_of_time_name, const int current_measurement) {
+  if (control_error->discontinuous_change_has_occurred()) {
+    control_error->reset();
+    averager->clear();
+
+    const std::deque<std::pair<double, double>> control_error_history =
+        control_error->control_error_history();
```
markscheel, line 50: Not really a change request, but here we need to be careful that the history obtained is the history of the new state, not the old state, if there has been a state change.

knelli2: The `control_error_history` function gets the history for the *current* state. The `udate_averager` function is called after we calculate the control error and potentially switch states so the control error should be in the *new* state. However, I'll add a note in `UpdateControlSystem` so that we are aware of this

knelli2:
@markscheel Squashed in some comments addressing your comments.

src/ControlSystem/UpdateControlSystem.hpp
wthrowe: @knelli2 How did this not get caught by the compiler?  Even if this is not easy to unit-test at the moment, I thought you were using it on your branch.

knelli2: So on my branch, I actually don't have the `update_averager` or `update_tuner` functions. I just put the bodies right here inside the `if constexpr`s. To easily test these new blocks of code, I factored them out into functions for this PR. But the `UpdateControlSystem` struct is never called with size control so it never tried to compile the `if constexpr`s.

wthrowe: Can you check that it compiles before we merge it?

knelli2:
@wthrowe I encountered some linking errors regarding the new functions when building the BBH executable with size control. Still unsure as to exactly why, but I think it has something to do with the `SizeControlErrors` lib in `src/ControlSystem/ControlErrors/Size` and forward declaring the functions in `UpdateControlSystem` rather than including the header (because of the other lib). I don't really know why this was its own lib in the first place, so I added an extra commit before the new one that just removes the lib so we just have one `ControlSystem` lib (and I was free to include the header in `UpdateControlSystem`.

src/ControlSystem/ControlErrors/Size/CMakeLists.txt
```diff
@@ -30,15 +26,3 @@ spectre_target_headers(
...
-target_link_libraries(
-  ${LIBRARY}
-  PUBLIC
-  ApparentHorizons
-  Boost::boost
-  DataStructures
-  ErrorHandling
-  GSL::gsl
-  Interpolation
-  Parallel
```
wthrowe, line 43: Presumably the library this was merged into needs these now.  (Some are already included.)

wthrowe (changes requested):
Not surprised about the linking errors.  The library dependency doesn't go away just because you skip the include.

Eventually we're going to want to find a way to not have system-specific code in the general control-system functions so that everything that uses a control system doesn't need, for example, ApparentHorizons.  The existing split doesn't help with that, though, because the dependency goes the wrong way, so no problem with removing it.

knelli2:
After looking it up, the only reason this depends on `ApparentHorizons` is because we use functions in `ApparentHorizons/StrahlkorperGr.hpp`. However, it looks like most (if not all?) of those functions can be moved to `NumericalAlgorithms/SphericalHarmonics/` as they have nothing specific to do with horizons and everything specific to do with Strahlkorpers.

So maybe as a follow up PR I'll move those.

markscheel (approved):
Looks good to me

nilsdeppe:
@wthrowe could you please take a look at this PR again?

wthrowe (approved):
ApparentHorizons was just an example.  While not our main focus, control systems make sense in systems that have nothing to do with GR or spheres or anything.  The library should be problem-agnostic.

But not something to deal with right now.

PR #5204, opened 2023-07-18
Add some docs on Py formatting

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/Hooks/pre-commit.sh
+++ b/tools/Hooks/pre-commit.sh
@@ -47,7 +47,7 @@ done
 
 # Python file checks
 if [ ${#python_files[@]} -ne 0 ]; then
-    # Use black to check Python file formatting, only if it is installed.
+    # Use black to check Python file formatting
     @Python_EXECUTABLE@ -m black --version > /dev/null
     if [ $? -eq 0 ]; then
         @Python_EXECUTABLE@ -m black --check --quiet ${python_files[@]}
@@ -56,6 +56,10 @@ if [ ${#python_files[@]} -ne 0 ]; then
             printf "Found Python formatting errors.\n"
             printf "Please run 'black .' in the repository.\n"
         fi
+    else
+        printf "Could not find 'black' Python formatter. Install with:\n"
+        printf "pip3 install -r support/Python/dev_requirements.txt\n"
+        exit 1
     fi
 
     # Use isort to check python import order, only if it is installed.
@@ -67,6 +71,10 @@ if [ ${#python_files[@]} -ne 0 ]; then
             printf "Found unsorted Python imports.\n"
             printf "Please run 'isort .' in the repository.\n"
         fi
+    else
+        printf "Could not find 'isort' Python formatter. Install with:\n"
+        printf "pip3 install -r support/Python/dev_requirements.txt\n"
+        exit 1
     fi
 fi
 
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -137,7 +137,9 @@ jobs:
       - name: Check Python formatting
         run: |
           cd $GITHUB_WORKSPACE
+          echo "Using 'black' to check Python formatting..."
           black --check .
+          echo "Using 'isort' to check Python formatting..."
           isort --check-only .
       - name: Test script
         run: |
--- a/docs/DevGuide/GitHubActions.md
+++ b/docs/DevGuide/GitHubActions.md
@@ -78,6 +78,8 @@ The following build jobs are launched:
   - The file `tests/Unit/RunSingleTest/CMakeLists.txt` wasn't modified
   In addition, the CHECK_FILES job tests Python formatting, the release
   workflow, and other tools in `tools/`.
+* "Check Python formatting" runs the `black` and `isort` formatters over the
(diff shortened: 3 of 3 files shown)


tools/Hooks/pre-commit.sh
```diff
@@ -47,7 +47,7 @@ done
 
 # Python file checks
 if [ ${#python_files[@]} -ne 0 ]; then
-    # Use black to check Python file formatting, only if it is installed.
+    # Use black to check Python file formatting
     @Python_EXECUTABLE@ -m black --version > /dev/null
     if [ $? -eq 0 ]; then
         @Python_EXECUTABLE@ -m black --check --quiet ${python_files[@]}
```
knelli2, line 53: What do you think of actually running black here like we actually run `clang-format` on c++ files? Maybe also putting them in the diff that `ClangFormat.py` creates?

knelli2 (commented):
What do you think about also doing this

```diff
diff --git a/docs/Installation/Installation.md b/docs/Installation/Installation.md
index 014cef7b70..b89fb99938 100644
--- a/docs/Installation/Installation.md
+++ b/docs/Installation/Installation.md
@@ -93,7 +93,7 @@ all of these dependencies.
   Alternatively, you can set `BOOTSTRAP_PY_DEPS=ON` when configuring a build
   with CMake to install missing Python packages into the build directory
   automatically.
-  <details>
+  <details open>
   \include support/Python/requirements.txt
   </details>
 
@@ -107,7 +107,7 @@ all of these dependencies.
   Install with `pip3 install -r support/Python/dev_requirements.txt`.
   Make sure you are working in a [Python venv](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment)
   before installing packages.
-  <details>
+  <details open>
   \include support/Python/dev_requirements.txt
   </details>
 * [Google Benchmark](https://github.com/google/benchmark) - to do
```

When I searched for hidden things on Chrome (linux), it actually opened the `Details` tab and highlighted the search. However, after some searching, apparently this behavior is browser dependent and not all browsers will search hidden content. So if we want to guarantee searchability in the installation docs, we can just have the details open initially (or just get rid of the hidden details and have the file included always).

tools/Hooks/pre-commit.sh
nilsdeppe: I think we could reasonably just run it if we can add the changes directly to the commit changes. I haven't looked into this.

@wthrowe how would you feel about this?

nilsdeppe (dismissed):
I'm happy with this. Thanks for making these changes! This would've been enough to install black and get formatting right :)

tools/Hooks/pre-commit.sh
wthrowe: Doing the same thing we do with clang-format makes sense.  I am agnostic on aborting the commit if the check fails.  I'm opposed to applying the changes during the commit process.

nilsdeppe: I agree, having something change code for you usually causes more challenges than it solves, especially in a case like this where the thing you need to do is 7 characters

nilsdeppe:
> What do you think about also doing this
> 
> ```diff
> diff --git a/docs/Installation/Installation.md b/docs/Installation/Installation.md
> index 014cef7b70..b89fb99938 100644
> --- a/docs/Installation/Installation.md
> +++ b/docs/Installation/Installation.md
> @@ -93,7 +93,7 @@ all of these dependencies.
>    Alternatively, you can set `BOOTSTRAP_PY_DEPS=ON` when configuring a build
>    with CMake to install missing Python packages into the build directory
>    automatically.
> -  <details>
> +  <details open>
>    \include support/Python/requirements.txt
>    </details>
>  
> @@ -107,7 +107,7 @@ all of these dependencies.
>    Install with `pip3 install -r support/Python/dev_requirements.txt`.
>    Make sure you are working in a [Python venv](https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/#creating-a-virtual-environment)
>    before installing packages.
> -  <details>
> +  <details open>
>    \include support/Python/dev_requirements.txt
>    </details>
>  * [Google Benchmark](https://github.com/google/benchmark) - to do
> ```
> 
> When I searched for hidden things on Chrome (linux), it actually opened the `Details` tab and highlighted the search. However, after some searching, apparently this behavior is browser dependent and not all browsers will search hidden content. So if we want to guarantee searchability in the installation docs, we can just have the details open initially (or just get rid of the hidden details and have the file included always).

I like this suggestion. @nilsvu would you be okay adding that?

nilsvu:
Well I put the requirements.txt file in `<details>` for a reason. You don't really need to know the list of packages, just `pip install -r [dev_]requirements.txt`. IMO showing the `<details>` is unnecessarily distracting on the installation page.
I added another hint how to install black and isort to the "running the checks locally" docs.

nilsdeppe:
But the point remains, if I search the installation page for any of our python dependencies they may or may not show up...

nilsdeppe (approved):
I'd rather get these improvements and worry about whether to detail or not later...

PR #5205, opened 2023-07-18
Add scalar sources

guilara:
## Proposed changes

Add a scalar field mass term to the rhs of the scalar equations. The idea is to add more complicate sources in the same way.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
+++ b/src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
@@ -1,0 +1,24 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp"
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/SetNumberOfGridPoints.hpp"
+
+namespace ScalarTensor {
+
+void add_scalar_source_to_dt_pi_scalar(
+    gsl::not_null<Scalar<DataVector>*> dt_pi_scalar,
+    const Scalar<DataVector>& scalar_source, const Scalar<DataVector>& lapse) {
+  get(*dt_pi_scalar) += get(lapse) * get(scalar_source);
+}
+
+void mass_source(
+    const gsl::not_null<Scalar<DataVector>*> scalar_source,
+    const Scalar<DataVector>& psi, const double mass_psi) {
+  get(*scalar_source) = square(mass_psi) * get(psi);
+}
+}  // namespace ScalarTensor
--- a/src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
+++ b/src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
@@ -1,0 +1,96 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "Evolution/Systems/ScalarTensor/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace ScalarTensor {
+
+/*!
+ * \brief Add in the source term to the \f$\Pi\f$
+ * evolved variable of the ::CurvedScalarWave system.
+ *
+ * \details The only source term in the wave equation
+ * \f[
+ *  \Box \Psi = \mathcal{S} ~,
+ *  \f]
+ *
+ * is in the equation for \f$\Pi\f$:
+ * \f[
+ *  \partial_t \Pi + \text{\{spatial derivative
+ * terms\}} = \alpha \mathcal{S}
+ * ~,
+ * \f]
+ *
+ * where \f$\mathcal{S}\f$ is the source term (e. g. in the Klein-Gordon
+ * equation, the source term is the derivative of the scalar potential
+ * \f$\mathcal{S} \equiv \partial V / \partial \Psi \f$.)
+ *
+ * This function adds that contribution to the existing value of `dt_pi_scalar`.
+ * The wave equation terms in the scalar equation should be computed before
+ * passing the `dt_pi_scalar` to this function for updating.
+ *
+ * \param dt_pi_scalar Time derivative terms of $\Pi$. The sourceless part
+ * should be computed before with ::CurvedScalarWave::TimeDerivative.
+ * \param scalar_source Source term $\mathcal{S}$ for the scalar equation.
+ * \param lapse Lapse $\alpha$.
+ *
+ * \see `CurvedScalarWave::TimeDerivative` for details about the source-less
+ * part of the time derivative calculation.
+ */
+void add_scalar_source_to_dt_pi_scalar(
+    gsl::not_null<Scalar<DataVector>*> dt_pi_scalar,
+    const Scalar<DataVector>& scalar_source, const Scalar<DataVector>& lapse);
+
+/*!
+ * \brief Computes the source term given by the mass of the scalar.
+ *
+ * \details For a scalar field with mass parameter \f$ m_\Psi \f$,
+ * the wave equation takes the form
+ * \f[
+ *   \Box \Psi = \mathcal{S} ~,
+ * \f]
+ *
+ * where the source is given by
+ * \f[
+ *   \mathcal{S} \equiv m^2_\Psi ~.
+ * \f]
+ *
+ * Here the mass parameter value is an option that needs to be specified in the
+ * input file.
+ *
+ * \param scalar_source Source term $\mathcal{S}$ for the scalar equation.
+ * \param psi Scalar field $\Psi$.
+ * \param mass_psi Mass of the scalar field $m_\Psi$.
+ *
+ * \see `ScalarTensor::Tags::ScalarMass` for details about the mass.
+ */
+void mass_source(gsl::not_null<Scalar<DataVector>*> scalar_source,
+                 const Scalar<DataVector>& psi, const double mass_psi);
+
+namespace Tags {
+
+/*!
+ * \brief Compute tag for the scalar source.
+ *
+ * \details Compute the scalar source from data box items using
+ * `mass_source`.
+ */
+struct ScalarSourceCompute : ScalarSource, db::ComputeTag {
+  using argument_tags =
+      tmpl::list<CurvedScalarWave::Tags::Psi, ScalarTensor::Tags::ScalarMass>;
+  using return_type = Scalar<DataVector>;
+  static constexpr void (*function)(const gsl::not_null<return_type*> result,
+                                    const Scalar<DataVector>&,
+                                    const double) = &mass_source;
+  using base = ScalarSource;
+};
+
+}  // namespace Tags
+}  // namespace ScalarTensor
--- a/src/Evolution/Systems/ScalarTensor/Tags.hpp
+++ b/src/Evolution/Systems/ScalarTensor/Tags.hpp
@@ -7,13 +7,14 @@
 #include "DataStructures/DataBox/Tag.hpp"
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
+#include "Options/String.hpp"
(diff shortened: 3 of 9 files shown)


src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
```diff
@@ -0,0 +1,25 @@
...
+
+#include "Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp"
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace ScalarTensor {
+
+void add_scalar_source_to_dt_pi_scalar(
```
nikwit, line 13: are you sure you need this function?

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
```diff
@@ -0,0 +1,25 @@
...
+    gsl::not_null<Scalar<DataVector>*> dt_pi_scalar,
+    const Scalar<DataVector>& scalar_source, const Scalar<DataVector>& lapse) {
+  dt_pi_scalar->get() += get(lapse) * scalar_source.get();
+}
+
+void compute_mass_source(
+    const gsl::not_null<Scalar<DataVector>*> scalar_source,
+    const Scalar<DataVector>& psi, const double mass_psi) {
+  // Make sure it has the same size
+  *scalar_source = make_with_value<Scalar<DataVector>>(psi, 0.);
```
nikwit: use `set_number_of_grid_points(scalar_source, psi)`

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
```diff
@@ -0,0 +1,25 @@
...
+  scalar_source->get() = square(mass_psi) * psi.get();
```
nikwit: use `get(*scalar_source)` and `get(psi)`

src/Evolution/Systems/ScalarTensor/Sources/Tags.hpp
```diff
@@ -0,0 +1,48 @@
...
+ * \f[
+ *   \Box \Psi = \mathcal{S} ~.
+ * \f]
+ */
+struct ScalarSource : db::SimpleTag {
+  using type = Scalar<DataVector>;
+};
+}  // namespace Tags
+
+namespace OptionTags {
```
nikwit: test these with `test_option_tag`.

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
```diff
@@ -0,0 +1,90 @@
...
+
+namespace Tags {
+
+/*!
+ * \brief Compute tag for the scalar source.
+ *
+ * \details Compute the scalar source from data box items using
+ * `compute_mass_source`.
+ */
+struct ScalarSourceCompute : ScalarSource, db::ComputeTag {
```
nikwit, line 85: test this with `test_compute_tag`

nikwit (changes requested):
some small comments

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
```diff
@@ -0,0 +1,90 @@
...
+#include "Utilities/Gsl.hpp"
+
+namespace ScalarTensor {
+
+/*!
+ * \brief Add in the source term to the \f$\Pi\f$
+ * evolved variable of the ::CurvedScalarWave system.
+ *
+ * \details The only source term in the wave equation
+ * \f[
```
nilsvu, line 20: FYI, you can use standard LaTeX syntax (without the silly `\f`s, just `\begin{equation}`). Up to you, I'm not requesting you change anything.

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
```diff
@@ -0,0 +1,90 @@
...
+ *  \partial_t \Pi + \text{\{spatial derivative
+ * terms\}} = \alpha \mathcal{S}
+ * ~,
+ * \f]
+ *
+ * where \f$\mathcal{S}\f$ is the source term (e. g. in the Klein-Gordon
+ * equation, the source term is the derivative of the scalar potential
+ * \f$\mathcal{S} \equiv \partial V / \partial \Psi \f$.)
+ *
+ * (Note that this function takes as argument the source term.)
```
nilsvu: Use `\param` to document each argument. Also in function below. Add a note if the return-by-reference argument is sized by the function.

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
```diff
@@ -0,0 +1,90 @@
...
+ * \f[
+ *   \mathcal{S} \equiv m^2_\Psi ~.
+ * \f]
+ *
+ * Here the mass parameter value is an option that needs to be specified in the
+ * input file.
+ *
+ * \see `ScalarTensor::Tags::ScalarMass` for details about the mass.
+ */
+void compute_mass_source(gsl::not_null<Scalar<DataVector>*> scalar_source,
```
nilsvu: No need to prefix functions with "compute". (They better compute something 😄 )

nilsvu (changes requested):

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
nilsvu, line 85: My recommendation is to avoid compute items if they're part of your algorithm. It's a lot easier to understand if algorithms are implemented in standard C++ functions rather than through a hierarchy of compute items. Computing things in a function can also be optimized better, reusing memory, etc. Compute items are mostly useful for observing derived quantities. So if you want to use a compute item here or not depends on your use case.

src/Evolution/Systems/ScalarTensor/Sources/Tags.hpp
```diff
@@ -0,0 +1,48 @@
...
+/*!
+ * \brief Tag holding the source term of the scalar equation.
+ *
+ * \details This tag hold the source term \f$ \mathcal{S} \f$,
+ * entering a wave equation of the form
+ * \f[
+ *   \Box \Psi = \mathcal{S} ~.
+ * \f]
+ */
+struct ScalarSource : db::SimpleTag {
```
nilsvu: I'd move all these tags up to ScalarTensor/Tags, as @nikwit suggested in the other PR.

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
guilara: Thanks, I actually didn't know. Thought it had to be this way because I always saw these funny symbols around

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
guilara: Thanks for the recommendation. This is actually an important point mainly for the following TimeDerivative PR https://github.com/sxs-collaboration/spectre/pull/5212 . I wrote this system using a quite few compute tags for some of the gr quantities and empirically it seems not to be that much slower than gh. I think this can be refined/optimized in a second pass later as it might need a little bit of care to implement.

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
guilara: Fixed

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
guilara: Added test

src/Evolution/Systems/ScalarTensor/Sources/Tags.hpp
guilara: Added

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
guilara: Done

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
guilara: Done

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
guilara: Wrote it to have parallel structure with the stress tensor part. But what it does is very simple

src/Evolution/Systems/ScalarTensor/Sources/Tags.hpp
guilara: Moved

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
nilsvu: Ok. Also keep in mind that compute items take up memory throughout the full simulation. If you need to re-compute a quantity in every time step and don't need to observe it (i.e. temporary / intermediate quantities), then prefer computing it directly in the time derivative function and reuse some of the temporary memory.

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
nilsvu: To clarify, do something like this:

```cpp
/*!
 * ...
 * \param dt_pi_scalar Time derivative terms of ..., should be computed before...
 * \param scalar_source Source term $\mathcal{S}$
 * \param lapse Lapse $\alpha$
 */
```

src/Evolution/Systems/ScalarTensor/Tags.hpp
```diff
@@ -7,13 +7,14 @@
...
+#include "Options/String.hpp"
 #include "Utilities/TMPL.hpp"
 
 /*!
  * \brief Tags for the scalar tensor system.
  */
-namespace ScalarTensor::Tags {
-
+namespace ScalarTensor {
+namespace Tags {
```
nilsvu, line 17: clang-tidy wants you to do `ScalarTensor::Tags`

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp
guilara: Added

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
```diff
@@ -0,0 +1,25 @@
...
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/SetNumberOfGridPoints.hpp"
+
+namespace ScalarTensor {
+
+void add_scalar_source_to_dt_pi_scalar(
+    gsl::not_null<Scalar<DataVector>*> dt_pi_scalar,
+    const Scalar<DataVector>& scalar_source, const Scalar<DataVector>& lapse) {
+  dt_pi_scalar->get() += get(lapse) * get(scalar_source);
```
nilsvu: `get(*dt_pi_scalar)` (fix only if you're fixing other things anyway)

Same in function below

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
```diff
@@ -0,0 +1,25 @@
...
+}
+
+void mass_source(
+    const gsl::not_null<Scalar<DataVector>*> scalar_source,
+    const Scalar<DataVector>& psi, const double mass_psi) {
+  set_number_of_grid_points(scalar_source, psi);
```
nilsvu: You don't need this line because you're assigning to it in the next line

nilsvu (dismissed):

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
guilara: I've changed it. There was a typo in the doc so I changed this as well

src/Evolution/Systems/ScalarTensor/Sources/ScalarSource.cpp
guilara: Removed the set_number... line

nilsvu (approved):

nikwit (approved):

PR #5212, opened 2023-07-19
Add time derivative computation to ScalarTensor

guilara:
## Proposed changes
We compute the time derivative of the combined ScalarTensor system by calling the computation from each of the individual (CSW and GH) systems. Additionally, we add any sources to the RHS of the equations to include the back-reaction of the spacetime and source terms in the wave equation of the scalar.

In the current approach, the geometrical quantities needed by CSW are provided by compute tags. An optimization of this might be to add them as temporaries (or to store them in cache?). However, we may do this in another PR.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ScalarTensor/TimeDerivative.cpp
+++ b/src/Evolution/Systems/ScalarTensor/TimeDerivative.cpp
@@ -1,0 +1,166 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/ScalarTensor/TimeDerivative.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/DataBox/PrefixHelpers.hpp"
+#include "DataStructures/DataBox/Prefixes.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/TaggedContainers.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/CurvedScalarWave/System.hpp"
+#include "Evolution/Systems/CurvedScalarWave/TimeDerivative.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/System.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/TimeDerivative.hpp"
+#include "Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp"
+#include "Evolution/Systems/ScalarTensor/StressEnergy.hpp"
+#include "Evolution/Systems/ScalarTensor/System.hpp"
+#include "Evolution/Systems/ScalarTensor/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Time/Tags/Time.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Literals.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace ScalarTensor {
+void TimeDerivative::apply(
+    // GH dt variables
+    const gsl::not_null<tnsr::aa<DataVector, dim>*> dt_spacetime_metric,
+    const gsl::not_null<tnsr::aa<DataVector, dim>*> dt_pi,
+    const gsl::not_null<tnsr::iaa<DataVector, dim>*> dt_phi,
+    // Scalar dt variables
+    const gsl::not_null<Scalar<DataVector>*> dt_psi_scalar,
+    const gsl::not_null<Scalar<DataVector>*> dt_pi_scalar,
+    const gsl::not_null<tnsr::i<DataVector, dim, Frame::Inertial>*>
+        dt_phi_scalar,
+
+    // GH temporal variables
+    const gsl::not_null<Scalar<DataVector>*> temp_gamma1,
+    const gsl::not_null<Scalar<DataVector>*> temp_gamma2,
+    const gsl::not_null<tnsr::a<DataVector, dim>*> temp_gauge_function,
+    const gsl::not_null<tnsr::ab<DataVector, dim>*>
+        temp_spacetime_deriv_gauge_function,
+    const gsl::not_null<Scalar<DataVector>*> gamma1gamma2,
+    const gsl::not_null<Scalar<DataVector>*> half_half_pi_two_normals,
+    const gsl::not_null<Scalar<DataVector>*> normal_dot_gauge_constraint,
+    const gsl::not_null<Scalar<DataVector>*> gamma1_plus_1,
+    const gsl::not_null<tnsr::a<DataVector, dim>*> pi_one_normal,
+    const gsl::not_null<tnsr::a<DataVector, dim>*> gauge_constraint,
+    const gsl::not_null<tnsr::i<DataVector, dim>*> half_phi_two_normals,
+    const gsl::not_null<tnsr::aa<DataVector, dim>*>
+        shift_dot_three_index_constraint,
+    const gsl::not_null<tnsr::aa<DataVector, dim>*>
+        mesh_velocity_dot_three_index_constraint,
+    const gsl::not_null<tnsr::ia<DataVector, dim>*> phi_one_normal,
+    const gsl::not_null<tnsr::aB<DataVector, dim>*> pi_2_up,
+    const gsl::not_null<tnsr::iaa<DataVector, dim>*> three_index_constraint,
+    const gsl::not_null<tnsr::Iaa<DataVector, dim>*> phi_1_up,
+    const gsl::not_null<tnsr::iaB<DataVector, dim>*> phi_3_up,
+    const gsl::not_null<tnsr::abC<DataVector, dim>*>
+        christoffel_first_kind_3_up,
+    const gsl::not_null<Scalar<DataVector>*> lapse,
+    const gsl::not_null<tnsr::I<DataVector, dim>*> shift,
+    const gsl::not_null<tnsr::II<DataVector, dim>*> inverse_spatial_metric,
+    const gsl::not_null<Scalar<DataVector>*> det_spatial_metric,
+    const gsl::not_null<Scalar<DataVector>*> sqrt_det_spatial_metric,
+    const gsl::not_null<tnsr::AA<DataVector, dim>*> inverse_spacetime_metric,
+    const gsl::not_null<tnsr::abb<DataVector, dim>*> christoffel_first_kind,
+    const gsl::not_null<tnsr::Abb<DataVector, dim>*> christoffel_second_kind,
+    const gsl::not_null<tnsr::a<DataVector, dim>*> trace_christoffel,
+    const gsl::not_null<tnsr::A<DataVector, dim>*> normal_spacetime_vector,
+
+    // Scalar temporal variables
+    const gsl::not_null<Scalar<DataVector>*> result_gamma1_scalar,
+    const gsl::not_null<Scalar<DataVector>*> result_gamma2_scalar,
+
+    // Extra temporal tags
+    const gsl::not_null<tnsr::aa<DataVector, dim>*> stress_energy,
+
+    // GH spatial derivatives
+    const tnsr::iaa<DataVector, dim>& d_spacetime_metric,
+    const tnsr::iaa<DataVector, dim>& d_pi,
+    const tnsr::ijaa<DataVector, dim>& d_phi,
+
+    // scalar spatial derivatives
+    const tnsr::i<DataVector, dim>& d_psi_scalar,
+    const tnsr::i<DataVector, dim>& d_pi_scalar,
+    const tnsr::ij<DataVector, dim>& d_phi_scalar,
+
+    // GH argument variables
+    const tnsr::aa<DataVector, dim>& spacetime_metric,
+    const tnsr::aa<DataVector, dim>& pi, const tnsr::iaa<DataVector, dim>& phi,
+    const Scalar<DataVector>& gamma0, const Scalar<DataVector>& gamma1,
+    const Scalar<DataVector>& gamma2,
+    const gh::gauges::GaugeCondition& gauge_condition, const Mesh<dim>& mesh,
+    double time,
+    const tnsr::I<DataVector, dim, Frame::Inertial>& inertial_coords,
+    const InverseJacobian<DataVector, dim, Frame::ElementLogical,
+                          Frame::Inertial>& inverse_jacobian,
+    const std::optional<tnsr::I<DataVector, dim, Frame::Inertial>>&
+        mesh_velocity,
+
+    // Scalar argument variables
+    const Scalar<DataVector>& pi_scalar,
+    const tnsr::i<DataVector, dim>& phi_scalar,
+    const Scalar<DataVector>& lapse_scalar,
+    const tnsr::I<DataVector, dim>& shift_scalar,
+    const tnsr::i<DataVector, dim>& deriv_lapse,
+    const tnsr::iJ<DataVector, dim>& deriv_shift,
+    const tnsr::II<DataVector, dim>& upper_spatial_metric,
+    const tnsr::I<DataVector, dim>& trace_spatial_christoffel,
+    const Scalar<DataVector>& trace_extrinsic_curvature,
+    const Scalar<DataVector>& gamma1_scalar,
+    const Scalar<DataVector>& gamma2_scalar,
+
+    const Scalar<DataVector>& scalar_source) {
+  // Compute sourceless part of the RHS of the metric equations
+  gh::TimeDerivative<dim>::apply(
+      // GH dt variables
+      dt_spacetime_metric, dt_pi, dt_phi,
+
+      // GH temporal variables
+      temp_gamma1, temp_gamma2, temp_gauge_function,
+      temp_spacetime_deriv_gauge_function, gamma1gamma2,
+      half_half_pi_two_normals, normal_dot_gauge_constraint, gamma1_plus_1,
+      pi_one_normal, gauge_constraint, half_phi_two_normals,
+      shift_dot_three_index_constraint,
+      mesh_velocity_dot_three_index_constraint, phi_one_normal, pi_2_up,
+      three_index_constraint, phi_1_up, phi_3_up, christoffel_first_kind_3_up,
+      lapse, shift, inverse_spatial_metric, det_spatial_metric,
+      sqrt_det_spatial_metric, inverse_spacetime_metric, christoffel_first_kind,
+      christoffel_second_kind, trace_christoffel, normal_spacetime_vector,
+
+      // GH argument variables
+      d_spacetime_metric, d_pi, d_phi, spacetime_metric, pi, phi, gamma0,
+      gamma1, gamma2, gauge_condition, mesh, time, inertial_coords,
+      inverse_jacobian, mesh_velocity);
+
+  // Compute sourceless part of the RHS of the scalar equation
+  CurvedScalarWave::TimeDerivative<dim>::apply(
+      // Scalar dt variables
+      dt_psi_scalar, dt_pi_scalar, dt_phi_scalar,
+
+      // Scalar temporal variables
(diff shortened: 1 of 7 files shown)


src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
```diff
@@ -0,0 +1,217 @@
...
+#include "Evolution/Systems/CurvedScalarWave/System.hpp"
+#include "Evolution/Systems/CurvedScalarWave/TimeDerivative.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/System.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/TimeDerivative.hpp"
+#include "Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp"
+#include "Evolution/Systems/ScalarTensor/StressEnergy.hpp"
+#include "Evolution/Systems/ScalarTensor/System.hpp"
+#include "Evolution/Systems/ScalarTensor/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+// Tag obtained from gh::TimeDerivative needs to be complete?
```
nikwit: whats this?

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
```diff
@@ -0,0 +1,217 @@
...
+ * `CurvedScalar::TimeDerivativeTerms` as a 'product system' -- each
+ * independently operating on its own subset of the supplied variable
+ * collections.
+ * The additional steps taken are to compute the trace-reversed stress energy
+ * tensor associated with the scalar part of the system and add its contribution
+ * to the \f$\partial_t \Pi_{a b}\f$ variable in the Generalized Harmonic
+ * system, as well as adding any scalar sources to the variable \f$\partial_t
+ * \Pi\f$.
+ */
+struct TimeDerivativeTerms {
```
nikwit: Just call this `TimeDerivative` to make it consistent with the other systems. Same for the filename

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
```diff
@@ -0,0 +1,217 @@
...
+#include "Time/Tags/Time.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Literals.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace ScalarTensor {
+/*!
+ * \brief Compute the RHS terms and flux values for both the Generalized
+ * Harmonic formulation of Einstein's equations and the scalar equations.
```
nikwit: This sounds like there is no coupling at all. how about "... for a combined system of the Generalized ..."

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
```diff
@@ -0,0 +1,217 @@
...
+ * The additional steps taken are to compute the trace-reversed stress energy
+ * tensor associated with the scalar part of the system and add its contribution
+ * to the \f$\partial_t \Pi_{a b}\f$ variable in the Generalized Harmonic
+ * system, as well as adding any scalar sources to the variable \f$\partial_t
+ * \Pi\f$.
+ */
+struct TimeDerivativeTerms {
+  using gh_dt_tags =
+      db::wrap_tags_in<::Tags::dt,
+                       typename gh::System<3_st>::variables_tag::tags_list>;
```
nikwit: I would template this class on Dim, or create a `static constexpr size_t Dim = 3` at the top to make it consistent with the other systems.

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
```diff
@@ -0,0 +1,217 @@
...
+  using scalar_dt_tags = db::wrap_tags_in<
+      ::Tags::dt,
+      typename CurvedScalarWave::System<3_st>::variables_tag::tags_list>;
+  using dt_tags = tmpl::append<gh_dt_tags, scalar_dt_tags>;
+  using scalar_flux_tags = tmpl::transform<
```
nikwit: you are not using this anywhere outside tests. from what I can tell. Also, `CSW` does not have any flux_variables, so I don't understand what you are trying to do here.

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
```diff
@@ -0,0 +1,217 @@
...
+      const tnsr::I<DataVector, 3_st>& shift_scalar,
+      const tnsr::i<DataVector, 3_st>& deriv_lapse,
+      const tnsr::iJ<DataVector, 3_st>& deriv_shift,
+      const tnsr::II<DataVector, 3_st>& upper_spatial_metric,
+      const tnsr::I<DataVector, 3_st>& trace_spatial_christoffel,
+      const Scalar<DataVector>& trace_extrinsic_curvature,
+      const Scalar<DataVector>& gamma1_scalar,
+      const Scalar<DataVector>& gamma2_scalar,
+
+      const Scalar<DataVector>& scalar_source) {
```
nikwit: move the definition to a .cpp file

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
```diff
@@ -0,0 +1,217 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <utility>
```
nikwit: are you using this?

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
```diff
@@ -0,0 +1,259 @@
...
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/MakeWithValue.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TypeTraits/IsA.hpp"
+
+namespace {
+template <typename ComputeVolumeTimeDerivativeTerms, size_t Dim,
+          typename EvolvedTagList, typename FluxTagList, typename TempTagList,
+          typename GradientTagList, typename ArgTagList>
+struct ComputeVolumeTimeDerivativeTermsHelper;
```
nikwit: can you please just call the three systems separately without a helper that is very hard to read? Tests in particular should always be very readable and it doesn't matter if they are a bit longer.

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
```diff
@@ -0,0 +1,259 @@
...
+      [[maybe_unused]] const gsl::not_null<FluxVariables*> volume_fluxes,
+      const gsl::not_null<TemporaryVariables*> temporaries,
+      const GradientVariables& partial_derivs,
+      const ArgumentVariables& time_derivative_args) {
+    ComputeVolumeTimeDerivativeTerms::apply(
+        make_not_null(&get<::Tags::dt<EvolvedTags>>(*dt_vars_ptr))...,
+        make_not_null(&get<FluxTags>(*volume_fluxes))...,
+        make_not_null(&get<TempTags>(*temporaries))...,
+        get<GradientTags>(partial_derivs)..., [](const auto& t) -> const auto& {
+          if constexpr (tt::is_a_v<std::unique_ptr,
```
nikwit: do the systems have any arg tags that are unique pointers?

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
```diff
@@ -0,0 +1,259 @@
...
+  using flux_variables_type =
+      Variables<tmpl::append<gh_flux_tags, scalar_flux_tags>>;
+
+  using gh_temp_tags = ScalarTensor::TimeDerivativeTerms::gh_temp_tags;
+  using scalar_temp_tags =
+      ScalarTensor::TimeDerivativeTerms::scalar_temp_tags;
+  using temp_variables_type = Variables<
+      typename ScalarTensor::TimeDerivativeTerms::temporary_tags>;
+
+  using gh_gradient_tags = tmpl::transform<
```
nikwit: doesn't `using gh_gradient_tags = ScalarTensor::TimeDerivativeTerms::gh_gradient_tags` work fine?

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+  using gh_arg_tags =
+      ScalarTensor::TimeDerivativeTerms::gh_arg_tags;
+  using scalar_arg_tags =
+      ScalarTensor::TimeDerivativeTerms::scalar_arg_tags;
+  using arg_variables_type = tuples::tagged_tuple_from_typelist<
+      tmpl::append<gh_arg_tags, scalar_arg_tags,
+                   tmpl::list<ScalarTensor::Tags::ScalarSource>>>;
+
+  const size_t element_size = 10_st;
```
nikwit: 10 works fine

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
```diff
@@ -0,0 +1,259 @@
...
+
+  temp_variables_type expected_temp_variables{element_size};
+  temp_variables_type temp_variables{element_size};
+
+  const auto gradient_variables =
+      make_with_random_values<gradient_variables_type>(
+          make_not_null(&gen), make_not_null(&dist), DataVector{element_size});
+
+  arg_variables_type arg_variables;
+  tmpl::for_each<tmpl::append<
```
nikwit: It' would be a lot more readable to just remove the `MeshVelocity` from the tag list and create it separately so you don't need all of this.

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
```diff
@@ -0,0 +1,259 @@
...
+      typename flux_variables_type::tags_list,
+      typename temp_variables_type::tags_list,
+      typename gradient_variables_type::tags_list,
+      typename arg_variables_type::tags_list>::apply(
+                              make_not_null(&dt_variables),
+                              make_not_null(&flux_variables),
+                              make_not_null(&temp_variables),
+                              gradient_variables, arg_variables);
+
+  // We compute the trace-reversed stress energy tensor fot the expected
```
nikwit: typo fot

nikwit (changes requested):
nice work, here are some suggestions

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
guilara: Apparently `gh::gauges::GaugeCondition` has a unique pointer

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
guilara: Without the include below this comment I get some error in the test that Time is an incomplete tag. Don't really understand this. It arises from something recently merged in develop it seems.

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
guilara: Changed name

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
guilara: Changed description to just say we compute the rhs of ScalarTensor

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
guilara: Added the `static constexp size_t Dim` param in the declaration, But in the .cpp file I still need to write `3_st` or sth ugly like `TimeDerivative::Dim` (I think), so left it with `3_st` in the .cpp file

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
guilara: Yeah, my implementation of all this is heavily based in GhMhd so thats why there are some of these seemegly strange and useless params sometimes. I removed all the flux arguments from there and from the test

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
guilara: Moved

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
guilara: Seems not. Removed `utility`

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
guilara: As above, I was trying to reuse as much as possible from GhMhd. This was coded up for the `TimeDerivative` test there and it seemed useful. I removed a few things from it that were not really used. 

... I guess I could try to do without it....

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
guilara: Tried it, doesn't seem to work for the Helper

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
guilara: Just 10 :)

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
guilara: Same comment as for the Helper. I guess I could also try to dismantle this...

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
guilara: Fixed

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivative.cpp
```diff
@@ -0,0 +1,223 @@
...
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/DerivSpatialMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/ExtrinsicCurvature.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/SpatialDerivOfLapse.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/SpatialDerivOfShift.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/MakeWithValue.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TypeTraits/IsA.hpp"
+
+#include "Parallel/Printf.hpp"
```
nikwit: remove

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
nikwit: ok, I think you can remove the comment.

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
nikwit: I think you can just write Dim in the cpp file.

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
nikwit: I think the test would be a lot more readable if you called them separately. It might be a bit wordy but should be quick to write... and this way you don't have to wrap your head around these helper and metafunctions when reading a simple test. Just call them similar as its done`CurvedScalarWave/TestTimeDerivative:104`

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivativeTerms.cpp
nikwit: ah right, I misread this. The gradient tags in the system don't actually have the derivative tag attached to it yet and this metafunction does that. fine as is

nikwit (changes requested):
Looks good, I think changing the test would be nice. Also, please do not squash fixup changes directly into your commits unless told otherwise by the reviewer.

guilara:
> Looks good, I think changing the test would be nice. Also, please do not squash fixup changes directly into your commits unless told otherwise by the reviewer.

Ah, sorry. That makes sense

tests/Unit/Evolution/Systems/ScalarTensor/Test_TimeDerivative.cpp
guilara: Removed

src/Evolution/Systems/ScalarTensor/TimeDerivativeTerms.hpp
guilara: Added. That worked

guilara:
I've rewritten the test and removed the Helper at all. It's a bit lengthy.

guilara:
I also removed the `_st`s

nikwit (approved):
Looks good, you can squash the commits together :)

src/Evolution/Systems/ScalarTensor/TimeDerivative.cpp
```diff
@@ -0,0 +1,163 @@
...
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Time/Tags/Time.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Literals.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace ScalarTensor {
+void TimeDerivative::apply(
+    // GH dt variables
+    gsl::not_null<tnsr::aa<DataVector, Dim>*> dt_spacetime_metric,
```
nilsdeppe: `gsl::` -> `const gsl::` (find-replace in the cpp will be easiest probably)

nilsdeppe (changes requested):
LGTM, one minor suggestion. I think you can just squash all the commits into 1, right? I don't need to see a fixup for the suggestion I'm making.

src/Evolution/Systems/ScalarTensor/TimeDerivative.cpp
guilara: I've changed it. Also squashed the commits

guilara:
I've also changed `Dim` to `dim` in `TimeDerivative` since its not a template parameter and to be consistent with #5235

nilsdeppe (approved):

PR #5213, opened 2023-07-19
Add missing sqrt in calculation of W in SpEC ID

ffoucart:
## Proposed changes

The Lorentz factor was calculated as (1-u^2) instead of sqrt(1-u^2) in SpecInitial data; this PR adds the missing sqrt.

### Upgrade instructions

None

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu (approved):
Ugh sorry this was my mistake. This is what happens when we don't test something.. Maybe we could move the `VariablesComputer` out of the SpEC ID class and test it independently (not requesting you to do that). It would need its own file then though..

PR #5215, opened 2023-07-19
Handle relative paths in GenerateXdmf

nilsvu:
## Proposed changes

Writes the correct path relative to the output file when passing data files in subdirectories like `spectre generate-xdmf Run/Volume0.h5 -o Run/Volume.xmf`. Also adds the option to write absolute paths.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/GenerateXdmf.py
+++ b/src/Visualization/Python/GenerateXdmf.py
@@ -272,6 +272,7 @@ def generate_xdmf(
     h5files,
     output: str,
     subfile_name: str,
+    relative_paths: bool = True,
     start_time: Optional[float] = None,
     stop_time: Optional[float] = None,
     stride: int = 1,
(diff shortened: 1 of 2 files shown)


knelli2 (commented):
I feel like absolute paths are safer, no? Obviously if you move the data everything is broken, but why shouldn't I be able to move the xmf file around without worrying that things will break (by default). I guess my question is, why rel paths as default and not absolute paths?

nilsvu:
If you run this on the cluster and then copy to your machine then absolute paths are broken. We've been writing relative paths in XMF files so far.

knelli2:
Hmm that's fair. Why have the absolute paths then?

nilsvu:
🤷‍♂️ I thought why not give the option. As you said, it does enable you to move the XMF file around. Not sure if it's useful, but can't hurt either.

knelli2 (dismissed):
Sure why not

nilsvu:
Fixed the test

knelli2 (approved):

PR #5222, opened 2023-07-21
Changes to get clang 16 to compile

kidder:
## Proposed changes

- Add new path for finding catch.  
- Don't build python documentation unless requested explicitly
- Add explicit instantiations for things that clang didn't produce symbols for

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Utilities/OptimizerHacks.cpp
+++ b/src/Utilities/OptimizerHacks.cpp
@@ -8,3 +8,18 @@ namespace optimizer_hacks {
 void indicate_value_can_be_changed(void* /*variable*/) {}
 }  // namespace optimizer_hacks
 #endif  /* defined(__clang__) && __clang__ < 11 */
+
+#if defined(__clang__) && __clang_major__ >= 15 && defined(__GLIBCXX__)
+#include <string>
+// clang v15+ fails to instantiate the following std library templates if
+// pre-compiled headers are used
+template std::__cxx11::basic_string<char, std::char_traits<char>,
+                                    std::allocator<char>>::pointer
+std::__cxx11::basic_string<char, std::char_traits<char>,
+                           std::allocator<char>>::_M_use_local_data();
+template std::__cxx11::basic_string<wchar_t, std::char_traits<wchar_t>,
+                                    std::allocator<wchar_t>>::pointer
+std::__cxx11::basic_string<wchar_t, std::char_traits<wchar_t>,
+                           std::allocator<wchar_t>>::_M_use_local_data();
+#endif /* defined(__clang__) && __clang_major__ >= 15 && defined(__GLIBCXX__) \
+        */
--- a/cmake/FindCatch.cmake
+++ b/cmake/FindCatch.cmake
@@ -29,7 +29,8 @@ endfunction()
 
 find_path(
   CATCH_INCLUDE_DIR
-  PATH_SUFFIXES single_include include catch catch2 include/catch2
+  PATH_SUFFIXES
+    single_include single_include/catch2 include catch catch2 include/catch2
   NAMES catch.hpp
   HINTS ${CATCH_ROOT}
   DOC "catch include dir"
--- a/cmake/SetupSphinx.cmake
+++ b/cmake/SetupSphinx.cmake
@@ -30,7 +30,7 @@ file(
   DESTINATION ${SPHINX_SOURCE}
 )
 
-add_custom_target(py-docs ALL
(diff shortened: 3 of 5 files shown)


src/Utilities/OptimizerHacks.cpp
```diff
@@ -8,3 +8,16 @@ namespace optimizer_hacks {
 void indicate_value_can_be_changed(void* /*variable*/) {}
 }  // namespace optimizer_hacks
 #endif  /* defined(__clang__) && __clang__ < 11 */
+
+#if defined(__clang__) && __clang_major__ >= 15
+// clang v15+ fails to instantiate the following std library templates if
+// pre-compiled headers are used
+template std::__cxx11::basic_string<char, std::char_traits<char>,
```
wthrowe, line 16: `#include <string>` (Have you tested this with the PCH disabled?)

Might be worth putting this behind some combination of `defined(__GLIBCXX__)` and whether the PCH is enabled.

wthrowe (commented):
I'll test this tomorrow.

wthrowe:
Compiled for me on clang 16 and the tests passed (or rather, no more tests failed than before).

kidder:
okay, pushed fixup add the include for string.  It compiles for me with both USE_PCH set to ON or OFF.  Is your comment about `__GLIBCXX__` related to using `USE_LIBCXX=ON` ?

wthrowe:
Related, but I don't think `USE_LIBCXX=OFF` necessarily means we're not using libc++.  We're not turning it on in that case, but it might be enabled for some other reason (user-specified flags or the compiler default).  I think `__GLIBCXX__` should reliably indicate libstdc++.

kidder:
okay, so you would like me to add `&& defined(__GLIBCXX__)`  to line 12

wthrowe:
Yes, that sounds good.

wthrowe (approved):

nilsvu:
@kidder did you need this just for AppleClang? The explicit instantiations fail to compile with clang 15 on CI:

```
../../../src/Utilities/OptimizerHacks.cpp:19:51: error: explicit instantiation of '_M_use_local_data' does not refer to a function template, variable template, member function, member class, or static data member
                           std::allocator<char>>::_M_use_local_data();
                                                  ^
../../../src/Utilities/OptimizerHacks.cpp:23:54: error: explicit instantiation of '_M_use_local_data' does not refer to a function template, variable template, member function, member class, or static data member
                           std::allocator<wchar_t>>::_M_use_local_data();
                                                     ^
```

knelli2:
Locally with clang-16, I also found that these explicit instantiations failed to compile.

nilsvu:
How did you fix it?

knelli2:
Building with clang-14 ;)

But to use clang-16 I found this worked
```diff
diff --git a/src/Utilities/OptimizerHacks.cpp b/src/Utilities/OptimizerHacks.cpp
index 1732b4a118..443e2dbd11 100644
--- a/src/Utilities/OptimizerHacks.cpp
+++ b/src/Utilities/OptimizerHacks.cpp
@@ -9,7 +9,7 @@ void indicate_value_can_be_changed(void* /*variable*/) {}
 }  // namespace optimizer_hacks
 #endif  /* defined(__clang__) && __clang__ < 11 */

-#if defined(__clang__) && __clang_major__ >= 15 && defined(__GLIBCXX__)
+#if defined(__clang__) && __clang_major__ == 15 && defined(__GLIBCXX__)
 #include <string>
 // clang v15+ fails to instantiate the following std library templates if
 // pre-compiled headers are used
```

nilsvu:
But @kidder needed this for clang 16?

kidder:
Yes, on my linux desktop; @wthrowe also successfully compiles with clang 15 on his linux desktop with this change

nilsvu:
Ubuntu 22.04 in the container fails 🤷‍♂️

wthrowe:
Looks like a different c++ standard library implementation than the fix was for.  What gcc version are you using?

nilsvu:
GCC 11.4.0

wthrowe:
Looks like this implementation changed in gcc 12, so we need to only apply it for newer libstdc++ versions.  Unfortunately, there is apparently no macro that indicates what version you have.

wthrowe:
Actually, apparently they finally added one in version 7: `_GLIBCXX_RELEASE`

PR #5227, opened 2023-07-22
Change observation value handling, allow any observation on failure

wthrowe:
This is primarily in preparation for observing on substeps, but generalizing the failure observation handling came almost for free.

There is a preexisting fixup commit to help reviewers.  It will be squashed into the preceding commit before merging.

### Upgrade instructions

The template for the type of observation value as been moved from Events to the actions that run them, so observation events that previously had an `OverrideObservationValue` option no longer do, and such behavior is handled at a higher level.  When used in `EventsAndTriggers` or `EventsAndDenseTriggers`, the behavior is as if `None` had been specified.  When used in `EventsRunAtCleanup`, the value is taken from the new, higher-level `ObservationValue` option.

In addition to removing any `OverrideObservationValue` options, any existing `EventsRunAtCleanup` section of input files should be changed to
```yaml
EventsRunAtCleanup:
  ObservationValue: <value, -1000.0 is used in our examples>
  Events:
    <list of events>
```
This is required even if the list of events is empty.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Event.hpp
+++ b/src/ControlSystem/Event.hpp
@@ -92,7 +92,8 @@ class Event : public ::Event {
   void operator()(const db::DataBox<DbTags>& box,
                   Parallel::GlobalCache<Metavariables>& cache,
                   const ArrayIndex& array_index,
-                  const Component* const component) const {
+                  const Component* const component,
+                  const ObservationValue& /*observation_value*/) const {
     const LinkedMessageId<double> measurement_id{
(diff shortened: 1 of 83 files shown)


nilsdeppe (commented):
LGTM, this is much easier to use than having to specify `None` for the value in a bunch of places! I'm happy for you to squash whenever

nilsdeppe (approved):

PR #5229, opened 2023-07-24
Add temperature to GRMHD related initial data

ermost:
Adds temperature to GRMHD initial data.
This is largely done through inherited member functions that call the equation of state.

Fishbone Moncrief seems to be missing a tag declaration at the moment.


isaaclegred:
@knelli2 has added the FM tag problem as an issue #5237

nilsdeppe:
Sorry, it looks like this conflicts with my changes to the disk ID to get it compiling. I'm happy to help with rebasing if there are any issues!

src/PointwiseFunctions/Hydro/Temperature.hpp
```diff
@@ -0,0 +1,41 @@
...
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "PointwiseFunctions/Hydro/TagsDeclarations.hpp"
+
+namespace hydro {
+
+template <typename DerivedSolution>
+class TemperatureInitialization {
```
nilsdeppe, line 18: Could you add a doxygen comment just so the class shows up in the documentation?

nilsdeppe (commented):
I'm happy with this implementation, so I think just the rebase and the compiler warning. I'm not really sure how to test this. Maybe something like
```cpp
template <typename EosType, size_t Dim>
class DummySolution {
 public:
  DummySolution(EosType eos) : eos_(std::move(eos)) {}

  template <typename DataType>
  tuples::TaggedTuple<hydro::Tags::RestMassDensity<DataType>>
  variables(const tnsr::I<DataType, Dim>& coords, 
                   tmpl::list<hydro::Tags::RestMassDensity<DataType>> /*meta*/) const {
    return {Scalar<DataType>(get<0>(coords).size(), 1.28e-3)};
  }

  // Additional ones for epsilon and Y_e

 private:
  EosType eos_{};
};

template <size_t Dim>
void test() {
  Polytrope polytrope{};
  DummySolution<Polytrope, Dim> solution{polytrope};
  const tnsr::I<DataVector, Dim> coords_dv{10, 3.0};
  const tnsr::I<double, Dim> coords_double{13.0};

  CHECK(get<hydro::Tags::Temperature<DataVector>>(solution.variables(coords_dv, tmpl::list<hydro::Tags::Temperature<DataVector>>)) == polytrope.temperature_from_density(Scalar<DataVector>{10, 1.28e-3));
  CHECK(get<hydro::Tags::Temperature<double>>(solution.variables(coords_double, tmpl::list<hydro::Tags::Temperature<double>>)) == polytrope.temperature_from_density(Scalar<double>{10, 1.28e-3));
  // TODO: maybe for ideal fluid too? Could do the table using the low-resolution table stored?
}
SPECTRE_TEST_CASE("...", "...") {
  test<1>();
  test<2>();
  test<3>();
}
```

nilsdeppe (dismissed):

src/PointwiseFunctions/AnalyticSolutions/GrMhd/BondiMichel.cpp
```diff
@@ -203,6 +203,14 @@ tuples::TaggedTuple<hydro::Tags::Pressure<DataType>> BondiMichel::variables(
                pow(vars.rest_mass_density, polytropic_exponent_)}}};
 }
 
+template <typename DataType>
```
nilsdeppe, line 206: From CI it looks like this needs an explicit instantiation at the end of this file.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/FishboneMoncriefDisk.cpp
```diff
@@ -237,6 +237,27 @@ FishboneMoncriefDisk::variables(
   return {std::move(specific_internal_energy)};
 }
 
+template <typename DataType, bool NeedSpacetime>
+tuples::TaggedTuple<hydro::Tags::Temperature<DataType>>
+FishboneMoncriefDisk::variables(
```
nilsdeppe, line 242: From CI it looks like this needs an explicit instantiation at the end of this file.

ermost:
not sure why it says "dismissed review" here... ? :)

there may be a deeper issue with modifying the tags. 
for the unit tests that compile, it seems that some of the ones for subcell are failing.

I need to investigate whether this is just because of the test or the initial data.
(There also seems to be a compiler depending issue clang is picking up on.)

nilsdeppe:
We just set up it up to dismiss reviews on push (a bit paranoid on the "don't let anyone sneak code in" side).

Let me know if you want/need any input from me!

src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
```diff
@@ -148,6 +149,15 @@ bool PrimitiveFromConservative<OrderedListOfPrimitiveRecoverySchemes,
       }
       get(*lorentz_factor)[s] = primitive_data.value().lorentz_factor;
       get(*pressure)[s] = primitive_data.value().pressure;
+
+      get(*specific_enthalpy)[s] = primitive_data.value().rho_h_w_squared /
+                                   (primitive_data.value().lorentz_factor *
```
nilsdeppe, line 196: should this divide out the Lorentz factor squared? I'm interpreting `rho_h_w_squared` as being $\rho h W^2$.

src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
```diff
@@ -148,6 +149,15 @@ bool PrimitiveFromConservative<OrderedListOfPrimitiveRecoverySchemes,
...
+                                    primitive_data.value().rest_mass_density);
+      get(*specific_internal_energy)[s] =
+          get(*specific_enthalpy)[s] - 1. -
```
nilsdeppe: Floating point wise, it might be better to do $\rho h - \rho - p$ but :shrug: Both seem like they'll go poorly in the low-density limit. The only advantage is not dividing by $\rho$, something we've talked about trying to eliminate since in general it seems like $\rho h$ is what shows up, not $h$ as much

nilsdeppe (changes requested):
LGTM! A couple small suggestions. I've attached the patch the fixes the linking errors on my machine. It's small, but I figured a patch is easier to share. 
[id_temp.patch](https://github.com/sxs-collaboration/spectre/files/12409946/id_temp.patch)

src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
ermost: yep! fixed it.

src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
ermost: I multiplied by rho.

One way might be to directly extract eps and T from the inversion, i.e., include in them in primitive_data where they will be computed most accurately.

nilsdeppe: Ah yea, that's a good idea! For another PR, but that should help a lot. I've been worried about lower-density limits because of this sort of math in places.

nilsdeppe (dismissed):

nilsdeppe:
Here's a patch that should fix the test failures.
[id_temp.patch](https://github.com/sxs-collaboration/spectre/files/12414384/id_temp.patch)

ermost:
Thank you, @nilsdeppe! 
I added some small fixUps that came up in CI.

nilsdeppe (commented):
LGTM! Feel free to squash whenever :)

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/SmoothFlow.hpp
```diff
@@ -92,6 +94,15 @@ class SmoothFlow : public evolution::initial_data::InitialData,
...
   using smooth_flow::variables;
 
+  template <typename DataType>
+  auto variables(const tnsr::I<DataType, Dim>& x, double t,
+                 tmpl::list<hydro::Tags::Temperature<DataType>> /*meta*/) const
+      -> tuples::TaggedTuple<hydro::Tags::Temperature<DataType>> {
+    return hydro::TemperatureInitialization<SmoothFlow<Dim>>::variables(
+        x, t, tmpl::list<hydro::Tags::Temperature<DataType>>{});
+  }
+  /// @}
```
nilsdeppe: CI documentation is complaining about this line. Just remove it :)

nilsdeppe (commented):
:tada: is works!!! Please rebase & squash, I just point out what the documentation check is unhappy about, hopefully that makes your life a bit easier :) Thank you so much for doing this :D

Edit: The clang-tidy failures are all in Boost TOMS748, which we just ignore because we can't change that file :(

ermost:
Thank you for your help with this! :)

ermost:
Rebased and squashed.

nilsdeppe (approved):
You're welcome! :)

PR #5240, opened 2023-07-25
Merge MutableGlobalCache into GlobalCache

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/GlobalCache.hpp
+++ b/src/Parallel/GlobalCache.hpp
@@ -8,11 +8,13 @@
 
 #include <charm++.h>
 #include <memory>
+#include <mutex>
 #include <numeric>
 #include <optional>
 #include <pup.h>
 #include <string>
 #include <tuple>
+#include <utility>
 #include <vector>
 
 #include "DataStructures/DataBox/Tag.hpp"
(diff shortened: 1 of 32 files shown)


src/Parallel/GlobalCache.hpp
```diff
@@ -634,16 +440,48 @@ template <typename GlobalCacheTag, typename Function, typename... Args>
...
+    callbacks = std::move(std::get<1>(tuples::get<tag>(mutable_global_cache_)));
+    std::get<1>(tuples::get<tag>(mutable_global_cache_)).clear();
+    std::get<1>(tuples::get<tag>(mutable_global_cache_)).shrink_to_fit();
+  }
+
+  // Invoke the callbacks.  Any new callbacks that are added to the
+  // list (if a callback calls mutable_cache_item_is_ready) will be
+  // saved and will not be invoked here.
+  for (auto& callback : callbacks) {
+    callback->invoke();
```
wthrowe, line 539: I think you need to hold the write lock for this entire function.  Here's the potential problem I see:
* We have two writers, A and B preparing to make independent changes to a cache object.
* We have an element E with a callback waiting for the change B is going to make.

Suppose the following sequence of events:
1. A mutates the object, copies the callback list, and starts the callback for element E.
2. E checks the current value and determines it is not ready.
3. B mutates the object, copies the empty callback list, and returns with nothing to do.
4. E returns a new callback, which is added to the callback list.
5. A returns.

We now have E waiting on a change that has already happened.

src/Parallel/GlobalCache.hpp
```diff
@@ -634,16 +440,48 @@ template <typename GlobalCacheTag, typename Function, typename... Args>
...
+  std::apply(
+      [this](const auto&... local_args) {
+        // First mutex is for value of mutable tag
+        std::mutex& mutex = tuples::get<MutexTag<tag>>(mutexes_).first;
+        // Scoped for lock guard
+        {
+          const std::lock_guard<std::mutex> lock(mutex);
+          Function::apply(make_not_null(&(std::get<0>(
+                              tuples::get<tag>(mutable_global_cache_)))),
+                          local_args...);
```
wthrowe, line 511: Should we dereference `unique_ptr`s here?  I guess not doing so is consistent with `db::mutate`, but you can't actually change what is being pointed to since that would violate the always-readable guarantee.

src/Parallel/GlobalCache.hpp
```diff
@@ -865,17 +667,21 @@ auto get_parallel_component(const GlobalCache<Metavariables>& cache)
...
-          *Parallel::local_branch(cache.mutable_global_cache_proxy_);
-      return local_mutable_cache.template get<GlobalCacheTag>();
+  using tag =
+      tmpl::conditional_t<is_mutable, MutableCacheTag<tmp_tag>, tmp_tag>;
+  if constexpr (is_mutable) {
+    // Tag is not in the const tags, so use mutable_global_cache_. No locks here
+    // because we require all mutable tags to be able to be read at all times
+    // (even when being written to)
+    if constexpr (tt::is_a_v<std::unique_ptr, typename tag::tag::type>) {
+      return *std::get<0>((tuples::get<tag>(cache.mutable_global_cache_)));
```
wthrowe: Nit: extra parens compared to two lines down.

tests/Unit/Parallel/Test_GlobalCache.cpp
```diff
@@ -562,28 +551,14 @@ void Test_GlobalCache<Metavariables>::exit_if_done(int index) {
...
 
 PUPable_def(UseCkCallbackAsCallback)
-// clang-tidy: possibly throwing constructor static storage
-// clang-tidy: false positive: redundant declaration
-PUP::able::PUP_ID Triangle::my_PUP_ID = 0;   // NOLINT
-PUP::able::PUP_ID Square::my_PUP_ID = 0;     // NOLINT
-PUP::able::PUP_ID Arthropod::my_PUP_ID = 0;  // NOLINT
+    // clang-tidy: possibly throwing constructor static storage
+    // clang-tidy: false positive: redundant declaration
+    PUP::able::PUP_ID Triangle::my_PUP_ID = 0;  // NOLINT
```
wthrowe: clang-format is dumb.  Don't let it make the formatting worse.

tests/Unit/Parallel/Test_GlobalCache.cpp
```diff
@@ -228,13 +228,24 @@ class UseCkCallbackAsCallback : public Parallel::Callback {
...
+// We have to define our expected mutable values statically out here because the
+// mutable part of the cache is once per node and we are only running on one
+// node. This means that when we run `run_test_one` on the first array element
+// and it mutates a mutable tag, that tag gets immediately changed inside the
+// mutable part of the cache. Then when the second array element runs
+// `run_test_one` this value has already been mutated, so we need to keep track
+// of these changes outside the functions with these static vars.
+static double expected_weight = 160.0;
+static std::string expected_email = "joe@somewhere.com";
+static size_t expected_number_of_legs = 6;
```
wthrowe: Can you explain in a github comment what is different from the old behavior that requires this?

tests/Unit/Parallel/Test_GlobalCache.cpp
```diff
@@ -290,18 +303,23 @@ void TestArrayChare<Metavariables>::run_test_two() {
           *Parallel::local_branch(global_cache_proxy_),
```
wthrowe: Comment a few lines up still refers to "150".  Haven't looked for other similar things.

wthrowe (changes requested):

src/Parallel/GlobalCacheDeclare.hpp
```diff
@@ -2,9 +2,7 @@
 // See LICENSE.txt for details.
 
 /// \file
-/// Forward-declares CProxy_GlobalCache which MutableGlobalCache needs, but
-/// GlobalCache is defined after MutableGlobalCache. Also forward declares
-/// ResourceInfo which the GlobalCache has an entry method for.
+/// Forward-declares ResourceInfo which the GlobalCache has an entry method for.
```
kidder, line 5: is the forward declaration of CProxy_GlobalCache still needed?

src/Parallel/Main.hpp
```diff
@@ -479,10 +470,9 @@ Main<Metavariables>::Main(CkArgMsg* msg) {
 
   // Now that the singleton map has been built, set the resource info in the
   // GlobalCache (if the tags exist). Since this info will be constant
-  // throughout a simulation, we opt for directly editing the tags in the const
+  // throughout a simulation, we opt for directly editing a const tag in the
   // GlobalCache before we pass it to any other parallel component rather than
-  // having the tags in the MutableGlobalCache and using a mutate call to set
-  // them.
+  // having the a mutable tag and using a mutate call to set them.
```
kidder: the a -> a

src/ParallelAlgorithms/Interpolation/InterpolationTarget.hpp
```diff
@@ -145,9 +145,9 @@ namespace intrp {
 ///   }
 /// } else {
 ///   set VerifyTemporalIdsAndSendPoints as a callback for the
-///   `domain::Tags::FunctionsOfTime` in MutableGlobalCache, so that
-///   once `FunctionsOfTime` are updated, VerifyTemporalIdsAndSendPoints
-///   is called.
+///   `domain::Tags::FunctionsOfTime` in mutable part of the GlobalCache, so
```
kidder: in the mutable

src/ParallelAlgorithms/Interpolation/InterpolationTarget.hpp
```diff
@@ -251,10 +251,8 @@ namespace intrp {
 /// and then the InterpolationTarget invokes another Action to send points
 /// to only those `Elements` that contain the points; this alternative
 /// uses less memory but much more communication. Another alternative would
-/// be to place the points in the MutableGlobalCache (so that there is one
-/// copy per core, rather than one copy per Element), or even in the
-/// GlobalCache (one copy per node) since the points need be computed only
-/// once.
+/// be to place the points in  the GlobalCache (one copy per node) since the
```
kidder: remove extra space between in and the

kidder (changes requested):

src/Parallel/GlobalCache.hpp
```diff
@@ -620,12 +402,36 @@ template <typename Metavariables>
...
-    return mutable_global_cache_
-        ->template mutable_cache_item_is_ready<GlobalCacheTag>(function);
+    optional_callback =
+        function(std::get<0>(tuples::get<tag>(mutable_global_cache_)));
+  }
+  if (optional_callback != nullptr) {
+    // Second mutex is for vector of callbacks
+    std::mutex& mutex = tuples::get<MutexTag<tag>>(mutexes_).second;
+    {
+      // Scoped for lock guard
```
nilsdeppe, line 434: Since this is all inside an `if` why not do:
```cpp
const std::lock_guard<std::mutex> lock(tuples::get<MutexTag<tag>>(mutexes_).second);
...
```
and avoid the extra indentation?

src/Parallel/GlobalCache.hpp
nilsdeppe: I'd be in favor of dereferences so that people can't change things they shouldn't be able to

nilsdeppe (commented):
LGTM, couple minor suggestions.

src/Parallel/GlobalCache.hpp
knelli2: Does the `return` affect this? Meaning that other destructors have to be called in addition to the lock guard destructor? Which order are they called and could that affect performance?

nilsdeppe: destructors are called in the opposite order of constructors, and are called before returning from the function. In terms of performance, there won't be a difference between:
```cpp
}
return blah;
```
and
```cpp
return blah;
```
The machine code should be identical in those two cases.

Does that answer your questions?

tests/Unit/Parallel/Test_GlobalCache.cpp
knelli2: So I originally left this part of the test alone, but found that it failed. Here is the flow of the `run_test_one` function:

1. Check value of tag is expected *starting* value in `run_test_one`
2. Mutate tag in `run_test_one`
3. Immediately call `run_test_two` inlined.

Now this is run on every element of the array parallel component (4 elements). What this means is that the mutable tag will be edited 4 times because `run_test_one` is run on all array elements (not a big deal because it's being mutated to the same value). It appears that in the old test, possibly because the mutate went through a proxy to the const cache and then a proxy to the mutable cache, the actual mutate call wasn't run until all 4 `run_test_one` functions were called. This means that step 1., the check that the tag is its *starting* value, always passed. However, now with my changes the mutate only goes through one proxy call to the cache. Apparently this caused the actual mutate to be run before all 4 of the `run_test_one` functions could run, so step 1. failed in at least one `run_test_one` because the tag is no longer at its starting value.

That's why I added the static variables. To keep track of the actual current state of the tags so that they can be compared in all `run_test_X` whenever they are run. I think the test before just got lucky that that's how the execution order worked out.

src/Parallel/GlobalCache.hpp
knelli2: Yup! Thank you!

tests/Unit/Parallel/Test_GlobalCache.cpp
wthrowe: I think that breaks a lot of the purpose of the test.  All the `mutable_cache_item_is_ready` calls are trivially being satisfied if everything is inline, right?  I don't think this is testing registering and calling callbacks anymore.

It might be necessary to have the different threads do different things to make them they actually have to wait on one another.

knelli2: What if I leave the functions as they originally were except I remove the calls to the other functions inside them. Then I call each one on a different element? Like this:

```cpp
array_proxy[3].run_test_five();
array_proxy[2].run_test_four();
array_proxy[1].run_test_three();
array_proxy[0].run_test_two();
array_proxy[3].run_test_one();
```

Since `run_test_one` is what does the mutate that kicks everything off, a bunch of callbacks should be registered and called.

wthrowe: Yes, that sounds good.  I don't see any reason they have to be chained.

knelli2:
@nilsdeppe @wthrowe @kidder Rebased and added a fixup commit. Also changed this from a draft to a real pull request because I believe we have the thread-safety things figured out.

nilsdeppe (commented):
I'm happy. No need to have me look again before merging!

tests/Unit/Parallel/Test_GlobalCache.cpp
```diff
@@ -220,32 +220,25 @@ struct TestMetavariables {
...
   void invoke() override { callback_.send(nullptr); }
-  void pup(PUP::er& /*p*/) override {
-    ERROR(
-        "Should not be pupping a UseCkCallbackAsCallback, since it is always "
-        "stored in the local GlobalCache");
+  void pup(PUP::er& p) override {
+    p | callback_;
+    // ERROR(
+    //     "Should not be pupping a UseCkCallbackAsCallback, since it is always
+    //     " "stored in the local GlobalCache");
```
wthrowe: Remove?

wthrowe (changes requested):
Otherwise looks good.

src/Parallel/GlobalCacheDeclare.hpp
knelli2: Yes I got errors if I didn't have it and didn't bother trying to track things down.

knelli2:
Had to fix just a couple more things since #5065 got merged (just `Test_Evolution.cpp` and `Test_DgDomain.cpp`)

wthrowe (approved):

kidder (approved):

PR #5235, opened 2023-07-25
Add boundary conditions for ScalarTensor

guilara:
## Proposed changes

Add product of boundary conditions for the ScalarTensor system. For now, only we allow for Ghost type boundary conditions for both systems and `demand_outgoing_char_speeds` for excision boundaries.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5243, opened 2023-07-26
Add Dirichlet boundary condition for CurvedScalarWave

guilara:
## Proposed changes

I propose to add this simple boundary condition (BC) for CurvedScalarWave (CSW) in order to be able to write a fully Ghost type boundary condition for the ScalarTensor system (#5235). This is because they are the simplest conditions to implement in the combined system and also because the CSW constraint preserving BCs may not trivially generalize to the ScalarTensor system (where the metric is evolving as well).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/AnalyticConstant.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/AnalyticConstant.cpp
@@ -1,0 +1,86 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/CurvedScalarWave/BoundaryConditions/AnalyticConstant.hpp"
+
+#include <cstddef>
+#include <memory>
+#include <optional>
+#include <pup.h>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/DotProduct.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Characteristics.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/MakeString.hpp"
+
+namespace CurvedScalarWave::BoundaryConditions {
+
+template <size_t Dim>
+std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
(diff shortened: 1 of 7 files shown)


src/Evolution/Systems/CurvedScalarWave/BoundaryConditions/AnalyticConstant.cpp
```diff
@@ -0,0 +1,87 @@
...
+  BoundaryCondition<Dim>::pup(p);
+  p | amplitude_;
+}
+
+template <size_t Dim>
+AnalyticConstant<Dim>::AnalyticConstant(CkMigrateMessage* const msg)
+    : BoundaryCondition<Dim>(msg) {}
+
+template <size_t Dim>
+AnalyticConstant<Dim>::AnalyticConstant(const double amplitude) {
```
nilsdeppe: You should prefer list initialization whenever possible:
```cpp
AnalyticConstant<Dim>::AnalyticConstant(const double amplitude) : amplitude_(amplitude) {}
```

tests/Unit/Evolution/Systems/CurvedScalarWave/BoundaryConditions/Test_AnalyticConstant.cpp
```diff
@@ -0,0 +1,66 @@
...
+          helpers::Tags::PythonFunctionName<
+              CurvedScalarWave::Tags::ConstraintGamma1>,
+          helpers::Tags::PythonFunctionName<
+              CurvedScalarWave::Tags::ConstraintGamma2>>{
+          "error", "psi_analytic_constant", "pi_analytic_constant",
+          "phi_analytic_constant", "lapse_analytic_constant",
+          "shift_analytic_constant", "constraint_gamma1_analytic_constant",
+          "constraint_gamma2_analytic_constant"},
+      "AnalyticConstant:\n"
+      "  Amplitude: 0.0\n",
```
nilsdeppe: what about setting the Amplitude to some non-zero & non-1 number, e.g. `2.3` and then just hard-code the same number in the python file

nilsdeppe (changes requested):
LGTM, please squash the minor suggestions immediately

guilara:
@nilsdeppe Thanks for the review. I've implemented the suggestions and rebased

nilsdeppe (approved):

PR #5251, opened 2023-07-27
Fix commit hook for older clang-format versions

### LINKED ISSUE #5249: git-clang-format `--diffstat` option added in clang-14
Versions of clang before 14 don't have the `--diffstat` option to `git-clang-format` (https://reviews.llvm.org/D84375, version tags listed [here](https://github.com/llvm/llvm-project/commit/191a395343b984d26d058c449cefe1a9ac0ff927#diff-b36a1ef270f7220995ade6ac916bed21677f2d5e3c5d1131052bba09e2ab92d3L506)), yet we support all the way back to clang-10. Clang-10 is what ships with ubuntu 20.04 and so by default our git hooks are broken on that (this is what I use).

nilsvu:
## Proposed changes

Fixes #5249 . @knelli2 could you confirm?

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/Hooks/pre-commit.sh
+++ b/tools/Hooks/pre-commit.sh
@@ -36,19 +36,20 @@ printf '%s\0' "${commit_files[@]}" | \
 # Use git-clang-format to check for any suspicious formatting of code.
 @CLANG_FORMAT_BIN@ --version > /dev/null
 if [ $? -eq 0 ]; then
-    clang_format_diffstat=$(@GIT_EXECUTABLE@ --no-pager \
-        clang-format --binary @CLANG_FORMAT_BIN@ --diffstat --quiet)
+    clang_format_diff=$(@GIT_EXECUTABLE@ --no-pager \
+        clang-format --binary @CLANG_FORMAT_BIN@ --diff --quiet)
     # Clang-format didn't always return the right exit code before version 15,
-    # so we check the diffstat output instead (see issue:
+    # so we check the diff output instead (see issue:
     # https://github.com/llvm/llvm-project/issues/54758)
-    if [ -n "$clang_format_diffstat" ]; then
-        @GIT_EXECUTABLE@ clang-format --binary @CLANG_FORMAT_BIN@ --diff \
-            > @CMAKE_SOURCE_DIR@/.clang_format_diff.patch
-        echo "Found C++ formatting errors:"
-        echo "$clang_format_diffstat"
+    if [ -n "$clang_format_diff" ] \
+        && [[ "$clang_format_diff" != *"no modified files to format"* ]] \
+        && [[ "$clang_format_diff" != *"did not modify any files"* ]]; then
+        echo "$clang_format_diff" > @CMAKE_SOURCE_DIR@/.clang_format_diff.patch
+        echo "Found C++ formatting errors."
         echo "Please run 'git clang-format' in the repository."
         echo "You can also apply the patch directly:"
         echo "git apply @CMAKE_SOURCE_DIR@/.clang_format_diff.patch"
+        echo ""
     fi
 fi
 


tools/Hooks/pre-commit.sh
```diff
@@ -36,17 +36,14 @@ printf '%s\0' "${commit_files[@]}" | \
 # Use git-clang-format to check for any suspicious formatting of code.
 @CLANG_FORMAT_BIN@ --version > /dev/null
 if [ $? -eq 0 ]; then
-    clang_format_diffstat=$(@GIT_EXECUTABLE@ --no-pager \
-        clang-format --binary @CLANG_FORMAT_BIN@ --diffstat --quiet)
+    clang_format_diff=$(@GIT_EXECUTABLE@ --no-pager \
+        clang-format --binary @CLANG_FORMAT_BIN@ --diff --quiet)
```
knelli2, line 40: Even with `--quiet`, if there are no changes this returns

```
no modified files to format
```

tools/Hooks/pre-commit.sh
```diff
@@ -36,17 +36,14 @@ printf '%s\0' "${commit_files[@]}" | \
...
+    # so we check the diff output instead (see issue:
     # https://github.com/llvm/llvm-project/issues/54758)
-    if [ -n "$clang_format_diffstat" ]; then
-        found_error=1
-        @GIT_EXECUTABLE@ clang-format --binary @CLANG_FORMAT_BIN@ --diff \
-            > @CMAKE_SOURCE_DIR@/.clang_format_diff.patch
-        echo "Found C++ formatting errors:"
-        echo "$clang_format_diffstat"
+    if [ -n "$clang_format_diff" ]; then
+        echo $clang_format_diff > @CMAKE_SOURCE_DIR@/.clang_format_diff.patch
```
knelli2: Need quotes around `$clang_format_diff` to preserve newlines and have the patch file formatted properly

knelli2 (changes requested):

tools/Hooks/pre-commit.sh
nilsvu: Ugh this works with clang-format v15

knelli2: Annoying that git-clang-format changed their behavior of `--quiet`

nilsvu:
Does this work with clang-10 now @knelli2?

tools/Hooks/pre-commit.sh
```diff
@@ -36,17 +36,16 @@ printf '%s\0' "${commit_files[@]}" | \
...
-        echo "Found C++ formatting errors:"
-        echo "$clang_format_diffstat"
+    if [ -n "$clang_format_diff" ] \
+        && [[ "$clang_format_diff" != *"no modified files to format"* ]] \
+        && [[ "$clang_format_diff" != *"did not modify any files"* ]]; then
+        echo "$clang_format_diff" > @CMAKE_SOURCE_DIR@/.clang_format_diff.patch
+        echo "Found C++ formatting errors."
         echo "Please run 'git clang-format' in the repository."
         echo "You can also apply the patch directly:"
         echo "git apply @CMAKE_SOURCE_DIR@/.clang_format_diff.patch"
```
knelli2, line 51: Can you add an `echo` after this last line so it separates the formatting error message from the actual commit message? Makes it easier to read.

knelli2 (changes requested):
Yes now this works, just one small thing.

tools/Hooks/pre-commit.sh
nilsvu: ok done

knelli2 (approved):

PR #5252, opened 2023-07-27
Fix GH harmonic gauge FPEs

nilsdeppe:
## Proposed changes

The constraint-preserving BCs need H_a and d_a H_b. We could propagate whether we are using harmonic gauge through, but that's a pretty big change I don't have time for right now. This will still elide _most_ of the work for (d_b) H_a=0, except actually setting the numerical values to zero. The boundary condition could be optimized separately.

As-is, I believe all GH simulations that use harmonic gauge and constraint preserving BCs are broken in develop.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
@@ -268,19 +268,6 @@ void TimeDerivative<Dim>::apply(
     for (size_t nu = 0; nu < Dim + 1; ++nu) {
       gauge_constraint->get(nu) += gauge_function->get(nu);
     }
-  } else {
-#ifdef SPECTRE_DEBUG
-    get(*sqrt_det_spatial_metric) =
-        std::numeric_limits<double>::signaling_NaN();
-    for (size_t mu = 0; mu < Dim + 1; ++mu) {
-      for (size_t nu = 0; nu < Dim + 1; ++nu) {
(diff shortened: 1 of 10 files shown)


src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
```diff
@@ -268,19 +267,6 @@ void TimeDerivative<Dim>::apply(
     for (size_t nu = 0; nu < Dim + 1; ++nu) {
       gauge_constraint->get(nu) += gauge_function->get(nu);
     }
-  } else {
-#ifdef SPECTRE_DEBUG
-    get(*sqrt_det_spatial_metric) =
```
knelli2, line 273: I think this block should be removed in the next commit, not this one. (this commit, at least from the name, doesn't sound like it should change functionality)

knelli2 (changes requested):

src/Evolution/Systems/GeneralizedHarmonic/TimeDerivative.cpp
nilsdeppe: I just looked at the parent code and it's different than I expected. I'll push a commit that does a couple things or drop this change. I was expecting the parent code in `ComputeTimeDerivative.hpp` to set the default value to NaN, but it doesn't right now, it just leaves it uninitialized.

knelli2 (approved):
Ok looks good.

nikwit (approved):

PR #5254, opened 2023-07-27
Add Python Bindings For Spectral And Enthalpy EoS

ctrandrsn:
## Proposed changes

Add python bindings for enthalpy and spectral equations of state.  We add this so we can create simulation domains and for post-processing.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/Hydro/EquationsOfState/Python/Enthalpy.cpp
+++ b/src/PointwiseFunctions/Hydro/EquationsOfState/Python/Enthalpy.cpp
@@ -1,0 +1,45 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "PointwiseFunctions/Hydro/EquationsOfState/Python/Enthalpy.hpp"
+
+#include <pybind11/numpy.h>
+#include <pybind11/operators.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <string>
+#include <vector>
+
+#include "PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.hpp"
+#include "PointwiseFunctions/Hydro/EquationsOfState/PolytropicFluid.hpp"
+#include "PointwiseFunctions/Hydro/EquationsOfState/Spectral.hpp"
+// low_density_eos uses Spectral and Polytropic Fluid type EoS.
+
+namespace py = pybind11;
+
+namespace EquationsOfState::py_bindings {
+
+template <typename LowDensityEoS>
+// NOLINTNEXTLINE(google-runtime-references)
+void bind_enthalpy_impl(py::module& m, const std::string& name) {
+  py::class_<Enthalpy<LowDensityEoS>, EquationOfState<true, 1>>(
+    m, name.c_str()).def(
+      py::init<double, double, double, double, std::vector<double>,
+               std::vector<double>, std::vector<double>, LowDensityEoS,
+               double>(),
+      py::arg("reference_density"), py::arg("max_density"),
+      py::arg("min_density"), py::arg("trig_scale"),
+      py::arg("polynomial_coefficients"), py::arg("sin_coefficients"),
+      py::arg("cos_coefficients"), py::arg("low_density_eos"),
+      py::arg("transition_delta_epsilon"));
+}
+
+void bind_enthalpy(py::module& m) {
+  bind_enthalpy_impl<PolytropicFluid<true>>(m, "EnthalpyPolytropicFluid");
+  bind_enthalpy_impl<Spectral>(m, "EnthalpySpectral");
+  bind_enthalpy_impl<Enthalpy<Spectral>>(m, "EnthalpyEnthalpySpectral");
+  bind_enthalpy_impl<Enthalpy<Enthalpy<Spectral>>>(
+      m, "EnthalpyEnthalpyEnthalpySpectral");
+}
+
+}  // namespace EquationsOfState::py_bindings
--- a/tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/CMakeLists.txt
+++ b/tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/CMakeLists.txt
@@ -2,6 +2,12 @@
 # See LICENSE.txt for details.
 
 spectre_add_python_bindings_test(
+  "Unit.Hydro.EquationsOfState.Python.Enthalpy"
+  "Test_Enthalpy.py"
+  "Unit;EquationsOfState;Python"
+  PyEquationsOfState)
+
+spectre_add_python_bindings_test(
   "Unit.Hydro.EquationsOfState.Python.PolytropicFluid"
   "Test_PolytropicFluid.py"
   "Unit;EquationsOfState;Python"
@@ -10,5 +16,11 @@ spectre_add_python_bindings_test(
 spectre_add_python_bindings_test(
   "Unit.Hydro.EquationsOfState.Python.PiecewisePolytropicFluid"
   "Test_PiecewisePolytropicFluid.py"
+  "Unit;EquationsOfState;Python"
+  PyEquationsOfState)
+
+spectre_add_python_bindings_test(
+  "Unit.Hydro.EquationsOfState.Python.Spectral"
+  "Test_Spectral.py"
   "Unit;EquationsOfState;Python"
   PyEquationsOfState)
--- a/tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
+++ b/tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
@@ -1,0 +1,96 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import unittest
+
+import numpy as np
+import numpy.testing as npt
+
+import spectre
+import spectre.PointwiseFunctions.Hydro.EquationsOfState as EoS
+from spectre.DataStructures import DataVector
+from spectre.DataStructures.Tensor import Scalar
+
+
+class TestEnthalpy(unittest.TestCase):
+    def test_enthalpy(self):
+        # Parameters are DBHF parameters from https://doi.org/10.48550/arXiv.2301.13818
+
+        low_eos_spectre = EoS.Spectral(
+            reference_density=4.533804669935759e-05,
+            reference_pressure=9.970647727158039e-08,
+            spectral_coefficients=[
+                1.2,
+                0.0,
+                1.34440187653529,
+                -0.46098357752567365,
+            ],
+            upper_density=0.0004533804669935759,
+        )
+
+        eos_spectre = EoS.EnthalpySpectral(
+            reference_density=0.00022669023349678794,
+            max_density=0.0031736632689550316,
+            min_density=0.0004533804669935759,
+            trig_scale=1.26426988871305,
+            polynomial_coefficients=[
+                1.0,
+                0.08063293075870805,
+                4.406887319408924e-26,
+                8.177895241388924e-22,
+                0.013558242085066733,
+                0.004117320982626606,
+                9.757362504479485e-26,
+                1.5646573325075753e-30,
+                0.00016253964205058317,
+            ],
+            sin_coefficients=[
+                0.0003763514388305583,
+                0.017968749910748837,
+                0.008140052979970034,
+                -0.003067418379116628,
+                -0.0008236601907322793,
+            ],
+            cos_coefficients=[
+                -0.01080996024705052,
+                -0.003421193490191067,
+                0.012325774692378716,
+                0.004367136076912163,
+                -0.00020374276952538073,
+            ],
+            low_density_eos=low_eos_spectre,
+            transition_delta_epsilon=0.0,
+        )
+        self.assertTrue(
+            isinstance(eos_spectre, EoS.RelativisticEquationOfState1D)
+        )
+        # Only testing p(rho) because the class is tested thoroughly in C++
+        density = np.geomspace(4.6e-4, 5e-3, 10)
+        pressure = eos_spectre.pressure_from_density(
+            Scalar[DataVector](density)
+        )
+        # These are the pressures calculated in python implementation
+        # testing the bindings. The python implementation will be added
+        # in a second merge request.
+
+        pressures_pyth = [
+            1.5302849088016283e-5,
+            3.1441124536164583e-5,
+            6.5907292284151015e-5,
+            1.4033564857885284e-4,
+            3.0098826697494873e-4,
+            6.4603643776028635e-4,
+            1.3821660234437354e-3,
+            2.9420820328679361e-3,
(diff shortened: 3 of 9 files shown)


src/PointwiseFunctions/Hydro/EquationsOfState/Python/Enthalpy.cpp
```diff
@@ -0,0 +1,67 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "PointwiseFunctions/Hydro/EquationsOfState/Python/Enthalpy.hpp"
+#include "PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.hpp"
```
nilsvu: move this include down (keep only the Python/Enthalpy.hpp as the first line here, since it's the header corresponding to this file)

src/PointwiseFunctions/Hydro/EquationsOfState/Python/Enthalpy.cpp
```diff
@@ -0,0 +1,67 @@
...
+
+// low_density_eos uses Spectral and Polytropic Fluid type EoS.
+#include <pybind11/numpy.h>
+#include <pybind11/operators.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+#include <string>
+#include <vector>
+#include "PointwiseFunctions/Hydro/EquationsOfState/PolytropicFluid.hpp"
```
nilsvu, line 14: add an empty line between above this line (between system and project includes)

src/PointwiseFunctions/Hydro/EquationsOfState/Python/Enthalpy.cpp
```diff
@@ -0,0 +1,67 @@
...
+#include "PointwiseFunctions/Hydro/EquationsOfState/Spectral.hpp"
+
+namespace py = pybind11;
+
+namespace EquationsOfState::py_bindings {
+
+void bind_enthalpy(py::module& m) {
+  py::class_<Enthalpy<PolytropicFluid<true>>, EquationOfState<true, 1>>(
```
nilsvu: Please combine these four functions into one. To do this, create a new function template `bind_enthalpy_impl<LowDensityEos>` that you call four times with different template parameters. You can pass the name of the class as a string to the function.

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/CMakeLists.txt
```diff
@@ -12,3 +12,10 @@ spectre_add_python_bindings_test(
   "Test_PiecewisePolytropicFluid.py"
   "Unit;EquationsOfState;Python"
   PyEquationsOfState)
+
+spectre_add_python_bindings_test(
+  "Unit.Hydro.EquationsOfState.Python.Spectral"
+  "Test_Spectral.py"
```
nilsvu, line 24: Register the Enthalpy test with CMake as well

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
```diff
@@ -0,0 +1,136 @@
...
+            -0.003421193490191067,
+            0.012325774692378716,
+            0.004367136076912163,
+            -0.00020374276952538073,
+        ]
+        transition_delta_epsilon = 0.0
+
+        ## Low Density Equation:
+
+        # Generate python spectral equation of state.
```
nilsvu: Is the idea here to re-implement the EOS in Python and compare to the C++ implementation? Isn't the EOS implementation tested sufficiently in C++ already? I think here in Python you only have to check that the bindings you wrote work, meaning that you can construct and invoke the EOS in Python. That's why the PolytropicFluid EOS has the comment "Only testing p(rho) because the class is tested thoroughly in C++". It's important that unit tests only test the particular unit of code they are written for (the bindings in this case, not the full EOS implementation). Otherwise the unit tests become harder to maintain. Therefore, I suggest you simplify these Python tests significantly. You could call the Python re-implementation in the C++ test if you think it would strengthen the test.

nilsvu (changes requested):

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
isaaclegred: I think it would be nice to have the python implementation available, like you said it probably should be called in the c++ tests.  I would say that the current EoS implementation is well tested in C++ but where the numbers in the tests come from is opaque.  Without the python implementation, if someone needed to add a new test, or test a new component, it would be an absolute nightmare.

nilsvu: Ok then I suggest you call the EOS from the C++ test and keep the Python test very simple (because the code it tests is very simple: it's just the binding code).

tests/Unit/Helpers/PointwiseFunctions/Hydro/EquationsOfState/TestHelpers.hpp
```diff
@@ -46,7 +46,7 @@ void check_impl(
     const MemberArgs&... member_args) {
   // Bounds for: density
   const std::array<std::pair<double, double>, 1> random_value_bounds{
-      {{1.0e-4, 4.0}}};
+      {{1.0e-4, 2.0}}};
```
isaaclegred: This range is better adapted to realistic EoSs, perhaps you should add a comment to this effect

tests/Unit/Helpers/PointwiseFunctions/Hydro/EquationsOfState/TestHelpers.hpp
```diff
@@ -74,7 +74,7 @@ void check_impl(
                                      "_rel_rest_mass_density_from_enthalpy")
                        : std::string(python_function_prefix +
                                      "_newt_rest_mass_density_from_enthalpy"),
-        {{{1, 1.0e4}}}, member_args_tuple, used_for_size);
+        {{{1, 5}}}, member_args_tuple, used_for_size);
```
isaaclegred: Same as above

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.py
```diff
@@ -0,0 +1,570 @@
...
+
+
+def alt(N):
+    a = np.empty((N,), int)
+    a[::2] = 1
+    a[1::2] = -1
+    return np.flip(a)
+
+
+class enthalpy_interpolator:
```
isaaclegred: Add a docstring to this class, it should say something about what this class does.  E.g.
```python
"""
A class to store the expansion of the enthalpy which is used by the enthalpy parametrization.  See the documentation for the Enthalpy EoS.
"""
```

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.py
```diff
@@ -0,0 +1,570 @@
...
+import numpy.polynomial.polynomial as polynomial
+import pandas as pd
+import scipy.interpolate as interpolate
+import scipy.optimize as optimize
+import Spectral as spectral
+from matplotlib import pyplot as plt
+from scipy.integrate import solve_ivp
+from scipy.special import factorial
+
+rhonuc = 2.8e14  # cgs
```
isaaclegred: Remove this line

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.py
```diff
@@ -0,0 +1,570 @@
...
+
+class enthalpy_interpolator:
+    def __init__(
+        self,
+        poly_coeffs,
+        sin_coeffs,
+        cos_coeffs,
+        min_density,
+        max_density,
+        rho_scale=0.5 * rhonuc,
```
isaaclegred: Remove the defaults for `rho_scale` and `trig_scale`

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.py
```diff
@@ -0,0 +1,570 @@
...
+
+    def __add__(self, other):
+        return enthalpy_interpolator(
+            self.poly_coeffs + other.poly_coeffs,
+            self.sin_coeffs + other.sin_coeffs,
+            self.cos_coeffs + other.cos_coeffs,
+            trig_scale=self.trig_scale,
+        )
+
+    @staticmethod
```
isaaclegred: This function should be somewhere in spectre, but it's unclear where to me, @nilsvu do you have ideas for this?

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.py
```diff
@@ -0,0 +1,570 @@
...
+                    x, self.sin_coeffs, self.trig_scale
+                )
+                + enthalpy_interpolator.eval_cos(
+                    x, self.cos_coeffs, self.trig_scale
+                )
+            )
+            + constant
+        )
+
+    def integ(self, m=1, k=[0], lbnd=0, make_expd=False):
```
isaaclegred: Point to equations in https://arxiv.org/abs/2301.13818.  This is Eqns. 15 + 16

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.py
```diff
@@ -0,0 +1,570 @@
...
+            fit_sin_coeffs,
+            fit_cos_coeffs,
+            valid_range=val_range,
+            rho_scale=scale,
+            trig_scale=trig_scale,
+            fit_cost=fit_cost,
+        )
+
+
+class eos_evaluator:
```
isaaclegred: Add a docstring saying that this class represents the enthalpy parametrization.   Say what the low density EoS is.

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Enthalpy.py
```diff
@@ -0,0 +1,570 @@
...
+            lambda x: self.eos.rho_scale * self.energy_density_dimless_from_x(x)
+        )
+
+        self.h_from_x = eos
+        self.pressure_from_x = lambda x: self.h_from_x(
+            x
+        ) * self.baryon_density_from_x(x) - self.energy_density_from_x(x)
+        self.chi_from_x = self.eos.deriv()
+
+    def pressure_from_rest_mass_density(self, rest_mass_density):
```
isaaclegred: Make these functions agree with the names from the c++ versions of the functions, i.e. `pressure_from_density(self, rest mass_density)`

isaaclegred (commented):
Looks good, just a few changes! Mainly just some changes to comments and variable names

nilsvu:
You can squash the commits. To make this PR more manageable, could you please split it into the two things it does, which is (1) add the pybindings (very little code and small test) and (2) extend the EOS test with the Python re-implementation? Please make these two separate commits. You can even omit (2) for now or open a separate PR for that, because then we can already merge (1).

isaaclegred:
@nilsvu  This has been narrowed in scope, can you mark it as ready for review?

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
```diff
@@ -0,0 +1,156 @@
...
+# See LICENSE.txt for details.
+
+import json
+import sys
+import unittest
+
+"These imports will be needed for second merge request"
+# sys.path.append("..")
+# import Enthalpy as enthalpy_python
+# import Enthalpy.eos_evaluator as enthalpy_eos
```
nilsvu: delete this for now please, and also the other commented-out code

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
```diff
@@ -0,0 +1,156 @@
...
+from spectre.DataStructures.Tensor import Scalar
+
+# import Spectral as spectral_python
+
+
+class TestEnthalpy(unittest.TestCase):
+    def test_enthalpy(self):
+        ##Set parameters:
+
+        low_eos_reference_density = 4.533804669935759e-05
```
nilsvu: just put these numbers directly into the constructors below, since you don't use them multiple times.

Also add a comment how you chose these numbers.

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
```diff
@@ -0,0 +1,156 @@
...
+        # Only testing p(rho) because the class is tested thoroughly in C++
+        density = np.geomspace(4.6e-4, 5e-3, 10)
+        pressure = eos_spectre.pressure_from_density(
+            Scalar[DataVector](density)
+        )
+        """
+        These are the pressures calculated in python implementation
+        testing the bindings. The python implementation will be added
+        in a second merge request.
+        """
```
nilsvu: make this a comment, not a string

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
```diff
@@ -0,0 +1,156 @@
...
+        pressures_pyth = [
+            0.00001530284908801628381003201440080374595709145069122314453125,
```
nilsvu: How did you get this many digits? :D And why do they all end with 5? :D Just limit this to at most 16 digits (double precision)

tests/Unit/PointwiseFunctions/Hydro/EquationsOfState/Python/Test_Enthalpy.py
```diff
@@ -0,0 +1,156 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import json
```
nilsvu: you don't use this, please delete. Also other things you don't use, like `pandas`

nilsvu (changes requested):
Comments apply to both test files.

ctrandrsn:
@nilsvu whenever you get a chance this is ready for review

nilsvu (approved):
Looks good! Thanks for adding this. Please squash your commits and rebase on latest develop.

ctrandrsn:
@nilsvu Looks like auto-merge was disabled but, it's ready to go when you get the chance.

nilsvu (dismissed):

nilsdeppe:
There are a couple small clang-tidy things.

nilsdeppe (approved):
Awesome! Thank you very much for doing this :D

PR #5255, opened 2023-07-28
Smooth horizon coefs in size control

knelli2:
## Proposed changes

These are the changes that were used to get us to common horizon.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size.hpp
+++ b/src/ControlSystem/ControlErrors/Size.hpp
@@ -258,16 +258,35 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
             QueueTags::InverseSpatialMetricOnExcisionSurface<Frame::Distorted>>(
             excision_quantities);
 
+    // lambda_00 is the quantity of the same name in ArXiv:1211.6079,
+    // and dt_lambda_00 is its time derivative.
+    // This is the map parameter that maps the excision boundary in the grid
+    // frame to the excision boundary in the distorted frame.
+    const auto map_lambda_and_deriv =
+        functions_of_time.at(function_of_time_name)->func_and_deriv(time);
+    const double lambda_00 = map_lambda_and_deriv[0][0];
+    const double dt_lambda_00 = map_lambda_and_deriv[1][0];
+
+    // horizon_00 is \hat{S}_00 in ArXiv:1211.6079,
+    // and dt_horizon_00 is its time derivative.
+    // These are coefficients of the horizon in the distorted frame.
+    const double horizon_00 = apparent_horizon.coefficients()[0];
+    const double dt_horizon_00 = time_deriv_apparent_horizon.coefficients()[0];
+
+    // This is needed for every state
+    const double control_error_delta_r = size::control_error_delta_r(
+        horizon_00, dt_horizon_00, lambda_00, dt_lambda_00,
+        grid_frame_excision_sphere_radius);
+
     info_.damping_time = min(tuner.current_timescale());
 
     const size::ErrorDiagnostics error_diagnostics = size::control_error(
         make_not_null(&info_), make_not_null(&char_speed_predictor_),
         make_not_null(&comoving_char_speed_predictor_),
-        make_not_null(&delta_radius_predictor_), time, apparent_horizon,
-        excision_surface, grid_frame_excision_sphere_radius,
-        time_deriv_apparent_horizon, lapse, shifty_quantity,
-        spatial_metric_on_excision, inverse_spatial_metric_on_excision,
-        functions_of_time.at(function_of_time_name));
+        make_not_null(&delta_radius_predictor_), time, control_error_delta_r,
+        dt_lambda_00, apparent_horizon, excision_surface, lapse,
+        shifty_quantity, spatial_metric_on_excision,
+        inverse_spatial_metric_on_excision);
 
     state_history_.store(time, info_, error_diagnostics.control_error_args);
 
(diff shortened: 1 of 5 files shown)


src/ControlSystem/ControlErrors/Size/Error.hpp
```diff
@@ -63,26 +60,20 @@ struct ErrorDiagnostics {
...
  * \param excision_boundary a Strahlkorper representing the excision
  *        boundary in frame Frame.  Note that the excision boundary is assumed
  *        to be a sphere in the grid frame.
- * \param grid_frame_excision_boundary_radius the radius of the (assumed
- *        spherical) excision boundary in the grid frame.
- * \param time_deriv_apparent_horizon Time derivative of apparent_horizon,
- *        where the time derivative is taken in frame Frame.  That is, this
- *        should simply be lim dt->0
- *        (apparent_horizon(t+dt)-apparent_horizon(t))/dt with no advective
- *        terms.
```
markscheel, line 76: Please add these dox to the new function `control_error_delta_r`.  In particular, we shouldn't lose the information about what frame the time derivative is taken in, and that the excision bdry is assumed spherical in the grid frame.

tests/Unit/ControlSystem/ControlErrors/Test_SizeError.cpp
```diff
@@ -203,7 +237,10 @@ void test_size_error_one_step(
     const auto control_error_history = error_class.control_error_history();
 
     // These should be identical because the control error class calls the
-    // control_error function
+    // control_error function. The horizon smoothing averager would normally
+    // make this slightly different, but it needs a few times before it can do
+    // anything. So in the meantime, the initial values for the horizon coef and
+    // its derivatives are the ones specified above
```
markscheel: I guess this means that the ah-smoothing-averager code is not tested, correct?  Probably should say this explicitly. Not sure how to easily test it (but testing it may be beyond the scope of this PR).

markscheel (commented):
Looks good! A few minor dox things.

tests/Unit/ControlSystem/ControlErrors/Test_SizeError.cpp
knelli2: Correct, it's not tested. I also wasn't sure how to explicitly test it, so I think it's something we'll have to add later.

knelli2:
@markscheel Squashed in the dox changes

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -230,35 +301,100 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
...
+      dt_horizon_00 = averaged_horizon_coef_at_average_time.value()[1][0];
+      const double d2t_horizon_00 =
+          averaged_horizon_coef_at_average_time.value()[2][0];
+
+      // Must account for time offset of averaged time. Do a simple Taylor
+      // expansion
+      const double time_diff = time - averaged_time;
+      horizon_00 += time_diff * dt_horizon_00;
+      dt_horizon_00 += 0.5 * square(time_diff) * d2t_horizon_00;
+
```
geoffrey4444, line 345: (optional) Is size control always second-order in time? If it was 3rd order, so you had a third time derivative, why not also update d2t_horizon_00 with the third derivative?

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -230,35 +301,100 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
             QueueTags::InverseSpatialMetricOnExcisionSurface<Frame::Distorted>>(
             excision_quantities);
```
geoffrey4444, line 304: (not a change request) why is `operator()` defined here in the hpp instead of implemented in the cpp file?

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -44,7 +46,34 @@ struct Distorted;
...
 
-namespace control_system::ControlErrors {
+namespace control_system {
+namespace size {
+/*!
+ * \brief Function that computes the control error for
+ * `control_system::size::States::DeltaR`.
+ *
+ * This is helpful to have calculated separately because other control errors
+ * may make use of this quantity. This is Q in Eq. 96 in \cite Hemberger2012jz.
```
geoffrey4444: (optional, if you agree my wording is clearer) The equation for the control error is given in Eq. (96) of \cite Hemberger2012jz.

geoffrey4444 (dismissed):
I'm happy to approve but have some optional comments for you to think about

src/ControlSystem/ControlErrors/Size.hpp
knelli2: The template parameters of `operator()` are not known.

src/ControlSystem/ControlErrors/Size.hpp
knelli2: The function of time for size is a 3rd order `PiecewisePolynomial`. However, regardless of the order, we only use the value and first derivative of the 00 component of the horizon below, not the second derivative, so we don't need to correct it.

src/ControlSystem/ControlErrors/Size.hpp
knelli2: @geoffrey4444 squashed this in

markscheel (approved):

knelli2:
In a meeting with @geoffrey4444, he said he didn't need to look at these changes again, so I'll merge this now. (Test failure unrelated)

PR #5265, opened 2023-07-31
Add list of compute tags for ScalarTensor initialization

guilara:
## Proposed changes

Add list of compute tags required for the initialization of ScalarTensor. These tags are added directly in the executable with `Initialization::Actions::AddComputeTags`.

Note: 
- For now, we add the trace of the extrinsic curvature and trace of christoffel as compute tags. In a future PR the idea would be to replace these for their computation directly in `ScalarTensor::TimeDerivative::apply`. (In testing, the current implementation did not seem significantly slower than just GH.)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ScalarTensor/Initialize.hpp
+++ b/src/Evolution/Systems/ScalarTensor/Initialize.hpp
@@ -1,0 +1,74 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <optional>
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/ConstraintDamping/Tags.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/Constraints.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/System.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
+#include "Evolution/Systems/ScalarTensor/Sources/ScalarSource.hpp"
+#include "Evolution/Systems/ScalarTensor/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Christoffel.hpp"
+#include "PointwiseFunctions/GeneralRelativity/DerivativesOfSpacetimeMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/DetAndInverseSpatialMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/ConstraintGammas.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/DerivSpatialMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/ExtrinsicCurvature.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/SpatialDerivOfLapse.hpp"
+#include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/SpatialDerivOfShift.hpp"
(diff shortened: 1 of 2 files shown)


wthrowe:
> * For now, we add the trace of the extrinsic curvature and trace of christoffel as compute tags. In a future PR the idea would be to replace these for their computation directly in `ScalarTensor::TimeDerivative::apply`. (In testing, the current implementation did not seem significantly slower than just GH.)

Just to clarify, the issue with compute tags isn't speed, it's memory use.

guilara:
> > * For now, we add the trace of the extrinsic curvature and trace of christoffel as compute tags. In a future PR the idea would be to replace these for their computation directly in `ScalarTensor::TimeDerivative::apply`. (In testing, the current implementation did not seem significantly slower than just GH.)
> 
> Just to clarify, the issue with compute tags isn't speed, it's memory use.

Oh, I see. For my isolated BH tests the memory seemed manageable. But I'm not sure about more challenging sims.

nilsdeppe:
Yea, the memory issue is sort of a problem of scale. 1 or 2 tensors isn't a big deal, but all the GH temporaries are ~250 tensor components, and that adds a lot of memory usage. When you include derivatives, etc. you get closer to 400 components, which then really does use a lot more memory. In terms of speed, I wouldn't expect there to be even a measurable difference between compute tag vs. in RHS.

src/Evolution/Systems/ScalarTensor/Initialize.hpp
```diff
@@ -0,0 +1,114 @@
...
+/// \cond
+namespace Parallel {
+template <typename Metavariables>
+class GlobalCache;
+}  // namespace Parallel
+/// \endcond
+
+namespace ScalarTensor::Actions {
+
+struct InitializeScalarTensorAnd3Plus1Variables {
```
nilsdeppe: Would you mind adding some doxygen documentation?

I think you can also turn this into a mutator instead of an action. Then you can use `Initialization::Actions::InitializeItems` in `ParallelAlgorithms/Actions/InitializeItems.hpp` It looks like examples are `evolution::dg::Initialization::Domain` `Initialization::TimeStepperHistory` and `Initialization::TimeStepping`. This should make adding a test quite easy, while you can then just using `db::tag_is_retrievable_v<Tag, db::DataBox<Tags>>` to verify the correct tags are added.

src/Evolution/Systems/ScalarTensor/Initialize.hpp
```diff
@@ -0,0 +1,114 @@
...
+  static constexpr size_t Dim = 3_st;
```
nilsdeppe: `Dim` -> `dim` (variable not template parameter)

nilsdeppe (changes requested):
You can squash immediately

src/Evolution/Systems/ScalarTensor/Initialize.hpp
guilara: Thanks for the suggestion. I was trying to implement this actually, but maybe a cleaner way is one @nikwit suggested to me of just adding these tags directly in the executable with `Initialization::Actions::AddComputeTags`. So maybe in this file what we could write is just the list of tags to add. What do you think?

nilsdeppe: That sounds good to me :)

nilsdeppe (commented):
I like this list approach. Do you just want to squash the two commits together? I think this is good to go then.

src/Evolution/Systems/ScalarTensor/Initialize.hpp
guilara: Good. Then I have changed the content of this PR to just include the list of compute tags to add.

nilsdeppe (approved):
🎉

guilara:
Thanks. It seems some tests timed out here

PR #5266, opened 2023-07-31
Add FoT names to CoordMaps

### LINKED ISSUE #5247: Coordinate maps know which functions of time their maps use
# Feature request:

The information about what functions of time are needed for each coordinate map are available and can be useful to know in the type-erased coordinate map. For example, we only need to check that the functions of time for our block are ready, not all functions of time.

This can be achieved by just adding a new member function to the type-erased base class that the derived class implements and retrieves the function of time names in a set.

knelli2:
## Proposed changes

The CoordinateMap(Base) class is now aware of what functions of time are being used for that map. This information can then be used by elements to make decisions based on which maps they have.

Fixes #5247.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/Composition.cpp
+++ b/src/Domain/CoordinateMaps/Composition.cpp
@@ -17,7 +17,9 @@ Composition<Frames, Dim, std::index_sequence<Is...>>::Composition(
     std::unique_ptr<
         CoordinateMapBase<tmpl::at<frames, tmpl::size_t<Is>>,
                           tmpl::at<frames, tmpl::size_t<Is + 1>>, Dim>>... maps)
-    : maps_{std::move(maps)...} {}
+    : maps_{std::move(maps)...},
+      function_of_time_names_(CoordinateMap_detail::initialize_names(
+          maps_, std::index_sequence<Is...>{})) {}
 
 template <typename Frames, size_t Dim, size_t... Is>
 Composition<Frames, Dim, std::index_sequence<Is...>>&
@@ -166,6 +168,12 @@ void Composition<Frames, Dim, std::index_sequence<Is...>>::pup(PUP::er& p) {
   // whenever possible. See `Domain` docs for details.
   if (version >= 0) {
     p | maps_;
+  }
+
+  // No need to pup this because it is uniquely determined by the maps
+  if (p.isUnpacking()) {
+    function_of_time_names_ = CoordinateMap_detail::initialize_names(
+        maps_, std::index_sequence<Is...>{});
   }
 }
 
@@ -350,10 +358,9 @@ template <typename Frames, size_t Dim, size_t... Is>
 bool Composition<Frames, Dim, std::index_sequence<Is...>>::is_equal_to(
     const CoordinateMapBase<SourceFrame, TargetFrame, Dim>& other) const {
   const auto& cast_of_other = dynamic_cast<const Composition&>(other);
-  bool result = true;
-  EXPAND_PACK_LEFT_TO_RIGHT(
-      (result = result and (*get<Is>(maps_) == *get<Is>(cast_of_other.maps_))));
-  return result;
+
+  return ((function_of_time_names_ == cast_of_other.function_of_time_names_) and
+          ... and (*get<Is>(maps_) == *get<Is>(cast_of_other.maps_)));
 }
 
 #define DIM(data) BOOST_PP_TUPLE_ELEM(0, data)
--- a/src/Domain/CoordinateMaps/Composition.cpp
+++ b/src/Domain/CoordinateMaps/Composition.cpp
@@ -359,8 +359,7 @@ bool Composition<Frames, Dim, std::index_sequence<Is...>>::is_equal_to(
     const CoordinateMapBase<SourceFrame, TargetFrame, Dim>& other) const {
   const auto& cast_of_other = dynamic_cast<const Composition&>(other);
 
-  return ((function_of_time_names_ == cast_of_other.function_of_time_names_) and
-          ... and (*get<Is>(maps_) == *get<Is>(cast_of_other.maps_)));
+  return (... and (*get<Is>(maps_) == *get<Is>(cast_of_other.maps_)));
 }
 
 #define DIM(data) BOOST_PP_TUPLE_ELEM(0, data)
--- a/src/Domain/CoordinateMaps/CoordinateMap.tpp
+++ b/src/Domain/CoordinateMaps/CoordinateMap.tpp
@@ -17,6 +17,7 @@
 #include <type_traits>
 #include <typeinfo>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 #include <vector>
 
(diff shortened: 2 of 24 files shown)


src/Domain/CoordinateMaps/Composition.cpp
```diff
@@ -350,7 +358,11 @@ template <typename Frames, size_t Dim, size_t... Is>
...
   const auto& cast_of_other = dynamic_cast<const Composition&>(other);
-  bool result = true;
+  bool result =
+      function_of_time_names_ == cast_of_other.function_of_time_names_;
+  if (not result) {
+    return false;
+  }
   EXPAND_PACK_LEFT_TO_RIGHT(
       (result = result and (*get<Is>(maps_) == *get<Is>(cast_of_other.maps_))));
   return result;
```
wthrowe: [optional] Not new in this PR, but this is set of statements is `return ((function_of_time_names_ == cast_of_other.function_of_time_names_) and ... and (*get<Is>(maps_) == *get<Is>(cast_of_other.maps_)));`.

src/Domain/CoordinateMaps/CoordinateMap.tpp
```diff
@@ -25,25 +26,72 @@
...
+using map_type_t = typename map_type<T>::type;
+
+template <typename... Maps, size_t... Is>
+std::unordered_set<std::string> initialize_names(
+    const std::tuple<Maps...>& maps, std::index_sequence<Is...> /*meta*/) {
+  std::unordered_set<std::string> function_of_time_names{};
+  const auto add_names = [&function_of_time_names, &maps](auto index) {
+    const auto& map = std::get<decltype(index)::value>(maps);
+    using TupleMap = std::decay_t<decltype(map)>;
+    constexpr bool map_is_unique_ptr = tt::is_a_v<std::unique_ptr, TupleMap>;
```
wthrowe, line 64: Why are these sometimes unique_ptrs and sometimes not?  What are they when they're not?

wthrowe (changes requested):
In the future, please separate large reformatting with no content change into separate commits.

src/Domain/CoordinateMaps/CoordinateMap.tpp
knelli2: The `Composition` coordinate map takes the maps as `unique_ptr<CoordinateMapBase>` while the actual `CoordinateMap` just takes maps as the individual maps (Affine, Rotation, etc...). That's why I needed to handle both here.

knelli2:
@wthrowe Added a new commit with `clang-format` stuff (sorry since that file had a lot of changes already I didn't realize I changed a lot of stuff that I didn't touch). Also squashed in your suggestion about the template pack.

wthrowe (dismissed):

knelli2:
@wthrowe Found a mistake in the `pup` function of `ProductOf2Maps` so added a quick fixup that handles it.

wthrowe:
That looks right.  Go ahead and squash.

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -398,6 +376,15 @@ void Shape::pup(PUP::er& p) {
   if (p.isUnpacking()) {
     ylm_ = ylm::Spherepack(l_max_, m_max_);
   }
+
+  if (version >= 1) {
+    p | f_of_t_names_;
+  } else if (p.isUnpacking()) {
+    f_of_t_names_ = std::unordered_set<std::string>{{shape_f_of_t_name_}};
```
nilsvu: Does this pattern of serializing the names work correctly? I think some names are still hardcoded, so if they're changed in the code and the old name is deserialized then there's an inconsistency between the hardcoded name and the name in the set. This means either you don't serialize the names at all but just return the hardcoded names (no need for the member variable even) and don't bother with serialization. Then the serialization will still be broken if names are changed because the domain/maps are inconsistent with the serialized FoT map in the file, but that's also the case on develop right now I think. Or you fix all this by serializing the hardcoded FoT name in the maps. Does that make sense or am I misunderstanding something?

knelli2: Conclusions of discussion with @nilsvu offline:

The "hardcoded" names are actually in the control system. The control system expects certain names in the functions of time (and the domain creators then must supply those names to be usable with control system). Meaning if we change the control system names (hardcoded in `src/ControlSystems`), then everything breaks. (so let's not do that)

The maps themselves actually take names in their constructors and already serialize them. So what's important here is that when we deserialize the maps and the functions of time, everything is consistent. So since we already serialize the names in the maps, we can just always reconstruct the `std::unordered_set` that is needed now from the serialized names and we don't need to change the versions of the pup functions. So I'll make a change to not serialize the `std::unordered_set` and only reconstruct it.

knelli2:
@nilsvu @wthrowe Made those changes in a fixup

nilsvu (dismissed):

src/Domain/CoordinateMaps/TimeDependent/Rotation.cpp
```diff
@@ -174,6 +150,11 @@ void Rotation<Dim>::pup(PUP::er& p) {
   if (version >= 0) {
     p | f_of_t_name_;
   }
+
+  // No need to pup this because it is uniquely determined by f_of_t_name_
+  if (p.isUnpacking()) {
+    f_of_t_names_.insert(f_of_t_name_);
```
wthrowe, line 157: I don't know whether we ever unpack into something that hasn't just been default-constructed, but probably best to clear this first just in case.

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -395,8 +372,13 @@ void Shape::pup(PUP::er& p) {
     p | transition_func_;
   }
 
+  // No need to pup these because they are uniquely determined by other members
   if (p.isUnpacking()) {
     ylm_ = ylm::Spherepack(l_max_, m_max_);
+    f_of_t_names_.insert(shape_f_of_t_name_);
```
wthrowe, line 379: Again, clear first.

src/Domain/CoordinateMaps/TimeDependent/SphericalCompression.cpp
```diff
@@ -296,6 +290,11 @@ void SphericalCompression<InteriorMap>::pup(PUP::er& p) {
     p | max_radius_;
     p | center_;
   }
+
+  // No need to pup this because it is uniquely determined by f_of_t_name_
+  if (p.isUnpacking()) {
+    f_of_t_names_.insert(f_of_t_name_);
```
wthrowe, line 297: Again, clear.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -286,6 +280,11 @@ void Translation<Dim>::pup(PUP::er& p) {
     f_of_r_ = nullptr;
     center_ = make_array<Dim, double>(0.0);
   }
+
+  // No need to pup this because it is uniquely determined by f_of_t_name_
+  if (p.isUnpacking()) {
+    f_of_t_names_.insert(f_of_t_name_);
```
wthrowe, line 261: Clear again.

wthrowe (changes requested):
Squash.

knelli2:
Ok squashed

nilsvu (approved):

wthrowe (approved):

wthrowe:
Clang-tidy failures are boost.

PR #5268, opened 2023-07-31
Rewrite translation map root-find

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
@@ -13,13 +13,13 @@
 #include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
 #include "DataStructures/Tensor/Identity.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
-#include "NumericalAlgorithms/RootFinding/RootBracketing.hpp"
 #include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
 #include "PointwiseFunctions/MathFunctions/MathFunction.hpp"
 #include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/DereferenceWrapper.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
 #include "Utilities/MakeArray.hpp"
 #include "Utilities/MakeWithValue.hpp"
 #include "Utilities/StdArrayHelpers.hpp"
@@ -87,17 +87,7 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
              << ").");
   double radial_function_value = 1.0;
   if (f_of_r_ != nullptr) {
-    std::array<double, Dim> distance_to_center = target_coords;
-    double root_factor = 0.;
-    for (size_t i = 0; i < Dim; i++) {
-      // shifted center
-      gsl::at(distance_to_center, i) -= gsl::at(center_, i);
-      // root_factor needed for root_finder
-      root_factor +=
-          -2.0 * gsl::at(distance_to_center, i) * function_of_time[i];
-    }
-    const double root = root_finder(magnitude(distance_to_center),
-                                    sqrNorm(function_of_time), root_factor);
+    const double root = root_finder(target_coords - center_, function_of_time);
     radial_function_value = (*f_of_r_)(root);
   }
   for (size_t i = 0; i < Dim; i++) {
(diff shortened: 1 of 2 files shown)


src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -236,24 +235,6 @@ double Translation<Dim>::root_finder(const double target_radius,
...
-      [this, &target_radius, &squared_function_of_time,
-       &root_factor](const double x) -> std::optional<double> {
-    return square(target_radius) +
-           ((*f_of_r_)(x)) *
-               (root_factor + squared_function_of_time * (*f_of_r_)(x)) -
-           square(x);
-  };
-  if (radius_bracket_function(lower_bound).value() <=
-      std::numeric_limits<double>::min()) {
-    return lower_bound;
```
AlexCarpenter46, line 252: I haven't tested this being taken out, but does this still cover the cases where the root could be or is close to 0?

AlexCarpenter46 (commented):
Looks good, didn't realize I was passing the wrong value for the function_at_upper_bound. If this covers the case of the root being 0 or close to, then LGTM

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
wthrowe: It's hard to analyze this code with so many user-specified functions (`f_of_r_` and multiple functions of time).  Can you give an example of the sort of case you're worried about?

I'm also not entirely sure what `root` represents.  It appears to be some sort of radius, but I don't know of what.

AlexCarpenter46: The `root` here represents the radius for the radial math function that will take the `target_coords` back to the `source_coords`. So if you have a point that gets translated in a way that looks very close to `gsl::at(result, i) += gsl::at(function_of_time_, i) * 1` (like the case when there's no radial falloff) then the root has to be 0. I think this can happen when the radius to your original coordinate is close to the origin and hasn't been translated very much yet.

wthrowe: I see.  Not sure if there might have been problems there, so I rewrote the root-find so that the function is manifestly non-positive at r=0.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -221,52 +215,36 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
...
+  const auto wrapper = [this, &distance_to_center,
+                        &function_of_time](const double r_squared) {
+    // This form makes it clear that wrapper(0.0) <= 0.0, even with
+    // roundoff, so it is guaranteed that the lower end of the bracket
+    // is valid even when we have to correct for negative radii.
+    const double f_of_r = (*f_of_r_)(sqrt(r_squared));
+    double expression_to_zero = r_squared;
+    for (size_t i = 0; i < Dim; ++i) {
+      expression_to_zero -=
+          square(distance_to_center[i] - f_of_r * function_of_time[i]);
```
AlexCarpenter46: square(gsl::at(distance_to_center, i) - f_of_r * gsl::at(function_of_time, i));

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -54,19 +54,16 @@ namespace domain::CoordinateMaps::TimeDependent {
...
  * \f{equation}{
- * \vec{\xi} = \vec{\bar{\xi}} - F(R)\vec{T}(t)
+ * \vec{\xi} = \vec{\bar{\xi}} - F(r)\vec{T}(t)
  * \f}
- * Where R is the root where \f$F(r)\f$ and \f$\vec{T}(t)\f$
- * cross zero. This is done by bracketing and finding the root of
+ * where \f$r^2\f$ is found as the root of
  * \f{equation}{
- * \bar{r}^2 + F(x) (A + |\vec{T}(t)| F(x)) - x^2
+ *   r^2 = \Big(\vec{\bar{\xi}} - \vec{c} - F(r) \vec{T}(t)\Big)^2.
```
AlexCarpenter46, line 65: should this be `r^2 = \vec{\bar{\xi}}^2 - \Big(\vec{c} - F(r)\vec{T}(t)\Big)^2`

AlexCarpenter46 (approved):
Ran 10,000 tests without a fail, LGTM once gls::at is in 😄! Also thanks for doing this, the root function is much easier to follow now

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
wthrowe: Did the first, but not the second.  We generally don't use `gsl::at` on `DataVector`s because it's redundant.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -88,16 +88,11 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
     for (size_t i = 0; i < Dim; i++) {
       // shifted center
       gsl::at(distance_to_center, i) -= gsl::at(center_, i);
-      // root_factor needed for root_finder
-      root_factor +=
-          -2.0 * gsl::at(distance_to_center, i) * function_of_time[i];
     }
-    const double root = root_finder(magnitude(distance_to_center),
-                                    sqrNorm(function_of_time), root_factor);
+    const double root = root_finder(distance_to_center, function_of_time);
```
knelli2: [Optional] No need for `distance_to_center` here. Just do

```
const double root = root_finder(target_coords - center_, function_of_time);
```

knelli2 (dismissed):

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
wthrowe: Good point.  Will do.

knelli2 (approved):

PR #5270, opened 2023-07-31
Inject ExcisionSphere time dep in BCO domains

knelli2:
## Proposed changes

In order to keep a versioning test passing, I had to reimplement a bit of the BCO domain creator code to actually make the domain.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Domain.cpp
+++ b/src/Domain/Domain.cpp
@@ -106,6 +106,20 @@ void Domain<VolumeDim>::inject_time_dependent_map_for_block(
 }
 
 template <size_t VolumeDim>
+void Domain<VolumeDim>::inject_time_dependent_map_for_excision_sphere(
+    const std::string& excision_sphere_name,
+    std::unique_ptr<
+        domain::CoordinateMapBase<Frame::Grid, Frame::Inertial, VolumeDim>>
+        moving_mesh_grid_to_inertial_map) {
+  ASSERT(excision_spheres_.count(excision_sphere_name) == 1,
+         "Cannot inject time dependent maps into excision sphere '"
+             << excision_sphere_name << "' because it does not exist.");
+
+  excision_spheres_.at(excision_sphere_name)
+      .inject_time_dependent_maps(std::move(moving_mesh_grid_to_inertial_map));
+}
+
+template <size_t VolumeDim>
 bool Domain<VolumeDim>::is_time_dependent() const {
   return alg::any_of(blocks_, [](const Block<VolumeDim>& block) {
     return block.is_time_dependent();
--- a/tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
+++ b/tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
@@ -565,6 +565,71 @@ void test_bbh_time_dependent_factory(const bool with_boundary_conditions,
   const std::vector<double> times_to_check{{1., 2.3}};
   const auto domain = TestHelpers::domain::creators::test_domain_creator(
       *binary_compact_object, with_boundary_conditions, false, times_to_check);
+
+  const auto& blocks = domain.blocks();
+  const auto& final_block = blocks[blocks.size() - 1];
+
+  std::unordered_map<std::string, ExcisionSphere<3>>
+      expected_excision_spheres{};
+  expected_excision_spheres.emplace(
+      "ExcisionSphereA",
+      ExcisionSphere<3>{1.0,
+                        tnsr::I<double, 3, Frame::Grid>{{3.0, 0.0, 0.0}},
+                        {{0, Direction<3>::lower_zeta()},
+                         {1, Direction<3>::lower_zeta()},
+                         {2, Direction<3>::lower_zeta()},
+                         {3, Direction<3>::lower_zeta()},
+                         {4, Direction<3>::lower_zeta()},
+                         {5, Direction<3>::lower_zeta()}}});
+  if (with_time_dependence) {
+    expected_excision_spheres.at("ExcisionSphereA")
+        .inject_time_dependent_maps(
+            final_block.moving_mesh_grid_to_inertial_map().get_clone());
+  }
+  if (excise_B) {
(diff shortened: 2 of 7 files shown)


src/Domain/Domain.cpp
```diff
@@ -105,6 +105,21 @@ void Domain<VolumeDim>::inject_time_dependent_map_for_block(
       std::move(moving_mesh_distorted_to_inertial_map));
 }
 
+template <size_t VolumeDim>
+void Domain<VolumeDim>::inject_time_dependent_map_for_excision_sphere(
+    const std::string& excision_sphere_name,
+    std::unique_ptr<
```
nilsvu, line 111: const

src/Domain/Domain.cpp
```diff
@@ -105,6 +105,21 @@ void Domain<VolumeDim>::inject_time_dependent_map_for_block(
...
+        domain::CoordinateMapBase<Frame::Grid, Frame::Inertial, VolumeDim>>
+        moving_mesh_grid_to_inertial_map) {
+  if (excision_spheres_.count(excision_sphere_name) != 1) {
+    ERROR("Cannot inject time dependent maps into excision sphere '"
```
nilsvu: `ASSERT`

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -346,6 +346,9 @@ void test_construction(const CylBCO& creator,
     CHECK(excision_sphere_a.center().get(i) == center_objectA.at(i));
     CHECK(excision_sphere_b.center().get(i) == center_objectB.at(i));
   }
+  const auto& block = domain.blocks()[0];
+  CHECK(block.is_time_dependent() == excision_sphere_a.is_time_dependent());
+  CHECK(block.is_time_dependent() == excision_sphere_b.is_time_dependent());
```
nilsvu, line 351: Please add a test to both BCO domain creators that the center of the excisions maps to the correct time-dependent inertial coords for hard-coded values.

Also consider adding a test to `tests/Unit/Helpers/Domain/Creators/TestHelpers.hpp` that `excision_sphere.is_time_dependent() == neighboring_block.is_time_dependent()` for all abutting directions.

nilsvu (changes requested):

src/Domain/Domain.cpp
knelli2: But I `std::move` it

nilsvu: Oh ok

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
knelli2: So I can't do the second thing because the Sphere domain creator has an excision sphere, but it can have a Shape map in the inner shell of blocks. However, we still don't have the proper transition functions to use for inside an excision. So I couldn't add an actual map to the excision sphere. I either had to 1) add an identity map to the excision sphere or 2) just not add a map, so I opted for 2. However, this means that the check that an excision sphere is time dependent if all blocks abutting it are time dependent isn't valid anymore so I got rid of it. We can add it once we figure out the sphere transition stuff.

knelli2:
@nilsvu ping

tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
```diff
@@ -255,8 +262,38 @@ void test_connectivity() {
...
+            .inject_time_dependent_maps(
+                final_block.moving_mesh_grid_to_inertial_map().get_clone());
+      }
+    }
+    const auto& excision_spheres = domain.excision_spheres();
+    CHECK(excision_spheres == expected_excision_spheres);
+
+    const auto check_excision_sphere_map =
+        [&binary_compact_object](const ExcisionSphere<3>& excision_sphere) {
+          const double time = 0.0;
```
nilsvu: I don't think this code runs at all because the `binary_compact_object` isn't time dependent here, right? Move to `test_bbh_time_dependent_factory` below?

knelli2: Done

nilsvu (approved):

tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
```diff
@@ -565,6 +565,71 @@ void test_bbh_time_dependent_factory(const bool with_boundary_conditions,
...
+        const auto functions_of_time =
+            binary_compact_object->functions_of_time();
+        const auto& center = excision_sphere.center();
+        const auto& map = excision_sphere.moving_mesh_grid_to_inertial_map();
+
+        const auto mapped_point = map(center, time, functions_of_time);
+
+        // Should be same point
+        for (size_t i = 0; i < 3; i++) {
+          CHECK(center.get(i) == approx(mapped_point.get(i)));
```
nilsvu, line 623: This still doesn't really check the map. While you're rebasing, maybe add the quick rotation check like you have in the cylindrical domain?

nilsvu (approved):
Please give this a quick rebase to run with latest CI.

tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
knelli2: I can't really do that because the cylindrical domain had the expansion map turned off so it made the map easy (just a rotation). This one has the expansion map so the mapped point is more complicated and I'd rather not reimplement the map here, or make a lot of edits to turn off the expansion map for the whole test.

Also I don't really see the purpose of testing the map? The maps are already tested elsewhere. What we really care about is that the map itself is the map we expect which is tested above when I check the expected excision spheres against the ones created in the domain creator.

nilsvu: ok

PR #5272, opened 2023-07-31
Enum class for grouping initial data in ForceFree system

yoonso0-0:
## Proposed changes

Add an enum specifying the category of initial data in ForceFree evolution system. The main purpose of this is to conveniently list a subset of initial data that requires a slightly different set of actions or treatments during evolution.

The idea is to enable something like below:
```
using InitialData = std::decay_t<decltype(*solution_or_data)>;
using single_ns_id_list = ForceFree::AnalyticData::available_data<SimulationType::IsolatedNsMagnetosphere>;

tmpl::list_contains_v<single_ns_id_list, InitialData>) {
     ... // (Apply some mutator specific to single NS simulations ..
    }
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ForceFree/CMakeLists.txt
+++ b/src/Evolution/Systems/ForceFree/CMakeLists.txt
@@ -23,6 +23,7 @@ spectre_target_headers(
   Characteristics.hpp
   ElectricCurrentDensity.hpp
   Fluxes.hpp
+  SimulationType.hpp
   Sources.hpp
   System.hpp
   Tags.hpp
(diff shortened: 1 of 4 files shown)


wthrowe:
Could this be detected at runtime instead?  It would be nice to not need to compile another executable.

yoonso0-0:
@wthrowe Uhh this is not accessible from the runtime (user) interface. We'll still use a single executable for running all kinds of initial data. This is used only inside the code to make developer's life easier, as a handy way to get a type list of initial data in different places of ForceFree code.

yoonso0-0:
I think the idea will be more clear in my upcoming PR, which I'll submit within a day

wthrowe (changes requested):
Alright.  Let's hold off on this until I've seen the intended use, then.

PR #5273, opened 2023-07-31
Add basic p-refinement criterion based on truncation error

nilsvu:
## Proposed changes

Just use the `truncation_error` estimate based on power monitors to p-refine to a target error. We can improve and expand this as we go. This p-refines the `ExportCoordinates2D` executable already pretty well when using the inertial coordinates to compute truncation errors and a Sphere domain in the input file.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
+++ b/src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
@@ -1,0 +1,198 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <optional>
+#include <pup.h>
+#include <string>
+#include <vector>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/DataBoxTag.hpp"
+#include "DataStructures/DataBox/ValidateSelection.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Amr/Flag.hpp"
+#include "Domain/Tags.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Options/Context.hpp"
+#include "Options/ParseError.hpp"
+#include "Options/String.hpp"
+#include "ParallelAlgorithms/Amr/Criteria/Criterion.hpp"
+#include "Utilities/Algorithm.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+template <size_t>
+class ElementId;
+/// \endcond
+
+namespace amr::Criteria {
+
+namespace TruncationError_detail {
+/*!
+ * \brief Apply the truncation error criterion to a single tensor component
+ *
+ * The `result` is the current decision in each dimension based on the previous
+ * tensor components. This function will update the flags if necessary. It takes
+ * the "max" of the current and new flags, where the "highest" flag is
+ * `Flag::IncreaseResolution`, followed by `Flag::DoNothing`, and then
+ * `Flag::DecreaseResolution`.
+ */
+template <size_t Dim>
+void max_over_components(
+    gsl::not_null<std::array<Flag, Dim>*> result,
+    const gsl::not_null<std::array<DataVector, Dim>*> power_monitors_buffer,
+    const DataVector& tensor_component, const Mesh<Dim>& mesh,
+    std::optional<double> target_abs_truncation_error,
+    std::optional<double> target_rel_truncation_error);
+}  // namespace TruncationError_detail
+
+/*!
+ * \brief Refine the grid towards the target truncation error
+ *
+ * - If any tensor component has a truncation error above the target value, the
+ *   element will be p-refined.
+ * - If all tensor components still satisfy the target even with one mode
+ *   removed, the element will be p-coarsened.
+ *
+ * For details on how the truncation error is computed see
+ * `PowerMonitors::truncation_error`.
+ *
+ * \tparam Dim Spatial dimension of the grid
+ * \tparam TensorTags List of tags of the tensors to be monitored
+ */
+template <size_t Dim, typename TensorTags>
+class TruncationError : public Criterion {
+ public:
+  struct VariablesToMonitor {
+    using type = std::vector<std::string>;
+    static constexpr Options::String help = {
+        "The tensors to monitor the truncation error of."};
+    static size_t lower_bound_on_size() { return 1; }
+  };
+  struct AbsoluteTargetTruncationError {
+    static std::string name() { return "AbsoluteTarget"; }
+    using type = Options::Auto<double, Options::AutoLabel::None>;
+    static constexpr Options::String help = {
+        "The absolute target truncation error. If any tensor component "
+        "has a truncation error above this value, the element will be "
+        "p-refined."};
+  };
+  struct RelativeTargetTruncationError {
+    static std::string name() { return "RelativeTarget"; }
+    using type = Options::Auto<double, Options::AutoLabel::None>;
+    static constexpr Options::String help = {
+        "The relative target truncation error. If any tensor component "
+        "has a truncation error above this value, the element will be "
+        "p-refined."};
+  };
+
+  using options = tmpl::list<VariablesToMonitor, AbsoluteTargetTruncationError,
+                             RelativeTargetTruncationError>;
+
+  static constexpr Options::String help = {
+      "Refine the grid towards the target truncation error"};
+
+  TruncationError() = default;
+
+  TruncationError(std::vector<std::string> vars_to_monitor,
+                  const std::optional<double> target_abs_truncation_error,
+                  const std::optional<double> target_rel_truncation_error,
+                  const Options::Context& context = {});
+
+  /// \cond
+  explicit TruncationError(CkMigrateMessage* msg);
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(TruncationError);  // NOLINT
+  /// \endcond
+
+  using compute_tags_for_observation_box = tmpl::list<>;
+
+  using argument_tags = tmpl::list<::Tags::DataBox>;
+
+  template <typename DbTagsList, typename Metavariables>
+  std::array<Flag, Dim> operator()(const db::DataBox<DbTagsList>& box,
+                                   Parallel::GlobalCache<Metavariables>& cache,
+                                   const ElementId<Dim>& element_id) const;
+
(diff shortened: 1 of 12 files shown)


src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
```diff
@@ -0,0 +1,113 @@
...
+  };
+
+  using options = tmpl::list<TargetTruncationError>;
+
+  static constexpr Options::String help = {
+      "Refine the grid towards the target truncation error"};
+
+  TruncationError() = default;
+
+  TruncationError(const double target_truncation_error)
```
nilsdeppe: Move out of line

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
```diff
@@ -0,0 +1,113 @@
...
+
+  using compute_tags_for_observation_box = tmpl::list<>;
+
+  using argument_tags = tmpl::list<::Tags::DataBox>;
+
+  template <typename DbTagsList, typename Metavariables>
+  auto operator()(const db::DataBox<DbTagsList>& box,
+                  Parallel::GlobalCache<Metavariables>& /*cache*/,
+                  const ElementId<Dim>& /*element_id*/) const {
+    auto result = make_array<Dim>(Flag::DoNothing);
```
nilsdeppe: Move out of line

Don't use auto return type if you know the return type

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
```diff
@@ -0,0 +1,113 @@
...
+    const auto& mesh = db::get<domain::Tags::Mesh<Dim>>(box);
+    tmpl::for_each<TensorTags>(
+        [&result, &box, &mesh,
+         target_truncation_error = target_truncation_error_](const auto tag_v) {
+          // Continue only if AMR flags are still unset
```
nilsdeppe: Why? should there not be some sort of priority procedure over all the flags? If not, this should be documented.

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
```diff
@@ -0,0 +1,113 @@
...
+ * element will be p-refined. For details on how the truncation error is
+ * computed see `PowerMonitors::truncation_error`.
+ *
+ * \tparam Dim Spatial dimension of the grid
+ * \tparam TensorTags List of tags of the tensors to be monitored
+ */
+template <size_t Dim, typename TensorTags>
+class TruncationError : public Criterion {
+ public:
+  struct TargetTruncationError {
```
nilsdeppe: Name this `TargetAbsoluteTruncationError`?

Why not include a relative error too?

nilsdeppe (changes requested):

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
kidder: So the way this criterion is written, AMR will never be able to decrease the resolution as returning DoNothing will always take priority over any criteria that wants to DecreaseResolution.  So this criterion should handle when to request to decrease the resolution as well...

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
```diff
@@ -107,7 +107,9 @@ class TruncationError : public Criterion {
   double target_truncation_error_{};
 };
 
+/// \cond
```
kidder: the documentation wrappers should be in the previous commit

kidder (changes requested):

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
nilsvu: Can I return `Undefined` to signal that this criterion makes no statement about what should happen in that dimension?

nilsdeppe: How can this ever not have an opinion on what to do? The error is either above or below tolerances.

nilsvu: Currently it doesn't try to coarsen the resolution, but I can add that. There should probably be a safety factor though or an acceptable absolute error range so it doesn't switch back and forth between resolutions at every AMR step?

nilsvu: As a first pass I was going to ignore coarsening. We have to experiment with this anyway.

nilsdeppe: Let's just have a meet about this. That's easier than going back and forth on gitHub comments :)

kidder: As AMR is currently implemented, you should return `Join` (the lowest priority flag) if you do not want to override the decisions of other criteria.  But as this is a p-refinement criterion, it should return one of `{IncreaseResolution, DoNothing, DecreaseResolution}` I would have expected this criterion to compute a third number besides the error and target error, such as the estimated error dropping the top mode, in order to determine if the resolution can be decreased.

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
nilsvu: Added with the coarsening

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
nilsvu: Added relative and absolute error

nilsvu:
Rewrote the implementation to work with relative and absolute errors, and do refinement+coarsening.

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
```diff
@@ -0,0 +1,212 @@
...
+ * - If all tensor components still satisfy the target even with one mode
+ *   removed, the element will be p-coarsened.
+ *
+ * For details on how the truncation error is computed see
+ * `PowerMonitors::truncation_error`.
+ *
+ * \tparam Dim Spatial dimension of the grid
+ * \tparam TensorTags List of tags of the tensors to be monitored
+ */
+template <size_t Dim, typename TensorTags>
```
nilsdeppe, line 68: [optional] Maybe in the future we will want to make it so that we can runtime choose how to deal with the tensor tags? E.g. we might want to monitor the L2 norm over the spacetime metric components, etc. It would also be nice to be able to have some runtime control over which of the tags we monitor, so we don't need to recompile to monitor different tags. E.g. we could just add all evolved and primitive tags to the list and then decide at runtime which we actually deal with. Again, not for this PR, but wanted to mention it and get your thoughts. I could see this being especially useful during testing/debugging times of the criterion.

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
```diff
@@ -0,0 +1,212 @@
...
+                              target_rel_truncation_error =
+                                  target_rel_truncation_error_](
+                                 const auto tag_v) {
+    // Stop if we have already decided to refine every dimension
+    if (result == make_array<Dim>(Flag::IncreaseResolution)) {
+      return;
+    }
+    using tag = tmpl::type_from<std::decay_t<decltype(tag_v)>>;
+    const auto& tensor = db::get<tag>(box);
+    for (const DataVector& tensor_component : tensor) {
```
nilsdeppe: what do you think about moving the work in a `cpp` file since this just acts on one DataVector? We just do _so_ much in hpp files and then tweaking anything means recompiling the executable. It would be nice for testing to be able to tweak the algorithm in the cpp file to only have to link. I think we could even extend to taking `Tensor::const_iterator` types and then do the loop over components in the cpp. This would mean that even things like L2 norms of components could be added and debugged with mostly only changing cpp files.

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
```diff
@@ -0,0 +1,212 @@
...
+          // this resolution (or increase resolution if another tensor component
+          // requested that).
+          const double rel_truncation_error_coarsened =
+              pow(10, -PowerMonitors::relative_truncation_error(
+                          modes, modes.size() - 1));
+          const double abs_truncation_error_coarsened =
+              umax * rel_truncation_error_coarsened;
+          if ((target_rel_truncation_error.has_value() and
+               rel_truncation_error_coarsened <=
+                   target_rel_truncation_error.value()) and
```
nilsdeppe: can you use `.value_or(0.0)` or `.value_or(negative_infinity)`? 2x

nilsdeppe (changes requested):
LGTM, a couple requests. You can squash immediately.

kidder (commented):
I have no requested changes

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
nilsvu: Agreed. Added the runtime control over which variables to monitor. I left out the norms over tensor components for now (not sure if we really need that because monitoring the L2 norm over something like the metric seems pretty much the same as monitoring each component. We can add this later if we want it.)

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
nilsvu: Moved to the cpp. Left the loop over components for now (not trivial because `Tensor::const_iterator` is templated and not sure if we need norms over components at all. Can be done later if needed.)

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
nilsvu: Tried it but I think it makes the logic a bit harder to understand because the two conditions for refine/coarsen have to use different `value_or` values.

nilsvu:
Pushed new first commit and a fixup

src/ParallelAlgorithms/Amr/Criteria/TruncationError.hpp
nilsdeppe: Sounds good. The advantage of monitoring an L2 norm over components would be fewer matrix multiplies :) But we can experiment with this in the futer.

nilsdeppe (commented):
LGTM, please squash

kidder (approved):

nilsdeppe (approved):

PR #5275, opened 2023-08-01
Remove Time from ParallelInterpolation

knelli2:
## Proposed changes

Use some template magic.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu (approved):
LGTM. I suppose another option would be to require all "temporal IDs" can convert to `double` (with `explicit operator double()`). @wthrowe what do you think?

wthrowe:
Interesting idea.  I don't see anything obviously wrong with it.  There's only one sane interpretation of a `TimeStepId` as a `double`, and there aren't many places where both types would be valid even if you did get confused about converting.

Go ahead if you want to try it.  Shouldn't be too hard to back out if we find it causes a problem.

PR #5232, opened 2023-07-25
Add a tag for masking neutron star interior in ForceFree evolution system

yoonso0-0:
## Proposed changes

When running neutron star magnetosphere simulations, our plan is to assign this masking variable at the initialization phase to identify grid points located inside the NS surface (e.g. r < 1). In evolution phase, one checks the value of this mask variable and overwrites the electric field to `- v x B` for grid points inside the NS surface.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5281, opened 2023-08-02
Add a rotating star magnetosphere initial data to ForceFree

yoonso0-0:
A rotating star with an initial dipole magnetic field. This is a toy model of a pulsar magnetosphere.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
+++ b/src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
@@ -1,0 +1,166 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.hpp"
+
+#include <cstddef>
+#include <memory>
+#include <pup.h>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/DotProduct.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Options/Options.hpp"
+#include "Options/ParseOptions.hpp"
+#include "PointwiseFunctions/InitialDataUtilities/InitialData.hpp"
+#include "Utilities/MakeWithValue.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace ForceFree::AnalyticData {
+
+RotatingDipole::RotatingDipole(const double vector_potential_amplitude,
+                               const double varpi0, const double delta,
+                               const double angular_velocity,
+                               const double tilt_angle,
+                               const Options::Context& context)
+    : vector_potential_amplitude_(vector_potential_amplitude),
+      varpi0_(varpi0),
+      delta_(delta),
+      angular_velocity_(angular_velocity),
+      tilt_angle_(tilt_angle) {
+  if (varpi0 < 0.0) {
+    PARSE_ERROR(context, "The length constant varpi0 ("
+                             << varpi0_ << ") cannot be negative");
+  }
+  if (delta < 0.0) {
+    PARSE_ERROR(context,
+                "The small number delta (" << delta_ << ") cannot be negative");
+  }
+  if (abs(angular_velocity) >= 1.0) {
+    PARSE_ERROR(context, "The rotation angular velocity ("
+                             << angular_velocity_
+                             << ") must be between -1.0 and 1.0");
+  }
+  if ((tilt_angle < 0.0) or (tilt_angle > M_PI)) {
+    PARSE_ERROR(context, "The rotator tilt angle ("
+                             << tilt_angle_ << ") must be between 0 and Pi");
+  }
+}
+
+RotatingDipole::RotatingDipole(CkMigrateMessage* msg) : InitialData(msg) {}
+
+std::unique_ptr<evolution::initial_data::InitialData>
+RotatingDipole::get_clone() const {
+  return std::make_unique<RotatingDipole>(*this);
+}
+
+void RotatingDipole::pup(PUP::er& p) {
+  InitialData::pup(p);
+  p | vector_potential_amplitude_;
+  p | varpi0_;
+  p | delta_;
+  p | angular_velocity_;
+  p | tilt_angle_;
+  p | background_spacetime_;
+}
+
+PUP::able::PUP_ID RotatingDipole::my_PUP_ID = 0;
+
+tuples::TaggedTuple<Tags::TildeE> RotatingDipole::variables(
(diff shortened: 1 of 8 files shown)


src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.hpp
```diff
@@ -0,0 +1,202 @@
...
+ *                    \delta^2)^{5/2}} .
+ * \end{align}
+ *
+ * Electric fields and charge density are set to zero everywhere.
+ *
+ * Rotation of the star is switched on at $t=0$ with the angular velocity
+ * specified in the input file. Inside the star we impose the MHD condition
+ * $\mathbf{E} + \mathbf{v} \times \mathbf{B} = 0$, where the velocity field is
+ * given by $\mathbf{v} \equiv \Omega \hat{z} \times \mathbf{r}$. (see docs of
+ * the mutator ImposeMhdConditionInsideNs). By this means, initial dipolar
```
wthrowe: This is to be added in a future PR, correct?  (Not a change request, just making sure that's why I can't find it.)

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
```diff
@@ -0,0 +1,137 @@
...
+  const auto& y = get<1>(coords);
+  const auto& z = get<2>(coords);
+  const DataVector x_prime = cos_alpha * x - sin_alpha * z;
+  const DataVector z_prime = sin_alpha * x + cos_alpha * z;
+
+  auto tilde_b_prime = make_with_value<tnsr::I<DataVector, 3>>(coords, 0.0);
+
+  // Regularized dipole field
+  const DataVector r_squared = get(dot_product(coords, coords));
+  const DataVector sqrt_r2_plus_d2 = sqrt(r_squared + square(delta_));
```
wthrowe: You only use this in expressions like `1 / pow<5>(sqrt_r2_plus_d2)`, so might as well store that instead.

wthrowe (changes requested):

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.hpp
yoonso0-0: Ah, yes. Sorry I forgot to mention that.

wthrowe:
Looks good.  Squash.

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
```diff
@@ -0,0 +1,138 @@
...
+  p | vector_potential_amplitude_;
+  p | varpi0_;
+  p | delta_;
+  p | angular_velocity_;
+  p | tilt_angle_;
+  p | background_spacetime_;
+}
+
+PUP::able::PUP_ID RotatingDipole::my_PUP_ID = 0;
+
```
ermost, line 68: I'm not sure it makes much of a difference, but in principle should E = B x (Omega x R) ?

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
```diff
@@ -0,0 +1,138 @@
...
+tuples::TaggedTuple<Tags::TildePsi> RotatingDipole::variables(
+    const tnsr::I<DataVector, 3>& coords, tmpl::list<Tags::TildePsi> /*meta*/) {
+  return {make_with_value<Scalar<DataVector>>(coords, 0.0)};
+}
+
+tuples::TaggedTuple<Tags::TildePhi> RotatingDipole::variables(
+    const tnsr::I<DataVector, 3>& coords, tmpl::list<Tags::TildePhi> /*meta*/) {
+  return {make_with_value<Scalar<DataVector>>(coords, 0.0)};
+}
+
```
ermost, line 120: Maybe add a comment that Q is in principle not zero, would follow from div E, and so would likely go like Omega*B?
In practice it shouldn't matter, because you are self-consistently fixing E in the implicit step outside, and then have Q relax through the evolution equation in the magnetosphere.

yoonso0-0:
Here is the fixup commit. As discussed in #5232 , I added a member function computing the optional masking variable. Updated test as well.

wthrowe:
New stuff looks good.  Squash.

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
yoonso0-0: We turn on the rotation at t=0 and overwrite `E = B x V` in every time steps, so practically this would not matter.

yoonso0-0: Rotation is turned on at t=0 and `E = B x V` is imposed in every time step, so I believe this doesn't have much practical difference.

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
yoonso0-0: If you were asking about the NS interior, we're overwriting `Q=0` inside the NS so this would be fine. If overwriting Q=0 inside the NS is something we shouldn't do, this should be fixed as you pointed out.

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
ermost: yes, for this way of evolving the ID, that's totally fine.
This comment (and the next one) are mainly about documentation.

Could you add a comment on this saying that it should be the ideal condition using v, but that this will be imposed during the evolution?

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
ermost: For the evolution that's totally fine, because Q decouples from the E and B evolution with the fixing.

Maybe add a comment on this, too?

yoonso0-0:
@ermost I'll add more comments on the documentation explaining why `q=E=0`

yoonso0-0:
Okay, pushed a fixup for documentation

ermost:
Changes look great!

ermost (approved):

yoonso0-0:
Squashed and rebased on develop. Thanks for the review!

nilsdeppe:
@wthrowe could you please take a look at this again?

wthrowe (approved):

PR #5295, opened 2023-08-02
Add a mutator initializing ForceFree::Tags::NsInteriorMask

yoonso0-0:
## Proposed changes

This is a follow up of #5232 , a mutator assigning the `std::optional` masking tag in the initialization phase for NS magnetosphere initial data.

Apart from the last one, other commits are from : 

- [ ] #5281 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
+++ b/src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
@@ -1,0 +1,77 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <optional>
+#include <type_traits>
+
+#include "DataStructures/DataBox/Protocols/Mutator.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Tags.hpp"
+#include "Evolution/DgSubcell/Tags/Coordinates.hpp"
+#include "Evolution/DgSubcell/Tags/Inactive.hpp"
+#include "Evolution/Systems/ForceFree/Tags.hpp"
+#include "PointwiseFunctions/InitialDataUtilities/InitialData.hpp"
+#include "PointwiseFunctions/InitialDataUtilities/Tags/InitialData.hpp"
+#include "Utilities/CallWithDynamicType.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TypeTraits/CreateIsCallable.hpp"
+
+namespace ForceFree {
+namespace detail {
+CREATE_IS_CALLABLE(interior_mask)
+CREATE_IS_CALLABLE_V(interior_mask)
+CREATE_IS_CALLABLE_R_V(interior_mask)
+}  // namespace detail
+
+/*!
+ * \brief Assign the masking scalar variable (see Tags::NsInteriorMask) at the
+ * initialization phase in NS magnetosphere simulations.
+ *
+ * Run the `interior_mask()` member function of the initial data if it is
+ * callable.
+ */
+template <typename Metavariables, bool UsedForFdGrid>
+struct MaskNeutronStarInterior : tt::ConformsTo<db::protocols::Mutator> {
+  using argument_tags = tmpl::list<
+      tmpl::conditional_t<
+          UsedForFdGrid,
+          evolution::dg::subcell::Tags::Coordinates<3, Frame::Inertial>,
+          domain::Tags::Coordinates<3, Frame::Inertial>>,
+      evolution::initial_data::Tags::InitialData>;
+
+  using return_tags = tmpl::list<tmpl::conditional_t<
+      UsedForFdGrid,
+      evolution::dg::subcell::Tags::Inactive<Tags::NsInteriorMask>,
+      Tags::NsInteriorMask>>;
+
+  static void apply(
+      const gsl::not_null<std::optional<Scalar<DataVector>>*>
+          neutron_star_interior_mask,
+      const tnsr::I<DataVector, 3, Frame::Inertial>& inertial_coords,
+      const evolution::initial_data::InitialData& solution_or_data) {
+    using all_data_and_solutions =
+        tmpl::at<typename Metavariables::factory_creation::factory_classes,
+                 evolution::initial_data::InitialData>;
+
+    call_with_dynamic_type<void, all_data_and_solutions>(
+        &solution_or_data, [&neutron_star_interior_mask,
+                            &inertial_coords](const auto* initial_data_ptr) {
+          using InitialData = std::decay_t<decltype(*initial_data_ptr)>;
+
+          if constexpr (detail::is_interior_mask_callable_r_v<
+                            std::optional<Scalar<DataVector>>, InitialData,
+                            tnsr::I<DataVector, 3, Frame::Inertial>>) {
+            (*neutron_star_interior_mask) =
+                (*initial_data_ptr).interior_mask(inertial_coords);
+          }
+        });
+  }
+};
+
+}  // namespace ForceFree
--- a/src/Evolution/Executables/ForceFree/EvolveForceFree.hpp
+++ b/src/Evolution/Executables/ForceFree/EvolveForceFree.hpp
@@ -40,6 +40,7 @@
 #include "Evolution/Systems/ForceFree/BoundaryCorrections/RegisterDerived.hpp"
 #include "Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp"
 #include "Evolution/Systems/ForceFree/ElectromagneticVariables.hpp"
+#include "Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp"
 #include "Evolution/Systems/ForceFree/System.hpp"
 #include "Evolution/Systems/ForceFree/Tags.hpp"
 #include "IO/Observer/Actions/RegisterEvents.hpp"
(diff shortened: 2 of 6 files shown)


src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
```diff
@@ -0,0 +1,106 @@
...
+namespace ForceFree {
+/*!
+ * \brief Assign the masking scalar variable (see Tags::NsInteriorMask) at the
+ * initialization phase in NS magnetosphere simulations.
+ *
+ * For an isolated NS magnetosphere simulations, center of the NS is located at
+ * the origin and coordinate radius of the NS is rescaled to 1. Grid points
+ * with $x^2 + y^2 + z^2 < 1.0$ are identified as interior, and are subject to
+ * correction of evolved variables in each time steps to enforce the MHD
+ * condition (see ImposeMhdConditionInsideNs)
```
yoonso0-0: `ImposeMhdConditionInsideNS` is the name of another mutator for which I'll submit another PR soon.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
```diff
@@ -0,0 +1,106 @@
...
+  static void apply(
+      const gsl::not_null<std::optional<Scalar<DataVector>>*>
+          neutron_star_interior_mask,
+      const tnsr::I<DataVector, 3, Frame::Grid>& grid_frame_coords,
+      const evolution::initial_data::InitialData& solution_or_data) {
+    using all_data_and_solutions =
+        tmpl::at<typename Metavariables::factory_creation::factory_classes,
+                 evolution::initial_data::InitialData>;
+
+    using single_star_id_list =
```
yoonso0-0: @wthrowe This is the actual usage of #5272 , I welcome any better idea for this.

wthrowe: How about adding
```c++
  static auto variables(const tnsr::I<DataVector, 3>& coords,
                        tmpl::list<Tags::NsInteriorMask> /*meta*/)
      -> tuples::TaggedTuple<Tags::NsInteriorMask>;
```
or some similar method to the initial data classes?

yoonso0-0: and change this mutator to just call the `variables` function?

wthrowe: Right.

yoonso0-0: Sounds good.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
```diff
@@ -0,0 +1,89 @@
...
+ * initialization phase in NS magnetosphere simulations.
+ *
+ * For an isolated NS magnetosphere simulations, center of the NS is located at
+ * the origin and coordinate radius of the NS is rescaled to 1. Grid points
+ * with $x^2 + y^2 + z^2 < 1.0$ are identified as interior, and are subject to
+ * correction of evolved variables in each time steps to enforce the MHD
+ * condition (see ImposeMhdConditionInsideNs)
+ *
+ * \note The location of NS interior (mask) is fixed in the Grid frame to avoid
+ * the NS surface crossing grid points during the evolution.
```
wthrowe: Use the inertial frame because that's how the fields are defined.  The result can't vary from call to call because initialization mutators are only called once.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
```diff
@@ -0,0 +1,89 @@
...
+                     ForceFree::Solutions::available_solutions<
+                         SimulationType::IsolatedNsMagnetosphere>>;
+
+    call_with_dynamic_type<void, all_data_and_solutions>(
+        &solution_or_data, [&neutron_star_interior_mask,
+                            &grid_frame_coords](const auto* initial_data_ptr) {
+          using InitialData = std::decay_t<decltype(*initial_data_ptr)>;
+
+          if constexpr (tmpl::list_contains_v<ns_magnetosphere_id_list,
+                                              InitialData>) {
```
wthrowe: Do the call unconditionally.  I think if you use `Tags::AnalyticSolutionOrData` in `argument_tags` you can even avoid the `call_with_dynamic_type`.  The function in `src/Evolution/Initialization/InitialData.hpp` might be useful, but I'm not sure on that one.

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
```diff
@@ -123,6 +123,35 @@ tuples::TaggedTuple<Tags::TildeQ> RotatingDipole::variables(
...
 }
 
+tuples::TaggedTuple<Tags::NsInteriorMask> RotatingDipole::variables(
+    const tnsr::I<DataVector, 3, Frame::Grid>& grid_frame_coords,
+    tmpl::list<Tags::NsInteriorMask> /*meta*/) {
+  std::optional<Scalar<DataVector>> result{};
+
+  DataVector r_squared = get(dot_product(grid_frame_coords, grid_frame_coords));
+
+  // NS radius rescaled to 1.0 on grid.
```
wthrowe: Is this true?  I don't see anything in the fields doing this, or is the NS region arbitrary?

wthrowe (changes requested):
Haven't reviewed the test.  Will do that once we have the design ironed out.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
yoonso0-0: Hmm maybe I'll remove this `\note` comment and use inertial frame coordinates for initialization.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
yoonso0-0: 1. `Forcefree` system is using runtime initial data based on the factory infrastructure, which was why I used `call_with_dynamic_type`. Not sure if there is a way around.
2. If run unconditionally, wouldn't this raise "tag is not retrievable" error for other types of initial data? e.g. simple 1D wave tests which we don't need masking. My idea was to have the 
```
  static auto variables(const tnsr::I<DataVector, 3>& coords,
                        tmpl::list<Tags::NsInteriorMask> /*meta*/)
      -> tuples::TaggedTuple<Tags::NsInteriorMask>;
```
only declared for NS magnetosphere initial data.

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
yoonso0-0: We set `c=1` in this evolution system so the length (=time) scale can be arbitrarily rescaled. I could add some option like `NsRadius`, but thought it'd be convenient to just fix `R_* = 1.0` and interpret all the results from this initial data in terms of `R_S`. For example, `t=1.0` in the code unit would correspond to `R_S / c` in physical unit.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
wthrowe: 1. Yes, you're right.  I was looking at code that hasn't been converted to run-time yet.
2. You'll have to either implement it for the others, or make it it's own function instead of a `variables` specialization and check if that function is callable.

src/PointwiseFunctions/AnalyticData/ForceFree/RotatingDipole.cpp
wthrowe: No that's fine.  I was expecting the fields to have a different form inside the star, but I see now that that isn't the case.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
yoonso0-0: Okay, I'll change this to something other than the `variables` specialization.

yoonso0-0:
I pushed the fixed version, and since there were multiple fixes I just squashed into one. The mutator itself actually became simpler.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
wthrowe: Use `CREATE_IS_CALLABLE` instead of hardcoding the solutions that need masks.

src/Evolution/Systems/ForceFree/MaskNeutronStarInterior.hpp
```diff
@@ -0,0 +1,84 @@
...
+#include "Utilities/TMPL.hpp"
+
+namespace ForceFree {
+/*!
+ * \brief Assign the masking scalar variable (see Tags::NsInteriorMask) at the
+ * initialization phase in NS magnetosphere simulations.
+ *
+ * For an isolated NS magnetosphere simulations, center of the NS is located at
+ * the origin and coordinate radius of the NS is rescaled to 1. Grid points
+ * with $x^2 + y^2 + z^2 < 1.0$ are identified as interior, and are subject to
```
wthrowe: Move this information to the solution, since that's where it is now implemented.  (Or just remove it if you think it's covered sufficiently there already.)

wthrowe (changes requested):

yoonso0-0:
Okay, I could remove the `enum` class thing!

wthrowe:
This is ready to be squashed, but I suggest waiting until #5281 is merged and you can rebase on it to reduce the stress on CI a bit.

yoonso0-0:
Okay, thanks for the review! I'll close #5272

wthrowe (approved):
@yoonso0-0 This is ready to go, yes?  I think the dependent label is just left over from before the last update.

yoonso0-0:
@wthrowe Yes

yoonso0-0:
Rebased on the latest develop after a couple of CI failures

PR #5296, opened 2023-08-02
Add observation of inertial center of mass for BNS

ffoucart:
## Proposed changes

Adds to the Bns executable the calculation of the inertial center of mass. More specifically, the code will output integrals of tildeD and tildeD * x^i  with x^i in the inertial frame. The integrands are set to zero for x>0 or x<0 in the grid frame, to follow each neutron star individually.

### Upgrade instructions

None

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

This is simpler than creating a new observation event with reduction of the appropriate integrand, but does have the disadvantage of adding new quantities to the observation databox.

### DIFF
--- a/src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
+++ b/src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
@@ -446,18 +446,22 @@ struct GhValenciaDivCleanTemplateBase<
               use_control_systems,
               tmpl::list<
                   hydro::Tags::TildeDInHalfPlaneCompute<
-                      DataVector, volume_dim, ::domain::ObjectLabel::A,
+                      DataVector, volume_dim,
+                      ::hydro::HalfPlaneIntegralMask::PositiveXOnly,
                       Events::Tags::ObserverCoordinates<3, Frame::Grid>>,
                   hydro::Tags::TildeDInHalfPlaneCompute<
-                      DataVector, volume_dim, ::domain::ObjectLabel::B,
+                      DataVector, volume_dim,
+                      ::hydro::HalfPlaneIntegralMask::NegativeXOnly,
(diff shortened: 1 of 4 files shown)


wthrowe:
I believe the plan discussed elsewhere is to modify this to not use `ObjectLabel`, but other than that this looks good.

ffoucart:
> I believe the plan discussed elsewhere is to modify this to not use `ObjectLabel`, but other than that this looks good.

I might have misunderstood the concerns; I thought that we wanted to avoid including these automatically for the non-BNS executable, which is why I waited for the BNS executable to be created (the observers are only used if we have a control system, which assumes two objects).
I don't mind using a different template though, if there is a preference for that.

wthrowe:
From @nilsdeppe, responding to me asking elsewhere whether I was misremembering the discussion:
> No, it was that, and the recommendation was to use a bool like `PositiveXOnly` or something

So, avoid using domain stuff as much as possible and switch to a bool for which object.

ffoucart:
Ok, sounds good. Probably still need some sort of enum as there are 3 cases, but that's not a problem.

nilsvu:
Can be a `std::optional<bool> positive_x_only` (or an enum)

wthrowe:
I don't think you can template on an `optional`.

nilsvu:
Doesn't need to be a template I think.

wthrowe:
I don't see how you could write the compute tag without a template argument.

nilsvu:
You're right, the compute tag needs a template parameter! Then I guess a new enum specific to this function is best?

wthrowe:
Sounds like a good way to go.

ffoucart:
It should be pretty easy; `PositiveXOnly, NegativeXOnly, AllSpace`... and if we ever need other half planes we can just add them. It's a bit duplicative of the ObjectLabel, but does allow us to stay away from that largely unrelated infrastructure.

wthrowe (approved):

PR #5301, opened 2023-08-03
Build PCH with CMake

nilsvu:
## Proposed changes

This is possible now because we require CMake 3.18, and CMake added support for this in 3.16.

Also enables PCH for Pybindings now.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


cmake/SetupCCache.cmake
```diff
@@ -10,9 +10,12 @@ if(USE_CCACHE)
       # In order to make ccache work with precompiled headers we need to:
       # - Hash the header file in the repo that will generate the precompiled
       #   header
-      # - Ignore the precompiled header in the build directory
+      # - In the past we had to ignore the precompiled header in the build
+      #   directory:
+      #   "CCACHE_IGNOREHEADERS=${CMAKE_BINARY_DIR}/SpectrePch.hpp:${CMAKE_BINARY_DIR}/SpectrePch.hpp.gch "
+      #   This doesn't seem to be necessary anymore, but may help if issues
```
nilsdeppe: Is there a reason to remove this? If not, I'd like to keep it as-is. I think actually CI caching was broken completely until these ignores were added. It's been working fine for many years, so why change it?

nilsdeppe (changes requested):
I think this is good. One request, though. Please squash right away

cmake/SetupCCache.cmake
nilsvu: Dropped this change, but removed the `SpectrePch.hpp.gch` part because it doesn't exist anymore. CMake generates the PCH somewhere in `CMakeFiles/`. I'll do some more tests to make sure ccache is working correctly.

nilsvu: I confirmed that the PCH works with caching on CI (see https://github.com/nilsvu/spectre/actions/runs/5864398940). To do this I changed some of the configuration here because the PCH file is generated by CMake somewhere in `CMakeFiles/` now. But I just followed the CCache docs (https://ccache.dev/manual/latest.html#_precompiled_headers) and that works fine. I updated this file (just `SetupCCache.cmake`), so please take a look.

cmake/SetupCCache.cmake
```diff
@@ -22,3 +32,9 @@ if(USE_CCACHE)
     message(STATUS "Could not find ccache")
   endif(CCACHE_FOUND)
 endif(USE_CCACHE)
+
+# Add `-fno-pch-timestamp` flag to Clang to support precompiled headers
+# (see https://ccache.dev/manual/4.8.2.html#_precompiled_headers)
+if(CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  list(APPEND CMAKE_CXX_COMPILE_OPTIONS_CREATE_PCH -Xclang -fno-pch-timestamp)
```
nilsdeppe, line 42: Overall LGTM.

This doesn't seem to be part of the public CMake API, so I'm a bit worried this could break. We should at least document that with a comment. See https://discourse.cmake.org/t/ccache-clang-and-fno-pch-timestamp/7253/6 for a discussion

nilsvu: Sounds good, added the comment

nilsdeppe (approved):

PR #5317, opened 2023-08-04
Avoid FPE for BBH observation horizon finds

knelli2:
## Proposed changes

See code comment for why this happens.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
+++ b/src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
@@ -155,6 +155,7 @@
 #include "Utilities/ErrorHandling/FloatingPointExceptions.hpp"
 #include "Utilities/ErrorHandling/SegfaultHandler.hpp"
 #include "Utilities/Functional.hpp"
+#include "Utilities/GetOutput.hpp"
 #include "Utilities/NoSuchType.hpp"
 #include "Utilities/ProtocolHelpers.hpp"
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
@@ -180,6 +181,30 @@ namespace Parallel {
 template <typename Metavariables>
 class CProxy_GlobalCache;
 }  // namespace Parallel
+
+// We need separate time tags for all horizon finders because of complicated
+// Interpolator internal details. So we just make a simple compute tag that
+// takes the actual time out of the box since we still want the actual time to
+// be the same, just a different tag.
+namespace Tags {
+template <size_t Index>
+struct AhObservationTime {
+  static std::string name() { return "AhObservationTime" + get_output(Index); }
+  using type = double;
+};
+
+template <size_t Index>
+struct AhObservationTimeCompute : AhObservationTime<Index>, db::ComputeTag {
+  using argument_tags = tmpl::list<::Tags::Time>;
+  using base = AhObservationTime<Index>;
+  using return_type = double;
+
+  static void function(const gsl::not_null<double*> ah_time,
+                       const double time) {
+    *ah_time = time;
+  }
+};
+}  // namespace Tags
 /// \endcond
 
 // Note: this executable does not use GeneralizedHarmonicBase.hpp, because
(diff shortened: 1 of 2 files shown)


src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
```diff
@@ -180,6 +180,28 @@ namespace Parallel {
 template <typename Metavariables>
 class CProxy_GlobalCache;
 }  // namespace Parallel
+
+// We don't use the Tags::TimeAndPreviousCompute in
```
nilsdeppe: Will this also be an issue for BNS when a BH forms? If so, should we handle this either in `TimeAndPrevious` or move this to `Time/Tags/`?

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
```diff
@@ -180,6 +180,28 @@ namespace Parallel {
...
+// observation horizon finds, we don't care about the previous time because
+// it'll never be used so we just set it to nullopt.
+template <size_t Index>
+struct TimeAndPreviousCompute : Tags::TimeAndPrevious<Index>, db::ComputeTag {
+  using argument_tags = tmpl::list<::Tags::Time>;
+  using base = Tags::TimeAndPrevious<Index>;
+  using return_type = LinkedMessageId<double>;
+
+  static void function(
+      gsl::not_null<LinkedMessageId<double>*> time_and_previous,
```
nilsdeppe: `const gsl`

nilsdeppe (changes requested):
You can squash any changes

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
knelli2: No this won't be an issue for BNS because there will only be one horizon finder there.

nilsdeppe: Okay, sounds good!

nilsdeppe (dismissed):

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
```diff
@@ -180,6 +180,28 @@ namespace Parallel {
...
 }  // namespace Parallel
+
+// We don't use the Tags::TimeAndPreviousCompute in
+// Time/Tags/TimeAndPrevious.hpp because it uses the
+// ::evolution::Tags::PreviousTriggerTime tag and that tag is set to NaN outside
+// of the dense triggers. This is going to be used in the observation horizon
+// finds which are typically run on standard triggers so using
+// ::evolution::Tags::PreviousTriggerTime would cause an FPE. For the
+// observation horizon finds, we don't care about the previous time because
+// it'll never be used so we just set it to nullopt.
```
wthrowe: This sounds like a workaround for a bug in the horizon finder.  Why would you ignore the second field in a `LinkedMessageId`?  That's the whole point of the class.  It's like using an atomic type and then never performing any atomic operations on it.

wthrowe (changes requested):

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
knelli2: This is precisely a workaround for a _feature_ of the horizon finder. I don't think trying to explain it here is a good idea, so let's talk about it at the Monday call.

knelli2:
@wthrowe Made the changes we discussed after the meeting today.

knelli2:
So many took 6+ hours...

knelli2:
Forgot to add a `name()` function to the new tag.

wthrowe (approved):
Looks fine.  This will probably all change if we switch to using counters to identify AH finds, so I'm not too worried about design decisions we make here.

wthrowe:
@nilsdeppe Do you want to look at this again?  I think it's been largely rewritten since your review.

nilsdeppe (approved):

PR #5318, opened 2023-08-05
Add protocols and docs for registration

nilsvu:
## Proposed changes

Document the element registration metavars, in the spirit of #3076.

The fixup commit is to help reviewers.

We can also consider retrieving the list of registrars directly from the `Parallel::Phase::Registration` phase of the parallel component.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/Protocols/ElementRegistrar.hpp
+++ b/src/Parallel/Protocols/ElementRegistrar.hpp
@@ -1,0 +1,37 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+namespace Parallel::protocols {
+/*!
+ * \brief Conforming types register and deregister array elements with other
+ * parallel components
+ *
+ * For example, array elements may have to register and deregister with
+ * interpolators or observers when the elements get created and destroyed during
+ * AMR or migrated during load balancing. They may do so by sending messages to
+ * the parallel components that notify of the creation and destruction of the
+ * array elements.
+ *
+ * Conforming classes have the following static member functions:
+ *
+ * ```cpp
+ * static void perform_registration<ParallelComponent>(const
+ *   db::DataBox<DbTagList>& box, Parallel::GlobalCache<Metavariables>& cache,
+ *   const ArrayIndex& array_index)
+ *
+ * static void perform_deregistration<ParallelComponent>(const
+ *   db::DataBox<DbTagList>& box, Parallel::GlobalCache<Metavariables>& cache,
+ *   const ArrayIndex& array_index)
+ * ```
+ *
+ * Here is an example implementation of this protocol:
+ *
+ * \snippet RegistrationHelpers.hpp element_registrar_example
+ */
+struct ElementRegistrar {
+  template <typename ConformingType>
+  struct test {};
+};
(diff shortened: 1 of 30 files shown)


src/Parallel/Protocols/ElementRegistrar.hpp
```diff
@@ -0,0 +1,34 @@
...
+ *
+ * For example, array elements may have to register and deregister with
+ * interpolators or observers when the elements get created and destroyed during
+ * AMR or migrated during load balancing. They may do so by sending messages to
+ * the parallel components that notify of the creation and destruction of the
+ * array elements.
+ *
+ * Conforming classes have the following static member functions:
+ *
+ * - `static void perform_registration<ParallelComponent>(const
```
knelli2: Make these `\code ... \endcode` or actual `\snippet`s? Makes the docs easier to read.

knelli2 (changes requested):
Just a small thing. Can be squashed in

nilsdeppe (commented):
I have no additional comments. No need to wait for my explicit approval to merge.

src/Parallel/Protocols/ElementRegistrar.hpp
nilsvu: There's a `\snippet` below. I think this list of requirements is fine as explicit text.

nilsvu:
Rebased

src/Parallel/Protocols/ElementRegistrar.hpp
knelli2: If you have the `\snippet` below, why do you need to rewrite the signature in a one-line comment? Just say "This protocol requires two functions, `perform_registration` and `perform_deregistration`, that follow the signatures given below in the example" or something to that effect.

In my opinion, the unformatted, single-color, single-line signature is hard to read.

nilsvu: Tried it, but then you have to look into the file where the snippet is to see the signature.

nilsvu: Switched to a code block now instead of the inline code

knelli2: > Tried it, but then you have to look into the file where the snippet is to see the signature.

Really? Then what does the code snippet actually show in the docs? Because I would have thought it showed the whole class (based on your changes) which should definitely have the signatures.

nilsvu: In the docs it prints the snippet, but in the code there's just the `\snippet` so you have to find it. But the code block prints well.

knelli2: Oh ok, I see now. Guess this is dependent on how we want people to discover/use documentation. For devs, I feel it's usually easier to go to the source file and read there, so in that case explicitly writing out the signature is helpful. However, anybody who uses the website (maybe contributors) will now see duplicated signatures (or possibly out of sync signatures if we forget to update the docs). Unclear which we should make the norm right now but that discussion is not for this PR. This looks good for now.

knelli2 (approved):

PR #5319, opened 2023-08-05
Mock global cache

nilsvu:
## Proposed changes

Allows to run most tests outside a Charm++ environment.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/FindCharm.cmake
+++ b/cmake/FindCharm.cmake
@@ -385,6 +385,15 @@ target_link_libraries(
   ${CHARM_LIB_ckmain}
   ${CHARM_LIB_conv-static}
   )
+# Add a preprocessing definition to indicate that a Charm++ main module is
+# available. This is used to conditionally compile Charm++-specific code
+# that assumes it is running in a Charm++ environment, such as the
+# `Parallel/CharmRegistration.hpp` code.
+target_compile_definitions(
+  Charmxx::main
+  INTERFACE
+  SPECTRE_CHARM_HAS_MAIN
+  )
 
 # Find and link MPI if necessary
 #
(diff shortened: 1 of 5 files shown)


cmake/FindCharm.cmake
```diff
@@ -385,6 +385,11 @@ target_link_libraries(
   ${CHARM_LIB_ckmain}
   ${CHARM_LIB_conv-static}
   )
+target_compile_definitions(
```
nilsdeppe, line 392: It might be a good idea to add a brief comment here what one of the intended uses for this definition is. basically just the long part of the commit message.

nilsdeppe (changes requested):
LGTM, one small suggestion, squash & rebase if you decide to make the change

cmake/FindCharm.cmake
nilsvu: ok, done

nilsdeppe (dismissed):

src/Parallel/GlobalCache.hpp
```diff
@@ -787,8 +814,10 @@ template <typename GlobalCacheTag, typename Function, typename Metavariables,
           typename... Args>
 void mutate(GlobalCache<Metavariables>& cache, Args&&... args) {
   if (cache.get_main_proxy().has_value()) {
-    cache.thisProxy.template mutate<GlobalCacheTag, Function>(
-        std::make_tuple<Args...>(std::forward<Args>(args)...));
+    if constexpr (not GlobalCache<Metavariables>::is_mocked) {
+      cache.thisProxy.template mutate<GlobalCacheTag, Function>(
+          std::make_tuple<Args...>(std::forward<Args>(args)...));
+    }
```
knelli2: So if the main proxy somehow has been set but the cache is mocked, this won't do anything?

knelli2 (changes requested):

src/Parallel/GlobalCache.hpp
nilsvu: I'll add an error to avoid silently hitting this for some reason

nilsvu: done

knelli2 (approved):

PR #5324, opened 2023-08-07
Allow FPEs to be caught as C++ exceptions

wthrowe:
This mostly worked on clang already, but gcc needed a command line flag.

~~For some reason that I haven't investigated, FPEs end up disabled after catching the exception.  This shouldn't be a big deal because in real use this will presumably be followed immediately by the failure cleanup code, which disables FPEs anyway.~~ - fixed

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/ErrorHandling/Test_FloatingPointExceptions.cpp
+++ b/tests/Unit/ErrorHandling/Test_FloatingPointExceptions.cpp
@@ -5,6 +5,7 @@
 
 #include <cmath>
 #include <limits>
+#include <stdexcept>
 
 #include "Utilities/ErrorHandling/FloatingPointExceptions.hpp"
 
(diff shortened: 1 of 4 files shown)


tests/Unit/ErrorHandling/Test_FloatingPointExceptions.cpp
```diff
@@ -18,58 +19,57 @@
...
+  asm("");
+  throw std::runtime_error("wrong");
+}
+}  // namespace
+
+SPECTRE_TEST_CASE("Unit.ErrorHandling.FloatingPointExceptions",
+                  "[ErrorHandling][Unit]") {
+#if defined(__APPLE__) and defined(__arm64__)
   CHECK(true);
+  return;
```
nilsdeppe: Do you need the `return` here? I don't remember if returning from a test case gives issues with Catch

nilsdeppe (commented):
LGTM, one question/suggestion. You can squash immediately

tests/Unit/ErrorHandling/Test_FloatingPointExceptions.cpp
wthrowe: It is unnecessary.  I'll remove it.

nilsvu:
> For some reason that I haven't investigated, FPEs end up disabled after catching the exception. This shouldn't be a big deal because in real use this will presumably be followed immediately by the failure cleanup code, which disables FPEs anyway.

Does this happen only in tests or also in executables? Catch might be messing with the signal handlers in tests.

wthrowe:
The signal handlers are set up correctly.  FPE trapping turns off, so we don't get a signal to handle.  Possibly they turn off once one is caught?

...which suggests the solution of just turning them back on in the code that catches them, which works.

nilsvu (dismissed):
Could you rebase just to make sure this still works with Catch2 v3?

nilsvu (approved):

nilsdeppe (approved):

PR #5326, opened 2023-08-07
Add events to CCE

knelli2:
## Proposed changes

Adds two new events, `Cce::Events::ObserveFields` and `Cce::Events::ObserveTimeStep`, which are different from the GH ones because CCE is run on a singleton and has volume data written differently. Then adds these two events and the `RunEventsAndTriggers` action to the `CharacteristicEvolution` component, but only when CCM is disabled.

### Upgrade instructions

All input files for the standalone CCE executable `CharacteristicExtract` must now have an
```yaml
EventsAndTriggers:
  - Trigger:
      ...
    Events:
      ...
```
block in the input file.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Evolution/Systems/Cce/Events/ObserveFields.hpp
```diff
@@ -0,0 +1,421 @@
...
+  }
+}
+}  // namespace detail
+
+/*!
+ * \brief Event to observe fields/variables in a characteristic evolution.
+ *
+ * \details Similar to `dg::Events::ObserveFields`, this event will write volume
+ * data from the characteristic domain to disk when triggered. However, there a
+ * several differences which are important to highlight.
```
wthrowe: "there are several"

src/Evolution/Systems/Cce/Events/ObserveFields.hpp
```diff
@@ -0,0 +1,421 @@
...
+ *
+ * First is the fields themselves. The DG event takes the fields to observe as
+ * template parameters because the event must work with many evolution systems.
+ * However, since this event is specific to the characteristic evolution system,
+ * we can hardcode the list of fields that are available to observe. The fields
+ * available to observe are the following tags along with their first and second
+ * `Cce::Tags::Dy` derivatives (see `Cce::Tags::Dy` for a definition of y`):
```
wthrowe: You have an unmatched backtick at the end.  Not sure if you are missing any other intended characters as well.

src/Evolution/Systems/Cce/Events/ObserveFields.hpp
```diff
@@ -0,0 +1,421 @@
...
+ * our typical way of writing/storing volume data won't work.
+ *
+ * This event writes its data in the following structure in the H5 file:
+ * `/Cce/VolumeData/TagName/CompactifiedRadius_X.dat`. Every field that is
+ * observed will get its own subgroup called `TagName`. In this subgroup, there
+ * will be N files corresponding to N radial grid points named
+ * `CompactifiedRadius_X.dat` where `X` here will range from 0 to `N-1`. We call
+ * these compactified radii because for a more "physical" radius, it goes to
+ * infinity at future-null infinity and we can't write that in a file. Instead,
+ * these N files will correspond to the compactified coordinate $y = 1 - 2R/r$
```
wthrowe, line 104: Your equations aren't rendering correctly.  I don't see an obvious cause.

src/Evolution/Systems/Cce/Events/ObserveTimeStep.hpp
```diff
@@ -0,0 +1,151 @@
...
+ * \brief %Observe the size of the time steps on the characteristic evolution.
+ *
+ * Writes reduction quantities:
+ * - `%Time`
+ * - `Time Step`
+ *
+ * The subfile will be written into the `/Cce` subgroup.
+ *
+ * All values are reported as positive numbers, even for backwards
+ * evolutions.
```
wthrowe: Does backwards CCE make sense?

wthrowe (changes requested):

src/Evolution/Systems/Cce/Events/ObserveTimeStep.hpp
knelli2: I don't think so? Considering these are future null slices. Must have been left over when I copied the DG event. I'll get rid of it.

src/Evolution/Systems/Cce/Events/ObserveFields.hpp
knelli2: This rendered locally for me once I fixed the missing back-tick above.

wthrowe: Yep, that did it.

wthrowe (dismissed):

wthrowe:
Note to core devs: This is waiting on a self-requested review.

src/Evolution/Systems/Cce/Events/ObserveFields.hpp
```diff
@@ -0,0 +1,420 @@
...
+ * modal set of data so we don't repeat it N times.
+ *
+ * All data will be written into the `observers::OptionTags::ReductionFileName`
+ * file.
+ */
+class ObserveFields : public Event {
+  template <typename Tag, bool IncludeSecondDeriv = true>
+  // clang-format off
+  using zero_one_two_derivs = tmpl::flatten<tmpl::list<
+      Tag,
```
Sizheng-Ma, line 137: It might be nice to be more specific, say 'radial deriv'

src/Evolution/Systems/Cce/Events/ObserveFields.hpp
```diff
@@ -0,0 +1,420 @@
...
+                 zero_one_two_derivs<Tags::BondiU>,
+                 zero_one_two_derivs<Tags::BondiQ>,
+                 zero_one_two_derivs<Tags::BondiW>,
+                 zero_one_two_derivs<Tags::BondiH, false>,
+                 zero_one_two_derivs<Tags::BondiJ>,
+                 zero_one_two_derivs<Tags::Du<Tags::BondiJ>>,
+                 Tags::BondiR,
+                 Tags::EthRDividedByR,
+                 Tags::DuRDividedByR,
+                 Tags::OneMinusY>>;
```
Sizheng-Ma: It's not reasonable to consider the model data of `Tags::OneMinusY` since it's just the radial coordinate of CCE. Although practically `Tags::OneMinusY` is constant on a sphere, which subsequently means that the `(0,0)` model data is the one we want, the mode decomposition makes the data interpretation redundant. 

`Tags::OneMinusY` remains constant over time once it's created at the beginning of the simulation (collocation points). So I feel it's more reasonable to dump it once at the very beginning and then we don't have to worry about it afterwards.

Sizheng-Ma (changes requested):

Sizheng-Ma:
For future reference, the modal decomposition implemented here is on a code time slice (known as 'partially flat gauge') rather than the actual physical time (known as 'Bondi time'). This may introduce some gauge effects in the dumped data. Users should be aware of this. Maybe add this comment somewhere in the code as a warning. 

A way to get around this is to use `Cce::ScriPlusInterpolationManager`, which interpolates the data to the same Bondi time slices before the decomposition, but it will make this PR more complicated.

knelli2:
Ok @Sizheng-Ma @wthrowe pushed a fixup so I only write `OneMinusY` once, and don't write it in modal form. I had to make a member variable `mutable` because the `operator()` of the event is const. Couldn't really think of another way.

wthrowe (changes requested):
No.  No state in Events.

Write the coordinates with every observation.  That's what we do for the volume evolution, and I see no reason for this to be special.

wthrowe:
For a little more context, mutables aren't just about thread safety.  We don't want to guarantee any particular behavior regarding recreating Events after a restart.  They may be serialized or they may be recreated from scratch, so those have to have the same result.

knelli2:
Ok removed the state. It does seem a bit redundant to write the exact same data every time, especially when it's guaranteed to never change, but it shouldn't matter for performance so it's alright.

Sizheng-Ma:
Looks good to me. Please squash.

wthrowe:
Squash.

You've made writing OneMinusY optional, so if you want to only write it once you can write it using a separate trigger that only fires once at the start and leave it out of the main observation.

knelli2:
Squashed.

Sizheng-Ma (approved):

wthrowe (approved):

wthrowe:
CI failure is something weird about the benchmark executable, which I believe to be unrelated to this PR.

PR #5327, opened 2023-08-07
Add thread-safety docs to some FoTs

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/PiecewisePolynomial.hpp
+++ b/src/Domain/FunctionsOfTime/PiecewisePolynomial.hpp
@@ -19,8 +19,13 @@
 
 namespace domain {
 namespace FunctionsOfTime {
-/// \ingroup ComputationalDomainGroup
-/// \brief A function that has a piecewise-constant `MaxDeriv`th derivative.
+/*!
+ * \ingroup ComputationalDomainGroup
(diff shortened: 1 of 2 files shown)


nilsvu (approved):
Thanks for adding these docs 👍

Would it be useful to make it more explicit that a class can be put in the mutable cache by adding a protocol? Then the mutable cache can check that all of its types conform to the protocol. So when you try to put an unsupported type in the mutable cache the compiler says "wait, make sure this type is thread safe please". Of course for derived classes the base class has to write in the docs that subclasses must be thread safe. Just a suggestion, up to you if you think it's useful (can also add an issue).

knelli2:
The thing with the functions of time is that only the `PiecewisePolynomial` and `QuaternionFunctionOfTime` can be updated inside the global cache. The other derived classes don't have a mechanism to be updated, so we can't really put this requirement on the FunctionOfTime base class. Also, the actual tag that gets mutated holds a `map<string, unique_ptr<FoT>>` as its type, and we can't really edit the standard...So it's sort of difficult to guarantee that the type of a tag is thread-safe with a protocol. So overall, I'd say it's fine as is. There really shouldn't be many more mutable cache tags (if any) so we can just enforce this ourselves through code review and requiring docs to be written in the appropriate places.

PR #5329, opened 2023-08-08
Add prefix tags for observation of CSW variables in ScalarTensor

guilara:
## Proposed changes

When specifying variables to observe in `ScalarTensor` input files, variables with matching names in the Generalized Harmonic and CurvedScalarWave (CSW) systems clash with each other. 

Here we write a prefix tag to wrap the CSW tags and add compute tags that retrieve the data from the CSW variables but that can be parsed with the prefix.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ScalarTensor/Tags.hpp
+++ b/src/Evolution/Systems/ScalarTensor/Tags.hpp
@@ -3,10 +3,15 @@
 
 #pragma once
 
+#include <cstddef>
+
 #include "DataStructures/DataBox/DataBoxTag.hpp"
 #include "DataStructures/DataBox/Tag.hpp"
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Constraints.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
 #include "Options/String.hpp"
 #include "Utilities/TMPL.hpp"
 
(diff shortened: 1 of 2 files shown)


nilsdeppe (commented):
This is fine by me. @nikwit could you give your opinion since you use this system?

src/Evolution/Systems/ScalarTensor/Tags.hpp
```diff
@@ -58,6 +61,108 @@ struct ScalarMass : db::SimpleTag {
...
+  }
+  using base = CSW<CurvedScalarWave::Tags::Psi>;
+};
+
+/*!
+ * \brief Compute tag for observing the evolved variable Pi.
+ * \details This tag copies the data from the evolved variable Tag
+ * CurvedScalarWave::Tags::Pi into the wrapped tag used for observation.
+ */
+struct CSWPiCompute : CSW<CurvedScalarWave::Tags::Pi>, db::ComputeTag {
```
nilsdeppe: I think you can combine all these into one:
```cpp
template <typename CswTag>
struct CswCompute : Csw<CswTag>, db::ComputeTag {
  using argument_tags = tmpl::list<CswTag>;
  using base = Csw<CswTag>;
  using return_type = typename base::type;
  static constexpr void function(const gsl::not_null<return_type*> result,
                                    const return_type& csw_var) {
    for (size_t i = 0; i < csw_var.size(); ++i) {
      make_const_view(make_not_null(&std::as_const((*result)[i], csw_var[i], 0, csw_var[i].size());
    }
  }
};
```

src/Evolution/Systems/ScalarTensor/Tags.hpp
```diff
@@ -58,6 +61,108 @@ struct ScalarMass : db::SimpleTag {
...
+
+/*!
+ * \brief Prefix tag to avoid ambiguities when observing variables with the same
+ * name in both parent systems.
+ * \note Since we also add compute tags for these quantities, we do not make
+ * this a derived class of `Tag`. Otherwise, we would have tags with repeated
+ * base tags in the `ObservationBox`.
+ */
+template <typename Tag>
+struct CSW : db::PrefixTag, db::SimpleTag {
```
nilsdeppe: `CSW` -> `Csw` for pascal case

nilsdeppe (commented):
LGTM, I think you can shorten the code quite considerably with the example I give :) Please squash immediately

src/Evolution/Systems/ScalarTensor/Tags.hpp
wthrowe: I've never used the feature, but does this work as a reference tag?  Something like
```c++
template <typename CswTag>
struct CswReference : Csw<CswTag>, db::ReferenceTag {
  using base = Csw<CswTag>;
  using argument_tags = tmpl::list<CswTag>;
  static const typename base::type& get(const typename base::type& tensor) {
    return tensor;
  }
};
```

knelli2: Might be worth trying the reference tag. The `ObservationBox` can handle reference tags, just like the `DataBox`. It'd be nice to avoid `const_cast` if we can.

src/Evolution/Systems/ScalarTensor/Tags.hpp
```diff
@@ -58,6 +61,108 @@ struct ScalarMass : db::SimpleTag {
...
+
+/*!
+ * \brief Compute tag for observing the evolved variable Psi.
+ * \details This tag copies the data from the evolved variable Tag
+ * CurvedScalarWave::Tags::Psi into the wrapped tag used for observation.
+ */
+struct CSWPsiCompute : CSW<CurvedScalarWave::Tags::Psi>, db::ComputeTag {
+  using argument_tags = tmpl::list<CurvedScalarWave::Tags::Psi>;
+  using return_type = Scalar<DataVector>;
+  static constexpr void function(const gsl::not_null<return_type*> result,
```
nikwit: I don't think these functions can be `constexpr`

nikwit (commented):
Looks good, other than the CI failure

src/Evolution/Systems/ScalarTensor/Tags.hpp
guilara: I have tried both options. I only managed to get @nilsdeppe solution to work immediately. Although I left the separate wrappers for the constraints since I want to avoid also adding the compute tags from CSW directly to the ObservationBox (as this would make them available for the user who might then trigger an error if they choose to observe the bare tags and no those with the Csw wrapper.)

Not sure why the ReferenceTag didn't work in the same way as the compute tag. I was getting an error saying that `CswReference<CurvedScalarWave::Tag::Psi>` could not be retrieved from the observation box. Not sure why it was not trying to retrieve the base tag instead.

src/Evolution/Systems/ScalarTensor/Tags.hpp
```diff
@@ -70,67 +72,46 @@ struct ScalarMass : db::SimpleTag {
...
+ * CurvedScalarWave constraints since we do not want to add the bare compute
+ * tags (::CurvedScalarWave::Tags::OneIndexConstraintCompute and
+ * ::CurvedScalarWave::Tags::TwoIndexConstraintCompute) directly in the
+ * ObservationBox, since that will make them observable and would lead to a
+ * clash with the ::gh constraint tags.
  */
 template <size_t Dim>
-struct CSWOneIndexConstraintCompute
-    : CSW<CurvedScalarWave::Tags::OneIndexConstraint<Dim>>,
+struct CswOneIndexConstraintCompute
```
nilsdeppe, line 113: Does the `CswCompute` tag not work on the constraints? Naively I would've expected it to work on any tag that holds some sort of tensor.

nilsdeppe (commented):
LGTM, one question, please go ahead and squash.

I don't understand enough about reference tags to know why it doesn't work and have always just done this non-owning DataVector business to get around it :shrug:

src/Evolution/Systems/ScalarTensor/Tags.hpp
guilara: It does work. But...
in the executable we would have
`using observe_fields = tmpl::list<..., CurvedScalarWave::Tags::TwoIndexConstraintCompute<volume_dim>, ScalarTensor::Tags::CswCompute<CurvedScalarWave::Tags::TwoIndexConstraint<volume_dim>>, ..., ::gh::Tags::TwoIndexConstraintCompute<volume_dim>, ...>`

So in principle someone could still write in the input file:
```
VariablesToObserve:
- Csw(TwoIndexConstraint)
- TwoIndexConstraint

```

and trigger an error when the code tries to write twice in the same file for the two index constraint where for the second variable it is still confused on which one to observe.

So I wanted to avoid adding at all `CurvedScalarWave::Tags::TwoIndexConstraintCompute<volume_dim>` in observe_fields.

At least that's what I understand is happening

nilsdeppe: Oh, I see... Hmm, I don't have a great solution for that at the moment, so doing what you are doing is fine :) Thanks!

nilsdeppe (approved):

knelli2 (commented):
The changes look good to me. Not sure what the changelog documentation error is though. @nilsvu do you know?

nilsvu:
Yea I fixed the changelog thing elsewhere

PR #5333, opened 2023-08-09
Add density dependent limit to magnitude of tildeS

ffoucart:
## Proposed changes

Introduces SpEC-like limits for the magnitude of the tildeS conserved variables (momentum flux). This is needed for stability of the evolution when using 2D equations of state in BNS inspiral.
The default parameters currently in the input file for BNS are what I used in my (only) run with a 2D EoS. They are likely not optimal, but they at least worked once.
All other input files using FixConservatives have been updated so that this new option does nothing. Again, this might not be the best choice, but it should hopefully minimize changes for others.

### Upgrade instructions

All input files using ValenciaDivClean or GhValenciaDivClean executables must add the new options
`SafetyFactorForSCutoffD, SafetyFactorForSSlope`
to
`FixConservatives`

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


ffoucart:
(On my branch, clang-tidy fails due to an error outside of Spectre)

nilsdeppe (approved):

PR #5334, opened 2023-08-10
Fix threadsafety of PiecewisePolynomial

knelli2:
## Proposed changes

Turns out `std::deque` didn't satisfy our requirements for thread-safety but a `std::list` does. Also changes QuaternionFunctionOfTime as well.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
+++ b/src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
@@ -3,7 +3,7 @@
 
 #include "Domain/FunctionsOfTime/FunctionOfTimeHelpers.hpp"
 
-#include <deque>
+#include <list>
 #include <pup.h>
 #include <pup_stl.h>
 
@@ -59,7 +59,7 @@ void reset_expiration_time(const gsl::not_null<double*> prev_expiration_time,
 template <size_t MaxDerivPlusOne, bool StoreCoefs>
 const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
     const double t,
-    const std::deque<StoredInfo<MaxDerivPlusOne, StoreCoefs>>& all_stored_infos,
+    const std::list<StoredInfo<MaxDerivPlusOne, StoreCoefs>>& all_stored_infos,
     const size_t all_info_size) {
   // this function assumes that the times in stored_info_at_update_times is
   // sorted, which is enforced by the update function of a piecewise polynomial.
@@ -68,54 +68,32 @@ const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
   // all_stored_infos because another thread may have added an entry to
   // all_stored_infos during a call to this function.
   ASSERT(all_info_size != 0,
-         "Deque of StoredInfos you are trying to access is empty. Was it "
+         "List of StoredInfos you are trying to access is empty. Was it "
          "constructed properly?");
 
-  // We can't use iterators because when a value is inserted into a std::deque,
-  // iterators are invalidated. We need this operation to be thread-safe so we
-  // only use references which are not invalidated by insertion at either end
-  // points. Because we require that `all_stored_infos` is ordered, we can
-  // safely assume that the index 0 always points to the same StoredInfo in the
-  // deque. This implementation of std::lower_bound is taken directly from
-  // https://en.cppreference.com/w/cpp/algorithm/lower_bound, replacing all
-  // iterators with size_t for indices.
-  const size_t upper_bound_stored_info_index = [&t, &all_stored_infos,
-                                                &all_info_size]() {
-    size_t first = 0;
-    size_t step = 0;
-    size_t count = all_info_size;
-    size_t current_index = 0;
-
-    while (count > 0) {
-      current_index = first;
-      step = count / 2;
-      current_index += step;
-
-      if (all_stored_infos[current_index].time < t) {
-        first = ++current_index;
-        count -= step + 1;
-      } else {
-        count = step;
-      }
+  auto upper_bound_stored_info = all_stored_infos.begin();
+  // Linear search is faster with std::list because of expensive
+  // std::advance/distance inside std::lower_bound
+  for (size_t i = 0; i < all_info_size; i++, upper_bound_stored_info++) {
+    if (not(upper_bound_stored_info->time < t)) {
+      break;
     }
+  }
 
-    return first;
-  }();
-
-  if (upper_bound_stored_info_index == 0) {
+  if (upper_bound_stored_info == all_stored_infos.begin()) {
     // all elements of times are greater than t
     // check if t is just less than the min element by roundoff
-    if (not equal_within_roundoff(all_stored_infos[0].time, t)) {
+    if (not equal_within_roundoff(upper_bound_stored_info->time, t)) {
       ERROR("requested time " << t << " precedes earliest time "
-                              << all_stored_infos[0].time << " of times.");
+                              << upper_bound_stored_info->time << " of times.");
     }
-    return all_stored_infos[0];
+    return *upper_bound_stored_info;
   }
   // t is either greater than all elements of times
   // or t is within the range of times.
-  // In both cases, 'upper_bound_deriv_info_index' is currently one greater than
+  // In both cases, 'upper_bound_stored_info' currently points to one index past
   // the desired index.
-  return all_stored_infos[upper_bound_stored_info_index - 1];
+  return *std::prev(upper_bound_stored_info);
 }
 
 template <size_t MaxDerivPlusOne, bool StoreCoefs>
@@ -168,10 +146,10 @@ GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3, 4, 5), (true, false))
 
 #undef INSTANTIATE
 
-#define INSTANTIATE(_, data)                                                   \
-  template const StoredInfo<DIM(data), STORECOEF(data)>&                       \
-  stored_info_from_upper_bound(                                                \
-      const double, const std::deque<StoredInfo<DIM(data), STORECOEF(data)>>&, \
+#define INSTANTIATE(_, data)                                                  \
+  template const StoredInfo<DIM(data), STORECOEF(data)>&                      \
+  stored_info_from_upper_bound(                                               \
+      const double, const std::list<StoredInfo<DIM(data), STORECOEF(data)>>&, \
       const size_t);
 
 GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3, 4, 5), (true, false))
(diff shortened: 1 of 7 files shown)


src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -214,13 +223,13 @@ template <size_t MaxDeriv>
 std::ostream& operator<<(
     std::ostream& os,
     const PiecewisePolynomial<MaxDeriv>& piecewise_polynomial) {
```
nilsdeppe, line 258: I'm not sure this is threadsafe. Haven't thought about it much, but I worry that calling `end` or `size` could be bad. I'm also not sure we care

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -266,9 +278,9 @@ std::ostream& operator<<(
     std::ostream& os,
     const QuaternionFunctionOfTime<MaxDeriv>& quaternion_f_of_t) {
   os << "Quaternion:\n";
-  for (size_t i = 0; i < quaternion_f_of_t.stored_quaternion_size_; ++i) {
-    os << quaternion_f_of_t.stored_quaternions_and_times_[i];
-    os << "\n";
+  for (auto it = quaternion_f_of_t.stored_quaternions_and_times_.begin();
```
nilsdeppe: Same note about threadsafety on stream

nilsdeppe (dismissed):

src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
```diff
@@ -68,54 +68,31 @@ const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
...
-        first = ++current_index;
-        count -= step + 1;
-      } else {
-        count = step;
-      }
-    }
-
-    return first;
-  }();
+  const auto upper_bound_stored_info = std::lower_bound(
```
wthrowe: `lower_bound` on a list is going to be slower than a linear search, particularly with an expensive `std::next` call as an argument.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -104,9 +105,10 @@ void PiecewisePolynomial<MaxDeriv>::update(
     const double time_of_update, DataVector updated_max_deriv,
     const double next_expiration_time) {
   if (time_of_update <=
-      deriv_info_at_update_times_
-          [deriv_info_size_.load(std::memory_order_relaxed) - 1]
-              .time) {
+      std::next(deriv_info_at_update_times_.begin(),
+                static_cast<long>(
+                    deriv_info_size_.load(std::memory_order_relaxed) - 1))
```
wthrowe: Is this `deriv_info_at_update_times_.back()`?

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
wthrowe: I don't think it's threadsafe.  If we care, you can loop `deriv_info_size_` iterations instead of using the end iterator.

src/Domain/FunctionsOfTime/PiecewisePolynomial.hpp
```diff
@@ -79,13 +79,13 @@ class PiecewisePolynomial : public FunctionOfTime {
   /// Returns the domain of validity of the function,
   /// including the extrapolation region.
   std::array<double, 2> time_bounds() const override {
-    return {{deriv_info_at_update_times_[0].time, expiration_time_}};
+    return {{deriv_info_at_update_times_.begin()->time, expiration_time_}};
```
wthrowe: [optional] (In various places.) I think `.front().` is somewhat more idiomatic than `.begin()->`, but this is a style choice so up to you.

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -113,11 +121,15 @@ void QuaternionFunctionOfTime<MaxDeriv>::update_stored_info() {
           << " stored quaternions.");
 
   // Final time, initial time, and quaternion
-  const double t = angle_deriv_info[last_index].time;
-  const double t0 = angle_deriv_info[last_index - 1].time;
+  const double t =
+      std::next(angle_deriv_info.begin(), static_cast<long>(last_index))->time;
```
wthrowe: Use `back()` or `end()` throughout this section.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
knelli2: Definitely not thread-safe. But I also don't think we care. I can easily change it to be thread-safe though

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -171,18 +173,25 @@ void PiecewisePolynomial<MaxDeriv>::reset_expiration_time(
...
     }
+  } else if (version == 1 and p.isUnpacking()) {
+    // An old version had a std::deque rather than a std::list
+    std::deque<FunctionOfTimeHelpers::StoredInfo<MaxDeriv + 1>> pupped_info{};
+    p | pupped_info;
+    for (auto& pupped_deriv_info : pupped_info) {
+      deriv_info_at_update_times_.emplace_back(std::move(pupped_deriv_info));
+    }
   } else {
     p | deriv_info_at_update_times_;
```
wthrowe: Do we need `pup` to be threadsafe?  I don't think we should ever be serializing the polynomial during an update, but what about sizing?

wthrowe (changes requested):

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
knelli2: Is that threadsafe? I guess since we aren't allowing two threads to update at the same time, it should be ok

wthrowe: Exactly.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
knelli2: No I don't think we should every be (or ever allow) updating during pupping (writing/reading). For sizing, I could just add another `if else (p.isSizing())` call that uses `deriv_info_size_` to only pup a threadsafe amount of things.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
knelli2: Ah yes because this is in the update part, I think it's just `back()`.

knelli2:
Ok pushed a fixup with the requested changes, and rewrote the `pup` function so I believe it is now thread safe for packing, unpacking, and sizing under the assumption that we don't update while serializing or vice versa.

src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
```diff
@@ -71,28 +71,31 @@ const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
...
-  const auto upper_bound_stored_info = std::lower_bound(
-      all_stored_infos.begin(),
-      std::next(all_stored_infos.begin(), static_cast<long>(all_info_size)), t,
-      [](const StoredInfo<MaxDerivPlusOne, StoreCoefs>& d, double t0) {
-        return d.time < t0;
-      });
-
-  if (upper_bound_stored_info == all_stored_infos.begin()) {
+  auto it = all_stored_infos.begin();
+  bool lower = false;
```
wthrowe: It doesn't look like you meaningfully use this.

src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
```diff
@@ -71,28 +71,31 @@ const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
...
 
-  const auto upper_bound_stored_info = std::lower_bound(
-      all_stored_infos.begin(),
-      std::next(all_stored_infos.begin(), static_cast<long>(all_info_size)), t,
-      [](const StoredInfo<MaxDerivPlusOne, StoreCoefs>& d, double t0) {
-        return d.time < t0;
-      });
-
-  if (upper_bound_stored_info == all_stored_infos.begin()) {
+  auto it = all_stored_infos.begin();
```
wthrowe: The old variable name seemed better.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -178,30 +176,66 @@ void PiecewisePolynomial<MaxDeriv>::pup(PUP::er& p) {
...
+      size_t size = 0;
+      p | size;
+      deriv_info_size_.store(size);
+      deriv_info_at_update_times_.clear();
+      deriv_info_at_update_times_.resize(
+          deriv_info_size_.load(std::memory_order_relaxed));
+      // Using .end() here is ok because we won't be updating while
+      // packing/unpacking
+      for (auto it = deriv_info_at_update_times_.begin();
+           it != deriv_info_at_update_times_.end(); it++) {
```
wthrowe: Range based loop?

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -69,30 +69,66 @@ void QuaternionFunctionOfTime<MaxDeriv>::pup(PUP::er& p) {
...
+      size_t size = 0;
+      p | size;
+      stored_quaternion_size_.store(size);
+      stored_quaternions_and_times_.clear();
+      stored_quaternions_and_times_.resize(
+          stored_quaternion_size_.load(std::memory_order_relaxed));
+      // Using .end() here is ok because we won't be updating while
+      // packing/unpacking
+      for (auto it = stored_quaternions_and_times_.begin();
+           it != stored_quaternions_and_times_.end(); it++) {
```
wthrowe: Range-based loop.

wthrowe (changes requested):

wthrowe:
Looks good.

knelli2:
@nilsdeppe Did you want to look at this again?

nilsdeppe:
Oh, sorry, no go ahead!

knelli2:
@wthrowe squashed and rebased

wthrowe (approved):

knelli2:
Rebased after all the CI changes, just to make sure this passes

nilsdeppe (approved):

PR #5340, opened 2023-08-11
Split up RunTests

nilsvu:
## Proposed changes

Also run most tests without Charm++. Only run in a Charm++ environment if needed.

I find significant reductions in unit test runtime with this PR (could be the reduced overhead from eliminating Charm++ and/or the smaller executable). The whole test suit runs more than twice as fast on CI, depending on the configuration, cutting down 10-20 minutes from some CI builds.

| CI Build            | Old             | New             |   Speedup |
|------------------|-----------------|-----------------|-----------|
| gcc-9 Debug      | 12:26 | 4:30 |   2.8x |
| gcc-9 Release    | 8:28 | 2:15 |   3.8x |
| gcc-10 Debug     | 16:19 | 4:16 |   3.8x |
| gcc-10 Release   | 10:28 | 1:51 |   5.7x |
| gcc-11 Debug     | 12:25 | 4:36 |   2.7x |
| gcc-11 Release   | 7:05 | 1:51 |   3.8x |
| clang-10 Release | 9:13 | 2:19 |   4.0x |
| clang-11 Debug   | 32:16 | 8:09 |   4.0x |
| clang-11 Release | 2:21 | 1:47 |   1.3x |
| clang-13 Debug   | 21:21 | 3:45 |   5.7x |

(compared to https://github.com/sxs-collaboration/spectre/actions/runs/5835823727/job/15828143351)

### Upgrade instructions

To run unit tests you don't need to compile the giant `RunTests` executable anymore (or edit `RunSingleTest`). Just build the individual test executable (like `Test_LinearOperators`) and run it (like `./bin/Test_LinearOperators`). See the [Catch2 docs](https://github.com/catchorg/Catch2/blob/devel/docs/command-line.md) for details on running the test executables. You can still build _all_ unit tests with the `unit-tests` target and run test cases with `ctest` (no change there).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


docs/DevGuide/BuildOptimization.md
```diff
@@ -130,12 +130,12 @@ the compilation time or memory usage.
```
nilsdeppe: I'd like to keep it until I have actually used the split executable approach. I have no experience to agree that this won't be needed anymore. I expect so, but I'm not sure.

tests/Unit/TestMain.cpp
```diff
@@ -0,0 +1,45 @@
...
+
+int main(int argc, char* argv[]) {
+  Catch::StringMaker<double>::precision =
+      std::numeric_limits<double>::max_digits10;
+  Catch::StringMaker<float>::precision =
+      std::numeric_limits<float>::max_digits10;
+
+  const int result = Catch::Session().run(argc, argv);
+
+  // In the case where we run all the non-failure tests at once we must ensure
```
nilsdeppe: [optional] I think this comment can get updated as we remove more of the old-style error tests and use `CHECK_THROWS_WITH` instead. Probably something like:
```cpp
// Clean up the python environment only after all tests have finished running,
// since there could be multiple tests run in a single executable launch.
```

If you decide to do this, the comment is also in the `TestCharmMain.cpp` file

nilsdeppe (changes requested):
Overall LGTM, one request: I'd like to keep RunSingleTest for now so I can try the split executables and be completely sure we don't need `RunSingleTest` anymore. Another small comment/wording suggestion. Please squash immediately.

Thanks for doing this big cleanup! :D

docs/DevGuide/BuildOptimization.md
nilsvu: Ok done

nilsdeppe (dismissed):

nilsdeppe:
Hmm, the macOS CI seems to be extremely unhappy. Not sure what is causing that, though.

nilsvu:
Fixed it (I think). Looks like on macOS we still have to make test helper libs static.

nilsdeppe (dismissed):

nilsvu:
Rebased on #5335

nilsdeppe (dismissed):

nilsdeppe (dismissed):

nilsdeppe:
Any idea about the segfault test failure on macOS? I can't tell what's going on

nilsvu:
Fixed the macOS build by running `Test_ErrorHandling` `WITH_CHARM` (like a couple other test execs that require a Charm++ environment).

nilsdeppe (approved):

nilsdeppe:
Thanks!

PR #5345, opened 2023-08-12
Update CMake policies

nilsvu:
## Proposed changes

Avoid a warning with CMake v3.27.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -32,6 +32,14 @@ unset(${SPECTRE_NAME}_VERSION_PATCH)
 if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.19)
   cmake_policy(SET CMP0110 NEW)
 endif ()
+# - We allow both uppercase and case-preserved hints in `find_package` calls, so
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -32,12 +32,6 @@ unset(${SPECTRE_NAME}_VERSION_PATCH)
 if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.19)
```
nilsdeppe, line 32: what do you mean by "now"? Which version of CMake?

nilsvu: Updated the commit message. (CMake policies work with the `cmake_minimum_required(VERSION)`, which is 3.18 for us at the moment. Any policy introduced before that uses the NEW behavior by default, and any policy introduced after uses OLD. CMP0094 was added in v3.15, so when we upgraded to v3.18 it became obsolete to toggle it to NEW.)

nilsdeppe (approved):

PR #5352, opened 2023-08-15
Print deadlock information in BBH exec

knelli2:
## Proposed changes

This should be all information necessary to debug why a deadlock happened. This information is

- `FunctionOfTime` time bounds
- Current control system measurement
- Element info
  - `BoundaryCorrectionAndGhostCellInbox` inbox
  - `NewSlabSize` inbox
  - `MortarNextTemporalId` data box tag
  - `MortarDataHistory` data box tag

Added functions to print all of this info. Most of this should be able to be reused for other executables, I've just only tested this with BBH (because that's the only one that deadlocks at the moment)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Parallel/OutputInbox.hpp
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * \tparam InboxTag Inbox tag with an `output_inbox` member function
+ * \tparam InboxTypes Types of all inboxes (deduced)
+ * \param inboxes All inboxes
+ * \param indent_pad Number of empty spaces to pad the message with incase
+ * indenting is needed.
+ * \return std::string The contents of the inbox
+ */
+template <typename InboxTag, typename... InboxTypes>
+std::string output_inbox(const tuples::TaggedTuple<InboxTypes...>& inboxes,
```
wthrowe, line 36: What is the advantage of this function over calling `InboxTag::output_inbox` directly?

src/Parallel/OutputInbox.hpp
```diff
@@ -0,0 +1,46 @@
...
+                         const size_t indent_pad) {
+  static_assert(tmpl::list_contains_v<tmpl::list<InboxTypes...>, InboxTag>);
+  static_assert(detail::is_output_inbox_callable_v<
+                    InboxTag, const typename InboxTag::type&, const size_t>,
+                "To output an inbox, there must be a static 'output_inbox' "
+                "function defined in the inbox tag.");
```
wthrowe: I generally dislike this sort of check, because I find that it makes error messages less useful.  Here you'll get a generic "there must be a static `output_inbox' function", whereas without this the compiler can specifically say the function doesn't exist or it's not static or it can't deduce a template parameter or the argument type is wrong or whatever.

src/Evolution/DgSubcell/Tags/InitialTciData.hpp
```diff
@@ -44,5 +47,24 @@ struct InitialTciData {
...
+  static std::string output_inbox(const type& inbox,
+                                  const size_t padding_size) {
+    std::stringstream ss{};
+    const std::string pad(padding_size, ' ');
+
+    ss << std::scientific << std::setprecision(16);
+    ss << pad << "InitialTciDataInbox:\n";
+    for (const auto& [current_time_step_id, hash_map] : inbox) {
+      ss << pad << " Current time: " << current_time_step_id.substep_time()
+         << "\n";
```
wthrowe: You probably want the whole id, not just the substep time, since the latter aren't uniquely identifying.

Same comment everywhere else.

src/Evolution/DgSubcell/Tags/InitialTciData.hpp
```diff
@@ -44,5 +47,24 @@ struct InitialTciData {
                << direction_and_element_id.second << " more than once");
     current_inbox.emplace(std::forward<ReceiveDataType>(data));
   }
+
+  static std::string output_inbox(const type& inbox,
```
wthrowe, line 51: All these functions need tests.

tests/Unit/Domain/FunctionsOfTime/Test_OutputTimeBounds.cpp
```diff
@@ -0,0 +1,110 @@
...
+    p | lower_bound_;
+    p | upper_bound_;
+  }
+
+ private:
+  double lower_bound_{};
+  double upper_bound_{};
+};
+
+void avoid_compiler_error() {
```
wthrowe: I assume this should say "warning", but usually we just
```c++
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunused-function"
  WRAPPED_PUPable_decl_template(Whatever);
#pragma GCC diagnostic pop
```

src/Time/BoundaryHistory.hpp
```diff
@@ -473,6 +475,27 @@ std::ostream& BoundaryHistory<LocalVars, RemoteVars, CouplingResult>::print(
   return os;
 }
 
+template <typename LocalVars, typename RemoteVars, typename CouplingResult>
+std::string BoundaryHistory<LocalVars, RemoteVars, CouplingResult>::
+    pretty_print_without_data(const size_t padding_size) const {
```
wthrowe: This is nearly identical to the preceding function.  Can you add a (template?) parameter to that one to control the data printing instead?

src/Evolution/DiscontinuousGalerkin/MortarData.cpp
```diff
@@ -261,6 +263,17 @@ void MortarData<Dim>::pup(PUP::er& p) {
...
 }
 
+template <size_t Dim>
+std::string MortarData<Dim>::pretty_print_current_buffer_no_data(
+    size_t padding_size) const {
+  std::stringstream ss{};
+  ss << std::scientific << std::setprecision(16);
+  const std::string pad(padding_size, ' ');
+  ss << pad << "Current buffer: " << mortar_index_
+     << ", time = " << time_step_id_[mortar_index_].substep_time() << "\n";
```
wthrowe: Another substep time you probably don't want.

src/Evolution/Executables/GeneralizedHarmonic/DeadlockActions.hpp
```diff
@@ -0,0 +1,118 @@
...
+        for (const auto& [key, single_mortar_data] : mortar_data) {
+          ss << "   Key: " << key << ", mortar data:\n";
+          ss << single_mortar_data.pretty_print_current_buffer_no_data(4_st);
+        }
+      }
+    } else {
+      ss << "\n";
+    }
+
+    Parallel::printf(ss.str());
```
wthrowe: `printf("%s", ss.str())`

src/Evolution/Executables/GeneralizedHarmonic/DeadlockActions.hpp
```diff
@@ -0,0 +1,118 @@
...
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/OutputInbox.hpp"
+#include "Parallel/Printf.hpp"
+#include "Time/Actions/ChangeSlabSize.hpp"
+#include "Time/Tags/Time.hpp"
+#include "Utilities/PrettyType.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace gh::deadlock {
+struct PrintControlSystemCurrentMeasurement {
```
wthrowe: We don't generally put code other than the executables themselves in "Executables" directories.  I could be convinced, but I'd like a rationale (as a GitHub comment).

wthrowe (changes requested):

src/Evolution/DgSubcell/Tags/InitialTciData.hpp
knelli2: Is there a `double` that *does* uniquely identify what time we are at? I ask because the printed version of a `TimeStepId` is quite difficult to interpret/understand at a glance, but a `double` is simple so I think it'd be easier for people (me) if I just compared `double`s. Also, we may need to compare these `TimeStepId`s to the FoT expirations (and potentially other times represented as `double`s) and that's difficult to do with the normal stream of a `TimeStepId`.

src/Evolution/Executables/GeneralizedHarmonic/DeadlockActions.hpp
knelli2: I didn't really know where to put these. Right now only the BBH will need them so I considered just putting them in the metavars file, but that makes the file even more crowded and it's already so complicated. However, eventually we'll want to use these (or similar/modified versions) for other executables that can deadlock. The control system one can easily go in the control sys library, however, the element one pulls on a lot of things (and with subcell we'll probably want something related to that as well) so I didn't know where to put it. A new library in the `Evolution/` directory? But how to keep dependencies straight if we have subcell stuff? Suggestions are welcome

src/Parallel/OutputInbox.hpp
knelli2: I thought

```cpp
Parallel::output_inbox<TagName>(inboxes, 0);
```

looked cleaner than

```cpp
TagName::output_inbox(tuples::get<TagName>(inboxes), 0);
```

for users because of the duplicate `TagName`. Also, the second way allows for the chance that you pass the wrong inbox to the wrong tag, even though they have the same type.

src/Evolution/DgSubcell/Tags/InitialTciData.hpp
wthrowe: There isn't, but the substep time is printed at the end of the TimeStepId output:
```
5:Slab[1.34257480385833006,4.67348735890730005]:1/4:1:2.59166701200169403
                                                      ^^^^^^^^^^^^^^^^^^^
```
Still a bunch of noise you have to ignore, but you don't have to do any calculations or anything.

src/Evolution/Executables/GeneralizedHarmonic/DeadlockActions.hpp
wthrowe: That's reasonable.  We can figure out if this is the correct choice when we want to use it more.

knelli2:
Ok @wthrowe added several things:

1. Many fixup commits for the various changes, named by which commit they are fixing up
2. Split up a previous commit that added the deadlock actions to that file in the executable directory into 3 new commits
  a. One to add the control system print
  b. One to add a new deadlock library with the element print action
  c. One to add those two to the BBH exec

Let me know if there's anything else

src/Evolution/DgSubcell/Tags/InitialTciData.hpp
```diff
@@ -56,9 +56,9 @@ struct InitialTciData {
     ss << std::scientific << std::setprecision(16);
     ss << pad << "InitialTciDataInbox:\n";
     for (const auto& [current_time_step_id, hash_map] : inbox) {
-      ss << pad << " Current time: " << current_time_step_id.substep_time()
-         << "\n";
+      ss << pad << " Action number: " << current_time_step_id << "\n";
```
wthrowe: I think "current_time_step_id" is an inaccurate name.  Looks like it's an int.

tests/Unit/Evolution/DgSubcell/Test_InitialTciData.cpp
```diff
@@ -35,6 +38,22 @@ void test() {
...
+                   << "  Action number: 1\n"
+                   << "   Key: " << id0 << ", TCI: 100\n"
+                   << "   Key: " << id1 << ", TCI: 200\n";
+  const std::string expected_inbox_output_v2 =
+      MakeString{} << " InitialTciDataInbox:\n"
+                   << "  Action number: 1\n"
+                   << "   Key: " << id1 << ", TCI: 200\n"
+                   << "   Key: " << id0 << ", TCI: 100\n";
+  // Even though this is a map, for some reason the order gets switch for
+  // different dimensions so we just just either of the two possibilities
```
wthrowe: It's an ordered map of unordered maps.

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_ChangeSlabSize.cpp
```diff
@@ -146,6 +148,17 @@ SPECTRE_TEST_CASE("Unit.CurvedScalarWave.Worldtube.ChangeSlabSize", "[Unit]") {
...
                     step_1);
+
+    const std::string inbox_output =
+        Tags::SphericalHarmonicsInbox<Dim>::output_inbox(worldtube_inbox, 2);
+    const std::string expected_inbox_output =
+        MakeString{} << std::scientific << std::setprecision(16)
+                     << "  SphericalHarmonicsInbox:\n"
+                     << "   Time: " << time_step_id_1 << "\n"
+                     << "    ElementId: " << element_ids.at(1) << "\n"
+                     << "    ElementId: " << element_ids.at(0) << "\n";
```
wthrowe, line 159: Order isn't guaranteed here.

src/Time/Actions/ChangeSlabSize.hpp
```diff
@@ -57,6 +60,22 @@ struct NewSlabSizeInbox
...
+  static std::string output_inbox(const type& inbox,
+                                  const size_t padding_size) {
+    std::stringstream ss{};
+    const std::string pad(padding_size, ' ');
+
+    ss << std::scientific << std::setprecision(16);
+    ss << pad << "NewSlabSizeInbox:\n";
+    for (const auto& [number_of_messages, slab_sizes] : inbox) {
+      (void)slab_sizes;
+      ss << pad << " Number of messages: " << number_of_messages
```
wthrowe: I think this is the slab number, not the number of messages.

src/Time/Actions/ChangeSlabSize.hpp
```diff
@@ -57,6 +60,22 @@ struct NewSlabSizeInbox
...
+
+  static std::string output_inbox(const type& inbox,
+                                  const size_t padding_size) {
+    std::stringstream ss{};
+    const std::string pad(padding_size, ' ');
+
+    ss << std::scientific << std::setprecision(16);
+    ss << pad << "NewSlabSizeInbox:\n";
+    for (const auto& [number_of_messages, slab_sizes] : inbox) {
+      (void)slab_sizes;
```
wthrowe: This is used just below.

tests/Unit/Time/Test_BoundaryHistory.cpp
```diff
@@ -220,46 +220,52 @@ void test_boundary_history() {
...
-                                             << "Data: (2)\n"
-                                             << "Time: Slab[3,3.5]:0/1\n"
-                                             << "Data: (3)\n";
+  std::string expected_output = MakeString{}
+                                << std::scientific << std::setprecision(16)
+                                << "Integration order: 2\n"
+                                << " Local Data:\n"
+                                << "  Time: " << make_time(-1.) << "\n"
+                                << "   Data: " << -1 << "\n"
+                                << "  Time: " << make_time(0.) << "\n"
```
wthrowe, line 229: I like the new indenting.

tests/Unit/Time/Test_BoundaryHistory.cpp
```diff
@@ -220,46 +220,52 @@ void test_boundary_history() {
...
+                   << "  Remote Data:\n"
+                   << "   Time: " << make_time(-2.) << "\n"
+                   << "   Time: " << make_time(-1.) << "\n"
+                   << "   Time: " << make_time(0.) << "\n"
+                   << "   Time: " << make_time(1.) << "\n"
+                   << "   Time: " << make_time(2.) << "\n"
+                   << "   Time: " << make_time(3.) << "\n";
+  std::stringstream ss{};
+  const std::string pretty_output = MakeString{}
+                                    << history.print<false>(ss, 1_st).rdbuf();
```
wthrowe:
```c++
std::stringstream ss{};
history.print<false>(ss, 1_st);
const std::string pretty_output = ss.str();
```

wthrowe (changes requested):
You can squash directly unless one of these turns out to be non-trivial.

knelli2:
Squashed everything.

wthrowe (dismissed):

src/Evolution/DiscontinuousGalerkin/InboxTags.hpp
```diff
@@ -176,6 +179,26 @@ struct BoundaryCorrectionAndGhostCellsInbox {
       }
     }
   }
+
+  static std::string output_inbox(const type& inbox,
```
nilsdeppe, line 183: Should we move these into `cpp` files whenever the template parameters of the tags are easy (`dim`)? Reduce compile time a bit

src/ControlSystem/Actions/PrintCurrentMeasurement.hpp
```diff
@@ -0,0 +1,27 @@
...
+#pragma once
+
+#include "ControlSystem/Tags/SystemTags.hpp"
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Printf.hpp"
+#include "Utilities/PrettyType.hpp"
+
+namespace control_system::Actions {
+struct PrintCurrentMeasurement {
```
nilsdeppe, line 18: add dox so it shows up in dox?

src/Evolution/Deadlock/PrintDgElementArray.hpp
```diff
@@ -0,0 +1,101 @@
...
+#include "Evolution/DiscontinuousGalerkin/InboxTags.hpp"
+#include "Evolution/DiscontinuousGalerkin/MortarTags.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/OutputInbox.hpp"
+#include "Parallel/Printf.hpp"
+#include "Time/Actions/ChangeSlabSize.hpp"
+#include "Time/Tags/Time.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace deadlock {
```
nilsdeppe, line 28: dox

src/Evolution/Deadlock/PrintDgElementArray.hpp
```diff
@@ -0,0 +1,101 @@
...
+struct PrintElementInfo {
```
nilsdeppe, line 47: dox

nilsdeppe (changes requested):
Please squash right away. Anything that's a question is optional

src/Evolution/DiscontinuousGalerkin/InboxTags.hpp
knelli2: Probably yes, but that requires some new files, cmake stuff, and instantiations. I'll try to do it later if I have time

knelli2:
@nilsdeppe squashed

src/Evolution/DiscontinuousGalerkin/InboxTags.hpp
nilsdeppe: Sounds good

nilsdeppe (approved):

PR #5353, opened 2023-08-15
Print more information when an exception is caught by DistributedObject

### LINKED ISSUE #5101: Add ElementId (and more?) to error messages
It would be helpful to supplement error messages with things like the element ID. I think this would be possible by catching exceptions somewhere in `DistributedObject` and then re-throwing them.

kidder:
## Proposed changes

Print the component, array index, phase, and algorithm step in addition to the exception message and type.

Fix #5101 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/DistributedObject.hpp
+++ b/src/Parallel/DistributedObject.hpp
@@ -1185,9 +1185,13 @@ void DistributedObject<ParallelComponent,
         "an issue.");
     sys::abort("");
   }
-  const std::string message = MakeString{}
-                              << "Message: " << exception.what() << "\nType: "
-                              << pretty_type::get_runtime_type_name(exception);
+  const std::string message =
+      MakeString{} << "Component: " << pretty_type::name<parallel_component>()
+                   << "\nArray Index: " << array_index_ << "\n"
+                   << "Phase: " << phase_ << "\n"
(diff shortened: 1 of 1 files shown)


nilsdeppe (dismissed):

src/Parallel/DistributedObject.hpp
```diff
@@ -1185,9 +1185,14 @@ void DistributedObject<ParallelComponent,
         "an issue.");
     sys::abort("");
   }
-  const std::string message = MakeString{}
-                              << "Message: " << exception.what() << "\nType: "
-                              << pretty_type::get_runtime_type_name(exception);
+  const std::string message =
+      MakeString{} << "Component: "
+                   << pretty_type::get_name<parallel_component>() << "\n"
```
knelli2: `pretty_type::name<parallel_component>()` or else all control system (or interpolation target) singletons will have the same name.

knelli2 (changes requested):

src/Parallel/DistributedObject.hpp
kidder: that's not true.  this prints out the full type name including all template arguments...
but `name` is less verbose so I'll make the change since one of the template arguments for DgElementArray is the PDAL

knelli2 (approved):

PR #5361, opened 2023-08-17
Use full path to ccache

nilsvu:
## Proposed changes

Previously, compiling would error if the `ccache` module used to configure the build wasn't available.

Also use the recommended `CMAKE_CXX_COMPILER_LAUNCHER` instead of `RULE_LAUNCH_COMPILE` (see https://cmake.org/cmake/help/latest/prop_gbl/RULE_LAUNCH_COMPILE.html).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupCCache.cmake
+++ b/cmake/SetupCCache.cmake
@@ -17,21 +17,27 @@
 # - Sharing caches between runs on CI works.
 
 option(USE_CCACHE "Use CCache if available to speed up builds" ON)
-if(USE_CCACHE)
-  find_program(CCACHE_FOUND ccache)
-  if(CCACHE_FOUND)
-    string(CONCAT CCACHE_COMMAND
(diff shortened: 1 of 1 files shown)


cmake/SetupCCache.cmake
```diff
@@ -17,21 +17,26 @@
...
+
+find_program(CCACHE_EXEC ccache)
+
+if (NOT CCACHE_EXEC)
+  message(STATUS "Could not find ccache")
+  return()
+endif()
+
+# Invoke compiler through ccache
+set(CMAKE_CXX_COMPILER_LAUNCHER
```
nilsdeppe, line 33: Can you also do this for C?

nilsdeppe (changes requested):

nilsdeppe (approved):

PR #5368, opened 2023-08-18
Add is_isotropic function for mesh, remove debugging CMake message

nilsdeppe:
## Proposed changes

I should've implemented this 6 years ago... It's used a lot in subcell...

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/NumericalAlgorithms/Spectral/Mesh.cpp
```diff
@@ -73,6 +73,17 @@ void Mesh<Dim>::pup(PUP::er& p) {
   p | quadratures_;
 }
 
+template <size_t Dim>
+bool is_isotropic(const Mesh<Dim>& mesh) {
```
nilsvu, line 77: I know you like free functions but this looks like it should be a member function :) mostly for discoverability.

src/NumericalAlgorithms/Spectral/Mesh.hpp
```diff
@@ -254,6 +254,15 @@ class Mesh {
...
 };
 
+/*!
+ * \ingroup DataStructuresGroup
+ * \brief Returns `true` if the mesh is isotropic, `false` otherwise.
+ *
+ * If `Dim` is zero, then `true` is always returned.
+ */
+template <size_t Dim>
+bool is_isotropic(const Mesh<Dim>& mesh);
```
knelli2, line 264: Why not make this a member function of `Mesh` so you can do `mesh.is_isotropic()`? Might be more efficient because you can calculate it in the constructor and just store a bool. Don't know if that actually affects performance though.

nilsdeppe (commented):
Member functions are generally bad. This is a well-known issue in CS. Unless you need access to private data, it really shouldn't be a member function. Here's a talk on why:

https://github.com/CppCon/CppCon2017/blob/master/Presentations/Free%20Your%20Functions/Free%20Your%20Functions%20-%20Klaus%20Iglberger%20-%20CppCon%202017.pdf

https://youtu.be/WLDT1lDOsb4

This does some comparisons and will almost exclusively be used inside ASSERT, so performance is really a non-issue. Having said that, I'm not convinced storing a bool would actually be better because this adds more work for serializing, which we will generally do more than this check. I also don't like the increased complexity of now having to keep a bool synced with 3 other member variables. In my opinion it's just asking for bugs.

src/NumericalAlgorithms/Spectral/Mesh.cpp
nilsvu: (replying on this thread rather than in a new PR comment)

I understand some of the points made in the talk, but not all of them. I think `std::begin()` and `std::end()` are bad examples chosen in the talk because I find those very cumbersome to work with. And I don't understand the counter-argument to "Programmers cannot find free functions as easily as member functions". I think that's actually quite true and is my main concern with the "free all functions" paradigm. I'd know where look for `mesh.is_isotropic()` but I would not know where to find `is_isotropic(mesh)`.

To be clear, I'm not blocking anything here. I would have implemented `is_isotropic` as a member function, but I can live with the free function too :)

nilsdeppe: I actually like `std::begin()`, I don't always use it, but that's because member versions are necessary for data encapsulation. The point is not to free _all_ functions, it is to free any function that doesn't directly need access to private state not otherwise accessible. This actually tightens the encapsulation. In terms of discoverability, this is actually a shortcoming of documentation, not of where to put the function. One solution would be to have a section in the class dox that lists associated free functions, then when someone adds a free function it gets listed. Remove one, it gets removed. I think it's important to not have shortcomings in documentation and documentation tooling (let's face it, Doxygen is okay but could be better) dictate how we write our code. There are major technical advantages to free functions over member functions, and so a solution should be found for documenting them, not choosing an inferior technical solution to document them.

I know :) I think it's a good discussion to have, though!

knelli2:
@kidder @nilsvu Did you want to look at this? If not, I'll just merge it. I still think it makes more sense (especially for the average contributor) to have this as a member function as that's the first logical place I'd look, but since I probably won't be using this ever I'm ok with this as a free function.

nilsvu (approved):

kidder (approved):

PR #5369, opened 2023-08-18
Update contributing guidelines

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -149,7 +149,8 @@ see the following sections in the [SpECTRE
 documentation](https://spectre-code.org/):
 
 * [Installing SpECTRE](installation.html)
-* [Running Status Checks Locally](github_actions_guide.html#perform-checks-locally)
+* [Running Status Checks
+  Locally](github_actions_guide.html#perform-checks-locally)
 
 ## Pull Requests {#pull-requests}
 
@@ -208,6 +209,13 @@ with the number of lines changed.  Therefore if your pull request is
 too large, we may ask you to break it up into several smaller pull
 requests.
 
+If you would like feedback on a pull request prior to it being ready
+for formal review, please open it
+[in draft mode](https://github.blog/2019-02-14-introducing-draft-pull-requests/)
+and [request reviews][request-reviews] from whomever you wish to get feedback
+from. As long as the PR is in draft mode it will not be reviewed, aside from the
+feedback requested.
+
 > Below, days mean business days, so if the time period includes the
 > weekend, add two days, and for major holidays add a day.
 > Furthermore, most of us are academics, and we occasionally go to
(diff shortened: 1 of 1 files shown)


nilsdeppe (dismissed):

nilsdeppe:
I still think we should move towards reducing review time further, but I'm okay with trying this. I have generally found that more reviews and longer review times actually doesn't end up significantly reducing the number of bugs we have. Sure, we can all point out a few places where maybe if someone else had looked at it we could've caught it, but that's sort of a red herring. In the majority of PRs, the bug slides through with pretty much any number of reviewers because they are subtle and only caught in weird edge cases (e.g. `std::deque` in the mutable cache).

docs/CONTRIBUTING.md
```diff
@@ -208,6 +209,12 @@ with the number of lines changed.  Therefore if your pull request is
 too large, we may ask you to break it up into several smaller pull
 requests.
 
+If you would like feedback on a pull request prior to it being ready
+for formal review, please open it
+[in draft mode](https://github.blog/2019-02-14-introducing-draft-pull-requests/)
+and [request reviews](request-reviews) from whomever you wish to get feedback
```
knelli2: The `request reviews` link doesn't show up properly. I think you need a preceding `#`.

docs/CONTRIBUTING.md
```diff
@@ -216,24 +223,18 @@ requests.
...
-  developer](#core-developers) will assign reviewers, removing the
-  `reviewers wanted` label.  If for some reason, no reviewers have
-  been assigned after three days, the author of the pull request
-  should feel free to [ping the core
-  developers](https://github.blog/2011-03-23-mention-somebody-they-re-notified/)
-  (e.g `@sxs-collaboration/spectre-core-devs please assign reviewers`)
+
+- Within two days, one of the [SpECTRE core developers](#core-developers) will
+  either review the PR or assign reviewers. If this has not happened after two
+  days, please [request a review](request-reviews)
```
knelli2: Same here.

docs/CONTRIBUTING.md
```diff
@@ -216,24 +223,18 @@ requests.
...
+  and select the `@sxs-collaboration/spectre-core-devs` team. Also feel free to
+  [ping the core developers](github-ping) (e.g
```
knelli2: Same here.

docs/CONTRIBUTING.md
```diff
@@ -249,29 +250,29 @@ following manner:
   the reviewer and ask for clarification.  Authors and reviewers are
   encouraged to talk to one another (in person, via Google hangout, or
   some other verbal method if possible) to resolve any issues.
-- Reviewers are encouraged to [ping others
-  \@GITHUB_USERNAME](https://github.blog/2011-03-23-mention-somebody-they-re-notified/)
-  to get opinions on code they are unsure about.
+- Reviewers are encouraged to [ping others \@GITHUB_USERNAME](github-ping) to
```
knelli2: Same here.

docs/CONTRIBUTING.md
```diff
@@ -249,29 +250,29 @@ following manner:
...
-- Once a pull request is updated, the reviewers should either request
-  further changes (removing the `updated` label) or tell the pull
-  request author to squash their commits.
-- Once all reviewers give the okay to squash, the author should
-  [rebase on develop and then squash their
+  should begin with `fixup`) and push them to the PR branch.
+  Fixup commits make reviews significantly faster because the reviewers don't
+  have to review the full PR again, but only the parts that changed.
+- After pushing fixup commits, the author should
+  [re-request reviews from the reviewers](request-reviews).
```
knelli2: Same here.

knelli2 (changes requested):

docs/CONTRIBUTING.md
nilsvu: The syntax is with square brackets. Fixed.

knelli2 (dismissed):

docs/CONTRIBUTING.md
```diff
@@ -208,6 +209,12 @@ with the number of lines changed.  Therefore if your pull request is
 too large, we may ask you to break it up into several smaller pull
 requests.
 
+If you would like feedback on a pull request prior to it being ready
+for formal review, please open it
+[in draft mode](https://github.blog/2019-02-14-introducing-draft-pull-requests/)
+and [request reviews][request-reviews] from whomever you wish to get feedback
+from. As long as the PR is in draft mode, no one should review the pull request.
```
geoffrey4444: "As long as the PR is in draft mode, it will not be reviewed, aside from the feedback requested."

Otherwise, this sounds like you're saying no one should look at it.

geoffrey4444 (commented):
I agree with @nilsdeppe's comment about hoping to work toward faster review times. Reviews should be long enough to catch the kinds of things they are good at catching but nothing more than that. I have one optional change to suggest, and then I'm happy with this PR and don't need to see it again.

nilsvu:
I also agree with @nilsdeppe that faster reviews would accelerate our overall progress. I would support a change in policy that removes any "minimum wait time" on PRs. I think we should instead be more proactive in requesting more reviews if needed.

kidder (approved):

knelli2:
@wthrowe Could you take a look at this?

wthrowe (approved):

PR #5372, opened 2023-08-21
Add h-refinement criterion based on power in highest modes

nilsvu:
## Proposed changes

Similar to the Persson TCI.

Figure 8

Bottom panel are the modes in each element. Top right axis and horizontal lines in top panel show the smoothness indicator.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Amr/Criteria/Persson.cpp
+++ b/src/ParallelAlgorithms/Amr/Criteria/Persson.cpp
@@ -1,0 +1,115 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ParallelAlgorithms/Amr/Criteria/Persson.hpp"
+
+#include <array>
+#include <cstddef>
+
+#include "DataStructures/ApplyMatrices.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Matrix.hpp"
+#include "Domain/Amr/Flag.hpp"
+#include "NumericalAlgorithms/Spectral/Filtering.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/SetNumberOfGridPoints.hpp"
+
+namespace amr::Criteria {
+
+template <size_t Dim>
+double persson_smoothness_indicator(
+    const gsl::not_null<DataVector*> filtered_component_buffer,
+    const DataVector& tensor_component, const Mesh<Dim>& mesh,
+    const size_t dimension, const size_t num_highest_modes) {
+  // Zero out the lowest modes in the given dimension
+  static const Matrix identity{};
+  auto matrices = make_array<Dim>(std::cref(identity));
+  gsl::at(matrices, dimension) = Spectral::filtering::zero_lowest_modes(
+      mesh.slice_through(dimension),
+      mesh.extents(dimension) - num_highest_modes);
+  set_number_of_grid_points(filtered_component_buffer,
+                            mesh.number_of_grid_points());
+  apply_matrices(filtered_component_buffer, matrices, tensor_component,
+                 mesh.extents());
+  // Take the L2 norm over all grid points
+  return blaze::l2Norm(*filtered_component_buffer) /
+         sqrt(filtered_component_buffer->size());
+}
+
+template <size_t Dim>
+std::array<double, Dim> persson_smoothness_indicator(
+    const DataVector& tensor_component, const Mesh<Dim>& mesh,
+    const size_t num_highest_modes) {
+  std::array<double, Dim> result{};
+  DataVector buffer{};
+  for (size_t d = 0; d < Dim; ++d) {
+    gsl::at(result, d) = persson_smoothness_indicator(
+        make_not_null(&buffer), tensor_component, mesh, d, num_highest_modes);
+  }
+  return result;
+}
+
+namespace Persson_detail {
+
+template <size_t Dim>
+void max_over_components(const gsl::not_null<std::array<Flag, Dim>*> result,
+                         const gsl::not_null<DataVector*> buffer,
+                         const DataVector& tensor_component,
+                         const Mesh<Dim>& mesh, const size_t num_highest_modes,
+                         const double alpha, const double absolute_tolerance,
+                         const double coarsening_factor) {
+  const double umax = max(abs(tensor_component));
+  for (size_t d = 0; d < Dim; ++d) {
+    // Skip this dimension if we have already decided to refine it
+    if (gsl::at(*result, d) == Flag::Split) {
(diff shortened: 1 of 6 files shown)


nilsvu:
@nilsdeppe @yoonso0-0 here's the Persson-based AMR criterion. Quite simple implementation. Pinging you because we talked about this, feel free to review if you want to.

src/ParallelAlgorithms/Amr/Criteria/Persson.cpp
```diff
@@ -0,0 +1,116 @@
...
+  auto matrices = make_array<Dim>(std::cref(identity));
+  gsl::at(matrices, dimension) = Spectral::filtering::zero_lowest_modes(
+      mesh.slice_through(dimension),
+      mesh.extents(dimension) - num_highest_modes);
+  set_number_of_grid_points(filtered_component_buffer,
+                            mesh.number_of_grid_points());
+  apply_matrices(filtered_component_buffer, matrices, tensor_component,
+                 mesh.extents());
+  // Take the L2 norm over all grid points
+  return sqrt(alg::accumulate(square(*filtered_component_buffer), 0.) /
```
nilsdeppe: I'm worried about expressions and template parameters here. Why not use Blaze L2 norm?

src/ParallelAlgorithms/Amr/Criteria/Persson.cpp
```diff
@@ -0,0 +1,116 @@
...
+void max_over_components(const gsl::not_null<std::array<Flag, Dim>*> result,
+                         const gsl::not_null<DataVector*> buffer,
+                         const DataVector& tensor_component,
+                         const Mesh<Dim>& mesh, const size_t num_highest_modes,
+                         const double relative_tolerance,
+                         const double absolute_tolerance,
+                         const double coarsening_factor) {
+  for (size_t d = 0; d < Dim; ++d) {
+    // Skip this dimension if we have already decided to refine it
+    if (gsl::at(*result, d) == Flag::IncreaseResolution) {
```
nilsdeppe: Is `Split` not higher than `IncreaseResolution`?

src/ParallelAlgorithms/Amr/Criteria/Persson.hpp
```diff
@@ -0,0 +1,237 @@
...
+/// @{
+/*!
+ * \brief Computes an anisotropic smoothness indicator based on the power in the
+ * highest modes
+ *
+ * This smoothness indicator is the L2 norm of the tensor component with the
+ * lowest N - M modes filtered out, where N is the number of grid points in the
+ * given dimension and M is `num_highest_modes`.
+ *
+ * This strategy is similar to the Persson troubled-cell indicator (see
```
nilsdeppe, line 43: Can you add some math here? This is similar to Persson in the sense of it uses spectral coefficients, but you have completely different parameters from Persson, so you really need to document this. I don't understand what value I should choose when compare to the alpha value of the Persson TCI. That needs to be made extremely clear because that value is central to the Persson TCI.

nilsdeppe (changes requested):

src/ParallelAlgorithms/Amr/Criteria/Persson.cpp
nilsvu: Blaze's L2 norm doesn't divide by the number of grid points, and I found the docs confusing. The docs say "computes the _squared_ L1/2/3/4 norm" (https://bitbucket.org/blaze-lib/blaze/wiki/Vector%20Operations#!norms). I found it clearer to do this explicit sum-divide-sqrt.

src/ParallelAlgorithms/Amr/Criteria/Persson.cpp
nilsvu: Ohh good catch, this should be `Split`.

src/ParallelAlgorithms/Amr/Criteria/Persson.cpp
nilsdeppe: Okay, but the performance here is unclear. Just use the Blaze one.

nilsvu: You mean `sqrt(square(blaze::l2Norm(*buffer)) / num_points)`?

nilsdeppe: Why not `blaze::l2Norm(*buffer) / sqrt(static_cast<double>(num_points));`?

nilsdeppe: or maybe: `(1.0 / sqrt(static_cast<double>(num_points))) * blaze::l2Norm(*buffer)`. :shrug:

nilsvu: Ok

src/ParallelAlgorithms/Amr/Criteria/Persson.hpp
nilsvu: relative tolerance = N^{-\alpha}

I'll just switch to `Alpha` as a parameter.

nilsvu:
Pushed a fixup

nilsdeppe (dismissed):
LGTM, please squash!

yoonso0-0:
I believe @nilsdeppe resolved technical parts, so just a small suggestion. What I see from here is that technically this is slightly different from the exact implementation of the Persson TCI, in that this AMR criteria uses `max(U)`, highest modes norm divided by `sqrt(N)` etc. So I think it'd be better to rename this criteria to something other than the term `Persson` to avoid potential confusion.

yoonso0-0:
From the SIAM conference I saw many people using the term "spectral shock sensor" or "modal sensor", I would pick something like `SpectralSensor` or `ModalSensor`. Welcome any opinions from Nils and Larry :)

src/ParallelAlgorithms/Amr/Criteria/TruncationError.cpp
```diff
@@ -29,6 +29,7 @@ void max_over_components(
   // still satisfy the target with the highest mode removed will the element
   // decrease p refinement in that dimension.
   PowerMonitors::power_monitors(power_monitors_buffer, tensor_component, mesh);
+  const double umax = max(abs(tensor_component));
```
kidder, line 32: this change should be in a separate commit

kidder (changes requested):

nilsvu:
@yoonso0-0 I think this is similar enough to the Persson TCI that sharing the name helps transfer some of the experience from the TCI to this AMR criterion. No strong opinion though.

nilsdeppe:
I personally think it's close enough to the Persson TCI (and might even be better in general) that the name is fine :shrug: we can rename later if need be :)

nilsdeppe (approved):

kidder (approved):

PR #5373, opened 2023-08-22
Add Docker container for Apple Silicon

geoffrey4444:
## Proposed changes

Add an arm64 docker container and corresponding development environment for use on Apple Silicon Macs. x86 docker files run extremely slowly on Apple Silicon Macs, because they use emulation; this PR adds a container that is Apple Silicon native.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/containers/Dockerfile.buildenv
+++ b/containers/Dockerfile.buildenv
@@ -3,17 +3,20 @@
 
 # If you change this file please push a new image to DockerHub so that the
 # new image is used. Docker must be run as root on your machine. There are 3
-# different images in this Dockerfile. Here is how to build them:
+# different images in this Dockerfile. Here is how to build them for x86_64
+# (for Apple Silicon, replace linux/amd64 with linux/arm64):
 # 1. `dev`
 #
 #   cd $SPECTRE_HOME
 #   docker build --target dev -t sxscollaboration/spectre:dev \
+#                 --platform linux/amd64 \
 #                 -f ./containers/Dockerfile.buildenv .
 #
 # 2. `ci` (this should use the `dev` image you just built for most of it)
 #
 #   cd $SPECTRE_HOME
 #   docker build --target ci -t sxscollaboration/spectre:ci \
+#                 --platform linux/amd64 \
 #                 -f ./containers/Dockerfile.buildenv .
 #
 # 3. `demo` To build this, it is recommended to first push the `dev` and `ci`
@@ -29,6 +32,7 @@
 #   cd $SPECTRE_HOME
 #   rm -rf build/
 #   docker build --target demo -t sxscollaboration/spectre:demo \
+#                 --platform linux/amd64 \
 #                 -f ./containers/Dockerfile.buildenv .
 #
 #    and then to push the `demo` image to DockerHub:
@@ -38,8 +42,24 @@
 # If you do not have permission to push to DockerHub please coordinate with
 # someone who does. Since changes to this image effect our testing
 # infrastructure it is important all changes be carefully reviewed.
+#
+FROM ubuntu:20.04 AS base
 
-FROM ubuntu:20.04 AS dev
+# See
+# https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope
+# for how TARGETARCH is defined.
+ARG TARGETARCH
+
+FROM base AS base-amd64
+ENV CHARM_ARCH=x86_64
+ENV TEX_ARCH=x86_64
+
+FROM base AS base-arm64
+ENV CHARM_ARCH=arm8
+ENV TEX_ARCH=aarch64
+
+FROM base-${TARGETARCH} AS dev
+ARG TARGETARCH
 
 ARG PARALLEL_MAKE_ARG=-j2
 ARG DEBIAN_FRONTEND=noninteractive
@@ -175,7 +195,8 @@ RUN wget https://github.com/Libsharp/libsharp/archive/v1.0.0.tar.gz -O libsharp.
     && tar -xzf libsharp.tar.gz \
     && mv libsharp-* libsharp_build \
     && cd libsharp_build \
-    && sed -i 's/march=native/march=x86-64/' configure.ac \
+    && { if [ "$TARGETARCH" != "arm64" ]; \
+        then sed -i 's/march=native/march=x86-64/' configure.ac; fi } \
     && autoconf \
     && ./configure --prefix=/usr/local --enable-pic --disable-openmp \
     && make $PARALLEL_MAKE_ARG \
@@ -185,14 +206,25 @@ RUN wget https://github.com/Libsharp/libsharp/archive/v1.0.0.tar.gz -O libsharp.
     && cd ../ \
     && rm -r libsharp*
 # - LibXSMM
-RUN wget https://github.com/hfp/libxsmm/archive/1.16.1.tar.gz -O libxsmm.tar.gz \
-    && tar -xzf libxsmm.tar.gz \
-    && mv libxsmm-* libxsmm \
-    && cd libxsmm \
-    && make $PARALLEL_MAKE_ARG PREFIX=/usr/local/ CXX=g++-9 CC=gcc-9 \
-        FC=gfortran-9 install \
-    && cd .. \
-    && rm -rf libxsmm libxsmm.tar.gz
+RUN if [ "$TARGETARCH" = "arm64" ] ; then \
+        git clone --single-branch --branch main --depth 1 \
+            https://github.com/libxsmm/libxsmm.git libxsmm \
+        && cd libxsmm \
+        && make $PARALLEL_MAKE_ARG PREFIX=/usr/local/ CXX=g++-9 CC=gcc-9 \
+            FC=gfortran-9 PLATFORM=1 install \
+        && cd .. \
+        && rm -rf libxsmm; \
+    else \
+        wget https://github.com/hfp/libxsmm/archive/1.16.1.tar.gz \
+            -O libxsmm.tar.gz \
+        && tar -xzf libxsmm.tar.gz \
+        && mv libxsmm-* libxsmm \
+        && cd libxsmm \
+        && make $PARALLEL_MAKE_ARG PREFIX=/usr/local/ CXX=g++-9 CC=gcc-9 \
+            FC=gfortran-9 install \
+        && cd .. \
+        && rm -rf libxsmm libxsmm.tar.gz; \
+    fi
 # - Yaml-cpp
 RUN wget https://github.com/jbeder/yaml-cpp/archive/yaml-cpp-0.6.3.tar.gz -O yaml-cpp.tar.gz \
     && tar -xzf yaml-cpp.tar.gz \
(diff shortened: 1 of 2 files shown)


geoffrey4444:
Two quick notes:

1. I don't (I think?) have permission to push to sxscollaboration on dockerhub. Would you prefer to give me permission or to push yourself after you're happy with this PR?
2. I will fix the whitespace CI failure when I respond to whatever change requests you have for this PR

nilsvu:
I can just pull your image, re-tag it, and push it to Dockerhub (or build it myself)

containers/Dockerfile_arm64.buildenv
```diff
@@ -0,0 +1,388 @@
...
+#
+#    and then to push the `demo` image to DockerHub:
+#
+#   docker push sxscollaboration/spectre_arm64:demo
+#
+# If you do not have permission to push to DockerHub please coordinate with
+# someone who does. Since changes to this image effect our testing
+# infrastructure it is important all changes be carefully reviewed.
+
+FROM --platform=linux/arm64 ubuntu:20.04 AS dev
```
nilsvu: Could you please try building this into the existing Dockerfile? It's pretty much the same, except for a few ARM modifications. Having to keep the two Dockerfiles in sync, and having to build and push separate images whenever the Dockerfile changes, would be quite the maintenance burden. There are some docs on multi-platform builds here: https://docs.docker.com/build/building/multi-platform/
The available variables are documented here: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

Note that the `--platform=$BUILDPLATFORM` in the docs linked above means that they compile on the host platform (whichever platform your machine is), but cross-compile for the `$TARGETPLATFORM`. That's a lot faster because no emulation is needed, but I have no experience when it's possible or not.

nilsvu (changes requested):

containers/Dockerfile_arm64.buildenv
nilsvu: Here's a relevant guide: https://www.docker.com/blog/faster-multi-platform-builds-dockerfile-cross-compilation-guide/

geoffrey4444:
> Could you please try building this into the existing Dockerfile? It's pretty much the same, except for a few ARM modifications. Having to keep the two Dockerfiles in sync, and having to build and push separate images whenever the Dockerfile changes, would be quite the maintenance burden. There are some docs on multi-platform builds here: https://docs.docker.com/build/building/multi-platform/
The available variables are documented here: https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope

I thought we discussed at a previous call that we would get this merged now, so there is a solution available for users running apple silicon, and then a future PR would take care of merging them. I'm happy to learn how to do this, but it will take me some time to figure it out, and I'd like to have a solution for users until I figure this out.

nilsdeppe:
I agree with Geoffrey, given the rapid increase in ARM Apple machines we should have a Docker solution, even if it's not the perfect one yet.

nilsvu:
It's a big maintenance burden for us to maintain an entirely separate second container. I think it would be quite quick to just put some `if [ "$TARGETARCH" = "arm64" ] &&` in front of `RUN` statements, along with some comments. That would be enough to not duplicate the entire Dockerfile. I'm not asking you to figure out how to cross-compile on the build platform or anything like that, just please don't duplicate everything so it becomes much harder to maintain.

Thanks for figuring out how to build the container for Apple Silicon! It's going to be very useful!

support/DevEnvironments/.devcontainer_arm64/devcontainer.json
```diff
@@ -0,0 +1,36 @@
...
+        "editor.formatOnSave": false,
+        "editor.formatOnPaste": true,
+        "editor.formatOnType": true,
+        "python.pythonPath": "/usr/bin/python3",
+        "python.formatting.provider": "black"
+    },
+    "remoteEnv": {
+        "SPECTRE_HOME": "${localWorkspaceFolder}"
+    },
+    "postCreateCommand": "git config --global --add safe.directory ${localWorkspaceFolder} && mkdir -p /root/.local/share/CMakeTools && cp ${localWorkspaceFolder}/support/DevEnvironments/.devcontainer/cmake-kits.json /root/.local/share/CMakeTools/cmake-tools-kits.json"
```
knelli2: Wrong dev container?

knelli2 (changes requested):
So @nilsvu I tried to do exactly this (have a combined x86/arm dockerfile) last summer and from my experience I don't think it's as trivial as you're making it out to be. There are some added/unforeseen complexities that could keep this PR open for weeks. I think we can assume the maintenance burden for a few weeks while @geoffrey4444 figures out how to do this.

Also one small change that can be squashed.

geoffrey4444:
@nilsvu @knelli2 and @geoffrey4444 discussed, and decided on the following:
* @geoffrey4444 will spend a week i) merging the two dockerfiles into one with if statements, and ii) exploring whether it's possible to have a single file uploaded to dockerhub, such that users on arm64 vs x86 will automatically pull down the correct image

support/DevEnvironments/.devcontainer_arm64/devcontainer.json
geoffrey4444: I'll check when I merge them

nilsvu:
I can deal with pushing the container to a single multi-platform image on Dockerhub (point ii). Just try to put the ARM changes into the existing Dockerfile with some `RUN [ "$TARGETARCH" = "arm64" ] && ...` conditionals or something like that (point i).

geoffrey4444:
@nilsvu I figured out how to use a single dockerfile that works on Apple Silicon (I checked on my machine) and (I think) on x86_64 (should be the default, but I don't have docker on an x86_64 machine handy to test).

Because you can't do conditional logic on ENV commands in dockerfiles, when building for Apple Silicon, there are some extra `--build-arg` options you must provide (see comments at the top of the docker file).

When you take a look at the fixup, please also double-check whether the dev environment for arm64 looks ok? I'm unsure e.g. if in the end the docker file to be pulled will have a different name, or if there are some other changes we'll need to make, since I'm not pushing the images to dockerhub

containers/Dockerfile.buildenv
```diff
@@ -38,8 +38,29 @@
 # If you do not have permission to push to DockerHub please coordinate with
 # someone who does. Since changes to this image effect our testing
 # infrastructure it is important all changes be carefully reviewed.
+#
+# NOTE: If building for Apple Silicon, then add the following arguments to
+# the docker build command in step 1 and in step 3 (building the ci target)
+# is not currently supported, since GitHub actions do not yet support
+# Apple Silicon:
```
knelli2: Even though CI can't support Apply Silicon, should we still modify the CI image to work with arm so it's still usable? Some devs may want to run with different compilers on arm and this would be how to do it.

support/DevEnvironments/.devcontainer_arm64/devcontainer.json
knelli2: I think this still needs to change because the cmake-kit has the arm version of CHARM_ROOT

knelli2 (changes requested):
You can squash these in.

containers/Dockerfile.buildenv
nilsvu: Would be nice but don't need to do it in this PR I think. I don't see a use case right now.

containers/Dockerfile.buildenv
```diff
@@ -38,8 +38,29 @@
...
+#
+#  --build-arg "CHARM_ARCH=arm8" --build-arg "TEX_ARCH=aarch64"
+#
 
-FROM ubuntu:20.04 AS dev
+FROM --platform=linux/$BUILDARCH ubuntu:20.04 AS dev
```
nilsvu: Remove the `--platform=linux/$BUILDARCH` here. Otherwise we can't build for x86 on ARM machines or vice-versa.

Details: The `--platform=linux/$BUILDARCH` would always create a container with the same architecture as the host that's building it. That's useful for cross-compiling on the host, but requires copying the result to a new stage with the correct target architecture. If you just delete the `--platform` then the correct base image for the target architecture will be used automatically, and emulated if necessary.

containers/Dockerfile.buildenv
```diff
@@ -38,8 +38,29 @@
...
+# Platform-dependent arguments. For building on Apple Silicon, instead of the
+# default options given here, CHARM_ARCH must be set to arm8, and
+# TEX_ARCH must be set to aarch64. Dockerfiles do not support conditional
+# logic for ARG and ENV commands, so these must be set on the command line.
+#
+# Note that on Apple Silicon, BUILDARCH is automatically set to 'arm64'. Also
+# note that building the ci target is currently only supported for x86_64, since
+# GitHub Actions does not support Apple Silicon.
+#
+ARG BUILDARCH
```
nilsvu: This should be `TARGETARCH` everywhere in this file. `BUILDARCH` is the architecture that is performing the build, see https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope. I suggest you link to these docs in a comment here.

containers/Dockerfile.buildenv
knelli2: Ok sounds good.

knelli2:
I built the x86 `dev` and `demo` images without issue so that works well

support/DevEnvironments/.devcontainer_arm64/cmake-kits.json
```diff
@@ -0,0 +1,20 @@
+[
+    {
+        "name": "Default",
+        "compilers": {
+            "C": "/usr/bin/clang-10",
+            "CXX": "/usr/bin/clang++-10",
+            "Fortran": "/usr/bin/gfortran-9"
+        },
+        "cmakeSettings": {
+            "CHARM_ROOT": "/work/charm_7_0_0/multicore-linux-arm8-gcc",
```
nilsvu: Is this line the only reason why the `.devcontainer` has to be duplicated? You can just delete this line because `CHARM_ROOT` is set as an environment variable. Then delete the separate `.devcontainer`.

nilsvu (changes requested):

containers/Dockerfile.buildenv
nilsvu: Avoid the extra `CHARM_ARCH` and `TEX_ARCH` build args like this:

```suggestion
FROM ubuntu:20.04 AS base

ARG TARGETARCH

FROM base AS base-amd64
ENV CHARM_ARCH=x86_64
ENV TEX_ARCH=x86_64

FROM base AS base-arm64
ENV CHARM_ARCH=arm8
ENV TEX_ARCH=aarch64

FROM base-${TARGETARCH} AS dev
```

This should enable building both containers with #5380.

nilsvu (changes requested):

support/DevEnvironments/.devcontainer_arm64/cmake-kits.json
geoffrey4444: Yes, this is the only reason, now that I know that the docker file itself will have the same name on both platforms. I'll delete the other container and delete this line

docs/DevGuide/QuickStartDockerVSCode.md
```diff
@@ -65,6 +65,10 @@ Now symlink the development environment:
 ```
 ln -s support/DevEnvironments/.devcontainer .devcontainer
 ```
+\note If you are using a Mac with Apple Silicon, instead do the following:
+```
+ln -s support/DevEnvironments/.devcontainer_arm64 .devcontainer
+```
```
nilsvu: delete

nilsvu (changes requested):
Looks good 👍 squash

knelli2 (commented):
squash

knelli2 (approved):

nilsvu (approved):

nilsvu:
Building the combined x86+ARM container here: https://github.com/nilsvu/spectre/actions/runs/6113113513/job/16591953188

PR #5376, opened 2023-08-23
Make preparations for control system expr time changes

knelli2:
## Proposed changes

Minor things needed in preparation for more major control system changes. The most important commit in this PR is "Add combined name for control systems" as it is one of the foundations for the changes to come.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5377, opened 2023-08-23
Add UpdateAggregator

knelli2:
## Proposed changes

Adds a class and action for aggregating/collecting all function of time updates before actually calling `Parallel::mutate`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #5376.~

### DIFF
--- a/src/ControlSystem/Tags/SystemTags.hpp
+++ b/src/ControlSystem/Tags/SystemTags.hpp
@@ -8,13 +8,17 @@
 #include <memory>
 #include <optional>
 #include <string>
+#include <unordered_map>
 #include <utility>
 
 #include "ControlSystem/Averager.hpp"
+#include "ControlSystem/CombinedName.hpp"
 #include "ControlSystem/Controller.hpp"
+#include "ControlSystem/Metafunctions.hpp"
 #include "ControlSystem/Protocols/ControlSystem.hpp"
 #include "ControlSystem/Tags/OptionTags.hpp"
 #include "ControlSystem/TimescaleTuner.hpp"
+#include "ControlSystem/UpdateFunctionOfTime.hpp"
 #include "DataStructures/DataBox/Tag.hpp"
 #include "Domain/Creators/DomainCreator.hpp"
 #include "ParallelAlgorithms/ApparentHorizonFinder/Tags.hpp"
@@ -24,9 +28,12 @@
 #include "Utilities/ProtocolHelpers.hpp"
 #include "Utilities/StdHelpers.hpp"
 #include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
 #include "Utilities/TypeTraits/CreateHasStaticMemberVariable.hpp"
 
 /// \cond
+template <class Metavariables, typename ControlSystem>
+struct ControlComponent;
 namespace control_system {
 template <typename ControlSystem>
 struct OptionHolder;
@@ -204,5 +211,48 @@ struct Verbosity : db::SimpleTag {
   static type create_from_options(const ::Verbosity verbosity) {
     return verbosity;
   }
+};
+
+/*!
+ * \brief Tag meant to be stored in the GlobalCache that stores a map between
+ * names of control systems and the "combined" name that that control system is
+ * part of.
+ *
+ * \details The "combined" name for each control system is computed using
+ * `control_system::system_to_combined_names` where the list of control systems
+ * is taken from the `component_list` type alias of the metavariables. Each
+ * "combined" name corresponds to a different
+ * `control_system::protocols::Measurement`.
+ */
+struct SystemToCombinedNames : db::SimpleTag {
+  using type = std::unordered_map<std::string, std::string>;
+
+  template <typename Metavariables>
+  using option_tags = tmpl::list<>;
+  static constexpr bool pass_metavariables = true;
+
+ private:
+  template <typename Component>
+  using system = typename Component::control_system;
+
+ public:
+  template <typename Metavariables>
+  static type create_from_options() {
+    using all_control_components =
+        metafunctions::all_control_components<Metavariables>;
+    using all_control_systems =
+        tmpl::transform<all_control_components, tmpl::bind<system, tmpl::_1>>;
+
+    return system_to_combined_names<all_control_systems>();
+  }
+};
+
+/*!
+ * \brief Map between "combined" names and the
+ * `control_system::UpdateAggregator`s that go with each.
+ */
+struct UpdateAggregators : db::SimpleTag {
+  using type =
+      std::unordered_map<std::string, control_system::UpdateAggregator>;
 };
 }  // namespace control_system::Tags
--- a/src/ControlSystem/UpdateFunctionOfTime.cpp
+++ b/src/ControlSystem/UpdateFunctionOfTime.cpp
@@ -3,13 +3,19 @@
 
 #include "ControlSystem/UpdateFunctionOfTime.hpp"
 
+#include <limits>
 #include <memory>
+#include <pup.h>
 #include <string>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
+#include <vector>
 
 #include "DataStructures/DataVector.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Utilities/Algorithm.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
 
 namespace control_system {
 void UpdateSingleFunctionOfTime::apply(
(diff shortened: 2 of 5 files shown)


src/ControlSystem/UpdateFunctionOfTime.cpp
```diff
@@ -44,4 +50,100 @@ void ResetFunctionOfTimeExpirationTime::apply(
     const std::string& f_of_t_name, const double new_expiration_time) {
   (*f_of_t_list).at(f_of_t_name)->reset_expiration_time(new_expiration_time);
 }
+
+UpdateAggregator::UpdateAggregator(
+    std::unordered_set<std::string> control_system_names)
+    : individual_names_(std::move(control_system_names)),
+      combined_name_([this]() {
```
knelli2: This makes an assumption that all control systems are active, which may not be the case. Rather than taking all control system names and then creating the combined name, pass the combined name and only the active control systems:

```cpp
UpdateAggregator::UpdateAggregator(
     std::string combined_name,
     std::unordered_set<std::string> active_control_system_names)
```

This will require the control system Initialization action (where this class gets constructed, in #5378) to know about all active/inactive control systems. Currently, that information is only available at option parsing. I will have to make a global cache tag to expose that information.

src/ControlSystem/Tags/SystemTags.hpp
```diff
@@ -25,12 +27,16 @@
...
+#include "Utilities/TaggedTuple.hpp"
 #include "Utilities/TypeTraits/CreateHasStaticMemberVariable.hpp"
 
 /// \cond
+template <class Metavariables, typename ControlSystem>
+struct ControlComponent;
 namespace control_system {
 template <typename ControlSystem>
 struct OptionHolder;
+struct UpdateAggregator;
```
wthrowe: Need a complete type for use in a `std::unordered_map`.

src/ControlSystem/Tags/SystemTags.hpp
```diff
@@ -206,4 +212,47 @@ struct Verbosity : db::SimpleTag {
...
+ * part of.
+ *
+ * \details The "combined" name for each control system is computed using
+ * `control_system::system_to_combined_names` where the list of control systems
+ * is taken from the `component_list` type alias of the metavariables. Each
+ * "combined" name corresponds to a different
+ * `control_system::protocols::Measurement`.
+ */
+struct SystemToCombinedNames : db::SimpleTag {
+  using type = std::unordered_map<std::string, std::string>;
```
wthrowe, line 228: `#include <unordered_map>`

src/ControlSystem/UpdateFunctionOfTime.hpp
```diff
@@ -58,4 +76,174 @@ struct ResetFunctionOfTimeExpirationTime {
...
+  const std::string& combined_name() const;
+
+  /*!
+   * \brief Once `is_ready` is true, returns a map between the control system
+   * name and a `std::pair` containing the `control_signal` that was passed to
+   * `insert` and the minimum of all the `new_fot_expiration_time`s passed to
+   * `insert`, respectively.
+   *
+   * \details This function is expected to only be called when `is_ready` is
+   * true, and only be called once. It also must be called before
```
wthrowe: Remove "and only be called once".  (The method doesn't have any side-effects.)

src/ControlSystem/UpdateFunctionOfTime.hpp
```diff
@@ -58,4 +76,174 @@ struct ResetFunctionOfTimeExpirationTime {
...
+   * `combined_measurement_expiration_time`. This function leaves all stored
+   * data in tact.
```
wthrowe: "intact", although that's generally expected behavior for a const member function so probably not really necessary to be there at all.

src/ControlSystem/UpdateFunctionOfTime.hpp
```diff
@@ -58,4 +76,174 @@ struct ResetFunctionOfTimeExpirationTime {
...
+  combined_fot_expiration_times() const;
+
+  /*!
+   * \brief Once `is_ready` is true, returns a `std::pair` containing the
+   * minimum of all `new_measurement_timescale`s passed to `insert` and the
+   * minimum of all `new_measurement_expiration_time`s passed to `insert`,
+   * respectively.
+   *
+   * \details This function is expected to only be called when `is_ready` is
+   * true, and only be called once. It also must be called after
```
wthrowe: "once per something" or something like that.  (I assume you don't create an entire new UpdateAggregator for every update.)

tests/Unit/ControlSystem/Test_UpdateFunctionOfTime.cpp
```diff
@@ -163,4 +184,261 @@ SPECTRE_TEST_CASE("Unit.ControlSystem.UpdateFunctionOfTime",
     CHECK(cache_quatfot == expected_quatfot);
   }
 }
+
+void test_update_aggregator() {
+  std::unordered_set<std::string> names{"Illinois", "California", "Wisconsin",
+                                        "Florida", "Texas"};
+  std::vector<std::string> sorted_names(names.begin(), names.end());
+  alg::sort(sorted_names);
```
wthrowe: You never use this in a way that cares if it is sorted.  Just use `names`.

tests/Unit/ControlSystem/Test_UpdateFunctionOfTime.cpp
```diff
@@ -163,4 +184,261 @@ SPECTRE_TEST_CASE("Unit.ControlSystem.UpdateFunctionOfTime",
...
+  CHECK_FALSE(aggregator.is_ready());
+
+  CHECK(combined_fot.size() == sorted_names.size());
+  for (size_t i = 0; i < sorted_names.size(); i++) {
+    CHECK(combined_fot.count(sorted_names[i]) == 1);
+    CHECK(combined_fot.at(sorted_names[i]) ==
+          std::make_pair(DataVector{double(i + 1)}, 1.0));
+  }
+
+  CHECK(combined_measurements == std::make_pair(1.0, 1.0));
```
wthrowe: Choose values so the two outputs aren't the same.

tests/Unit/ControlSystem/Test_UpdateFunctionOfTime.cpp
```diff
@@ -163,4 +184,261 @@ SPECTRE_TEST_CASE("Unit.ControlSystem.UpdateFunctionOfTime",
...
+  CHECK_FALSE(box_aggregators_component1.empty());
+  CHECK(box_aggregators_component2.empty());
+  CHECK(box_aggregators_component3.empty());
+
+  const control_system::UpdateAggregator& box_aggregator13 =
+      box_aggregators_component1.at("FoT1FoT3");
+  const control_system::UpdateAggregator& box_aggregator2 =
+      box_aggregators_component1.at("FoT2");
+
+  //   const double time = 1.0;
```
wthrowe: Remove?

tests/Unit/ControlSystem/Test_UpdateFunctionOfTime.cpp
```diff
@@ -163,4 +184,261 @@ SPECTRE_TEST_CASE("Unit.ControlSystem.UpdateFunctionOfTime",
...
+  // ready, they are reset immediately during the simple action
+  CHECK_FALSE(box_aggregator13.is_ready());
+  CHECK_FALSE(box_aggregator2.is_ready());
+  CHECK(box_aggregators_component2.empty());
+  CHECK(box_aggregators_component3.empty());
+
+  // Update one of the two functions of time for the 13 measurement
+  ActionTesting::simple_action<component1,
+                               control_system::AggregateUpdate<System1>>(
+      make_not_null(&runner), /*time,*/ 0_st, new_measurement_timescale1,
```
wthrowe: Remove?  (And several more later.)

wthrowe (changes requested):
Only a partial review since this needs to be updated for other PRs that have been merged.  I'll go through the whole thing after that.

knelli2:
@wthrowe Updated and no longer dependent. Also squashed in those changes you suggested.

wthrowe:
There are some CI failures, including an ASAN error.  I haven't tried to identify the problem.

knelli2:
@wthrowe Figured it out. In a test I got a reference to a map element, then cleared the map, but used the reference after.

src/ControlSystem/UpdateFunctionOfTime.hpp
```diff
@@ -58,4 +76,174 @@ struct ResetFunctionOfTimeExpirationTime {
...
+   * calculated during that update.
+   * \param control_signal New highest derivative for the function of time
+   * calculated during that update (will be `std::move`ed).
+   * \param new_fot_expiration_time New function of time expiration time
+   * calculated for during that update
+   */
+  void insert(const std::string& control_system_name,
+              const DataVector& new_measurement_timescale,
+              const double new_measurement_expiration_time,
+              DataVector&& control_signal,
```
wthrowe: Take by value.  Rvalue reference arguments seem to be out of favor these days.  I believe because it's annoying to call such functions with lvalues and moves are generally cheap.

src/ControlSystem/UpdateFunctionOfTime.hpp
```diff
@@ -58,4 +76,174 @@ struct ResetFunctionOfTimeExpirationTime {
...
+   * \brief Checks if `insert` has been called for all control systems that this
+   * class was constructed with.
+   */
+  bool is_ready() const;
+
+  /*!
+   * \brief Returns a sorted concatenation of the control system names this
+   * class was constructed with.
+   *
+   * The sorting is done with `std::sort`.
```
wthrowe: Is this important for some reason?  Seems like an implementation detail.

src/ControlSystem/UpdateFunctionOfTime.hpp
```diff
@@ -58,4 +76,174 @@ struct ResetFunctionOfTimeExpirationTime {
...
+   */
+  const std::string& combined_name() const;
+
+  /*!
+   * \brief Once `is_ready` is true, returns a map between the control system
+   * name and a `std::pair` containing the `control_signal` that was passed to
+   * `insert` and the minimum of all the `new_fot_expiration_time`s passed to
+   * `insert`, respectively.
```
wthrowe: Optional grammar nit: "respectively" relates entries in parallel lists, which you don't have here.  Same thing occurs again later.

tests/Unit/ControlSystem/Test_UpdateFunctionOfTime.cpp
```diff
@@ -163,4 +189,347 @@ SPECTRE_TEST_CASE("Unit.ControlSystem.UpdateFunctionOfTime",
...
+
+  control_system::UpdateAggregator unserialized_aggregator{combined_name,
+                                                           names};
+  control_system::UpdateAggregator aggregator =
+      serialize_and_deserialize(unserialized_aggregator);
+
+  CHECK(aggregator.combined_name() == combined_name);
+
+  aggregator.insert("Wisconsin", DataVector{1.0}, 10.0,
+                    std::move(signals.at("Wisconsin")), 5.0);
```
wthrowe: [optional] Once you change the rvalue reference argument these can be non-moved and then you don't need two copies of the signals map.  (This is why rvalue reference arguments are annoying.)

wthrowe (changes requested):
Squash directly.

knelli2:
@wthrowe Squashed and rebased

wthrowe (approved):

PR #5380, opened 2023-08-23
Add GitHub workflow to build Docker containers

nilsvu:
## Proposed changes

@geoffrey4444 once #5373 works we can add ARM to the list of platforms here to automatically build and push both architectures to DockerHub.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/BuildDockerContainer.yaml
+++ b/.github/workflows/BuildDockerContainer.yaml
@@ -62,3 +62,11 @@ jobs:
           target: ci
           tags: ${{ inputs.image_name }}:ci
           platforms: linux/amd64 #,linux/arm64
+
+  run_tests:
+    name: Test
+    uses: ./.github/workflows/Tests.yaml
+    needs: build_container
+    secrets: inherit
+    with:
+      container: ${{ inputs.image_name }}:ci
--- a/.github/workflows/BuildDockerContainer.yaml
+++ b/.github/workflows/BuildDockerContainer.yaml
@@ -1,0 +1,64 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+# This workflow can be run manually on any branch to build and push our Docker
+# containers.
+name: Build Docker container
+
+on:
+  workflow_dispatch:
+    inputs:
+      image_name:
+        description: >
+          Image name to push to DockerHub
+        required: true
+        default: 'sxscollaboration/spectre'
+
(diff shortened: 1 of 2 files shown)


.github/workflows/BuildDockerContainer.yaml
```diff
@@ -0,0 +1,61 @@
...
+      # - name: Login to GitHub container registry
+      #   uses: docker/login-action@v2
+      #   with:
+      #     registry: ghcr.io
+      #     username: ${{ github.actor }}
+      #     password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build dev container
+        uses: docker/build-push-action@v4
+        with:
+          push: true
```
knelli2, line 50: Is there a way to have this wait for another approval before pushing? Sort of like our releases? Just so we can inspect things if need be

knelli2 (commented):
We'll need to be careful when we run this and how exactly we set up automatic builds, but this'll be a good start towards that.

.github/workflows/BuildDockerContainer.yaml
nilsvu: Waiting for another approval between build and push would be more complicated because it needs to be split into separate jobs. See the `environment` option in line 21. We can configure protection rules for the environment, e.g. that it requires an approval, but it applies to the whole job.

I intend to use this on a branch to push to my own Dockerhub (or some other staging area) by setting the `image_name` to something like `nilsvu/spectre`. Then I can test it before pushing to `sxscollaboration/spectre` manually.

knelli2: Do you intend to use this workflow on develop then (in addition to your fork)?

nilsvu: Probably not? Usually we want to push the container to a staging area before merging a PR.

knelli2: If we don't plan on using this on develop really, is there a way to only add this to your fork?

nilsvu: No I think adding it to the repo is the right thing to do. It allows anyone of us to trigger the build from the PR branch. And it allows to run the tests over the new container (see new commit).

nilsdeppe (approved):

knelli2 (approved):

PR #5382, opened 2023-08-23
Add compute tags for computing and observing force-free constraint violations

yoonso0-0:
## Proposed changes

Add two compute tags (`E.B` and `E^2-B^2`) for observation purpose, and three compute tags (`TildeB^2`, `TildeE^2`, `TildeE.TildeB`) frequently being evaluated in evolution phase.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/ForceFree/EvolveForceFree.hpp
+++ b/src/Evolution/Executables/ForceFree/EvolveForceFree.hpp
@@ -40,6 +40,7 @@
 #include "Evolution/Systems/ForceFree/BoundaryCorrections/RegisterDerived.hpp"
 #include "Evolution/Systems/ForceFree/ElectricCurrentDensity.hpp"
 #include "Evolution/Systems/ForceFree/ElectromagneticVariables.hpp"
+#include "Evolution/Systems/ForceFree/ForceFreeConstraints.hpp"
 #include "Evolution/Systems/ForceFree/System.hpp"
 #include "Evolution/Systems/ForceFree/Tags.hpp"
 #include "IO/Observer/Actions/RegisterEvents.hpp"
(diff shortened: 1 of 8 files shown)


src/Evolution/Executables/ForceFree/EvolveForceFree.hpp
```diff
@@ -219,7 +222,10 @@ struct EvolutionMetavars {
           domain::Tags::Coordinates<3, Frame::ElementLogical>>,
 
       Initialization::Actions::AddComputeTags<
-          tmpl::list<ForceFree::Tags::ComputeTildeJ>>,
+          tmpl::list<ForceFree::Tags::TildeESquaredCompute,
+                     ForceFree::Tags::TildeBSquaredCompute,
+                     ForceFree::Tags::TildeEDotTildeBCompute,
```
nilsvu, line 227: Maybe just inline the dot product where you need it? Remember that compute tags need memory on every element. So we usually compute intermediate quantities like this in temporary storage and discard it when the time step is done.

src/Evolution/Systems/ForceFree/CMakeLists.txt
```diff
@@ -12,6 +12,7 @@ spectre_target_sources(
   ElectricCurrentDensity.cpp
   ElectromagneticVariables.cpp
   Fluxes.cpp
+  ForceFreeConstraints.cpp
```
nilsvu, line 15: Just `Constraints.*pp` (don't prefix every file with `ForceFree`)

nilsvu (changes requested):

src/Evolution/Executables/ForceFree/EvolveForceFree.hpp
yoonso0-0: @nilsvu If memory is not an issue for this evolution system (which is the case for the GH that have >100 variables to store), I'd like to have this as a tag since this term is computed fairly frequently during the time step. It's used for both flux and sources.

yoonso0-0: But I don't have a strong opinion on this. It could be that this is a bit too much for optimization.

knelli2:
@yoonso0-0  Since it's been a while, if you'd like to have these changes merged into spectre, please comment here or let the core-devs know. If you would no longer like to have these changes merged, or we don't hear from you by next Wed Aug 28 1pm PDT, we will close this PR. If you change your mind in the future, we can always re-open the PR.

PR #5388, opened 2023-08-24
Add Piecewise Function to Translation Map

AlexCarpenter46:
## Proposed changes

Adds the Translation Map functions needed for the inspiral portion of evolutions.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

I did not know how to change the test in the TestMapHelpers to be a different custom approx sometimes.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
@@ -14,15 +14,18 @@
 #include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
 #include "DataStructures/Tensor/Identity.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "NumericalAlgorithms/RootFinding/QuadraticEquation.hpp"
 #include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
 #include "PointwiseFunctions/MathFunctions/MathFunction.hpp"
 #include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/DereferenceWrapper.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 #include "Utilities/MakeArray.hpp"
 #include "Utilities/MakeWithValue.hpp"
+#include "Utilities/Serialization/PupStlCpp17.hpp"
 #include "Utilities/StdArrayHelpers.hpp"
 #include "Utilities/StdHelpers.hpp"
 
@@ -32,6 +35,18 @@ template <size_t Dim>
 Translation<Dim>::Translation(std::string function_of_time_name)
     : f_of_t_name_(std::move(function_of_time_name)),
       f_of_t_names_({f_of_t_name_}),
+      inner_radius_(std::nullopt),
+      outer_radius_(std::nullopt),
+      f_of_r_(nullptr),
+      center_(make_array<Dim, double>(0.0)) {}
+
+template <size_t Dim>
+Translation<Dim>::Translation(std::string function_of_time_name,
+                              double inner_radius, double outer_radius)
+    : f_of_t_name_(std::move(function_of_time_name)),
+      f_of_t_names_({f_of_t_name_}),
+      inner_radius_(inner_radius),
+      outer_radius_(outer_radius),
       f_of_r_(nullptr),
       center_(make_array<Dim, double>(0.0)) {}
 
@@ -42,6 +57,8 @@ Translation<Dim>::Translation(
     std::array<double, Dim>& center)
     : f_of_t_name_(std::move(function_of_time_name)),
       f_of_t_names_({f_of_t_name_}),
+      inner_radius_(std::nullopt),
+      outer_radius_(std::nullopt),
       f_of_r_(std::move(radial_function)),
       center_(center) {}
 
@@ -49,6 +66,8 @@ template <size_t Dim>
 Translation<Dim>::Translation(const Translation<Dim>& Translation_Map)
     : f_of_t_name_(Translation_Map.f_of_t_name_),
       f_of_t_names_(Translation_Map.f_of_t_names_),
+      inner_radius_(Translation_Map.inner_radius_),
+      outer_radius_(Translation_Map.outer_radius_),
       center_(Translation_Map.center_) {
   if (Translation_Map.f_of_r_ == nullptr) {
     f_of_r_ = nullptr;
@@ -64,7 +83,11 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::operator()(
     const std::unordered_map<
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time) const {
-  return coord_helper(source_coords, time, functions_of_time, 0);
+  if (inner_radius_.has_value()) {
+    return piecewise_helper(source_coords, time, functions_of_time, 0);
+  } else {
+    return math_function_helper(source_coords, time, functions_of_time, 0);
+  }
 }
 
 template <size_t Dim>
@@ -84,17 +107,79 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
              << function_of_time.size()
              << ") does not match the dimension of the translation map (" << Dim
              << ").");
-  double radial_function_value = 1.0;
-  if (f_of_r_ != nullptr) {
-    const double root = root_finder(target_coords - center_, function_of_time);
-    radial_function_value = (*f_of_r_)(root);
+  // If an inner radius specified then take the inverse of the
+  // piecewise specific translation.
+  if (inner_radius_.has_value()) {
+    auto target_radius = magnitude(target_coords);
+    double non_translated_radius = 0.;
+    for (size_t i = 0; i < Dim; i++) {
+      non_translated_radius +=
+          square(gsl::at(target_coords, i) - function_of_time[i]);
+    }
+    non_translated_radius = sqrt(non_translated_radius);
+    // Assumes no translation if the target radius is greater than the
+    // outer radius.
+    if (target_radius >= outer_radius_.value()) {
+      return result;
+    } else if (non_translated_radius <= inner_radius_.value()) {
+      for (size_t i = 0; i < Dim; i++) {
+        gsl::at(result, i) -= function_of_time[i];
+      }
+    } else {
+      // We need to solve a quadratic of the form w^2a + wb + c where
+      // a = T(t)^2 - (R_{out} - R_{in})^2,
+      // b = 2(R_{out}(R_{out} - R_{in}) - T(t)\vec{\bar{\xi}}), and
+      // c = \vec{\bar{\xi}}^2 - R_{out}^2
+      double a = 0.;
+      double b = 0.;
+      double c = 0.;
+      for (size_t i = 0; i < Dim; i++) {
+        a += square(function_of_time[i]);
+        b -= 2 * function_of_time[i] * gsl::at(target_coords, i);
+        c += square(gsl::at(target_coords, i));
+      }
+      a -= square(outer_radius_.value() - inner_radius_.value());
+      b += 2 * outer_radius_.value() *
+           (outer_radius_.value() - inner_radius_.value());
+      c -= square(outer_radius_.value());
+      auto roots = real_roots(a, b, c);
+      double radial_falloff_factor = 0.;
+      ASSERT(
+          roots.has_value() and roots.value()[0] >= 0 and roots.value()[1] >= 0,
+          "One of the roots is negative, this should not happen.");
+      if (roots.value()[0] <= 1.0) {
+        ASSERT(roots.value()[1] > 1.0,
+               "Singular map: two solutions between 0 and 1");
+        radial_falloff_factor = roots.value()[0];
+      } else if (roots.value()[1] <= 1.0) {
+        radial_falloff_factor = roots.value()[1];
+      } else if (roots.value()[0] <=
+                     std::numeric_limits<double>::epsilon() + 1.0 or
+                 roots.value()[1] <=
+                     std::numeric_limits<double>::epsilon() + 1.0) {
+        radial_falloff_factor = 1.0;
+      }
+      for (size_t i = 0; i < Dim; i++) {
+        gsl::at(result, i) -=
+            gsl::at(function_of_time, i) * radial_falloff_factor;
+      }
+    }
+    return result;
+    // Otherwise take the inverse of the radial MathFunction translation.
+  } else {
+    double radial_function_value = 1.0;
+    // If a MathFunction is specified, do a root find.
+    if (f_of_r_ != nullptr) {
+      const double root =
+          root_finder(target_coords - center_, function_of_time);
+      radial_function_value = (*f_of_r_)(root);
+    }
+    for (size_t i = 0; i < Dim; i++) {
+      gsl::at(result, i) -= function_of_time[i] * radial_function_value;
+    }
+    return result;
   }
-  for (size_t i = 0; i < Dim; i++) {
-    gsl::at(result, i) -= function_of_time[i] * radial_function_value;
-  }
-  return result;
 }
-
 template <size_t Dim>
 template <typename T>
 std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
@@ -102,7 +187,11 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::frame_velocity(
     const std::unordered_map<
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time) const {
-  return coord_helper(source_coords, time, functions_of_time, 1);
+  if (inner_radius_.has_value()) {
+    return piecewise_helper(source_coords, time, functions_of_time, 1);
+  } else {
+    return math_function_helper(source_coords, time, functions_of_time, 1);
+  }
 }
 
 template <size_t Dim>
@@ -113,40 +202,73 @@ Translation<Dim>::jacobian(
     const std::unordered_map<
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time) const {
-  if (f_of_r_ == nullptr) {
-    return identity<Dim>(dereference_wrapper(source_coords[0]));
-  } else {
-    std::array<tt::remove_cvref_wrap_t<T>, Dim> distance_to_center{};
-    for (size_t i = 0; i < Dim; i++) {
-      gsl::at(distance_to_center, i) =
-          gsl::at(source_coords, i) - gsl::at(center_, i);
-    }
-    auto radius = magnitude(distance_to_center);
-
+  // If inner radius has a value, then calculate the jacobian for the piecewise
+  // translation.
+  if (inner_radius_.has_value()) {
+    auto radius = magnitude(source_coords);
     const DataVector function_of_time =
         functions_of_time.at(f_of_t_name_)->func(time)[0];
-
     auto result = make_with_value<
         tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>>(
         dereference_wrapper(source_coords[0]), 0.0);
     for (size_t i = 0; i < Dim; i++) {
       for (size_t j = 0; j < Dim; j++) {
         for (size_t k = 0; k < get_size(radius); k++) {
-          if (get_element(radius, k) > 1.e-13) {
-            result.get(i, j) = (*f_of_r_).first_deriv(get_element(radius, k)) *
-                               gsl::at(function_of_time, i) *
-                               gsl::at(distance_to_center, j) /
-                               get_element(radius, k);
-          } else {
-            result.get(i, j) = (*f_of_r_).second_deriv(get_element(radius, k)) *
-                               gsl::at(function_of_time, i) *
-                               gsl::at(distance_to_center, j);
+          // The jacobian is the identity for regions not between the inner and
+          // outer radius.
+          if (get_element(radius, k) > inner_radius_.value() and
+              get_element(radius, k) < outer_radius_.value()) {
+            // using the derivative of the radial falloff factor as
+            // \frac{dw}{dr} = \frac{-1.0}{R_{out} - R{in}}
+            get_element(result.get(i, j), k) =
+                (-1.0 / (outer_radius_.value() - inner_radius_.value())) *
+                gsl::at(function_of_time, i) *
+                get_element(dereference_wrapper(gsl::at(source_coords, j)), k) /
+                get_element(radius, k);
           }
         }
       }
       result.get(i, i) += 1.0;
     }
     return result;
+    // Otherwise calculate the jacobian for the MathFunction translation.
+  } else {
+    if (f_of_r_ == nullptr) {
+      return identity<Dim>(dereference_wrapper(source_coords[0]));
+    } else {
+      std::array<tt::remove_cvref_wrap_t<T>, Dim> distance_to_center{};
+      for (size_t i = 0; i < Dim; i++) {
+        gsl::at(distance_to_center, i) =
+            gsl::at(source_coords, i) - gsl::at(center_, i);
+      }
+      auto radius = magnitude(distance_to_center);
+      const DataVector function_of_time =
+          functions_of_time.at(f_of_t_name_)->func(time)[0];
+
+      auto result = make_with_value<
+          tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>>(
+          dereference_wrapper(source_coords[0]), 0.0);
+      for (size_t i = 0; i < Dim; i++) {
+        for (size_t j = 0; j < Dim; j++) {
+          for (size_t k = 0; k < get_size(radius); k++) {
+            if (get_element(radius, k) > 1.e-13) {
+              get_element(result.get(i, j), k) =
+                  (*f_of_r_).first_deriv(get_element(radius, k)) *
+                  gsl::at(function_of_time, i) *
+                  get_element(gsl::at(distance_to_center, j), k) /
+                  get_element(radius, k);
+            } else {
+              ASSERT(equal_within_roundoff(
+                         (*f_of_r_).first_deriv(get_element(radius, k)), 0.),
+                     "Non-smooth map, MathFunction derivative should be 0.0 at "
+                     "map center");
+            }
+          }
+        }
+        result.get(i, i) += 1.0;
+      }
+      return result;
+    }
   }
 }
 template <size_t Dim>
@@ -157,7 +279,7 @@ Translation<Dim>::inv_jacobian(
     const std::unordered_map<
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time) const {
-  if (f_of_r_ == nullptr) {
+  if (f_of_r_ == nullptr and not inner_radius_.has_value()) {
     return identity<Dim>(dereference_wrapper(source_coords[0]));
   } else {
     return determinant_and_inverse(
@@ -168,7 +290,8 @@ Translation<Dim>::inv_jacobian(
 
 template <size_t Dim>
 template <typename T>
-std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
+std::array<tt::remove_cvref_wrap_t<T>, Dim>
+Translation<Dim>::math_function_helper(
     const std::array<T, Dim>& source_coords, double time,
     const std::unordered_map<
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
@@ -200,6 +323,60 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
   for (size_t i = 0; i < Dim; i++) {
     gsl::at(result, i) +=
         gsl::at(func_or_deriv_of_time, i) * radial_function_value;
+  }
+  return result;
+}
+
+template <size_t Dim>
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::piecewise_helper(
+    const std::array<T, Dim>& source_coords, double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time,
+    const size_t function_or_deriv_index) const {
+  const auto func_or_deriv_of_time =
+      gsl::at(functions_of_time.at(f_of_t_name_)->func_and_deriv(time),
+              function_or_deriv_index);
+  ASSERT(func_or_deriv_of_time.size() == Dim,
+         "The dimension of the function of time ("
+             << func_or_deriv_of_time.size()
+             << ") does not match the dimension of the translation map (" << Dim
+             << ").");
+  std::array<tt::remove_cvref_wrap_t<T>, Dim> result{};
+  // sizing the result and getting the radial function value
+  for (size_t i = 0; i < Dim; i++) {
+    gsl::at(result, i) = gsl::at(source_coords, i);
+  }
+  auto radius = magnitude(result);
+  if (function_or_deriv_index == 1) {
+    for (size_t i = 0; i < Dim; i++) {
+      gsl::at(result, i) = 0.0;
+    }
+  }
+  for (size_t k = 0; k < get_size(radius); k++) {
+    if (get_element(radius, k) <= inner_radius_.value()) {
+      for (size_t i = 0; i < Dim; i++) {
+        get_element(gsl::at(result, i), k) += gsl::at(func_or_deriv_of_time, i);
+      }
+      ASSERT(max(magnitude(result)) <= outer_radius_.value(),
+             "Coordinates translated outside outer radius, This should not "
+             "happen");
+    } else if (get_element(radius, k) > inner_radius_.value() and
+               get_element(radius, k) < outer_radius_.value()) {
+      // this is the linear falloff factor w in the documentation of the form
+      // w = (R_{out} - r) / (R_{out} - R_{in})
+      const double radial_falloff_factor =
+          (outer_radius_.value() - get_element(radius, k)) /
+          (outer_radius_.value() - inner_radius_.value());
+      for (size_t i = 0; i < Dim; i++) {
+        get_element(gsl::at(result, i), k) +=
+            gsl::at(func_or_deriv_of_time, i) * radial_falloff_factor;
+      }
+      ASSERT(max(magnitude(result)) <= outer_radius_.value(),
+             "Coordinates translated outside outer radius, This should not "
+             "happen");
+    }
   }
   return result;
 }
@@ -239,7 +416,7 @@ double Translation<Dim>::root_finder(
 
 template <size_t Dim>
 void Translation<Dim>::pup(PUP::er& p) {
-  size_t version = 1;
+  size_t version = 2;
   p | version;
   // Remember to increment the version number when making changes to this
   // function. Retain support for unpacking data written by previous versions
@@ -253,6 +430,13 @@ void Translation<Dim>::pup(PUP::er& p) {
   } else if (p.isUnpacking()) {
     f_of_r_ = nullptr;
     center_ = make_array<Dim, double>(0.0);
+  }
+  if (version >= 2) {
+    p | inner_radius_;
+    p | outer_radius_;
+  } else if (p.isUnpacking()) {
+    inner_radius_ = std::nullopt;
+    outer_radius_ = std::nullopt;
   }
 
   // No need to pup this because it is uniquely determined by f_of_t_name_
@@ -268,7 +452,9 @@ bool operator==(const Translation<Dim>& lhs, const Translation<Dim>& rhs) {
          (lhs.f_of_r_ == nullptr) == (rhs.f_of_r_ == nullptr) and
          ((lhs.f_of_r_ == nullptr and rhs.f_of_r_ == nullptr) or
           *lhs.f_of_r_ == *rhs.f_of_r_) and
-         lhs.center_ == rhs.center_;
+         lhs.center_ == rhs.center_ and
+         lhs.inner_radius_ == rhs.inner_radius_ and
+         lhs.outer_radius_ == rhs.outer_radius_;
 }
 
 // Explicit instantiations
--- a/src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
@@ -27,23 +27,46 @@ class er;
 namespace domain::CoordinateMaps::TimeDependent {
 /*!
  * \ingroup CoordMapsTimeDependentGroup
- * \brief Translation map defined by \f$\vec{x} = \vec{\xi}+F(r)\vec{T}(t)\f$.
+ * \brief Translation map defined by \f$\vec{x} = \vec{\xi}+F(r)\vec{T}(t)\f$
+ * where $F(r)$ takes on different forms based on which constructor is used.
  *
- * \details The map adds a translation, \f$F(r)\vec{T}(t)\f$, to the coordinates
- * \f$\vec{\xi}\f$, where \f$\vec{T}(t)\f$ is a FunctionOfTime and\f$F(r)\f$ is
- * a 1D radial MathFunction. The radius of each point is found by subtracting
- * the center map argument from the coordinates \f$\vec{\xi}\f$ or the target
- * coordinates \f$\vec{\bar{\xi}}\f$. The Translation Map class is overloaded so
- * that the user can choose if a radial depedence is needed for their problem.
- * If a radial dependence is not specified, this sets \f$F(r) = 1\f$.
+ * \details The map adds a translation to the coordinates $\vec{\xi}$ based on
+ * what type of translation is needed. For the piecewise translation, a
+ * translation $F(r)\vec{T}(t)$ is added to $\vec{\xi}$ based on what region
+ * $|\vec{\xi}|$ is in. For coordinates within the inner radius, $F(r) = 1$
+ * causing a uniform translation. Coordinates in between the inner and outer
+ * radius have a linear radial falloff applied to them. Coordinates beyond the
+ * outer radius have no translation applied to them $F(r) = 0$. The piecewise
+ * translation assumes that the center of your map is at (0., 0., 0.). For the
+ * radial MathFunction translation, a radial translation \f$F(r)\vec{T}(t)\f$ is
+ * added to the coordinates \f$\vec{\xi}\f$, where \f$\vec{T}(t)\f$ is a
+ * FunctionOfTime and\f$F(r)\f$ is a 1D radial MathFunction. The radius of each
+ * point is found by subtracting the center map argument from the coordinates
+ * \f$\vec{\xi}\f$ or the target coordinates \f$\vec{\bar{\xi}}\f$. The
+ * Translation Map class is overloaded so that the user can choose between a
+ * piecewise translation, radial translation or a uniform translation based on
+ * their problem. If a radial dependence is not specified, this sets \f$F(r) =
+ * 1\f$.
  *
  * ### Mapped Coordinates
- *
- * The translation map translates the coordinates \f$\vec{\xi}\f$ to the
- * target coordinates
+ * The piecewise translation translates the coordinates $\vec{\xi}$
+ * to the target coordinates $\vec{\bar{\xi}}$ based on the region $\vec{\xi}$
+ * is in.
  * \f{equation}{
- * \vec{\bar{\xi}} = \vec{\xi} + F(r)\vec{T}(t)
+ * \vec{\bar{\xi}} = \left\{\begin{array}{ll}\vec{\xi} + \vec{T}(t), &
+ * |\vec{\xi}| \leq R_{in}, \\ \vec{\xi} + wT(t), &  R_{in} < |\vec{\xi}| <
+ * R_{out}, \\ \vec{\xi}, & |\vec{\xi}| \geq R_{out} \end{array}\right.
  * \f}
+ *
+ * Where $R_{in}$ is the inner radius, $R_{out}$ is the outer radius, and $w$ is
+ * the radial falloff factor found through
+ * \f{equation}{
+ * w = \frac{R_{out} - |\vec{\xi}|}{R_{out} - R_{in}}
+ * \f}
+ *
+ * The radial MathFunction translation translates the coordinates
+ * \f$\vec{\xi}\f$ to the target coordinates \f{equation}{\vec{\bar{\xi}} =
+ * \vec{\xi} + F(r)\vec{T}(t) \f}
  *
  * If you only supply a FunctionOfTime to the constructor of this class, the
  * radial function will be set to 1.0 causing a uniform translation for your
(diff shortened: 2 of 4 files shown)


src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -27,18 +27,52 @@ class er;
 namespace domain::CoordinateMaps::TimeDependent {
 /*!
  * \ingroup CoordMapsTimeDependentGroup
- * \brief Translation map defined by \f$\vec{x} = \vec{\xi}+F(r)\vec{T}(t)\f$.
+ * \brief Translation map defined by either a piecewise [$\vec{x} =
```
knelli2: I would keep the original brief. It works for all 3 versions of this class, it's just $F(r)$ takes on different functional forms.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -27,18 +27,52 @@ class er;
...
  *
- * \details The map adds a translation, \f$F(r)\vec{T}(t)\f$, to the coordinates
- * \f$\vec{\xi}\f$, where \f$\vec{T}(t)\f$ is a FunctionOfTime and\f$F(r)\f$ is
- * a 1D radial MathFunction. The radius of each point is found by subtracting
- * the center map argument from the coordinates \f$\vec{\xi}\f$ or the target
- * coordinates \f$\vec{\bar{\xi}}\f$. The Translation Map class is overloaded so
- * that the user can choose if a radial depedence is needed for their problem.
- * If a radial dependence is not specified, this sets \f$F(r) = 1\f$.
+ * \details The map adds a translation based on what portion of the evolution is
+ * in (inspiral, merger, ringdown). During the inspiral, a piecewise translation
```
knelli2: Throughout this PR, I think you should remove all references to `inspiral`, `merger`, and `ringdown` (both docs and variable/function names). Even though we as devs know that the different constructors will be used for different parts of the evolution, this class doesn't really care about that. Furthermore, if we decide we want to use a different constructor for a different part of the evolution than is stated here, then these docs would be wrong. You explanations for what each constructor does are good, though.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -27,18 +27,52 @@ class er;
...
+ * is added to $\vec{\xi}$, $\vec{x} = \vec{\xi}+\vec{T}(t): \vec{\xi} <= i$,
+ * $\vec{x} = \vec{\xi}+w\vec{T}(t): i < \vec{\xi} < o$, $\vec{x} = \vec{\xi}$
```
knelli2: Instead of $i$ and $o$, can you use $R_{\mathrm{in}}$ and $R_{\mathrm{out}}$ for the inner and outer radii?

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -27,18 +27,52 @@ class er;
...
+ * to the target coordinates based on the region $\vec{\xi}$ is in.
+ * \f{equation}{
+ * \vec{\bar{\xi}} = \vec{\xi} + \vec{T}(t) if |\vec{\xi}| <= i
+ * \f}
+ * \f{equation}{
+ * \vec{\bar{\xi}} = \vec{\xi} + wT(t) if i < |\vec{\xi}| < o
+ * \f}
+ * \f{equation}{
+ * \vec{\bar{\xi}} = \vec{\xi} if |\vec{xi}| >= o
+ * \f}
```
knelli2: For the linear fall-off constructor, can you make these docs more like `src/Domain/CoordinateMaps/TimeDependent/SphericalCompression.hpp73-80` with nice alignment and it's own equation block? The formatting is a bit odd here if you look at the rendered version.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
knelli2: I don't think you need to add these equations here. You'll be specifying them below anyways.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -172,9 +252,9 @@ class Translation {
...
       const Translation<LocalDim>& lhs, const Translation<LocalDim>& rhs);
 
-  // This helper function computes the translated coordinates or frame velocity
-  // based on the option passed in, 0 for translated coordinates, and frame
-  // velocity for any other number.
+  // These 2 helper functions compute the translated coordinates or frame
+  // velocity based on the option passed in, 0 for translated coordinates, and
+  // frame velocity for any other number.
   template <typename T>
   std::array<tt::remove_cvref_wrap_t<T>, Dim> coord_helper(
```
knelli2: Maybe name this `math_function_helper` now that you'll have another coordinate helper for a different constructor?

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -184,11 +264,22 @@ class Translation {
...
+          functions_of_time,
+      size_t function_or_deriv_index) const;
+
   double root_finder(const std::array<double, Dim>& distance_to_center,
                      const DataVector& function_of_time) const;
 
   std::string f_of_t_name_{};
   std::unordered_set<std::string> f_of_t_names_;
+  double inner_radius_{};
+  double outer_radius_{};
```
knelli2: Throughout the code, you use `std::isnan(inner_radius_)` to check if you're using the linear falloff constructor and I believe that is causing the FPEs in your tests, since you set them to a signaling NaN. C++ actually has a data-type that does exactly what you want called `std::optional<double>`. Basically wherever you have the `std::isnan` you replace with `inner_radius_.has_value()` and then whenever you need the actual value you do `inner_radius_.value()`. We follow this pattern in a lot of other places in the code as well; where we have a value, but only sometimes, and we need to check if we actually have a value.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -84,25 +105,73 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
+    }
+    non_translated_radius = sqrt(non_translated_radius);
+    if (target_radius >= outer_radius_) {
+      return result;
+    } else if (non_translated_radius <= inner_radius_) {
+      for (size_t i = 0; i < Dim; i++) {
+        gsl::at(result, i) -= function_of_time[i];
+      }
+    } else {
+      // Solving a quadratic of known quantities.
```
knelli2: Can you add more to this code comment explaining which exact quadratic equation you are solving and what `a`, `b`, `c` are explicitly? That'll go a long way in the future when somebody else has to look at this (like you are in SpEC right now!).

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -84,25 +105,73 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
+        b -= 2 * function_of_time[i] * gsl::at(target_coords, i);
+        c += square(gsl::at(target_coords, i));
+      }
+      a -= square(outer_radius_ - inner_radius_);
+      b += 2 * outer_radius_ * (outer_radius_ - inner_radius_);
+      c -= square(outer_radius_);
+      auto roots = real_roots(a, b, c);
+      double radial_falloff_factor = 0.;
+      if (roots.value()[0] >= 0.0 && roots.value()[0] <= 1.0) {
+        ASSERT(roots.value()[1] < 0.0 || roots.value()[1] > 1.0,
```
knelli2: We use `and` and `or` instead of `&&` and `||`. It's just easier for people to read. Same for the rest of this PR.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -84,25 +105,73 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
+               "Singular map: two solutions between 0 and 1");
+        radial_falloff_factor = roots.value()[0];
+      } else if (roots.value()[1] >= 0.0 && roots.value()[1] <= 1.0) {
+        radial_falloff_factor = roots.value()[1];
```
knelli2: Do you need a corresponding `ASSERT` here for `[0]` like you have in the first branch for `[1]`?

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -84,25 +105,73 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
              << ") does not match the dimension of the translation map (" << Dim
              << ").");
-  double radial_function_value = 1.0;
-  if (f_of_r_ != nullptr) {
-    const double root = root_finder(target_coords - center_, function_of_time);
-    radial_function_value = (*f_of_r_)(root);
-  }
-  for (size_t i = 0; i < Dim; i++) {
-    gsl::at(result, i) -= function_of_time[i] * radial_function_value;
+  if (std::isnan(inner_radius_)) {
```
knelli2: Can you add some code comments for all these `if` statements saying what is actually happening? Like "Math function specific. doing a root find" or "inner/outer radius specific". That way it's a bit easier for devs to parse all the `if`s and `for`s

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -204,6 +297,65 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
...
+    const std::array<T, Dim>& source_coords, double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time,
+    const size_t function_or_deriv_index) const {
+  ASSERT(functions_of_time.count(f_of_t_name_) == 1,
+         "The function of time '" << f_of_t_name_
+                                  << "' is not one of the known functions of "
+                                     "time. The known functions of time are: "
+                                  << keys_of(functions_of_time));
```
knelli2: We don't need to add this anymore. I made some changes that does this check one level above in the actual coordinate map.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -204,6 +297,65 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
...
+  const auto func_or_deriv_of_time =
+      gsl::at(functions_of_time.at(f_of_t_name_)->func_and_deriv(time),
+              function_or_deriv_index);
+  ASSERT(func_or_deriv_of_time.size() == Dim,
```
knelli2, line 341: Keep this one though.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -204,6 +297,65 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
...
+    }
+  }
+  for (size_t k = 0; k < get_size(radius); k++) {
+    if (get_element(radius, k) <= inner_radius_) {
+      for (size_t i = 0; i < Dim; i++) {
+        get_element(gsl::at(result, i), k) += gsl::at(func_or_deriv_of_time, i);
+      }
+      ASSERT(max(magnitude(result)) <= outer_radius_,
+             "Coordinates translated outside outer radius, This should not"
+                 << " happen");
```
knelli2: Don't need the `<<` here. Just join the string. (Same below)

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -204,6 +297,65 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
...
+    } else if (get_element(radius, k) > inner_radius_ &&
+               get_element(radius, k) < outer_radius_) {
+      // this factor comes from the SpEC implementation
```
knelli2: Can you write out in the code comment what the function form of w is and that it's a linear falloff? I don't think you need to mention SpEC here since a linear falloff is simple enough.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -150,7 +201,15 @@ void test_translation() {
                   f_of_t_list);
     test_inv_jacobian(radial_translation_map_deserialized, point_xi, t,
                       f_of_t_list);
-
+    // This is for an edge case where the numerical derivative taken in the
+    // TestMapHelpers jacobian pushes the coordinate to the next boundary
+    // causing it to be translated improperly.
+    if (inspiral_radius <= 0.9997) {
```
knelli2: Could you explain a bit more what causes this? (github comment) I'm not sure I understand.

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
```diff
@@ -360,16 +360,17 @@ void test_coordinate_map_argument_types(
     }
     {
       INFO("Test inverse Jacobian");
+      Approx custom_approx = Approx::custom().epsilon(1.e-11);
```
knelli2, line 500: Hmm I'm not sure changing the approx only here is a good idea, but I'm also not sure how else to avoid some slightly larger differences.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -93,16 +103,52 @@ void test_translation() {
...
     std::array<double, Dim> radial_frame_vel{};
+    std::array<double, Dim> inspiral_frame_vel{};
+    std::array<double, Dim> far_inspiral_frame_vel{0.};
     for (size_t i = 0; i < Dim; i++) {
       gsl::at(frame_vel, i) = f_of_t->func_and_deriv(t)[1][i];
       gsl::at(radial_frame_vel, i) =
           f_of_t->func_and_deriv(t)[1][i] * gaussian(radius);
+      if (inspiral_radius <= inner_radius) {
+        gsl::at(inspiral_frame_vel, i) = f_of_t->func_and_deriv(t)[1][i];
+      } else if (inspiral_radius > inner_radius &&
```
knelli2: I don't think this needs to be an `else if`, just an `else` because the point distribution for the inner radius is always less than the outer radius and you just checked if the point was inside the inner radius (same below)

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -160,8 +219,11 @@ void test_translation() {
   CHECK_FALSE(radial_translation_map != radial_translation_map);
   CHECK_FALSE(radial_translation_map_deserialized !=
               radial_translation_map_deserialized);
+  CHECK_FALSE(inspiral_translation_map != inspiral_translation_map);
+  CHECK_FALSE(inspiral_translation_map_deserialized !=
+              inspiral_translation_map_deserialized);
 
-  // Check serialization
+  //   // Check serialization
```
knelli2: Drop change?

knelli2 (changes requested):
Thank you for working out all the kinks in this! Let me know if you want to chat about any of the changes I requested.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
AlexCarpenter46: Ah okay, that makes sense, yeah I was wondering if the copy constructor was causing the fpes. That sounds perfect though, will change :)

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: okay, should I take this out of the other spots in the translation as well?

knelli2: If there are any other places that check if the function of time name is in the functions of time, yes. But I think this is the only one?

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
AlexCarpenter46: Oh yeah, so in the numerical derivative function used in the jacobian, it calls the operator with tiny steps forward something like x1 = x + dx, x2 = x1 + dx and x3 = x2 + dx where dx = 1e-4 so it moves the point_xi up to 3e-4 forward and calls the operator with each different value. When it does that, it enters the region where the translation is different causing the jacobian to not be the identity matrix anymore. However, when it calls the jacobian, it still uses the original point_xi so there's a mixup between which region it's in. So to stop this from causing problems I just made it so that the test doesn't get called if your radius > 0.9997. I think this fix is fine because this problem can only arise when comparing the analytic jacobian to the numerical one, so this shouldn't affect evolutions or be thought of as a bug I don't think. However, this did remind me that I meant to put `or inspiral_radius > 1.0`

knelli2: Ahh ok that makes sense, thank you! I'd suggest a couple things:

- Instead of hard-coding `0.9997`, do `inspiral_radius <= 0.99 * inner_radius`. This gives you a safe-enough buffer and also doesn't hard code a value.
- I think it can also be `or inspiral_radius >= inner_radius`
- Can you add basically this exact github comment into the code? It was a very helpful explanation and I think the test readability would benefit from it.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: oh whoops yeah, I think I'm remembering from when the coord helper had it

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: I don't think so cause if the if statement above isn't true then that means the [0] is either below 0 or above 1 so it couldn't be a singluar map

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -31,14 +31,16 @@ namespace domain {
 namespace {
 template <size_t Dim>
 void test_translation() {
-  MAKE_GENERATOR(gen);
+  MAKE_GENERATOR(gen, 1927162167);
```
AlexCarpenter46: need to take this out, was testing an edge case after previous comments.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -27,24 +27,45 @@ class er;
...
+ * For the radial MathFunction translation, a radial translation
+ * \f$F(r)\vec{T}(t)\f$ is added to the coordinates \f$\vec{\xi}\f$, where
+ * \f$\vec{T}(t)\f$ is a FunctionOfTime and\f$F(r)\f$ is a 1D radial
+ * MathFunction. The radius of each point is found by subtracting the center map
+ * argument from the coordinates \f$\vec{\xi}\f$ or the target coordinates
+ * \f$\vec{\bar{\xi}}\f$. The Translation Map class is overloaded so that the
+ * user can choose between a piecewise translation, radial translation or a
+ * uniform translation based on their problem. If a radial dependence is not
+ * specified, this sets \f$F(r) = 1\f$. If no inner or outer radius is specified
+ * then they'll be set to NaNs.
```
knelli2: This is no longer true

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -27,24 +27,45 @@ class er;
...
  *
  * ### Mapped Coordinates
+ * The piecewise translation translates the coordinates $\vec{\xi}$
+ * to the target coordinates based on the region $\vec{\xi}$ is in.
+ * \f{equation}{
+ * \vec{\bar{\xi}} &= \left\{\begin{array}{ll}\vec{\xi} + \vec{T}(t), &
+ * |\vec{\xi}| \leq R_{in}, \\ \vec{\xi} + wT(t), &  R_{in} < |\vec{\xi}| <
+ * R_{out}, \\ \vec{\xi}, & |\vec{\xi}| \geq R_{out} \end{array}\right. \f}
```
knelli2: This isn't rendering properly. You can use `make doc` locally to build documentation and view it beforehand. Also, when putting blocks in like this, make sure there's a blank line above and below the block. Both `\f` should be on their own lines as well.

I also think you can use latex syntax like `\begin{equation} ... \end{equation}`. I believe that's been added.

You should apply this to all the dox changes in this PR.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -27,24 +27,45 @@ class er;
...
  *
- * The translation map translates the coordinates \f$\vec{\xi}\f$ to the
- * target coordinates
+ * Where $R_{in}$ is the inner radius, $R_{out}$ is the outer radius, and $w$ is
+ * the radial falloff factor found through
  * \f{equation}{
- * \vec{\bar{\xi}} = \vec{\xi} + F(r)\vec{T}(t)
+ * w = \frac{R_{out} - |\vec{xi}|}{R_{out} - R_{in}}
```
knelli2: Missing `\` for $xi$ -> $\xi$

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -55,9 +76,28 @@ namespace domain::CoordinateMaps::TimeDependent {
...
+ * The piecewise inverse translates the coordinates
+ * \f$\vec{\bar{\xi}}\f$ to the original coordinates based on what region
+ * $\vec{\bar{\xi}}$ is in.
+ * \f{equation}{
+ * \vec{\xi} &= \left\{\begin{array}{ll}\vec{\bar{\xi}} -
+ * \vec{T}(t), & |\vec{\bar{\xi}}| \leq R_{in}, or, |\vec{\bar{\xi}} - T(t)|
+ * \leq R_{in}, \\ \vec{\bar{\xi}} - wT(t), &  R_{in} < |\vec{\bar{\xi}}| <
+ * R_{out}, \\ \vec{\bar{\xi}}, & |\vec{\bar{\xi}}| \geq R_{out}
+ * \end{array}\right.
+ * \f}
```
knelli2: This also doesn't render properly.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -67,18 +107,30 @@ namespace domain::CoordinateMaps::TimeDependent {
  * \f}
  *
  * ### Frame Velocity
+ * For the piecewise translation, the frame velocity is found through
+ * \f{equation}{
+ * \vec{v} &= \left\{\begin{array}{ll}\frac{\vec{dT}(t)}{dt}, & |\vec{\xi}| \leq
+ * R_{in}, \\ w\frac{\vec{dT}(t)}{dt}, &  R_{in} < |\vec{\xi}| < R_{out}, \\ 0,
+ * & |\vec{\xi}| \geq R_{out} \end{array}\right. \f}
```
knelli2: Also doesn't render properly.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -172,11 +227,20 @@ class Translation {
...
+      const std::array<T, Dim>& source_coords, double time,
+      const std::unordered_map<
+          std::string,
+          std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+          functions_of_time,
+      size_t function_or_deriv_index) const;
+
   template <typename T>
-  std::array<tt::remove_cvref_wrap_t<T>, Dim> coord_helper(
+  std::array<tt::remove_cvref_wrap_t<T>, Dim> inspiral_helper(
```
knelli2: Rename to `piecewise_helper`?

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -32,6 +34,18 @@ template <size_t Dim>
 Translation<Dim>::Translation(std::string function_of_time_name)
     : f_of_t_name_(std::move(function_of_time_name)),
       f_of_t_names_({f_of_t_name_}),
+      inner_radius_(),
+      outer_radius_(),
```
knelli2: A default constructed `std::optional` is a `std::nullopt`, so you can either not include these two lines since the variables will already be what you want, or if you want to explicitly keep them here, do `inner_radius_(std::nullopt)`. Same for the oher constructor below.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -64,7 +82,11 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::operator()(
     const std::unordered_map<
         std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
         functions_of_time) const {
-  return coord_helper(source_coords, time, functions_of_time, 0);
+  if (!inner_radius_.has_value()) {
```
knelli2: Two things:

1. We only use `!` when it's in `!=`. If you're just using it to negate a bool, use `not` instead. Just like for `&&` -> `and`, `!` -> `and`.
2. [optional] Usually it's easier for people to read `if (condition) { ... } else { ... }` rather than `if (not condition) { ... } else { ... }`. I'd suggest flipping this so it's just `if (inner_radius_.has_value()`. If you decide to make this change, do it for all `if` statements in this PR that you added.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -84,25 +106,83 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
+      return result;
+    } else if (non_translated_radius <= inner_radius_.value()) {
+      for (size_t i = 0; i < Dim; i++) {
+        gsl::at(result, i) -= function_of_time[i];
+      }
+    } else {
+      // We need to solve a quadratic of the form w^2a + wb + c where
+      // a = T(t)^2 - (R_{out} - R_{in})^2,
+      // b = 2(R_{out}(R_{out} - R_{in}) - T(t)\vec{\bar{\xi}}), and
+      // c = \vec{\bar{\xi}}^2 - R_{out}^2
```
knelli2, line 132: `c` is different than in the docs in the hpp. Which is correct?

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -204,6 +315,60 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::coord_helper(
...
+    gsl::at(result, i) = gsl::at(source_coords, i);
+  }
+  auto radius = magnitude(result);
+  if (function_or_deriv_index == 1) {
+    for (size_t i = 0; i < Dim; i++) {
+      gsl::at(result, i) = 0.0;
+    }
+  }
+  for (size_t k = 0; k < get_size(radius); k++) {
+    if (get_element(radius, k) <= inner_radius_) {
```
knelli2: `inner_radius_.value()` Same for the rest of this function.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -268,7 +440,11 @@ bool operator==(const Translation<Dim>& lhs, const Translation<Dim>& rhs) {
          (lhs.f_of_r_ == nullptr) == (rhs.f_of_r_ == nullptr) and
          ((lhs.f_of_r_ == nullptr and rhs.f_of_r_ == nullptr) or
           *lhs.f_of_r_ == *rhs.f_of_r_) and
-         lhs.center_ == rhs.center_;
+         lhs.center_ == rhs.center_ and
+         (!lhs.inner_radius_.has_value() == !rhs.inner_radius_.has_value() or
```
knelli2: A `std::optional` actually does all this checking in its `operator==` so you just have to do `lhs.inner_radius_ == rhs.inner_radius_ and lhs.outer_radius_ == rhs.outer_radius_`

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -93,16 +103,51 @@ void test_translation() {
...
     CHECK_ITERABLE_APPROX(translation_map(point_xi, t, f_of_t_list),
                           point_xi + translation);
+    if (inspiral_radius <= inner_radius) {
+      CHECK_ITERABLE_APPROX(inspiral_translation_map(point_xi, t, f_of_t_list),
+                            point_xi + translation);
+      CHECK_ITERABLE_APPROX(inspiral_translation_map
+                                .inverse(point_xi + translation, t, f_of_t_list)
+                                .value(),
+                            point_xi);
+    } else if (inspiral_radius > inner_radius &&
```
knelli2: `&&` -> `and`

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -93,16 +103,51 @@ void test_translation() {
       radius += square(distance_to_center[i]);
     }
     radius = sqrt(radius);
+    double inspiral_radius = magnitude(point_xi);
+    double radial_falloff_factor =
```
knelli2: const

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -55,14 +57,20 @@ void test_translation() {
...
   fill_with_random_values(make_not_null(&point_xi), make_not_null(&gen),
                           make_not_null(&dist_double));
   fill_with_random_values(make_not_null(&center), make_not_null(&gen),
                           make_not_null(&dist_double));
+  fill_with_random_values(make_not_null(&far_point_xi), make_not_null(&gen),
+                          make_not_null(&far_dist_double));
   const CoordinateMaps::TimeDependent::Translation<Dim> translation_map{
       "translation"};
+  const CoordinateMaps::TimeDependent::Translation<Dim>
+      inspiral_translation_map{"translation", inner_radius, outer_radius};
```
knelli2: Rename to `piecewise_translation_map`

knelli2 (changes requested):
The previous round of changes should have been included as a fixup to make it easier to review. Unless the changes are trivial (or the reviewer says to squash directly), please keep everything in fixups. This next round of changes should also be in a fixup.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -84,25 +106,83 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
+        b -= 2 * function_of_time[i] * gsl::at(target_coords, i);
+        c += square(gsl::at(target_coords, i));
+      }
+      a -= square(outer_radius_.value() - inner_radius_.value());
+      b += 2 * outer_radius_.value() *
+           (outer_radius_.value() - inner_radius_.value());
+      c -= square(outer_radius_.value());
+      auto roots = real_roots(a, b, c);
+      double radial_falloff_factor = 0.;
+      if (roots.value()[0] >= 0.0 and roots.value()[0] <= 1.0) {
```
wthrowe: Both roots should always be non-negative, so ASSERT that and then you don't need the checks here and below.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -84,25 +106,83 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
+        ASSERT(roots.value()[1] < 0.0 or roots.value()[1] > 1.0,
+               "Singular map: two solutions between 0 and 1");
+        radial_falloff_factor = roots.value()[0];
+      } else if (roots.value()[1] >= 0.0 and roots.value()[1] <= 1.0) {
+        radial_falloff_factor = roots.value()[1];
+      }
```
wthrowe, line 161: Handle the case where both roots are greater than one.  Shouldn't happen analytically, but there might be roundoff issues.  There's nothing numerically unstable or FPE-causing here, so just setting `radial_falloff_factor` to one if a root is epsilon over should be fine.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -84,25 +106,83 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
...
+    auto target_radius = magnitude(target_coords);
+    double non_translated_radius = 0.;
+    for (size_t i = 0; i < Dim; i++) {
+      non_translated_radius +=
+          square(gsl::at(target_coords, i) - function_of_time[i]);
+    }
+    non_translated_radius = sqrt(non_translated_radius);
+    // Assumes no translation if the target radius is greater than the
+    // outer radius.
+    if (target_radius >= outer_radius_.value()) {
```
wthrowe, line 122: [optional] This is `c >= 0` (for the `c` calculated below).  I don't think there are any numerical issues here, just might be cleaner to not calculate it twice.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -113,34 +193,64 @@ Translation<Dim>::jacobian(
...
+          for (size_t k = 0; k < get_size(radius); k++) {
+            if (get_element(radius, k) > 1.e-13) {
+              result.get(i, j) =
+                  (*f_of_r_).first_deriv(get_element(radius, k)) *
+                  gsl::at(function_of_time, i) *
+                  gsl::at(distance_to_center, j) / get_element(radius, k);
+            } else {
+              result.get(i, j) =
+                  (*f_of_r_).second_deriv(get_element(radius, k)) *
+                  gsl::at(function_of_time, i) * gsl::at(distance_to_center, j);
```
wthrowe: This limit is incorrect, because `distance_to_center` and `radius` aren't independent.  The map is non-smooth if `(*f_of_r_).first_deriv(0.0)` is not zero.  Documentation also needs to be fixed.

It isn't obvious to me what the no-inner-radius case is supposed to do, so I don't know what the correct resolution is here.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -113,34 +193,64 @@ Translation<Dim>::jacobian(
...
+          functions_of_time.at(f_of_t_name_)->func(time)[0];
 
+      auto result = make_with_value<
+          tnsr::Ij<tt::remove_cvref_wrap_t<T>, Dim, Frame::NoFrame>>(
+          dereference_wrapper(source_coords[0]), 0.0);
+      for (size_t i = 0; i < Dim; i++) {
+        for (size_t j = 0; j < Dim; j++) {
+          for (size_t k = 0; k < get_size(radius); k++) {
+            if (get_element(radius, k) > 1.e-13) {
+              result.get(i, j) =
```
wthrowe: Write a test that would have caught the missing `get_element(..., k)` here and in various similar places.

src/Domain/CoordinateMaps/TimeDependent/Translation.hpp
```diff
@@ -27,24 +27,45 @@ class er;
 namespace domain::CoordinateMaps::TimeDependent {
 /*!
  * \ingroup CoordMapsTimeDependentGroup
- * \brief Translation map defined by \f$\vec{x} = \vec{\xi}+F(r)\vec{T}(t)\f$.
+ * \brief Translation map defined by \f$\vec{x} = \vec{\xi}+F(r)\vec{T}(t)\f$
+ * where $F(r)$ takes on different forms based on which constructor is used.
```
wthrowe, line 31: I may be missing it, but these forms should be given somewhere.

wthrowe (changes requested):

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: This c is correct, will correct the hpp

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: does it also check the values inside?

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: The no-inner-radius case is for when you're using the mathfunction translation instead of the piecewise translation. I think this map will be smooth since the mathfunction provided will be a gaussian with a max at the center map argument . Also this limit is taken from `/spec/Dust/Domain/SpatialCoordMaps/SpatiallyDependentTranslation.cpp` lines 304-320

wthrowe: Just because SpEC uses the wrong equation doesn't mean we should too.  I'll file an issue over there.

But erroring out if the map isn't smooth is probably the right thing to do, anyway.  I don't know where the radial function comes from, so I don't know  if it should be an ASSERT or an ERROR.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
knelli2: Yup!

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: I think the function `test_coordinate_map_argument_types` catches the missing `get_element(..., k)` and I use it for all the different maps.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
AlexCarpenter46: Okay, so then I think the correct thing to do is`ASSERT(equal_within_roundoff(*f_of_r_).first_deriv(get_element(radius, k)), 0.)` then make the jacobian at that point 0.0

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
wthrowe: Except it apparently doesn't, because the test is passing?

AlexCarpenter46: hmmm, I don't understand what you're asking then

wthrowe: The LHS of this assignment should be wrapped in `get_element`, but the tests are passing even though it isn't.

AlexCarpenter46: oh okay, that makes more sense

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -219,10 +219,11 @@ Translation<Dim>::jacobian(
               get_element(radius, k) < outer_radius_.value()) {
             // using the derivative of the radial falloff factor as
             // \frac{dw}{dr} = \frac{-1.0}{R_{out} - R{in}}
-            result.get(i, j) =
+            get_element(result.get(i, j), k) =
                 (-1.0 / (outer_radius_.value() - inner_radius_.value())) *
-                gsl::at(function_of_time, i) *
-                (gsl::at(source_coords, j) / get_element(radius, k));
+                get_element(gsl::at(function_of_time, i), k) *
```
wthrowe: This is a double.  No `get_element`.

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -250,16 +251,16 @@ Translation<Dim>::jacobian(
         for (size_t j = 0; j < Dim; j++) {
           for (size_t k = 0; k < get_size(radius); k++) {
             if (get_element(radius, k) > 1.e-13) {
-              result.get(i, j) =
+              get_element(result.get(i, j), k) =
                   (*f_of_r_).first_deriv(get_element(radius, k)) *
-                  gsl::at(function_of_time, i) *
-                  gsl::at(distance_to_center, j) / get_element(radius, k);
+                  get_element(gsl::at(function_of_time, i), k) *
```
wthrowe: No `get_element`.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -209,13 +246,41 @@ void test_translation() {
...
       test_jacobian(piecewise_translation_map, point_xi, t, f_of_t_list);
       test_inv_jacobian(piecewise_translation_map, point_xi, t, f_of_t_list);
+      test_jacobian(piecewise_translation_map_deserialized, point_xi, t,
+                    f_of_t_list);
+      test_inv_jacobian(piecewise_translation_map_deserialized, point_xi, t,
+                        f_of_t_list);
+    }
+    // Checking if any radius in the datavector is within the range above.
+    bool jac_flag = true;
+    for (size_t i = 0; i < get_size(get_element(point_xi_dv, 0)); i++) {
```
wthrowe: Use `gsl::at` when the type is known to be an `array`, instead of `get_element`.
Use `.size()` instead of `get_size` when the type is known to be a container.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -209,13 +246,41 @@ void test_translation() {
...
+      if (get_element(piecewise_radius_dv, i) > inner_radius * .99 and
+          get_element(piecewise_radius_dv, i) < inner_radius * 1.01) {
```
wthrowe: Use `[]` or `gsl::at` instead of `get_element`.

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
```diff
@@ -147,6 +147,68 @@ void test_jacobian(
...
+void test_jacobian(const Map& map,
+                   const std::array<DataVector, Map::dim>& test_point) {
+  INFO("Test Jacobian");
+  CAPTURE(test_point);
+  // Our default approx value is too stringent for this test
+  Approx local_approx = Approx::custom().epsilon(1e-10).scale(1.0);
+  const double dx = 1e-4;
+  const auto jacobian = map.jacobian(test_point);
+  for (size_t i = 0; i < Map::dim; ++i) {
+    for (size_t k = 0; k < get_size(get_element(test_point, 0)); k++) {
```
wthrowe: Again `gsl::at(test_point, 0).size()` (or `test_point.front().size()`).

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
```diff
@@ -147,6 +147,68 @@ void test_jacobian(
...
+      const auto numerical_deriv_i =
+          numerical_derivative(map, get_element(test_point, k), i, dx);
```
wthrowe: Again `gsl::at`.

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
```diff
@@ -147,6 +147,68 @@ void test_jacobian(
...
+      for (size_t j = 0; j < Map::dim; ++j) {
+        INFO("i: " << i << " j: " << j);
+        CHECK(get_element(jacobian.get(j, i), k) ==
```
wthrowe: Again `at` or `[]`.

Change all the ones below as well.

wthrowe (changes requested):
I believe this sufficiently tests for the DataVector indexing problems.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Translation.cpp
```diff
@@ -257,9 +257,9 @@ void test_translation() {
     }
     // Checking if any radius in the datavector is within the range above.
     bool jac_flag = true;
-    for (size_t i = 0; i < get_size(get_element(point_xi_dv, 0)); i++) {
-      if (get_element(piecewise_radius_dv, i) > inner_radius * .99 and
-          get_element(piecewise_radius_dv, i) < inner_radius * 1.01) {
+    for (size_t i = 0; i < get_size(gsl::at(point_xi_dv, 0)); i++) {
```
wthrowe: `get_size` -> `.size()`

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
```diff
@@ -220,6 +282,83 @@ void test_inv_jacobian(
...
+                       const std::array<DataVector, Map::dim>& test_point) {
+  INFO("Test inverse Jacobian");
+  CAPTURE(test_point);
+  const auto jacobian = map.jacobian(test_point);
+  const auto inv_jacobian = map.inv_jacobian(test_point);
+
+  const auto expected_identity = [&jacobian, &inv_jacobian]() {
+    auto identity =
+        make_with_value<tnsr::Ij<DataVector, Map::dim, Frame::NoFrame>>(
+            get_size(jacobian.get(0, 0)), 0.0);
```
wthrowe: `.size()`

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
```diff
@@ -220,6 +282,83 @@ void test_inv_jacobian(
...
+  CAPTURE(test_point);
+  CAPTURE(time);
+  const auto jacobian = map.jacobian(test_point, time, functions_of_time);
+  const auto inv_jacobian =
+      map.inv_jacobian(test_point, time, functions_of_time);
+
+  const auto expected_identity = [&jacobian, &inv_jacobian]() {
+    auto identity =
+        make_with_value<tnsr::Ij<DataVector, Map::dim, Frame::NoFrame>>(
+            get_size(jacobian.get(0, 0)), 0.0);
```
wthrowe: `.size()`

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
```diff
@@ -220,6 +282,83 @@ void test_inv_jacobian(
...
+    for (size_t i = 0; i < Map::dim; ++i) {
+      for (size_t j = 0; j < Map::dim; ++j) {
+        for (size_t l = 0; l < jacobian.get(0, 0).size(); l++) {
+          for (size_t k = 0; k < Map::dim; ++k) {
+            get_element(identity.get(i, j), l) +=
```
wthrowe: `gsl::at`

wthrowe (commented):
Missed a few, but otherwise looks good.  This can be squashed (including these last changes) as far as I am concerned, but wait for the go-ahead from @knelli2.

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
AlexCarpenter46: I'm getting an error when I try to add this saying `assignment of read only location`.

wthrowe: What compiler are you using?  Seems OK with clang 16.

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
knelli2: Isn't the datatype of the jacobian here a  `DataVector`, not `std::array`? So I think `[]` is best if you know that it's `DataVector`. (I rarely see `gsl::at` or `.at` with a `DataVector`)

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
knelli2: For reasons stated in my other comment, maybe try `identity.get(i, j)[k]` since it'll be a `DataVector`. Hopefully that'll fix your error?

src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
```diff
@@ -156,19 +144,39 @@ std::optional<std::array<double, Dim>> Translation<Dim>::inverse(
       c -= square(outer_radius_.value());
       auto roots = real_roots(a, b, c);
       double radial_falloff_factor = 0.;
-      if (roots.value()[0] >= 0.0 and roots.value()[0] <= 1.0) {
-        ASSERT(roots.value()[1] < 0.0 or roots.value()[1] > 1.0,
+      ASSERT(roots.value()[0] >= 0 and roots.value()[1] >= 0,
```
knelli2: Maybe also add an `ASSERT(roots.has_value(), ...)` as an extra precaution?

knelli2 (changes requested):
Ok I'm happy with all the changes now. You can squash everything together now into one commit, including these last changes.

knelli2:
@AlexCarpenter46 Can you rebase this on the latest develop? A new CI container just got pushed

knelli2 (approved):
Looks good! @wthrowe did you want to take another look? I think we can ignore the clang-tidy errors for this.

wthrowe (approved):
clang-tidy is all warnings we plan to disable.
gcc-11 is presumably running out of memory, although the logs are screwed up so I can't verify that.

PR #5389, opened 2023-08-25
Add CcsnCollapse as Gh executable

pajkosmi:
## Proposed changes
Allow CcsnCollapse to run for the Gh system.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


pajkosmi:
Nils, b/c a new template parameter was added to `struct DerivCompute`, I had to add a Meshtag argument to `::Tags::DerivCompute<>` in the *.hpp files in different systems like `EvolveCurvedScalarWave`, `EvolveGhBinaryBlackHole`, etc. Is there a better way I can adjust this, so the CcsnCollapse problem doesn't affect other systems, or is this technical debt small enough?

nilsdeppe (commented):
Nope, this is a necessary change because we assume we would only ever have 1 numerical scheme/1 mesh 🙃 We've had a number of sweeping changes like this, and it's just inevitable cleanup/improvements.

The code LGTM, I will ask you to spilt it into a few commits (we can jump on a hangout and go over git surgery if you want, just Slack me). I suggest:
1. Add mesh tag template parameter to deriv compute. This will include all the executable changes where the tag gets added.
2. Add dt_lapse, dt_shift, and dt_spatial_metric to CCSN collapse.
3. Add CcsnCollapse GH executable (this is just the CMake file and the fwd decls)

knelli2:
I think @nilsdeppe's splits for the commit are good. One question for @pajkosmi, for the CCSN physical problem, are you ever going to need to extract waveforms? If you will, then you're changes now are good, but if you won't then I'd remove the `BondiSachs` argument when adding the CCSN executable in the cmake file.

nilsdeppe:
One of the big reasons to use spectre for CCSN is to get GW waveforms :)

knelli2:
Ok! Then definitely leave the `BondiSachs` in

pajkosmi:
@nilsdeppe I have split the commits.

nilsdeppe (approved):

knelli2 (approved):

PR #5394, opened 2023-08-27
Update Bond Sachs interpolation to run on GhMhd executables

### LINKED ISSUE #5391: BondiSachs is broken in GhValencia executables
@isaaclegred found that if you tried to run the `BondiSachsInterpolation` Event in any of the GhValencia executables, that you wouldn't get any output. We traced this issue back to the actual event that's added in the `GhValenciaDivCleanBase` is incorrect. Currently, it uses the `interp::Events::Interpolate` event which is really only used for horizon finds. What actually should be used is `interp::Events::InterpolateWithoutInterpComponent`.

However, `interp::Events::InterpolateWithoutInterpComponent` needs the metavars as a template, and we found that the derived metavars that are passed in wasn't a complete type, so the compiler complained. The only reason it's templated on the metavars is because a tag needs the metavars. But there's a base tag for that other tag so we can use that instead and then remove the metavars template.

The last thing to figure out is how this will (if at all) impact the `WithHorizon` GhValencia executables because they *do* need the `interp::Events::Interpolate` event.

isaaclegred:
## Proposed changes

This PR modifies  the interpolation done to compute Bondi-Sachs worldtubes in the GhMhd executable to be `InterpolateWithoutInterpComponent`.  This allows the `BondiSachsInterpolation` to run (and therefore extract waveforms from the GhMhd executables).    This PR, if it works, would close #5391.
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
+++ b/src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
@@ -163,6 +163,7 @@
 #include "ParallelAlgorithms/Interpolation/Actions/TryToInterpolate.hpp"
 #include "ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp"
 #include "ParallelAlgorithms/Interpolation/Events/Interpolate.hpp"
+#include "ParallelAlgorithms/Interpolation/Events/InterpolateWithoutInterpComponent.hpp"
 #include "ParallelAlgorithms/Interpolation/InterpolationTarget.hpp"
 #include "ParallelAlgorithms/Interpolation/Interpolator.hpp"
 #include "ParallelAlgorithms/Interpolation/Protocols/InterpolationTargetTag.hpp"
@@ -635,8 +636,9 @@ struct GhValenciaDivCleanTemplateBase<
                                           non_tensor_compute_tags>,
                 Events::time_events<system>,
                 control_system::control_system_events<control_systems>,
-                intrp::Events::Interpolate<3, InterpolationTargetTags,
-                                           interpolator_source_vars>...>>>,
+                intrp::Events::InterpolateWithoutInterpComponent<
+                    volume_dim, InterpolationTargetTags,
+                    interpolator_source_vars>...>>>,
         tmpl::pair<
             grmhd::GhValenciaDivClean::BoundaryConditions::BoundaryCondition,
             boundary_conditions>,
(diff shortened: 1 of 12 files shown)


knelli2:
There's probably several places in the tests where this needs to change. I'd suggest building locally to find all of them before pushing

knelli2:
The event changes look good. I think you need to rebase to fix a conflict. Also, several of your files have formatting changes that are unrelated to the event changes. Could you factor those out into a different commit before the event changes? You can also squash your fixups

isaaclegred:
Hm by factoring out the formatting changes do you mean literally introduce a commit which is just running `clang-format` over the existing files @knelli2?

knelli2:
@isaaclegred Yup, exactly!

isaaclegred:
Alright squashed and factored, some of the formatting changes didn't really respond the way I thought they would (like `EvolveCurvedScalarWave`), but I guess they are the way they are.

knelli2 (dismissed):

knelli2:
Looks like you missed the `EvolveScalarTensorSingleBlackHole` executable. Wonder if that was added while you rebased.

isaaclegred:
Indeed it was, sneaky sneaky

isaaclegred:
@knelli2 whenever you get a chance :)

knelli2 (approved):
@nilsvu Did you want to look at this or should I just merge?

src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
```diff
@@ -635,8 +636,9 @@ struct GhValenciaDivCleanTemplateBase<
                                           non_tensor_compute_tags>,
                 Events::time_events<system>,
                 control_system::control_system_events<control_systems>,
-                intrp::Events::Interpolate<3, InterpolationTargetTags,
-                                           interpolator_source_vars>...>>>,
+                intrp::Events::InterpolateWithoutInterpComponent<
+                    volume_dim, InterpolationTargetTags,
+                    interpolator_source_vars>...>>>,
```
nilsvu, line 641: I think you can remove at least the `intrp::Interpolator` component below in the `component_list`, maybe more things (@knelli2?). Looks like it can be removed in `EvolveValenciaDivClean` as well (maybe also in other executables, haven't checked).

nilsvu (changes requested):

src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
knelli2: It can't be removed in `GhValencia` because the "with horizon" executables need it for horizon finding (maybe we could add another template bool to include it or not). It could probably be removed from `Valencia` because I don't think it makes sense to find a horizon in a static spacetime.

Regardless, let's do those in a followup so @isaaclegred can actually get waveforms from his BNS runs now.

nilsvu (approved):

PR #5402, opened 2023-08-29
Fix race condition inside PiecewisePolynomial

knelli2:
## Proposed changes

A code comment inside the `PiecewisePolynomial::update` function explains how this occurred before. Also update QuaternionFunctionOfTime to fix the same issue. This became an issue once #5240 was merged, and was missed in 
#5198 and #5334.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
+++ b/src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
@@ -45,17 +45,6 @@ void StoredInfo<MaxDerivPlusOne, false>::pup(PUP::er& p) {
   p | stored_quantities;
 }
 
-void reset_expiration_time(const gsl::not_null<double*> prev_expiration_time,
-                           const double next_expiration_time) {
-  if (next_expiration_time < *prev_expiration_time) {
-    ERROR("Attempted to change expiration time to "
-          << next_expiration_time
-          << ", which precedes the previous expiration time of "
-          << *prev_expiration_time << ".");
-  }
-  *prev_expiration_time = next_expiration_time;
-}
-
 template <size_t MaxDerivPlusOne, bool StoreCoefs>
 const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
     const double t,
--- a/src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
+++ b/src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
@@ -35,6 +35,26 @@ PiecewisePolynomial<MaxDeriv>::PiecewisePolynomial(
 
 template <size_t MaxDeriv>
 PiecewisePolynomial<MaxDeriv>::PiecewisePolynomial(
+    PiecewisePolynomial<MaxDeriv>&& rhs) {
+  *this = std::move(rhs);
+}
+
+template <size_t MaxDeriv>
+PiecewisePolynomial<MaxDeriv>& PiecewisePolynomial<MaxDeriv>::operator=(
+    PiecewisePolynomial<MaxDeriv>&& rhs) {
+  if (this == &rhs) {
+    return *this;
+  }
+  deriv_info_at_update_times_ = std::move(rhs.deriv_info_at_update_times_);
+  expiration_time_ = std::move(rhs.expiration_time_);
+  deriv_info_size_.store(
+      rhs.deriv_info_size_.exchange(0, std::memory_order_acq_rel),
+      std::memory_order_release);
+  return *this;
+}
+
+template <size_t MaxDeriv>
+PiecewisePolynomial<MaxDeriv>::PiecewisePolynomial(
     const PiecewisePolynomial<MaxDeriv>& rhs) {
   *this = rhs;
 }
--- a/src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
+++ b/src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
@@ -9,6 +9,7 @@
 #include <iterator>
 #include <list>
 #include <memory>
+#include <mutex>
 #include <ostream>
 #include <pup.h>
 #include <pup_stl.h>
@@ -31,6 +32,7 @@ PiecewisePolynomial<MaxDeriv>::PiecewisePolynomial(
     : deriv_info_at_update_times_{{t, std::move(initial_func_and_derivs)}},
       expiration_time_(expiration_time) {
   deriv_info_size_.store(1, std::memory_order_release);
+  expiration_time_.store(expiration_time, std::memory_order_release);
 }
 
 template <size_t MaxDeriv>
@@ -46,7 +48,9 @@ PiecewisePolynomial<MaxDeriv>& PiecewisePolynomial<MaxDeriv>::operator=(
     return *this;
   }
   deriv_info_at_update_times_ = std::move(rhs.deriv_info_at_update_times_);
-  expiration_time_ = std::move(rhs.expiration_time_);
+  expiration_time_.store(
+      rhs.expiration_time_.exchange(0, std::memory_order_acq_rel),
+      std::memory_order_release);
   deriv_info_size_.store(
       rhs.deriv_info_size_.exchange(0, std::memory_order_acq_rel),
       std::memory_order_release);
@@ -67,7 +71,8 @@ PiecewisePolynomial<MaxDeriv>& PiecewisePolynomial<MaxDeriv>::operator=(
     return *this;
   }
   deriv_info_at_update_times_ = rhs.deriv_info_at_update_times_;
-  expiration_time_ = rhs.expiration_time_;
+  expiration_time_.store(rhs.expiration_time_.load(std::memory_order_acquire),
+                         std::memory_order_release);
   deriv_info_size_.store(rhs.deriv_info_size_.load(std::memory_order_acquire),
                          std::memory_order_release);
   return *this;
@@ -82,8 +87,10 @@ std::unique_ptr<FunctionOfTime> PiecewisePolynomial<MaxDeriv>::get_clone()
 template <size_t MaxDeriv>
 template <size_t MaxDerivReturned>
 std::array<DataVector, MaxDerivReturned + 1>
-PiecewisePolynomial<MaxDeriv>::func_and_derivs(const double t) const {
-  if (t > expiration_time_) {
+PiecewisePolynomial<MaxDeriv>::func_and_derivs(
+    const double t, const bool check_expiration_time) const {
+  if (check_expiration_time and
+      t > expiration_time_.load(std::memory_order_acquire)) {
     ERROR("Attempt to evaluate PiecewisePolynomial at a time "
           << t << " that is after the expiration time " << expiration_time_
           << ". The difference between times is " << t - expiration_time_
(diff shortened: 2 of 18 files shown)


src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -47,7 +47,7 @@ PiecewisePolynomial<MaxDeriv>& PiecewisePolynomial<MaxDeriv>::operator=(
     return *this;
   }
   deriv_info_at_update_times_ = rhs.deriv_info_at_update_times_;
-  expiration_time_ = rhs.expiration_time_;
+  expiration_time_.store(rhs.expiration_time_.load(std::memory_order_relaxed));
```
wthrowe: It looks a bit weird to be specifying memory orders for the loads but not the stores.  Is this intentional?  (Happens in a bunch of places.)

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -213,7 +236,16 @@ void PiecewisePolynomial<MaxDeriv>::pup(PUP::er& p) {
...
       // However, for v2+, we store expiration time, size, then data for
       // thread-safety reasons
-      p | expiration_time_;
+
+      // For v3+ the expiration time is atomic. While for 2 it isn't
+      if (version >= 3) {
+        double loaded_expiration_time = 0.0;
+        p | loaded_expiration_time;
+        expiration_time_.store(loaded_expiration_time,
+                               std::memory_order_relaxed);
```
wthrowe: Pup `operator|` is defined for `std::atomic` in `Utilities/Serialization/PupStlCpp11.hpp`.

It's also compatible with the serialization of the non-atomic contained type, so you don't need to bump the version.

wthrowe (changes requested):

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -112,13 +114,15 @@ void PiecewisePolynomial<MaxDeriv>::update(
           << ", which precedes the previous update time of "
           << deriv_info_at_update_times_.back().time << ".");
   }
-  if (next_expiration_time < expiration_time_) {
+  const double current_expiration_time =
```
nilsdeppe: move up to line ~66 to use it in another `if`?

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
nilsdeppe: I believe since we don't expect to copy the classes, any ordering should be fine. That's not to disagree with your request of consistency. The default is the strongest, but should be okay unless we copy this class a lot (I hope not)

src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
```diff
@@ -45,15 +46,18 @@ void StoredInfo<MaxDerivPlusOne, false>::pup(PUP::er& p) {
   p | stored_quantities;
 }
 
-void reset_expiration_time(const gsl::not_null<double*> prev_expiration_time,
-                           const double next_expiration_time) {
-  if (next_expiration_time < *prev_expiration_time) {
+void reset_expiration_time(
+    const gsl::not_null<std::atomic<double>*> prev_expiration_time,
+    const double next_expiration_time) {
+  const double loaded_prev_expr_time =
```
nilsdeppe: This all isn't safe because the atomic can get change after the `if` check. You want a compare and swap: https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -157,8 +158,30 @@ void PiecewisePolynomial<MaxDeriv>::update(
...
+  // old values at a time past the old expiration time on one thread, but new
+  // values at a time past the old expiration time on a different thread. Thus
+  // we must update the expiration time last.
+  //
+  // We also want to ensure that we are accessing only valid deriv infos. Thus
+  // we need to update the valid size *after* we emplace the updated deriv info
+  // to ensure everything is consistent. This gives us the ordering above.
   deriv_info_at_update_times_.emplace_back(time_of_update, std::move(func));
   deriv_info_size_.fetch_add(1, std::memory_order_acq_rel);
+  expiration_time_.store(next_expiration_time, std::memory_order_release);
```
nilsdeppe: To be extra safe that there isn't a race condition, you can do a https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange I _think_ there are still edge cases where there would be a race condition, but it at least adds another check.

nilsdeppe (changes requested):

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
wthrowe: Line 66 is in a different function.

src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
wthrowe: It's safe because we don't allow concurrent writing.  A lot of things will break if we allow that.

wthrowe:
Unrelated to this change, these classes need move constructors and assignment operators defined because `std::atomic` isn't movable.  Do you want to do that while you're touching everything else or file an issue?

src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
nilsdeppe: Given the number of issues we've had with these sorts of assumptions, I don't think it's worth the risk of not making the code threadsafe by switching to a compare and swap.

wthrowe: Fixing it here is reasonable, but the other case is in a function that is intentionally not threadsafe for performance reasons.

nilsdeppe: Compare and swap mostly is an issue when scaling in a highly contended environment. If we were updating the FoTs ~1 per RHS evaluation per element it _might_ matter, but given the number of bugs this file alone has produced, I still think it's better to do the check and actually have some way of knowing we hit a threading issue than not.

I also don't remember discussing making the function not threadsafe for performance reasons, I thought it was because of the underlying physics/design of the control system as a whole. I remember not wanting to use locks because of performance reasons, but that's separate for a CAS on write.

Edit: One option is to do a CAS inside `#ifdef SPECTRE_DEBUG`. Then when we get mystery issues we can at least run in release but with checks enabled to try and catch these things while the run isn't too slow.

wthrowe: You're right, I was conflating the two things.  We could lock the function to make it threadsafe without a significant performance penalty.

Edit: The _other_ function.  No locks needed here, obviously.

knelli2: So I don't think this function specifically is actually used at all. From a quick search, it's only actually used in tests.

nilsdeppe: Two separate things:

Should we just remove this function? If not, we can keep it and just do the CAS since this function sort of acts like that's what it wants to do. I don't have a preference since I don't know enough about why this was added in the first place.

----

@wthrowe a few thoughts on this:
1. What do you think about adding a `SPECTRE_DEBUG_THREADS` macro that we can enable to enable various threading-related checks? We might even want to have a dox page/section listing some of the possible checks to do inside this macro guard that help with detecting threading issues.
2. Assuming 1. (or using `SPECTRE_DEBUG`) we could add a lock only in the threading-debug context, then do a `try_lock` on it and `ERROR` if the lock can't be obtained. This would tell us we are somehow violating an assumption we've baked into the code. I worry we (I?) will break this subtly with nodegroup elements by accident, so having a sanity check could be great.

wthrowe: Let's just do it unconditionally.  Managing a lock with no contention has to be a lot cheaper than the multiple memory allocations and several O(total number of updates)-cost operations already done by that function.

Annoyingly, `try_lock` is allowed to just randomly return false because it feels like it.  I imagine that in practice it only happens when multiple threads are manipulating the lock in close succession, but we don't have a guarantee.  If we read the current expiration time at the start before locking and then do a CAS at the end I think that catches everything we can.

(Don't care if we keep this function.)

knelli2: I think I'll remove this function because I don't see a use for it. Presumably this was added very early on when we didn't know how to deal with expiration times, so being able to change only the expiration time (and not the derivative) seemed like a good idea at the time. However, now we know (more or less) how everything is going to work, and it doesn't seem like we'll need this function.

Now for the `update` function, I'll get the current expiration time at the beginning of the function and then do a CAS at the end. This seems like the best we can do? Anything else I should add?

nilsdeppe: Sounds good. Yea, it's probably some remnant of that.

I just remembered that I _think_ the mutable cache locks all FoTs during updates of any, right? In that case I guess we don't have to worry about locking the individual FoTs, right? The only extra safeguard if we want (I think it would be good) is the CAS at the end, just to be sure nothing weird has happened. @wthrowe what's your thoughts on this if the mutable cache does locking?

wthrowe: Let's make it thread-safe.  It's, I think, three lines and maybe a comment, and maybe we'll remove the lock in the cache someday to support something that can handle simultaneous writes.

So do the CAS stuff discussed above and then just stick a lock_guard at the start of the function.

nilsdeppe: That sounds great, thank you!

knelli2:
@nilsdeppe @wthrowe Made changes.

Added two fixups for the original two commits. For the consistency of `std::memory_order` I chose to just always use `acquire/release` for `load/store` no matter what function I was in. Then for other operations that require an acquire and release, I use `acq_rel`. I don't use `relaxed` ever now (if I forgot one let me know and I'll fix it). I think I did the CAS change correctly, but I'm not 100% sure.

I also added two new commits that remove the `reset_expiration_time` function, and also add move constructor/assignment to PiecewisePolynomial and QuaternionFunctionOfTime.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -206,13 +208,14 @@ void PiecewisePolynomial<MaxDeriv>::update(
   // to ensure everything is consistent. This gives us the ordering above.
   deriv_info_at_update_times_.emplace_back(time_of_update, std::move(func));
   deriv_info_size_.fetch_add(1, std::memory_order_acq_rel);
-  expiration_time_.store(next_expiration_time, std::memory_order_release);
+  expiration_time_.compare_exchange_strong(
```
nilsdeppe: Yep, then check that the result is `true`:
```cpp
if (not expiration_time_.compare_exchange_strong(
    current_expiration_time, next_expiration_time, std::memory_order_acq_rel)) {
  ERROR("...");
}
```

nilsdeppe (commented):
One small suggestion, but this addresses my concerns. I think the move operations are fine, but I haven't thought about moving atomics too deeply, other than I don't really know what it means to do that, hence why they're deleted :P

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
knelli2: Squashed into the current fixup for @wthrowe to view.

nilsdeppe:
@wthrowe could you please look at the fixups on this?

wthrowe (dismissed):
Looks good.  Squash.

wthrowe (approved):

nilsdeppe (approved):

PR #5403, opened 2023-08-29
Move Basis, Quadrature, and Mesh from Spectral to SpatialDiscretization

kidder:
## Proposed changes

This will decouple things that only depend upon a Mesh from the Spectral library and contents of Spectral.hpp

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
Why are you making this change?

kidder:
see proposed changes above...
is there a reason not to?

nilsvu:
Seems fine, it's just a very low-level change with a big impact (lots of rebase conflicts). I don't love factoring Mesh into basically its very own library, with a somewhat long and unwieldy namespace name (this PR adds ~1k lines, presumably mostly because the namespace name is long). I agree that Spectral has grown recently with the addition of FD things so we may have to restructure it a bit.

If you just want to avoid including all of Spectral.hpp with Mesh (good idea!), then I suggest you just move `Basis` and `Quadrature` into their own headers. That would a very uncontroversial change with a very small diff and no rebase conflicts. Happy to discuss further restructurings of Spectral (maybe rename the whole thing to SpatialDiscretization or something shorter so it can include FD things? not sure).

kidder:
- I'm fine with shortening `SpatialDiscretization` if we can agree on it
- Decoupling libraries leads to shorter build times
- Plan to also move all the swsh stuff to its own library

nilsvu:
My suggestion is to just split Basis and Quadrature out of Spectral.hpp into their own files.

> Decoupling libraries leads to shorter build times

I don't think that applies here because all executables depend on Spectral anyway, so it is built regardless.

kidder:
re-implemented in #5421

PR #5404, opened 2023-08-29
Optionally write Bns control sys centers to disk

knelli2:
## Proposed changes

We already write center measurements to disk for BBHs, so we should also do it for the BNS. It can help with debugging control system errors by ruling out bad center measurements.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Measurements/BNSCenterOfMass.hpp
+++ b/src/ControlSystem/Measurements/BNSCenterOfMass.hpp
@@ -5,6 +5,8 @@
 
 #include <cstddef>
 #include <string>
+#include <tuple>
+#include <vector>
 
 #include "ControlSystem/Component.hpp"
 #include "ControlSystem/Protocols/Measurement.hpp"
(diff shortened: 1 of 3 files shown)


ffoucart (commented):
Sounds good to me. A couple of questions that are likely just related to my lack of knowledge of the way this code works:
* Are we guaranteed that `PostReductionSendBNSStarCentersToControlSystem` is called on node 0?
* Do we need to add `control_system::Tags::WriteDataToDisk` in the global cache of all executables that use `BothBNSCenters`?

knelli2:
- No, `PostReductionSendBNSStarCentersToControlSystem` will be called on whatever component you told it to be run on in the `Parallel::contribute_to_reduction` call. In our case, (new) line 160 of BNSCenterOfMass.hpp. However, what I think you're referring to is the `writer_proxy[0]` part? This means we're going to send the data that is to be written to the 0th node of the `ObserverWriter` parallel component, which is always guaranteed to exist (we always have at least 1 node).
- Technically, yes. However, `control_system::Tags::WriteDataToDisk` is actually needed for all control systems regardless, so any executable that has a control system parallel component in it will automatically have this tag in the cache. This is why I didn't worry about adding it anywhere in this PR.

ffoucart (approved):

wthrowe:
> * Technically, yes. However, `control_system::Tags::WriteDataToDisk` is actually needed for all control systems regardless, so any executable that has a control system parallel component in it will automatically have this tag in the cache. This is why I didn't worry about adding it anywhere in this PR.

To be explicit, it's added by `control_system::Actions::Initialize` (which is actually an initialization mutator, not an action), so it's safe to assume it's there.

wthrowe (approved):

knelli2:
@wthrowe Any idea why the gcc-11 Release build failed? It looks like the BBH executable failed to build. I usually see that type of build error when it runs out of memory.

wthrowe:
Yeah, that looks like OOM.  Nothing in this PR should have affected that executable, though.  Try it again and if it works write it off as some weird GitHub glitch?

nilsdeppe:
I restarted the failed GCC-11 job.

wthrowe:
Same failure in #5400.

PR #5395, opened 2023-08-28
Encourage formaline CI test to reuse compiler cache

nilsvu:
## Proposed changes

Make sure to use the same compiler flags as the full build.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5408, opened 2023-08-30
Remove some unnecessary global CMake include dirs

nilsvu:
## Proposed changes

Found this while debugging #5395. Both PRs invalidate CI caches, so merging them around the same time would minimize CI downtime.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


CMakeLists.txt
```diff
@@ -182,14 +182,11 @@ endif()
```
nilsdeppe, line 181: Not a change request, but for my own knowledge. Do you know if the recommendation nowadays is to handle include dirs through targets? E.g. should we have a `SpectreSrcInclude` target that we use? Similar for tests? Or should the `spectre_add_library`-type calls set the include dir for each target instead of setting global ones?

nilsdeppe (dismissed):

CMakeLists.txt
nilsvu: I looked at that, but haven't figured out a great solution. We want to include headers with the full relative path to `src/`, which means the include dir is the same for all libraries. I haven't found a best-practice example for using `target_include_directories` effectively for this, so I kept the global `include_directories` for now (but limited the tests include dir to just the tests).

nilsdeppe: Ah right... Alright, sounds good to me :) thanks!

nilsdeppe (approved):

PR #5413, opened 2023-09-01
Link libxsmm only in Release mode on ARM Macs

nilsvu:
## Proposed changes

libxsmm is currently disabled in Debug mode, but still linked. On my ARM Mac, libxsmm triggers SIGILL signals that makes debugging with lldb difficult. Just not linking libxsmm fixes this issue, and doesn't change any functionality because libxsmm is disabled in Debug mode anyway.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Utilities/CMakeLists.txt
+++ b/src/Utilities/CMakeLists.txt
@@ -88,8 +88,20 @@ target_link_libraries(
   Boost::boost
   Brigand
   ErrorHandling
-  Libxsmm
   )
+
+# Libxsmm is disabled in Debug mode (see Blas.hpp).
+# As of Sep 1, 2023 libxsmm has an issue with lldb on ARM Macs (detecting
+# the CPU architecture raises SIGILL signals when loading the libxsmm dynamic
+# library). Just not linking libxsmm in Debug mode works around that.
+if (NOT ENABLE_SPECTRE_DEBUG
+    OR NOT (APPLE AND CMAKE_HOST_SYSTEM_PROCESSOR STREQUAL "arm64"))
+  target_link_libraries(
+    ${LIBRARY}
+    PUBLIC
+    Libxsmm
+    )
+endif()
 
 add_subdirectory(ErrorHandling)
 add_subdirectory(System)


src/Utilities/CMakeLists.txt
```diff
@@ -88,9 +88,16 @@ target_link_libraries(
   Boost::boost
   Brigand
   ErrorHandling
-  Libxsmm
   )
 
+if (NOT ENABLE_SPECTRE_DEBUG)
```
knelli2: Maybe add a small comment here saying "libxsmm is disabled in debug mode, so only link it in release" just so future people know why this was done

knelli2 (changes requested):

src/Utilities/CMakeLists.txt
nilsvu: done

knelli2 (dismissed):

nilsdeppe (changes requested):
This sounds like it's an ARM problem, so why not only do it on ARM macs? There have been bugs in libxsmm that were much easier to track down when building in debug mode..

nilsvu:
Sure I can just do it on ARM only. Note though that libxsmm is currently linked into Debug builds but not used, which seems wrong. If you want to debug something with libxsmm you have to delete some `#ifndef SPECTRE_DEBUG` in Blas.hpp. At that point it wouldn't be too much to ask to also delete the `if` in the CMakeLists.txt I think?

nilsdeppe:
I just don't see the benefit of removing it in the CMake. I don't remember why it's disable in debug mode at all, actually

knelli2:
#790

nilsdeppe:
Okay, well, I still think if the issue is with a particular arch & OS we should limit things to that :Shrug:

nilsvu:
So you prefer to link in an unused library? :D

nilsdeppe:
Yes.

nilsvu:
Fine, pushed an updated commit

nilsdeppe (approved):

knelli2 (approved):
The changes themselves are ok, but I'm still a bit confused why we want to link in an unused library? However, we can have that discussion later. We should get this in.

PR #5415, opened 2023-09-03
Add a few more labels to release notes

nilsvu:
## Proposed changes

Cleans up the long list of PRs in release notes a bit, so it is easier to read.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/CompileReleaseNotes.py
+++ b/tools/CompileReleaseNotes.py
@@ -60,8 +60,20 @@ class PullRequest:
     upgrade_instructions: Optional[str] = None
 
 
-# These should correspond to labels on GitHub
-PULL_REQUEST_GROUPS = ["new feature", None, "bugfix"]
+# The groups into which pull-requests are sorted.
+# - The keys in this mapping should correspond to labels on GitHub. The values
+#   are the group headers.
+# - The order of the groups is the order in which they will appear in the
+#   release notes (dicts are ordered in Python 3.7+).
+PULL_REQUEST_GROUPS = {
+    "new feature": "New features",
+    "cli/pybindings": "CLI & Python bindings",
+    None: "General changes",
+    "documentation": "Documentation",
+    "bugfix": "Bugfixes",
+    "build system": "Build system",
+    "ci/cd": "Continuous integration & deployment",
+}
 
 
 def get_merged_pull_requests(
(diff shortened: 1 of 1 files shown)


knelli2:
While you're doing this, could you also add a `Python` label? If the goal of this is to organize release notes, and we want to make the CLI/py-bindings a very usable public interface, seems like they should get their own chunk of the release notes so users can track changes.

nilsvu:
Ok done.

tools/CompileReleaseNotes.py
```diff
@@ -182,7 +195,10 @@ def format_list_of_prs(prs):
         for group, prs_iterator in grouped_prs:
             group_header = {
                 "new feature": "New features",
+                "documentation": "Documentation",
                 "bugfix": "Bugfixes",
+                "build system": "Build system",
+                "ci/cd": "CI/CD",
                 None: "General changes",
```
knelli2: Do we need to add `CLI/Pybindings` here?

tools/CompileReleaseNotes.py
```diff
@@ -329,9 +345,11 @@ def format_list_of_prs(prs):
             pr.author = pr_gh.user.login
             # Add missing metadata to PR
             pr.url = pr_gh.html_url
-            # Add group information to PR
+            # Add group information to PR. Iterate over groups in reverse order
+            # so that PRs with both, say, "bugfix" and "cli/pybindings" labels
+            # are listed under "bugfix" rather than "cli/pybindings".
```
knelli2, line 343: Would it be easy to just include it twice? Once under `bugfix` and once under `cli/pybindings`? (and more generally for many groups). That way we don't have to assign a kind of priority to a label? If not then don't worry about it

tools/CompileReleaseNotes.py
nilsvu: Oh I just forgot to replace this dict with the one above, thanks for catching.

tools/CompileReleaseNotes.py
nilsvu: My intention is to move things like bugfixes to the bottom of the list because you don't really care about those when reading the changelog. This should declutter the list so you can more easily spot the changes that may be breaking your code.

knelli2:
Swapping the dictionary seems to have broken a formatting test

knelli2 (approved):

nilsdeppe (approved):

PR #5416, opened 2023-09-03
Write and read surface spherical harmonic data to and from disk

### LINKED ISSUE #5183: Read and write Strahlkorper coefficients from disk
We need to load Strahlkorper YLM coefficients from a file to initialize the shape of our excision surfaces, both for high-spin inspirals and for ringdowns.

macedo22:
## Proposed changes
Fixes #5183

This PR implements writing and reading surface spherical harmonic data to and from disk. The writing currently assumes that `l_max` for a surface is constant but was implemented in a way to make it easy to change in the future if and when `l_max` can change over the course of the simulation. The reading, however, is already generalized to handle reading in rows of data with different `l_max` values.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/SphericalHarmonics/IO/CMakeLists.txt
+++ b/src/NumericalAlgorithms/SphericalHarmonics/IO/CMakeLists.txt
@@ -1,0 +1,30 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+set(LIBRARY SphericalHarmonicsIO)
+
+add_spectre_library(${LIBRARY})
+
+spectre_target_sources(
+  ${LIBRARY}
+  PRIVATE
+  ReadSurfaceYlm.cpp
+  )
+
+spectre_target_headers(
+  ${LIBRARY}
+  INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/src
+  HEADERS
+  ReadSurfaceYlm.hpp
+  )
+
+target_link_libraries(
+  ${LIBRARY}
+  PRIVATE
+  DataStructures
+  ErrorHandling
+  H5
+  Utilities
+  PUBLIC
+  SphericalHarmonics
+  )
--- a/src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
@@ -1,0 +1,213 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp"
+
+#include <array>
+#include <cmath>
+#include <cstddef>
+#include <numeric>
+#include <string>
+#include <vector>
+
+#include "DataStructures/Matrix.hpp"
+#include "DataStructures/ModalVector.hpp"
+#include "DataStructures/Tensor/IndexType.hpp"
+#include "IO/H5/AccessType.hpp"
+#include "IO/H5/Dat.hpp"
+#include "IO/H5/File.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/SpherepackIterator.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/GetOutput.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeString.hpp"
+
+namespace ylm {
+namespace {
+const size_t num_non_coef_headers = 5;
+
+// check that the headers of the legend are what we expect them to be
+template <typename Frame>
+void check_legend(const std::vector<std::string>& legend) {
+  const std::string expected_frame{get_output(Frame{})};
+
+  // check format and ordering of non-coefficient headers
+  const auto check_header = [&legend](const size_t column,
+                                      const std::string& expected_header) {
+    const std::string& read_header = gsl::at(legend, column);
+    if (read_header != expected_header) {
+      ERROR("In column " << column << " of the Ylm legend, expected header "
+                         << expected_header << " but got header " << read_header
+                         << ".");
+    }
+  };
+  check_header(0, "Time");
+  check_header(1, expected_frame + "ExpansionCenter_x");
+  check_header(2, expected_frame + "ExpansionCenter_y");
+  check_header(3, expected_frame + "ExpansionCenter_z");
+  check_header(4, "Lmax");
+
+  // check format and ordering of coefficient headers
+  int expected_l = 0;
+  int expected_m = 0;
+  size_t coef_column_number = num_non_coef_headers;
+  while (coef_column_number < legend.size()) {
+    const std::string& coefficient_header = legend[coef_column_number];
+    const std::string expected_coefficient_header = {
+        MakeString{} << "coef(" << std::to_string(expected_l) << ","
+                     << std::to_string(expected_m) << ")"};
+    if (coefficient_header != expected_coefficient_header) {
+      ERROR(MakeString{}
+            << "In column " << coef_column_number
+            << " of the Ylm legend, expected header "
+            << expected_coefficient_header << " but got header "
+            << coefficient_header
+            << ".\n\n Coefficient headers are expected to take the form "
+               "coef(l,m) and be ordered first by ascending l and then for "
+               "each l, ordered by ascending m, i.e. coef(0,0), coef(1,-1), "
+               "coef(1,0), coef(1,1), coef(2,-2), ...");
+    }
+
+    if (expected_l == expected_m) {
+      expected_l++;
+      expected_m = -expected_l;
+    } else {
+      expected_m++;
+    }
+
+    coef_column_number++;
+  }
+}
+
+// checks if a double is a nonnegative integer
+bool is_nonnegative_int(const double n) {
+  return n == abs(n) and n == floor(n);
+}
+
+template <typename Frame>
+Strahlkorper<Frame> read_surface_ylm_row(const Matrix& ylm_data,
+                                         const size_t row_number) {
+  const size_t l_max_column_number = 4;
+  const std::array<double, 3> expansion_center{{ylm_data(row_number, 1),
+                                                ylm_data(row_number, 2),
+                                                ylm_data(row_number, 3)}};
+  const double l_max_from_file = ylm_data(row_number, l_max_column_number);
+  if (not is_nonnegative_int(l_max_from_file)) {
+    ERROR("Row " << row_number << " of the Ylm data has an invalid Lmax value ("
+                 << l_max_from_file << ") in column " << l_max_column_number
+                 << ". The value of Lmax should be a nonnegative integer.");
+  }
+
+  const auto l_max = static_cast<size_t>(l_max_from_file);
+  // l_max == m_max
+  const size_t spectral_size = Spherepack::spectral_size(l_max, l_max);
+  // We only write and store half of the coefficients. This is the minimum
+  // number of coefficients needed to describe the strahlkorper given the l_max,
+  // i.e. columns of 0.0 for higher coefficients are okay
+  const size_t min_expected_num_coefficients = spectral_size / 2;
+  const size_t min_expected_num_columns =
+      min_expected_num_coefficients + num_non_coef_headers;
+  const size_t actual_num_columns = ylm_data.columns();
+
+  if (actual_num_columns < min_expected_num_columns) {
+    ERROR("Row " << row_number
+                 << " of the Ylm data does not have enough coefficients for "
+                    "the Lmax. For Lmax = "
+                 << l_max << ", expected at least "
+                 << min_expected_num_coefficients << " coefficient columns and "
+                 << min_expected_num_columns << " total columns.");
+  }
+
+  ModalVector spectral_coefficients(spectral_size, 0.0);
+  size_t coef_column_number = num_non_coef_headers;
+  SpherepackIterator iter(l_max, l_max);
+  // read in expected coefficients for the given l_max
+  for (size_t l = 0; l <= l_max; l++) {
+    for (int m = -static_cast<int>(l); m <= static_cast<int>(l); m++) {
+      iter.set(l, m);
+
+      const double coefficient = ylm_data(row_number, coef_column_number);
+      spectral_coefficients[iter()] = coefficient;
+
+      coef_column_number++;
+    }
+  }
+  // make sure any higher order coefficients (l > l_max) present in the data
+  // that was read in are 0.0
+  while (coef_column_number < actual_num_columns) {
+    if (ylm_data(row_number, coef_column_number) != 0.0) {
+      ERROR("Row " << row_number << " of the Ylm data has Lmax = " << l_max
+                   << " but non-zero coefficients for l > Lmax.");
+    }
+    coef_column_number++;
+  }
+
+  Strahlkorper<Frame> strahlkorper(l_max, l_max, spectral_coefficients,
+                                   expansion_center);
+
+  return strahlkorper;
+}
+}  // namespace
+
+template <typename Frame>
+std::vector<Strahlkorper<Frame>> read_surface_ylm(
+    const std::string& file_name, const std::string& surface_subfile_name,
+    const size_t requested_number_of_times_from_end) {
+  ASSERT(requested_number_of_times_from_end > 0,
+         "Must request to read in at least one row (time) of Ylm data.");
+
+  h5::H5File<h5::AccessType::ReadOnly> file{file_name};
+  const std::string ylm_subfile_name{std::string{"/"} + surface_subfile_name};
+  const auto& ylm_file = file.get<h5::Dat>(ylm_subfile_name);
+
+  // number of rows available to read in
+  const size_t total_number_of_times = gsl::at(ylm_file.get_dimensions(), 0);
+  if (total_number_of_times == 0) {
+    ERROR("The Ylm data to read from contain 0 rows (times) of data.");
+  }
+
+  if (requested_number_of_times_from_end > total_number_of_times) {
+    ERROR("The requested number of time values ("
+          << requested_number_of_times_from_end
+          << ") is more than the number of rows in the Ylm data that was read "
+             "in ("
+          << total_number_of_times << ")");
+  }
+
+  const auto& ylm_legend = ylm_file.get_legend();
+  check_legend<Frame>(ylm_legend);
+
+  std::vector<size_t> columns(ylm_legend.size());
+  std::iota(std::begin(columns), std::end(columns), 0);
+  // grab all columns of the last requested_number_of_times_from_end rows
+  const auto ylm_data_subset = ylm_file.get_data_subset(
+      columns, total_number_of_times - requested_number_of_times_from_end,
+      requested_number_of_times_from_end);
+
+  std::vector<Strahlkorper<Frame>> strahlkorpers(
+      requested_number_of_times_from_end);
+  for (size_t i = 0; i < requested_number_of_times_from_end; i++) {
+    strahlkorpers[i] = read_surface_ylm_row<Frame>(ylm_data_subset, i);
+  }
+
+  file.close_current_object();
+  return strahlkorpers;
+}
+}  // namespace ylm
+
+#define FRAMETYPE(data) BOOST_PP_TUPLE_ELEM(0, data)
+
+#define INSTANTIATE(_, data)                                       \
+  template std::vector<ylm::Strahlkorper<FRAMETYPE(data)>>         \
+  ylm::read_surface_ylm<>(const std::string& file_name,            \
+                          const std::string& surface_subfile_name, \
+                          size_t requested_number_of_times_from_end);
+
+GENERATE_INSTANTIATIONS(INSTANTIATE,
+                        (Frame::Grid, Frame::Inertial, Frame::Distorted))
+
+#undef INSTANTIATE
+#undef FRAMETYPE
--- a/src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
@@ -1,0 +1,33 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <string>
+#include <vector>
+
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+
+namespace ylm {
+/// \ingroup SurfacesGroup
+/// \brief Returns a list of `ylm::Strahlkorper`s constructed from reading in
+/// spherical harmonic data for a surface at a requested list of times
+///
+/// \details The `ylm::Strahlkorper`s are constructed by reading in data from
+/// an H5 subfile that is expected to be in the format described by
+/// `intrp::callbacks::ObserveSurfaceData`. It is assumed that
+/// \f$l_{max} = m_{max}\f$.
+///
+/// \param file_name name of the H5 file containing the surface's spherical
+/// harmonic data
+/// \param surface_subfile_name name of the subfile (with no leading slash nor
+/// the `.dat` extension) within `file_name` that contains the surface's
+/// spherical harmonic data to read in
+/// \param requested_number_of_times_from_end the number of times to read in
+/// starting backwards from the final time found in `surface_subfile_name`
+template <typename Frame>
+std::vector<ylm::Strahlkorper<Frame>> read_surface_ylm(
+    const std::string& file_name, const std::string& surface_subfile_name,
+    size_t requested_number_of_times_from_end);
+}  // namespace ylm
--- a/src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
@@ -3,12 +3,16 @@
 
 #include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
 
+#include <algorithm>
 #include <cmath>
 #include <ostream>
 #include <pup.h>
 #include <pup_stl.h>
 #include <utility>
 
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/ModalVector.hpp"
+#include "DataStructures/VectorImpl.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/SpherepackIterator.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/StdArrayHelpers.hpp"
@@ -39,12 +43,26 @@ Strahlkorper<Frame>::Strahlkorper(
     : l_max_(l_max),
       m_max_(m_max),
       ylm_(l_max, m_max),
-      // clang-tidy: do not std::move trivially constructable types
-      center_(std::move(center)),  // NOLINT
+      center_(center),
       strahlkorper_coefs_(ylm_.phys_to_spec(radius_at_collocation_points)) {
   ASSERT(radius_at_collocation_points.size() == ylm_.physical_size(),
          "Bad size " << radius_at_collocation_points.size() << ", expected "
                      << ylm_.physical_size());
+}
+
+template <typename Frame>
+Strahlkorper<Frame>::Strahlkorper(const size_t l_max, const size_t m_max,
+                                  const ModalVector& spectral_coefficients,
+                                  std::array<double, 3> center)
+    : l_max_(l_max), m_max_(m_max), ylm_(l_max, m_max), center_(center) {
+  ASSERT(spectral_coefficients.size() == ylm_.spectral_size(),
+         "Bad size " << spectral_coefficients.size() << ", expected "
+                     << ylm_.spectral_size());
+
+  strahlkorper_coefs_.destructive_resize(ylm_.spectral_size());
+  // We can remove this copy once strahlkorper_coefs_ are stored as ModalVector
+  std::copy(spectral_coefficients.begin(), spectral_coefficients.end(),
+            strahlkorper_coefs_.begin());
 }
 
 template <typename Frame>
--- a/src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
@@ -7,6 +7,7 @@
 #include <cstddef>
 
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/ModalVector.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Spherepack.hpp"
 #include "Options/String.hpp"
 #include "Utilities/ForceInline.hpp"
@@ -73,9 +74,18 @@ class Strahlkorper {
   /// spectral coefficients, it is not possible to choose spectral
   /// coefficients to exactly match all points in
   /// `radius_at_collocation_points`.
-  Strahlkorper(size_t l_max, size_t m_max,
-               const DataVector& radius_at_collocation_points,
-               std::array<double, 3> center);
+  explicit Strahlkorper(size_t l_max, size_t m_max,
+                        const DataVector& radius_at_collocation_points,
+                        std::array<double, 3> center);
+
+  /// \brief Construct a Strahlkorper from a `ModalVector` containing the
+  /// spectral coefficients
+  ///
+  /// \details The spectral coefficients should be in the form defined by
+  /// `ylm::Strahlkorper::coefficients() const`.
+  explicit Strahlkorper(size_t l_max, size_t m_max,
+                        const ModalVector& spectral_coefficients,
+                        std::array<double, 3> center);
 
   /// Prolong or restrict another surface to the given `l_max` and `m_max`.
   Strahlkorper(size_t l_max, size_t m_max,
--- a/src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
@@ -3,7 +3,6 @@
 
 #pragma once
 
-#include <array>
 #include <cstddef>
 #include <string>
 #include <utility>
@@ -13,6 +12,7 @@
 #include "DataStructures/DataBox/TagName.hpp"
 #include "IO/H5/TensorData.hpp"
 #include "IO/Observer/ObserverComponent.hpp"
+#include "IO/Observer/ReductionActions.hpp"
 #include "IO/Observer/Tags.hpp"
 #include "IO/Observer/VolumeActions.hpp"
 #include "NumericalAlgorithms/Spectral/Basis.hpp"
@@ -34,10 +34,26 @@
 
 namespace intrp {
 namespace callbacks {
+namespace detail {
+// Fills the legend and row of spherical harmonic data to write to disk
+//
+// The number of coefficients to write is based on `max_l`, the maximum value
+// that the input `strahlkorper` could possibly have. When
+// `strahlkorper.l_max() < max_l`, coefficients with \f$l\f$ higher than
+// `strahlkorper.l_max()` will simply be zero. Assuming the same `max_l` is
+// always used for a given surface, we will always write the same number of
+// columns for each row, as `max_l` sets the number of columns to write
+template <typename Frame>
+void fill_ylm_legend_and_data(gsl::not_null<std::vector<std::string>*> legend,
+                              gsl::not_null<std::vector<double>*> data,
+                              const ylm::Strahlkorper<Frame>& strahlkorper,
+                              double time, size_t max_l);
+}  // namespace detail
 
-/// \brief post_interpolation_callback that outputs
-/// 2D "volume" data on a surface.
+/// \brief post_interpolation_callback that outputs 2D "volume" data on a
+/// surface and the surface's spherical harmonic data
 ///
+/// \details
 /// Uses:
 /// - Metavariables
 ///   - `temporal_id`
@@ -48,6 +64,27 @@ namespace callbacks {
 ///
 /// For requirements on InterpolationTargetTag, see
 /// intrp::protocols::InterpolationTargetTag
+///
+/// The columns of spherical harmonic data written take the form
+///
+/// \code
+/// [Time, {Frame}ExpansionCenter_x, {Frame}ExpansionCenter_y,
+/// {Frame}ExpansionCenter_z, Lmax, coef(0,0), ... coef(Lmax,Lmax)]
+/// \endcode
+///
+/// where `coef(l,m)` refers to the strahlkorper coefficients stored and defined
+/// by `ylm::Strahlkorper::coefficients() const`. It is assumed that
+/// \f$l_{max} = m_{max}\f$.
+///
+/// \note Currently, \f$l_{max}\f$ for a given surface does not change over the
+/// course of the simulation, which means that the total number of columns of
+/// coefficients that we need to write is also constant. The current
+/// implementation of writing the coefficients at one time assumes \f$l_{max}\f$
+/// of a surface remains constant. If and when in the future functionality for
+/// an adaptive \f$l_{max}\f$ is added, the implementation for writing the
+/// coefficients will need to be updated to account for this. One possible way
+/// to address this is to have a known maximum \f$l_{max}\f$ for a given surface
+/// and write all coefficients up to that maximum \f$l_{max}\f$.
 template <typename TagsToObserve, typename InterpolationTargetTag,
           typename HorizonFrame>
 struct ObserveSurfaceData
@@ -82,6 +119,7 @@ struct ObserveSurfaceData
       tensor_components.push_back(
           {"InertialCoordinates_z"s, get<2>(inertial_strahlkorper_coords)});
     }
+
     // Output each tag if it is a scalar. Otherwise, throw a compile-time
     // error. This could be generalized to handle tensors of nonzero rank by
     // looping over the components, so each component could be visualized
@@ -108,9 +146,9 @@ struct ObserveSurfaceData
         2, Spectral::Basis::SphericalHarmonic};
     const std::vector<Spectral::Quadrature> quadratures_vector{
         {Spectral::Quadrature::Gauss, Spectral::Quadrature::Equiangular}};
-    const observers::ObservationId& observation_id = observers::ObservationId(
-        InterpolationTarget_detail::get_temporal_id_value(temporal_id),
-        subfile_path + ".vol");
+    const double time =
+        InterpolationTarget_detail::get_temporal_id_value(temporal_id);
+    const observers::ObservationId observation_id{time, subfile_path + ".vol"};
 
     auto& proxy = Parallel::get_parallel_component<
         observers::ObserverWriter<Metavariables>>(cache);
@@ -123,6 +161,29 @@ struct ObserveSurfaceData
         std::vector<ElementVolumeData>{{surface_name, tensor_components,
                                         extents_vector, bases_vector,
                                         quadratures_vector}});
+
+    std::vector<std::string> ylm_legend{};
+    std::vector<double> ylm_data{};
+    // The number of coefficients written will be (l_max + 1)^2 where l_max is
+    // the current value of l_max for this surface's strahlkorper. Because l_max
+    // remains constant, the number of coefficient columns written does, too. In
+    // the future when l_max is adaptive, instead of passing in the current
+    // l_max of the strahlkorper, we could pass in the maximum value that l_max
+    // could be to ensure that we (a) have enough columns to write all the
+    // coefficients regardless of the current value of l_max and (b) write a
+    // constant number of columns for each row of data regardless of the current
+    // l_max.
+    detail::fill_ylm_legend_and_data(make_not_null(&ylm_legend),
+                                     make_not_null(&ylm_data), strahlkorper,
+                                     time, strahlkorper.l_max());
+
+    const std::string ylm_subfile_name{std::string{"/"} + surface_name +
+                                       "_Ylm"};
+
+    Parallel::threaded_action<
+        observers::ThreadedActions::WriteReductionDataRow>(
+        proxy[0], ylm_subfile_name, std::move(ylm_legend),
+        std::make_tuple(std::move(ylm_data)));
   }
 };
 }  // namespace callbacks
--- a/tests/Unit/NumericalAlgorithms/SphericalHarmonics/IO/Test_ReadSurfaceYlm.cpp
+++ b/tests/Unit/NumericalAlgorithms/SphericalHarmonics/IO/Test_ReadSurfaceYlm.cpp
@@ -1,0 +1,361 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <array>
+#include <cstddef>
+#include <limits>
+#include <random>
+#include <string>
+#include <vector>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/IndexType.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "Helpers/DataStructures/MakeWithRandomValues.hpp"
+#include "IO/H5/AccessType.hpp"
+#include "IO/H5/Dat.hpp"
+#include "IO/H5/File.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp"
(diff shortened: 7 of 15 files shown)


src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
```diff
@@ -18,6 +18,12 @@ class er;
 /// \endcond
 
 namespace ylm {
+/// \brief The kind of data points used to construct a `Strahlkorper`
+enum class StrahlkorperConstructorData {
+  RadiusAtCollocationPoints,
+  SpectralCoefficients
```
macedo22: An enum class like this for extending the constructor was requested and I initially chose to name the class and values here very explicitly to make them clear, but I will say that as-is they are verrry long to type (basically have to copy+paste). Suggestions on better (shorter) names are very welcome, e.g. maybe `StrahlkorperData::Radius` and `StrahlkorperData::Coef`?

macedo22:
Oh no, the tests failing CI all passed locally 😅 (with the exception of `clang-tidy`, which @knelli2 told me is an expected error that can be ignored). Apologies, I will have to look into why they're failing on CI and not ocean

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -32,10 +35,100 @@
...
+///
+/// \param legend the legend to fill for the strahlkorper data we want to write
+/// \param data the spherical harmonic data to fill that we want to write
+/// \param strahlkorper the strahlkorper from which to get the spherical
+/// harmonic data to fill `legend` and `data`
+/// \param time the current time
+/// \param max_l the maximum value that \f$l_max\f$ could be (sets number of
+/// columns to write)
+template <typename Frame>
+void fill_ylm_legend_and_data(
```
knelli2: `detail` functions shouldn't get doxygen documentation. If you'd like (and would probably be helpful) you can keep the `\details` paragraph as a regular code comment

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -32,10 +35,100 @@
...
+  ylm::SpherepackIterator iter(l_max, l_max);
+  for (size_t l = 0; l <= l_max; l++) {
+    for (int m = -l; m < 0; m++) {
+      legend->push_back(MakeString{} << "Im(" << l << "," << m << ")");
+
+      iter.set(l, m);
+      data->push_back(ylm_coefficients[iter()]);
+    }
+    for (int m = 0; m <= static_cast<int>(l); m++) {
+      legend->push_back(MakeString{} << "Re(" << l << "," << m << ")");
```
knelli2: Even though the $Y_{lm}$ functions themselves are complex, here we're only writing their coefficients, which are always real, so we don't need `Im` and `Re`. I don't think we have a defined format for how to write spherical harmonic column headers, but for shape control I do `l2m-2`. Maybe not the best, but we should try to be consistent. A lot of other places in this PR assume the `Im/Re` header format so they'll have to be changed as well.

You can then also combine these two inner loops into one.

Edit: So actually what I said above is incorrect. The coefficients are actually complex. However, like you have in the code, we only write the real part of the coefficient for $m\ge 0$ and the imaginary part for $m<0$ (see [Strahlkorper docs](https://spectre-code.org/classylm_1_1Strahlkorper.html#a81c2bb59da9016c4252467180fc4b481)). However, when we talk about the $l=2,m=-2$ coefficient and refer to it in the vector, it's understood that it's the imaginary part. We hardly ever work with individual components so this rarely comes up in practice. I revise my recommendation, though, and think that `coef(2,-2)` would be a good legend so it's clear that it's a coefficient.

The comment about combining the loops still stands then.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -32,10 +35,100 @@
...
+  data->emplace_back(expansion_center[2]);
+  legend->emplace_back("Lmax");
+  data->emplace_back(l_max);
+
+  const DataVector ylm_coefficients = strahlkorper.coefficients();
+  // fill coefficients for l in [0, l_max]
+  // l_max == m_max
+  ylm::SpherepackIterator iter(l_max, l_max);
+  for (size_t l = 0; l <= l_max; l++) {
+    for (int m = -l; m < 0; m++) {
```
knelli2: Inside the `for (int m = -l; m < 0; m++) {`, cast `l` to an `int` like you do below as well. I think this loop pops up a few times in this PR so make sure to get those as well.

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
knelli2: Since this is only used internally inside `Strahlkorper`, I think you can just make this an member enum called `ContructorDataType`. Then the two values could be `Radius` (or `Radii`) and `SpectralCoefs`. How does that sound?

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Test_Strahlkorper.cpp
```diff
@@ -75,6 +79,27 @@ void test_invert_spec_phys_transform() {
...
+  std::uniform_real_distribution<double> distribution(0.0, 1.0);
+  MAKE_GENERATOR(generator);
+  const auto radius = make_with_random_values<DataVector>(
+      make_not_null(&generator), distribution,
+      DataVector(physical_size, std::numeric_limits<double>::signaling_NaN()));
+
+  const Strahlkorper<Frame::Inertial> s_physical(
+      l_max, l_max, radius, center,
+      ylm::StrahlkorperConstructorData::RadiusAtCollocationPoints);
+  const Strahlkorper<Frame::Inertial> s_spectral(
```
knelli2: Maybe `strahlkorper_radius` and `strahlkorper_spectral`?

src/NumericalAlgorithms/SphericalHarmonics/IO/CMakeLists.txt
```diff
@@ -0,0 +1,15 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+spectre_target_sources(
```
knelli2, line 8: I believe you're getting undefined symbol issues because this should be its own library. You'll need

```cmake
set(LIBRARY SphericalHarmonicIO)

add_spectre_library(${LIBRARY})
```

above this function and

```cmake
target_link_libraries(
  ${LIBRARY}
  ...
)
```

at the bottom below everything where the `...` is all the libraries you need to link in (one of them being `H5`).

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
```diff
@@ -0,0 +1,31 @@
...
+/// \param file_name name of the H5 file containing the surface's spherical
+/// harmonic data
+/// \param surface_subfile_name name of the subfile within `file_name` that
+/// contains the surface's spherical harmonic data to read in
+/// \param requested_number_of_time_values the number of times to read in
+/// starting backwards from the final time found in `surface_subfile_name`
+template <typename Frame>
+std::vector<ylm::Strahlkorper<Frame>> read_surface_ylm(
+    const std::string& file_name, const std::string& surface_subfile_name,
+    size_t requested_number_of_time_values);
```
knelli2: Maybe `number_of_times_from_end`? Then just from the variable name it's clear that you are counting from the end, not the beginning.

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,159 @@
...
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+
+namespace ylm {
+namespace {
+const size_t l_max_column_number = 4;
+const size_t num_non_coef_headers = 5;
```
knelli2: These are only used in one function so they should probably go into the function.

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,159 @@
...
+}  // namespace
+
+template <typename Frame>
+std::vector<Strahlkorper<Frame>> read_surface_ylm(
+    const std::string& file_name, const std::string& surface_subfile_name,
+    const size_t requested_number_of_time_values) {
+  h5::H5File<h5::AccessType::ReadOnly> file{file_name};
+  const std::string ylm_subfile_name{std::string{"/"} + surface_subfile_name};
+  const auto& ylm_file = file.get<h5::Dat>(ylm_subfile_name);
+  const auto& ylm_data = ylm_file.get_data();
```
knelli2: `ylm_file` actually has a member function called `get_data_subset(vector_of_column_numbers, first_row, num_rows)` that'll only grab the rows you specifically want. This'll be more efficient if this file has a lot of rows because then we won't have to read in the entire file, only a subset of it. Some of your checks below will probably have to move above this call and can use `ylm_file.get_dimensions()` to check dimension size.

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
```diff
@@ -0,0 +1,31 @@
...
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+
+namespace ylm {
+/// \ingroup SurfacesGroup
+/// \brief Returns a list of `ylm::Strahlkorper`s constructed from reading in
+/// spherical harmonic data for a surface at a requested list of times
+///
+/// \details The `ylm::Strahlkorper`s are constructed by reading in data from
+/// an H5 subfile that is expected to be in the format described by
+/// `intrp::callbacks::ObserveSurfaceData`.
```
knelli2: Maybe also note that it's assumed that `l_max == m_max` somewhere. Either here or in `ObserveSurfaceData`.

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+#define FRAMETYPE(data) BOOST_PP_TUPLE_ELEM(0, data)
+
+#define INSTANTIATE(_, data)                                       \
+  template std::vector<ylm::Strahlkorper<FRAMETYPE(data)>>         \
+  ylm::read_surface_ylm<>(const std::string& file_name,            \
+                          const std::string& surface_subfile_name, \
+                          size_t requested_number_of_time_values);
+
+GENERATE_INSTANTIATIONS(INSTANTIATE, (Frame::Grid, Frame::Inertial))
```
knelli2: Also `Frame::Distorted`

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Test_ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,295 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
```
knelli2: This file should go in an `IO/` directory with it's own test library just like the source files.

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Test_ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,295 @@
...
+  std::uniform_real_distribution<> distribution(0.1, 2.0);
+
+  std::array<ylm::Strahlkorper<Frame>, NumTimes> strahlkorpers{};
+  for (size_t i = 0; i < NumTimes; i++) {
+    const auto radius = make_with_random_values<DataVector>(
+        make_not_null(&generator), distribution,
+        DataVector(ylm::Spherepack::physical_size(l_maxes[i], l_maxes[i]),
+                   std::numeric_limits<double>::signaling_NaN()));
+    strahlkorpers[i] = ylm::Strahlkorper<Frame>(
+        l_maxes[i], l_maxes[i], radius, expansion_center,
```
knelli2: `gsl::at(array, i)` for all the arrays. Same for the rest of this file

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Test_ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,295 @@
...
+// what it generates
+template <typename Frame, size_t NumTimes>
+void write_test_strahlkorpers(
+    const std::array<ylm::Strahlkorper<Frame>, NumTimes>& strahlkorpers,
+    const std::string& test_filename, const std::string& subfile_name,
+    const std::array<double, NumTimes> times, const size_t max_l) {
+  std::vector<std::vector<std::string>> legends(NumTimes);
+  std::vector<std::vector<double>> data(NumTimes);
+  for (size_t i = 0; i < NumTimes; i++) {
+    intrp::callbacks::detail::fill_ylm_legend_and_data(
```
knelli2: You could just have one `std::vector<std::string>` and always pass that to `intrp::callbacks::detail::fill_ylm_legend_and_data`. Sure you'll be overwriting multiple times but oh well. Then no need for the check below.

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Test_ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,295 @@
...
+                          3);
+                    }()),
+                    Catch::Matchers::ContainsSubstring(
+                        "Row 2 of the Ylm data does not have the expected "
+                        "format. For Lmax = 4, expected at least 30 columns"));
+
+  CHECK_THROWS_WITH(
+      ([&filename, &legend, &good_data]() {
+        std::vector<std::vector<double>> bad_data = good_data;
+        bad_data[0][legend.size() - 1] = 0.8;
```
knelli2:
```cpp
bad_data[0].back() = 0.8;
```

knelli2 (changes requested):
Thank you for doing this, it's been a big help in getting towards ringdown! I think I was able to figure out why you had symbol errors in some of the tests. See the comment below.

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
macedo22: I'm good with the values `Radius`/`Radii` and `SpectralCoefs` but the type is mostly used outside of `Strahlkorper`, so then if I did move it inside, externally this would mean the change `StrahlkorperConstructorDataType::Radius` -> `Strahlkorper::ConstructorDataType::Radius` which is ~ same diff in terms of keystrokes (with a couple extra). Let me know if I'm misunderstanding or if you have a preference, but I will at least shorten the values to `Radius`/`Radii` and `SpectralCoefs`!

knelli2: Ah yeah that's my bad. By "inside" I meant that it's only used within the context/constructor of this class and nowhere else. We actually already have a similar example of an enum used only in a class functions in `src/NumericalAlgorithms/SphericalHarmonics/SpherepackIterator.hpp`, so I'd say move it inside the `Strahlkorper` class. It keeps everything together.

src/NumericalAlgorithms/SphericalHarmonics/IO/CMakeLists.txt
macedo22: Will do this! A general question: can you not have subdirectories that are part of a library defined in the parent folder? i.e. can you not make a subfolder without also making it its own library?

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -32,10 +35,100 @@
...
+/// \param time the current time
+/// \param max_l the maximum value that \f$l_max\f$ could be (sets number of
+/// columns to write)
+template <typename Frame>
+void fill_ylm_legend_and_data(
+    const gsl::not_null<std::vector<std::string>*> legend,
+    const gsl::not_null<std::vector<double>*> data,
+    const ylm::Strahlkorper<Frame>& strahlkorper, const double time,
+    const size_t max_l) {
+  const size_t l_max = strahlkorper.l_max();
```
nilsvu: Please name this so it can't be confused with `max_l` or just inline as `strahlkorper.l_max()` where you use it.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -32,10 +35,100 @@
...
+  ASSERT(max_l >= l_max,
+         "The Lmax of the Ylm data to write ("
+             << l_max << ") is larger than the maximum value that l can be ("
+             << max_l << ").");
+  const std::array<double, 3> expansion_center =
+      strahlkorper.expansion_center();
+  // number of terms in
+  // \sum_{l=0}^{max\_l} \sum_{m=-l}^{l} F^{lm} Y^{lm}(\theta,\phi) is the
+  // sum of the first (max_l + 1) odd numbers, which is (max_l + 1)^2
+  const size_t num_coefficients = square(max_l + 1);
```
nilsvu: Use `ylm::Spherepack::spectral_size` or `strahlkorper.coefficients().size()` or `ylm::SpherepackIterator::spherepack_array_size()` please and remove the comment (rely on the Spherepack docs and perhaps improve it if you find it needs improving).

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -32,10 +35,100 @@
...
+  data->reserve(num_columns);
+
+  legend->emplace_back("Time");
+  data->emplace_back(time);
+  legend->emplace_back("ExpansionCenter_x");
+  data->emplace_back(expansion_center[0]);
+  legend->emplace_back("ExpansionCenter_y");
+  data->emplace_back(expansion_center[1]);
+  legend->emplace_back("ExpansionCenter_z");
+  data->emplace_back(expansion_center[2]);
```
nilsvu: Please encode the frame of the coordinates in the legend. Maybe `{Frame}ExpansionCenter_x`, where `{Frame}` is `get_output(Frame{})`.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -32,10 +35,100 @@
...
+  legend->emplace_back("Lmax");
+  data->emplace_back(l_max);
+
+  const DataVector ylm_coefficients = strahlkorper.coefficients();
```
nilsvu: `const DataVector&` (otherwise this will create a copy!)

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -32,10 +35,100 @@
...
+  for (size_t l = l_max + 1; l <= max_l; l++) {
+    for (int m = -l; m < 0; m++) {
+      legend->push_back(MakeString{} << "Im(" << l << "," << m << ")");
+    }
+    for (int m = 0; m <= static_cast<int>(l); m++) {
+      legend->push_back(MakeString{} << "Re(" << l << "," << m << ")");
+    }
+  }
+
+  ASSERT(legend->size() == data->size(),
```
nilsvu: Move this check to the unit tests.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -121,6 +238,28 @@ struct ObserveSurfaceData
...
+    detail::fill_ylm_legend_and_data(make_not_null(&ylm_legend),
+                                     make_not_null(&ylm_data), strahlkorper,
+                                     time, strahlkorper.l_max());
+
+    const std::string ylm_subfile_name{std::string{"/"} + surface_name +
+                                       "_Ylm"};
+
+    Parallel::threaded_action<
+        observers::ThreadedActions::WriteReductionDataRow>(
+        proxy[0], ylm_subfile_name, ylm_legend, std::make_tuple(ylm_data));
```
nilsvu: `std::move(ylm_legend)` and `std::move(ylm_data)` (not sure if that avoids a copy but can't hurt I think)

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -106,9 +222,10 @@ struct ObserveSurfaceData
         2, Spectral::Basis::SphericalHarmonic};
     const std::vector<Spectral::Quadrature> quadratures_vector{
         {Spectral::Quadrature::Gauss, Spectral::Quadrature::Equiangular}};
-    const observers::ObservationId& observation_id = observers::ObservationId(
-        InterpolationTarget_detail::get_temporal_id_value(temporal_id),
-        subfile_path + ".vol");
+    const double time =
+        InterpolationTarget_detail::get_temporal_id_value(temporal_id);
+    const observers::ObservationId& observation_id =
+        observers::ObservationId(time, subfile_path + ".vol");
```
nilsvu:
```cpp
const observers::ObservationId observation_id{time, subfile_path + ".vol"};
```

(I know you didn't write this but please fix while you're touching this line)

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
```diff
@@ -34,17 +34,27 @@ Strahlkorper<Frame>::Strahlkorper(const size_t l_max, const size_t m_max,
 template <typename Frame>
 Strahlkorper<Frame>::Strahlkorper(
     const size_t l_max, const size_t m_max,
-    const DataVector& radius_at_collocation_points,
-    std::array<double, 3> center)
+    const DataVector& collocation_radii_or_spectral_coefs,
+    std::array<double, 3> center, const StrahlkorperConstructorData data_type)
```
nilsvu: To avoid passing this enum to every construction of a Strahlkorper you could distinguish the new constructor by taking a `ModalVector spectral_coefs`. Then the correct constructor gets selected automatically, and `ModalVector` is the right data structure for the coefficients anyway.

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
nilsvu: See my comment to the Strahlkorper constructor

nilsvu (changes requested):
Haven't reviewed the third commit yet.

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
knelli2: I think choosing this method will lead to a lot of unrelated changes to this PR that are not necessary. Namely that if we add this constructor, we should also change the other two constructors that take spectral coefs as `DataVector`s and change them to `ModalVector`s, and consequently everywhere those constructors are used. However, this will require a lot of copying between `DataVector` and `ModalVector` both externally and internally. Then there is the question of if we change the internal storage of the coefs to `ModalVector` instead (since that's actually the right data structure). This would mean several `Spherepack` functions would need to change, everywhere that uses those functions, the `coefficients()` function of `Strahlkorper` and everywhere that uses that.

Basically, if we go with this change, I'll insist we do it right and not part way which adds so many unrelated changes. Maybe we should make this an issue instead?

nilsvu: I'm all for minimal changes, which is why I'm proposing this to avoid the added enum to _every_ construction of a Strahlkorper. Maybe there's a middle ground that avoids having to refactor much of Strahlkorper. I'd be fine with copying the ModalVector to the internal DataVector in the new constructor for now, with a comment that the copy can be avoided when we switch the internal storage to ModalVector. The new constructor is only used when loading the Strahlkorper from disk at the moment, so the copy doesn't hurt.

nilsvu: Then we open an issue to do the refactor to ModalVector and do it as a follow-up.

knelli2: Ok yeah I'm fine with copying the `ModalVector` into the internal `DataVector`. We'll probably need to make both the new `ModalVector` version and old `DataVector` version of that constructor `explicit` to avoid any unwanted conversion.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
macedo22: Just removed this because it is already `CHECK`ed in the unit test

macedo22:
I believe the only merge conflicts are includes but let me know if you'd like me to rebase before rereview!

knelli2:
@macedo22 Yes please rebase (but keep the fixups). CI unfortunately won't run while there are conflicts

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -78,28 +71,22 @@ void fill_ylm_legend_and_data(
 
   legend->emplace_back("Time");
   data->emplace_back(time);
-  legend->emplace_back("ExpansionCenter_x");
+  legend->emplace_back(MakeString{} << frame << "ExpansionCenter_x");
```
nilsvu: Does `frame + "ExpansionCenter_x"` not work?

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -143,14 +121,16 @@ void fill_ylm_legend_and_data(
 /// The columns of spherical harmonic data written take the form
 ///
 /// \code
-/// [Time, ExpansionCenter_x, ExpansionCenter_y, Expansion_Center_z, Lmax,
-/// Re(0,0), Im(1,-1), Re(1,0), Re(1,1), Im(2,-2), Im(2,-1), Re(2,0), Re(2,1),
-/// Re(2,2), ..., Re(Lmax,Lmax)]
+/// [Time, {Frame}ExpansionCenter_x, {Frame}ExpansionCenter_y,
+/// {Frame}Expansion_Center_z, Lmax, coef(0,0),coef(1,-1), coef(1,0), coef(1,1),
+/// coef(2,-2), coef(2,-1), coef(2,0), coef(2,1), coef(2,2), ...,
+/// coef(Lmax,Lmax)]
```
nilsvu: Maybe just write `coef(l,m) ... coef(Lmax,Lmax)` or something like that? And describe the ordering of `l` and `m` if you feel that's useful? Better than having to infer it from this list of indices.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -259,7 +238,8 @@ struct ObserveSurfaceData
 
     Parallel::threaded_action<
         observers::ThreadedActions::WriteReductionDataRow>(
-        proxy[0], ylm_subfile_name, ylm_legend, std::make_tuple(ylm_data));
+        proxy[0], std::move(ylm_subfile_name), std::move(ylm_legend),
```
nilsvu: Can't `std::move` the `ylm_subfile_name` because it's constant. But `std::move` the `ylm_data` below please.

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
```diff
@@ -34,27 +37,33 @@ Strahlkorper<Frame>::Strahlkorper(const size_t l_max, const size_t m_max,
...
+
+template <typename Frame>
+Strahlkorper<Frame>::Strahlkorper(const size_t l_max, const size_t m_max,
+                                  const ModalVector& spectral_coefficients,
+                                  std::array<double, 3> center)
+    : l_max_(l_max),
+      m_max_(m_max),
+      ylm_(l_max, m_max),
+      // clang-tidy: do not std::move trivially constructable types
+      center_(std::move(center)) {  // NOLINT
```
nilsvu: No need to `std::move` here. Clang-tidy is correct :) Same in the constructor above.

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
```diff
@@ -34,27 +37,33 @@ Strahlkorper<Frame>::Strahlkorper(const size_t l_max, const size_t m_max,
...
+  ASSERT(spectral_coefficients.size() == strahlkorper_coefs_.size(),
+         "Bad size " << spectral_coefficients.size() << ", expected "
+                     << strahlkorper_coefs_.size());
+  std::copy(spectral_coefficients.begin(), spectral_coefficients.end(),
```
nilsvu, line 64: Add a comment above this line please with something like: "We can remove this copy once the `spectral_coeffs_` are stored as `ModalVector`"

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
```diff
@@ -34,27 +37,33 @@ Strahlkorper<Frame>::Strahlkorper(const size_t l_max, const size_t m_max,
...
+template <typename Frame>
+Strahlkorper<Frame>::Strahlkorper(const size_t l_max, const size_t m_max,
+                                  const ModalVector& spectral_coefficients,
+                                  std::array<double, 3> center)
+    : l_max_(l_max),
+      m_max_(m_max),
+      ylm_(l_max, m_max),
+      // clang-tidy: do not std::move trivially constructable types
+      center_(std::move(center)) {  // NOLINT
+  ASSERT(spectral_coefficients.size() == strahlkorper_coefs_.size(),
```
nilsvu: Don't you have to resize the `strahlkorper_coefs_` first?

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
```diff
@@ -63,32 +58,34 @@ class Strahlkorper {
...
+  /// coefficients.  Specifically, `radius_at_collocation_points` has
+  /// \f$(l_{\rm max} + 1) (2 m_{\rm max} + 1)\f$ degrees of freedom,
+  /// but because there are only
   /// \f$m_{\rm max}^2+(l_{\rm max}-m_{\rm max})(2m_{\rm max}+1)\f$
-  /// spectral coefficients, it is not possible to choose spectral coefficients
-  /// to exactly match all points in `collocation_radii_or_spectral_coefs`.
+  /// spectral coefficients, it is not possible to choose spectral
+  /// coefficients to exactly match all points in
+  /// `radius_at_collocation_points`.
+  Strahlkorper(size_t l_max, size_t m_max,
```
nilsvu: Better mark this and new new constructor `explicit` as @knelli2 suggested so the `DataVector` and `ModalVector` won't be confused.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -24,18 +26,87 @@
...
+         "The Lmax of the Ylm data to write ("
+             << strahlkorper.l_max()
+             << ") is larger than the maximum value that l can be (" << max_l
+             << ").");
+  const std::array<double, 3> expansion_center =
+      strahlkorper.expansion_center();
+  const std::string frame{get_output(Frame{})};
+  // we only store half and thus only write half of the coefficients
+  const size_t num_coefficients =
+      ylm::Spherepack::spectral_size(max_l, max_l) / 2;
```
nilsvu: I'm still not quite sure this number is correct (I find the Strahlkorper/Spherepack docs a bit confusing). You do a `resize()` of the data below where you fill it with zeros, so it needs to be correct. Please add a test where `max_l` is larger than `strahlkorper.l_max()`.

nilsvu (changes requested):
You can squash once @knelli2 has seen the fixups. I still have to review the reading commit.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
macedo22: I didn't try this but I will!

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
macedo22: Ah so if you look at the documentation for `Strahlkorper::coefficients()`, my understanding is that we store half of the Spherepack coefficients (i.e. half of `spectral_size()`), so that is what is written. We could also dump everything (the whole, not just the half) but then we would just write a bunch of zeros (the other half, whose actual values it seems we don't store due to the simple complex conjugation relationship described in the docs I mentioned). `Test_ReadSurfaceData.cpp` tests this filling function for `max_l > strahlkorper.l_max()` but let me know if you want me to add another test or add additional documentation!

macedo22: And please let me know if you have a different understanding of those docs!

macedo22:
Note that the most recent push was just the rebase that @knelli2 requested

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
macedo22: Apologies - I think it's not actually half, but I think what I originally had in my original commit is correct: (max_l + 1)^2. I am not able to check this now but I will, just wanted to amend my above statement. Essentially, I think the number of coefficients is the sum of the first (max_l + 1)  odd numbers, which is (max_l + 1)^2. I will check this again, apologies!

nilsvu: The result of `ylm::Spherepack::spectral_size() / 2` is `(l_max + 1)^2` and I agree with your statement. So I think all is good. Thanks for checking! Just add a test where `l_max` is larger than `strahlkorper.l_max()` to `Test_ObserveTimeSeries` (in addition to the one in `Test_ReadSurfaceData`).

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
macedo22: I was under the impression I didn't have to since I thought it would already be sized by [this line](https://github.com/sxs-collaboration/spectre/blob/7630eea057a54396e086f7a84cf7ba3fb948cca9/src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp#L167) that sizes `strahlkorper_coefs_`, but let me know if I have that wrong

nilsvu: Does that get evaluated after the l_max is set, or with the default value of 2? Seems safer to resize explicitly, but either way this should show up in tests.

nilsvu: I looked this up and I think it's valid C++. The members are initialized in the order that they are declared, so it uses the correct l_max. Bit obscure though.

macedo22: I was going to say that we could change the assert to `ASSERT(spectral_coefficients.size() == ylm_.spectral_size()` but I guess the way it is now does also give the extra security that we know that `strahlkorper_coefs_` does not need to be resized, otherwise it would have failed the tests, as you said. But I agree, definitely an obscure detail

knelli2: Hmm I'm leaning towards removing the `ASSERT` and just explicitly doing `strahlkorper_coefs_.destructive_resize(...)` even though the current code is technically valid. This way it's very clear on the lines below that the container is the correct size and we don't have to rely on obscure knowledge on how default variables are initialized.

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Test_Strahlkorper.cpp
```diff
@@ -86,18 +84,25 @@ void test_phys_spec_constructor_consistency() {
...
-
-  CHECK(s_physical == s_spectral);
+  const Strahlkorper<Frame::Inertial> strahlkorper_physical(l_max, l_max,
+                                                            radius, center);
+
+  // create Strahlkorper using spectral coefficients
+  const size_t spectral_size = ylm::Spherepack::spectral_size(l_max, l_max);
+  ModalVector spectral_coefficients(
+      spectral_size, std::numeric_limits<double>::signaling_NaN());
+  std::copy(strahlkorper_physical.coefficients().begin(),
```
knelli2, line 99: [optional] You can avoid this extra copy by making `spectral_coefficients` non-owning

```cpp
const DataVector& coefs = strahlkorper_physical.coefficients();
const ModalVector spectral_coefficients{coefs.data(), coefs.size()};
const Strahlkorper<Frame::Inertial> strahlkorper_spectral(
      l_max, l_max, spectral_coefficients, center);
```

src/NumericalAlgorithms/SphericalHarmonics/IO/CMakeLists.txt
knelli2: In general yes you can have subdirectories that are part of the library of the parent folder. That's what you had originally. However, we don't want that in this case because we don't want the `SphericalHarmonics` library to depend on the `IO` library. That's the main reason to make the subdirectory is because we actually wanted a new library.

knelli2 (changes requested):
Ok I'm happy with all the changes! You can squash in these couple things if you decide to do them and then I'll take a look once it's all squashed together.

knelli2:
@macedo22 When squashing, can you rebase on the latest develop? A new CI container just got pushed

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
macedo22: So I was just about to add this but then I noticed that in the [docs](https://spectre-code.org/classVectorImpl.html#ad55c34f0c923e4d76ad94638c469c417) it says that if we expect it to be the right size, we shouldn't use it because it will be slow. I guess right now this constructor is only ever being called when we read in data, but nevertheless, it will be called during a time when we know that it will be correct every time and thus slower than it needs to be. Let me know if this changes your opinion or not!

macedo22: Oh wait, I completely misread it. It shouldn't be used when it's expected to resize - my bad! We're good 👍

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Test_Strahlkorper.cpp
macedo22: I unfortunately was not able to get this to compile so I left it as-is

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.cpp
macedo22: @knelli2 I amended the `ASSERT` so it's still useful in printing the error message it does during debug runs (just like the other constructor that takes the radius does) but also added the `destructive_resize` you requested

macedo22:
I'll mention that after rebasing I saw a lot of new types of clang-tidy errors that I hadn't seen before that even flagged existing code (e.g. having arguments of the same type next to each other), so I wasn't sure if these are things I needed to fix or not since they weren't enforced in the past, but let me know if I need to address these. There is also one for cognitive complexity for the `read_surface_ylm` function I added but again since it wasn't there before, I wasn't sure if I needed to address this or not

knelli2 (dismissed):
Ok I'm happy with this! Thank you for putting in all the effort.

As for clang-tidy, the new errors came up because we updated the container. We're still in the process of deciding what to do about them, so I think for now we can ignore these. They also don't seem like particularly helpful errors in my opinion.

@nilsvu Did you want to take another look?

nilsvu:
Don't worry about the new clang-tidy issues, just this one:

> /__w/spectre/spectre/src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp:49:24: error: narrowing conversion from 'double' to 'size_t' (aka 'unsigned long') [bugprone-narrowing-conversions,cppcoreguidelines-narrowing-conversions,-warnings-as-errors]
  const size_t l_max = ylm_data(row_number, l_max_column_number);

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
```diff
@@ -26,18 +28,87 @@
...
+// Fills the legend and row of spherical harmonic data to write to disk
+//
+// The number of coefficients to write is based on `max_l`, the maximum value
+// that the input `strahlkorper` could possibly have. When
+// `strahlkorper.l_max() < max_l`, coefficients with \f$l\f$ higher than
+// `strahlkorper.l_max()` will simply be zero. Assuming the same `max_l` is
+// always used for a given surface, we will always write the same number of
+// columns for each row, as `max_l` sets the number of columns to write
+template <typename Frame>
+void fill_ylm_legend_and_data(
```
nilsvu: Please move the implementation of this function to the cpp file, and instantiate it explicitly for `Frame::Inertial` (and maybe other frames @knelli2?)

src/NumericalAlgorithms/SphericalHarmonics/IO/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+  ReadSurfaceYlm.hpp
+  )
+
+target_link_libraries(
+  ${LIBRARY}
+  PRIVATE
+  DataStructures
+  H5
+  SphericalHarmonics
+  Utilities
```
nilsvu, line 27: Add `ErrorHandling`

src/NumericalAlgorithms/SphericalHarmonics/IO/CMakeLists.txt
```diff
@@ -0,0 +1,28 @@
...
+  HEADERS
+  ReadSurfaceYlm.hpp
+  )
+
+target_link_libraries(
+  ${LIBRARY}
+  PRIVATE
+  DataStructures
+  H5
+  SphericalHarmonics
```
nilsvu: This should be `PUBLIC` (you include `Strahlkorper.hpp` in the header)

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
```diff
@@ -0,0 +1,32 @@
...
+///
+/// \details The `ylm::Strahlkorper`s are constructed by reading in data from
+/// an H5 subfile that is expected to be in the format described by
+/// `intrp::callbacks::ObserveSurfaceData`. It is assumed that
+/// \f$l_{max} = m_{max}\f$.
+///
+/// \param file_name name of the H5 file containing the surface's spherical
+/// harmonic data
+/// \param surface_subfile_name name of the subfile within `file_name` that
+/// contains the surface's spherical harmonic data to read in
```
nilsvu: "without a leading slash, but including the '.dat' extension"

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,194 @@
...
+                                                ylm_data(row_number, 2),
+                                                ylm_data(row_number, 3)}};
+  const double l_max_from_file = ylm_data(row_number, l_max_column_number);
+  if (not is_nonnegative_int(l_max_from_file)) {
+    ERROR("Row " << row_number << " of the Ylm data has an invalid Lmax value ("
+                 << l_max_from_file << ") in column " << l_max_column_number
+                 << ". The value of Lmax should be a nonnegative integer.");
+  }
+
+  const size_t l_max = ylm_data(row_number, l_max_column_number);
```
nilsvu:
```cpp
const size_t l_max = static_cast<size_t>(l_max_from_file);
```

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,194 @@
...
+  SpherepackIterator iter(l_max, l_max);
+  // read in expected coefficients for the given l_max
+  for (size_t l = 0; l <= l_max; l++) {
+    for (int m = -static_cast<int>(l); m <= static_cast<int>(l); m++) {
+      iter.set(l, m);
+
+      const double coefficient = ylm_data(row_number, coef_column_number);
+      spectral_coefficients[iter()] = coefficient;
+
+      coef_column_number++;
```
nilsvu, line 135: This relies on the order of the columns. Please check the legend to make sure that the coefficient you are reading corresponds to the `(l,m)` that you are expecting here. (It would be quite hard to debug if these numbers were read in the wrong order.)

nilsvu (changes requested):

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveSurfaceData.hpp
knelli2: Yeah probably best to do all 3

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
knelli2: Maybe do this in the calling function `read_surface_ylm` so you don't have to check for every row of data?

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
```diff
@@ -27,6 +27,61 @@
 
 namespace ylm {
 namespace {
+const size_t num_non_coef_headers = 5;
+
+// check that the headers of the legend are what we expect them to be
+template <typename Frame>
+void check_legend(const std::vector<std::string>& legend) {
```
knelli2, line 34: Instantiate this function as well for all frames?

src/NumericalAlgorithms/SphericalHarmonics/IO/CMakeLists.txt
knelli2: Ah, he meant that only the `SphericalHarmonics` library should be public. The rest should be private.

The distinction for if a library dependency is public, private, or interface is this:

- INTERFACE: Files from this library dependency are included in `hpp` files *only*
- PRIVATE: Files from this library dependency are included in `cpp` files *only*
- PUBLIC: Files from this library dependency are included in both `hpp` and `cpp` files.

knelli2 (commented):
Looks good! You can squash these in once @nilsvu gives his ok.

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.cpp
macedo22: I think we don't have to do this because the instantiations of `read_surface_ylm<Frame>` instantiate `check_legend<Frame>` for all frames since it calls `check_legend<Frame>`. Let me know if you'd still like for me to add instantiations for `check_legend`, despite this!

knelli2: Oh I didn't know nested functions would also be explicitly instantiated. No need to make this change then

macedo22:
Hmm as far as I can tell, the clang-tidy errors are reporting incorrect warnings (e.g. unused variables that are actually used), but let me know if I'm missing something!

tests/Unit/NumericalAlgorithms/SphericalHarmonics/IO/Test_ReadSurfaceYlm.cpp
```diff
@@ -0,0 +1,358 @@
...
+
+  // Temporary file with test data to read in
+  const std::string test_filename{"TestReadYlm.h5"};
+  if (file_system::check_if_file_exists(test_filename)) {
+    file_system::rm(test_filename, true);
+  }
+
+  // first test surface
+  using frame_a = Frame::Grid;
+  const std::string subfile_name_a = "SurfaceA_Ylm";
```
nilsvu: (optional, you can do this while squashing) You can avoid the `_a` etc suffixes by using scopes:

```cpp
{
  INFO("Test A");
  using frame = Frame::Grid;
  ...
}
{
  INFO("Test B");
  // Here you can use the same variable names
}
```

This also gives better test output because you can use `INFO` and `CAPTURE` limited to the scope.

nilsvu (dismissed):
Looks good! Please squash. The clang-tidy errors look like false positives.

nilsvu (dismissed):

nilsvu:
To fix the CI failure, move `Observer` in `ParallelAlgorithms/Interpolation/CMakeLists.txt` from `INTERFACE` to `PUBLIC`.

knelli2 (approved):

PR #5428, opened 2023-09-07
Add maximum specific thermal energy for 2d eos

ffoucart:
## Proposed changes

Add an upper bound on the thermal energy of the fluid for 2D EoS

### Upgrade instructions

Add
`MaxThermalSpecificEnergy: 1.0e4`
to the options given to `FixToAtmosphere` (or any other value appropriate to your simulation)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/VariableFixing/FixToAtmosphere.cpp
+++ b/src/Evolution/VariableFixing/FixToAtmosphere.cpp
@@ -3,8 +3,9 @@
 
 #include "Evolution/VariableFixing/FixToAtmosphere.hpp"
 
-#include <limits>
 #include <pup.h>  // IWYU pragma: keep
+
+#include <limits>
 
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
@@ -105,9 +106,21 @@ void FixToAtmosphere<Dim>::operator()(
 
       // We probably need a better maximum temperature as well, but this is not
       // as well defined. To be discussed once implementation needs improvement.
-      if (const double max_temperature =
-              equation_of_state.temperature_upper_bound();
-          get(*temperature)[i] > max_temperature) {
+      const double max_specific_energy =
+          get(equation_of_state
+                  .specific_internal_energy_from_density_and_temperature(
+                      Scalar<double>{rest_mass_density->get()[i]},
+                      Scalar<double>{0.0})) +
+          max_thermal_specific_energy_;
+      double max_temperature =
+          get(equation_of_state.temperature_from_density_and_energy(
+              Scalar<double>{rest_mass_density->get()[i]},
+              Scalar<double>{max_specific_energy}));
+      max_temperature =
+          equation_of_state.temperature_upper_bound() > max_temperature
+              ? equation_of_state.temperature_upper_bound()
+              : max_temperature;
+      if (get(*temperature)[i] > max_temperature) {
         get(*temperature)[i] = max_temperature;
         changed_temperature = true;
       }
--- a/src/Evolution/VariableFixing/FixToAtmosphere.cpp
+++ b/src/Evolution/VariableFixing/FixToAtmosphere.cpp
@@ -20,14 +20,17 @@
 namespace VariableFixing {
 
 template <size_t Dim>
-FixToAtmosphere<Dim>::FixToAtmosphere(
-    const double density_of_atmosphere, const double density_cutoff,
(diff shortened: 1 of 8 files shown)


ffoucart:
It's not entirely clear what the "right" way to do this. SpEC has 3 different ways to impose an upper bound on the temperature, depending on the chosen equation of state (bound on temperature, specific enthalpy, or polytropic constant).
For 2D EoS with a ideal gas thermal component, what SpEC calls "temperature" is just the specific thermal energy (up to a multiplicative constant), and that is what I used in my BNS simulations so far... but we will likely have to revisit this in the future.

ermost:
For consistency with the 3D eos, would it makes sense to ultimately have this as a global temperature bound in the EOS?

(or do you use a stricter limit here, than what you would generally want to allow)?

ffoucart:
I didn't use "temperature" because this is meant mostly for the "cold EoS + thermal part" equations of state, and for those EoS what we call "temperature" is just the specific thermal energy. If we can figure out a way to have a robust definition of an upper bound on temperature across EoS I would be happy to use that instead though.

ermost:
OK. Maybe there is way: 

the temperature_from_... function for the Hybrid EOS uses the same definition as above.

[HybridEos<ColdEquationOfState>::temperature_from_density_and_energy_impl](https://github.com/sxs-collaboration/spectre/blob/046396555459b131ae10141885bf5bb3c7d9098d/src/PointwiseFunctions/Hydro/EquationsOfState/HybridEos.cpp#L124)

The 2D eos also has a temperature_upper_bound function, which currently is not set to a meaningful value. 
https://github.com/sxs-collaboration/spectre/blob/046396555459b131ae10141885bf5bb3c7d9098d/src/PointwiseFunctions/Hydro/EquationsOfState/EquationOfState.hpp#L473

would it (eventually) make sense to put the limit in temperature_upper_bound() and then just call that? 
this would generalize to the 3D eos, too.

ffoucart:
Worth thinking about.
I'm not sure whether it would be sufficient for all cases. The temperature bound might need to be dependent on density, or physical setup, possibly in a way that also depends on where in the simulation we are located... so I'm not sure whether a hard-coded bound would work (but I'm not sure that it would not work either...).
My take is that we should wait to run some sims with the 3d EoS and see what we need to keep those stable... then maybe we can come up with a consistent strategy.

ermost:
that makes sense. agree in terms of having the option for more flexible atmosphere. that can be very useful.

src/Evolution/VariableFixing/FixToAtmosphere.cpp
```diff
@@ -95,9 +99,19 @@ void FixToAtmosphere<Dim>::operator()(
...
+                      Scalar<double>{rest_mass_density->get()[i]},
+                      Scalar<double>{0.0})) + max_thermal_specific_energy_;
+      const double max_temperature =
+          get(equation_of_state.temperature_from_density_and_energy(
+              Scalar<double>{rest_mass_density->get()[i]},
+              Scalar<double>{max_specific_energy}));
 
-      if (temperature < min_temperature) {
-        temperature = min_temperature;
+      if (temperature < min_temperature || temperature > max_temperature) {
```
nilsdeppe: `||` -> `or`

src/Evolution/VariableFixing/FixToAtmosphere.hpp
```diff
@@ -104,21 +104,32 @@ class FixToAtmosphere {
         "below TransitionDensityCutoff."};
   };
 
+  /// \brief The maximum value of the thermal specific energy
+  /// Used for 2D equations of state
```
nilsdeppe: It might be good to document what a "reasonable" value is for different problems. I'm not sure where is the best place to do that, but here seems not terrible. Maybe:
```cpp
///
/// Reasonable values are:
/// BNS merger G=C=M_sun=1: 1.0e4
```
we can add more (e.g. disks, etc.) later. What do you think?

nilsdeppe (changes requested):
LGTM, couple small suggestions

ffoucart:
Rebased against develop... but this PR may no longer be needed.

ffoucart:
@nilsdeppe I think we should just archive this PR and not merge it.

nilsdeppe:
Sounds good! I'll close the PR, which archives it. We can revive it if necessary.

PR #5429, opened 2023-09-07
Build multi-platform x86 + ARM containers on CI

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/BuildDockerContainer.yaml
+++ b/.github/workflows/BuildDockerContainer.yaml
@@ -13,6 +13,11 @@ on:
           Image name to push to DockerHub
         required: true
         default: 'sxscollaboration/spectre'
+      build_arm:
+        type: boolean
+        description: >
+          Build ARM container in addition to x86_64
+        default: true
 
 jobs:
   build_container:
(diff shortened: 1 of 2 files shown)


knelli2:
How long will the eumulation take? I'm worried about this taking longer than 6 hours

nilsvu:
Let's wait for https://github.com/nilsvu/spectre/actions/runs/6117406457 :)

.github/workflows/BuildDockerContainer.yaml
```diff
@@ -52,7 +51,7 @@ jobs:
           file: "./containers/Dockerfile.buildenv"
           target: dev
           tags: ${{ inputs.image_name }}:dev
-          platforms: linux/amd64 #,linux/arm64
+          platforms: linux/amd64,linux/arm64
```
knelli2: Can we build the container on 2 cores by setting `PARALLEL_MAKE_ARG=-j2` somehow?

nilsvu: I think it already builds on 2 cores because that's the default in the container

knelli2: Ah yes you are right.

nilsvu:
The ARM container made it to building Charm++, then timed out at 6 hours. https://github.com/sxs-collaboration/spectre/pull/5432 might speed it up sufficiently (it eliminates 2.5 hours just building CMake!). I don't think we have to depend this PR on the speedup PRs since the automatic container deployment isn't critical. And the speedup PRs are easier to test once this is on develop.

knelli2 (dismissed):

nilsvu:
I added a toggle for building the ARM container rather than always building it. This way we can build the ARM container somewhere else if we don't want to wait for the emulation.

nilsdeppe (approved):

Issue #5384, opened 2023-08-24
CI Failure: InputFiles Burgers Step Yaml Execute SegFault


pajkosmi:
# Bug reports:

While waiting for CI to pass on a local, recently rebased branch, only on **Unit Tests (gcc-9, Debug)** does the "Test formaline tar can be built" step fail.

### Expected behavior:

CI completes successfully.

### Current behavior:

Only on **Unit Tests (gcc-9, Debug)**, during the "Test formaline tar can be built" step, `2 - InputFiles.Burgers.Step.yaml.execute (Failed)` occurs with a Segmentation Fault.  The entire error message is posted below under "Detailed Discussion".

### Environment:

This occurs on CI.

# Feature request:

### Component:

- [ ] Code
- [ ] Documentation
- [ ] Build system
- [x] Continuous integration

### Desired feature:

- [ ] Detail 1
- [ ] Detail 2
- [ ] Detail 3

### Detailed discussion:

Error file below:
[burgers_error.txt](https://github.com/sxs-collaboration/spectre/files/12432259/burgers_error.txt)

I don't know if they're related, but this same test failed (among others) mentioned in issue #1031, specific to the Travis system.

nilsvu:
I don't immediately see what might be causing this. I've seen it on other CI builds as well (clang-10 Debug most recently), also in the "formaline" test. AFAICT the formaline build configuration is the same as the full build..

nilsdeppe:
Did this start when we switched to formaline sharing ccache with the first build?

nilsvu:
No it happened before :(

nilsvu:
Here it happened while running input file tests (not in formaline): https://github.com/sxs-collaboration/spectre/actions/runs/6193508313/job/16815188173#step:20:1
Also in Burgers though...

nilsvu:
This is happening a lot. @wthrowe any idea what might be going on? It seems to only affect the Burgers executable. Recent failure: https://github.com/sxs-collaboration/spectre/actions/runs/6489244786/job/17623141811?pr=5546

wthrowe:
No.  I've seen it in CI a few times, but don't recall ever seeing it locally, although I don't use older gcc builds or builds of formaline.  From googling apparently the "constprop" in the backtrace means there's some sort of inlining going on, but that doesn't mean it's relevant.  The  original report doesn't have that frame, but it might have just inlined more for some reason.

PR #5432, opened 2023-09-08
Speed up building Docker containers, upgrade container to Ubuntu 22.04, test Clang 14

nilsvu:
## Proposed changes

- Installing CMake from source takes a long time (2.5h when emulating ARM on CI!), so download prebuilt binaries instead.
- Doxygen takes another 1.5h to compile when emulating ARM on CI. Instead, it is now cross-compiled for ARM on the host x86 machine, then copied over to the ARM container. (took me a while to figure out, but not as long as I expected given how complicated cross-compiling can be. The result is surprisingly straightforward, and can be used to cross-compile other packages in the future as well if needed.)
- Upgrade container to Ubuntu 22.04 and install packages through apt if available. Those are already available for x86 and ARM so we don't have to build them from source (most importantly HDF5, which takes another 1.5h to compile when emulating ARM on CI).
- Test clang 14 on CI because it's in the container now, so this was quick to add.

These speedups help build the ARM container on CI, which has to emulate ARM. Here's a CI run: https://github.com/nilsvu/spectre/actions/runs/6129266583

The other dependencies that take a long time are:

- [x] ccache (Ubuntu 22.04 has v4.5.1, which is probably enough)
- [x] Doxygen (need v1.9.3+, which is available only on Ubuntu 23.04)
- [x] HDF5 (Ubuntu 22.04 has v1.10.7, which is probably enough)
- [ ] Charm++ (twice: multicore and MPI version)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


containers/Dockerfile.buildenv
```diff
@@ -342,6 +339,12 @@ RUN wget -O - https://apt.llvm.org/llvm-snapshot.gpg.key | apt-key add - \
     && add-apt-repository 'deb http://apt.llvm.org/focal/ llvm-toolchain-focal-13 main' \
     && apt-get update -y && apt-get install -y clang-13
 
+# Install minimum required CMake version so that we test compatibility with it
+RUN wget https://github.com/Kitware/CMake/releases/download/v3.18.2/cmake-3.18.2-linux-${CMAKE_ARCH}.sh -O cmake-install.sh \
```
nilsdeppe: Sorry, what line above causes this to be skipped on ARM?

containers/Dockerfile.buildenv
```diff
@@ -304,19 +247,16 @@ RUN git clone --single-branch --branch v7.0.0 --depth 1 \
     && cd /work/charm_7_0_0 \
     && git checkout v7.0.0 \
     && git apply /work/v7.0.0.patch \
-    && ./build LIBS multicore-linux-${CHARM_ARCH} gcc \
-      ${PARALLEL_MAKE_ARG} -g -O2 --build-shared \
+    && ./build LIBS multicore-linux-${CHARM_ARCH} clang \
```
nilsdeppe: sorry why switch to clang?

nilsdeppe (changes requested):
LGTM, just a couple clarifying questions. Happy to approve once I understand :)

containers/Dockerfile.buildenv
nilsvu: This isn't skipped on ARM. Currently we don't build the CI container on ARM at all. This step (installing CMake v3.18) would be one of the steps that fail (because Kitware doesn't publish a v3.18 ARM binary). I haven't actually tried getting the CI container to build on ARM.

containers/Dockerfile.buildenv
nilsvu: I may have had a reason but don't remember, maybe just consistency with the MPI Charm build below. Prefer to drop this change?

containers/Dockerfile.buildenv
nilsdeppe: Ah okay, thanks!

containers/Dockerfile.buildenv
nilsdeppe: I think I intentionally wanted different builds just to make sure we could do both GCC & clang with charm++ :shrug: So slight preference to drop

nilsvu: ok, dropped 👍

containers/Dockerfile.buildenv
nilsvu: I added a conditional to install this only on x86 and checked that the CI container builds on ARM. We still don't build it on CI or push it to Dockerhub though, because currently GitHub doesn't have any ARM runners.

nilsvu:
Made the small change above and reduced the number of clang builds on CI. The clang-14 build isn't marked as "required" so we can merge PRs fine without rebasing.

nilsdeppe (dismissed):

nilsdeppe:
I assume I should be expecting the clang-14 failure, right? When do we want to merge this?

nilsvu:
Yes, clang-14 will fail until the upgraded container is pushed. I'm testing CI in the container on my fork. I'll message you when it's ready.

knelli2:
I was able to build the `dev` and `ci` containers on my local linux machine. I know @nilsvu is still debugging an issue on running CI right now, so we should wait on that. @nilsvu do you have a preference on if #5447 should me merged before or after this PR since it'll create conflicts?

nilsvu:
Merge https://github.com/sxs-collaboration/spectre/pull/5447 first because that's ready to go.

nilsvu:
Ok this is ready for review now. I think I worked out all the issues that came from upgrading the container to Ubuntu 22.04.

Plan:

- [x] Review
- [x] Push upgraded container
- [x] Run CI on this PR, then merge
- [ ] Rebase PRs

nilsdeppe (approved):

knelli2:
@nilsvu Can you paste a link to the CI on your fork that passed with the new container?

nilsvu:
https://github.com/nilsvu/spectre/actions/runs/6181442942

knelli2:
The Burgers executable in gcc-10 Debug failed the execute input file test with a seg fault. Do you know if that's from the container?

nilsvu:
No that's https://github.com/sxs-collaboration/spectre/issues/5384

knelli2 (approved):

knelli2:
New container being built [here](https://github.com/nilsvu/spectre/actions/runs/6188302743/job/16800016645). It will push to `sxscollaboration/spectre` when it is finished (if all goes according to plan)

PR #5433, opened 2023-09-08
Drop clang-10

nilsvu:
## Proposed changes

Preparation for upgrading the container to Ubuntu 22.04 (which doesn't bundle clang-10). Also a step toward some more C++20 support.

### Upgrade instructions

We don't guarantee compatibility with Clang 10 anymore. Upgrade to Clang 11+ or use GCC 9+.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -405,21 +405,6 @@ jobs:
           - compiler: gcc-10
             build_type: Debug
             TEST_TIMEOUT_FACTOR: 2
-          # Add a test without PCH to the build matrix, which only builds core
-          # libraries. Building all the tests without the PCH takes very long
-          # and the most we would catch is a missing include of something that's
-          # in the PCH.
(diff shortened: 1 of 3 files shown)


nilsdeppe (approved):

knelli2 (approved):
Can you add this to the upgrade instructions? Seems like an important thing to make users aware of. Also, I think there are a few places where we have `#if defined(__clang__)` that pertain to versions 10 and lower. Do you want to get rid of those or just leave them?

nilsvu:
We can delete workarounds for clang 10 when we encounter them and deleting them helps somehow. I see no reason to intentionally break compatibility with clang 10 in this PR.

nilsvu:
@wthrowe and @kidder do you want to look at this?

wthrowe (approved):

kidder (approved):

PR #5434, opened 2023-09-08
Make KastaunHydro stable against unphysical energy states

ermost:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/KastaunEtAlHydro.cpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/KastaunEtAlHydro.cpp
@@ -27,7 +27,7 @@ struct Primitives {
   const double specific_internal_energy;
 };
 
-template <size_t ThermodynamicDim>
+template <size_t ThermodynamicDim, bool EnforcePhysicality>
 class FunctionOfZ {
  public:
   FunctionOfZ(const double total_energy_density,
@@ -41,11 +41,27 @@ class FunctionOfZ {
       : q_(total_energy_density / rest_mass_density_times_lorentz_factor - 1.0),
         r_squared_(momentum_density_squared /
                    square(rest_mass_density_times_lorentz_factor)),
+        r_(std::sqrt(r_squared_)),
         rest_mass_density_times_lorentz_factor_(
             rest_mass_density_times_lorentz_factor),
         electron_fraction_(electron_fraction),
         equation_of_state_(equation_of_state) {
-    r_ = std::sqrt(r_squared_);
+
+    // Internal consistency check
+    //
+    const double rho_min =
+        rest_mass_density_times_lorentz_factor_ / lorentz_max_;
+    const double eps_min =
+        equation_of_state_.specific_internal_energy_lower_bound(rho_min);
+
+    if constexpr (EnforcePhysicality) {
+      q_ = std::max(q_, eps_min);
+      r_ = std::min(r_, kmax_ * (q_ + 1.));
+    } else {
+      if ((q_ < eps_min) or (r_ > kmax_ * (q_ + 1.))) {
+        state_is_unphysical = true;
+      }
+    }
   }
 
   std::pair<double, double> root_bracket();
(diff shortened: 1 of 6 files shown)


ermost:
@nilsdeppe @yoonso0-0 

I'm trying to figure out why the TciOnDgGrid fails with this PR. It looks like there is an option that triggers on primitive recovery failure. With this PR primitive recovery does not throw any error for hydro conservatives, since internal conservative fixing is applied. This is because the KastaunHydro routine will give FPEs if the input is not physical (S^2 > (tau + D)^2).
On FD subcell this seems fine to do, but on DG you probably want to fail when you violate physicality.

Do you have any thoughts on how to best handle this?

yoonso0-0:
Just took a quick look, and it seems that `Test_TciOnDgGrid.cpp` fails. If you look around the line 210, there is a code doing

```
  } else if (test_this == TestThis::PrimRecoveryFailed) {
    db::mutate<grmhd::ValenciaDivClean::Tags::TildeS<Frame::Inertial>>(
        [point_to_change](const auto tilde_s_ptr) {
          // Manipulate one of the conserved variable in an (very) inconsistent
          // way so that primitive recovery fails.
          get<0>(*tilde_s_ptr)[point_to_change] = 1e3;
        },
        make_not_null(&box));
```

which changes the value of one of the conserved variable (TildeS) in a random way so that the prim recovery fails 'on purpose'.

`Test_TciOnDgGrid.cpp` uses only `NewmanHamlin` for the prim recovery failing case, so this must be related to any change made in `PrimitiveFromConservative`...?

src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
```diff
@@ -178,9 +178,9 @@ bool PrimitiveFromConservative<OrderedListOfPrimitiveRecoverySchemes,
         equal_within_roundoff(
             total_energy_density[s],
             get(magnetic_field_squared)[s] * 0.5 + total_energy_density[s])) {
```
yoonso0-0, line 180: So here is what happens.

At the line 88 of `Test_TciOnDgGrid.cpp`, magnetic field as a prim variable is set to be very very small (not sure why we chose to use a hydro regime case for testing the TCI?). So even if the following (line 319)
```
  const std::tuple<int, evolution::dg::subcell::RdmpTciData> result =
      db::mutate_apply<grmhd::ValenciaDivClean::subcell::TciOnDgGrid<
          grmhd::ValenciaDivClean::PrimitiveRecoverySchemes::NewmanHamlin>>(
          make_not_null(&box), persson_exponent, element_stays_on_dg);
```
is executed with the `NewmanHamlin` scheme as the template parameter, due to this `equal_within_roundoff` consistency check that it'll run the `KastaunEtAlHydro` scheme.

src/Evolution/Systems/GrMhd/ValenciaDivClean/KastaunEtAlHydro.cpp
```diff
@@ -46,6 +46,17 @@ class FunctionOfZ {
         electron_fraction_(electron_fraction),
         equation_of_state_(equation_of_state) {
     r_ = std::sqrt(r_squared_);
+
+    // Internal consistency check
+
+    if constexpr (EnforcePhysicality) {
```
yoonso0-0, line 57: (continued from the last comment)

but here in the `hpp` file `EnforcePhysicality` was set to `true`. I think the enforcing physicality here puts the insane value of momentum density
```
get<0>(*tilde_s_ptr)[point_to_change] = 1e3;
```
(line 213 of Test_TciOnDgGrid.cpp`) back to a correct physical value, then the prim recovery actually succeeds.

yoonso0-0: If my guess is correct, we need to fix the line 213 of `Test_TcionDgGrid` so that it puts another, well-failing (?) combination of conserved variables so that the prim recovery can successfully fail. Not sure if it's easier to increase the magnetic field in the test.

src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
ermost: That's correct. KastaunHydro is now default regardless of the choice of MHD inversion.

src/Evolution/Systems/GrMhd/ValenciaDivClean/KastaunEtAlHydro.cpp
ermost: Yep, that's exactly what happens. Currently the routine fixes too aggressively, when it's not supposed to do that. I added a missing commit that should address this.

ermost:
Thanks a lot for your comments and analysis @yoonso0-0 :)

I pushed another commit that couples the ErrorOnFailure template parameter with KastaunHydro and lets the routine fail if the input is unphysical. Setting this option in TciOnDg seems to restore the expected behavior.

ermost:
The only remaining question is how to now handle ErrorOnFailure:
Currently this is supposed to immediately abort the code (without throwing a C++ error that could be caught?).

I have just commented this out for now, but this should be handled somehow. Possible solutions could be to either throw an actual error that could be handled, make the abort option dependent on another compile time parameter, or pass this in differently. Thoughts?

ermost:
I moved the template parameter to the apply function as discussed in the call.

tests/Unit/Evolution/Systems/GrMhd/ValenciaDivClean/Test_PrimitiveFromConservative.cpp
```diff
@@ -284,6 +284,26 @@ void test_primitive_from_conservative_known(const DataVector& used_for_size) {
   CHECK_ITERABLE_APPROX(expected_magnetic_field, magnetic_field);
   CHECK_ITERABLE_APPROX(expected_divergence_cleaning_field,
                         divergence_cleaning_field);
+
+  if constexpr (!UseMagneticField) {
```
nilsdeppe: `!` -> `not`

nilsdeppe (commented):
LGTM, one minor style-guide thing you can squash immediately!

ermost:
I fixed the style issue and squashed.

nilsdeppe (approved):

PR #5441, opened 2023-09-08
Add CircularDeque container

wthrowe:
This will be used in the rewritten BoundaryHistory.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/StaticDeque.hpp
+++ b/src/DataStructures/StaticDeque.hpp
@@ -6,6 +6,8 @@
 #include <algorithm>
 #include <cstddef>
 #include <iterator>
+#include <limits>
+#include <memory>
 #include <pup.h>
 #include <type_traits>
 
@@ -39,6 +41,9 @@ template <typename T, size_t Capacity>
 class StaticDeque
     : public stl_boilerplate::RandomAccessSequence<StaticDeque<T, Capacity>,
                                                    T, true> {
+  // Capacity = size_t max is special and dynamically allocates, but
+  // we don't advertise that here and only intend it to be used
+  // through CircularDeque.
  public:
   // Aliases needed in the class definition.  The rest can just be
   // inherited.
@@ -72,10 +77,12 @@ class StaticDeque
   void assign(std::initializer_list<T> init);
 
   size_type size() const { return size_; }
-  size_type max_size() const { return Capacity; }
+  size_type capacity() const { return data_.capacity; }
+  static constexpr size_type max_size() { return Capacity; }
+  void reserve(size_type count);
   void resize(size_type count);
   void resize(size_type count, const T& value);
-  void shrink_to_fit() {}
+  void shrink_to_fit();
 
   T& operator[](size_type n) { return *entry_address(n); }
   const T& operator[](size_type n) const { return *entry_address(n); }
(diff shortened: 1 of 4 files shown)


src/DataStructures/StaticDeque.hpp
```diff
@@ -155,14 +166,39 @@ class StaticDeque
...
+    std::byte* get() { return data; }
+    const std::byte* get() const { return data; }
+  };
+
+  // gcc improperly forbids full specializations
+  // (https://gcc.gnu.org/bugzilla/show_bug.cgi?id=85282)
+  template <typename Void>
+  struct Data<std::numeric_limits<size_t>::max(), Void> {
+    static constexpr bool variable_capacity = true;
+    // There doesn't seem to be any alignment guarantee on the
```
nilsdeppe, line 191: for AVX/simd types we might want 64-byte alignment. Is the `data` itself going to be aligned, or is it the objects stored in the `data` that get aligned? In the second case I'm thinking we might have wasted space. If the second, should/could we do some sort of forced-alignment when allocating this pointer?

src/DataStructures/StaticDeque.hpp
```diff
@@ -523,8 +614,44 @@ void StaticDeque<T, Capacity>::pup(PUP::er& p) {
   }
 }
 
+template <typename T, size_t Capacity>
+StaticDeque<T, Capacity>::StaticDeque(
+    StaticDeque<T, Capacity>::WithCapacityTag /*meta*/, const size_t capacity)
+    : data_{std::unique_ptr<std::byte[]>(new std::byte[capacity * sizeof(T)]),
```
nilsdeppe, line 620: Re alignment: can we make the `new` call here allocate to be aligned with `T`?

nilsdeppe (changes requested):
LGTM, a couple questions for my understanding

src/DataStructures/StaticDeque.hpp
wthrowe: This is the alignment of `*data` matching the alignment of `T` (or actually of any type with a fundamental alignment because none of the allocation code uses `T` except to get the size).  The standard guarantees that arrays of `char`, `unsigned char`, and `std::byte` (but apparently not `signed char`?) created with `operator new[]` will be properly aligned for any type with fundamental alignment, intended for exactly what we are doing here allocating other objects into a buffer.  I couldn't find any guarantee that `std::vector<std::byte>` also guaranteed that, although I'd be pretty shocked if it didn't do so in practice.

According to cppreference this is usually either 8 or 16 byte alignment.  For stricter alignment we could use `std::aligned_alloc` instead of `operator new[]` (and `free` instead of `delete[]`).  I don't think any other changes would be required.  If such types' requirements are properly reported by `alignof` that would be easy.

src/DataStructures/StaticDeque.hpp
wthrowe: Mostly covered by the discussion on the other comment, but short answer: it does if `T` has a fundamental alignment, and we should modify it to handle the general case if it's easy.

src/DataStructures/StaticDeque.hpp
nilsdeppe: Okay, that sounds good! The only types where I know stricter alignment _can_ be useful SIMD, but even that's not really the case right now. It used to be alignment matter, but a lot less so nowadays :shrug:

nilsdeppe (approved):

PR #5453, opened 2023-09-14
Fix race in list of StoredInfos

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
+++ b/src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
@@ -61,28 +61,51 @@ const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
          "constructed properly?");
 
   auto upper_bound_stored_info = all_stored_infos.begin();
+
+  // If 'upper_bound_stored_info' points to the beginning of the list, we can't
+  // use std::prev
+  const auto check_for_beginning = [&t, &all_stored_infos,
+                                    &upper_bound_stored_info]()
+      -> const StoredInfo<MaxDerivPlusOne, StoreCoefs>& {
+    if (upper_bound_stored_info == all_stored_infos.begin()) {
+      // all elements of times are greater than t
+      // check if t is just less than the min element by roundoff
+      if (not equal_within_roundoff(upper_bound_stored_info->time, t)) {
+        ERROR("requested time " << t << " precedes earliest time "
+                                << upper_bound_stored_info->time
+                                << " of times.");
+      }
+      return *upper_bound_stored_info;
+    } else {
+      return *std::prev(upper_bound_stored_info);
+    }
+  };
+
   // Linear search is faster with std::list because of expensive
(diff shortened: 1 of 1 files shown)


wthrowe (approved):

kidder:
Is it possible to add a test that would have caught the bug?

knelli2:
You're asking if it's possible to consistently produce a race condition?

kidder:
At the very least, I assume there is some inputs to this function that now give different results before/after the changes in this PR; therefore you should be able to add a test that checks that you get the new result

knelli2:
No the same inputs should still give the same output, before/after the change. The issue was what happened if the list that was passed to the function was appended to *during* the function call. Before this change, there was a race condition. After the change, there shouldn't be. If the length of the list didn't change during the function call, then all was/is good. So if we wanted to test this, I'd somehow have to consistently edit the passed in list during the function call from a different core and I'm not sure how to do that.

src/Domain/FunctionsOfTime/FunctionOfTimeHelpers.cpp
```diff
@@ -61,28 +61,51 @@ const StoredInfo<MaxDerivPlusOne, StoreCoefs>& stored_info_from_upper_bound(
...
-  // std::advance/distance inside std::lower_bound
-  for (size_t i = 0; i < all_info_size; i++, upper_bound_stored_info++) {
+  // std::advance/distance inside std::lower_bound. Only search until size - 1
+  // so we don't send the iterator past-the-end. A past-the-end iterator for a
+  // list *stays* past-the-end even if there was an insert at the end. If
+  // condition is satisfied, t is within the range of times so
+  // 'upper_bound_stored_info' currently points to one index past the desired
+  // index. Here it's safe to use std::prev because we aren't pointing
+  // past-the-end.
+  for (size_t i = 0; i < all_info_size - 1; i++, upper_bound_stored_info++) {
```
kidder, line 92: I would find this code much easier to understand if you factored out the `i=0` and started this loop with `i=1`
Then you: 
- know you are at the beginning of the list before entering the loop
- can always return `std::prev` inside the loop
- can return `std::prev` or the last element after the loop as you know you're not at the beginning of the list

knelli2: This gets messy if `all_info_size == 1` because the condition that I usually check (`if (not(iter->time < t))`) will sometimes be false with only one element. So if I check `i=0` before the loop, it wouldn't do anything. Then I'd get to a loop that goes like `for (i=1; i<0; i++)` which won't actually do anything. Then you get to the final check, but I can't `++` the iterator otherwise it'll be past-the-end and we're right back where we started. So I *again* have to check if the iterator is the beginning before the `++` to ensure I do the right thing. To me, this seems even more complicated.

My current implementation takes care of these considerations, and I think it is fairly succinct. Let me know if you'd still like me to rework this.

kidder:
Ignoring usual CI failures

PR #5461, opened 2023-09-15
Rewrite BoundaryHistory to support substeps

wthrowe:
If reviewers prefer, the second and third commits could be moved to a separate PR.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -86,6 +86,10 @@ Checks: '*,
 CheckOptions:
   - key: performance-move-const-arg.CheckTriviallyCopyableMove
     value: false
+  - key: cppcoreguidelines-special-member-functions.AllowSoleDefaultDtor
+    value: true
+  - key: cppcoreguidelines-special-member-functions.AllowMissingMoveFunctions
+    value: true
 WarningsAsErrors: '*'
 # It is unclear if the header filter actually works or how to use it so
 # just include all headers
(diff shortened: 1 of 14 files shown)


kidder (commented):
- I like the rewrite and I have no requested changes.
- clang-tidy has flagged some moves of trivially-copyable type
- gcc11 is not compiling the print function, maybe need a `using ::operator<<;` or an include in the test?

wthrowe:
I can reproduce the gcc 11 problem locally, but I'm not having any luck fixing it.  Adding `using ::operator<<;` immediately before the problematic line and including `Utilities/StdHelpers.hpp` as the first thing in the compilation unit doesn't do anything.  Calling `operator<<` directly either doesn't do anything of gives an error about ambiguity (I think in a different call) depending on exactly how I do it.

wthrowe:
gcc 11 fixed.  Search for `print_stl` if you want to see how.

kidder (dismissed):

kidder:
Hi @wthrowe some of the clang-tidy errors are ones we have fixed in the past 
```
error: parameter 'entry' is const-qualified in the function declaration; const-qualification of parameters only has an effect in function definitions [readability-avoid-const-params-in-decls,-warnings-as-errors]
```

wthrowe:
Disabled some checks in a new commit, fixed the rest.  Also deleted an internal function that was never used.

nilsdeppe (approved):

nilsdeppe:
I confirmed with @kidder that this can get merged

PR #5454, opened 2023-09-14
Ignore some more clang-tidy warnings

knelli2:
### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

Issue #5245, opened 2023-07-26
Add size control states 3,4,5


knelli2:
Tests of both head-on BBHs and inspirals lead us to the conclusion that we need this now.

State 5 can be added independently of 3 and 4. State 3 needs a new quantity, the gradient of the comoving characteristic speed: $\nabla_r \lambda_{\mathrm{comov}}$.

Action list:
- [x] Add state 5
- [ ] Add $\nabla_r \lambda_{\mathrm{comov}}$
- [x] Add state 3 and 4

geoffrey4444:
Issue is finished, but we need to test State 3, see why it's not being used

PR #5462, opened 2023-09-15
Add DeltaRDriftOutward (State 5) to size control.

markscheel:
This state reacts to an excision boundary too deep inside the horizon, and slowly drifts the excision boundary outward.
Input files are set so that state 5 is off by default, as in SpEC.
The tests of the logic mirror the identical tests in SpEC.

### Upgrade instructions

Input files using size control need to add a new option:
 * DeltaRDriftOutwardOptions: None

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size.hpp
+++ b/src/ControlSystem/ControlErrors/Size.hpp
@@ -28,6 +28,7 @@
 #include "NumericalAlgorithms/Interpolation/ZeroCrossingPredictor.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
+#include "Options/Auto.hpp"
 #include "Options/String.hpp"
 #include "Parallel/GlobalCache.hpp"
 #include "Parallel/Printf.hpp"
@@ -189,8 +190,46 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
         "TimescaleTuner for smoothing horizon measurements."};
   };
 
+  struct DeltaRDriftOutwardOptions {
+    using type =
+        Options::Auto<DeltaRDriftOutwardOptions, Options::AutoLabel::None>;
+    static constexpr Options::String help{
+        "Options for State DeltaRDriftOutward. Specify 'None' to disable State "
+        "DeltaRDriftOutward."};
+    struct MaxAllowedRadialDistance {
+      using type = double;
+      static constexpr Options::String help{
+          "Drift excision boundary outward if distance from horizon to "
+          "excision exceeds this."};
+    };
+    struct OutwardDriftVelocity {
+      using type = double;
+      static constexpr Options::String help{
+          "Constant drift velocity term, if triggered by "
+          "MaxAllowedRadialDistance."};
+    };
+    struct OutwardDriftTimescale {
+      using type = double;
+      static constexpr Options::String help{
+          "Denominator in non-constant drift velocity term, if triggered by "
+          "MaxAllowedRadialDistance."};
+    };
+    using options = tmpl::list<MaxAllowedRadialDistance, OutwardDriftVelocity,
+                               OutwardDriftTimescale>;
+    void pup(PUP::er& p) {
+      p | max_allowed_radial_distance;
+      p | outward_drift_velocity;
+      p | outward_drift_timescale;
+    }
+
+    double max_allowed_radial_distance{};
+    double outward_drift_velocity{};
+    double outward_drift_timescale{};
+  };
+
   using options = tmpl::list<MaxNumTimesForZeroCrossingPredictor,
-                             SmoothAvgTimescaleFraction, SmootherTuner>;
+                             SmoothAvgTimescaleFraction, SmootherTuner,
+                             DeltaRDriftOutwardOptions>;
   static constexpr Options::String help{
       "Computes the control error for size control. Will also write a "
       "diagnostics file if the control systems are allowed to write data to "
@@ -211,7 +250,8 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
    * is moved inside this class.
    */
   Size(const int max_times, const double smooth_avg_timescale_frac,
-       TimescaleTuner smoother_tuner);
+       TimescaleTuner smoother_tuner,
+       std::optional<DeltaRDriftOutwardOptions> delta_r_drift_outward_options);
 
   /// Returns the internal `control_system::size::Info::suggested_time_scale`. A
   /// std::nullopt means that no timescale is suggested.
@@ -358,6 +398,16 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
     const double control_error_delta_r = size::control_error_delta_r(
         horizon_00, dt_horizon_00, lambda_00, dt_lambda_00,
         grid_frame_excision_sphere_radius);
+    const std::optional<double> control_error_delta_r_outward =
+        delta_r_drift_outward_options_.has_value()
+            ? std::optional<double>(control_error_delta_r -
+                                    delta_r_drift_outward_options_.value()
+                                        .outward_drift_velocity -
+                                    (lambda_00 + horizon_00 -
+                                     grid_frame_excision_sphere_radius / Y00) /
+                                        delta_r_drift_outward_options_.value()
+                                            .outward_drift_timescale)
+            : std::nullopt;
 
     info_.damping_time = min(tuner.current_timescale());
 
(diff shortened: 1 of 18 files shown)


markscheel:
clang-tidy fails because of silly new critera like disallowing `f(double,double)`.

knelli2:
Don't worry about clang-tidy. That's dealt with in #5454.

knelli2:
This PR adds part of #5245

src/ControlSystem/ControlErrors/Size/DeltaRDriftOutward.hpp
```diff
@@ -0,0 +1,32 @@
...
+ public:
+  DeltaRDriftOutward() = default;
+  std::string name() const override { return "DeltaRDriftOutward"; }
+  size_t number() const override { return 5; }
+  std::unique_ptr<State> get_clone() const override;
+  std::string update(const gsl::not_null<Info*> info,
+                     const StateUpdateArgs& update_args,
+                     const CrossingTimeInfo& crossing_time_info) const override;
+  /// The return value is Q from Eq. 96 of \cite Hemberger2012jz, plus
+  /// a velocity term.
```
knelli2: `an outward velocity term`

src/ControlSystem/ControlErrors/Size/DeltaRDriftOutward.cpp
```diff
@@ -0,0 +1,90 @@
...
+    // AhSpeed.  It needs to be slightly greater than unity, but the
+    // control system should not be sensitive to the exact
+    // value. The value of 1.01 was chosen arbitrarily in SpEC and
+    // never needed to be changed.
+    constexpr double non_oscillation_factor = 1.01;
+    info->discontinuous_change_has_occurred = true;
+    info->state = std::make_unique<States::AhSpeed>();
+    info->target_char_speed =
+        update_args.min_char_speed * non_oscillation_factor;
+    ss << " Switching to AhSpeed.\n";
```
knelli2: This can be appended to the `ss` above.

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -189,8 +189,31 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
         "TimescaleTuner for smoothing horizon measurements."};
   };
 
+  struct MaxAllowedRadialDistance {
```
knelli2: How about we add these three options to their own struct `State5Options` so it's easier to disable? Doing this

```cpp
  struct State5Options {
    struct MaxAllowedRadialDistance {}
    struct OutwardDriftVelocity {}
    struct OutwardDriftTimescale {}
    // rest of the option stuff and member variables
    
    using type = Options::Auto<State5Options, Options::AutoLabel::None>;
    static constexpr Options::String help = ...;
  };

  using options = tmpl::list<..., State5Options, ...>;

  Size(..., const std::optional<State5Options>& state_5_options, ...);
```

lets us do in the input file

```yaml
ControlSystems:
  Size:
    ControlError:
      ...
      State5Options: None
```

in order to disable state 5. I think this is clearer for users than having to specify 0 velocity or a super large timescale in order to disable it. This might mean a few other `double`s in this change have to now be `std::optional<double>`

src/ControlSystem/ControlErrors/Size/DeltaR.cpp
```diff
@@ -96,6 +103,14 @@ std::string DeltaR::update(const gsl::not_null<Info*> info,
        << std::abs(update_args.control_error_delta_r) << " > threshold "
        << delta_r_control_error_threshold << ". Staying in DeltaR.\n";
     ss << " Suggested timescale = " << info->suggested_time_scale;
+  } else if (update_args.average_radial_distance >
```
knelli2: If you make the change about `Options::Auto`, then you need an additional check here `else if (update_args.state_5_stuff.has_value() and ...)` (and in `Initial`)

knelli2 (changes requested):
You can squash these changes in directly.

src/ControlSystem/ControlErrors/Size.hpp
markscheel: Yes, this is way better!

markscheel:
made changes and squashed.

Also, added stuff to pup which I forgot last time.

src/ControlSystem/ControlErrors/Size/Error.hpp
```diff
@@ -64,6 +65,11 @@ struct ErrorDiagnostics {
  * \param time the current time.
  * \param control_error_delta_r the control error for the DeltaR state. This is
  *        used in other states as well.
+ * \param control_error_delta_r_outward the control error for the
+ *        DeltaRDriftOutward state.
+ * \param max_allowed_radial_distance the maximum average radial distance
+ *        between the horizon and the excision boundary that is allowed without
+ *        triggering the DeltaRDriftOutward state.
```
knelli2: Maybe note in these two that `std::nullopt` means `DeltaRDriftOutward` won't be used.

src/ControlSystem/ControlErrors/Size/State.hpp
```diff
@@ -34,6 +35,13 @@ struct StateUpdateArgs {
   /// is in state Label::DeltaR.
   /// This is Q in Eq. 96 of \cite Hemberger2012jz.
   double control_error_delta_r;
+  /// average_radial_distance is the average distance between the horizon and
+  /// the excision boundary.
+  std::optional<double> average_radial_distance;
+  /// max_allowed_radial_distance is the minimum distance between the horizon
+  /// and the excision boundary that will trigger state
+  /// Label::DeltaRDriftOutward.
+  std::optional<double> max_allowed_radial_distance;
```
knelli2: I think saying what `nullopt` means here would also be good.

src/ControlSystem/ControlErrors/Size/State.hpp
```diff
@@ -43,6 +51,12 @@ struct StateUpdateArgs {
 struct ControlErrorArgs {
   double min_char_speed;
   double control_error_delta_r;
+  /*!
+   * \brief control_error_delta_r_outward is the control error in the
+   * case where there is an outward drift caused by too large separation
+   * between the horizon and the excision boundary.
+   */
+  std::optional<double> control_error_delta_r_outward;
```
knelli2, line 62: And here

tests/Unit/ControlSystem/ControlErrors/Test_SizeError.cpp
```diff
@@ -166,9 +171,20 @@ void test_size_error_one_step(
           lambda_dt_lambda[0][0], lambda_dt_lambda[1][0],
           grid_excision_boundary_radius);
 
+  const double Y00 = 0.25 * M_2_SQRTPI;
+  const std::optional<double> control_error_delta_r_outward =
```
knelli2: Can this just be default constructed because `DeltaRDriftOutward` is turned off for this test?

knelli2 (changes requested):
Couple dox things and one other thing. Please squash!

Edit: Also change the Upgrade Instructions in the PR description.

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -189,8 +190,44 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
         "TimescaleTuner for smoothing horizon measurements."};
   };
 
-  using options = tmpl::list<MaxNumTimesForZeroCrossingPredictor,
-                             SmoothAvgTimescaleFraction, SmootherTuner>;
+  struct State5Options {
```
wthrowe: Up to you, but since internally this is called `DeltaRDriftOutward`, maybe name the option that too.  Seems more user-friendly.

knelli2: Oh yeah, probably a good idea. `5` is synonymous in my head with `DeltaRDriftOutward` which is why I put that

markscheel:
Made changes (including renaming State5Options -> DeltaRDriftOutwardOptions) and squashed.

knelli2:
@markscheel Forgot to push? I don't see the changes

markscheel:
(yeah forgot to push)

wthrowe (dismissed):
Looks fine except CI is legitimately failing.  Don't need another look from me unless there are significant changes.

tests/Unit/ControlSystem/ControlErrors/Test_SizeError.cpp
```diff
@@ -92,6 +93,10 @@ void test_size_error_one_step(
   const double initial_damping_time = 0.1;
   const double initial_target_drift_velocity = 0.0;
   const double initial_suggested_time_scale = 0.0;
+  // Set constants so that State::DeltaROutward is turned off.
+  const std::optional<double> max_allowed_radial_distance{};
+  const std::optional<double> outward_drift_velocity{};
+  const std::optional<double> outward_drift_timescale{};
```
knelli2: Oh yeah these two aren't used anymore.

knelli2 (changes requested):

tests/Unit/ControlSystem/ControlErrors/Test_SizeError.cpp
markscheel: Removed and squashed.

knelli2 (approved):

knelli2:
Ignoring clang-tidy and gcc-11 Release for the usual reasons. Unsure why gcc-9 Release failed building the executables, but something got messed up with the logs so can't figure out why. However, this change seems unlikely to have caused the issue, so merging anyways.

PR #5460, opened 2023-09-15
Print backtrace in FixedHashMap exceptions, abbreviate function name in ERROR

nilsvu:
## Proposed changes

Print backtrace in out-of-range errors that occur in commonly used containers, like `DirectionMap`.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5467, opened 2023-09-18
Add ERROR_AS for custom exception types on errors

nilsdeppe:
## Proposed changes

- Allow for custom exception types using `ERROR_AS`
- Add `SpectreFpe` exception type
- Move `ErrorHandling` tests into `Utilities`

Some questions:
1. Do we want more exception types? (Adding them changes the PCH so best to add as many as we think we'll need at once)
2. Do we want to namespace all of our exceptions?

**Note** I've canceled CI because it's such a big build that it won't complete _and_ we should first decide what we want.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Utilities/ErrorHandling/Error.hpp
```diff
@@ -66,7 +68,29 @@ template <typename F>
...
+
+/*!
+ * \ingroup ErrorHandlingGroup
+ * \brief Same as ERROR but will throw `EXCEPTION_TYPE` instead of
+ * `SpectreError`.
+ *
+ * \note Any exception types used must have an associated explicit
+ * instantiation in `Utilities/ErrorHandling/AbortWithErrorMessage.cpp`
+ */
+#define ERROR_WITH(m, EXCEPTION_TYPE)                                        \
```
nilsvu: Quick note: Catch2 calls the corresponding check `CHECK_THROWS_AS( expression, exception type )`, so you could do `ERROR_AS` to be consistent with that (if you want):
https://github.com/catchorg/Catch2/blob/devel/docs/assertions.md#exceptions

src/Utilities/ErrorHandling/AbortWithErrorMessage.hpp
```diff
@@ -26,6 +26,7 @@
 /// We try to demangle and format the backtrace. Long symbol names are
 /// abbreviated, unless you set the `SPECTRE_SHOW_FULL_BACKTRACE_SYMBOLS`
 /// environment variable to a non-empty value (e.g. "1").
+template <typename ExceptionTypeToThrow>
```
knelli2, line 29: Can't reach with Github comments, but maybe document on `abort_with_error_message_no_trace` that it always throws a `SpectreError` even though it won't actually print the exception type?

src/Utilities/ErrorHandling/Exceptions.hpp
```diff
@@ -19,6 +19,13 @@ class SpectreError : public std::runtime_error {
   explicit SpectreError(const std::string& message) : runtime_error(message) {}
 };
 
+/// \ingroup ErrorHandlingGroup
+/// Exception indicating an ERROR was triggered because of an FPE
+class SpectreFpe : public std::runtime_error {
```
wthrowe, line 28: Note that developers can't rely on this being thrown in all configurations, so it can't be used to try to recover from the error.

wthrowe (changes requested):

src/Utilities/ErrorHandling/Exceptions.hpp
knelli2: What's the fallback if this can't be raised?

wthrowe: Depending on the setup, it may abort the program (maybe call `std::terminate` or one of those things?) without reliably catching the exception, or the FPE may be ignored and execution will continue.

knelli2: Oh that second option is super fun! Yeah A note or even a warning would probably be good here.

wthrowe: The first one should hopefully be rare, since it is basically a compiler-support issue and I think everything we support supports it.  That one also applies to segfaults.

The second is a hardware issue (or maybe some really low-level software, not sure), so nothing we can do about it.  It affects ARM.  (But not segfault reporting.)

src/Utilities/ErrorHandling/Error.hpp
nilsdeppe: Seems reasonable!

kidder (commented):
I have no additional requested changes

src/Utilities/ErrorHandling/Exceptions.hpp
nilsdeppe: We might be able to make some progress on macOS. From this https://stackoverflow.com/questions/69059981/how-to-trap-floating-point-exceptions-on-m1-macs (and the associated long discussion) it seems FPEs do get trapped on Aarch64 on Linux, but macOS FPEs raise SIGILL. That seems to be an issue with macOS rather than AArch64. Note that the ARM toolchain doesn't support FPE trapping with AArch64, but it seems clang & gcc do.(?)

nilsdeppe:
rebased, squashed and included https://github.com/sxs-collaboration/spectre/pull/5460 since that's also related to the same files.

knelli2 (approved):

nilsvu (dismissed):

kidder (approved):

src/Utilities/ErrorHandling/FloatingPointExceptions.cpp
```diff
@@ -30,7 +31,7 @@ const int exception_flags = FE_DIVBYZERO | FE_INVALID | FE_OVERFLOW;
 #elif SPECTRE_FPE_FENV
   feenableexcept(exception_flags);
 #endif
-  ERROR("Floating point exception!");
+  ERROR_WITH("Floating point exception!", SpectreFpe);
```
knelli2: Oops, should be `ERROR_AS` with the new changes

knelli2 (changes requested):
Could you also change the PR description?

kidder (changes requested):
it isn't compiling...

nilsdeppe:
Rebased and squash in the change

nilsvu (dismissed):

knelli2 (dismissed):
Looks like we may have a couple more clang-tidy errors to add to the list to ignore. I think we can for sure add `abseil-*` since we don't use it. Unsure if we should disable `concurrency-mt-unsafe` and `misc-no-recursion`.

Either way, shouldn't hold this PR up I don't think.

tests/Unit/DataStructures/Test_FixedHashMap.cpp
```diff
@@ -128,14 +128,12 @@ void test_direction_key() {
...
-      CHECK(e.what() == get_output(dir3) + " not in FixedHashMap");
-    } catch (...) {
-      CHECK(false);
-    }
+    CHECK_THROWS_WITH(passed_map.at(dir3),
+                      Catch::Matchers::ContainsSubstring(
+                          get_output(dir3) + " not in FixedHashMap"));
+    // Can also test that the exception type is `std::out_of_range` once that's
+    // implemented:
+    // CHECK_THROWS_AS(map.at(dir3), std::out_of_range);
```
wthrowe: It's implemented, so do this or delete it.

wthrowe (changes requested):
On clang-tidy:
* `abseil-*` should go, obviously
* `misc-no-recursion` is a false positive here.  (The recursive call is in an unreachable branch.)  We can see how much it flags, either correctly or not, before deciding.
* `concurrency-mt-unsafe` is probably actually something we should be paying attention to.  (Not required to fix for this PR, certainly.)

tests/Unit/DataStructures/Test_FixedHashMap.cpp
nilsdeppe: Done and rebased.

knelli2 (approved):

wthrowe (approved):

nilsvu (approved):

kidder (approved):

PR #5470, opened 2023-09-18
Build charm `--with-production` in container

nilsvu:
## Proposed changes

Should we prefer the speedup from `--with-production` over Charm++ debug settings?

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2 (approved):
I'm personally fine with this. I think most of the error messages we would want to get are also in production (like unregistered PUP ID)

nilsdeppe (approved):

PR #5473, opened 2023-09-19
Fix bug in deserialization of old QuaternionFoTs

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


tests/Unit/Domain/FunctionsOfTime/Test_PiecewisePolynomial.cpp
```diff
@@ -179,6 +182,37 @@ void test_func_and_derivs() {
       quartic(std::array<DataVector, 5>{{c0, c1, c2, c3, c4}}));
 }
 
+void test_serialization_versioning() {
+  using Poly = FunctionsOfTime::PiecewisePolynomial<1>;
+  register_classes_with_charm<Poly>();
+  const std::unique_ptr<FunctionsOfTime::FunctionOfTime> poly(
+      new Poly(2.0, {{{3.0}, {4.0}}}, 5.0));
```
nilsvu: `std::make_unique<Poly>(...)` instead of `new`?

tests/Unit/Domain/FunctionsOfTime/Test_PiecewisePolynomial.cpp
```diff
@@ -179,6 +182,37 @@ void test_func_and_derivs() {
...
+  poly->update(5.0, {6.0}, 7.0);
+
+  // Because of the implementation-defined sign, there's no way to
+  // write char literals that won't cause narrowing errors.
+  const auto vector_char = [](auto... values) {
+    return std::vector<char>{static_cast<char>(values)...};
+  };
+
+  const auto serialization_v2 = vector_char(
```
nilsvu, line 203: Maybe add a note how this data representation can be obtained when the test is updated? I think this kind of test is a great idea, and it would be nice to make it easy for people to add these tests to more domain-related classes.

nilsvu (changes requested):

nilsvu (approved):

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -126,6 +126,8 @@ void QuaternionFunctionOfTime<MaxDeriv>::pup(PUP::er& p) {
 
       // Same for v0 and v1
       p | angle_f_of_t_;
+      expiration_time_.store(angle_f_of_t_.time_bounds()[1],
+                             std::memory_order_relaxed);
```
knelli2, line 130: For this file and `PiecewisePolynomial.cpp` I wanted to always use `std::memory_order_{acquire/release}` rather than `std::memory_order_relaxed` (even though in `pup` it doesn't really make a difference). However, I realize I missed a spot a several lines below which is presumably why you used it here, so that's my bad. I'll update this in a follow-up PR.

knelli2 (approved):

PR #5476, opened 2023-09-19
Use consistent memory order in QuatFoT

knelli2:
## Proposed changes

Follow up for #5473

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
+++ b/src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
@@ -127,7 +127,7 @@ void QuaternionFunctionOfTime<MaxDeriv>::pup(PUP::er& p) {
       // Same for v0 and v1
       p | angle_f_of_t_;
       expiration_time_.store(angle_f_of_t_.time_bounds()[1],
-                             std::memory_order_relaxed);
+                             std::memory_order_release);
 
       if (version == 0) {
         stored_quaternion_size_.store(stored_quaternions_and_times_.size(),
@@ -146,7 +146,7 @@ void QuaternionFunctionOfTime<MaxDeriv>::pup(PUP::er& p) {
         p | expiration_time_;
       } else {
         expiration_time_.store(angle_f_of_t_.time_bounds()[1],
-                               std::memory_order_relaxed);
+                               std::memory_order_release);
       }
 
       size_t size = 0;


wthrowe (approved):
Looks like the only ones that were missed.

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -146,7 +146,7 @@ void QuaternionFunctionOfTime<MaxDeriv>::pup(PUP::er& p) {
         p | expiration_time_;
       } else {
         expiration_time_.store(angle_f_of_t_.time_bounds()[1],
-                               std::memory_order_relaxed);
```
nilsvu, line 149: This passed tests because versions before 3 aren't tested?

knelli2: Are you referring to the memory order or just the fact that there was a bug in the serialization before? The memory order doesn't have an impact on the tests. But yes, I believe the tests passed before because we don't test versions before 3 which had the bug

nilsvu: Reading in v2 data before this fix failed I guess. Have you tried reading in old data? If not, I'd say it's unlikely that it will work :) Note that you could just drop support for reading in old data and error out to simplify the code, if no old data is around anymore anyway. (not asking you to do that here)

knelli2: No haven't tried reading in old data. Not sure if I know where any is. Yeah I think it's worth considering just dropping old versions. I'll think about it (not for this PR)

nilsvu (approved):

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
wthrowe: I considered dropping support in #5473 instead of fixing it, but it turns out the test of Domain includes an old data set (didn't check if it was v0 or v1).  It doesn't check enough to be useful in verifying that the code works, but that would have to be regenerated with an old version of SpECTRE if the format were dropped.

nilsvu: Ok good (I guess). That test checks the domain by equality I think, so it catches things that are included in the equality operator. Should be all public state I think?

wthrowe: Only the parts that were changed from the default value before serialization.  It didn't fail before #5473.

PR #5477, opened 2023-09-19
Disable clang-tidy trivially-copyable move check

### LINKED ISSUE #470: Silence clang-tidy warning about trivially copyable moves when that becomes possible
clang-tidy recently [added an option](https://github.com/llvm-mirror/clang-tools-extra/commit/d0bf766a6ea03e8499ca9b6f51a348387cc07b7a) to not warn on moving trivially copyable types, which is one of the most common things we NOLINT (excluding pointer arithmetic and pup functions).  It will probably be a while before we can use it, but we should remember to enable it when it becomes available (I'm guessing LLVM 6).

wthrowe:
Fixes #470 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -82,6 +82,9 @@ Checks: '*,
 -readability-static-accessed-through-instance,
 # we are okay with lower case,
 -readability-uppercase-literal-suffix,'
+CheckOptions:
+  - key: performance-move-const-arg.CheckTriviallyCopyableMove
+    value: false
 WarningsAsErrors: '*'
 # It is unclear if the header filter actually works or how to use it so
 # just include all headers


.clang-tidy
```diff
@@ -82,6 +82,9 @@ Checks: '*,
 -readability-static-accessed-through-instance,
 # we are okay with lower case,
 -readability-uppercase-literal-suffix,'
+CheckOptions:
+  - key: performance-move-const-arg.CheckTriviallyCopyableMove
+    value: 0
```
knelli2: [optional] [The docs](https://clang.llvm.org/extra/clang-tidy/checks/performance/move-const-arg.html) say to use *true* to enable it, so should we use *false* to disable it? Or is it more standard to use `0`? I wasn't able to find anything after a quick search online.

wthrowe: Looks like they changed it at some point, probably version 11 or 12 is my guess.  We're about to drop support for those anyway (and it's not clear whether that was even a behavior change or just a doc change), so I'll switch to the new form.

knelli2: Looks like it changed in [clang-12](https://releases.llvm.org/12.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/performance-move-const-arg.html) because [clang-11](https://releases.llvm.org/11.0.1/tools/clang/tools/extra/docs/clang-tidy/checks/performance-move-const-arg.html) has the old method.

knelli2 (approved):

nilsvu:
Why not just do what clang-tidy says and remove the trivial move? (The issue doesn't say)

wthrowe:
Nor does clang-tidy give any explanation of why this is bad.  The rest of the warning appears to be based on [standards from perforce](https://www.perforce.com/resources/qac/high-integrity-cpp-coding-standard/standard-library#paragraph-29026), but this part was just made up by one of the reviewers [when the code was submitted to LLVM](https://reviews.llvm.org/D12031).

nilsvu:
Did a part of your comment before the "nor" disappear?

wthrowe:
No, it was a response to your comment:

> The issue doesn't say [why we should allow this]

> Nor does clang-tidy give any explanation of why this is bad.

Standard-conforming code is allowed by default.  There needs to be a reason for a ban, not a reason to allow it.

nilsvu:
Ah ok, thanks for clarifying 👍

nilsvu (approved):

Issue #5463, opened 2023-09-15
Add Weyl scalars in volume of characteristic domain


knelli2:
As we start to do more with CCE (like eventually experiment more with initial data), it would be nice if we could observe the Weyl scalars $\Psi_0$ through $\Psi_4$ in the volume of the characteristic domain to more easily debug things. Currently, $\Psi_0$ is already implemented in `src/Evolution/Systems/Cce/NewmanPenrose.hpp` in the `VolumeWeyl<Tags::Psi0>` struct. We would basically want to just add more specializations of this struct `VolumeWeyl<Tags::Psi1>` ... and so on.

The expressions for all of the Weyl scalars are contained in equations 90 and 91 of https://arxiv.org/pdf/2007.01339.pdf.

$\Psi_1$ is already expressed in terms of the spin weighted Bondi quantities we evolve and calculate in SpECTRE ($J, K, \beta, Q, r, \partial_y, \eth, ...$) so that will be mostly translating an expression into code.

For $\Psi_2, \Psi_3, \Psi_4$ (equation 90 in the paper), the expressions aren't in terms of those spin weighted Bondi quantities. However, equations 86 and 88 have expressions for converting the existing terms in equation 90 to spin weighted Bondi quantities. So for these Weyl scalars, there will be an extra step of doing some simplifications on paper before coding up the expressions. Luckily, there is a [mathematica notebook](https://github.com/moxcodes/cce_formalism_companion) implemented by @moxcodes that does some of these simplifications already, since the expressions can get quite long. So it'd be worth taking a look at that.

So actionable steps to calculate $\Psi_1$:

- Translate equation 91b from the above paper into code.
- This will go in `src/Evolution/Systems/Cce/NewmanPenrose.hpp` in a struct called `VolumeWeyl<Tags::Psi1>`.
- All tags for the Bondi quantities can be found in `src/Evolution/Systems/Cce/Tags.hpp` except for $\eth\beta$ and $\bar{\eth}\beta$.
- These two tags will be `Spectral::Swsh::Tags::Derivative<Tags::BondiBeta, Spectral::Swsh::Tags::Eth>` and `Spectral::Swsh::Tags::Derivative<Tags::BondiBeta, Spectral::Swsh::Tags::Ethbar>`, respectively.
- I don't think the second one is implemented, so that may also have to be added.
[...]

PR #5480, opened 2023-09-20
Add calculation of $\Psi_1$ in the CCE volume

duetosymmetry:
## Proposed changes

Adding another Weyl scalar to the volume (partially addresses #5463)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/NewmanPenrose.cpp
+++ b/src/Evolution/Systems/Cce/NewmanPenrose.cpp
@@ -37,6 +37,48 @@ void weyl_psi0_impl(
              0.5 * bondi_j * (1.0 + square(bondi_k)) * dy_j * conj(dy_j)) /
                 square(bondi_k));
 }
+
+void weyl_psi1_impl(
+    const gsl::not_null<SpinWeighted<ComplexDataVector, 1>*> psi_1,
+    const SpinWeighted<ComplexDataVector, 2>& bondi_j,
+    const SpinWeighted<ComplexDataVector, 2>& dy_j,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_k,
+    const SpinWeighted<ComplexDataVector, 1>& bondi_q,
+    const SpinWeighted<ComplexDataVector, 1>& dy_q,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_r,
+    const SpinWeighted<ComplexDataVector, 1>& eth_r_divided_by_r,
+    const SpinWeighted<ComplexDataVector, 0>& dy_beta,
+    const SpinWeighted<ComplexDataVector, 1>& eth_beta,
+    const SpinWeighted<ComplexDataVector, 1>& eth_dy_beta,
+    const SpinWeighted<ComplexDataVector, 0>& one_minus_y) {
+
+  const double prefac = 1./sqrt(128.); // compile time const, but
+                                       // sqrt is not constexpr yet
+  const auto one_plus_k = 1. + bondi_k;
+  const auto eth_beta_plus_half_q = eth_beta + 0.5 * bondi_q;
+  const auto conj_j_times_dy_j = conj(bondi_j) * dy_j;
+
+  const auto inner_expr =
+    bondi_j
+    * (-2. * conj(dy_q)
+       + conj(dy_j) * (2. * eth_beta_plus_half_q
+                       + bondi_j * conj(eth_beta_plus_half_q)))
+    + one_plus_k
+    * (eth_beta_plus_half_q * (conj_j_times_dy_j
+                               - conj(conj_j_times_dy_j))
+       + 2. * (dy_q + bondi_j * conj(dy_q))
+       - one_plus_k * (2. * dy_q + dy_j * conj(eth_beta_plus_half_q)));
+
+  *psi_1 = prefac * square(one_minus_y) / (square(bondi_r) * sqrt(one_plus_k))
+    * (bondi_j * conj(eth_beta_plus_half_q)
+       - one_plus_k * eth_beta_plus_half_q
+       + one_minus_y
+       * (eth_dy_beta * one_plus_k
+          - bondi_j * conj(eth_dy_beta)
+          + dy_beta * (one_plus_k * eth_r_divided_by_r
+                       - bondi_j * conj(eth_r_divided_by_r))
+          + 0.25 * inner_expr / bondi_k));
+}
 }  // namespace
 
 void VolumeWeyl<Tags::Psi0>::apply(
@@ -49,6 +91,25 @@ void VolumeWeyl<Tags::Psi0>::apply(
     const Scalar<SpinWeighted<ComplexDataVector, 0>>& one_minus_y) {
   weyl_psi0_impl(make_not_null(&get(*psi_0)), get(bondi_j), get(dy_j),
                  get(dy_dy_j), get(bondi_k), get(bondi_r), get(one_minus_y));
+}
+
+void VolumeWeyl<Tags::Psi1>::apply(
+    const gsl::not_null<Scalar<SpinWeighted<ComplexDataVector, 1>>*> psi_1,
+    const Scalar<SpinWeighted<ComplexDataVector, 2>>& bondi_j,
+    const Scalar<SpinWeighted<ComplexDataVector, 2>>& dy_j,
+    const Scalar<SpinWeighted<ComplexDataVector, 0>>& bondi_k,
+    const Scalar<SpinWeighted<ComplexDataVector, 1>>& bondi_q,
+    const Scalar<SpinWeighted<ComplexDataVector, 1>>& dy_q,
+    const Scalar<SpinWeighted<ComplexDataVector, 0>>& bondi_r,
+    const Scalar<SpinWeighted<ComplexDataVector, 1>>& eth_r_divided_by_r,
+    const Scalar<SpinWeighted<ComplexDataVector, 0>>& dy_beta,
+    const Scalar<SpinWeighted<ComplexDataVector, 1>>& eth_beta,
+    const Scalar<SpinWeighted<ComplexDataVector, 1>>& eth_dy_beta,
+    const Scalar<SpinWeighted<ComplexDataVector, 0>>& one_minus_y) {
+  weyl_psi1_impl(make_not_null(&get(*psi_1)), get(bondi_j), get(dy_j),
+                 get(bondi_k), get(bondi_q), get(dy_q),
+                 get(bondi_r), get(eth_r_divided_by_r), get(dy_beta),
+                 get(eth_beta), get(eth_dy_beta), get(one_minus_y));
 }
 
 void TransformBondiJToCauchyCoords::apply(
--- a/src/Evolution/Systems/Cce/NewmanPenrose.hpp
+++ b/src/Evolution/Systems/Cce/NewmanPenrose.hpp
@@ -95,6 +95,93 @@ struct Psi0Compute : Tags::Psi0, db::ComputeTag {
 }  // namespace Tags
 
 /*!
+ * \brief Compute the Weyl scalar \f$\Psi_1\f$ in the volume according to the
+ * standard set of Newman-Penrose vectors.
+ *
+ * \details Our convention is \f$\Psi_1 =
+ * l^\alpha n^\beta l^\mu m^\nu C_{\alpha
+ * \beta \mu \nu}\f$.
+ *
+ * \f{align*}{
+ * \Psi_1 =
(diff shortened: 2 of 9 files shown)


duetosymmetry:
Sorry for the many force pushes, just found some typos between my LaTeX and the code. Should be ok now.

duetosymmetry:
gcc-11 unit tests are failing for an unrelated reason: While building EvolveGhBinaryBlackHole, got:
```
g++-11: fatal error: Killed signal terminated program cc1plus
compilation terminated.
```

knelli2:
Yeah building the BBH executable on gcc-11 Release keeps running out of memory on the CI runners. Don't worry about it.

src/Evolution/Systems/Cce/NewmanPenrose.cpp
```diff
@@ -37,6 +37,48 @@ void weyl_psi0_impl(
...
+    * (-2. * conj(dy_q)
+       + conj(dy_j) * (2. * eth_beta_plus_half_q
+                       + bondi_j * conj(eth_beta_plus_half_q)))
+    + one_plus_k
+    * (eth_beta_plus_half_q * (conj_j_times_dy_j
+                               - conj(conj_j_times_dy_j))
+       + 2. * (dy_q + bondi_j * conj(dy_q))
+       - one_plus_k * (2. * dy_q + dy_j * conj(eth_beta_plus_half_q)));
+
+  *psi_1 = prefac * square(one_minus_y) / square(bondi_r) / sqrt(one_plus_k)
```
knelli2: `square(one_minus_y) / (square(bondi_r) * sqrt(one_plus_k))`

src/Evolution/Systems/Cce/NewmanPenrose.cpp
```diff
@@ -37,6 +37,48 @@ void weyl_psi0_impl(
...
+    * (4. * bondi_j * conj(eth_beta_plus_half_q)
+       - 4. * one_plus_k * eth_beta_plus_half_q
+       + one_minus_y
+       * (4. * eth_dy_beta * one_plus_k
+          - 4. * bondi_j * conj(eth_dy_beta)
+          + 4. * dy_beta * (one_plus_k * eth_r_divided_by_r
+                            - bondi_j * conj(eth_r_divided_by_r))
+          + inner_expr / bondi_k));
```
knelli2: There's a `4.0` multiplying every term except for this last one. Maybe factor the `4.0` out into the prefactor and multiply this last term by `0.25`? Less overall calculations.

src/Evolution/Systems/Cce/NewmanPenrose.hpp
```diff
@@ -69,6 +69,65 @@ struct VolumeWeyl<Tags::Psi0> {
...
+ * \beta \mu \nu}\f$.
+ *
+ * \f{align*}{
+ * \Psi_1 =
+ * &\frac{(1-y)^2}{\sqrt{2048} \sqrt{1 + K} R^2}\Bigg\{
+ * 4 J(\bar{\eth }\beta + \tfrac{1}{2} \bar{Q})
+ * - 4(1 + K) (\eth \beta + \tfrac{1}{2} Q) \\
+ * &+(1-y)\Bigg[
+ * 4 (1 + K)\eth\partial_{y}\beta -4 J\bar{\eth }\partial_{y}\beta
+ * + 4 \left(- J \frac{\bar{\eth }' R}{R} + (1 + K) \frac{\eth ' R}{R}\right)
```
knelli2: Why do these $\eth$ s have primes?

src/Evolution/Systems/Cce/NewmanPenrose.hpp
```diff
@@ -69,6 +69,65 @@ struct VolumeWeyl<Tags::Psi0> {
...
+ * 2 (\partial_{y}Q + J \partial_{y}\bar{Q}) + (\bar{J} \partial_{y}J - J
+ * \partial_{y}\bar{J}) (\eth\beta + \tfrac{1}{2} Q)
+ * \\
+ * &\qquad\quad - (1+K)
+ * [2 \partial_{y}Q + \partial _{y}J (\bar{\eth }\beta + \tfrac{1}{2} \bar{Q})]
+ * \Big\}\Bigg)\Bigg]\Bigg\}
+ * \f}
+ */
+template <>
+struct VolumeWeyl<Tags::Psi1> {
```
knelli2, line 128: This needs to be tested. In `tests/Unit/Evolution/Systems/Cce/Test_NewmanPenrose.cpp` we calculate everything for $\Psi_0$ using a Kerr-Schild solution and check that we get zero. You can probably just add the $\Psi_1$ calculation to that.

knelli2 (changes requested):
I only checked that the code matches the dox. Didn't go through and do the calculation myself yet to confirm the expression is correct.

We may be able to optimize the function a bit more than what I suggested below, but I don't think that's necessary at the moment.

src/Evolution/Systems/Cce/NewmanPenrose.hpp
duetosymmetry: This notation is coming from @moxcodes's Mma notebook, and it distinguishes $\eth$ in Bondi-like coordinates from $\breve\eth$ in numerical coordinates (see Eq. (46c) of 2007.01339). Now that I think about it, it's possible that every $\eth$ in this expression should actually be a $\breve\eth$... Does anybody know off the top of their head? Maybe @Sizheng-Ma?

src/Evolution/Systems/Cce/NewmanPenrose.hpp
duetosymmetry: Oops, that's what I get for only grepping inside of `src/`! Ok, I kind of don't want to duplicate all the setup for the psi0 test (which itself is doing a bunch of duplication of work from Test_GaugeTransformBoundaryData). Please advise if I should just ignore this and just duplicate, or if there's a better approach.

knelli2: Since we're going to need to test the other Weyl scalars, and those will require similar spin weighted bondi variables as $\Psi_1$, I'd say make a new function similar to `compute_psi0_of_bh_on_wt()` in `Test_NewmanPenrose.cpp`, except this is only concerned with volume weyl scalars (doesn't do all the boundary stuff necessary for `compute_psi0_of_bh_on_wt()`). Then test $\Psi_0$ and $\Psi_1$ with this new function for now. You'll have to initialize some of the volume variables before hand. I'd have to take a look at how to do that as I don't know off the top of my head

src/Evolution/Systems/Cce/NewmanPenrose.hpp
Sizheng-Ma: In the code, `Eth` refers to $\eth$, instead of $\breve{\eth}$ (100% sure).

Sizheng-Ma: Just to clarify, here `Eth` means tags, rather than the operation.

Sizheng-Ma: Also, please avoid using primes because that refers to Cauchy coordinates --- There are four coordinate systems, so it might be good to stick to the same notation.

duetosymmetry: Then I'm confused about why in Jordan's paper, in Eq. (46c), he has $(\breve\eth R)/R$, whereas in the code, the quantity is named `eth_r_divided_by_r`. Do we happen to have $\eth R = \breve\eth R$, since `R` does not depend on `y`?

Sizheng-Ma: Yes, exactly. `R` (Bondi radius of the worldtube) is a surface variable. Eq. (46c) in https://arxiv.org/pdf/2007.01339.pdf shows that $\eth R=\breve{\eth} R$. Similarly, $\partial_u R = \partial_{\breve{u}}R$.

duetosymmetry: 🤦  doh, I should have figured that out earlier, it's been a long day... Thanks, @Sizheng-Ma. I'll update the documentation to get rid of the prime symbols.

src/Evolution/Systems/Cce/NewmanPenrose.hpp
duetosymmetry: Is it possible to factor out all the common code into another helper in `tests/Unit/Helpers/Evolution/Systems/Cce/BoundaryTestHelpers.hpp`? I'm not yet sure what this would look like.

knelli2: Or since everything will still be tested in the same `Test_NewmanPenrose.cpp` file, just factor into free functions in that file?

duetosymmetry: I was suggesting `BoundaryTestHelpers.hpp` because a bunch of @Sizheng-Ma's test is copied from a different module, `tests/Unit/Evolution/Systems/Cce/Test_GaugeTransformBoundaryData.cpp`, and both of these modules use helpers from `BoundaryTestHelpers.hpp`; so I wanted to consolidate the code that's common to both tests. But I defer to your/Sizheng's advice for where the common code should live.

knelli2: Ok yeah consolidation is good. Since, with your additions, the boundary helper file won't be used for boundary stuff only anymore, why don't we rename it `BoundaryTestHelpers.hpp` -> `VolumeAndBoundaryTestHelpers.hpp`. Then you can factor out all the common code in there that you need.

knelli2:
@duetosymmetry I believe your student Subhayu was going to take this over?

duetosymmetry:
@subhob have you taken a look at this at any point?

SubhoB:
> @SubhoB have you taken a look at this at any point?

Hey @knelli2 @duetosymmetry yeah, I think Kyle and I decided at the collab meeting to look over this together, it needs some minor changes as far as I remember, I am reaching out to him on Slack

duetosymmetry:
TODO: Add Psi1 to tests/Unit/Evolution/Systems/Cce/Events/Test_ObserveFields.cpp .

duetosymmetry:
Psi1 has now been added to tests/Unit/Evolution/Systems/Cce/Events/Test_ObserveFields.cpp .

duetosymmetry:
Unrelated test timeouts only in configuration (clang-14, Debug, ON, OFF, JEMALLOC, 2)
```
The following tests FAILED:
	 70 - InputFiles.Xcts.BinaryBlackHole.yaml.parse (Timeout)
	 71 - InputFiles.Xcts.HeadOnBns.yaml.parse (Timeout)
```

tests/Unit/Helpers/Evolution/Systems/Cce/VolumeTestHelpers.hpp
```diff
@@ -116,8 +116,18 @@ using matching_variables_tag = ::Tags::Variables<
 using volume_spin_weighted_variables_tag = ::Tags::Variables<
     tmpl::list<Tags::BondiJ, Tags::Dy<Tags::BondiJ>, Tags::BondiJCauchyView,
                Tags::Dy<Tags::BondiJCauchyView>,
+               Tags::Dy<Tags::Dy<Tags::BondiJ>>,
                Tags::Dy<Tags::Dy<Tags::BondiJCauchyView>>, Tags::Psi0Match,
-               Tags::OneMinusY, Tags::Dy<Tags::Psi0Match>, Tags::BondiU>>;
+               Tags::OneMinusY, Tags::Dy<Tags::Psi0Match>, Tags::BondiU,
```
knelli2: I think these changes can be dropped now since you're not modifying the volume test?

knelli2 (changes requested):
You can squash all the Psi1 commits into a single commit

duetosymmetry:
Do you want to keep the refactoring commit separate, i.e. two commits in this PR? Or squash all commits in this PR into one?

knelli2:
Yes, keep the refactoring commit separate so there are two commits.

tests/Unit/Helpers/Evolution/Systems/Cce/VolumeTestHelpers.hpp
knelli2: Ping on this

duetosymmetry: Ah shoot sorry, my rebase was too simplistic!

duetosymmetry: Done!

knelli2 (approved):
Thanks for doing this!

PR #5484, opened 2023-09-21
Only allow one MutableCache callback per ArrayComponentId

knelli2:
## Proposed changes

Currently, whenever an element checks if a mutable cache tag is ready and it isn't, it registers a callback with the mutable cache. However, if an element is restarted for reasons other than the mutable cache tag being updated (like neighbor communication) then an additional callback is registered. This could potentially have a ton of callbacks registered for the same element.

Instead, this PR will only register the first callback for the element. If more callbacks would be registered, the mutable cache checks if there's already a callback registered, and if so, doesn't register another. This has the side effect, that the first callback to be registered is the winner and all subsequent ones will be ignored, so we can't have different types of callbacks for the same element (perform algorithm vs simple action)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Parallel/ArrayComponentId.hpp
```diff
@@ -57,6 +59,12 @@ bool operator!=(const ArrayComponentId& lhs, const ArrayComponentId& rhs);
 
 std::ostream& operator<<(std::ostream& os,
                          const ArrayComponentId& array_component_id);
+
+template <typename ParallelComponent, typename ArrayIndexType>
```
nilsdeppe, line 66: add dox so it shows up?

Also, `ArrayComponentId` is no longer in `ObserverGroup` (Doxygen) so update that in previous commit

src/Parallel/GlobalCache.hpp
```diff
@@ -277,9 +280,15 @@ class GlobalCache
    *
    * \warning The `function` may be called twice so it should not modify any
    * state in its scope.
+   * \n
```
nilsdeppe: doesn't just a blank line in the comment work? (I haven't tried `\n` in doxygen)

src/Parallel/GlobalCache.hpp
```diff
@@ -486,22 +495,26 @@ bool GlobalCache<Metavariables>::mutable_cache_item_is_ready(
...
     // This function returns true if no callback was registered and false if one
     // was registered.
-    return not callback_was_registered();
+    const bool cache_item_is_ready = not callback_was_registered();
+    if (cache_item_is_ready) {
+      const std::lock_guard<std::mutex> lock(mutex);
+      std::unordered_map<Parallel::ArrayComponentId, std::unique_ptr<Callback>>&
+          callbacks = std::get<1>(tuples::get<tag>(mutable_global_cache_));
+
+      if (callbacks.count(array_component_id) == 1) {
```
nilsdeppe: You don't need this check, I think `erase` just removes the element if it exists

tests/Unit/Framework/Tests/Test_ActionTesting.cpp
```diff
@@ -1171,21 +1172,19 @@ SPECTRE_TEST_CASE("Unit.ActionTesting.MutableGlobalCache", "[Unit]") {
...
+  const auto array_component_id =
+      Parallel::make_array_component_id<component>(static_cast<int>(0));
 
   CHECK(Parallel::mutable_cache_item_is_ready<CacheTag>(
-      cache, [](const int /*value*/) {
+      cache, array_component_id, [](const int /*value*/) {
         return std::unique_ptr<Parallel::Callback>{};
       }));
 
+  // We cannot have two different types of callbacks at once in the mutable
```
nilsdeppe, line 1183: Should we add a `\warning` somewhere to the documentation that states this clearly and gives a brief summary of what's needed to generalize?

nilsdeppe (changes requested):
LGTM, a few suggestions. I'm happy for you to squash immediately.

src/Parallel/GlobalCache.hpp
knelli2: No, for `\warning` and `\note`, if you only put a blank line between two `\warning`s back to back, it'll just squash both into a single warning.

nilsvu:
@knelli2 could you put some context/reasoning in the PR description?

src/Parallel/GlobalCache.hpp
nilsdeppe: Okay, sounds good!

wthrowe (commented):

src/Parallel/GlobalCache.hpp
knelli2: So `\n` didn't work, but adding `\parblock ... \endparblock` did work.

nilsdeppe (approved):

PR #5486, opened 2023-09-21
List tested compiler versions in docs

nilsvu:
## Proposed changes

Be more precise about supported versions.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Installation/Installation.md
+++ b/docs/Installation/Installation.md
@@ -57,10 +57,12 @@ all of these dependencies.
 
 #### Required:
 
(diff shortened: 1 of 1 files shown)


wthrowe:
I think we should still list the supported versions in addition to the tested versions.  Neither clang-12 nor clang-16 is tested, but you're going to get very different responses from us if you have an issue related to one versus the other.

nilsvu:
You're right. I think I just close this and we keep listing the supported versions.

PR #5491, opened 2023-09-22
Isolate multithreading in FoTs

wthrowe:
This reduces the number of places where thread-safety problems could arise, and should also improve performance after a large number of updates (although I have not tested this empirically).

Also makes obtaining some information needed for substep methods with control systems easier.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/FunctionsOfTime/ThreadsafeList.hpp
```diff
@@ -0,0 +1,317 @@
...
+  std::unique_ptr<Interval> interval_list_{};
+  alignas(64) std::atomic<const Interval*> most_recent_interval_{};
+  // Pad memory to avoid false-sharing when accessing most_recent_interval_
+  // NOLINTNEXTLINE(modernize-avoid-c-arrays)
+  char unused_padding_[64 - (sizeof(std::atomic<const Interval*>) % 64)] = {};
+};
+
+template <typename T>
+void ThreadsafeList<T>::Interval::pup(PUP::er& p) {
+  p | expiration;
```
knelli2, line 148: Since this is going to be stored inside a function of time that's going to be serialized into an h5 file, should we add a version?
```cpp
  size_t version = 0;
  p | version;
  // Remember to increment the version number when making changes to this
  // function. Retain support for unpacking data written by previous versions
  // whenever possible. See `Domain` docs for details.
```

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
```diff
@@ -0,0 +1,317 @@
...
+    ERROR("Tried to insert at time "
+          << update_time << ", which is not the old expiration time "
+          << old_expiration);
+  }
+  if (expiration_time <= update_time) {
+    ERROR("Expiration time " << expiration_time << " is not after update time "
+                             << update_time);
+  }
+  std::unique_ptr<Interval> new_interval(new Interval{
+      expiration_time, std::move(data), std::move(interval_list_)});
```
knelli2, line 203: Maybe `std::make_unique`? Same for the rest of the `new`s. I suggest this only for consistency because we rarely ever use `new` itself, and always `std::make_unique`

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
```diff
@@ -0,0 +1,317 @@
...
+  if (interval == nullptr) {
+    ERROR("Attempt to access an empty function of time.");
+  }
+  if (time > interval->expiration or
+      (interval_after_boundary and time == interval->expiration)) {
+    ERROR("Attempt to evaluate at time "
+          << time << ", which is after the expiration time "
+          << interval->expiration);
+  }
+  for (;;) {
```
knelli2: I think some docs for this loop would be helpful. Something like
```
Recursively loop through previous intervals until we find the proper interval
that satisfies the `interval_after_boundary` condition. Returns the initial interval
if time == initial_time_
```

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
```diff
@@ -0,0 +1,317 @@
...
+namespace domain::FunctionsOfTime::FunctionOfTimeHelpers {
+/// A list of time intervals that allows safe access to existing
+/// elements even during modification.
+///
+/// Concurrent modification is not supported.  All non-modifying
+/// operations except for serialization can be safely performed in
+/// parallel with each other and with `insert` and will return a
+/// consistent state.
+template <typename T>
+class ThreadsafeList {
```
knelli2, line 29: Question: While this is currently only used in the functions of time, do you think we should move this to `DataStructures` so it's available generally? Would we have another use case for this in the future?

tests/Unit/Domain/FunctionsOfTime/Test_ThreadsafeList.cpp
```diff
@@ -0,0 +1,216 @@
...
+  CHECK(list.initial_time() == 1.0);
+  CHECK(list.expiration_time() == 1.0);
+
+  CHECK(list == ThreadsafeList<int>(1.0));
+  CHECK_FALSE(list != ThreadsafeList<int>(1.0));
+  CHECK(list != ThreadsafeList<int>(2.0));
+  CHECK_FALSE(list == ThreadsafeList<int>(2.0));
+
+  {
+    auto copy = list;
```
knelli2: Maybe `test_{copy,move}_semantics()` for these?

tests/Unit/Domain/FunctionsOfTime/Test_ThreadsafeList.cpp
```diff
@@ -0,0 +1,216 @@
...
+  }
+
+  {
+    const ThreadsafeList<int> list2(1.0);
+    CHECK(list2.begin() == list2.end());
+    CHECK_FALSE(list2.begin() != list2.end());
+  }
+  // Test iterators
+  {
+    decltype(list)::iterator it = list.begin();
```
knelli2, line 107: There's also `test_iterators()` and `test_reverse_iterators()` in the test helpers. Would those be useful?

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -124,152 +72,137 @@ PiecewisePolynomial<MaxDeriv>::func_and_derivs(const double t) const {
...
+  ASSERT(p.isUnpacking(), "Can't serialize old version");
+
+  std::vector<LegacyStoredInfo<MaxDeriv + 1>> info_vector{};
+  double expiration_time{};
+
+  // For versions 0 and 1, we stored the data first, then expiration time,
+  // then possibly the size
+  if (version <= 1) {
+    if (version == 0) {
+      // Version 0 had a std::vector rather than a std::list
```
knelli2: Remove `rather than a std::list` (this function doesn't know anything about lists). Same in QuatFoT

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -124,152 +72,137 @@ PiecewisePolynomial<MaxDeriv>::func_and_derivs(const double t) const {
...
+      p | info_vector;
+    } else {
+      // Version 1 had a std::deque rather than a std::list
```
knelli2: Again remove `rather than a std::list`

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -282,17 +215,28 @@ template <size_t MaxDeriv>
...
-  const size_t writable_size =
-      piecewise_polynomial.deriv_info_size_.load(std::memory_order_acquire);
-  auto it = piecewise_polynomial.deriv_info_at_update_times_.begin();
-  // Can't use .end() because of thread-safety and don't want to do expensive
-  // std::next() call, so we just loop over allowed size while incrementing an
-  // iterator
-  for (size_t i = 0; i < writable_size; i++, it++) {
-    os << *it;
-    // writable_size guaranteed to be >= 1. No need to worry about negative
-    // size_t
```
knelli2, line 294: Can you keep this comment in?

src/Utilities/Serialization/PupBoost.hpp
```diff
@@ -53,4 +54,23 @@ template <class T, size_t N>
...
+  T component_1 = quaternion.R_component_1();
+  T component_2 = quaternion.R_component_2();
+  T component_3 = quaternion.R_component_3();
+  T component_4 = quaternion.R_component_4();
+  p | component_1;
+  p | component_2;
+  p | component_3;
+  p | component_4;
+  quaternion = boost::math::quaternion<T>(component_1, component_2, component_3,
+                                          component_4);
```
knelli2: Should this be wrapped in `if (p.isUnpacking())`? Seems a bit odd to modify an existing object that you are just packing.

knelli2 (changes requested):
Thanks for doing this! This looks a lot simpler than before

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
wthrowe: I debated about that.  In the end, I decided that having a bunch of stuff tied to expiration times made this FoT specific.  Can move it if you think otherwise.  (Also easy to move later.)

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
wthrowe: I'll add it to the list itself.  I don't want to add it to every interval.

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
wthrowe: `make_unique` doesn't work on aggregates until C++20.

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
wthrowe: I'll come up with something.  This function doesn't reference `initial_time_`, though, so I'm not sure where you got that.

tests/Unit/Domain/FunctionsOfTime/Test_ThreadsafeList.cpp
wthrowe: No, they won't work on this class.

tests/Unit/Domain/FunctionsOfTime/Test_ThreadsafeList.cpp
wthrowe: I don't like that they don't test `!=`, but that really means I should fix them to do that instead of avoiding them.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
wthrowe: `writable_size` doesn't exist anymore.  I also don't really understand what the comment was trying to tell me.

wthrowe (commented):
Missed one.

tests/Unit/Domain/FunctionsOfTime/Test_ThreadsafeList.cpp
knelli2: Yeah if we already have them I think it'd be good to use. Just add another small commit

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
knelli2: You have
```cpp
    if (entry != iters.size() - 1) {
      os << "\n";
    }
```
and if `iters.size() == 0` this is bad. But we know that can't happen for other reasons. So replace `writable_size` with `iters.size()` in the comment

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
knelli2: Yeah the expiration time makes it not super general. I think it's alright here for now.

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
knelli2: Maybe add a comment that this can be changed once we switch to C++20 since we might be doing that soon.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
wthrowe: Why would that be bad?  If this was a `<` test or similar it could be confusing, but `!=` does exactly what you'd expect.

Somewhat unrelated: Now that I look at it more, I should probably comment about why we know `std::next([...].begin())` is OK elsewhere in the function.

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
```diff
@@ -0,0 +1,317 @@
...
+
+template <typename T>
+double ThreadsafeList<T>::expiration_after(const double time) const {
+  return find_interval(time, true).expiration;
+}
+
+template <typename T>
+void ThreadsafeList<T>::pup(PUP::er& p) {
+  p | initial_time_;
+  p | interval_list_;
```
wthrowe: This needs to be fixed to be threadsafe.

knelli2: I think it's safe to assume that if we are serializing or deserializing, we won't be updating/inserting. The only operation that has to be threadsafe is sizing since that can be done at any time. However, we should discuss this to confirm.

wthrowe: We serialize during IO.
https://github.com/sxs-collaboration/spectre/blob/04376d11fdd83829d72508ca1db9f7674fb0c148/src/IO/Observer/VolumeActions.hpp#L378

wthrowe:
Several fixups and one new commit "Test move and != in test_copy_semantics".

src/Domain/FunctionsOfTime/ThreadsafeList.hpp
```diff
@@ -233,11 +235,35 @@ double ThreadsafeList<T>::expiration_after(const double time) const {
...
+  // whenever possible. See `Domain` docs for details.
+
+  if (version != 0) {
+    ERROR("Unrecognized version " << version);
+  }
+
   p | initial_time_;
-  p | interval_list_;
   if (p.isUnpacking()) {
+    bool empty{};
```
knelli2: Can you add a comment here that the reason we can't just do `p | interval_list_;` is because of thread-safety?

knelli2 (commented):
One small thing. You can squash everything together

knelli2 (approved):

PR #5494, opened 2023-09-23
Remove some catastrophic cancellation in prim recovery

nilsdeppe:
## Proposed changes

I'm still worried about Palenzuela+ and NewmanHamlin, but the Kastaun schemes are better anyway. With the Kastaun scheme I'm a little worried that it couldn't be improved in some limits since it does use enthalpy and the Lorentz factor, so low-density low-velocity could see precision loss, but I haven't looked at this in detail.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
@@ -69,17 +69,15 @@ bool PrimitiveFromConservative<OrderedListOfPrimitiveRecoverySchemes,
   for (size_t i = 0; i < 3; ++i) {
     magnetic_field->get(i) = tilde_b.get(i) / get(sqrt_det_spatial_metric);
   }
-  const size_t size = get<0>(tilde_b).size();
+  const size_t number_of_points = get<0>(tilde_b).size();
   Variables<
       tmpl::list<::Tags::TempScalar<0>, ::Tags::TempScalar<1>,
                  ::Tags::TempScalar<2>, ::Tags::TempScalar<3>,
                  ::Tags::TempScalar<4>, ::Tags::TempI<5, 3, Frame::Inertial>>>
-      temp_buffer(size);
+      temp_buffer(number_of_points);
 
-  DataVector& total_energy_density =
-      get(get<::Tags::TempScalar<0>>(temp_buffer));
-  total_energy_density =
-      (get(tilde_tau) + get(tilde_d)) / get(sqrt_det_spatial_metric);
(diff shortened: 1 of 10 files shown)


src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
```diff
@@ -180,10 +180,8 @@ bool PrimitiveFromConservative<OrderedListOfPrimitiveRecoverySchemes,
 
       // Check consistency
       if (use_hydro_optimization and
-          equal_within_roundoff(
-              tau[s] + rest_mass_density_times_lorentz_factor[s],
-              get(magnetic_field_squared)[s] * 0.5 + tau[s] +
-                  rest_mass_density_times_lorentz_factor[s])) {
+          (get(magnetic_field_squared)[s] <
+           100.0 * std::numeric_limits<double>::epsilon() * tau[s])) {
```
nilsdeppe, line 184: @ermost is this the comparison I should be making? How large the b-energy is compared to the non-rest mass energy?

ermost: I think this is fine. This should be a more conservative bound, since in the relevant regions rho*epsilon < rho.

ermost:
I think for KastaunHydro this should be fine, for KastaunMHD I'm not sure, re: precision loss.

Newman-Hamlin with tables will likely still have to compute an intermediate eps from h, since only h and p are known in the iteration, and eps cannot in general be recovered from h and p alone when using a table.

I agree that Kastaun should be the way to go. I think they are more robust anyway.

src/Evolution/Systems/GrMhd/ValenciaDivClean/PrimitiveFromConservative.cpp
nilsdeppe: Okay, great, thank you!

nilsdeppe:
Yea, I'm hoping @wthrowe can take a look at the KastaunMHD once the time steppers stuff he's working on now is done. He's really good at figuring out ways to get rid of catastrophic precision loss :) The KastaunMHD paper does talk about this, so clearly they thought about this, but an independent look could be good. I'd really like to only have (h-1) show up in recover to eliminate precision loss from `h=1+small`, but it's not obvious to me that's possible.

Yea, and with NewmanHamlin there's this cubed root term that basically makes catastrophic precision loss really difficult to avoid, unfortunately. So 🤞 KastaunMHD + entropy switch will be enough :)

kidder (dismissed):

kidder (approved):

PR #5495, opened 2023-09-24
Work around boost failure to build in clang 15

geoffrey4444:
## Proposed changes

[Workaround](https://github.com/boostorg/functional/pull/21) for boost not building with clang 15: adds `-D_HAS_AUTO_PTR_ETC=0` to `CMAKE_CXXX_FLAGS` if clang version >= 15.0.0. This fix is necessary to build with latest Xcode/AppleClang.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupBoost.cmake
+++ b/cmake/SetupBoost.cmake
@@ -120,3 +120,13 @@ set_property(TARGET Boost::boost
   APPEND PROPERTY
   INTERFACE_COMPILE_DEFINITIONS
   $<$<COMPILE_LANGUAGE:CXX>:${BOOST_MULTI_ARRAY_TYPES_HEADER_GUARD}>)
+
+# Work around boost not building with clang 15 (fixed in boost 1.83.0)
+# (https://github.com/boostorg/functional/pull/21)
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND BOOST_VERSION VERSION_LESS 1.83.0)
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "15.0.0" OR
+      CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "15.0.0")
+        target_compile_definitions(Boost::boost INTERFACE _HAS_AUTO_PTR_ETC=0)
+  endif()
+endif()
+


cmake/SetupCxxFlags.cmake
```diff
@@ -8,6 +8,16 @@ option(OVERRIDE_ARCH "The architecture to use. Default is native." OFF)
...
+# Work around boost not building with clang 15
+# (https://github.com/boostorg/functional/pull/21)
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "15.0.0")
+    set(CMAKE_CXX_FLAGS "-D_HAS_AUTO_PTR_ETC=0 ${CMAKE_CXX_FLAGS}")
+  elseif(CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "15.0.0")
+    set(CMAKE_CXX_FLAGS "-D_HAS_AUTO_PTR_ETC=0 ${CMAKE_CXX_FLAGS}")
+  endif()
+endif()
+
```
geoffrey4444: I use if/elseif here, even though it duplicates a line, to avoid lines over 80 characters. I tried continuing the if line with \, but make complained

wthrowe: I believe CMake ignores newlines outside of strings (except as generic whitespace), so you should be able to wrap it just like C++.

(To any other readers: The previous comment contains a backslash before the last comma, but GitHub isn't displaying it.)

cmake/SetupCxxFlags.cmake
```diff
@@ -8,6 +8,16 @@ option(OVERRIDE_ARCH "The architecture to use. Default is native." OFF)
 option(ENABLE_SPECTRE_DEBUG "Enable ASSERTs and other SPECTRE_DEBUG options"
   OFF)
 
+# Work around boost not building with clang 15
+# (https://github.com/boostorg/functional/pull/21)
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "15.0.0")
+    set(CMAKE_CXX_FLAGS "-D_HAS_AUTO_PTR_ETC=0 ${CMAKE_CXX_FLAGS}")
+  elseif(CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "15.0.0")
+    set(CMAKE_CXX_FLAGS "-D_HAS_AUTO_PTR_ETC=0 ${CMAKE_CXX_FLAGS}")
```
nilsvu: Add this to SetupBoost.cmake please, to the boost target using `target_compile_definitions`.

nilsvu (changes requested):

geoffrey4444:
Ready for another look when you get a chance...thanks for the reviews! It's such a short PR that I squashed immediately.

nilsvu (changes requested):
This only became a problem with recent versions of boost (the code builds fine with boost 1.74 and clang 15 in the container, though I don't see an obvious reason why). It is fixed in boost 1.83. Therefore I suggest to add the condition `BOOST_VERSION VERSION_LESS 1.83.0`.

geoffrey4444:
> This only became a problem with recent versions of boost (the code builds fine with boost 1.74 and clang 15 in the container, though I don't see an obvious reason why). It is fixed in boost 1.83. Therefore I suggest to add the condition `BOOST_VERSION VERSION_LESS 1.83.0`.

Note that brew has not yet updated boost to 1.83.0, so until they do, this workaround still matters for macOS users running the latest AppleClang

cmake/SetupBoost.cmake
```diff
@@ -120,3 +120,13 @@ set_property(TARGET Boost::boost
   APPEND PROPERTY
   INTERFACE_COMPILE_DEFINITIONS
   $<$<COMPILE_LANGUAGE:CXX>:${BOOST_MULTI_ARRAY_TYPES_HEADER_GUARD}>)
+
+# Work around boost not building with clang 15 (fixed in boost 1.83.0)
+# (https://github.com/boostorg/functional/pull/21)
+if (CMAKE_CXX_COMPILER_ID MATCHES "Clang" AND BOOST_VERSION VERSION_LESS 1.83.0)
+  if(CMAKE_CXX_COMPILER_VERSION VERSION_GREATER "15.0.0" OR
+      CMAKE_CXX_COMPILER_VERSION VERSION_EQUAL "15.0.0")
```
nilsvu, line 128: Note that `GREATER_EQUAL` also exists, and this could also be combined with the above `if` (not requesting this).

nilsvu (approved):

nilsvu:
Thanks for fixing this @geoffrey4444 !

PR #5499, opened 2023-09-25
Remove layer of autogenerated code from executables

nilsvu:
## Proposed changes

Define executables in a cpp file instead of auto-generating that cpp file. This removes the need for a `Fwd.hpp` file, removes static variables with special names, and makes it (IMO) much easier to understand how an executable is compiled (there's a cpp file with a main function now).

As a side product this PR also makes executable headers compatible with clang-tidy because they don't contain undefined symbols like `metavariables` anymore.

### Upgrade instructions

If you're adding or editing executables in the build system, apply these changes (see one of the existing executables for an example):

- `add_spectre_parallel_executable` was deleted. Use `add_spectre_executable` instead and link `Charmxx::main`.
- Delete the `Fwd.hpp` file.
- Add a cpp file with a main function (specifically, a `CkRegisterMainModule()` function).
- Move the `charm_init_node_funcs` and `charm_init_proc_funcs` to the call to `Parallel::charmxx::register_init_node_and_proc` in the cpp file. Remove the error handling functions from these lists because they are registered automatically.
- If needed, pass template parameters for the metavars from CMake to the cpp file with `target_compile_definitions`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Elliptic/Executables/Punctures/SolvePunctures.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+extern "C" void CkRegisterMainModule() {
+  Parallel::charmxx::register_main_module<Metavariables>();
+  Parallel::charmxx::register_init_node_and_proc(
+      {&domain::creators::register_derived_with_charm,
+       &register_derived_classes_with_charm<
+           Metavariables::solver::schwarz_smoother::subdomain_solver>,
+       &elliptic::subdomain_preconditioners::register_derived_with_charm,
+       &register_factory_classes_with_charm<Metavariables>},
+      {});
```
nilsdeppe, line 21: why not add the init_proc? I just noticed this on this file, but probably happened in others too

nilsdeppe (changes requested):

nilsdeppe:
I did not get through all the files...

src/Elliptic/Executables/Punctures/SolvePunctures.cpp
nilsvu: The error handling is added by default (see CharmMain.tpp). We dont have any other proc init functions.

nilsdeppe (approved):
LGTM, please squash :) Thanks for doing this!

PR #5503, opened 2023-09-27
Optimize interpolation onto sphere targets

knelli2:
## Proposed changes

Due to the internals of the `InterpolateWithoutInterpComponent` event, on every element we find the block logical coordinates of *all* target points. For the sphere target this could be a significant amount depending on the $\ell$ of the surface and the number of concentric surfaces. For a sphere target of $\ell = 16$ and 4 concentric surfaces, this amounted to a 3-25x slowdown of a BNS depending on how frequent the event was run.

In order to optimize this, the sphere target is handled specially in `InterpolateWithoutInterpComponent`. Instead of getting block logical coords for every point of the sphere target on every element, we do some checks first:

1. Check if any radii of any of the spheres are in our element. If not, then we end here and don't do an interpolation.
2. If our element contains any number of radii, build a bounding box in x, y, and z.
3. Check all points of the radii in our element. If a point is within this bounding box, get its block logical coordinates. If not, skip that point
4. Once we have all the block logical coords of the points in our element, proceed as before with the interpolation

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

In order to make this optimization, a feature had to be bypassed for the sphere target. The `InterpolateWithoutInterpComponent` event calls the `InterpolationTargetVarsFromElement` simple action, passing along the block logical coordinates. The first time `InterpolationTargetVarsFromElement` is run (for a given temporal id), it originally did a check for points outside the computational domain (invalid points). It was able to do so just once because every element computed the same block logical coordinates for all the target points, so only the first was necessary to do this check. An invalid point was represented by a `std::nullopt` rather than the block logical coordinates. So `InterpolationTargetVarsFromElement` checked for `std::nullopt`s, and if there were any, filled the variables of those specific points with NaNs (technically what value to fill with is a compile time choice, but it's almost always NaNs).

However, for the sphere target, we aren't computing the block logical coords for all target points on every element, yet we still need to pass along a vector the length of the all the target points. But with the sphere target, we only populate this vector with the block logical coords of the target points within out element. Thus, in this way, the `std::nullopt` for the rest of the target points doesn't represent invalid points, just points outside of this element. So the first vector of block logical coords that arrives at `InterpolationTargetVarsFromElement` can't be used to check for invalid points (and neither can the rest of the vectors sent from the rest of the elements).

Thus, checking for invalid points is the feature that is bypassed when using the sphere target now. This is a known bug/issue, but is only a problem if a user has a sphere target and places one of the radii outside the computational domain. I'll create a GitHub issue once this is merged, but I felt the performance gain that we get from this optimization was important enough to put off this subtlety.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Events/InterpolateWithoutInterpComponent.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Events/InterpolateWithoutInterpComponent.hpp
@@ -3,32 +3,52 @@
 
 #pragma once
 
+#include <array>
 #include <cstddef>
+#include <limits>
+#include <optional>
 #include <pup.h>
+#include <set>
+#include <utility>
+#include <vector>
 
 #include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/IdPair.hpp"
+#include "DataStructures/Index.hpp"
 #include "Domain/Creators/Tags/Domain.hpp"
+#include "Domain/Domain.hpp"
 #include "Domain/ElementLogicalCoordinates.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/Structure/BlockId.hpp"
 #include "Domain/Structure/ElementId.hpp"
+#include "Domain/Tags.hpp"
 #include "NumericalAlgorithms/Interpolation/IrregularInterpolant.hpp"
 #include "Options/String.hpp"
 #include "Parallel/GlobalCache.hpp"
 #include "Parallel/Invoke.hpp"
+#include "Parallel/ParallelComponentHelpers.hpp"
 #include "ParallelAlgorithms/EventsAndTriggers/Event.hpp"
 #include "ParallelAlgorithms/Interpolation/Actions/InterpolationTargetVarsFromElement.hpp"
 #include "ParallelAlgorithms/Interpolation/Events/GetComputeItemsOnSource.hpp"
 #include "ParallelAlgorithms/Interpolation/PointInfoTag.hpp"
+#include "ParallelAlgorithms/Interpolation/Targets/Sphere.hpp"
 #include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/Algorithm.hpp"
+#include "Utilities/CartesianProduct.hpp"
+#include "Utilities/ConstantExpressions.hpp"
 #include "Utilities/PrettyType.hpp"
 #include "Utilities/Serialization/CharmPupable.hpp"
 #include "Utilities/TMPL.hpp"
 #include "Utilities/TypeTraits/CreateGetStaticMemberVariableOrDefault.hpp"
+#include "Utilities/TypeTraits/IsA.hpp"
 
 /// \cond
 namespace Events::Tags {
 template <size_t Dim>
 struct ObserverMesh;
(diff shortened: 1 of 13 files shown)


nilsdeppe (dismissed):
:100: Nice work! I've verified this speeds up the BNS with BondiSachs be the required amount. I now can't notice any overhead from observing GW data.

knelli2:
@nilsdeppe Fixed a missing instantiation and an unused variable warning. Squashed directly.

nilsdeppe (dismissed):

src/ParallelAlgorithms/Interpolation/Events/InterpolateWithoutInterpComponent.hpp
```diff
@@ -77,28 +108,175 @@ class InterpolateWithoutInterpComponent<VolumeDim, InterpolationTargetTag,
...
                                                        tmpl::list<>>;
 
-  using argument_tags =
-      tmpl::list<typename InterpolationTargetTag::temporal_id,
-                 Tags::InterpPointInfoBase,
-                 ::Events::Tags::ObserverMesh<VolumeDim>, SourceVarTags...>;
+  using argument_tags = tmpl::list<
+      typename InterpolationTargetTag::temporal_id, Tags::InterpPointInfoBase,
+      ::Events::Tags::ObserverMesh<VolumeDim>,
+      ::Events::Tags::ObserverCoordinates<VolumeDim, frame>, SourceVarTags...>;
```
nilsdeppe: Something I just realized might be a problem: In FD we have gaps between grid points, so if one of the target points ends up between grid points of neighboring elements, would this cause it not to interpolate? The solution would probably be to take the DG coords as the bounds, and then interpolate with `ObserverMesh`

knelli2: Yeah I think this would cause it not to call the post interpolation callback because it'd be waiting to receive all the target points and just wouldn't receive one. Is there a tag for only the DG coords that is accessible to non-subcell code?

nilsdeppe: `domain::Tags::Coordinates` always gives the DG coords in all executables :)

knelli2: Oh ok seems simple enough to just swap out `::Events::Tags::ObserveCoordinates` for `domain::Tags::Coordinates`. And the observer mesh is already used for the interpolation so that's taken care of.

nilsdeppe: Yep! You want the `ObserverMesh`, though, since that's what the interpolation code uses :)

knelli2: Is the `::Events::Tags::ObserverMesh<VolumeDim>` that's already there not what we want? I thought it was

nilsdeppe: Yes, sorry, I was trying to say "Don't change that one, only the coordinates"

knelli2: Ok done, squashed, and added a comment

nilsdeppe (approved):

PR #5504, opened 2023-09-27
Fix bug in damped harmonic gauge

### LINKED ISSUE #4594: Evolve a BBH in damped harmonic gauge
Currently, BBHs require harmonic gauge, because starting in damped harmonic gauge at t=0 requires characteristic speed control. Once characteristic speed control is working, the BinaryBlackHole executable should start in damped harmonic gauge instead of harmonic gauge.

Depends on #4593, #4592, and #4591.

geoffrey4444:
## Proposed changes

Fixes a bug in the time derivative of the damped harmonic gauge source function. One term in the calculation was missing a factor of 1/lapse.

The python unit test didn't catch this because the term in question numerically was never much bigger than 1e-12, the default epsilon for `check_with_random_values`. I increased the maximum magnitude of the random inputs by a factor of 10 to catch such bugs in the future, and I verified that after this change, the unit test would fail if the missing factor of 1/lapse were removed once again.

Fixes #4594.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
@@ -358,7 +358,7 @@ void damped_harmonic_impl(
     }
   }
 
-  // d_t lapse / lapse = 0.5 * n^a n^b (lapse Pi_{ab} - shift^i Phi_{iab})
+  // d_t lapse = lapse * (0.5 * n^a n^b (lapse Pi_{ab} - shift^i Phi_{iab}))
   get<0>(d4_muS_over_lapse) = -get<0>(shift) * get<0>(half_phi_two_normals);
   for (size_t i = 1; i < SpatialDim; ++i) {
     get<0>(d4_muS_over_lapse) -= shift.get(i) * half_phi_two_normals.get(i);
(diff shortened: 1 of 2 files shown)


src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
```diff
@@ -390,7 +390,7 @@ void damped_harmonic_impl(
   //  0.5 * n^a n^b Pi_{ab}
   //  0.5 * n^a n^b Phi_{iab}
   // so we reuse that work by taking them as arguments.
-  get<0>(d4_muS_over_lapse) *= -get(mu_S);
+  get<0>(d4_muS_over_lapse) *= -get(mu_S) * get(one_over_lapse);
```
knelli2, line 393: Don't multiply by one over lapse here. I think you can just remove the multiplication of lapse on line 367 above.

knelli2 (changes requested):

nilsdeppe (commented):
I have no additional comments. Nice catch! :D

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
nilsvu: That would be consistent with the comment in line 361. But the value is used in line 372, so that needs a factor of the lapse then.

nilsvu (dismissed):

src/Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/DampedHarmonic.cpp
knelli2: Oops missed that. @geoffrey4444 could you fix the comment on line 361 then?

geoffrey4444: The equation in that comment was correct but not what was actually being computed, so I updated it to reflect that here we're computing d_t lapse, not dt_lapse / lapse. Squashed

nilsvu (approved):

knelli2 (approved):

PR #5505, opened 2023-09-27
Fix Kastaun recovery for 1D EOS

jyoo1042:
## Proposed changes
replace the recovered $\epsilon$ (specific internal energy) and $\rho h W^2$ with values computed using the equation of state with the recovered rest mass density and Lorentz factor when using Kastaun recovery scheme for 1D equation of state.


ermost:
Good catch!

ermost:
Taking a second look: It seems this should have been in principle catched by the limits on epsilon.
Ideally, you would think that eps_min(rho) = eps_max (rho) = eps (rho) for the pure polytrope.

However, the EOS implementation currently doesn't seem to do this. See here:

```
 double specific_internal_energy_lower_bound(
      const double /* rest_mass_density */) const override {
    return 0.0;
  }
``` 

Should this be fixed as well for consistency? And similarly the temperature bound, which may matter for reconstruction, if those values are used there.

jyoo1042:
@ermost, I just made the change to the specific internal energy lower and upper bound for PolytropicFluid.

ermost:
Looks great! Thank you!

PR #5506, opened 2023-09-27
Add CAPTURE_FOR_ERROR to print info on failure

wthrowe:
Example:
```c++
    const Scalar<DataVector> scalar{DataVector{0.0, 1.0, 5.0, 4.0}};
    const int integer = 12;
    CAPTURE_FOR_ERROR(scalar);
    CAPTURE_FOR_ERROR(integer);
    ERROR("Boom");
```
Leads to:
```
[backtrace and stuff]

Boom

Captured variables:
scalar = T()=(0,1,5,4)
integer = 12
############ ERROR ############
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Utilities/ErrorHandling/CaptureForError.hpp
+++ b/src/Utilities/ErrorHandling/CaptureForError.hpp
@@ -1,0 +1,102 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <ostream>
+#include <vector>
+
+namespace CaptureForError_detail {
+class CaptureForErrorBase {
+ protected:
+  CaptureForErrorBase() = default;
+  ~CaptureForErrorBase() = default;
+
+ public:
+  CaptureForErrorBase(CaptureForErrorBase&&) = delete;
+  CaptureForErrorBase(const CaptureForErrorBase&) = delete;
+  CaptureForErrorBase& operator=(CaptureForErrorBase&&) = delete;
+  CaptureForErrorBase& operator=(const CaptureForErrorBase&) = delete;
+
+  virtual void print(std::ostream& stream) const = 0;
+};
+
+class CaptureList {
+ public:
+  static CaptureList& the_list();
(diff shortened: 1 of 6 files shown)


src/Utilities/ErrorHandling/CaptureForError.hpp
```diff
@@ -0,0 +1,102 @@
...
+
+  // Capturing an rvalue would lead to a dangling reference.
+  CaptureForError(const char* /*name*/, const T&& /*capture*/) = delete;
+
+  ~CaptureForError() noexcept(false) {
+    CaptureList::the_list().deregister_self(this);
+  }
+
+  void print(std::ostream& stream) const override {
+    stream << name_ << " = " << capture_ << "\n";
```
nilsdeppe, line 71: maybe add `using ::operator<<;` to deal with STL as best as we can?

nilsdeppe (approved):
This is awesome :D Thank you for implementing this so quickly! One small question, squash immediately if you agree :)

nilsdeppe:
Marked priority because this will help tremendously with debugging BNS simulations

src/Utilities/ErrorHandling/CaptureForError.hpp
nilsdeppe: And, maybe more important, call `disable_floating_point_exceptions()` (and re-enable)? Or maybe that's handle by `ERROR` itself

wthrowe: `using ::operator<<;` won't compile unless I include something that defines such a function.  I can add it with an include of `StdHelpers.hpp` if you think that's a good idea.

`ERROR` and `ASSERT` have FPEs disabled when they call this.

nilsdeppe: Ah okay, nah let's just leave as-is... Thanks!

PR #5507, opened 2023-09-27
Add Render3D command with simple "clip" rendering

nilsvu:
## Proposed changes

Uses ParaView to render a 2D clip of 3D simulation data. We can add more rendering commands here later for specialized use cases (like BNS, BBH simulations, etc. Example: #4656). The "clip" rendering is an example and useful to get a quick look into the volume data.

I've tested this on CaltechHPC and Wheeler. It should "just work" with the ParaView module in this PR. Try it like this:

```
spectre render-3d clip --help
```
---

For GPU nodes this should also work:

1. Request a GPU node:

```sh
srun --gres=gpu:1 --nodes 1 -t 02:00:00 --pty /bin/bash
```

2. `spectre_load_modules` and load ParaView (I have installed this in the `sxs` group):

```sh
module load paraview/5.11.1-egl
```

3. `spectre render-3d clip --help`

- [x] We should think about how to add tests for these scripts. E.g. we could install ParaView in one of the CI builds somehow, detect the ParaView installation in CMake, and run the rendering tests if ParaView was found. Edit: I'll add the CI test once #5528 is merged because it needs a specific Python version.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
@sxs-collaboration/spectre-core-devs could one of you review this? You can try it on CaltechHPC or Wheeler if you want. Here's an example rendering:

```sh
./bin/spectre render-3d clip Bbh.xmf -o test.png -y Lapse --show-grid --zoom 50
```

test

support/Environments/wheeler_gcc.sh
```diff
@@ -59,6 +60,7 @@ spectre_load_modules() {
     module load hdf5/1.12.2
     module load libbacktrace/1.0
     module load spec-exporter/2023-07
+    module load paraview/5.11.1-osmesa
```
knelli2, line 62: Our [sxs docs for wheeler](https://github.com/sxs-collaboration/WelcomeToSXS/wiki/Wheeler#paraview-for-visualizations) give instructions using `5.10.1`. Maybe we should use that?

knelli2 (dismissed):
The changes look fine, I'm just not sure how useful this will be. Even for a quick visualization, you have to be able to choose everything you want on the command line and if you don't know any of the arguments (specifically the time or the zoom or the variable you want to observe) you have to either look into the volume data itself, or just open paraview. And I feel like often times, even if people do want just a quick visualization, they wind up exploring other things as well which is easier in a GUI.

I won't hold up the PR, but I think this feature is very limited and even if we add more capability, it'd just be easier to do all of this in the GUI in the first place.

support/Environments/wheeler_gcc.sh
nilsvu: That module uses an old Python 3.7. ParaView must use the same version as the rest of the build. Updated the docs.

nilsvu:
Yea I agree using the GUI is the most useful. However, downloading all volume data to you local machine quickly becomes infeasible. So we're left with either connecting to a ParaView server running on the cluster or connecting to a window server running the GUI on the cluster. Neither is particularly easy in many cases, so people still end up downloading many gigabytes of data, which can be prohibitively slow, fill up your whole disk, etc, if it's possible at all. The only solutions that I can think of are:

1. ParaView scripts like this PR that do specific renderings (not for exploration).
2. Making it easy to connect to a ParaView server running on the cluster. I haven't figured that out yet for CaltechHPC, though I _have_ figured out how to run the GUI on CaltechHPC in a VM in the browser and documented it here: https://github.com/sxs-collaboration/WelcomeToSXS/wiki/CaltechHPC.

knelli2:
I think of those options (download vol data, specific scripts, pvserver), using `pvserver` is the way to go as this allows the most flexibility. But yeah, this is then specific to every machine and we have to figure it out

nilsvu:
Absolutely. As I said, this isn't always easy though so people don't do it. I think there's merit to both options I listed above, so we should do both. The specific renderings could also be useful for things like the status CLI where we monitor production runs (maybe).

.github/workflows/Tests.yaml
```diff
@@ -552,6 +556,23 @@ jobs:
...
+        run: |
+          apt-get install -y libglvnd-dev  # Needed to find 'libglapi.so'
+          wget -O paraview.tar.gz --no-check-certificate "https://www.paraview.org/paraview-downloads/download.php?submit=Download&version=v5.10&type=binary&os=Linux&downloadFile=ParaView-5.10.1-osmesa-MPI-Linux-Python3.9-x86_64.tar.gz"
+          tar -xzf paraview.tar.gz
+          rm paraview.tar.gz
+          mv ParaView-* /opt/paraview
+          # Make 'paraview' Python package available
+          PYTHONPATH=/opt/paraview/lib/python3.9/site-packages:$PYTHONPATH
+          # Give system-installed Python packages priority over ParaView's
+          PYTHONPATH=/usr/local/lib/python3.9/dist-packages:$PYTHONPATH
```
knelli2: Just put the paraview python at the end of `$PYTHONPATH`?

Also, in the gcc-10 Debug build, looks like an H5 python test is using the `h5py` from paraview and not from the system. So I imagine that the paraview python is maybe used for other tests as well and I don't think we want that.

knelli2 (changes requested):

.github/workflows/Tests.yaml
nilsvu: > Just put the paraview python at the end of $PYTHONPATH?

Doesn't work because the `PYTHONPATH` is empty by default and system locations are searched automatically, so to give system packages priority you have to prepend them to the path.

knelli2 (approved):

PR #5510, opened 2023-09-28
Check FoT before post intrp callbacks and UpdateMessageQueue

knelli2:
## Proposed changes

Consider the following sequence of events:

1. An element checks that FoTs are invalid. Waits for them to be updated.
2. Once they are updated, the element does an interpolation to some target points and sends the data to the target component. This target component lives on a different node than the element.
3. Upon receiving all the data, the target component wants to do some operation that includes the FoTs in a post interpolation callback.
4. The FoTs have not been updated on this node yet, so they are invalid.
5. There is no check on the target component for the validity of the FoTs, so this leads to a runtime error (usually that the FoTs were evaluated after the expiration time)

This PR fixes this bug by adding a check of the FoTs before we call the post interpolation callback.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Actions/InterpolationTargetReceiveVars.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Actions/InterpolationTargetReceiveVars.hpp
@@ -5,12 +5,14 @@
 
 #include <cstddef>
 #include <memory>
+#include <optional>
 #include <unordered_set>
 #include <utility>
 #include <vector>
 
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "DataStructures/VariablesTag.hpp"
+#include "Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp"
 #include "Parallel/GlobalCache.hpp"
 #include "Parallel/Invoke.hpp"
 #include "ParallelAlgorithms/Interpolation/Actions/SendPointsToInterpolator.hpp"
@@ -24,7 +26,9 @@
 #include "Utilities/TypeTraits.hpp"
 
 /// \cond
-// IWYU pragma: no_forward_declare db::DataBox
+namespace domain::Tags {
+struct FunctionsOfTime;
+}  // namespace domain::Tags
 namespace intrp {
 namespace Tags {
 template <typename TemporalId>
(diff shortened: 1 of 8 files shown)


src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -470,6 +475,43 @@ void add_received_variables(
       box);
 }
 
+template <typename SimpleAction, typename Metavariables, typename ArrayIndex,
+          typename Component, typename TemporalId, typename... Args>
+bool functions_of_time_are_ready(Parallel::GlobalCache<Metavariables>& cache,
```
knelli2: I'm going to factor this out into `FunctionsOfTimeAreReady.hpp` in a separate PR because I'll need this elsewhere.

wthrowe: When you do see if you can share code with the existing `functions_of_time_are_ready`, since they are very similar.

src/ParallelAlgorithms/Interpolation/Actions/InterpolationTargetReceiveVars.hpp
```diff
@@ -124,11 +129,24 @@ struct InterpolationTargetReceiveVars {
...
     if (InterpolationTarget_detail::have_data_at_all_points<
             InterpolationTargetTag>(box, temporal_id)) {
       // All the valid points have been interpolated.
+
+      // Check if functions of time are ready on this component
+      if (not InterpolationTarget_detail::functions_of_time_are_ready<
+              InterpolationTargetReceiveVars>(
+              cache, array_index,
+              std::add_pointer_t<ParallelComponent>{nullptr}, temporal_id,
+              vars_src, global_offsets, temporal_id, true)) {
```
wthrowe: You never use `vars_src` in the action this calls, and I'm guessing it could be a large object that is getting copied.  Perhaps send an empty vector or something?

wthrowe (changes requested):

knelli2:
@wthrowe Rebased and no longer dependent. I also just did the control systems as well since it wasn't that much extra code.

wthrowe (approved):

Issue #3544, opened 2021-09-23
Change name of  tensor `size()` method to `number_of_independent_components()`


nikwit:
It has happened a few times to me now that I initialized something with `scalar.size()` instead of `get(scalar).size()` which resulted in some hard to debug out of range accessor bugs. 

I would suggest to change this method name to `number_of_independent_components` and perhaps also add a `number_of_grid_points` method. After all, `size` is fairly ambiguous when it comes to Tensors. These are the same names that `Variables` use. How much effort would it be to refactor this?

nilsdeppe:
I would estimate the effort to be enormous. Probably 1 week full time. I don't see this being a reasonable refactor and certainly not the effort. 

Aside from that, calling it "size()" is consistent with the STL naming conventions. I don't see adding a `number_of_grid_points()` method as a good thing since `Tensor` can hold a `double` or pretty much any other type (even modal data). `Variables` are always data on a mesh, so it makes sense to talk about the number of grid points.

PR #5511, opened 2023-09-28
Correct tabulated eos indexing bug

pajkosmi:
## Proposed changes

Ensures that the tabulated EoS updates the entire datavector.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

The current version of tabulated eos takes doubles and datavectors as inputs.  However, only the first element of the datavector is updated, as opposed to all entries.  This behavior arises from loop upper limits that run from `0` to `electron_fraction.size()`, which is a value of `1`.  Instead, the upper limit should be `get(electron_fraction).size()`, (note the extra `get()`)---similar to issue #3544 .


ermost:
Thanks for fixing this! I definitely get confused by `get()`

nilsvu (approved):

nilsdeppe (approved):

PR #5513, opened 2023-09-28
Add FoTAreReady function for simple actions

knelli2:
## Proposed changes

The previous version of `functions_of_time_are_ready()` only used a `PerformAlgorithmCallback`. But sometimes we want to use a `SimpleActionCallback`. This generalizes `functions_of_time_are_ready()` to allow both.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
+++ b/src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
@@ -10,6 +10,7 @@
 #include <string>
 #include <tuple>
 #include <unordered_map>
+#include <unordered_set>
 #include <utility>
 
 #include "DataStructures/DataBox/DataBox.hpp"
@@ -18,6 +19,7 @@
 #include "Parallel/ArrayComponentId.hpp"
 #include "Parallel/Callback.hpp"
 #include "Parallel/GlobalCache.hpp"
+#include "Parallel/ParallelComponentHelpers.hpp"
 #include "Utilities/Algorithm.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/StdHelpers.hpp"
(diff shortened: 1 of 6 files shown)


src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
```diff
@@ -36,55 +38,102 @@ class TaggedTuple;
...
+///
+/// Check that functions of time in \p CacheTag with names in \p
+/// functions_to_check are ready at time \p time.  If  \p functions_to_check is
+/// a `std::nullopt`, checks all functions in \p CacheTag.  If any function is
+/// ready, schedules a `Parallel::SimpleActionCallback` with the GlobalCache
+/// which calls the simple action passed in as a template parameter. The `Args`
+/// are moved into the callback.
+template <typename CacheTag, typename SimpleAction, typename Metavariables,
+          typename ArrayIndex, typename Component, typename... Args>
+bool functions_of_time_are_ready(
```
wthrowe: Not sure I'm a fan of distinguishing the two overloads only by the existence of a template parameter.

src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
```diff
@@ -36,55 +38,102 @@ class TaggedTuple;
...
+    Parallel::GlobalCache<Metavariables>& cache, const ArrayIndex& array_index,
+    const Component* component_p, const double time,
+    const std::optional<std::unordered_set<std::string>>& functions_to_check,
+    Args... args) {
```
wthrowe: `Args&&` and forward below.

src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
```diff
@@ -36,55 +38,102 @@ class TaggedTuple;
...
 
 namespace domain {
+namespace detail {
+template <typename CacheTag, typename Callback, typename Metavariables,
+          typename ArrayIndex, typename Component, typename... Args>
+bool functions_of_time_are_ready_impl(
+    Parallel::GlobalCache<Metavariables>& cache, const ArrayIndex& array_index,
+    const Component* /*meta*/, const double time,
+    const std::optional<std::unordered_set<std::string>>& functions_to_check,
+    Args... args) {
```
wthrowe: `Args&&` and forward below.

wthrowe (changes requested):
You also have some clang-tidy warnings.  The gcc9 failures appear to be unrelated and will hopefully work after a new push.

src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
knelli2: I tried this first, but ran into build errors in `SimpleActionCallback` due to the `const &`ness of some of the args that were passed in. I wasn't sure how to get around that because `SimpleActionCallback` needs to actually store the data locally (not references to it which may go out of scope). That's why I chose a copy initially and then `std::move` from then on.

src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
knelli2: Any suggestions? I did it like this because the different callbacks have different template parameters. Didn't make a lot of sense to have to specify an extra template for the perform algorithm callback that would never be used.

src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
wthrowe: That's a bug in `SimpleActionCallback`.  It shouldn't be taking it's arguments by reference, but by value and then moving.

src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
wthrowe: Give them different names, maybe.

src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp
knelli2: Ok added a new commit that fixes `SimpleActionCallback`

src/Parallel/Callback.hpp
```diff
@@ -37,9 +37,10 @@ class SimpleActionCallback : public Callback {
  public:
   WRAPPED_PUPable_decl_template(SimpleActionCallback);  // NOLINT
   SimpleActionCallback() = default;
-  SimpleActionCallback(Proxy proxy, Args&&... args)
+  // NOLINTNEXTLINE(google-explicit-constructor)
+  SimpleActionCallback(Proxy proxy, std::decay_t<Args>... args)
```
wthrowe, line 41: Why do you need `decay_t` all over the place in this class?

src/Parallel/Callback.hpp
```diff
@@ -37,9 +37,10 @@ class SimpleActionCallback : public Callback {
...
       : proxy_(proxy),
-        args_(std::make_tuple<Args...>(std::forward<Args>(args)...)) {}
+        args_(std::make_tuple<std::decay_t<Args>...>(std::move(args)...)) {}
```
wthrowe: [optional] `make_tuple` doesn't do anything here.  It can be removed.

src/Parallel/Callback.hpp
knelli2: Well `Args...` keeps all the `const &` so I had to remove those for the declaration of the member tuple so it wasn't a tuple of references, but I left everything else as `Args...`. However, my test failed with that because a `DataVector` I was using to test actually got moved into this callback instead of copied (not `std::move` from the calling code, but from the `std::move` one line below) and this caused subsequent tests to fail. When I did `std::decay_t<Args>...` in the constructor, everything worked fine.

wthrowe: I see.  I think you're right and this is the correct way to do it here.

Optionally, you might want to add `decay_t` to the calling code as well, just to keep the number of equivalent types the compiler has to deal with down.  Probably not many instantiations of this class though, so not very important.

wthrowe (commented):
That leaves only optional things.  Let me know if you want to do any or if this is ready.

src/Parallel/Callback.hpp
knelli2: Other than the test, the calling code (in the FoT are ready functions) is usually `std::forward<Args>(args)...`. So I'll leave that as is.

knelli2:
Squashed in the optional remove of `std::make_tuple`

wthrowe (approved):

PR #5516, opened 2023-09-29
Add ComovingCharSpeedDerivative

markscheel:
This is a function that computes the derivative of the comoving characteristic speed with respect to the map parameter. The sign of this derivative is used in the control system for state DeltaRDriftInward. Basically, if a particular state will cause the comoving characteristic speed to decrease, then we don't enter that state.

Note that this is not exactly the same quantity computed in SpEC (but I think that SpEC should be computing this quantity and not what it currently computes).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
+++ b/src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
@@ -1,0 +1,153 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp"
+
+#include <cmath>
+#include <cstddef>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tags/TempTensor.hpp"
+#include "DataStructures/TempBuffer.hpp"
+#include "DataStructures/Tensor/IndexType.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace control_system::size {
+void comoving_char_speed_derivative(
+    const gsl::not_null<Scalar<DataVector>*> result, const double lambda_00,
+    const double dt_lambda_00, const double horizon_00,
+    const double dt_horizon_00, const double grid_frame_excision_sphere_radius,
+    const tnsr::i<DataVector, 3, Frame::Distorted>& excision_rhat,
+    const tnsr::i<DataVector, 3, Frame::Distorted>& excision_normal_one_form,
+    const Scalar<DataVector>& excision_normal_one_form_norm,
+    const tnsr::I<DataVector, 3, Frame::Distorted>&
+        distorted_components_of_grid_shift,
+    const tnsr::II<DataVector, 3, Frame::Distorted>&
+        inverse_spatial_metric_on_excision_boundary,
+    const tnsr::Ijj<DataVector, 3, Frame::Distorted>&
+        spatial_christoffel_second_kind,
+    const tnsr::i<DataVector, 3, Frame::Distorted>& deriv_lapse,
+    const tnsr::iJ<DataVector, 3, Frame::Distorted>& deriv_of_distorted_shift,
+    const InverseJacobian<DataVector, 3, Frame::Grid, Frame::Distorted>&
+        inverse_jacobian_grid_to_distorted) {
+  const double Y00 = 0.25 * M_2_SQRTPI;
+
+  // Define temporary storage.
+  using excision_normal_vector_tag =
+      ::Tags::TempI<1, 3, Frame::Distorted, DataVector>;
+  using deriv_normal_one_form_tag =
+      ::Tags::Tempi<2, 3, Frame::Distorted, DataVector>;
+  TempBuffer<tmpl::list<excision_normal_vector_tag, deriv_normal_one_form_tag>>
+      buffer(get<0>(excision_rhat).size());
+  auto& excision_normal_vector = get<excision_normal_vector_tag>(buffer);
+  auto& deriv_normal_one_form = get<deriv_normal_one_form_tag>(buffer);
+
+  // excision_rhat is a tnsr:i when it is returned from a Strahlkorper.
+  // But excision_rhat is a coordinate quantity, not a physical tensor, so
+  // it can also be used as a tnsr::I.  Here we create a tnsr::I called
+  // excision_rhat_vector that points into excision_rhat.
+  const tnsr::I<DataVector, 3, Frame::Distorted> excision_rhat_vector{};
+  for (size_t i = 0; i < 3; ++i) {
+    // Is there a way to do this without the const_casts?
+    // Note that excision_rhat_vector must be non-const because
+    // we are changing it (by calling set_data_ref).
+    // And set_data_ref expects a non-const argument.
+    const_cast<DataVector*>(&excision_rhat_vector.get(i))  // NOLINT
+        ->set_data_ref(
+            const_cast<DataVector*>(&excision_rhat.get(i)));  // NOLINT
+  }
+
+  tenex::evaluate<ti::I>(
+      make_not_null(&excision_normal_vector),
+      excision_normal_one_form(ti::j) *
+          inverse_spatial_metric_on_excision_boundary(ti::J, ti::I));
+
+  // Fill result temporarily with all the terms in d/dlambda00 (n_hati)
+  // that are proportional to n_hati.
+  //   First, fill result with s_p s_j gamma^{pk} xi^i Gamma^j_{ki},
+  //   which is (almost) the last term in d/dlambda00(n_hati).
+  tenex::evaluate<>(
+      result, excision_normal_vector(ti::K) * excision_normal_one_form(ti::j) *
+                  excision_rhat_vector(ti::I) *
+                  spatial_christoffel_second_kind(ti::J, ti::k, ti::i));
+  //   Second, add to result s^k s_j InvJac^j_k / r_EB, which is
+  //   (almost) the second term in d/dlambda00(n_hati)
+  //     Note that the for the contraction s_j InvJac^j_k, the j on the s is
+  //     a distorted-frame index but the j on the InvJac is a grid-frame index.
+  //     This is really weird but it happens because some of the things are
+  //     not tensors. (In particular, the map itself looks like
+  //     x^{i_distorted} = x^{i_grid} * stuff, which equates grid and
+  //     distorted incides).
+  for (size_t j = 0; j < 3; ++j) {
+    for (size_t k = 0; k < 3; ++k) {
+      get(*result) += excision_normal_vector.get(k) *
+                      excision_normal_one_form.get(j) *
+                      inverse_jacobian_grid_to_distorted.get(j, k) /
+                      grid_frame_excision_sphere_radius;
+    }
+  }
+  //   Third, scale by norm^3 so that result contains
+  //   1/a (n^k n_j InvJac^j_k / r_EB + n_p n_j gamma^{pk} xi^i Gamma^j_{ki}),
+  //   which is (almost) the last two terms of d/dlambda00(n_hati).
+  get(*result) /= cube(get(excision_normal_one_form_norm));
+
+  // Set deriv_normal_one_form to the first two terms of d/dlambda00 (n_hati).
+  // Possible memory optimization: excision_normal_vector isn't used anymore,
+  // so that storage could be used for deriv_normal_one_form.
+  tenex::evaluate<ti::i>(make_not_null(&deriv_normal_one_form),
+                         -Y00 * (*result)() * excision_normal_one_form(ti::i));
+
+  // Add the first term to deriv_normal_one_form, so that deriv_normal_one_form
+  // contains the entire d/dlambda00 (n_hati).
+  //     Note that the for the contraction s_j InvJac^j_i, the j on the s is
+  //     a distorted-frame index but the j on the InvJac is a grid-frame index.
+  //     This is really weird but it happens because some of the things are
+  //     not tensors. (In particular, the map itself looks like
+  //     x^{i_distorted} = x^{i_grid} * stuff, which equates grid and
+  //     distorted incides).
+  for (size_t i = 0; i < 3; ++i) {
+    for (size_t j = 0; j < 3; ++j) {
+      deriv_normal_one_form.get(i) +=
+          excision_normal_one_form.get(j) *
+          inverse_jacobian_grid_to_distorted.get(j, i) * Y00 /
+          (grid_frame_excision_sphere_radius *
+           get(excision_normal_one_form_norm));
+    }
+  }
+
+  // Now put the actual result, i.e. d/dlambda00 (v_c), into result.
+  // Do this term by term.  Ignore the overall factor of Y00 until the
+  // end.
+  //   First do the dt_horizon_00 term (without the normalization)
+  tenex::evaluate<>(result, -excision_normal_one_form(ti::i) *
+                                excision_rhat_vector(ti::I) * dt_horizon_00 /
+                                horizon_00);
+  //   Next do the shift term (again without the normalization)
+  tenex::update<>(result,
+                  (*result)() - excision_normal_one_form(ti::i) *
+                                    excision_rhat_vector(ti::J) *
+                                    deriv_of_distorted_shift(ti::j, ti::I));
+
+  // Put in the norm factor.
+  get(*result) /= get(excision_normal_one_form_norm);
+
+  // Add the dlapse term (without the Y00 factor).
+  tenex::update<>(
+      result, (*result)() + deriv_lapse(ti::i) * excision_rhat_vector(ti::I));
+
+  // Put in the Y00 factor.
+  get(*result) *= Y00;
+
+  // Add the final term to result
+  tenex::update<>(
+      result,
+      (*result)() +
+          deriv_normal_one_form(ti::i) *
+              (Y00 * dt_lambda_00 * excision_rhat_vector(ti::I) +
+               distorted_components_of_grid_shift(ti::I) -
+               excision_rhat_vector(ti::I) * (dt_horizon_00 / horizon_00) *
+                   (Y00 * lambda_00 - grid_frame_excision_sphere_radius)));
+}
+}  // namespace control_system::size
--- a/src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
+++ b/src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
@@ -1,0 +1,418 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Utilities/Gsl.hpp"
+
+/// \cond
+class DataVector;
+namespace Frame {
+struct Distorted;
+struct Grid;
+}  // namespace Frame
+/// \endcond
+
+namespace control_system::size {
+
+/*!
+ * \brief Computes the derivative of the comoving characteristic speed
+ * with respect to the size map parameter.
+ *
+ * \param result the derivative of the comoving char speed
+ *        \f$d v_c/d\lambda_{00}\f$, which is computed here using
+ *        Eq. (\f$\ref{eq:result}\f$).
+ * \param lambda_00 the map parameter \f$\lambda_{00}\f$. This is the usual
+ *        spherical harmonic coefficient, not a Spherepack value.
+ * \param dt_lambda_00 the time derivative of the map parameter
+ * \param horizon_00 the average coefficient of the horizon \f$\hat{S}_{00}\f$.
+ *        This is the usual spherical harmonic coefficient, not a Spherepack
+ *        value.
+ * \param dt_horizon_00 the time derivative of horizon_00
+ * \param grid_frame_excision_sphere_radius radius of the excision boundary
+ *        in the grid frame, \f$r_{\mathrm{EB}}\f$.
+ * \param excision_rhat the direction cosine \f$\xi_\hat{i}\f$. Not a
+ *        spacetime tensor: it is raised/lowered with \f$\delta_{ij}\f$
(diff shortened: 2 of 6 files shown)


src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+}  // namespace Frame
+/// \endcond
+
+namespace control_system::size {
+
+/*!
+ * \brief Computes the derivative of the comoving characteristic speed
+ * with respect to the size map parameter.
+ *
+ * \tparam Frame should be ::Frame::Distorted if ::Frame::Distorted exists.
```
wthrowe: And if it doesn't?

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Utilities/Gsl.hpp"
+
+/// \cond
+class DataVector;
+namespace Frame {
+struct Distorted;
+struct Inertial;
+}  // namespace Frame
```
wthrowe, line 14: Move `Frame` declarations to cpp.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  for (size_t i = 0; i < 3; ++i) {
+    excision_normal_vector.get(i) =
+        get<0>(excision_normal_one_form) *
+            inverse_spatial_metric_on_excision_boundary.get(0, i) +
+        get<1>(excision_normal_one_form) *
+            inverse_spatial_metric_on_excision_boundary.get(1, i) +
+        get<2>(excision_normal_one_form) *
+            inverse_spatial_metric_on_excision_boundary.get(2, i);
+  }
```
wthrowe: [optional] Consider
```c++

tenex::evaluate<ti::I>(
    make_not_null(&excision_normal_vector),
    excision_normal_one_form(ti::j) *
    inverse_spatial_metric_on_excision_boundary(ti::J, ti::I));
```
and perhaps similar things below.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
```diff
@@ -0,0 +1,140 @@
...
+                        spatial_christoffel_second_kind.get(j, k, m);
+      }
+    }
+  }
+  // scale so result contains most of the terms in
+  // d/dlambda00 (n_hati)
+  get(*result) /= cube(get(excision_normal_one_form_norm));
+
+  // Now overwrite excision_normal_vector to contain
+  // d/dlambda00 (n_hati).
```
wthrowe: This isn't a vector.  Make a temporary of the right type.

wthrowe (changes requested):
I found that the variable reuse made this function very hard to understand, particularly when values were being stored in inappropriate tensor types.  I am skeptical that this function needs that kind of optimization.  If you do decide to reuse variables, create new references to give them different names.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+ * grid frame (where the excision boundary is a sphere). Here
+ * \f$B\f$ represents terms that depend on angles but are independent of
+ * \f$\lambda_{00}\f$.  These \f$B\f$ terms depend on \f$\lambda_{\ell m}\f$ for
+ * \f$\ell>0\f$ but the precise form of these terms will not be important here,
+ * because below we will be differentiating the map with respect
+ * to \f$\lambda_{00}\f$.
+ *
+ * The comoving characteristic speed is
+ * \f{align}
+ *     v_c &= -\alpha -\hat{n}_\hat{i}\beta^\hat{i}
```
knelli2: Is this supposed to be $+\hat{n}_\hat{i}\beta^\hat{i}$? The previous equation for the regular char speed on the excision boundary has a + and the transformations you describe below don't seem like they'd add a minus sign.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+ * negative forever, then size control will fail.  Therefore \f$v_c\f$ is
+ * used in making decisions in size control. We care about
+ * \f$d v_c/d\lambda_{00}\f$ because the sign of that quantity tells us
+ * whether \f$v_c\f$ will increase or decrease when we increase or decrease
+ * the map parameter \f$\lambda_{00}\f$; this information will be used to
+ * decide whether to transition between different size control states.
+ *
+ * ## Derivation of comoving_char_speed_derivative
+ *
+ * The function comoving_char_speed_derivative computes
```
knelli2: You don't need the function name. We're already reading the docs for this function. Maybe

```
## Derivation of derivative of comoving characteristic speed

This function computes ...
```

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+ *  \frac{d}{d\lambda_{00}}\frac{\partial \hat{x}^i}{\partial x^j} &=
+ *  -\frac{Y_{00}}{r_{\mathrm{EB}}} \delta^i_j.
+ * \f}
+ *
+ * By taking the derivative of the identity
+ * \f{align}
+ *  \frac{\partial \hat{x}^\hat{i}}{\partial x^k}
+ *  \frac{\partial x^k}{\partial \hat{x}^\hat{j}} &= \delta^\hat{i}_\hat{j}
+ * \f}
+ * we can derive
```
knelli2: `we can derive the derivative of the inverse Jacobian as`

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+ * \f{align}
+ *  \frac{d}{d\lambda_{00}}\frac{\partial x^i}{\partial \hat{x}^j} &=
+ *  +\frac{Y_{00}}{r_{\mathrm{EB}}} \delta^i_j.
+ * \f}
```
knelli2: Why the sign change?

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+ *
+ * ### Derivative of a function of space
+ *
+ * Assume we have arbitrary function of space \f$f(\hat{x}^i)\f$.
```
knelli2: `an arbitrary function`

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+ *  \frac{d}{d\lambda_{00}} a^2 &= 2 s_i\frac{Y_{00}}{r_{\mathrm{EB}}}
+ *  \delta^i_\hat{i} \hat{s}_{\hat j} \gamma^{\hat{i} \hat{j}}
+ *  + \hat{s}_{\hat i} \hat{s}_{\hat j}
+ *  \gamma^{\hat{i} \hat{k}} \gamma^{\hat{j} \hat{l}}
+ *  \xi^\hat{m} Y_{00} \partial_{\hat m} \gamma_{\hat{k} \hat{l}}.
+ * \f}
+ * Here we have used Eq. (\f$\ref{eq:derivf}\f$) to differentiate the 3-metric.
+ * We have also refrained from raising and lowering indices
+ * to alleviate potential confusion over whether to raise or lower using
+ * \f$\gamma_{\hat{i} \hat{j}}\f$ or using \f$\delta_{\hat{i}\hat{j}}\f$.
```
knelli2: You say you refrain from raising/lowering indices, but you have $\gamma_{\hat{k} \hat{l}}$ instead of $\gamma^{\hat{i} \hat{j}}$ ?

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+ *  \delta^i_\hat{k} \hat{s}_{\hat j} \gamma^{\hat{k} \hat{j}}
+ *  - \hat{s}_{\hat i} \frac{Y_{00}}{2a^3} \hat{s}_{\hat p}
+ *  \hat{s}_{\hat j} \gamma^{\hat{p} \hat{k}}
+ *  \gamma^{\hat{j} \hat{l}}
+ *  \xi^\hat{m} \partial_{\hat m} \gamma_{\hat{k} \hat{l}} \\
+ *  &=
+ *  \xi_\hat{i}\frac{Y_{00}}{r_{\mathrm{EB}}}
+ *  - \hat{s}_{\hat i} \frac{1}{a^3} \xi_\hat{k}\frac{Y_{00}}{r_{\mathrm{EB}}}
+ *    \hat{s}_{\hat j} \gamma^{\hat{k} \hat{j}}
+ *  - \hat{s}_{\hat i} \frac{Y_{00}}{2a^3} \hat{s}_{\hat p}
```
knelli2: Did this factor of 2 get absorbed into $\Gamma^\hat{j}_{\hat{k}\hat{m}}$?

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,347 @@
...
+    double grid_frame_excision_sphere_radius,
+    const tnsr::i<DataVector, 3, Frame>& excision_rhat,
+    const tnsr::i<DataVector, 3, Frame>& excision_normal_one_form,
+    const Scalar<DataVector>& excision_normal_one_form_norm,
+    const tnsr::I<DataVector, 3, Frame>& frame_components_of_grid_shift,
+    const tnsr::II<DataVector, 3, Frame>&
+        inverse_spatial_metric_on_excision_boundary,
+    const tnsr::Ijj<DataVector, 3, Frame>& spatial_christoffel_second_kind,
+    const tnsr::i<DataVector, 3, Frame>& deriv_lapse,
+    const tnsr::iJ<DataVector, 3, Frame>& deriv_shift);
```
knelli2: Even though you define `deriv_shift` in the docs, it's still a bit confusing that's it's not the derivative of `frame_components_of_grid_shift`. What about the name `deriv_of_frame_shift`?

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
knelli2: If we don't expect this to be used in any frame other than the distorted frame, I think we should just hardcode `Frame::Distorted` everywhere.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
knelli2: Yeah I think a lot of this can be simplified with tensor expressions. Would make things easier to read.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
```diff
@@ -0,0 +1,140 @@
...
+    get(*result) += excision_normal_vector.get(k) * excision_rhat.get(k) /
+                    grid_frame_excision_sphere_radius;
+  }
+  // Add n_p n_j gamma^{pk} xi^m Gamma^j_{km} to result
+  for (size_t k = 0; k < 3; ++k) {
+    for (size_t j = 0; j < 3; ++j) {
+      for (size_t m = 0; m < 3; ++m) {
+        get(*result) += excision_normal_vector.get(k) *
+                        excision_normal_one_form.get(j) * excision_rhat.get(m) *
+                        spatial_christoffel_second_kind.get(j, k, m);
```
knelli2: The passed in `excision_rhat` is a `tnsr::i` and I believe it's used properly previously. But here you're multiplying it by `spatial_christoffel_second_kind` which is `tnsr::Ijj` and I'm pretty sure the `I` part is contracted with the `excision_normal_one_form` and the first `j` is contracted with the `excision_normal_vector`. That leaves the second `j` to be contracted with a `tnsr::i`? Should `excision_rhat` actually be `excision_rhat_vector` here? That'd be consistent with the dox.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
```diff
@@ -0,0 +1,140 @@
...
+    excision_normal_vector.get(i) =
+        Y00 * (excision_rhat.get(i) / grid_frame_excision_sphere_radius -
+               get(*result) * excision_normal_one_form.get(i));
+  }
+
+  // Overwrite result term by term. First do the n_i xi^i term
+  // (without the Y00 or norm factor)
+  get(*result) = get<0>(excision_normal_one_form) * get<0>(excision_rhat);
+  for (size_t i = 1; i < 3; ++i) {
+    get(*result) += excision_normal_one_form.get(i) * excision_rhat.get(i);
```
knelli2: Same here. Both `excision_normal_one_form` and `excision_rhat` are `tnsr::i`.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
```diff
@@ -0,0 +1,140 @@
...
+  }
+  get(*result) *= -dt_horizon_00 / horizon_00;
+
+  // Add the n_i xi^j partial_j \beta^i term (without the Y00 or norm
+  // factor)
+  for (size_t i = 0; i < 3; ++i) {
+    for (size_t j = 0; j < 3; ++j) {
+      get(*result) += excision_rhat.get(j) * excision_normal_one_form.get(i) *
```
knelli2: Same here.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
```diff
@@ -0,0 +1,140 @@
...
+                      deriv_shift.get(j, i);
+    }
+  }
+
+  // Put in the norm factor.
+  get(*result) /= get(excision_normal_one_form_norm);
+
+  // Add the dlapse term (without the Y00 factor).
+  for (size_t i = 0; i < 3; ++i) {
+    get(*result) -= deriv_lapse.get(i) * excision_rhat.get(i);
```
knelli2: Same here.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
```diff
@@ -0,0 +1,140 @@
...
+
+  // Put in the Y00 factor.
+  get(*result) *= Y00;
+
+  // Add the final factor to result
+  for (size_t i = 0; i < 3; ++i) {
+    get(*result) += excision_normal_vector.get(i) *
+                    (Y00 * dt_lambda_00 * excision_rhat.get(i) -
+                     frame_components_of_grid_shift.get(i) -
+                     excision_rhat.get(i) * (dt_horizon_00 / horizon_00) *
```
knelli2: Same here.

knelli2 (changes requested):
Yeah I agree, the temporary storage was a bit hard to follow.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
markscheel: Yes, it should be a plus sign here. Good catch!  The sign propagates to the last equation, eq. 28, where both shift terms should have signs switched [please make sure you agree]. There is another sign error in eq 28; the deriv-of-lapse term should have a plus sign because of the minus sign in eq. 12. [please make sure you agree]

knelli2: Yes in the final equation I agree that

1. The sign of the derivative of the lapse should should be switched from minus to plus
2. The signs of both the shift terms should be flipped from what they currently are.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
markscheel: Yes

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
markscheel: Uh oh. When I redid the calculation more carefully (making sure all the hats are in the right place), I still get the sign change but I get an extra factor of Jacobian.  This will make all the subsequent formulas more complicated, and may mean that yet another quantity might need to be interpolated.

knelli2: Hmm yes I can see that...Is the possible new quantity the jacobian itself?

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
markscheel: Added many tenex things. Cannot do everywhere because of frame differences (see comment).

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
markscheel: I think I need them now in hpp since I use explicit frames.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
markscheel: I made a new temp.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
markscheel: Now pass in a Jacobian.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
markscheel: Reworded.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
markscheel: reworded.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
markscheel: reworded.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.cpp
```diff
@@ -0,0 +1,153 @@
...
+  // excision_rhat_vector that points into excision_rhat.
+  const tnsr::I<DataVector, 3, Frame::Distorted> excision_rhat_vector{};
+  for (size_t i = 0; i < 3; ++i) {
+    // Is there a way to do this without the const_casts?
+    // Note that excision_rhat_vector must be non-const because
+    // we are changing it (by calling set_data_ref).
+    // And set_data_ref expects a non-const argument.
+    const_cast<DataVector*>(&excision_rhat_vector.get(i))  // NOLINT
+        ->set_data_ref(
+            const_cast<DataVector*>(&excision_rhat.get(i)));  // NOLINT
```
wthrowe, line 59: Answering the question: There isn't a better way than what you have here.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,415 @@
...
+ * \brief Computes the derivative of the comoving characteristic speed
+ * with respect to the size map parameter.
+ *
+ * \param result the derivative of the comoving char speed
+ *        \f$d v_c/d\lambda_{00}\f$, which is computed here using
+ *        Eq. (\f$\ref{eq:result}\f$).
+ * \param lambda_00 the map parameter \f$\lambda_{00}\f$
+ * \param dt_lambda_00 the time derivative of the map parameter
+ * \param horizon_00 the average coefficient of the horizon \f$\hat{S}_{00}\f$
+ * \param dt_horizon_00 the time derivative of horizon_00
```
knelli2: Are these values the regular spherical harmonic values, or are they `Spherepack` values? We should be explicit because I've been confused by this in the past where we don't specify.

wthrowe (dismissed):
Looks good.

src/ControlSystem/ControlErrors/Size/ComovingCharSpeedDerivative.hpp
```diff
@@ -0,0 +1,415 @@
...
+ * \param spatial_christoffel_second_kind the Christoffel symbols
+ *        \f$\Gamma^\hat{k}_{\hat{i}\hat{j}}\f$
+ * \param deriv_lapse the spatial derivative of the lapse
+ *        \f$\partial_\hat{i} \alpha\f$
+ * \param deriv_of_distorted_shift the spatial derivative of the shift in the
+ *        distorted frame
+ *        \f$\partial_\hat{j} \hat{\beta}^\hat{i}\f$. This is not the
+ *        derivative of distorted_components_of_grid_shift.
+ * \param inverse_jacobian_grid_to_distorted the quantity
+ *        \f$J^i_\hat{k}=partial_\hat{k} x^i\f$,
```
knelli2: `\partial`

knelli2 (changes requested):
Couple small things. Go ahead and squash everything in.

markscheel:
Made small changes and squashed.

wthrowe (approved):

knelli2 (approved):

PR #5525, opened 2023-10-02
Move internals of ThreadsafeList to a tpp file

wthrowe:
I don't think this class is particularly expensive.  This was mostly an experiment into how difficult it is to move a class to a tpp when it is not used in the interface of any other class.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
+++ b/src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
@@ -18,6 +18,7 @@
 #include <vector>
 
 #include "DataStructures/DataVector.hpp"
+#include "Domain/FunctionsOfTime/ThreadsafeList.tpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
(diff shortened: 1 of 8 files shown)


src/Domain/FunctionsOfTime/ThreadsafeList.tpp
```diff
@@ -0,0 +1,263 @@
...
+template <typename T>
+void Interval<T>::pup(PUP::er& p) {
+  p | expiration;
+  p | data;
+  p | previous;
+}
+}  // namespace ThreadsafeList_detail
+
+template <typename T>
+ThreadsafeList<T>::ThreadsafeList() = default;
```
knelli2, line 37: Thoughts on explicit instantiation in a `cpp` for this if we only plan to use it in the functions of time? I know for generic containers that's not the best because then you have to have an instantiation for each possible type you'll need. But if we only plan to use these in the functions of time, I could see it working for this (even though you're right, this wouldn't take up much compile time in an `hpp/tpp`).

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -18,13 +18,36 @@
 #include <vector>
 
 #include "DataStructures/DataVector.hpp"
+#include "Domain/FunctionsOfTime/ThreadsafeList.tpp"
```
knelli2, line 21: `hpp` and `tpp`? I don't think we have a convention for this because I've seen it done both ways.

src/Domain/FunctionsOfTime/ThreadsafeList.tpp
wthrowe: That was basically my reasoning for _not_ doing this as explicit instantiations.  It's only used in three translation units, so we don't need to avoid implicit instantiations for performance, and it hurts maintainability.

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
wthrowe: Yeah, we don't have a convention for this.  I've gone through several iterations myself on how I decide what headers to include where.  At the moment I'm going with something like "Include the hpp if the type is named (or sometimes non-obviously implicitly named, annoyingly) in the file.".  Here the type is named in the hpp, but not the cpp, so that's where the include is.  If it were named in both, both would get it.

tpp files are pretty unambiguous, at least.

knelli2: Ok sounds good. We should probably make a convention for this at some point just so we have something everybody can refer to.

knelli2 (approved):

PR #5528, opened 2023-10-03
Install CI compilers etc on demand to reduce disk space usage on CI

### LINKED ISSUE #5517: Remove CI container, install compiler versions on demand
The CI container is so large (6-7GB) because it includes all compilers, Python, and CMake versions that we test on CI. But they're just downloaded through apt, which we can just do on CI instead. The build that needs the particular compiler, Python, or CMake version can download it. There's no advantage to "pre-downloading" these things in the Docker container, since the container needs to be downloaded on CI anyway.

This means to reproduce a specific CI build in the container the compiler has to be installed first, like `apt-get install clang-13`. I think that's fine, since it leads to way smaller container sizes, hence faster downloads of the container, and more disk space available on CI where disk space is very constrained. It also allows us to adjust compiler versions on CI more easily since we don't have to rebuild the container.

Only compiling the MPI version of Charm++ shouldn't be done on CI because it takes a long time. We can include it in the dev container.

nilsvu:
## Proposed changes

Fixes #5517. CI container still exists, but only for the MPI build of Charm. This makes it a lot easier to add clang 15 and 16 builds to CI. Also allows installing ParaView+Python on demand for #5507.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/actions/parse-compiler/action.yaml
+++ b/.github/actions/parse-compiler/action.yaml
@@ -1,0 +1,36 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+name: Parse compiler
+
+description: |
+  Parse the compiler name and version from the input and set the environment
+  variables CC, CXX, FC, COMPILER_ID, and COMPILER_VERSION.
+
+inputs:
+  compiler:
+    description: Compiler to parse, e.g. 'gcc-10'
+    required: true
+
+runs:
+  using: "composite"
+  steps:
+    - run: |
+        if [[ ${{ inputs.compiler }} =~ (gcc|clang)-([0-9\.]+) ]]; then
+          COMPILER_ID=${BASH_REMATCH[1]}
+          COMPILER_VERSION=${BASH_REMATCH[2]}
+          CC=${COMPILER_ID}-${COMPILER_VERSION};
+          if [[ $COMPILER_ID = gcc ]]; then
+            CXX=g++-${COMPILER_VERSION};
+            FC=gfortran-${COMPILER_VERSION};
+          else
+            CXX=clang++-${COMPILER_VERSION};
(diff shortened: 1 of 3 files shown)


knelli2:
Have you checked how this impacts how long CI runs when you actually have to install the compilers?

nilsvu:
No because the container still has those installed. Should be faster than downloading the container with all compilers installed though. It's just downloading binaries: either the container with all compilers from Dockerhub, or just the one compiler you need from apt.

nilsdeppe:
Is the argument for that that the total number of bytes transferred is less? That's not obviously the issue to me, because while you download less you download from a different source and you actually have to do an installation. I have no real intuition how those effect this

nilsvu:
I'll do a quick test on my fork. I think either way the time to download the container and/or the compilers isn't relevant (it's a couple of minutes). More important is (1) the flexibility of changing the compilers we test, (2) same for Python versions, e.g. for testing ParaView we need to use a Python version compatible with the ParaView binaries, and (3) the reduced disk space used by the container, so we can do things like build the coverage report again etc, which is currently running out of disk space.

nilsdeppe:
Sounds good! Would the Dockerfile preamble  be a natural place to stick what you just said? That makes the design decision very clear :)

nilsvu:
Ok added a comment to the `ci` stage of the Dockerfile

nilsvu:
Here's the test on my fork: https://github.com/nilsvu/spectre/actions/runs/6462446918

Downloading the container takes 1m down from ~1m 50s and installing compilers takes a couple of seconds.

nilsdeppe (approved):
LGTM

knelli2 (approved):

PR #5531, opened 2023-10-03
Reinterpolate to all elements for new target points (fix BBH deadlocks)

### LINKED ISSUE #5487: Fix BBH evolution deadlocks
BBH evolutions sometimes deadlock. Why? How to fix?

knelli2:
## Proposed changes

This is believed to be what was causing a lot of deadlocks in BBH simulations. Between horizon find iterations, a target point of the new horizon surface would move from one element on one core, to a different element on a second core. Because of some implicit assumption made about message ordering (which wasn't consciously made, but just overlooked), this target point never got interpolated to. Thus the target was waiting for the interpolated data from this point while the interpolator was waiting for its next set of points and a deadlock happened.

Fixes #5487.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
@@ -12,6 +12,7 @@
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "ParallelAlgorithms/Interpolation/Actions/TryToInterpolate.hpp"
 #include "ParallelAlgorithms/Interpolation/InterpolatedVars.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/Gsl.hpp"
 #include "Utilities/Requires.hpp"
 #include "Utilities/TaggedTuple.hpp"
@@ -45,6 +46,17 @@ namespace Actions {
 /// After receiving the points, interpolates volume data onto them
 /// if it already has all the volume data.
 ///
+/// The `iteration` parameter is used to order receives of
+/// `block_logical_coords`. Because of the asynchronous nature of communication,
+/// it is possible that a more recent set of points arrives before an older set.
+/// It is assumed that if a more recent set arrives, then the old set is no
+/// longer needed. This `iteration` parameter tags each communication as "more
+/// recent" or "older" so if we receive an older set of points after a more
+/// recent set, we don't overwrite the more recent set.
+///
+/// \note If the interpolator receives points with the same iteration, an ERROR
+/// will occur.
+///
 /// Uses:
 /// - Databox:
 ///   - `Tags::NumberOfElements`
@@ -69,9 +81,10 @@ struct ReceivePoints {
       std::vector<std::optional<
           IdPair<domain::BlockId,
                  tnsr::I<double, VolumeDim, typename ::Frame::BlockLogical>>>>&&
-          block_logical_coords) {
+          block_logical_coords,
+      const size_t iteration = 0_st) {
     db::mutate<intrp::Tags::InterpolatedVarsHolders<Metavariables>>(
-        [&temporal_id, &block_logical_coords](
+        [&temporal_id, &block_logical_coords, &iteration](
             const gsl::not_null<typename intrp::Tags::InterpolatedVarsHolders<
                 Metavariables>::type*>
                 vars_holders) {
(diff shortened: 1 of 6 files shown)


src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
```diff
@@ -80,12 +80,32 @@ struct ReceivePoints {
...
-                  std::move(block_logical_coords)}));
+          // Add the new target interpolation points at this temporal_id. If we
+          // already have some target points at this temporal_id, we overwrite
+          // them with the new target points. This is because the interpolation
+          // has already finished for the previous set of target points if we
+          // are receiving new target points. This core must not have had any
+          // elements that contained any of the previous target points. However,
+          // these elements may now contain some new target points.
+          if (vars_infos.count(temporal_id) == 0) {
+            vars_infos.emplace(std::make_pair(
```
wthrowe: `vars_infos.insert_or_assign`?  (Haven't checked exactly what type this is, but most similar classes have it.)

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
```diff
@@ -80,12 +80,32 @@ struct ReceivePoints {
...
-              intrp::Vars::Info<VolumeDim, typename InterpolationTargetTag::
-                                               vars_to_interpolate_to_target>{
-                  std::move(block_logical_coords)}));
+          // Add the new target interpolation points at this temporal_id. If we
+          // already have some target points at this temporal_id, we overwrite
+          // them with the new target points. This is because the interpolation
+          // has already finished for the previous set of target points if we
+          // are receiving new target points. This core must not have had any
+          // elements that contained any of the previous target points. However,
+          // these elements may now contain some new target points.
```
wthrowe: Could this happen in the opposite order?  We receive the second set of points (which need interpolation) first, and then overwrite them with the first set (which didn't)?

wthrowe (changes requested):

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
knelli2: Hmm let me think if we need to keep around the other members of the `Info` class when we receive a new set of points. I'm thinking no.

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
knelli2: Yes, I believe that could technically happen in the opposite order. I wonder if we should somehow send along the horizon find iteration so we know which set of points is the correct one (always the higher iteration)

knelli2:
@wthrowe Pushed a fixup that covers the scenario of receiving points in the wrong order.

wthrowe:
That should do it.  Go ahead and squash.

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
```diff
@@ -80,12 +89,34 @@ struct ReceivePoints {
...
+          // and we definitely didn't have any of the new target points in our
+          // element by the fact that we have already received the next
+          // iteration of points.
+          //
+          // Whenever we overwrite the target points, we also empty the
+          // `interpolation_is_done_for_these_elements` (by virtue of a default
+          // constructed `intrp::Vars::Info`) so that we always check every
+          // element for this new set of target points.
+          if (vars_infos.count(temporal_id) == 0 or
+              vars_infos.at(temporal_id).iteration < iteration) {
```
knelli2, line 117: Make it error if the new iteration is the same as the old iteration. This is a bug.

knelli2: @wthrowe Pushed a fixup for this

wthrowe: Looks good.

knelli2:
@wthrowe Squashed

AlexCarpenter46:
I've been testing this branch for a few different head on BBH's (different resolutions, enabling checkpoints and quiet/debug output) and they've all successfully reached a common horizon :). So I would say this has fixed the current deadlocks we've been encountering. These runs are on Ocean at `/home/alexcarpenter/BBH_Headon/BBH_develop_with_fix`

nilsvu:
Hoorayyy 🙌

nilsdeppe:
@wthrowe can you please take a look at this again?

wthrowe:
It's waiting on @markscheel?

nilsdeppe:
> @wthrowe Squashed

?

knelli2:
I removed Mark because this is a pretty important bug fix that I think we should get in quickly.

wthrowe:
If the PR author requests a review, I'm going to respect that.

wthrowe (approved):

nilsdeppe:
> If the PR author requests a review, I'm going to respect that.

That doesn't mean you can't give feedback post-squash, though.

PR #5533, opened 2023-10-03
Allow DenseTriggers mutable access to the DataBox

wthrowe:
This will be used to coordinate between the control systems and time-stepping code.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/EventsAndDenseTriggers/DenseTriggers/Filter.hpp
+++ b/src/Evolution/EventsAndDenseTriggers/DenseTriggers/Filter.hpp
@@ -11,6 +11,7 @@
 #include "Evolution/EventsAndDenseTriggers/DenseTrigger.hpp"
 #include "Options/String.hpp"
 #include "ParallelAlgorithms/EventsAndTriggers/Trigger.hpp"
+#include "Utilities/Gsl.hpp"
 #include "Utilities/Serialization/CharmPupable.hpp"
 #include "Utilities/TMPL.hpp"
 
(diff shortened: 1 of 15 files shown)


src/Evolution/EventsAndDenseTriggers/DenseTriggers/Filter.hpp
```diff
@@ -62,29 +63,31 @@ class Filter : public DenseTrigger {
   explicit Filter(std::unique_ptr<DenseTrigger> trigger,
                   std::unique_ptr<Trigger> filter);
 
-  using is_triggered_argument_tags = tmpl::list<Tags::DataBox>;
+  using is_triggered_return_tags = tmpl::list<Tags::DataBox>;
+  using is_triggered_argument_tags = tmpl::list<>;
```
knelli2, line 67: If `is_triggered` is being called with a `db::mutate_apply` now, and you pass in the box as a return tag, are you allowed to do `db::mutate` inside `is_triggered`? I thought no?

wthrowe: It's a special case added in #5478.  The `db::mutate` and `db::mutate_apply` functions don't lock the box if they don't actually mutate anything.  So you can take the box as a return, but you can't take anything else at the same time.

knelli2 (approved):

knelli2:
Ignoring clang-tidy (it's inside charm) and failed tests.

Issue #5520, opened 2023-09-30
RadiallyFallingFloor and ParameterizedDeloptonization do not adjust temperature


nilsdeppe:
# Bug reports:

### Expected behavior:

They should both be modifying _all_ the primitive variables for consistency. `FixToAtmosphere` is also broken, but I'm fixing that right now.

### Current behavior:

After the fixing the temperature is in an inconsistent state with the other primitives.

PR #5534, opened 2023-10-03
Temperature updates for param delep 3D EoS

pajkosmi:
## Proposed changes

Make parameterized deleptonization compatible with 3D eos and update the temperature in the 3D eos case.   Fixes 2nd half of issue #5520.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

The temperature only needs to be updated for 3D eos cases because a change in electron fraction will not affect the other primitives for 1D/2D eos cases.

### DIFF
--- a/src/Evolution/VariableFixing/ParameterizedDeleptonization.cpp
+++ b/src/Evolution/VariableFixing/ParameterizedDeleptonization.cpp
@@ -66,24 +66,27 @@ void ParameterizedDeleptonization::operator()(
     const gsl::not_null<Scalar<DataVector>*> electron_fraction,
     const gsl::not_null<Scalar<DataVector>*> pressure,
     const gsl::not_null<Scalar<DataVector>*> specific_enthalpy,
+    const gsl::not_null<Scalar<DataVector>*> temperature,
     const Scalar<DataVector>& rest_mass_density,
     const EquationsOfState::EquationOfState<true, ThermodynamicDim>&
         equation_of_state) const {
   if (LIKELY(enable_)) {
     for (size_t i = 0; i < rest_mass_density.get().size(); i++) {
       correct_electron_fraction(specific_internal_energy, electron_fraction,
-                                pressure, specific_enthalpy, rest_mass_density,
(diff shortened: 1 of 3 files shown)


src/Evolution/VariableFixing/ParameterizedDeleptonization.cpp
```diff
@@ -122,6 +124,15 @@ void ParameterizedDeleptonization::correct_electron_fraction(
         get(equation_of_state.pressure_from_density_and_energy(
```
nilsdeppe, line 118: In the 1d and 2d case, is there even an EOS call needed? And why don't you need to adjust pressure in the 3d case? If I remember correctly, ultimately you want to keep entropy constant and change Y_e, which would mean pressure, epsilon, and T change. Here it looks like you're wanting to fix epsilon and then vary T and p

tests/Unit/Evolution/VariableFixing/Test_ParameterizedDeleptonization.cpp
```diff
@@ -71,22 +84,63 @@ void test_variable_fixer(
...
+  auto specific_enthalpy = hydro::relativistic_specific_enthalpy(
+      density, specific_internal_energy, pressure);
+
+  // check Ye values that will naturally decrease
+  variable_fixer(&specific_internal_energy, &electron_fraction, &pressure,
+                 &specific_enthalpy, &temperature, density, equation_of_state);
+
+  const Scalar<DataVector> expected_electron_fraction{
+      DataVector{0.49, 0.3077746733472282, 0.2}};
+
```
nilsdeppe, line 134: Also check that the values you expect to change, change. E.g. epsilon is the same and T&p change or something

nilsdeppe (changes requested):

nilsdeppe (commented):
LGTM please squash!

pajkosmi:
Squashed

nilsdeppe (approved):

PR #5536, opened 2023-10-04
Don't require FFTW unless we are linking FUKA

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/FindFUKA.cmake
+++ b/cmake/FindFUKA.cmake
@@ -24,17 +24,19 @@ find_library(
 
 # Link MPI (should be the same MPI that FUKA was built with)
 find_package(MPI COMPONENTS C)
-find_package(FFTW)
 
-if (FUKA_LIB AND MPI_C_FOUND AND FFTW_FOUND)
-  add_library(FUKA::Exporter INTERFACE IMPORTED)
-  target_link_libraries(
-    FUKA::Exporter
-    INTERFACE
-    MPI::MPI_C
-    FFTW::FFTW
-    ${FUKA_LIB}
-  )
+if (FUKA_LIB AND MPI_C_FOUND)
+  find_package(FFTW)
+  if (FFTW_FOUND)
+    add_library(FUKA::Exporter INTERFACE IMPORTED)
+    target_link_libraries(
+      FUKA::Exporter
+      INTERFACE
+      MPI::MPI_C
+      FFTW::FFTW
+      ${FUKA_LIB}
+    )
+  endif()
 endif()
 
 include(FindPackageHandleStandardArgs)


cmake/FindFUKA.cmake
```diff
@@ -24,17 +24,19 @@ find_library(
...
-  add_library(FUKA::Exporter INTERFACE IMPORTED)
-  target_link_libraries(
-    FUKA::Exporter
-    INTERFACE
-    MPI::MPI_C
-    FFTW::FFTW
-    ${FUKA_LIB}
-  )
+if (FUKA_LIB AND MPI_C_FOUND)
+  find_package(FFTW)
```
nilsvu, line 29: What issue did you have here specifically?

The change seems fine, in fact I haven't found best-practice examples how to handle dependencies like this. There's [`find_dependency`](https://cmake.org/cmake/help/latest/module/CMakeFindDependencyMacro.html) but that's not usable here (there's even a note in the docs)...

nilsvu (approved):

cmake/FindFUKA.cmake
nilsdeppe: I just don't have FFTW installed and since I don't have/use FUKA on my machine I'd like to minimize what we/I need to install for spectre. Basically, trying to minimize dependencies that aren't actually necessary :)

nilsvu: Sure, you didn't get any warnings though, right? CMake just tried to find FFTW, didn't find it, and printed a status message (not a warning)?

nilsdeppe: Yes, that's right. CMake errored with not being able to find FFTW!

nilsvu: Wait it errored? Not just printed a status message? That's odd because FFTW is not marked `REQUIRED`...

nilsvu: In the container it does this:

```
-- Found MPI_C: /usr/lib/x86_64-linux-gnu/openmpi/lib/libmpi.so (found version "3.1") 
-- Found MPI: TRUE (found version "3.1") found components: C 
-- Could NOT find FFTW (missing: FFTW_INCLUDE_DIR FFTW_LIB) 
-- Could NOT find FUKA (missing: FUKA_LIB) 
```

(https://github.com/sxs-collaboration/spectre/actions/runs/6396902065/job/17363695098)

Then happily continues. That didn't happen on your machine?

nilsdeppe: Oof, yea, now I don't remember 😅 It at least looked very unhappy. I thought it failed, but maybe I'm wrong. It was definitely intrusive enough that I got the impression things broke

nilsvu: Ok this change makes it better either way

PR #5543, opened 2023-10-06
Add constraint energy normalization in GH

acifajkya:
## Proposed changes

Add the generalized harmonic constraint energy normalization.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/Constraints.cpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/Constraints.cpp
@@ -10,6 +10,7 @@
 #include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
 #include "PointwiseFunctions/GeneralRelativity/GeneralizedHarmonic/Christoffel.hpp"
 #include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 #include "Utilities/MakeWithValue.hpp"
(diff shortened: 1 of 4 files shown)


src/Evolution/Systems/GeneralizedHarmonic/Constraints.cpp
```diff
@@ -1325,6 +1326,60 @@ void constraint_energy(
...
 }
+
+template <typename DataType, size_t SpatialDim, typename Frame>
+Scalar<DataType> constraint_energy_normalization(
+    const tnsr::iaa<DataType, SpatialDim, Frame>& d_spacetime_metric,
+    const tnsr::iaa<DataType, SpatialDim, Frame>& d_pi,
+    const tnsr::ijaa<DataType, SpatialDim, Frame>& d_phi,
+    const tnsr::II<DataType, SpatialDim, Frame>& inverse_spatial_metric,
+    const Scalar<DataType>& sqrt_spatial_metric_determinant,
+    const double dimensional_constant) {
```
AlexCarpenter46, line 1337: At some point we'll need to add a function for computing the dimensional constant for evolutions where the total mass doesn't equal 1, but for now defaulting to 1.0 seems fine.

tests/Unit/Evolution/Systems/GeneralizedHarmonic/Test_Constraints.cpp
```diff
@@ -748,6 +748,52 @@ void test_constraint_energy_analytic(const Solution& solution,
...
+          const tnsr::ijaa<DataType, SpatialDim, Frame>&,
+          const tnsr::II<DataType, SpatialDim, Frame>&, const Scalar<DataType>&,
+          double)>(
+          &gh::constraint_energy_normalization<DataType, SpatialDim, Frame>),
+      "TestFunctions", "constraint_energy_normalization", {{{-1.0, 1.0}}},
+      used_for_size);
+}
+
+// Test the return-by-value normalized constraint energy function for Minkowski
+void test_constraint_energy_normalization_minkowski(
```
AlexCarpenter46, line 769: I don't think this test is needed since the random values will cover it. Also, I'm not sure if the dimensional constant is defined as you go to flat space.

AlexCarpenter46 (changes requested):
Just one tiny thing! Overall looks good to me, great work :)

tests/Unit/Evolution/Systems/GeneralizedHarmonic/Test_Constraints.cpp
wthrowe: Don't drop known-special-case tests just because there are python tests.  The python tests rely on there not being the same bug in the C++ and python implementations, which we have no way to verify (and which has happened).

(I haven't followed this PR and don't know whether this specific test makes sense.)

knelli2: Yes I agree keeping this specific test is a good idea.

src/Evolution/Systems/GeneralizedHarmonic/Constraints.hpp
```diff
@@ -551,6 +551,44 @@ void constraint_energy(
     double four_index_constraint_multiplier = 1.0);
 /// @}
 
+/// @{
+/*!
+ * \brief Computes the generalized-harmonic normalized constraint energy.
+ *
+ * \details Computes the generalized-harmonic normalized constraint energy.
+ * [Eq. (70) of \cite Lindblom2005qh with \f$m^{ab}=\delta^{ab}\f$,
```
knelli2: Extra `[` or missing `]` here?

src/Evolution/Systems/GeneralizedHarmonic/Constraints.hpp
```diff
@@ -551,6 +551,44 @@ void constraint_energy(
...
+ * \f{eqnarray}{
+ * ||E||=\gamma^{ij}\delta^{ab}\delta^{cd}
+ * (\Lambda^{2}\partial_{i}g_{ac}\partial_{j}g_{bd}
+ * + \partial_{i}\Pi_{ac}\partial_{j}\Pi_{bd} +
+ * \gamma^{kl}\partial_{i}\Phi_{kac}\partial_{j}\Phi_{lbd})\gamma^{1/2}
```
knelli2, line 564: This equation looks like it compute the integrand of equation 70 of the reference above, not the integral itself. If that's what was intended, add to the docs that this is the integrand, not the integral.

src/Evolution/Systems/GeneralizedHarmonic/Constraints.hpp
```diff
@@ -551,6 +551,44 @@ void constraint_energy(
...
+ * with the Kronecker delta.
+ */
+template <typename DataType, size_t SpatialDim, typename Frame>
+Scalar<DataType> constraint_energy_normalization(
+    const tnsr::iaa<DataType, SpatialDim, Frame>& d_spacetime_metric,
+    const tnsr::iaa<DataType, SpatialDim, Frame>& d_pi,
+    const tnsr::ijaa<DataType, SpatialDim, Frame>& d_phi,
+    const tnsr::II<DataType, SpatialDim, Frame>& inverse_spatial_metric,
+    const Scalar<DataType>& sqrt_spatial_metric_determinant,
+    double dimensional_constant = 1.0);
```
knelli2: I would remove the default here. I can imagine a scenario where we use this in a compute tag, but assume equal mass and then try a run with unequal mass but forget to pass in the new parameter and this normalization is then wrong. (Also update the docs)

knelli2 (changes requested):
You can squash these changes in directly.

src/Evolution/Systems/GeneralizedHarmonic/Constraints.hpp
knelli2: Ping on this.

knelli2 (changes requested):

knelli2 (approved):

AlexCarpenter46 (approved):

PR #5544, opened 2023-10-06
Add expiration_after method to FunctionOfTime

wthrowe:
Necessary for deterministically coupling the evolution step size to the control system updates.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/FunctionOfTime.hpp
+++ b/src/Domain/FunctionsOfTime/FunctionOfTime.hpp
@@ -55,6 +55,16 @@ class FunctionOfTime : public PUP::able {
   /// interval.
   virtual std::array<double, 2> time_bounds() const = 0;
 
+  /// \brief The first expiration time after \p time.
+  ///
+  /// \details For non-updatable functions, this returns infinity. For
+  /// updatable functions, the first expiration time after \p time is
+  /// found by determining the update immediately before \p time. The
+  /// expiration time of this update is what is returned. If \p time
+  /// happens to be an update itself, then the expiration of that
+  /// update is returned.
+  virtual double expiration_after(double time) const = 0;
+
   /// Updates the maximum derivative of the FunctionOfTime at a given time while
   /// also resetting the expiration. By default, a FunctionOfTime cannot be
   /// updated.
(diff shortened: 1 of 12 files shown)


src/Domain/FunctionsOfTime/FunctionOfTime.hpp
```diff
@@ -55,6 +55,12 @@ class FunctionOfTime : public PUP::able {
   /// interval.
   virtual std::array<double, 2> time_bounds() const = 0;
 
+  /// The smallest value `time_bounds()[1]` could have that is greater
+  /// than \p time.  For updatable functions of time, this will be the
+  /// expiration time of the last update before or at \p time.  For
+  /// non-updatable functions, this will be infinity.
+  virtual double expiration_after(double time) const = 0;
```
knelli2: Maybe don't make this pure virtual. Make the default return infinity like how we have the `update` function error by default. Then you don't have to make changes to `FixedSpeedCubic` or `SettleToConstant`.

src/Domain/FunctionsOfTime/FunctionOfTime.hpp
```diff
@@ -55,6 +55,12 @@ class FunctionOfTime : public PUP::able {
   /// interval.
   virtual std::array<double, 2> time_bounds() const = 0;
 
+  /// The smallest value `time_bounds()[1]` could have that is greater
+  /// than \p time.  For updatable functions of time, this will be the
+  /// expiration time of the last update before or at \p time.  For
+  /// non-updatable functions, this will be infinity.
```
knelli2: I found this description a bit confusing because it was too compressed. What about something like

```
\brief The first expiration time after \p time.

\details For non-updatable functions, this returns infinity. For updatable
functions, the first expiration time after \p time is found by determining the
update immediately before \p time. The expiration time of this update is what
is returned. If \p time happens to be an update itself, then the expiration of
that update is returned
```

knelli2 (changes requested):

src/Domain/FunctionsOfTime/FunctionOfTime.hpp
wthrowe: Used your version.

src/Domain/FunctionsOfTime/FunctionOfTime.hpp
wthrowe: I'd rather not, on the general defaults-are-bad reasoning, and since you're only net saving one three-line function.  If you forget to implement `update` it's going to be really obvious.  This is much more subtle.

wthrowe (commented):
Squashed in the doc change.

knelli2 (approved):

Issue #5021, opened 2023-05-19
Automate BBH simulations


nilsvu:
Continuing discussion from #2987.

Remaining tasks are:

- [ ] Continue ringdown from the last inspiral segment. Probably first by running a new executable. We can either continue the segments or run in a separate ringdown directory.
- [ ] Sort out directory structure for eccentricity reduction and initial data. This should remain quite flexible (scripts shouldn't make any assumptions about this directory structure).
- [ ] Propagate scheduling options like num nodes.

nilsvu:
Discussion with @nilsdeppe @markscheel @wthrowe:

- We converged on a concept of "pipelines" for BBH, BNS, BHNS, supernovae, etc. A pipeline is essentially an automation driver that runs after a simulation has finished and calls into other spectre Python modules to do the actual work (like scheduling simulations, archiving, etc).
- We might call a step in a pipeline an "epoch" (or just "step" or "stage"?).
- The pipeline continues epochs, segments, levs, etc. It uses Python modules to share logic between different pipelines (eccentricity reduction, etc). The pipeline is responsible for choosing directory names etc.

The directory structure for a BNS pipelines might look like this:

```sh
# Ecc reduction
Epoch_001_Id/  # (with or without suffixes)
Epoch_002_Ev/  # Runs at Lev_01
Epoch_003_Id/
Epoch_003_Ev/  # Runs at Lev_05
# Start all levs from end of Epoch_003 (PBandJ)
# Start at t = t_pbandj > 0 (don't relabel to t=0)
Epoch_004_Ev/Lev_01...04
    Lev_01/Segment_0000
         Segment_0001
         # ...
         # Switch to single BH
         Segment_0045
    Lev_02/Segment_0000
         Segment_0001
         # ...
         # Switch to single NS
         Segment_0048
         # ...
         # Switch to single BH
         Segment_0089
    # Or organize with stages when switching executables:
    Lev_02/
        Stage0/
            Segment_0000
            Segment_0001
            # ...
        # Switch to single NS
        Stage1/
            Segment_0000
            # ...
        # Switch to single BH
        Stage2/
            Segment_0000
Epoch_005_AssembleForArchive/Lev*
# Maybe a higher-level tool is responsible for the rest of the archiving:
Epoch_006_Archive/
Epoch_007_PushToCatalog/
```

- When a simulation ends with exit code 0 the submit script calls the pipeline specified in the input file metadata:

```yaml
# Metadata in input files:
[...]

Issue #4600, opened 2023-01-12
Ringdown: continue an evolution from common horizon


geoffrey4444:
# Feature request:

Common horizon finding items:
- [x] DenseTrigger/Event for finding/dumping AhC.
    - This dense trigger also ideally will terminate the simulation when AhC is spherical enough, or when constraints are getting to big, as in SpEC. 
    - For now, just wait until it dies.
- [x] Store volume data when common horizon find is triggered. If no horizon is found, callback will delete the stored volume data (we don't need it). If horizon is found, then we keep the volume data.
    - For now, we know when the common horizon forms, so we can just turn on AhC + volume data output right when we need it at the end of the inspiral, after AhC forms
    - Don't delete AhC volume data for now. We can consider deleting it after we know ringdown is successful, but not necessary.

SpEC Termination criterion (not needed for q=1, zero spin BBH):
- A certain number of common horizon finds
- Constraints don't get too big
- Lmax of the common horizon gets below a threshold (is spherical enough)

Initializing ringdown domain (in post-processing script):
- [x] Dump all map coefficients
- [x] Need several AhC finds (as described above) to initialize shape and translation map coefs (and derivatives) for ringdown domain.
- [x] For shape, need #5183 .
- [x] Translation and quaternion rotation function of time are `SettleToConstant`s and cannot be updated, need #4598 #4599
- [x] Also need Expansion/Scaling #4598 #4599 and #5670
- [ ] Need to compute excision radius of new domain (degenerate with 00 component of shape map, SpEC chooses 00 component = 0). #5670
- [ ] Need c++ function that knows about old domain and new domain and does iterative procedure to find excision radius. #5670

Deal with time stepper history initialization:
- Don't. Use RK instead.
- Could use RK initially and use that to initialize history for LMS methods.

### Desired feature:

- [ ] After spectre can initialize a ringdown from common horizon initial data (#4598, #4599), combine these pieces so that the single black hole executable can successfully start the evolution.

### Detailed discussion:
Note: Until spectre can evolve to common horizon, this can be tested by interpolating SpEC data with a common horizon.

nilsvu:
- We start with doing this in a separate executable.
- Loading numeric data already works.
- With the two separate executables we can automate initializing control systems etc. by writing a ringdown input file from measurements of the evolution. We need to initialize Strahlkorpers from a file specified in the input file.
- Use pipeline infrastructure discussed in #5021.
[...]

PR #5546, opened 2023-10-07
Add skeleton BBH pipeline

nilsvu:
## Proposed changes

Tools to `generate-id` > `start-inspiral` > `start-ringdown`. No automation yet and barely any smart parameter choices, but this should provide a good starting point to flesh out the transition from inspiral to ringdown. See https://github.com/sxs-collaboration/spectre/issues/5021 for details on future plans.

Try it like this:

```sh
./bin/spectre bbh generate-id -q 1 -D 20 -w 0.015 -o ./ID
# wait...
./bin/spectre bbh start-inspiral ./ID/InitialData.yaml -O ./Ev/
# wait...
./bin/spectre bbh start-ringdown ./Ev/Inspiral.yaml -O ./Ringdown/
```

(this won't really work yet because the inspiral won't finish properly, but that's #4600)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/InitialData.py
+++ b/support/Pipelines/Bbh/InitialData.py
@@ -1,0 +1,151 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+from pathlib import Path
+from typing import Union
+
+import click
+from rich.pretty import pretty_repr
+
+from spectre.support.Schedule import schedule, scheduler_options
+
+logger = logging.getLogger(__name__)
+
+ID_INPUT_FILE_TEMPLATE = Path(__file__).parent / "InitialData.yaml"
+
+
+def id_parameters(
+    mass_ratio: float,
+    separation: float,
+    orbital_angular_velocity: float,
+    refinement_level: int,
+    polynomial_order: int,
+):
+    """Determine initial data parameters from options.
+
+    These parameters fill the 'ID_INPUT_FILE_TEMPLATE'.
+
+    Arguments:
+      mass_ratio: Defined as q = M_A / M_B >= 1.
+      separation: Coordinate separation D of the black holes.
+      orbital_angular_velocity: Omega_0.
+      refinement_level: h-refinement level.
+      polynomial_order: p-refinement level.
+    """
+
+    # Sanity checks
+    assert mass_ratio >= 1.0, "Mass ratio is defined to be >= 1.0."
+
+    # Determine initial data parameters from options
+    M_A = mass_ratio / (1.0 + mass_ratio)
+    M_B = 1.0 / (1.0 + mass_ratio)
+    x_A = separation / (1.0 + mass_ratio)
+    x_B = x_A - separation
+    return {
+        "MassRight": M_A,
+        "MassLeft": M_B,
+        "XRight": x_A,
+        "XLeft": x_B,
+        "ExcisionRadiusRight": 0.89 * 2.0 * M_A,
+        "ExcisionRadiusLeft": 0.89 * 2.0 * M_B,
+        "OrbitalAngularVelocity": orbital_angular_velocity,
+        # Resolution
+        "L": refinement_level,
+        "P": polynomial_order,
+    }
+
+
+def generate_id(
+    mass_ratio: float,
+    separation: float,
+    orbital_angular_velocity: float,
+    refinement_level: int,
+    polynomial_order: int,
+    id_input_file_template: Union[str, Path] = ID_INPUT_FILE_TEMPLATE,
+    **scheduler_kwargs,
+):
+    """Generate initial data for a BBH simulation.
+
+    Parameters for the initial data will be inserted into the
+    'id_input_file_template'. The remaining options are forwarded to the
+    'schedule' command. See 'schedule' docs for details.
+    """
+    logger.warning(
+        "The BBH pipeline is still experimental. Please review the"
+        " generated input files."
+    )
+
+    # Determine initial data parameters from options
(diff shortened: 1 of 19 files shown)


support/Pipelines/Bbh/InitialData.py
```diff
@@ -0,0 +1,143 @@
...
+    )
+    logger.debug(f"Initial data parameters: {pretty_repr(id_params)}")
+
+    # Schedule!
+    return schedule(id_input_file_template, **id_params, **scheduler_kwargs)
+
+
+@click.command(name="generate-id", help=generate_id.__doc__)
+@click.option(
+    "--mass-ratio",
```
knelli2, line 95: How do you feel about using another yaml to specify these? As we add more and more capability, having to write a new click option for every nob we want to tweak will probably get pretty annoying. This could simplify the command down to

```sh
spectre bbh generate-id InitialParams.yaml
```

where `InitialParams.yaml` is

```yaml
mass-ratio: 1.5
separation: 16
orbital-angular-velocity: 0.01
polynomial-order: 7
```

and then all the other parameters we didn't specify will be default.

This suggestion applies to all phases of the BBH pipeline.

support/Pipelines/Bbh/InitialData.py
```diff
@@ -0,0 +1,143 @@
...
+    "-w",
+    type=float,
+    help="Orbital angular velocity Omega_0.",
+)
+@click.option(
+    "--refinement-level",
+    "-L",
+    type=int,
+    help="h-refinement level.",
+    default=0,
```
knelli2, line 120: Not to be figured out for this PR, but do we want to allow defaults? We explicitly don't in our executable input files for a variety of reasons, and this sort of gets around that.

support/Pipelines/Bbh/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+    """Pipeline for binary black hole simulations."""
+    pass
+
+
+bbh_pipeline.add_command(generate_id_command)
+bbh_pipeline.add_command(start_inspiral_command)
+bbh_pipeline.add_command(start_ringdown_command)
+
+if __name__ == "__main__":
+    bbh_pipeline(help_option_names=["-h", "--help"])
```
knelli2, line 22: Maybe print a pretty click warning somewhere that this is still very experimental and shouldn't be used for actual runs yet?

knelli2 (commented):
I think this is a great start! We'll definitely need to tune input file parameters and such, but that can be done as we go.

support/Pipelines/Bbh/InitialData.py
nilsvu: Ohh no, adding another "meta" input file seems like it could be a disaster :D I'd rather keep the knobs here down, and even remove the hp refinement options once we have AMR. If you want full control over options you can write your own input file and run it. The steps in the pipeline still work if you're starting them from a simulation that you didn't run with the pipeline.

support/Pipelines/Bbh/InitialData.py
nilsvu: My plan for these resolution options is to replace them with something like "lev" or "target truncation error" once we have AMR. I can remove the default here for now, no strong opinion. BTW a big reason for avoiding defaults in input files is that they're so hard to discover and can change your simulation in subtle ways. Here they're just a `--help` away. CLIs use defaults all the time (still have to be careful!).

support/Pipelines/Bbh/__init__.py
nilsvu: It's not more experimental than the BBH input file that you would copy from tests/InputFiles :D

support/Pipelines/Bbh/InitialData.py
nilsvu: I would be fine with adding an option to supply the CLI option from a configuration file rather than on the command line if you want. We  already do something like that with the user config file and the status CLI. It's just an alternative way of passing the options to the CLI. There's a pretty generic way to do that with click.

knelli2: Yeah I think that would be nice.

support/Pipelines/Bbh/__init__.py
knelli2: It's not the input file that's experimental, it's the pipeline. My idea of a pipeline is smooth seamless transitions with little input required from the user. At the moment, this isn't that. But I hope it will be!

nilsvu: Ah ok, agreed! Added some warnings.

support/Pipelines/Bbh/InitialData.py
knelli2: Did you want to make this change now or in a follow up PR?

knelli2 (commented):
This is a great start!

nilsvu:
Fixed the MPI failure (I hope)

support/Pipelines/Bbh/InitialData.py
nilsvu: I wasn't planning to do this right away, since I don't have a use case right now

nilsdeppe: I agree, could be nice, first need to see how all this works in practice :)

nilsdeppe (dismissed):
I didn't read it all in fine detail, but this seems like a good way to structure and organize all this.

nilsvu:
Rebased

knelli2 (approved):

PR #5565, opened 2023-10-13
Allow events to mutate the DataBox

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/DataBox/ObservationBox.hpp
+++ b/src/DataStructures/DataBox/ObservationBox.hpp
@@ -5,7 +5,10 @@
 
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "DataStructures/DataBox/Item.hpp"
+#include "DataStructures/DataBox/TagTraits.hpp"
+#include "Utilities/CleanupRoutine.hpp"
 #include "Utilities/ErrorHandling/StaticAssert.hpp"
+#include "Utilities/Gsl.hpp"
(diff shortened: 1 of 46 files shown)


src/DataStructures/DataBox/ObservationBox.hpp
```diff
@@ -45,6 +48,9 @@ class ObservationBox<tmpl::list<ComputeTags...>, DataBoxType>
  public:
   /// A list of all the compute item tags
   using compute_item_tags = tmpl::list<ComputeTags...>;
+  /// A list of all the compute item tags that are compute items
+  using actual_compute_item_tags =
```
knelli2: Can we come up with a better naming scheme than `thing` and `actual_thing`? If the original `compute_item_tags` isn't only compute items, maybe rename it to just `item_tags`? The class template pack should probably also change then since they aren't necessarily `ComputeTags`.

knelli2 (changes requested):

src/DataStructures/DataBox/ObservationBox.hpp
wthrowe: Since the new alias was only used in one place, I just inlined it.

I think renaming to `item_tags` would be confusing as there is already a `tags_list` that means something else.

knelli2 (approved):

PR #5566, opened 2023-10-13
Store control-system measurement times in the DataBox

wthrowe:
Knowing the times will be necessary for implementing time-step adjustment to avoid deadlocking.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/FutureMeasurements.cpp
+++ b/src/ControlSystem/FutureMeasurements.cpp
@@ -1,0 +1,67 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ControlSystem/FutureMeasurements.hpp"
+
+#include <limits>
+#include <pup_stl.h>
+
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/StdHelpers.hpp"
+
+namespace control_system {
+FutureMeasurements::FutureMeasurements(const size_t measurements_per_update,
+                                       const double first_measurement_time)
+    : measurements_{std::numeric_limits<double>::signaling_NaN(),
+                    first_measurement_time},
+      measurements_until_update_(measurements_per_update - 1),
+      measurements_per_update_(measurements_per_update) {
+  ASSERT(measurements_per_update > 0, "Cannot update without measurements.");
+}
+
+std::optional<double> FutureMeasurements::next_measurement() const {
+  if (measurements_.size() <= 1) {
+    return std::nullopt;
+  }
+  // First entry is an old measurement.
+  return measurements_[1];
+}
+
+std::optional<double> FutureMeasurements::next_update() const {
+  if (measurements_.size() <= measurements_until_update_ + 1) {
+    return std::nullopt;
+  }
+  return {measurements_[measurements_until_update_ + 1]};
+}
+
+void FutureMeasurements::pop_front() {
+  ASSERT(measurements_.size() > 1, "Popped an empty container.");
+  measurements_.pop_front();
+  if (measurements_until_update_ == 0) {
+    measurements_until_update_ = measurements_per_update_ - 1;
+  } else {
+    --measurements_until_update_;
+  }
+}
+
+void FutureMeasurements::update(
+    const domain::FunctionsOfTime::FunctionOfTime& measurement_timescale) {
+  ASSERT(not measurements_.empty(), "Don't know current measurement time.");
+  // Limit the number of stored measurements to a few cycles.  We
+  // shouldn't need more than one cycle, but tuning the exact size
+  // isn't necessary.
+  while (measurement_timescale.time_bounds()[1] >= measurements_.back() and
+         measurements_.size() < 3 * measurements_per_update_) {
+    measurements_.emplace_back(
+        measurements_.back() +
+        measurement_timescale.func(measurements_.back())[0][0]);
+  }
+}
+
+void FutureMeasurements::pup(PUP::er& p) {
+  p | measurements_;
+  p | measurements_until_update_;
+  p | measurements_per_update_;
+}
+}  // namespace control_system
--- a/src/ControlSystem/FutureMeasurements.hpp
+++ b/src/ControlSystem/FutureMeasurements.hpp
@@ -1,0 +1,66 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <deque>
+#include <optional>
+#include <pup.h>
+#include <utility>
+
+/// \cond
+namespace domain::FunctionsOfTime {
+class FunctionOfTime;
+}  // namespace domain::FunctionsOfTime
+/// \endcond
+
+namespace control_system {
+/// Class for computing the upcoming measurement times for a control
+/// system measurement.
+///
+/// At any time, the \ref control_system::Tags::MeasurementTimescales
+/// "measurement timescales" can be queried to find the spacing
+/// between control system measurements.  Most uses, however, require
+/// the actual measurement times, and sometimes need to know times of
+/// control-system update triggers.  This class calculates these
+/// quantities from the timescales.
+class FutureMeasurements {
+ public:
+  FutureMeasurements() = default;
+
+  FutureMeasurements(size_t measurements_per_update,
+                     double first_measurement_time);
+
+  /// Next measurement time, if known.
+  std::optional<double> next_measurement() const;
+  /// Next measurement that triggers an update, if known.
+  std::optional<double> next_update() const;
+
+  /// Remove the earliest measurement form the list, generally because
+  /// it has been performed.
+  void pop_front();
+
+  /// Calculate and store measurement times up through the expiration
+  /// time of the argument.
+  ///
+  /// Given a measurement time \f$t_i\f$, the next measurement will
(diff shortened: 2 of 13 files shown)


src/ControlSystem/FutureMeasurements.hpp
```diff
@@ -0,0 +1,54 @@
...
+
+/// \cond
+namespace domain::FunctionsOfTime {
+class FunctionOfTime;
+}  // namespace domain::FunctionsOfTime
+/// \endcond
+
+namespace control_system {
+/// Information about upcoming measurements.
+class FutureMeasurements {
```
knelli2: Can you expand on this a bit? What could this potentially be used for (what is it intended to be used for)? Maybe say that this is also aware of updates as well as measurements?

src/ControlSystem/FutureMeasurements.hpp
```diff
@@ -0,0 +1,54 @@
...
+ public:
+  FutureMeasurements() = default;
+
+  FutureMeasurements(size_t measurements_per_update,
+                     double first_measurement_time);
+
+  // Next measurement time, if known.
+  std::optional<double> next_measurement() const;
+  // Next measurement that triggers an update, if known.
```
knelli2: Doxygen for both of these?

src/ControlSystem/FutureMeasurements.cpp
```diff
@@ -0,0 +1,65 @@
...
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/StdHelpers.hpp"
+
+namespace control_system {
+FutureMeasurements::FutureMeasurements(const size_t measurements_per_update,
+                                       const double first_measurement_time)
+    : measurements_{std::numeric_limits<double>::signaling_NaN(),
+                    first_measurement_time},
+      measurements_until_update_(measurements_per_update - 1),
```
knelli2, line 18:
```cpp
ASSERT(measurements_per_update > 0, ...)
```
? Obviously other things should break if this was actually false, but would be good as a safety check.

tests/Unit/ControlSystem/Actions/Test_InitializeMeasurements.cpp
```diff
@@ -234,11 +257,25 @@ SPECTRE_TEST_CASE("Unit.ControlSystem.InitializeMeasurements",
...
-  CHECK(Submeasurement<tmpl::list<SystemA, SystemB>>::call_count == 1);
-  CHECK(Submeasurement<tmpl::list<SystemC>>::call_count == 1);
+  CHECK(Submeasurement<tmpl::list<SystemA, SystemB>>::call_count ==
+        (ab_active ? 1 : 0));
+  CHECK(Submeasurement<tmpl::list<SystemC>>::call_count == (c_active ? 1 : 0));
+}
+
+SPECTRE_TEST_CASE("Unit.ControlSystem.InitializeMeasurements",
+                  "[ControlSystem][Unit]") {
+  test_initialize_measurements(true, true);
```
knelli2, line 287: [optional] I know it's the same number of lines, but it's more easy to generalize in the future and it's clear what the two arguments are for.
```cpp
  for (const auto& [ab_active, c_active] :
       cartesian_product(make_array(true, false), make_array(true, false))) {
    test_initialize_measurements(ab_active, c_active);
  }
```

src/ControlSystem/Trigger.hpp
```diff
@@ -83,125 +75,92 @@ class Trigger : public DenseTrigger {
...
+          measurement_times,
+      const double time) {
+    if (measurement_times->next_measurement() == std::optional(time)) {
+      measurement_times->pop_front();
     }
 
-    const bool triggered = time == *next_trigger_;
-    if (triggered) {
-      *next_trigger_ = next_measurement(time, measurement_timescales);
+    if (not measurement_times->next_measurement().has_value()) {
```
knelli2, line 119: Will this work? Currently, every element has to wait for every FoT to be ready. Here, you're only checking the functions of time when these specific control systems need to be updated. But what about the other control systems? If we only check these specific FoTs on this element, then I believe your current implementation would work.

knelli2 (changes requested):

src/ControlSystem/Trigger.hpp
wthrowe: These are measurement timescales, not FoTs.

tests/Unit/ControlSystem/Actions/Test_InitializeMeasurements.cpp
wthrowe: I prefer the current version.

(And why `make_array(true, false)` instead of `std::array{true, false}`?)

src/ControlSystem/FutureMeasurements.hpp
wthrowe: I've expanded a little, but I don't really have much to say.  I generally prefer to avoid listing what uses a class in its documentation unless things are really tightly coupled, because that's not a property of the class and can end up out-of-date very easily.

Do you have a suggested wording for a statement about updates?  The "time of an update" seems to be ambiguous, since the SpEC update got split into two things happening at different times and I haven't observed a consensus about which that phrase would refer to.

knelli2: I think your fixup was a good `\brief` for a class, but it basically just repeated what the namespace and the class name already said: `control_system::FutureMeasurements` = class that knows about future control system measurements. In my opinion that's not super helpful documentation.

I view the docs as me (or an anthropomorphized version of the class) giving a newer dev of the code a (brief) explanation of how this class ties in to the surrounding infrastructure. You don't have to give specific uses of this class like "this class will be used in the control_system::Trigger class for ...". I agree that's hard to maintain. But you can give a broader idea of why this class is useful.

Here's how I would have documented the class. If you decide to use it, feel free to change up some wording. Hopefully the last paragraph removes the ambiguity of what a "time of an update" means.

```
Class for computing the upcoming measurement times for a control
system measurement.

The measurement timescales are stored in the GlobalCache. At a given simulation time,
they can be used to get the current measurement timescale. However, it is
sometimes advantageous to know when future measurements will occur in
case preemptive decisions have to be made. It is difficult to know how to
calculate these future measurements without additional information and
knowledge about the structure of the measurement timescales. This class
has this additional knowledge and takes care of that calculation. See the
`control_system::Tags::MeasurementTimescales` tag for this additional
information.

In addition to knowing about the simulation time of future measurements,
this class also knows the simulation time of control system/FunctionOfTime
updates as well.
```

tests/Unit/ControlSystem/Actions/Test_InitializeMeasurements.cpp
knelli2: No real reason. Just copied this from elsewhere that used the `cartesian_product`

knelli2 (commented):
You can squash the docs changes directly if you decide to make them.

src/ControlSystem/FutureMeasurements.hpp
wthrowe: > I think your fixup was a good `\brief` for a class, but it basically just repeated what the namespace and the class name already said: `control_system::FutureMeasurements` = class that knows about future control system measurements. In my opinion that's not super helpful documentation.

I added that it computes the values, which is an important point that was missing.

> ```
> The measurement timescales are stored in the GlobalCache.
> ```

I don't want to talk about the GlobalCache here.  Avoiding use of the GlobalCache in this class was an intentional decision I made.

> ```
> At a given simulation time,
> they can be used to get the current measurement timescale. However, it is
> sometimes advantageous to know when future measurements will occur in
> case preemptive decisions have to be made.
> ```

These two sentences are good.  I'll use something like them.

> ```
> It is difficult to know how to
> calculate these future measurements without additional information and
> knowledge about the structure of the measurement timescales. This class
> has this additional knowledge and takes care of that calculation. See the
> `control_system::Tags::MeasurementTimescales` tag for this additional
> information.
> ```

It's not factored out because it's difficult or confusing.  It's just `t + timescale(t)`.  It's factored out because it seems like a bad idea to calculate and store all the values in two places.

Although, now that I poke around, I don't actually see `t + timescale(t)` documented anywhere, and this does seem like a pretty good place for it.

Linking to the tag is a good idea, too.

>  ``` 
> In addition to knowing about the simulation time of future measurements,
> this class also knows the simulation time of control system/FunctionOfTime
> updates as well.
> ```

That's actually the "time of update" that it doesn't know about.  It knows about the measurement half.

wthrowe:
Pushed fixups.  Changes:
* Documentation update discussed above.  I decided to put the update formula on the `update` method.
* Changed inactive control systems to report one measurement per update.  I found this more convenient for code in a later PR.
* Changed the assertion about the list growing long to a non-error limit on the list size.  The class never uses more than the first cycle, but storing a bit more means no thought has to go into the exact number of entries required.  I also got worried that it might try to store every measurement time during a wave-zone step in LTS.

knelli2 (commented):
Looks good. Squash

knelli2 (approved):

knelli2:
Not sure what some of the test failures are because I can't see the output. But they failed building the executables

wthrowe:
For gcc-9 Release, "View raw logs" shows "g++-9: fatal error: Killed signal terminated program cc1plus", so presumably OOM.  Doesn't work for gcc-10 Release, though.

knelli2:
Ok sounds good. Nothing to block this PR

PR #5567, opened 2023-10-16
Explicitly state CCE ordering

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Tutorials/CCE.md
+++ b/docs/Tutorials/CCE.md
@@ -196,7 +196,15 @@ provided as spherical harmonic modes with the following datasets:
 - `DtLapse.dat`
 
 In this format, each row must start with the time stamp, and the remaining
-values are the complex modes in m-varies-fastest format.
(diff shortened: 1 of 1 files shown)


markscheel (approved):
This appears to be the current ordering of the `CceR*.h5` files (at least for the one file I looked at, from a q8_7d run).

Note that the `CceR*.h5` files contain "dat" objects, and each such object has a legend that specifies the ordering.  The safer thing to do would be to read the legend and assume the ordering in the legend, since someone might change the ordering, e.g. someone might decide to eliminate things like `Lapse_Im(0,0)` which are always zero.

But that said, it looks like ReduceCceWorldtube.cpp currently reads the legend but ignores it, so currently the ordering in this PR is what (presumably; it is nontrivial to check) is assumed by ReduceCceWorldtube.cpp.  Also, SpEC no longer outputs `CceR*.h5` files so at least for SpEC the ordering is not likely to change.

PR #5569, opened 2023-10-18
Allow multiple post interpolation callbacks

knelli2:
## Proposed changes

Doesn't actually add any, but adds support for adding more.

For reviewers, most of the important logic changes are in `InterpolationTargetDetail.hpp`. The rest is mostly just changing the name of the type alias and wrapping them in a `tmpl::list<>` .

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
+++ b/src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
@@ -95,38 +95,77 @@ double get_temporal_id_value(const T& id) {
 }
 
 CREATE_IS_CALLABLE(apply)
-CREATE_IS_CALLABLE_V(apply)
 
-// apply_callback accomplishes the overload for the
-// two signatures of callback functions.
 template <typename T, typename DbTags, typename Metavariables,
           typename TemporalId>
-bool apply_callback(
+bool apply_callbacks(
     const gsl::not_null<db::DataBox<DbTags>*> box,
     const gsl::not_null<Parallel::GlobalCache<Metavariables>*> cache,
     const TemporalId& temporal_id) {
-  if constexpr (is_apply_callable_v<typename T::post_interpolation_callback,
-                                    decltype(*box), decltype(*cache),
-                                    decltype(temporal_id)>) {
-    T::post_interpolation_callback::apply(*box, *cache, temporal_id);
+  using callbacks = typename T::post_interpolation_callbacks;
+  constexpr bool all_simple_callbacks =
+      tmpl::all<callbacks,
+                is_apply_callable<tmpl::_1, tmpl::pin<decltype(*box)>,
+                                  tmpl::pin<decltype(*cache)>,
+                                  tmpl::pin<decltype(temporal_id)>>>::value;
+  if constexpr (all_simple_callbacks) {
+    tmpl::for_each<callbacks>([&](const auto callback_v) {
+      using callback = tmpl::type_from<decltype(callback_v)>;
+      callback::apply(*box, *cache, temporal_id);
+    });
 
     // For the simpler callback function, we will always clean up volume data,
     // so we return true here.
     return true;
   } else {
-    return T::post_interpolation_callback::apply(box, cache, temporal_id);
+    static_assert(tmpl::size<callbacks>::value == 1,
+                  "Can only have 1 post_interpolation_callbacks that mutates "
+                  "the target box.");
+    return tmpl::front<callbacks>::apply(box, cache, temporal_id);
   }
 }
 
 CREATE_HAS_STATIC_MEMBER_VARIABLE(fill_invalid_points_with)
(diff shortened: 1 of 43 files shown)


src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -97,36 +97,85 @@ double get_temporal_id_value(const T& id) {
...
-// two signatures of callback functions.
+template <typename Callbacks, typename Box, typename Cache, typename TemporalId>
+struct AllSimpleCallbacks;
+
+template <typename... Callbacks, typename Box, typename Cache,
+          typename TemporalId>
+struct AllSimpleCallbacks<tmpl::list<Callbacks...>, Box, Cache, TemporalId>
+    : public std::integral_constant<
+          bool, tmpl2::flat_all_v<is_apply_callable_v<Callbacks, Box, Cache,
+                                                      TemporalId>...>> {};
```
wthrowe: [optional] You can just inherit from `tmpl2::flat_all<...>` and skip the `type -> bool -> type` conversion, since that already inherits from an `integral_constant`.  

As another alternative, you could use a fold expression.  I don't know if anyone's compared performance between those and `flat_all`, but these are small lists so it shouldn't matter.

And yet another alternative:  Is
```c++
AllSimpleCallbacks<callbacks, decltype(*box), decltype(*cache),
                   decltype(temporal_id)>
```
sufficiently clearer than
```c++
tmpl::all<callbacks,
          is_apply_callable<tmpl::_1, tmpl::pin<decltype(*box)>,
                            tmpl::pin<decltype(*cache)>,
                            tmpl::pin<decltype(temporal_id)>>
```
to be worth the extra struct?

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -97,36 +97,85 @@ double get_temporal_id_value(const T& id) {
...
+struct get_value {
+  using type =
+      typename with_value<Callback,
+                          has_fill_invalid_points_with_v<Callback>>::type;
+};
+
+template <typename Callbacks>
+struct get_invalid_fill {
+  using type =
+      tmpl::find<tmpl::flatten<tmpl::transform<Callbacks, get_value<tmpl::_1>>>,
```
wthrowe: What is being flattened here?  I think the argument is something along the lines of `tmpl::list<NoSuchType, DoubleWrapper<T>>`

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -97,36 +97,85 @@ double get_temporal_id_value(const T& id) {
...
+                  "the target box.");
+    return tmpl::front<callbacks>::apply(box, cache, temporal_id);
   }
 }
 
 CREATE_HAS_STATIC_MEMBER_VARIABLE(fill_invalid_points_with)
 CREATE_HAS_STATIC_MEMBER_VARIABLE_V(fill_invalid_points_with)
 
+template <typename T, bool B>
+struct with_value;
```
wthrowe: This namespace isn't super-specific, so name these structs something more descriptive.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -97,36 +97,85 @@ double get_temporal_id_value(const T& id) {
...
+  using type =
+      typename with_value<Callback,
+                          has_fill_invalid_points_with_v<Callback>>::type;
+};
+
+template <typename Callbacks>
+struct get_invalid_fill {
+  using type =
+      tmpl::find<tmpl::flatten<tmpl::transform<Callbacks, get_value<tmpl::_1>>>,
+                 tt::is_a<DoubleWrapper, tmpl::_1>>;
```
wthrowe: Is it worth checking for conflicting fill requests?  The code to do that might be kind of a pain.

wthrowe (changes requested):
Also CI is unhappy.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
knelli2: Oops. I have no clue why, but when I wrote this I thought `tmpl::flatten` did a `tmpl::filter` instead. I'll just remove it

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
knelli2: No I don't think it's worth it because all of the current callbacks that specify this fill value set it to `quiet_NaN`. And yeah it was going to be a pain to make it general so I just didn't.

wthrowe:
Looks good.  Squash.

wthrowe (approved):

PR #5573, opened 2023-10-18
Enable p-refinement for ScalarWave executables

kidder:
## Proposed changes

Enable p-refinement for ScalarWave executables.
Currently only works for global time-stepping.
Only the example input files for 1D actually do AMR during the evolution.  See them if you want to use AMR in 2D or 3D.

### Upgrade instructions

ScalarWave input files need to have additional options, as well as now using global time-steppers instead of local time-steppers.  (See the diff of `tests/InputFiles/ScalarWave/PlaneWave1D.yaml`)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
+++ b/src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
@@ -16,6 +16,8 @@
 #include "DataStructures/DataBox/DataBoxTag.hpp"
 #include "DataStructures/DataBox/Prefixes.hpp"
 #include "DataStructures/Variables.hpp"
+#include "Domain/Amr/Info.hpp"
+#include "Domain/Amr/Tags/NeighborFlags.hpp"
 #include "Domain/Creators/Tags/InitialExtents.hpp"
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/Element.hpp"
@@ -32,6 +34,7 @@
 #include "ParallelAlgorithms/Amr/Protocols/Projector.hpp"
 #include "ParallelAlgorithms/Initialization/MutateAssign.hpp"
 #include "Time/TimeStepId.hpp"
+#include "Utilities/Gsl.hpp"
 #include "Utilities/TMPL.hpp"
 
 /// \cond
@@ -183,7 +186,8 @@ struct ProjectMortars : tt::ConformsTo<amr::protocols::Projector> {
                  evolution::dg::Tags::NormalCovectorAndMagnitude<dim>,
                  Tags::MortarDataHistory<dim, typename dt_variables_tag::type>>;
   using argument_tags =
-      tmpl::list<domain::Tags::Mesh<dim>, domain::Tags::Element<dim>>;
+      tmpl::list<domain::Tags::Mesh<dim>, domain::Tags::Element<dim>,
+                 amr::Tags::NeighborInfo<dim>>;
 
   static void apply(
       const gsl::not_null<::dg::MortarMap<dim, evolution::dg::MortarData<dim>>*>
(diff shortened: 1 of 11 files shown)


src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
```diff
@@ -216,7 +221,20 @@ struct ProjectMortars : tt::ConformsTo<amr::protocols::Projector> {
       for (const auto& neighbor : neighbors) {
         const auto mortar_id = std::make_pair(direction, neighbor);
         mortar_data->emplace(mortar_id, MortarData<dim>{1});
-        mortar_mesh->emplace(mortar_id, Mesh<dim - 1>{});
+        const auto& new_neighbor_mesh = neighbor_info.at(neighbor).new_mesh;
```
nilsvu:
```
orientation.inverse()(neighbor_info.at(neighbor_id).new_mesh)
```

Then you can remove the mesh orientation and construction code below (note that you didn't orient the basis/quadrature below).

src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
```diff
@@ -216,7 +221,20 @@ struct ProjectMortars : tt::ConformsTo<amr::protocols::Projector> {
       for (const auto& neighbor : neighbors) {
         const auto mortar_id = std::make_pair(direction, neighbor);
         mortar_data->emplace(mortar_id, MortarData<dim>{1});
```
nilsvu, line 223: What's the 1 here actually?

src/Evolution/Executables/ScalarWave/EvolveScalarWave.hpp
```diff
@@ -197,6 +216,32 @@ struct EvolutionMetavars {
...
+        Initialization::ProjectTimeStepping<volume_dim>,
+        evolution::dg::Initialization::ProjectDomain<volume_dim>,
+        Initialization::ProjectTimeStepperHistory<EvolutionMetavars>,
+        ::amr::projectors::ProjectVariables<volume_dim,
+                                            typename system::variables_tag>,
+        ::amr::projectors::ProjectTensors<volume_dim,
+                                          ::ScalarWave::Tags::ConstraintGamma2>,
+        evolution::dg::Initialization::ProjectMortars<EvolutionMetavars>,
+        evolution::Actions::ProjectRunEventsAndDenseTriggers,
+        ::amr::projectors::DefaultInitialize<
```
nilsvu, line 230: Can you move some/all of these to the corresponding projectors? Would be easier if these are handled by the code that adds the tags, so you don't have to worry about them in the metavars.

wthrowe (dismissed):

nilsvu (changes requested):

src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
kidder: the number of buffers used by a MortarData...

src/Evolution/Executables/ScalarWave/EvolveScalarWave.hpp
kidder: I can, but will defer doing so until I implement the h-refinement overloads for the projectors as what is default initialized may change...

nilsvu: I'm not sure I understand, the way these are handled might change with adding h-refinement, but moving the responsibility to the projectors (which live next to the initializer that adds the tag) already now seems easier than doing so later.

kidder: It is psychological; replacing 5 lines of code with multiple commits adding dozens of lines now seems a waste of my time.  On the other hand, I can easily make the changes across multiple PRs when implementing h-refinement.

nilsvu: ok!

nilsvu: We have to add these to every executable though when we add support for AMR. If you prefer that fine with me, just checking.

nilsvu (approved):

src/Evolution/Executables/ScalarWave/EvolveScalarWave.hpp
nilsvu: Merging now, this can be done ina follow-up as well when adding AMR to more executables

PR #5575, opened 2023-10-19
Support domain deformations in elliptic solver

nilsvu:
## Proposed changes

This enables horizon-conforming (ellipsoidal) excision surfaces for spinning BH initial data. It will also allow for surface-fitting coordinates for BNS initial data. For reference, here's my thesis branch that had this implemented in a different way (using static deformed maps, not time dependence): https://github.com/sxs-collaboration/spectre/compare/develop...nilsvu:spectre:thesis

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CMakeLists.txt
+++ b/src/Domain/CMakeLists.txt
@@ -70,6 +70,7 @@ target_link_libraries(
   DomainBoundaryConditions
   DomainStructure
   ErrorHandling
+  FunctionsOfTime
   Options
   Spectral
   Utilities
@@ -77,8 +78,6 @@ target_link_libraries(
   LinearOperators
   RootFinding
   SphericalHarmonics
-  INTERFACE
-  FunctionsOfTime
 )
 
 add_subdirectory(Amr)
--- a/src/Domain/ElementMap.cpp
+++ b/src/Domain/ElementMap.cpp
@@ -3,8 +3,10 @@
 
 #include "Domain/ElementMap.hpp"
 
+#include "Domain/CoordinateMaps/Composition.hpp"
 #include "Domain/CoordinateMaps/CoordinateMap.hpp"  // IWYU pragma: keep
 #include "Domain/Structure/Side.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/OptimizerHacks.hpp"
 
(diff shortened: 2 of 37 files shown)


src/Domain/ElementMap.cpp
```diff
@@ -53,6 +55,52 @@ ElementMap<Dim, TargetFrame>::ElementMap(
...
+template <size_t Dim, typename TargetFrame>
+ElementMap<Dim, TargetFrame>::ElementMap(ElementId<Dim> element_id,
+                                         const Block<Dim>& block)
+    : ElementMap(
+          std::move(element_id),
+          [&block]() -> std::unique_ptr<domain::CoordinateMapBase<
+                         Frame::BlockLogical, TargetFrame, Dim>> {
+            if constexpr (std::is_same_v<TargetFrame, Frame::Inertial>) {
+              if (block.is_time_dependent()) {
+                if (block.has_distorted_frame()) {
```
nilsdeppe: Do you actually need this case? Shouldn't the else branch work even if there's a distorted frame? I'm fairly confident the evolution code at least requires grid->inertial to always be a valid map, independent of whether or not there's a distorted frame.

src/Domain/CMakeLists.txt
```diff
@@ -70,15 +70,14 @@ target_link_libraries(
   DomainBoundaryConditions
```
nilsdeppe, line 70: In the elliptic code do you actually evaluate the maps outside of compute items? If not, this commit, if not the previous as well could be replaced by adding a grid->inertial map into the elliptic box and using the time-dependent compute tags we use in the evolution code. Those tags don't do any extra work for "static" domains. I guess they currently require a `Tags::Time` to be in the box, but we could reasonably template on a bool or something to elide the time if it's not needed.

src/Elliptic/DiscontinuousGalerkin/Initialization.cpp
```diff
@@ -50,29 +50,26 @@ void InitializeGeometry<Dim>::operator()(
...
   *element = domain::Initialization::create_initial_element(element_id, block,
                                                             initial_refinement);
   // Element map
-  *element_map = ElementMap<Dim, Frame::Inertial>{
-      element_id, block.stationary_map().get_clone()};
+  *element_map = ElementMap<Dim, Frame::Inertial>{element_id, block};
   // Coordinates
   *logical_coords = logical_coordinates(*mesh);
-  *inertial_coords = element_map->operator()(*logical_coords);
+  *inertial_coords =
```
nilsdeppe, line 68: Right, if you don't need the map in other places, this seems like the perfect place to just call the grid->inertial map once

nilsdeppe (changes requested):
Looks reasonable, I do have one design question. Might be easier to discuss briefly on a call, up to you. Squash any changes directly, please.

src/Domain/ElementMap.cpp
nilsvu: You're right 👍 removed the unnecessary case.

src/Domain/CMakeLists.txt
nilsvu: Currently the map is only needed during initialization; in multiple actions though (domain init, subdomain init, faces & mortars). I thought about removing the ElementMap tag from the DataBox, but kept it for now so I don't have to recreate the ElementMap in these places. Could be done though I think. When we update the map parameters we have to re-evaluate the map as well though, which might be easier if the ElementMap is available. I haven't thought that part through yet.

nilsdeppe: Alright, that sounds good. Thanks for the explanation!

nilsdeppe (dismissed):

nilsvu:
Fixed a use-after-move clang-tidy complaint in the ElementMap constructor `ASSERT`. Nice to have clang tidy to catch these things.

nilsdeppe (approved):

PR #5576, opened 2023-10-19
Enable time dependent translation with transition for Sphere domain

guilara:
## Proposed changes

Add a translation map acting on the outer most shell of the spherical domain. If there is no radial partition, apply a uniform translation (since the inner boundary of the translation needs to be spherical and the shape map could also be active).

Note: The test `Unit.Domain.Creators.Sphere` failed because of roundoff error in assets 

`ASSERT(max(magnitude(result)) <= outer_radius_.value(),
             "Coordinates translated outside outer radius, This should not "
             "happen");`

 in https://github.com/sxs-collaboration/spectre/blob/06b239527bec8f6e2bfb6464ec881a98adf2af2b/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp#L376

When I printed the values during a test, the assert failed by a factor of order 10e-15 of the `outer_radius_` and seemed to remain of the same order in time. Changed the assert to allow for roundoff error.

Update: issue above fixed in #5601

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/Sphere.cpp
+++ b/src/Domain/Creators/Sphere.cpp
@@ -267,11 +267,23 @@ Sphere::Sphere(
       // Build the maps. We only apply the maps in the inner-most shell. The
       // inner radius is what's passed in, but the outer radius is the outer
       // radius of the inner-most shell so we have to see how many shells we
-      // have
+      // have.
+      // The translation map linear transition occurs only in the outer-most
+      // shell, such that the outer boundary is not translated. Require at least
+      // one radial partition, as the transition requires spherical shape.
+
+      if (radial_partitioning_.empty()) {
+        PARSE_ERROR(context,
+                    "The hard-coded translation map requires at least two "
+                    "shells. Use at least one radial partition.");
+      }
+
       std::get<sphere::TimeDependentMapOptions>(time_dependent_options_.value())
           .build_maps(std::array{0.0, 0.0, 0.0}, inner_radius_,
-                      radial_partitioning_.empty() ? outer_radius_
-                                                   : radial_partitioning_[0]);
+                      radial_partitioning_[0],
+                      // Translation inner radius
+                      std::pair<double, double>{radial_partitioning_.back(),
+                                                outer_radius_});
     }
   }
 }
@@ -374,9 +386,14 @@ Domain<3> Sphere::create_domain() const {
               time_dependent_options_.value());
 
       // First shell gets the distorted maps.
+      // Last shell gets translation with transition region
       for (size_t block_id = 0; block_id < num_blocks_; block_id++) {
         const bool include_distorted_map_in_first_shell =
             block_id < num_blocks_per_shell_;
+        // False if block_id is in the last shell
+        const bool use_rigid_translation =
+            block_id + num_blocks_per_shell_ + (fill_interior_ ? 1 : 0) <
+            num_blocks_;
         block_maps_grid_to_distorted[block_id] =
             hard_coded_options.grid_to_distorted_map(
                 include_distorted_map_in_first_shell);
@@ -385,7 +402,7 @@ Domain<3> Sphere::create_domain() const {
                 include_distorted_map_in_first_shell);
         block_maps_grid_to_inertial[block_id] =
             hard_coded_options.grid_to_inertial_map(
-                include_distorted_map_in_first_shell);
+                include_distorted_map_in_first_shell, use_rigid_translation);
       }
     } else {
       const auto& time_dependence = std::get<std::unique_ptr<
--- a/src/Domain/Creators/SphereTimeDependentMaps.cpp
+++ b/src/Domain/Creators/SphereTimeDependentMaps.cpp
@@ -26,11 +26,14 @@
 #include "Utilities/ErrorHandling/Error.hpp"
 
 namespace domain::creators::sphere {
+
 TimeDependentMapOptions::TimeDependentMapOptions(
-    const double initial_time, const ShapeMapOptions& shape_map_options)
+    const double initial_time, const ShapeMapOptions& shape_map_options,
+    const TranslationMapOptions& translation_map_options)
     : initial_time_(initial_time),
       initial_l_max_(shape_map_options.l_max),
-      initial_shape_values_(shape_map_options.initial_values) {}
+      initial_shape_values_(shape_map_options.initial_values),
+      initial_translation_values_(translation_map_options.initial_values) {}
 
 std::unordered_map<std::string,
                    std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
@@ -46,7 +49,8 @@ TimeDependentMapOptions::create_functions_of_time(
   // their initial expiration time to infinity (i.e. not expiring)
   std::unordered_map<std::string, double> expiration_times{
       {size_name, std::numeric_limits<double>::infinity()},
-      {shape_name, std::numeric_limits<double>::infinity()}};
+      {shape_name, std::numeric_limits<double>::infinity()},
+      {translation_name, std::numeric_limits<double>::infinity()}};
 
   // If we have control systems, overwrite these expiration times with the ones
   // supplied by the control system
@@ -101,12 +105,32 @@ TimeDependentMapOptions::create_functions_of_time(
                                  {0.0}}},
       expiration_times.at(size_name));
 
+  DataVector initial_translation_center_temp{3, 0.0};
+  DataVector initial_translation_velocity_temp{3, 0.0};
+  for (size_t i = 0; i < 3; i++) {
+    initial_translation_center_temp[i] =
+        gsl::at(initial_translation_values_.front(), i);
+    initial_translation_velocity_temp[i] =
+        gsl::at(initial_translation_values_.back(), i);
+  }
+
+  // TranslationMap FunctionOfTime
+  result[translation_name] =
+      std::make_unique<FunctionsOfTime::PiecewisePolynomial<2>>(
+          initial_time_,
+          std::array<DataVector, 3>{
+              {std::move(initial_translation_center_temp),
+               std::move(initial_translation_velocity_temp),
+               {3, 0.0}}},
+          expiration_times.at(translation_name));
+
   return result;
 }
 
-void TimeDependentMapOptions::build_maps(const std::array<double, 3>& center,
-                                         const double inner_radius,
-                                         const double outer_radius) {
+void TimeDependentMapOptions::build_maps(
+    const std::array<double, 3>& center, const double inner_radius,
+    const double outer_radius,
+    std::pair<double, double> translation_transition_radii) {
   std::unique_ptr<domain::CoordinateMaps::ShapeMapTransitionFunctions::
                       ShapeMapTransitionFunction>
       transition_func =
@@ -115,17 +139,22 @@ void TimeDependentMapOptions::build_maps(const std::array<double, 3>& center,
   shape_map_ = ShapeMap{center,         initial_l_max_,
                         initial_l_max_, std::move(transition_func),
                         shape_name,     size_name};
+
+  rigid_translation_map_ = TranslationMap{translation_name};
+
+  linear_falloff_translation_map_ =
+      TranslationMap{translation_name, translation_transition_radii.first,
+                     translation_transition_radii.second};
 }
 
 // If you edit any of the functions below, be sure to update the documentation
 // in the Sphere domain creator as well as this class' documentation.
 TimeDependentMapOptions::MapType<Frame::Distorted, Frame::Inertial>
 TimeDependentMapOptions::distorted_to_inertial_map(
-    const bool include_distorted_map) {
+    const bool include_distorted_map) const {
   if (include_distorted_map) {
-    return std::make_unique<
-        IdentityForComposition<Frame::Distorted, Frame::Inertial>>(
-        IdentityMap{});
+    return std::make_unique<DistortedToInertialComposition>(
+        rigid_translation_map_);
   } else {
     return nullptr;
   }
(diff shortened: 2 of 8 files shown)


src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -26,11 +26,23 @@
...
-    const double initial_time, const ShapeMapOptions& shape_map_options)
+    const double initial_time, const ShapeMapOptions& shape_map_options,
+    const std::array<double, 3>& translation_velocity,
+    const double translation_inner_radius,
+    const double translation_outer_radius)
     : initial_time_(initial_time),
       initial_l_max_(shape_map_options.l_max),
-      initial_shape_values_(shape_map_options.initial_values) {}
+      initial_shape_values_(shape_map_options.initial_values),
+      translation_velocity_(translation_velocity),
```
nilsvu: `initial_translation_velocity`

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -131,15 +139,47 @@ struct TimeDependentMapOptions {
     std::optional<std::variant<KerrSchildFromBoyerLindquist>> initial_values{};
   };
 
-  using options = tmpl::list<InitialTime, ShapeMapOptions>;
+  struct TranslationMapVelocity {
+    static std::string name() { return "Velocity"; }
```
nilsvu: `InitialVelocity`

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -131,15 +139,47 @@ struct TimeDependentMapOptions {
...
+    using type = std::array<double, 3>;
+    static constexpr Options::String help = {
+        "Velocity of the translation map."};
+    using group = TranslationMapOptions;
+  };
+
+  struct TranslationMapInnerRadius {
```
nilsvu: Don't take the inner and outer radii as options. Just apply the transition in the outer shell of the sphere, so inner radius is the last number in `radial_partitions` and outer radius is the outer radius of the sphere. Add this translation map with the transition _only_ to the blocks that make up the outer shell. Apply a rigid translation (no inner/outer radii) to all other blocks. If there is no partition (just one shell) and the interior of the sphere is filled with a cube then either error out or apply a rigid translation (because the transition only works with spherical inner and outer radii).

nilsvu (changes requested):

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -115,17 +133,21 @@ void TimeDependentMapOptions::build_maps(const std::array<double, 3>& center,
   shape_map_ = ShapeMap{center,         initial_l_max_,
                         initial_l_max_, std::move(transition_func),
                         shape_name,     size_name};
+  if (translation_inner_radius == translation_outer_radius) {
```
nilsvu: Take the translation radii as `std::optional<std::pair<double, double>>` then you can either pass `std::nullopt` or both radii.

You also have to build a second rigid translation map with no radii, which you can apply to the inner blocks. Then you don't need to remove asserts in the Translation map. I think it is necessary to have these two different maps because their Jacobian is discontinuous at the `translation_inner_radius`.

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -101,12 +105,26 @@ TimeDependentMapOptions::create_functions_of_time(
...
+  for (size_t i = 0; i < 3; i++) {
+    initial_velocity_temp[i] = gsl::at(initial_velocity_, i);
+  }
+
+  // TranslationMap FunctionOfTime
+  result[translation_name] =
+      std::make_unique<FunctionsOfTime::PiecewisePolynomial<2>>(
+          initial_time_,
+          std::array<DataVector, 3>{
+              {{3, 0.0}, initial_velocity_temp, {3, 0.0}}},
```
nilsvu: `std::move(initial_velocity_temp)`

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -115,17 +133,21 @@ void TimeDependentMapOptions::build_maps(const std::array<double, 3>& center,
...
 // in the Sphere domain creator as well as this class' documentation.
 TimeDependentMapOptions::MapType<Frame::Distorted, Frame::Inertial>
 TimeDependentMapOptions::distorted_to_inertial_map(
-    const bool include_distorted_map) {
+    const bool include_distorted_map) const {
   if (include_distorted_map) {
-    return std::make_unique<
-        IdentityForComposition<Frame::Distorted, Frame::Inertial>>(
-        IdentityMap{});
+    return std::make_unique<DistortedToInertialComposition>(translation_map_);
```
nilsvu: Here you add the rigid translation (not the transition)

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -145,10 +167,10 @@ TimeDependentMapOptions::MapType<Frame::Grid, Frame::Inertial>
 TimeDependentMapOptions::grid_to_inertial_map(
     const bool include_distorted_map) const {
   if (include_distorted_map) {
-    return std::make_unique<GridToInertialComposition>(shape_map_);
+    return std::make_unique<GridToInertialComposition>(shape_map_,
+                                                       translation_map_);
   } else {
-    return std::make_unique<
-        IdentityForComposition<Frame::Grid, Frame::Inertial>>(IdentityMap{});
+    return std::make_unique<GridToInertialSimple>(translation_map_);
```
nilsvu: Take another bool like `transition_translation_map` to determine if this block should use the transition or the rigid translation.

nilsvu (changes requested):

guilara:
@nilsvu  I think addressed all the comments

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -131,15 +146,24 @@ struct TimeDependentMapOptions {
     std::optional<std::variant<KerrSchildFromBoyerLindquist>> initial_values{};
   };
 
-  using options = tmpl::list<InitialTime, ShapeMapOptions>;
+  struct InitialVelocity {
+    static std::string name() { return "InitialVelocity"; }
```
knelli2: We probably also want to be able to specify the initial translation itself, not just the initial velocity, so we'll need an option for that. To keep things organized, make the translation options like the shape options. One struct that has

```cpp
struct TranslationMapOptions {
  using type = TranslationMapOptions;
  static std::string name() { return "TranslationMap"; }

  struct InitialValues {
    using type = std::array<std::array<double, 3>, 2>;
    ...
  };

  ...

  using options = tmpl::list<InitialValues>;
  std::array<std::array<double, 3>, 2> initial_values{};
};
```

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -195,16 +223,21 @@ struct TimeDependentMapOptions {
...
   inline static const std::string size_name{"Size"};
   inline static const std::string shape_name{"Shape"};
+  inline static const std::string translation_name{"Translation"};
 
  private:
   double initial_time_{std::numeric_limits<double>::signaling_NaN()};
   size_t initial_l_max_{};
   ShapeMap shape_map_{};
+  TranslationMap rigid_translation_map_{};
+  TranslationMap translation_map_{};
```
knelli2: Use `linear_falloff_translation_map_` instead of `translation_map_` so it's clear. Or just make it a `std::pair` of translation maps.

src/Domain/Creators/Sphere.cpp
```diff
@@ -267,11 +267,20 @@ Sphere::Sphere(
       // Build the maps. We only apply the maps in the inner-most shell. The
       // inner radius is what's passed in, but the outer radius is the outer
       // radius of the inner-most shell so we have to see how many shells we
-      // have
+      // have.
+      // The translation map linear transition occurs only in the outer-most
+      // shell, such that the outer boundary is not translated. If there
+      // is no radial partition, a uniform translation is applied instead, as
+      // the transition requires spherical shape.
```
knelli2: I don't think it makes sense to allow only a single spherical shell while also having a translation. I don't think we have boundary conditions that can handle that. We should probably just error if that happens and require we have two spherical shells (one radial partition).

src/Domain/Creators/Sphere.cpp
```diff
@@ -267,11 +267,20 @@ Sphere::Sphere(
...
       std::get<sphere::TimeDependentMapOptions>(time_dependent_options_.value())
           .build_maps(std::array{0.0, 0.0, 0.0}, inner_radius_,
                       radial_partitioning_.empty() ? outer_radius_
-                                                   : radial_partitioning_[0]);
+                                                   : radial_partitioning_[0],
+                      // Translation inner radius
+                      radial_partitioning_.empty()
+                          ? std::nullopt
+                          : std::optional<std::pair<double, double>>{
+                                {radial_partitioning_.back(), outer_radius_}});
```
knelli2: Then these don't have to be optional, but can always be passed.

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -115,17 +136,28 @@ void TimeDependentMapOptions::build_maps(const std::array<double, 3>& center,
   shape_map_ = ShapeMap{center,         initial_l_max_,
                         initial_l_max_, std::move(transition_func),
                         shape_name,     size_name};
+
+  rigid_translation_map_ = TranslationMap{translation_name};
+
+  if (translation_transition_radii.has_value()) {
```
knelli2: And then here you just always have a rigid translation map and a falloff translation map.

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -535,11 +537,14 @@ void test_parse_errors() {
           "an outflow-type boundary condition, you must use that."));
   using TDMO = domain::creators::sphere::TimeDependentMapOptions;
```
knelli2, line 584: The [sphere test failed](https://github.com/sxs-collaboration/spectre/actions/runs/6664974239/job/18113638093?pr=5576) for one of the builds still.

knelli2 (changes requested):

guilara:
For some reason I am still getting the asserts (in the description) triggered with the Xcts Kerr input file. In particular for the inner shell. Maybe there is still a bug somewhere here or in Translation.

As for the rest, I implemented the changes requested @knelli2

src/Domain/Creators/Sphere.cpp
```diff
@@ -374,9 +386,13 @@ Domain<3> Sphere::create_domain() const {
               time_dependent_options_.value());
 
       // First shell gets the distorted maps.
+      // Last shell gets translation with transition region
       for (size_t block_id = 0; block_id < num_blocks_; block_id++) {
         const bool include_distorted_map_in_first_shell =
             block_id < num_blocks_per_shell_;
+        // False if block_id is in the last shell
+        const bool use_rigid_translation =
+            block_id + num_blocks_per_shell_ < num_blocks_;
```
knelli2: I think this is incorrect if `fill_interior_` is true since the central cube is append at the end of the block list

knelli2 (changes requested):
The fixups look good. You can squash the existing changes plus these new ones. As for the error in the translation map, if the values are really off by 1e-14 then we just need to account for roundoff. Basically do this in two places in the translation map

```diff
diff --git a/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp b/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
index de3f73ed2d..debb09ef7a 100644
--- a/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
@@ -373,7 +373,8 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::piecewise_helper(
         get_element(gsl::at(result, i), k) +=
             gsl::at(func_or_deriv_of_time, i) * radial_falloff_factor;
       }
-      ASSERT(max(magnitude(result)) <= outer_radius_.value(),
+      const double eps = std::numeric_limits<double>::epsilon() * 100.0;
+      ASSERT(max(magnitude(result)) - eps <= outer_radius_.value(),
              "Coordinates translated outside outer radius, This should not "
              "happen");
     }
```

If that doesn't work, then we'll need to investigate more. While you're at it, please also print the numbers in that ASSERT message so we can see how off they are.

Also, please fix all clang-tidy errors when squashing.

guilara:
> The fixups look good. You can squash the existing changes plus these new ones. As for the error in the translation map, if the values are really off by 1e-14 then we just need to account for roundoff. Basically do this in two places in the translation map
> 
> ```diff
> diff --git a/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp b/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
> index de3f73ed2d..debb09ef7a 100644
> --- a/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
> +++ b/src/Domain/CoordinateMaps/TimeDependent/Translation.cpp
> @@ -373,7 +373,8 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> Translation<Dim>::piecewise_helper(
>          get_element(gsl::at(result, i), k) +=
>              gsl::at(func_or_deriv_of_time, i) * radial_falloff_factor;
>        }
> -      ASSERT(max(magnitude(result)) <= outer_radius_.value(),
> +      const double eps = std::numeric_limits<double>::epsilon() * 100.0;
> +      ASSERT(max(magnitude(result)) - eps <= outer_radius_.value(),
>               "Coordinates translated outside outer radius, This should not "
>               "happen");
>      }
> ```
> 
> If that doesn't work, then we'll need to investigate more. While you're at it, please also print the numbers in that ASSERT message so we can see how off they are.
> 
> Also, please fix all clang-tidy errors when squashing.

Running the Xcts input file with one partition I get:
1) For the assert checking blocks outside the inner radius I get:
`equal within roundoff (max(magnitude(result)), outer_radius_.value())) = 1 
1.0e15*(outer_radius_.value() - max(magnitude(result)))/outer_radius_.value() =  -0.177636`
or -0.177636e-15

whereas 
2) for the assert checking blocks that have at least one point as falling within the inner radius I get
` < inner_radius_value()... 
get_element(gsl::at(result, index), k) = -0.752477 ,
get_element(gsl::at(result, index), k) = -2.053507 ,
get_element(gsl::at(result, index), k) = 2.053507 ,
get_element(radius, k) = 3.000000 ,
equal within roundoff (max(magnitude(result)), outer_radius_.value())) = 1 ,
 1.0e15*(outer_radius_.value() - max(magnitude(result)))/outer_radius_.value() =  -0.177636`
 
 which is actually the same number -0.177636e-15

guilara:
@nilsvu Would it be ok if I set the spin to zero and remove the time dep maps in the Xcts KerrSchild test? In this PR I require at least two shells for hardcoded time dep maps and that test is timing out. 

Or how much can I increase the timeout time?

knelli2:
@nilsvu Can you take another look at this?

knelli2:
@guilara I believe we can close this because it is superseded by #5758?

knelli2:
Closing because this feature has already been implemented in #5758.

PR #5578, opened 2023-10-20
Optimize and vectorize TOMS748

nilsdeppe:
## Proposed changes

I've had this on a branch for a long time, decided to clean it up. This is necessary for vectorizing FixConservatives and primitive recovery. FixConservatives was ~10x after vectorization, primitive recovery I don't remember if I did as thorough of a comparison, but I remember at least a 5x speedup. Our GRMHD simulations (haven't checked GHMHD) spend ~50% of their time in primitive recovery, so hopefully once that's all in we should see a big reduction in runtime there.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


cmake/SetupXsimd.cmake
```diff
@@ -6,6 +6,10 @@ option(USE_XSIMD "Use xsimd if it is available" OFF)
 if(USE_XSIMD)
   find_package(xsimd REQUIRED)
 
+  if(xsimd_VERSION VERSION_LESS 11.0.1)
+    message(FATAL_ERROR "xsimd must be at least 11.0.1, got ${xsimd_VERSION}")
```
wthrowe, line 10: Update installation requirements documentation.

cmake/SetupXsimd.cmake
```diff
@@ -26,21 +26,23 @@ if(USE_XSIMD)
...
     xsimd
     )
 
-  target_link_libraries(
-    Blaze
-    INTERFACE
-    xsimd
-    )
+  # target_link_libraries(
+  #   Blaze
```
wthrowe, line 47: Just want to make sure leaving this was intentional.

src/DataStructures/Tensor/Tensor.hpp
```diff
@@ -108,7 +109,8 @@ class Tensor<X, Symm, IndexList<Indices...>> {
           std::is_same_v<X, ComplexModalVector> or
           std::is_same_v<X, DataVector> or std::is_same_v<X, ModalVector> or
           is_spin_weighted_of_v<ComplexDataVector, X> or
-          is_spin_weighted_of_v<ComplexModalVector, X>,
+          is_spin_weighted_of_v<ComplexModalVector, X> or
+          simd::is_batch<X>::value,
```
wthrowe, line 113: Update the error message.  (Maybe remove the list of types from it so the information isn't duplicated.)

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -6,15 +6,465 @@
 
 #pragma once
 
-#include <boost/math/tools/roots.hpp>
```
wthrowe, line 9: [arbitrary line] This is derived from boost code, so you need to include the boost license and copyright statement somewhere.

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -6,15 +6,465 @@
...
+    }
+    return num / denom;
+  } else {
+    // return num / denom without overflow, return r if overflow would occur.
+    const auto mask0 = fabs(denom) < static_cast<T>(1);
+    // Note: if denom >= 1.0 you get an FPE because of overflow from
+    // `max() * (1. + value)`
+    const auto mask = fabs(simd::select(mask0, denom, static_cast<T>(0)) *
+                           std::numeric_limits<T>::max()) <= fabs(num);
+    return simd::select(mask0 and mask, r, num / denom);
```
wthrowe: Does `num / denom` not FPE here?

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -6,15 +6,465 @@
...
+
+template <bool AssumeFinite, class F, class T, class Tol>
+std::pair<T, T> toms748_solve(F f, const T& ax, const T& bx, const T& fax,
+                              const T& fbx, Tol tol,
+                              const simd::mask_type_t<T>& ignore_filter,
+                              size_t& max_iter) {
+  // Main entry point and logic for Toms Algorithm 748
+  // root finder.
+  if (UNLIKELY(simd::any(ax >= bx))) {
+    throw std::domain_error("Lower bound is larger than upper bound");
```
wthrowe, line 305: Would allowing `ax == bx` cause any problems with the implementation or need special handling?  If it would it's probably not worth allowing, but if not it seems mathematically OK (as long as the single-point bracket is a root, of course).

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -6,15 +6,465 @@
...
+      break;
+    }
+
+    // Now we take a double-length secant step:
+    const auto fabs_fa_less_fabs_fb_mask =
+        (fabs(fa) < fabs(fb)) and incomplete_mask;
+    u = simd::select(fabs_fa_less_fabs_fb_mask, a, b);
+    fu = simd::select(fabs_fa_less_fabs_fb_mask, fa, fb);
+    const T b_minus_a = b - a;
+    // need to mask out fb==fa case
```
wthrowe: Is this a TODO?  I don't see any code handling this.

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -6,15 +6,465 @@
...
+    c = simd::fnma(static_cast<T>(2) * (fu / (fb - fa)), b_minus_a, u);
+    // c = u - static_cast<T>(2) * (fu / (fb - fa)) * b_minus_a;
```
wthrowe: Remove comment or move before the line it's documenting.

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -114,20 +581,47 @@ double toms748(const Function& f, const double lower_bound,
...
  * \requires Function `f` be callable with a `double` and a `size_t`
  *
+ * \note if `AssumeFinite` is true than the code assumes all numbers are
+ * finite and that `a > 0` is equivalent to `sign(a) > 0`. This reduces
+ * runtime but will cause bugs if the numbers aren't finite. It also assumes
+ * that products like `fa * fb` are also finite.
+ *
  * \throws `convergence_error` if, for any index, the requested tolerance is not
- * met after `max_iterations` iterations.
+p * met after `max_iterations` iterations.
```
wthrowe: Oops.

cmake/SetupXsimd.cmake
```diff
@@ -6,6 +6,10 @@ option(USE_XSIMD "Use xsimd if it is available" OFF)
 if(USE_XSIMD)
   find_package(xsimd REQUIRED)
 
+  if(xsimd_VERSION VERSION_LESS 11.0.1)
+    message(FATAL_ERROR "xsimd must be at least 11.0.1, got ${xsimd_VERSION}")
+  endif()
```
nilsvu, line 11: `find_package(xsimd REQUIRED 11.0.1)`

wthrowe (changes requested):
And as a bonus, this should get rid of most of the stupid clang-tidy warnings from Boost.

src/PointwiseFunctions/AnalyticSolutions/Xcts/CMakeLists.txt
```diff
@@ -42,6 +42,7 @@ target_link_libraries(
   Options
   RelativisticEulerSolutions
   Serialization
+  Simd
```
nilsvu: Make `PRIVATE` and add `PRIVATE RootFinding` please.

tests/Unit/NumericalAlgorithms/RootFinding/Test_TOMS748.cpp
```diff
@@ -167,8 +178,12 @@ SPECTRE_TEST_CASE("Unit.Numerical.RootFinding.TOMS748",
...
 
         const DataVector constant{2.0, 4.0, 2.0, 4.0};
-        const auto f_lambda = [&constant](const double x, const size_t i) {
-          return constant[i] - square(x);
+        const auto f_lambda = [&constant](const auto x, const size_t i) {
+          if constexpr (simd::is_batch<std::decay_t<decltype(x)>>::value) {
+            return simd::load_unaligned(&constant[i]) - square(x);
+          } else {
+            return constant[i] - square(x);
+          }
```
nilsvu, line 239: [random line] maybe add a benchmark somewhere? You optimized this code, so it would be nice to be able to benchmark it. https://github.com/catchorg/Catch2/blob/devel/docs/benchmarks.md

nilsvu (changes requested):

cmake/SetupXsimd.cmake
nilsdeppe: Yes, I'll add a comment. Basically, I haven't got all bugs out when doing Blaze+XSIMD in our case. I'm not sure if it's some padding assumption or what.

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
nilsdeppe: I have not had an issue, but I'll add a comment with code that could be changed to prevent that. This might have something to do with retiring instructions, etc. but SIMD things like that is something I don't have a complete grasp on. If we have issues, it'll be easy to guard against :) 

Edit: I just stuck it in... It's one mask, not super expensive to compute.

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
nilsdeppe: I'm assuming something subtle breaks in the implementation. The Boost implementation doesn't allow it so I just did that rather than possibly creating some weird edge case

cmake/SetupXsimd.cmake
nilsdeppe: I did not have this work correctly for xsimd. Feel free to give it a try later, but I spent enough time with it that I can't put more in

nilsdeppe:
Pushed fixups. Thanks for the reviews!

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -17,6 +17,15 @@
 
 namespace RootFinder {
 namespace toms748_detail {
+// Original implementation of TOMS748 is from Boost:
+//  (C) Copyright John Maddock 2006.
+//  Use, modification and distribution are subject to the
+//  Boost Software License, Version 1.0. (See accompanying file
+//  LICENSE_1_0.txt or copy at http://www.boost.org/LICENSE_1_0.txt)
```
wthrowe, line 29: You're missing the actual license file.

wthrowe (changes requested):
Other changes look good.

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
nilsdeppe: I added the file. Much to my surprise it looks like installed versions of Boost also don't necessarily include it, but I agree we should include the license for it.

nilsdeppe:
I added a couple more fixups that change the way the `size()` function works and also adds a `make_sequence` function

nilsdeppe:
@nilsvu could you please take a look at the fixups? @wthrowe I changed a couple small features in the simd lib in the last fixups (and a linking issue) if you have a chance to take a look. The linking issue was what was causing CI to fail.

wthrowe:
Looks good.

src/Evolution/Systems/GrMhd/ValenciaDivClean/LICENSE_1_0.txt
```diff
@@ -0,0 +1,23 @@
+Boost Software License - Version 1.0 - August 17th, 2003
```
nilsvu: This file should probably be in the TOMS748 directory rather than in GrMHD?

src/PointwiseFunctions/AnalyticSolutions/Xcts/Schwarzschild.cpp
```diff
@@ -104,10 +105,16 @@ double kerr_schild_areal_radius_from_isotropic(const double isotropic_radius,
 
 DataVector kerr_schild_areal_radius_from_isotropic(
     const DataVector& isotropic_radius, const double mass) {
-  return RootFinder::toms748(
-      [&isotropic_radius, &mass](const double areal_radius, const size_t i) {
-        return kerr_schild_isotropic_radius_from_areal(areal_radius, mass) -
-               isotropic_radius[i];
+  return RootFinder::toms748<true>(
```
nilsvu, line 108: Question: is this a test if the vectorization is working? Any reason you chose this particular function? Are you planning to go through and do some profiling and vectorize other toms748 calls in the code?

tests/Unit/NumericalAlgorithms/RootFinding/Test_TOMS748.cpp
```diff
@@ -142,21 +149,75 @@ void test_convergence_error_datavector() {
...
+    const auto f_at_lower_bound = f(lower_bound);
+    const auto f_at_upper_bound = f(upper_bound);
+    meter.measure([&f, &lower_bound, &upper_bound, &f_at_lower_bound,
+                   &f_at_upper_bound]() {
+      return RootFinder::toms748(f, lower_bound, upper_bound, f_at_lower_bound,
+                                 f_at_upper_bound, 1.0e-14, 1.0e-12);
+    });
+  };
+
+  for (size_t i = 0; i < 4; ++i) {
```
nilsvu, line 206: remove the loop and the `bound_type`s or add some comments?

nilsvu (changes requested):
you can squash

nilsvu:
Any idea why the root finder in the tabulated EOS test failed to converge in the macOS CI test?

src/Evolution/Systems/GrMhd/ValenciaDivClean/LICENSE_1_0.txt
nilsdeppe: I don't know. @wthrowe  what do you think?

src/PointwiseFunctions/AnalyticSolutions/Xcts/Schwarzschild.cpp
nilsdeppe: 1. More or less. I just wanted to make sure it works on code other than the stuff I've already done and this seemed like an easy candidate.
2. This seemed like an easy candidate.
3. Sort of. FixConservatives and primitive recovery yes, beyond that, no unless it ends up being something that slows me down.

nilsdeppe:
No, I have no idea why it doesn't work on macOS. I tried reproducing on my machine (Linux) with the same seeds, etc. and can't.

Edit: I'm pushing a fixup commit that should print out more info (tolerances and `result`) so we can get some idea of how well it did.

nilsdeppe:
@nilsvu @wthrowe this should be ready now...

src/Evolution/Systems/GrMhd/ValenciaDivClean/LICENSE_1_0.txt
wthrowe: I think so.  Why wouldn't RootFinding be correct?

wthrowe (changes requested):
Can you give a (very short) summary of what you changed since the last time we looked at this?  Was it just adjusting tolerances or similar?

src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -35,61 +498,84 @@ namespace RootFinder {
...
+                     T(absolute_tolerance));
   };
-  // clang-tidy: internal boost warning, can't fix it.
-  auto result = boost::math::tools::toms748_solve(  // NOLINT
+  auto result = toms748_detail::toms748_solve<AssumeFinite>(
       f, lower_bound, upper_bound, f_at_lower_bound, f_at_upper_bound, tol,
-      max_iters);
+      ignore_filter, max_iters);
   if (max_iters >= max_iterations) {
     throw convergence_error(
```
nilsdeppe, line 565: I only change this throw to have extra info in it.

nilsdeppe:
Okay, I moved the file. Sorry, this was a goof on my end. I don't know why I put it in the wrong directory. I then interpreted @nilsvu's comment as "Why not put it inside the header file?" which is clearly not what it asked. Sorry for the confusion!

wthrowe (dismissed):

tests/Unit/NumericalAlgorithms/RootFinding/Test_TOMS748.cpp
nilsvu: This is still here, were you going to remove it?

Also benchmarks are disabled by default when running `ctest` so you can remove the `bool enable` if you want.

nilsvu:
I ran the tabulated EOS test 1e4 times without fail on macOS (Apple Silicon though, not GitHub's Intel chips). Not sure what caused the convergence issue on CI but I can't reproduce it so can't do anything about it :)

nilsvu:
Edit: GitHub isn't surfacing this: https://github.com/sxs-collaboration/spectre/pull/5578#discussion_r1376531864

tests/Unit/NumericalAlgorithms/RootFinding/Test_TOMS748.cpp
nilsdeppe: I added a comment explaining why those are present (hard-coded vector widths)

nilsdeppe:
Rebased on latest develop since this was a bit old

wthrowe (dismissed):

nilsdeppe:
I've pushed a fixup that resolves the random failures in some of the tests like `Test_Tabulated3D.cpp`. I also pushed a commit that adds several `CAPTURE`s to `test_Tabulated3D.cpp`. I rebased as well.

nilsvu (approved):

wthrowe (approved):
Looks good.  Squash.

nilsdeppe:
Rebased and squashed. Thanks for the reviews!

nilsdeppe:
CI failures are:
- 2x executable build timeouts
- 1x Burgers Step failure

PR #5583, opened 2023-10-22
Allow timescale tuner to not decrease timescale

knelli2:
## Proposed changes

For size control, the `DecreaseFactor` of the `TimescaleTuner` has to be 1.0 because other logic takes care of decreasing the timescale. If it isn't 1.0, the run will crash almost instantly. Therefore, it doesn't really make sense to allow this as an option for size control. This PR templates `TimescaleTuner` on a bool to `AllowDecrease`. All other control systems have this bool as `true`, but size control has it as `false`.

### Upgrade instructions

If your executable has a Size control system, remove the `DecreaseThreshold:` and `DecreaseFactor:` options from the TimescaleTuner of the Size control systems. Note that you should *keep* these options for the `SmootherTuner` option of Size control.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Protocols/ControlError.hpp
+++ b/src/ControlSystem/Protocols/ControlError.hpp
@@ -12,9 +12,26 @@
 #include "Utilities/TMPL.hpp"
 #include "Utilities/TaggedTuple.hpp"
 
+template <bool AllowDecrease>
 struct TimescaleTuner;
 
 namespace control_system::protocols {
+namespace detail {
+
+struct DummyMetavariables {
+  using component_list = tmpl::list<>;
+};
+struct DummyTupleTags {
+  using type = int;
+};
+
+template <typename T, bool AllowDecrease>
+struct has_signature
+    : std::is_invocable_r<DataVector, T, const ::TimescaleTuner<AllowDecrease>&,
+                          const Parallel::GlobalCache<DummyMetavariables>&,
+                          const double, const std::string&,
+                          const tuples::TaggedTuple<DummyTupleTags>&> {};
+}  // namespace detail
 /// \brief Definition of a control error
 ///
 /// A control error is used within a control system to compute how far off the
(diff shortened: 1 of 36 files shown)


src/ControlSystem/Protocols/ControlError.hpp
```diff
@@ -12,9 +12,57 @@
...
+
+template <typename T, typename = std::void_t<>>
+struct has_signature_1 : std::false_type {};
+
+template <typename T, typename = std::void_t<>>
+struct has_signature_2 : std::false_type {};
+
+template <typename T>
+struct has_signature_1<
+    T, std::void_t<decltype(
```
wthrowe: [optional] This looks very similar to `std::is_invocable_r`.  Might be able to simplify.

src/ControlSystem/Tags/OptionTags.hpp
```diff
@@ -24,6 +25,11 @@ namespace control_system {
 /// their public member names and assigned to their corresponding DataBox tags.
 template <typename ControlSystem>
 struct OptionHolder {
+ private:
+  static constexpr bool is_size =
+      ::control_system::size::is_size_v<ControlSystem>;
```
wthrowe, line 30: [future work] It would be nice to not have everything in the control system infrastructure special-cased on size control.  At some point maybe we should have control systems report what features they want with something like `ControlSystem::allow_timescale_decrease`.

src/ControlSystem/TimescaleTuner.cpp
```diff
@@ -33,12 +34,12 @@ struct TimescaleCreator {
...
 
-TimescaleTuner::TimescaleTuner(
+template <bool AllowDecrease>
+TimescaleTuner<AllowDecrease>::TimescaleTuner(
     const typename InitialTimescales::type& initial_timescale,
     const double max_timescale, const double min_timescale,
-    const double decrease_timescale_threshold,
     const double increase_timescale_threshold, const double increase_factor,
-    const double decrease_factor)
+    const double decrease_timescale_threshold, const double decrease_factor)
```
wthrowe, line 42: Do some sort of check that these are reasonable in the no-decrease case.  Check `decrease_factor == 1.0`, at least.

tests/Unit/ControlSystem/Test_TimescaleTuner.cpp
```diff
@@ -564,9 +578,14 @@ void test_errors() {
 
 SPECTRE_TEST_CASE("Unit.ControlSystem.TimescaleTuner",
                   "[ControlSystem][Unit]") {
-  test_increase_or_decrease();
-  test_no_change_to_timescale();
-  test_create_from_options();
-  test_equality_and_serialization();
+  tmpl::for_each<tmpl::list<std::true_type, std::false_type>>(
+      [](const auto bool_v) {
+        constexpr bool AllowDecrease = tmpl::type_from<decltype(bool_v)>::value;
```
wthrowe: Snake-case.

wthrowe (changes requested):

src/ControlSystem/Protocols/ControlError.hpp
knelli2: That made things a lot cleaner, thanks!

src/ControlSystem/Tags/OptionTags.hpp
knelli2: I agree. I'll hopefully put this into the control system protocol eventually, but was a bit too much work at the moment.

src/ControlSystem/TimescaleTuner.cpp
knelli2: Only did the `decrease_factor == 1.0` check because I pass `limits::max()` for the decrease threshold and that's less intuitive of a default.

knelli2:
@wthrowe Squashed in the changes

wthrowe (dismissed):

wthrowe (approved):

PR #3384, opened 2021-07-24
Overlay options on restart from checkpoint

fmahebert:
## Proposed changes

Adds support for updating simulation parameters at restarts from checkpoint.

The interface is fairly rigid, but should be sufficient for a first pass:
- Looks for file `OriginalInputFile.overlay000123.yaml` when restarting from checkpoint `000123`.
- Allows certain GlobalCache tags only to be updated

Things I'm not sure about:
- The names of everything. The word "overlay" makes good sense at the file parser level, but not at the user-level when describing the feature "let's update parameters when restarting from checkpoint". I picked some names for the phase and functions but probably they can be improved on.
- The way Charm++ restarts work, the code doesn't know it's restarting from checkpoint vs just continuing as part of a longer run. In other words the phase sequence "WriteCheckpoint -> UpdateOptions -> Evolve" looks the same whether it's happening 10 hours into a run or 0 hours into a run. Right now I've added the reparsing phase into the wallclock-aware phase control, which lets us reparse only when restarting. But if someone writes `VisitAndReturn(WriteCheckpoint), VisitAndReturn(UpdateOptions)` phase control, the code might try to reparse input file every N slabs for example. I don't love it, but maybe this is just how the feature works...

The code isn't totally done: some tests still need to be written. I'm putting in this draft PR so we can start discussing the design, and because I will likely not have time to write the tests in the next few days anyway... In this draft the tests will fail because I've tweaked the Burgers input files to demonstrate how the feature works, but this is not currently in a "good state".

To try the feature yourself:
- compile `EvolveBurgersStep`
- `bin/EvolveBurgersStep ../tests/InputFiles/Burgers/Step.yaml`<- runs for ~1 minute, checkpoints and exits
- `bin/EvolveBurgersStep +restart SpectreCheckpoint000000` <- no overlay file, proceeds without changes
- `bin/EvolveBurgersStep +restart SpectreCheckpoint000001` <- overlay file changes obs frequency
- `bin/EvolveBurgersStep +restart SpectreCheckpoint000002` <- overlay file changes obs frequency again
- continue ad nauseum <- no more overlay files, proceeds without changes

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
[...]

PR #5585, opened 2023-10-24
Overlay options on restart

wthrowe:
Rebase of #3384.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/AddInputFileTests.cmake
+++ b/cmake/AddInputFileTests.cmake
@@ -122,6 +122,9 @@ function(add_input_file_tests INPUT_FILE_DIR INPUT_FILE_WHITELIST)
     if (${INPUT_FILE} IN_LIST INPUT_FILE_WHITELIST)
       continue()
     endif()
+    if (INPUT_FILE MATCHES "\\.overlay_[0-9]*\\.yaml$")
+      continue()
+    endif()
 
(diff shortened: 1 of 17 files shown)


knelli2:
Keep open until we have more time and infrastructure isn't changing so much.

wthrowe:
Rebased and variously cleaned up and updated.  I added some basic support for checkpointing to the input file tests and set one of the tests (ScalarWave/PlaneWave2D) to do an overlayed restart.  As a use case specifically mentioned by @nilsdeppe, I also marked `PhaseChangeAndTriggers` as overlayable so that the `CheckpointAndExitAfterWallclock` options can be changed on a restart, with the caveat that completely removing `CheckpointAndExitAfterWallclock` does not work (although setting the timeout to None does, as would, I believe but haven't tested, changing it to exist but never be triggered).

Also added a few Time tags as overlayable, just because I'm most familiar with them.  More things can easily be marked in small follow-up PRs.

nilsdeppe (approved):
Awesome! Thank you and @fmahebert (too many years ago!) for doing this!!!

fmahebert:
That was an unexpected Github notification 😂 

Thanks @wthrowe for finishing this up — I'm happy to see this was a useful development!

PR #5588, opened 2023-10-24
Add spherical shell coord maps

nilsvu:
## Proposed changes

First steps towards spherical shells. These maps will be used to map the logical axes [-1, 1] to ${\theta, \phi, r}$ and then to Cartesian coordinates.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/PolarAngle.cpp
+++ b/src/Domain/CoordinateMaps/PolarAngle.cpp
@@ -1,0 +1,77 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/CoordinateMaps/PolarAngle.hpp"
+
+#include <cmath>
+#include <cstddef>
+#include <pup.h>
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/DereferenceWrapper.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/MakeWithValue.hpp"
+
+namespace domain::CoordinateMaps {
+
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, 1> PolarAngle::operator()(
+    const std::array<T, 1>& xi) const {
+  return {{acos(-xi[0])}};
+}
+
+std::optional<std::array<double, 1>> PolarAngle::inverse(
+    const std::array<double, 1>& theta) const {
+  return {{-cos(theta[0])}};
+}
+
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, 1, Frame::NoFrame> PolarAngle::jacobian(
+    const std::array<T, 1>& xi) const {
+  using ReturnType = tt::remove_cvref_wrap_t<T>;
+  tnsr::Ij<ReturnType, 1, Frame::NoFrame> result{};
+  get<0, 0>(result) = 1. / sqrt(1. - square(xi[0]));
+  return result;
+}
+
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, 1, Frame::NoFrame>
+PolarAngle::inv_jacobian(const std::array<T, 1>& xi) const {
+  using ReturnType = tt::remove_cvref_wrap_t<T>;
+  tnsr::Ij<ReturnType, 1, Frame::NoFrame> result{};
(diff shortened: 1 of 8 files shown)


src/Domain/CoordinateMaps/PolarAngle.cpp
```diff
@@ -0,0 +1,77 @@
...
+    const std::array<double, 1>& theta) const {
+  return {{-cos(theta[0])}};
+}
+
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, 1, Frame::NoFrame> PolarAngle::jacobian(
+    const std::array<T, 1>& xi) const {
+  using ReturnType = tt::remove_cvref_wrap_t<T>;
+  tnsr::Ij<ReturnType, 1, Frame::NoFrame> result{};
+  get<0, 0>(result) = 1. / sqrt(1. - square(xi[0]));
```
teukolsky, line 34: Are we guaranteed never to call this at 1 or -1?

nilsvu: I added a warning to the docs. In practice we never call this on the poles because we use Gauss-Legendre points in cos(theta).

src/Domain/CoordinateMaps/PolarAngle.hpp
```diff
@@ -0,0 +1,68 @@
...
+ * \ingroup CoordinateMapsGroup
+ * \brief Map from the logical interval $\xi \in [-1, 1]$ to the polar angle
+ * $\theta \in [0, \pi]$, where $\xi = -\cos(\theta)$ as used by
+ * `ylm::Spherepack`.
+ *
+ * \warning The Jacobian of this map is ill-defined at the endpoints
+ * $\xi = -1, 1$, which correspond to the poles $\theta = 0, \pi$.
+ * Therefore it is typically used with Gauss-Legendre collocation points.
+ */
+class PolarAngle {
```
kidder, line 32: I don't understand why you are adding this map.   We want the logical coordinates to be those of the basis functions which are theta, phi, and \xi for the radial direction.

src/Domain/CoordinateMaps/SphericalToCartesian.hpp
```diff
@@ -0,0 +1,90 @@
...
+ * prepend the following maps:
+ * - `r`: Any 1D radial map that maps [-1, 1] to [inner_radius, outer_radius],
+ *   such as `domain::CoordinateMaps::Interval`.
+ * - `theta`: `domain::CoordinateMaps::PolarAngle` to map [-1, 1] to [0, pi]
+ *   with the relation $\xi = cos(\theta)$.
+ * - `phi`: Affine map [-1, 1] to [0, 2pi), e.g.
+ *   `domain::CoordinateMaps::Affine` or `domain::CoordinateMaps::Interval`
+ */
+template <size_t Dim>
+class SphericalToCartesian {
```
kidder, line 39: This map will not work with SPHEREPACK.  It needs the spherical Pfaffian map from SpEC.

kidder (changes requested):
We need to discuss a plan for adding a spherical basis.  This is not how I would do it.

src/Domain/CoordinateMaps/PolarAngle.cpp
markscheel: You should never divide by powers of `sin(theta)` .  Even though you don't actually have a point at the poles, you will get convergence problems because the points approach the poles as you increase resolution.  (This was a problem with the original version of SpEC's CCE for example.)

If you use Pfaffian derivatives you will find that you never need to divide by powers of `sin(theta)` because all the powers of `sin(theta)` will cancel out when you compute anything that is regular at the poles (although sometimes it might require a bit of algebra to eliminate them).

nilsvu:
It occurred to me that we probably have a misunderstanding: I don't intend for these maps to be used in the numerical scheme. We'll use YlmSpherepack etc for that. These maps just have to be consistent with YlmSpherepack in that the coordinates are identical. Then things like visualization of grid points, evaluating pointwise functions on the grid, interpolation to points, etc, work without changing other infrastructure to deal with spherical shells. At least that's my intention. I put this into the ExportCoordinates executable and that worked well without taking derivatives or doing DG things. Happy to discuss of course - we have lots of experience with Ylms from SpEC available.

knelli2:
Superseded by conversations at SXS-Con about adding a spherical harmonic basis (true spherical shells)

PR #5589, opened 2023-10-24
Add transform for History

wthrowe:
IMEX preparation.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/History.hpp
+++ b/src/Time/History.hpp
@@ -879,6 +879,22 @@ void History<Vars>::pup(PUP::er& p) {
   // Don't serialize the allocation cache.
 }
 
+template <typename Vars>
+std::ostream& History<Vars>::print(std::ostream& os) const {
+  using ::operator<<;
+  os << "Integration order: " << integration_order_ << "\n";
+  os << "Step values:\n";
+  for (auto& record : *this) {
+    record.print(os);
+  }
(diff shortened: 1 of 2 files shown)


src/Time/History.hpp
```diff
@@ -1008,4 +1008,92 @@ template <typename Vars>
...
+                record.time_step_id, History<DestVars>::no_value,
+                [&](const auto result) {
+                  derivative_transformer(result, record.derivative);
+                });
+          }
+        } else {
+          static_assert(
+              std::is_invocable_v<ValueTransformer, const SourceVars&>,
+              "Transform function must either be callable to mutate entries "
+              "or to generate them.");
```
nilsdeppe: I found "to generate them" didn't make it obvious that the function should return the transformed state by value. Maybe something like "or return the transformed state by value."?

nilsdeppe (changes requested):
LGTM, one suggestion. Let me know if you agree or not and please squash immediately if you decide to do some sort of reword

src/Time/History.hpp
wthrowe: Done as you suggested.

nilsdeppe (approved):

kidder:
ignoring known CI issues

PR #5593, opened 2023-10-25
Limit time step to prevent control-system deadlocks

wthrowe:
### Upgrade instructions

Executables with control systems and global time-stepping will now work with any time stepper, not just AdamsBashforth.  When using a non-AdamsBashforth time stepper the step size will gradually decrease, so you should include the ChangeSlabSize event to increase it again.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Actions/LimitTimeStep.hpp
+++ b/src/ControlSystem/Actions/LimitTimeStep.hpp
@@ -1,0 +1,248 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <algorithm>
+#include <memory>
+#include <optional>
+
+#include "ControlSystem/CombinedName.hpp"
+#include "ControlSystem/Metafunctions.hpp"
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/ArrayComponentId.hpp"
+#include "Parallel/Callback.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Time/ChangeSlabSize.hpp"
+#include "Time/Tags/HistoryEvolvedVariables.hpp"
+#include "Time/Tags/TimeStepper.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+namespace Tags {
+struct TimeStep;
+struct TimeStepId;
+}  // namespace Tags
+namespace control_system::Tags {
+template <typename ControlSystems>
+struct FutureMeasurements;
+struct MeasurementTimescales;
+}  // namespace control_system::Tags
+namespace domain::Tags {
+struct FunctionsOfTime;
+}  // namespace domain::Tags
+namespace tuples {
+template <typename... Tags>
+class TaggedTuple;
+}  // namespace tuples
+/// \endcond
+
+namespace control_system::Actions {
+/// \ingroup ControlSystemsGroup
+/// \brief Limit the step size in a GTS evolution to prevent deadlocks from
+/// control system measurements.
+///
+/// \details Most time steppers require evaluations of the coordinates
+/// at several times during the step before they can produce dense
+/// output.  If any of those evaluations require a function-of-time
+/// update depending on a measurement within the step, the evolution
+/// will deadlock.  This action reduces the step size if necessary to
+/// prevent that from happening.
+///
+/// Specifically:
+/// 1. The chosen step will never be longer than the unmodified step,
+///    and will be short enough to avoid relevant function-of-time
+///    expirations.
+/// 2. Given the previous, the step will cover as many control-system
+///    updates as possible.
+/// 3. If the next step is likely to be limited by this action, adjust
+///    the length of the current step so that this step and the next
+///    step will be as close as possible to the same size.
+template <typename ControlSystems>
+struct LimitTimeStep {
+ private:
+  using control_system_groups =
+      tmpl::transform<metafunctions::measurements_t<ControlSystems>,
+                      metafunctions::control_systems_with_measurement<
+                          tmpl::pin<ControlSystems>, tmpl::_1>>;
+
+  template <typename Group>
+  struct GroupExpiration {
+    using type = double;
+  };
+
+ public:
+  template <typename DbTagsList, typename... InboxTags, typename Metavariables,
+            typename ArrayIndex, typename ActionList,
+            typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTagsList>& box,
+      const tuples::TaggedTuple<InboxTags...>& /*inboxes*/,
+      Parallel::GlobalCache<Metavariables>& cache,
+      const ArrayIndex& array_index, ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    static_assert(not Metavariables::local_time_stepping,
+                  "The control system LimitTimeStep action is only for global "
+                  "time stepping.");
+    const auto& time_step_id = db::get<::Tags::TimeStepId>(box);
+    if (time_step_id.substep() != 0) {
+      return {Parallel::AlgorithmExecution::Continue, std::nullopt};
+    }
+
+    const auto& time_stepper = db::get<::Tags::TimeStepper<>>(box);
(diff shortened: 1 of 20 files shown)


src/ControlSystem/Actions/LimitTimeStep.hpp
```diff
@@ -0,0 +1,245 @@
...
+  template <typename Group>
+  struct GroupExpiration {
+    using type = double;
+  };
+
+ public:
+  template <typename DbTagsList, typename... InboxTags, typename Metavariables,
+            typename ArrayIndex, typename ActionList,
+            typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
```
knelli2, line 82: Add
```cpp
static_assert(Metavariables::local_time_stepping, "The control system LimitTimeStep
    Action can only be used with global time stepping");
```
inside the action?

src/ControlSystem/Actions/LimitTimeStep.hpp
```diff
@@ -0,0 +1,245 @@
...
+
+    bool ready = true;
+    // Calculate group_expiration_times
+    tmpl::for_each<control_system_groups>([&](auto group_v) {
+      if (not ready) {
+        return;
+      }
+      using group = tmpl::type_from<decltype(group_v)>;
+
+      auto& measurement_times =
```
knelli2: Rename to `future_measurements`? Having both `measurement_times` and `measurement_timescales` was a bit confusing.

src/ControlSystem/Actions/LimitTimeStep.hpp
```diff
@@ -0,0 +1,245 @@
...
+    // Step time that produces a balanced step with the following
+    // step, ignoring the restrictions on the current step.
+    double preferred_step_time = orig_step_end;
+
+    tmpl::for_each<control_system_groups>([&](auto group_v) {
+      using group = tmpl::type_from<decltype(group_v)>;
+
+      // This was used above, so it is not nullopt.
+      const double group_update =
+          *db::get<Tags::FutureMeasurements<group>>(box).next_update();
```
knelli2: [optional] `.value()` instead?

tests/Unit/ControlSystem/Actions/Test_LimitTimeStep.cpp
```diff
@@ -0,0 +1,327 @@
...
+          const std::vector<std::pair<double, double>>& fot_updatesC,
+          std::unique_ptr<TimeStepper> stepper =
+              std::make_unique<TimeSteppers::Rk3HesthavenSsp>(),
+          TimeSteppers::History<Var::type> history = {}) {
+  INFO(test_label);
+  ASSERT(measurement_updatesA.size() > 1, "Bad argument");
+  ASSERT(measurement_updatesBC.size() > 1, "Bad argument");
+  ASSERT(fot_updatesA.size() > 1, "Bad argument");
+  ASSERT(fot_updatesB.size() > 1, "Bad argument");
+  ASSERT(fot_updatesC.size() > 1, "Bad argument");
```
knelli2, line 119: Do you really need these? Usually for a test we just assume function arguments are correct because we are the ones creating a test. If you decide we do want these because it makes debugging easier, maybe use `REQUIRE` since we want to catch this in Release mode as well? The errors are probably self explanatory and "Bad argument" isn't a great error message.

tests/Unit/ControlSystem/Actions/Test_LimitTimeStep.cpp
```diff
@@ -0,0 +1,327 @@
...
+       {{0.0, arbitrary}, {infinity, nan}});
+  test("Step much shorter than limit", 1.0, 5.0, 5.0,
+       {{0.0, 10.0}, {50.0, nan}},
+       {{0.0, arbitrary}, {50.0, nan}},  // goal range [20, 50]
+       {{0.0, infinity}, {infinity, nan}},
+       {{0.0, arbitrary}, {infinity, nan}},
+       {{0.0, arbitrary}, {infinity, nan}});
+  test("Limited by expiration", 1.0, 5.0, 4.0,
+       {{0.0, 1.0}, {5.0, nan}},
+       {{0.0, arbitrary}, {4.0, nan}},  // goal range [2, 4]
```
knelli2, line 225: Question: Does it matter that the measurement timescales expire before the FoT expiration time? Or that the number of measurements technically don't line up with the measurement timescale and measurement expiration? Or are you just forcing it to test the FoT expiration time code? (Same question for several others below)

tests/Unit/ControlSystem/Actions/Test_LimitTimeStep.cpp
```diff
@@ -0,0 +1,327 @@
...
+  // clang-format off
+  test("No control systems", 1.0, 5.0, 5.0,
+       {{0.0, infinity}, {infinity, nan}},
+       {{0.0, arbitrary}, {infinity, nan}},
+       {{0.0, infinity}, {infinity, nan}},
+       {{0.0, arbitrary}, {infinity, nan}},
+       {{0.0, arbitrary}, {infinity, nan}});
+  test("Step much shorter than limit", 1.0, 5.0, 5.0,
+       {{0.0, 10.0}, {50.0, nan}},
+       {{0.0, arbitrary}, {50.0, nan}},  // goal range [20, 50]
```
knelli2, line 219: What is this goal range? The 50 seems somewhat self explanatory, but the 20 doesn't. Could you add a comment before this about what the goal range is supposed to represent?

knelli2 (changes requested):
Have you tested this will all the executables you've changed here that have control systems? (Namely BH, BBH, BNS, STBH). While the code looks ok, I'm worried some strange deadlock could occur if this isn't fully tested for each with large numbers of elements.

tests/Unit/ControlSystem/Actions/Test_LimitTimeStep.cpp
wthrowe: As the one writing the test, I got the arguments wrong (requiring two entries is unusual), so I put in an assertion.  We always use ASSERT to check function arguments.  I agree the message isn't useful, but the message field is required and I have nothing to add to the text of the test.

tests/Unit/ControlSystem/Actions/Test_LimitTimeStep.cpp
wthrowe: This code is not supposed to care when the measurement timescales expire (except that it needs enough to evaluate them).  Making it late means I didn't have to tune it to be between the update and next measurement each time, particularly in the later cases covering multiple FoT updates.

The measurements and expirations won't line up exactly in the real case because of floating-point error.  Testing this way makes any problem from that obvious.

wthrowe (commented):
Fixup pushed.

> Have you tested this will all the executables you've changed here that have control systems? (Namely BH, BBH, BNS, STBH). While the code looks ok, I'm worried some strange deadlock could occur if this isn't fully tested for each with large numbers of elements.

I have tested with `EvolveGhSingleBlackHole`.  I don't know how to run BBH or BNS.

As far as I am aware, this problem is independent of the domain decomposition.

knelli2:
It's not domain composition that directly affects this. If there was a roundoff bug regarding expiration times (not saying that there is), then it's possible we wouldn't catch it with a smaller number of elements/nodes. This is because there'd be less charm communication latency and so FoTs would be updated before any elements reached that time. But with more nodes/elements this problem gets worse (as I've seen for BBH).

I think it would be a good idea to have @geoffrey4444 and @ffoucart do some standard runs for their respective problems with GTS RK (or DP) methods. Maybe also include the event that periodically resets the timestep size based on the cfl.

(You can squash the fixup in)

nilsdeppe (approved):
BNS mergers seem to still run. I have an AB run going still, but also tried an RK3. The RK3 run was about 10x _slower_ than AB3. It's here `/home/nils/ds09/spectre/BnsRuns/GtsControlSystem` My guess is that for some reason the control system is way too tight for rotation, but I don't know how to check that or how to adjust that. AB run is here `/home/nils/ds09/spectre/BnsRuns/GtsControlSystemAb`. both are on Wheeler

@knelli2 @wthrowe any insights into the huge slowdown of RK methods would be appreciated. Naively I would've expected these runs not to be limited by the control system

knelli2 (approved):
Ok so after discussing with @nilsdeppe, he just had to increase the initial damping timescale and the run was back to it's normal speed. So it appears the BNS is working as expected. I also know @geoffrey4444 has a student testing the BBH with GTS, but since the BBH executable is still LTS on develop, this PR won't affect it. Therefore, I'm satisfied this PR works and it can be merged.

Ignoring known test failures.

PR #5607, opened 2023-10-30
Register derived parallel callbacks

knelli2:
## Proposed changes

The GlobalCache never really gets pupped, but it is sized by the MemoryMonitor. Therefore, all it's members need to be registered with charm. Before the derived callbacks weren't registered, so we got the `Unrecognized PUP::able::PUP_ID` error. This PR will register the callbacks only once so that doesn't happen.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Framework/ActionTesting.hpp
+++ b/tests/Unit/Framework/ActionTesting.hpp
@@ -6,6 +6,7 @@
 #include <algorithm>
 #include <charm++.h>
 #include <cstddef>
+#include <memory>
 #include <optional>
 #include <tuple>
 #include <unordered_map>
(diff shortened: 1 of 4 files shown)


tests/Unit/Framework/ActionTesting.hpp
```diff
@@ -6,6 +6,7 @@
 #include <algorithm>
```
wthrowe, line 6: Wrong class name in the commit message.  (Should be MockDistributedObjectProxy.)

wthrowe (changes requested):

wthrowe (approved):

PR #5610, opened 2023-11-01
Add boost to KerrSchild

guilara:
## Proposed changes

Add a boost to the KerrSchild solution. This is done by boosting the coordinates, null form and derivatives of the H function defining the KerrSchild solution. (For the case when there is a boost the solution depends on time as well, but we don't code up the time dependence of the coordinates for simplicity.)

Evolving this solution will require time dependent translation maps to keep the excision surface inside the apparent horizon (#5576)

This reuses items from https://github.com/sxs-collaboration/spectre/compare/develop...nilsvu:spectre:kerr_boost

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.cpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.cpp
@@ -20,8 +20,11 @@
 #include "PointwiseFunctions/GeneralRelativity/ExtrinsicCurvature.hpp"
 #include "PointwiseFunctions/GeneralRelativity/InverseSpacetimeMetric.hpp"
 #include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/TimeDerivativeOfSpatialMetric.hpp"
+#include "PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp"
 #include "Utilities/ConstantExpressions.hpp"
 #include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 #include "Utilities/MakeWithValue.hpp"
@@ -34,13 +37,14 @@ KerrSchild::KerrSchild(CkMigrateMessage* /*msg*/) {}
 KerrSchild::KerrSchild(const double mass,
                        const std::array<double, 3>& dimensionless_spin,
                        const std::array<double, 3>& center,
+                       const std::array<double, 3>& boost_velocity,
                        const Options::Context& context)
     : mass_(mass),
-      // clang-tidy: do not std::move trivial types.
-      dimensionless_spin_(dimensionless_spin),  // NOLINT
-      // clang-tidy: do not std::move trivial types.
+      dimensionless_spin_(dimensionless_spin),
       center_(center),
-      zero_spin_(dimensionless_spin_ == std::array<double, 3>{{0., 0., 0.}}) {
+      boost_velocity_(boost_velocity),
+      zero_spin_(dimensionless_spin_ == std::array<double, 3>{{0., 0., 0.}}),
+      zero_velocity_(boost_velocity_ == std::array<double, 3>{{0., 0., 0.}}) {
   const double spin_magnitude = magnitude(dimensionless_spin_);
   if (spin_magnitude > 1.0) {
     PARSE_ERROR(context, "Spin magnitude must be < 1. Given spin: "
@@ -56,7 +60,9 @@ void KerrSchild::pup(PUP::er& p) {
   p | mass_;
   p | dimensionless_spin_;
   p | center_;
+  p | boost_velocity_;
   p | zero_spin_;
+  p | zero_velocity_;
 }
 
 template <typename DataType, typename Frame>
@@ -68,11 +74,23 @@ template <typename DataType, typename Frame>
 void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
     const gsl::not_null<tnsr::I<DataType, 3, Frame>*> x_minus_center,
     const gsl::not_null<CachedBuffer*> /*cache*/,
-    internal_tags::x_minus_center<DataType, Frame> /*meta*/) const {
+    internal_tags::x_minus_center_unboosted<DataType, Frame> /*meta*/) const {
   *x_minus_center = x_;
   for (size_t d = 0; d < 3; ++d) {
     x_minus_center->get(d) -= gsl::at(solution_.center(), d);
   }
+}
+
+template <typename DataType, typename Frame>
+void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
+    const gsl::not_null<tnsr::I<DataType, 3, Frame>*> x_minus_center_boosted,
+    const gsl::not_null<CachedBuffer*> cache,
+    internal_tags::x_minus_center<DataType, Frame> /*meta*/) const {
+  const auto& x_minus_center = cache->get_var(
+      *this, internal_tags::x_minus_center_unboosted<DataType, Frame>{});
+  // Boost the coordinates to the rest frame
+  sr::lorentz_boost(x_minus_center_boosted, x_minus_center, 0.,
+                    solution_.boost_velocity());
 }
 
 template <typename DataType, typename Frame>
@@ -251,9 +269,9 @@ void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
 
 template <typename DataType, typename Frame>
 void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
-    const gsl::not_null<tnsr::i<DataType, 3, Frame>*> deriv_H,
+    const gsl::not_null<tnsr::a<DataType, 3, Frame>*> deriv_H,
     const gsl::not_null<CachedBuffer*> cache,
-    internal_tags::deriv_H<DataType, Frame> /*meta*/) const {
+    internal_tags::deriv_H_unboosted<DataType, Frame> /*meta*/) const {
   const auto& deriv_log_r =
       cache->get_var(*this, internal_tags::deriv_log_r<DataType, Frame>{});
   const auto& deriv_H_temp1 =
@@ -262,10 +280,21 @@ void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
       get(cache->get_var(*this, internal_tags::deriv_H_temp2<DataType>{}));
 
   const auto spin_a = solution_.dimensionless_spin() * solution_.mass();
+  get<0>(*deriv_H) = 0.;
   for (size_t i = 0; i < 3; ++i) {
-    deriv_H->get(i) =
+    deriv_H->get(i + 1) =
         deriv_H_temp1 * deriv_log_r.get(i) - deriv_H_temp2 * gsl::at(spin_a, i);
   }
+}
+
+template <typename DataType, typename Frame>
+void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
+    const gsl::not_null<tnsr::a<DataType, 3, Frame>*> deriv_H_boosted,
+    const gsl::not_null<CachedBuffer*> cache,
+    internal_tags::deriv_H<DataType, Frame> /*meta*/) const {
+  const auto& deriv_H = cache->get_var(
+      *this, internal_tags::deriv_H_unboosted<DataType, Frame>{});
+  sr::lorentz_boost(deriv_H_boosted, deriv_H, -solution_.boost_velocity());
 }
 
 template <typename DataType, typename Frame>
@@ -297,9 +326,9 @@ void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
 
 template <typename DataType, typename Frame>
 void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
-    const gsl::not_null<tnsr::i<DataType, 3, Frame>*> null_form,
+    const gsl::not_null<tnsr::a<DataType, 3, Frame>*> null_form,
     const gsl::not_null<CachedBuffer*> cache,
-    internal_tags::null_form<DataType, Frame> /*meta*/) const {
+    internal_tags::null_form_unboosted<DataType, Frame> /*meta*/) const {
   const auto& a_dot_x_over_r =
       get(cache->get_var(*this, internal_tags::a_dot_x_over_r<DataType>{}));
   const auto& r = get(cache->get_var(*this, internal_tags::r<DataType>{}));
@@ -309,53 +338,69 @@ void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
       get(cache->get_var(*this, internal_tags::denom<DataType>{}));
 
   const auto spin_a = solution_.dimensionless_spin() * solution_.mass();
+  get<0>(*null_form) = 1.;
   for (size_t i = 0; i < 3; ++i) {
     const size_t cross_product_index_1 = (i + 1) % 3;
     const size_t cross_product_index_2 = (i + 2) % 3;
-    null_form->get(i) = denom * (r * x_minus_center.get(i) -
-                                 gsl::at(spin_a, cross_product_index_1) *
-                                     x_minus_center.get(cross_product_index_2) +
-                                 gsl::at(spin_a, cross_product_index_2) *
-                                     x_minus_center.get(cross_product_index_1) +
-                                 a_dot_x_over_r * gsl::at(spin_a, i));
+    null_form->get(i + 1) =
+        denom * (r * x_minus_center.get(i) -
+                 gsl::at(spin_a, cross_product_index_1) *
+                     x_minus_center.get(cross_product_index_2) +
+                 gsl::at(spin_a, cross_product_index_2) *
+                     x_minus_center.get(cross_product_index_1) +
+                 a_dot_x_over_r * gsl::at(spin_a, i));
   }
 }
 
 template <typename DataType, typename Frame>
 void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
-    const gsl::not_null<tnsr::ij<DataType, 3, Frame>*> deriv_null_form,
+    const gsl::not_null<tnsr::a<DataType, 3, Frame>*> null_form_boosted,
     const gsl::not_null<CachedBuffer*> cache,
-    internal_tags::deriv_null_form<DataType, Frame> /*meta*/) const {
+    internal_tags::null_form<DataType, Frame> /*meta*/) const {
+  const auto& null_form = cache->get_var(
+      *this, internal_tags::null_form_unboosted<DataType, Frame>{});
+  sr::lorentz_boost(null_form_boosted, null_form, -solution_.boost_velocity());
+}
+
+template <typename DataType, typename Frame>
+void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
+    const gsl::not_null<tnsr::ab<DataType, 3, Frame>*> deriv_null_form,
+    const gsl::not_null<CachedBuffer*> cache,
+    internal_tags::deriv_null_form_unboosted<DataType, Frame> /*meta*/) const {
   const auto spin_a = solution_.dimensionless_spin() * solution_.mass();
   const auto& denom =
       get(cache->get_var(*this, internal_tags::denom<DataType>{}));
   const auto& r = get(cache->get_var(*this, internal_tags::r<DataType>{}));
   const auto& x_minus_center =
       cache->get_var(*this, internal_tags::x_minus_center<DataType, Frame>{});
-  const auto& null_form =
-      cache->get_var(*this, internal_tags::null_form<DataType, Frame>{});
+  const auto& null_form = cache->get_var(
+      *this, internal_tags::null_form_unboosted<DataType, Frame>{});
   const auto& a_dot_x_over_rsquared = get(
       cache->get_var(*this, internal_tags::a_dot_x_over_rsquared<DataType>{}));
   const auto& deriv_log_r =
       cache->get_var(*this, internal_tags::deriv_log_r<DataType, Frame>{});
 
+  for (size_t d = 0; d < 4; ++d) {
+    deriv_null_form->get(d, 0) = 0.;
+    deriv_null_form->get(0, d) = 0.;
+  }
   for (size_t i = 0; i < 3; i++) {
     for (size_t j = 0; j < 3; j++) {
-      deriv_null_form->get(j, i) =
+      deriv_null_form->get(j + 1, i + 1) =
           denom * (gsl::at(spin_a, i) * gsl::at(spin_a, j) / r +
-                   (x_minus_center.get(i) - 2.0 * r * null_form.get(i) -
+                   (x_minus_center.get(i) - 2.0 * r * null_form.get(i + 1) -
                     a_dot_x_over_rsquared * gsl::at(spin_a, i)) *
                        deriv_log_r.get(j) * r);
       if (i == j) {
-        deriv_null_form->get(j, i) += denom * r;
+        deriv_null_form->get(j + 1, i + 1) += denom * r;
       } else {  //  add denom*epsilon^ijk a_k
         size_t k = (j + 1) % 3;
         if (k == i) {  // j+1 = i (cyclic), so choose minus sign
           k++;
           k = k % 3;  // and set k to be neither i nor j
-          deriv_null_form->get(j, i) -= denom * gsl::at(spin_a, k);
+          deriv_null_form->get(j + 1, i + 1) -= denom * gsl::at(spin_a, k);
         } else {  // i+1 = j (cyclic), so choose plus sign
-          deriv_null_form->get(j, i) += denom * gsl::at(spin_a, k);
+          deriv_null_form->get(j + 1, i + 1) += denom * gsl::at(spin_a, k);
         }
       }
     }
@@ -364,16 +409,31 @@ void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
 
 template <typename DataType, typename Frame>
 void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
+    const gsl::not_null<tnsr::ab<DataType, 3, Frame>*> deriv_null_form_boosted,
+    const gsl::not_null<CachedBuffer*> cache,
+    internal_tags::deriv_null_form<DataType, Frame> /*meta*/) const {
+  const auto& deriv_null_form = cache->get_var(
+      *this, internal_tags::deriv_null_form_unboosted<DataType, Frame>{});
+  // Inverse-boost the first index, because it represents the derivative which
+  // is given in the boosted frame
+  sr::lorentz_boost(deriv_null_form_boosted, deriv_null_form,
+                    -solution_.boost_velocity());
+}
+
+template <typename DataType, typename Frame>
+void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
     const gsl::not_null<Scalar<DataType>*> lapse_squared,
     const gsl::not_null<CachedBuffer*> cache,
     internal_tags::lapse_squared<DataType> /*meta*/) const {
-  if (solution_.zero_spin()) {
+  if (solution_.zero_spin() and solution_.zero_velocity()) {
     const auto& r = get(cache->get_var(*this, internal_tags::r<DataType>{}));
     get(*lapse_squared) = 1.0 / (1.0 + 2.0 * solution_.mass() / r);
     return;
   }
   const auto& H = get(cache->get_var(*this, internal_tags::H<DataType>{}));
-  get(*lapse_squared) = 1.0 / (1.0 + 2.0 * square(null_vector_0_) * H);
+  const auto& null_form =
+      cache->get_var(*this, internal_tags::null_form<DataType, Frame>{});
+  get(*lapse_squared) = 1.0 / (1.0 + 2.0 * square(get<0>(null_form)) * H);
 }
 
 template <typename DataType, typename Frame>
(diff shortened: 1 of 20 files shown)


src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
```diff
@@ -214,6 +214,16 @@ namespace Solutions {
  *
  * Right now we use (\f$\ref{eq:sphertocartsimple}\f$), but we may
  * wish to use the other transformation in the future.
+ *
+ * ## Boost of the Kerr-Schild solution
+ *
+ * We add initial momentum to the solution by applying a Lorentz boost to the
+ * metric. Given the form of the Kerr-Schild metric in terms of a covariant
```
nikwit: Can you rephrase this to something more readable? Perhaps "The Kerr-Schild metric can be expressed covariantly in terms of (...). We construct the boosted metric by boosting these object individually."

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
```diff
@@ -368,143 +406,175 @@ class KerrSchild : public AnalyticSolution<3_st>,
...
+        const;
+
+    void operator()(
+        gsl::not_null<tnsr::I<DataType, 3, Frame>*> x_minus_center_boosted,
+        gsl::not_null<CachedBuffer*> /*cache*/,
         internal_tags::x_minus_center<DataType, Frame> /*meta*/) const;
 
-    void operator()(const gsl::not_null<Scalar<DataType>*> a_dot_x,
-                    const gsl::not_null<CachedBuffer*> cache,
+    void operator()(gsl::not_null<Scalar<DataType>*> a_dot_x,
```
nikwit, line 434: why did you delete all these `const` specifiers?

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.cpp
```diff
@@ -391,8 +450,7 @@ void KerrSchild::IntermediateComputer<DataType, Frame>::operator()(
   const auto& lapse = get(cache->get_var(*this, gr::Tags::Lapse<DataType>{}));
   const auto& lapse_squared =
       get(cache->get_var(*this, internal_tags::lapse_squared<DataType>{}));
-  get(*deriv_lapse_multiplier) =
-      -square(null_vector_0_) * lapse * lapse_squared;
+  get(*deriv_lapse_multiplier) = -lapse * lapse_squared;
```
nikwit, line 457: why can you set `square(null_vector_0_)` to 1 here but not other places?

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.cpp
```diff
@@ -58,6 +59,80 @@ void lorentz_boost_matrix(
     (*boost_matrix).get(i + 1, i + 1) += 1.0;
   }
 }
+
+template <size_t SpatialDim>
+tnsr::Ab<double, SpatialDim, Frame::NoFrame> lorentz_boost_matrix(
+    const std::array<double, SpatialDim>& velocity) {
+  tnsr::I<double, SpatialDim, Frame::NoFrame> velocity_as_tensor{};
```
nikwit: `tnsr::I<double, SpatialDim, Frame::NoFrame> velocity_as_tensor(velocity)` should work I think

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
```diff
@@ -53,5 +56,51 @@ template <size_t SpatialDim>
...
+template <size_t SpatialDim>
+tnsr::Ab<double, SpatialDim, Frame::NoFrame> lorentz_boost_matrix(
+    const std::array<double, SpatialDim>& velocity);
+/// @}
+
+/// @{
+/*!
+ * \ingroup SpecialRelativityGroup
+ * \brief Apply a Lorentz boost to the spatial part of a one form.
+ * \details The zero component of the one form is also specified.
```
nikwit: Maybe "This requires passing the 0th component of the one form as an additional argument.

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
```diff
@@ -53,5 +56,51 @@ template <size_t SpatialDim>
...
+    const double vector_component_0,
+    const std::array<double, SpatialDim>& velocity);
+/// @}
+
+/// @{
+/*!
+ * \ingroup SpecialRelativityGroup
+ * \brief Apply a Lorentz boost to a one form.
+ *
+ * Note that the Lorentz matrix is symmetric.
```
nikwit: I don't think this is relevant for this function.

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
```diff
@@ -53,5 +56,51 @@ template <size_t SpatialDim>
...
+void lorentz_boost(
+    const gsl::not_null<tnsr::a<DataType, SpatialDim, Frame>*> result,
+    const tnsr::a<DataType, SpatialDim, Frame>& vector,
+    const std::array<double, SpatialDim>& velocity);
+/// @}
+
+/// @{
+/*!
+ * \ingroup SpecialRelativityGroup
+ * \brief Apply a Lorentz boosts for each index of a rank-2 tensor with
```
nikwit: maybe something like: "(...) boost to each component"

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
```diff
@@ -53,5 +56,51 @@ template <size_t SpatialDim>
...
+
+template <size_t SpatialDim>
+tnsr::Ab<double, SpatialDim, Frame::NoFrame> lorentz_boost_matrix(
+    const std::array<double, SpatialDim>& velocity);
+/// @}
+
+/// @{
+/*!
+ * \ingroup SpecialRelativityGroup
+ * \brief Apply a Lorentz boost to the spatial part of a one form.
```
nikwit: You mean vector here instead of one form, right?

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
```diff
@@ -53,5 +56,51 @@ template <size_t SpatialDim>
...
+/*!
+ * \ingroup SpecialRelativityGroup
+ * \brief Apply a Lorentz boost to a one form.
+ *
+ * Note that the Lorentz matrix is symmetric.
+ */
+template <typename DataType, size_t SpatialDim, typename Frame>
+void lorentz_boost(
+    const gsl::not_null<tnsr::a<DataType, SpatialDim, Frame>*> result,
+    const tnsr::a<DataType, SpatialDim, Frame>& vector,
```
nikwit: one_form

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.cpp
```diff
@@ -58,6 +59,80 @@ void lorentz_boost_matrix(
...
+  for (size_t i = 0; i < SpatialDim + 1; ++i) {
+    result->get(i) = 0.;
+    for (size_t j = 0; j < SpatialDim + 1; ++j) {
+      result->get(i) += boost_matrix.get(i, j) * vector.get(j);
+    }
+  }
+}
+
+template <typename DataType, size_t SpatialDim, typename Frame>
+void lorentz_boost(
```
nikwit, line 109: is there a sensible situation where different indices of a tensor would be boosted with different velocities? Is this generalization necessary?

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.cpp
```diff
@@ -58,6 +59,80 @@ void lorentz_boost_matrix(
     (*boost_matrix).get(i + 1, i + 1) += 1.0;
   }
 }
+
+template <size_t SpatialDim>
+tnsr::Ab<double, SpatialDim, Frame::NoFrame> lorentz_boost_matrix(
```
nikwit, line 64: can you make this a symmetric tensor?

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.cpp
```diff
@@ -58,6 +59,80 @@ void lorentz_boost_matrix(
...
+    const std::array<double, SpatialDim>& velocity_first_index,
+    const std::array<double, SpatialDim>& velocity_second_index) {
+  if (velocity_first_index == make_array<SpatialDim>(0.) and
+      velocity_second_index == make_array<SpatialDim>(0.)) {
+    *result = tensor;
+    return;
+  }
+  const auto boost_matrix_1 = lorentz_boost_matrix(velocity_first_index);
+  const auto boost_matrix_2 = lorentz_boost_matrix(velocity_second_index);
+  for (size_t i = 0; i < SpatialDim + 1; ++i) {
```
nikwit, line 118: [optional] `tenex::evaluate<ti::a, ti::b>(result, boost_matrix_1(ti::C, ti::a) * boost_matrix_2(ti::D, ti::b) * tensor(ti::c, ti::d)` is quite concise.

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
```diff
@@ -53,5 +56,51 @@ template <size_t SpatialDim>
...
+
+/// @{
+/*!
+ * \ingroup SpecialRelativityGroup
+ * \brief Apply a Lorentz boost to a one form.
+ *
+ * Note that the Lorentz matrix is symmetric.
+ */
+template <typename DataType, size_t SpatialDim, typename Frame>
+void lorentz_boost(
```
nikwit, line 82: Ideally, there would be one implementation that works for a generic tensor type so do not have all of these overloads. If you think that would be overkill for now, maybe just add a comment that this can be done in the future.

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_WrappedGr.cpp
```diff
@@ -44,7 +44,7 @@ void compare_different_wrapped_solutions(const double mass,
 template <typename SolutionType>
 void test_copy_and_move(const SolutionType& solution) {
   test_copy_semantics(solution);
-  auto solution_copy = solution;
+  const auto& solution_copy = solution;
```
nikwit: this is just a reference now. Can you just pass in `solution` directly?

tests/Unit/Helpers/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/VerifyGrSolution.hpp
```diff
@@ -426,5 +426,55 @@ void verify_consistency(const Solution& solution, const double time,
                                               derivative_delta),
       derivative_approx);
 }
+
+/// Check the consistency of dependent quantities returned by a
```
nikwit: dependent on what?

nikwit (changes requested):
nice work overall. Two points:

1.) I think we discussed that with a boost, `KerrSchild` is technically time dependent but you have only implemented it for `t=0`. It is therefore technically no longer a `AnalyticSolution`.  Can you make sure to document this?

2.) As we discussed, could you write a test that computes tensors like the extrinsic curvature and the spacetime metric once boosted and once unboosted  and make sure they correspond to a manual transformation?

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
guilara: clang-tidy was complaining. Said only has effect in the cpp

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.cpp
guilara: I don't think its set to one, just that it seems not in the formula for the spatial derivative of the lapse.

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.cpp
guilara: Can I? This has one upper and one lower index

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
guilara: done

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.cpp
guilara: Done. Seems to work

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
guilara: Done.

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
guilara: I think I wouldn't want to delay this anymore. Added comment.

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_WrappedGr.cpp
guilara: Done.

tests/Unit/Helpers/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/VerifyGrSolution.hpp
guilara: On other metric quantities. Changed doc.

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
guilara: One form. This is because here and in SpEC they chose (rather confusingly) the lorentz boost matrix to be the usual _inverse_ lorentz boost matrix (or that for one forms).

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
guilara: Done

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
guilara: Okay

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
guilara: Okay. Removed comment

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.cpp
guilara: Also inherited from the implementation in SpEC, maybe to save overloads when boosting tensors with mixed up/down indices. Then one just changes the sign of the velocity in the relevant entry.

guilara:
@nikwit I've added one more test to manually boost the spacetime metric and check that I get the same as the implementation. Not entirely sure how to do it for the extrinsic curvature. I think I need to upgrade it to 4d tensor (not sure about this) then boost it and compare a 4d tensor obtained from the implementation.

guilara:
@nikwit was suggesting moving the KerrBoost to AnalyticData instead of AnalyticSolutions while I suggest leaving it here and add a warning in the doc.  

The reasoning is mainly because the Kerr boost as it is currently implemented (and as it is implemented in SpEC) is only usable for initial data and some assumptions are being made. For instance, we are not coding up the time dependence at all (the solution now is both function of position and time since its moving, but we set t = 0), and assumptions on the time derivative of the lapse and shift (which will be rewritten anyway when evolved).  

@nilsvu any thoughts on this?

nilsvu:
@geoffrey4444 Could you review this one too, since you wrote the Lorentz transformations here and boosts in SpEC?

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
```diff
@@ -214,6 +214,28 @@ namespace Solutions {
  *
  * Right now we use (\f$\ref{eq:sphertocartsimple}\f$), but we may
  * wish to use the other transformation in the future.
+ *
+ * ## Boost of the Kerr-Schild solution
+ *
+ * We add initial momentum to the solution by applying a Lorentz boost to the
+ * metric. Since the Kerr-Schild metric can be expressed covariantly in terms of
+ * the Minkowski metric, a scalar function and a one form, we constructing the
+ * metric by boosting the these objects individually. Notice that we also need
```
nilsvu: <s>the</s>

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
```diff
@@ -214,6 +214,28 @@ namespace Solutions {
...
+ * to appropriately boost the coordinates to the boosted frame.
```
nilsvu: I believe we're evaluating the boosted Kerr-Schild metric in the boosted frame, meaning that in these coordinates the black hole does not move, correct? So we inverse-boost the input coordinates to get to the unboosted frame, evaluate Kerr-Schild quantities in the unboosted frame, then boost these quantities. Please clarify the docs here.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
```diff
@@ -214,6 +214,28 @@ namespace Solutions {
...
+ *
+ * \warning While technically the boosted Kerr-Schild metric is dependent on
+ * both the time and space coordinates, we have implemented it only at $t = 0$
+ * as in SpEC. Therefore it is technically not an analytic solution and should
+ * not be used to compute errors with respect to it.
```
nilsvu, line 231: The solution should be stationary in the boosted frame; so shouldn't all time derivatives be zero and this be a solution at any $t$? Am I misunderstanding something?

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.cpp
nilsvu: Not currently supported.

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
nilsvu: The argument type is a vector (`tnsr::I`), so I also think this documentation should read `vector`. Refer to the docs of `lorentz_boost_matrix` above for details and math. I also find those docs confusing though. I believe they say that $\Lambda^a_\bar{a}$ is computed, which boosts a vector from barred to unbarred or a one-form from unbarred to barred (consistent with Blandford/Thorne Sec. 2.7). So my suggestion is:

- The docs and argument for this function should read `vector` (they boost a vector)
- The function for the one-form below should apply the inverse Lorentz transformation (by inverting the sign of the velocity). This means you add a minus sign to the function below and remove the minus signs from where the function is used in Kerr-Schild. Make sure to clarify this in the docs of the function.

@geoffrey4444 can you confirm or clarify this?

tests/Unit/Helpers/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/VerifyGrSolution.hpp
```diff
@@ -426,5 +426,55 @@ void verify_consistency(const Solution& solution, const double time,
                                               derivative_delta),
       derivative_approx);
 }
+
+/// Check the consistency of quantities depenent on other metric quantities
```
nilsvu: typo: dependent

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_KerrSchild.cpp
```diff
@@ -366,13 +380,116 @@ void test_zero_spin_optimization(const DataType& used_for_size) {
       });
 }
 
+template <typename Frame, typename DataType>
+void test_boosted_for_zero_velocity(const DataType& used_for_size) {
+  const std::array<double, 3> boost_velocity{{0.0, 0.0, 1.0e-50}};
```
nilsvu: Maybe `std::numeric_limits<double>::epsilon()`?

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_KerrSchild.cpp
```diff
@@ -366,13 +380,116 @@ void test_zero_spin_optimization(const DataType& used_for_size) {
       });
 }
 
+template <typename Frame, typename DataType>
+void test_boosted_for_zero_velocity(const DataType& used_for_size) {
```
nilsvu, line 396: Nice test!

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_KerrSchild.cpp
```diff
@@ -366,13 +380,116 @@ void test_zero_spin_optimization(const DataType& used_for_size) {
...
+  const std::array<double, 3> spin{{0.0, 0.0, 0.0}};
+  const std::array<double, 3> center{{0.0, 0.0, 0.0}};
+  gr::Solutions::KerrSchild solution(mass, spin, center);
+
+  // Boosted KerrSchild solution
+  const std::array<double, 3> velocity{{0.5, 0.0, 0.0}};
+  gr::Solutions::KerrSchild boosted_solution(mass, spin, center, velocity);
+
+  // Need to evaluate at the boosted coordinates
+  // Boost coordinates
```
nilsvu, line 445: These are the unboosted coords, no? (related to comments above)

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_KerrSchild.cpp
```diff
@@ -366,13 +380,116 @@ void test_zero_spin_optimization(const DataType& used_for_size) {
...
+      [&all_tags_zero_velocity, &all_tags_tiny_velocity](auto tag_v) {
+        using tag = tmpl::type_from<decltype(tag_v)>;
+        CHECK_ITERABLE_APPROX(get<tag>(all_tags_zero_velocity),
+                              get<tag>(all_tags_tiny_velocity));
+      });
+}
+
+template <typename Frame, typename DataType>
+void test_boosted_spacetime_metric(const DataType& used_for_size) {
+  // Test that the extrinsic curvature of the boosted solution indeed
```
nilsvu: extrinsic curvature -> spacetime metric? Also test extrinsic curvature?

tests/Unit/PointwiseFunctions/SpecialRelativity/Test_LorentzBoostMatrix.cpp
```diff
@@ -86,3 +167,9 @@ SPECTRE_TEST_CASE(
   d = large_velocity_squared;
   CHECK_FOR_DOUBLES(test_lorentz_boost_matrix_analytic, (1, 2, 3));
 }
+
+SPECTRE_TEST_CASE("Unit.PointwiseFunctions.SpecialRelativity.LorentzBoost",
```
nilsvu: combine with test case above

nilsvu (changes requested):
Yes please keep this class in `AnalyticSolution` with the warning in the docs.

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
guilara: You are right Nils. My mistake here, Niko. It is the transformation for a vector. But I should take the inverse Lorentz matrix (or add a sign to the velocity) as they transform with the inverse matrix wrt one forms.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
guilara: Will clarify. I got confused about it myself at some point. We start with coordinates in the frame where the black hole is moving. Then boost them to the frame where the black hole is at rest, and where we can construct the stationary solution. Then we boost all covariant quantities back to the frame where it is moving.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
guilara: I believe it is stationary in the rest frame, which is not the frame we want. In the frame we want, the black hole is moving and so the metric will be dependent on t and x. I didn't code this t dependence  since I didn't see this in SpEC --it seems to me it should only enter in the function that computes the radius to the center. 

I also didn't code the time derivatives of the lapse and shift in the frame where the black hole is moving (which are zero in the rest frame of course) but left them as zero. This should be fine because I think they are replaced by the gauge functions we use to evolve this.

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_KerrSchild.cpp
guilara: Not yet sure how to do it. I think I would need to elevate the extrinsic curvature to a 4d tensor in both frames.

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_KerrSchild.cpp
guilara: Maybe the terminology ''boosted/unboosted'' is a bit confusing.  Rest frame (of the BH) coords I would say.

tests/Unit/PointwiseFunctions/SpecialRelativity/Test_LorentzBoostMatrix.cpp
guilara: Combined

tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_KerrSchild.cpp
guilara: Okay

tests/Unit/Helpers/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/VerifyGrSolution.hpp
guilara: Corrected

src/PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp
nilsvu: I think the sign for this function is correct, but the one below (for one-forms) should have a minus sign.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
nilsvu: Hm, for initial data I thought we're solving in a frame where the BHs are at rest.. @geoffrey4444 can you clarify how this works in SpEC?

geoffrey4444:
> @geoffrey4444 Could you review this one too, since you wrote the Lorentz transformations here and boosts in SpEC?

I'll try but really swamped this week.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
nilsvu: Talked about this in the meeting today. I think you're correct. For initial data we go into a rotating frame that gets rid of the translation but keeps the Lorentz contraction.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
nilsvu: Ok I think you're correct. Let me look at the SpEC implementation too.

guilara:
@nilsvu  I have rebased this  and added an assert that would throw an error if we call a tag I can't compute for the boost.

geoffrey4444 (approved):
@nilsvu does this look ready to go to you?

nilsvu (approved):
Looks good. Thanks for fixup this up. Pleas squash your commits. Also to keep the git blame record straight, could you add this to your commit message:

```sh
# Two empty lines


Co-authored-by: Nils Vu <nilsvu@caltech.edu>
```

nilsvu:
It's two empty lines, no `>` symbols

guilara:
> It's two empty lines, no `>` symbols

Oh, sure

nilsvu:
Thank you for getting this done @guilara !

guilara:
> Thank you for getting this done @guilara !

Thank you for all the detailed help with it

PR #5614, opened 2023-11-03
Add ErrorIfDataTooBig Event

wthrowe:
@geoffrey4444 I haven't tested this much since several executables don't compile, but the BBH executable does if you want to try it.  I'll leave this as a draft until I've figured out the other issues.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Events/ErrorIfDataTooBig.hpp
+++ b/src/ParallelAlgorithms/Events/ErrorIfDataTooBig.hpp
@@ -1,0 +1,172 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cmath>
+#include <cstddef>
+#include <limits>
+#include <optional>
+#include <pup.h>
+#include <string>
+#include <type_traits>
+#include <unordered_set>
+#include <vector>
+
+#include "DataStructures/DataBox/ObservationBox.hpp"
+#include "DataStructures/DataBox/TagName.hpp"
+#include "DataStructures/DataBox/ValidateSelection.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/ExtractPoint.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Options/String.hpp"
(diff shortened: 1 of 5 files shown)


src/ParallelAlgorithms/Events/ErrorIfDataTooBig.hpp
```diff
@@ -0,0 +1,163 @@
...
+template <size_t Dim, typename Tensors, typename NonTensorComputeTags>
+class ErrorIfDataTooBig : public Event {
+ public:
+  /// \cond
+  explicit ErrorIfDataTooBig(CkMigrateMessage* /*unused*/) {}
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(ErrorIfDataTooBig);  // NOLINT
+  /// \endcond
+
+  struct VariablesToCheck {
```
nilsdeppe, line 71: Thanks for doing this! :D 
Would it be worth making a threshold per variable? I could see having different thresholds for e.g. constraints and say det(gamma) 🤷 Up to you! I'm fine either way. I think this is a very valuable addition and we shouldn't hold it up trying to perfect it :)

knelli2: Maybe make it a variant of a single number and a vector so people can decide in the input file? Then just check that the vectors are the same size in the constructor

wthrowe: I don't think it's worth it since you can already do this:
```yaml
Events:
  - ErrorIfDataTooBig:
      VariablesToCheck: [Constraint]
      Threshold: 0.1
  - ErrorIfDataTooBig:
      VariablesToCheck: [det(gamma)]
      Threshold: 10.0
```
I doubt people will want a huge number of different thresholds.

wthrowe (commented):
I removed printing the time.  It was nice, but it didn't compile in elliptic executables and the place this is used doesn't have information about what the appropriate tag to access is.

nilsdeppe (commented):
LGTM!

knelli2 (approved):

nilsdeppe (approved):

Issue #5600, opened 2023-10-26
BBH domain needs distorted frame in object cubes


knelli2:
In runs that @geoffrey4444 has been doing, we have found that the horizon radius in the distorted frame is growing (which it should be). However, we only have a distorted frame in the inner shell around each of the objects. Right before merger, the horizon radius grows so large that the horizon surface moves outside this inner shell and into the cube around the object. However, this cube doesn't have a distorted frame. Thus when we ask to convert the distorted frame points of the horizon to the block logical frame using the `block_logical_coordinates` function, because the block doesn't have a distorted frame, we just say the point isn't in the domain. This then causes the horizon finder to fail because it needs all block logical coordinates to find the horizon.

The solution is to create a new `ShapeMapTransitionFunction` that radially linearly goes from being 1 at the inner spherical surface to 0 at the the cubical outer surface. If it is easy, it should be generalized so that both the inner and outer surface can be any shape (either spherical or flat).

knelli2:
Closing this because it was completed in #5619.

PR #5616, opened 2023-11-07
Add wedge transition for shape map

knelli2:
## Proposed changes

A new transition function for the shape map that is to be used in concentric wedge blocks. The inner and outer boundaries of these wedges can be spherical, flat, or in-between just like the Wedge coordinate map. This transition will mostly be used for the rectangular BinaryCompactObject domain for the shells and cubes around each object (PR to come for that).

Needed for #5600.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
@@ -1,0 +1,342 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp"
+
+#include <array>
+#include <cmath>
+#include <optional>
+#include <pup.h>
+
+#include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "NumericalAlgorithms/RootFinding/QuadraticEquation.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/CaptureForError.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/MakeArray.hpp"
+#include "Utilities/MakeWithValue.hpp"
+#include "Utilities/StdArrayHelpers.hpp"
+#include "Utilities/StdHelpers.hpp"
+
+namespace domain::CoordinateMaps::ShapeMapTransitionFunctions {
+template <typename T>
+T Wedge::Surface::distance(const std::array<T, 3>& coords) const {
+  // Short circuit if it's a sphere. Then the distance is trivially the radius
+  // of this surface
+  if (sphericity == 1.0) {
+    return make_with_value<T>(coords[0], radius);
+  }
+
+  ASSERT(coords[2] != 0.0,
+         "For wedge transition, rotated coords cannot have z component equal "
+         "to zero.");
+
+  // D = R * [ (1 - s) / (sqrt(3) * cos(theta)) + s]
+  // cos(theta) = z / r
+  return radius *
+         ((1.0 - sphericity) / (coords[2] * sqrt(3.0)) * magnitude(coords) +
+          sphericity);
+}
+
+void Wedge::Surface::pup(PUP::er& p) {
+  p | radius;
+  p | sphericity;
+}
+
+bool Wedge::Surface::operator==(const Wedge::Surface& other) const {
+  return radius == other.radius and sphericity == other.sphericity;
+}
+
+bool Wedge::Surface::operator!=(const Wedge::Surface& other) const {
+  return not(*this == other);
+}
+
+Wedge::Wedge(double inner_radius, double outer_radius, double inner_sphericity,
+             double outer_sphericity, OrientationMap<3> orientation_map)
+    : inner_surface_(Surface{inner_radius, inner_sphericity}),
+      outer_surface_(Surface{outer_radius, outer_sphericity}),
+      orientation_map_(std::move(orientation_map)),
+      direction_(orientation_map_.inverse_map()(Direction<3>::upper_zeta())) {}
+
+double Wedge::operator()(const std::array<double, 3>& source_coords) const {
+  return call_impl<double>(source_coords);
+}
+DataVector Wedge::operator()(
+    const std::array<DataVector, 3>& source_coords) const {
+  return call_impl<DataVector>(source_coords);
+}
+
+std::optional<double> Wedge::original_radius_over_radius(
+    const std::array<double, 3>& target_coords, double distorted_radius) const {
+  const double radius = magnitude(target_coords);
+  CAPTURE_FOR_ERROR(target_coords);
+  CAPTURE_FOR_ERROR(distorted_radius);
+  CAPTURE_FOR_ERROR(radius);
+  CAPTURE_FOR_ERROR(orientation_map_);
+  CAPTURE_FOR_ERROR(direction_);
+
+  // Couple protections that would make a point completely outside of the domain
+  // of validity for any wedge
+  if (equal_within_roundoff(radius, 0.0) or
+      equal_within_roundoff(distorted_radius, 1.0) or distorted_radius > 1.0) {
+    return std::nullopt;
+  }
+
+  // First we check the extremal case of being outside the outermost radius. We
+  // can check the outermost radius because its surface doesn't move. We can't
+  // check the innermost surface because the inner bound is the origin which we
+  // already checked above.
+  if (radius - eps_ > outer_surface_.radius) {
+    return std::nullopt;
+  }
+
+  // Rotate the coordinates so they are aligned with +z. This makes checking if
+  // the target_coord is in the proper wedge very simple since we only have to
+  // check if the target coord is in the direction of the +z wedge
+  const std::array<double, 3> rotated_coords =
+      discrete_rotation(orientation_map_.inverse_map(), target_coords);
+  CAPTURE_FOR_ERROR(rotated_coords);
+
+  // This comparison only works because the opening angle of the wedge is pi/2
+  // in both x and y
+  if (rotated_coords[2] <
+      std::max(abs(rotated_coords[0]), abs(rotated_coords[1]))) {
+    return std::nullopt;
+  }
+
+  // If distorted radius is 0, this means the map is the identity so the radius
+  // and the original radius are equal. Also we don't want to divide by 0 below
+  if (equal_within_roundoff(distorted_radius, 0.0)) {
+    return std::optional{1.0};
+  }
+
+  const double outer_distance = outer_surface_.distance(rotated_coords);
+  const double inner_distance = inner_surface_.distance(rotated_coords);
+  const double distance_difference = outer_distance - inner_distance;
+
+  // If we are at the overall outer distance, then the transition function is 0
+  // so the map is again the identity so the radius and original radius are
+  // equal. We can't check the overall inner distance because that has been
+  // distorted so we don't know where the mapped inner distance is.
+  if (equal_within_roundoff(radius, outer_distance)) {
+    return std::optional{1.0};
+  }
+
+  // Solving equation of the form rtil*x^2 + ((D_o - D_i)/SumYlm - D_o)*x - (D_o
+  // - D_i)/SumYlm = 0
+  const double a = radius;
+  const double c = -distance_difference / distorted_radius;
+  const double b = -c - outer_distance;
+
+  std::optional<std::array<double, 2>> roots = real_roots(a, b, c);
+
+  if (roots.has_value()) {
+    for (const double root : roots.value()) {
+      // Check if the root is positive and within the inner and outer distance
+      // (divided by the mapped radius)
+      if (root > 0.0 and root + eps_ >= inner_distance / radius and
+          root - eps_ <= outer_distance / radius) {
+        return std::optional{root};
+      }
+    }
+    return std::nullopt;
+  } else {
+    return std::nullopt;
+  }
+}
+
+double Wedge::map_over_radius(
+    const std::array<double, 3>& source_coords) const {
+  return map_over_radius_impl<double>(source_coords);
+}
+DataVector Wedge::map_over_radius(
+    const std::array<DataVector, 3>& source_coords) const {
+  return map_over_radius_impl<DataVector>(source_coords);
+}
+
+std::array<double, 3> Wedge::gradient(
+    const std::array<double, 3>& source_coords) const {
+  return gradient_impl<double>(source_coords);
+}
+std::array<DataVector, 3> Wedge::gradient(
+    const std::array<DataVector, 3>& source_coords) const {
+  return gradient_impl<DataVector>(source_coords);
+}
+
+template <typename T>
+T Wedge::call_impl(const std::array<T, 3>& source_coords) const {
+  const std::array<T, 3> rotated_coords =
+      discrete_rotation(orientation_map_.inverse_map(), source_coords);
+  check_distances(rotated_coords);
+  T outer_distance = outer_surface_.distance(rotated_coords);
+
+  return (outer_distance - magnitude(rotated_coords)) /
+         (outer_distance - inner_surface_.distance(rotated_coords));
+}
+
+template <typename T>
+T Wedge::map_over_radius_impl(const std::array<T, 3>& source_coords) const {
+  const std::array<T, 3> rotated_coords =
+      discrete_rotation(orientation_map_.inverse_map(), source_coords);
+  check_distances(rotated_coords);
+  const T radius = magnitude(rotated_coords);
+  const T outer_distance = outer_surface_.distance(rotated_coords);
+
+  return (outer_distance - radius) /
+         ((outer_distance - inner_surface_.distance(rotated_coords)) * radius);
+}
+
+template <typename T>
+std::array<T, 3> Wedge::gradient_impl(
+    const std::array<T, 3>& source_coords) const {
+  // If both surfaces are spherical then we short circuit because the distances
+  // are constant and we only need to take a derivative of r.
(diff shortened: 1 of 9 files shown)


src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Sphere.hpp
```diff
@@ -72,4 +72,16 @@ class SphereTransition final : public ShapeMapTransitionFunction {
...
+/// \cond
+// The Sphere class was renamed from SphereTransition at some point. This is
+// only an issue when it comes to pupping. Specifically deserializing, because
+// the name of the class is serialized along with its data. So if we only
+// register the Sphere derived class, but try to read in data that was
+// serialized with the SphereTransition name, then we will get an unregistered
+// PUP_ID error. Therefore, we add this class with the SphereTransition name
+// which is identical to the Sphere class for purposes of deserialization. This
+// class is not to be used otherwise.
+class SphereTransition : public Sphere {};
```
knelli2: Not super happy about this, but I would like to rename the class because the original name is redundant. If reviewers are strongly against this, I can forgo the name change

src/Domain/Structure/OrientationMap.hpp
```diff
@@ -95,6 +95,11 @@ class OrientationMap {
   /// The corresponding Orientation of the host in the frame of the neighbor.
   OrientationMap<VolumeDim> inverse_map() const;
 
+  /// The stored map directions
+  const std::array<Direction<VolumeDim>, VolumeDim>& mapped_directions() const {
```
wthrowe: If you're exposing this, document what it means.  I think dropping this commit and using `operator()` is clearer, though.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Sphere.hpp
wthrowe: Renaming seems fine.  We could add a way to override the name used for the charm registration if we hit this often.

That said, are you sure this doesn't need its own `my_PUP_ID` and similar?

tests/Unit/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/CMakeLists.txt
```diff
@@ -0,0 +1,7 @@
+# Distributed under the MIT License.
```
wthrowe, line 1: Add directory to parent CMakeLists.txt.  Looks like done in wrong commit.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp
```diff
@@ -0,0 +1,245 @@
...
+  template <typename T>
+  void check_distances(const std::array<T, 3>& coords) const;
+
+  Surface overall_inner_surface_{};
+  Surface overall_outer_surface_{};
+  Surface this_wedge_inner_surface_{};
+  Surface this_wedge_outer_surface_{};
+  OrientationMap<3> orientation_map_{};
+  Direction<3> direction_{};
+  static constexpr double eps_ = std::numeric_limits<double>::epsilon() * 100;
```
wthrowe, line 237: Include `<limits>`.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp
```diff
@@ -0,0 +1,245 @@
...
+ * \frac{\partial f}{\partial x_i} = \frac{-\frac{x_i}{r}}{D_{\text{out}} -
+ * D_{\text{in}}}
+ * \end{equation}
+ */
+class Wedge final : public ShapeMapTransitionFunction {
+  struct Surface {
+    double radius{};
+    double sphericity{};
+    template <typename T>
+    T distance(const std::array<T, 3>& coords) const;
```
wthrowe, line 163: Distance to what?

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp
```diff
@@ -0,0 +1,245 @@
...
+ * \begin{equation}
+ * f(r, \theta, \phi) = \frac{D_{\text{out}}(r, \theta, \phi) -
+ * r}{D_{\text{out}}(r, \theta, \phi) - D_{\text{in}}(r, \theta, \phi)},
+ * \label{eq:transition_func}
+ * \end{equation}
+ *
+ * where
+ *
+ * \begin{equation}
+ * D(r, \theta, \phi) = R\left(\frac{1 - s}{\sqrt{3}\cos{\theta}} + s\right),
```
wthrowe, line 35: It looks like D doesn't depend on r (or phi, but that's less important).  I think the previous equation would be easier to understand if the r-dependence were explicitly gone, since it's surprising to see a complicated dependence on r when the equation is linear.

Reading more, I see that you treat this as a field a lot, which is more natural with all the coordinates included.  Maybe move the \note paragraph below up here to emphasize the lack of r-dependence at the start?

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -0,0 +1,484 @@
...
+          Surface{this_wedge_outer_radius, this_wedge_outer_sphericity}),
+      orientation_map_(std::move(orientation_map)),
+      direction_(orientation_map_.mapped_directions()[1]) {
+  // Getting the direction this way is only works because of the rotations from
+  // orientations_for_sphere_wrappings() for the wedges
+  if (direction_.axis() == Direction<3>::Axis::Eta) {
+    direction_ = Direction<3>{Direction<3>::Axis::Zeta, direction_.side()};
+  } else if (direction_.axis() == Direction<3>::Axis::Zeta) {
+    direction_ = Direction<3>{Direction<3>::Axis::Eta, direction_.side()};
+  }
```
wthrowe: Shouldn't this be something like `orientation_map_.inverse_map()(Direction<3>::upper_zeta())`?

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -0,0 +1,484 @@
...
+  // First we check the extremal case of being outside the outermost radius. We
+  // can check the outermost radius because its surface doesn't move. We can't
+  // check the innermost surface because the inner bound is the origin which we
+  // already checked above.
+  if (radius - eps_ > overall_outer_surface_.radius) {
+    return std::nullopt;
+  }
+
+  // Before calculating inner/outer distance, we must determine if the point is
+  // in our block direction or not (we'll determine if it's actually in our
```
wthrowe: I haven't worked through the next 100 lines of logic, but aren't we in the +x wedge if `x > max(|y|, |z|)` and similar for the other 5?

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -0,0 +1,484 @@
...
+    }
+  } else {
+    ERROR("Inconsistent angles.");
+  }
+
+  // We are in the proper wedge direction. It is safe to assume the coordinates
+  // can be rotated from their original orientation to now "aligned" with +z.
+  // Therefore we can continue with computing the original radius.
+  const std::array<double, 3> rotated_coords =
+      discrete_rotation(orientation_map_.inverse_map(), target_coords);
```
wthrowe, line 101: You could rotate first, and then you only have to check if you're in the +z wedge.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -0,0 +1,484 @@
...
+  CAPTURE_FOR_ERROR(rotated_coords);
+
+  // If distorted radius is 0, this means the map is the identity so the radius
+  // and the original radius are equal. Also we don't want to divide by 0 below
```
wthrowe, line 112: I'm confused about how the radius can be 0 inside a wedge.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -0,0 +1,484 @@
...
+           sqrt(3.0);
+  };
+
+  // We can make some simplifications if either of the surfaces are spherical
+  // because then the derivative of the distance is zero since it's constant. In
+  // the first two branches, it's safe to assume the other sphericity isn't 1
+  // because of the above check.
+  if (overall_outer_surface_.sphericity == 1.0) {
+    const T total_factor = one_over_denom * outer_distance_minus_radius *
+                           make_factor(overall_inner_surface_);
```
wthrowe: Since you seem concerned with such things in this function, pretty sure `T total_factor = make_factor(...); total_factor *= ...;` saves you an allocation.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -0,0 +1,484 @@
...
+
+    for (size_t i = 0; i < 2; i++) {
+      gsl::at(result, i) -=
+          total_factor * rotated_coords[2] * gsl::at(rotated_coords, i);
+    }
+
+    result[2] += total_factor * (square(radius) - square(rotated_coords[2]));
+  } else if (overall_inner_surface_.sphericity == 1.0) {
+    std::array<T, 3> outer_deriv_distance =
+        make_array<3>(make_factor(overall_outer_surface_));
```
wthrowe: Rewriting this like the previous case where you compute an overall factor instead of an array will save allocations and FLOPs.

Probably also for the case below, but I haven't worked that one out in enough detail to be sure.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -0,0 +1,484 @@
...
+}
+
+bool Wedge::operator!=(const ShapeMapTransitionFunction& other) const {
+  return not(*this == other);
+}
+
+// checks that the magnitudes are all between `r_min_` and `r_max_`
+template <typename T>
+void Wedge::check_distances([
+    [maybe_unused]] const std::array<T, 3>& coords) const {
```
wthrowe, line 299: Is this actually what clang-format wants?  Yuck.

wthrowe (changes requested):

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Sphere.hpp
knelli2: The reason I added this was because of the versioning test in `Test_Domain` where we deserialize a BCO domain from H5 that was serialized with the old transition function name. Not sure if there's a good way to override charm registration name when reading something in that was already written. Open to suggestions

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
knelli2: Damn, you're right. And I spent so much time thinking about angles...

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
knelli2: Ah, this is a poor name for the variable then. The "distorted radius" is $\sum_{lm} \lambda_{lm}(t)Y_{lm}$. It's documented in the base class. So if this quantity is 0, then there is no distortion.

knelli2:
@wthrowe Pushed a fixup. I'll remove that one commit and fix the wrong commit CMakeLists.txt thing when I squash

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Sphere.hpp
wthrowe: I meant we could provide a way to override the second argument in
```c++
    PUPable_reg2(class_to_register,
                 registration_name<class_to_register>().c_str());
```
but if it's only for one thing probably not worth it.

Ping on the question about my_PUP_ID.

wthrowe (changes requested):
Pinged on one thing.  You can squash the current fixup.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Sphere.hpp
knelli2: I tried compiling with
```cpp
class SphereTransition : public Sphere {};
PUP::able::PUP_ID SphereTransition::my_PUP_ID = 0;
```
and got an error about `no member named 'my_PUP_ID'`. The goal is to have the `SphereTransition` class be identical to the `Sphere` class in every way except name, so I think it makes sense to not give it its own PUP_ID.

wthrowe: Yes, you'd need to add the stuff in the class body as well.

> I think it makes sense to not give it its own PUP_ID.

But you are assigning two `PUP_ID`s, since you register twice, just one is overwriting the other.  Looking through the charm code I *think* this has the effect that you cannot serialize the first one registered through a base class pointer.  Or maybe you can but it is recorded as being the other type, not sure.  If you want to do this you definitely need tests that the serialization works the way you expect it to.

knelli2: Hmm ok. This is getting too complicated and I'd rather not run the risk of breaking serialization things like this, so I'll just forgo the name change. It's not a huge deal, I'll just name future transition functions properly. If there ever comes a time where we drop support for certain versions of maps, then we can revisit this.

wthrowe (dismissed):

tests/Unit/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Test_Wedge.cpp
```diff
@@ -0,0 +1,350 @@
...
+                      overall_inner_radius / sqrt(3.0));
+      check_bad_point("Too big radius", this_wedge_outer_radius * 10.0,
+                      this_wedge_outer_radius * 50.0);
+    }
+  }
+}
+
+SPECTRE_TEST_CASE("Unit.Domain.CoordinateMaps.Shape.Wedge", "[Domain][Unit]") {
+  // test_only_transition();
+  (void)test_only_transition;
```
nilsvu: Looks like you disabled this for debugging? Also the other test function should be called.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp
```diff
@@ -0,0 +1,252 @@
...
+   * inner surface
+   * \param this_wedge_outer_sphericity Sphericity of this particular wedge's
+   * outer surface
+   * \param orientation_map The same orientation map that was passed into the
+   * corresponding `domain::CoordinateMaps::Wedge` block.
+   */
+  Wedge(double overall_inner_radius, double overall_outer_radius,
+        double overall_inner_sphericity, double overall_outer_sphericity,
+        double this_wedge_inner_radius, double this_wedge_outer_radius,
+        double this_wedge_inner_sphericity, double this_wedge_outer_sphericity,
```
nilsvu: Question: what do you need the "this wedge" quantities for? I thought the transition only depends on the inner and outer radius/sphericity.

nilsvu (changes requested):

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp
knelli2: For the inverse function, we have to return `nullopt` if we are outside of the given block so we need to know this block's bounds

nilsvu: That's not a requirement we currently impose on maps. Many maps return logical coordinates outside of [-1, 1] from the inverse function and only return `nullopt` if they failed to do the inverse.

knelli2: Question that's tangentially related, do we actually use the inverse mapping at all in the evolution equations? Or is is just for horizon finds/control system things/observing in different frames?

tests/Unit/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Test_Wedge.cpp
knelli2: Oops. And the other test function is called at the end of `test_only_transition`

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp
nilsvu: The inverse is only used for things like interpolations I think.

knelli2: Hmm ok I'll change it to just check if it's inside the range of the transition function not the block itself

knelli2: @nilsvu ok pushed a fixup

nilsvu (approved):

wthrowe:
Please don't squash fixups before I have a chance to look at them.

knelli2:
Oh sorry @wthrowe that's totally my bad. I can `git reflog` and re-push the fixup if that'd be easier for you?

wthrowe (approved):

PR #5617, opened 2023-11-07
Add IMEX time-stepper interface and Heun2 IMEX

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/TimeSteppers/ImexRungeKutta.hpp
+++ b/src/Time/TimeSteppers/ImexRungeKutta.hpp
@@ -1,0 +1,75 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <vector>
+
+#include "Time/TimeSteppers/ImexTimeStepper.hpp"
+#include "Time/TimeSteppers/RungeKutta.hpp"
+#include "Utilities/Gsl.hpp"
+
+/// \cond
+class TimeDelta;
+namespace TimeSteppers {
+template <typename T>
+class ConstUntypedHistory;
+template <typename T>
+class MutableUntypedHistory;
+}  // namespace TimeSteppers
+/// \endcond
+
+namespace TimeSteppers {
+/*!
+ * \ingroup TimeSteppersGroup
+ * Intermediate base class implementing a generic IMEX Runge-Kutta
+ * scheme.
+ *
+ * Implements most of the virtual methods of ImexTimeStepper for a
+ * generic Runge-Kutta method.  Derived classes must implement the
+ * requirements of the `RungeKutta` base class, as well as
+ * `imex_order()`, `implicit_stage_order()`, and
+ * `implicit_butcher_tableau()`.
+ */
+class ImexRungeKutta : public virtual RungeKutta,
+                       public virtual ImexTimeStepper {
+ public:
+  /// Implicit part of the Butcher tableau.  Most parts of the tableau
+  /// must be the same as the explicit part, and so are omitted.
+  struct ImplicitButcherTableau {
+    /*!
+     * The coefficient matrix of the substeps.  We can only reasonably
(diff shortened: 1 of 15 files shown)


src/Time/TimeSteppers/ImexTimeStepper.hpp
```diff
@@ -0,0 +1,177 @@
...
+ *   Y_{n+1} =
+ *   Y_{n,\text{explicit}} + Y_{n,\text{inhomogeneous}} + w_n S(Y_{n+1})
+ * \f}
+ *
+ * Here \f$Y_{n,\text{explicit}}\f$ is the value of the evolved
+ * variables before the implicit substep is applied,
+ * \f$Y_{n,\text{inhomogeneous}}\f$ is the contribution of the past
+ * values from the implicit derivative history, and \f$S(\cdot)\f$ is
+ * the implicit portion of the right-hand-side (generally source
+ * terms for PDEs).  We call \f$w_n\f$ the \emph{implicit weight}.
```
wthrowe: Doxygen doesn't like `\emph`.  Need to replace it with something else.  (Not sure how I messed that locally).

src/Time/TimeSteppers/ImexRungeKutta.hpp
```diff
@@ -0,0 +1,71 @@
...
+class ConstUntypedHistory;
+template <typename T>
+class MutableUntypedHistory;
+}  // namespace TimeSteppers
+/// \endcond
+
+namespace TimeSteppers {
+/*!
+ * \ingroup TimeSteppersGroup
+ * Intermediate base class implementing a generic IMEX Runge-Kutta
```
nilsdeppe, line 25: [optional] One of the best discussions (not to say I find it crystal clear) of IM(EX) RK methods I came across is this one from NASA: https://ntrs.nasa.gov/api/citations/20160005923/downloads/20160005923.pdf Might be worth linking/referencing if you think it could be useful to people? Numerical Recipes doesn't really discuss this in a lot of detail so having something else for folks to learn about it with could be good. Maybe also reference the second Hairer+ book?

src/Time/TimeSteppers/ImexRungeKutta.hpp
```diff
@@ -0,0 +1,71 @@
...
+ */
+class ImexRungeKutta : public virtual RungeKutta,
+                       public virtual ImexTimeStepper {
+ public:
+  /// Implicit part of the Butcher tableau.  Most parts of the tableau
+  /// must be the same as the explicit part, and so are omitted.
+  struct ImplicitButcherTableau {
+    /*!
+     * The coefficient matrix of the substeps.  We can only reasonably
+     * support EDIRK methods, so the tableau must be lower-triangular
```
nilsdeppe: What about ESDIRK and QESDIRK? I don't know the detailed differences, so I would find it helpful to list those explicitly as either being okay or not. I think this rules out collocation-type methods, at least as the most obvious ones to me. Is that right? Might also be worth stating explicitly

nilsdeppe (changes requested):
LGTM, a couple documentation questions. At group meeting you said you were writing a tutorial so if you think the dox suggestions should go in that instead, just let me know :)

src/Time/TimeSteppers/ImexRungeKutta.hpp
wthrowe: I added the NASA reference here, since it explains the notation well.  I'll include Hairer in the tutorial, since that has a bit of discussion of properties of implicit methods and Hairer covers that in depth.

nilsdeppe (commented):
LGTM! That clarifies things for me, thank you! Please go ahead and squash :)

nilsdeppe (approved):

PR #5619, opened 2023-11-09
Use wedge shape map transition in BCO domain

### LINKED ISSUE #5600: BBH domain needs distorted frame in object cubes

knelli2:
## Proposed changes

Adds the ability to use the wedge transition in the sphere and cube blocks around each object in the BCO domain. This option is not added for the cylindrical BCO domain because different transition functions would be required.

Fixes #5600.

### Upgrade instructions

For the `BinaryCompactObject` domain creator, you must add the following two options to each of the shape map blocks

```yaml
BinaryCompactObject:
  ...
  TimeDependentMaps:
    ...
    ShapeMapA:
      ...
      TransitionEndsAtCube: true
    ShapeMapB: #Same as A
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #5616.~

### DIFF
--- a/src/Domain/Creators/BinaryCompactObject.cpp
+++ b/src/Domain/Creators/BinaryCompactObject.cpp
@@ -286,7 +286,7 @@ BinaryCompactObject::BinaryCompactObject(
 }
 
 BinaryCompactObject::BinaryCompactObject(
-    std::optional<bco::TimeDependentMapOptions> time_dependent_options,
+    std::optional<bco::TimeDependentMapOptions<false>> time_dependent_options,
     typename ObjectA::type object_A, typename ObjectB::type object_B,
     double envelope_radius, double outer_radius,
     const typename InitialRefinement::type& initial_refinement,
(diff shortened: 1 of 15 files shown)


knelli2:
No longer dependent and rebased on develop

wthrowe (approved):

PR #5621, opened 2023-11-11
Add DirectionId and DirectionIdMap

kidder:
## Proposed changes

These are used throughout the subcell code, and I also want to use them in some AMR code.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/Structure/DirectionId.hpp
```diff
@@ -0,0 +1,53 @@
...
+
+#pragma once
+
+#include <cstddef>
+#include <iosfwd>
+
+#include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/ElementId.hpp"
+
+namespace PUP {
```
nilsdeppe: wrap in `/// \cond`

src/Domain/Structure/DirectionId.cpp
```diff
@@ -0,0 +1,75 @@
...
+
+template <size_t VolumeDim>
+size_t hash_value(const DirectionId<VolumeDim>& id) {
+  size_t h = 0;
+  boost::hash_combine(h, id.direction);
+  boost::hash_combine(h, id.id);
+  return h;
+}
+
+// clang-tidy: do not modify namespace std
```
nilsdeppe: can we `NOLINT(SPECIFIC_ERROR)` here?

src/Domain/Structure/DirectionIdMap.hpp
```diff
@@ -0,0 +1,36 @@
...
+
+#include "DataStructures/FixedHashMap.hpp"
+#include "Domain/Structure/DirectionId.hpp"
+#include "Domain/Structure/MaxNumberOfNeighbors.hpp"
+
+/// An optimized map with DirectionId keys
+template <size_t Dim, typename T>
+class DirectionIdMap
+    : public FixedHashMap<maximum_number_of_neighbors(Dim), DirectionId<Dim>, T,
+                          boost::hash<DirectionId<Dim>>> {
```
nilsdeppe: Why not just leave the hasher defaulted? You added a `std::hash` for `DirectionId`

nilsdeppe (changes requested):
Couple minor things. Could you also do a `git grep "boost::hash<DirectionId"`? those should all be replaceable with the defaulted `std::hash`

Edit: I'm happy to have the `boost::hash` removal be a separate commit after these two to make it easier :)

src/Domain/Structure/DirectionId.hpp
```diff
@@ -0,0 +1,55 @@
...
+#include "Domain/Structure/ElementId.hpp"
+
+/// \cond
+namespace PUP {
+class er;
+}  // namespace PUP
+/// \endcond
+
+template <size_t VolumeDim>
+struct DirectionId {
```
nilsvu: Two thoughts: (1) DirectionId sounds like the ID for a direction, which made me wonder why we need an ID for it until I realized it stands for "Direction and neighbor ID". Related: (2) this same type is defined as MortarId in MortarHelpers.hpp. So maybe name it MortarId? It's not always used to identify DG mortars but IMO that's fine. If you don't like it maybe we can give it a thought on the Monday call.

nilsdeppe: Just as a note in case I can't make the meeting for some unexpected reason, we used to have this data structure and called it `DirectionalElementId`.

I don't love the subcell code using `MortarId` since there are no mortars, really. Most of the places that would use it are actually indexing volume data. I worry that the subcell code is already too complicated with my best efforts to use clear names, so using a name like `MortarId` for indexing volume data, volume fluxes, neighbor RDMP data, etc. would make it extra confusing :(

nilsvu: `DirectionalElementId` sounds good enough to me. We could alias `using MortarId = DirectionalElementId` in the `dg` namespace (not requesting this change). Why was the data structure removed (I don't remember it).

nilsdeppe: Sounds good to both! My memory is that at some point we were only using it in a couple places and figured a `std::pair` was easier than having an extra class (turns out whatever group of us made that decision was wrong :P)

kidder: I won't be able to make the meeting tomorrow.  I'll do whatever the consensus is, but  I'll suggest DirectionalId as that is shorter.   I presume you want the same name change for DirectionIdMap as well.

src/Domain/Structure/DirectionalId.hpp
```diff
@@ -16,7 +16,7 @@ class er;
 /// \endcond
 
 template <size_t VolumeDim>
-struct DirectionId {
+struct DirectionalId {
```
nilsdeppe, line 22: sorry I missed this on my last review. Could you document this a bit so it shows up in doxygen?

nilsdeppe (commented):
LGTM, one minor thing you can squash

src/Domain/Structure/DirectionalId.hpp
kidder: done

nilsdeppe (approved):

kidder:
ignoring test timeout

PR #5622, opened 2023-11-11
Minor cleanup of testing framework code

kidder:
## Proposed changes

Small changes aimed at reducing code complexity in order to improve compile time and reduce code coupling.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Utilities/System/CMakeLists.txt
+++ b/src/Utilities/System/CMakeLists.txt
@@ -9,6 +9,7 @@ spectre_target_sources(
   ${LIBRARY}
   PRIVATE
   Abort.cpp
+  Exit.cpp
   ParallelInfo.cpp
   Prefetch.cpp
   )
--- a/src/Utilities/System/Exit.cpp
+++ b/src/Utilities/System/Exit.cpp
@@ -1,0 +1,18 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Utilities/System/Exit.hpp"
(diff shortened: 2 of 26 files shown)


src/Utilities/System/CMakeLists.txt
```diff
@@ -9,6 +9,7 @@ spectre_target_sources(
   ${LIBRARY}
   PRIVATE
   Abort.cpp
+  Exit.cpp
```
nilsvu, line 12: Did you intend to actually make the dependence on Charm `PRIVATE` in this file as the commit message promises? If so I think it would be good to also add a comment there why you made it private so it won't accidentally be made public again later.

nilsvu (changes requested):

src/Utilities/System/CMakeLists.txt
kidder: I did, but I couldn't get it to work because we still have some cyclic library dependencies.  I'll remove the comment from the commit message

kidder:
@nilsvu any other comments or requested changes?

src/Utilities/System/CMakeLists.txt
nilsvu: Do you understand the cyclic dependency? How can making a dependency private introduce a cyclic dependency?

nilsvu (approved):
LGTM

PR #5623, opened 2023-11-11
Fix interpolation framework when there are no valid points to interpolate

kidder:
## Proposed changes

If there were no valid interpolation points (i.e. all points were outside the Domain), the interpolation target was not informed.  This led to a growing list of pending interpolations and uninterpolated volume data for sequential interpolation targets such as the apparent horizon finder.

Now I added a check for this case and send a message to the interpolation target that allows the interpolation target to do whatever it did when there was a mixture of valid and invalid interpolation points.

I also fixed a few minor things I encountered debugging this, as well as added tests that would have failed.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


tests/Unit/ParallelAlgorithms/Interpolation/Test_InterpolationTargetSphere.cpp
```diff
@@ -227,9 +234,12 @@ SPECTRE_TEST_CASE("Unit.NumericalAlgorithms.InterpolationTarget.Sphere",
...
-    test_interpolation_target_sphere(make_not_null(&gen), num_spheres,
-                                     intrp::AngularOrdering::Cce);
-    test_interpolation_target_sphere(make_not_null(&gen), num_spheres,
-                                     intrp::AngularOrdering::Strahlkorper);
+    test_interpolation_target_sphere<InterpTargetTestHelpers::ValidPoints::All>(
+        make_not_null(&gen), num_spheres, intrp::AngularOrdering::Cce);
+    test_interpolation_target_sphere<InterpTargetTestHelpers::ValidPoints::All>(
+        make_not_null(&gen), num_spheres, intrp::AngularOrdering::Strahlkorper);
+    test_interpolation_target_sphere<
+        InterpTargetTestHelpers::ValidPoints::None>(
```
wthrowe, line 242: Should this test have a "Some" check?

tests/Unit/ParallelAlgorithms/Interpolation/Test_ObserveTimeSeriesAndSurfaceData.cpp
```diff
@@ -815,27 +835,44 @@ SPECTRE_TEST_CASE(
...
+  // This check also uses no interpolated data
   check_ylm_data_with_greater_max_l();
 }
+
+SPECTRE_TEST_CASE(
+    "Unit.NumericalAlgorithms.Interpolator.ObserveTimeSeriesAndSurfaceData",
+    "[Unit]") {
+  domain::creators::register_derived_with_charm();
+  run_test<InterpTargetTestHelpers::ValidPoints::All>();
+  run_test<InterpTargetTestHelpers::ValidPoints::None>();
```
wthrowe, line 876: Should this test test "Some"?

wthrowe (changes requested):

tests/Unit/ParallelAlgorithms/Interpolation/Test_InterpolationTargetSphere.cpp
kidder: I can add one, it would be non-trivial as the target spheres are randomly generated.  I might be able to do it with an offset spherical domain.

tests/Unit/ParallelAlgorithms/Interpolation/Test_ObserveTimeSeriesAndSurfaceData.cpp
kidder: This would be more difficult to test as I would need to pass in which points are invalid.  I didn't think it was worth the effort as the only difference between All and Some would be the invalid points print nan which is tested by None.  I can add a comment saying so, or put in the effort for the additional test if you would prefere

wthrowe: You can just add a comment.  Same for the other test I commented on.

wthrowe (dismissed):

wthrowe:
Needs a rebase.

wthrowe (approved):

PR #5627, opened 2023-11-17
Change behavior of sphere shape transition function

knelli2:
## Proposed changes

In SpEC, the shape map is (for centered coordinates)

$$
x^i = \xi^i \left(1 - \frac{f(r, \theta, \phi)}{r} \sum_{\ell,m} \lambda_{\ell m}Y_{\ell m}(\theta, \phi)\right)
$$

However, in SpECTRE, the shape map currently is

$$
x^i = \xi^i \left(1 - h(r, \theta, \phi) \sum_{\ell,m} \lambda_{\ell m}Y_{\ell m}(\theta, \phi)\right)
$$

In SpEC, we have that $0 \leq f(r, \theta, \phi) \leq 1$, while in SpECTRE we have $0 \leq h(r, \theta, \phi) \leq 1$. This slight difference of a factor of $r$ is a fundamentally different coordinate mapping.

The actual solution to this is to change our shape coordinate mapping to match the one in SpEC. All of the control systems in SpECTRE make/use assumptions based on the SpEC formula because I thought the shape map in SpECTRE was implemented exactly the same in SpEC. I will make this actual change in a future PR, because it's more involved.

This PR is simply a band-aid fix in the meantime. We can actually make use of the fact that the current `SphereTransition` transition function (which is $h(r, \theta, \phi)$ ) that is used in SpECTRE is

$$
h(r, \theta, \phi) = \frac{ar+b}{r}
$$

which can be rewritten in the SpEC way as

$$
h(r, \theta, \phi) = \frac{ar+b}{r} = \frac{f(r, \theta, \phi)}{r}
$$

The coefficients $a$ and $b$ in the `SphereTransition` class are chosen so $h$ is 1 at the excision boundary and 0 at the outer radius of the inner shell around the excision. However, if we just tweak the coefficients a little, we can make it so that $f$ is 1 at the excision boundary and falls off to 0 while $h$ is $1/r_{\text{Ex}}$ at the excision boundary and falls off to 0. This then makes the SpECTRE coordinate mapping identical to the SpEC one.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
@@ -26,8 +26,8 @@ SphereTransition::SphereTransition(const double r_min, const double r_max)
         "r_max =  "
         << r_max << ", and r_min = " << r_min);
   }
-  a_ = -r_min / (r_max - r_min);
-  b_ = r_min * r_max / (r_max - r_min);
+  a_ = -1.0 / (r_max - r_min);
+  b_ = -a_ * r_max;
 }
 
 double SphereTransition::operator()(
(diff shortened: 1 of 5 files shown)


wthrowe:
Do you want this reviewed and merged, or do you want to wait on the better fix?

knelli2:
I'd like this reviewed and merged please. I hope to have the better fix in a PR before Thanksgiving, so I imagine it won't get merged until after the holiday. This will just allow runs that @geoffrey4444 and his students are doing to actually have a correct shape map in the time being.

wthrowe (approved):

PR #5636, opened 2023-11-21
Make shape map exactly like the one in SpEC

### LINKED ISSUE #5635: Shape map test fails on CI
# Bug reports:

### Expected behavior:

test passes

### Current behavior:

test fails wtih
```
'radius < 1.' violated!
  Coordinates mapped through the center!
```

### Environment:

https://github.com/sxs-collaboration/spectre/actions/runs/6933538336/job/18859882389

knelli2:
## Proposed changes

This is a follow up to #5627.

In SpEC, the shape map is (for centered coordinates)

$$
x^i = \xi^i \left(1 - \frac{f(r, \theta, \phi)}{r} \sum_{\ell,m} \lambda_{\ell m}Y_{\ell m}(\theta, \phi)\right)
$$

However, in SpECTRE, the shape map currently is

$$
x^i = \xi^i \left(1 - h(r, \theta, \phi) \sum_{\ell,m} \lambda_{\ell m}Y_{\ell m}(\theta, \phi)\right)
$$

In SpEC, we have that $0 \leq f(r, \theta, \phi) \leq 1$, while in SpECTRE we have $0 \leq h(r, \theta, \phi) \leq 1$. This slight difference of a factor of $r$ is a fundamentally different coordinate mapping.

This PR edits the shape map to be identical to the one in SpEC and consequently edits both of the transition functions to be from 1 to 0.

The wedge transition map also has additional edits that fixes some issues it had before.

This PR also fixes #5635.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
@@ -94,7 +94,8 @@ std::array<tt::remove_cvref_wrap_t<T>, 3> Shape::operator()(
 #ifdef SPECTRE_DEBUG
   using ReturnType = tt::remove_cvref_wrap_t<T>;
   const ReturnType shift_radii =
-      distorted_radii * transition_func_->operator()(centered_coords);
+      distorted_radii * transition_func_->operator()(centered_coords) *
+      check_and_compute_one_over_radius(centered_coords);
   if constexpr (std::is_same_v<ReturnType, double>) {
     ASSERT(shift_radii < 1., "Coordinates mapped through the center!");
   } else {
@@ -104,9 +105,11 @@ std::array<tt::remove_cvref_wrap_t<T>, 3> Shape::operator()(
   }
 #endif  // SPECTRE_DEBUG
 
-  return center_ + centered_coords *
-                       (1. - distorted_radii *
-                                 transition_func_->operator()(centered_coords));
+  return center_ +
+         centered_coords *
+             (1. - distorted_radii *
+                       transition_func_->operator()(centered_coords) *
+                       check_and_compute_one_over_radius(centered_coords));
 }
 
 std::optional<std::array<double, 3>> Shape::inverse(
@@ -145,7 +148,8 @@ std::array<tt::remove_cvref_wrap_t<T>, 3> Shape::frame_velocity(
   ylm_.interpolate_from_coefs(make_not_null(&radii_velocities), coef_derivs,
                               interpolation_info);
   return -centered_coords * radii_velocities *
-         transition_func_->operator()(centered_coords);
+         transition_func_->operator()(centered_coords) *
+         check_and_compute_one_over_radius(centered_coords);
 }
 
 template <typename T>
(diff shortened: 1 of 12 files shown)


src/Domain/CoordinateMaps/TimeDependent/Shape.hpp
```diff
@@ -93,49 +86,95 @@ namespace domain::CoordinateMaps::TimeDependent {
...
+ *
+ * \f{equation}{
+ * \frac{r}{\tilde{r}} =
+ * \frac{1}{1-f(r,\theta,\phi)\sum\lambda_{lm}(t)Y_{lm}(\theta,\phi) / r}
+ * \f}
+ *
+ * For more details, see
+ * \link
+ * domain::CoordinateMaps::ShapeMapTransitionFunctions::ShapeMapTransitionFunction::original_radius_over_radius
+ * ShapeMapTransitionFunction::original_radius_over_radius \endlink.
```
knelli2: Wasn't sure what to do about this. Even without the `\link` it would be over 80 characters.

wthrowe: This looks like a good case of a long line.  There's a whitelist in `tools/FileTestDefs.sh` you can add to.  You can either add this file or whitelist `\link` in general.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp
```diff
@@ -89,15 +89,17 @@ class ShapeMapTransitionFunction : public PUP::able {
       double distorted_radius) const = 0;
   /*!
    * Evaluate the transition function at the Cartesian coordinates divided by
-   * the radius. All divisions over the radius which could potentially be zero
-   * divisions are left to the transition function which uses its knowledge
-   * of the domain to do this safely or throw an appropriate error.
+   * the radius to a power. All divisions over the radius which could
+   * potentially be zero divisions are left to the transition function which
+   * uses its knowledge of the domain to do this safely or throw an appropriate
+   * error.
```
wthrowe: The general feature isn't new, but it's being significantly expanded upon so I want to understand it.

What is this for?  Looking at the math, the place a division-by-zero could potentially be analytically handled is in the shape map, not the transition function (and I don't think it actually gives a usable map even then).  All the implementations are just dividing by r, without any of the checks mentioned here.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -179,19 +173,27 @@ T Wedge::call_impl(const std::array<T, 3>& source_coords) const {
...
-  return (outer_distance - radius) /
-         ((outer_distance - inner_surface_.distance(rotated_coords)) * radius);
+  if (power == 1) {
+    return (outer_distance - radius) /
+           ((outer_distance - inner_surface_.distance(rotated_coords)) *
+            radius);
+  } else {
+    return (outer_distance - radius) /
+           ((outer_distance - inner_surface_.distance(rotated_coords))) *
+           integer_pow(1.0 / radius, power);
```
wthrowe: You're already doing a division.  Put the new factors inside the denominator instead of doing another.  (Particularly since I'm not sure whether `integer_pow(1.0 / radius, 3)` does one division or three.)

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -244,51 +245,90 @@ tnsr::Ij<tt::remove_cvref_wrap_t<T>, 3, Frame::NoFrame> Shape::jacobian(
...
-  auto& transition_func = get<1>(theta_phis);
-  transition_func = transition_func_->operator()(centered_coords);
+  auto& transition_func_over_square_radius = get<1>(theta_phis);
+
   // leave it to the domain specific transition function to divide by the radii
   // as it can safely do this
-  const auto transition_func_over_radius =
-      transition_func_->map_over_radius(centered_coords);
+  transition_func_over_square_radius =
+      transition_func_->map_over_radius(centered_coords, 2);
```
wthrowe: This re-use is not an optimization.  It doesn't avoid an allocation and may prevent some form of copy elision.

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -244,51 +245,90 @@ tnsr::Ij<tt::remove_cvref_wrap_t<T>, 3, Frame::NoFrame> Shape::jacobian(
...
+  auto& target_gradient_z_times_spatial_part = target_gradient_z;
+  target_gradient_x_times_spatial_part *= transition_func_over_square_radius;
+  target_gradient_y_times_spatial_part *= transition_func_over_square_radius;
+  target_gradient_z_times_spatial_part *= transition_func_over_square_radius;
+
+  // re-use allocation
+  auto& transition_func_over_cube_radius = get<1>(theta_phis);
+
+  transition_func_over_cube_radius =
+      transition_func_->map_over_radius(centered_coords, 3);
```
wthrowe: Again, not an optimization.

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -244,51 +245,90 @@ tnsr::Ij<tt::remove_cvref_wrap_t<T>, 3, Frame::NoFrame> Shape::jacobian(
...
-                     target_gradient_z * transition_func_over_radius);
+      -z_centered * ((z_transition_gradient_over_radius -
+                      z_centered * transition_func_over_cube_radius) *
+                         distorted_radii +
+                     target_gradient_z_times_spatial_part);
+
+  // re-use allocation
+  auto& transition_func_over_radius = get<1>(theta_phis);
+  transition_func_over_radius =
+      transition_func_->map_over_radius(centered_coords);
```
wthrowe: Again, not an optimization.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -25,22 +25,36 @@
...
-          sphericity);
+  if (axis.has_value()) {
+#ifdef SPECTRE_DEBUG
+    for (size_t i = 0; i < get_size(coords[0]); i++) {
+      if (get_element(gsl::at(coords, axis.value()), i) == 0.0) {
+        ERROR(
+            "The Wedge transition map was called with coordinates outside of "
+            "its wedge. The "
+            << (axis.value() == 0_st ? "x" : (axis.value() == 1_st ? "y" : "z"))
+            << "-coordinate shouldn't be 0 but it is.");
```
wthrowe: Print the coordinate or the index or something.

wthrowe (changes requested):

src/Domain/CoordinateMaps/TimeDependent/Shape.hpp
knelli2: Whitelisting `\link` in general seems like the best approach and is helpful in the future as well.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp
knelli2: Yes, there is one place in the shape map where dividing by r could be taken care of, but my best guess as to why the transition function is responsible for checking the `1/r` is because the transition function could be anything and so it may also need to account for a `1/r` somewhere itself. And, I think, because the only place that `1/r` shows up is with the transition function itself, it made sense to just have the transition function do all of the checks.

Both the `Wedge` and `Sphere` transition functions do check the magnitude of the point (in debug mode) before dividing by it. So these checks do exist.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
knelli2: I changed this in the commit after, (no clue why) but I'll move it to the shape map edit commit

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp
wthrowe: You're right that I missed the checks in the transition functions, but why not move them to the shape map itself to avoid duplication?  The shape map isn't valid at r=0, so there can't be any issues with dividing by r.

knelli2: Ok yeah, I'm convinced this should be in the shape map now. If the transition function itself needs to divide by r it can just assume it's not zero since the shape map took care of it.

wthrowe: If we want to consider using the map in the opposite sense, to remove distortion when going inwards, then you could get a valid map at the origin where I think you might want to remove cancellations in the transition function, but I still don't think you'd want to do it with varying powers in the fetch functions.  You'd want the interface to only provide the valid quantities.

knelli2:
@wthrowe Pushed some fixups. I'll make the whitelist commit the first commit when I squash. I'll also squash the new changes into the appropriate commits as well since I edited the Wedge transition in both original commits.

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -375,6 +377,24 @@ void Shape::check_size(const gsl::not_null<DataVector*>& coefs,
   }
 }
 
+template <typename T>
+T Shape::check_and_compute_one_over_radius(
+    const std::array<T, 3>& centered_coords) const {
+#ifdef SPECTRE_DEBUG
+  const T radius = magnitude(centered_coords);
```
wthrowe, line 383: [optional] Move up one line and return `1.0 / radius`.

wthrowe (commented):
Looks good.  Squash (and fix clang-tidy).

wthrowe (dismissed):
Looks good.  GitHub Actions is having problems (in general, not just for us) at the moment, but they usually resolve that sort of thing quickly.

wthrowe (approved):

PR #5640, opened 2023-11-27
WorldtubeBufferUpdater and WorldtubeDataManager for CCE

Sizheng-Ma:
## Proposed changes

This PR aims to pave the path for dealing with the worldtube data of the Klein-Gordon CCE system. Two major changes are introduced to the current infrastructure

1. The current `WorldtubeDataManager` is hard-coded to handle `Tags::characteristic_worldtube_boundary_tags<Tags::BoundaryValue>`. This can be relaxed so that it can be used to deal with other worldtube data (say a scalar field).
2. There are repeated contents in `WorldtubeBufferUpdater` and `WorldtubeDataManager`, and will be more once they are extended to handle the Klein-Gordon system. I extract some common parts into functions to avoid more copy and paste.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/WorldtubeBufferUpdater.hpp
+++ b/src/Evolution/Systems/Cce/WorldtubeBufferUpdater.hpp
@@ -81,6 +81,22 @@ std::string dataset_name_for_component(std::string base_name,
 std::pair<size_t, size_t> create_span_for_time_value(
     double time, size_t pad, size_t interpolator_length, size_t lower_bound,
     size_t upper_bound, const DataVector& time_buffer);
+
+// retrieves the extraction radius from the specified file.
+// We assume that the filename has the extraction radius encoded as an
+// integer between the last occurrence of 'R' and the last occurrence of
+// '.'. This is the format provided by SpEC.
+std::string get_text_radius(const std::string& cce_data_filename);
+
+// retrieves time stamps and lmax the from the specified file.
+void set_time_buffer_and_lmax(gsl::not_null<DataVector*> time_buffer,
+                              size_t& l_max, const h5::Dat& data);
+
+// retrieves modal data from Bondi or Klein-Gordon worldtube H5 file.
+void update_buffer_with_modal_data(
+    gsl::not_null<ComplexModalVector*> buffer_to_update,
+    const h5::Dat& read_data, size_t computation_l_max, size_t l_max,
+    size_t time_span_start, size_t time_span_end, bool is_real);
 }  // namespace detail
 
 /// the full set of tensors to be extracted from the worldtube h5 file
--- a/src/Evolution/Systems/Cce/WorldtubeDataManager.cpp
+++ b/src/Evolution/Systems/Cce/WorldtubeDataManager.cpp
@@ -234,8 +234,9 @@ bool MetricWorldtubeDataManager::populate_hypersurface_boundary_data(
   return true;
 }
 
-std::unique_ptr<WorldtubeDataManager> MetricWorldtubeDataManager::get_clone()
-    const {
+std::unique_ptr<WorldtubeDataManager<
+    Tags::characteristic_worldtube_boundary_tags<Tags::BoundaryValue>>>
+MetricWorldtubeDataManager::get_clone() const {
   return std::make_unique<MetricWorldtubeDataManager>(
       buffer_updater_->get_clone(), l_max_, buffer_depth_,
       interpolator_->get_clone(), fix_spec_normalization_);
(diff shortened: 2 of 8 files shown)


Sizheng-Ma:
I've tested the executable `CharactersiticExtract` with a BBH system. All the waveform outputs are not impacted by the changes.

src/Evolution/Systems/Cce/WorldtubeBufferUpdater.hpp
```diff
@@ -91,6 +91,12 @@ std::string get_text_radius(const std::string& cce_data_filename);
 // retrieves time stamps and lmax the from the specified file.
 void get_time_buffer_and_lmax(DataVector& time_buffer, size_t& l_max,
                               const h5::Dat& data);
+
+// retrieves modal data from Bondi or Klein-Gordon worldtube H5 file.
+void get_modal_data_for_bondi_and_klein_gordon(
```
knelli2: We don't want to have to keep renaming this function every time a new CCE system is added. Maybe `update_buffer_with_modal_data`?

src/Evolution/Systems/Cce/WorldtubeBufferUpdater.hpp
```diff
@@ -81,6 +81,16 @@ std::string dataset_name_for_component(std::string base_name,
...
+
+// retrieves the extraction radius from the specified file.
+// We assume that the filename has the extraction radius encoded as an
+// integer between the last occurrence of 'R' and the last occurrence of
+// '.'. This is the format provided by SpEC.
+std::string get_text_radius(const std::string& cce_data_filename);
+
+// retrieves time stamps and lmax the from the specified file.
+void get_time_buffer_and_lmax(DataVector& time_buffer, size_t& l_max,
+                              const h5::Dat& data);
```
knelli2: This should probably be `set_time_buffer_and_lmax` since this function doesn't return anything. (The other function `get_text_radius` is fine because it actually returns something)

src/Evolution/Systems/Cce/WorldtubeDataManager.hpp
```diff
@@ -48,6 +48,7 @@ class BondiWorldtubeDataManager;
  *   underlying data source. This is primarily used for monitoring the frequency
  *   and size of the buffer updates.
  */
+template <typename BoundaryTags>
 class WorldtubeDataManager : public PUP::able {
```
knelli2, line 69: Add a line in the dox that `BoundaryTags` is a `tmpl::list` of tags on the worldtube boundary

src/Evolution/Systems/Cce/WorldtubeDataManager.hpp
```diff
@@ -20,6 +20,20 @@
 
 namespace Cce {
 
+namespace detail {
+
+template <typename InputTags>
+void puper(size_t& time_span_start, size_t& time_span_end,
```
knelli2: Rename to `set_non_pupped_members`. Also since this doesn't actually use the pupper, take all the reference arguments as `gsl::not_null`.

src/Evolution/Systems/Cce/WorldtubeDataManager.hpp
```diff
@@ -20,6 +20,20 @@
...
+           Variables<InputTags>& coefficients_buffers,
+           Variables<InputTags>& interpolated_coefficients, size_t buffer_depth,
+           size_t interpolator_length, size_t l_max);
+
+template <typename InputTags>
+void constructor(size_t& buffer_depth,
```
knelli2: Rename to `initialize_buffers`. Also take the reference arguments as `gsl::not_null`

knelli2 (changes requested):

Sizheng-Ma:
Done!

src/Evolution/Systems/Cce/WorldtubeDataManager.cpp
```diff
@@ -28,17 +28,19 @@ namespace Cce {
...
-           Variables<InputTags>& interpolated_coefficients,
-           const size_t buffer_depth, const size_t interpolator_length,
-           const size_t l_max) {
-  time_span_start = 0;
-  time_span_end = 0;
+void set_non_pupped_members(
+    gsl::not_null<size_t*> time_span_start,
+    gsl::not_null<size_t*> time_span_end,
+    gsl::not_null<Variables<InputTags>*> coefficients_buffers,
+    gsl::not_null<Variables<InputTags>*> interpolated_coefficients,
```
knelli2: Make these `const gsl::not_null` here in the cpp file only.

src/Evolution/Systems/Cce/WorldtubeDataManager.cpp
```diff
@@ -45,10 +45,11 @@ void set_non_pupped_members(
 }
 
 template <typename InputTags>
-void constructor(size_t& buffer_depth,
-                 Variables<InputTags>& coefficients_buffers,
-                 const size_t buffer_size, const size_t interpolator_length,
-                 const size_t l_max) {
+void initialize_buffers(
+    gsl::not_null<size_t*> buffer_depth,
+    gsl::not_null<Variables<InputTags>*> coefficients_buffers,
```
knelli2: Same here, `const gsl::not_null` only in cpp.

knelli2 (changes requested):
Looks good. Squash all the fixups including these last few changes.

Sizheng-Ma:
Done. I noticed that the first argument of `set_time_buffer_and_lmax` was also a reference. I switched it to `gsl::not_null` in the last commit.

knelli2 (commented):
Nice catch. You can squash the final fixup

Sizheng-Ma:
Ok, squashed.

knelli2 (approved):

PR #5641, opened 2023-11-27
Add option for runtime element distribution

knelli2:
## Proposed changes

Now we can choose how we distribute elements at runtime so you can experiment with what is best for your runs. Currently all the options are different weights for a space-filling curve except for a round-robin distribution. The distributions I added to each of the input files is exactly what it was before, so in practice the distribution should be the same as before.

### Upgrade instructions

Evolution and elliptic executables must now have a new option in the input file called
```yaml
Parallelization:
  ElementDistribution: NumGridPoints
```

An exception to this are the `*CharacteristicExtract` executables. They do not need this new option.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Tags/ElementDistribution.hpp
+++ b/src/Domain/Tags/ElementDistribution.hpp
@@ -1,0 +1,48 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <memory>
+#include <optional>
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/ElementDistribution.hpp"
+#include "Options/Auto.hpp"
+#include "Options/String.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+namespace Parallel::OptionTags {
+struct Parallelization;
+}  // namespace Parallel::OptionTags
+/// \endcond
+
+namespace domain {
+namespace OptionTags {
+/// \ingroup OptionTagsGroup
+/// \ingroup ComputationalDomainGroup
+struct ElementDistribution {
+  struct RoundRobin {};
+  using type = Options::Auto<ElementWeight, RoundRobin>;
+  static constexpr Options::String help = {
+      "Weighting pattern to use for ZCurve element distribution. Specify "
+      "RoundRobin to just place each element on the next core."};
+  using group = Parallel::OptionTags::Parallelization;
+};
+}  // namespace OptionTags
+
+namespace Tags {
+struct ElementDistribution : db::SimpleTag {
+  using type = std::optional<ElementWeight>;
+  using option_tags = tmpl::list<OptionTags::ElementDistribution>;
+
+  static constexpr bool pass_metavariables = false;
+  static type create_from_options(const type& element_distribution) {
+    return element_distribution;
+  }
+};
+}  // namespace Tags
+}  // namespace domain
--- a/src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
+++ b/src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
@@ -23,6 +23,8 @@
 #include "Domain/Structure/ElementId.hpp"
 #include "Domain/Structure/InitialElementIds.hpp"
 #include "Domain/Tags.hpp"
+#include "Domain/Tags/ElementDistribution.hpp"
+#include "Elliptic/DiscontinuousGalerkin/Tags.hpp"
 #include "Parallel/Algorithms/AlgorithmArray.hpp"
 #include "Parallel/GlobalCache.hpp"
 #include "Parallel/Info.hpp"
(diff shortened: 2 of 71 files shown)


src/Elliptic/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -83,39 +87,67 @@ struct DefaultElementsAllocator
...
-            .insert(global_cache, initialization_items, target_proc);
-
-        const size_t target_node =
-            Parallel::node_of<size_t>(target_proc, local_cache);
-        ++elements_per_core[target_proc];
-        ++elements_per_node[target_node];
-        grid_points_per_core[target_proc] += grid_points_per_element;
-        grid_points_per_node[target_node] += grid_points_per_element;
+      if (element_weight.has_value()) {
+        for (const auto& element_id : element_ids) {
```
nilsdeppe, line 122: Maybe add a comment here and in the else to the effect of `// Distribute by estimated cost of the element. E.g., a weighted space-filling curve.` and then `// Do a round-robin distribution of the elements`?

src/Evolution/DiscontinuousGalerkin/DgElementArray.hpp
```diff
@@ -110,33 +112,26 @@ void DgElementArray<Metavariables, PhaseDepActionList>::allocate_array(
...
-      domain::get_element_costs(
-          blocks, initial_refinement_levels, initial_extents,
-          local_time_stepping
-              ? domain::ElementWeight::NumGridPointsAndGridSpacing
-              : domain::ElementWeight::NumGridPoints,
-          quadrature);
-  const domain::BlockZCurveProcDistribution<volume_dim> element_distribution{
-      element_costs,   num_of_procs_to_use, blocks, initial_refinement_levels,
-      initial_extents, procs_to_ignore};
+  // Only need this if the element weight has a value
```
nilsdeppe: `this` -> `element_distribution` (so that when someone accidentally moves the comment or code, at least we have some hope of putting them back together :D )

Maybe expand comment `... since then we have to distribute the elements in a more complicated way than just round robin`?

src/ParallelAlgorithms/LinearSolver/Multigrid/ElementsAllocator.hpp
```diff
@@ -167,19 +177,34 @@ struct ElementsAllocator
...
-          element_costs,   num_of_procs_to_use,
-          blocks,          initial_refinement_levels,
-          initial_extents, procs_to_ignore};
-      for (const auto& element_id : element_ids) {
-        const size_t target_proc =
-            element_distribution.get_proc_for_element(element_id);
-        element_array(element_id)
-            .insert(global_cache, initialization_items, target_proc);
+      if (element_weight.has_value()) {
+        const std::unordered_map<ElementId<Dim>, double> element_costs =
```
nilsdeppe, line 182: Please add same comments here that you do to the other file

src/Domain/Tags/ElementDistribution.hpp
```diff
@@ -0,0 +1,41 @@
...
+#include "Domain/ElementDistribution.hpp"
+#include "Options/Auto.hpp"
+#include "Options/String.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace domain {
+namespace OptionTags {
+/// \ingroup OptionTagsGroup
+/// \ingroup ComputationalDomainGroup
+struct ElementDistribution {
```
nilsdeppe, line 27: Should this go into the `ResourceInfo` group? I can't decide whether or not I like that. I don't love having a bunch of lowest-level options if possible :Shrug:

nilsdeppe (changes requested):
LGTM, a few small suggestions for comments and clarifications. Please squash any changes immediately

src/Domain/Tags/ElementDistribution.hpp
knelli2: `ResourceInfo` isn't actually an options group, it's just a struct. That being said, I can put it into the struct and just add a member function for accessing it if you think that's better. The `ResourceInfo` is available in the initialization items, so I don't have strong opinions where we place the option

nilsdeppe: Ah okay. Hmm, yea, if it's easy to do maybe it's worth doing? My only hesitation is then if we have an executable that uses ResourceInfo but no elements things will be confusing. In a sense we probably want something like a `Parallelization` group, but that seems beyond the scope of this PR and should probably have input from the other core devs :shrug: So I will let you make a decision on what you think is best and we'll just go with that :)

knelli2: Oh actually, this is true for the CCE executable. It doesn't have a `DgElementArray`, so with the current setup you never specify an element distribution. But if it was in the `ResourceInfo`, then you'd have to, but it'd be a useless option.

I like the idea of a `Parallelization` group. I'll put the element distribution in there, and we can add to it later with whatever else we come up with.

I guess we could technically put the `ResourceInfo` into `Parallelization`, but I think that's out of the scope of this PR.

nilsdeppe: That sounds like a good plan to me!

nilsdeppe (commented):
LGTM, please squash!

knelli2:
@nilsdeppe Squashed. @nilsvu could you take a look at the elliptic parts that I edited? I had to change more things than expected so I'd like your sign off

nilsvu (approved):
Looks good 👍 Thanks for adding this.

nilsdeppe (approved):

Issue #5472, opened 2023-09-18
Reduce memory usage of compiling executables


nilsvu:
Compiling executables needs a lot of RAM. Currently building the BBH executable with gcc 11 in Release mode on CI exceeds the 7GB of RAM available on the Github-hosted runners and fails. Also building in parallel often exceeds the available RAM on clusters and personal machines. Therefore, it would be very helpful to reduce the memory usage of our builds.

nilsvu:
Here's a recent profile I did with [clang's build analyzer](https://github.com/aras-p/ClangBuildAnalyzer): [profile.txt](https://github.com/sxs-collaboration/spectre/files/12653341/profile.txt)

kidder:
Here are the top offenders in compile-time and memory usage for a release build using gcc 13.1 on my desktop (with DEBUG_SYMBOLS=OFF, BUILD_SHARED_LIBS=ON) :
```
302.120 sec Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.cpp
297.600 sec Evolution/Executables/GrMhd/GhValenciaDivClean/EvolveGhValenciaDivCleanBondiMichel.cpp
295.980 sec Evolution/Executables/GrMhd/GhValenciaDivClean/EvolveGhValenciaDivCleanTovStar.cpp
261.480 sec Evolution/Executables/GrMhd/GhValenciaDivClean/EvolveGhValenciaDivCleanBns.cpp
244.160 sec Evolution/Executables/GeneralizedHarmonic/EvolveGhSingleBlackHole.cpp
241.690 sec src/Evolution/DgSubcell/Matrices.cpp
232.310 sec Elliptic/Executables/Xcts/SolveXcts.cpp
229.270 sec tests/Unit/ControlSystem/Systems/Test_RotScaleTrans.cpp
228.700 sec Evolution/Executables/ScalarTensor/EvolveScalarTensorSingleBlackHole.cpp
216.450 sec tests/Unit/DataStructures/Tensor/Expressions/Test_EvaluateRank3Symmetric.cpp
201.240 sec tests/Unit/DataStructures/Tensor/Expressions/Test_EvaluateRank3NonSymmetric.cpp
190.220 sec Evolution/Executables/GrMhd/ValenciaDivClean/EvolveValenciaDivCleanFishboneMoncriefDisk.cpp
[...]

PR #5647, opened 2023-12-04
Reduce compilation memory of executables

macedo22:
Addresses #5472

## Proposed changes

This PR includes several changes that were found to decrease peak RAM usage during compilation of `EvolveGhBinaryBlackHole`. See **Further comments** below for details on the compilation impact of each commit and overall.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments
**Effects of changes on compiling `EvolveGhBinaryBlackHole`:**
The original commits that generated the stats below are backed up on [this branch](https://github.com/macedo22/spectre/tree/bbh_compilation_improvements_1_og_pr), where **clang-13 Release** and **gcc-9 Release** were used to compile on wheeler014.

Each commit below is applied on top of the one above it, so the relative effect of one commit applied on top of the ones before it is the difference from the commit above it. For example, applying `32a2f1d` after `9769906` saved an additional 0.2 GB of RAM for clang and 1.54 GB for gcc.

```
Modification                                    | Commit hash | clang peak RAM (GB) | gcc peak RAM (GB) | clang user time (mm:ss) | gcc user time (mm:ss)
---------------------------------------------------------------------------------------------------------------------------------------------------------
(baseline)                                        b479ee1       11.36                 19.00                8:32                     13:50
Factor ObserveNorms::operator()                   9769906       11.15                 18.84                8:07                     13:22
Rework Databox::reset_compute_items_after_mutate  32a2f1d       10.95                 17.30                8:05                     13:37
Factor get_template_parameters_as_string          8bb55a9       10.85                 17.14                7:52                     13:38
Rework Options help text printing                 3448d08       10.69                 16.38                7:49                     13:21
Factor Options::Parser::parse                     3ad5784        9.98                 15.60                6:55                     11:04
---------------------------------------------------------------------------------------------------------------------------------------------------------
Total saved from baseline commit                                 1.38                  3.40                1:37                      2:46
```

**Effects of changes on CI:**
Some build times and the sizes of the build directories improved a little. See effects on CI recorded [here](https://docs.google.com/spreadsheets/d/1bDai4-JoYYlcjl3hmTEDeO5DF6p1GnmdyAKcXmoiM2o/edit#gid=1692422746).

### DIFF
--- a/src/DataStructures/DataBox/DataBox.hpp
+++ b/src/DataStructures/DataBox/DataBox.hpp
@@ -651,10 +651,11 @@ db::DataBox<tmpl::list<Tags...>>::reset_compute_items_after_mutate(
   using current_tags_to_reset = tmpl::list<TagsOfImmutableItemsToReset...>;
   using next_compute_tags_to_reset = tmpl::list_difference<
       tmpl::remove_duplicates<tmpl::transform<
-          tmpl::append<
-              tmpl::filter<typename DataBox<tmpl::list<Tags...>>::edge_list,
-                           std::is_same<tmpl::pin<TagsOfImmutableItemsToReset>,
-                                        tmpl::get_source<tmpl::_1>>>...>,
+          tmpl::filter<
+              typename DataBox<tmpl::list<Tags...>>::edge_list,
+              tmpl::lazy::list_contains<
+                  tmpl::pin<tmpl::list<TagsOfImmutableItemsToReset...>>,
+                  tmpl::get_source<tmpl::_1>>>,
           tmpl::get_destination<tmpl::_1>>>,
       current_tags_to_reset>;
   reset_compute_items_after_mutate(next_compute_tags_to_reset{});
--- a/src/ParallelAlgorithms/Events/ObserveNorms.hpp
+++ b/src/ParallelAlgorithms/Events/ObserveNorms.hpp
@@ -4,7 +4,6 @@
 #pragma once
 
 #include <cstddef>
-#include <limits>
 #include <optional>
 #include <pup.h>
 #include <string>
@@ -38,7 +37,6 @@
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/Functional.hpp"
-#include "Utilities/Numeric.hpp"
 #include "Utilities/OptionalHelpers.hpp"
 #include "Utilities/Serialization/CharmPupable.hpp"
 #include "Utilities/TMPL.hpp"
@@ -217,6 +215,17 @@ class ObserveNorms<tmpl::list<ObservableTensorTags...>,
   using return_tags = tmpl::list<>;
   using argument_tags = tmpl::list<::Tags::ObservationBox>;
 
+  template <typename TensorToObserveTag, typename ComputeTagsList,
+            typename DataBoxType, size_t Dim>
+  void observe_norms_impl(
+      gsl::not_null<
+          std::unordered_map<std::string, std::pair<std::vector<double>,
+                                                    std::vector<std::string>>>*>
+          norm_values_and_names,
+      const ObservationBox<ComputeTagsList, DataBoxType>& box,
+      const Mesh<Dim>& mesh, const DataVector& det_jacobian,
+      size_t number_of_points) const;
+
   template <typename ComputeTagsList, typename DataBoxType,
             typename Metavariables, size_t VolumeDim,
             typename ParallelComponent>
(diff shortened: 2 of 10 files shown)


src/DataStructures/DataBox/DataBox.hpp
```diff
@@ -650,10 +650,11 @@ db::DataBox<tmpl::list<Tags...>>::reset_compute_items_after_mutate(
   using current_tags_to_reset = tmpl::list<TagsOfImmutableItemsToReset...>;
   using next_compute_tags_to_reset = tmpl::list_difference<
       tmpl::remove_duplicates<tmpl::transform<
-          tmpl::append<
-              tmpl::filter<typename DataBox<tmpl::list<Tags...>>::edge_list,
-                           std::is_same<tmpl::pin<TagsOfImmutableItemsToReset>,
-                                        tmpl::get_source<tmpl::_1>>>...>,
+          tmpl::filter<
+              typename DataBox<tmpl::list<Tags...>>::edge_list,
+              tmpl::bind<tmpl::list_contains,
```
wthrowe: [optional] I recently added a `tmpl::lazy::list_contains`, if you want to avoid the `tmpl::bind`.  I don't know whether bind has a performance impact.

wthrowe (changes requested):
Very cool.  I would not have guessed most of that would matter.  (The DataBox one makes sense in hindsight, at least.)

My only suggested change is optional, so if you don't want to make it let me know and I'll approve.

knelli2 (dismissed):
Also surprised how much moving the `PARSE_ERROR`s from `ParseOptions.hpp` to the cpp makes a difference.

src/DataStructures/DataBox/DataBox.hpp
macedo22: I compiled with this change it seemed to be the same or ever so slightly less RAM (but the difference could just be noise), so I just threw it in

macedo22:
Applied suggested fixup and rebased!

knelli2 (dismissed):
Could you also edit the title of the PR? Your changes here will actually benefit all of our executables, not just the BBH one.

src/ParallelAlgorithms/Events/ObserveNorms.hpp
```diff
@@ -321,6 +330,55 @@ ObserveNorms<tmpl::list<ObservableTensorTags...>,
...
+            std::string,
+            std::pair<std::vector<double>, std::vector<std::string>>>*>
+            norm_values_and_names,
+        const ObservationBox<ComputeTagsList, DataBoxType>& box,
+        const Mesh<Dim>& mesh, const DataVector& det_jacobian,
+        const size_t number_of_points) const {
+  const std::string tensor_name = db::tag_name<TensorToObserveTag>();
+  // Loop over ObservableTensorTags and see if it was requested to be observed.
+  // This approach allows us to delay evaluating any compute tags until they're
+  // actually needed for observing.
```
nilsvu: This comment should remain at the `ObservableTensorTags` pack expansion below.

Is there actually any benefit in factoring out this `impl` function? It's still templated on the whole box and on the `ObservableTensorTags`. If it makes no compile-time difference then I'd keep this code inlined in the `tmpl::for_each` lambda function below to reduce complexity.

wthrowe (dismissed):

nilsvu (commented):
About the changes in CI runtime that you mentioned in the PR description: Keep in mind that CI runtimes depend heavily on caches from previous runs. To measure it accurately you could (1) delete all your GitHub Actions caches on your fork, (2) push current develop to a new branch (not called "develop"), (3) push your changes to a new branch. CI will run on both new branches without using caches. (I'm not requesting you do this, just suggesting a method to measure the CI runtime accurately)

src/ParallelAlgorithms/Events/ObserveNorms.hpp
macedo22: Regarding the lambda, I pulled it out because it was a big part of the memory savings for the `ObserveNorms` changes relative to its total savings, though the total savings weren't huge. With the current change I have (pulling out the `tmpl::for_each` into a function AND factoring out `check_norm_is_observable` and `fill_norm_values_and_names`), it saved 0.21 GB and 24.5s for clang-13 Release and 0.17GB and 27.9s for gcc-9 Release. Now if I put the `tmpl::for_each` back in (but still keep `check_norm_is_observable` and `fill_norm_values_and_names` factored out), it adds back 0.08GB (of 0.21GB) and 13s (of 24.5s) for clang and 0.12 GB (of 0.17GB) and 14s (of 27.9s) for gcc. (note: The timing may have some noise, I didn't run it multiple times to get an average)

Let me know how you feel about the tradeoff, but I will move the comment back either way!

nilsdeppe:
Please don't spend time trying to measure CI. That's a lot of effort for specific numbers that I don't think we need. The results are quite clear: this makes major improvements. I think the effort is better spent getting this merged and working on more improvements.

src/ParallelAlgorithms/Events/ObserveNorms.hpp
nilsdeppe: Keep as-is. the `tmpl::for_each` would be a 50%+ regression, which is huge.

nilsvu: Yep. Sounds like `tmpl::for_each` is a lot more expensive than a parameter pack expansion. Might be an avenue for more optimizations in other places.

nilsvu: If you want to merge this now, go ahead and move the comment back to the right place in a follow-up.

macedo22: Sounds good. I moved the comment back down

wthrowe (approved):

knelli2 (approved):

nilsvu (approved):

PR #5651, opened 2023-12-05
Add serialization-versioning test utility

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


kidder:
@wthrowe I can reproduce the CI failure on my MacBook; if you're around tomorrow afternoon we can try and figure out what is going on...

kidder:
@nilvu @nilsdeppe do either of you want to review this?

nilsdeppe:
I can take a look, but if others are happy with it I'm fine to let others lead the code review :)

tests/Unit/Domain/FunctionsOfTime/Test_PiecewisePolynomial.cpp
```diff
@@ -194,52 +195,9 @@ void test_serialization_versioning() {
...
-          dynamic_cast<const Poly&>(*deserialized));
-  }
-  {
-    std::unique_ptr<FunctionsOfTime::FunctionOfTime> deserialized{};
-    deserialize(make_not_null(&deserialized), serialization_v3.data());
-    CHECK(dynamic_cast<const Poly&>(*poly) ==
-          dynamic_cast<const Poly&>(*deserialized));
-  }
+  TestHelpers::serialization::test_versioning<Poly>(
+      "Domain/FunctionsOfTime/PiecewisePolynomial.serializations", "version 3",
```
nilsvu, line 199: Can you keep the test for v2 around as well?

tests/Unit/Domain/FunctionsOfTime/Test_QuaternionFunctionOfTime.cpp
```diff
@@ -28,132 +29,9 @@ void test_serialization_versioning() {
...
-          dynamic_cast<const QuatFoT&>(*deserialized));
-  }
-  {
-    std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime> deserialized{};
-    deserialize(make_not_null(&deserialized), serialization_v4.data());
-    CHECK(dynamic_cast<const QuatFoT&>(*func) ==
-          dynamic_cast<const QuatFoT&>(*deserialized));
-  }
+  TestHelpers::serialization::test_versioning<QuatFoT>(
+      "Domain/FunctionsOfTime/QuaternionFunctionOfTime.serializations",
```
nilsvu, line 33: Same question about the other versions

tests/Unit/Helpers/Utilities/Serialization/Versioning.hpp
```diff
@@ -0,0 +1,140 @@
...
+/// Old serializations of the class are stored in \p filename
+/// (relative to the unit test source directory).  Each stored version
+/// has a label, where the current label must match \p current_label.
+/// This function checks that
+///
+/// * The last entry matches \p current_label and matches the result
+///   of serializing \p current_object.
+///
+/// * Old lines in the file are successfully deserialized and give
+///   expected values.
```
nilsvu, line 62: Ah I see how this works now. While an interesting idea, I suggest to simplify this function to just deserialize the data with the given label and check equality with the given object. That makes the unit tests easier to understand, gives the flexibility to test variations of objects, and allows handling of newer options. The `.serializations` file is just a dictionary of serializations then.

nilsvu (changes requested):

tests/Unit/Domain/FunctionsOfTime/Test_PiecewisePolynomial.cpp
wthrowe: ?  It's there.

tests/Unit/Helpers/Utilities/Serialization/Versioning.hpp
wthrowe: I think the current design allows all that, and additionally ensures that you can't forget to test any entries and that the file is up-to-date.

tests/Unit/Helpers/Utilities/Serialization/Versioning.hpp
```diff
@@ -0,0 +1,140 @@
...
+///
+/// By default, old versions are expected to deserialize to \p
+/// current_object, but if the object represented by the serialized
+/// entries has changed (for example, because the \p current_object
+/// has newer options enabled), objects can be passed in \p
+/// old_objects, keyed by the last version they should match.
+///
+/// By default, objects are directly compared for equality.  If the
+/// objects being tested are stored in `std::unique_ptr`s the \p
+/// Compare template parameter can be passed on order to dereference
```
nilsdeppe: `passed on order` -> `passed in to`?

tests/Unit/Helpers/Utilities/Serialization/Versioning.hpp
nilsdeppe: I agree with @wthrowe . I think to do what you want you could just have `current_label` be an older version, then set up the objects as you want in the map. The only thing that's currently not clear to me on how we want to handle is the case where deserialization of an old version is just not possible and we throw an `ERROR`. Could you add to the dox how to test that an appropriate error is thrown in that case?

nilsdeppe (changes requested):
I think the design is great! Thank you for doing this! One question about a specific edge case that we will probably hit at some point. I'm happy for you to squash those changes immediately, even if they're a bit bigger.

tests/Unit/Helpers/Utilities/Serialization/Versioning.hpp
wthrowe: Hadn't thought of the ERROR case.  I've made the old versions into a variant so you can check for error messages instead of success.

If you want to test multiple sets of parameters you can call the function multiple times with different file names.

wthrowe:
Since I squashed directly, I should probably mention two other minor changes:
```patch
     const bool generate_new_entry = false) {
   const versioning_detail::ToCompare<Compare> to_compare{};
 
+  {
+    INFO("Current version should not be listed in old_objects.");
+    CHECK(old_objects.count(current_label) == 0);
+  }
+
   const auto serializations_to_test =
       versioning_detail::read_serializations(filename);
 
```
```patch
            (current_label != serializations_to_test.back().first and
             current_serialization != serializations_to_test.back().second)));
     }
-    CHECK(to_compare(deserialize<Serialize>(current_serialization.data())) ==
-          to_compare(current_object));
+    REQUIRE(to_compare(deserialize<Serialize>(current_serialization.data())) ==
+            to_compare(current_object));
     versioning_detail::write_serialization(filename, current_label,
                                            current_serialization);
   } else {
```
The first should be self-explanatory.  The second prevents writing a new record if that record is already failing the test.

nilsdeppe (approved):
LGTM! Thank you for doing this :D

tests/Unit/Helpers/Utilities/Serialization/Versioning.hpp
nilsvu: Ok

nilsvu (approved):

PR #5653, opened 2023-12-05
Add IMEX SolveImplicitSector mutator

wthrowe:
This is the actual solver for the IMEX code.  The remaining stuff to submit is mostly actions and other similar wrappers.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Imex/SolveImplicitSector.hpp
+++ b/src/Evolution/Imex/SolveImplicitSector.hpp
@@ -1,0 +1,122 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <tuple>
+#include <type_traits>
+
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Evolution/Imex/Mode.hpp"
+#include "Evolution/Imex/Protocols/ImplicitSector.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+class ImexTimeStepper;
+class TimeDelta;
+template <typename TagsList>
+class Variables;
+namespace Tags {
+struct TimeStep;
+template <typename StepperInterface>
+struct TimeStepper;
+}  // namespace Tags
+namespace TimeSteppers {
+template <typename Vars>
+class History;
+}  // namespace TimeSteppers
+namespace imex::Tags {
+template <typename ImplicitSector>
+struct ImplicitHistory;
+struct Mode;
+template <typename Sector>
+struct SolveFailures;
+struct SolveTolerance;
+}  // namespace imex::Tags
+/// \endcond
+
+namespace imex {
+namespace solve_implicit_sector_detail {
+template <typename Tags>
+using ForwardTuple = tmpl::wrap<
+    tmpl::transform<Tags, std::add_lvalue_reference<std::add_const<
+                              tmpl::bind<tmpl::type_from, tmpl::_1>>>>,
+    std::tuple>;
+}  // namespace solve_implicit_sector_detail
+
+/// Perform the implicit solve for one implicit sector.
+///
+/// This will update the tensors in the implicit sector and clean up
+/// the corresponding time stepper history.  A new history entry is
+/// not added, because that should be done with the same values of the
+/// variables used for the explicit portion of the time derivative,
+/// which may still undergo variable-fixing-like corrections.
+///
+/// \warning
+/// This will use the value of `::Tags::Time` from the DataBox.  Most
+/// of the time, the value appropriate for evaluating the explicit RHS
+/// is stored there, so it will likely need to be set to the
+/// appropriate value for the implicit RHS for the duration of this
+/// mutation.
+template <typename SystemVariablesTag, typename ImplicitSector>
+struct SolveImplicitSector {
+  static_assert(
+      tt::assert_conforms_to_v<ImplicitSector, protocols::ImplicitSector>);
+
+ public:
+  using SystemVariables = typename SystemVariablesTag::type;
+  using SectorVariables = Variables<typename ImplicitSector::tensors>;
+
+ private:
+  template <typename Attempt>
+  struct get_tags_from_evolution {
+    using type = typename Attempt::tags_from_evolution;
+  };
+
+  using tags_for_each_attempt =
+      tmpl::transform<typename ImplicitSector::solve_attempts,
+                      get_tags_from_evolution<tmpl::_1>>;
+  // List of tags used for the initial guess followed by lists of tags
+  // used for each solve attempt.
+  using evolution_data_tags =
+      tmpl::push_front<tags_for_each_attempt,
+                       typename ImplicitSector::initial_guess::argument_tags>;
+
+  using EvolutionDataTuple = solve_implicit_sector_detail::ForwardTuple<
+      tmpl::join<evolution_data_tags>>;
+
+  static void apply_impl(
+      gsl::not_null<SystemVariables*> system_variables,
+      gsl::not_null<TimeSteppers::History<SectorVariables>*> implicit_history,
+      gsl::not_null<Scalar<DataVector>*> solve_failures,
+      const ImexTimeStepper& time_stepper, const TimeDelta& time_step,
+      Mode implicit_solve_mode, double implicit_solve_tolerance,
+      const EvolutionDataTuple& joined_evolution_data);
+
+ public:
+  using return_tags = tmpl::list<SystemVariablesTag,
+                                 imex::Tags::ImplicitHistory<ImplicitSector>,
+                                 Tags::SolveFailures<ImplicitSector>>;
+  using argument_tags = tmpl::append<
+      tmpl::list<::Tags::TimeStepper<ImexTimeStepper>, ::Tags::TimeStep,
+                 Tags::Mode, Tags::SolveTolerance>,
+      tmpl::join<evolution_data_tags>>;
+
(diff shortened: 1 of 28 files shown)


src/Evolution/Imex/Tags/Mode.hpp
```diff
@@ -0,0 +1,29 @@
...
+#pragma once
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "Evolution/Imex/Mode.hpp"
+#include "Evolution/Tags.hpp"
+#include "Options/Options.hpp"
+
+namespace imex {
+namespace OptionTags {
+struct ImexMode {
```
nilsdeppe: Do you want to add Doxygen comments for the tags, mostly for discoverability?

src/Evolution/Imex/Tags/Mode.hpp
```diff
@@ -0,0 +1,29 @@
...
+  static constexpr Options::String help{"IMEX implementation to use"};
+  using type = ::imex::Mode;
+  using group = evolution::OptionTags::Group;
```
nilsdeppe: Would it be useful to add an `Imex` subgroup into `Evolution` to group the options together? [later in a separate PR is totally fine if you think it's a good idea!]

src/Evolution/Imex/Tags/SolveTolerance.hpp
```diff
@@ -0,0 +1,30 @@
...
+
+#pragma once
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "Evolution/Tags.hpp"
+#include "Options/Options.hpp"
+
+namespace imex {
+namespace OptionTags {
+struct ImexImplicitSolveTolerance {
```
nilsdeppe: same doxygen question

src/Evolution/Imex/SolveImplicitSector.hpp
```diff
@@ -0,0 +1,118 @@
...
+ private:
+  template <typename Attempt>
+  struct get_tags_from_evolution {
+    using type = typename Attempt::tags_from_evolution;
+  };
+
+  using attempt_tags = tmpl::transform<typename ImplicitSector::solve_attempts,
+                                       get_tags_from_evolution<tmpl::_1>>;
+  using evolution_data_tags =
+      tmpl::push_front<attempt_tags,
```
nilsdeppe: Is it possible `tmpl::append` here to combine the two lists? If not, maybe a comment why?

src/Evolution/Imex/SolveImplicitSector.tpp
```diff
@@ -0,0 +1,520 @@
...
+               initial_guess_arguments, tmpl::type_<ImplicitSector>{});
+  if (not equation.do_solve()) {
+    return;
+  }
+
+  const size_t number_of_grid_points = get(*solve_failures).size();
+  // Only allocated if used.
+  Matrix semi_implicit_jacobian{};
+
+  bool had_failure = true;
```
nilsdeppe: I think I would find reading the code easier if this was `solve_succeeded` or something like that. `not had_failure` I think is `success` and I would find it easier to read
```cpp
if (solve_succeeded) {
  return;
}
```

src/Evolution/Imex/SolveImplicitSector.tpp
```diff
@@ -0,0 +1,520 @@
...
+              had_failure = true;
+              continue;
+            } else {
+              throw;
+            }
+          }
+          break;
+        }
+        case Mode::SemiImplicit: {
+          auto correction_array = solver(initial_guess);
```
nilsdeppe: I'm a bit unclear on what memory allocations are being done in the `SemiImplicit` solve. Does this `solver` call and the `jacobian()` call below do memory allocations? I think the `lapack::general_matrix_linear_solve` call also does a `std::vector` allocation inside. If we are doing those three allocations, could they be combine or shared between grid points? If that's already being done, would a comment or two in this part of the switch statement be useful?

Edit: from looking at the above code I think there aren't any allocations, other than maybe inside `general_matrix_linear_solve`? Maybe a few comments here would help make that clear?

src/Evolution/Imex/SolveImplicitSector.tpp
```diff
@@ -0,0 +1,520 @@
...
+           "initial_guess must return one GuessResult per point or an empty "
+           "vector for GuessResult::InitialGuess everywhere.");
+  }
+
+  double implicit_weight() const { return implicit_weight_; }
+  const SectorVariables& inhomogeneous_terms() const {
+    return inhomogeneous_terms_;
+  }
+
+  bool do_solve() const { return implicit_weight_ != 0.0; }
```
nilsdeppe: the function name makes me think it does some computation. Maybe rename to something like `solved_needed`?

src/Evolution/Imex/SolveImplicitSector.tpp
```diff
@@ -0,0 +1,520 @@
...
+      db::compute_databox_type<tmpl::append<simple_tags, compute_tags>>;
+
+ public:
+  template <typename PassedEvolutionData>
+  ImplicitSolver(const ImplicitEquation<SectorVariables>& implicit_equation,
+                 PassedEvolutionData&& data_from_evolution)
+      : solve_box_(db::create<simple_tags, compute_tags>()),
+        implicit_equation_(&implicit_equation) {
+    static_assert(std::is_same_v<PassedEvolutionData, const EvolutionData&>,
+                  "ImplicitSolver was passed a temporary.");
```
nilsdeppe: Could you expand this error message? I'm not sure what I should do if I were to get this.

src/Evolution/Imex/SolveImplicitSector.tpp
```diff
@@ -0,0 +1,520 @@
...
+  }
+
+ private:
+  double implicit_weight_;
+  SectorVariables inhomogeneous_terms_{};
+  std::vector<GuessResult> initial_guess_types_{};
+};
+
+template <typename ImplicitSector, typename SolveAttempt>
+class ImplicitSolver {
```
nilsdeppe, line 137: Would some overview documentation be useful here? Not necessarily for doxygen, but to get a quick overview of what the different functions do and why? Maybe also for `ImplicitEquation`?

tests/Unit/Helpers/Evolution/Imex/TestSector.hpp
```diff
@@ -0,0 +1,209 @@
...
+template <typename Tag>
+struct ReadOnlySource : ::db::SimpleTag {
+  using type = typename Tag::type;
+};
+
+template <typename Tag>
+struct ReadOnly : Tag, ::db::ComputeTag {
+  using base = Tag;
+  using argument_tags = tmpl::list<ReadOnlySource<Tag>>;
+  static void function(gsl::not_null<typename ReadOnly::type*> dest,
```
nilsdeppe: `const gsl::not_null`?

tests/Unit/Evolution/Imex/Test_SolveImplicitSector.cpp
```diff
@@ -0,0 +1,750 @@
...
+    CHECK(not performing_step_with_no_implicit_term);
+
+    std::pair current_values{var2, var3};
+    CHECK((*prep_run_values)[static_cast<size_t>(Prep)] != current_values);
+    (*prep_run_values)[static_cast<size_t>(Prep)] = std::move(current_values);
+  }
+};
+// End stuff only used for DataBox handling
+
+// [initial_guess]
```
nilsdeppe, line 138: Do you want this `initial_guess` to be around the `AnalyticSolution` or the `InitialGuess` class below? Right now I think it's around the `AnalyticSolution`

nilsdeppe (changes requested):
This is great! Thanks for all your hard work on the IMEX code! I have a few suggestions/comments/questions. Please feel free to rebase and squash directly since I don't think any of them are major.

src/Evolution/Imex/SolveImplicitSector.hpp
wthrowe: Renamed a thing and added some comments to try to make this clearer.  Not sure if it's enough.

src/Evolution/Imex/SolveImplicitSector.tpp
wthrowe: You should stop passing a temporary.  I've added a brief explanation of why.

src/Evolution/Imex/SolveImplicitSector.tpp
wthrowe: There weren't supposed to be any after the first point, but you're right that I missed one in the LAPACK wrapper, so I fixed that.

I've made the types of some variables and temporaries explicitly arrays, so it should be clear that there are no allocations in them.

tests/Unit/Evolution/Imex/Test_SolveImplicitSector.cpp
wthrowe: This is intentional.  They are both valid "initial_guess" structs (the example chooses between them with a `tmpl::conditional_t`), but this one does a reasonable calculation while `InitialGuess` just does some nonsensical modifications to the data.

src/Evolution/Imex/Tags/Mode.hpp
wthrowe: Done, and renamed some tags both to avoid duplicating the group name and for consistency.

wthrowe (commented):
This is now dependent on (and includes the commits from) #5657.

nilsdeppe (commented):
LGTM, please go ahead and squash!

wthrowe:
@nilsdeppe I think this is ready to go.

nilsdeppe (approved):

PR #5657, opened 2023-12-07
Convert Tags::TimeStepper to not be a base tag

wthrowe:
First commit is IMEX changes that would have major conflicts with this redesign.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/Tags/TimeStepper.hpp
+++ b/src/Time/Tags/TimeStepper.hpp
@@ -13,18 +13,15 @@
 namespace Tags {
 /// \ingroup DataBoxTagsGroup
 /// \ingroup TimeGroup
-/// \brief Tag for a ::TimeStepper of type `StepperType`.
+/// The evolution TimeStepper.  The template parameter should be one
+/// of the time stepper base classes, such as `TimeStepper` or
+/// `LtsTimeStepper`.
 ///
-/// Leaving the template parameter unspecified gives a base tag.
-template <typename StepperType = void>
-struct TimeStepper;
-
-/// \cond
-template <>
-struct TimeStepper<> : db::BaseTag {};
-
+/// For the contained object to be used, the reference tags listed in
+/// `time_stepper_ref_tags<StepperType>` will also need to be added to
+/// the DataBox.
 template <typename StepperType>
-struct TimeStepper : TimeStepper<>, db::SimpleTag {
+struct ConcreteTimeStepper : db::SimpleTag {
   using type = std::unique_ptr<StepperType>;
   using option_tags = tmpl::list<::OptionTags::TimeStepper<StepperType>>;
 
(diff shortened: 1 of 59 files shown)


src/Time/Tags/TimeStepper.hpp
```diff
@@ -34,5 +25,34 @@ struct TimeStepper : TimeStepper<>, db::SimpleTag {
...
+  using base = TimeStepper<StepperInterface>;
+  using argument_tags = tmpl::list<ConcreteTimeStepper<StepperType>>;
+  static const StepperInterface& get(const StepperType& stepper) {
+    return stepper;
+  }
+};
 }  // namespace Tags
+
+/// \ingroup TimeGroup
+/// List of Tags::TimeStepperRef specializations needed then adding a
```
kidder: then -> when

src/Time/Tags/TimeStepper.hpp
```diff
@@ -13,18 +13,9 @@
...
-///
-/// Leaving the template parameter unspecified gives a base tag.
-template <typename StepperType = void>
-struct TimeStepper;
-
-/// \cond
-template <>
-struct TimeStepper<> : db::BaseTag {};
-
+/// The evolution TimeStepper.
```
kidder: Add something like:
`/// \detail StepperType can be either a concrete type (e.g. in a test) or the abstract base type (e.g. in an executable that lets you select the concrete type from the input file)`

src/Time/Tags/TimeStepper.hpp
```diff
@@ -34,5 +25,34 @@ struct TimeStepper : TimeStepper<>, db::SimpleTag {
     return deserialize<type>(serialize<type>(time_stepper).data());
   }
 };
-/// \endcond
+
+/// \ingroup DataBoxTagsGroup
+/// \ingroup TimeGroup
+/// Access to a time stepper through the `StepperInterface` interface.
+template <typename StepperInterface>
```
kidder, line 45: add something like 
`/// \detail This tag should only be used for fetching (i.e. calling get or in an argument_tags list).  This tag should not be directly added to a DataBox or the ConstGlobalCache.  Instead add the appropriate ConcreteTimeStepper and TimeStepperRef tags (which can be obtained with time_stepper_ref_tags).`

kidder (changes requested):
you can squash the changes

nilsdeppe:
I have no additional comments. Please merge once Larry's happy. No need to wait for me :)

src/Time/Tags/TimeStepper.hpp
wthrowe: I decided to stop using concrete types, since it was unnecessarily confusing.  (The only benefit was that you didn't need static casts when creating DataBoxes.)

knelli2:
I was trying to get the GH+CCE executable compiling again an ran into an issue with two time steppers being added to the box (one for GH, one for CCE) that had the same base tag `Tags::TimeStepper<>`. I think this PR will fix my issue.

wthrowe (commented):
I also changed the tests to use the `time_stepper_ref_tags` alias instead of adding the tags directly, and added a test of that.

kidder (commented):
fixup looks good, squash

src/Time/Tags/TimeStepper.hpp
```diff
@@ -34,5 +31,42 @@ struct TimeStepper : TimeStepper<>, db::SimpleTag {
...
   }
 };
-/// \endcond
+
+/// \ingroup DataBoxTagsGroup
+/// \ingroup TimeGroup
+/// Access to a time stepper through the `StepperInterface` interface
+/// (such as `TimeStepper` or `LtsTimeStepper`).
+///
+/// \detail This tag cannot be added directly to the DataBox of
```
kidder: `\details`

kidder (changes requested):
sorry missed this

kidder (approved):

PR #5662, opened 2023-12-10
Worldtube data manager for Klein-Gordon CCE

Sizheng-Ma:
## Proposed changes

This PR constructs `KleinGordonWorldtubeDataManager` along with its unit test. Also, it adds option tags for future development.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/OptionTags.hpp
+++ b/src/Evolution/Systems/Cce/OptionTags.hpp
@@ -141,6 +141,14 @@ struct BoundaryDataFilename {
   using group = Cce;
 };
 
+struct KleinGordonBoundaryDataFilename {
+  using type = std::string;
+  static constexpr Options::String help{
+      "H5 file to read the Klein-Gordon wordltube data from. It could be the "
+      "same as/different from `BoundaryDataFilename`."};
+  using group = Cce;
+};
+
 struct H5LookaheadTimes {
   using type = size_t;
   static constexpr Options::String help{
@@ -279,7 +287,8 @@ struct CceEvolutionPrefix : Tag {
   }
 };
 
-/// A tag that constructs a `MetricWorldtubeDataManager` from options
+/// A tag that constructs a `MetricWorldtubeDataManager` or
+/// `BondiWorldtubeDataManager` from options
 struct H5WorldtubeBoundaryDataManager : db::SimpleTag {
   using type = std::unique_ptr<WorldtubeDataManager<
       Tags::characteristic_worldtube_boundary_tags<Tags::BoundaryValue>>>;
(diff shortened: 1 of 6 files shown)


src/Evolution/Systems/Cce/WorldtubeDataManager.hpp
```diff
@@ -49,9 +49,16 @@ void populate_hypersurface_boundary_data(
     size_t l_max, size_t buffer_depth);
 }  // namespace detail
 
+namespace Tags {
+using klein_gordon_worldtube_boundary_tags =
+    tmpl::list<Tags::BoundaryValue<Tags::KleinGordonPsi>,
+               Tags::BoundaryValue<Tags::KleinGordonPi>>;
+} // namespace Tags
```
knelli2: Should these tags go somewhere else? Where are the tags for the other two WorldtubeDataManagers kept?

src/Evolution/Systems/Cce/OptionTags.hpp
```diff
@@ -320,6 +329,28 @@ struct H5WorldtubeBoundaryDataManager : db::SimpleTag {
   }
 };
 
+/// A tag that constructs a `KleinGordonWorldtubeDataManager` from options
+struct KleinGordonH5WorldtubeBoundaryDataManager : db::SimpleTag {
```
knelli2, line 333: Why not combine this tag with the existing one? It doesn't need any extra parameters. Then instead of having a separate option for `KleinGordonBoundaryDataFilename`, you change the `H5IsBondiData` tag to be an enum (which you'd have to add) like

```cpp
enum class H5DataType { CartesianMetric, Bondi, KleinGordon };
```

and just use that to decide which `WorldtubeDataManager` to construct.

src/Evolution/Systems/Cce/OptionTags.hpp
```diff
@@ -141,6 +141,14 @@ struct BoundaryDataFilename {
   using group = Cce;
 };
 
+struct KleinGrodonBoundaryDataFilename {
```
knelli2: Mispelled `Grodon` -> `Gordon`.

knelli2 (changes requested):
Just a couple things. Also, the first commit message is too long. Shorten it to 72 characters or less.

src/Evolution/Systems/Cce/OptionTags.hpp
Sizheng-Ma: This manager is designed to handle only the Klein-Gordon tags (`Tags::klein_gordon_worldtube_boundary_tags`). The other tensor tags (`Tags::characteristic_worldtube_boundary_tags<Tags::BoundaryValue>`) are still managed by the current `H5WorldtubeBoundaryDataManager`. In other words, there will be two managers in the scalar-tensor CCE system, in order to deal with boundary data separately (`Tags::klein_gordon_worldtube_boundary_tags`+`Tags::characteristic_worldtube_boundary_tags<Tags::BoundaryValue>`).

This can lead to two major benefits. First, on the SpEC side, we might dump tensor and scalar data into different files. Having two managers offers us more freedom to handle this scenario. Second, having a separate manager for the Klein-Gordon variables alone allows us to construct a standalone scalar CCE system (rather than the current coupled scalar+tensor CCE).

src/Evolution/Systems/Cce/WorldtubeDataManager.hpp
Sizheng-Ma: They are kept in `src/Evolution/Systems/Cce/BoundaryData.hpp`,  where the tags are computed from spacetime metric, etc. It's not relevant here. The only other place I could think of is to move them to `src/Evolution/Systems/Cce/Tags.hpp`.

Sizheng-Ma:
> Just a couple things. Also, the first commit message is too long. Shorten it to 72 characters or less.

Done!

src/Evolution/Systems/Cce/OptionTags.hpp
knelli2: Ahh I see. I did not realize this new manager would work in parallel with the old one. I thought it was just a replacement for the KG system. That all makes sense. Thanks!

src/Evolution/Systems/Cce/WorldtubeDataManager.hpp
knelli2: The `characteristic_worldtube_boundary_tags` list in `BoundaryData.hpp` seems to be basically the same as this `klein_gordon_worldtube_boundary_tags` list, just with more (and different) tags and potentially a different prefix. So I believe it makes sense to keep these two lists together so we don't have to go looking in two different places for a boundary tags list.

Also, having the list in a more "central" place allows us to reuse the type alias if we need (no idea if that's the plan or not).

knelli2 (changes requested):
You can squash everything if you agree with this last change.

Sizheng-Ma:
Sounds good. I've moved it to `BoundaryData.hpp`.

knelli2 (approved):

PR #1650, opened 2019-08-05
Improve handling of options for initialization

kidder:
## Proposed changes

This PR improves the handling of options for initialization of the DataBoxes of components.
It fixes a potential bug as there was no guarantee that the options passed into the components
corresponded to the specified tags (types were checked by the compiler, but options of the same
type potentially could be mixed up).
This PR will allow the initialization items to have different types from the options constructed from the input file.

### Types of changes:

- [x] Bugfix
- [x] New feature

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Code review checklist

- [ ] The PR passes all checks, including unit tests and `clang-tidy`.
  For instructions on how to perform the CI checks locally refer to the [Dev
  guide on the Travis CI](https://spectre-code.org/travis_guide.html).
- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).

### Further comments

I will add a commit to update the parallel infrastructure documentation.
In a forthcoming PR, I will update the ConstGlobalCache to use this infrastructure so the Domain
can be placed in the ConstGlobalCache.

PR #5664, opened 2023-12-11
Add copy of vector of EaDT from options to cache

knelli2:
## Proposed changes

Even though the EventsAndDenseTriggers object itself is mutated inside the element DataBox, the actual dense triggers and events aren't changed and can therefore, in theory, be available in the GlobalCache. This PR adds a new mutable cache tag that stores the vector of dense triggers and events. The reason it is mutable is because EaDT has a method `add_trigger_and_events` where you can add more triggers/events during execution. This means that in order to keep the cache tag up to date, it will have to be mutated. The mutation is left to whatever adds the new triggers/events. Currently, this PR doesn't add any of the special mutations that are needed (basically the control system triggers/events).

This PR will be needed for interpolation framework changes.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/EventsAndDenseTriggers/Tags.hpp
+++ b/src/Evolution/EventsAndDenseTriggers/Tags.hpp
@@ -17,7 +17,7 @@ namespace OptionTags {
  * OptionTags::EventsAndTriggers
  */
 struct EventsAndDenseTriggers {
-  using type = evolution::EventsAndDenseTriggers;
+  using type = std::vector<evolution::DenseTriggerAndEventsConstruction>;
   static constexpr Options::String help = "Events to run at arbitrary times";
 };
 }  // namespace OptionTags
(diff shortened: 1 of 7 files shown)


src/Evolution/EventsAndDenseTriggers/Tags.hpp
```diff
@@ -31,8 +31,43 @@ struct EventsAndDenseTriggers : db::SimpleTag {
...
-    return deserialize<type>(serialize(events_and_triggers).data());
+  static type create_from_options(
+      const std::vector<evolution::DenseTriggerAndEventsConstruction>&
+          construction) {
+    // The deserialize(serialize()) is to deal with unique_ptrs in the data
+    // structures since we don't have `get_clone` implemented for
+    // triggers/events
+    return deserialize<type>(
+        serialize(
+            type{deserialize<
```
wthrowe, line 42: I don't think you need to go through the serialization framework twice.  You already have an rvalue after the inner deserialize.

wthrowe (changes requested):
> the actual dense triggers and events aren't changed

I think what you meant here is that the identities of the dense triggers and events aren't changed?  Because the dense trigger objects themselves are.

I'd like more information about how you intend to use this.  What will update it, and what will read it?  It can't be used in the way we've used mutable cache tags in the past, since there is no way to handle an is_ready call.

Do you actually need the dense triggers, or would just a list of events be enough?

knelli2:
> I think what you meant here is that the identities of the dense triggers and events aren't changed? Because the dense trigger objects themselves are.

Ah yes that is what I meant, and is much clearer.

So at the end of the day, what I need is a vector of events (interpolation events, which I have a way of distinguishing)  that are going to be run during the simulation. I will then use this vector of events in the `allocate_array` function of an Array chare to initialize one element of the array for each event (using a unique string for each event as the array index). Basically the only thing that can be guaranteed to exist in `allocate_array` is the GlobalCache. Since the EaTs are already stored in the cache, it is simple to get the EaTs, check if the event is an interpolation, and add it to the vector. The EaDTs are not as simple because we can add more dense triggers/events during the initialization phase with the `add_trigger_and_events` method of EaDT.

So yes, I technically only need the events from the EaDTs available in the `allocate_array` function. I believe it's too complicated to try and create more array elements after initialization has started which is why I would like to get everything done in the `allocate_array` function, but I'm not really sure that is possible.

wthrowe:
The `allocate_array` functions used to be able to take arguments, but that appears to have been removed in #1650 (since `Main` and `DistributedObject` now both use `component::simple_tags_from_options`).  I think that's really what you want here?  @kidder I haven't read through that entire PR discussion, but do you remember if there was a reason for that decision?

kidder:
- The `allocate_array` functions are passed a tagged tuple of initialization_items which can be constructed from the input file options (and the metavariables) via the `simple_tags_from_options` typelist.  Is there other information needed?  If so, where is it coming from?

wthrowe:
Unless I'm misreading the code, that same list is also used to compute the DataBox, though.

kidder:
- yes, that is true.  It could be easily changed if needed.  In the past there was a separate `array_allocation_tags` that could be reinstated for anything only needed inside the `allocate_array` function.  
- anything in the const global cache is also available during the `allocate_array` function
- I'm not clear on what this array component is, and what information you want to pass the `allocate_array` function

knelli2:
Closing this in favor of re-implementing the `array_allocation_tags` method. This was discussed with the reviewers offline.

PR #5666, opened 2023-12-13
Remove 2nd order worldtube evolution

nikwit:
## Proposed changes
Removes the second order evolution of the worldtube system. This is because we will add the effect of the scalar self force shortly but are not planning to implement it beyond first order in the near future.

### Upgrade instructions
The second order worldtube evolution is disabled, use orders 0 or 1.


nikwit:
I don't really understand the clang-tidy error since I did not edit the file it is referring to.

knelli2 (approved):

wthrowe (approved):
I've seen the hdf5.h clang-tidy error on another PR, but haven't gotten around to investigating.  It's unrelated to this.

PR #5677, opened 2023-12-15
Generalize `InertialParticlePosition` compute tag

nikwit:
The compute tag was fixed for circular orbits. Here, I generalize it to output the position and velocity for arbitrary orbits by calling the functions of time on the grid coordinates of the particle.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
@@ -69,7 +69,8 @@ void FaceCoordinatesCompute<Dim, Frame, Centered>::function(
         result,
     const ::ExcisionSphere<Dim>& excision_sphere, const Element<Dim>& element,
     const tnsr::I<DataVector, Dim, ::Frame::Inertial>& inertial_coords,
-    const Mesh<Dim>& mesh, const tnsr::I<double, Dim>& particle_position) {
+    const Mesh<Dim>& mesh,
+    const std::array<tnsr::I<double, Dim>, 2>& particle_position_velocity) {
   if constexpr (not(Centered and std::is_same_v<Frame, ::Frame::Inertial>)) {
     ERROR("Should be centered in inertial frame");
   }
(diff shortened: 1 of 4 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -91,7 +92,7 @@ void FaceCoordinatesCompute<Dim, Frame, Centered>::function(
                   mesh.extents(), direction.value().dimension(),
                   index_to_slice_at(mesh.extents(), direction.value()));
     for (size_t i = 0; i < Dim; ++i) {
-      result->value().get(i) -= particle_position.get(i);
+      result->value().get(i) -= particle_position_velocity.at(0).get(i);
```
knelli2, line 95: Question: You only use the position here. Presumably in a future PR you'll use the velocity either here or somewhere else?

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_Tags.cpp
```diff
@@ -296,31 +296,49 @@ void test_compute_face_coordinates() {
...
+                                   0.}};
+    tnsr::I<double, Dim> velocity{
+        {-orbit_radius * angular_velocity * sin(angular_velocity * time),
+         orbit_radius * angular_velocity * cos(angular_velocity * time), 0.}};
+    return std::array<tnsr::I<double, Dim>, 2>{std::move(position),
+                                               std::move(velocity)};
+  };
+  // There is an error introduced here which presumable comes from the
+  // quaternion function of time integrating the orbit
+  Approx apprx = Approx::custom().epsilon(1e-8).scale(1.0);
```
knelli2: Yes it's because it's trying to integrate an ODE from 0 to a large time which the quaternion is changing an appreciable amount. To get a more accurate answer, limit the time distribution from 0 to 1.

[Nitpick] Also typo: `presumable` -> `presumably`.

knelli2 (changes requested):
Also please fix the clang-tidy error. You can squash all changes directly.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
nikwit: Yes, for example in #5681

knelli2 (approved):

PR #5682, opened 2023-12-18
Remove control system Event

knelli2:
## Proposed changes

The control system event used to correspond to a measurement and loop over all submeasurements and call an `apply` function. For most submeasurements, the `apply` function just copied code from a different event, or actually called a different event altogether. Additionally, for upcoming interpolation infrastructure changes, it is useful/necessary to get all event being called from the factory creation list in the metavars. Therefore, these two factors motivated this change of getting rid of the control system event, and instead just having the submeaurements specify an event to run. Now, instead of one event being run per measurement, many events will be run.

I had to make an event for the BNS submeasurement, but no calculation was changed. Just how the code was called.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #5665.~


knelli2:
Rebased on develop and no longer dependent.

src/ControlSystem/Actions/InitializeMeasurements.hpp
```diff
@@ -99,16 +98,24 @@ struct InitializeMeasurements {
...
+                using measurement = tmpl::type_from<decltype(measurement_v)>;
                 using control_system_group =
                     metafunctions::control_systems_with_measurement_t<
-                        ControlSystems,
-                        typename tmpl::type_from<decltype(measurement_v)>>;
+                        ControlSystems, measurement>;
+                using events = tmpl::transform<
+                    typename measurement::submeasurements,
+                    tmpl::bind<metafunctions::event_from_submeasurement_t,
+                               tmpl::pin<control_system_group>, tmpl::_1>>;
```
wthrowe: Use the lazy version to avoid the need for bind.

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
```diff
@@ -91,6 +91,133 @@ void center_of_mass_integral_on_element(
...
+            ControlSystems>>(std::move(reduction_data), my_proxy,
+                             reduction_target_proxy);
+  }
+
+  using is_ready_argument_tags = tmpl::list<::Tags::Time>;
+
+  template <typename Metavariables, typename ArrayIndex, typename Component>
+  bool is_ready(const double time, Parallel::GlobalCache<Metavariables>& cache,
+                const ArrayIndex& array_index,
+                const Component* const component) const {
```
wthrowe: This function is redundant.  The same check is done before calling it, so it will always return true.

src/ControlSystem/Metafunctions.hpp
```diff
@@ -130,5 +109,40 @@ template <typename Metavariables>
...
+template <typename AllControlSystems, typename Measurement>
+struct events_from_measurement {
+  using submeasurements = submeasurements_t<Measurement>;
+  using control_systems_with_measurement =
+      control_systems_with_measurement_t<AllControlSystems, Measurement>;
+
+  using type = tmpl::transform<
+      submeasurements,
+      tmpl::bind<event_from_submeasurement_t,
+                 tmpl::pin<control_systems_with_measurement>, tmpl::_1>>;
```
wthrowe: Use the lazy version.

src/ControlSystem/Metafunctions.hpp
```diff
@@ -130,5 +109,40 @@ template <typename Metavariables>
...
+}  // namespace detail
+
+/// \ingroup ControlSystemGroup
+/// The list of events needed for measurements for a list of control
+/// systems.
+template <typename ControlSystems>
+using control_system_events = tmpl::flatten<
+    tmpl::transform<metafunctions::measurements_t<ControlSystems>,
+                    tmpl::bind<detail::events_from_measurement_t,
+                               tmpl::pin<ControlSystems>, tmpl::_1>>>;
```
wthrowe: Use lazy version.

src/ControlSystem/Protocols/Submeasurement.hpp
```diff
@@ -8,45 +8,39 @@
...
+ * A conforming struct must provide
+ *
+ * - An `interpolation_target_tag` type alias templated on the \ref
+ *   ControlSystem "control systems" using this submeasurement.  (This template
+ *   parameter must be used in the call to `RunCallbacks` discussed below.) This
+ *   alias may be `void` if the submeasurement does not use an interpolation
+ *   target tag.  This is only used to collect the tags that must be registered
+ *   in the metavariables.
+ * - An `event` type alias also templated on the \ref ControlSystem "control
+ *   systems" using this submeasurement which is an `::Event`. This is because
```
wthrowe: The line break seems to be screwing up the rendering of this link.

wthrowe: What is the "This" near the end of this line referring to?

src/ControlSystem/Protocols/Submeasurement.hpp
```diff
@@ -8,45 +8,39 @@
...
+ * These structs are referenced from structs conforming to the
+ * Measurement protocol.  They define independent parts of a control
+ * system measurement, such as individual horizon-finds in a
+ * two-horizon measurement.
+ *
+ * A conforming struct must provide
+ *
+ * - An `interpolation_target_tag` type alias templated on the \ref
+ *   ControlSystem "control systems" using this submeasurement.  (This template
+ *   parameter must be used in the call to `RunCallbacks` discussed below.) This
```
wthrowe, line 24: There's no mention of this in the discussion of RunCallbacks below.

wthrowe (changes requested):

src/ControlSystem/Measurements/BNSCenterOfMass.hpp
knelli2: Where is the check done? I looked at the Event base class, EventsAnd[Dense]Triggers, and RunEventsAnd[Dense]Triggers, and I didn't find it. Other events that the control system use have basically the same `is_ready` function.

wthrowe: For dense triggers, `domain::CheckFunctionsOfTimeAreReadyPostprocessor` in `src/Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp` will cause a check for any event with `needs_evolved_variables()` returning true.  For non-dense triggers, `domain::Actions::CheckFunctionsOfTimeAreReady` in the same file will have been run.

Some events had checks added before the general dense-trigger check existed and I didn't think to remove the now-redundant code when it was added.

knelli2: Ok I just changed this function to only return `true`. I think the larger change is beyond the scope of this PR

knelli2:
Pushed a fixup

wthrowe:
Looks good.  Squash.

wthrowe (approved):

PR #5683, opened 2023-12-19
Boundary component for Klein-Gordon CCE

Sizheng-Ma:
## Proposed changes

This PR adds a boundary parallel component `KleinGordonH5WorldtubeBoundary` for Klein-Gordon CCE. The component has a single initialization action `Actions::InitializeWorldtubeBoundary<KleinGordonH5WorldtubeBoundary<Metavariables>>`, which initializes two tag lists `Metavariables::cce_boundary_communication_tags` and `Metavariables::klein_gordon_boundary_communication_tags` ( the metavariable will be added in the next PR). The third and fourth commits of this PR add the initialization action and the component respectively.

The initialization action is constructed from a base class `InitializeWorldtubeBoundaryBase`, which currently holds a single tag list. This is fine for the current CCE system since the code treats `Metavariables::cce_boundary_communication_tags` as one `::Tags::Variables`. However, for the upcoming Klein-Gordon CCE system, we want to handle `Metavariables::cce_boundary_communication_tags` and `Metavariables::klein_gordon_boundary_communication_tags` separately, so we initialize them as two different chunks of `::Tags::Variables`. In the first commit, a tag pack is added to `InitializeWorldtubeBoundaryBase` for this purpose.

The second commit is a minor change. The function `create_fake_time_varying_klein_gordon_data` is moved to a new file so that it can be used by other unit tests.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/Actions/InitializeWorldtubeBoundary.hpp
+++ b/src/Evolution/Systems/Cce/Actions/InitializeWorldtubeBoundary.hpp
@@ -45,13 +45,13 @@ namespace Actions {
 
 namespace detail {
 template <typename Initializer, typename ManagerTags,
-          typename BoundaryCommunicationTagsList>
+          typename... BoundaryCommunicationTagsList>
 struct InitializeWorldtubeBoundaryBase {
   using simple_tags_from_options = ManagerTags;
   using const_global_cache_tags = tmpl::list<Tags::LMax>;
 
   using simple_tags =
-      tmpl::list<::Tags::Variables<BoundaryCommunicationTagsList>>;
+      tmpl::list<::Tags::Variables<BoundaryCommunicationTagsList>...>;
 
   template <typename DataBoxTagsList, typename... InboxTags,
             typename ArrayIndex, typename Metavariables, typename ActionList,
@@ -78,11 +78,11 @@ struct InitializeWorldtubeBoundaryBase {
       }
     }
     const size_t l_max = db::get<Tags::LMax>(box);
-    Variables<BoundaryCommunicationTagsList> boundary_variables{
-        Spectral::Swsh::number_of_swsh_collocation_points(l_max)};
 
-    Initialization::mutate_assign<simple_tags>(make_not_null(&box),
-                                               std::move(boundary_variables));
+    Initialization::mutate_assign<simple_tags>(
+        make_not_null(&box),
+        Variables<BoundaryCommunicationTagsList>{
+            Spectral::Swsh::number_of_swsh_collocation_points(l_max)}...);
     return {Parallel::AlgorithmExecution::Continue, std::nullopt};
   }
 };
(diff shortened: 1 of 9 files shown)


src/Evolution/Systems/Cce/Actions/InitializeWorldtubeBoundary.hpp
```diff
@@ -45,13 +45,16 @@ namespace Actions {
 
 namespace detail {
 template <typename Initializer, typename ManagerTags,
-          typename BoundaryCommunicationTagsList>
+          typename BoundaryCommunicationTagsList1,
+          typename... BoundaryCommunicationTagsList2>
```
knelli2: With just this commit, this actually adds a Variables tag with an empty list since there are no extra tag lists yet (which is not something we want). I'd recommend something like

```cpp
template <typename Initializer, typename ManagerTags,
          typename... BoundaryCommunicationTagsList> {
  // other stuff

  simple_tags = tmpl::list<::Tags::Variables<BoundaryCommunicationTagsList>...>;

  // inside apply function
    Initialization::mutate_assign<simple_tags>(make_not_null(&box),
                     Variables<BoundaryCommunicationTagsList>{
        Spectral::Swsh::number_of_swsh_collocation_points(l_max)}...);
};
```

Not that we'll likely need more than 2 lists, but this also avoids the annoying `1` and `2` suffixes.

tests/Unit/Helpers/Evolution/Systems/Cce/CMakeLists.txt
```diff
@@ -7,6 +7,7 @@ set(LIBRARY "CceHelpers")
 
 set(LIBRARY_SOURCES
   CceComputationTestHelpers.cpp
+  KleinGordonBoundaryTestHelpers.cpp
```
knelli2, line 10: You need to add the `hpp` too.

src/Evolution/Systems/Cce/Actions/InitializeWorldtubeBoundary.hpp
```diff
@@ -140,6 +142,51 @@ struct InitializeWorldtubeBoundary<H5WorldtubeBoundary<Metavariables>>
...
 };
 
+/*!
+ * \ingroup ActionsGroup
+ * \brief Initializes a KleinGordonH5WorldtubeBoundary
+ *
+ * \details Uses:
+ * - initialization tag
+ * `Cce::Tags::H5WorldtubeBoundaryDataManager`,
+ * `Cce::Tags::KleinGordonH5WorldtubeBoundaryDataManager`.
```
knelli2: We don't have the `initialization_tag` type alias anymore, it's called `simple_tags_from_options` now. I know it's incorrect for the other ones as well. Could you please fix those dox in a separate commit?

knelli2 (changes requested):

src/Evolution/Systems/Cce/Actions/InitializeWorldtubeBoundary.hpp
Sizheng-Ma: Good idea!

src/Evolution/Systems/Cce/Actions/InitializeWorldtubeBoundary.hpp
Sizheng-Ma: Done!

knelli2 (commented):
Looks good. You can squash

Sizheng-Ma:
Done!

knelli2 (approved):
Test timeouts unrelated.

PR #5679, opened 2023-12-16
Support weak form in elliptic DG

nilsvu:
## Proposed changes

Almost gets the elliptic solver to work on large sphere domains! The attached `Lorentzian.yaml` configuration works, only BBH ID needs some more treatment (next PR).

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5667, opened 2023-12-13
Switch elliptic DG to a primal formulation 

nilsvu:
## Proposed changes

Handle auxiliary boundary corrections through a weak divergence rather than a Schur complement. This changes how Jacobians are applied on boundaries. It turns out this primal formulation is more stable when the Jacobians are large (for large outer boundaries).

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5684, opened 2023-12-20
Solve for psi-1 in XCTS system

### LINKED ISSUE #4597: Initial data: can we make BBH initial data with a large outer boundary?
Is it possible to make initial data that satisfies the following criteria?
1. converges,
2. completes within 24 hours of wall time, and
3. is at least as accurate as initial data interpolated from SpEC?

If so, great. If not, what's the trouble, and can it be fixed in a ~month timescale?

nilsvu:
## Proposed changes

Psi-1 approaches zero at the outer boundary rather than 1, hence has a more accurate numerical derivative on the stretched outer wedges. Combined with the weak primal form of the elliptic DG scheme (#5667 and #5679) this enables BBH ID with a large outer radius and fixes #4597 (some optimizations can still be done to accelerate the solves).

Try it like this:

```
./bin/spectre bbh generate-id -q 1 -D 16 -w 0.0144 -L 0 -P 5 -o ./ID
```

### Upgrade instructions

In XCTS input files replace `ConformalFactor` and `LapseTimesConformalFactor` with `ConformalFactorMinusOne` and `LapseTimesConformalFactorMinusOne` in analytic boundary conditions and error observations.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe:
I can review once CI is happy :)

nilsvu:
CI is mostly ok now. I don't think I can silence the clang-tidy warning.

src/Utilities/TypeTraits/GetFundamentalType.hpp
```diff
@@ -7,8 +7,15 @@
 
 #include "Utilities/NoSuchType.hpp"
 #include "Utilities/TMPL.hpp"
+#include "Utilities/TypeTraits/CreateHasTypeAlias.hpp"
 
 namespace tt {
+namespace detail {
+CREATE_HAS_TYPE_ALIAS(ElementType)
+CREATE_HAS_TYPE_ALIAS_V(ElementType)
```
nilsdeppe, line 16: [optional] One idea to try is `[[maybe_unused]]` on this line. If that fails to build, we could mark the ones in the macro as that? I'm also fine ignoring and have no strong preference.

src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
```diff
@@ -190,8 +190,9 @@ void negative_expansion_quantities(
```
nilsdeppe, line 190: [optional, random line] there are a lot of divisions by the conformal factor. I'm not sure if it matters because this is a BC, but it may be possible to factor those out and do only one. Probably doesn't matter, though.

src/Elliptic/Systems/Xcts/Equations.cpp
```diff
@@ -61,60 +61,61 @@ void add_hamiltonian_sources(
     const gsl::not_null<Scalar<DataVector>*> hamiltonian_constraint,
```
nilsdeppe, line 61: [musing] It might be worth revisiting this in a followup performance review of the elliptic code. It might be possible to precompute the conformal factor and then pass it around. I think this is low priority, but if you have a list somewhere to keep track of these things, maybe add this?

nilsdeppe (dismissed):
LGTM, couple ideas/thoughts. Feel free to respond how you prefer (e.g., just merging is acceptable :P)

Also, fantastic job on figuring all this out and getting it to work! These are in my experience the most challenging types of bugs to fix and are very frustrating. It's a huge accomplishment and I think you should be really proud! This took a lot of hard work and perseverance! 💯

src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
nilsvu: Haven't profiled this recently, but I'll keep this in mind when doing a round of optimizations 👍 Not much use to do this blindly I think.

src/Elliptic/Systems/Xcts/Equations.cpp
nilsvu: Yes, will do. I have no idea if precomputing it is faster or not.

src/Utilities/TypeTraits/GetFundamentalType.hpp
nilsvu: Added a nolint block to silence this warning (new feature since clang-format 14). There's still another warning about an infinite recursion that cannot happen, which I don't think I can silence since I don't want to nolint the general code in `CachedTempBuffer`.

nilsvu:
Thank you!

nilsdeppe (approved):

PR #5685, opened 2023-12-28
Automate BBH pipeline

nilsvu:
## Proposed changes

Run the inspiral automatically after ID generation. Also continue with ringdown automatically. Implements parts of #5021. Things like eccentricity control etc can be added here as well.

Try it like this (e.g. on Wheeler):

```
./bin/spectre bbh generate-id -q 1 -D 16 -w 0.0144 -d ./BbhTest --evolve
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/InitialData.py
+++ b/support/Pipelines/Bbh/InitialData.py
@@ -3,7 +3,7 @@
 
 import logging
 from pathlib import Path
-from typing import Union
+from typing import Optional, Union
 
 import click
 from rich.pretty import pretty_repr
@@ -67,6 +67,10 @@ def generate_id(
     refinement_level: int,
     polynomial_order: int,
     id_input_file_template: Union[str, Path] = ID_INPUT_FILE_TEMPLATE,
+    evolve: bool = False,
+    pipeline_dir: Optional[Union[str, Path]] = None,
+    run_dir: Optional[Union[str, Path]] = None,
+    segments_dir: Optional[Union[str, Path]] = None,
     **scheduler_kwargs,
 ):
     """Generate initial data for a BBH simulation.
@@ -80,6 +84,28 @@ def generate_id(
         " generated input files."
     )
 
+    # Resolve directories
+    if pipeline_dir:
+        pipeline_dir = Path(pipeline_dir).resolve()
+    assert segments_dir is None, (
+        "Initial data generation doesn't use segments at the moment. Specify"
+        " '--run-dir' / '-o' or '--pipeline-dir' / '-d' instead."
+    )
+    if evolve:
+        assert pipeline_dir is not None, (
+            "Specify a '--pipeline-dir' / '-d' to evolve the initial data."
+            " Don't specify a '--run-dir' / '-o' because it will be created in"
+            " the 'pipeline_dir' automatically."
+        )
+        assert run_dir is None, (
+            "Specify the '--pipeline-dir' / '-d' rather than '--run-dir' / '-o'"
+            " when evolving the initial data. Directories for the initial data,"
+            " evolution, etc will be created in the 'pipeline_dir'"
+            " automatically."
+        )
+    if pipeline_dir and not run_dir:
+        run_dir = pipeline_dir / "001_InitialData"
+
     # Determine initial data parameters from options
     id_params = id_parameters(
         mass_ratio=mass_ratio,
@@ -92,7 +118,15 @@ def generate_id(
     logger.debug(f"Initial data parameters: {pretty_repr(id_params)}")
 
     # Schedule!
-    return schedule(id_input_file_template, **id_params, **scheduler_kwargs)
+    return schedule(
+        id_input_file_template,
+        **id_params,
+        **scheduler_kwargs,
+        evolve=evolve,
+        pipeline_dir=pipeline_dir,
+        run_dir=run_dir,
+        segments_dir=segments_dir,
+    )
 
 
 @click.command(name="generate-id", help=generate_id.__doc__)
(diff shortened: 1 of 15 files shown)


support/Python/RunNext.py
```diff
@@ -0,0 +1,126 @@
...
+    show_default="directory of the INPUT_FILE_PATH",
+)
+def run_next_command(input_file_path, input_run_dir):
+    if input_run_dir is None:
+        input_run_dir = Path(input_file_path).resolve().parent
+    with open(input_file_path) as open_input_file:
+        metadata = next(yaml.safe_load_all(open_input_file))
+    if "Next" not in metadata:
+        logger.info(
+            f"The input file lists no 'Next' entrypoint. Nothing to do."
```
knelli2: `The input file metadata lists no 'Next'...` just to be specific

support/Python/__main__.py
```diff
@@ -35,6 +35,7 @@ def list_commands(self, ctx):
             "render-1d",
             "render-3d",
             "resubmit",
+            "run-next",
             "schedule",
```
knelli2, line 39: I'm worried that people will misinterpret this as running the next segment of an evolution, and not the next portion of a pipeline. It's not obvious to me that you'd have to use `schedule` to run the next segment. No real suggestion at the moment, just wanted to note there might be confusion on this.

tests/support/Python/Test_RunNext.py
```diff
@@ -0,0 +1,64 @@
...
+
+    def test_cli(self):
+        runner = CliRunner()
+        result = runner.invoke(
+            run_next_command,
+            [str(self.input_file_path)],
+            catch_exceptions=False,
+        )
+        self.assertEqual(result.exit_code, 0)
+        self.assertIn(spectre.__version__, result.stdout)
```
knelli2, line 59: Will this test pass in `PY_DEV_MODE`?

support/Pipelines/Bbh/InitialData.py
```diff
@@ -76,6 +80,28 @@ def generate_id(
...
+            " the 'pipeline_dir' automatically."
+        )
+        assert run_dir is None, (
+            "Specify the '--pipeline-dir' / '-d' rather than '--run-dir' / '-o'"
+            " when evolving the initial data. Directories for the initial data,"
+            " evolution, etc will be created in the 'pipeline_dir'"
+            " automatically."
+        )
+    if pipeline_dir and not run_dir:
+        run_dir = pipeline_dir / "001_ID"
```
knelli2: One index or zero index?

Also, not for right now, but once we have eccentricity reduction, we're going to maybe have to change the numbering since we'll need to have multiple ID/Ev rounds

support/Pipelines/Bbh/Ringdown.py
```diff
@@ -144,6 +158,15 @@ def start_ringdown(
...
+    "-d",
+    type=click.Path(
+        writable=True,
+        path_type=Path,
+    ),
+    help="Directory where steps in the pipeline are created.",
+)
 @scheduler_options
 def start_ringdown_command(**kwargs):
     _rich_traceback_guard = True  # Hide traceback until here
```
knelli2, line 172: [Random line] Is it easy to have some tests that all the pipeline directory stuff works, even if it's just `--no-submit` and checking some input files?

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -114,6 +114,16 @@ def start_inspiral(
...
 
+    # Determine resource allocation
+    if (
+        scheduler_kwargs.get("scheduler") is not None
+        and scheduler_kwargs.get("num_procs") is None
+        and scheduler_kwargs.get("num_nodes") is None
+    ):
+        # Just run on 4 nodes for now, because 1 is surely not enough. We can
+        # make this smarter later (e.g. scale with the number of elements).
+        scheduler_kwargs["num_nodes"] = 4
```
knelli2, line 131: Can you add this to the help string above so users are aware of it?

Also, I think this is something that should be machine specific? 4 wheeler nodes are not equivalent to 4 Caltech HPC nodes. Not saying to implement this now, 4 seems like an ok default for all our clusters, but for future consideration.

knelli2 (changes requested):
Overall looks good! I think this is coming together nicely!

support/Python/__main__.py
nilsvu: Hm all I can think of is to look at the help string:

```
$ ./bin/spectre --help
...
  resubmit               Create the next segment in the SEGMENTS_DIR and...
  run-next               Run the next entrypoint specified in the input...
...
$ ./bin/spectre resubmit -h
Usage: spectre resubmit [OPTIONS] [SEGMENTS_DIRS]...

  Create the next segment in the SEGMENTS_DIR and schedule it
...
$ ./bin/spectre run-next -h
Usage: spectre run-next [OPTIONS] INPUT_FILE_PATH

  Run the next entrypoint specified in the input file metadata
...
```

That makes it pretty clear I think.

tests/support/Python/Test_RunNext.py
nilsvu: yep

support/Pipelines/Bbh/InitialData.py
nilsvu: Yep this is just a preliminary hardcoded numbering from 1 to 3. With ecc control etc we can make this dynamic and increment the number with each ecc step.

support/Pipelines/Bbh/Ringdown.py
nilsvu: I couldn't come up with a good test of the full pipeline because the steps actually need data from the previous steps. Of course we can and should add integration/regression tests that run the pipeline on a cluster. I added some tests of the generated input files though as you suggested.

support/Pipelines/Bbh/Inspiral.py
nilsvu: Added to the docs. Yes this has to scale with the node size somehow. Maybe we can specify the number of cores (rather than nodes) and make the scheduler select the number of nodes.

knelli2 (commented):
Ok looks good. Sqash

knelli2 (approved):

PR #5696, opened 2024-01-08
Add RotScaleTrans Map and Tests

### LINKED ISSUE #5673: Combine translation, rotation, expansion map
Needed for ringdown in BBH, inspiral for unequal mass BBH

AlexCarpenter46:
## Proposed changes

Adds the RotScaleTrans map from SpEC. Currently, this map can be used for any combination of rotation, expansion and translation. 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
@@ -1,0 +1,993 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/CoordinateMaps/TimeDependent/RotScaleTrans.hpp"
+
+#include <array>
+#include <cstddef>
+#include <memory>
+#include <optional>
+#include <ostream>
+#include <pup.h>
+#include <pup_stl.h>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <utility>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Matrix.hpp"
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
+#include "DataStructures/Tensor/Identity.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/CoordinateMaps/TimeDependent/RotationMatrixHelpers.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "NumericalAlgorithms/RootFinding/QuadraticEquation.hpp"
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/DereferenceWrapper.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeArray.hpp"
+#include "Utilities/MakeWithValue.hpp"
+#include "Utilities/Serialization/PupStlCpp17.hpp"
+#include "Utilities/StdArrayHelpers.hpp"
+#include "Utilities/StdHelpers.hpp"
+#include "Utilities/TypeTraits/RemoveReferenceWrapper.hpp"
+
+namespace domain::CoordinateMaps::TimeDependent {
+
+template <size_t Dim>
+RotScaleTrans<Dim>::RotScaleTrans(
+    std::optional<std::pair<std::string, std::string>> scale_f_of_t_names,
+    std::optional<std::string> rot_f_of_t_name,
+    std::optional<std::string> trans_f_of_t_name, double inner_radius,
+    double outer_radius, BlockRegion region)
+    : inner_radius_(inner_radius),
+      outer_radius_(outer_radius),
+      region_(region) {
+  if (scale_f_of_t_names.has_value()) {
+    scale_f_of_t_a_ = std::move(scale_f_of_t_names.value().first);
+    scale_f_of_t_b_ = std::move(scale_f_of_t_names.value().second);
+    f_of_t_names_.insert(scale_f_of_t_a_.value());
+    f_of_t_names_.insert(scale_f_of_t_b_.value());
+  }
+  if (rot_f_of_t_name.has_value()) {
+    rot_f_of_t_ = std::move(rot_f_of_t_name.value());
+    f_of_t_names_.insert(rot_f_of_t_.value());
+  }
+  if (trans_f_of_t_name.has_value()) {
+    trans_f_of_t_ = std::move(trans_f_of_t_name.value());
+    f_of_t_names_.insert(trans_f_of_t_.value());
+  }
+}
+
+template <size_t Dim>
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, Dim> RotScaleTrans<Dim>::operator()(
+    const std::array<T, Dim>& source_coords, const double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time) const {
+  std::array<tt::remove_cvref_wrap_t<T>, Dim> result{};
+  for (size_t i = 0; i < Dim; i++) {
+    gsl::at(result, i) = gsl::at(source_coords, i);
+  }
+  const tt::remove_cvref_wrap_t<T> radius = magnitude(result);
+  // Rotation Map
+  if (rot_f_of_t_.has_value()) {
+    const Matrix rot_matrix = rotation_matrix<Dim>(
+        time, *(functions_of_time.at(rot_f_of_t_.value())));
+    for (size_t i = 0; i < Dim; i++) {
+      gsl::at(result, i) = rot_matrix(i, 0) * source_coords[0];
+      for (size_t j = 1; j < Dim; j++) {
+        gsl::at(result, i) += rot_matrix(i, j) * gsl::at(source_coords, j);
+      }
+    }
+  }
+  // Expansion Map
+  if (scale_f_of_t_a_.has_value()) {
+    const double scale_a_of_t =
+        functions_of_time.at(scale_f_of_t_a_.value())->func(time)[0][0];
+    const double scale_b_of_t =
+        functions_of_time.at(scale_f_of_t_b_.value())->func(time)[0][0];
+    if (region_ == BlockRegion::Inner) {
+      for (size_t i = 0; i < Dim; i++) {
+        gsl::at(result, i) *= scale_a_of_t;
+      }
+    } else if (region_ == BlockRegion::Transition) {
+      for (size_t k = 0; k < get_size(radius); k++) {
+        // Optimization from SpEC to reduce roundoff.
+        // Closer to outer radius
+        if (1.0 - get_element(radius, k) / (inner_radius_ + outer_radius_) <
+            .5) {
+          // Expansion falloff factor w_E in the documentation of the form
+          // w_E = \frac{R_{in}(R_{out} - r)(E_{a}(t) - E_{b}(t))}{r(R_{out} -
+          // R_{in})}
+          double radial_scaling_factor =
+              ((outer_radius_ - get_element(radius, k)) *
+               (scale_a_of_t - scale_b_of_t) * inner_radius_) /
+              ((outer_radius_ - inner_radius_) * get_element(radius, k));
+          for (size_t i = 0; i < Dim; i++) {
+            get_element(gsl::at(result, i), k) *=
+                (scale_b_of_t + radial_scaling_factor);
+          }
+          // Closer to inner radius
+        } else {
+          // Expansion falloff factor w_E in the documentation of the form
+          // w_E = \frac{R_{out}(R_{in} - r)(E_{a}(t) - E_{b}(t))}{r(R_{out} -
+          // R_{in})}
+          double radial_scaling_factor =
+              ((inner_radius_ - get_element(radius, k)) *
+               (scale_a_of_t - scale_b_of_t) * outer_radius_) /
+              ((outer_radius_ - inner_radius_) * get_element(radius, k));
+          for (size_t i = 0; i < Dim; i++) {
+            get_element(gsl::at(result, i), k) *=
+                (scale_a_of_t + radial_scaling_factor);
+          }
+        }
+      }
+    } else {
+      for (size_t i = 0; i < Dim; i++) {
+        gsl::at(result, i) *= scale_b_of_t;
+      }
+    }
+  }
+  // Translation map
+  if (trans_f_of_t_.has_value()) {
+    const DataVector trans_func_of_time =
+        functions_of_time.at(trans_f_of_t_.value())->func(time)[0];
+    ASSERT(trans_func_of_time.size() == Dim,
+           "The dimension of the function of time ("
+               << trans_func_of_time.size()
+               << ") does not match the dimension of the map (" << Dim << ").");
+    if (region_ == BlockRegion::Inner) {
+      for (size_t i = 0; i < Dim; i++) {
+        gsl::at(result, i) += gsl::at(trans_func_of_time, i);
+      }
+    } else if (region_ == BlockRegion::Transition) {
+      for (size_t k = 0; k < get_size(radius); k++) {
+        // closer to outer radius
+        if (1.0 - get_element(radius, k) / (inner_radius_ + outer_radius_) <
+            .5) {
+          // Translation falloff factor w_T in the documentation of the
+          // form w_T = (R_{out} - r) / (R_{out} - R_{in})
+          const double radial_translation_factor =
+              (outer_radius_ - get_element(radius, k)) /
+              (outer_radius_ - inner_radius_);
+          for (size_t i = 0; i < Dim; i++) {
+            get_element(gsl::at(result, i), k) +=
+                gsl::at(trans_func_of_time, i) * radial_translation_factor;
+          }
+        } else {
+          // Translation falloff factor w_T in the documentation of the
+          // form w_T = (R_{in} - r) / (R_{out} - R_{in})
+          const double radial_translation_factor =
+              (inner_radius_ - get_element(radius, k)) /
+                  (outer_radius_ - inner_radius_) +
+              1.0;
+          for (size_t i = 0; i < Dim; i++) {
+            get_element(gsl::at(result, i), k) +=
+                gsl::at(trans_func_of_time, i) * radial_translation_factor;
+          }
+        }
+      }
+      ASSERT(max(magnitude(result)) <= outer_radius_ or
+                 equal_within_roundoff(max(magnitude(result)), outer_radius_),
+             "Coordinates translated outside outer radius, this should "
+             "not happen: "
+                 << max(magnitude(result))
+                 << " outer radius: " << outer_radius_);
+    }
+  }
+  return result;
+}
+
+template <size_t Dim>
+std::optional<std::array<double, Dim>> RotScaleTrans<Dim>::inverse(
+    const std::array<double, Dim>& target_coords, double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+        functions_of_time) const {
+  std::array<double, Dim> result{};
+  for (size_t i = 0; i < Dim; i++) {
+    gsl::at(result, i) = gsl::at(target_coords, i);
+  }
+  const double radius = magnitude(result);
+  // This variable is used when taking the inverse of rotation which needs the
+  // source coords after the inverse expansion and translation have been done.
+  std::array<double, Dim> pre_rotation_result = result;
+
+  // Inverse translation without expansion
+  if (trans_f_of_t_.has_value() and not scale_f_of_t_a_.has_value()) {
+    const DataVector trans_func_of_time =
+        functions_of_time.at(trans_f_of_t_.value())->func(time)[0];
+    ASSERT(trans_func_of_time.size() == Dim,
+           "The dimension of the function of time ("
+               << trans_func_of_time.size()
+               << ") does not match the dimension of the map (" << Dim << ").");
+      double non_translated_radius_squared = 0.;
+      for (size_t i = 0; i < Dim; i++) {
+        non_translated_radius_squared +=
+            square(gsl::at(target_coords, i) - gsl::at(trans_func_of_time, i));
+      }
+      if (radius >= outer_radius_) {
+        // no translation applied
+      } else if (non_translated_radius_squared <= square(inner_radius_)) {
+        for (size_t i = 0; i < Dim; i++) {
+          gsl::at(pre_rotation_result, i) -= gsl::at(trans_func_of_time, i);
+        }
+      } else {
+        // We need to solve a quadratic of the form w^2a + wb + c where
+        // a, b, and c change if you're closer to the inner radius. This
+        // is an optimization to reduce roundoff from SpEC.
+        double a = 0.;
+        double b = 0.;
+        double c = 0.;
+        double radial_translation_factor = 0.;
+        for (size_t i = 0; i < Dim; i++) {
+          a += square(gsl::at(trans_func_of_time, i));
+        }
+        a -= square(outer_radius_ - inner_radius_);
+        // When closer to the outer radius the quadratic has the form
+        // a = T(t)^2 - (R_{out} - R_{in})^2,
+        // b = 2(R_{out}(R_{out} - R_{in}) - T(t)\vec{\bar{\xi}}), and
+        // c = \vec{\bar{\xi}}^2 - R_{out}^2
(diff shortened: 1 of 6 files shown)


AlexCarpenter46:
The fixup is to simplify the constructor so that we don't need so many overloads as well as making it easier to implement in the BinaryCompactObject and Sphere domain.

knelli2:
Offline review done with @knelli2 and @AlexCarpenter46. He will post a fixup with the discussed changes.

AlexCarpenter46:
Okay, I think this is ready for another round of reviews! I have a few comments though. For the changes to the local_approx for points farther/close to the $R_{out}E_{b}(t)$ boundary, the jacobian and frame velocity values were sometimes < 1.0 so I was seeing deviation on the order of 10^-9 for jacobian and 10^-8 for frame velocity.

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.hpp
```diff
@@ -104,83 +108,88 @@ namespace domain::CoordinateMaps::TimeDependent {
...
- * 2w(T(t)(T(t) - \vec{\bar{\xi}}) + R_{in}(R_{out} - R_{in})
- * + (T(t) - \vec{\bar{\xi}})^2 - R_{in}^2
+ * w_{T}^2(T(t)^2 - (R_{out} - R_{in})^2) -
+ * 2w_{T}(\vec{T}(t) \cdot (\vec{T}(t) - \vec{\bar{\xi}}) + R_{in}(R_{out} -
+ * R_{in})
+ * + (\vec{T}(t) - \vec{\bar{\xi}})^2 - R_{in}^2
  * \f}
  *
+ * If Expansion and Translation are supplied, then the inverse is given by
+ * equation 4, with no transpose of rotation applied.
```
knelli2: Don't hard code equation numbers because they might change. You can add
```tex
\label{eq:full_inverse}
```
inside the `\f{equation}{..` and then here do `given by Eq. \ref{eq:full_inverse}` and it'll automatically add in the equation numbers. Even if they change.

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.hpp
```diff
@@ -85,17 +88,18 @@ namespace domain::CoordinateMaps::TimeDependent {
...
  * \vec{\xi} = \left\{\begin{array}{ll}R^{T}(t)(\frac{(\vec{\bar{\xi}} -
- * \vec{T}(t))}{E_{a}(t)}), & |\vec{\xi}| \leq R_{in}E_{a}(t),
- * \\ R^{T}(t)\frac{\vec{\bar{\xi}} - w_{T}T(t)}{w_{E}}, & R_{in} < |\vec{\xi}|
- * < 0.5 R_{out}, \\ R^{T}(t)\frac{\vec{\bar{\xi}} - (1.0 - w_{T})T(t)}{w_{E}},
- * & 0.5 R_{out} \leq |\vec{\xi}| < R_{out},
- * \\ R^{T}(t)\frac{\vec{\bar{\xi}}}{E_{b}(t)}, & |\vec{\xi}| \geq R_{out}
- * \end{array}\right.
+ * \vec{T}(t))}{E_{a}(t)}), & |\vec{\bar{\xi}}| \leq R_{in}E_{a}(t),
+ * \\ R^{T}(t)\frac{\vec{\bar{\xi}} - w_{T}\vec{T}(t)}{w_{E}},
+ * & R_{in} < |\vec{\bar{\xi}}| < 0.5 R_{out}E_{b}(t),
```
knelli2: Inner bound should be $R_{in}(t)E_a(t)$?

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.hpp
```diff
@@ -191,50 +200,51 @@ namespace domain::CoordinateMaps::TimeDependent {
...
- * \vec{v} = \left\{\begin{array}{ll}\vec{\xi}(dR(t)E_{a}(t) +
- * R(t)dE_{a}(t)) + \vec{dT}(t), & |\vec{\xi}| \leq R_{in},
- * \\ \vec{\xi}(dR(t)(E_{a}(t) + w_{E}) + R(t)(dE_{a}(t) + dw_{E})) + T(t) +
- * w_{T}T(t), & R_{in} < |\vec{\xi}| < 0.9 R_{out},
- * \\ \vec{\xi}(dR(t)(E_{b}(t) + w_{E}) + R(t)(dE_{b}(t) + dw_{E})) + w_{T}T(t),
- * & 0.9 R_{out} \leq |\vec{\xi}| < R_{out}, \\ \vec{\xi}(dR(t)E_{b}(t) +
- * R(t)dE_{b}(t)), & |\vec{\xi}| \geq R_{out} \end{array}\right.
+ * \vec{v} = \left\{\begin{array}{ll}(E_{a}(t)dR(t) +
+ * dE_{a}(t)R(t))\vec{\xi} + d \vec{T}(t), & |\vec{\xi}| \leq R_{in},
+ * \\ ((E_{a}(t)dR(t) + w_{E})dR(t) + (dE_{a}(t) + dw_{E})R(t))\vec{\xi} + (1 +
```
knelli2: Extra $dR(t)$ here?

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
```diff
@@ -156,31 +154,31 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> RotScaleTrans<Dim>::operator()(
...
-        if (get_element(radius, k) <= inner_radius_.value()) {
+        if (get_element(radius, k) <= inner_radius_) {
           for (size_t i = 0; i < Dim; i++) {
             get_element(gsl::at(result, i), k) +=
                 gsl::at(trans_func_of_time, i);
           }
-          ASSERT(max(magnitude(result)) <= outer_radius_.value(),
+          ASSERT(max(magnitude(result)) <= outer_radius_,
                  "Coordinates translated outside outer radius, this should "
                  "not happen");
```
knelli2: Move the `ASSERT`s out of the `for` loop

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
```diff
@@ -1032,18 +993,17 @@ double RotScaleTrans<Dim>::root_helper(
   } else if (roots.value()[1] >= 0 and roots.value()[1] <= 1.0) {
     return roots.value()[1];
   } else if ((roots.value()[0] > 1.0 and
-              roots.value()[0] <=
-                  std::numeric_limits<double>::epsilon() + 1.0) or
+              equal_within_roundoff(roots.value()[0], 1.0)) or
              (roots.value()[1] > 1.0 and
```
knelli2: You actually don't need the `roots.value()[0] > 1.0` check because above you already return if it's `<= 1.0` so it's guaranteed to be `> 1.0` here. Same thing for comparing to `0.0` below in the next check.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
```diff
@@ -165,77 +177,56 @@ void test_RotScaleTrans() {
...
-        deriv_translation);
-    CHECK_ITERABLE_APPROX(
-        rot_trans_map_rigid.frame_velocity(point_xi, t, f_of_t_list),
-        expected_rotation_deriv + deriv_translation);
-    CHECK_ITERABLE_APPROX(
-        scale_trans_map_rigid.frame_velocity(point_xi, t, f_of_t_list),
-        point_xi * (scale_a + d_scale_a) + deriv_translation);
     // Inverse
     CHECK_ITERABLE_APPROX(
         rot_map.inverse(expected_rotation, t, f_of_t_list).value(), point_xi);
```
knelli2: I think you can get rid of all the inverse tests and just use `test_inverse_map` from `TestMapHelpers`. As long as the `operator()` is tested separately like you have above, then `test_inverse_map` should be sufficient. That'll hopefully make the test shorter

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
```diff
@@ -556,26 +462,34 @@ void test_RotScaleTrans() {
...
+      test_frame_velocity(rot_scale_map_non_rigid, point_xi, t, f_of_t_list);
+      test_frame_velocity(rot_trans_map_rigid, point_xi, t, f_of_t_list);
+      test_frame_velocity(rot_trans_map_non_rigid, point_xi, t, f_of_t_list);
+      test_frame_velocity(scale_trans_map_rigid, point_xi, t, f_of_t_list);
+      test_frame_velocity(scale_trans_map_non_rigid, point_xi, t, f_of_t_list);
+      test_frame_velocity(rot_scale_trans_map_rigid, point_xi, t, f_of_t_list);
+      test_frame_velocity(rot_scale_trans_map_non_rigid, point_xi, t,
+                          f_of_t_list);
+      // jacobian/inv jacobian
       test_jacobian(rot_map, point_xi, t, f_of_t_list);
```
knelli2: You repeat this block of tests several times. You could use a lambda to condense this

```cpp
const auto check_all_maps = [&](const auto& point_to_check) {
  test_frame_velocity(rot_map, point_to_check, t, f_of_t_list);
  test_frame_velocity(...);
  ...
  test_jacobian(...);
  test_inv_jacobian(...);
  ...
  };
```

And then just call it below like

```cpp
check_all_maps(point_xi);
...
check_all_maps(point_xi_dv);
...
check_all_maps(far_point_xi);
...
```

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
```diff
@@ -136,7 +136,7 @@ void test_jacobian(
         return map(point, time, functions_of_time);
       };
   // Our default approx value is too stringent for this test
-  Approx local_approx = Approx::custom().epsilon(1e-10).scale(1.0);
+  Approx local_approx = Approx::custom().epsilon(1e-9).scale(1.0);
```
knelli2: I don't think we should change these `approx`s. If the jacobian/frame velocity isn't hitting 1e-10, we should probably investigate that. Try making the final time of the test closer to the initial time. The quaternion FoT has trouble when it doesn't have updates and needs to integrate far into the future

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.hpp
```diff
@@ -28,52 +28,55 @@ namespace domain::CoordinateMaps::TimeDependent {
...
- * w_{T}T(t), & R_{in} < |\vec{\xi}| < 0.9 R_{out}, \\ R(t)(w_{E} +
- * E_{b}(t))\vec{\xi} + w_{T}T(t), & 0.9 R_{out} < |\vec{\xi}| < R_{out},
- * \\ R(t)E_{b}(t)\vec{\xi}, &
- * |\vec{\xi}| \geq R_{out} \end{array}\right.
+ * \vec{\bar{\xi}} = \left\{\begin{array}{ll}E_{a}(t)R(t)\vec{\xi} + \vec{T}(t),
+ * & |\vec{\xi}| \leq R_{in}, \\ (w_{E} + E_{a}(t))R(t)\vec{\xi} + (1 +
+ * w_{T})\vec{T}(t), & R_{in} < |\vec{\xi}| \leq 0.9 R_{out}, \\ (w_{E} +
+ * E_{b}(t))R(t)\vec{\xi} + w_{T}\vec{T}(t), & 0.9 R_{out} < |\vec{\xi}| <
+ * R_{out}, \\ E_{b}(t)R(t)\vec{\xi}, & |\vec{\xi}| \geq R_{out}
+ * \end{array}\right.
```
knelli2: [Discussed offline] Change all the factors to be $0.5 (R_{in} + R_{out})$ so we are guaranteed to be within the region. Using $0.5 R_{out}$ doesn't account for where the inner radius is.

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
```diff
@@ -1032,18 +993,17 @@ double RotScaleTrans<Dim>::root_helper(
...
     return 1.0;
   } else if ((roots.value()[0] < 0.0 and
-              roots.value()[0] > -std::numeric_limits<double>::epsilon()) or
+              equal_within_roundoff(roots.value()[0], 0.0)) or
              (roots.value()[1] < 0.0 and
-              roots.value()[1] > -std::numeric_limits<double>::epsilon())) {
+              equal_within_roundoff(roots.value()[0], 0.0))) {
     return 0.0;
   }
+  ASSERT(false, "Root helper couldn't find correct root");
```
knelli2: Make this an `ERROR` please. If we get here, something is going very wrong and returning zero as the root in production mode won't help us debug it.

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
```diff
@@ -284,45 +281,44 @@ std::optional<std::array<double, Dim>> RotScaleTrans<Dim>::inverse(
...
         functions_of_time.at(scale_f_of_t_a_.value())->func(time)[0][0];
     const double scale_b_of_t =
         functions_of_time.at(scale_f_of_t_b_.value())->func(time)[0][0];
     // if not rigid expansion do piecewise expansion
     if (not rigid_) {
-      if (radius >= scale_b_of_t * outer_radius_.value()) {
+      if (radius >= scale_b_of_t * outer_radius_) {
         for (size_t i = 0; i < Dim; i++) {
-          gsl::at(result, i) /= scale_b_of_t;
+          gsl::at(pre_rotation_result, i) /= scale_b_of_t;
```
knelli2: Add an `ASSERT` here like you did below.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
```diff
@@ -94,70 +94,85 @@ void test_RotScaleTrans() {
   const std::pair<std::string, std::string> scale_pair{"expansion_a",
                                                        "expansion_b"};
   // Rotation, Scaling, Translation Non-Rigid
-  const domain::CoordinateMaps::TimeDependent::RotScaleTrans<Dim>
+  domain::CoordinateMaps::TimeDependent::RotScaleTrans<Dim>
       rot_scale_trans_map_non_rigid{scale_pair,    "rotation_angle",
                                     "translation", inner_radius,
                                     outer_radius,  false};
+  rot_scale_trans_map_non_rigid =
```
knelli2: Instead of overwriting the variable, just serialize/deserialize immediately:

```cpp
  const domain::CoordinateMaps::TimeDependent::RotScaleTrans<Dim>
     serialize_and_deserialize(rot_scale_trans_map_non_rigid{scale_pair,    "rotation_angle",
                                    "translation", inner_radius,
                                    outer_radius,  false});
```
Same for the rest as well.

knelli2 (changes requested):
This looks really good! Thanks for doing this! Just a few more things. I also noticed in a lot of places in the `cpp` file that you have parentheses around things that might not need them, like `a * (b)` or `c + (a * b)`. Just take a look and see if you can simplify things in a few places. Too many parentheses can get a little confusing.

Please squash all the current fixups into one commit and add new fixups for these next changes.

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.hpp
```diff
@@ -39,11 +39,11 @@ namespace domain::CoordinateMaps::TimeDependent {
...
  * expanded by the outer expansion function of time $E_{b}(t)$. This map assumes
- * that the center of the map is at (0., 0., 0.). If translation or expansion is
- * specified, then an inner and outer radius must be specified as well. There is
- * a rigid flag which when set to true treats the coordinates as if they're
- * within the inner radius. This can be used as an optimization if you know
- * which blocks in the domain will always be within the inner radius.
+ * that the center of the map is at (0., 0., 0.). There is an enum class to set
+ * which region your block is in. Specifying Inner treats coordinates as if
+ * they're inside the inner radius, transition treats the points as if they're
+ * between the inner and outer radius. Outer treats points as if they're on or
```
knelli2: Make Inner be `RotScaleTrans::BlockRegion::Inner`. Same for Transition and Outer

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
```diff
@@ -150,50 +145,43 @@ std::array<tt::remove_cvref_wrap_t<T>, Dim> RotScaleTrans<Dim>::operator()(
                << trans_func_of_time.size()
                << ") does not match the dimension of the map (" << Dim << ").");
     // if not rigid translation do piecwise translation
```
knelli2: Remove comment? Or change update it

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
```diff
@@ -281,94 +267,78 @@ std::optional<std::array<double, Dim>> RotScaleTrans<Dim>::inverse(
...
+      if (1.0 - radius / (inner_radius_ * scale_a_of_t +
+                          outer_radius_ * scale_b_of_t) <
+          .5) {
+        radial_scaling_factor = scale_b_of_t * outer_radius_;
+        b = 2.0 * radial_scaling_factor *
+            (scale_a_of_t * inner_radius_ - radial_scaling_factor);
+        c = square(scale_b_of_t * outer_radius_) - radius_squared;
+        std::optional<std::array<double, 2>> roots = real_roots(a, b, c);
+        radial_scaling_factor = root_helper(roots);
+        radial_scaling_factor =
```
knelli2: You reuse `radial_scaling_factor` to represent different intermediate quantities. This can be somewhat confusing. Could you make some intermediate variable names for these instead and only set `raidal_scaling_factor` at the very end? Same for the `else` part of this `if`.

src/Domain/CoordinateMaps/TimeDependent/RotScaleTrans.cpp
```diff
@@ -981,18 +884,16 @@ double RotScaleTrans<Dim>::root_helper(
...
+
+             equal_within_roundoff(roots.value()[1], 1.0)) {
     return 1.0;
-  } else if ((roots.value()[0] < 0.0 and
-              equal_within_roundoff(roots.value()[0], 0.0)) or
-             (roots.value()[1] < 0.0 and
-              equal_within_roundoff(roots.value()[0], 0.0))) {
+  } else if (equal_within_roundoff(roots.value()[0], 0.0) or
+
+             equal_within_roundoff(roots.value()[0], 0.0)) {
```
knelli2: Should be `roots.value()[1]` here?

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
```diff
@@ -93,89 +96,121 @@ void test_RotScaleTrans() {
...
+  domain::CoordinateMaps::TimeDependent::RotScaleTrans<Dim> trans_map_outer =
+      map_serialize_and_deserialize(std::nullopt, std::nullopt, "translation",
+                                    BlockRegion::Outer);
+
+  double far_double_1 = 0;
+  double far_double_2 = 0;
+  if (Dim == 2) {
+    far_double_1 = 30.0;
+    far_double_2 = 35.36;
+  } else {
```
knelli2: Could you explain why this is necessary? I'm not sure I understand.

Also, if we keep it, these can be one-liners
```cpp
const double far_double_1 = Dim == 2 ? 30.0 : 25.0;
const double far_double_2 = Dim == 2 ? 35.36 : 28.87;
```

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
```diff
@@ -214,427 +250,288 @@ void test_RotScaleTrans() {
...
+      test_inv_jacobian(rot_scale_trans_map_transition, point_to_check, t,
+                        f_of_t_list);
+      test_jacobian(rot_scale_trans_map_outer, point_to_check, t, f_of_t_list);
+      test_inv_jacobian(rot_scale_trans_map_outer, point_to_check, t,
                         f_of_t_list);
+    };
+
+    if (radius <= inner_radius * .99) {
+      check_inner_maps(point_xi);
+    } else if (radius >= inner_radius * 1.01) {
```
knelli2: Why the buffer? I thought that was only needed for the jacobian tests. Also, isn't `radius >= inner_radius * 1.01` excluding the specific part of the map we want to test the most? Where it goes bad because of roundoff?

knelli2 (changes requested):
Only a few more things! Please put these changes into a separate fixup commit

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
AlexCarpenter46: This was a mistake, I meant to take this out since we added the enum class. Now the jacobian and operator will always do the same thing regardless of what region the radius is actually in for the tests since the enum determines which region you're in.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
AlexCarpenter46: I wanted points that were pretty close to the outer boundary, but not too far beyond. This helped fix the roundoff errors for the jacobian tests. After some more testing though, the frame velocity still has trouble reaching 1e-10 super consistently though when rotation is specified so the quaternion is still causing some trouble.

knelli2: But the outer radius is `50` from above?

AlexCarpenter46: Oh maybe my variable naming is confusing, these numbers go in as the far distribution for far_point_xi. So it gives far_point_xi 2 numbers between 30 and 35 in 2D so that when you take the magnitude of it it's pretty close to 50 and similar for 3D

knelli2: Ohhhh. Ok I got it. If you want to have more control over the radius, I suggest using a random distribution from 0 to 50 (or a bit outside) for the radius, and a random dist for the azimuthal angle. Then for 3D you can have another random dist for the polar angle. That way you don't have to guess for the x/y/z components..

tests/Unit/Helpers/Domain/CoordinateMaps/TestMapHelpers.hpp
AlexCarpenter46: Jacobian tests were failing with accuracy of 1e-9 near the outer boundary. Did a test to make sure that the numbers matched with spec this test can be found on Ocean in /home/alexcarpenter/spectre_spec_comparison.cpp the numbers agree to 1e-16 between spec and spectre so it seems safe to increase this tolerance

knelli2: Yes I agree it's safe to increase the tolerance, but only for this test. Your current change will increase it for all domain tests. Instead, in a new commit, edit these functions to take an `approx` as an argument and default it to 1e-10 (the current value). Then for your test, you can pass in a 1e-9 instead.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
```diff
@@ -455,19 +447,21 @@ void test_RotScaleTrans() {
                         f_of_t_list);
     };
 
-    if (radius <= inner_radius * .99) {
-      check_inner_maps(point_xi);
-    } else if (radius >= inner_radius * 1.01) {
-      check_transition_maps(point_xi);
+    if (radius <= inner_radius) {
+      check_inner_maps_inverse(point_xi);
+    } else if (radius >= inner_radius) {
```
knelli2: This can just be `else`.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_RotScaleTrans.cpp
```diff
@@ -455,19 +447,21 @@ void test_RotScaleTrans() {
...
     check_all_maps_jacobian(point_xi_dv);
 
-    if (far_radius <= outer_radius * .99) {
-      check_transition_maps(far_point_xi);
-    } else if (far_radius >= outer_radius * 1.01) {
-      check_outer_maps(far_point_xi);
+    if (far_radius <= outer_radius) {
+      check_transition_maps_inverse(far_point_xi);
+    } else if (far_radius >= outer_radius) {
+      check_outer_maps_inverse(far_point_xi);
```
knelli2: Same here.

knelli2 (commented):
Ok everything looks good! These last couple things can just be squashed in. Please make the changes to `TestMapHelpers` into its own commit. Everything else can go into another commit. Excellent work!

knelli2 (approved):

PR #5697, opened 2024-01-09
Add Klein-Gordon CCE evolution component, metavariable, and executable 

Sizheng-Ma:
## Proposed changes

This PR adds a metavariable, executable, and evolution component for the Klein-Gordon CCE system. For the moment, the evolution component is almost empty. Actions will be added in the following PRs.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/Cce/CMakeLists.txt
+++ b/src/Evolution/Executables/Cce/CMakeLists.txt
@@ -16,11 +16,11 @@ set(LIBS_TO_LINK
   Utilities
   )
 
-function(add_cce_executable EXECUTABLE BOUNDARY_COMPONENT)
+function(add_cce_executable EXECUTABLE BOUNDARY_COMPONENT SYSTEM)
   add_spectre_executable(
     ${EXECUTABLE}
     EXCLUDE_FROM_ALL
-    CharacteristicExtract.cpp
+    ${SYSTEM}.cpp
   )
   target_compile_definitions(
     ${EXECUTABLE}
@@ -33,8 +33,15 @@ endfunction()
 add_cce_executable(
   CharacteristicExtract
   Cce::H5WorldtubeBoundary
+  CharacteristicExtract
   )
 add_cce_executable(
   AnalyticTestCharacteristicExtract
   Cce::AnalyticWorldtubeBoundary
+  CharacteristicExtract
+  )
+add_cce_executable(
+  KleinGordonCharacteristicExtract
+  Cce::KleinGordonH5WorldtubeBoundary
+  KleinGordonCharacteristicExtract
   )
--- a/src/Evolution/Executables/Cce/KleinGordonCharacteristicExtract.cpp
+++ b/src/Evolution/Executables/Cce/KleinGordonCharacteristicExtract.cpp
@@ -1,0 +1,41 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Executables/Cce/KleinGordonCharacteristicExtract.hpp"
+
+#include <vector>
+
+#include "Evolution/Systems/Cce/Initialize/RegisterInitializeJWithCharm.hpp"
+#include "Evolution/Systems/Cce/WorldtubeBufferUpdater.hpp"
+#include "Evolution/Systems/Cce/WorldtubeDataManager.hpp"
+#include "NumericalAlgorithms/Interpolation/BarycentricRationalSpanInterpolator.hpp"
+#include "NumericalAlgorithms/Interpolation/CubicSpanInterpolator.hpp"
+#include "NumericalAlgorithms/Interpolation/LinearSpanInterpolator.hpp"
+#include "NumericalAlgorithms/Interpolation/SpanInterpolator.hpp"
+#include "Parallel/CharmMain.tpp"
+#include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
+
+// Parameters chosen in CMakeLists.txt
+using metavariables = EvolutionMetavars<BOUNDARY_COMPONENT>;
+
+extern "C" void CkRegisterMainModule() {
+  Parallel::charmxx::register_main_module<metavariables>();
+  Parallel::charmxx::register_init_node_and_proc(
+      {&Cce::register_initialize_j_with_charm<
+           metavariables::evolve_ccm, metavariables::cce_boundary_component>,
+       &register_derived_classes_with_charm<
+           Cce::WorldtubeBufferUpdater<Cce::cce_metric_input_tags>>,
+       &register_derived_classes_with_charm<
+           Cce::WorldtubeBufferUpdater<Cce::cce_bondi_input_tags>>,
+       &register_derived_classes_with_charm<
+           Cce::WorldtubeBufferUpdater<Cce::klein_gordon_input_tags>>,
+       &register_derived_classes_with_charm<Cce::WorldtubeDataManager<
+           Cce::Tags::characteristic_worldtube_boundary_tags<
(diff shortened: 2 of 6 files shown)


src/Evolution/Systems/Cce/CMakeLists.txt
```diff
@@ -47,6 +47,7 @@ spectre_target_headers(
   SpecBoundaryData.hpp
   SwshDerivatives.hpp
   System.hpp
+  KleinGordonSystem.hpp
```
knelli2, line 50: It seems like this was forgotten in a previous PR, so move this to it's own commit

src/Evolution/Systems/Cce/Components/KleinGordonCharacteristicEvolution.hpp
```diff
@@ -0,0 +1,78 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <type_traits>
+
+#include "Evolution/Systems/Cce/Components/CharacteristicEvolution.hpp"
+#include "Evolution/Systems/Cce/KleinGordonSystem.hpp"
```
knelli2, line 9:
```cpp
#include "Parallel/GlobalCache.hpp"
#include "Parallel/Local.hpp"
#include "Parallel/Phase.hpp"
#include "Time/Actions/SelfStartActions.hpp"
#include "Utilities/TMPL.hpp"
```

src/Evolution/Executables/Cce/CMakeLists.txt
```diff
@@ -16,11 +16,11 @@ set(LIBS_TO_LINK
   Utilities
   )
 
-function(add_cce_executable EXECUTABLE BOUNDARY_COMPONENT)
+function(add_cce_executable EXECUTABLE BOUNDARY_COMPONENT SYSTEM)
   add_spectre_executable(
     ${EXECUTABLE}
     EXCLUDE_FROM_ALL
-    CharacteristicExtract.cpp
+    ${SYSTEM}
```
knelli2: Maybe do
```cpp
${SYSTEM}.cpp
```

so you don't always have to add the `.cpp` in the function calls below.

src/Evolution/Executables/Cce/KleinGordonCharacteristicExtract.cpp
```diff
@@ -0,0 +1,39 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Executables/Cce/KleinGordonCharacteristicExtract.hpp"
+
+#include <vector>
+
+#include "Evolution/Systems/Cce/Initialize/RegisterInitializeJWithCharm.hpp"
```
knelli2, line 8:
```cpp
#include "Evolution/Systems/Cce/WorldtubeBufferUpdater.hpp"
#include "Evolution/Systems/Cce/WorldtubeDataManager.hpp"
```

src/Evolution/Executables/Cce/KleinGordonCharacteristicExtract.hpp
```diff
@@ -0,0 +1,105 @@
...
+#include "Time/Triggers/TimeTriggers.hpp"
+
+/// \cond
+namespace PUP {
+class er;
+}  // namespace PUP
+/// \endcond
+
+template <template <typename> class BoundaryComponent>
+struct EvolutionMetavars : CharacteristicExtractDefaults<false> {
```
knelli2, line 36: Have you tried compiling this executable yet?

knelli2 (changes requested):

src/Evolution/Systems/Cce/Components/KleinGordonCharacteristicEvolution.hpp
Sizheng-Ma: Out of curiosity, these header files are already included in `Evolution/Systems/Cce/Components/CharacteristicEvolution.hpp`. I think it should be fine to omit them here, don't you think?

src/Evolution/Executables/Cce/KleinGordonCharacteristicExtract.hpp
Sizheng-Ma: Yes, it works. But thinking more carefully about it, I feel it's more convenient to include all the actions of `CharacteristicEvolution<Metavariables>` in this commit, since the GR (tensor) part of the two executables should be the same. Actions for the scalar part will be added later. I've just pushed one more commit that does this. For the moment, the executable `KleinGordonCharacteristicEvolution` does the same thing as `CharacteristicEvolution`, except that `Actions::RequestBoundaryData`, `Actions::ReceiveWorldtubeData`, `Actions::RequestNextBoundaryData` are not added, because they are not ready for the `KleinGordonH5WorldtubeBoundary` boundary component. If you are happy with this change, I will squash it with the second commit.

src/Evolution/Systems/Cce/Components/KleinGordonCharacteristicEvolution.hpp
knelli2: Technically, yes, it's ok to omit them here because they are included in the other `hpp`. However, if somebody in the future decided to remove one of them from `CharacteristicEvolution.hpp`, then this file would fail to compile and somebody would have to spend time trying to figure out why.

So as a rule of thumb, we just always include all files that we need. Even if they may have been included via another file.

src/Evolution/Systems/Cce/Components/KleinGordonCharacteristicEvolution.hpp
```diff
@@ -51,14 +51,65 @@ struct KleinGordonCharacteristicEvolution
...
+      // note that the initialization will only actually happen on the
+      // iterations immediately following restarts
+      Actions::InitializeFirstHypersurface<
+          evolve_ccm, typename Metavariables::cce_boundary_component>,
+      tmpl::conditional_t<
+          tt::is_a_v<AnalyticWorldtubeBoundary,
+                     typename Metavariables::cce_boundary_component>,
+          Actions::UpdateGauge<false>, Actions::UpdateGauge<evolve_ccm>>,
+      Actions::PrecomputeGlobalCceDependencies,
+      tmpl::conditional_t<evolve_ccm,
```
knelli2, line 73: Do you plan on ever doing a KG CCM system?

knelli2 (commented):
Just one question

src/Evolution/Systems/Cce/Components/KleinGordonCharacteristicEvolution.hpp
Sizheng-Ma: I don't have a plan for the next few months, but I can see two things that could motivate the KG CCM system. One is to simulate ringdown tails. The tail in the tensor sector (i.e., strain and $\Psi_4$) decays faster than the one in the scalar field. Therefore, digging out tails from the scalar field should be much easier than the tensor fields. If it turns out the SpEC CCM has a hard time doing the tensor tail, then I may have to switch to the scalar field first. 

The other one is related to Niko's worldtube projection. I have been talking with him about using this KG CCE system to extract GWs at future null infinity. But the usefulness of the CCM remains unclear.

So I feel it's good to keep those options in case they are needed.

knelli2: Ok that sounds like  good plan. Let's keep these compile-time switches in here, but they probably won't be used for a while.

You can squash your commits.

Sizheng-Ma:
Done!

knelli2 (approved):

knelli2:
Test failure unrelated

PR #5703, opened 2024-01-17
Add SettleToConstantQuaternion

geoffrey4444:
## Proposed changes

Add new function of time like SettleToConstant but for quaternions; the function of time always returned a unit quaternion (i.e., a rotation).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
+++ b/src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
@@ -1,0 +1,140 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/FunctionsOfTime/SettleToConstantQuaternion.hpp"
+
+#include <cmath>
+#include <limits>
+#include <memory>
+
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeArray.hpp"
+#include "Utilities/StdArrayHelpers.hpp"
+
+namespace {
+template <size_t MaxDerivReturned>
+std::array<double, MaxDerivReturned + 1> quat_norm_and_derivs(
+    const std::array<DataVector, MaxDerivReturned + 1>&
+        unnormalized_func_and_derivs) {
+  std::array<double, MaxDerivReturned + 1> result =
+      make_array<MaxDerivReturned + 1>(0.0);
+  gsl::at(result, 0) = norm(gsl::at(unnormalized_func_and_derivs, 0));
+  if constexpr (MaxDerivReturned > 0) {
+    gsl::at(result, 1) = sum(gsl::at(unnormalized_func_and_derivs, 0) *
+                             gsl::at(unnormalized_func_and_derivs, 1)) /
+                         gsl::at(result, 0);
+    if constexpr (MaxDerivReturned > 1) {
+      gsl::at(result, 2) =
+          sum(square(gsl::at(unnormalized_func_and_derivs, 1)) +
+              gsl::at(unnormalized_func_and_derivs, 0) *
+                  gsl::at(unnormalized_func_and_derivs, 2)) /
+          gsl::at(result, 0);
+    }
+  }
+  return result;
+}
+}  // namespace
+
+namespace domain::FunctionsOfTime {
+SettleToConstantQuaternion::SettleToConstantQuaternion(
+    const std::array<DataVector, 3>& initial_func_and_derivs,
+    const double match_time, const double decay_time)
+    : unnormalized_function_of_time_(initial_func_and_derivs, match_time,
+                                     decay_time),
+      match_time_(match_time),
+      decay_time_(decay_time) {
+  if (initial_func_and_derivs[0].size() != 4) {
+    ERROR(
+        "SettleToConstantQuaternion requires the initial function and its time "
+        "derivatives to be quaternions stored as DataVectors of size 4, not "
+        "size "
+        << initial_func_and_derivs[0].size());
+  }
+  if (not equal_within_roundoff(
+          gsl::at(quat_norm_and_derivs<2>(initial_func_and_derivs), 0), 1.0)) {
+    ERROR(
+        "SettleToConstantQuaternion requires that the initial quaternion "
+        "should be a quaternion with a norm of 1.0, not "
+        << gsl::at(quat_norm_and_derivs<2>(initial_func_and_derivs), 0));
+  }
+}
+
+std::unique_ptr<FunctionOfTime> SettleToConstantQuaternion::get_clone() const {
+  return std::make_unique<SettleToConstantQuaternion>(*this);
+}
+
+template <size_t MaxDerivReturned>
+std::array<DataVector, MaxDerivReturned + 1>
+SettleToConstantQuaternion::func_and_derivs(const double t) const {
+  static_assert(MaxDerivReturned < 3, "The maximum available derivative is 2.");
+
+  std::array<DataVector, MaxDerivReturned + 1> result{};
+  if constexpr (MaxDerivReturned == 0) {
+    result = unnormalized_function_of_time_.func(t);
+  } else if constexpr (MaxDerivReturned == 1) {
+    result = unnormalized_function_of_time_.func_and_deriv(t);
+  } else if constexpr (MaxDerivReturned == 2) {
+    result = unnormalized_function_of_time_.func_and_2_derivs(t);
+  }
+  std::array<double, MaxDerivReturned + 1> norm_and_derivs =
+      quat_norm_and_derivs<MaxDerivReturned>(result);
+  result /= gsl::at(norm_and_derivs, 0);
+  if constexpr (MaxDerivReturned > 0) {
+    gsl::at(result, 1) -= gsl::at(result, 0) * gsl::at(norm_and_derivs, 1) /
+                          gsl::at(norm_and_derivs, 0);
+    if constexpr (MaxDerivReturned > 1) {
+      gsl::at(result, 2) -= gsl::at(result, 1) * 2.0 *
+                            gsl::at(norm_and_derivs, 1) /
+                            gsl::at(norm_and_derivs, 0);
+      gsl::at(result, 2) -= gsl::at(result, 0) * gsl::at(norm_and_derivs, 2) /
+                            gsl::at(norm_and_derivs, 0);
+    }
+  }
+
+  return result;
+}
+
+void SettleToConstantQuaternion::pup(PUP::er& p) {
+  FunctionOfTime::pup(p);
+  size_t version = 0;
+  p | version;
+  // Remember to increment the version number when making changes to this
+  // function. Retain support for unpacking data written by previous versions
+  // whenever possible. See `Domain` docs for details.
+  if (version >= 0) {
+    p | unnormalized_function_of_time_;
+    p | match_time_;
+    p | decay_time_;
+  }
+}
+
+bool operator==(const SettleToConstantQuaternion& lhs,
+                const SettleToConstantQuaternion& rhs) {
+  return lhs.unnormalized_function_of_time_ ==
+             rhs.unnormalized_function_of_time_ and
+         lhs.match_time_ == rhs.match_time_ and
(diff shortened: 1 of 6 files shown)


src/Domain/FunctionsOfTime/SettleToConstantQuaternion.hpp
```diff
@@ -0,0 +1,94 @@
...
+  friend bool operator==(const SettleToConstantQuaternion& lhs,
+                         const SettleToConstantQuaternion& rhs);
+
+  template <size_t MaxDerivReturned = 2>
+  std::array<DataVector, MaxDerivReturned + 1> func_and_derivs(double t) const;
+
+  SettleToConstant unnormalized_function_of_time_;
+  double match_time_{std::numeric_limits<double>::signaling_NaN()};
+  double decay_time_{std::numeric_limits<double>::signaling_NaN()};
+  size_t size_;
```
knelli2: I don't think `size_` is necessary to store. It's only used in two places so I think it can just be hardcoded. If you want to keep it, you can make it a `static constexpr`

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,136 @@
...
+                  gsl::at(unnormalized_func_and_derivs, 2)) /
+          gsl::at(result, 0);
+    }
+  }
+  return result;
+}
+};  // namespace
+
+namespace domain::FunctionsOfTime {
+SettleToConstantQuaternion::SettleToConstantQuaternion(
```
knelli2, line 42: Since you say in the docs that the input quaternion must be unit, you should probably check that here in the constructor. You can probably just call `quat_norm_and_derivs()` on the `initial_func_and_derivs` and check if both are `equal_within_roundoff`.

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,136 @@
...
+    const std::array<DataVector, 3>& initial_func_and_derivs,
+    const double match_time, const double decay_time)
+    : unnormalized_function_of_time_(initial_func_and_derivs, match_time,
+                                     decay_time),
+      match_time_(match_time),
+      decay_time_(decay_time),
+      size_(initial_func_and_derivs[0].size()) {
+  if (size_ != 4) {
+    ERROR(
```
knelli2: This is a good check, but I don't think you need a member variable. Just check 
```cpp
if (initial_func_and_derivs[0].size() != 4) { ...
```

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,136 @@
...
+  return std::make_unique<SettleToConstantQuaternion>(*this);
+}
+
+template <size_t MaxDerivReturned>
+std::array<DataVector, MaxDerivReturned + 1>
+SettleToConstantQuaternion::func_and_derivs(const double t) const {
+  static_assert(MaxDerivReturned < 3, "The maximum available derivative is 2.");
+
+  std::array<DataVector, MaxDerivReturned + 1> result =
+      make_array<MaxDerivReturned + 1>(DataVector(size_, 0.0));
```
knelli2: I think you can just default construct this. Below, you're just overwriting the zeros you initialized here.

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,136 @@
...
+
+namespace {
+template <size_t MaxDerivReturned>
+std::array<double, MaxDerivReturned + 1> quat_norm_and_derivs(
+    const std::array<DataVector, MaxDerivReturned + 1>&
+        unnormalized_func_and_derivs) {
+  std::array<double, MaxDerivReturned + 1> result =
+      make_array<MaxDerivReturned + 1>(0.0);
+  gsl::at(result, 0) =
+      sqrt(sum(square(gsl::at(unnormalized_func_and_derivs, 0))));
```
knelli2: This can just be
```cpp
gsl::at(result, 0) = norm(gsl::at(unnormalized_func_and_derivs, 0));
```

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,136 @@
...
+  if constexpr (MaxDerivReturned == 0) {
+    result = unnormalized_function_of_time_.func(t);
+  } else if constexpr (MaxDerivReturned == 1) {
+    result = unnormalized_function_of_time_.func_and_deriv(t);
+  } else if constexpr (MaxDerivReturned == 2) {
+    result = unnormalized_function_of_time_.func_and_2_derivs(t);
+  }
+  std::array<double, MaxDerivReturned + 1> norm_and_derivs =
+      quat_norm_and_derivs<MaxDerivReturned>(result);
+  gsl::at(result, 0) /= gsl::at(norm_and_derivs, 0);
```
knelli2: [optional] You can avoid repeating the divide by including `#include "Utilities/StdArrayHelpers.hpp` doing
```cpp
result /= gsl::at(norm_and_derivs, 0);
```
This will divide everything at once.

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,136 @@
...
+  FunctionOfTime::pup(p);
+  size_t version = 0;
+  p | version;
+  // Remember to increment the version number when making changes to this
+  // function. Retain support for unpacking data written by previous versions
+  // whenever possible. See `Domain` docs for details.
+  if (version >= 0) {
+    p | unnormalized_function_of_time_;
+    p | match_time_;
+    p | decay_time_;
```
knelli2, line 110: If you choose to keep the `size_` variable, then you'll have to either pup it, or set it from `unnormalized_function_of_time_`. However, if you get rid of `size_` then there's nothing to do.

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,136 @@
...
+  }
+}
+
+bool operator==(const SettleToConstantQuaternion& lhs,
+                const SettleToConstantQuaternion& rhs) {
+  return lhs.unnormalized_function_of_time_ ==
+             rhs.unnormalized_function_of_time_ and
+         lhs.match_time_ == rhs.match_time_ and
+         lhs.decay_time_ == rhs.decay_time_;
```
knelli2, line 119: Same thing here. If you're keeping `size_`, add it here. Otherwise, all good.

tests/Unit/Domain/FunctionsOfTime/Test_SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,181 @@
...
+  CHECK_ITERABLE_APPROX(lambdas1[0], f_t0);
+  CHECK_ITERABLE_APPROX(lambdas1[1], dtf_t0);
+
+  const auto lambdas2 = f_of_t->func(match_time);
+  CHECK_ITERABLE_APPROX(lambdas2[0], f_t0);
+
+  // check that asymptotic values approach f = A/|A| = const.
+  const auto lambdas3 = f_of_t->func_and_2_derivs(1.0e5);
+  const DataVector zero{0.0, 0.0, 0.0, 0.0};
+  const DataVector normalized_A = A / sqrt(sum(A * A));
```
knelli2: `A / norm(A)`

tests/Unit/Domain/FunctionsOfTime/Test_SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,181 @@
...
+  using SettleToConstantQuaternion =
+      domain::FunctionsOfTime::SettleToConstantQuaternion;
+
+  domain::FunctionsOfTime::register_derived_with_charm();
+
+  const double match_time = 10.0;
+  const double decay_time = 5.0;
+  // Choose for an initial quaternion
+  // q = (cos \theta/2, \hat{n} \sin \theta/2), where
+  // \theta = \Omega(t-t_0) + \alpha(t-t_0)^2/2 _ \phi
```
knelli2: Is the underscore supposed to be a minus?

tests/Unit/Domain/FunctionsOfTime/Test_SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,181 @@
...
+  const DataVector A = f_t0 - B;
+  const std::array<DataVector, 3> init_func{{f_t0, dtf_t0, d2tf_t0}};
+
+  INFO("Test with base class construction.");
+  const std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime> f_of_t =
+      std::make_unique<domain::FunctionsOfTime::SettleToConstantQuaternion>(
+          init_func, match_time, decay_time);
+  test(f_of_t, match_time, decay_time, f_t0, dtf_t0, d2tf_t0, A);
+
+  // FIXME: SEG FAULTS FOR SOME REASON
```
knelli2: Remove.

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
```diff
@@ -0,0 +1,136 @@
...
+template <size_t MaxDerivReturned>
+std::array<double, MaxDerivReturned + 1> quat_norm_and_derivs(
+    const std::array<DataVector, MaxDerivReturned + 1>&
+        unnormalized_func_and_derivs) {
+  std::array<double, MaxDerivReturned + 1> result =
+      make_array<MaxDerivReturned + 1>(0.0);
+  gsl::at(result, 0) =
+      sqrt(sum(square(gsl::at(unnormalized_func_and_derivs, 0))));
+  if constexpr (MaxDerivReturned > 0) {
+    gsl::at(result, 1) = sum(gsl::at(unnormalized_func_and_derivs, 0) *
```
knelli2, line 26: [Optional] For `std::array`s, you only have to use `gsl::at()` when accessing with a non-const variable (like `i`). If you're accessing with a constant expression (i.e. a number like `0` or `1`), then you can use `[]` to access.

knelli2 (changes requested):
Looks good. Please also fix the clang-tidy and compilation errors as well.

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.hpp
geoffrey4444: I'll just hard-code it

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
geoffrey4444: I got rid of it

tests/Unit/Domain/FunctionsOfTime/Test_SettleToConstantQuaternion.cpp
geoffrey4444: should be a +...thanks for catching this!

tests/Unit/Domain/FunctionsOfTime/Test_SettleToConstantQuaternion.cpp
geoffrey4444: 🤦‍♂️

src/Domain/FunctionsOfTime/SettleToConstantQuaternion.cpp
geoffrey4444: Unless others insist, I'd rather stick with gsl::at(), but good to know for the future

knelli2: Yeah that's alright.

knelli2 (commented):
Looks good, you can squash. (you can optionally wait for #5706 to be merged and rebase so the tests pass)

geoffrey4444:
Thanks again for the review, @knelli2 ! I squashed. A GrMhd test failure on gcc-9 release is unrelated to this PR (looks like tolerances might be too strict?)

knelli2 (approved):

knelli2:
Test failures unrealted

PR #5704, opened 2024-01-17
Initialize Klein-Gordon CCE variables

Sizheng-Ma:
## Proposed changes

This PR adds an action to initialize Klein-Gordon CCE variables on the evolution component. The corresponding unit test is also included. The unit test examines all the initialization actions on the component, including both the tensor and scalar parts. The test for tensor variables is the same as the existing CCE system, so I move it to a function in the first commit.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/Actions/InitializeKleinGordonVariables.hpp
+++ b/src/Evolution/Systems/Cce/Actions/InitializeKleinGordonVariables.hpp
@@ -1,0 +1,65 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "Evolution/Systems/Cce/OptionTags.hpp"
+#include "ParallelAlgorithms/Initialization/MutateAssign.hpp"
+
+namespace Cce::Actions {
+
+/*!
+ * \ingroup ActionsGroup
+ * \brief Initialize the data storage for the scalar field in the
+ * `KleinGordonCharacteristicExtract` component, which is the singleton that
+ * handles the main evolution system for Klein-Gordon CCE computations.
+ *
+ * \details Sets up the \ref DataBoxGroup to be ready to take data from the
+ * worldtube component, calculate initial data, and start the hypersurface
+ * computations.
+ *
+ * \ref DataBoxGroup changes:
+ * - Modifies: nothing
+ * - Adds:
+ *  - `Tags::Variables<metavariables::klein_gordon_boundary_communication_tags>`
+ *  - `Tags::Variables<metavariables::klein_gordon_gauge_boundary_tags>`
+ *  - `Tags::Variables<metavariables::klein_gordon_scri_tags>`
+ * - Removes: nothing
+ */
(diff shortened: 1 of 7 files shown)


src/Evolution/Systems/Cce/Actions/InitializeKleinGordonVariables.hpp
```diff
@@ -0,0 +1,70 @@
...
+      const tuples::TaggedTuple<InboxTags...>& /*inboxes*/,
+      const Parallel::GlobalCache<Metavariables>& /*cache*/,
+      const ArrayIndex& /*array_index*/, const ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    initialize_impl(make_not_null(&box));
+    return {Parallel::AlgorithmExecution::Continue, std::nullopt};
+  }
+
+  template <typename TagList>
+  static void initialize_impl(const gsl::not_null<db::DataBox<TagList>*> box) {
```
knelli2: You don't really need an `impl` here. It's only used once, so I think this code can just go in the `apply` function.

tests/Unit/Evolution/Systems/Cce/Actions/Test_InitializeKleinGordonCharacteristicEvolution.cpp
```diff
@@ -0,0 +1,230 @@
...
+  using phase_dependent_action_list =
+      tmpl::list<Parallel::PhaseActions<Parallel::Phase::Initialization,
+                                        initialize_action_list>,
+                 Parallel::PhaseActions<Parallel::Phase::Evolve, tmpl::list<>>>;
+  using const_global_cache_tags =
+      Parallel::get_const_global_cache_tags_from_actions<
+          phase_dependent_action_list>;
+};
+
+struct metavariables {
```
knelli2: Can't many of these type aliases in this metavars be inherited from `CharacteristicExtractBase.hpp`?

knelli2 (changes requested):

src/Evolution/Systems/Cce/Actions/InitializeKleinGordonVariables.hpp
Sizheng-Ma: I was thinking about SpEC CCM, where we have to avoid all the `apply` functions. But I guess linking this KG system to SpEC won't happen in the near future. I've removed it.

tests/Unit/Evolution/Systems/Cce/Actions/Test_InitializeKleinGordonCharacteristicEvolution.cpp
Sizheng-Ma: Nice idea...

knelli2 (commented):
Looks good. Squash.

Sizheng-Ma:
Done.

knelli2 (approved):

PR #5715, opened 2024-01-24
Add exporter library

nilsvu:
## Proposed changes

Adds a function to interpolate volume data files to arbitrary points, which has two use cases:

1. Replaces the Python function with the same functionality. It's been used for things like generating data for paper plots. The new C++ implementation supports OpenMP parallelization over data files, which should make it faster for large data sets (@yoonso0-0 has complained about this). I haven't done a speed test yet. Happy to do so if someone has a suitable data set (@yoonso0-0?).
2. The library can be linked into external programs to load SpECTRE data, e.g. to load SpECTRE initial data into SpEC or other codes.

### Upgrade instructions

The Python module `Visualization.InterpolateToCoords` has moved to `IO.Exporter.InterpolateToPoints`. It has a slightly different (easier) interface, is implemented in C++ now, and supports parallelization, which should make it considerably faster.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupOpenMP.cmake
+++ b/cmake/SetupOpenMP.cmake
@@ -1,0 +1,8 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+option(ENABLE_OPENMP "Enable OpenMP in some parts of the code" OFF)
+
+if (ENABLE_OPENMP)
+  find_package(OpenMP COMPONENTS CXX)
+endif()
--- a/src/IO/Exporter/Exporter.cpp
+++ b/src/IO/Exporter/Exporter.cpp
@@ -1,0 +1,276 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "IO/Exporter/Exporter.hpp"
+
+#include <csignal>  // For Blaze error handling without PCH
+#ifdef _OPENMP
+#include <omp.h>
+#endif  // _OPENMP
+
+#include "Domain/BlockLogicalCoordinates.hpp"
+#include "Domain/Creators/RegisterDerivedWithCharm.hpp"
+#include "Domain/Creators/TimeDependence/RegisterDerivedWithCharm.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/ElementLogicalCoordinates.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/RegisterDerivedWithCharm.hpp"
+#include "IO/H5/File.hpp"
+#include "IO/H5/TensorData.hpp"
+#include "IO/H5/VolumeData.hpp"
+#include "NumericalAlgorithms/Interpolation/IrregularInterpolant.hpp"
+#include "Utilities/FileSystem.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/GetOutput.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Overloader.hpp"
+#include "Utilities/Serialization/Serialize.hpp"
+
+// Ignore OpenMP pragmas when OpenMP is not enabled
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunknown-pragmas"
+
+namespace spectre::Exporter {
+
+namespace {
+template <size_t Dim>
+void interpolate_to_points(
+    const gsl::not_null<std::vector<std::vector<double>>*> result,
+    const gsl::not_null<std::vector<bool>*> filled_data,
+    const std::string& filename, const std::string& subfile_name,
+    const size_t obs_id, const std::vector<std::string>& tensor_components,
+    const std::vector<std::optional<
+        IdPair<domain::BlockId, tnsr::I<double, Dim, Frame::BlockLogical>>>>&
+        block_logical_coords) {
+  std::vector<std::string> grid_names;
+  std::vector<std::vector<size_t>> all_extents;
+  std::vector<std::vector<Spectral::Basis>> all_bases;
+  std::vector<std::vector<Spectral::Quadrature>> all_quadratures;
+  std::vector<DataVector> tensor_data{};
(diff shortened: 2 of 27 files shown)


cmake/SetupOpenMP.cmake
```diff
@@ -0,0 +1,8 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+option(ENABLE_OPENMP "Enable OpenMP in some parts of the code" OFF)
```
nilsdeppe, line 4: Is there any drawback to making `ON` the default? That way tools can always parallelize and will hopefully lead to a better default user experience :shrug: @wthrowe @kidder thoughts?

yoonso0-0:
Nice! I can try this for interpolating data onto a plane to draw figures.

src/IO/Exporter/Exporter.hpp
```diff
@@ -0,0 +1,52 @@
...
+ * not specified, OpenMP will determine the number of threads automatically.
+ * It's also possible to set the number of threads using the environment
+ * variable OMP_NUM_THREADS. It's an error to specify num_threads if OpenMP is
+ * not linked in. Set num_threads to 1 to disable OpenMP.
+ * \return std::vector<std::vector<double>> The interpolated data. The first
+ * dimension corresponds to the selected tensor components, and the second
+ * dimension corresponds to the target points.
+ */
+template <size_t Dim>
+std::vector<std::vector<double>> interpolate_to_points(
```
nilsdeppe, line 44: I remember you talked about a potential C/Fortran interface. Is that planned in addition to this?

src/IO/Exporter/Exporter.cpp
```diff
@@ -0,0 +1,275 @@
...
+#include "NumericalAlgorithms/Interpolation/IrregularInterpolant.hpp"
+#include "Utilities/FileSystem.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/GetOutput.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Overloader.hpp"
+#include "Utilities/Serialization/Serialize.hpp"
+
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wunknown-pragmas"
```
nilsdeppe, line 31: maybe add a comment that this is aimed to ignore OpenMP pragmas when we don't have OpenMP?

src/IO/Exporter/Exporter.cpp
```diff
@@ -0,0 +1,275 @@
...
+  // HDF5 operations at a time
+#pragma omp critical
+  {
+    const h5::H5File<h5::AccessType::ReadOnly> h5file(filename);
+    const auto& volfile = h5file.get<h5::VolumeData>(subfile_name);
+    grid_names = volfile.get_grid_names(obs_id);
+    all_extents = volfile.get_extents(obs_id);
+    all_bases = volfile.get_bases(obs_id);
+    all_quadratures = volfile.get_quadratures(obs_id);
+    // Pre-load the tensor data because it's stored contiguously for all grids
```
nilsdeppe: I don't know what this means, sorry. What is being "pre-loaded"?

nilsdeppe (changes requested):
A few suggestions, feel free to squash if you do them :)

cmake/SetupOpenMP.cmake
nilsvu: I wanted to be cautious :) I had a small issue on Wheeler where a default OpenMP was found and then conflicted with Anaconda (like the zlib issue that you also had). So I chose to make this opt-in. Not too hard to enable by default though I think.

nilsdeppe: ah okay, sounds good!

src/IO/Exporter/Exporter.hpp
nilsvu: Yea maybe if needed? I don't want to waste time writing it without a use case. Should be as easy as compiling a C-compatible wrapper function with a C++ compiler right?

nilsdeppe: Yes, I think it would mean making things `double**` and that's about it. You'd need to be careful to use `malloc` and not `new` but that's fine. Could probably also just have a thin wrapper that takes the C types and translates them into C++ types. But agreed, can do this later :)

cmake/SetupOpenMP.cmake
wthrowe: I worry about our dependencies trying to be smart and automatically starting to use multithreading if it's available.  We've already had to disable that in blaze, and I see a bunch of checks on _OPENMP in the libxsmm headers, although those are too complicated for me to easily tell what they are doing.

nilsvu: Setting `OMP_NUM_THREADS=1` should reliably turn all that off.

wthrowe: It gets rid of the thread creation, but are you sure OpenMP code with `OMP_NUM_THREADS=1` is as fast as non-OpenMP code?  These are the low-level matrix multiplication routines, so performance is important.  Googling finds some people claiming it's a significant slowdown and some claiming it's irrelevant, and it's not obvious to me who to believe.

nilsvu: Let's make sure to test it. Keeping OpenMP disabled by default seems the safer choice for now anyway. Note that even with OpenMP enabled it should only get linked into the exporter lib at the moment (that's why I disabled PCH for the exporter lib, so OpenMP doesn't have to get linked into the PCH).

src/IO/Exporter/Exporter.cpp
nilsvu: Reworded to "Load the tensor data for all grids in the file because it's stored contiguously".

cmake/SetupOpenMP.cmake
nilsdeppe: Alright, let's keep it off :) We can reassess if we ever need it. Thanks @wthrowe !

nilsdeppe (approved):

PR #5718, opened 2024-01-24
Instantiate inertial->distorted for strahlkorper_in_different_frame

geoffrey4444:
## Proposed changes

For ringdown, we will need to transform Strahlkorpers from the ringdown inertial frame to the ringdown distorted frame. But `strahlkorper_in_different_frame()`  currently assumes the initial frame is the grid frame. This PR adds support for transforming from the inertial frame to the distorted frame.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


tests/Unit/Domain/Test_StrahlkorperTransformations.cpp
```diff
@@ -92,13 +93,18 @@ void test_strahlkorper_in_different_frame() {
             2.0, ylm.theta_phi_points(), 1.0,
             std::array<double, 3>{{0.1, 0.2, 0.3}}));
     strahlkorper_expected.reset(new ylm::Strahlkorper<DestFrame>(
-        l_max, l_max, new_radius, strahlkorper_grid_center));
+        l_max, l_max, new_radius, strahlkorper_src_center));
+  } else if constexpr (std::is_same_v<SrcFrame, ::Frame::Inertial> and
+                       std::is_same_v<DestFrame, ::Frame::Distorted>) {
+    strahlkorper_expected.reset(new ylm::Strahlkorper<DestFrame>(
```
geoffrey4444, line 99: Clang tidy fails on this line, but I don't understand what it's mad about

geoffrey4444: Discussion on 1/25/24 call: we agree to disable the check that this trips. If I need to make changes before merging this PR (I'm guessing that will be the case), I can disable that check as part of this PR once someone tells me how

src/Domain/StrahlkorperTransformations.cpp
```diff
@@ -67,18 +68,36 @@ void coords_to_different_frame(
...
             "Strahlkorper lies outside of distorted-frame region");
       }
       const auto& grid_to_distorted_map =
           block.moving_mesh_grid_to_distorted_map();
       x_dest = grid_to_distorted_map(x_src, time, functions_of_time);
+    } else if constexpr (std::is_same_v<DestFrame, ::Frame::Distorted> and
+                         std::is_same_v<SrcFrame, ::Frame::Inertial>) {
+      if (not block.has_distorted_frame()) {
+        throw std::runtime_error(
+            "Strahlkorper lies outside of distorted-frame region");
```
knelli2, line 84: Not a change request because you're just copying what was already done in the file, but these should probably be converted to `ERROR`s at some point.

src/Domain/StrahlkorperTransformations.cpp
```diff
@@ -183,27 +202,50 @@ void strahlkorper_in_different_frame(
...
+    static_assert(
+        std::is_same_v<SrcFrame, ::Frame::Grid> or
+            std::is_same_v<SrcFrame, ::Frame::Inertial>,
+        "Source frame must currently be Grid frame or Inertial frame");
+    static_assert(std::is_same_v<DestFrame, ::Frame::Inertial> or
+                      std::is_same_v<DestFrame, ::Frame::Distorted>,
+                  "Destination frame must currently be Inertial frame or "
+                  "Distorted frame");
+    if constexpr (std::is_same_v<SrcFrame, ::Frame::Grid>) {
+      const auto& x_src =
```
knelli2: To reduce code duplication, maybe do

```cpp
tnsr::I<double, 3, SrcFrame> x_src{};

if constexpr (std::is_same_v<SrcFrame, ::Frame::Grid>) {
  x_src = // logical to grid map
} else {
  x_src = // logical to grid to inertial map
}

const double r_src = ...;
...
```

tests/Unit/Domain/Test_StrahlkorperTransformations.cpp
knelli2: Yeah an extra commit sounds good.

knelli2 (changes requested):
Couple small things. You can squash directly

src/Domain/StrahlkorperTransformations.cpp
geoffrey4444: Thanks for this suggestion!

knelli2 (approved):

PR #5723, opened 2024-01-30
Add CLI to plot size control diagnostics

knelli2:
## Proposed changes

This is just a quick utility that plots most of the size control diagnostics necessary to quickly see what has gone wrong in a simulation. Not intended for a super detailed analysis, but you can change the x-range on the plot to zoom in on a specific time (and have the y-axes of all plots rescale accordingly)

To quickly plot size control diagnostics, try
```
spectre plot-size-control --object-label A --output SizeControlDiag --x-label "Time (M)" --title "AhA size control diagnostics" ReductionsFile.h5
```

This gives the result of [SizeControlDiagA.pdf](https://github.com/sxs-collaboration/spectre/files/14091278/SizeControlDiagA.pdf)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/PlotSizeControl.py
+++ b/src/Visualization/Python/PlotSizeControl.py
@@ -1,0 +1,229 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+import os
+from typing import Optional, Sequence
+
+import click
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+from spectre.Visualization.ReadH5 import available_subfiles, to_dataframe
+
+logger = logging.getLogger(__name__)
+
+
+@click.command(name="plot-size-control")
+@click.argument(
+    "reduction_files",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    nargs=-1,
+)
+@click.option(
+    "--object-label",
+    "-d",
+    required=True,
+    help=(
+        "Which object to plot. This is either 'A', 'B', or 'None'. 'None' is"
+        " used when there is only one black hole in the simulation."
+    ),
+)
+@click.option(
+    "--output",
+    "-o",
+    type=click.Path(file_okay=True, dir_okay=False, writable=True),
+    help=(
+        "Name of the output image. The '--object-label' will be added"
+        " automatically to the end of the output. Also the suffix '.pdf' will"
+        " be added if necessary. If unspecified, the plot is shown"
+        " interactively, which only works on machines with a window server."
+    ),
+)
+# Plotting options
+@click.option(
+    "--x-bounds",
+    type=float,
+    nargs=2,
+    help="The lower and upper bounds of the x-axis.",
+)
+@click.option(
+    "--x-label",
+    help="The label on the x-axis.",
+)
+@click.option(
+    "--title",
+    "-t",
+    help="Title of the graph.",
+)
+@click.option(
+    "--stylesheet",
+    "-s",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    envvar="SPECTRE_MPL_STYLESHEET",
+    help=(
+        "Select a matplotlib stylesheet for customization of the plot, such "
+        "as linestyle cycles, linewidth, fontsize, legend, etc. "
+        "The stylesheet can also be set with the 'SPECTRE_MPL_STYLESHEET' "
+        "environment variable."
+    ),
+)
+def plot_size_control_command(
+    reduction_files: Sequence[str],
+    object_label: str,
+    output: Optional[str],
+    x_bounds: Optional[Sequence[float]],
+    x_label: Optional[str],
+    title: Optional[str],
+    stylesheet,
+):
+    """
+    Plot diagnostic information regarding the Size control system.
+
+    This tool assumes there is a subfile in each of the "reduction-files" with
+    the path `/ControlSystems/Size{LABEL}/Diagnostics.dat`, where `{LABEL}` is
+    replaced with the "object-label" input option.
+    """
+    object_label = (
+        "" if object_label.lower() == "none" else object_label.upper()
+    )
+
+    def check_diagnostics_file(h5_filename):
+        with h5py.File(h5_filename, "r") as h5file:
+            # See src/ControlSystem/ControlErrors/Size.cpp for path
+            diagnostics_file_name = (
+                f"ControlSystems/Size{object_label}/Diagnostics.dat"
+            )
+            diagnostics_data = h5file.get(diagnostics_file_name)
+            if diagnostics_data is None:
+                raise click.UsageError(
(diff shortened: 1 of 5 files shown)


src/Visualization/Python/PlotSizeControl.py
```diff
@@ -0,0 +1,361 @@
...
+    help=(
+        "Name of the output image. The '--object-label' will be added"
+        " automatically to the end of the output. Also the suffix '.pdf' will"
+        " be added if necessary. If unspecified, the plot is shown"
+        " interactively, which only works on machines with a window server."
+    ),
+)
+# Plotting options
+@click.option(
+    "--xlim",
```
nilsvu: Mind making this consistent with `PlotDatFile.py`, i.e. `--x-bounds`, `--x-label`? (just for consistency of the CLI)

src/Visualization/Python/PlotSizeControl.py
```diff
@@ -0,0 +1,361 @@
...
+    stylesheet,
+):
+    """
+    Plot diagnostic information regarding the Size control system.
+
+    This tool assumes there is a subfile in each of the "reduction-files" with
+    the path `/ControlSystems/Size{LABEL}/Diagnostics.dat`, where `{LABEL}` is
+    replaced with the "object-label" input option.
+    """
+    _rich_traceback_guard = True  # Hide traceback until here
```
nilsvu: delete this line (it will also hide tracebacks from the rest of this function)

src/Visualization/Python/PlotSizeControl.py
```diff
@@ -0,0 +1,361 @@
...
+
+    object_label = (
+        "" if object_label.lower() == "none" else object_label.upper()
+    )
+
+    times = np.array([])
```
nilsvu: This is actually almost trivial with pandas:

```py
import pandas as pd
from spectre.Visualization.ReadH5 import to_dataframe

data = pd.concat(to_dataframe(h5py.File(filename)[subfile_name]) for filename in filenames)

# That's all! Replaces these ~90 lines.
```

nilsvu (changes requested):
Will finish the review later

src/Visualization/Python/PlotSizeControl.py
```diff
@@ -0,0 +1,361 @@
...
+                [damping_timescale, diagnostics_data[:, 22]]
+            )
+
+    # Apply stylesheet
+    if stylesheet is not None:
+        plt.style.use(stylesheet)
+
+    # Set up plots
+    fig, axes = plt.subplots(7, 1, sharex=True)
+    plt.rcParams["text.usetex"] = True
```
nilsvu: Delete this line (should be left up to the user / the stylesheet)

src/Visualization/Python/PlotSizeControl.py
nilsvu: Note: if the amount of data becomes too large you can also do `to_dataframe(..., slice=np.s_[::step])` to load only every `step` observation.

src/Visualization/Python/PlotSizeControl.py
```diff
@@ -0,0 +1,361 @@
...
+
+    # Plot the selected quantities. Groupings are based both on relationship
+    # of quantities and also relative scales of quantities
+    axes = list(axes)
+    # Damping times are sort of the odd ones out. Put it with control error but
+    # on a different y-axis
+    top1 = axes[0].plot(times, control_error, label="Control Error")
```
nilsvu: Does it make sense to plot `abs(control_error)` on a log scale?

src/Visualization/Python/PlotSizeControl.py
```diff
@@ -0,0 +1,361 @@
...
+    axes[5].plot(times, horizon_coef_00, label="Horizon coef 00")
+
+    # Time deriv of lambda and horizon
+    axes[6].plot(times, dt_function_of_time, label="dtFunction of time")
+    axes[6].plot(times, raw_dt_horizon_coef_00, label="Raw dtHorizon coef 00")
+    axes[6].plot(times, avg_dt_horizon_coef_00, label="Avg dtHorizon coef 00")
+
+    # Configure the axes and legends
+    fig.set_size_inches(10, len(axes) * 2)
+    fig.set_dpi(100)
```
nilsvu: The dpi should also be left to the user / the stylesheet I think

src/Visualization/Python/PlotSizeControl.py
```diff
@@ -0,0 +1,361 @@
...
+    if xlabel:
+        # Bottom plot. -1 is the second y-axis for top plot
+        axes[-2].set_xlabel(xlabel)
+    if xlim:
+        plt.xlim(*xlim)
+    if title:
+        plt.title(title)
+    for i in range(len(axes) - 1):
+        autoscale(axes[i], margin=0.05)
```
nilsvu: Can't you just rely on matplotlib's autoscaling? I think a way to also avoid plotting all the data outside the bounds is with `data = data[data["Time"] >= xlim[0] & data["Time"] <= xlim[1]]` right after loading the data.

tests/Unit/Visualization/Python/Test_PlotSizeControl.py
```diff
@@ -0,0 +1,131 @@
...
+                "--xlabel",
+                "Time (M)",
+                "--title",
+                "Size Control First File",
+            ],
+            catch_exceptions=False,
+        )
+
+        output_file_name = os.path.join(self.work_dir, "SingleFileB.pdf")
+        print(result.stdout)
```
nilsvu: If you want to print output on error, do `self.assertEqual(result.exit_code, 0, result.output)`

tests/Unit/Visualization/Python/Test_PlotSizeControl.py
```diff
@@ -0,0 +1,131 @@
...
+        self.assertEqual(result.exit_code, 0)
+        self.assertTrue(os.path.exists(output_file_name))
+
+        # Just one h5 file
```
nilsvu: multiple, right?

nilsvu (changes requested):

src/Visualization/Python/PlotSizeControl.py
knelli2: No, not really. I think we'd lose a lot of information about small oscillations that way.

src/Visualization/Python/PlotSizeControl.py
knelli2: I found that when I limited the x-range, that the y-axes didn't rescale automatically. It kept the autoscaling for the full x-range. However, I agree if we limit the data right after loading, this problem goes away. I'll try that

src/Visualization/Python/PlotSizeControl.py
nilsvu: How so? You see small oscillations well on a log scale. I suggest trying a log scale, and/or plotting both. Anyway, this is up to you.

nilsvu (approved):

PR #5725, opened 2024-01-30
Update CaltechHpc environment files

knelli2:
## Proposed changes

This PR does 4 things (1-3 are done in the first commit simultaneously, 4 is done in the second commit):

1. Use the SXS-installed modules in the current environment rather than the system installed ones.
2. Rename the current environment to include `skylake` in the file names
3. Use AVX2 in the current environment instead of AVX512.
4. Add a new environment (icelake nodes, modules installed by me). Adds environment file, machine file, and submit script. Has `icelake` in the filenames.

### Upgrade instructions

On CaltechHPC, the environment file `caltech_hpc_gcc.sh` has been renamed to `caltech_hpc_gcc_skylake.sh`.

There is also a new environment file
- `caltech_hpc_gcc_icelake.sh`

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


support/Environments/caltech_hpc_skylake_homegrown_gcc.sh
```diff
@@ -0,0 +1,88 @@
...
+    cmake -D CMAKE_C_COMPILER=gcc \
+          -D CMAKE_CXX_COMPILER=g++ \
+          -D CMAKE_Fortran_COMPILER=gfortran \
+          -D CHARM_ROOT=$CHARM_ROOT \
+          -D CMAKE_BUILD_TYPE=Release \
+          -D MEMORY_ALLOCATOR=JEMALLOC \
+          -D BUILD_PYTHON_BINDINGS=ON \
+          -D MACHINE=CaltechHpcSkylake \
+          -D OVERRIDE_ARCH=skylake \
+          -D YAMLCPP_ROOT=/central/groups/sxs/knelli/libraries/yaml-cpp/0.8.0 \
```
nilsvu: Can you delete this line? the `CMAKE_PREFIX_PATH` in the module file should be enough. If it isn't (which it should be), add `YAMLCCP_ROOT` as an env variable in the module file.

support/Environments/caltech_hpc_skylake_homegrown_gcc.sh
```diff
@@ -0,0 +1,88 @@
...
+    # these issues. It is still unclear *why* the SYSTEM allocator caused these
+    # issues. --Kyle Nelli
+    cmake -D CMAKE_C_COMPILER=gcc \
+          -D CMAKE_CXX_COMPILER=g++ \
+          -D CMAKE_Fortran_COMPILER=gfortran \
+          -D CHARM_ROOT=$CHARM_ROOT \
+          -D CMAKE_BUILD_TYPE=Release \
+          -D MEMORY_ALLOCATOR=JEMALLOC \
+          -D BUILD_PYTHON_BINDINGS=ON \
+          -D MACHINE=CaltechHpcSkylake \
```
nilsvu: Do you intend to keep all three environments on CaltechHPC maintained? E.g. I don't see an fftw module in this file, nor libbacktrace (for nicer backtraces) or paraview.

support/Environments/caltech_hpc_icelake.sh
```diff
@@ -0,0 +1,91 @@
...
+    cmake -D CMAKE_C_COMPILER=gcc \
+          -D CMAKE_CXX_COMPILER=g++ \
+          -D CMAKE_Fortran_COMPILER=gfortran \
+          -D CHARM_ROOT=$CHARM_ROOT \
+          -D CMAKE_BUILD_TYPE=Release \
+          -D MEMORY_ALLOCATOR=JEMALLOC \
+          -D BUILD_PYTHON_BINDINGS=ON \
+          -D MACHINE=CaltechHpcIcelake \
+          -D OVERRIDE_ARCH=icelake-server \
+          -D YAMLCPP_ROOT=/central/groups/sxs/knelli/libraries/yaml-cpp/0.8.0 \
```
nilsvu: same comment

nilsvu (changes requested):

support/Environments/caltech_hpc_icelake.sh
```diff
@@ -0,0 +1,91 @@
...
+    fi
+    spectre_load_modules
+
+    # Note that choosing the memory allocator to be JEMALLOC is important. When
+    # using the SYSTEM allocator, during BBH runs dumping volume data too often
+    # would result in memory corruption errors. Switching to JEMALLOC fixed
+    # these issues. It is still unclear *why* the SYSTEM allocator caused these
+    # issues. --Kyle Nelli
+    cmake -D CMAKE_C_COMPILER=gcc \
+          -D CMAKE_CXX_COMPILER=g++ \
```
nilsvu: this is picking up `/central/groups/sxs/knelli/compilers/clang/16.0.6/bin/g++`, which I think is just an alias but still weird.

support/Environments/caltech_hpc_skylake_homegrown_gcc.sh
knelli2: I'll try `YAMLCPP_ROOT` because just the `CMAKE_PREFIX_PATH` wasn't enough when I tried

support/Machines/CaltechHpcIcelake.yaml
```diff
@@ -0,0 +1,15 @@
...
+# See LICENSE.txt for details.
+
+Machine:
+  Name: CaltechHpcIcelake
+  Description: |
+    Public supercomputer at Caltech using the icelake intel nodes and
+    dependencies installed by Kyle Nelli.
+    More information:
+    https://www.hpc.caltech.edu/documentation
+  # Split one physical 56 core node into two charm nodes
```
nilsvu: 64 cores

support/Environments/caltech_hpc_skylake_homegrown_gcc.sh
knelli2: For now, yes I intend to have all three environments. However, the sys admins will be switching the entire machine over to a new OS soon (not exactly sure when) so the `caltech_hpc_skylake_gcc.sh` env will be obsolete because those are system installed modules on the current (soon to be old) OS. I plan to remove that env when the OS switch happens, then we'll just have skylake vs icelake with my installed modules.

As for the fftw module, that was needed because of the spec-exporter on the icelake nodes. The spec-exporter for skylake currently uses the system fftw (annoying). I can unify things when the OS switch happens.

For libbacktrace and paraview, I just didn't install those yet because they weren't of highest priority to get the code running. I can do those in a followup.

support/Environments/caltech_hpc_skylake_homegrown_gcc.sh
```diff
@@ -0,0 +1,88 @@
...
+    module unload libraries/brigand/master
+    module unload libraries/catch/3.4.0
+    module unload oneapi/mpi-2021.11
+    module unload libraries/charm/7.0.0
+    module unload libraries/gsl/2.7
+    module unload libraries/hdf5/1.12.3
+    module unload libraries/jemalloc/5.3.0
+    module unload libraries/libsharp/1.0.0
+    module unload libraries/libxsmm/1.16.1-avx2
+    module unload libraries/openblas/0.3.25-avx2
```
nilsvu: I can confirm that the issue with random FPEs from MKL avx512 in the current CaltechHPC Skylake environment is fixed here 👍 So should we just upgrade the environments in place (delete the other one)?

nilsvu (changes requested):

support/Environments/caltech_hpc_icelake.sh
knelli2: Yeah I noticed that too. It is just an alias to the correct gcc, so I didn't think much of it. Not sure how to fix this without specifying the full path. Maybe there is some way to add only a `clang-format` module?

support/Environments/caltech_hpc_skylake_homegrown_gcc.sh
knelli2: This would be an argument to just forgo the old system-installed env. I don't have an issue with this. Should we wait for more people to try this out?

support/Environments/caltech_hpc_skylake_homegrown_gcc.sh
nilsvu: quick note: the new [Paraview 5.12](https://www.paraview.org/download/) prebuilt binaries use py3.10, so should be compatible with your installation.

knelli2:
Ok @nilsvu, I think this is ready for another look. I simplified things a lot

support/Environments/caltech_hpc_skylake_homegrown_gcc.sh
nilsdeppe: I don't think we should use the prebuilt binaries. On wheeler for example the 5.11.1 paraview doesn't work on compute nodes, while the 5.10 version I compiled from scratch does. We really should have ParaView be built against the MPI and HDF5 we use, ideally the python too.

nilsvu (approved):

PR #5732, opened 2024-02-01
Limit initial time step to avoid control system deadlocks

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Actions/InitializeMeasurements.hpp
+++ b/src/ControlSystem/Actions/InitializeMeasurements.hpp
@@ -3,6 +3,7 @@
 
 #pragma once
 
+#include <algorithm>
 #include <cstddef>
 #include <limits>
 #include <memory>
@@ -23,11 +24,20 @@
 #include "Parallel/AlgorithmExecution.hpp"
 #include "Parallel/GlobalCache.hpp"
 #include "ParallelAlgorithms/EventsAndTriggers/Event.hpp"
+#include "Time/ChooseLtsStepSize.hpp"
+#include "Time/Slab.hpp"
+#include "Time/Time.hpp"
 #include "Utilities/Gsl.hpp"
 #include "Utilities/MakeVector.hpp"
 #include "Utilities/TMPL.hpp"
 
 /// \cond
+namespace Tags {
+struct TimeStep;
+}  // namespace Tags
+namespace domain::Tags {
+struct FunctionsOfTime;
+}  // namespace domain::Tags
 namespace evolution::Tags {
 struct EventsAndDenseTriggers;
 }  // namespace evolution::Tags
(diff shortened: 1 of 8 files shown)


src/ControlSystem/Actions/InitializeMeasurements.hpp
```diff
@@ -119,6 +129,40 @@ struct InitializeMeasurements {
...
+    for (const auto& fot :
+         Parallel::get<domain::Tags::FunctionsOfTime>(cache)) {
+      earliest_expiration =
+          std::min(earliest_expiration, fot.second->time_bounds()[1]);
+    }
+    const auto& time_step = db::get<::Tags::TimeStep>(box);
+    const auto start_time = time_step.slab().start();
+    if ((start_time + time_step).value() > earliest_expiration) {
+      TimeDelta new_step{};
+      if constexpr (Metavariables::local_time_stepping) {
```
knelli2: [optional] You could put the `if constexpr` inside the `db::mutate`. Avoids the intermediate variable.

tests/Unit/ControlSystem/Actions/Test_InitializeMeasurements.cpp
```diff
@@ -266,9 +297,13 @@ void test_initialize_measurements(const bool ab_active, const bool c_active) {
 
 SPECTRE_TEST_CASE("Unit.ControlSystem.InitializeMeasurements",
                   "[ControlSystem][Unit]") {
-  test_initialize_measurements(true, true);
-  test_initialize_measurements(true, false);
-  test_initialize_measurements(false, true);
-  test_initialize_measurements(false, false);
+  test_initialize_measurements<false>(true, true);
+  test_initialize_measurements<false>(true, false);
```
knelli2: [optional]

```cpp
  for (const auto& [ab_active, c_active] :
       cartesian_product(std::array{true, false}, std::array{true, false})) {
    test_initialize_measurements<false>(ab_active, c_active);
    test_initialize_measurements<true>(ab_active, c_active);
  }
```

knelli2 (commented):
In the BBH metavariables, the `InitializeMeasurements` action is at the end of the list, so it's fine that we change the step size there. But what prevents somebody from putting another initialization action after it that again changes the step size to something undesirable by the control system? Not saying this needs to hold up this PR, but something we should think about.

wthrowe:
Initialization actions are only allowed to decrease the time step, otherwise they could destabilize the evolution.  That won't interfere with this code.

The one exception I could imagine is a hypothetical initial run of some StepChoosers to pick a time step, but that's not really any worse than the existing restrictions on the step action ordering.

knelli2 (approved):
Looks good

PR #5735, opened 2024-02-02
Add domain rendering script

nilsvu:
Shows elements and grid lines. A second high-resolution dataset can be used to draw smooth element outlines. This rendering is a starting point for visualizations of the domain geometry, e.g. for publications.

For example:

punctures_domain

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


yoonso0-0:
output

yoonso0-0:
It's working really well. Thanks for implementing this.

The above output image is what I get when I fix the clip normal to `0.0, 0.0, 1.0` (Z axis, default) but shifted the clip origin into `10.0, 0.0, 0.0`. In principle it should show the same image as below : 

output


Is this a paraview behavior or a bug is somewhere?

nilsvu:
Should be fixed (see fixup commit). Camera positioning is a bit tricky with ParaView... I'm sure this can be made more robust, but I think this is ok for a start.

yoonso0-0:
Checked the fixup solved issue. Please squash

src/Visualization/Python/Render3D/Domain.py
```diff
@@ -0,0 +1,192 @@
...
+    pv.ColorBy(outline_display, None)
+
+    # Set resolution
+    layout = pv.GetLayout()
+    layout.SetSize(1200, 1200)
+
+    # Configure camera
+    camera = pv.GetActiveCamera()
+    camera.SetFocalPoint(*clip_origin)
+    camera.SetPosition(*(np.array(clip_origin) + np.array(clip_normal) * 10.0))
```
knelli2, line 178: This factor of 10 seems a bit arbitrary. I could imagine this not working for a BBH with an outer boundary of like 1000.

knelli2 (approved):
Let me know if you want to do anything about my suggestion. Not necessary though

src/Visualization/Python/Render3D/Domain.py
nilsvu: Actually ParaView ignores this factor, so I kept it at 10 :D It sets the zoom factor automatically to show the entire domain. As I said above, we can surely make the camera handling better here, but I haven't played around with it enough to understand what ParaView actually does with the camera.

PR #5737, opened 2024-02-02
Support a different elastic material in each block

nilsvu:
## Proposed changes

Getting code merged that's used in the thermal noise paper (https://arxiv.org/abs/2111.06893). This PR adds support for layered materials, but the DG scheme still needs an update to support this (next PR).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/BlockGroups.hpp
+++ b/src/Domain/Creators/BlockGroups.hpp
@@ -1,0 +1,45 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+/// \file
+/// Utilities to work with block names and groups
+
+#pragma once
+
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+namespace domain {
+
+/*!
+ * \brief Check if a block is in the given group
+ *
+ * \param block_name The name of the block to check
+ * \param block_or_group_name The group name we're testing. Returns true if
+ * the block is in this group. This can also be the name of the block itself.
+ * \param block_groups All block groups.
+ */
+bool block_is_in_group(
+    const std::string& block_name, const std::string& block_or_group_name,
+    const std::unordered_map<std::string, std::unordered_set<std::string>>&
+        block_groups);
+
+/*!
+ * \brief Expand a list of block or group names into a list of block names
+ *
+ * \param block_or_group_names Block or group names to expand
+ * \param all_block_names All block names in the domain
+ * \param block_groups Block groups used to expand the names
+ * \return std::unordered_set<std::string> List of block names that appear in
+ * \p all_block_names. If one of the input names was a group, then all block
+ * names from that group are included. Overlaps between groups are allowed.
+ */
+std::unordered_set<std::string> expand_block_groups_to_block_names(
+    const std::vector<std::string>& block_or_group_names,
+    const std::vector<std::string>& all_block_names,
+    const std::unordered_map<std::string, std::unordered_set<std::string>>&
+        block_groups);
+
+}  // namespace domain
--- a/src/Domain/Creators/ExpandOverBlocks.hpp
+++ b/src/Domain/Creators/ExpandOverBlocks.hpp
@@ -10,23 +10,44 @@
 #include <unordered_set>
 #include <vector>
 
+#include "Utilities/MakeArray.hpp"
+
 namespace domain {
+
+namespace ExpandOverBlocks_detail {
+template <typename T, typename U, typename = std::void_t<>>
+struct is_value_type {
+  static constexpr bool value = false;
+};
+
+template <typename T, typename U>
+struct is_value_type<T, U, std::void_t<typename T::value_type>> {
+  static constexpr bool value = std::is_same_v<U, typename T::value_type>;
+};
+}  // namespace ExpandOverBlocks_detail
+
 /*!
- * \brief Produce a distribution of type `T` over all blocks and dimensions in
- * the domain, based on values `T` of variable isotropy and homogeneity.
+ * \brief Produce a std::vector<T> over all blocks of the domain
  *
  * This class is useful to option-create values for e.g. the initial refinement
  * level or initial number of grid points for domain creators. It can be used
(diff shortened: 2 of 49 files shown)


.clang-tidy
```diff
@@ -78,6 +78,8 @@ Checks: '*,
 -readability-magic-numbers,
 # Same as llvm-qualified-auto above.,
 -readability-qualified-auto,
+# Access specifies can be useful to structure code,
```
geoffrey4444: specifies -> specifiers?

src/Domain/Creators/ExpandOverBlocks.hpp
```diff
@@ -41,40 +67,151 @@ namespace domain {
...
+  }
 
  private:
   size_t num_blocks_;
   std::vector<std::string> block_names_;
   std::unordered_map<std::string, std::unordered_set<std::string>>
       block_groups_;
 };
+
+template <typename T>
```
nilsdeppe: Could we move the definitions to a `tpp` and instantiate STL types maybe in `ExpandOverBlocks.cpp`? We can then also instantiate specific other types in other places as necessary. I'd us to continue to reduce code in header files. In this case I worry a bit that it'll be transitively included in a decent number of places.

src/Domain/Creators/ExpandOverBlocks.hpp
```diff
@@ -41,40 +67,151 @@ namespace domain {
...
+std::vector<T> ExpandOverBlocks<T>::operator()(
+    const std::vector<T>& value) const {
+  if (value.size() != num_blocks_) {
+    throw std::length_error{"You supplied " + std::to_string(value.size()) +
+                            " values, but the domain creator has " +
+                            std::to_string(num_blocks_) + " blocks."};
+  }
+  if constexpr (tt::is_a_v<std::unique_ptr, T>) {
+    std::vector<T> expanded{};
```
nilsdeppe: `Utilities/CloneUniquePtrs.hpp` has code to do this. Can you just use that?

src/Domain/Creators/ExpandOverBlocks.hpp
```diff
@@ -41,40 +67,151 @@ namespace domain {
...
+template <typename T>
+std::vector<T> ExpandOverBlocks<T>::operator()(
+    const std::unordered_map<std::string, T>& value) const {
+  ASSERT(num_blocks_ == block_names_.size(),
+         "Construct 'ExpandOverBlocks' with block names to use the "
+         "map-over-block-names feature.");
+  // Expand group names
+  auto value_per_block = [&value]() {
+    if constexpr (tt::is_a_v<std::unique_ptr, T>) {
+      std::unordered_map<std::string, T> copy{};
```
nilsdeppe: Again, can we use `Utilities/CloneUniquePtrs.hpp`?

src/ParallelAlgorithms/Events/ObserveNorms.hpp
```diff
@@ -38,6 +38,7 @@
 #include "Utilities/ErrorHandling/Error.hpp"
```
nilsdeppe, line 38: I don't understand how this change allows observing anything per material layer. It just allows you to name `ObserveNorms` something else. Couldn't this be achieved with compute tags filtering on the layer?

This needs tests regardless, but I'm not sure this is the right solution to the problem. Maybe we can discuss in the tech call?

src/Domain/Creators/BlockGroups.hpp
```diff
@@ -0,0 +1,45 @@
...
+ * \brief Expand a list of block or group names into a list of block names
+ *
+ * \param block_or_group_names Block or group names to expand
+ * \param all_block_names All block names in the domain
+ * \param block_groups Block groups used to expand the names
+ * \return std::unordered_set<std::string> List of block names that appear in
+ * \p all_block_names. If one of the input names was a group, then all block
+ * names from that group are included. Overlaps between groups are allowed.
+ */
+std::unordered_set<std::string> expand_block_groups(
```
nilsdeppe: maybe add `_to_block_names`? It isn't clear to me from the current name what it does, and while documented, the extra clarity seems worth it to me :shrug:

src/Elliptic/Systems/Elasticity/Actions/InitializeConstitutiveRelation.hpp
```diff
@@ -0,0 +1,201 @@
+// Distributed under the MIT License.
```
nilsdeppe, line 1: Are there tests for these?

nilsdeppe (changes requested):

geoffrey4444 (changes requested):
Just one quick thing I saw so far…haven’t finished reviewing yet

geoffrey4444 (changes requested):
Maybe I missed it, but is there a unit test anywhere that actually uses different constitutive relations in different blocks? If not, would it be hard to add one>?

src/ParallelAlgorithms/Events/ObserveNorms.hpp
nilsvu: As discussed in a meeting recently, yes, this only provides a name for the event in the input file. It's needed when you have multiple `ObserveNorms` events in the same executable (here you have one for norms over the full domain and one for norms over each material layer). I think passing along a label struct is consistent with other parts of the code where we do that, e.g. `Options::Auto<T, Label>`. I added some docs to explain.

src/Domain/Creators/BlockGroups.hpp
nilsvu: squashed this rename in

nilsvu:
I squashed in the split of ExpandOverBlocks into a tpp file, and added a test.

nilsdeppe (commented):
LGTM, please squash!

nilsvu:
@geoffrey4444 do you want to take another look?

geoffrey4444 (approved):

nilsdeppe (approved):

Issue #3844, opened 2022-03-03
Segfaults with Numpy 1.22+


nilsvu:
# Bug reports:

In the container I get segfaults in Pypp `check_with_random_values` tests when I `pip install numpy==1.22.2` (upgrading from 1.21.5). I haven't been able to reproduce this on macOS yet. If anyone runs into this issue as well and can provide any insight, that would be much appreciated.

PR #5744, opened 2024-02-07
Python 12 fixes

kidder:
## Proposed changes

Also fix a shadow warning from clang v17 and increase some test timeouts

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Framework/SetupLocalPythonEnvironment.cpp
+++ b/tests/Unit/Framework/SetupLocalPythonEnvironment.cpp
@@ -26,8 +26,10 @@ void prepend_python_path(const std::string& new_path) {
     ERROR_NO_TRACE("Trying to add path '"
                    << new_path
                    << "' to the python environment during setup "
-                      "but this directory does not exist. Maybe "
-                      "you have a typo in your path?");
+                      "but this directory does not exist. Either "
+                      "you have a typo in your path, or you need "
+                      "to create the appropriate site-packages "
+                      "directory after upgrading python.");
   }
   PyObject* sys_path = PySys_GetObject("path");
   PyList_Insert(sys_path, 0, PyUnicode_FromString(new_path.c_str()));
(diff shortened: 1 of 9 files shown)


kidder:
@nilsvu I also could not use the old version of numpy (1.22.0) with python 3.12, but I do not get the claimed segfaults in Pypp tests with version 1.24.4. (See #3844)

tests/Unit/Framework/SetupLocalPythonEnvironment.cpp
```diff
@@ -26,8 +26,10 @@ void prepend_python_path(const std::string& new_path) {
     ERROR_NO_TRACE("Trying to add path '"
                    << new_path
                    << "' to the python environment during setup "
-                      "but this directory does not exist. Maybe "
-                      "you have a typo in your path?");
+                      "but this directory does not exist. Either "
+                      "you have a typo in your path, or you need "
+                      "to create the appropriate site-packages "
+                      "directory after upgrading python.");
```
nilsvu, line 32: Maybe just delete this error message altogether? Should be ok to add nonexistent directories to the path.

nilsvu (approved):

nilsvu:
So are you suggesting that https://github.com/sxs-collaboration/spectre/issues/3844 is fixed for newer numpy or Python versions? If you have an idea which versions work and which are faulty it would be good to update the constraints in the `support/Python/requirements.txt` file.

tests/Unit/Framework/SetupLocalPythonEnvironment.cpp
kidder: I assume someone put in the current message because they had a typo in their path and it led to confusion.  So I'd prefer to keep it

kidder:
> So are you suggesting that #3844 is fixed for newer numpy or Python versions? If you have an idea which versions work and which are faulty it would be good to update the constraints in the `support/Python/requirements.txt` file.

Yes possibly; the issue only referenced the container though so I haven't tested anything.  But if python 12 isn't compatible with the pinned numpy version, something will have to change

PR #5746, opened 2024-02-09
Fix a minus sign in puncture equations

nilsvu:
## Proposed changes

The docs have the correct equation: Eq. (1) here https://spectre-code.org/namespacePunctures.html#details
The source term is moved to the LHS (because it depends on the field $u$ that is being solved for), so the term should have a minus sign.
I found this when comparing to the TwoPunctures code https://arxiv.org/abs/gr-qc/0404056. Can't really catch this in a unit test I think.
I can pretty easily reproduce their Fig. 5 now with nice exponential convergence.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (approved):
I also don't have a great suggestion for a test that would've caught this :shrug: We've had something like this before, and in that case testing that an analytic solution was computed correctly would've caught it, but I'm not familiar enough with the elliptic code to be able to give any ideas in that direction. I'll merge this now and if you think of a way to test it, I'll review that PR.

PR #5748, opened 2024-02-09
Add TaggedTuple support to pypp

nilsdeppe:
## Proposed changes

This will be used for simplifying our python tests of boundary conditions.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Framework/Pypp.hpp
+++ b/tests/Unit/Framework/Pypp.hpp
@@ -16,7 +16,9 @@
 #include <type_traits>
 
 #include "DataStructures/ComplexDataVector.hpp"
+#include "DataStructures/DataBox/TagName.hpp"
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Metafunctions.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Framework/PyppFundamentals.hpp"
 #include "Utilities/ErrorHandling/FloatingPointExceptions.hpp"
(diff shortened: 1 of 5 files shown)


tests/Unit/Framework/Pypp.hpp
```diff
@@ -209,6 +211,50 @@ struct ContainerPackAndUnpack<std::tuple<Tensor<T, Symm, Indices>...>,
...
+  struct UnpackedTag {
+    static std::string name() { return db::tag_name<Tag>(); }
+    using type = typename ContainerPackAndUnpack<
+        typename Tag::type, ConversionClassList>::unpacked_container;
+  };
+  using unpacked_container = tuples::TaggedTuple<UnpackedTag<Tags>...>;
+  using packed_container = tuples::TaggedTuple<Tags...>;
+  // This assumes the first member of the TaggedTuple sets the packed
+  // type. I'm (Nils Deppe) not sure how else to handle this right now.
+  using packed_type =
```
kidder, line 231: so does this mean the tagged tuple must have the same packed_type for each element?  Is there a way to static_assert this?

kidder (approved):
One question that you can squash any changes if you decide to address it in any way

tests/Unit/Framework/Pypp.hpp
nilsdeppe: I don't know how to check it (easily?) since other types are okay, as long as they aren't packed. I.e. if you want to pass in or out an int or double that's fine, it just can't be the first entry in the tuple. I don't know exactly how to assert this check since it's not different packed types that are bad.

kidder: okay, then I am fine with the PR being merged

PR #5749, opened 2024-02-09
Add CLI tool to plot volume data along a line, group plotting commands in CLI

nilsvu:
## Proposed changes

Useful for quick inspection of simulation output. E.g. here's a quick plot reproducing Fig. 5 in https://arxiv.org/abs/gr-qc/0404056:

line

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/PlotAlongLine.py
+++ b/src/Visualization/Python/PlotAlongLine.py
@@ -33,7 +33,7 @@ def _parse_point(ctx, param, value):
     return np.array(list(map(float, value.split(","))))
 
 
-@click.command(name="plot-along-line")
+@click.command(name="along-line")
 @click.argument(
     "h5_files",
     nargs=-1,
--- a/src/Visualization/Python/PlotAlongLine.py
+++ b/src/Visualization/Python/PlotAlongLine.py
@@ -1,0 +1,277 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+import os
+
+import click
+import matplotlib.pyplot as plt
+import numpy as np
+import rich
+
+import spectre.IO.H5 as spectre_h5
+from spectre.IO.Exporter import interpolate_to_points
+
+logger = logging.getLogger(__name__)
+
+
+def _parse_step(ctx, param, value):
+    if value is None:
+        return None
+    if value.lower() == "first":
+        return 0
+    if value.lower() == "last":
+        return -1
+    return int(value)
+
+
+def _parse_point(ctx, param, value):
+    if not value:
+        return None
+    return np.array(list(map(float, value.split(","))))
+
+
+@click.command(name="plot-along-line")
+@click.argument(
+    "h5_files",
+    nargs=-1,
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+)
+@click.option(
+    "--subfile-name",
+    "-d",
+    help="Name of subfile within h5 file containing volume data to plot.",
+)
+@click.option(
+    "--list-vars",
+    "-l",
+    is_flag=True,
+    help="Print available variables and exit.",
+)
+@click.option(
+    "--var",
+    "-y",
+    "vars",
+    multiple=True,
+    help=(
+        "Variables to plot. List any tensor components "
+        "in the volume data file, such as 'Shift_x'."
+    ),
+)
+@click.option(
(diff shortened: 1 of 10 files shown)


src/Visualization/Python/PlotAlongLine.py
```diff
@@ -0,0 +1,249 @@
...
+    "vars",
+    multiple=True,
+    help=(
+        "Variables to plot. List any tensor components "
+        "in the volume data file, such as 'Shift_x'."
+    ),
+)
+@click.option(
+    "--step",
+    type=int,
```
knelli2: You have code in `Clip.py` to allow `first` and `last` as values too. Maybe use that to be consistent?

src/Visualization/Python/PlotAlongLine.py
```diff
@@ -0,0 +1,249 @@
...
+        "Name of the output plot file. If unspecified, the plot is "
+        "shown interactively, which only works on machines with a "
+        "window server."
+    ),
+)
+@click.option(
+    "--line-start",
+    "-A",
+    callback=_parse_point,
+    help="Coordinates of the start of the line through the volume data.",
```
knelli2: Could you specify that these must be a comma separated list? Also, I think this differs from how we specify the origin in `Clip.py`. We should probably make these consistent.

src/Visualization/Python/PlotAlongLine.py
```diff
@@ -0,0 +1,249 @@
...
+    if line_start is None or line_end is None:
+        raise click.UsageError(
+            f"Specify '--line-start' / '-A' and '--line-end' / '-B'."
+        )
+    if len(line_start) != dim or len(line_end) != dim:
+        raise click.UsageError(
+            "Both '--line-start' / '-A' and '--line-end' / '-B' must have"
+            f" {dim} values (the dimension of the volume data)."
+        )
+    line_parameter = np.linspace(0, 1, 200)
```
knelli2: Should we allow the number of points to be an input? I don't really have a preference.

knelli2 (changes requested):
Just a couple consistency things

src/Visualization/Python/PlotAlongLine.py
nilsvu: Added the comment. In `Clip.py` this is hard-coded to 3D so we can use `nargs=3`, but here the dimension is determined from the data files so `nargs` doesn't work and we can only check the number of coordinates later.

src/Visualization/Python/PlotAlongLine.py
nilsvu: Sure why not, added the parameter.

src/Visualization/Python/PlotAlongLine.py
knelli2: What to you think about changing `Clip.py` (not necessarily in this PR) to use comma separated lists as well and then checking the number of args? I think it looks a bit cleaner on the command line

nilsvu: Yea probably better to prefer consistency over the slight advantage that `nargs` is handled natively by `click` (better help string and error msg).

src/Visualization/Python/PlotAlongLine.py
```diff
@@ -52,7 +62,7 @@ def _parse_point(ctx, param, value):
 )
 @click.option(
     "--step",
-    type=int,
+    callback=_parse_step,
     help="Observation step number. Specify '-1' for the last step in the file.",
 )
```
knelli2: So that it's consistent with `Clip`?
```py
@click.option(
    "--time-step",
    "-t",
    callback=_parse_step,
    help=(
        "Select a time step. Specify '-1' or 'last' to select the last time"
        " step. You can also specify 'first' for the first time step."
    ),
)
```

knelli2 (changes requested):
You can squash everything

src/Visualization/Python/PlotAlongLine.py
nilsvu: Currently it's consistent with `PlotPowerMonitors`. Seems like we should make the `render-3d` commands consistent with the `plot` commands.

knelli2 (approved):
Ok yeah there seems to be too much inconsistency between the `plot` and `render-3d` commands to address in this PR. Let's fix those in a follow-up.

nilsvu:
unrelated timeout

PR #5709, opened 2024-01-19
Databox runtime retrieval

nilsdeppe:
## Proposed changes

This adds a method of getting a type-erased DataBox. Some retrieval errors become runtime, but this will allow us to move more code into cpp files, which will hopefully decrease compile time.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5756, opened 2024-02-13
Revert DataBox runtime retrieval to fix wrong numbers in elliptic solver

nilsvu:
## Proposed changes

Somehow #5709 broke the elliptic solver, in that the numbers are slightly changed after a few iterations and eventually differ enough that the solver terminates with a wrong result. I'll look into this more tomorrow, but I'm worried that this also affects other simulations, so maybe it would be better to merge this revert until it's sorted out.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2:
In a recent meeting, I believe @nilsvu determined that this was caused by a very specific case of sub-items/compute items which is tested, but not fixed, in the second commit

knelli2:
@nilsvu Can this be closed now that #5814 is merged?

nilsvu:
Yes!

PR #5758, opened 2024-02-13
Add rotation, expansion and translation maps to Sphere domain

guilara:
## Proposed changes

We add the RotScaleTrans time-dependent maps as options for the Sphere domain

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/Sphere.cpp
+++ b/src/Domain/Creators/Sphere.cpp
@@ -267,11 +267,25 @@ Sphere::Sphere(
       // Build the maps. We only apply the maps in the inner-most shell. The
       // inner radius is what's passed in, but the outer radius is the outer
       // radius of the inner-most shell so we have to see how many shells we
-      // have
+      // have.
+      // The transition region for the rotation, expansion and translation maps
+      // occurs only in the outer-most shell. We require at least one radial
+      // partition, as the transition requires spherical shape.
+
+      if (radial_partitioning_.empty()) {
+        PARSE_ERROR(context,
+                    "The hard-coded translation map requires at least two "
+                    "shells. Use at least one radial partition.");
+      }
+
       std::get<sphere::TimeDependentMapOptions>(time_dependent_options_.value())
-          .build_maps(std::array{0.0, 0.0, 0.0}, inner_radius_,
-                      radial_partitioning_.empty() ? outer_radius_
-                                                   : radial_partitioning_[0]);
+          .build_maps(
+              std::array{0.0, 0.0, 0.0},
+              // Inner shell radii for the Shape map
+              std::pair<double, double>{inner_radius_, radial_partitioning_[0]},
+              // Outer shell radii for the transition region
+              std::pair<double, double>{radial_partitioning_.back(),
+                                        outer_radius_});
     }
   }
 }
@@ -374,9 +388,15 @@ Domain<3> Sphere::create_domain() const {
               time_dependent_options_.value());
 
       // First shell gets the distorted maps.
+      // Last shell gets the transition region for the rotation, expansion and
+      // translation maps
       for (size_t block_id = 0; block_id < num_blocks_; block_id++) {
         const bool include_distorted_map_in_first_shell =
             block_id < num_blocks_per_shell_;
+        // False if block_id is in the last shell
+        const bool use_rigid =
+            block_id + num_blocks_per_shell_ + (fill_interior_ ? 1 : 0) <
+            num_blocks_;
         block_maps_grid_to_distorted[block_id] =
             hard_coded_options.grid_to_distorted_map(
                 include_distorted_map_in_first_shell);
@@ -385,7 +405,7 @@ Domain<3> Sphere::create_domain() const {
                 include_distorted_map_in_first_shell);
         block_maps_grid_to_inertial[block_id] =
             hard_coded_options.grid_to_inertial_map(
-                include_distorted_map_in_first_shell);
+                include_distorted_map_in_first_shell, use_rigid);
       }
     } else {
       const auto& time_dependence = std::get<std::unique_ptr<
--- a/src/Domain/Creators/SphereTimeDependentMaps.cpp
+++ b/src/Domain/Creators/SphereTimeDependentMaps.cpp
@@ -21,16 +21,24 @@
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
 #include "Domain/FunctionsOfTime/PiecewisePolynomial.hpp"
 #include "Domain/FunctionsOfTime/QuaternionFunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/SettleToConstant.hpp"
+#include "Domain/FunctionsOfTime/SettleToConstantQuaternion.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Spherepack.hpp"
 #include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrHorizon.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 
 namespace domain::creators::sphere {
+
 TimeDependentMapOptions::TimeDependentMapOptions(
-    const double initial_time, const ShapeMapOptions& shape_map_options)
+    const double initial_time, const ShapeMapOptions& shape_map_options,
+    std::optional<RotationMapOptions> rotation_map_options,
+    std::optional<ExpansionMapOptions> expansion_map_options,
+    std::optional<TranslationMapOptions> translation_map_options)
     : initial_time_(initial_time),
-      initial_l_max_(shape_map_options.l_max),
-      initial_shape_values_(shape_map_options.initial_values) {}
+      shape_map_options_(shape_map_options),
+      rotation_map_options_(rotation_map_options),
+      expansion_map_options_(expansion_map_options),
+      translation_map_options_(translation_map_options) {}
 
 std::unordered_map<std::string,
                    std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
@@ -46,7 +54,8 @@ TimeDependentMapOptions::create_functions_of_time(
   // their initial expiration time to infinity (i.e. not expiring)
   std::unordered_map<std::string, double> expiration_times{
       {size_name, std::numeric_limits<double>::infinity()},
-      {shape_name, std::numeric_limits<double>::infinity()}};
+      {shape_name, std::numeric_limits<double>::infinity()},
+      {translation_name, std::numeric_limits<double>::infinity()}};
 
   // If we have control systems, overwrite these expiration times with the ones
   // supplied by the control system
@@ -55,16 +64,19 @@ TimeDependentMapOptions::create_functions_of_time(
   }
 
   DataVector shape_zeros{
-      ylm::Spherepack::spectral_size(initial_l_max_, initial_l_max_), 0.0};
+      ylm::Spherepack::spectral_size(shape_map_options_.l_max,
+                                     shape_map_options_.l_max),
+      0.0};
   DataVector shape_func{};
   DataVector size_func{1, 0.0};
 
-  if (initial_shape_values_.has_value()) {
+  if (shape_map_options_.initial_values.has_value()) {
     if (std::holds_alternative<KerrSchildFromBoyerLindquist>(
-            initial_shape_values_.value())) {
-      const ylm::Spherepack ylm{initial_l_max_, initial_l_max_};
-      const auto& mass_and_spin =
-          std::get<KerrSchildFromBoyerLindquist>(initial_shape_values_.value());
+            shape_map_options_.initial_values.value())) {
+      const ylm::Spherepack ylm{shape_map_options_.l_max,
+                                shape_map_options_.l_max};
+      const auto& mass_and_spin = std::get<KerrSchildFromBoyerLindquist>(
+          shape_map_options_.initial_values.value());
       const DataVector radial_distortion =
           1.0 - get(gr::Solutions::kerr_schild_radius_from_boyer_lindquist(
                     inner_radius, ylm.theta_phi_points(), mass_and_spin.mass,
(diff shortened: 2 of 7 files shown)


src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -142,12 +142,65 @@ struct TimeDependentMapOptions {
...
+    using type = Options::Auto<RotationMapOptions, Options::AutoLabel::None>;
+    static std::string name() { return "RotationMap"; }
+    static constexpr Options::String help = {
+        "Options for a time-dependent rotation map about an arbitrary axis. "
+        "Specify 'None' to not use this map."};
+
+    struct InitialAngularVelocity {
+      using type = std::array<double, 3>;
+      static constexpr Options::String help = {"The initial angular velocity."};
+    };
```
knelli2: So for this, it's going to have to be a little different than the BBH domain. Instead of initial angular velocity, we'll need initial quaternion and its first two time derivatives. This is because of the `SettleToConstantQuaterion` FoT.

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -142,12 +142,65 @@ struct TimeDependentMapOptions {
...
+    using type = Options::Auto<ExpansionMapOptions, Options::AutoLabel::None>;
+    static std::string name() { return "ExpansionMap"; }
+    static constexpr Options::String help = {
+        "Options for the expansion map. Specify 'None' to not use this map."};
+
+    struct InitialValues {
+      using type = std::array<double, 2>;
+      static constexpr Options::String help = {
+          "Initial value and deriv of expansion."};
+    };
```
knelli2: Same for expansion. For both the actual expansion, and the outer boundary expansion, each need an initial value and two time derivatives because of the `SettleToConstant`s.

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -196,10 +251,9 @@ struct TimeDependentMapOptions {
    * - Shape: `Shape` (with a size function of time)
    * - Translation: `Translation`
    */
-  void build_maps(
-      const std::array<double, 3>& center, double inner_radius,
-      double outer_radius,
-      std::pair<double, double> translation_transition_radii);
+  void build_maps(const std::array<double, 3>& center, double inner_radius,
+                  double outer_radius,
+                  std::pair<double, double> translation_transition_radii);
```
knelli2: You probably only need one `transition_radius` since the outer boundary radius is already specified. Also this transition radius also applies for expansion, so maybe remove the `translation_` part.

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -223,28 +277,32 @@ struct TimeDependentMapOptions {
...
-   * \brief This will construct the map from `Frame::Grid` to `Frame::Inertial`.
+   * \brief This will construct the map from `Frame::Grid` to
+   * `Frame::Inertial`.
    *
-   * If the argument `include_distorted_map` is true, then this map will have a
-   * `Shape` map (with a size function of time). If it is false, then there will
-   * only be an identity map.
+   * If the argument `include_distorted_map` is true, then this map will
+   * have a `Shape` map (with a size function of time). If it is false, then
+   * there will only be an identity map.
```
knelli2: This is no longer true. There will be rotation, expansion, and translation maps now. Same for the other `map` functions.

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -223,28 +277,32 @@ struct TimeDependentMapOptions {
...
    */
   MapType<Frame::Grid, Frame::Inertial> grid_to_inertial_map(
       bool include_distorted_map, bool use_rigid_translation) const;
 
   inline static const std::string size_name{"Size"};
   inline static const std::string shape_name{"Shape"};
+  inline static const std::string rotation_name{"Rotation"};
+  inline static const std::string expansion_name{"Expansion"};
+  inline static const std::string expansion_outer_name{"ExpansionOuter"};
```
knelli2: Please use `ExpansionOuterBoundary` for the outer name so that it's consistent with the BBH.

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -50,6 +55,9 @@ TimeDependentMapOptions::create_functions_of_time(
   std::unordered_map<std::string, double> expiration_times{
       {size_name, std::numeric_limits<double>::infinity()},
       {shape_name, std::numeric_limits<double>::infinity()},
+      {expansion_name, std::numeric_limits<double>::infinity()},
+      {expansion_outer_name, std::numeric_limits<double>::infinity()},
+      {rotation_name, std::numeric_limits<double>::infinity()},
```
knelli2: For the sphere domain, expansion and rotation won't need expiration times because of the `SettleToConst` FoTs.

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -105,24 +116,63 @@ TimeDependentMapOptions::create_functions_of_time(
                                  {0.0}}},
       expiration_times.at(size_name));
 
+  // ExpansionMap FunctionOfTime
+  // Note: Missing expiration times in SettleToConstant
```
knelli2: You can remove the note. This is expected.

src/Domain/Creators/Sphere.cpp
```diff
@@ -374,9 +386,14 @@ Domain<3> Sphere::create_domain() const {
               time_dependent_options_.value());
 
       // First shell gets the distorted maps.
+      // Last shell gets translation with transition region
       for (size_t block_id = 0; block_id < num_blocks_; block_id++) {
         const bool include_distorted_map_in_first_shell =
             block_id < num_blocks_per_shell_;
+        // False if block_id is in the last shell
+        const bool use_rigid_translation =
```
knelli2: Remove `_translation`

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -195,16 +230,21 @@ struct TimeDependentMapOptions {
    * only be an identity map.
    */
   MapType<Frame::Grid, Frame::Inertial> grid_to_inertial_map(
-      bool include_distorted_map) const;
+      bool include_distorted_map, bool use_rigid_translation) const;
```
knelli2: Also remove `_translation`

tests/InputFiles/Xcts/KerrSchild.yaml
```diff
@@ -60,8 +60,9 @@ DomainCreator:
         InitialValues:
           Mass: *KerrMass
           Spin: *KerrSpin
-      TranslationMap:
-        InitialValues: [*Center, [0., 0., 0.]]
+      RotationMap: None
+      ExpansionMap: None
+      TranslationMap: None
```
knelli2: Do you still need the large timeout when all of these are `None`?

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -131,15 +142,35 @@ struct TimeDependentMapOptions {
...
+    static std::string name() { return "TranslationMap"; }
+    static constexpr Options::String help = {
+        "Options for a time-dependent translation map in that keeps the outer "
+        "boundary fixed."};
+
+    struct InitialValues {
+      using type = std::array<std::array<double, 3>, 2>;
+      static constexpr Options::String help = {
+          "Initial values for the translation map."};
+    };
```
knelli2: I think we'll also need a second derivative here as well.

knelli2 (changes requested):
Thank you for doing this! Please add these changes in a fixup.

tests/InputFiles/Xcts/KerrSchild.yaml
guilara: Yeah, I think it's mostly because of the extra radial partition. We could modify de example to not use the shape map (?)

knelli2: We could use the shape time dependence instead of the hard coded shape maps. It'd look like this (with a couple extra yaml anchors you need to add):

```yaml
    TimeDependentMaps:
      Shape:
        InitialTime: 0.0
        LMax: 10
        Mass: *KerrMass
        Spin: *KerrSpin
        Center: *KerrCenter
        InnerRadius: *InnerRadius
        OuterRadius: *OuterRadius
```

@nilsvu Does that seem ok?

nilsvu: Why did you have to add the extra radial partition? Would be better to keep the domain the same here (log map around the horizon seems important).

knelli2: So with the hardcoded maps, the shape map cannot extend into the region where the expansion/translation map are transitioning to zero because the shape map changes the radius of coordinates and in the transition region the exp/trans map make an assumption that the coords they are using are the grid coordinates. Not some distorted coordinates. The easiest way to avoid this is to just require at least one partition. The shape map is in the innermost shell, and the exp/trans transition is in the outermost shell.

For this test though, if we use the shape time dependence, we can avoid an extra partition and leave it as it was originally because there is no exp/trans maps at all. I think this is the easiest solution for leaving the test as unchanged as possible. And I believe it'll avoid the long timeouts

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: I've added them as InitialValues as in the other cases

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: Added one more initial value for the 2nd derivative

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: Edited docs con maps

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: Renamed

src/Domain/Creators/SphereTimeDependentMaps.cpp
guilara: Removed. I had to leave the one for the unused outer region.

src/Domain/Creators/SphereTimeDependentMaps.cpp
guilara: Removed

src/Domain/Creators/Sphere.cpp
guilara: Removed

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: Removed

tests/InputFiles/Xcts/KerrSchild.yaml
guilara: I went back to the old test and implemented @knelli2 suggestion for the shape map

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: Added

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: Not sure how to remove this. The outer radius variable there is for the shape map. It seems to mean the outer radius of the first shell. Still need the outer radius of the whole sphere domain.

src/Domain/Creators/SphereTimeDependentMaps.cpp
```diff
@@ -117,60 +115,66 @@ TimeDependentMapOptions::create_functions_of_time(
...
                 {{gsl::at(expansion_map_options_.value().initial_values, 0)},
                  {gsl::at(expansion_map_options_.value().initial_values, 1)},
-                 {0.0}}},
+                 {gsl::at(expansion_map_options_.value().initial_values, 2)}}},
             initial_time_, expansion_map_options_.value().decay_timescale);
 
     // ExpansionMap in the Outer regionFunctionOfTime
-    result[expansion_outer_name] =
+    result[expansion_outer_boundary_name] =
         std::make_unique<FunctionsOfTime::PiecewisePolynomial<2>>(
```
knelli2: Ah sorry, I wasn't clear here. The expansion outer boundary FoT *also* needs to be a `SettleToConstant`. Therefore it won't have an expiration time and needs input file options similar to the expansion map.

src/Domain/Creators/SphereTimeDependentMaps.hpp
knelli2: Oh yes I see what it means now. That's actually pretty confusing having `outer_radius` really mean the outer radius of the inner shell. What do you think about a signature like this

```cpp
  void build_maps(const std::array<double, 3>& center, double inner_radius,
                  std::pair<double, double> inner_shell_radii,
                  std::pair<double, double> outer_shell_radii);
```

Feel free to suggest something else if this still isn't clear enough.

knelli2 (changes requested):
Last couple things

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: This notation seems fine. Implemented

src/Domain/Creators/SphereTimeDependentMaps.cpp
guilara: Changed to SettleToConstant and added input file options

knelli2 (commented):
Ok everything looks good! Please squash everything down into one commit. Thank you for doing this!

src/Domain/Creators/SphereTimeDependentMaps.hpp
```diff
@@ -81,21 +83,30 @@ struct TimeDependentMapOptions {
...
+                            RotScaleTransMap>;
 
  public:
   using maps_list =
       tmpl::list<IdentityForComposition<Frame::Grid, Frame::Inertial>,
                  IdentityForComposition<Frame::Grid, Frame::Distorted>,
                  IdentityForComposition<Frame::Distorted, Frame::Inertial>,
-                 GridToDistortedComposition, GridToInertialComposition>;
+                 GridToDistortedComposition, GridToInertialSimple,
+                 GridToInertialComposition, DistortedToInertialComposition>;
```
knelli2: Oh I missed this before, but we can't subtract from this list, but only add to it. This has to do with charm registration stuff and being able to load in old domains (it's annoying, but we don't want to break things). So you need to keep the `domain::CoordinateMap<Frame::Grid, Frame::Inertial, ShapeMap>` in this list.

I should add a comment for this type alias (and the one in BBH). I'll do that after this PR is merged.

knelli2 (changes requested):

src/Domain/Creators/SphereTimeDependentMaps.hpp
guilara: Readded

knelli2 (approved):
Thank you for adding this! Ignoring unrelated test failure.

knelli2:
@guilara Could you please add UPDATE INSTRUCTIONS in the PR description? This is info on how users need to update their input files to use your new changes

PR #5763, opened 2024-02-14
Prevent events used in control system from being factory creatable

### LINKED ISSUE #5762: Control system events show up in available events in input file
# Bug reports:

### Expected behavior:

The control system events shouldn't be available for a user to choose from the input file.

### Current behavior:

Currently they are. I believe this is from #5682. The control system Event wasn't constructible from options, but when I removed it and just used regular events, those *are* constructible from options and I didn't add a way to specify that they *not* be constructible from options.

knelli2:
## Proposed changes

Fixes #5762

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Measurements/BothHorizons.hpp
+++ b/src/ControlSystem/Measurements/BothHorizons.hpp
@@ -6,6 +6,7 @@
 #include <cstddef>
 #include <string>
 
+#include "ControlSystem/Measurements/NonFactoryCreatable.hpp"
 #include "ControlSystem/Protocols/Measurement.hpp"
 #include "ControlSystem/Protocols/Submeasurement.hpp"
 #include "ControlSystem/RunCallbacks.hpp"
@@ -92,9 +93,8 @@ struct BothHorizons : tt::ConformsTo<protocols::Measurement> {
     using interpolation_target_tag = InterpolationTarget<ControlSystems>;
 
     template <typename ControlSystems>
-    using event =
-        intrp::Events::Interpolate<3, InterpolationTarget<ControlSystems>,
-                                   ::ah::source_vars<3>>;
+    using event = NonFactoryCreatableWrapper<intrp::Events::Interpolate<
+        3, InterpolationTarget<ControlSystems>, ::ah::source_vars<3>>>;
   };
 
   using submeasurements = tmpl::list<FindHorizon<::domain::ObjectLabel::A>,
--- a/src/ControlSystem/Measurements/NonFactoryCreatable.hpp
+++ b/src/ControlSystem/Measurements/NonFactoryCreatable.hpp
@@ -1,0 +1,39 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <pup.h>
+
+#include "Utilities/Serialization/CharmPupable.hpp"
+
+/*!
+ * \brief Simple struct that changes a class from being factory creatable, to
+ * non-factory creatable.
+ *
+ * It does this by defining and setting a static constexpr bool
+ * `factory_creatable = false`.
+ *
+ * \tparam FactoryCreatableClass Any class that is already factory creatable.
+ */
+template <typename FactoryCreatableClass>
+struct NonFactoryCreatableWrapper : public FactoryCreatableClass {
+  static constexpr bool factory_creatable = false;
+  using FactoryCreatableClass::FactoryCreatableClass;
+
+  using factory_creatable_class = FactoryCreatableClass;
+
+  /// \cond
+  explicit NonFactoryCreatableWrapper(CkMigrateMessage* /*unused*/) {}
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(NonFactoryCreatableWrapper);  // NOLINT
+  /// \endcond
+};
+
+/// \cond
+// NOLINTBEGIN
+template <typename FactoryCreatableClass>
+PUP::able::PUP_ID NonFactoryCreatableWrapper<FactoryCreatableClass>::my_PUP_ID =
+    0;
+// NOLINTEND
+/// \endcond
--- a/src/ParallelAlgorithms/Interpolation/Events/Interpolate.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Events/Interpolate.hpp
@@ -50,6 +50,7 @@ template <size_t VolumeDim, typename InterpolationTargetTag,
           typename... InterpolatorSourceVarTags>
 class Interpolate<VolumeDim, InterpolationTargetTag,
                   tmpl::list<InterpolatorSourceVarTags...>> : public Event {
+ public:
   /// \cond
   explicit Interpolate(CkMigrateMessage* /*unused*/) {}
   using PUP::able::register_constructor;
--- a/tests/Unit/ControlSystem/Test_Measurements.cpp
+++ b/tests/Unit/ControlSystem/Test_Measurements.cpp
@@ -3,6 +3,9 @@
 
 #include "Framework/TestingFramework.hpp"
 
+#include <memory>
+#include <typeinfo>
+
 #include "ControlSystem/Measurements/BNSCenterOfMass.hpp"
 #include "ControlSystem/Measurements/BothHorizons.hpp"
 #include "ControlSystem/Measurements/CharSpeed.hpp"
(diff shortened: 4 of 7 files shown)


wthrowe:
Possible alternate design that could be handled completely within the control system code:
```c++
template <typename CreatableEvent>
struct NonCreatableWrapper : CreatableEvent {
  static constexpr bool factory_creatable = false;
  using CreatableEvent::CreatableEvent;
  // pup stuff...
};
```
Thoughts?

knelli2:
Yeah I like that, and I think it'll fit well with the other interpolation changes I'm working on. Do you think the `NonCreatableWrapper` should be generally available next to the `Event` base class, or specifically in the control system? Currently only the control system has events that shouldn't be factory creatable, but I can't think of any other scenario where we'd have events that aren't factory creatable.

wthrowe:
Let's keep it in the control system for now.  Easy enough to move it later.  With the template parameter renamed, it could even move to `Options`, since there's nothing specific to Events there, but let's only do that if we have a use.

We may need other non-creatable stuff occasionally, but it should be rare that we need to transform a creatable class to a non-creatable one.

knelli2:
@wthrowe Good suggestion. This was a lot more self-contained.

src/ControlSystem/Measurements/BothHorizons.hpp
```diff
@@ -6,7 +6,9 @@
 #include <cstddef>
 #include <string>
 
+#include "ControlSystem/Measurements/NonFactoryCreatable.hpp"
 #include "ControlSystem/Protocols/Measurement.hpp"
+
```
wthrowe: Remove blank line.

src/ControlSystem/Measurements/NonFactoryCreatable.hpp
```diff
@@ -0,0 +1,19 @@
...
+ * It does this by defining and setting a static constexpr bool
+ * `factory_creatable = false`.
+ *
+ * \tparam FactoryCreatableClass Any class that is already factory creatable.
+ */
+template <typename FactoryCreatableClass>
+struct NonFactoryCreatableWrapper : public FactoryCreatableClass {
+  static constexpr bool factory_creatable = false;
+  using FactoryCreatableClass::FactoryCreatableClass;
+};
```
wthrowe, line 31: I think you need `WRAPPED_PUPable_decl_template(NonFactoryCreatableWrapper)` for serialization to work properly.

wthrowe (changes requested):
Squash directly.

src/ControlSystem/Measurements/NonFactoryCreatable.hpp
knelli2: Currently, these events don't have member data or `pup` functions. I tried adding some member data and a `pup` function to test the serialization, and the serialized (and deserialized) class was the same. I wonder if it's because the `NonFactoryCreatableWrapper` doesn't hold any data, so it just inherits the pup stuff from the `FactoryCreatableClass` and that works ok.

wthrowe: I think you run into trouble if both `T` and `NonFactoryCreatableWrapper<T>` are registered with charm.  The wrapper will inherit the pup stuff, but that will include the mapping of PUP_ID to class, so one of them will deserialize to the wrong type when serialized through a base-class pointer.

You don't need a `pup` function, since the inherited version of that is correct.

knelli2: Have any thoughts about a good way to test this? Only thing I can think of is to register both `T` and `NonFactoryCreatableWrapper<T>` with charm and try to serialize both through a `unique_ptr<::Event>`.

wthrowe: That is how I would do it.

knelli2:
For some reason, the `Interpolate` event had all its members marked `private`, so I made things `public`

tests/Unit/ControlSystem/Test_Measurements.cpp
```diff
@@ -100,6 +105,37 @@ static_assert(
...
+
+  // No equivalence operators so just check dynamic casting
+  CHECK(dynamic_cast<const non_factory_creatable_event*>(
+            non_factory_event.get()) != nullptr);
+  CHECK(dynamic_cast<const factory_creatable_event*>(factory_event.get()) !=
+        nullptr);
+  CHECK(dynamic_cast<const non_factory_creatable_event*>(factory_event.get()) ==
+        nullptr);
+  // Can't test the other way around because the non factory creatable class is
+  // derived from the factory creatable one.
```
wthrowe:
```c++
#include <typeinfo>
...
CHECK(typeid(*serialized_non_factory_event) == typeid(non_factory_creatable_event));
CHECK(typeid(*serialized_factory_event) == typeid(factory_creatable_event));
```
(Also you're currently checking the wrong variables.)

src/ParallelAlgorithms/Interpolation/Events/Interpolate.hpp
```diff
@@ -50,6 +50,7 @@ template <size_t VolumeDim, typename InterpolationTargetTag,
           typename... InterpolatorSourceVarTags>
 class Interpolate<VolumeDim, InterpolationTargetTag,
                   tmpl::list<InterpolatorSourceVarTags...>> : public Event {
+ public:
```
wthrowe, line 53: Worked before because `WRAPPED_PUPable_decl_template` contains a `public:`, but we should definitely not be relying on that.

wthrowe (changes requested):

knelli2:
Ok squashed in the changes in. The test passes with the `PUP_ID` changes and fails without it, as expected

tests/Unit/ControlSystem/Test_Measurements.cpp
```diff
@@ -18,70 +21,119 @@
...
+      serialize_and_deserialize(factory_event);
+
+  const auto get_typeid = [&](const EventPtr& event_ptr) {
+    const auto& event_ref = *event_ptr.get();
+    return typeid(event_ref).name();
+  };
+
+  CHECK(get_typeid(serialized_non_factory_event) ==
+        get_typeid(non_factory_event));
+  CHECK(get_typeid(serialized_factory_event) == get_typeid(factory_event));
```
wthrowe: Why did you do this instead of the thing I suggested?  I don't think this is guaranteed to work.

wthrowe (changes requested):

tests/Unit/ControlSystem/Test_Measurements.cpp
knelli2: When I did your suggestion I got a compilation warning from clang (which we have as an error):
```
error: expression with side effects will be evaluated despite being used as an operand to 'typeid' [-Werror,-Wpotentially-evaluated-expression]
```
I googled it and it's something related to actually calling `.get()` and dereferencing the pointer inside `typeid()`. However, I was not aware that `typeid().name()` wouldn't give unique results

wthrowe: Just silence the warning.  It's just clang being dumb.

wthrowe: For clarification, the thing I'm mostly worried about is that these are c strings, so you may end up with something like how the result of `"Hello" == "Hello"` is unspecified.  But since they give us an equality operator for the objects we should just use it anyway.

tests/Unit/ControlSystem/Test_Measurements.cpp
```diff
@@ -18,70 +21,122 @@
...
+      serialize_and_deserialize(factory_event);
+
+#ifdef __clang__
+#pragma clang diagnostic push
+#pragma clang diagnostic ignored "-Wpotentially-evaluated-expression"
+#endif  // __clang__
+  CHECK(typeid(*serialized_non_factory_event.get()) ==
+        typeid(*non_factory_event.get()));
+  CHECK(typeid(*serialized_factory_event.get()) ==
+        typeid(*factory_event.get()));
```
wthrowe, line 135: This is OK, but why all the redundant `.get()` calls?

wthrowe (approved):

tests/Unit/ControlSystem/Test_Measurements.cpp
knelli2: 🤷🏻

Artifact of previous changes

PR #5768, opened 2024-02-15
Support older control system output in size plotting

knelli2:
## Proposed changes

Also add a missing SPHEREPACK factor to the calculation of the average delta R (and a comment explainin why).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size.hpp
+++ b/src/ControlSystem/ControlErrors/Size.hpp
@@ -395,6 +395,10 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
               DataVector{time_deriv_apparent_horizon.coefficients()[0]}});
     }
 
+    // This is needed because the horizon_00 (and dt) are spherepack coefs, not
+    // spherical harmonic coefs.
+    const double spherepack_factor = sqrt(0.5 * M_PI);
+
     // This is needed for every state
     const double control_error_delta_r = size::control_error_delta_r(
         horizon_00, dt_horizon_00, lambda_00, dt_lambda_00,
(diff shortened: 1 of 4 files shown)


src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -404,7 +408,8 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
             ? std::optional<double>(control_error_delta_r -
                                     delta_r_drift_outward_options_.value()
                                         .outward_drift_velocity -
-                                    (lambda_00 + horizon_00 -
+                                    (lambda_00 +
+                                     spherepack_factor * horizon_00 -
```
nilsvu, line 412: Can you add a test that would have caught this?

nilsvu (changes requested):

src/ControlSystem/ControlErrors/Size.hpp
knelli2: Testing this is extremely difficult because it's just in the mess that's the size control error. The test currently turns off the `delta_r_drift_outward` because there's not a good way to know the expected value of the size control error if we turn it on. Also, it's very hard to come up with parameters such that we transition to the correct state that this quantity will be used in.

So in short, no not really, unless I spend days on this single test.

nilsvu:
Hmm seems like this needs to be made testable, since there was actually a bug here.

knelli2:
I agree it *should* be tested, but it's very non-trivial to test.

src/ControlSystem/ControlErrors/Size.hpp
knelli2: Ok @nilsvu I added a test

nilsvu (approved):
To make testing easier it often helps to separate logic from control flow / parallel stuff. It's easy to test free functions, but hard to test actions etc. No change request for this PR.

src/ControlSystem/ControlErrors/Size.hpp
nilsvu, line 412: Question: how severe is this bug? what impact did it have on simulations?

knelli2: None. In all of my runs I've disabled the `delta_r_drift_outward` option since we haven't needed it.

nilsvu: Oh ok good 👍

PR #5772, opened 2024-02-15
Fix a few bugs related to step-size changing

wthrowe:
Tracked down while testing LTS Adams-Moulton, but at least "Use correct id to check can_change_step_size" can be hit on develop.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/Actions/AdvanceTime.hpp
+++ b/src/Time/Actions/AdvanceTime.hpp
@@ -84,21 +84,27 @@ struct AdvanceTime {
            const gsl::not_null<TimeDelta*> next_time_step,
            const gsl::not_null<AdaptiveSteppingDiagnostics*> diags,
            const TimeStepper& time_stepper, const bool using_error_control) {
+          const bool new_step = next_time_id->substep() == 0;
           if (time_id->slab_number() != next_time_id->slab_number()) {
             ++diags->number_of_slabs;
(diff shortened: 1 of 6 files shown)


src/Time/Actions/AdvanceTime.hpp
```diff
@@ -84,21 +84,27 @@ struct AdvanceTime {
...
           *time_id = *next_time_id;
-          *time_step = next_time_step->with_slab(time_id->step_time().slab());
+
+          if (new_step) {
+            if (time_step->fraction() != next_time_step->fraction()) {
+              ++diags->number_of_step_fraction_changes;
+            }
+
+            *time_step = next_time_step->with_slab(time_id->step_time().slab());
+            *next_time_step = *time_step;
```
knelli2, line 106: Wait are the next time step and the current time step are the same?

wthrowe: Yes.  What did you expect?

knelli2: Naively, I wouldn't expect them to be the same? However, I must be missing something conceptually because this was true before your changes as well. This shouldn't hold up this PR.

knelli2 (approved):

kidder (approved):

PR #5775, opened 2024-02-16
Add EvolvedPosition/Velocity Tags

nikwit:
These tags will be used to integrate the orbit of the scalar charge.

~depends on #5774~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
@@ -192,7 +192,8 @@ struct ObserveCoefficientsTrigger : db::SimpleTag {
 /// @{
 /*!
  * \brief The position and velocity of the scalar charge particle orbiting a
- * central black hole given in inertial coordinates.
+ * central black hole given in inertial coordinates. This compute tag is meant
+ * to be used by the elements.
  */
 template <size_t Dim>
 struct ParticlePositionVelocity : db::SimpleTag {
@@ -208,7 +209,7 @@ struct ParticlePositionVelocityCompute : ParticlePositionVelocity<Dim>,
                                    domain::Tags::FunctionsOfTime>;
   static void function(
       gsl::not_null<std::array<tnsr::I<double, Dim, Frame::Inertial>, 2>*>
-          position,
+          position_velocity,
       const ::ExcisionSphere<Dim>& excision_sphere, double time,
       const std::unordered_map<
           std::string,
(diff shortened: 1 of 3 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -217,6 +218,41 @@ struct ParticlePositionVelocityCompute : ParticlePositionVelocity<Dim>,
 };
 /// @}
 
+/*!
+ * \brief The position of the scalar charge evolved by the worldtube singleton.
+ */
+template <size_t Dim>
+struct EvolvedPosition : db::SimpleTag {
```
knelli2, line 226: It's a bit confusing that you have duplicate tags for the position/velocity of the scalar charge. The notes on the compute tags are helpful, but I'm wondering if you could reuse the `ParticlePositionVelocity` simple tag on both the elements and the worldtube singleton, but then only have different compute tags? That way it's clear the position/velocity are the same thing, but they are just computed differently on different components (and maybe explain how they are computed differently?)

knelli2 (changes requested):
Let me know if changing this will make things more difficult down the line with your upcoming changes (I haven't looked at the other PRs yet)

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: Yes, I agree this is not optimal. One of the things to consider is that the `EvolvedPosition` and `EvolvedVelocity` tags are tensors of size 1 DataVectors as the time stepper can not advance doubles afaik. I try and hide this conversion from DataVectors to doubles in the compute tags but if I were to use this tag on the element components they would have to do this manually.

knelli2: @wthrowe Correct me if I'm wrong, but I thought the time stepper and history were able to handle `double`s?

wthrowe: I believe the full list is:
* `double`
* `std::complex<double>`
* `DataVector`
* `ComplexDataVector`
* `std::array<T, N>`, where I'm skeptical that it will work with `T` anything but `double`
* `Variables`

(i.e., the types with `make_math_wrapper` overloads).

So yes, but really no.  In particular, `Tensor` is missing from the list and `Variables` only holds `DataVector`s and similar.  Adding `Tensor<double>` wouldn't be hard, but that would still only allow you to evolve one `Tensor`, not a system.  The vague plan for ODE systems has been to use single-point `Variables` objects, but this hasn't been tried in practice before.

knelli2: Hmm ok. Seems like it's not trivial to combine them then. In that case, please add a note to the `EvolvedPosition` and `EvolvedVelocity` tag docs that these tags are intended to be used on the worldtube singleton, not the elements.

knelli2 (approved):
Test timeout unrelated

PR #5776, opened 2024-02-16
Add `InitialPositionVelocity` tag

nikwit:
Adds the `InitialPositionAndVelocity` tag. Also, the worldtube system is changed to evolve the position and velocity by using the new tag to initialize them.

~depends on #5774 #5775~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
@@ -36,6 +36,9 @@
 namespace Tags {
 struct Time;
 }  // namespace Tags
+namespace OptionTags {
+struct InitialTime;
+}  // namespace OptionTags
 /// \endcond
 
 namespace CurvedScalarWave::Worldtube {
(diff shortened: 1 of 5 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -189,6 +192,28 @@ struct ObserveCoefficientsTrigger : db::SimpleTag {
...
+    const auto initial_fot = domain_creator->functions_of_time();
+    const auto domain = domain_creator->create_domain();
+    const auto& excision_sphere =
+        domain.excision_spheres().at(excision_sphere_name);
+    ASSERT(excision_sphere.is_time_dependent(),
+           "excision_sphere not time dependent");
+    const auto& maps = excision_sphere.moving_mesh_grid_to_inertial_map();
+    const auto mapped_tuple = maps.coords_frame_velocity_jacobians(
+        excision_sphere.center(), initial_time, initial_fot);
+    return {std::get<0>(mapped_tuple), std::get<3>(mapped_tuple)};
```
knelli2, line 218: At `t = 0` I believe we assume that the grid frame and the inertial frame are the same, so we don't need to call the time dependent maps, you can just use the grid centers and copy them into an inertial tensor. Does the curved scalar wave system assume something different?

knelli2 (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: There are a few time dependent domain creators like `RotationAboutZAxis` or `SphericalCompression` that allow you to set `InitialRotationAngle` or `InitialValue` for the compression factor. 

In most cases the grid and inertial frame should coincide at t=0, but given that this is not enforced I would prefer to initialize it like this - it doesn't cost anything after all.

knelli2: Ok that's fine. In that case, could you add a comment above the `domain_creator->functions_of_time()` to say that we don't care about expiration times here since we're evaluating at the initial time?

nikwit: done and also added a sentence for the docs. thanks for the review!

knelli2 (approved):

knelli2:
Ignoring unrelated test errors

PR #5779, opened 2024-02-16
Use multiple highest modes for Persson TCI and remove initial data TCI

yoonso0-0:
## Proposed changes

Extends functionality of the Persson TCI to take multiple highest modes when computing the L2 norm of a filtered solution. This resembles the algorithm of `amr::Criteria::Persson`.

Also wipe out initial data TCI from input files.

### Upgrade instructions

All input files for executables using DG-FD (subcell) should make following changes in SpatialDiscretization-DiscontinuousGalerkin group
* Remove `InitialData` group, since we are not using initial data TCI anymore.
* Add `PerssonTci` group, to which `PerssonExponent` is moved as the `Exponent` option.
* In the `PerssonTci` group, a new option named `NumHighestModes` is added. This enables users to control how many highest modes to be left in the filtered solution of which norm is computed in the Persson TCI. Use 1 by default since this is the configuration that we have used so far.

Below is an example.

Before:
```
    Subcell:
      RdmpDelta0: 1.0e-7
      RdmpEpsilon: 1.0e-3
      PerssonExponent: 4.0
      InitialData:
        RdmpDelta0: 1.0e-7
        RdmpEpsilon: 1.0e-3
        PerssonExponent: 4.0
```

After:
```
    Subcell:
      PerssonTci:
        Exponent: 4.0
        NumHighestModes: 1
      RdmpDelta0: 1.0e-7
      RdmpEpsilon: 1.0e-3
```


src/Evolution/DgSubcell/PerssonTci.cpp
```diff
@@ -76,7 +74,11 @@ bool persson_tci_impl(const gsl::not_null<DataVector*> filtered_component,
     // to pow<double,double> already provides a certain amount of speed-up, it
     // would be sufficient to stick to the case 2 for now.
     //
-    if (pow(dg_mesh.extents(d) - 1, alpha) * l2Norm(*filtered_component) >
+
+    const double component_norm{l2Norm(component)};
+
+    if (pow(dg_mesh.extents(d) - num_highest_modes, alpha) *
```
knelli2, line 80: You should have a check somewhere (probably in option parsing/initialization) that `num_highest_modes < dg_mesh.extents(i)` for all dims. Otherwise this will likely FPE

knelli2 (changes requested):
I didn't look at anything else in the PR

nilsdeppe (approved):

knelli2 (approved):

PR #5777, opened 2024-02-16
Add `UpdateAcceleration` mutator

nikwit:
The worldtube singleton now integrates the position and velocity of the charge using the geodesic acceleration.

~depends on #5774 #5775 #5776~

PR #5785, opened 2024-02-20
Add some self-force option tags

nikwit:
Adds the tags `Charge`, `Mass` and `ApplySelfForce` which are needed to compute the value of the scalar self-force onto the particle

~depends on #5777~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
@@ -64,6 +64,14 @@ struct Charge {
 };
 
 /*!
+ * \brief Options for the scalar self-force
+ */
+struct SelfForceOptions {
+  static constexpr Options::String help = {"Options for the scalar self-force"};
+  using group = Worldtube;
+};
+
+/*!
  * \brief Name of the excision sphere designated to act as a worldtube
  */
 struct ExcisionSphere {
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
@@ -82,6 +82,17 @@ struct Mass {
   using group = SelfForceOptions;
 };
 
+/*!
+ * \brief Whether to apply the acceleration due to the scalar self-force to the
+ * particle
+ */
+struct ApplySelfForce {
+  using type = bool;
+  static constexpr Options::String help{
+      "Whether to apply the acceleration due to the scalar self-force to the "
+      "particle"};
+  using group = SelfForceOptions;
+};
 
 /*!
  * \brief Name of the excision sphere designated to act as a worldtube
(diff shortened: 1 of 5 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -53,6 +53,15 @@ struct Worldtube {
   static constexpr Options::String help = {"Options for the Worldtube"};
 };
 
+/*!
+ * \brief The value of the scalar charge
+ */
+struct Charge {
+  using type = double;
+  static constexpr Options::String help{"The value of the scalar charge."};
```
knelli2: I'm presuming this is in units of electron charge?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -70,6 +70,17 @@ struct SelfForceOptions {
   using group = Worldtube;
 };
 
+/*!
+ * \brief The mass of the scalar particle
+ */
+struct Mass {
+  using type = double;
+  static constexpr Options::String help{"The mass of the scalar particle."};
```
knelli2: What are the units of this mass? Is it really in terms of M of the BH?

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
```diff
@@ -134,6 +134,7 @@ Worldtube:
   ExpansionOrder: 1
   Charge: 0.5
   SelfForceOptions:
+    ApplySelfForce: false
     Mass: 0.5
```
knelli2: Not for this PR, but if you plan on adding more options to `SelfForceOptions` it may be worth considering rearranging options so the input file can look like

```yaml
Worldtube:
  SelfForceOptions: None
```

when you want to turn off self-force instead of having to supply a bunch of options that won't be used.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: This is a scalar charge, not electric charge. It is also given in units of black hole mass. I added the units to make it clear.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: Yes, I added a comment

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
nikwit: Yes, I was thinking about doing that later. Would I have to make an option creatable-class for this?

nikwit:
squashed in your comments, thanks for the review!

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
knelli2: Yes you'd have to make a class that's option creatable, then use that class wrapped in an `Options::Auto<ClassName, Options::AutoLabel::None>` as the type of option to create.

knelli2 (approved):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
knelli2: Wow 🤦‍♀️. Maybe I shouldn't review PRs in the evening

PR #5771, opened 2024-02-15
Mc logical coordinates

ffoucart:
## Proposed changes

MC packets use logical coordinates instead of inertial coordinates.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

This will make is easier to determine which cell a packet belongs to when using curved meshes.

PR #5786, opened 2024-02-20
Mc emission function

ffoucart:
## Proposed changes

Add a function to emit MC packets according to a given emission rate. The packets are randomly distributed within a FD cell, with isotropic distribution in the fluid frame.

### Upgrade instructions

NA

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Particles/MonteCarlo/EmitPackets.tpp
+++ b/src/Evolution/Particles/MonteCarlo/EmitPackets.tpp
@@ -1,0 +1,143 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp"
+
+#include <cmath>
+
+#include "Evolution/Particles/MonteCarlo/Packet.hpp"
+
+namespace Particles::MonteCarlo {
+
+template <size_t EnergyBins, size_t NeutrinoSpecies>
+void TemplatedLocalFunctions<EnergyBins, NeutrinoSpecies>::emit_packets(
+    const gsl::not_null<std::vector<Packet>*> packets,
+    const gsl::not_null<std::mt19937*> random_number_generator,
+    const double& time_start_step, const double& time_step,
+    const Mesh<3>& mesh,
+    const std::array<std::array<DataVector, EnergyBins>, NeutrinoSpecies>&
+        emission_in_cell,
+    const std::array<DataVector, NeutrinoSpecies>& single_packet_energy,
+    const std::array<double, EnergyBins>& energy_at_bin_center,
+    const Jacobian<DataVector, 4, Frame::Inertial, Frame::Fluid>&
+        inertial_to_fluid_jacobian,
+    const InverseJacobian<DataVector, 4, Frame::Inertial, Frame::Fluid>&
+        inertial_to_fluid_inverse_jacobian) {
+  std::uniform_real_distribution<double> rng_uniform_zero_to_one(0.0, 1.0);
+  // We begin by determining how many packets to create for each cell, neutrino
+  // species, and energy group.
+  const size_t grid_size = mesh.number_of_grid_points();
+  std::array<std::array<std::vector<size_t>, EnergyBins>, NeutrinoSpecies>
+      number_of_packets_to_create_per_cell;
+  size_t number_of_packets_to_create_total = 0;
+  for (size_t s = 0; s < NeutrinoSpecies; s++) {
+    for (size_t g = 0; g < EnergyBins; g++) {
+      gsl::at(gsl::at(number_of_packets_to_create_per_cell, s), g)
+          .resize(grid_size);
+      for (size_t i = 0; i < grid_size; i++) {
+        const double packets_to_create_double =
+            gsl::at(gsl::at(emission_in_cell, s), g)[i] /
+            gsl::at(single_packet_energy, s)[i];
+        size_t packets_to_create_int = floor(packets_to_create_double);
+        if (rng_uniform_zero_to_one(*random_number_generator) <
+            packets_to_create_double -
+                static_cast<double>(packets_to_create_int)) {
+          packets_to_create_int++;
+        }
+        gsl::at(gsl::at(number_of_packets_to_create_per_cell, s), g)[i] =
+            packets_to_create_int;
+        number_of_packets_to_create_total += packets_to_create_int;
+      }
+    }
+  }
+
+  // With the number of packets known, resize the vector holding the packets
+  Packet default_packet(0, 0.0, 0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0, 0.0);
+  const size_t initial_size = packets->size();
+  packets->resize(initial_size + number_of_packets_to_create_total,
+                  default_packet);
+
+  // Allocate memory for some temporary data.
+  double time_normalized = 0.0;
+  std::array<double, 3> coord_normalized{{0.0, 0.0, 0.0}};
+  std::array<double, 3> three_momentum_normalized{{0.0, 0.0, 0.0}};
+  std::array<double, 4> four_momentum_fluid_frame{{0.0, 0.0, 0.0, 0.0}};
+  const Index<3> extents = mesh.extents();
+  const std::array<double, 3> logical_dx{2.0 / static_cast<double>(extents[0]),
+                                         2.0 / static_cast<double>(extents[1]),
+                                         2.0 / static_cast<double>(extents[2])};
+  std::array<double, 3> coord_bottom_cell{{0.0, 0.0, 0.0}};
+
+  // Now create the packets
+  size_t next_packet_index = initial_size;
+  for (size_t x_idx = 0; x_idx < extents[0]; x_idx++) {
+    gsl::at(coord_bottom_cell, 0) =
+        -1.0 + 2.0 * static_cast<double>(x_idx) * gsl::at(logical_dx, 0);
+    for (size_t y_idx = 0; y_idx < extents[1]; y_idx++) {
+      gsl::at(coord_bottom_cell, 1) =
+          -1.0 + 2.0 * static_cast<double>(y_idx) * gsl::at(logical_dx, 1);
+      for (size_t z_idx = 0; z_idx < extents[2]; z_idx++) {
+        const size_t idx = mesh.storage_index(Index<3>{{x_idx, y_idx, z_idx}});
+        gsl::at(coord_bottom_cell, 2) =
+            -1.0 + 2.0 * static_cast<double>(z_idx) * gsl::at(logical_dx, 2);
+        for (size_t s = 0; s < NeutrinoSpecies; s++) {
+          for (size_t g = 0; g < EnergyBins; g++) {
+            for (size_t p = 0;
+                 p < gsl::at(gsl::at(number_of_packets_to_create_per_cell, s),
+                             g)[idx];
+                 p++) {
+              detail::draw_single_packet(&time_normalized, &coord_normalized,
+                                         &three_momentum_normalized,
+                                         random_number_generator);
+              Packet& current_packet = (*packets)[next_packet_index];
+              current_packet.species = s;
+              current_packet.time =
+                  time_start_step + time_normalized * time_step;
+              current_packet.number_of_neutrinos =
+                  gsl::at(gsl::at(single_packet_energy, s), idx) /
+                  gsl::at(energy_at_bin_center, g);
+              current_packet.index_of_closest_grid_point = idx;
+              // 4_momentum_fluid_frame contains p^mu in an orthornormal
+              // coordinate system moving with the fluid.
+              gsl::at(four_momentum_fluid_frame, 0) =
+                  gsl::at(energy_at_bin_center, g);
+              for (size_t d = 0; d < 3; d++) {
+                current_packet.coordinates.get(d) =
+                    gsl::at(coord_bottom_cell, d) +
+                    gsl::at(coord_normalized, d) * gsl::at(logical_dx, d);
+                gsl::at(four_momentum_fluid_frame, d + 1) =
+                    gsl::at(three_momentum_normalized, d) *
+                    gsl::at(four_momentum_fluid_frame, 0);
+              }
+
+              // The packet stores p^t and p_i in the inertial frame. We
+              // transform the momentum accordingly.
+              current_packet.momentum_upper_t = 0.0;
+              for (size_t d = 0; d < 4; d++) {
+                current_packet.momentum_upper_t +=
+                    inertial_to_fluid_inverse_jacobian.get(0, d)[idx] *
+                    gsl::at(four_momentum_fluid_frame, d);
+              }
+              for (size_t d = 0; d < 3; d++) {
+                // Multiply by -1.0 because p_t = -p^t in an orthonormal frame
+                current_packet.momentum.get(d) =
+                    (-1.0) * gsl::at(four_momentum_fluid_frame, 0) *
+                    inertial_to_fluid_jacobian.get(d + 1, 0)[idx];
+                for (size_t dd = 0; dd < 3; dd++) {
+                  current_packet.momentum.get(d) +=
+                      gsl::at(four_momentum_fluid_frame, dd + 1) *
+                      inertial_to_fluid_jacobian.get(d + 1, dd + 1)[idx];
+                }
+              }
+              next_packet_index++;
+            }
+          }
+        }
+      }
+    }
+  }
+}
+
+}  // namespace Particles::MonteCarlo
--- a/src/Evolution/Particles/MonteCarlo/EvolvePackets.cpp
+++ b/src/Evolution/Particles/MonteCarlo/EvolvePackets.cpp
@@ -1,9 +1,9 @@
 // Distributed under the MIT License.
 // See LICENSE.txt for details.
 
-#include "EvolveMonteCarloPackets.hpp"
+#include "Evolution/Particles/MonteCarlo/EvolvePackets.hpp"
 
-#include "Evolution/Particles/MonteCarlo/MonteCarloPacket.hpp"
+#include "Evolution/Particles/MonteCarlo/Packet.hpp"
 
 namespace Particles::MonteCarlo {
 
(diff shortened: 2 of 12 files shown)


ffoucart:
Depends on #5771 
Only the last commit is new.

ffoucart:
@nilsdeppe  @kidder 
A draft of the function creating (emitting) MC packets. There are a few things worth discussing here, which I think provide useful example of choices we have to make for the MC code:
(a) How do we deal with the number of species and number of energy groups? Currently, I templated the function on those two parameters. This allows me to use std::array for things like the emissivity... but we'll need to add macros to generate the function for a range of reasonable choices. Alternatively, I could use std::vector everywhere, and not use templates.
(b) Related point: MC will have a lot of data structures like the emissivity, which depends on species, neutrino energy, and spatial location. Here, I used
`const std::array<std::array<DataVector, EnergyBins>, NeutrinoSpecies>& emission_in_cell`
but I'm not married to that choice... It would be good to agree on a format before creating many more such functions though.
(c) I need the Jacobian and InverseJacobian for the map from inertial to fluid coordinates, and vice-versa. I'm not 100% sure of the conventions for how these are defined. What I coded so far seems internally self-consistent (the main thing the test of the function does so far is verified that the energy in the fluid frame is indeed correct), but maybe not consistent with standard spectre conventions. It doesn't stricly need to be defined as a jacobian of course, but that make identifying what the data means easier...
(d) I am passing a gsl random number generator to the emission function for now. If that's not how we want to handle RNG, that'd be good to know.
(e) In the test , I impose a seed of 0 to the RNG, for reproducibility. We could let it be more random though, if you prefer.

ffoucart:
This PR also adds a 'species' member to the Packet class. This should have been there from the start, but I forgot to include it in previous PRs (as it only becomes important when accessing emission/absorption/scattering rates...)

ffoucart:
@nilsdeppe @kidder I made the structural changes discussed at our meeting last week; this should now be ready for actual code review.

src/Evolution/Particles/MonteCarlo/EmitMonteCarloPackets.tpp
```diff
@@ -0,0 +1,162 @@
...
+
+#include "Evolution/Particles/MonteCarlo/MonteCarloPacket.hpp"
+
+namespace Particles::MonteCarlo {
+
+namespace detail {
+// Draw a single packet with homogeneouse spatial distribution
+// in [0,1]x[0,1]x[0,1], energy of 1, isotropic momentum
+// distribution, and time in [0,1].
+void DrawSinglePacket(double& time, std::array<double, 3>& coord,
```
nilsdeppe: `DrawSinglePacket` -> `draw_single_packet`

Can you make the quantities you change be passed in as `const gsl::not_null` and everything else be `const&`?

Move this function to the `.cpp` file because otherwise when you have another place you include this `tpp` you will get a linking error

src/Evolution/Particles/MonteCarlo/EmitMonteCarloPackets.tpp
```diff
@@ -0,0 +1,162 @@
...
+  const double sin_theta = sqrt(1.0 - cos_theta * cos_theta);
+  gsl::at(momentum, 0) = sin_theta * cos(phi);
+  gsl::at(momentum, 1) = sin_theta * sin(phi);
+  gsl::at(momentum, 2) = cos_theta;
+}
+}  // namespace detail
+
+template <size_t EnergyBins, size_t NeutrinoSpecies>
+void MonteCarlo<EnergyBins,NeutrinoSpecies>::emit_packets(
+    gsl::not_null<std::vector<Packet>*> packets, const double& time_start_step,
```
nilsdeppe: `const gsl::not_null`

src/Evolution/Particles/MonteCarlo/MonteCarloDeclaration.hpp
```diff
@@ -0,0 +1,60 @@
...
+                      const gsl_rng& random_number_generator);
+} // namespace detail
+
+
+// Structure containing Monte-Carlo function templated on EnergyBins
+// and/or NeutrinoSpecies
+template <size_t EnergyBins, size_t NeutrinoSpecies>
+struct MonteCarlo{
+
+// Function emitting packets of total energy emission_in_cell for each grid
```
nilsdeppe: Could you change this to a doxygen comment? (`///`)

Could you also write up/copy paste the equations being solved? This makes them easier to review and also will help someone else find bugs in the future

src/Evolution/Particles/MonteCarlo/MonteCarloDeclaration.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+namespace detail{
+
+void DrawSinglePacket(double& time, std::array<double, 3>& coord,
+                      std::array<double, 3>& momentum,
+                      const gsl_rng& random_number_generator);
+} // namespace detail
+
+
+// Structure containing Monte-Carlo function templated on EnergyBins
```
nilsdeppe: Could you change this to a doxygen comment? (`///`)

src/Evolution/Particles/MonteCarlo/MonteCarloDeclaration.hpp
```diff
@@ -0,0 +1,60 @@
...
+
+/// Items related to the evolution of particles
+/// Items related to Monte-Carlo radiation transport
+namespace Particles::MonteCarlo {
+
+struct Packet;
+
+namespace detail{
+
+void DrawSinglePacket(double& time, std::array<double, 3>& coord,
```
nilsdeppe: `DrawSinglePacket` -> `draw_single_packet`

src/Evolution/Particles/MonteCarlo/MonteCarloDeclaration.hpp
```diff
@@ -0,0 +1,60 @@
+// Distributed under the MIT License.
```
nilsdeppe: Can you rename this and the `cpp` file to `MonteCarloPackets.hpp`? Given everything in this directory is for MC (hence the name of the directory), maybe we can just name things `EmitPackets`, `Packets`, and `Packet`?

src/Evolution/Particles/MonteCarlo/MonteCarloDeclaration.hpp
```diff
@@ -0,0 +1,60 @@
...
+void DrawSinglePacket(double& time, std::array<double, 3>& coord,
+                      std::array<double, 3>& momentum,
+                      const gsl_rng& random_number_generator);
+} // namespace detail
+
+
+// Structure containing Monte-Carlo function templated on EnergyBins
+// and/or NeutrinoSpecies
+template <size_t EnergyBins, size_t NeutrinoSpecies>
+struct MonteCarlo{
```
nilsdeppe: Would `Packets` be a better name? Unclear, but then at least `emit_packets` could just be `emit` since `Packets::emit` is reasonably clear, I think.

tests/Unit/Evolution/Particles/MonteCarlo/CMakeLists.txt
```diff
@@ -4,6 +4,7 @@
 set(LIBRARY "Test_MonteCarlo")
 
 set(LIBRARY_SOURCES
+  Test_MonteCarloEmission.cpp
```
nilsdeppe: Should have the same name as the `tpp` file. E.g. `Test_EmitMonteCarloPackets.cpp` or `Test_EmitPackets.cpp` if you decide to rename

src/Evolution/Particles/MonteCarlo/MonteCarloDeclaration.hpp
```diff
@@ -0,0 +1,60 @@
...
+#pragma once
+
+#include <gsl/gsl_rng.h>
+#include <vector>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/Gsl.hpp"
+
```
nilsdeppe:
```
/// \cond
namespace Frame {
struct Fluid;
} // namespace Frame
/// \endcond
```

nilsdeppe (changes requested):

ffoucart:
Last commit has a lot of changes mainly due to file being moved around... but I think that it gets to all the comments in the previous review.
(If it's easier for me to just squash the four commits by now, let me know...)

src/Evolution/Particles/MonteCarlo/EmitPackets.tpp
```diff
@@ -3,42 +3,19 @@
 
 #pragma once
 
-#include "MonteCarloDeclaration.hpp"
+#include "TemplatedLocalFunctions.hpp"
```
nilsdeppe: Can you add the full path here? Just to avoid any possible issues with name collisions in the future

ffoucart:
Last commit attempts to add const gsl::not_null in cpp. 
Also rebased against develop to fix failing MC test.

src/Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp
```diff
@@ -0,0 +1,72 @@
...
+  /// energy single_packet_energy in the fluid frame. The number of packets
+  /// is, in each bin b and for each species s,
+  /// N = emission_in_cell[s][b] / single_packet_energy[s]
+  /// and randomly rounded up or down to get integer number of packets (i.e.
+  /// if we want N=2.6 packets, there is a 60 percent chance of creating a 3rd
+  /// packet). The position of the packets is drawn from a homogeneous
+  /// distribution in the logical frame, and the direction of propagation of the
+  /// neutrinos is drawn from an isotropic distribution in the fluid frame.
+  /// Specifically, the 4-momentum of a packet of energy nu in the fluid frame
+  /// is p^t = nu; p^x = nu * sin(theta) * cos(phi); p^y = nu * sin(theta) *
```
nilsdeppe: Can you render the equations as latex? E.g.
```cpp
/// \f$p^t = \nu\f$
/// \f{align}{
/// p^x &= \nu \sin(\theta) \cos(\phi) \\
/// p^y &= \nu \sin(\theta) \sin(\phi)
/// \f}
```

src/Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp
```diff
@@ -0,0 +1,72 @@
...
+  /// sin(phi) p^z = nu * cos(theta) with cos(theta) drawn from a uniform
+  /// distribution in [-1,1] and phi from a uniform distribution in [0,2*pi]. We
+  /// transform to the inertial frame p_t and p^x,p^y,p^z using the
+  /// jacobian/inverse jacobian passed as option. The number of neutrinos in
+  /// each packet is defined as n = single_packet_energy[s] /
```
nilsdeppe: For code blocks can you use ticks, e.g.
```cpp
/// defined as `n = single_packet_energy[s] / energy_at_bin_center[b]` but note
```
You might have to be careful about how this gets split across lines. I don't remember if Doxygen is okay or not with that.

You can run `make doc` or `make doc-check` (the latter does some simple textual sanity checks) to render the documentation. I usually then tar the directory and download it to my laptop if I rendered it on Wheeler.

src/Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.cpp
```diff
@@ -0,0 +1,37 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "TemplatedLocalFunctions.hpp"
```
nilsdeppe: Can you use the full path? We want to avoid possible collisions in the future. E.g.
```cpp
#include "Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp"
```

src/Evolution/Particles/MonteCarlo/Packet.hpp
```diff
@@ -32,6 +34,10 @@ struct Packet {
     momentum[2] = p_z_;
   }
 
+  // Species of neutrinos (in the code, just an index used to access the
+  // right interaction rates; typically 0=nu_e, 1=nu_a, 2=nu_x)
```
nilsdeppe: Not a concrete suggestion, but we will probably want some easier way to index this. E.g. an `enum` or something with the tags order. Just something that we will want to think about in the future

src/Evolution/Particles/MonteCarlo/Packet.cpp
```diff
@@ -1,7 +1,7 @@
 // Distributed under the MIT License.
 // See LICENSE.txt for details.
 
-#include "MonteCarloPacket.hpp"
+#include "Packet.hpp"
```
nilsdeppe:
```cpp
#include "Evolution/Particles/MonteCarlo/Packet.hpp"
```
for same reason as other place

tests/Unit/Evolution/Particles/MonteCarlo/Test_EmitPackets.cpp
```diff
@@ -0,0 +1,91 @@
...
+
+  const double time = 0.0;
+  const double time_step = 0.1;
+
+  MAKE_GENERATOR(generator);
+  const Mesh<3> mesh(3, Spectral::Basis::FiniteDifference,
+                     Spectral::Quadrature::CellCentered);
+
+  // Zero vector for tensor creations
+  DataVector zero_dv(27);
```
nilsdeppe: `const DataVector zero_dv(27, 0.0);` then delete next line

tests/Unit/Evolution/Particles/MonteCarlo/Test_EmitPackets.cpp
```diff
@@ -0,0 +1,91 @@
...
+      inverse_jacobian = make_with_value<
+          InverseJacobian<DataVector, 4, Frame::Inertial, Frame::Fluid>>(
+          zero_dv, 0.0);
+  Jacobian<DataVector, 4, Frame::Inertial, Frame::Fluid> jacobian =
+      make_with_value<Jacobian<DataVector, 4, Frame::Inertial, Frame::Fluid>>(
+          zero_dv, 0.0);
+
+  // Set data
+  for (size_t s = 0; s < 2; s++) {
+    for (size_t i = 0; i < 27; i++) {
```
nilsdeppe: `27` -> `single_packet_energy[0].size()` just in case the size gets adjusted at some point

tests/Unit/Evolution/Particles/MonteCarlo/Test_EmitPackets.cpp
```diff
@@ -0,0 +1,91 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
```
nilsdeppe, line 3: I believe we ask that include order here is:
```cpp
#include "Framework/TestingFramework.hpp"

...
```
The reason is that _sometimes_ the order matters for tests and it's easier to just always do it that way than to remember when it does and when it doesn't (I have forgotten)

tests/Unit/Evolution/Particles/MonteCarlo/Test_Packet.cpp
```diff
@@ -3,8 +3,8 @@
 
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
-#include "Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.hpp"
-#include "Evolution/Particles/MonteCarlo/MonteCarloPacket.hpp"
+#include "Evolution/Particles/MonteCarlo/EvolvePackets.hpp"
+#include "Evolution/Particles/MonteCarlo/Packet.hpp"
 #include "Framework/TestingFramework.hpp"
```
nilsdeppe: Can you move this up in this file too for the same reason?

nilsdeppe (changes requested):
A few more minor things. Would it make sense to just squash this into one commit?

src/Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp
```diff
@@ -0,0 +1,80 @@
...
+/// \f$p^y = \nu * sin(\theta) * sin(\phi)\f$
+/// \f$p^z = \nu * cos(theta)\f$
+/// with \f$cos(\theta)\f$ drawn from a uniform distribution in [-1,1] and
+/// \f$\phi\f$ from a uniform distribution in \f$[0,2*\pi]\f$. We
+/// transform to the inertial frame \f$p_t\f$ and \f$p^x,p^y,p^z\f$ using the
+/// jacobian/inverse jacobian passed as option. The number of neutrinos in
+/// each packet is defined as
+/// `n = single_packet_energy[s] / energy_at_bin_center[b]`
+/// Note that the packet energy is in code units and energy of a bin in MeV.
+///
```
ffoucart, line 64: I'm still having some issue with this doxygen comment.
(1) For some reason it doesn't appear on the `TemplatedLocalFunctions` page
(2) If I use
```
/// `\f{aligh}
/// p^x =& A \\
/// p^y =& B
/// \f`
```
clang-tidy complains about a multi-line comment, but if instead I use
```
/*!
 * Comment here
 * More comment
 * /
```
I get a clang-tidy error saying that `/*!` is bad syntax (I was copying the syntax from the HLL file...)

ffoucart:
Ok, I think all comments are addressed and merged in a single commit. I am still having minor issues with the rendering of Doxygen on my local version (the equations don't seem to be processed by TeX..), but the rest looks ok now.

ffoucart:
Fix multiline comment.

kidder (dismissed):
you'll need to rebase

ffoucart:
It seems to have merged with the other MC PR without conflict...

src/Evolution/Particles/MonteCarlo/EvolvePackets.cpp
```diff
@@ -1,9 +1,9 @@
 // Distributed under the MIT License.
 // See LICENSE.txt for details.
 
-#include "EvolveMonteCarloPackets.hpp"
+#include "EvolvePackets.hpp"
```
kidder: oops, we missed changing this include to a full path

kidder (changes requested):
you can squash the one change

kidder (approved):

nilsdeppe (approved):

PR #5787, opened 2024-02-20
Add `BackgroundQuantities` tag

nikwit:
Adds a `TimeDilationFactor` simple  tag as well as a `BackgroundQuantities` compute tag. The worldtube will run an iterative scheme when it applies the scalar self-force, where several iterations of computations are done each time step (to be added shortly). The  `BackgroundQuantities` tag is there to cache all the quantities that do not change between iterations, i.e. only depend on the position and velocity of the scalar charge.

~depends on #5785~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
@@ -20,6 +20,8 @@
 #include "NumericalAlgorithms/Spectral/Quadrature.hpp"
 #include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp"
 #include "PointwiseFunctions/GeneralRelativity/GeodesicAcceleration.hpp"
+#include "PointwiseFunctions/GeneralRelativity/InverseSpacetimeMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/SpacetimeMetric.hpp"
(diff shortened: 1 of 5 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -173,9 +175,43 @@ void GeodesicAccelerationCompute<Dim>::function(
...
+  double temp = metric.get(0, 0);
+  for (size_t i = 0; i < Dim; ++i) {
+    temp += 2. * metric.get(i + 1, 0) * velocity.get(i);
+    for (size_t j = 0; j < Dim; ++j) {
+      temp += metric.get(i + 1, j + 1) * velocity.get(i) * velocity.get(j);
+    }
+  }
+  get(get<TimeDilationFactor>(*result)) = sqrt(-1. / temp);
+  get<gr::Tags::SpacetimeMetric<double, Dim>>(*result) = metric;
+  get<gr::Tags::InverseSpacetimeMetric<double, Dim>>(*result) = inverse_metric;
```
knelli2: `std::move`?

knelli2 (changes requested):
You can squash this in. For the clang-tidy fix, can you name the commit something like "Ignore clang-tidy errors from inside HDF5"?

nikwit:
squashed in, thanks for the review!

knelli2 (approved):

PR #5788, opened 2024-02-20
Use hydro speeds in GRMHD HLL where magnetic field is zero

nilsdeppe:
## Proposed changes

With taggedtuple support in pypp the boundary condition python files can be cleaned up, but I don't have the energy right now to do that huge refactor... 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Evolution/Systems/GrMhd/ValenciaDivClean/BoundaryCorrections/Hll.cpp
```diff
@@ -61,26 +68,116 @@ double Hll::dg_package_data(
...
       get(*packaged_largest_ingoing_char_speed) =
-          -get(lapse) - get(shift_dot_normal) - get(*normal_dot_mesh_velocity);
-    } else {
+          (get(v0) * (1.0 - get(sound_speed_squared)) - discriminant) /
+              (1.0 - get(v_squared) * get(sound_speed_squared)) -
+          get(shift_dot_normal);
+
+      for (size_t i = 0; i < get(rest_mass_density).size(); i++) {
+        if (get(rest_mass_density)[i] < 1.e-08) {
+          double alpha = 0.0;  // get(rest_mass_density)[i]/1.e-08;
```
kidder: confusion... why is alpha 0 and then we are doing math with alpha?

kidder (changes requested):

src/Evolution/Systems/GrMhd/ValenciaDivClean/BoundaryCorrections/Hll.cpp
nilsdeppe: @ffoucart do we still need this?

ffoucart: It used to be necessary for stability of the atmosphere. Effectively, it makes the code use the speed of light as characteristic speed at low density. I have not tried to remove that correction though, and it's probably not a great idea to keep the hard-coded 1.e-8 in the final version.
My preference, if possible, would be to get rid of the alpha, just set the speeds to the speed of light at low density, and have the density threshold as an specifiable input that we can test on merger simulations and maybe lower / remove at a later time if it proves unnecessary.

ffoucart:
The only difference with what I used for BNS mergers seems to be that the sound speed is now calculated within the equation of state class. Assuming that this was already reviewed separately, I'm good with this PR (up to the discussion above on the treatment of low-density regions).

ffoucart:
If you want me to try some simulations with different thresholds for the low-density regions, let me know.

nilsdeppe:
Okay, great! I'm making a few other minor cleanup changes. I'll add a parameter for the density cutoff for atmosphere and will just use the light speeds there. Then we can test disabling it if we want but can re-enable later.

nilsdeppe:
I pushed fixups. @ffoucart I changed the calculation, it didn't agree with what was in the Rezzolla book as far as I could tell

ffoucart:
The formula we had in spec comes from Font. I don't have the Rezzola book,
but I can have a look if there is a disagreement.

nilsdeppe:
Here's the tex for the equations I switched to: https://spectre-code.org/namespaceRelativisticEuler_1_1Valencia.html#a8a63b3bfe1b17683401ba9abd4d6f7f4

Could this be related to grid vs. inertial frame solves?

src/Evolution/Systems/GrMhd/ValenciaDivClean/BoundaryCorrections/Hll.cpp
```diff
@@ -109,63 +139,44 @@ double Hll::dg_package_data(
...
+      get(sound_speed_squared) = clamp(get(sound_speed_squared), 0.0, 1.0);
 
-      // Get v0 from Font
-      Scalar<DataVector> v0 = tilde_d;
-      dot_product(make_not_null(&v0), spatial_velocity, normal_covector);
-      get(v0) *= get(lapse);
+      // Compute v_dot_normal, v^i n_i
+      dot_product(make_not_null(&v_dot_normal), spatial_velocity,
+                  normal_covector);
+      get(v_dot_normal) *= get(lapse);
```
ffoucart: I think our two original formulae are the same, but you are having additional lapse factors by multiplying v_dot_normal here (giving you my v0), then still multiplying, but using this as v_n in the discriminant and still multiplying by the lapse in the final expression. In your formulation, I think you don't want L147. Then you should get an expression equivalent to mine, if I am not mistaken.

nilsdeppe: You're right, I've removed this!

ffoucart (dismissed):
Looks good to me. As a sanity check, you could run the first few hours of evolution of the lowest resolution BNS system, e.g. on wheeler:
/panfs/ds09/sxs/ffoucart/Runs/Spectre/BNS_Gamma2/ConvergenceTest/5Points
The maximum density in particular is quite sensitive to the HLL solver, so checking that you get the same result as me would be a good way to verify consistency of the new implementation.

src/Evolution/Systems/GrMhd/ValenciaDivClean/BoundaryCorrections/Hll.cpp
```diff
@@ -61,26 +77,118 @@ double Hll::dg_package_data(
...
+      get(sound_speed_squared) = clamp(get(sound_speed_squared), 0.0, 1.0);
+
+      // Compute v_dot_normal, v^i n_i
+      dot_product(make_not_null(&v_dot_normal), spatial_velocity,
+                  normal_covector);
+
+      // Compute v^2=v^i v_i
+      dot_product(make_not_null(&v_squared), spatial_velocity,
+                  spatial_velocity_one_form);
+      get(v_squared) = clamp(get(v_squared), 0.0, 1.0 - 1.0e8);
```
kidder: wait what? e-08?

src/Evolution/Systems/GrMhd/ValenciaDivClean/BoundaryCorrections/Hll.cpp
```diff
@@ -61,26 +77,118 @@ double Hll::dg_package_data(
...
+              get(lapse_over_one_minus_v2_cs2)[i] *
+                  (get(v_dot_normal_times_one_minus_cs2)[i] - discriminant[i]) -
+              get(shift_dot_normal)[i];
+        }
+      }
+    }
+
+    // Correct for mesh velocity
     if (normal_dot_mesh_velocity.has_value()) {
       get(*packaged_largest_outgoing_char_speed) =
```
kidder: is there a reason not to use -=?

kidder (changes requested):

kidder (dismissed):

kidder:
@nilsdeppe is there a reason this didn't get merged?  If not, it needs a rebase...

knelli2:
@nilsdeppe will look into this

nilsdeppe:
I've rebased this and made sure it agrees with the paper. Here's an image of the inspiral trajectories from a few different runs I've done trying to improve the domain decomposition and tuning other parameters. The point is they all agree.
image

kidder (approved):

PR #5789, opened 2024-02-21
Support discontinuous fluxes in elliptic DG

nilsvu:
## Proposed changes

Three changes to the elliptic DG scheme, mainly to support elasticity with layered materials as published in https://arxiv.org/abs/2111.06893. See commit messages for details.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/InputFiles/Elasticity/SingleCoatingMirror.yaml
+++ b/tests/InputFiles/Elasticity/SingleCoatingMirror.yaml
@@ -1,0 +1,174 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+Executable: SolveElasticity3D
+Testing:
+  Check: parse
+ExpectedOutput:
+  - MirrorReductions.h5
+  - MirrorVolume0.h5
+OutputFileChecks:
+  - Label: Volume integrals
+    Subfile: VolumeIntegrals.dat
+    FileGlob: MirrorReductions.h5
+    SkipColumns: [0, 1, 2]
+    ExpectedData: [[1.47912989657707e-02]]
+    AbsoluteTolerance: 1.e-4
+  - Label: Coating integrals
+    Subfile: VolumeIntegralsLayer0.dat
+    FileGlob: MirrorReductions.h5
+    SkipColumns: [0, 1, 2]
+    ExpectedData: [[3.12266515383387e-04]]
+    AbsoluteTolerance: 1.e-4
+  - Label: Substrate integrals
+    Subfile: VolumeIntegralsLayer1.dat
+    FileGlob: MirrorReductions.h5
+    SkipColumns: [0, 1, 2]
+    ExpectedData: [[1.44790324503873e-02]]
+    AbsoluteTolerance: 1.e-4
+
+---
+
+Background:
+  # Only used to compute errors, which quantify how far away the solution is
+  # from the half-space mirror solution. Has no effect on the numerical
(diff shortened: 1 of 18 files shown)


tests/Unit/Elliptic/DiscontinuousGalerkin/Test_DgOperator.cpp
```diff
@@ -1107,6 +1109,141 @@ SPECTRE_TEST_CASE("Unit.Elliptic.DG.Operator", "[Unit][Elliptic]") {
...
+              elliptic::BoundaryConditions::AnalyticSolution<system>>(
+              analytic_solution.get_clone(),
+              elliptic::BoundaryConditionType::Dirichlet)};
+      const ElementId<3> center_id{6};
+      const ElementId<3> wedge_id{0};
+      using Vars = Variables<tmpl::list<Var<Poisson::Tags::Field>>>;
+      using OperatorVars =
+          Variables<tmpl::list<DgOperatorAppliedTo<Var<Poisson::Tags::Field>>>>;
+      Vars vars_rnd_center{27};
+      get(get<Var<Poisson::Tags::Field>>(vars_rnd_center)) = DataVector{
```
nilsdeppe, line 1142: Can you make these approximate comparisons? We will almost inevitably run on different hardware or compilers that will break exact comparison.

nilsvu:
@nilsdeppe would you be able to review this?

nilsdeppe:
Sure, once it passes CI I can look

nilsvu:
CI should be fixed now. Either way, can you do #5791 first? I rebased on that to have the extra regression tests here.

tests/InputFiles/Elasticity/SingleCoatingMirror.yaml
```diff
@@ -0,0 +1,174 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+Executable: SolveElasticity3D
```
nilsdeppe, line 4: Does this mean we build yet another executable in CI? If so, can we replace an existing build instead?

nilsdeppe (changes requested):
I don't have a concrete suggestion here, just a thing to note comparing to the evolution code. We remove a lot of memory allocations that used to be done on interfaces/mortars, which helped quite a bit with performance in multithreaded cases

tests/Unit/Elliptic/DiscontinuousGalerkin/Test_DgOperator.cpp
nilsvu: The comparisons are done with our standard approx. Here I just assign the random test data, the comparisons are elsewhere in this file.

tests/InputFiles/Elasticity/SingleCoatingMirror.yaml
nilsvu: No the HalfSpaceMirror.yaml already builds this exec.

nilsvu:
Re memory allocations: there's a comment at the start of the function about that. I'll deal with it when I do a performance optimization pass over the elliptic code.

nilsvu:
Rebased because two CI tests failed in an unrelated way

nilsdeppe (approved):
Sounds good!

nilsvu:
CI failure looks like #5832

PR #5790, opened 2024-02-21
Wrap RotatingStar to be usable in GhValenciaDivClean

ncorsobh:
## Proposed changes

Adds the `RelativisticEuler` solution for `RotatingStar` to the list of `WrappedGR` solutions usable in `GhValenciaDivClean`. Adds time derivatives to the solution (all of which identically 0) to allow it to be compatible with this function.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
@@ -770,6 +770,15 @@ RotatingStar::variables(
 
 template <typename DataType>
 auto RotatingStar::variables(
+    const gsl::not_null<IntermediateVariables<DataType>*> /*vars*/,
+    const tnsr::I<DataType, 3>& x,
+    tmpl::list<::Tags::dt<gr::Tags::Lapse<DataType>>> /*meta*/) const
+    -> tuples::TaggedTuple<::Tags::dt<gr::Tags::Lapse<DataType>>> {
+  return make_with_value<Scalar<DataType>>(get<0>(x), 0.0);
+}
+
+template <typename DataType>
+auto RotatingStar::variables(
     const gsl::not_null<IntermediateVariables<DataType>*> vars,
     const tnsr::I<DataType, 3>& x,
     tmpl::list<DerivLapse<DataType>> /*meta*/) const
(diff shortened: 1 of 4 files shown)


src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -768,6 +768,16 @@ RotatingStar::variables(
                                 vars->metric_data->alpha, vars->phi);
 }
 
+template <typename DataType>
+auto RotatingStar::variables(
+    const gsl::not_null<IntermediateVariables<DataType>*> vars,
+    const tnsr::I<DataType, 3>& x,
+    tmpl::list<::Tags::dt<gr::Tags::Lapse<DataType>>> /*meta*/) const
+    -> tuples::TaggedTuple<::Tags::dt<gr::Tags::Lapse<DataType>>> {
+  interpolate_deriv_vars_if_necessary(vars);
```
nilsdeppe: I don't think you need this since the answer is identically zero, even without interpolation. Same for the other 2 you're adding.

nilsdeppe (changes requested):
Thanks for doing this! You can squash and rebase immediately

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
ncorsobh: So I made this change, but now it's failing to compile in the GitHub tests because `vars` is not used in the function. Should I replace this line with something else so that `vars` is still used?

nilsdeppe: You can just comment out the variable name in the function arguments: `/*vars*/`

nilsdeppe: It doesn't look like this was done?

ncorsobh: Sorry, I guess some weird versioning thing happened that reverted it to what it was before. I'll try again

nilsdeppe:
CI failure is unrelated to this PR

nilsdeppe (approved):

PR #5793, opened 2024-02-22
Compute scalar self-force in `UpdateAcceleration`

nikwit:
Changes the `UpdateAcceleration` mutator to compute the acceleration due to the scalar self-force.

~depends on #5787~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SelfForce.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SelfForce.cpp
@@ -1,0 +1,65 @@
+
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/SelfForce.hpp"
+
+#include <cstddef>
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/Gsl.hpp"
+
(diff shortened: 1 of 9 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/SelfForce.cpp
```diff
@@ -0,0 +1,64 @@
...
+    const tnsr::i<double, Dim>& psi_dipole,
+    const tnsr::I<double, Dim>& particle_velocity, const double particle_charge,
+    const double particle_mass, const tnsr::AA<double, Dim>& inverse_metric,
+    const Scalar<double>& dilation_factor) {
+  const double u0_squared = square(get(dilation_factor));
+  for (size_t i = 0; i < Dim; ++i) {
+    self_force_acc->get(i) =
+        (inverse_metric.get(i + 1, 0) -
+         particle_velocity.get(i) * inverse_metric.get(0, 0)) *
+        get(dt_psi_monopole) * particle_charge / particle_mass / u0_squared;
```
knelli2: [optional] Factor out the computation of `particle_charge / particle_mass / u0_squared` so it isn't recomputed. (useful also for the loop below)

knelli2 (dismissed):
Only one optional thing if you want

nikwit:
Done, thanks for the review!

knelli2 (approved):

PR #5795, opened 2024-02-22
Log Distribution Fix

AlexCarpenter46:
## Proposed changes

Changes the Interval in Log Distribution for frustums to be the even if the x-coord of the objects aren't the same. This was causing parts of the envelope to not line up properly for unequal mass runs (shown below) which caused bad constraints even at t=0. The second pic is with the fix.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

image
image

### DIFF
--- a/src/Domain/CoordinateMaps/Frustum.cpp
+++ b/src/Domain/CoordinateMaps/Frustum.cpp
@@ -130,6 +130,10 @@ Frustum::Frustum(const std::array<std::array<double, 2>, 4>& face_vertices,
     w_plus_ = w_delta + 1.0;
     w_minus_ = w_delta - 1.0;
   } else {
+    if (zeta_distribution_ == Distribution::Logarithmic) {
+      zeta_distribution_value_ = distribution_value.value_or(
+          -(radius_ + inner_radius_) / (radius_ - inner_radius_));
+    }
     w_plus_ = 2.0;
     w_minus_ = 0.0;
   }
@@ -150,12 +154,7 @@ std::array<tt::remove_cvref_wrap_t<T>, 3> Frustum::operator()(
     cap_zeta = zeta;
   } else if (zeta_distribution_ == Distribution::Logarithmic) {
     Interval log_mapping{
-        -1.0,
-        1.0,
-        -1.0,
-        1.0,
-        zeta_distribution_,
-        -(radius_ + inner_radius_) / (radius_ - inner_radius_)};
+        -1.0, 1.0, -1.0, 1.0, zeta_distribution_, zeta_distribution_value_};
     cap_zeta = log_mapping(std::array<T, 1>{{zeta}})[0];
   } else {
     ERROR(
@@ -238,12 +237,7 @@ std::optional<std::array<double, 3>> Frustum::inverse(
                         (w_plus_ - w_minus_ * logical_coords[2]);
   } else if (zeta_distribution_ == Distribution::Logarithmic) {
     Interval log_mapping{
-        -1.0,
-        1.0,
-        -1.0,
-        1.0,
-        zeta_distribution_,
-        -(radius_ + inner_radius_) / (radius_ - inner_radius_)};
+        -1.0, 1.0, -1.0, 1.0, zeta_distribution_, zeta_distribution_value_};
     logical_coords[2] =
         log_mapping.inverse(std::array<double, 1>{{logical_coords[2]}})
             .value()[0];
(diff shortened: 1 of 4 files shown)


src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -130,6 +130,10 @@ Frustum::Frustum(const std::array<std::array<double, 2>, 4>& face_vertices,
     w_plus_ = w_delta + 1.0;
     w_minus_ = w_delta - 1.0;
   } else {
+    zeta_distribution_value_ =
+        distribution_value.has_value()
+            ? distribution_value.value()
+            : -(radius_ + inner_radius_) / (radius_ - inner_radius_);
```
knelli2: This can be

```cpp
    zeta_distribution_value_ =
        distribution_value.value_or(
            -(radius_ + inner_radius_) / (radius_ - inner_radius_));
```

knelli2 (changes requested):
Just one thing

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -589,6 +578,9 @@ void Frustum::pup(PUP::er& p) {
     // data from an old version.
     inner_radius_ = std::numeric_limits<double>::signaling_NaN();
   }
+  if (version >= 3) {
+    p | zeta_distribution_value_;
```
nilsvu, line 582: Add an `else` clause which runs when unpacking data written by previous version of the code. It should set `zeta_distribution_value_` to the value use before this change.

src/Domain/CoordinateMaps/Frustum.hpp
```diff
@@ -335,6 +335,7 @@ class Frustum {
   bool with_equiangular_map_{false};
   bool is_identity_{false};
   Distribution zeta_distribution_ = Distribution::Linear;
+  double zeta_distribution_value_{std::numeric_limits<double>::signaling_NaN()};
```
nilsvu, line 338: also add to `==` operator (careful with the `NaN` value)

nilsvu (changes requested):

src/Domain/CoordinateMaps/Frustum.hpp
AlexCarpenter46: When adding to the `==` operator, it seems to break the test in `Test_Domain.cpp` and I'm not exactly sure why. I code --diff'd the 2 domains being compared and they seem to be the same (from the test output). Does the `SerializedDomain.h5` need to be updated?

nilsvu: I think this is because `inner_radius_` is unpacked to NaN from the data written by the old code. I suggest you set the `zeta_distribution_value_` only if the distribution is set to logarithmic, because only then do you actually need it. (both in the constructor and in pup)

nilsdeppe:
I just want to say nice job on finding and fixing this, @AlexCarpenter46 ! These sorts of things are not always easy to find!

knelli2:
No additional comments

src/Domain/CoordinateMaps/Frustum.cpp
```diff
@@ -612,10 +616,10 @@ bool operator==(const Frustum& lhs, const Frustum& rhs) {
          lhs.half_opening_angle_ == rhs.half_opening_angle_ and
          lhs.one_over_tan_half_opening_angle_ ==
          rhs.one_over_tan_half_opening_angle_ and
-         std::isnan(lhs.zeta_distribution_value_) ==
-             std::isnan(rhs.zeta_distribution_value_) and
-         ((std::isnan(lhs.zeta_distribution_value_) and
-           std::isnan(rhs.zeta_distribution_value_)) or
+         ((std::isnan(lhs.zeta_distribution_value_) or
```
nilsvu: why this change? I found the previous easier to understand

nilsvu (dismissed):
Pleas squash, and also run clang-format over the files please

src/Domain/CoordinateMaps/Frustum.cpp
AlexCarpenter46: Oh for some reason I thought the logic in the other one was causing the difference, so I tried this out. I can revert it back if you want.

AlexCarpenter46: Reverted it back

nilsvu (approved):
Nice 👍

knelli2 (approved):
Ignoring unrelated test failure

PR #5799, opened 2024-02-23
Start of class BinaryWithGravitationalWaves for XCTS solver

joaorebelo-megum:
## Proposed changes

Creation of the class BinaryWithGravitationalWaves for XCTS solver, that computes initial data for binary black hole with realistic wave content.
The conformal data is obtained from Post-Newtonian Theory.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

Some terms are still not computed (set to zero), they will be added in a later PR.


docs/References.bib
```diff
@@ -1217,6 +1217,20 @@ @Article{Hou2007
   url     = "https://doi.org/10.1007/s10915-006-9105-9"
 }
 
+@article{Jaranowski1997ky,
+    author = "Jaranowski, Piotr and Schaefer, Gerhard",
+    title = "{Third postNewtonian higher order ADM Hamilton dynamics for
+             two-body point mass systems}",
+    eprint = "gr-qc/9712075",
+    archivePrefix = "arXiv",
+    doi = "10.1103/PhysRevD.57.7274",
```
hrueter, line 1285: Add the `url` field in the form `https://doi.org/<doi>` for this entry and the other new entries.

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
```diff
@@ -0,0 +1,627 @@
...
+ * \tilde{\pi}_{(3)}^{ij})^{TT} \Bigr] + O(\epsilon^6). \f}
+ *
+ * where
+ *
+ * \f{equation}{
+ *  \tilde{\pi}_{(3)}^{i j}=\frac{1}{16 \pi} \sum_a p_{a}^k\{-\delta_{i
+ * j}(\frac{1}{r_a})_{, k}+2[\delta_{i k}(\frac{1}{r_a})_{, j}+\delta_{j
+ * k}(\frac{1}{r_a})_{, i}]-\frac{1}{2} r_{a, i j k}\}. \f}
+ */
+class BinaryWithGravitationalWaves
```
nilsvu, line 481: Nice! Since this is specific to binary black holes, what do you think about naming this `PostNewtonianBbh`?

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
```diff
@@ -0,0 +1,627 @@
...
+  void add_integral_term_to_radiative(
+      const gsl::not_null<tnsr::ii<DataType, Dim>*> radiative_term,
+      const gsl::not_null<Cache*> cache) const;
+};
+
+}  // namespace detail
+
+/*!
+ * \brief   Binary black hole initial data with realistic wave background,
+ * constructed in Post-Newtonian approximations.
```
nilsvu, line 354: Could you add a list/overview of assumptions that go into this approach, with some notes on limitations and possible plans to deal with them? Specifically: spins, coordinates of the isolated BH solutions, choice of time slicing.

Also, could you add some notes on the goals of this approach, e.g. reduce junk radiation?

nilsvu, line 354: Question: have you already thought about black hole boundary conditions? Do the standard apparent horizon boundary conditions work with this?

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
```diff
@@ -0,0 +1,627 @@
...
+    static constexpr Options::String help =
+        "The coordinates on the x-axis of the right black hole.";
+    using type = double;
+  };
+  struct YMomentumLeft {
+    static constexpr Options::String help =
+        "The y-axis-componet of the linear momentum of the left black hole.";
+    using type = double;
+  };
+  struct YMomentumRight {
```
nilsvu: Is it possible to specify an orbital angular velocity instead of these momenta?

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
hrueter: The formalism is quite general and would also apply to other objects (neutron stars, ...), so it could be generalised like it is done for the `Binary` class.

Furthermore one could use other approximations than Post-Newtonian for the past evolution. Therefore we should not put "Post-Newtonian" in the name.

hrueter (changes requested):

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
joaorebelo-megum: The time slicing is comming from the trace K which we choose to be the same as in Post-Newtonian ADMTT gauge.
I'll also add the others to the documentation.

joaorebelo-megum: Right now we see a very slow convergence, but it converges. We'll have to look into that when we finish all the computations needed for the conformal data.
Still I'll add a warning in the documentation.

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
joaorebelo-megum: We are going to start testing with circular orbits, so this option will desapear. Later, we can then add orbital angular velocity as an option for other orbits in the past evolution. (later merge request)

joaorebelo-megum:
@nilsvu The clang-tidy and unit tests errors should be fixed now.
Should I squash the fixups?

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
```diff
@@ -0,0 +1,628 @@
...
+  const double mass_right;
+  const double xcoord_left;
+  const double xcoord_right;
+  const double ymomentum_left;
+  const double ymomentum_right;
+  const double attenuation_parameter;
+  const double separation = xcoord_right - xcoord_left;
+  const std::array<double, 3> normal_lr{{-1., 0., 0.}};
+  tnsr::I<DataType, 3> momentum_left = x;
+  tnsr::I<DataType, 3> momentum_right = x;
```
nilsvu: Should these be `tnsr::I<double, 3>`? Then you also don't have to initialize them by copying `x`.

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
```diff
@@ -0,0 +1,756 @@
...
+void BinaryWithGravitationalWavesVariables<DataType>::operator()(
+    const gsl::not_null<tnsr::ii<DataType, 3>*> radiative_term,
+    const gsl::not_null<Cache*> cache,
+    detail::Tags::RadiativeTerm<DataType> /*meta*/) const {
+  get<0, 0>(*radiative_term) = 0.;
+  get<0, 1>(*radiative_term) = 0.;
+  get<0, 2>(*radiative_term) = 0.;
+  get<1, 1>(*radiative_term) = 0.;
+  get<1, 2>(*radiative_term) = 0.;
+  get<2, 2>(*radiative_term) = 0.;
```
nilsvu: you can do: `std::fill(radiative_term->begin(), radiative_term->end(), 0.)`. Also in some more cases below.

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
```diff
@@ -0,0 +1,756 @@
...
+    detail::Tags::NearZoneTerm<DataType> /*meta*/) const {
+  const auto& distance_left =
+      get(cache->get_var(*this, detail::Tags::DistanceLeft<DataType>{}));
+  const auto& distance_right =
+      get(cache->get_var(*this, detail::Tags::DistanceRight<DataType>{}));
+  const auto& normal_left =
+      cache->get_var(*this, detail::Tags::NormalLeft<DataType>{});
+  const auto& normal_right =
+      cache->get_var(*this, detail::Tags::NormalRight<DataType>{});
+  const auto s = distance_left + distance_right + separation;
```
nilsvu: Best to write `const DataType s = ` here, to make sure the addition is actually carried out. Otherwise, the result might be just the unevaluated expression, which can have unexpected effects. This is because we use the [Blaze](https://bitbucket.org/blaze-lib/blaze/src/master/) library for linear algebra.

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
```diff
@@ -0,0 +1,756 @@
...
+  const auto& distance_left =
+      get(cache->get_var(*this, detail::Tags::DistanceLeft<DataType>{}));
+  const auto& normal_left =
+      cache->get_var(*this, detail::Tags::NormalLeft<DataType>{});
+  std::array<std::array<double, 3>, 3> delta{
+      {{{1., 0., 0.}}, {{0., 1., 0.}}, {{0., 0., 1.}}}};
+  for (size_t i = 0; i < 3; ++i) {
+    for (size_t j = 0; j < 3; ++j) {
+      for (size_t k = 0; k < 3; ++k) {
+        deriv_3_distance_left->get(i, j, k) =
```
nilsvu, line 97: (optional) You could make some of these expressions easier to implement using our tensor expressions framework. See `src/Evolution/Systems/CurvedScalarWave/Worldtube/KerrSchildDerivatives.cpp` for examples. I don't think a Kronecker delta is supported yet though, right @macedo22?

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
```diff
@@ -0,0 +1,756 @@
...
+template <typename DataType>
+void BinaryWithGravitationalWavesVariables<DataType>::operator()(
+    const gsl::not_null<tnsr::ii<DataType, 3>*> conformal_metric,
+    const gsl::not_null<Cache*> cache,
+    Xcts::Tags::ConformalMetric<DataType, 3, Frame::Inertial> /*meta*/) const {
+  const auto& distance_left =
+      get(cache->get_var(*this, detail::Tags::DistanceLeft<DataType>{}));
+  const auto& distance_right =
+      get(cache->get_var(*this, detail::Tags::DistanceRight<DataType>{}));
+  const auto E_left =
```
nilsvu: These are `double`s, right? Best make that explicit.

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
```diff
@@ -0,0 +1,756 @@
...
+  add_radiative_term_PN_of_conformal_metric(conformal_metric, cache);
+}
+
+template <typename DataType>
+void BinaryWithGravitationalWavesVariables<DataType>::operator()(
+    const gsl::not_null<tnsr::ijj<DataType, 3>*> deriv_conformal_metric,
+    const gsl::not_null<Cache*> /*cache*/,
+    ::Tags::deriv<Xcts::Tags::ConformalMetric<DataType, 3, Frame::Inertial>,
+                  tmpl::size_t<Dim>, Frame::Inertial> /*meta*/) const {
+  std::fill(deriv_conformal_metric->begin(), deriv_conformal_metric->end(), 0.);
```
nilsvu: Is this missing? I don't think the derivative of the conformal metric is zero, or is it?

tests/InputFiles/Xcts/BinaryWithGravitationalWaves.yaml
```diff
@@ -0,0 +1,226 @@
...
+            Rotation: [0., 0., 0.]
+            Lapse: &bh_right
+              Schwarzschild:
+                Mass: *mass_right
+                Coordinates: Isotropic
+            NegativeExpansion: *bh_right
+      UseLogarithmicMap: True
+    ObjectB:
+      InnerRadius: 0.4
+      OuterRadius: 0.5
```
nilsvu: This is very small. Is this intentional?

tests/InputFiles/Xcts/BinaryWithGravitationalWaves.yaml
```diff
@@ -0,0 +1,226 @@
...
+                Coordinates: Isotropic
+            NegativeExpansion: *bh_left
+      UseLogarithmicMap: True
+    Envelope:
+      Radius: &outer_shell_inner_radius 14.
+      RadialDistribution: Linear
+    OuterShell:
+      Radius: &outer_radius 20.
+      RadialDistribution: &outer_shell_distribution Linear
+      OpeningAngle: 90.0
```
nilsvu: Probably better to set this to `120`.

tests/InputFiles/Xcts/BinaryWithGravitationalWaves.yaml
```diff
@@ -0,0 +1,226 @@
...
+      BoundaryCondition: Flatness
+    UseEquiangularMap: True
+    # This h-refinement is set up so spherical wedges have equal angular size.
```
nilsvu, line 75: If you set `OpeningAngle: 120` you can remove all this and use `InitialRefinement: 0`, or maybe some more manual radial refinement based on your own experimentation.

tests/InputFiles/Xcts/BinaryWithGravitationalWaves.yaml
```diff
@@ -0,0 +1,226 @@
...
+      OuterShellUpperYLeft:       [0, 1, 0]
+      OuterShellLowerYLeft:       [0, 1, 0]
+      OuterShellLowerX:           [1, 1, 0]
+      OuterShellUpperZRight:      [0, 1, 0]
+      OuterShellLowerZRight:      [0, 1, 0]
+      OuterShellUpperYRight:      [0, 1, 0]
+      OuterShellLowerYRight:      [0, 1, 0]
+      OuterShellUpperX:           [1, 1, 0]
+    # This p-refinement represents a crude manual optimization of the domain. We
+    # will need AMR to optimize the domain further.
```
nilsvu, line 104: Also here you probably have to do your own experimentation.

tests/InputFiles/Xcts/BinaryWithGravitationalWaves.yaml
```diff
@@ -0,0 +1,226 @@
...
+      ObjectAShell:   [6, 6, 10]
+      ObjectBShell:   [6, 6, 10]
+      ObjectACube:    [6, 6, 7]
+      ObjectBCube:    [6, 6, 7]
+      Envelope:       [6, 6, 6]
+      OuterShell:     [6, 6, 6]
+
+Discretization:
+  DiscontinuousGalerkin:
+    PenaltyParameter: 1.
```
nilsvu: I recommend setting this to 1.5 to be a bit safer. And/or check for a few values between 1 and 10 to find the value where the linear solver converges in the fewest iterations.

tests/InputFiles/Xcts/BinaryWithGravitationalWaves.yaml
```diff
@@ -0,0 +1,226 @@
...
+  ReductionFileName: "BwGWReductions"
+
+NonlinearSolver:
+  NewtonRaphson:
+    ConvergenceCriteria:
+      MaxIterations: 20
+      RelativeResidual: 1.e-10
+      AbsoluteResidual: 1.e-11
+    SufficientDecrease: 1.e-4
+    MaxGlobalizationSteps: 40
```
nilsvu: You can reduce this to something like 10. For anything higher the solver probably won't ever converge anyway.

tests/Unit/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.py
```diff
@@ -0,0 +1,464 @@
...
+# See LICENSE.txt for details.
+
+import numpy as np
+
+xcoords = [-4.5, 10.2]
+mass_left = 1.1
+mass_right = 0.9
+atenuation_parameter = 0.99
+outer_radius = 21.0
+write_evolution_option = False
```
nilsvu: What is this / is it used?

nilsvu (commented):
Very nice 👍 Couple of suggestions. Yes please squash the existing commits and push new ones if you make any further changes.

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
joaorebelo-megum: It is like that so that I can use the dot_product() in my expressions. Using  double it doesn't compile.

nilsvu: Ok I suggest you use `tnsr::I<double, 3>` and do the dot product manually then, because otherwise you're creating a tensor over the entire grid filled with a constant number. That seems a bit too wasteful :)

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
macedo22: Correct, Kronecker delta is not yet supported

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.hpp
joaorebelo-megum: I decided to go with a std::array<double, 3> (and make a dot_product function). But on the final stage of the code this variable will no longer be used.

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
joaorebelo-megum: Done

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
joaorebelo-megum: Indeed I had some trouble with this not long ago, thank you! It is now changed

src/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.cpp
joaorebelo-megum: Yes, it was missing. It is implemented now.

tests/InputFiles/Xcts/BinaryWithGravitationalWaves.yaml
joaorebelo-megum: Yes, the horizon in isotropic coordinates is at 0.5 M.

tests/InputFiles/Xcts/BinaryWithGravitationalWaves.yaml
joaorebelo-megum: Done, thank you. Still I didn't test if the solver works, I'm using this to output and see the Background Data at the inicial iteration. I will start working on the solving after I have every term of the metric computed + time derivatives.

tests/Unit/PointwiseFunctions/AnalyticData/Xcts/BinaryWithGravitationalWaves.py
joaorebelo-megum: No, it is not, thank you for noticing.

joaorebelo-megum:
@nilsvu I think everything is taken care of. If you happen to spot anything else, please tell me!

PR #5805, opened 2024-02-25
Add PlotSlice command

nilsvu:
## Proposed changes

Allows to plot a quick slice through volume data (without ParaView).

Example from the unit test:

plot

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/Exporter/Exporter.cpp
+++ b/src/IO/Exporter/Exporter.cpp
@@ -122,13 +122,37 @@ void interpolate_to_points(
     }
   }
 }
+
+// Determines the selected observation ID in the volume data file, given either
+// an `ObservationId` directly or an `ObservationStep`.
+struct SelectObservation {
+  size_t operator()(const ObservationId observation_id) const {
+    return observation_id.value;
+  }
+  size_t operator()(const ObservationStep observation_step) const {
+    int step = observation_step.value;
+    const auto obs_ids = volfile.list_observation_ids();
+    if (step < 0) {
+      step += static_cast<int>(obs_ids.size());
+    }
+    if (step < 0 or static_cast<size_t>(step) >= obs_ids.size()) {
+      ERROR_NO_TRACE("Invalid observation step: "
+                     << observation_step.value << ". There are "
+                     << obs_ids.size() << " observations in the file.");
+    }
+    return obs_ids[static_cast<size_t>(step)];
+  }
+  const h5::VolumeData& volfile;
+};
+
 }  // namespace
 
 template <size_t Dim>
 std::vector<std::vector<double>> interpolate_to_points(
     const std::variant<std::vector<std::string>, std::string>&
         volume_files_or_glob,
-    std::string subfile_name, int observation_step,
+    std::string subfile_name,
+    const std::variant<ObservationId, ObservationStep>& observation,
     const std::vector<std::string>& tensor_components,
     const std::array<std::vector<double>, Dim>& target_points,
     const std::optional<size_t> num_threads) {
@@ -166,21 +190,13 @@ std::vector<std::vector<double>> interpolate_to_points(
     ERROR_NO_TRACE("Mismatched dimensions: expected "
                    << Dim << "D volume data, but got " << dim << "D.");
   }
-  // Get observation ID. This currently assumes that all volume files contain
-  // the same observations. For generalizing to volume files across multiple
-  // segments, see the Python function `Visualization.ReadH5:select_observation`
-  // and possibly move it to C++.
-  const auto obs_ids = first_volfile.list_observation_ids();
-  if (observation_step < 0) {
-    observation_step += static_cast<int>(obs_ids.size());
-  }
-  if (observation_step < 0 or
-      static_cast<size_t>(observation_step) >= obs_ids.size()) {
-    ERROR_NO_TRACE("Invalid observation step: "
-                   << observation_step << ". There are " << obs_ids.size()
-                   << " observations in the file.");
-  }
-  const size_t obs_id = obs_ids[static_cast<size_t>(observation_step)];
+  // Get observation ID
+  // This currently assumes that all volume files contain the same observations,
+  // so we only look into the first file. For generalizing to volume files
+  // across multiple segments, see the Python function
+  // `Visualization.ReadH5:select_observation` and possibly move it to C++.
+  const size_t obs_id =
+      std::visit(SelectObservation{first_volfile}, observation);
   // Get domain, time, functions of time
   const auto domain =
       deserialize<Domain<Dim>>(first_volfile.get_domain(obs_id)->data());
@@ -262,7 +278,8 @@ std::vector<std::vector<double>> interpolate_to_points(
   template std::vector<std::vector<double>> interpolate_to_points<DIM(data)>( \
       const std::variant<std::vector<std::string>, std::string>&              \
           volume_files_or_glob,                                               \
-      std::string subfile_name, int observation_step,                         \
+      std::string subfile_name,                                               \
+      const std::variant<ObservationId, ObservationStep>& observation,        \
       const std::vector<std::string>& tensor_components,                      \
       const std::array<std::vector<double>, DIM(data)>& target_points,        \
       const std::optional<size_t> num_threads);
--- a/src/IO/Exporter/Python/InterpolateToPoints.py
+++ b/src/IO/Exporter/Python/InterpolateToPoints.py
@@ -9,51 +9,16 @@ import rich
 
 import spectre.IO.H5 as spectre_h5
 from spectre.IO.Exporter import interpolate_to_points
+from spectre.Visualization.OpenVolfiles import (
+    open_volfiles_command,
+    parse_points,
+)
 
 logger = logging.getLogger(__name__)
 
 
-def parse_points(ctx, param, values):
-    if not values:
-        return None
-    points = [list(map(float, value.split(","))) for value in values]
-    dim = len(points[0])
-    if any([len(point) != dim for point in points]):
-        raise click.BadParameter(
-            "All specified points must have the same dimension"
-        )
-    return np.array(points)
-
-
 @click.command(name="interpolate-to-points")
-@click.argument(
-    "h5_files",
-    nargs=-1,
-    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
-)
-@click.option(
-    "--subfile-name",
-    "-d",
-    help=(
-        "Name of subfile within h5 file containing volume data to interpolate."
-    ),
-)
-@click.option(
-    "--list-vars",
-    "-l",
-    is_flag=True,
-    help="Print available variables and exit.",
-)
-@click.option(
-    "--var",
-    "-y",
-    "vars",
-    multiple=True,
-    help=(
-        "Variables to interpolate. List any tensor components "
-        "in the volume data file, such as 'Shift_x'."
-    ),
-)
+@open_volfiles_command(obs_id_required=True, multiple_vars=True)
 @click.option(
     "--target-coords-file",
     "-t",
@@ -74,19 +39,6 @@ def parse_points(ctx, param, values):
         "List target coordinates explicitly, e.g. '0,0,0'. "
         "Can be specified multiple times to quickly interpolate "
         "to a couple of target points."
-    ),
-)
-# Note: support for files with different observations can be added if needed by
-# porting `Visualization.ReadH5:select_observation` to C++. That function would
-# also support a `time` option, as an alternative to `step`.
-@click.option(
-    "--step",
-    type=int,
-    help=(
-        "Observation step number. Specify '-1' for the last step in the file. "
-        "All files must contain the same set of observations. Support for "
-        "files with different observations (e.g. from multiple segments of a "
-        "simulation) can be added if needed."
     ),
 )
 @click.option(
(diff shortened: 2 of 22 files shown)


nilsvu:
@knelli2 could you review this one?

src/IO/Exporter/Exporter.cpp
```diff
@@ -165,21 +165,29 @@ std::vector<std::vector<double>> interpolate_to_points(
...
-  }
-  if (observation_step < 0 or
-      static_cast<size_t>(observation_step) >= obs_ids.size()) {
-    ERROR_NO_TRACE("Invalid observation step: "
-                   << observation_step << ". There are " << obs_ids.size()
-                   << " observations in the file.");
-  }
-  const size_t obs_id = obs_ids[static_cast<size_t>(observation_step)];
+  // Get observation ID
+  const size_t obs_id = std::visit(
```
knelli2: This feels like it should have a test. Is that easy to add?

src/IO/Exporter/Python/InterpolateToPoints.py
```diff
@@ -26,34 +27,7 @@ def parse_points(ctx, param, values):
...
-    multiple=True,
-    help=(
-        "Variables to interpolate. List any tensor components "
-        "in the volume data file, such as 'Shift_x'."
-    ),
-)
+@open_volfiles_command
 @click.option(
     "--target-coords-file",
     "-t",
```
knelli2, line 24: Can't reach it with Github comments, but replace the `parse_points` below in `target-coords` with the new one you wrote?

src/Visualization/Python/PlotSlice.py
```diff
@@ -0,0 +1,250 @@
...
+    """Returns points on a rectangular slice in 3D.
+
+    Parameters:
+      slice_origin: Coordinates of the center of the slice.
+      slice_radius: Visible radius around the center of the slice.
+      slice_normal: Direction of the normal of the slice.
+      slice_up: Up-direction of the slice.
+      num_points: Number of points to return in each dimension.
+
+    Returns: An array of shape (3, num_points, num_points) with uniformly
```
knelli2: Why an extra dimension of length 3? I thought this was a 2D slice?

src/Visualization/Python/PlotSlice.py
```diff
@@ -0,0 +1,250 @@
...
+    "--slice-origin",
+    "-C",
+    callback=parse_point,
+    help=(
+        "Coordinates of the center of the slice through the volume "
+        "data. Specify as comma-separated list, e.g. '0,0,0'."
+    ),
+)
+@click.option(
+    "--slice-radius",
```
knelli2: Calling this `slice-radius` when it's a rectangle is a bit confusing. Maybe `slice-extents`? Or `slice-range`? Something like that.

src/Visualization/Python/PlotSlice.py
```diff
@@ -0,0 +1,250 @@
...
+    stylesheet,
+):
+    """Plot variables on a slice through volume data
+
+    Interpolates the volume data in the H5_FILES to a slice and plots the
+    selected variables. You choose the slice by specifying three corners of a
+    rectangle.
+    """
+    if obs_id is None:
+        raise click.UsageError("Specify '--step' / '-s' or '--time' / '-t'.")
```
knelli2: Even though this is short, this exact code-block is used in at least one other file I think? Maybe worth factoring out in case the obs_id becomes more complicated later on?

src/Visualization/Python/PlotSlice.py
```diff
@@ -0,0 +1,250 @@
...
+        raise click.UsageError("Only one variable can be plotted at a time.")
+    var_name = vars[0]
+
+    # Interpolate the selected variable to the slice
+    if (
+        slice_origin is None
+        or slice_radius is None
+        or slice_normal is None
+        or slice_up is None
+    ):
```
knelli2: Make them click-required options?

src/Visualization/Python/PlotSlice.py
```diff
@@ -0,0 +1,250 @@
...
+    contours = plt.contour(
+        x, y, data, colors="white", linewidths=0.5, alpha=0.6
+    )
+    plt.clabel(contours)
+    plt.gca().set_aspect("equal")
+    plt.xlabel(x_label)
+    plt.ylabel(y_label)
+
+    if output:
+        plt.savefig(output, format="pdf", bbox_inches="tight")
```
knelli2: Check `output` for the `.pdf` extension like the other CLI plotting tools. Better yet, maybe another thing to factor out

knelli2 (changes requested):

src/Visualization/Python/PlotSlice.py
nilsvu: These are the 3D coordinates of the points on the slice.

src/Visualization/Python/PlotSlice.py
nilsvu: Then you'd have to specify them even if you don't know the subfile name or the variable name yet and want the available names printed out.

src/Visualization/Python/PlotSlice.py
nilsvu: Ok I tried something else, see fixup.

src/Visualization/Python/PlotSlice.py
```diff
@@ -23,41 +23,46 @@
...
-      slice_normal: Direction of the normal of the slice.
-      slice_up: Up-direction of the slice.
-      num_points: Number of points to return in each dimension.
+      slice_origin: Coordinates of the center of the slice. Mutually exclusive
+        with slice_center.
+      slice_center: Coordinates of the center of the slice. Mutually exclusive
+        with slice_origin.
+      slice_x: Horizontal axis of the slice. Must be orthogonal to slice_y.
+      slice_y: Vertical axis of the slice. Must be orthogonal to slice_x.
+      num_points: Number of points to return in each dimension (tuple or int).
```
knelli2: I actually think this is more confusing now than it was before. Even though it's explained in the CLI docs, I'm going to get confused every time what the difference between `slice_origin` and `slice_center` are. And I'm still a bit confused on what `slice_{x,y}` are. Also if you want to do anything besides a (1,0,0)/(0,1,0) slice, then it's going to be hard to ensure x/y are orthogonal.

I think the `slice_origin`, `slice_normal`, and `slice_up` are easier to understand and easier for users to input (in the case of a rotated xy-plane slice, you keep the normal the same and just change the "up" direction. With the new method, you have to calculate the vectors yourself before you run the CLI). What about replacing the name `slice_radius` with `slice_width` but keeping the functionality the same?

src/Visualization/Python/PlotSlice.py
knelli2: Ping

knelli2 (changes requested):

src/Visualization/Python/PlotSlice.py
nilsvu: ok, pushed a new fixup

src/Visualization/Python/PlotSlice.py
nilsvu: Pushed a fixup that factors these commands out. Helps a lot making them consistent with each other. Note that the format and extension don't need to be specified. If no extension is specified, matplotlib defaults to png. If an extension is specified, matplotlib will detect that.
Note also that the fixup removes the automatic object label suffix from the memory monitor output file. I think it's generally better to write to the exact file that the user has specified, not modify the name unexpectedly. This makes the output easier to find and, more importantly, avoids accidentally overwriting existing files. Also simplifies the code.

src/Visualization/Python/OpenVolfiles.py
```diff
@@ -61,130 +74,151 @@ def open_volfiles_command(f):
...
+        )
+        @click.option(
+            "--var",
+            "-y",
+            "vars_patterns",
+            multiple=multiple_vars,
+            help=(
+                "Variable to plot. List any tensor components "
+                "in the volume data file, such as 'Shift_x'. "
+                "Also accepts glob patterns like 'Shift_*'."
```
knelli2, line 113: Will the glob work if you don't allow multiple vars?

src/Visualization/Python/Plot.py
```diff
@@ -0,0 +1,87 @@
...
+
+    def decorator(f):
+        @click.option(
+            "--output",
+            "-o",
+            type=click.Path(file_okay=True, dir_okay=False, writable=True),
+            help=(
+                "Name of the output plot file. If unspecified, the plot is "
+                "shown interactively, which only works on machines with a "
+                "window server."
```
knelli2: Maybe add, "If no extension is specified, a png will be saved"

src/Visualization/Python/Plot.py
```diff
@@ -0,0 +1,87 @@
...
+            ),
+        )
+        # Preserve the original function's name and docstring
+        @functools.wraps(f)
+        def command(output, **kwargs):
+            # Call the original function
+            f(**kwargs)
+            # Show or save the plot
+            if output:
+                plt.savefig(output)
```
knelli2, line 78: Maybe keep the `bbox_inches="tight"`? I think some plots that have the legend outside the plot won't render properly without it

src/Visualization/Python/PlotSlice.py
```diff
@@ -83,11 +82,13 @@ def points_on_slice(
     ),
 )
 @click.option(
-    "--slice-radius",
-    "-R",
+    "--slice-extent",
+    "-X",
```
knelli2, line 86: Oh yeah, this looks good.

knelli2 (changes requested):
Ok you can squash everything down into a few commits (while fixing the test errors)

src/Visualization/Python/PlotSlice.py
knelli2: I think you mean the `PlotSizeControl` test? But ok yeah that makes sense, just be sure to edit the test

src/Visualization/Python/OpenVolfiles.py
nilsvu: The glob works, but you'll get an error if it matches more than 1 variable. That's useful behavior I think (don't have to type the full name).

src/Visualization/Python/Plot.py
nilsvu: It's part of the default stylesheet (plots.mplstyle in this directory).

nilsvu:
Done. You should actually try out the `-s` option for the stylesheet, it's pretty fun :D Just do something like `-s seaborn-dark` or `-s classic`. See https://matplotlib.org/stable/gallery/style_sheets/style_sheets_reference.html for all built-in styles.

classic:

plot

src/IO/Exporter/Exporter.cpp
knelli2: Would this be easy to add? Also, do we need to worry about automatic conversion between `size_t` and `int`? Currently I think it's only tested for `int`? Actually I'm not sure because of the conversion thing.

src/Visualization/Python/PlotAlongLine.py
```diff
@@ -90,30 +83,18 @@ def points_on_line(
...
-    envvar="SPECTRE_MPL_STYLESHEET",
-    help=(
-        "Select a matplotlib stylesheet for customization of the plot, such "
-        "as linestyle cycles, linewidth, fontsize, legend, etc. "
-        "The stylesheet can also be set with the 'SPECTRE_MPL_STYLESHEET' "
-        "environment variable."
-    ),
-)
+@apply_stylesheet_command()
+@show_or_save_plot_command()
```
knelli2, line 87: Are the order of these two important? If so, we should not that somewhere

src/Visualization/Python/PlotSlice.py
```diff
@@ -0,0 +1,220 @@
...
+    if y_axis is not None:
+        y = target_coords[y_axis]
+        y_label = "xyz"[y_axis]
+    else:
+        y = np.linspace(0, 1, num_samples[1])
+        y_label = None
+
+    # Plot the slice
+    plt.contourf(x, y, data)
+    contours = plt.contour(
```
knelli2, line 209: Probably not for this PR, but we may want to offer other types of plots besides contours.

knelli2 (commented):
Last few comments/questions. Feel free to ignore them

src/Visualization/Python/PlotAlongLine.py
nilsvu: No, not important

src/Visualization/Python/PlotSlice.py
nilsvu: If someone wants to do anything more than look at a quick slice through the data, then they can do that themselves in Python. We can't build all the possible things into the CLI, it's just to get a quick idea of what's going on with a simulation.

src/IO/Exporter/Exporter.cpp
nilsvu: Pushed a fixup that adds names to the call site of the function

nilsvu:
@knelli2 please take a look at the new fixups

src/Visualization/Python/PlotAlongLine.py
```diff
@@ -103,12 +103,11 @@ def plot_along_line_command(
     points.
     """
     # Interpolate the selected quantities to the line
-    dim = len(line_start)
     if line_start is None or line_end is None:
         raise click.UsageError(
             f"Specify '--line-start' / '-A' and '--line-end' / '-B'."
         )
-    if len(line_start) != dim or len(line_end) != dim:
```
knelli2, line 206: Is there a way to actually check that the start/end point has the correct number of components?

knelli2 (commented):
Don't know if you can fix the clang-tidy error or if we should ignore it. You can squash everything in

src/Visualization/Python/PlotAlongLine.py
nilsvu: Not without opening one of the h5 files again. You'll get a "mismatched dimensions" error from the interpolation function, which is not as nice as catching this earlier (here) but enough for now I think.

nilsvu:
Rebased. Thanks for reviewing! No we can't fix the clang-tidy issue.

knelli2 (approved):

knelli2:
Ignoring clang-tidy errors

PR #5807, opened 2024-02-25
Add CLI to plot memory monitors

knelli2:
## Proposed changes

Examples of output:

[no_color.pdf](https://github.com/sxs-collaboration/spectre/files/14397432/no_color.pdf)

[with_color.pdf](https://github.com/sxs-collaboration/spectre/files/14397430/with_color.pdf)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/PlotMemoryMonitors.py
+++ b/src/Visualization/Python/PlotMemoryMonitors.py
@@ -1,0 +1,232 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+import logging
+import os
+from math import inf
+from typing import Optional, Sequence
+
+import click
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+from spectre.Visualization.ReadH5 import available_subfiles, to_dataframe
+
+logger = logging.getLogger(__name__)
+
+
+@click.command(name="memory-monitors")
+@click.argument(
+    "reduction_files",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    nargs=-1,
+)
+@click.option(
+    "--output",
+    "-o",
+    type=click.Path(file_okay=True, dir_okay=False, writable=True),
+    help=(
+        "Name of the output plot file. If unspecified, the plot is "
+        "shown interactively, which only works on machines with a "
+        "window server."
+    ),
+)
+# Plotting options
+@click.option(
+    "--use-mb/--use-gb",
+    default=False,
+    show_default=True,
+    help="Plot the y-axis in Megabytes or Gigabytes",
+)
+@click.option(
+    "--x-label",
+    help="The label on the x-axis.",
+    show_default="name of the x-axis column",
+)
+@click.option(
+    "--x-bounds",
+    type=float,
+    nargs=2,
+    help="The lower and upper bounds of the x-axis.",
+)
+@click.option(
+    "--stylesheet",
+    "-s",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    envvar="SPECTRE_MPL_STYLESHEET",
+    help=(
+        "Select a matplotlib stylesheet for customization of the plot, such "
+        "as linestyle cycles, linewidth, fontsize, legend, etc. "
+        "The stylesheet can also be set with the 'SPECTRE_MPL_STYLESHEET' "
+        "environment variable."
+    ),
+)
+def plot_memory_monitors_command(
+    reduction_files: Sequence[str],
+    output: Optional[str],
+    use_mb: bool,
+    x_label: Optional[str],
+    x_bounds: Optional[Sequence[float]],
+    stylesheet,
+):
+    """
+    Plot the memory usage of a simulation from the MemoryMonitors data in the
+    Reductions H5 file.
+
+    This tool assumes there is a group in each of the "reduction-files" with the
+    path "/MemoryMonitors/" that holds dat files for each parallel component
+    that was monitored.
+
+    Note that the totals plotted here are not necessary the total memory usage
+    of the simulation. The memory monitors only capture what is inside
+    'pup'functions. Any memory that cannot be captured by a 'pup' function will
+    not by represented by this plot.
+    """
+
+    # Given an h5 file, make sure that the "MemoryMonitors" group exists. Then
+    # return the list of subfiles in that group
+    def check_memory_monitor_dir(h5_filename: str):
+        h5file = h5py.File(h5_filename, "r")
+        memory_monitor_dir = h5file.get("MemoryMonitors")
+        if memory_monitor_dir is None:
+            raise click.UsageError(
+                "Unable to open group 'MemoryMonitors' from h5 file"
+                f" {h5_filename}. Available subfiles are:\n"
+                f" {available_subfiles(h5file, extension='.dat')}"
+            )
+
+        return list(memory_monitor_dir.keys())
+
+    # Given an h5 file and a subfile, make sure that the subfile exists inside
+    # the h5 file. Then return the subfile as a DataFrame (with "Time" as the
+    # index)
+    def check_memory_monitor_file(
+        h5_filename: str, h5file: h5py.File, subfile_name: str
+    ):
+        subfile_path = f"/MemoryMonitors/{subfile_name}"
+        subfile = h5file.get(subfile_path)
+        if subfile_path is None:
+            raise click.UsageError(
+                f"Unable to open memory file '{subfile_path}'"
+                f" from h5 file {h5_filename}. Available subfiles are:\n"
+                f" {available_subfiles(h5file, extension='.dat')}"
+            )
+
+        return to_dataframe(subfile).set_index("Time")
+
+    # Given a DataFrame, sum all the columns that list totals for each node
+    def total_over_nodes(df: pd.DataFrame):
+        cols_to_sum = [
+            col
+            for col in df.columns
+            if col == "Size (MB)" or "Size on node" in col
+        ]
+        return df[cols_to_sum].sum(axis=1)
+
+    # Get a list of all components that we have monitored from the first
+    # reductions file
+    memory_filenames = check_memory_monitor_dir(reduction_files[0])
+
+    # Open every h5file. For each h5file, turn each subfile into a DataFrame.
+    # Then concat all DataFrames together into one that's indexed by the
+    # subfile/component name.
+    totals_df = pd.DataFrame()
+    for reduction_file in reduction_files:
+        h5file = h5py.File(reduction_file)
+        local_totals_df = pd.DataFrame()
+
+        for subfile_name in memory_filenames:
+            df = check_memory_monitor_file(reduction_file, h5file, subfile_name)
+            local_totals_df[subfile_name] = total_over_nodes(df)
+
+        totals_df = pd.concat([totals_df, local_totals_df])
+
+    # Restrict to x-bounds if there are any
+    for subfile in memory_filenames:
+        if x_bounds is not None:
+            totals_df[subfile] = totals_df[subfile][
+                (totals_df.index >= x_bounds[0])
+                & (totals_df.index <= x_bounds[1])
+            ]
+
+    # Need .dat because all other components have that extension
+    the_rest_str = "The Rest.dat"
+
+    # Sum all the columns that list totals for each node to get a grand total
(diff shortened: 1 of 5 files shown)


tests/Unit/Visualization/Python/CMakeLists.txt
```diff
@@ -44,6 +50,11 @@ if(${BUILD_PYTHON_BINDINGS})
     PROPERTIES
     TIMEOUT 10
     )
+  set_tests_properties(
+    "Unit.Visualization.Python.PlotMemoryMonitors"
+    PROPERTIES
+    TIMEOUT 40
```
knelli2: For some reason, the command itself takes a very long time to run. I wasn't able to determine why, even with `spectre --profile`. Any suggestions of where I'm going wrong are welcome.

nilsdeppe:
I'll let @nilsvu review the actual code, but I have a feature request: would it be easy to add a flag like `--top-n 5` to show only the 5 most significant users? On the plots you showed it was tricky to figure out from the legend which were the 3 big ones shown.

nilsvu:
Good idea, and/or how about a stacked area plot like this: https://python-graph-gallery.com/stacked-area-plot/
Then only show the legend entry for the big ones (above 5% or something like that)

nilsdeppe:
I'd be fine with adding that as an option. I do find it easier in the current style to read off how many GB something is using. I think the stacked plot makes relative comparisons nice, though. So maybe a CLI flag to enable the stacking? E.g. `--stacked-plot`?

knelli2:
I'm personally not a big fan of the stacked plot. It's a bit ambiguous to me whether to interpret a stack as relative to zero, or relative to the stack below it. I do like the `--top-n` option though. I'll add that.

nilsvu:
True. Maybe just combine the entries with only a couple percent into a single entry, then you don't need an extra CLI option and you can't see those small entries in the plot anyway. Also sort the entries.

knelli2:
Examples of output now with fixup

no_color

with_color

nilsdeppe:
LGTM, once @nilsvu is happy with the python code, please go ahead and merge. No need for my approval.

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,288 @@
...
+    type=click.Path(file_okay=True, dir_okay=False, writable=True),
+    help=(
+        "Name of the output plot file. If unspecified, the plot is "
+        "shown interactively, which only works on machines with a "
+        "window server."
+    ),
+)
+# Plotting options
+@click.option(
+    "--use-mb/--use-gb",
```
nilsvu, line 39: I'd remove this. Either make it always GB or switch to MB if the total memory is below 1 GB. This is just another option the person using this has to learn and set.

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,288 @@
...
+    default=True,
+    show_default=True,
+    help="Plot the y-axis in Megabytes or Gigabytes",
+)
+@click.option(
+    "--color-components",
```
nilsvu: Also remove this. The stylesheet sets the color cycle (see PlotPowerMonitors.py how to evaluate the color cycle if needed).

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,288 @@
...
+                totals_df[component] / divisor,
+                linewidth=0.2,
+                # Remove .dat extension
+                label=component[:-4],
+            )
+        else:
+            ax.plot(
+                totals_df["Time"],
+                totals_df[component] / divisor,
+                color="grey",
```
nilsvu: There's no way to distinguish these lines (no legend, no colors or other property cycles). Just delete the nocolor option and have the stylesheet define the style.

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,288 @@
...
+        ax.set_xlabel(x_label)
+    if color_components:
+        # Total is always first
+        maxes = [inf]
+        for component in components_to_plot:
+            maxes.append(totals_df[component].max())
+        # Now that we have all the maxes, we need to sort them and get the
+        # permutation of the sort for the legend so it's in order of largest at
+        # the top, smallest at the bottom
+        maxes_tuple = [(maxes[i], i) for i in range(len(maxes))]
```
nilsvu: use `np.argsort` to get the permutation. you can also do this earlier and then plot in the right order so you don't have to reorder the legend.

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,288 @@
...
+
+        # Restrict to x-bounds if there are any
+        if x_bounds is not None:
+            df_dict[memory_filename] = df_dict[memory_filename][
+                (df_dict[memory_filename]["Time"] >= x_bounds[0])
+                & (df_dict[memory_filename]["Time"] <= x_bounds[1])
+            ]
+
+    # Given a DataFrame, sum all the columns that list totals for each node
+    def compute_individual_total(df: pd.DataFrame):
```
nilsvu: Maybe `total_over_nodes`?

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,288 @@
...
+            # "Size (MB)" is for a singleton. "Size on node" is for groups,
+            # nodegroups, and arrays
+            if column == "Size (MB)" or "Size on node" in column:
+                if total is not None:
+                    total += df[column]
+                else:
+                    # This copy is important because we want a new column/array.
+                    # Otherwise we'd just be storing a reference and we'd change
+                    # the values in 'df'.
+                    total = df[column].copy()
```
nilsvu: You read every H5 file multiple times, once for each parallel component. Not sure how much data we'll deal with here, but to fix this and read each h5file only once you can do this: `pd.concat(pd.DataFrame({subfile: total_over_nodes(to_dataframe(subfile)) for subfile in memory_filenames}) for h5file in h5files)` (pseudocode, maybe not do this in 1 line :D). This also gets you a single DataFrame of all parallel components over time, so you don't have to deal with dicts anymore.

nilsvu (changes requested):

src/Visualization/Python/PlotMemoryMonitors.py
knelli2: I'd prefer to keep this one. I did make the default GB though since that's more likely

nilsvu: Ok

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,269 @@
...
+                components_to_plot.append(component_name)
+
+        if the_rest is not None:
+            components_to_plot.append(the_rest_str)
+
+        return total, the_rest, components_to_plot
+
+    # Now add the total and the rest to the totals DataFrame
+    total_series, the_rest_series, components_to_plot = (
+        compute_overall_total_and_the_rest(totals_df)
```
nilsvu: You do this only once, so you don't need the function I think? Also update the comment about "dicts" above.

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,269 @@
...
+    # Given a dictionary of DataFrames, sum all the columns from all the
+    # DataFrames that list totals for each node to get a grand total memory
+    # usage. Then return a DataFrame with the time and this new total
+    def compute_overall_total_and_the_rest(df: pd.DataFrame):
+        total = None
+        for component_name in memory_filenames:
+            if total is not None:
+                total += df[component_name]
+            else:
+                total = df[component_name].copy()
```
nilsvu: This is just `total = df.sum(index=1)`

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,269 @@
...
+        the_rest = None
+        for component_name in memory_filenames:
+            component_total = df[component_name]
+            if component_total.max() < 0.01 * average_max:
+                if the_rest is not None:
+                    the_rest += component_total
+                else:
+                    the_rest = component_total.copy()
+            else:
+                components_to_plot.append(component_name)
```
nilsvu:
```py
rest_threshold = 0.01 * total.mean()
components_to_plot = df.columns[df.max() > rest_threshold]
rest_columns = [col for col in df.columns if col not in cols]
the_rest = df[rest_columns].sum(axis=1)
```

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,269 @@
...
+    ax.plot(
+        totals_df["Time"],
+        totals_df["Total"] / divisor,
+        color="black",
+        label="Total",
+    )
+
+    # Determine plotting order
+    maxes = [totals_df[component].max() for component in components_to_plot]
+    permutation = np.argsort(maxes)[::-1]
```
nilsvu: (optional)
```py
maxes = totals_df[components_to_plot].max()
components_to_plot = components_to_plot[np.argsort(maxes)[::-1]]
```

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,269 @@
...
+
+    # Plot the individual components
+    for idx in permutation:
+        component = components_to_plot[idx]
+        ax.plot(
+            totals_df["Time"],
+            totals_df[component] / divisor,
+            linewidth=0.2,
```
nilsvu, line 199: Is the line width needed because the time sampling is so dense? Generally better to leave style up to the stylesheet.

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,269 @@
...
+        subfile_path = f"/MemoryMonitors/{subfile_name}"
+        subfile = h5file.get(subfile_path)
+        if subfile_path is None:
+            raise click.UsageError(
+                f"Unable to open memory file '{subfile_path}'"
+                f" from h5 file {h5_filename}. Available subfiles are:\n"
+                f" {available_subfiles(h5file, extension='.dat')}"
+            )
+
+        return to_dataframe(subfile)
```
nilsvu: `return to_dataframe(subfile).set_index("Time")` makes everything below easier

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,269 @@
...
+            if column == "Size (MB)" or "Size on node" in column:
+                if total is not None:
+                    total += df[column]
+                else:
+                    # This copy is important because we want a new column/array.
+                    # Otherwise we'd just be storing a reference and we'd change
+                    # the values in 'df'.
+                    total = df[column].copy()
+
+        return total
```
nilsvu: Want to sum over all columns? Then just do `df.sum(axis=1)`. Otherwise `df[cols_to_sum].sum(axis=1)`. Suggestion:

```py
def total_over_nodes(df):
    cols_to_sum = [col == "Size (MB)" or "Size on node" in col for col in df.columns]
    return df[cols_to_sum].sum(axis=1)
```

src/Visualization/Python/PlotMemoryMonitors.py
```diff
@@ -0,0 +1,269 @@
...
+    # subfile/component name.
+    totals_df = pd.DataFrame()
+    for reduction_file in reduction_files:
+        h5file = h5py.File(reduction_file)
+        local_totals_df = pd.DataFrame()
+
+        for subfile_name in memory_filenames:
+            df = check_memory_monitor_file(reduction_file, h5file, subfile_name)
+            if "Time" not in local_totals_df:
+                local_totals_df["Time"] = df["Time"]
```
nilsvu: can delete this

nilsvu (changes requested):
Some simplifications with pandas. You can squash everything.

src/Visualization/Python/PlotMemoryMonitors.py
knelli2: I did this so the total line would stand out in comparison to the other lines

nilsvu: Ah ok, maybe the color is enough? Up to you.

nilsvu (approved):

PR #5820, opened 2024-03-02
Skeleton of full MC time step

ffoucart:
## Proposed changes

Lays down the main steps required to evolve Monte-Carlo packets within a given element, once we include absorption and elastic scattering. The PR does not actually change the behavior of the code so far (most steps are placeholders that need to be implemented), but it serves to remind me of what needs to be done to match the SpEC algorithm.

### Upgrade instructions

NA

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
+++ b/src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
@@ -97,10 +97,27 @@ void evolve_single_packet_on_geodesic(
   packet->time += time_step;
 }
 
+// Functions to be implemented to complete implementation of Monte-Carlo
+// time step
+double compute_fluid_frame_energy(const Packet& /*packet*/) { return -1.0; }
+void compute_opacities(const gsl::not_null<double*> absorption_opacity,
+                       const gsl::not_null<double*> scattering_opacity,
+                       const double& /*fluid_frame_energy*/) {
+  *absorption_opacity = 0.0;
+  *scattering_opacity = 0.0;
+}
+void scatter_packet(const gsl::not_null<Packet*> /*packet*/) {
+  // To be implemented
+}
+void diffuse_packet(const gsl::not_null<Packet*> /*packet*/,
+                    const double& /*time_step*/) {
+  // To be implemented
+}
+
 }  // namespace detail
 
 void evolve_packets(
-    gsl::not_null<std::vector<Packet>*> packets, const double& time_step,
+    gsl::not_null<std::vector<Packet>*> packets, const double& final_time,
     const Mesh<3>& mesh,
     const tnsr::I<DataVector, 3, Frame::ElementLogical>& mesh_coordinates,
     const Scalar<DataVector>& lapse,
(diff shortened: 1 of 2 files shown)


src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
```diff
@@ -97,10 +97,26 @@ void evolve_single_packet_on_geodesic(
   packet->time += time_step;
 }
 
+// Functions to be implemented to complete implementation of Monte-Carlo
+// time step
+double compute_fluid_frame_energy(const Packet& /*packet*/) { return -1.0; }
+void compute_opacities(double& absorption_opacity, double& scattering_opacity,
```
nilsdeppe: `double&` -> `const gsl::not_null<double*>` (but not the `const double&` below)

src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
```diff
@@ -97,10 +97,26 @@ void evolve_single_packet_on_geodesic(
...
+                       const double& /*fluid_frame_energy*/) {
+  absorption_opacity = 0.0;
+  scattering_opacity = 0.0;
+}
+void scatter_packet(gsl::not_null<Packet*> /*packet*/) {
```
nilsdeppe: `const gsl...`

src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
```diff
@@ -97,10 +97,26 @@ void evolve_single_packet_on_geodesic(
...
+  // To be implemented
+}
+void diffuse_packet(gsl::not_null<Packet*> /*packet*/,
```
nilsdeppe: `const gsl...`

src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
```diff
@@ -122,25 +138,90 @@ void evolve_packets(
...
+      // Get fluid frame energy of neutrinos in packet, then retrieve
+      // opacities
+      fluid_frame_energy = detail::compute_fluid_frame_energy(packet);
+      detail::compute_opacities(absorption_opacity, scattering_opacity,
+                                fluid_frame_energy);
+
+      // Determine time to next events
+      // TO DO: Implement when including opacities!
+      dt_min = dt_end_step;
+      dt_cell_check = 10.0 * dt_end_step;
```
nilsdeppe, line 178: Could you add a C++ code comment where the `10.0 * X` numbers come from? Specifically, why 10? Even if the answer is "it seems to work well in practice," just so that if we edit this file in the future we have some guidance for these numbers

src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
```diff
@@ -122,25 +138,90 @@ void evolve_packets(
...
+      dt_min = std::min(dt_cell_check, dt_min);
+      dt_absorption = 10.0 * dt_end_step;
+      dt_min = std::min(dt_absorption, dt_min);
+      dt_scattering = 10.0 * dt_end_step;
+      dt_min = std::min(dt_scattering, dt_min);
+
+      // Absorption comes first; delete the packet
```
nilsdeppe: Could you elaborate just to be extra clear:
```cpp
// Handle absorption before emission
//
// To remove this packet we swap the current and last packet,
// then pop the last packet from the end of the vector of packets.
// We then decrease the counter `p` so that we check the packet
// from the end that we just swapped into the current slot.
//
// Note: This works fine even if p==0 since unsigned ints (size_t)
// wrap at zero.
```

src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
```diff
@@ -122,25 +138,90 @@ void evolve_packets(
...
+      if (dt_min == dt_absorption) {
+        (*packets)[p] = (*packets)[n_packets - 1];
```
nilsdeppe: `std::swap((*packets)[p], (*packets)[n_packets - 1]);`

src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
```diff
@@ -122,25 +138,90 @@ void evolve_packets(
...
+        dt_end_step -= dt_min;
+        dt_absorption -= dt_min;
+        dt_cell_check -= dt_min;
+        dt_min = dt_end_step;
+        dt_min = std::min(dt_cell_check, dt_min);
+        dt_min = std::min(dt_absorption, dt_min);
+        // Correct for relativistic effects when using opacities
+        const double scattering_optical_depth = dt_min * scattering_opacity;
+        // High optical depth: use approximate diffusion method to move packet
+        if (scattering_optical_depth > 3.0) {
```
nilsdeppe, line 220: Can you expand on the above comment why the number 3? Even if it's just "in practice this works well" that would be good to have noted

nilsdeppe (dismissed):
Looks good! A few minor suggestions. Please squash right away :)

Reviewers: @kidder @nilsdeppe

src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
ffoucart: This one is part of the "TO DO"; this is not what we want, but depends on the opacities which I have not implemented yet. For now, it just needs to be > dt_end_step

src/Evolution/Particles/MonteCarlo/EvolveMonteCarloPackets.cpp
ffoucart: I'll put a comment here; practically this is a parameter of the "diffusion" method for high-scattering region, which will be coded as a const member of a struct in a coming PR (the next one to be submitted)

ffoucart:
Addressed comments and squashed.

kidder (approved):

PR #5811, opened 2024-02-27
Add InverseJacobianInertialToFluidCompute

ffoucart:
## Proposed changes

Implement a ComputeTag that calculates the inverse jacobian of the map between the inertial frame and an orthonormal tetrad comoving with the fluid.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

I needed to add a new frame (Frame::Fluid) -- should it be in the same file as the other frames, or defined separately in the MC code?

PR #5821, opened 2024-03-02
Calculate packet energy for Monte-Carlo

ffoucart:
## Proposed changes

Add on-the-fly calculation of neutrino energy in fluid frame (needed to calculate interaction rates and perform scattering).

### Upgrade instructions

NA

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Particles/MonteCarlo/EvolvePackets.cpp
+++ b/src/Evolution/Particles/MonteCarlo/EvolvePackets.cpp
@@ -11,8 +11,8 @@ namespace detail {
 
 // Time derivative of p_i in a packet, according to geodesic equation
 void time_derivative_momentum_geodesic(
-    gsl::not_null<std::array<double, 3>*> dt_momentum, const Packet& packet,
-    const Scalar<DataVector>& lapse,
+    const gsl::not_null<std::array<double, 3>*> dt_momentum,
+    const Packet& packet, const Scalar<DataVector>& lapse,
     const tnsr::i<DataVector, 3, Frame::Inertial>& d_lapse,
     const tnsr::iJ<DataVector, 3, Frame::Inertial>& d_shift,
     const tnsr::iJJ<DataVector, 3, Frame::Inertial>& d_inv_spatial_metric) {
@@ -34,7 +34,7 @@ void time_derivative_momentum_geodesic(
 }
 
 void evolve_single_packet_on_geodesic(
-    gsl::not_null<Packet*> packet, const double& time_step,
+    const gsl::not_null<Packet*> packet, const double& time_step,
     const Scalar<DataVector>& lapse,
     const tnsr::I<DataVector, 3, Frame::Inertial>& shift,
     const tnsr::i<DataVector, 3, Frame::Inertial>& d_lapse,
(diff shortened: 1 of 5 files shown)


ffoucart:
Depends on #5820 ; a proper test can be incorporated as part of the test to be added with #5811 , which currently calculates the neutrino energy in the test function itself.

src/Evolution/Particles/MonteCarlo/MonteCarloPacket.hpp
```diff
@@ -60,4 +60,11 @@ struct Packet {
       const Scalar<DataVector>& lapse);
 };
 
+// Calculate energy of neutrinos in a frame comoving with the fluid
```
nilsdeppe: Could you promote this to doxygen? E.g.
```cpp
/*!
 * \brief Calculate energy of neutrinos in a frame comoving with the fluid
 * 
 * \f{align}{
 * E=W\alpha p^t - \gamma^{ij}u_i p_i
 * \f}
 */
```

src/Evolution/Particles/MonteCarlo/MonteCarloPacket.cpp
```diff
@@ -20,4 +20,22 @@ void Packet::renormalize_momentum(
...
+    const Scalar<DataVector>& lapse,
+    const tnsr::II<DataVector, 3, Frame::Inertial>& inv_spatial_metric) {
+  const size_t idx = packet.index_of_closest_grid_point;
+  double fluid_frame_energy =
+      get(lorentz_factor)[idx] * get(lapse)[idx] * packet.momentum_upper_t;
+  for (size_t i = 0; i < 3; i++) {
+    for (size_t j = 0; j < 3; j++) {
+      fluid_frame_energy -= inv_spatial_metric.get(i, j)[idx] *
+                            lower_spatial_four_velocity.get(i)[idx] *
+                            packet.momentum[idx];
```
nilsdeppe: Should this by `j` instead of `idx`? Could you add a test to check for that? A few ideas that come to mind:
1. Set up a packet with `p^t=0` and `p_i={0.1, 0.35, 0.78}` and then check that you get the expected number out
2. Set up a packet with `p^t=0.9` and `p_i={0,0,0}` and then check you get the expected number out

In general just avoid 1 (or zero unless it's deliberate) since a multiplication bug isn't caught if you use 1 (2 is similarly bad, but slightly better). This is why we chose `p^t=0.9`

nilsdeppe (changes requested):

src/Evolution/Particles/MonteCarlo/MonteCarloPacket.cpp
ffoucart: Oh yeah, good catch...

ffoucart:
Added less trivial test for the functions within Packet.hpp, in addition to requested code changes.
The tests of Packet.hpp have also been factorized into a function to separate them from the tests of propagation (which I'll need to factorize and improve in a future PR as well...)

ffoucart:
If anyone understands what could be happening with clang-14 Debug, let me know...

src/Evolution/Particles/MonteCarlo/EvolvePackets.cpp
```diff
@@ -165,7 +169,10 @@ void evolve_packets(
     while (dt_end_step > 0.05 * (final_time - initial_time)) {
       // Get fluid frame energy of neutrinos in packet, then retrieve
       // opacities
-      fluid_frame_energy = detail::compute_fluid_frame_energy(packet);
```
kidder, line 168: remove the detail::compute_fluid_frame_energy

kidder (changes requested):
you can squash this into one commit, including the one new requested change

kidder (approved):

nilsdeppe (approved):

PR #5829, opened 2024-03-05
Simplify pole connectivity when writing surface files

knelli2:
## Proposed changes

Simplifies the pole connectivity by just using a fan of triangles. This prevents triangles from overlapping with each other and avoids weird rendering artifacts when the sphere is partially transparent. See before/after images.

Old outline:
image

New outline:
image

Old transparent (tilted slightly):
image

New transparent (tilted slightly):
image

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/IO/H5/Test_VolumeData.cpp
+++ b/tests/Unit/IO/H5/Test_VolumeData.cpp
@@ -33,11 +33,10 @@
 
 namespace {
 void test_strahlkorper() {
-  constexpr size_t l_max = 12;
-  constexpr size_t m_max = 12;
+  constexpr size_t l_max = 8;
   constexpr double sphere_radius = 4.0;
   constexpr std::array<double, 3> center{{5.0, 6.0, 7.0}};
-  const ylm::Strahlkorper<Frame::Inertial> strahlkorper{l_max, m_max,
+  const ylm::Strahlkorper<Frame::Inertial> strahlkorper{l_max, l_max,
                                                         sphere_radius, center};
   const ylm::Spherepack& ylm = strahlkorper.ylm_spherepack();
   const std::array<DataVector, 2> theta_phi = ylm.theta_phi_points();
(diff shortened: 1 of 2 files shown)


tests/Unit/IO/H5/Test_VolumeData.cpp
```diff
@@ -33,11 +33,10 @@
 
 namespace {
 void test_strahlkorper() {
-  constexpr size_t l_max = 12;
-  constexpr size_t m_max = 12;
+  constexpr size_t l_max = 4;
```
wthrowe: Why this change?

Edit: OK, I see why, but I think going all the way to 4 is a bit extreme.

tests/Unit/IO/H5/Test_VolumeData.cpp
```diff
@@ -99,6 +98,37 @@ void test_strahlkorper() {
...
+  // Store file connectivity in vector like expected_connectivity
+  std::vector<size_t> file_connectivity(expected_connectivity.size(), 0);
+
+  for (size_t i = 0; i < expected_connectivity.size(); i++) {
+    file_connectivity[i] = static_cast<size_t>(connectivity_data[i]);
+  }
+
+  // Sort to check connectivity is the same since elementwise comparison is not
+  // required or accurate
+  std::sort(file_connectivity.begin(), file_connectivity.end());
```
wthrowe: Perhaps I'm misunderstanding something, but isn't the order of the connectivity data extremely important?  It's concatenated triples.

tests/Unit/IO/H5/Test_VolumeData.cpp
```diff
@@ -99,6 +98,37 @@ void test_strahlkorper() {
...
+    strahlkorper_file.close_current_object();
+  }
+
+  // Every 3 numbers is a triangle:
+  // root, root + n * (l + 1), root * (n + 1) * (l + 1)
+  std::vector<size_t> expected_connectivity = {0, 5, 10, 0, 10, 15, 0, 15, 20,
+                                               4, 9, 14, 4, 14, 19, 4, 19, 24};
+
+  // Store file connectivity in vector like expected_connectivity
+  std::vector<size_t> file_connectivity(expected_connectivity.size(), 0);
```
wthrowe: Check that the size is as expected.

tests/Unit/IO/H5/Test_VolumeData.cpp
```diff
@@ -99,6 +98,37 @@ void test_strahlkorper() {
...
+
+  for (size_t i = 0; i < expected_connectivity.size(); i++) {
+    file_connectivity[i] = static_cast<size_t>(connectivity_data[i]);
+  }
```
wthrowe: Define the expected value with the correct type so you don't have to do these conversions.

wthrowe (changes requested):
Alternatively, the algorithm the old code is attempting to implement is good, it's just that the port from SpEC is wrong.  The correct algorithm is `FillPolarHole_TriangularReduction` in `SpEC/Utils/IO/TopologyHelpers.cpp`.  Whichever you prefer.

tests/Unit/IO/H5/Test_VolumeData.cpp
knelli2: I had just copied this from a similar test later in the file. I can remove the sorting

knelli2:
If we were actually doing anything with the topology in some kind of 3D software besides just plotting it, I'd spend the time to implement that SpEC algorithm. But since we are just filling a pole so it looks nice in Paraview, I'll keep the simple method so it's not as hard for future people to understand. It can always be edited if our needs change.

wthrowe (dismissed):
Sounds good.  Squash.

wthrowe (approved):

PR #5830, opened 2024-03-06
Remove some libraries' implicit dependence on Parallel

knelli2:
## Proposed changes

The following files depended on `Parallel` but were in a library that didn't make sense to depend on `Parallel`. Alleviate this by moving the files to different libraries.

- Move `DiagnosticInfo.hpp` to `Parallel/DomainDiagnosticInfo.hpp`
- Move the following three to `ParallelAlgorithms/Actions`
  - `Domain/FunctionsOfTime/FunctionsOfTimeAreReady.hpp`
  - `NumericalAlgorithms/LinearOperators/FilterAction.hpp`
  - `Evolution/DiscontinuousGalerkin/Limiters/LimiterActions.hpp`

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/ParallelDomain/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
...
+add_spectre_library(${LIBRARY} INTERFACE)
+
+spectre_target_headers(
+  ${LIBRARY}
+  INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/src
+  HEADERS
+  DiagnosticInfo.hpp
+  FilterAction.hpp
+  FunctionsOfTimeAreReady.hpp
+  LimiterActions.hpp
```
knelli2: So these files don't really have a lot in common, and also don't necessarily have anything in common with Domain. However, each one depends on `Parallel` and was previously in a library that should *not* have depended on `Parallel`. So the "least bad" solution I could come up with was just put them all in their own library so that that library could depend on `Parallel`. However, I don't particularly like this solution.

Another option is to put each file into its own separate one-file library (in their respective directories) and just have those libraries depend on `Parallel`. Also not great.

Suggestions are welcome because I think this needs to be fixed.

nilsvu: While not great, is there a problem depending on `Parallel`? It's very low level and shouldn't depend on `Domain`, so there's no cyclic dependency I think?

knelli2: Basically I interpret it as `Parallel` == Charm++ and the 4 libraries I moved these files from have nothing to do with charm at all (in my opinion) except for these actions.

Also, I think @nilsdeppe will have `Parallel` depend on `Domain` (or maybe `DomainStructure`?) for the nodegroup changes. I'm not sure on that.

kidder: So here is my suggestions:
-  DiagnosticInfo:  should just be in Parallel.  Instead of passing in the Domain, just pass in the number of blocks
- the other three files I would move to `ParallelAlgorithms/Actions`

knelli2: I like that suggestion. Seems like the best option so far. Any objections @nilsvu @wthrowe @nilsdeppe ?

wthrowe: No objections.

nilsdeppe: Sounds fine to me!

knelli2: Ok pushed new commits to make these changes

kidder (approved):

PR #5831, opened 2024-03-06
Add `FaceQuantitiesCompute` tag

nikwit:
In the iterative scheme that I will open PRs for shortly, the puncture field needs to be evaluated several times each time step and `Actions::SendToWorldtube` is run each time. This new compute tag caches all the quantities on the element faces adjacent to the worldtube that do not need to be re-computed. Doing so, it simplifies the action `SendToWorldtube` quite a bit by off-loading the work to the compute tag.

It also uses the C++20 feature of templated lambdas because we can now ;)

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
@@ -76,8 +76,6 @@ struct SendToWorldtube {
                  domain::Tags::InverseJacobian<Dim, Frame::ElementLogical,
                                                Frame::Inertial>>;
 
-  using inbox_tags = tmpl::list<Worldtube::Tags::SphericalHarmonicsInbox<Dim>>;
-
   template <typename DbTagsList, typename... InboxTags, typename Metavariables,
             typename ArrayIndex, typename ActionList,
             typename ParallelComponent>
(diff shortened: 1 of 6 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -471,6 +472,45 @@ struct FaceCoordinatesCompute : FaceCoordinates<Dim, Frame, Centered>,
 };
 /// @}
 
+/// @{
+/*!
+ * \brief The value of the scalar field and its time derivative on element faces
```
knelli2, line 477: Can you add in the `\details` that if an element isn't on the face of the worldtube boundary that this will return a `nullopt`?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -209,6 +210,61 @@ void BackgroundQuantitiesCompute<Dim>::function(
...
+void FaceQuantitiesCompute::function(
+    gsl::not_null<return_type*> result, const Scalar<DataVector>& psi,
+    const Scalar<DataVector>& pi, const tnsr::i<DataVector, Dim>& phi,
+    const tnsr::I<DataVector, Dim>& shift, const Scalar<DataVector>& lapse,
+    const InverseJacobian<DataVector, Dim, Frame::ElementLogical,
+                          Frame::Inertial>& inv_jacobian,
+    const ::ExcisionSphere<Dim>& excision_sphere, const Element<Dim>& element,
+    const Mesh<Dim>& mesh) {
+  const auto direction = excision_sphere.abutting_direction(element.id());
+  if (direction.has_value()) {
```
knelli2: [optional] Though it's probably default constructed to a `std::nullopt`, probably good to explicitly set the result if we aren't on the worltube boundary.

```cpp
if (not direction.has_value()) {
  result->reset();
  return;
}
...
```
Also reduces the indent and makes things less squished.

src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
```diff
@@ -87,62 +85,44 @@ struct SendToWorldtube {
...
-        db::get<Worldtube::Tags::PunctureField<Dim>>(box);
-    if (puncture_field.has_value()) {
-      const auto& element_id = db::get<domain::Tags::Element<Dim>>(box).id();
-      const auto& excision_sphere = db::get<Tags::ExcisionSphere<Dim>>(box);
-      const auto direction = excision_sphere.abutting_direction(element_id);
-      ASSERT(direction.has_value(), "Should be abutting");
+    const auto& element_id = db::get<domain::Tags::Element<Dim>>(box).id();
+    const auto& excision_sphere = db::get<Tags::ExcisionSphere<Dim>>(box);
+    const auto direction = excision_sphere.abutting_direction(element_id);
+    if (direction.has_value()) {
```
knelli2: [optional] You can avoid the indent by just returning early if the direction doesn't have a value

knelli2 (changes requested):
You can squash directly. Just as a general note, we are noticing that for some executables our memory usage is pretty high so keep that in mind when adding compute tags. This use should be ok because (I believe) you save some re-computation in future PRs

nikwit:
squashed in your changes, thanks for the review :)

Regarding the memory usage, the compute tags I use (like this one) are mostly optionals that only have values if the element is abutting worldtube and then only hold tensor data for the face abutting the worldtube. So I do not think this will have any measurable impact on memory compared to the global volume data.

knelli2 (approved):
Ah yes I forgot this will only hold data for a small portion of the elements. Sounds good! :)

PR #5835, opened 2024-03-07
Add gallery to docs

nilsvu:
## Proposed changes

Adds a gallery page so we can start collecting some nice visualizations. We can move this to a nicer landing page later. It currently looks like this:

Bildschirmfoto 2024-03-07 um 12 13 20

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Gallery.md
+++ b/docs/Gallery.md
@@ -1,0 +1,37 @@
+\cond NEVER
+Distributed under the MIT License.
+See LICENSE.txt for details.
+\endcond
+
+\cond NEVER
+Instructions for adding images or videos:
+
+1. Send your media files to a SpECTRE core dev (see CONTRIBUTING.md for details
+   and communication channels). They will place the files in this public Google
+   Drive folder:
+   https://drive.google.com/drive/folders/1xvLplF_pPlkADGfgFNaW_ByVATFv61m-
+2. Open the media files in Google Drive and get their file IDs. To do this,
+   select [...] > Open in new window > Copy the <FILE_ID> in the URL of the form
+   https://drive.google.com/file/d/<FILE_ID>/view.
+3. Add a section to this page.
+   - Embed images like this:
+     <img src="https://drive.google.com/thumbnail?id=<FILE_ID>&sz=w<WIDTH>"/>
+     Replace <WIDTH> with the width you want your image to have on the page.
+     Google Drive will automatically provide an image of this size.
+   - Embed videos like this:
+     \htmlonly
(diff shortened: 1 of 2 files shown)


docs/Gallery.md
```diff
@@ -0,0 +1,35 @@
...
+See LICENSE.txt for details.
+\endcond
+
+\cond NEVER
+Instructions for adding images or videos:
+
+1. Send your media files to a SpECTRE core dev (see CONTRIBUTING.md for details
+   and communication channels). They will place the files in this public Google
+   Drive folder:
+   https://drive.google.com/drive/folders/16txwJiFB_0jTCfAY0WYiINj2m2J3NEy1
```
nilsdeppe: My understanding was we would just use a GitHub repo for images. Who owns this drive folder? If it's not the sxs-collaboration GMail then I think it's a bad idea to use it since if (when) someone leaves science, then we might lose everything in it. A GitHub repo seems by far the safest method...

nilsdeppe (changes requested):

docs/Gallery.md
nilsvu: Yea GitHub + YouTube is another option, but not as good I think. I chose Google Drive for these reasons:

- Main reason: I don't want to deal with resizing images. We need a small version of each image for the gallery page to load reasonably quickly (can't have loads of hi-res images on one page). Google Drive provides a preview image of any requested size automatically. That's very convenient actually.
- Permissions seem easy to manage. Core devs all have write access to the folder (accept the invitation in your email). Managing permissions for Google Drive isn't any better or worse than YouTube.
- Easier to put images and videos in one place, than to manage both a GitHub repo (which is not ideal for binary files in the first place) and YouTube.
- An embedded video player is also provided by Google Drive.
- When we get a landing page on the SXS website we can drop the folder in there and build a Jekyll workflow that resizes images automatically or something like that. But I don't want to deal with that until then.

So the shared folder seems just easier to manage overall.

nilsdeppe: Okay, but who owns this drive folder? I'm still worried if it's owned by an individual that we as a collaboration actually don't own it.

nilsvu: Yep it's owned by the organization. Larry and I set that up just now. It's a folder in the organization's shared drive (hence owned by the organization). The folder has public read permissions and core-dev write permissions.

nilsdeppe (dismissed):
Awesome! Thanks for setting this up :)

nilsvu:
Updated to allow fullscreen videos

nilsdeppe (approved):

PR #5842, opened 2024-03-12
Update Caltech HPC Skylake env

knelli2:
Our SXS reservation was moved from the old CentOS to the new RHEL OS and this required rebuilding some modules

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Environments/caltech_hpc_gcc_skylake.sh
+++ b/support/Environments/caltech_hpc_gcc_skylake.sh
@@ -5,7 +5,7 @@
 
 spectre_setup_modules() {
     module use /central/groups/sxs/modules/
-    echo "Place the following line in you '~/.bashrc' so you don't have to "
+    echo "Place the following line in your '~/.bashrc' so you don't have to "
     echo "run 'spectre_setup_modules' every time you log in:"
     echo ""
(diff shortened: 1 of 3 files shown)


support/Environments/caltech_hpc_gcc_skylake.sh
```diff
@@ -13,12 +13,12 @@ spectre_setup_modules() {
 
 spectre_load_modules() {
     module use /central/groups/sxs/modules/
-    module load libraries/spectre-deps/skylake-2024-02
+    module load libraries/spectre-deps/skylake-2024-03
```
nilsvu, line 16: Comment about the _Icelake_ environment: there's a `> /dev/null 2>&1` that silences any error messages. This also silences any helpful output when I have conflicting modules already loaded and try to run `spectre_load_modules`. Maybe just remove?

knelli2: Ok removed. I originally had that because the intel MPI module kept printing stuff whenever it was loaded/unloaded which I found annoying, but I found a way to silence it in the intel MPI module file

nilsvu (approved):

PR #5844, opened 2024-03-13
Move EventsAndDenseTriggers lib to ParallelAlgorithms

knelli2:
## Proposed changes

EaDT mirrors EaT very closely so it makes sense that these two should be side-by-side in the directory tree. Unfortunately we can't combine EaDT and EaT because 1) The elliptic code depends on EaT 2) EaDT depends on `Time` 3)  We don't want elliptic code to depend on `Time`. Therefore they have to stay separate. I believe this was the original reason it was put in evolution. Also, I will be using EaDT in my interpolation framework changes, and it felt wrong to include files from the evolution directory in the ParallelAlgorithms directory.

This also moves EaDT out of the `evolution` namespace as well, similar to EaT.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/IO/Observer/CMakeLists.txt
```diff
@@ -46,6 +46,7 @@ target_link_libraries(
   Utilities
   INTERFACE
   EventsAndTriggers
+  EventsAndDenseTriggers
```
wthrowe, line 48: Alphabetize.

tests/Unit/ParallelAlgorithms/LinearSolver/Gmres/CMakeLists.txt
```diff
@@ -16,6 +16,7 @@ target_link_libraries(
   Convergence
   DataStructures
   Logging
+  LinearSolverHelpers
```
wthrowe, line 18: Alphabetize.

src/Time/Tags/TimeAndPrevious.hpp
```diff
@@ -17,9 +17,9 @@
 namespace Tags {
 struct Time;
 }  // namespace Tags
-namespace evolution::Tags {
+namespace Tags {
```
wthrowe: [optional] Merge the duplicate `Tags` forward-declaration namespaces.  They occur in other files as well.

wthrowe (changes requested):

knelli2:
Squashed in the changes because there were only a couple

wthrowe (approved):

PR #5845, opened 2024-03-13
Enable printing containers holding non-streamable objects

kidder:
Instead of failing to compile, this will print UNSTREAMABLE as the value of a non-streamable object.

This is useful for debugging.  For example, you can print the contents of a DataBox without having to write stream operators for all the types of items in the DataBox.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Utilities/PrintHelpers.hpp
+++ b/src/Utilities/PrintHelpers.hpp
@@ -10,6 +10,21 @@
 #include <utility>
 #include <vector>
 
+#include "Utilities/Requires.hpp"
+#include "Utilities/TypeTraits/IsStreamable.hpp"
+
+/// \ingroup UtilitiesGroup
+/// \brief Either streams `value` or the string "UNSTREAMABLE"
+template <typename StreamType, typename T>
+StreamType& print_value(StreamType& os, const T& value) {
+  if constexpr (tt::is_streamable_v<StreamType, T>) {
+    os << value;
+  } else {
+    os << "UNSTREAMABLE";
+  }
+  return os;
+}
+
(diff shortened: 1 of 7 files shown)


src/Utilities/GetOutput.hpp
```diff
@@ -6,13 +6,15 @@
...
+
 /*!
  * \ingroup UtilitiesGroup
  * \brief Get the streamed output of `t` as a `std::string`
  */
 template <typename T>
 std::string get_output(const T& t) {
   std::ostringstream os;
-  os << t;
+  print_value(os, t);
```
wthrowe: Similar to the discussion in #4794, I don't like modifying the behavior of `get_output` because it is used in tests to emulate `operator<<`.  And if you do change it, update the documentation.

src/Utilities/PrintHelpers.hpp
```diff
@@ -10,6 +10,27 @@
...
 #include <vector>
 
+#include "Utilities/Requires.hpp"
+#include "Utilities/TypeTraits/IsStreamable.hpp"
+
+/// @{
+/// \ingroup UtilitiesGroup
+/// \brief Either streams `value` or the string "UNSTREAMABLE"
+template <typename StreamType, typename T,
+          Requires<tt::is_streamable<StreamType, T>::value> = nullptr>
```
wthrowe: Prefer `if constexpr`.

wthrowe (changes requested):

wthrowe (approved):

PR #5847, opened 2024-03-13
Add tags for worldtube iterations

nikwit:
Changes the self force options to an `Auto` class and introduces the tags `MaxIterations` and `CurrentIteration` which will be required for the iterative scheme of the worldtube.

~depends on #5831~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/InitializeCurrentIteration.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/InitializeCurrentIteration.hpp
@@ -1,0 +1,30 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+
+#include "DataStructures/DataBox/Protocols/Mutator.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+
+namespace CurvedScalarWave::Worldtube::Initialization {
+
+/*!
+ * \brief Sets the initial value of `CurrentIteration to 0.
+ */
+struct InitializeCurrentIteration : tt::ConformsTo<db::protocols::Mutator> {
+  using return_tags =
+      tmpl::list<CurvedScalarWave::Worldtube::Tags::CurrentIteration>;
+  using argument_tags = tmpl::list<>;
+  using simple_tags = return_tags;
+  using compute_tags = tmpl::list<>;
+  using simple_tags_from_options = tmpl::list<>;
+  using const_global_cache_tags = tmpl::list<>;
+  using mutable_global_cache_tags = tmpl::list<>;
+  static void apply(const gsl::not_null<size_t*> current_iteration) {
+    *current_iteration = 0;
+  }
+};
+}  // namespace CurvedScalarWave::Worldtube::Initialization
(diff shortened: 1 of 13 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -246,13 +244,33 @@ struct Charge : db::SimpleTag {
...
+};
+
+/*!
+ * \brief The mass of the scalar charge. Only has a value is `ApplySelfForce` is
+ * true.
  */
 struct Mass : db::SimpleTag {
-  using type = double;
-  using option_tags = tmpl::list<OptionTags::Mass>;
+  using type = std::optional<double>;
```
knelli2, line 260: An idea to maybe avoid having to hold individual optionals in tags for self force things and check if they are available: You could just store the `SelfForceOptions` that were constructed from the input file in a tag. Then that tag can be used instead of `ApplySelfForce` (you just check the optional), and also wherever you need it's parameters. Thoughts?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -274,16 +283,18 @@ struct Mass : db::SimpleTag {
...
 /*!
- * \brief Whether to apply the acceleration due to the scalar self-force to the
- * particle
+ * \brief The maximum number of iterations that will be applied to the
+ * acceleration of the particle.
  */
-struct ApplySelfForce : db::SimpleTag {
-  using type = bool;
-  using option_tags = tmpl::list<OptionTags::ApplySelfForce>;
+struct MaxIterations : db::SimpleTag {
```
knelli2, line 275: With this change, you now implicitly convert a `size_t` to a `bool` wherever you had a check for `ApplySelfForce`. Could you change those to explicitly check if the `size_t > 0`? Also, depending on your thoughts about my other comment, this may be moot.

src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/InitializeCurrentIteration.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include "DataStructures/DataBox/Protocols/Mutator.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+
+namespace CurvedScalarWave::Worldtube::Initialization {
+
+/*!
+ * \brief Sets the initial value of `CurrentIteration to 0.
+ */
+struct InitializeCurrentIteration : tt::ConformsTo<db::protocols::Mutator> {
```
knelli2, line 17: Did you want to add this to the executable?

knelli2 (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/InitializeCurrentIteration.hpp
nikwit: yes, it is added in #5855

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: I removed `ApplySelfForce` now in favor of just checking whether `MaxIterations` is greater than 0. I think I would prefer to hide the extraction of these quantities in the OptionParsing, rather than having to get them from the struct each time step.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: Removed `ApplySelfForce` in fixup

knelli2 (commented):
Looks good. You can squash

knelli2:
@nikwit Could you give this a quick rebase just to run CI again? Otherwise it looks good and I'll merge it after CI finishes.

knelli2 (approved):

PR #5852, opened 2024-03-14
Allow multi-line names for spectre test cases

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SpectreAddCatchTests.cmake
+++ b/cmake/SpectreAddCatchTests.cmake
@@ -143,6 +143,12 @@ function(spectre_parse_file SOURCE_FILE TEST_TARGET)
     # Strip wrapping quotation marks of each element of the list
     # TEST_STRINGS
     string(REGEX REPLACE "^\"(.*)\"$" "\\1" TEST_STRINGS "${TEST_STRINGS}")
+    # If a test name went on multiple lines, there will either be a sequence
+    # .";" or ";". from the line break (after replacing all the whitespace with
+    # ;) Remove these from the test string
+    string(REPLACE ".\";\"" "." TEST_STRINGS "${TEST_STRINGS}")
+    string(REPLACE "\";\"." "." TEST_STRINGS "${TEST_STRINGS}")
+    # Remove the last set of quotation marks
     string(REPLACE "\";\"" ";" TEST_STRINGS "${TEST_STRINGS}")
 
     # Validate that a test name and tags have been provided
--- a/cmake/SpectreParseTests.py
+++ b/cmake/SpectreParseTests.py
@@ -99,10 +99,18 @@ def parse_source_file(file_name):
         # Capture the name of the test into the first group and the tags into
         # the second. For example,
         # "Unit.My.Test", "[Unit][My]"
-        # group(1) == Unit.My.Test
+        # test_name == Unit.My.Test
         # group(2) == [Unit][My]
-        parsed_name = re.search(r'"(.*)",[\s]*"(.*)"', test_name)
-        test_name = parsed_name.group(1)
+        #
+        # Also handle cases where the name is very long and goes onto multiple
+        # lines,
+        # "Unit.My.Test.Name."
+        #     "Is.Very.Long",
+        # "[Unit][My]"
+        # test_name == Unit.My.Test.Name.Is.Very.Long
+        # group(2) == [Unit][My]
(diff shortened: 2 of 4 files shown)


cmake/SpectreParseTests.py
```diff
@@ -101,8 +101,18 @@ def parse_source_file(file_name):
         # "Unit.My.Test", "[Unit][My]"
         # group(1) == Unit.My.Test
         # group(2) == [Unit][My]
-        parsed_name = re.search(r'"(.*)",[\s]*"(.*)"', test_name)
-        test_name = parsed_name.group(1)
+        #
+        # Also handle cases where the name is very long and goes onto multiple
+        # lines (this requires the name to have parentheses because of how C++
+        # macros work),
```
wthrowe: I don't understand this comment.  That's not something I would expect the preprocessor to care about.

wthrowe (changes requested):
Can you give an example of how to use this?  I can't get it to work.

cmake/SpectreParseTests.py
knelli2: This is the use case I found that didn't work (on a branch, not develop):

```cpp
SPECTRE_TEST_CASE(("Unit.ParallelAlgorithms.Interpolation.Runtime.Callbacks."
                   "ObserveTimeSeriesOnSurface"),
                  "[Unit]") {
  test_option_parsing();
}
```

When I didn't have the `()` surrounding the test name, CMake complained. Also in other instances of macros, I found that I sometimes had to wrap a multi-line argument in `()`

wthrowe: If CMake is complaining that's not a problem with macros.  CMake isn't running the preprocessor.

Have you checked that your test actually runs?  If I try something like what you have the test is silently ignored.

knelli2: Yup my test runs

wthrowe: Did you re-run cmake?  If I copy-paste your example into a file it's silently ignored.

knelli2: Ok now it works with the test case

```cpp
SPECTRE_TEST_CASE("Unit.ParallelAlgorithms.Interpolation.Runtime.Callbacks."
                  "ObserveTimeSeriesOnSurface",
                  "[Unit]") {
  test_option_parsing();
}
```

I wasn't actually running the test through `ctest`. I was just running the executable directly and the test did run which is why I thought it had worked. Now I checked with `ctest`

nilsdeppe: I think the regex I wrote in CMake is probably the part that's failing (or maybe it's regex in python, I forget what parses these nowadays). Just an FYI

knelli2: Turned out to be both :)

cmake/SpectreAddCatchTests.cmake
```diff
@@ -143,6 +143,8 @@ function(spectre_parse_file SOURCE_FILE TEST_TARGET)
     # Strip wrapping quotation marks of each element of the list
     # TEST_STRINGS
     string(REGEX REPLACE "^\"(.*)\"$" "\\1" TEST_STRINGS "${TEST_STRINGS}")
+    string(REPLACE ".\";\"" "." TEST_STRINGS "${TEST_STRINGS}")
+    string(REPLACE "\";\"." "." TEST_STRINGS "${TEST_STRINGS}")
```
wthrowe, line 150: Update the comment on this block of replacements or put the ones to join the list entries in a separate block.

cmake/SpectreParseTests.py
wthrowe: Seems to work without parentheses now, but it is silently ignored with parentheses.  This is probably fine, but remove anything related to handling parentheses.

cmake/SpectreParseTests.py
```diff
@@ -101,8 +101,18 @@ def parse_source_file(file_name):
...
+        # ("Unit.My.Test.Name."
+        #     "Is.Very.Long"),
+        # "[Unit][My]"
+        # group(1) == Unit.My.Test.Name.Is.Very.Long
```
wthrowe: `group(1)` includes some quotes and whitespace.

wthrowe (changes requested):

knelli2:
Squashed all the changes because they were short

wthrowe (approved):

PR #5853, opened 2024-03-15
Add dev containers for all of our images

knelli2:
## Proposed changes

Makes opening the repo in the different containers significantly easier than having to edit a single dev container file every time.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2:
@nilsvu What do you think of these changes? I think they will be useful for SXS-con and also in general for people who want to use different containers to try things.

nilsvu:
Yea seems useful, I'm only worried about maintaining all these files. I see some merge logic noted in the docs https://containers.dev/implementors/spec/, so is it possible to have the different files inherit most fields from a parent file?

knelli2:
From a quick glance it looks like several fields we use are missing merge logic so we'd still have to copy a decent amount. Not sure if it's worth it to rely on merge logic.

knelli2:
@knelli2 and @nilsvu will discuss more

PR #5854, opened 2024-03-15
Add RotScaleTrans Map to BCO

AlexCarpenter46:
## Proposed changes

Adds the RotScaleTrans map to the BCO domain. Also enables the translation control and translation options in the BBH/BNS executable.

### Upgrade instructions

Will need to add the translation map options in the TimeDependentMaps portion of the yamls using the BCO domain as well as add the control system for it.
```yaml
      TranslationMap:  # in domain options
        InitialValues:
          - [0.0, 0.0, 0.0]
          - [0.0, 0.0, 0.0]
          - [0.0, 0.0, 0.0]
# ...
  Translation:  # in control system options
    IsActive: true
    Averager:
      AverageTimescaleFraction: 0.25
      Average0thDeriv: false
    Controller:
      UpdateFraction: 0.3
    TimescaleTuner:
      InitialTimescales: [0.2, 0.2, 0.2]
      MinTimescale: 1.0e-2
      MaxTimescale: 20.0
      IncreaseThreshold: 2.5e-4
      DecreaseThreshold: 1.0e-3
      IncreaseFactor: 1.01
      DecreaseFactor: 0.98
    ControlError:
```
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/BinaryCompactObject.hpp
+++ b/src/Domain/Creators/BinaryCompactObject.hpp
@@ -114,10 +114,10 @@ namespace creators {
  * \par Time dependence:
  * The following time-dependent maps are applied:
  *
- * - A `CubicScale` expansion and a `Rotation` applied to all blocks from the
- *   Grid to the Inertial frame. However, if there is a size map in the block
- *   (defined below), then the expansion and rotation maps go from the Distorted
- *   to the Inertial frame.
+ * - A piecewise `Expansion`, a `Rotation` and a piecewise `Translation` is
+ * applied to all blocks from the Grid to the Inertial frame. However, if there
+ * is a shape map in the block (defined below), then the expansion, rotation,
+ * and translation maps go from the Distorted to the Inertial frame.
  * - If an object is excised, then the corresponding shell has a
  *   `Shape` map. The shape map goes from the Grid to the Distorted frame.
  *
@@ -504,6 +504,7 @@ class BinaryCompactObject : public DomainCreator<3> {
   double length_inner_cube_{};
   double length_outer_cube_{};
   size_t number_of_blocks_{};
+  size_t first_outer_shell_block_{};
   std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
       outer_boundary_condition_;
   std::vector<std::string> block_names_{};
--- a/src/Domain/Creators/BinaryCompactObjectHelpers.hpp
+++ b/src/Domain/Creators/BinaryCompactObjectHelpers.hpp
@@ -13,6 +13,7 @@
 #include "Domain/CoordinateMaps/CoordinateMap.hpp"
 #include "Domain/CoordinateMaps/Identity.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/CubicScale.hpp"
+#include "Domain/CoordinateMaps/TimeDependent/RotScaleTrans.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/Rotation.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/Shape.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp"
@@ -127,6 +128,7 @@ struct TimeDependentMapOptions {
   // Time-dependent maps
   using Expansion = domain::CoordinateMaps::TimeDependent::CubicScale<3>;
   using Rotation = domain::CoordinateMaps::TimeDependent::Rotation<3>;
+  using RotScaleTrans = domain::CoordinateMaps::TimeDependent::RotScaleTrans<3>;
   using Shape = domain::CoordinateMaps::TimeDependent::Shape;
   using Identity = domain::CoordinateMaps::Identity<3>;
 
@@ -143,7 +145,11 @@ struct TimeDependentMapOptions {
                                Rotation>,
       detail::produce_all_maps<Frame::Grid, Frame::Distorted, Shape>,
       detail::produce_all_maps<Frame::Distorted, Frame::Inertial, Expansion,
-                               Rotation>>;
+                               Rotation>,
+      detail::produce_all_maps<Frame::Grid, Frame::Inertial, Shape,
+                               RotScaleTrans>,
+      detail::produce_all_maps<Frame::Distorted, Frame::Inertial,
+                               RotScaleTrans>>;
 
   /// \brief The initial time of the functions of time.
   struct InitialTime {
@@ -206,6 +212,27 @@ struct TimeDependentMapOptions {
     std::array<double, 3> initial_angular_velocity{};
   };
 
+  /// \brief Options for the Translation Map, the outer radius is always set to
+  /// the outer boundary of the Domain, so there's no option needed for outer
+  /// boundary.
+  struct TranslationMapOptions {
+    using type = Options::Auto<TranslationMapOptions, Options::AutoLabel::None>;
+    static std::string name() { return "TranslationMap"; }
+    static constexpr Options::String help = {
+        "Options for a time-dependent translation map. Specify 'None' to not "
+        "use this map."};
+
+    struct InitialValues {
+      using type = std::array<std::array<double, 3>, 3>;
+      static constexpr Options::String help = {
+          "Initial position, velocity and acceleration."};
+    };
+
+    using options = tmpl::list<InitialValues>;
+
+    std::array<std::array<double, 3>, 3> initial_values{};
+  };
+
   // We use a type alias here instead of defining the ShapeMapOptions struct
   // because there appears to be a bug in clang-10. If the definition of
   // ShapeMapOptions is here inside TimeDependentMapOptions, on clang-10 there
@@ -221,7 +248,7 @@ struct TimeDependentMapOptions {
 
   using options =
       tmpl::list<InitialTime, ExpansionMapOptions, RotationMapOptions,
-                 ShapeMapOptions<domain::ObjectLabel::A>,
+                 TranslationMapOptions, ShapeMapOptions<domain::ObjectLabel::A>,
                  ShapeMapOptions<domain::ObjectLabel::B>>;
   static constexpr Options::String help{
       "The options for all time dependent maps in a binary compact object "
@@ -233,6 +260,7 @@ struct TimeDependentMapOptions {
       double initial_time,
       std::optional<ExpansionMapOptions> expansion_map_options,
       std::optional<RotationMapOptions> rotation_options,
+      std::optional<TranslationMapOptions> translation_options,
       std::optional<ShapeMapOptions<domain::ObjectLabel::A>> shape_options_A,
       std::optional<ShapeMapOptions<domain::ObjectLabel::B>> shape_options_B,
       const Options::Context& context = {});
@@ -246,6 +274,7 @@ struct TimeDependentMapOptions {
    * - Expansion: `PiecewisePolynomial<2>`
    * - ExpansionOuterBoundary: `FixedSpeedCubic`
    * - Rotation: `QuaternionFunctionOfTime<3>`
+   * - Translation: `PiecewisePolynomial<3>`
    * - SizeA/B: `PiecewisePolynomial<3>`
    * - ShapeA/B: `PiecewisePolynomial<2>`
    */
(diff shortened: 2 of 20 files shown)


AlexCarpenter46:
The Test_Domain.cpp test will fail because the maps have changed so the SerializedDomain we test against in the versioning   has the old domain. Not sure if I should change that file or code the expected domain we should get.

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -257,6 +263,27 @@ struct TimeDependentMapOptions {
     std::array<double, 3> initial_angular_velocity{};
   };
 
+  // \brief Options for the Translation Map, the outer radius is always set to
+  // the outer boundary of the Domain, so there's no option needed for outer
+  // boundary.
```
knelli2: This won't show up in doxygen. Needs and extra "/"

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -257,6 +263,27 @@ struct TimeDependentMapOptions {
...
+  struct TranslationMapOptions {
+    using type = Options::Auto<TranslationMapOptions, Options::AutoLabel::None>;
+    static std::string name() { return "TranslationMap"; }
+    static constexpr Options::String help = {
+        "Options for a time-dependent translation map. Specify 'None' to not "
+        "use this map."};
+
+    struct InitialValues {
+      using type = std::array<std::array<double, 3>, 2>;
```
knelli2: Can you also add the initial acceleration as well? That way it's consistent with the Sphere domain

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -296,6 +324,7 @@ struct TimeDependentMapOptions {
    * - Expansion: `PiecewisePolynomial<2>`
    * - ExpansionOuterBoundary: `FixedSpeedCubic`
    * - Rotation: `QuaternionFunctionOfTime<3>`
+   * - Translation: `PiecewisePolynomial<3>`
```
knelli2, line 277: The docs for all these `map` functions are now wrong since you aren't using the individual maps, but the combined one. Could you update them?

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -114,10 +114,10 @@ namespace creators {
  * \par Time dependence:
  * The following time-dependent maps are applied:
  *
- * - A `CubicScale` expansion and a `Rotation` applied to all blocks from the
- *   Grid to the Inertial frame. However, if there is a size map in the block
- *   (defined below), then the expansion and rotation maps go from the Distorted
- *   to the Inertial frame.
+ * - A piecewise expansion, a `Rotation` and a piecewise translation is applied
+ *   to all blocks from the Grid to the Inertial frame. However, if there is a
+ *   size map in the block (defined below), then the expansion, rotation, and
```
knelli2: `shape map` not `size map`

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -394,19 +398,22 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
...
     PARSE_ERROR(context,
                 "To use the CylindricalBBH domain with time-dependent maps, "
-                "you must include the inner spheres for both objects. "
+                "you must include the inner spheres for both objects and "
+                "the outer sphere. "
                 "Currently, one or both objects is missing the inner spheres.");
   }
 
   if (time_dependent_options_.has_value()) {
+    const double inner_radius_C = 3.0 * (center_A_[2] - center_B_[2]);
```
knelli2: What is `C` here? Could you add a more descriptive name?

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -895,41 +902,47 @@ Domain<3> CylindricalBinaryCompactObject::create_domain() const {
     // map from the grid to inertial frame. No maps to the distorted frame
     grid_to_inertial_block_maps[0] =
         time_dependent_options_
-            ->grid_to_inertial_map<domain::ObjectLabel::None>(false);
+            ->grid_to_inertial_map<domain::ObjectLabel::None>(false, true);
+
+    if (include_outer_sphere_) {
```
knelli2: Doesn't this always have to be true?

tests/Unit/Domain/Creators/Test_BinaryCompactObject.cpp
```diff
@@ -368,7 +368,9 @@ std::string create_option_string(
             "      AsymptoticVelocityOuterBoundary: -0.1\n"
             "      DecayTimescaleOuterBoundaryVelocity: 5.0\n"
             "    RotationMap:\n"
-            "      InitialAngularVelocity: [0.0, 0.0, -0.2]\n"s +
+            "      InitialAngularVelocity: [0.0, 0.0, -0.2]\n"
+            "    TranslationMap:\n"
+            "      InitialValues: [[0.0, 0.0, 0.0], [0.0, 0.0, 0.]]\n"s +
```
knelli2: [nitpick] `0.0` for consistency

tests/Unit/Domain/Creators/Test_BinaryCompactObjectHelpers.cpp
```diff
@@ -241,6 +267,7 @@ void test(const bool include_expansion, const bool include_rotation,
 
   const std::array<std::array<double, 3>, 2> centers{
       std::array{5.0, 0.01, 0.02}, std::array{-5.0, -0.01, -0.02}};
+  const double domain_envelope_radius = 5.0;
```
knelli2: This radius is within the cubes since their centers are at 5.0. Maybe make this something bigger like 15?

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -245,6 +245,9 @@ std::string create_option_string(
                             "    ExpansionMap: None\n"
                             "    RotationMap:\n"
                             "      InitialAngularVelocity: [0.0, 0.0, -0.2]\n"
+                            "    TranslationMap: None\n"
+                            //  "      InitialValues: [[0.0, 0.0, 0.0], [0.0,
+                            //  0.0, 0.0]]\n"
```
knelli2: Remove?

tests/Unit/Domain/Test_Domain.cpp
```diff
@@ -464,6 +475,592 @@ void test_3d_rectilinear_domains() {
   }
 }
 
+// The version testing requires an old version of the TimeDependentMapOptions
```
knelli2, line 478: Let's put all this into an anonymous namespace so we don't have to worry about it conflicting with anything. No need to keep it in the `bco` namespace

tests/InputFiles/GeneralizedHarmonic/BinaryBlackHole.yaml
```diff
@@ -350,6 +352,22 @@ ControlSystems:
       IncreaseFactor: 1.01
       DecreaseFactor: 0.98
     ControlError:
+  Translation:
+    IsActive: true
```
knelli2: Maybe keep this deactivated by default until we've done more testing with it.

knelli2 (changes requested):
Looks really good! Only a few things I noticed. Also please fix the clang-tidy errors as well as the failing input file tests. I believe you just missed some input files that use the BCO domain.

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
AlexCarpenter46: I think I thought what if there's a scenario where someone just wants the rotation map so they technically don't need to specify the outer sphere for that, buuuuuut that's not really useful

AlexCarpenter46: After trying to take it out, I realized I kept this in for the tests, which tests both true and false for include_outer_sphere_. Should I change the test to make sure it errors if you use time-dependent maps without the outer sphere?

knelli2: Yeah no sense in trying to account for the special case of only a rotation. I believe the error in your constructor should catch this.

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -332,7 +330,7 @@ struct TimeDependentMapOptions {
    * `Frame::Inertial`
    *
    * If we are including a shape map, then this will be a composition of an
-   * `CubicScale` and `Rotation` map. If we aren't, this returns a `nullptr`.
+   * `RotScaleTrans` map. If we aren't, this returns a `nullptr`.
```
knelli2: This isn't a composition anymore. Just a single map

src/Evolution/Executables/GeneralizedHarmonic/EvolveGhBinaryBlackHole.hpp
```diff
@@ -16,9 +16,9 @@
 #include "ControlSystem/Metafunctions.hpp"
 #include "ControlSystem/Systems/Expansion.hpp"
 #include "ControlSystem/Systems/Rotation.hpp"
-#include "ControlSystem/Systems/Translation.hpp"
 #include "ControlSystem/Systems/Shape.hpp"
 #include "ControlSystem/Systems/Size.hpp"
+#include "ControlSystem/Systems/Translation.hpp"
```
knelli2, line 21: Make sure this goes into the second commit, not the first. Same for the rest of the changes related to adding the translation system

support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -81,6 +81,8 @@ DomainCreator:
         DecayTimescaleOuterBoundaryVelocity: 50.0
       RotationMap:
         InitialAngularVelocity: [0.0, 0.0, {{ InitialAngularVelocity }}]
+      TranslationMap:
+        InitialValues: [[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]
```
knelli2, line 85: [optional] You could just make this `None` since we aren't using translation control at the moment. Same comment for the other input files

tests/Unit/Domain/Test_Domain.cpp
```diff
@@ -1130,8 +1128,6 @@ Domain<3> create_serialized_domain() {
 
   const std::vector<CoordinateMaps::Distribution> object_A_radial_distribution{
       CoordinateMaps::Distribution::Logarithmic};
-  const std::vector<CoordinateMaps::Distribution> object_B_radial_distribution =
-      object_A_radial_distribution;
```
knelli2: If you're going to keep this change, I'd rename `object_A_radial_distribution` to just `radial_distribution` so it's not confusing why below you have an `_A` thing used in a function call for `_B`.

knelli2 (changes requested):
Only a couple more things. You can squash the previous changes and these new ones in directly

support/Pipelines/Bbh/Inspiral.yaml
AlexCarpenter46: I think I'll keep it in since the map functions change when you choose to not use a certain map, this way it works exactly like spec at the moment.

support/Pipelines/Bbh/InitialData.yaml
```diff
@@ -115,6 +115,7 @@ DomainCreator:
       InitialTime: 0.
       ExpansionMap: None
       RotationMap: None
+      TranslationMap: None
```
knelli2, line 118: Oh actually wait. All the input file changes that have `TranslationMap` should go in the first commit, otherwise it won't compile correctly on its own. The only input file changes in the second commit should be the translation control system. That was my bad for saying the wrong thing above

knelli2 (changes requested):

support/Pipelines/Bbh/InitialData.yaml
AlexCarpenter46: Oh whoops yeah, I didn't think about that either lol, squashed in that fix :)

knelli2 (dismissed):

AlexCarpenter46:
I missed adding Translation control to the CylindricalBBH, but it's fixed now!

knelli2 (approved):

PR #5855, opened 2024-03-16
Add iterative scheme to worldtube

nikwit:
Adds the iterative scheme to the worldtube, as described in section V A of https://arxiv.org/abs/2403.08864.

The worldtube sends the updated acceleration to the adjacent elements which use it to compute the updated puncture field and then the updated regular field. This is send back to the worldtube and used to update the acceleration which can be repeated an arbitrary number of times. An action test of this iterative scheme is included.

~depends on #5847~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
@@ -22,6 +22,7 @@
 #include "Domain/TagsTimeDependent.hpp"
 #include "Evolution/Systems/CurvedScalarWave/System.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/ReceiveWorldtubeData.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
@@ -114,7 +115,9 @@ struct SendToWorldtube {
     const auto& area_element =
         get<gr::surfaces::Tags::AreaElement<DataVector>>(face_quantities);
     const auto& puncture_field =
-        db::get<Worldtube::Tags::PunctureField<Dim>>(box).value();
+        db::get<Tags::CurrentIteration>(box) > 0
+            ? db::get<Tags::IteratedPunctureField<Dim>>(box).value()
+            : db::get<Tags::PunctureField<Dim>>(box).value();
     const auto& psi_puncture = get<CurvedScalarWave::Tags::Psi>(puncture_field);
     const auto& dt_psi_puncture =
         get<::Tags::dt<CurvedScalarWave::Tags::Psi>>(puncture_field);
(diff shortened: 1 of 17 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Inboxes.hpp
```diff
@@ -94,4 +94,15 @@ struct RegularFieldInbox
     return ss.str();
   }
 };
+
+/*!
+ * \brief Inbox of the element chares that contains the current iteration of the
+ * acceleration of the particle.
+ */
+template <size_t Dim>
+struct SelfForceInbox : Parallel::InboxInserters::Value<SelfForceInbox<Dim>> {
```
knelli2, line 103: Add the `output_inbox` function and a test that it writes the correct thing

src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
```diff
@@ -161,8 +162,26 @@ struct SendToWorldtube {
     Parallel::receive_data<Worldtube::Tags::SphericalHarmonicsInbox<Dim>>(
         worldtube_component, db::get<::Tags::TimeStepId>(box),
         std::make_pair(element_id, std::move(Ylm_coefs)));
+    if (db::get<Tags::CurrentIteration>(box) <
+        db::get<Tags::MaxIterations>(box) - 1) {
```
knelli2, line 168: Should this logic just go in `IteratePunctureField` instead? I think it makes more sense for `IteratePunctureField` to be the one to increase the `current_iteration` and determine if it actually needs to compute another iteration of the puncture field or go to BCs

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ReceiveElementData.hpp
```diff
@@ -126,7 +128,28 @@ struct ReceiveElementData {
         ylm::ylm_to_stf_0(dt_psi_ylm_l0), std::move(psi_stf_l1),
         std::move(dt_psi_stf_l1));
     inbox.erase(time_step_id);
-    return {Parallel::AlgorithmExecution::Continue, std::nullopt};
+    if (db::get<Tags::CurrentIteration>(box) <
+        db::get<Tags::MaxIterations>(box) - 1) {
```
knelli2, line 132: Same comment here about moving this to `IterateAccelerationTerms`

knelli2 (changes requested):
Also please fix the clang-tidy errors

src/Evolution/Systems/CurvedScalarWave/Worldtube/ElementActions/SendToWorldtube.hpp
nikwit: That's what I had originally but I think this way is better because it automatically handles the cases where `MaxIterations` is equal to 0 (geodesic) or 1 and the `IteratePunctureField` and `IterateAccelerationTerms` actions are skipped entirely.

knelli2: Ahh I see. I guess I was a bit confused because checking the iterations didn't seem related to sending data to the worldtube. But the beginning of the iterate actions can determine if an iteration needs to happen or not. I don't have a strong opinion on this so I'll defer to you.

knelli2 (commented):
Squash

nikwit:
@knelli2 thanks, I decided to leave the iteration decision as is. I noticed that I forgot to actually use the iterated puncture field which I addressed in another fixup commit if you would like to take a look.

knelli2 (commented):
Looks good. There are a couple more clang-tidy errors, so you can squash those in as well.

knelli2 (approved):

knelli2:
Ignoring unrelated test failure

PR #5858, opened 2024-03-17
Add an observer for tracking the size of DataBox items

kidder:
## Proposed changes

This adds an event that can be used to track the sizes of DataBox items.  It is not added to any executable, as it seems to add some compilation time.  To use it, add it as an Event similar to the MonitorMemory event.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Events/ObserveDataBox.hpp
+++ b/src/ParallelAlgorithms/Events/ObserveDataBox.hpp
@@ -1,0 +1,166 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <map>
+#include <pup.h>
+#include <tuple>
+#include <vector>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "IO/Observer/ObserverComponent.hpp"
+#include "IO/Observer/ReductionActions.hpp"
+#include "Options/String.hpp"
+#include "Parallel/DistributedObject.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Invoke.hpp"
+#include "Parallel/Reduction.hpp"
+#include "ParallelAlgorithms/EventsAndTriggers/Event.hpp"
+#include "Utilities/PrettyType.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace Events {
+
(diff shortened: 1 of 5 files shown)


src/ParallelAlgorithms/Events/ObserveDataBox.hpp
```diff
@@ -0,0 +1,159 @@
...
+  template <typename ParallelComponent, typename DbTags, typename Metavariables,
+            typename ArrayIndex>
+  static void apply(db::DataBox<DbTags>& /*box*/,
+                    Parallel::GlobalCache<Metavariables>& cache,
+                    const ArrayIndex& /*array_index*/, const double time,
+                    const std::map<std::string, size_t>& item_sizes) {
+    auto& observer_writer_proxy = Parallel::get_parallel_component<
+        observers::ObserverWriter<Metavariables>>(cache);
+    const std::string subfile_name =
+        "/ObserveDataBox/" + pretty_type::name<ContributingComponent>();
```
nilsdeppe: Should we make this file name be specified in `ObserveDataBox` like we do for other observations? Maybe otherwise name this `ObserveDataBoxSizeInMb`? Long, but extremely unlikely to conflict with anything

src/ParallelAlgorithms/Events/ObserveDataBox.hpp
```diff
@@ -0,0 +1,159 @@
...
+          target_proxy[0], time, item_sizes);
+    } else {
+      Parallel::contribute_to_reduction<ReduceDataBoxSize<ParallelComponent>>(
+          ReductionType{time, item_sizes}, my_proxy, target_proxy[0]);
+    }
+  }
+};
+}  // namespace detail
+
+class ObserveDataBox : public Event {
```
nilsdeppe, line 105: Could you add some doxygen comment so this shows up?

How difficult would it be to write a simple test that just verifies this works with say 1 singleton, 1 nodegroup, and 1 array (with 3 elements?)? If it's difficult, I'm happy to defer since this feature is very necessary

nilsdeppe (changes requested):
A few minor things and some optionals in question form. Let me know what you think, please squash directly :)

nilsdeppe (commented):
the fixup LGTM. Please squash!

nilsdeppe (approved):

PR #5860, opened 2024-03-17
Add 0th order acceleration terms

nikwit:
The terms are added to the puncture field for non-geodesic orbits. They are described in Section IV of https://arxiv.org/abs/2403.08864.


src/Evolution/Systems/CurvedScalarWave/Worldtube/AccelerationTermsOrder0.cpp
```diff
@@ -0,0 +1,292 @@
...
+  const double rp = get(magnitude(particle_position));
+  const double rpdot = (xp * xpdot + yp * ypdot) / rp;
+
+  const auto& Dx = get<0>(centered_coords);
+  const auto& Dy = get<1>(centered_coords);
+  const auto& z = get<2>(centered_coords);
+
+  const double M = bh_mass;
+
+  DynamicBuffer<DataVector> temps(56, grid_size);
```
knelli2, line 48: I think we want to have some guidelines regarding auto-generated code since it's not always the most optimal, just so we aren't baffled in 2 years when we look back at this. Could you

1. Add a comment here saying that the following code was auto-generated using Mathematica (or Sympy, or whatever you used)
2. If you have the code that generated this in a public repo/place, could you also add a comment pointing to it here? This way we can reproduce what you did.
3. Take a cursory look at the auto-generated code and if you notice any quick/simple optimizations that can be made, make those and add a comment where you did so we can note the difference. (I tried to do a couple below)

src/Evolution/Systems/CurvedScalarWave/Worldtube/AccelerationTermsOrder0.cpp
```diff
@@ -0,0 +1,292 @@
...
+  dv_72 = dv_32 * dv_67;
+
+  get(get<CurvedScalarWave::Tags::Psi>(*result)) = -d_23 * dv_27 / dv_12;
+  get(get<::Tags::dt<CurvedScalarWave::Tags::Psi>>(*result)) =
+      -0.5 * dv_28 *
+      (-3. * d_25 * d_42 * dv_27 * dv_38 * dv_39 +
+       d_25 * d_42 * dv_29 *
+           (4. * M * d_14 * d_20 * dv_2 * dv_38 * ft +
+            2. * d_0 * d_14 * dv_14 * dv_38 * fx +
+            2. * d_0 * d_14 * dv_17 * dv_38 * fy - d_20 * dv_24 * dv_7 * dt_ft -
```
knelli2: Factor out `2.0 * dv_14 * dv_38` from the first 3 terms

src/Evolution/Systems/CurvedScalarWave/Worldtube/AccelerationTermsOrder0.cpp
```diff
@@ -0,0 +1,292 @@
...
+            dv_10 * dv_40 - dv_10 * dv_43 - dv_10 * dv_8 * dt_ft -
+            dv_15 * dv_44 - dv_18 * dv_45 - dv_20 * dv_36 - dv_20 * dv_42 -
+            dv_21 * dv_50 - 14 * dv_22 * dv_35 * ft * rpdot - dv_22 * dv_40 -
+            dv_22 * dv_43 - dv_22 * dv_44 * fx - dv_22 * dv_45 * fy -
+            dv_23 * dv_50 - dv_25 * dv_41 -
+            4 * dv_35 * dv_51 * ft * rp * rp * rp * rp * rp * rp * rp -
```
knelli2: `rp * square(cube(rp))`

src/Evolution/Systems/CurvedScalarWave/Worldtube/AccelerationTermsOrder0.cpp
```diff
@@ -0,0 +1,292 @@
...
+  DataVector& dv_62 = temps.at(49);
+  dv_62 = dv_31 * dv_39;
+  DataVector& dv_63 = temps.at(50);
+  dv_63 = 6. * d_8 * dv_2 * dv_20;
+  DataVector& dv_64 = temps.at(51);
+  dv_64 = d_46 * dv_29 * dv_7;
+  DataVector& dv_65 = temps.at(16);
+  dv_65 = 3. * dv_16;
+  DataVector& dv_66 = temps.at(18);
+  dv_66 = 3. * dv_19;
```
knelli2: `dv_63`, `dv_65`, and `dv_66` only appear in the combo `dv_63 + dv_65 + dv_66` below. Combine them here instead and only use `dv_66` below
```
dv_66 = dv_63 + dv_65 + 3. * dv_19
```

src/Evolution/Systems/CurvedScalarWave/Worldtube/AccelerationTermsOrder0.cpp
```diff
@@ -0,0 +1,292 @@
...
+            dv_49 * dv_52) +
+       9. * d_25 * dv_27 * dv_29 * rpdot / (rp * sqrt(rp)) -
+       2. * d_33 * dv_27 * dv_29 * rp * rp * sqrt(rp) / d_25) /
+      (d_24 * sqrt(d_24) * rp * square(square(rp)) * sqrt(rp));
+  get<0>(get<::Tags::deriv<CurvedScalarWave::Tags::Psi, tmpl::size_t<3>,
+                           Frame::Inertial>>(*result)) =
+      -dv_69 * (d_35 * dv_55 + d_36 * dv_53 + d_45 * dv_54 + d_45 * dv_57 +
+                d_47 * dv_58 + d_47 * dv_59 - dv_21 * dv_68 - dv_23 * dv_68 -
+                dv_31 * dv_64 + dv_47 * dv_61 + dv_48 * dv_61 + dv_49 * dv_61 -
+                dv_62 * dv_63 - dv_62 * dv_65 - dv_62 * dv_66);
```
knelli2: (Here for example)

knelli2 (changes requested):
Also address clang-tidy errors please

src/Evolution/Systems/CurvedScalarWave/Worldtube/AccelerationTermsOrder0.cpp
nikwit: I added a description and a link to a new public repo. I incorporated your changes but do not think that all the changes need to be added in the comments, there are quite a few. A diff to the auto-generated file should show them.

knelli2 (approved):
Looks good! Go ahead and squash

nikwit:
Thank you!

knelli2 (approved):

PR #5861, opened 2024-03-17
Add 1st order acceleration terms

nikwit:
Add the first order acceleration terms to the worldtube. 

~depends on #5860~


tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_AccelerationTerms.cpp
```diff
@@ -34,68 +34,102 @@ using puncture_vars =
 // calculation
 void test_derivative() {
   MAKE_GENERATOR(gen);
-  Approx local_approx = Approx::custom().epsilon(1e-12).scale(1.0);
+  Approx local_approx = Approx::custom().epsilon(1e-9).scale(1.0);
```
knelli2, line 37: Why such a large increase?

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp
```diff
@@ -117,4 +117,32 @@ void acceleration_terms_0(
     const tnsr::I<double, 3>& particle_velocity,
     const tnsr::I<double, 3>& particle_acceleration, double ft, double fx,
     double fy, double dt_ft, double dt_fx, double dt_fy, double bh_mass);
+
+/*!
+ * \brief Computes the acceleration terms of a puncture/singular field
```
knelli2, line 122: Are these *only* the first order terms? Or do they include 0th order terms as well? Might be good to clarify if this function has to be called after `acceleration_terms_0` or if it's a standalone function.

knelli2 (changes requested):
There may be some common subexpressions you could factor out, but not really sure if it's worth it. I tried looking and got a headache almost immediately lol. So I'm happy without any changes to the cpp.

src/Evolution/Systems/CurvedScalarWave/Worldtube/PunctureField.hpp
nikwit: I added a small comment

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_AccelerationTerms.cpp
nikwit: The comparison to finite difference very rarely produces an error larger than 1e-10. I assume this is because the first order field is less smooth than the zeroth order field so the finite difference scheme carries a larger error.

nikwit:
Thanks. Yes, the sympy CSE is pretty good but certainly not perfect. I doubt that additional manual optimization will make a measurable difference considering the sheer number of flops in this function

knelli2 (approved):

PR #5862, opened 2024-03-18
Make Newtonian Euler sources runtime and fix circular dependency

nilsdeppe:
## Proposed changes

This is done as 3 commits since logically they are different things. However, the first commit enforces the cyclic dependency we currently ignore in develop. The next two commits correct this circular dependency, at least I'm very sure. It's gone with these and several other commits on my branch that I believe are irrelevant to the circular dependency 🤞 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


kidder (commented):
LGTM

nilsdeppe:
Squashed! Thanks for the review :)

kidder (approved):

PR #5864, opened 2024-03-18
Generalize LTS communication handling

wthrowe:
The old version made assumptions based on the Adams-Bashforth method, but that won't be true for other methods.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/TimeStepId.hpp
+++ b/src/Time/TimeStepId.hpp
@@ -69,12 +69,19 @@ class TimeStepId {
   double substep_time_{};
 };
 
+/// These comparisons define a total order on all TimeStepIds with the
+/// same direction of time flow.  For any two objects representing
+/// states of the same element, the ordering will match the
+/// computational ordering of those states.  For others, (e.g.,
+/// overlapping steps) it is arbitrary.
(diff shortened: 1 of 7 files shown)


src/Time/TimeStepId.hpp
```diff
@@ -69,12 +69,19 @@ class TimeStepId {
   double substep_time_{};
 };
 
+/// These comparisons define a total order on all TimeStepIds with the
+/// same direction of time flow.  For any two objects representing
+/// states of the same element, the ordering will match the
+/// computational ordering of those states.  For others, (e.g.,
+/// overlapping steps) it is arbitrary.
```
knelli2, line 76: Should we maybe say what the arbitrary decision is? Just that for the arbitrary case, we check which step size is smaller?

wthrowe: It's not that simple.  That's just the last tie-breaker.  The part of the algorithm that does the meaningful ordering also distinguishes some arbitrary cases.

kidder (approved):

PR #5865, opened 2024-03-18
Add derivatives of Kerr Schild quantities

nikwit:
The acceleration terms of the worldtube require first and second spatial derivatives of Kerr Schild quantities.  I computed these with Mathematica for a spin-free Kerr-Schild background of mass 1, centered on the coordinate system. As this is a very specific, specialised calculation which I do not expect others to need, I simply added them to the `Worldtube` namespace.

All derivatives have a unit test that checks the analytical derivative against a finite difference numerical derivative.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/KerrSchildDerivatives.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/KerrSchildDerivatives.cpp
@@ -166,4 +166,26 @@ tnsr::iAbb<double, 3> spatial_derivative_christoffel(
   return di_christoffel;
 }
 
+tnsr::iA<double, 3> spatial_derivative_ks_contracted_christoffel(
+    const tnsr::I<double, 3>& pos) {
+  const double r_sq = get(dot_product(pos, pos));
+  const double r = sqrt(r_sq);
+  const double one_over_r = 1. / r;
+  const double one_over_r_2 = 1. / r_sq;
+  const double one_over_r_3 = cube(one_over_r);
+  const double one_over_r_4 = square(one_over_r_2);
+  const double one_over_r_5 = one_over_r_4 * one_over_r;
+
+  tnsr::iA<double, 3> di_contracted_christoffel{};
+  for (size_t i = 0; i < 3; ++i) {
+    di_contracted_christoffel.get(i, 0) = 4. * pos.get(i) * one_over_r_4;
+    for (size_t j = 0; j < 3; ++j) {
+      di_contracted_christoffel.get(i, j + 1) =
+          -6. * pos.get(i) * pos.get(j) * one_over_r_5;
+    }
+    di_contracted_christoffel.get(i, i + 1) += 2. * one_over_r_3;
+  }
+  return di_contracted_christoffel;
+}
+
 }  // namespace CurvedScalarWave::Worldtube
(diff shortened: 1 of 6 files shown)


tests/Unit/Framework/TestHelpers.hpp
```diff
@@ -306,9 +309,14 @@ numerical_derivative(const Invocable& function,
   const std::array<double, VolumeDim> x_1behind = x - dx;
   const std::array<double, VolumeDim> x_2behind = x_1behind - dx;
   const std::array<double, VolumeDim> x_3behind = x_2behind - dx;
-  return (1.0 / (60.0 * delta)) * (function(x_3ahead) - function(x_3behind)) -
-         (3.0 / (20.0 * delta)) * (function(x_2ahead) - function(x_2behind)) +
-         (0.75 / delta) * (function(x_1ahead) - function(x_1behind));
+  return (1.0 / (60.0 * delta)) *
+             (function(x_3ahead, std::forward<Args>(args)...) -
+              function(x_3behind, std::forward<Args>(args)...)) -
```
wthrowe: You shouldn't forward the same thing multiple times.  Forwarding implies that the object may moved from.

It's probably cleaner to wrap your function in a lambda, rather than modifying this helper.  For your uses you can just move your existing wrappers into the loops and capture the loop variables.

src/Evolution/Systems/CurvedScalarWave/Worldtube/KerrSchildDerivatives.cpp
```diff
@@ -0,0 +1,56 @@
...
+#include "DataStructures/Tensor/EagerMath/Trace.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace CurvedScalarWave::Worldtube {
+tnsr::iAA<double, 3> spatial_derivative_inverse_ks_metric(
+    const tnsr::I<double, 3>& pos) {
+  const double r = get(magnitude(pos));
+  const double one_over_r = 1. / r;
+  const double one_over_r_2 = square(one_over_r);
```
wthrowe: Calculate `r^2` as `r.r`, not `magnitude(r)^2 = sqrt(r.r)^2`.

src/Evolution/Systems/CurvedScalarWave/Worldtube/KerrSchildDerivatives.cpp
```diff
@@ -0,0 +1,56 @@
...
+  for (size_t i = 0; i < 3; ++i) {
+    delta_ll.get(i, i) = 1.;
+    delta_ul.get(i, i) = 1.;
+    pos_lower.get(i) = pos.get(i);
+  }
+
+  const auto d_imetric_ij = tenex::evaluate<ti::i, ti::J, ti::K>(
+      6. * pos(ti::J) * pos(ti::K) * pos_lower(ti::i) * one_over_r_5 -
+      2. * delta_ul(ti::J, ti::i) * pos(ti::K) * one_over_r_3 -
+      2. * delta_ul(ti::K, ti::i) * pos(ti::J) * one_over_r_3);
```
wthrowe: [optional] Consider using a Horner-like evaluation here and elsewhere: instead of `a * (1/r^5) + b * (1/r^3)`, do `(a * (1/r^2) + b) * (1/r^3)`.

wthrowe (changes requested):

tests/Unit/Framework/TestHelpers.hpp
nikwit: good catch, I removed the commit.

nikwit:
Thanks for the review, I pushed a fixup

wthrowe:
Looks good except for some gcc warnings.  Squash, including fixing those.

wthrowe (approved):

PR #5871, opened 2024-03-19
Allow Strahlkorper construction from different frame

knelli2:
This is helpful for both runtime choosing of frames and for transition to ringdown

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
@@ -11,6 +11,7 @@
 #include "NumericalAlgorithms/SphericalHarmonics/Spherepack.hpp"
 #include "Options/String.hpp"
 #include "Utilities/ForceInline.hpp"
+#include "Utilities/StdArrayHelpers.hpp"
 
 /// \cond
 namespace PUP {
@@ -49,6 +50,10 @@ class Strahlkorper {
 
   // Pup needs default constructor
   Strahlkorper() = default;
+  Strahlkorper(const Strahlkorper&) = default;
+  Strahlkorper(Strahlkorper&&) = default;
+  Strahlkorper& operator=(const Strahlkorper&) = default;
+  Strahlkorper& operator=(Strahlkorper&&) = default;
 
   /// Construct a sphere of radius `radius` with a given center.
   Strahlkorper(size_t l_max, size_t m_max, double radius,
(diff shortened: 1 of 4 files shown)


src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
```diff
@@ -88,18 +89,51 @@ class Strahlkorper {
...
 
   /// Move-construct a Strahlkorper from another Strahlkorper,
   /// explicitly specifying the coefficients.
-  Strahlkorper(DataVector coefs, Strahlkorper&& another_strahlkorper);
+  template <typename OtherFrame>
+  Strahlkorper(DataVector coefs,
+               Strahlkorper<OtherFrame>&& another_strahlkorper)
+      : l_max_(another_strahlkorper.l_max()),
+        m_max_(another_strahlkorper.m_max()),
+        ylm_(std::move(another_strahlkorper.ylm_spherepack())),
```
wthrowe: `ylm_spherepack()` returns a const reference, so you can't move out of it.

This move was the only useful thing this constructor did, so either restore it somehow or remove the constructor in favor of the const-ref one.

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
```diff
@@ -88,18 +89,51 @@ class Strahlkorper {
                         std::array<double, 3> center);
 
   /// Prolong or restrict another surface to the given `l_max` and `m_max`.
+  template <typename OtherFrame>
```
wthrowe, line 109: For all of these, document that `OtherFrame` is ignored.

wthrowe, line 109: [Arbitrary line] Do you want a frame-change constructor that doesn't do prolongation or similar?

wthrowe (changes requested):

src/NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp
knelli2: The `Strahlkorper{coefs, other_strahlkorper}` achieves that if you use the `other_strahlkorper.coefficients()` for the first argument. I don't really want to deal with the template gymnastics I'll have to do to avoid a conflict between the regular copy constructor, and a "copy constructor" for a different frame.

wthrowe:
```c++
Strahlkorper(const Strahlkorper&) = default;
Strahlkorper(Strahlkorper&&) = default;

template <typename OtherFrame>
Strahlkorper(const Strahlkorper<OtherFrame>& other) : /* ... */ {}
```
?

(Probably `explicit`.)

knelli2:
@wthrowe rebased and added a fixup

wthrowe:
Looks good.  Squash.

wthrowe (approved):

PR #5886, opened 2024-03-25
Store slab-size change info in DataBox instead of inboxes

wthrowe:
It used to be that Events could not modify the DataBox, so the only way to store information was to synchronously send yourself a message to an inbox.  This is no longer true, so this changes to a cleaner organization.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/ControlSystem/Actions/LimitTimeStep.hpp
```diff
@@ -14,7 +14,7 @@
 #include "Parallel/ArrayComponentId.hpp"
 #include "Parallel/Callback.hpp"
 #include "Parallel/GlobalCache.hpp"
-#include "Time/ChangeSlabSize.hpp"
+#include "Time/ChangeSlabSize/ChangeSlabSize.hpp"
```
kidder, line 17: commit message has a typo (mush -> much)

src/Time/ChangeSlabSize/Projector.hpp
```diff
@@ -0,0 +1,67 @@
...
+            children_items.begin()->second);
+    for (const auto& child : children_items) {
+      ASSERT(*number_of_expected_messages ==
+                 tuples::get<::Tags::ChangeSlabSize::NumberOfExpectedMessages>(
+                     child.second),
+             "Different elements expecting different numbers of "
+             "ChangeSlabSize messages.");
+    }
+  }
+};
```
kidder: I believe this projector is currently the same as the generic projector `ParallelAlgorithms/Amr/Projectors/CopyFromCreatorOrLeaveAsIs.hpp` 
I leave it up to you as to whether you want to keep this or use the generic projector...

kidder (changes requested):
- one typo in a commit message, which you can squash
- one optional change to consider

wthrowe:
Used the generic projector and added a commit adding documentation to the AMR namespaces so they show up.

Out of scope, but the metavariables seem like the wrong place for these things.  The projectors should be specified in the same place that the tags are added to the DataBox.

kidder (approved):

kidder:
> Out of scope, but the metavariables seem like the wrong place for these things. The projectors should be specified in the same place that the tags are added to the DataBox.

Yes, we should probably have a core-dev discussion about this...

PR #5895, opened 2024-03-29
Add script to compute eccentricity removal updates for SpECTRE ID

guilara:
## Proposed changes

This script computes updates to the orbital velocity and expansion velocity (adot) for Xcts solves based on fits to the coordinate distance between Objects A and B from evolution of the Xcts data.

It is based on a SpEC script and seems to work for SpECTRE initial data as well (have tried Superposed Harmonic Schwarzschild data evolved in Harmonic gauge and SKS in Damped Harmonic gauge).

Screenshot from 2024-03-29 16-36-46
(Harmonic run)

Usage:

Make sure the Control Systems are writing the trajectories to disk.
`ControlSystems:
  WriteDataToDisk: true`

The data should be stored in the reductions file, e.g. `GhBinaryBlackHoleReductionData.h5`

The user can specify the time window for to apply the algorithm (tmin to exclude junk, tmax to use up to 2-3 orbits), and also the values used in the input xcts file. (If no values are specified, the script just returns the update that needs to be applied to the parameters in the Xcts input file. On the other hand, if they are specified, the script also returns the updated parameters).

Select a good update from the different models by examining the output plot and rms values. A good fit will have small residuals (especially at early times) and a small rms value.

Command example:
`spectre eccentricity-control GhBinaryBlackHoleReductionData.h5 -A 'ApparentHorizons/ControlSystemAhA_Centers.dat' -B 'ApparentHorizons/ControlSystemAhB_Centers.dat' --tmin 100.0 --tmax 1400.0 --angular-velocity-from-xcts 0.012452602615 --expansion-from-xcts -0.0000195689969705  --output EccRemoval.pdf`

An example output is:
```
...
==== Function fitted to dOmega/dt: B*cos(w*t+phi)+quadratic      ,  rms = 6.38e-06  ====
Fit parameters:
Oscillatory part: (B, w, phi)=(-1.67e-05, 0.0105, 0.3298), Polynomial part: (-0.00048, -4.2e-08-1.9e-12)
Eccentricity based on fit: -0.000089
Suggested updates based on fit:
(dOmega, dadot) = (-0.0000001505, +8.79665e-07)
Updated Xcts values based on fit:
(Omega, adot) = ( 0.0124524521, -1.86893317e-05)
```

In the Xcts input file, replace the angular velocity and expansion to the updated values:
```   
Background:
  Binary:
    ...
    AngularVelocity: 0.0124524521
    Expansion: -1.86893317e-05
    ...
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/EccentricityControl/EccentricityControl.py
+++ b/support/Pipelines/EccentricityControl/EccentricityControl.py
@@ -1,0 +1,617 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+import os
+import sys
+
+import click
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+import rich
+import scipy
+from scipy import io, optimize
+
+from spectre.Visualization.Plot import (
+    apply_stylesheet_command,
+    show_or_save_plot_command,
+)
+from spectre.Visualization.ReadH5 import available_subfiles
+
+logger = logging.getLogger(__name__)
+
+
+# Read .dat files from Reductions data: Inertial centers, Mass, Spin
+def extract_data_from_file(h5_file, subfile_name, functions, x_axis):
+    """Extract data from '.dat' datasets in H5 files"""
+
+    with h5py.File(h5_file, "r") as h5file:
+        # Print available subfiles and exit
+        if subfile_name is None:
+            import rich.columns
+
+            rich.print(
+                rich.columns.Columns(
+                    available_subfiles(h5file, extension=".dat")
+                )
+            )
+            return
+
+        # Open subfile
+        if not subfile_name.endswith(".dat"):
+            subfile_name += ".dat"
+        dat_file = h5file.get(subfile_name)
+        if dat_file is None:
+            raise click.UsageError(
+                f"Unable to open dat file '{subfile_name}'. Available "
+                f"files are:\n {available_subfiles(h5file, extension='.dat')}"
+            )
+
+        # Read legend from subfile
+        legend = list(dat_file.attrs["Legend"])
+
+        # Select x-axis
+        if x_axis is None:
+            x_axis = legend[0]
+        elif x_axis not in legend:
+            raise click.UsageError(
+                f"Unknown x-axis '{x_axis}'. Available columns are: {legend}"
+            )
+
+        num_obs = len(dat_file[:, legend.index(x_axis)])
+        out_table = np.reshape(dat_file[:, legend.index(x_axis)], (num_obs, 1))
+
+        # Assemble in table
+        for function, label in functions:
+            if function not in legend:
+                raise click.UsageError(
+                    f"Unknown function '{function}'. "
+                    f"Available functions are: {legend}"
+                )
+
+            out_table = np.append(
+                out_table,
+                np.reshape(dat_file[:, legend.index(function)], (num_obs, 1)),
+                axis=1,
+            )
+
+        return out_table
+
+
+# Compute separation norm
+def compute_separation(h5_file, subfile_name_aha, subfile_name_ahb):
+    """Compute coordinate separation"""
+
+    functions = [
+        ["InertialCenter_x", "x"],
+        ["InertialCenter_y", "y"],
+        ["InertialCenter_z", "z"],
+    ]
+    x_axis = "Time"
+
+    # Extract data
+    ObjectA_centers = extract_data_from_file(
+        h5_file=h5_file,
+        subfile_name=subfile_name_aha,
+        functions=functions,
+        x_axis=x_axis,
+    )
+
+    ObjectB_centers = extract_data_from_file(
+        h5_file=h5_file,
+        subfile_name=subfile_name_ahb,
+        functions=functions,
+        x_axis=x_axis,
+    )
+
+    if (
+        subfile_name_aha is None
+        or subfile_name_ahb is None
+        or subfile_name_aha == subfile_name_ahb
+    ):
+        raise click.UsageError(
+            f"Dat files '{subfile_name_aha}' and '{subfile_name_aha}' are the"
+            " same or at least one of them is missing. Choose files for"
+            " different objects. Available files are:\n"
+            f" {available_subfiles(h5_file, extension='.dat')}"
+        )
+
+    # Compute separation
+    num_obs = len(ObjectA_centers[:, 0])
+
+    if len(ObjectB_centers[:, 0]) < num_obs:
+        num_obs = len(ObjectB_centers[:, 0])
+
+    # Separation vector
+    separation_vec = (
+        ObjectA_centers[:num_obs, 1:] - ObjectB_centers[:num_obs, 1:]
+    )
+
+    # Compute separation norm
+    separation_norm = np.zeros((num_obs, 2))
+    separation_norm[:, 0] = ObjectA_centers[:, 0]
+    separation_norm[:, 1] = np.linalg.norm(separation_vec, axis=1)
+
+    return separation_norm
+
+
+def compute_time_derivative_of_separation_in_window(data, tmin=None, tmax=None):
+    """Compute time derivative of separation on time window"""
+    traw = data[:, 0]
+    sraw = data[:, 1]
+
+    # Compute separation derivative
+    dsdtraw = (sraw[2:] - sraw[0:-2]) / (traw[2:] - traw[0:-2])
+
+    trawcut = traw[1:-1]
+
+    # Apply time window: select values in [tmin, tmax]
+    if tmin == None and tmax == None:
+        if traw[-1] < 200:
+            which_indices = trawcut > 20
+        else:
+            which_indices = trawcut > 60
+    elif tmax == None:
+        which_indices = trawcut > tmin
+    else:
+        which_indices = np.logical_and(trawcut > tmin, trawcut < tmax)
+
+    dsdt = dsdtraw[which_indices]
+    t = trawcut[which_indices]
+
+    return t, dsdt
+
+
+def fit_model(x, y, model):
+    """Fit coordinate separation"""
+    F = model["function"]
+    inparams = model["initial guess"]
+
+    errfunc = lambda p, x, y: F(p, x) - y
+    p, success = optimize.leastsq(errfunc, inparams[:], args=(x, y))
+
+    # Compute rms error of fit
+    e2 = (errfunc(p, x, y)) ** 2
+    rms = np.sqrt(sum(e2) / np.size(e2))
+
+    return dict([("parameters", p), ("rms", rms), ("success", success)])
+
+
+def compute_coord_sep_updates(
+    x, y, model, initial_separation, initial_xcts_values=None
+):
+    """Compute updates for eccentricity control"""
+    fit_results = fit_model(x=x, y=y, model=model)
+
+    amplitude, omega, phase = fit_results["parameters"][:3]
+
+    # Compute updates for Omega and expansion and compute eccentricity
+    dOmg = amplitude / 2.0 / initial_separation * np.sin(phase)
+    dadot = -amplitude / initial_separation * np.cos(phase)
+
+    # The amplitude could be negative due to degeneracy with phase shifts
+    # of pi/2
+    # Make sure eccentricity estimate is positive
+    eccentricity = np.abs(amplitude / initial_separation / omega)
+
+    fit_results["eccentricity"] = eccentricity
+    fit_results["xcts updates"] = dict(
+        [("omega update", dOmg), ("expansion update", dadot)]
+    )
+
+    # Update xcts parameters if given
+    if initial_xcts_values is not None:
+        xcts_omega, xcts_expansion = initial_xcts_values
+        fit_results["previous xcts values"] = dict(
+            [("omega", xcts_omega), ("expansion", xcts_expansion)]
+        )
+        updated_xcts_omega = xcts_omega + dOmg
+        updated_xcts_expansion = xcts_expansion + dadot
+        fit_results["updated xcts values"] = dict(
+            [
+                ("omega", updated_xcts_omega),
+                ("expansion", updated_xcts_expansion),
+            ]
+        )
+
+    return fit_results
+
+
+def coordinate_separation_eccentricity_control_digest(
+    h5_file, x, y, data, functions, output=None
+):
+    """Print and output for eccentricity control"""
+
+    if output is not None:
+        traw = data[:, 0]
+        sraw = data[:, 1]
+        # Plot coordinate separation
+        fig, axes = plt.subplots(2, 2)
+        ((ax1, ax2), (ax3, ax4)) = axes
+        fig.suptitle(
+            h5_file,
+            color="b",
+            size="large",
+        )
+        ax2.plot(traw, sraw, "k", label="s", linewidth=2)
+        ax2.set_title("coordinate separation " + r"$ D $")
+
+        # Plot derivative of coordinate separation
+        ax1.plot(x, y, "k", label=r"$ dD/dt $", linewidth=2)
+        ax1.set_title(r"$ dD/dt $")
+
+        ax4.set_axis_off()
+
+    logger.info("Eccentricity control summary")
+
+    for name, func in functions.items():
+        expression = func["label"]
+        rms = func["fit result"]["rms"]
+
+        logger.info(
+            f"==== Function fitted to dD/dt: {expression:30s},  rms ="
+            f" {rms:4.3g}  ===="
+        )
+
+        style = "--"
+        F = func["function"]
+        eccentricity = func["fit result"]["eccentricity"]
+
+        # Print fit parameters
+        p = func["fit result"]["parameters"]
+        logger.info("Fit parameters:")
+        if np.size(p) == 3:
+            logger.info(
+                f"Oscillatory part: (B, w)=({p[0]:4.3g}, {p[1]:7.4f}), ",
+                f"Polynomial part: ({p[2]:4.2g})",
+            )
+        else:
+            logger.info(
+                f"Oscillatory part: (B, w, phi)=({p[0]:4.3g}, {p[1]:6.4f},"
+                f" {p[2]:6.4f}), "
+            )
+            if np.size(p) >= 4:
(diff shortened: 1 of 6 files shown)


support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
+#!/usr/bin/env python
```
nilsvu: I suggest to place this in `support/Pipelines/EccentricityControl.py`.

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+
+        # Select x-axis
+        if x_axis is None:
+            x_axis = legend[0]
+        elif x_axis not in legend:
+            raise click.UsageError(
+                f"Unknown x-axis '{x_axis}'. Available columns are: {legend}"
+            )
+
+        num_obs = dat_file[:, legend.index(x_axis)].size
```
nilsvu: `len(dat_file)` should suffice

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+        h5_file, subfile_name, functions, x_axis
+    )
+
+    subfile_name = "/ApparentHorizons/ControlSystemAhB_Centers.dat"
+    ObjectB_centers = extract_data_from_file(
+        h5_file, subfile_name, functions, x_axis
+    )
+
+    # Compute separation
+    num_obs = ObjectA_centers[:, 0].size
```
nilsvu: Below you take the difference of the two arrays, so you'll get an error if their length differs. Just assert that their lengths are equal?

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+    # Separation vector
+    separation_vec = ObjectA_centers[:, 1:] - ObjectB_centers[:, 1:]
+
+    # Compute separation norm
+    separation_norm = np.zeros((num_obs, 2))
+    separation_norm[:, 0] = ObjectA_centers[:, 0]
+    for i in range(0, num_obs):
+        separation_norm[i, 1] = np.sqrt(
+            np.dot(separation_vec[i], separation_vec[i])
+        )
```
nilsvu:
```py
separation_norm[:, 1] = np.linalg.norm(separation_vec, axis=1)
```

This avoids the slow Python loop.

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+def fit(x, y, F, inparams, name, style, sinitial):
+    """Fit and plot coordinate separation"""
+    errfunc = lambda p, x, y: F(p, x) - y
+    p, success = optimize.leastsq(errfunc, inparams[:], args=(x, y))
+
+    # Compute rms error of fit
+    e2 = (errfunc(p, x, y)) ** 2
+    np.shape(e2)
```
nilsvu: delete line

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+import rich
+import scipy
+from scipy import io, optimize
+
+from spectre.Visualization.ReadH5 import available_subfiles
+
+# logger = logging.getLogger(__name__)
```
nilsvu: Use this `logger` to replace `print`s please. Use different levels (debug, info, etc). That gives us control over how much output we want when calling the script. More info: https://docs.python.org/3/howto/logging.html#

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+# @click.command(name="dat")
+# @click.argument(
+#     "h5_file",
+#     type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+# )
+
+
+# Read .dat files from Reductions data: Inertial centers, Mass, Spin
+def extract_data_from_file(h5_file, subfile_name, functions, x_axis):
+    """Extract data from '.dat' datasets in H5 files"""
```
nilsvu: Add more docs to all of these functions please. In particular, the docs should describe each argument and the return value. It can also be helpful to add type annotations to the arguments, like `h5_file: str`.

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+    return p
+
+
+def ecc_remove(h5_file, tmin):
+    """Extract data from '.dat' datasets in H5 files"""
+    data = compute_separation(h5_file)
+    traw = data[:, 0]
+    sraw = data[:, 1]
+
+    dsdtraw = (sraw[2:] - sraw[0:-2]) / (traw[2:] - traw[0:-2])
```
nilsvu: Why do you offset this by 2 and not just 1?

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+
+    trawcut = traw[1:-1]
+
+    if tmin == None:
```
nilsvu: Could you add some comments as to the reasoning here?

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+    errfunc = lambda p, x, y: F(p, x) - y
+    p, success = optimize.leastsq(errfunc, inparams[:], args=(x, y))
+
+    # Compute rms error of fit
+    e2 = (errfunc(p, x, y)) ** 2
+    np.shape(e2)
+    rms = np.sqrt(sum(e2) / np.size(e2))
+
+    # Compute updates & eccentricity
+    dOmg = p[0] / 2.0 / sinitial * np.sin(p[2])
```
nilsvu: Could you add some comments where these Jacobians come from? It would also be helpful to name the parameters: `amplitude, omega, offset = p` or something like that.

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+    )
+    plt.legend(loc=(1.1, -1.3))
+
+    # plot residual
+    plt.subplot(2, 2, 3)
+    plt.plot(x, errfunc(p, x, y), style, label=name)
+    plt.title("residual")
+
+    if np.size(p) == 3:
+        print("(B,w)=(%4.3g, %7.4f)," % (p[0], p[1]), "poly=(%4.2g)" % p[2])
```
nilsvu: Use f-strings everywhere, like `f"(B, w) = ({p[0]:4.3g}, {p[1]:7.4f})"`, or `.format()`.

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+            p[3],
+            p[4],
+            0,
+        ],
+        "B*cos(w*t+phi)+quadratic",
+        "-",
+        sraw[0],
+    )
+
+    plt.savefig("EccRemoval.pdf")
```
nilsvu: Take the plot filename as an argument to the function. Only plot and save if a plot filename is provided.

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+
+
+if __name__ == "__main__":
+    h5file = "GhBinaryBlackHoleReductionData.h5"
+    tmin = 100
+    ecc_remove(h5file, tmin)  # help_option_names=["-h", "--help"])
```
nilsvu: Please register this with the CLI so we can just do `spectre ecc-reduce H5_FILE --plot-filename PLOT_FILE`. To do this, follow the guidelines here: https://spectre-code.org/spectre_writing_python_bindings.html#autotoc_md149.

support/Python/EccentricityRemoval.py
```diff
@@ -0,0 +1,270 @@
...
+    plt.title("coordinate separation " + r"$D$")
+
+    plt.subplot(2, 2, 1)
+    plt.plot(t, dsdt, "k", label=r"$ dD/dt $", linewidth=2)
+    plt.title(r"$ dD/dt $")
+
+    # ==== RESTRICTED FIT ====
+    p = fit(
+        t,
+        dsdt,
```
nilsvu: Use named arguments when calling a function whenever it helps with clarity. It would help a lot here.

nilsvu (changes requested):
Thanks for adding this!

AlexCarpenter46:
Tried using this and got confused with the output, it wasn't clear to me that you need to add those corrections to your current parameters. Was wondering if a comment could be added specifying that you need to update your parameters and add the corrections to them or if you could output the corrected omega and adot terms?

support/Python/EccentricityRemoval.py
nilsdeppe: I like this suggestion (also all the others, but I don't know enough about python other than to agree with Nils :D ). I'm hoping as the eccentricity control gets robust in SpEC we can port that and always use the same algorithm or at least CLI/entry point independent of whether our target ecc is 0 or 0.1234.

nilsdeppe (commented):
I don't have anything meaningful really to add, but I want to say thank you for doing this! It won't be long until this is needed (and I assume you already do!) for all our merger simulations. I like Nils's suggestions, especially in terms of documentation and clarity. The SpEC scripts that do things like this were something I always found a bit difficult to follow and don't always have references to papers, so having those would be great!

I'm not "requesting changes" so absolutely no need to ask me to look at this. @nilsvu just merge when you're happy! Thanks for reviewing!

haraldp271:
I assume this is the old SpEC script @guilara and I used a few days ago to start doing ecc removal of this binary in Spectre.   The techniques used here work for non-precessing binaries, but the code has been replaced in SpEC by OmegaDoEccRemoval.py.  
- While this code is useful for the near future, I expect that it gets replaced by more generic and more robust code down the road.  Therefore, I suggest to choose a less all-encompassing name, that makes it easy to replace with a better script.   Perhaps `CoordSepEccReduction.py`.
- It looks like tmin=100 is hardcoded.  I didn't see any tmax, so I assume the script just uses all data.  Both these choices are very simplistic.  In the newer SpEC scripts we have improved on tmin/tmax determination, with more ongoing work by one of @markscheel students.  I think it's ok to use this for the time being, but I strongly recommend to describe these shortcomings clearly in the help-text.

nilsvu:
Thanks! Yes we can merge this with some of my suggestions above and then update it with the new techniques in SpEC very soon. I'd put it in `support/Pipelines/EccentricityControl.py` and update it later.

geoffrey4444:
Can you please output the recommended new parameters and not just differences from the current iteration?

haraldp271:
That requires the script to know the current params (which is what the fancier scripts do).  In this most basic version this wasn't included.

guilara:
I've been working on the corrections. I could include the omega from the fit plus the correction @geoffrey4444  but I'm not sure that will be good enough @haraldp271 . I'm also including tmax specification by the user.

guilara:
@geoffrey4444 @haraldp271 @AlexCarpenter46 I've added an option for the user to specify the parameters they are using in the input Xcts file so that the script returns the updated parameters and not just the corrections to be applied to them.

support/Pipelines/EccentricityControl/EccentricityControl.py
```diff
@@ -0,0 +1,419 @@
...
+        print("Updated Xcts values based on fit:")
+        print(
+            f"(Omega, adot) = ({(xcts_omega + dOmg):13.10f},"
+            f" {(xcts_expansion + dadot):13.10g})"
+        )
+
+    return p
+
+
+@click.command(name="ecc-remove")
```
nilsvu: This name should match the name you use in `__main__.py`. I think `ecc-remove` or `ecc-reduce` or `ecc-control` or something like that is all fine.

support/Pipelines/EccentricityControl/EccentricityControl.py
```diff
@@ -0,0 +1,419 @@
...
+)
+@click.option(
+    "--expansion-from-xcts",
+    type=float,
+    nargs=1,
+    help="Value of the expansion velocity (adot) used in the Xcts file.",
+)
+@apply_stylesheet_command()
+@show_or_save_plot_command()
+def eccentricity_control_command(
```
nilsvu, line 563: Can you split this into a function that can be called from outside the CLI command, so we can use it in the automated ecc control? The function should (1) return the delta updates, (2) not print anything unless through `logging`, (3) not plot anything unless given a plot output filename or given a matplotlib `fig` or `ax`.

I'd also be happy merging this PR as-is and factor out the function later, if you prefer. I just didn't want to merge it and then refactor your code without checking in with you :)

guilara: I can give it a try. I could easily do (2) and (3), I think. For (1) maybe I could collect the updates from all models and output that --since there is no single update here.

nilsvu: Sounds good. It's fine to return any or all of the possible updates, then we can use it already to build the automation pipeline and work on improving the fits later or at the same time.

guilara:
@nilsvu I've addressed all of the comments.

nilsvu (commented):
Looks good! Only a test is missing. It's fine to make the test as minimal as possible, just to check that this code runs without error. You can write an H5 file in the test with some fake data, see `tests/tools/Test_Status.py` for an example. You can also squash your commits.

AzerAfram:
I ran the latest version of `EccentricityControl.py` with a merged BBH on SpECTRE and got negative, sporadic eccentricities (also, the changes for omega and expansion were sporadic especially when comparing H1 with H3 and H2 with H4). This was not just with one BBH or interval but for every input that I tried. This may be because `omega/omega0` gives odd values (just as in SpEC, models with these values should not be allowed to pass: `OmegaDotEccRemoval.py` line 343). This is not an issue with the general algorithm itself though as `OmegaDotEccRemoval.py` refactored to take in SpECTRE data outputs plausible values for eccentricity, omega, and expansion across all four models with the same input (data and intervals) . I think the issue is in some part due to the current implementation of `fit_model()` which looks similar to `old_fit()` in `OmegaDotEccRemoval.py` but `OmegaDotEccRemoval.py` no longer uses `old_fit()` by default (and when it is forced, errors for out of bound `omega/omega0` values are triggered). Also we need to decide how to implement a function similar to `GetVarsFromSpinData()` in `OmegaDotEccRemoval.py` since this function requires the `dimensionFullInertialSpins` in the x, y and z axis which are stored by default using SpEC but are not stored whatsoever using SpECTRE (only the magnitude of `dimensionFullInertialSpins` can be obtained). `dimensionFullInertialSpins` cannot simply be zeroed out or ignored as this causes significant errors in the output.

guilara:
@AzerAfram Yes, Indeed the approach here uses the old_fit of SpEC which is not the best. And we only got it to work with a lot of supervision, meaning, only accepting updates from models that look good in the residual plots (which may be only a few and only after adjusting the time bounds manually a few times). I assume the values from the newer fit function are better and more consistent because there is an algorithm to select good time bounds.

@nilsvu I've been on holiday. I'll get back to this later in the week when I'm back in the office.

nilsvu (commented):
Hi @guilara I would like to merge this because @vtommasini and I have tested it a bit and it gets us to reasonably small eccentricities, with the caveats that you and @AzerAfram described. Could you rebase it? If you manage to add a small test as well that would be nice (maybe write sinusoids to a file and run the script over it, similar to Test_PlotTrajectories.py). Then at least we test on CI that there's no obvious problem with the code. No need to test any details, because as we're all aware we will replace this implementation either by porting over the current SpEC implementation, or calling SpEC once it is refactored according to #5937, or use @AzerAfram's #5966 to wrap the SpEC scripts.

guilara:
> Hi @guilara I would like to merge this because @vtommasini and I have tested it a bit and it gets us to reasonably small eccentricities, with the caveats that you and @AzerAfram described. Could you rebase it? If you manage to add a small test as well that would be nice (maybe write sinusoids to a file and run the script over it, similar to Test_PlotTrajectories.py). Then at least we test on CI that there's no obvious problem with the code. No need to test any details, because as we're all aware we will replace this implementation either by porting over the current SpEC implementation, or calling SpEC once it is refactored according to #5937, or use @AzerAfram's #5966 to wrap the SpEC scripts.

OK, I'll rebase it and try to write the test

guilara:
@nilsvu I've rebased and added a test

tests/support/Pipelines/EccentricityControl/CMakeLists.txt
```diff
@@ -2,13 +2,21 @@
 # See LICENSE.txt for details.
 
 if (SpEC_FOUND)
+  spectre_add_python_bindings_test(
+    "support.Pipelines.EccentricityControl.EccentricityControl"
+    Test_InitialOrbitalParameters.py
```
nilsvu: Fix this name

tests/support/Pipelines/EccentricityControl/Test_EccentricityControl.py
```diff
@@ -0,0 +1,171 @@
...
+            # Populate dataset with AhB data
+            for data_point in AhB_data:
+                dataset_AhB.append(data_point)
+            # Close dataset for AhB
+            h5_file.close_current_object()
+        logging.info(
+            f"Successfully created and populated HDF5 file: {self.h5_filename}"
+        )
+
+    def test_cli(self):
```
nilsvu, line 150: You have this test with the mock data set up, but don't actually check if the returned eccentricity and param updates make sense. Could you add a test function that invokes the non-CLI function and check the return values are reasonable?

tests/support/Pipelines/EccentricityControl/CMakeLists.txt
```diff
@@ -2,13 +2,21 @@
 # See LICENSE.txt for details.
 
 if (SpEC_FOUND)
+  spectre_add_python_bindings_test(
```
nilsvu, line 12: The test doesn't depend on SpEC, so move it out of the `if` please.

nilsvu (changes requested):

nilsvu (dismissed):
LGTM please squash everything into one commit 👍

guilara:
> LGTM please squash everything into one commit 👍

Good! Done. Thanks for the feedback

nilsvu (approved):
Thanks! Let's merge this and refer to #5937 for the next steps.

PR #5896, opened 2024-03-29
BinaryCompactObject domain: support initial Kerr horizon shape

nilsvu:
## Proposed changes

Allows to deform excision surfaces in the initial data to conform to spinning Kerr horizons.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/Creators/BinaryCompactObjectHelpers.cpp
```diff
@@ -420,13 +464,7 @@ TimeDependentMapOptions<IsCylindrical>::grid_to_inertial_map(
...
       return std::make_unique<detail::gi_map<Rotation>>(rotation_map_.value());
     } else {
-      ERROR(
-          "Requesting grid to inertial map without a distorted frame and "
-          "without a Rotation or Expansion map for object "
-          << Object
-          << ". This means there are no time dependent maps. If you don't want "
-             "time dependent maps, specify 'None' for TimeDependentMapOptions. "
-             "Otherwise specify at least one time dependent map.");
+      return nullptr;
```
knelli2, line 467: Why this change?

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
```diff
@@ -289,7 +301,9 @@ struct TimeDependentMapOptions {
    */
   std::unordered_map<std::string,
                      std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
-  create_functions_of_time(const std::unordered_map<std::string, double>&
+  create_functions_of_time(const std::optional<double>& inner_radius_A,
+                           const std::optional<double>& inner_radius_B,
```
knelli2: I believe this information is already set from the `build_maps` function, correct? Why pass it in again here?

knelli2 (commented):
Couple questions

src/Domain/Creators/BinaryCompactObjectHelpers.cpp
nilsvu: Expansion and rotation maps are both null, only a shape map is set. That triggered this error. So I handle the `nullptr` returned here for the grid-to-inertial maps in `create_domain`.

src/Domain/Creators/BinaryCompactObjectHelpers.hpp
nilsvu: The information isn't stored AFAICT.

knelli2: Should we store it? I think I'd prefer that over changing this function signature since right now the signature of `create_functions_of_time` is the same as `functions_of_time` that the domain creators have

nilsvu: Ok pushed a fixup

knelli2 (approved):
Looks good. You can squash

nilsvu:
Done. Thanks for reviewing!

PR #5914, opened 2024-04-16
Fix race condition when updating FoTs

knelli2:
This occurs when updates are received out of order because of asynchronous communication.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2:
Also pinging @nikwit just so you are aware of this (no need to review unless you really want to)

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -141,6 +154,31 @@ void PiecewisePolynomial<MaxDeriv>::update(const double time_of_update,
...
+
+  // This update time came from the backlog. Remove it and return so we don't
+  // keep recursing.
+  if (update_backlog_.contains(time_of_update)) {
+    update_backlog_.erase(time_of_update);
+    return;
+  }
+
+  // We will only get here in the "highest level" of the recursion
+  for (auto it = update_backlog_.begin(); it != update_backlog_.end();) {
```
nikwit: would a simple while loop be easier to read, rather than recursion?

```
while (not update_backlog_.empty() and
         update_backlog_.begin()->first == time_bounds()[1]) {
  const double exp_time = time_bounds()[1];
  store_entry(exp_time, std::move(update_backlog_[exp_time].first), update_backlog_[exp_time].second);
  update_backlog_.erase(exp_time);
}
```
This loop also uses the fact that we know the backlog is ordered so if the first item cannot be applied, the following items cannot be either.
This way `update` can do all checks and then either (add it to the backlog and return) or (store entry and apply the backlog which has already been checked).

tests/Unit/Domain/FunctionsOfTime/Test_QuaternionFunctionOfTime.cpp
```diff
@@ -31,7 +31,70 @@ void test_serialization_versioning() {
...
+void test_out_of_order_update() {
+  using QuatFoT = domain::FunctionsOfTime::QuaternionFunctionOfTime<2>;
+  register_classes_with_charm<QuatFoT>();
+  const DataVector quat_dv{1.0, 0.0, 0.0, 0.0};
+  const DataVector zero_dv{3, 0.0};
+  const std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime> func(
+      std::make_unique<QuatFoT>(
+          0.0, std::array<DataVector, 1>{quat_dv},
+          std::array<DataVector, 3>{zero_dv, zero_dv, zero_dv}, 0.5));
+
```
nikwit, line 46: can you add another update `func->update(2.5, zero_dv, 3.0)` here? This way we ensure that the `update_backlog` can store multiple entries that are given out of order. This should work because the `std::map` is guaranteed to be ordered but it's not immediately obvious.

nikwit (changes requested):

wthrowe:

knelli2:
@nikwit I implemented your suggestions in a fixup (much better than what I had)

wthrowe:
Other than NOLINTing the clang-tidy false positive, looks good to me.

nikwit (approved):
looks good, you can squash that in :)

wthrowe (approved):

PR #5919, opened 2024-04-18
Bind horizon finder in Python

nilsvu:
## Proposed changes

This will allow finding horizons in output data in Python. For context, we can use this to find horizons on the initial data slice, write out the Ylm coefficients, and use them to initialize the inspiral domain.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/Exporter/Python/__init__.py
+++ b/src/IO/Exporter/Python/__init__.py
@@ -1,4 +1,49 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
+from typing import Sequence, Type
+
 from ._Pybindings import *
+
+
+def interpolate_tensors_to_points(
+    *args,
+    tensor_names: Sequence[str],
+    tensor_types: Sequence[Type],
+    **kwargs,
+):
+    """Wrapper around 'interpolate_to_points' for tensors.
+
+    Interpolates volume data to points by calling 'interpolate_to_points'.
+    However, instead of passing a list of tensor components
+    (like 'SpatialMetric_xx') you can pass a list of tensor names (like
+    'SpatialMetric') and their corresponding types (like
+    'tnsr.ii[DataVector, 3]') and this function will assemble the tensors
+    and return them as a list.
+
+    Note: It would be nice to move this function into C++, but that's not
+    trivial because the tensor types are not known at compile time. This can be
+    revisited when we have Python support for Variables.
+    """
+    from spectre.DataStructures import DataVector
+
+    tensors = {
+        tensor_name: tensor_type()
+        for tensor_name, tensor_type in zip(tensor_names, tensor_types)
+    }
+    tensor_components = []
+    for name, tensor in tensors.items():
+        tensor_components += [
+            name + tensor.component_suffix(i) for i in range(tensor.size)
+        ]
+    data = interpolate_to_points(
+        *args,
+        **kwargs,
+        tensor_components=tensor_components,
+    )
+    j = 0
+    for tensor in tensors.values():
+        for i in range(tensor.size):
+            tensor[i] = DataVector(data[j])
+            j += 1
+    return tensors.values()
--- a/src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
@@ -5,12 +5,15 @@
 
 #include <array>
 #include <cstddef>
+#include <pybind11/operators.h>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
 
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/ModalVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/IO/FillYlmLegendAndData.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
 
 namespace py = pybind11;
@@ -57,5 +60,25 @@ void bind_strahlkorper(pybind11::module& m) {  // NOLINT
       .def(py::self == py::self)
       // NOLINTNEXTLINE(misc-redundant-expression)
       .def(py::self != py::self);
+  m.def(
+      "ylm_legend_and_data",
+      [](const ylm::Strahlkorper<Frame::Inertial>& strahlkorper,
+         const double time, const size_t max_l)
+          -> std::pair<std::vector<std::string>, std::vector<double>> {
+        std::vector<std::string> legend{};
+        std::vector<double> data{};
+        ylm::fill_ylm_legend_and_data(make_not_null(&legend),
+                                      make_not_null(&data), strahlkorper, time,
+                                      max_l);
+        return std::make_pair(legend, data);
+      },
+      py::arg("strahlkorper"), py::arg("time"), py::arg("max_l"));
+  m.def("read_surface_ylm", &ylm::read_surface_ylm<Frame::Inertial>,
+        py::arg("file_name"), py::arg("surface_subfile_name"),
+        py::arg("requested_number_of_times_from_end"));
+  m.def("read_surface_ylm_single_time",
+        &ylm::read_surface_ylm_single_time<Frame::Inertial>,
+        py::arg("file_name"), py::arg("surface_subfile_name"), py::arg("time"),
+        py::arg("relative_epsilon"));
 }
 }  // namespace ylm::py_bindings
--- a/src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
@@ -1,0 +1,61 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.hpp"
+
+#include <array>
+#include <cstddef>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/ModalVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+
+namespace py = pybind11;
+
+namespace ylm::py_bindings {
+void bind_strahlkorper(pybind11::module& m) {  // NOLINT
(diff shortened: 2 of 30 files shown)


src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
```diff
@@ -35,5 +36,19 @@ void bind_strahlkorper(pybind11::module& m) {  // NOLINT
                              &ylm::Strahlkorper<Frame::Inertial>::l_max)
       .def_property_readonly("m_max",
                              &ylm::Strahlkorper<Frame::Inertial>::m_max);
+  m.def(
+      "ylm_legend_and_data",
```
knelli2, line 64: #5921 should help with the dependencies of this

nilsvu: Thanks, rebased in this and added bindings for both reading and writing

src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
```diff
@@ -0,0 +1,44 @@
...
+           py::arg("l_max"), py::arg("m_max"), py::arg("radius"),
+           py::arg("center"))
+      .def(py::init<size_t, size_t, const DataVector&, std::array<double, 3>>(),
+           py::arg("l_max"), py::arg("m_max"),
+           py::arg("radius_at_collocation_points"), py::arg("center"))
+      .def(
+          py::init<size_t, size_t, const ylm::Strahlkorper<Frame::Inertial>&>(),
+          py::arg("l_max"), py::arg("m_max"), py::arg("another_strahlkorper"))
+      .def(
+          py::init<size_t, size_t, const ModalVector&, std::array<double, 3>>(),
```
knelli2, line 34: If both `DataVector` and `ModalVector` can be represented as numpy arrays, how does it know which constructor to use if a numpy array is passed? Or do you have to specify the explicit data type?

src/NumericalAlgorithms/SphericalHarmonics/Python/StrahlkorperFunctions.cpp
```diff
@@ -0,0 +1,27 @@
...
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp"
+
+namespace py = pybind11;
+
+namespace ylm::py_bindings {
+void bind_strahlkorper_functions(pybind11::module& m) {  // NOLINT
+  m.def(
+      "cartesian_coords",
+      [](const ylm::Strahlkorper<Frame::Inertial>& strahlkorper) {
```
knelli2: [optional] I've actually found myself wanting this exact function in c++ several times (where you only pass a strahlkorper and get back the coords). Add it in c++ as well then bind that?

src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
knelli2: Are the following functions maybe better placed in the `StrahlkorperFunctions.cpp` bindings file since they do things with Strahlkorper data?

src/IO/Exporter/Python/__init__.py
```diff
@@ -1,4 +1,49 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
+from typing import Sequence, Type
+
 from ._Pybindings import *
+
+
+def interpolate_tensors_to_points(
```
knelli2, line 9: Why not put this next to `interpolate_to_points` so it's exposed?

support/Pipelines/Bbh/FindHorizon.py
```diff
@@ -0,0 +1,231 @@
...
+    obs_time: float,
+    l_max: int,
+    initial_radius: float,
+    center: Sequence[float],
+    output: Optional[str] = None,
+    output_coeffs_subfile: Optional[str] = None,
+    output_coords_subfile: Optional[str] = None,
+    tensor_names: Optional[Sequence[str]] = None,
+):
+    """Find an apparent horizon in volume data.
```
knelli2, line 63: Please make a note that both the horizon and the volume data must be in the Inertial frame. Should we add support for other frames? We could bind (in a future PR) the transformation of the strahlkorper coords to a different frame given a domain.

support/Pipelines/Bbh/FindHorizon.py
```diff
@@ -0,0 +1,231 @@
...
+
+    Arguments:
+      h5_files: List of H5 files containing volume data or glob pattern.
+      subfile_name: Name of the volume data subfile in the 'h5_files'.
+      obs_id: Observation ID in the volume data.
+      obs_time: Time of the observation.
+      l_max: Maximum l-mode for the horizon search.
+      initial_radius: Initial coordinate radius of the horizon.
+      center: Coordinate center of the horizon.
```
knelli2: This is the expansion center of the horizon, not a physical center. Should note that somehow

support/Pipelines/Bbh/FindHorizon.py
```diff
@@ -0,0 +1,231 @@
...
+def find_horizon(
+    h5_files: Union[str, Sequence[str]],
+    subfile_name: str,
+    obs_id: int,
+    obs_time: float,
+    l_max: int,
+    initial_radius: float,
+    center: Sequence[float],
+    output: Optional[str] = None,
+    output_coeffs_subfile: Optional[str] = None,
```
knelli2, line 58: [nitpick] I think in almost all the c++ code we use the shortened `coefs` instead of `coeffs`. I realize that a single "f" isn't how it's spelled, but for consistency

support/Pipelines/Bbh/FindHorizon.py
```diff
@@ -0,0 +1,231 @@
...
+        ]
+    fast_flow = FastFlow(
+        FlowType.Fast,
+        alpha=1.0,
+        beta=0.5,
+        abs_tol=1e-12,
+        truncation_tol=0.01,
+        divergence_tol=1.2,
+        divergence_iter=5,
+        max_its=100,
```
knelli2: Should we allow any of these to be specifiable? These work for relatively spherical horizons, but not sure how they'll do for more distorted ones

support/Pipelines/Bbh/FindHorizon.py
```diff
@@ -0,0 +1,231 @@
...
+@click.option(
+    "--center",
+    "-C",
+    nargs=3,
+    type=float,
+    required=True,
+    help="Coordinate center of the horizon.",
+)
+@click.option(
+    "--output",
```
knelli2, line 208: Should this be required for the CLI? Not sure what the point of omitting it would be here...nothing would happen

tests/support/Pipelines/Bbh/Test_FindHorizon.py
```diff
@@ -0,0 +1,42 @@
...
+    def test_find_horizon(self):
+        obs_ids, obs_times = list_observations(
+            open_volfiles([self.h5_filename], "/element_data")
+        )
+        # Can't test more than that this code runs until it errors until we have
+        # a way to generate data with an apparent horizon.
+        with self.assertRaisesRegex(
+            RuntimeError, "Failed to open dataset 'InverseSpatialMetric_xx'"
+        ):
+            find_horizon(
```
knelli2: For some reason gcc isn't liking this.

knelli2 (changes requested):

src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
nilsvu: They do Strahlkorper IO, so I think this is a good place for them.

src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
nilsvu: Not sure, maybe it selects the first constructor that works. You can pass an explicit DataVector or ModalVector though.

src/IO/Exporter/Python/__init__.py
nilsvu: This is next to `interpolate_to_points`, which is a binding brought in through the `from ._Pybindings import *` in this file. Both can be imported as `from spectre.IO.Exporter import interpolate_[tensors_]_to_points`.

support/Pipelines/Bbh/FindHorizon.py
nilsvu: We can add support for other frames if needed. I just didn't know if that would be necessary.

support/Pipelines/Bbh/FindHorizon.py
nilsvu: Ugh ok, a quick search shows 589 occurrences of `coefs` and 460 occurrences of `coeffs`, so pretty evenly split 😅 So I'll leave it like this.

support/Pipelines/Bbh/FindHorizon.py
nilsvu: Yea we can expose any or all of these in the CLI, I just didn't know which were necessary. We can expose them when we need them, unless you already know which ones are necessary. I added the option to pass in a `FastFlow` object to the Python function, so programmatically you can set all these parameters now.

support/Pipelines/Bbh/FindHorizon.py
nilsvu: You get a logging message telling you that a horizon was found and some info about it. I found that useful to check quickly for a horizon in some initial data without having to write a file.

src/NumericalAlgorithms/SphericalHarmonics/Python/StrahlkorperFunctions.cpp
nilsvu: ok done

tests/support/Pipelines/Bbh/Test_FindHorizon.py
nilsvu: tried another way of calling this

nilsvu: didn't work. I just disabled this test for gcc (see comment in cmake)

src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
knelli2: I know you *can* but I'm worried users won't know right away to do that and there will be some weird results by just assuming a numpy array will work and it using the wrong constructor. I don't really have a change request but something to keep aware of.

knelli2 (approved):
Ok looks good. You can squash

nilsvu:
Rebased and squashed. Thanks for reviewing!

knelli2 (approved):

PR #5851, opened 2024-03-14
Diffusion scattering mc

ffoucart:
## Proposed changes

Add method for packet diffusion in high scattering regions

### Upgrade instructions

NA

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5903, opened 2024-04-09
Add hydro coupling bookkeeping to emission function in MC

ffoucart:
## Proposed changes

This PR begins the process of keeping track of energy/momentum/lepton number exchange between neutrinos and the fluid. Here, I only consider the "emit" function, as it is already merged.
My current plan is to keep the total energy/momentum/lepton number in the various coupling_* containers, then divide by the coordinate volume at the end of a timestep to get the actual coupling terms used in the GhValenciaDivClean equations.

### Upgrade instructions

None

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5920, opened 2024-04-18
Bookkeeping of hydro-mc coupling in packet evolution

ffoucart:
## Proposed changes

Add computation of coupling between MC packet and fluid variables in EvolvePacketInElement. This is only bookkeeping for now, and not used to feedback on the fluid. We will also need to add contribution from the packets using the diffusion approximation after merging #5851 

### Upgrade instructions

None

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp
+++ b/src/Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp
@@ -31,6 +31,29 @@ void draw_single_packet(
     gsl::not_null<std::mt19937*> random_number_generator);
 }  // namespace detail
 
+
+/// \brief Contribution to the neutrino-matter coupling
+/// terms from evolving a packet by time dt, given
+/// absorption and scattering opacities absorption_opacity,
+/// scattering_opacity.
+///
+/// \details We consider total momentum exchances, not
+/// densities; thus when coupling to the evolution of the
+/// energy/momentum density, division by the spatial volume is
+/// necessary. We calculate the source terms of Eqs (62-64) of
+/// \cite Foucart:2021mcb, with integrals calculated as in
+/// Eqs (6-10) of that manuscript (except that we do not
+/// divide by the coordinate volume V).
+void AddCouplingTermsForPropagation(
+    gsl::not_null<Scalar<DataVector>*> coupling_tilde_tau,
+    gsl::not_null<tnsr::i<DataVector, 3, Frame::Inertial>*>
+        coupling_tilde_s,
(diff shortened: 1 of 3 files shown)


tests/Unit/Evolution/Particles/MonteCarlo/Test_EvolvePackets.cpp
```diff
@@ -94,10 +94,23 @@ SPECTRE_TEST_CASE("Unit.Evolution.Particles.MonteCarloEvolution",
       std::array<DataVector, 2>{{zero_dv, zero_dv}},
       std::array<DataVector, 2>{{zero_dv, zero_dv}}};
 
+  // Set non-zero value that should nearly never lead
+  // to interaction, to get non-zero interaction terms.
```
ffoucart: I'll update the comment after code review -- but in fact, due to the use of a minimum value of 1.e-100 in the RNG when determining the time to the next event, we are explicitly guaranteed that no scattering / absorption will happen.

src/Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp
```diff
@@ -24,12 +24,36 @@ struct Packet;
 
 namespace detail {
 
+const double proton_mass_MeV = 938.272046;
```
kidder: I made a request in #5903 to move this to the Units.hpp file; just want to make sure it doesn't get add here

src/Evolution/Particles/MonteCarlo/TemplatedLocalFunctions.hpp
```diff
@@ -24,12 +24,36 @@ struct Packet;
...
+
 void draw_single_packet(gsl::not_null<double*> time,
                         gsl::not_null<std::array<double, 3>*> coord,
                         gsl::not_null<std::array<double, 3>*> momentum,
                         gsl::not_null<std::mt19937*> random_number_generator);
 }  // namespace detail
 
+
+// Contribution to the neutrino-matter coupling terms from
```
kidder: Make this a doxygen comment with the first sentence as \brief and the rest as \details

kidder (changes requested):
you can squash changes, but will need to wait until #5903 is merged in order to rebase so that you can get the proton mass from the correct place.

ffoucart:
Ok, rebased against develop, including #5903 .
I'm also realizing that I'll need to move the calculation of the coupling for the evolution along geodesics in a different file for #5927 to work... but I can deal with that in the relevant PR.

kidder (approved):

PR #5921, opened 2024-04-18
Add functionality to Ylm IO

knelli2:
## Proposed changes

This is the first of 3 PRs to allow reading horizon coefficients from a file into the shape map coefficients in a domain creator. This PR just factors out a function into a more accessible part of the code and adds the ability to read Ylm coefs at a specific time from a file.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
@@ -30,4 +30,29 @@ template <typename Frame>
 std::vector<ylm::Strahlkorper<Frame>> read_surface_ylm(
     const std::string& file_name, const std::string& surface_subfile_name,
     size_t requested_number_of_times_from_end);
+
+/*!
+ * \brief Similar to `ylm::read_surface_ylm`, this reads in spherical harmonic
+ * data for a surface and constructs a `ylm::Strahlkorper`. However, this
+ * function only does it at a specific time and returns a single
+ * `ylm::Strahlkorper`.
+ *
+ * \note If two times are found within \p epsilon of the \p time, then an error
+ * will occur. Similarly, if no \p time is found within the \p epsilon, then an
+ * error will occur as well.
+ *
+ * \param file_name name of the H5 file containing the surface's spherical
+ * harmonic data
+ * \param surface_subfile_name name of the subfile (with no leading slash nor
+ * the `.dat` extension) within `file_name` that contains the surface's
+ * spherical harmonic data to read in
+ * \param time Time to read the coefficients at.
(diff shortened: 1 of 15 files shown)


tests/Unit/NumericalAlgorithms/SphericalHarmonics/IO/Test_FillYlmLegendAndData.cpp
```diff
@@ -0,0 +1,105 @@
...
+      CHECK(data[i] == expected_coefs[i - 5]);
+    }
+  }
+
+  legend.clear();
+  data.clear();
+  iter.reset();
+
+  // Where max_l is larger than strahlkorper.l_max()
+  {
```
wthrowe, line 73: [optional] My understanding is that the goal of this function is that the output is effectively a prolonged version of the input strahlkorper to the requested l_max.  If so, a simpler test for this part would be to check that `fill_ylm_legend_and_data(strahkorper, larger_l_max)` and `fill_ylm_legend_and_data(Strahlkorper(larger_l_max, larger_l_max, strahkorper), larger_l_max)` give the same result.

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
```diff
@@ -30,4 +30,29 @@ template <typename Frame>
...
+ *
+ * \param file_name name of the H5 file containing the surface's spherical
+ * harmonic data
+ * \param surface_subfile_name name of the subfile (with no leading slash nor
+ * the `.dat` extension) within `file_name` that contains the surface's
+ * spherical harmonic data to read in
+ * \param time Time to read the coefficients at.
+ * \param epsilon How much error is allowed when looking for a specific time.
+ * This is useful so users don't have to know the specific time to machine
+ * precision.
```
wthrowe, line 52: What is the expected use case where the caller doesn't know the exact time?

Mention that `epsilon` is a relative error, possibly by renaming it to indicate that.

wthrowe (changes requested):

nilsvu (dismissed):

src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp
knelli2: Say I've been dumping data every 0.5M and I want to read it it at time 123.5M. However, the actual time of the output is something like 123.499999999999748. The epsilon accounts for this. The rename sounds like a good idea though

knelli2:
@wthrowe pushed fixups

wthrowe:
OK.  Squash and rebase.

wthrowe (approved):

PR #5922, opened 2024-04-18
Factor out Shape options from BCO and Sphere domain

knelli2:
## Proposed changes

This is the second of 3 PRs to allow reading horizon coefficients from a file into the shape map coefficients in a domain creator. This PR factors out the shape map options from the BCO and Sphere domains into a common place since they are shared. In doing so, this also adds the capability of specifying if you want a shape map at runtime in the Sphere domain.

### Upgrade instructions

When using the hardcoded time dependent maps in the Sphere domain, you must now also specify `SizeInitialValues:` in the shape map options:
```yaml
ShapeMap:
  LMax: 10
  InitialValues: Spherical
  SizeInitialValues: Auto  # <-- This is new
```
These are either `Auto` or an array of length 3 for the value and it's two time derivatives

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

~Depends on and includes #5921.~

PR #5923, opened 2024-04-18
Add ability to read horizon coefs from file into shape map

### LINKED ISSUE #4598: Ringdown: initialize size map
# Feature request:

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] Given a BBH with a common apparent horizon, initialize the size map to avoid incoming char speeds when starting the ringdown evolution.
 
### Detailed discussion:
Note: Until spectre can evolve to common horizon, this can be tested by interpolating SpEC data with a common horizon.

### LINKED ISSUE #5541: Read in shape and size coefficients from file to Sphere domain creator
For BBH ringdown, we'll need to start on a Sphere domain with new coefficients for the shape map. We'll want to read these in from an h5 file. We already have a function that can read in a vector of Ylm coefficients from horizon finds which was added in #5416. The function is stored in `src/NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp`. We'll want something similar to this, but slightly different in that we only want to read in one specific time, not a vector. However, I believe much of the implementation in `ReadSurfaceYlm.cpp` can be reused for this purpose.

So the steps for this will be
- Add a new function `read_surface_ylm_single_time` to `ReadSurfaceYlm.hpp`
  - This function will take an h5 file name, a subfile name, the time to read in, and maybe an epsilon factor so the user doesn't have to copy the exact time and there is a little wiggle room.
  - Loop over the available times in the subfile and find the one within epsilon of the input time. Get the row number for this time.
  - Reuse the `read_surface_ylm_row` function in the `cpp` given this row number
- Use this new function in the SphereTimeDependentMaps
  - Add a new variant to the `ShapeMapInitialValues` option in `src/Domain/Creators/SphereTimeDependentMaps.hpp`.
  - This variant will be a struct that takes all the arguments that the `read_surface_ylm_single_time` will need.
  - In the `create_functions_of_time` function of the SphereTimeDepOpts, use these new options and call the `read_surface_ylm_single_time` function.
  - Use the coefficients of the read in Strahlkorper to initialize the coefficients of the shape and size map.

We'll also need support for reading in the time derivative(s) of the shape coefficients. These will probably be stored in a different subfile so the procedure for reading them in will be the same.

### LINKED ISSUE #5669: Modifications to Sphere Domain for Ringdown Evolution
# Feature request:

The Sphere domain requires a few changes so that the evolution can proceed from inspiral/merger to ringdown, including (but not limited to):

- [ ] Reading the ylm coefficients and their time derivatives for the initialization of the Shape map functions of time
- [ ] Initialization of the rotation and scaling maps (using SettleToConstant / SettleToConstantDeriv, as SpEC does) 

These will be added by myself.

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] Detail 1
- [ ] Detail 2
- [ ] Detail 3

knelli2:
## Proposed changes

This is the third and final of 3 PRs to allow reading horizon coefficients from a file into the shape map coefficients in a domain creator.

Fixes #5541
Fixes #5669
Fixes #4598 

### Upgrade instructions

Specify reading in the shape coefficients from a file like so
```yaml
ShapeMap:
  LMax: 8
  SizeInitialValues: Auto
  InitialValues:
    H5Filename: NameOfFile.h5
    SubfileNames:
      - NameOfCoefSubfile
      - NameOfdtCoefSubfile
      - NameOfd2tCoefSubfile
    MatchTime: 49.8
    MatchTimeEpsilon: 1e-10 # <-- Use 'Auto' for a default of 1e-12
    SetL2CoefsToZero: True
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #5921 and #5922.~

### DIFF
--- a/src/Domain/Creators/ShapeMapOptions.cpp
+++ b/src/Domain/Creators/ShapeMapOptions.cpp
@@ -12,6 +12,8 @@
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Domain/Structure/ObjectLabel.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Spherepack.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/SpherepackIterator.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
 #include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrHorizon.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/StdArrayHelpers.hpp"
@@ -47,6 +49,45 @@ initial_shape_and_size_funcs(
       size_funcs[0][0] = shape_funcs[0][0] * sqrt(0.5 * M_PI);
       // Set l=0 for shape map to 0 because size is going to be used
       shape_funcs[0][0] = 0.0;
+    } else if (std::holds_alternative<YlmsFromFile>(
+                   shape_options.initial_values.value())) {
+      const auto& files =
+          std::get<YlmsFromFile>(shape_options.initial_values.value());
+      const std::string& h5_filename = files.h5_filename;
+      const std::vector<std::string>& subfile_names = files.subfile_names;
+      const double match_time = files.match_time;
+      const double match_time_epsilon =
+          files.match_time_epsilon.value_or(1e-12);
+      const bool set_l1_coefs_to_zero = files.set_l1_coefs_to_zero;
+      const size_t l_max = shape_options.l_max;
+      ylm::SpherepackIterator iter{l_max, l_max};
+
+      for (size_t i = 0; i < subfile_names.size(); i++) {
+        // Frame doesn't matter here
+        const ylm::Strahlkorper<Frame::Distorted> file_strahlkorper =
+            ylm::read_surface_ylm_single_time<Frame::Distorted>(
+                h5_filename, gsl::at(subfile_names, i), match_time,
+                match_time_epsilon);
+        const ylm::Strahlkorper<Frame::Distorted> this_strahlkorper{
+            shape_options.l_max, 1.0, std::array{0.0, 0.0, 0.0}};
+
+        // The coefficients in the shape map are stored as the negative
+        // coefficients of the strahlkorper, so we need to multiply by -1 here.
+        gsl::at(shape_funcs, i) =
+            -1.0 * file_strahlkorper.ylm_spherepack().prolong_or_restrict(
+                       file_strahlkorper.coefficients(),
+                       this_strahlkorper.ylm_spherepack());
+        // Transform from SPHEREPACK to actual Ylm for size func
+        gsl::at(size_funcs, i)[0] =
+            gsl::at(shape_funcs, i)[0] * sqrt(0.5 * M_PI);
+        // Set l=0 for shape map to 0 because size is going to be used
+        gsl::at(shape_funcs, i)[0] = 0.0;
+        if (set_l1_coefs_to_zero) {
+          for (int m = -1; m <= 1; m++) {
+            gsl::at(shape_funcs, i)[iter.set(1_st, m)()] = 0.0;
+          }
+        }
+      }
     }
   }
 
--- a/src/Domain/Creators/ShapeMapOptions.hpp
+++ b/src/Domain/Creators/ShapeMapOptions.hpp
@@ -12,7 +12,10 @@
 
 #include "DataStructures/DataVector.hpp"
 #include "Domain/Structure/ObjectLabel.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
 #include "Options/Auto.hpp"
+#include "Options/Context.hpp"
 #include "Options/String.hpp"
 #include "Utilities/Gsl.hpp"
 #include "Utilities/TMPL.hpp"
@@ -53,6 +56,63 @@ struct KerrSchildFromBoyerLindquist {
 /// Label for shape map options
 struct Spherical {};
 
+struct YlmsFromFile {
+  struct H5Filename {
+    using type = std::string;
+    static constexpr Options::String help =
+        "Path to the data file containing the ylm coefficients and their "
+        "derivatives.";
+  };
+
+  struct SubfileNames {
+    using type = std::vector<std::string>;
+    static constexpr Options::String help =
+        "Subfile names for the different order derivatives of the ylm "
+        "coefficients. You must specify the subfile name for the ylm "
+        "coefficients themselves, and can optionally specify the subfile name "
+        "for the first and second time derivatives as well, in that order. If "
+        "you don't specify a derivative subfile, those coefficients will be "
+        "defaulted to zero.";
+    static size_t lower_bound_on_size() { return 1; }
+    static size_t upper_bound_on_size() { return 3; }
+  };
+
+  struct MatchTime {
+    using type = double;
+    static constexpr Options::String help =
+        "Time in the H5File to get the coefficients at. Will likely be the "
+        "same as the initial time";
+  };
+
+  struct MatchTimeEpsilon {
+    using type = Options::Auto<double>;
+    static constexpr Options::String help =
+        "Look for times in the H5File within this epsilon of the match time. "
+        "This is to avoid having to know the exact time to all digits. Default "
+        "is 1e-12.";
+  };
+
+  struct SetL1CoefsToZero {
+    using type = bool;
+    static constexpr Options::String help =
+        "Whether to set the L=1 coefs to zero or not. This may be desirable "
+        "because L=1 is degenerate with a translation of the BH.";
+  };
+
+  using options = tmpl::list<H5Filename, SubfileNames, MatchTime,
+                             MatchTimeEpsilon, SetL1CoefsToZero>;
+
+  static constexpr Options::String help = {
+      "Read the Y_lm coefficients of a Strahlkorper from file and use those to "
+      "initialize the coefficients of a shape map."};
+
+  std::string h5_filename{};
+  std::vector<std::string> subfile_names{};
+  double match_time{};
+  std::optional<double> match_time_epsilon{};
+  bool set_l1_coefs_to_zero{};
+};
+
 /*!
  * \brief Class to be used as an option for initializing shape map coefficients.
  *
@@ -79,7 +139,8 @@ struct ShapeMapOptions {
 
   struct InitialValues {
     using type =
-        Options::Auto<std::variant<KerrSchildFromBoyerLindquist>, Spherical>;
+        Options::Auto<std::variant<KerrSchildFromBoyerLindquist, YlmsFromFile>,
+                      Spherical>;
     static constexpr Options::String help = {
         "Initial Ylm coefficients for the shape map. Specify 'Spherical' for "
         "all coefficients to be initialized to zero."};
@@ -109,7 +170,8 @@ struct ShapeMapOptions {
                           common_options>;
 
   size_t l_max{};
-  std::optional<std::variant<KerrSchildFromBoyerLindquist>> initial_values{};
+  std::optional<std::variant<KerrSchildFromBoyerLindquist, YlmsFromFile>>
+      initial_values{};
   std::optional<std::array<double, 3>> initial_size_values{};
   bool transition_ends_at_cube{false};
 };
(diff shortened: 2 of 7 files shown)


src/Domain/Creators/ShapeMapOptions.cpp
```diff
@@ -47,6 +49,41 @@ initial_shape_and_size_funcs(
...
+        // Frame doesn't matter here
+        const ylm::Strahlkorper<Frame::Distorted> file_strahlkorper =
+            ylm::read_surface_ylm_single_time<Frame::Distorted>(
+                h5_filename, gsl::at(subfile_names, i), match_time,
+                match_time_epsilon);
+        const ylm::Strahlkorper<Frame::Distorted> this_strahlkorper{
+            shape_options.l_max, 1.0, std::array{0.0, 0.0, 0.0}};
+
+        // The coefficients in the shape map are stored as the negative
+        // coefficients of the strahlkorper, so we need to multiply by -1 here.
```
nilsvu, line 75: I didn't find this in the documentation of the Shape map. Could you check and edit the docs to reflect this?

src/Domain/Creators/ShapeMapOptions.cpp
```diff
@@ -47,6 +49,41 @@ initial_shape_and_size_funcs(
...
+        gsl::at(shape_funcs, i) =
+            -1.0 * file_strahlkorper.ylm_spherepack().prolong_or_restrict(
+                       file_strahlkorper.coefficients(),
+                       this_strahlkorper.ylm_spherepack());
+        gsl::at(shape_funcs, i)[0] = 0.0;
```
nilsvu, line 84: Need to set `size_funcs[0][0]`?

src/Domain/Creators/ShapeMapOptions.hpp
```diff
@@ -53,6 +56,61 @@ struct KerrSchildFromBoyerLindquist {
...
+struct YlmsFromFile {
+  struct H5Filename {
+    using type = std::string;
+    static constexpr Options::String help =
+        "Path to the data file containing the ylm coefficients and their "
+        "derivatives.";
+  };
+
+  struct SubfileNames {
+    using type = std::array<std::string, 3>;
```
nilsvu: What if I don't have the time derivatives (yet), i.e. just starting from a horizon found in initial data? Should these be optionals?

src/Domain/Creators/ShapeMapOptions.hpp
```diff
@@ -53,6 +56,61 @@ struct KerrSchildFromBoyerLindquist {
...
+  };
+
+  using options = tmpl::list<H5Filename, SubfileNames, MatchTime,
+                             MatchTimeEpsilon, SetL1CoefsToZero>;
+
+  static constexpr Options::String help = {
+      "Read the Y_lm coefficients of a Strahlkorper from file and use those to "
+      "initialize the coefficients of a shape map. The 00 coefficients (for "
+      "all derivs) are set to zero automatically. Note that the coefficients "
+      "that are read in will be multiplied by -1 to conform with how we store "
```
nilsvu: The -1 thing is an implementation detail, remove.

tests/Unit/Domain/Creators/Test_ShapeMapOptions.cpp
```diff
@@ -113,11 +174,100 @@ void test_funcs() {
...
+            shape_map_options, inner_radius);
+
+    ylm::SpherepackIterator iter{l_max, l_max};
+    ylm::SpherepackIterator file_iter{file_l_max, file_l_max};
+    for (size_t i = 0; i < shape_funcs.size(); i++) {
+      const size_t expected_size = ylm::Spherepack::spectral_size(l_max, l_max);
+      // Make sure they were restricted properly
+      CHECK(gsl::at(shape_funcs, i).size() == expected_size);
+
+      // Loop pointwise so we only check the coefficients that matter. Also
```
nilsvu: Also what? :D

nilsvu (changes requested):

tests/Unit/Domain/Creators/Test_ShapeMapOptions.cpp
knelli2: Could you expand on what the confusion is? The coefficients as stored in spherepack have a ton of zeros in them which we don't need to check

src/Domain/Creators/ShapeMapOptions.cpp
knelli2: Added a new commit with these changes

tests/Unit/Domain/Creators/Test_ShapeMapOptions.cpp
nilsvu: Oh sorry there's just an "Also" at the end of this line with nothing following it, sort of creating suspense :)

src/Domain/Creators/ShapeMapOptions.hpp
knelli2: For some reason, option parsing didn't like `std::array<Options::Auto<std::string>, 3>` and I didn't feel like going down the rabbit hole of why options didn't interpret this correctly so I just went with `std::vector<std::string>` instead and using upper/lower bounds on the size. I think it's much less common to need the time deriv of the coefs and not the coefs themselves. If somebody does need that then we can revisit this

tests/Unit/Domain/Creators/Test_ShapeMapOptions.cpp
knelli2: Ohhhhhhhh. lol oops. I'll get that when I squash

src/Domain/Creators/ShapeMapOptions.cpp
nilsvu: Can you add a test that would have caught this bug?

knelli2: Done

tests/Unit/Domain/Creators/Test_ShapeMapOptions.cpp
```diff
@@ -215,46 +215,78 @@ void test_funcs(const gsl::not_null<Generator*> generator) {
...
+      const auto [shape_funcs, size_funcs] =
+          domain::creators::time_dependent_options::
+              initial_shape_and_size_funcs(shape_map_options, inner_radius);
+
+      CHECK(size_funcs ==
+            std::array{DataVector{-1.0 * strahlkorpers[0].coefficients()[0] *
+                                  sqrt(0.5 * M_PI)},
+                       DataVector{-1.0 * strahlkorpers[1].coefficients()[0] *
+                                  sqrt(0.5 * M_PI)},
+                       DataVector{0.0}, DataVector{0.0}});
```
nilsvu, line 291: Also check zero coeffs of shape funcs.

Maybe a stronger test would be to write out a strahlkorper, load it back in, and check that the xi=-1 logical coord of neighboring blocks map to the cartesian coords of the strahlkorper. I added such a test for the sphere and BBH domains for the Kerr-horizon-conforming case, so you can copy most of that. That test would catch any bugs with factors of pi or minus signs or anything like that.

nilsvu (changes requested):

knelli2:
@nilsvu Could you take another look at this? Maybe before we merge #5953 since it'll be easier to resolve conflicts there than here

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -784,42 +771,121 @@ void test_kerr_horizon_conforming(const bool use_time_dependence) {
...
+                                           num_points)}};
+
+  const double time = 0.7;
+  const size_t l_max = 32;
+  const double mass = 0.8;
+  const std::array<double, 3> spin{{0.0, 0.0, 0.9}};
+  const double r_plus = mass * (1. + sqrt(1. - dot(spin, spin)));
+  const double inner_radius = r_plus;
+  std::string info_string{
+      "Check that inner radius is deformed to constant Boyer-Lindquist radius"};
```
nilsvu: Don't pass this to functions. Just use a scope and set `INFO("...")` for each test.

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -784,42 +771,121 @@ void test_kerr_horizon_conforming(const bool use_time_dependence) {
...
+
+  DataVector radius =
+      get(gr::Solutions::kerr_schild_radius_from_boyer_lindquist(
+          inner_radius, theta_phi, mass, spin));
+  tnsr::I<DataVector, 3> x{};
+  const auto set_x = [&x, &radius, &theta_phi]() {
```
nilsvu: This is confusing :D Just take radius, theta_phi as arguments.

nilsvu (changes requested):

knelli2:
Squashed everything and rebased

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -784,42 +770,131 @@ void test_kerr_horizon_conforming(const bool use_time_dependence) {
...
+  }
+
+  {
+    INFO("Check reading in Ylms from file");
+    const std::string h5_filename{"StrahlkorperCoefsFile.h5"};
+    const std::string subfile_name{"Ylm_coefs"};
+    if (file_system::check_if_file_exists(h5_filename)) {
+      file_system::rm(h5_filename, true);
+    }
+    ylm::Strahlkorper<Frame::Distorted> strahlkorper{l_max, inner_radius,
```
nilsvu: This doesn't actually deform anything. Aren't all coefficients that get multiplied with -1 for the shape map zero? Just construct the strahlkorper from the Kerr-Schild radius you construct above, then all three tests here do the same deformation just in different ways.

nilsvu (changes requested):

knelli2:
@nilsvu #5953 seems like it's almost ready to be merged

src/Domain/Creators/ShapeMapOptions.hpp
```diff
@@ -53,6 +56,64 @@ struct KerrSchildFromBoyerLindquist {
...
+    static constexpr Options::String help =
+        "Path to the data file containing the ylm coefficients and their "
+        "derivatives.";
+  };
+
+  struct SubfileNames {
+    using type = std::vector<std::string>;
+    static constexpr Options::String help =
+        "Subfile names for the different order derivatives of the ylm "
+        "coefficients without the preceding '/' and without the '.dat' "
```
nilsvu: I believe this condition is relaxed now

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -784,42 +770,130 @@ void test_kerr_horizon_conforming(const bool use_time_dependence) {
...
+  const ylm::Spherepack ylm{l_max, l_max};
+  const std::array<DataVector, 2> theta_phi = ylm.theta_phi_points();
+
+  const double time = 0.7;
+  const double mass = 0.8;
+  const std::array<double, 3> spin{{0.0, 0.0, 0.9}};
+  const double r_plus = mass * (1. + sqrt(1. - dot(spin, spin)));
+  const double inner_radius = r_plus;
+
+  const auto set_x = [](const DataVector& local_radius,
```
nilsvu: You don't need the lambda anymore, just define x here

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -784,42 +770,130 @@ void test_kerr_horizon_conforming(const bool use_time_dependence) {
...
+
+    time_dependent_options = domain::creators::sphere::TimeDependentMapOptions{
+        time,
+        {{l_max,
+          domain::creators::time_dependent_options::YlmsFromFile{
+              h5_filename, std::vector{subfile_name}, time, std::nullopt,
+              false},
+          // Constructing a strahlkorper from collocation radii will not exactly
+          // match the collocation points (see Strahlkorper constructor docs).
+          // For this reason, the 00 coef we calculate is not exact. This is why
```
nilsvu, line 862: Does this mean the round trip (write shape coeffs to file, then read them back in to shape map) is not exact? If so please note that caveat down in docs and option help, and we should probably fix that.

nilsvu (commented):
You can squash

tests/Unit/Domain/Creators/Test_Sphere.cpp
knelli2: No it's not the writing/reading, it's just constructing the strahlkorper from collocation radii that apparently isn't exact. It's noted in the Strahlkorper constructor and I noted it here

nilsvu: Oh and in the round trip we write out the Strahlkorper coefficients directly. Sounds good then.

nilsvu (approved):

PR #5924, opened 2024-04-18
Normalize subfile path in H5 lib

nilsvu:
This has bothered me for a long time. A leading slash and _no_ extension for subfile names was required before, and failing to do this lead to rather useless error messages. Now, the leading slash and the extension are handled within the H5 lib, so it should be much easier to use.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/H5/File.hpp
+++ b/src/IO/H5/File.hpp
@@ -209,7 +209,7 @@ class H5File {
 
   template <typename ObjectType>
   std::tuple<bool, detail::OpenGroup, std::string> check_if_object_exists(
-      const std::string& path) const;
+      std::string path) const;
 
   std::string file_name_;
   // NOLINTNEXTLINE(spectre-mutable)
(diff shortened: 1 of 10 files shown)


src/IO/H5/File.hpp
```diff
@@ -392,7 +392,16 @@ const ObjectType& H5File<Access_t>::convert_to_derived(
 template <AccessType Access_t>
 template <typename ObjectType>
 std::tuple<bool, detail::OpenGroup, std::string>
-H5File<Access_t>::check_if_object_exists(const std::string& path) const {
+H5File<Access_t>::check_if_object_exists(std::string path) const {
+  // Normalize path to have a leading slash and the correct extension
+  if (path.front() != '/') {
+    path = '/' + path;
```
knelli2, line 398: Can you add a test of this normalization in the c++ tests?

knelli2 (changes requested):
@nilsdeppe Could you take a quick look at this? I want to merge this before #5919 because it's used in there

src/IO/H5/File.hpp
nilsvu: done and squashed in

nilsdeppe:
No need to wait for explicit approval from me.

knelli2 (approved):

PR #5929, opened 2024-04-19
Add CLI command to plot elliptic solver convergence

nilsvu:
## Proposed changes

Example:

convergence

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/PlotEllipticConvergence.py
+++ b/src/Visualization/Python/PlotEllipticConvergence.py
@@ -1,0 +1,137 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+from typing import List
+
+import click
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+from matplotlib.ticker import MaxNLocator
+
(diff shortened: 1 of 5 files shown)


src/Visualization/Python/PlotEllipticConvergence.py
```diff
@@ -0,0 +1,137 @@
...
+            )
+            m += len(residuals) - 1
+    # Plot linear solver residuals
+    for i, residuals in enumerate(linear_residuals):
+        plt.plot(
+            residuals.index + cumulative_linsolv_iterations[i],
+            residuals["Residual"] / norm,
+            color="black",
+            label="Linear residual" if i == 0 else None,
+            marker="." if len(residuals) < 20 else None,
```
knelli2, line 97: Why not just always have a marker?

knelli2 (approved):

src/Visualization/Python/PlotEllipticConvergence.py
nilsvu: It's too busy if you have so many markers that you can't distinguish them anymore. And inversely, you should have markers if the line has visible kinks. So 20 seems like a reasonable but fairly arbitrary value in between.

PR #5931, opened 2024-04-20
Compute a worldtube constraint for Klein-Gordon Cce

Sizheng-Ma:
## Proposed changes

Both Cauchy and characteristic systems evolve the scalar field independently. The results have to be consistent at the worldtube. This yields a worldtube constraint (`Cce::Tags::KleinGordonWorldtubeConstraint`) that needs to be monitored during evolution, which reads:

$C=\psi^{\rm Cauchy}-\psi^{\rm Characteristic}$,

where $\psi^{\rm Cauchy}$ is from Cauchy's worldtube data (`Cce::Tags::BoundaryValue<Cce::Tags::KleinGordonPsi>`) and $\psi^{\rm Characteristic}$ is the evolved variable (`Cce::Tags::KleinGordonPsi`). 

This PR adds an action to compute `Cce::Tags::KleinGordonWorldtubeConstraint` and dump it to the disk. The first commit adds its tag. The second one initializes the tag in the databox. The third commit adds the action. And the last one dumps the tag to the disk.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/GaugeTransformBoundaryData.hpp
+++ b/src/Evolution/Systems/Cce/GaugeTransformBoundaryData.hpp
@@ -848,4 +848,29 @@ struct InitializeGauge {
       gsl::not_null<Scalar<SpinWeighted<ComplexDataVector, 0>>*> omega,
       size_t l_max);
 };
+
+/*!
+ * \brief Compute a worldtube constraint for Klein-Gordon Cce
+ *
+ * \details Both Cauchy and characteristic systems evolve the scalar field
+ * independently. The results have to be consistent at the worldtube. The
+ * difference (`Tags::KleinGordonWorldtubeConstraint`) between the Cauchy
+ * worldtube data `Tags::BoundaryValue<Tags::KleinGordonPsi>` and the evolved
+ * volume data `Tags::KleinGordonPsi` (at the worldtube)
+ * indicates the accuracy of the simulation.
+ */
+struct ComputeKGWorldtubeConstraint {
+  using return_tags = tmpl::list<Tags::KleinGordonWorldtubeConstraint>;
+  using argument_tags = tmpl::list<
+      Tags::BoundaryValue<Tags::KleinGordonPsi>,
+      Spectral::Swsh::Tags::SwshInterpolator<Tags::CauchyAngularCoords>,
+      Tags::LMax, Tags::KleinGordonPsi>;
+
+  static void apply(
+      gsl::not_null<Scalar<SpinWeighted<ComplexDataVector, 0>>*>
+          evolution_kg_constraint,
+      const Scalar<SpinWeighted<ComplexDataVector, 0>>& cauchy_kg_psi,
+      const Spectral::Swsh::SwshInterpolator& interpolator, size_t l_max,
+      const Scalar<SpinWeighted<ComplexDataVector, 0>>& volume_psi);
+};
 }  // namespace Cce
--- a/src/Evolution/Systems/Cce/GaugeTransformBoundaryData.hpp
+++ b/src/Evolution/Systems/Cce/GaugeTransformBoundaryData.hpp
@@ -848,29 +848,4 @@ struct InitializeGauge {
       gsl::not_null<Scalar<SpinWeighted<ComplexDataVector, 0>>*> omega,
       size_t l_max);
 };
-
-/*!
- * \brief Compute a worldtube constraint for Klein-Gordon Cce
- *
- * \details Both Cauchy and characteristic systems evolve the scalar field
- * independently. The results have to be consistent at the worldtube. The
- * difference (`Tags::KleinGordonWorldtubeConstraint`) between the Cauchy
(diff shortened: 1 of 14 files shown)


src/Evolution/Systems/Cce/GaugeTransformBoundaryData.hpp
```diff
@@ -848,4 +848,29 @@ struct InitializeGauge {
...
+ * \brief Compute a worldtube constraint for Klein-Gordon Cce
+ *
+ * \details Both Cauchy and characteristic systems evolve the scalar field
+ * independently. The results have to be consistent at the worldtube. The
+ * difference (`Tags::KleinGordonWorldtubeConstraint`) between the Cauchy
+ * worldtube data `Tags::BoundaryValue<Tags::KleinGordonPsi>` and the evolved
+ * volume data `Tags::KleinGordonPsi` (at the worldtube)
+ * indicates the accuracy of the simulation.
+ */
+struct ComputeKGWorldtubeConstraint {
```
wthrowe: This doesn't appear to have anything to do with gauge transformations, so it should probably go in a different file.

wthrowe: This needs a test that actually checks that the result.  The current test just checks that `db::mutate_apply<ComputeKGWorldtubeConstraint>` and `::Actions::MutateApply<ComputeKGWorldtubeConstraint>` give the same answer, which doesn't give much evidence that it is the correct answer.

wthrowe (changes requested):

src/Evolution/Systems/Cce/GaugeTransformBoundaryData.hpp
Sizheng-Ma: Testing the physical aspect of the calculation is a quite nontrivial task. In fact, it will be one of the main focuses of the upcoming method paper for this CCE system --- in the paper, a variety of systems are used to show that the constraint converges exponentially with resolution (I've obtained many solid results to verify this is indeed the case). Another PR #5932 initiates the steps to build worldtube data for those systems. The goal therein is to add a new executable (in parallel with `AnalyticTestCharacteristicExtract`) that can justify all the physical aspects of the new CCE system, including the worldtube constraint in this PR.

Sizheng-Ma: I've moved the mutator to a new file.

wthrowe: I'm not concerned about the physics, here.  I want some evidence that the variable slicing is doing what you expect.  At the moment, all the test does is checks that the function will run without crashing.

knelli2: @Sizheng-Ma I believe what @wthrowe is looking for is just a test where you give the `apply` function a random cauchy KG Psi, volume Psi, and an interpolator, and you check that you get out an expected value (since it's just a subtraction).

knelli2 (changes requested):

src/Evolution/Systems/Cce/GaugeTransformBoundaryData.hpp
wthrowe: Yes, that would be fine.

Sizheng-Ma: I think the test added in the third commit (`Add ComputeKGWorldtubeConstraint`) already achieved the goal. See `tests/Unit/Evolution/Systems/Cce/Actions/Test_KleinGordonCceCalculations.cpp`

wthrowe: You need a test that would fail if I changed the body of the function so it computed the wrong thing.  For example, I just checked out this PR and made the change
```diff
diff --git a/src/Evolution/Systems/Cce/WorldtubeConstraint.cpp b/src/Evolution/Systems/Cce/WorldtubeConstraint.cpp
index 9f99d5daf..594f041d8 100644
--- a/src/Evolution/Systems/Cce/WorldtubeConstraint.cpp
+++ b/src/Evolution/Systems/Cce/WorldtubeConstraint.cpp
@@ -17,6 +17,6 @@ void ComputeKGWorldtubeConstraint::apply(
   const SpinWeighted<ComplexDataVector, 0> surface_psi;
   make_const_view(make_not_null(&surface_psi), get(volume_psi), 0,
                   Spectral::Swsh::number_of_swsh_collocation_points(l_max));
-  get(*evolution_kg_constraint) = evolution_kg_psi - surface_psi;
+  get(*evolution_kg_constraint) = evolution_kg_psi + surface_psi;
 }
 }  // namespace Cce
```
and the test still passes.

Sizheng-Ma: Okay, I see. The test has been updated, where the expected result is now computed manually.

PR #5932, opened 2024-04-20
Add KleinGordonWorldtubeData and its option tag

Sizheng-Ma:
## Proposed changes

This PR adds an abstract base class `KleinGordonWorldtubeData` that handles analytic worldtube data for Klein-Gordon Cce. The future Cce executable built upon it aims to provide a battery of tests for the system, e.g., bouncing Schwarzschild black hole in Sec. V of [Moxon et al. ](https://arxiv.org/pdf/2110.08635.pdf)

The content of `KleinGordonWorldtubeData` is fully in parallel with `WorldtubeData`

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/AnalyticSolutions/WorldtubeData.hpp
+++ b/src/Evolution/Systems/Cce/AnalyticSolutions/WorldtubeData.hpp
@@ -40,6 +40,11 @@ using cce_analytic_solutions_cache_tags = tmpl::list<
     Tags::Dr<gr::Tags::Shift<DataVector, 3>>,
     Tags::Dr<gr::Tags::Lapse<DataVector>>, Tags::News>;
 
+/// The collection of cache tags for `KleinGordonWorldtubeData`
+using kg_cce_analytic_solutions_cache_tags =
+    tmpl::list<Tags::CauchyCartesianCoords, Tags::KleinGordonPsi,
+               Tags::KleinGordonPi>;
+
 /// \cond
 class BouncingBlackHole;
 class GaugeWave;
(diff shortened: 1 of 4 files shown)


src/Evolution/Systems/Cce/AnalyticSolutions/WorldtubeData.hpp
```diff
@@ -288,5 +293,121 @@ struct WorldtubeData
...
+ * \warning This class is not intended to be threadsafe! Therefore, using
+ * instances of this class placed into the const global cache results in
+ * undefined behavior. The analytic data for CCE is not easily represented as a
+ * full closed-form solution for the scalar field over the
+ * domain, so this class and its descendants perform numerical calculations.
+ * Instead, it makes best sense to compute the global solution over the
+ * extraction sphere, and cache intermediate steps to avoid repeating
+ * potentially expensive calculations.
+ */
+struct KleinGordonWorldtubeData
```
knelli2, line 330: Is there an analogous test for the regular WorldtubeData that you can use to model a test for this class after?

src/Evolution/Systems/Cce/OptionTags.hpp
```diff
@@ -192,6 +192,13 @@ struct AnalyticSolution {
   using group = Cce;
 };
 
+struct KleinGordonAnalyticSolution {
+  using type = std::unique_ptr<Solutions::KleinGordonWorldtubeData>;
```
knelli2, line 198: Add a test for this next to the regular `AnalyticSolution` test in `tests/Unit/Evolution/Systems/Cce/Test_OptionTags.cpp`

knelli2 (changes requested):

src/Evolution/Systems/Cce/OptionTags.hpp
Sizheng-Ma: The problem is that this abstract class now doesn't have any derived class (which will be added via following PRs). Do you have any suggestions on how to add the test?

PR #5941, opened 2024-04-24
Observe walltime in elliptic solver

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/PlotEllipticConvergence.py
+++ b/src/Visualization/Python/PlotEllipticConvergence.py
@@ -3,6 +3,7 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
+import datetime
 import logging
 from typing import List
 
@@ -17,7 +18,7 @@ from spectre.Visualization.Plot import (
     apply_stylesheet_command,
     show_or_save_plot_command,
 )
-from spectre.Visualization.ReadH5 import to_dataframe
+from spectre.Visualization.ReadH5 import available_subfiles, to_dataframe
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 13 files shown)


nilsdeppe (dismissed):

knelli2 (changes requested):
Looks ok except for the tests

src/Visualization/Python/PlotEllipticConvergence.py
```diff
@@ -73,39 +75,71 @@ def plot_elliptic_convergence(
...
                 residuals["Residual"] / norm,
                 color="black",
                 ls="dotted",
                 marker=".",
-                label="Nonlinear residual" if i == 0 else None,
+                label="Nonlinear solver" if i == 0 else None,
+            )
+            ax_time.plot(
+                cumulative_linsolv_iterations[m : m + len(residuals) - 1],
+                np.diff(residuals["Walltime"]),
```
knelli2: Check if the "Walltime" was dumped first. Otherwise this won't work on any existing ID runs. We can remove the check later once enough time has passed and all runs we are interested in output the walltime.

knelli2 (changes requested):

src/Visualization/Python/PlotEllipticConvergence.py
nilsvu: Ok pushed a fixup

knelli2 (dismissed):
Looks good. You can squash

knelli2 (commented):
The fixups look ok, but MacOS CI is complaining about an empty file? Not sure what that's about since it doesn't look like you touched that file

knelli2 (approved):

PR #5944, opened 2024-04-25
Change translation control to work like SpEC

AlexCarpenter46:
## Proposed changes

Changes translation control to work more like SpEC and take into account both objects instead of just the primary.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Translation.hpp
+++ b/src/ControlSystem/ControlErrors/Translation.hpp
@@ -41,12 +41,12 @@ namespace ControlErrors {
  * domain::CoordinateMaps::TimeDependent::Translation Translation \endlink
  * coordinate map
  *
- * \details Computes the error in how much the system has translated by using
- * Eq. (42) from \cite Ossokine2013zga. The equation is
- *
- * \f[ \left(0, \delta\vec{T}\right) = a\mathbf{q}\left(\mathbf{x}_A -
- * \mathbf{c}_A - \mathbf{\delta q}\wedge\mathbf{c}_A - \frac{\delta
- * a}{a}\mathbf{c}_A \right)\mathbf{q}^*
+ * \details Computes the error in how much the system has translated by using a
+ * modified version of Eq. (42) from \cite Ossokine2013zga. The equation is
+ * \f[ \left(0, \delta\vec{T}\right) = a\mathbf{q}\left(\frac{1}{2}(\mathbf{x}_A
+ * + \mathbf{x}_B - \frac{1}{2}(\mathbf{c}_A + \mathbf{c}_B)) - \mathbf{\delta
+ * q}\wedge\frac{1}{2}(\mathbf{c}_A + \mathbf{c}_B) - \frac{\delta
+ * a}{a}\frac{1}{2}(\mathbf{c}_A + \mathbf{c}_B) \right)\mathbf{q}^*
  * \f]
  *
  * where object A is located on the positive x-axis in the grid frame, bold face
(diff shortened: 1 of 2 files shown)


src/ControlSystem/ControlErrors/Translation.hpp
```diff
@@ -109,30 +111,62 @@ struct Translation : tt::ConformsTo<protocols::ControlError> {
...
+        Parallel::get<domain::Tags::ObjectCenter<domain::ObjectLabel::B>>(
+            cache);
+    const DataVector grid_position_of_B{{grid_position_of_B_tnsr[0],
+                                         grid_position_of_B_tnsr[1],
+                                         grid_position_of_B_tnsr[2]}};
+    const DataVector& current_position_of_B = get<center_B>(measurements);
+
+    const DataVector& grid_position_average =
+        0.5 * (grid_position_of_A + grid_position_of_B);
+    const DataVector& current_position_average =
```
knelli2: These shouldn't be references since you are computing something new. Same for the separations.

src/ControlSystem/ControlErrors/Translation.hpp
```diff
@@ -109,30 +111,62 @@ struct Translation : tt::ConformsTo<protocols::ControlError> {
...
+    const DataVector& current_separation =
+        current_position_of_A - current_position_of_B;
+
+    // These quantities come from the translation control implementation in SpEC
+    double current_separation_dot_grid_separation = 0.0;
+    double current_separation_dot_grid_average = 0.0;
+    double grid_separation_dot_grid_average = 0.0;
+    double grid_separation_dot_grid_separation = 0.0;
+    for (size_t i = 0; i < 3; i++) {
+      current_separation_dot_grid_separation +=
```
knelli2: You can use `dot(left, right)` for all of these. No need for a loop.

src/ControlSystem/ControlErrors/Translation.hpp
```diff
@@ -109,30 +111,62 @@ struct Translation : tt::ConformsTo<protocols::ControlError> {
...
+          grid_separation[i] * grid_separation[i];
+    }
+
+    // From eq. 42 in 1304.3067 where the grid and current position are swapped
+    // from only object A to the average grid and current positions of both
+    // objects.
+    DataVector translation_control =
+        make_with_value<DataVector>(grid_position_of_A, 0.0);
+    for (size_t i = 0; i < 3; i++) {
+      translation_control[i] =
```
knelli2: Also no need for a loop here. This can just be
```cpp
DataVector translation_control = 
      expansion_factor / grid_separation_dot_grid_separation * (...);
```
(I believe you did it like this in the test)

knelli2 (changes requested):
You can squash directly

src/ControlSystem/ControlErrors/Translation.hpp
```diff
@@ -109,30 +111,54 @@ struct Translation : tt::ConformsTo<protocols::ControlError> {
...
+        dot(grid_separation, grid_position_average);
+    double grid_separation_dot_grid_separation =
+        dot(grid_separation, grid_separation);
+
+    // From eq. 42 in 1304.3067 where the grid and current position are swapped
+    // from only object A to the average grid and current positions of both
+    // objects.
+    DataVector translation_control =
+        make_with_value<DataVector>(grid_position_of_A, 0.0);
+    translation_control =
```
knelli2: [nitpick] No need to allocate first, then assign. Just do it all in one step

knelli2 (changes requested):

src/ControlSystem/ControlErrors/Translation.hpp
AlexCarpenter46: oh whoops yeah, forgot to take that out with the loop

knelli2 (approved):

PR #5947, opened 2024-04-26
Parallelize block_logical_coords with OpenMP if enabled

nilsvu:
## Proposed changes

The majority of time spent in interpolation and loading numeric initial data is spent in `block_logical_coordinates`, mapping
all grid points through the domain. It's very easily parallelized though (literally 1 line of OpenMP for perfect strong scaling).

OpenMP is disabled in our charm++ execs unless explicitly requested, so this doesn't have an effect there. It currently only applies to pybindings and the exporter lib. However, we are loading numeric initial data on a nodegroup, so we may even want to enable the OpenMP parallelization there. I haven't tested it yet though.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SpectreAddCatchTests.cmake
+++ b/cmake/SpectreAddCatchTests.cmake
@@ -42,11 +42,11 @@ add_custom_target(unit-tests)
 spectre_define_test_timeout_factor_option(UNIT "unit")
 
 # Environment variables for test
-set(_TEST_ENV_VARS "")
+set(_CATCH_TEST_ENV_VARS "")
 # - Disable ASAN's leak sanitizer because Charm++ has false positives
-list(APPEND _TEST_ENV_VARS "ASAN_OPTIONS=detect_leaks=0")
+list(APPEND _CATCH_TEST_ENV_VARS "ASAN_OPTIONS=detect_leaks=0")
 # - Set PYTHONPATH to find Python modules
-list(APPEND _TEST_ENV_VARS "PYTHONPATH=${PYTHONPATH}")
+list(APPEND _CATCH_TEST_ENV_VARS "PYTHONPATH=${PYTHONPATH}")
 
 # Main function - the only one designed to be called from outside this module.
 function(spectre_add_catch_tests TEST_TARGET)
@@ -225,7 +225,7 @@ function(spectre_parse_file SOURCE_FILE TEST_TARGET)
         TIMEOUT ${TIMEOUT}
         PASS_REGULAR_EXPRESSION "${OUTPUT_REGEX}"
         LABELS "${TAGS}"
-        ENVIRONMENT "${_TEST_ENV_VARS}")
+        ENVIRONMENT "${_CATCH_TEST_ENV_VARS}")
       set(FAILURE_TESTS "\"~${CTEST_NAME}\";${FAILURE_TESTS}")
     else ()
       set_tests_properties(
@@ -233,7 +233,7 @@ function(spectre_parse_file SOURCE_FILE TEST_TARGET)
         FAIL_REGULAR_EXPRESSION "No tests ran"
         TIMEOUT ${TIMEOUT}
         LABELS "${TAGS}"
-        ENVIRONMENT "${_TEST_ENV_VARS}")
+        ENVIRONMENT "${_CATCH_TEST_ENV_VARS}")
     endif ()
   endforeach ()
   set_property(GLOBAL PROPERTY SPECTRE_FAILURE_TESTS ${FAILURE_TESTS})
--- a/src/Domain/BlockLogicalCoordinates.cpp
+++ b/src/Domain/BlockLogicalCoordinates.cpp
@@ -11,26 +11,17 @@
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "Domain/Block.hpp"
 #include "Domain/Domain.hpp"  // IWYU pragma: keep
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
 #include "Domain/Structure/BlockId.hpp"
 #include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 
-namespace {
-// Define this alias so we don't need to keep typing this monster.
-template <size_t Dim>
-using block_logical_coord_holder =
-    std::optional<IdPair<domain::BlockId,
-                         tnsr::I<double, Dim, typename ::Frame::BlockLogical>>>;
-using functions_of_time_type = std::unordered_map<
-    std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>;
-}  // namespace
-
 template <size_t Dim, typename Frame>
 std::optional<tnsr::I<double, Dim, ::Frame::BlockLogical>>
 block_logical_coordinates_single_point(
     const tnsr::I<double, Dim, Frame>& input_point, const Block<Dim>& block,
-    const double time, const functions_of_time_type& functions_of_time) {
+    const double time, const domain::FunctionsOfTimeMap& functions_of_time) {
   std::optional<tnsr::I<double, Dim, ::Frame::BlockLogical>> logical_point{};
   if (block.is_time_dependent()) {
     if constexpr (std::is_same_v<Frame, ::Frame::Inertial>) {
@@ -143,11 +134,11 @@ block_logical_coordinates_single_point(
 }
 
 template <size_t Dim, typename Frame>
-std::vector<block_logical_coord_holder<Dim>> block_logical_coordinates(
+std::vector<BlockLogicalCoords<Dim>> block_logical_coordinates(
     const Domain<Dim>& domain, const tnsr::I<DataVector, Dim, Frame>& x,
-    const double time, const functions_of_time_type& functions_of_time) {
+    const double time, const domain::FunctionsOfTimeMap& functions_of_time) {
   const size_t num_pts = get<0>(x).size();
-  std::vector<block_logical_coord_holder<Dim>> block_coord_holders(num_pts);
+  std::vector<BlockLogicalCoords<Dim>> block_coord_holders(num_pts);
   for (size_t s = 0; s < num_pts; ++s) {
     tnsr::I<double, Dim, Frame> x_frame(0.0);
     for (size_t d = 0; d < Dim; ++d) {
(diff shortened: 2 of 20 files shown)


cmake/SetupPch.cmake
```diff
@@ -50,6 +50,9 @@ target_link_libraries(
   hdf5::hdf5
   SpectreFlags
   )
+if(TARGET OpenMP::OpenMP_CXX)
+  target_link_libraries(SpectrePchFlags INTERFACE OpenMP::OpenMP_CXX)
```
wthrowe: Have you measured the overhead of this?  I would not expect running OpenMP sections with a single thread to be as fast as compiling them without OpenMP, but I don't know if the change is significant for us.

wthrowe (changes requested):
The code looks fine, aside from the things CI says.  I am worried about performance implications if you haven't benchmarked, though.

cmake/SetupPch.cmake
nilsvu: I added a benchmark with some results. I didn't find any measurable overhead when running on a single thread with OpenMP linked.

wthrowe: Sounds good, then.

wthrowe (approved):

wthrowe (changes requested):
Based on CI, this appears to have caused a significant slowdown in the Unit.NumericalAlgorithms.Interpolator.InterpolateEventNoInterpolator test.

knelli2:
Let's hold off on merging this until @nilsvu does some testing on mbot. It's possible the slow interpolation is a CaltechHPC thing and not our algorithm

nilsvu:
This fixup should fix it @wthrowe.

@knelli2 This is independent and can be merged. This PR adds the OMP parallelization for pybindings and exporter. Probably won't be used in our parallel executables.

cmake/SpectreAddCatchTests.cmake
```diff
@@ -42,11 +42,15 @@ add_custom_target(unit-tests)
...
 # - Disable ASAN's leak sanitizer because Charm++ has false positives
-list(APPEND _TEST_ENV_VARS "ASAN_OPTIONS=detect_leaks=0")
+list(APPEND _CATCH_TEST_ENV_VARS "ASAN_OPTIONS=detect_leaks=0")
 # - Set PYTHONPATH to find Python modules
-list(APPEND _TEST_ENV_VARS "PYTHONPATH=${PYTHONPATH}")
+list(APPEND _CATCH_TEST_ENV_VARS "PYTHONPATH=${PYTHONPATH}")
+# - Disable OpenMP multithreading in ctest because it runs multiple tests in
+#   parallel. Still allows to set this environment variable in the shell when
+#   running test executables manually to benchmark multithreaded performance.
+list(APPEND _CATCH_TEST_ENV_VARS "OMP_NUM_THREADS=1")
```
wthrowe: I don't like this solution, but I haven't thought of a better one so far.  This means that running the test executables directly isn't going to do what they are intended to do or what people expect, and there's not going to be an easy way for people to figure out what's going on.

I've been confused in similar ways in the past by tests that do benchmarking, but at least for those the benchmarking code is obvious in the test.

nilsvu: The alternative is to disable multithreading in the test main function, which is very easy to do (I can push a quick commit tomorrow). I didn't do that because I wanted to control the benchmark with the environment variable, but it's certainly more predictable. I can add a line to the benchmark to set the num threads at compile time.

nilsvu: pushed

nilsdeppe (changes requested):
I'm blocking so I can look at this. A few quick notes:

1. I remember we discussed in the past not linking the entire code base against OpenMP.
2. It was definitely not communicated at all that everything would depend on OpenMP. I'm not sure I like the idea of require OpenMP everywhere. 
3. This also seems much more complicated than originally advertised, but I'll need to look more carefullly..

src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -143,11 +141,23 @@ block_logical_coordinates_single_point(
...
+  if (num_threads.has_value()) {
+    ERROR("OpenMP is not available, so num_threads cannot be specified.");
+  }
+#endif  // _OPENMP
+
   const size_t num_pts = get<0>(x).size();
-  std::vector<block_logical_coord_holder<Dim>> block_coord_holders(num_pts);
+  std::vector<BlockLogicalCoords<Dim>> block_coord_holders(num_pts);
+#pragma omp parallel for num_threads(resolved_num_threads)
   for (size_t s = 0; s < num_pts; ++s) {
```
nilsdeppe, line 142: Did you ever try just doing:
```cpp
#include <thread>

const auto lambda = [&x, time, &functions_of_time](const size_t start, const size_t end) {
  for (size_t s = start; s < end; ++s) {
    tnsr::I<double, Dim, Frame> x_frame(0.0);
    for (size_t d = 0; d < Dim; ++d) {
      x_frame.get(d) = x.get(d)[s];
    }
    // Check which block this point is in. Each point will be in one
    // and only one block, unless it is on a shared boundary.  In that
    // case, choose the first matching block (and this block will have
    // the smallest block_id).
    for (const auto& block : domain.blocks()) {
      std::optional<tnsr::I<double, Dim, ::Frame::BlockLogical>> x_logical =
          block_logical_coordinates_single_point(x_frame, block, time,
                                                 functions_of_time);
      if (x_logical.has_value()) {
        // Point is in this block.  Don't bother checking subsequent
        // blocks.
        block_coord_holders[s] = make_id_pair(domain::BlockId(block.id()),
                                              std::move(x_logical.value()));
        break;
      }
    }
  }
};

if (num_threads.value_or(0) == 0) {
  lambda(0, num_pts);
} else {
  std::vector<std::thread> threads(num_threads.value());
  // Spawn threads to do the interpolation.
  for (auto& thread : threads) {
    // Note: you'll want to do the math so that the remainder grid points are
    // approximately evenly distributed.
    thread = std::thread{lambda, this_thread_start, this_thread_end};
  }
  // Wait for threads to finish the interpolation.
  for (auto& thread : threads) {
    thread.join();
  }
}
```
I'm not sure if you'll get "perfect strong scaling", and it's more than "just a pragma", but it also doesn't require _any_ changes to other files, no env vars, etc. Plus, this doesn't even require a compiler that supports OpenMP (for a long time Clang didn't!)

nilsvu: Thanks for offering the suggestion! Seems very reasonable. OpenMP is so easy though :D The parallelization of `block_logical_coordinates` is indeed just one line ("as advertised"). The parallelization of the exporter lib is a bit more complicated (still very easy though), but that's a separate lib that's only used for pybindings and exporter lib.

I limited linking OpenMP out of an abundance of caution before. I think that's still possible by compiling a parallelized version of `block_logical_coordinates` into the exporter lib, and leaving the domain lib untouched. I can try that.

Note that this parallelization speeds up the exporter by a factor 10 at least, and this is actually very noticeable when interpolating data for plots or analysis, and essential for importing data into another code. That's why I'd like to find a path for merging this.

I think independent of this we should disable OpenMP multithreading in our executables, so no dependencies accidentally try to do multithreading (like OpenBLAS, which we already disable, or libsharp). I think I'll just add a line `OMP_NUM_THREADS=1` to our submit script template, just to be safe.

nilsdeppe: You're welcome! I don't see how you get to 1 line. The OpenMP-specific changes in BlockLogicalCoordinates are
```cpp
#ifdef _OPENMP
#include <omp.h>
#endif  // _OPENMP

// Ignore OpenMP pragmas when OpenMP is not enabled
#pragma GCC diagnostic push
#pragma GCC diagnostic ignored "-Wunknown-pragmas"

  // Resolve number of threads to use in OpenMP parallelization
#ifdef _OPENMP
  const size_t resolved_num_threads =
      num_threads.value_or(omp_get_max_threads());
#else
  if (num_threads.has_value()) {
    ERROR("OpenMP is not available, so num_threads cannot be specified.");
  }
#endif  // _OPENMP

#pragma omp parallel for num_threads(resolved_num_threads)

#pragma GCC diagnostic pop
```
I count ~17 lines of code. While not huge, these are somewhat annoying "if" macros, which are hard to read. Macros also do not play well with C++ modules, so this adds a potential future challenge. I've also not had great success with the `-Wunknown-pragmas` ignoring on mbot (I haven't looked into why) where the warnings are printed. Furthermore, as Will pointed out, now the tests have to have an env variable set.

Additionally, putting another parallelization method into the code that literally over 99% of the source tree doesn't use is something I really don't like. Interoperability with other frameworks is always difficult, and once we merge something like this, it becomes someone else's problem if they find issues. I've had this happen several times and it's cost me a lot of time. During our telecon the discussion that led to "yes, go ahead with OpenMP" was restricted to the exporter and clearly stated that nothing else should depend on OpenMP. Thus, I currently don't see a path forward with linking OpenMP into the entire code base that I would be comfortable with. C++ threads are a language feature and control of them is very easy. I'll note that the diff I gave you has roughly the same number of lines changed in the file, the lines are contained to 1 place (excluding the include), there are no macros (good for C++ modules compatibility), and still runs in parallel. You might still get a 10x speedup, and the entire diff could shrink by a non-trivial amount. Certainly the CMake would be a lot easier.

Adding the line to submit scripts seems fine. We currently build libsharp without OpenMP and at least I try to use serial OpenBLAS implementations in addition to us disabling it explicitly for OpenBLAS.

nilsvu: As noted above I think it's possible to remove the OpenMP linking from the domain lib (and thus also from the PCH and the rest of the code). I may not have written this clearly above. I can move the loop with its parallelization into the exporter lib and only call into the domain lib in the body of the loop. Then there's no parallelization in the domain lib at all.

nilsdeppe: I think if you restrict the OMP pragmas to only the exporter lib it will be fine. Given the way Kokkos works, I will likely have Kokkos run in OpenMP mode while I debug our implementation before trying to run it on a GPU. That means if _any_ code that we use in our executables uses OpenMP, there's a problem. That means the domain files cannot have any OpenMP includes or directives, etc. in them.

nilsvu:
Updated. Please take another look @wthrowe @nilsdeppe.

nilsdeppe (approved):
Yep, this looks good to me! Thank you!

wthrowe (approved):

PR #5956, opened 2024-04-30
Download dependencies on demand

nilsvu:
## Proposed changes

This seems to work ok... putting this out here to get feedback. This approach could be used for some of the easy to install but tedious dependencies.

Suggestion by @nilsdeppe and me:
 
Bundle in `external`:

- [x] Brigand (#5967)
- [x] libsharp (#5967)

Try CMake FetchContent (if not found on system):

- [x] Blaze
- [x] Catch2
- [x] libxsmm
- [x] yaml-cpp
- [x] Python dependencies (set `BOOTSTRAP_PY_DEPS=ON`)

System modules:

- git
- cmake
- gsl
- hdf5
- boost
- mpi
- python

Install manually:

- Charm++

Optional:

- jemalloc
- libbacktrace

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
<s>Also the py env can be added by cmake I think. Either do `python -m venv build_dir/env` or (maybe even easier) install py deps into the build directory and set the pythonpath.</s> Edit: 🤦‍♂️ I actually added this feature a long time ago: `BOOTSTRAP_PY_DEPS`.

wthrowe:
Looking at CI, it failed to detect that the dependency was already installed, tried and failed to download a new copy, and then didn't give an error message saying why it failed to download.  This would not be a good user experience.

If we do do this, it should be opt-in, because in most circumstances it seems like a bad idea.  I assume this will rebuild the dependencies for every SpECTRE build-directory, so you wouldn't want to use it on a development machine or a supported cluster.

nilsdeppe:
So we tried something like this many years ago, maybe 8-ish, with dg-charm. I don't remember if we always auto-downloaded or if you could point to existing installations. However, universally it wasn't a very good experience (probably more accurately described as terrible) and we intentionally moved away from it. I think we actually took the exact approach done here, and one of the issues was a proliferation of libs. E.g. if you have different build directories you'd build the deps once per build directory. Maybe this can be done better nowadays, I don't know. Another struggle then was that you could get into the state where you needed to redownload the deps even though nothing really changed (maybe they somehow got deleted? I don't remember). This would be very inconvenient if you are without internet, eg on an airplane. These factors together made us abandon this approach back then. Again, maybe it's better nowadays.

nilsvu:
I think everything you two point out is true. Except note that deps are only fetched if not found on the system, so you can still install manually.

I'm just trying to find a solution for the trouble we have building the dependency stack on a new machine. It's so tedious and error-prone to go through each of the dependencies on the installation page, look up installation instructions in their various (existing or nonexistent) documentations, make sure important flags are set like `-fPIC` (I regularly get messages about that and have to tell people to rebuild yaml-cpp with `-fPIC` or shared libs, even though there's a note on that on the installation page), and move the installation artifacts to good locations on whichever system you're on.

The only alternative I can think of really is Spack, which can be annoying in different ways (often builds loads of low-level system packages, and has a bit of a learning curve as well).

nilsdeppe:
Yes, I completely agree! I didn't mean to sound negative about this approach! I wanted to share what I could remember from when we tried it. Have you looked at what tools like Conan do to try to add "package management" to CMake? I'm not sure how well those types of tools work, but could at least be worth reading their overview :Shrug:

wthrowe:
I think the correct solution on, say, a national supercomputer is probably very different from the correct solution on a developer's laptop, and it's not clear which problem you are trying to solve here.  What cases are you trying to address?

nilsvu:
> which problem you are trying to solve here

Getting started quickly on any machine, supercomputer or not.

wthrowe:
As I said in my previous comment, I don't think that problem has a general solution.  From your comments, it sounds like you've seen common trouble cases that we should work on, but if you won't tell me what they are I can't help.

nilsvu:
Rebased. This is ready for review / testing.

wthrowe:
Since my input was explicitly requested through other channels:

I still think this is a bad idea, but I won't try to block it as long as it is off by default.

nilsdeppe (dismissed):
This works great. I tested it on mbot, both with modules loaded and without. It fetches the content as needed.

One optional consideration: yaml-cpp and catch2 give warnings like:
```
 CMake Warning at cmake/SetupYamlCpp.cmake:4 (find_package):
   By not providing "Findyaml-cpp.cmake" in CMAKE_MODULE_PATH this project has
   asked CMake to find a package configuration file provided by "yaml-cpp",
   but CMake did not find one.

   Could not find a package configuration file provided by "yaml-cpp" with any
   of the following names:

     yaml-cppConfig.cmake
     yaml-cpp-config.cmake

   Add the installation prefix of "yaml-cpp" to CMAKE_PREFIX_PATH or set
   "yaml-cpp_DIR" to a directory containing one of the above files.  If
   "yaml-cpp" provides a separate development package or SDK, be sure it has
   been installed.
 Call Stack (most recent call first):
   CMakeLists.txt:162 (include)
```

I've found that this works at least with CMake 3.24 and newer:
```cmake
if (SPECTRE_FETCH_MISSING_DEPS)
  include(FetchContent)
  FetchContent_Declare(Catch2
    GIT_REPOSITORY https://github.com/catchorg/Catch2.git
    GIT_TAG v3.4.0
    GIT_SHALLOW TRUE
    ${SPECTRE_FETCHCONTENT_BASE_ARGS}
    FIND_PACKAGE_ARGS
  )
  FetchContent_MakeAvailable(Catch2)
else()
  find_package(Catch2 3.4.0)
  if (NOT Catch2_FOUND AND NOT SPECTRE_FETCH_MISSING_DEPS)
    message(FATAL_ERROR "Could not find Catch2. If you want to fetch "
      "missing dependencies automatically, set SPECTRE_FETCH_MISSING_DEPS=ON.")
  endif()
endif()
```
Another option could be to set `QUIET` as the flag to `find_package` (maybe only if fetch content is on?)

yaml-cpp also gives me:
```
 CMake Deprecation Warning at /home/fs01/nd357/spectre_build/clang_cmake_fetch/_deps/yaml-cpp-src/CMakeLists.txt:2 (cmake_minimum_required):
   Compatibility with CMake < 3.5 will be removed from a future version of
   CMake.

   Update the VERSION argument <min> value or use a ...<max> suffix to tell
   CMake that the project does not need compatibility with older versions.
```
I'm confused because I think we require like 3.18 or something

PR #5953, opened 2024-04-29
Binary Input File Tuning

AlexCarpenter46:
## Proposed changes

Adjusting the input files to be more realistic. 

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #5957, opened 2024-04-30
Automate choice of some params in BBH pipeline

knelli2:
## Proposed changes

Some of the initial parameters in our input file depend on the parameters used in the ID solve, and it'd be very tedious to try and update these manually in our pipeline. This PR adds logic to determine some constraint damping parameters and control system parameters automatically based on the masses, mass ratio, spins, and separation of the binary. These formulas were taken from `DoMultipleRuns.input` in SpEC.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

This PR should be merged after #5953

### DIFF
--- a/support/Pipelines/Bbh/Inspiral.py
+++ b/support/Pipelines/Bbh/Inspiral.py
@@ -2,7 +2,6 @@
 # See LICENSE.txt for details.
 
 import logging
-import subprocess
 from pathlib import Path
 from typing import Optional, Union
 
@@ -10,11 +9,52 @@ import click
 import yaml
 from rich.pretty import pretty_repr
 
+import spectre.IO.H5 as spectre_h5
 from spectre.support.Schedule import schedule, scheduler_options
+from spectre.Visualization.ReadH5 import to_dataframe
 
 logger = logging.getLogger(__name__)
 
 INSPIRAL_INPUT_FILE_TEMPLATE = Path(__file__).parent / "Inspiral.yaml"
+
+
+# These parameters come from empirically tested values in SpEC and SpECTRE
+def _control_system_params(
+    mass_left: float,
+    mass_right: float,
+    spin_magnitude_left: float,
+    spin_magnitude_right: float,
+) -> dict:
+    total_mass = mass_left + mass_right
+    mass_ratio = max(mass_left, mass_right) / min(mass_left, mass_right)
+    if spin_magnitude_left > 0.9 or spin_magnitude_right > 0.9:
+        damping_time_base = 0.1
+        decrease_threshold_base = 2e-4
+        max_damping_timescale = 10.0
+    else:
+        damping_time_base = 0.2
+        decrease_threshold_base = 2e-3
+        max_damping_timescale = 20.0
+
+    kinematic_timescale = damping_time_base * total_mass
+    decrease_threshold = (
+        0.1 * decrease_threshold_base / (mass_ratio + 1.0 / mass_ratio)
+    )
+    increase_threshold_fraction = 0.25
+
+    return {
+        "MaxDampingTimescale": max_damping_timescale,
+        "KinematicTimescale": kinematic_timescale,
+        "SizeATimescale": damping_time_base * 0.2 * mass_right,
+        "SizeBTimescale": damping_time_base * 0.2 * mass_left,
+        "ShapeATimescale": 5.0 * kinematic_timescale,
+        "ShapeBTimescale": 5.0 * kinematic_timescale,
+        "SizeIncreaseThreshold": 1e-3,
+        "DecreaseThreshold": decrease_threshold,
+        "IncreaseThreshold": increase_threshold_fraction * decrease_threshold,
+        "SizeBMaxTimescale": 10 if spin_magnitude_left > 0.9 else 20,
+        "SizeAMaxTimescale": 10 if spin_magnitude_right > 0.9 else 20,
+    }
 
 
 def inspiral_parameters(
@@ -36,7 +76,31 @@ def inspiral_parameters(
     """
     id_domain_creator = id_input_file["DomainCreator"]["BinaryCompactObject"]
     id_binary = id_input_file["Background"]["Binary"]
-    return {
+
+    # ID parameters
+    horizons_filename = Path(id_run_dir) / "Horizons.h5"
+    with spectre_h5.H5File(
+        str(horizons_filename.resolve()), "r"
+    ) as horizons_file:
+        aha_quantities = to_dataframe(horizons_file.get_dat("AhA.dat")).iloc[-1]
+        mass_right = aha_quantities["ChristodoulouMass"]
+        spin_magnitude_right = aha_quantities["DimensionlessSpinMagnitude"]
+
+        horizons_file.close_current_object()
+        ahb_quantities = to_dataframe(horizons_file.get_dat("AhB.dat")).iloc[-1]
+        mass_left = ahb_quantities["ChristodoulouMass"]
+        spin_magnitude_left = ahb_quantities["DimensionlessSpinMagnitude"]
+
+    # Uncomment when we are confident that we can make the total mass 1. For
+    # now, allow total mass != 1
+    # total_mass = (
+    #     horizon_quantities_A["ChristodoulouMass"]
+    #     + horizon_quantities_B["ChristodoulouMass"]
+    # )
+    # if total_mass != 1.0:
+    #     raise ValueError(f"Total mass must 1.0, not {total_mass}.")
+
+    params = {
         # Initial data files
         "IdFileGlob": str(
             Path(id_run_dir).resolve()
@@ -54,6 +118,18 @@ def inspiral_parameters(
         "L": refinement_level,
         "P": polynomial_order,
     }
+
+    # Control system
+    params.update(
+        _control_system_params(
+            mass_left=mass_left,
+            mass_right=mass_right,
+            spin_magnitude_left=spin_magnitude_left,
+            spin_magnitude_right=spin_magnitude_right,
+        )
+    )
+
+    return params
 
 
 def _load_spec_id_params(id_params_file: Path) -> dict:
@@ -94,7 +170,13 @@ def inspiral_parameters_spec(
       refinement_level: h-refinement level.
       polynomial_order: p-refinement level.
     """
-    return {
+
+    mass_left = id_params["ID_MB"]
+    mass_right = id_params["ID_MA"]
+    spin_magnitude_left = id_params["ID_chiBMagnitude"]
+    spin_magnitude_right = id_params["ID_chiAMagnitude"]
+
+    params = {
         # Initial data files
         "SpecDataDirectory": str(Path(id_run_dir).resolve()),
         # Domain geometry
@@ -110,6 +192,18 @@ def inspiral_parameters_spec(
         "L": refinement_level,
         "P": polynomial_order,
     }
+
+    # Control system
+    params.update(
+        _control_system_params(
+            mass_left=mass_left,
+            mass_right=mass_right,
+            spin_magnitude_left=spin_magnitude_left,
+            spin_magnitude_right=spin_magnitude_right,
+        )
+    )
+
+    return params
 
 
 def start_inspiral(
--- a/support/Pipelines/Bbh/Inspiral.py
+++ b/support/Pipelines/Bbh/Inspiral.py
@@ -57,6 +57,24 @@ def _control_system_params(
     }
 
 
+def _constraint_damping_params(
+    mass_left: float,
+    mass_right: float,
+    initial_separation: float,
+) -> dict:
+    total_mass = mass_left + mass_right
+    return {
+        "Gamma0Constant": 0.001 / total_mass,
+        "Gamma0LeftAmplitude": 4.0 / mass_left,
+        "Gamma0LeftWidth": 7.0 * mass_left,
+        "Gamma0RightAmplitude": 4.0 / mass_right,
+        "Gamma0RightWidth": 7.0 * mass_right,
+        "Gamma0OriginAmplitude": 0.075 / total_mass,
+        "Gamma0OriginWidth": 2.5 * initial_separation,
+        "Gamma1Width": 10.0 * initial_separation,
+    }
+
+
 def inspiral_parameters(
     id_input_file: dict,
     id_run_dir: Union[str, Path],
@@ -79,6 +97,10 @@ def inspiral_parameters(
 
     # ID parameters
     horizons_filename = Path(id_run_dir) / "Horizons.h5"
+    initial_separation = (
+        id_domain_creator["ObjectA"]["XCoord"]
+        - id_domain_creator["ObjectB"]["XCoord"]
+    )
     with spectre_h5.H5File(
         str(horizons_filename.resolve()), "r"
     ) as horizons_file:
@@ -118,6 +140,15 @@ def inspiral_parameters(
         "L": refinement_level,
         "P": polynomial_order,
     }
+
+    # Constraint damping parameters
+    params.update(
+        _constraint_damping_params(
+            mass_left=mass_left,
+            mass_right=mass_right,
+            initial_separation=initial_separation,
+        )
+    )
 
     # Control system
     params.update(
@@ -175,6 +206,7 @@ def inspiral_parameters_spec(
     mass_right = id_params["ID_MA"]
     spin_magnitude_left = id_params["ID_chiBMagnitude"]
     spin_magnitude_right = id_params["ID_chiAMagnitude"]
+    initial_separation = id_params["ID_d"]
 
     params = {
         # Initial data files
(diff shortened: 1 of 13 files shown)


support/Pipelines/Bbh/Inspiral.py
```diff
@@ -35,6 +36,89 @@ def inspiral_parameters(
...
+    )
+    total_mass = mass_left + mass_right
+    mass_ratio = max(mass_left, mass_right) / min(mass_left, mass_right)
+    center_left, center_right = id_binary["XCoords"]
+    initial_separation = np.linalg.norm(
+        np.array(center_right) - np.array(center_left)
+    )
+
+    ###########################################################################
+    # Control system parameters
```
nilsvu: How about this for structuring the code better than SpEC: You put each of these sections into their own function that returns a dict. Then you call them like this:

```py
params = {
  # ...
}
params.update(
  control_system_parameters(total_mass=total_mass, ...)
)
return params
```

support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -130,42 +130,28 @@ EvolutionSystem:
         SpatialDecayWidth: 17.0152695482514 # From SpEC run: 100.0 / sqrt(34.54)
         Amplitudes: [1.0, 1.0, 1.0]         # From SpEC run: damped harmonic
         Exponents: [2, 2, 2]                # From SpEC run
-    DampingFunctionGamma0:
+    DampingFunctionGamma0: &ConstraintDampingTripleGaussian
       TimeDependentTripleGaussian:
-        Constant: 0.001             # 0.001 / (m_A + m_B)
+        Constant: {{ Gamma0Constant }}
```
nilsvu, line 180: Note that you _can_ do simple math in these templates, e.g. `Constant: {{ 0.001 / (m_A + m_B) }}`. Maybe that's better than doing it in Python and passing it down? It would certainly make easy expressions a lot less verbose, and probably easier to understand.

nilsvu (changes requested):

support/Pipelines/Bbh/Inspiral.yaml
knelli2: I saw we already did that for some other parameters in the input file and I didn't like it because it's basically a python variable in a yaml file. And if we have math in the yaml like that it's a bit confusing. I'd prefer to do all the math in the python script and just treat the jinja templates as a find/replace with values. I think that's clearer for users if they ever look at these files (and clearer for me). Only one place to look for math rather than scattered through an input file

knelli2:
Rebased and squashed in the suggestion about factoring out functions

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -16,6 +17,66 @@
...
+) -> dict:
+    damping_time_base = 0.2
+    decrease_threshold_base = 2e-3
+    max_damping_timescale = 20.0
+    if spin_magnitude_left > 0.9 or spin_magnitude_right > 0.9:
+        damping_time_base = 0.1
+        decrease_threshold_base = 2e-4
+        max_damping_timescale = 10.0
+
+    params = {}
```
nilsvu: (nitpick) For consistency with the code below I'd do this syntax, but that's just personal preference:

```py
return {
  "MaxDampingTimescale": max_damping_timescale,
  "KinematicTimescale": damping_time_base * total_mass,
  # ...
}
```

Have to precompute values used multiple times, but maybe that's better anyway.

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -35,7 +96,51 @@ def inspiral_parameters(
     """
     id_domain_creator = id_input_file["DomainCreator"]["BinaryCompactObject"]
     id_binary = id_input_file["Background"]["Binary"]
-    return {
+    left_object = id_binary["ObjectLeft"]
+    right_object = id_binary["ObjectRight"]
+
+    assert (
+        len(left_object.keys()) == 1
```
nilsvu: (nitpick) don't need the `.keys()`

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -35,7 +96,51 @@ def inspiral_parameters(
...
+    ), "Expected only one option for 'ObjectLeft'"
+    assert (
+        len(right_object.keys()) == 1
+    ), "Expected only one option for 'ObjectRight'"
+
+    left_object_type, left_object_params = list(left_object.items())[0]
+    right_object_type, right_object_params = list(right_object.items())[0]
+
+    available_solutions = [
```
nilsvu: `supported_solutions`?

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -35,7 +96,51 @@ def inspiral_parameters(
...
+    )
+    spin_magnitude_right = np.linalg.norm(
+        right_object_params.get("Spin", [0.0, 0.0, 0.0])
+    )
+    total_mass = mass_left + mass_right
+    mass_ratio = max(mass_left, mass_right) / min(mass_left, mass_right)
+    center_left, center_right = id_binary["XCoords"]
+    initial_separation = np.linalg.norm(
+        np.array(center_right) - np.array(center_left)
+    )
```
nilsvu: (nitpick) can be just `center_right - center_left`

support/Pipelines/Bbh/Inspiral.yaml
nilsvu: 🤷‍♂️ don't know we can try it this way. I don't think doing easy math in the input file template is confusing, and it would reduce the number of parameters being computed and passed to the template. But I'm fine with doing it either way for now.

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -16,6 +17,62 @@
...
+    mass_right: float,
+    total_mass: float,
+    mass_ratio: float,
+    spin_magnitude_left: float,
+    spin_magnitude_right: float,
+) -> dict:
+    damping_time_base = 0.2
+    decrease_threshold_base = 2e-3
+    max_damping_timescale = 20.0
+    if spin_magnitude_left > 0.9 or spin_magnitude_right > 0.9:
```
nilsvu, line 30: (more nitpicks) use `if else` here. In Python the scoping is more relaxed than in C++, so you don't need to define the variables before the `if`.

nilsvu (dismissed):
you can squash

nilsvu (dismissed):

nilsvu:
@markscheel @geoffrey4444 could one of you take a look at this as well?

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -16,6 +17,63 @@
 INSPIRAL_INPUT_FILE_TEMPLATE = Path(__file__).parent / "Inspiral.yaml"
 
 
+# These parameters come from empirically tested values in SpEC and SpECTRE
+def _control_system_params(
+    mass_left: float,
+    mass_right: float,
+    total_mass: float,
+    mass_ratio: float,
```
markscheel: Do we need these as parameters since they are functions of mass_left and mass_right?
(a potential bug is someone might pass in a `total_mass` that is not equal to `mass_left+mass_right`, and this relationship isn't tested in this function.)  If we really want all 4 parameters, then there should be `assert`s that they are correct.

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -16,6 +17,63 @@
...
+        "IncreaseThreshold": increase_threshold_fraction * decrease_threshold,
+        "SizeBMaxTimescale": 10 if spin_magnitude_left > 0.9 else 20,
+        "SizeAMaxTimescale": 10 if spin_magnitude_right > 0.9 else 20,
+    }
+
+
+def _constraint_damping_params(
+    mass_left: float,
+    mass_right: float,
+    total_mass: float,
```
markscheel: Similar; do we need total_mass?

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -35,7 +93,45 @@ def inspiral_parameters(
...
+        f" solutions are: {supported_solutions}"
+    )
+    assert right_object_type in supported_solutions, (
+        f"Unknown solution for 'ObjectRight': '{right_object_type}'. Supported"
+        f" solutions are: {supported_solutions}"
+    )
+
+    # ID parameters
+    mass_left = left_object_params["Mass"]
+    mass_right = right_object_params["Mass"]
```
knelli2: These masses are incorrect. They are the conformal masses for the Xcts solve. We actually need the Christodoulou/ADM masses for this. We can only get those from horizon finds after the ID solve.

I've been testing some runs with these (incorrect) masses and they don't matter too much for the control system params, but they are bad for the constraint damping params. My bbh runs die quickly. So I think we should hold off on this temporarily until we have horizon finds after the ID solve. Either I or @nilsvu probably need to do that.

support/Pipelines/Bbh/Inspiral.py
knelli2: See my other comment. But yeah we should probably only take mass left/right and just recompute the total mass and mass ratio.

support/Pipelines/Bbh/Inspiral.py
nilsvu: Yea I'll get you the horizon masses, though for equal-mass nonspinning you know they are ~0.5. Does the ~20% difference to the conformal masses matter so much for the constraint damping? Does the constraint damping work if you set the masses to 0.5?

knelli2: Yes it matters quite a bit for the constraint damping parameters. Using the correct masses of 0.5 allowed the simulation to run much longer than with the incorrect masses.

support/Pipelines/Bbh/Inspiral.yaml
geoffrey4444: As someone who in SpEC took every opportunity to do things in input files instead of code, I agree with Kyle: I've learned my lesson, and it's better to discourage users from doing math not in code, because it's easier for them to introduce mistakes, there are often fewer checks and safeguards, and it's more likely to end up with a yaml file not in version control than code

support/Pipelines/Bbh/Inspiral.py
geoffrey4444: I agree with @knelli2 that you need to set the black-hole masses from the results of a horizon find, not the conformal masses

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -35,7 +93,45 @@ def inspiral_parameters(
...
+    spin_magnitude_left = np.linalg.norm(
+        left_object_params.get("Spin", [0.0, 0.0, 0.0])
+    )
+    spin_magnitude_right = np.linalg.norm(
+        right_object_params.get("Spin", [0.0, 0.0, 0.0])
+    )
```
geoffrey4444, line 103: I'm not sure where you are using the spins, but these should be the true spins from a horizon find, not the bare, conformal spins

geoffrey4444 (changes requested):

support/Pipelines/Bbh/Inspiral.py
knelli2: They are used for some control system parameters above. But yes I agree, this is included with using the correct masses from the horizon finds.

support/Pipelines/Bbh/Inspiral.py
nilsvu: I have a branch to get horizon quantities from initial data here: https://github.com/nilsvu/spectre/commits/id_horizon_quantities/

You can call `find_horizon` on the initial data here in this function to get the quantities. We can also store these alongside the initial data later, maybe in the reductions file.

nilsvu: Here's the PR to get the horizon masses and spins: https://github.com/sxs-collaboration/spectre/pull/6024

knelli2: Once #6024 is merged, I'll use that here to get the correct masses and spins

nilsvu: #6024 is merged now, so you can go ahead with this 👍

knelli2:
@nilsvu Rebased and just combined changes since a decent amount changed. The tests will fail and I'm not really sure how to test this now though since it requires volume data to be written and be correct enough to find two horizons...

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -3,23 +3,161 @@
...
+from spectre.SphericalHarmonics import Strahlkorper
 from spectre.support.Schedule import schedule, scheduler_options
+from spectre.Visualization.ReadH5 import select_observation
 
 logger = logging.getLogger(__name__)
 
 INSPIRAL_INPUT_FILE_TEMPLATE = Path(__file__).parent / "Inspiral.yaml"
 
 
+def _find_horizon(
```
nilsvu: I moved this to a `postprocess_id` step that runs automatically after ID generation: #6082. Then the masses and spins are just stored alongside the ID in `BbhReductions.h5:Ah{A,B}.dat` and you can just read them here. That's better, right?

knelli2: Naively, it does seem better. But we should probably discuss more offline about how this post-process step will interact with ID root find, ecc reduction, PB&J (which we'll need to do more planning about), and whether we want yet another step in the pipeline (too few and we don't have enough flexibility; too many and it gets confusing)

knelli2:
Still need to test this in an actual pipeline

support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -106,17 +106,29 @@ DomainCreator:
         InitialValues: [[0.0, 0.0, 0.0], [0.0, 0.0, 0.0], [0.0, 0.0, 0.0]]
       ShapeMapA:
         LMax: &LMax 10
-        InitialValues: Spherical
+        InitialValues:
+          H5Filename: &HorizonsFile {{ HorizonsFile }}
```
nilsvu: Have you tried this? I think this won't be wrapped in quotes, which could be a problem for paths that have spaces, and is definitely a problem in the list of subfile names below.

support/Python/Schedule.py
```diff
@@ -110,9 +111,18 @@ def _copy_submit_script_template(
...
 
 
-# Write `pathlib.Path` objects to YAML as plain strings
-def _path_representer(dumper: yaml.Dumper, path: Path) -> yaml.nodes.ScalarNode:
-    return dumper.represent_scalar("tag:yaml.org,2002:str", str(path))
+# Write following objects to YAML
+#  - `pathlib.Path`
+#  - np.float64
+def _path_representer(dumper: yaml.Dumper, to_dump) -> yaml.nodes.ScalarNode:
+    if isinstance(to_dump, Path):
```
nilsvu: add a separate function for `np.float64` instead of these conditionals. The function is then added with `add_multi_representer` below.

support/Python/Schedule.py
```diff
@@ -110,9 +111,18 @@ def _copy_submit_script_template(
...
+        return dumper.represent_scalar("tag:yaml.org,2002:str", str(to_dump))
+    elif isinstance(to_dump, np.float64):
+        return dumper.represent_scalar(
+            "tag:yaml.org,2002:str", str(float(to_dump))
```
nilsvu: why string? should probably be represented as a float in yaml, like Python's `float`

nilsvu (changes requested):

support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -393,100 +391,108 @@ ControlSystems:
...
+      IncreaseThreshold: *IncreaseThreshold
+      DecreaseThreshold: *DecreaseThreshold
+      IncreaseFactor: *IncreaseFactor
+      DecreaseFactor: *DecreaseFactor
+    ControlError:
+  SizeA:
+    IsActive: true
+    Averager: *SizeShapeAverager
+    Controller: &SizeController
       UpdateFraction: 0.3
```
nilsvu: This was 0.2 before. I think that was important.

nilsvu (changes requested):

support/Python/Schedule.py
knelli2: Because I tried it with just `float` and got an error

nilsvu: If you write it as a string it will read back as a string, which is wrong (and could be achieved simply by casting the float to a string before writing). I think the only acceptable behavior is if the `np.float64` reads back in as a `np.float64` or a `float`.

support/Pipelines/Bbh/Inspiral.yaml
knelli2: Yeah this worked just fine. However, I didn't think about the spaces. I'll put these in quotes to account for that

knelli2:
@nilsvu Ok I think this is ready now. The first commit has what worked for `np.float64`. I also added the control system and constraint damping params to when we are starting from SpEC ID because I realized I had forgot that and it'd fail without it. Note that from SpEC ID, the shape map still starts as a sphere, but this will change in a future PR once I have a lot more time dep option features added to the BCO domains.

nilsvu (approved):

Issue #3886, opened 2022-03-23
UCX error failed to allocate memory


nilsdeppe:
# Bug reports:

The error appears on Frontera with a UCX build and also on Wheeler when using OpenMPI built against UCX. Unfortunately, while mostly reproducible, it is not consistent. The exact error on Wheeler is:
```
[1648009459.439548] [wheeler004:32674:0]           ib_md.c:349  UCX  ERROR ibv_reg_mr(address=0x7fc7603eafe0, length=52848, access=0xf) failed: Resource temporarily unavailable
[1648009459.439628] [wheeler004:32674:0]          ucp_mm.c:164  UCX  ERROR failed to register address 0x7fc7603eafe0 mem_type bit 0x1 length 52848 on md[3]=mlx4_0: Input/output error (md reg_mem_types 0x1)
[1648009459.439636] [wheeler004:32674:0]     ucp_request.c:501  UCX  ERROR failed to register user buffer datatype 0x8 address 0x7fc7603eafe0 len 52848: Input/output error
[wheeler004:32674] pml_ucx.c:860  Error: ucx send failed: Input/output error
[wheeler004:32674] *** An error occurred in MPI_Isend
[wheeler004:32674] *** reported by process [1564344321,1]
[wheeler004:32674] *** on communicator MPI COMMUNICATOR 3 DUP FROM 0
[wheeler004:32674] *** MPI_ERR_OTHER: known error not in list
[wheeler004:32674] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[wheeler004:32674] ***    and potentially your MPI job)
--------------------------------------------------------------------------
MPI_ABORT was invoked on rank 0 in communicator MPI COMMUNICATOR 3 DUP FROM 0
with errorcode 1.

NOTE: invoking MPI_ABORT causes Open MPI to kill all MPI processes.
You may or may not see output from other processes, depending on
exactly when Open MPI kills them.
--------------------------------------------------------------------------
------------- Processor 92 Exiting: Caught Signal ------------
Reason: Terminated
[92] Stack Traceback:
  [92:0] EvolveGhBinaryBlackHole 0x123822b
  [92:1] libpthread.so.0 0x7f54f8973100
  [92:2] libmlx4-rdmav2.so 0x7f54e31c8d03
  [92:3] libmlx4-rdmav2.so 0x7f54e31c8dca
  [92:4] libuct_ib.so.0 0x7f54e8d3638e
[...]

PR #5959, opened 2024-05-01
Fix CaltechHPC environment

nilsvu:
- The previous environment had an issue with loading numeric initial data (took ~40min, now takes 16 seconds).
- It also ran BBH evolutions extremely slowly for some reason (2M in 24 hours).
- It was actually built for cascadelake, not skylake.

Kyle and I reinstalled the environment on skylake, so we should be able to run on all available nodes on CaltechHPC now. I confirmed that the IO issues are fixed and a BBH runs at reasonable speed.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
Marking priority because BBH evolutions are essentially broken on CaltechHPC at the moment.

knelli2:
I'm testing this now. I'll report back once I'm confident everything is working smoothly (or as non-roughly as possible)

support/Environments/caltech_hpc_gcc_skylake.sh
```diff
@@ -28,21 +28,32 @@ spectre_run_cmake() {
...
+          -D BLA_VENDOR=OpenBLAS \
           -D CMAKE_BUILD_TYPE=Release \
+          -D BUILD_DOCS=OFF \
+          -D DEBUG_SYMBOLS=OFF \
           -D MEMORY_ALLOCATOR=JEMALLOC \
           -D BUILD_PYTHON_BINDINGS=ON \
           -D MACHINE=CaltechHpcSkylake \
           -D OVERRIDE_ARCH=skylake \
-          -D ENABLE_PARAVIEW=ON \
+          -D ENABLE_PARAVIEW=OFF \
```
knelli2: Were there issues finding paraview? This is really only an issue for the CLI so not an immediate priority.

support/Environments/caltech_hpc_gcc_skylake.sh
```diff
@@ -13,12 +13,12 @@ spectre_setup_modules() {
 
 spectre_load_modules() {
```
knelli2: [random line] Should we remove the `caltech_hpc_gcc_icelake.sh` env now and rename this one to just `caltech_hpc_gcc.sh`? I think that'll be less confusing for users.

knelli2 (commented):
Ok I have tested this out and am now confident that this will work on all nodes of CaltechHPC (skylake, cascadelake, icelake).

Note that we had to add the environment variable `FI_PROVIDER=tcp` to the `skylake-2024-04` environment because the mpi module kept trying to use UCX for the fabric but because of #3886, we can't use UCX, so we go with TCP instead.

support/Environments/caltech_hpc_gcc_skylake.sh
nilsvu: No issues, I sent a request to the CaltechHPC admins to build ParaView with the same Python etc as the rest of the build. They haven't responded yet.

support/Environments/caltech_hpc_gcc_skylake.sh
nilsvu: Yes I think so. Now or later?

knelli2: I'd say now. Just rip the Band-Aid off. Be sure to also change the cluster installation instructions, maybe adding a little description that whatever node you build on, you can only run on that type of node or newer.

nilsvu: ok done

knelli2 (approved):

PR #5961, opened 2024-05-01
Improve static linking of GSL, HDF5, libgfortran, & libquadmath. Add CMake & executable code to print licenses and copyrights of 3PLs

nilsdeppe:
## Proposed changes

- Fix the order we set up GSL and Blaze
- Allow static linking of GNU Scientific Library
- Fix issues with static linking HDF5
- Allow static linking of libgfortran, libquadmath. I've sometimes hit version issues with libgfortran
- Add third party library (3PL) licenses and CMake support to read them.
- Add flag to executables to print copyright & licenses for 3PLs.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Parallel/Main.hpp
```diff
@@ -395,6 +399,9 @@ Main<Metavariables>::Main(CkArgMsg* msg) {
     if (parsed_command_line_options.count("dump-only") != 0) {
       sys::exit();
     }
+    if (parsed_command_line_options.count("copyright-and-licenses") != 0) {
+      Parallel::printf("%s\n", copyright_and_license_info());
+    }
```
wthrowe: Swap the order of these two ifs.

cmake/SetupLicenseInfo.cmake
```diff
@@ -0,0 +1,74 @@
...
+# See LICENSE.txt for details.
+
+set(_LIST_OF_3PLS
+  Blaze
+  Boost
+  Brigand
+  Catch2
+  Gsl
+  Hdf5
+  Libbacktrace
```
wthrowe, line 12: Should jemalloc be here?  I think that's the only missing link dependency we have listed (other than charm).

wthrowe, line 12: [optional] If you're feeling fancy, you can exclude optional dependencies that weren't enabled.  I don't think always including them is a problem, though.

wthrowe (changes requested):
I believe this will satisfy our license requirements for attribution and such, once the charm situation is dealt with.

cmake/SetupHdf5.cmake
```diff
@@ -38,7 +38,11 @@ endif()
 # are listed here:
 # https://github.com/HDFGroup/hdf5/blob/develop/doc/file-locking.md
 include(CheckCXXSourceCompiles)
-set(CMAKE_REQUIRED_LIBRARIES hdf5::hdf5)
+if(HDF5_USE_STATIC_LIBRARIES)
+  set(CMAKE_REQUIRED_LIBRARIES ${HDF5_C_LIBRARIES})
```
nilsvu: This will break if the HDF5 headers are not in some global include path. You can set `CMAKE_REQUIRED_INCLUDES` as well I think: https://cmake.org/cmake/help/latest/module/CheckCXXSourceCompiles.html.
Why is this necessary? Does the `hdf5::hdf5` target carry the wrong (non-static) libs?

cmake/SetupLicenseInfo.cmake
```diff
@@ -0,0 +1,74 @@
...
+  Xsimd
+  YamlCpp
+)
+
+file(READ "${CMAKE_SOURCE_DIR}/LICENSE.txt" SPECTRE_MIT_LICENSE)
+file(READ "${CMAKE_SOURCE_DIR}/external/Licenses/SpectreGpl3License.txt"
+      SPECTRE_GPL3_LICENSE)
+set(SPECTRE_LICENSE_CXX_STRING "
+SpECTRE source code is distributed under the MIT License, while SpECTRE
+executables and libraries, generally machine code, is distributed and the GNU
```
nilsvu: `and` -> `under`

nilsvu (commented):
Thanks for collecting all the licenses!

cmake/SetupHdf5.cmake
nilsdeppe: Something like that must be the case yes. It failed to link against libz when I used static libs and only did `hdf5::hdf5`

nilsvu: Hm then it should fail in our executables too. My guess is that in the executables zlib is pulled in from charm instead, which also links zlib. Here's an upstream issue: https://forum.hdfgroup.org/t/cmake-zlib-dependency-not-properly-propagted-by-hdf5-targets/5149/8

nilsdeppe: Okay, I'll think about this a bit to see what I can think of in terms of finding a portable solution that doesn't require Charm++ to link in libz. Thanks for finding that issue!

cmake/SetupLicenseInfo.cmake
nilsdeppe: We don't always link it, but I'll add it anyway. That way we don't have to worry about it when distributing executables :)

kidder (commented):
I have no requested changes

cmake/SetupHdf5.cmake
nilsdeppe: Aha, if I link against `HDF5::HDF5` it works but not if I do `hdf5::hdf5`

nilsdeppe:
I rebased and pushed fixups. Thanks for the reviews!

cmake/SetupLicenseInfo.cmake
nilsdeppe: I thought about this before already and then decided I was too lazy 🙃

nilsvu (approved):

wthrowe (approved):
Looks good.  Squash.

PR #5963, opened 2024-05-03
Add Cce H5 file type

knelli2:
## Proposed changes

This new H5 file type will help streamline the usage of waveform output from the CCE executable within the [sxs](https://github.com/sxs-collaboration/sxs) and [scri](https://github.com/moble/scri) python packages.

These new Cce subfile is basically just a group that holds datasets for the bondi variables. The group and the datasets have attributes that the python packages will look for.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/H5/Cce.cpp
+++ b/src/IO/H5/Cce.cpp
@@ -1,0 +1,264 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "IO/H5/Cce.hpp"
+
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <hdf5.h>
+#include <string>
+#include <unordered_map>
+#include <unordered_set>
+#include <vector>
+
+#include "DataStructures/Matrix.hpp"
+#include "IO/H5/AccessType.hpp"
+#include "IO/H5/CheckH5.hpp"
+#include "IO/H5/Header.hpp"
+#include "IO/H5/Helpers.hpp"
+#include "IO/H5/OpenGroup.hpp"
+#include "IO/H5/Version.hpp"
+#include "IO/H5/Wrappers.hpp"
+#include "Utilities/Algorithm.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/MakeString.hpp"
+#include "Utilities/StdHelpers.hpp"
+
+namespace h5 {
+Cce::Cce(const bool exists, detail::OpenGroup&& group, const hid_t /*location*/,
+         const std::string& name, const size_t l_max, const uint32_t version)
+    : group_(std::move(group)),
+      name_(extension() == name.substr(name.size() > extension().size()
+                                           ? name.size() - extension().size()
+                                           : 0)
+                ? name
+                : name + extension()),
+      path_(group_.group_path_with_trailing_slash() + name),
+      version_(version),
+      l_max_(l_max),
+      legend_([this]() {
+        std::vector<std::string> legend;
+        legend.reserve(2 * square(l_max_ + 1) + 1);
+        legend.emplace_back("time");
+        for (int i = 0; i <= static_cast<int>(l_max_); ++i) {
+          for (int j = -i; j <= i; ++j) {
+            legend.push_back(MakeString{} << "Real Y_" << i << "," << j);
+            legend.push_back(MakeString{} << "Imag Y_" << i << "," << j);
+          }
+        }
+        return legend;
+      }()),
+      cce_group_(group_.id(), name_, h5::AccessType::ReadWrite) {
+  for (const std::string& bondi_var : bondi_variables_) {
+    // We will set the id below
+    bondi_datasets_[bondi_var] = DataSet{-1, {0, legend_.size()}};
+  }
+
+  if (exists) {
+    {
+      // We treat this as an internal version for now. We'll need to deal with
+      // proper versioning later.
+
+      // Check if the version exists before calling the open_version.
+      const htri_t version_exists = H5Aexists(cce_group_.id(), "version.ver");
+      if (version_exists != 0) {
+        const Version open_version(true, detail::OpenGroup{}, cce_group_.id(),
+                                   "version");
+        version_ = open_version.get_version();
+      }
+    }
+    {
+      const htri_t header_exists = H5Aexists(cce_group_.id(), "header.hdr");
+      if (header_exists != 0) {
+        const Header header(true, detail::OpenGroup{}, cce_group_.id(),
+                            "header");
+        header_ = header.get_header();
+      }
+    }
+
+    for (const std::string& bondi_var : bondi_variables_) {
+      DataSet& dataset = bondi_datasets_.at(bondi_var);
+      dataset.id =
+          H5Dopen2(cce_group_.id(), bondi_var.c_str(), h5::h5p_default());
+      CHECK_H5(dataset.id, "Failed to open dataset");
+
+      hid_t space_id = H5Dget_space(dataset.id);
+      std::array<hsize_t, 2> max_dims{};
+      if (2 != H5Sget_simple_extent_dims(space_id, dataset.size.data(),
+                                         max_dims.data())) {
+        ERROR("Invalid number of dimensions in cce file " << bondi_var
+                                                          << " on disk.");
+      }
+      CHECK_H5(H5Sclose(space_id), "Failed to close dataspace");
+
+      if (legend_ != read_rank1_attribute<std::string>(dataset.id, "Legend"s)) {
+        ERROR("l_max from cce file " << bondi_var
+                                     << " does not match l_max in constructor");
+      }
+    }
+  } else {  // file does not exist
+    {
+      Version open_version(false, detail::OpenGroup{}, cce_group_.id(),
+                           "version", version_);
+    }
+    {
+      Header header(false, detail::OpenGroup{}, cce_group_.id(), "header");
+      header_ = header.get_header();
+    }
+    // So this file is compatible with the sxs/scri python packages offered by
+    // the SXS collaboration
+    write_to_attribute(cce_group_.id(), sxs_format_str_, sxs_version_str_);
+
+    for (const std::string& bondi_var : bondi_variables_) {
+      DataSet& dataset = bondi_datasets_.at(bondi_var);
+      dataset.id = h5::detail::create_extensible_dataset(
+          cce_group_.id(), bondi_var, dataset.size,
+          std::array<hsize_t, 2>{{4, legend_.size()}},
+          {{h5s_unlimited(), legend_.size()}});
+      CHECK_H5(dataset.id, "Failed to create dataset");
+
+      write_to_attribute(dataset.id, "Legend"s, legend_);
+      // So this dataset is compatible with the sxs/scri python packages offered
+      // by the SXS collaboration
+      write_to_attribute(dataset.id, sxs_format_str_, sxs_version_str_);
+    }
+  }
+}
+
+Cce::~Cce() {
+#if defined(__GNUC__) and not defined(__clang__)
+// The pragma here is used to suppress the warning that the compile time branch
+// of `ERROR` will always call `terminate()` because it throws an error.
+// Throwing that error is a code path that will never actually be entered at
+// runtime, so we suppress the warning here.
+#pragma GCC diagnostic push
+#pragma GCC diagnostic ignored "-Wterminate"
+#endif
+  // We don't use structured bindings because there was a bug that was fixed in
+  // C++20 about capturing structured bindings in a lambda (the lambda is in the
+  // internals of CHECK_H5), so older compilers that we support may not have
+  // fixed this bug.
+  for (const auto& name_and_dataset : bondi_datasets_) {
+    const auto& name = name_and_dataset.first;
+    const auto& dataset = name_and_dataset.second;
+    CHECK_H5(H5Dclose(dataset.id), "Failed to close dataset " << name);
+  }
+#if defined(__GNUC__) and not defined(__clang__)
+#pragma GCC diagnostic pop
+#endif
+}
+
+void Cce::append(
+    const std::unordered_map<std::string, std::vector<double>>& data) {
+  for (const std::string& bondi_var : bondi_variables_) {
+    if (not data.contains(bondi_var)) {
+      ERROR("Passed in data does not contain the bondi variable " << bondi_var);
+    }
+    DataSet& dataset = bondi_datasets_.at(bondi_var);
+
+    const std::vector<double>& vec_data = data.at(bondi_var);
+    if (vec_data.size() != dataset.size[1]) {
+      ERROR("Cannot add columns to Cce files. Current number of columns is "
+            << dataset.size[1] << " but received " << vec_data.size()
+            << " entries.");
+    }
+
+    dataset.size =
+        h5::append_to_dataset(dataset.id, bondi_var, vec_data, 1, dataset.size);
+  }
+}
+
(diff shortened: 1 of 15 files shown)


nilsdeppe (dismissed):

src/IO/H5/Cce.cpp
```diff
@@ -0,0 +1,262 @@
...
+      write_to_attribute(dataset.id, "Legend"s, legend_);
+      // So this dataset is compatible with the sxs/scri python packages offered
+      // by the SXS collaboration
+      write_to_attribute(dataset.id, sxs_format_str_, sxs_version_str_);
+    }
+  }
+}
+
+Cce::~Cce() {
+#ifdef __clang__
```
wthrowe: Why are we avoiding structured bindings with clang and using them with gcc?  If the point of this `#ifdef` is just to protect the pragmas, just put it around them and use the same normal c++ for both versions, otherwise comment what's going on.

src/IO/H5/Cce.cpp
```diff
@@ -0,0 +1,262 @@
...
+  for (const auto& name_and_dataset : bondi_datasets_) {
+    const auto& name = name_and_dataset.first;
+    const auto& dataset = name_and_dataset.second;
+    CHECK_H5(H5Dclose(dataset.id), "Failed to close dataset " << name);
+  }
+#else
+// The pragma here is used to suppress the warning that the compile time branch
+// of `ERROR` will always call `terminate()` because it throws an error.
+// Throwing that error is a code path that will never actually be entered at
+// runtime, so we suppress the warning here.
```
wthrowe, line 135: Why can't it happen?

src/IO/H5/Cce.cpp
```diff
@@ -0,0 +1,262 @@
...
+  if (these_ell.empty()) {
+    return {num_rows, 0, 0.0};
+  }
+
+  // Always grab the time
+  std::vector<size_t> these_columns{0};
+
+  // For a given ell, we nedd the first column index and the last column index.
+  // For a given ell, the total number of coefs is given by
+  // f(l) = 2 * (l + 1)^1 (first factor of 2 comes from having both real and
```
wthrowe: `^1` -> `^2`

tests/Unit/IO/H5/Test_Cce.cpp
```diff
@@ -0,0 +1,579 @@
...
+#include "IO/H5/Wrappers.hpp"
+#include "Informer/InfoFromBuild.hpp"
+#include "Utilities/Algorithm.hpp"
+#include "Utilities/FileSystem.hpp"
+#include "Utilities/Formaline.hpp"
+#include "Utilities/GetOutput.hpp"
+#include "Utilities/MakeString.hpp"
+#include "Utilities/TMPL.hpp"
+
+#include "Parallel/Printf/Printf.hpp"
```
wthrowe: Remove.

tests/Unit/IO/H5/Test_Cce.cpp
```diff
@@ -0,0 +1,579 @@
...
+  CHECK(cce_file.get_version() == version_number);
+
+  // Check getting the header from Cce file
+  std::stringstream ss;
+  ss << "# ";
+  auto build_info = info_from_build();
+  ss << std::regex_replace(build_info, std::regex{"\n"}, "\n# ");
+  const auto& header = cce_file.get_header();
+  CHECK("#\n# File created on "s ==
+        header.substr(0, header.find("File created on ") + 16));
```
wthrowe: I think this is just `header.starts_with("#\n# File created on ")`

tests/Unit/IO/H5/Test_Cce.cpp
```diff
@@ -0,0 +1,579 @@
...
+  CHECK(ss.str() == header.substr(header.find("# SpECTRE")));
```
wthrowe: `header.ends_with(ss.str())`?

knelli2:
Hmm the MacOS CI failed for the test I added. Not really sure how to debug without a Mac...

wthrowe (changes requested):
The test segfaulted on the macOS build.  I didn't catch anything that would cause that reading the code, but it's probably a real failure.

nilsvu:
I can look at the segfault at macos

src/IO/H5/Cce.cpp
knelli2: My compiler (clang 14) complained about trying to capture structured bindings in a lambda (in the ERROR macro I believe, which CHECK_H5 calls). GCC didn't care about this. The `#ifdef` was more for the pragmas. I can just avoid the structured bindings, add a comment explaining why, and only add the `#ifdef` for the pragmas.

src/IO/H5/Cce.cpp
knelli2: To be honest, I just copied this from the `Dat.cpp` file, so not really sure. @macedo22 added this when she made the ERROR macros usable at compile time.

src/IO/H5/Cce.cpp
nilsdeppe: I think there is a C++17-ish bug about this that gets fixed in C++20 (or 23), so whether or not we can directly capture struct bindings in lambdas currently depends on the compiler :(

src/IO/H5/Cce.cpp
macedo22: The warning being suppressed is for the branch taken in `ERROR` when we have a constant evaluated context (the code is being evaluated at compile time). This branch of `ERROR`, by construction, will never run at runtime because in this branch, we will get the error thrown at compile time.

macedo22: See the branch [here](https://github.com/sxs-collaboration/spectre/blob/a9ee6aa7c65ba703dbdc6dfb79436bb5378d4465/src/Utilities/ErrorHandling/Error.hpp#L68)

macedo22: At least, that is my understanding from when we made this change

src/IO/H5/Cce.cpp
wthrowe: Yes, just `#ifdef` the pragmas.

src/IO/H5/Cce.cpp
wthrowe: Ah, right.  That's fine then.

knelli2:
Fixup doesn't include anything for MacOS failure.

wthrowe:
Fixups look good.  Just need to figure out the Mac thing.

nilsvu:
Here's a fix for the macOS issue:

```diff
diff --git a/src/IO/H5/Helpers.cpp b/src/IO/H5/Helpers.cpp
index 6a6b5b6ca5..cf8001cbf2 100644
--- a/src/IO/H5/Helpers.cpp
+++ b/src/IO/H5/Helpers.cpp
@@ -473,8 +473,15 @@ void write_to_attribute(const hid_t location_id, const std::string& name,
   const hid_t att_id = H5Acreate2(location_id, name.c_str(), h5_type<Type>(),
                                   space_id, h5p_default(), h5p_default());
   CHECK_H5(att_id, "Failed to create attribute '" << name << "'");
-  CHECK_H5(H5Awrite(att_id, h5_type<Type>(), static_cast<const void*>(&value)),
-           "Failed to write value: " << value);
+  if constexpr (std::is_same_v<Type, std::string>) {
+    const auto cstr = value.c_str();
+    CHECK_H5(H5Awrite(att_id, h5_type<Type>(), &cstr),
+             "Failed to write value: " << value);
+  } else {
+    CHECK_H5(
+        H5Awrite(att_id, h5_type<Type>(), static_cast<const void*>(&value)),
+        "Failed to write value: " << value);
+  }
   CHECK_H5(H5Aclose(att_id), "Failed to close attribute '" << name << "'");
   CHECK_H5(H5Sclose(space_id), "Unable to close dataspace");
 }
```

Note that we don't currently store any other single string attributes in H5 files, maybe because of issues like this. Even the input file is stores as single-element vector, which seems wrong. Maybe with this fix the input file can be stored as a plain string?

nilsvu:
Can you also add Python bindings for the CCE file type? I expect we'll want to access those in Python quite a lot.

knelli2:
1. Rebased
2. Added new commit for writing string attributes using C-style string (did not change current uses like in input_source. That can be done in a different PR)
3. Squashed fixup into "Add Cce" commit
4. Added new commit that binds the Cce h5 file in python

nilsdeppe (dismissed):

src/IO/H5/Helpers.cpp
```diff
@@ -473,8 +473,15 @@ void write_to_attribute(const hid_t location_id, const std::string& name,
   const hid_t att_id = H5Acreate2(location_id, name.c_str(), h5_type<Type>(),
                                   space_id, h5p_default(), h5p_default());
   CHECK_H5(att_id, "Failed to create attribute '" << name << "'");
-  CHECK_H5(H5Awrite(att_id, h5_type<Type>(), static_cast<const void*>(&value)),
-           "Failed to write value: " << value);
+  if constexpr (std::is_same_v<Type, std::string>) {
+    const auto cstr = value.c_str();
+    CHECK_H5(H5Awrite(att_id, h5_type<Type>(), &cstr),
+             "Failed to write value: " << value);
```
wthrowe: I'm very confused.  This seems to work, but every example I can find online passes this function a `char *`, while this passes a `char **`.  What's going on?

src/IO/H5/Helpers.cpp
```diff
@@ -473,8 +473,15 @@ void write_to_attribute(const hid_t location_id, const std::string& name,
   const hid_t att_id = H5Acreate2(location_id, name.c_str(), h5_type<Type>(),
                                   space_id, h5p_default(), h5p_default());
   CHECK_H5(att_id, "Failed to create attribute '" << name << "'");
-  CHECK_H5(H5Awrite(att_id, h5_type<Type>(), static_cast<const void*>(&value)),
-           "Failed to write value: " << value);
+  if constexpr (std::is_same_v<Type, std::string>) {
+    const auto cstr = value.c_str();
+    CHECK_H5(H5Awrite(att_id, h5_type<Type>(), &cstr),
```
wthrowe: From the `h5_type` docs: "\note For strings, the returned type must be released with H5Tclose"

wthrowe (changes requested):

src/IO/H5/Helpers.cpp
knelli2: This seems like it isn't always done in the `Helpers.cpp`

src/IO/H5/Helpers.cpp
knelli2: I tried removing the `&` and got a seg fault

src/IO/H5/Helpers.cpp
wthrowe: Yes, `read_value_attribute` also gets it wrong.  The other calls aren't strings.

src/IO/H5/Helpers.cpp
wthrowe: Yes, this is the form that seems to work.  I'm just wondering if you know of any documentation that actually explains how to use the function or an explanation of why every example I can find is apparently wrong.  SpEC does the same thing as this, which makes me happier.

knelli2: No I don't know of any documentation for this. I just copied this from Nils Vu's patch he sent, and didn't question it.

nilsdeppe: I can't find any explicit HDF5 documentation on this. A couple examples I found online actually don't write the attribute out because they use `H5S_NULL` instead of `H5S_SCALAR`. Below is an example code if you want to try digging into this. My best guess is that because the `H5S_SCALAR` is a "scalar string" that you must pass a pointer to the string, i.e. a pointer to the pointer of characters. Here's the `H5Awrite` documentation https://docs.hdfgroup.org/hdf5/develop/group___h5_a.html#title37 with the `void * buf` being what I interpret as the "pointer to scalar" which if the scalar is of type `char*` then I guess you'd want a `char**`. In the below example, if I switch from `H5S_SCALAR` to `H5S_NULL` the code runs but `h5dump ./Attributes.h5` shows the attribute isn't written. If I try `H5S_SIMPLE` everything explodes with cryptic HDF5 internals errors that I don't understand. So what I can tell you is that empirically, this is the right way to write a null-terminated variable-length string as an HDF5 attribute. It may be different for datasets, I don't remember. I do remember years ago when writing the helpers that getting strings working was a pain because online examples were essentially useless...

```cpp
 #include <hdf5.h>
 #include <string>

 #define H5FILE_NAME "Attributes.h5"
 #define ATTR_NAME "VarLenAttr" /* Name of the string attribute */

 #define DATASET_RANK  1   /* Rank and size of the dataset  */
 #define DATASET_SIZE  7

 int main() {
   hid_t file, dataset, attr;       /* File and dataset identifiers */
   hid_t dataset_space, attr_space; /* Dataset's and Attribute's dataspace id */
   hid_t attr_type;                 /* Attribute type */
   herr_t status;                   /* Return value */

   hsize_t fdim[] = {DATASET_SIZE};
   std::string data{"ASD"};

   file = H5Fcreate(H5FILE_NAME, H5F_ACC_TRUNC, H5P_DEFAULT, H5P_DEFAULT);

   dataset_space = H5Screate(H5S_SIMPLE);
   status = H5Sset_extent_simple(dataset_space, DATASET_RANK, fdim, NULL);

   dataset = H5Dcreate2(file, "my_dataset", H5T_NATIVE_INT, dataset_space,
                        H5P_DEFAULT, H5P_DEFAULT, H5P_DEFAULT);

   attr_space = H5Screate(H5S_SCALAR);
   attr_type = H5Tcopy(H5T_C_S1);
   status = H5Tset_size(attr_type, H5T_VARIABLE);
   status = H5Tset_strpad(attr_type, H5T_STR_NULLTERM);
   status = H5Tset_cset(attr_type, H5T_CSET_ASCII);

   attr = H5Acreate2(dataset, ATTR_NAME, attr_type, attr_space, H5P_DEFAULT,
                     H5P_DEFAULT);

   const char* attr_string = data.c_str();
   status = H5Awrite(attr, attr_type, &attr_string);
   H5Aclose(attr);
   H5Tclose(attr_type);
   H5Sclose(attr_space);

   H5Fclose(file);
 }
```


Is this enough for you to be okay with this part of the diff or is there some explicit change request you still have on this part?

src/IO/H5/Helpers.cpp
nilsdeppe: Will is correct, this is a latent bug in the other code as well. It only matters for strings since we set some properties on the string to make it nicer to deal with.

src/IO/H5/Helpers.cpp
wthrowe: Thanks for the code.  Yes, I don't think anything needs to change for this.  I just spent a lot of time confused about how the interface worked and hoped someone else knew better than me.

nilsdeppe: You're welcome! Okay, sonuds good! I tried digging into the HDF5 internals but got stuck where they try to use structs with function pointer members as a way of mocking classes: https://github.com/HDFGroup/hdf5/blob/18321dee40b21e19c95bf0ef2f61e3ec71233fe9/src/H5VLconnector.h#L837 So I never got to the final "write the string to disk" stage since I have no idea where the "constructor" for this class is :(

src/IO/H5/Helpers.cpp
knelli2: Ok so shall I change it here, or do a follow-up?

wthrowe: Fix it in this PR.

knelli2: Squashed it in to the "Write H5" commit since it was only a few lines

nilsdeppe (approved):

wthrowe (approved):

Issue #5937, opened 2024-04-23
Initial orbital parameters and eccentricity control


nilsvu:
Currently we use some old SpEC scripts to do basic eccentricity control, both for setting the initial parameters and for the iterations. Specifically, we use `ZeroEccParamsFromPN.py` for the initial parameters (#5933) and old eccentricity fits for the iterations (#5895). We have several options for improving these:

- Use the eccentricity control based on EOB models that has been tested in SpEC.
- Use the new varpro fits that have been tested in SpEC.
- Extend the PN formulae to more terms and to include eccentricity.
- Extend either of these methods to work better for BNS/BHNS as well.

We probably want a combination of these methods, until it's clear that one works much better than the others.

These improvements are orthogonal to the eccentricity control automation procedure, which is tracked in #5909.

nilsvu:
Ping @PunJustice @haraldp271 @vtommasini @shabibti @taknapp @kchatziioannou @guilara since you've all been involved in eccentricity control. Let me know if you have any input :)

knelli2:
@markscheel as well just so you're aware

nilsvu:
In a meeting today with @markscheel @nilsdeppe @knelli2 we came up with an interface that can be used for eccentricity control. We think that we can use any eccentricity control method as long as it conforms to this interface. We can also import Python code that's merged into SpEC as long as it conforms to this interface. Just note that SpEC is private, so code merged into SpEC can only be used within the collaboration, whereas code merged into SpECTRE is publicly available.

Just two functions are needed:

```py
class OrbitalParameters:
    """These parameters define a binary orbit.

    They can be used to generate initial data and evolve it.
    Object A is initially at positive x, and object B is at negative x.

    Of the orbital parameters, only a subset may be specified. It is the responsibility
    of the functions below to check for consistency of these parameters
    and fill in the rest.
    """
    # Intrinsic parameters
    mass_ratio: float  # q = M_A / M_B >= 1
    dimensionless_spin_a: Sequence[float]
    dimensionless_spin_b: Sequence[float]
    # Orbital parameters
    separation: Optional[float]  # D_0
    orbital_angular_velocity: Optional[float]  # Omega_0
    radial_expansion_velocity: Optional[float]  # adot_0
    eccentricity: Optional[float]
    mean_anomaly_fraction: Optional[float]  # The mean anomaly divided by 2 pi, so a value between 0 and 1.
    num_orbits: Optional[float]
    time_to_merger: Optional[float]

[...]

PR #5966, opened 2024-05-05
Add OmegaDotEccentricityControl.py for SpECTRE

AzerAfram:
## Proposed changes
This script (originally from SpEC) takes in a directory that contains `GhBinaryBlackHoleReductionData.h5` that was outputted from a binary black hole run on SpECTRE (ideally, the binary black hole ran till merger).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments
Several niche functions have been added to extract the necessary data from the reduction data. If necessary, the script can be altered to utilize only one function for extraction, though this would make the main functions more difficult to read. Additionally, omega, eccentricity, and coordinate separation must be modified (refer to "INPUT" in the file). If necessary, the script can be adjusted to accept these values as parameters instead. If either `/ObservationAhA.dat` or `/ObservationAhB.dat` in `GhBinaryBlackHoleReducitonData.h5` contain less than ~10 entries, refer to `dimensionFullSpins()`.

### DIFF
--- a/src/Visualization/Python/Render3D/Bbh.py
+++ b/src/Visualization/Python/Render3D/Bbh.py
@@ -54,6 +54,7 @@ def render_bbh(
     time_step: int = 0,
     animate: bool = False,
     camera_angle: str = "Side",
+    zoom_factor: float = 1.0,
     color_map: str = "Rainbow Uniform",
     show_grid: bool = False,
     show_time: bool = False,
@@ -186,6 +187,9 @@ def render_bbh(
         ]
         render_view.CameraViewUp = [0.0, 0.0, 1.0]
         render_view.CameraParallelScale = 519.6152422706632
+    camera = pv.GetActiveCamera()
+    pv.ResetCamera()
+    camera.Zoom(zoom_factor)
 
     # Simulation time
     if show_time:
@@ -262,6 +266,7 @@ def render_bbh(
         " inbetween Side and Top view"
     ),
 )
+@click.option("zoom_factor", "--zoom", help="Zoom factor.", default=1.0)
 @click.option(
     "--color-map",
     "-m",


geoffrey4444:
@AzerAfram I'm glad you got this to work! But just merging this script would mean any future updates to the spec script would not make it into spectre.

I wonder if you could modify this so that instead of putting in the entire script, you read in the files as you do here, then output an H5 file with the same data in the places the spec script expects? Then you can just use the spec script, rather than using a copy that would have to be maintained every time the spec script gets updated

nilsvu:
Quick note: it's also possible to `import` a script from spec, as is done in `support/Pipelines/EccentricityControl/InitialOrbitalParameters.py`. I can take a closer look at this later.

nilsvu:
Another quick note: there's some discussion on this topic in #5937.

AzerAfram:
@geoffrey4444 Yes, I can and will do that instead. Thank you, @nilsvu.

AzerAfram:
Okay, I've made it so that the script alters `Horizons.h5` where necessary, and also creates an `ID_Params.perl` file with only the required arguments.

geoffrey4444 (changes requested):
Please squash into one commit. I can help you do this you're note sure how.

nilsvu:
Hi @AzerAfram , we've made some progress on the SpEC side of the interface. This PR is about to be merged: https://github.com/sxs-collaboration/spec/pull/2673. It allows you to pass in the trajectories as numpy arrays and returns the updates to the orbital parameters. Could you update this PR to use it, so you don't have to write/edit any H5 files anymore and you don't have to write a Params perl file anymore? Just import and call the `performAllFits` function.

AzerAfram:
Ok, I'll do that.

PR #5968, opened 2024-05-05
Animate PlotSlice and PlotAlongLine

nilsvu:
## Proposed changes

Allows to easily write out a gif. Support for things like time slicing can be added later.

Example:

slice

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/Plot.py
+++ b/src/Visualization/Python/Plot.py
@@ -6,7 +6,9 @@ import logging
 import os
 
 import click
+import matplotlib.animation
 import matplotlib.pyplot as plt
+import rich.progress
 
 logger = logging.getLogger(__name__)
 
@@ -49,9 +51,11 @@ def apply_stylesheet_command():
 def show_or_save_plot_command():
     """Add an 'output' CLI option and show or save the plot accordingly
 
-    Apply this decorator to a CLI command that generates a plot. At the end of
-    the command, the plot is either shown interactively or saved to a file,
-    depending on the user-specified 'output' option.
+    Apply this decorator to a CLI command that generates a plot or animation.
+    Return the `matplotlib.Figure` or `matplotlib.animation.Animation` from your
+    command. At the end of the command, the plot or animation is either shown
+    interactively or saved to a file, depending on the user-specified 'output'
+    option.
     """
 
     def decorator(f):
@@ -63,7 +67,8 @@ def show_or_save_plot_command():
                 "Name of the output plot file. If unspecified, the plot is "
                 "shown interactively, which only works on machines with a "
                 "window server. If a filename is specified, its extension "
-                "determines the file format, e.g. 'plot.png' or 'plot.pdf'. "
+                "determines the file format, e.g. 'plot.png' or 'plot.pdf' "
+                "or 'animation.gif'. "
                 "If no extension is given, the file format depends on the "
                 "system settings (see matplotlib.pyplot.savefig docs)."
             ),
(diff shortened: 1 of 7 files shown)


src/Visualization/Python/Plot.py
```diff
@@ -63,7 +67,8 @@ def decorator(f):
                 "Name of the output plot file. If unspecified, the plot is "
                 "shown interactively, which only works on machines with a "
                 "window server. If a filename is specified, its extension "
-                "determines the file format, e.g. 'plot.png' or 'plot.pdf'. "
+                "determines the file format, e.g. 'plot.png' or 'plot.pdf' "
+                "or 'animation.gif'. "
```
knelli2: What about `.mp4`? If users specify `mp4`, does that work too? I'm thinking that people may not always want a gif

src/Visualization/Python/PlotAlongLine.py
```diff
@@ -97,14 +112,30 @@ def plot_along_line_command(
...
 ):
     """Plot variables along a line through volume data
 
     Interpolates the volume data in the H5_FILES to a line and plots the
     selected variables. You choose the line by specifying the start and end
     points.
+
+    Either select a specific observation in the volume data with '--step' or
+    '--time', or omit these options to produce an animation over all
+    observations.
```
knelli2: I'm not sure about making the default be an animation. I think the default should still be that you have to specify a time/step *or* pass a flag like `--animate` to get an animation. Otherwise the command just erros and prints the help message

src/Visualization/Python/PlotAlongLine.py
```diff
@@ -144,19 +167,56 @@ def plot_along_line_command(
...
+            observation_id=obs_id,
+            tensor_components=vars,
+            target_points=target_coords,
+            num_threads=num_threads,
+        )
+        for y, var_name, line in zip(vars_on_line, vars, lines):
+            line.set_data(x, y)
+        time_label.set_text(f"t = {obs_time:g}")
+        ax.relim()
+        ax.autoscale_view()
```
knelli2, line 208: Not a change request, but this could potentially create weird movies if the bounds are changing quickly. We'd have to do something more sophisticated to fix this though.

src/Visualization/Python/PlotSlice.py
```diff
@@ -157,14 +175,30 @@ def plot_slice_command(
...
 
     Interpolates the volume data in the H5_FILES to a slice and plots the
-    selected variables. You choose the slice by specifying three corners of a
-    rectangle.
+    selected variables. You choose the slice by specifying its center, extents,
+    normal, and up direction.
+
+    Either select a specific observation in the volume data with '--step' or
+    '--time', or omit these options to produce an animation over all
+    observations.
```
knelli2: Same here about the default being an animation

src/Visualization/Python/PlotSlice.py
```diff
@@ -213,17 +237,92 @@ def coord_axis(vec):
...
+        fig,
+        update_plot,
+        init_func=list,
+        frames=len(obs_ids),
+        interval=interval,
+        blit=False,
+    )
 
 
 if __name__ == "__main__":
```
knelli2, line 331: [random line] Test both of these please

knelli2 (changes requested):

src/Visualization/Python/Plot.py
nilsvu: Yes works too but only if ffmpeg is available

src/Visualization/Python/PlotAlongLine.py
nilsvu: Sounds good will add the flag

src/Visualization/Python/PlotAlongLine.py
nilsvu: The alternative is looking through all files and observations to determine bounds. I had to do that for PlotSlice so colors remain the same, but here I thought it's better to autoscale and have the user fix bounds if needed.

src/Visualization/Python/Plot.py
knelli2: Ok nice. Maybe add a note about `mp4` and `ffmpeg` then?

nilsvu:
pushed a fixup

knelli2 (commented):
Looks good. You can squash

knelli2 (approved):

PR #5972, opened 2024-05-07
Static libsz and libaec, switch to HDF5::HDF5 in cmake, rename ENABLE_SPECTRE_DEBUG to SPECTRE_DEBUG in cmake, fix incorrect CCE assert

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupHdf5.cmake
+++ b/cmake/SetupHdf5.cmake
@@ -52,6 +52,34 @@ if(HDF5_IS_PARALLEL)
   endif()
 endif()
 
+if(HDF5_USE_STATIC_LIBRARIES)
+  # If we are using static libraries for HDF5, try to use static libraries
+  # for the libsz and transitive libaec dependency. These aren't very
+  # common, while other dependencies like libz are, so we leave those
+  # as shared library dependencies.
+  get_target_property(
+    _HDF5_INTERFACE_LINK_LIBS HDF5::HDF5 INTERFACE_LINK_LIBRARIES)
+  string(FIND
+    "${_HDF5_INTERFACE_LINK_LIBS}" "libsz.so" _LOCATION_OF_LIBSZ)
+  if (NOT ${_LOCATION_OF_LIBSZ} EQUAL -1)
+    find_library(_libsz NAMES libsz.a)
+    if(_libsz)
+      string(REPLACE
+        "libsz.so" "libsz.a" _HDF5_INTERFACE_LINK_LIBS
+        "${_HDF5_INTERFACE_LINK_LIBS}")
+      find_library(_libaec NAMES libaec.a)
+      if(_libaec)
+        list(APPEND _HDF5_INTERFACE_LINK_LIBS ${_libaec})
+      endif()
+    endif()
+  endif()
+  set_target_properties(
+    HDF5::HDF5
+    PROPERTIES
+    INTERFACE_LINK_LIBRARIES "${_HDF5_INTERFACE_LINK_LIBS}"
+  )
+endif()
+
 # Check if file locking API is available. The versions supporting this feature
 # are listed here:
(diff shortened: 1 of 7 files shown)


kidder (approved):

cmake/SetupHdf5.cmake
```diff
@@ -52,6 +52,34 @@ if(HDF5_IS_PARALLEL)
...
+  # If we are using static libraries for HDF5, try to use static libraries
+  # for the libsz and transitive libaec dependency. These aren't very
+  # common, while other dependencies like libz are, so we leave those
+  # as shared library dependencies.
+  get_target_property(
+    _HDF5_INTERFACE_LINK_LIBS HDF5::HDF5 INTERFACE_LINK_LIBRARIES)
+  string(FIND
+    "${_HDF5_INTERFACE_LINK_LIBS}" "libsz.so" _LOCATION_OF_LIBSZ)
+  if (NOT ${_LOCATION_OF_LIBSZ} EQUAL -1)
+    find_library(_libsz NAMES libsz.a)
```
nilsvu, line 65: libsz is an optional dependency of hdf5 that you specify when building hdf5. Also when building hdf5 you can specify if you want to build static libs, shared libs, or both. So if a static build of everything is important, can we just make sure to build hdf5 in an all-static way instead of trying to patch transitive dependencies in the spectre build system? For an all-static build we have to control the environment very carefully anyway, probably in a container, so we can make sure to build hdf5 correctly as well.

nilsdeppe: I'm not sure what you're looking for in terms of changes or are suggesting. This works in the container for producing a portable statically linked executable. I linked against static HDF5 and then found out that `libsz` was linked dynamically for that. Are you suggesting we compile HDF5 ourselves statically in the container?

nilsvu: Yes that's what I thought, in fact I thought we had to build a modified container where everything is statically linked anyway?

nilsdeppe: Not that I know of, but I haven't thought about the containers in a while. I don't really have time to reconfigure the way we build everything in the container, so from my side it's either this or I can't distribute CCE for the ETK workshop...

knelli2:
That's one mouthful of a PR title

nilsdeppe:
Yea... But I didn't want to "sneak" anything in 📦

knelli2:
But that's what the description is for!

nilsdeppe:
Oh, I view that as "details" :P And I couldn't think of how to summarize 4 tangential things into one catch short word :P

src/Evolution/Systems/Cce/WorldtubeBufferUpdater.cpp
```diff
@@ -39,9 +39,11 @@ std::pair<size_t, size_t> create_span_for_time_value(
       lower_bound < upper_bound,
       "The supplied `lower_bound` is greater than `upper_bound`, which is not "
       "permitted");
-  ASSERT(2 * interpolator_length + pad < upper_bound,
+  ASSERT(2 * interpolator_length + pad <= upper_bound,
```
wthrowe, line 42: I can't figure out what this function does, but this new check is what is required to avoid underflow in the math, so it's probably correct.

wthrowe (approved):

src/Evolution/Systems/Cce/WorldtubeBufferUpdater.cpp
nilsdeppe: Yea, I traced this back a decent bit through the code, never fully understood it but understood it well enough to expect several other ASSERTs to trigger if it was wrong :)

PR #5976, opened 2024-05-09
CLI for plotting trajectories of binaries

vtommasini:
### Code review checklist

- [ ] This CLI plots trajectories from an inspiral run, concatenating the partial trajectories from each segment, and plots the full trajectories.
- [ ] The output can be a .pdf, .png, or graphic window if no output file type is sepcified.
- [ ] Example of output file:
Screenshot 2024-05-09 at 8 17 13 AM

### DIFF
--- a/src/Visualization/Python/PlotTrajectories.py
+++ b/src/Visualization/Python/PlotTrajectories.py
@@ -1,0 +1,185 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+# Plot trajectory from inspiral run, concatenating all segments
+
+import logging
+
+import click
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+
+logger = logging.getLogger(__name__)
+
+from spectre.Visualization.Plot import (
+    apply_stylesheet_command,
+    show_or_save_plot_command,
+)
+
+
+def import_A_and_B(filenames, subfile_name_aha, subfile_name_ahb):
+    A_data = []
+    B_data = []
+
+    for filename in filenames:
+        with h5py.File(filename, "r") as file:
+            A_data.append(
+                np.array(
+                    file[subfile_name_aha][:, [4, 5, 6]]
+                )  # 0 ->time,  4 -> x, 5 -> y, 6 -> z
+            )
+            B_data.append(
+                np.array(
+                    file[subfile_name_ahb][:, [4, 5, 6]]
+                )  # 0 ->time,  4 -> x, 5 -> y, 6 -> z
+            )
+
+    if A_data and B_data:
+        A = np.concatenate(A_data)
+        B = np.concatenate(B_data)
+        min_length = min(len(A), len(B))
+        return A[:min_length], B[:min_length]
+    else:
+        return None, None
+
+
+def plot_trajectory(AhA, AhB, sample_rate, figsize):
+    """
+    Plot concatenated trajectories in inspiral simulation.
+
+    Outputs a 2x2 figure with 4 plots:
+    1st row: A and B trajectories in 3D, coordinate separation in 3D
+    2nd row: A and B trajectories in 2D xy plane, coordinate separation in 2D
+    xy plane. The 3D plots use a subsampled version of the data, with default
+    value of 15, to speed up the plots and avoid memory error.
+
+    Arguments:
+    AhA: Array of shape (num_points, 3) with the coordinates of
+    the first object.
+    AhB: Array of shape (num_points, 3) with the coordinates of
+    the second object.
+    sample_rate: Integer specifying the downsampling rate for 3D plots.
+    figsize: Tuple of two floats specifying the figure size as
+    width and height in inches.
+    """
+
+    fig = plt.figure(figsize=figsize)
+
+    # Sample data for 3D plots
+    AhA_sampled = AhA[::sample_rate]
+    AhB_sampled = AhB[::sample_rate]
+
+    # Plot 3D trajectories
+    ax1 = fig.add_subplot(2, 2, 1, projection="3d")
+    ax1.plot(
+        *AhA_sampled.T,
+        color="C0",
+        label="AhA",
+    )
+    ax1.plot(
+        *AhB_sampled.T,
+        color="C1",
+        label="AhB",
+    )
+    ax1.set_xlabel("X")
+    ax1.set_ylabel("Y")
+    ax1.set_zlabel("Z")
+    ax1.set_title("3D Trajectories")
+    ax1.legend()
+
+    # Calculate coordinate separation in 3D
+    separation_3d = AhA_sampled - AhB_sampled
+
+    # Plot 3D coordinate separation
+    ax2 = fig.add_subplot(2, 2, 2, projection="3d")
+    ax2.plot(
+        *separation_3d.T,
+        color="C2",
+    )
+    ax2.set_xlabel("X diff")
+    ax2.set_ylabel("Y diff")
+    ax2.set_zlabel("Z diff")
+    ax2.set_title("Coordinate separation (3D)")
+
+    # Plot 2D trajectories
+    ax3 = fig.add_subplot(2, 2, 3)
+    ax3.plot(*AhA[:, 0:2].T, label="AhA", color="C0")
+    ax3.plot(*AhB[:, 0:2].T, label="AhB", color="C1")
+    ax3.set_xlabel("x")
+    ax3.set_ylabel("y")
+    ax3.legend()
+    ax3.set_title("Trajectories (2D)")
+    ax3.set_aspect("equal")  # Set aspect ratio to 1
+    ax3.grid(True)  # Add gridlines
+
+    # Calculate coordinate separation in 2D
+    separation_2d = AhA[:, 0:2] - AhB[:, 0:2]
+    # Plot coordinate separation in 2D
+    ax4 = fig.add_subplot(2, 2, 4)
+    ax4.plot(*separation_2d.T, color="C2")
+    ax4.set_xlabel("x1 - x2")
+    ax4.set_ylabel("y1 - y2")
+    ax4.set_title("Coordinate separation (2D)")
+    ax4.set_aspect("equal")  # Set aspect ratio to 1
+    ax4.grid(True)  # Add gridlines
+
+    plt.tight_layout()
+    plt.subplots_adjust(wspace=0.3, hspace=0.3)
+
+
+@click.command(name="trajectories")
+@click.argument(
+    "h5_files",
+    nargs=-1,
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+)
+@click.option(
+    "--subfile-name-aha",
+    "-A",
+    default="ApparentHorizons/ControlSystemAhA_Centers.dat",
+    help=(
+        "Name of subfile containing the apparent horizon centers for object A."
+    ),
+)
+@click.option(
+    "--subfile-name-ahb",
+    "-B",
+    default="ApparentHorizons/ControlSystemAhB_Centers.dat",
+    help=(
+        "Name of subfile containing the apparent horizon centers for object B."
+    ),
+)
+@click.option(
+    "--sample-rate",
+    type=int,
+    default=10,
+    show_default=True,
+    help="Downsampling value for 3D plots",
+)
+@click.option(
+    "--figsize",
+    nargs=2,
+    type=float,
+    default=(10.0, 10.0),
+    show_default=True,
+    help="Figure size as width and height in inches",
+)
+@apply_stylesheet_command()
+@show_or_save_plot_command()
+def plot_trajectories_command(
+    h5_files, subfile_name_aha, subfile_name_ahb, sample_rate, figsize=(10, 10)
+):
+    """Plot trajectories in inspiral simulation
+
+    Concatenates partial trajectories from each h5 reductions files and plots
+    full trajectories.
+    """
+    reductionfiles = h5_files
+    AhA, AhB = import_A_and_B(
+        reductionfiles, subfile_name_aha, subfile_name_ahb
+    )
+    plot_trajectory(AhA, AhB, sample_rate, figsize)
+
+
+if __name__ == "__main__":
+    plot_trajectories_command(help_option_names=["-h", "--help"])
--- a/tests/Unit/Visualization/Python/Test_PlotTrajectories.py
+++ b/tests/Unit/Visualization/Python/Test_PlotTrajectories.py
@@ -1,0 +1,137 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+# unit test for plot trajectories
+
+import logging
+import os
+import shutil
+import unittest
+
+import numpy as np
+from click.testing import CliRunner
+
+import spectre.IO.H5 as spectre_h5
+from spectre.Informer import unit_test_build_path, unit_test_src_path
+from spectre.Visualization.PlotTrajectories import plot_trajectories_command
+
+
+class TestPlotTrajectories(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = os.path.join(
+            unit_test_build_path(), "Visualization", "PlotTrajectories"
+        )
+        self.h5_filename = os.path.join(
+            self.test_dir, "TestPlotTrajectoriesReductions.h5"
+        )
+        shutil.rmtree(self.test_dir, ignore_errors=True)
+        os.makedirs(self.test_dir, exist_ok=True)
+        self.create_h5_file()
+
+    def tearDown(self):
+        shutil.rmtree(self.test_dir)
+
+    def create_h5_file(self):
+        """Create the H5"""
+        logging.info(f"Creating HDF5 file: {self.h5_filename}")
+
+        # Generate mock-up inspiral data
+        nsamples = 100
+        dt = 0.02
+        x0 = 0.35
+        y0 = 0.35
+        z0 = 0
+        z1 = -9.0e-6
+        cosAmp = 7.43
+        sinAmp = 7.43
+        cosFreq = 0.0173
+        sinFreq = 0.0172
+
+        # Define the spirals as functions
+        def SpiralA(t):
+            return np.array(
+                [
+                    x0 + cosAmp * np.cos(-cosFreq * t),
+                    y0 - sinAmp * np.sin(sinFreq * t),
+                    z0 + z1 * (1 - 0.1) * t,
+                ]
+            )
+
+        def SpiralB(t):
+            return np.array(
+                [
+                    -x0 + cosAmp * np.cos(np.pi + cosFreq * t),
+                    -y0 + sinAmp * np.sin(sinFreq * t),
+                    z0 + z1 * (1 + 0.1) * t,
+                ]
+            )
+
+        # Generate time samples
+        tTable = np.arange(0, (nsamples + 1) * dt, dt)
+
+        # Map time to spiral points
+        AhA_data = np.array([[t, *SpiralA(t), *SpiralA(t)] for t in tTable])
+        AhB_data = np.array([[t, *SpiralB(t), *SpiralB(t)] for t in tTable])
+
+        with spectre_h5.H5File(self.h5_filename, "w") as h5_file:
+            # Insert dataset for AhA
+            dataset_AhA = h5_file.insert_dat(
+                "ApparentHorizons/ControlSystemAhA_Centers.dat",
+                legend=[
+                    "Time",
+                    "GridCenter_x",
+                    "GridCenter_y",
+                    "GridCenter_z",
+                    "InertialCenter_x",
+                    "InertialCenter_y",
+                    "InertialCenter_z",
+                ],  # Legend for the dataset
+                version=0,  # Version number
+            )
+            # Populate dataset with AhA data
+            for data_point in AhA_data:
+                dataset_AhA.append(data_point)
+            # Close dataset for AhA
+            h5_file.close_current_object()
+
+            # Insert dataset for AhB
+            dataset_AhB = h5_file.insert_dat(
+                "ApparentHorizons/ControlSystemAhB_Centers.dat",
+                legend=[
+                    "Time",
+                    "GridCenter_x",
+                    "GridCenter_y",
+                    "GridCenter_z",
+                    "InertialCenter_x",
+                    "InertialCenter_y",
+                    "InertialCenter_z",
+                ],  # Legend for the dataset
+                version=0,  # Version number
+            )
+            # Populate dataset with AhB data
+            for data_point in AhB_data:
+                dataset_AhB.append(data_point)
+            # Close dataset for AhB
+            h5_file.close_current_object()
+        logging.info(
+            f"Successfully created and populated HDF5 file: {self.h5_filename}"
+        )
+
+    def test_cli(self):
+        output_filename = os.path.join(self.test_dir, "output.pdf")
+        runner = CliRunner()
+        result = runner.invoke(
+            plot_trajectories_command,
+            [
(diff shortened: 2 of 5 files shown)


src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+    type=str,
+    help=(
+        'Output filename. Choose ".pdf" '
+        'or ".png". If not provided, the plot will be displayed in a '
+        "graphic window."
+    ),
+)
+@click.option(
+    "--force", is_flag=True, help="Force overwriting existing output file."
+)
```
nilsvu: Replace these two options with:
```py
@show_or_save_plot_command()
@apply_stylesheet_command()
```
imported from `spectre.Visualization.Plot`. The first will take care of saving the plot or displaying it on screen for you. The second will apply a default configuration to the plots, and also allows the user to adjust the style (try it, e.g. `-s seaborn-dark`).

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+import os
+
+import click
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+@click.command()
+@click.argument("simulation_name", type=str)
```
nilsvu: Replace this option with:

```py
@click.argument(
    "h5_files",
    nargs=-1,
    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
)
```

This means the user won't pass a simulation directory to the CLI, but they will pass the reduction files directly as a glob pattern like this:

```
spectre plot trajectories path/to/Segment_*/BbhReductions.h5
```

This is more flexible because no assumptions are made on the directory structure and the user can choose which segments/files to plot. It's listed in the guidelines here: https://spectre-code.org/spectre_writing_python_bindings.html#autotoc_md155

Then you don't need the `select_reduction_files` function anymore.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+                    B_data.append(
+                        np.array(
+                            file[
+                                "ApparentHorizons/ControlSystemAhB_Centers.dat"
+                            ][:]
+                        )
+                    )
+        except Exception as e:
+            click.echo(f"Error reading file '{filename}': {e}")
+            return None, None
```
nilsvu: You don't need this `try: except:`. If an error occurs then it will just propagate to the user, it'll give as much information as this `click.echo`.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+
+def plot_trajectory(AhA, AhB, plotlabel="", output=None, force=False):
+    indexs = [4, 5]  # 4 -> x, 5 -> y
+
+    if output and os.path.exists(output) and not force:
+        click.echo(
+            f"Warning: Output file '{output}' already exists. "
+            "Use --force to overwrite."
+        )
+        return
```
nilsvu: You can delete this when you use `@show_or_save_plot_command`. It's a good idea though, and we should consider adding this functionality to `@show_or_save_plot_command` in a different PR.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+
+    plt.figure(figsize=(10, 5))
```
nilsvu: Take the `figsize` as an additional option to the CLI command

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+    plt.subplot(1, 2, 1)
+    plt.plot(
+        [AhA[i][indexs[0]] for i in range(len(AhA))],
```
nilsvu: Use Numpy array slicing (always prefer that over loops, it's so much faster):

`[AhA[i][indexs[0]] for i in range(len(AhA))]` -> `AhA[:, indexs[0]]`

same in multiple places below

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+        [AhA[i][indexs[1]] for i in range(len(AhA))],
+        label="AhA",
+        color="blue",
```
nilsvu: `color="C0"` (that's the default and can be configured by the user)

Suggestion: I think it would be useful to fade the color from opacity 0 to 1 based on the time parameter, or maybe from some opacity < 1 to 1. That would give a sense of the time evolution. Maybe also plot `marker="."` at the very end of the trajectory to highlight the current position.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+    )
+    plt.plot(
+        [AhB[i][indexs[0]] for i in range(len(AhB))],
+        [AhB[i][indexs[1]] for i in range(len(AhB))],
+        label="AhB",
+        color="red",
```
nilsvu: `color="C1"`

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+        A = np.concatenate(A_data)
+        B = np.concatenate(B_data)
+        min_length = min(len(A), len(B))
+        return A[:min_length], B[:min_length]
+    else:
+        return None, None
+
+
+def plot_trajectory(AhA, AhB, plotlabel="", output=None, force=False):
+    indexs = [4, 5]  # 4 -> x, 5 -> y
```
nilsvu: What do you think about adding another row of plots that show the x-z projection?

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+
+    plt.subplot(1, 2, 2)
+    x_diff = [AhA[i][indexs[0]] - AhB[i][indexs[0]] for i in range(len(AhA))]
+    y_diff = [AhA[i][indexs[1]] - AhB[i][indexs[1]] for i in range(len(AhA))]
+    plt.plot(x_diff, y_diff, color="black")
+    plt.xlabel("x1-x2")
+    plt.ylabel("y1-y2")
+    plt.title("Coordinate separation" + " " + plotlabel)
+
+    plt.tight_layout()
```
nilsvu, line 126: You can delete this line, it's taken care of by the default plot configuration.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+            click.echo(
+                f"Error: Output file '{output}' already exists. "
+                "Use --force to overwrite."
+            )
+            return
+        plt.savefig(output)
+        click.echo(f"Plot saved as {output}")
+        plt.close()
+    else:
+        plt.show()  # No output file specified, display in graphic window
```
nilsvu: This is taken care of by `@show_or_save_plot_command`, so you can delete this part :)

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+import glob
+import os
+
+import click
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+
+
+@click.command()
```
nilsvu: add `name="trajectories"` so it appears properly in help texts.

Also name the function `plot_trajectories_command` for consistency.

nilsvu (changes requested):
This looks great! Thanks for writing this script. Please squash all your commits including changes you make in response to my comments below into a single commit, then force-push to this branch.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,155 @@
...
+
+
+def import_A_and_B(filenames):
+    A_data = []
+    B_data = []
+
+    for filename in filenames:
+        try:
+            with h5py.File(filename, "r") as file:
+                if "ApparentHorizons/ControlSystemAhA_Centers.dat" in file:
```
nilsvu: Take the subfile names as options to the CLI command that have this as the default value.

nilsvu (changes requested):

src/Visualization/Python/PlotSizeControl.py
```diff
@@ -27,15 +27,6 @@
     type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
     nargs=-1,
 )
-@click.option(
```
nilsvu: Oops don't delete this :)

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,157 @@
...
+    full trajectories.
+    """
+    reductionfiles = h5_files
+    AhA, AhB = import_A_and_B(
+        reductionfiles, subfile_name_aha, subfile_name_ahb
+    )
+    if AhA is not None and AhB is not None:
+        plot_trajectory(AhA, AhB)
+    else:
+        click.echo("Error: Unable to import data or not valid data.")
```
nilsvu: You don't need this `if: else:`. If an error happens then it will be displayed.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def import_A_and_B(filenames, subfile_name_aha, subfile_name_ahb):
+    A_data = []
+    B_data = []
+
+    for filename in filenames:
+        with h5py.File(filename, "r") as file:
+            if subfile_name_aha in file:
```
nilsvu: Remove these `if` conditions as well. It's great that you're considering what happens when errors happen! I agree that's very important. You always want to give the user enough information and guidance. In this case just access the subfile so an error is thrown when it doesn't exists. That should be enough, and it's better than silently ignoring the file.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,159 @@
...
+                A_data.append(np.array(file[subfile_name_aha][:]))
+            if subfile_name_ahb in file:
+                B_data.append(np.array(file[subfile_name_ahb][:]))
+
+    if A_data and B_data:
```
nilsvu, line 37: You can remove this `if` as well

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,159 @@
...
+
+
+def import_A_and_B(filenames, subfile_name_aha, subfile_name_ahb):
+    A_data = []
+    B_data = []
+
+    for filename in filenames:
+        with h5py.File(filename, "r") as file:
+            if subfile_name_aha in file:
+                A_data.append(np.array(file[subfile_name_aha][:]))
```
nilsvu: I suggest you import only the `time` and `xyz` columns of the data here, then you don't have to do any special indexing later. Like this: `file[subfile_name_aha][:, [0,4,5,6]]`

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,158 @@
...
+    "--subfile-name-ahb",
+    "-B",
+    default="ApparentHorizons/ControlSystemAhB_Centers.dat",
+    help=(
+        "Name of subfile containing the apparent horizon centers for object B."
+    ),
+)
+@show_or_save_plot_command()
+@apply_stylesheet_command()
+def plot_trajectories_command(h5_files, subfile_name_aha, subfile_name_ahb):
```
knelli2: This looks great! The image that this produces is really cool!

trajectories

Could you also add some tests for this? Similar to the test in `tests/Unit/Visualization/Python/Test_PlotSizeControl.py`

knelli2 (changes requested):

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,158 @@
...
+    AhA_sampled = AhA[::sample_rate]
+    AhB_sampled = AhB[::sample_rate]
+
+    # Plot 3D trajectories
+    ax1 = fig.add_subplot(2, 2, 1, projection="3d")
+    ax1.scatter(
+        AhA_sampled[:, indexs[0]],
+        AhA_sampled[:, indexs[1]],
+        AhA_sampled[:, indexs[2]],
+        color="blue",
```
nilsvu: `color="C0"`

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,158 @@
...
+    if A_data and B_data:
+        A = np.concatenate(A_data)
+        B = np.concatenate(B_data)
+        min_length = min(len(A), len(B))
+        return A[:min_length], B[:min_length]
+    else:
+        return None, None
+
+
+def plot_trajectory(AhA, AhB, sample_rate=15):
```
nilsvu: Please add documentation to this function that describes what the function does and also describes each argument. You find examples in the other Python files in this directory.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,158 @@
...
+    ax1.set_xlabel("X")
+    ax1.set_ylabel("Y")
+    ax1.set_zlabel("Z")
+    ax1.set_title("3D Trajectories")
+    ax1.legend()
+
+    # Calculate coordinate separation in 3D
+    x_diff_3d = AhA_sampled[:, indexs[0]] - AhB_sampled[:, indexs[0]]
+    y_diff_3d = AhA_sampled[:, indexs[1]] - AhB_sampled[:, indexs[1]]
+    z_diff_3d = AhA_sampled[:, indexs[2]] - AhB_sampled[:, indexs[2]]
```
nilsvu: This is just:

```
separation_3d = AhA_sampled[:, 1:] - AhB_sampled[:, 1:]
```

This computes all 3 components at once. For 2d it's `[:, 1:3]`.

nilsvu (changes requested):

src/Visualization/Python/PlotTrajectories.py
vtommasini, line 1: Here is my revised version, with all the current comments/suggestions incorporated. I made the code more compact. The suggested as well as new options are documented in the help file. I'll keep working on the Test script while I wait for more feedback.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,188 @@
...
+    value of 15, to speed up the plots and avoid memory error
+    """
+
+    # Scaling factor for font size based on figure size
+    base_figsize = (9.0, 9.0)  # Base figure size for reference
+    scale_factor = np.sqrt(
+        (figsize[0] * figsize[1]) / (base_figsize[0] * base_figsize[1])
+    )
+    base_fontsize = 10  # Base font size for labels
+    fontsize = base_fontsize * scale_factor
```
nilsvu: Is your intention here to make the font bigger when the figure is bigger? You don't need to do that. It's better to stick with the default and let the user increase the fontsize if they want (e.g. with a stylesheet).

Generally when plotting: If you want bigger labels, decrease the figsize (for reference, a paper typically has 3.4 inch wide columns). If you want higher resolution, increase the DPI (in a stylesheet).

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,188 @@
...
+    fig = plt.figure(figsize=figsize)
+
+    # Sample data for 3D plots
+    AhA_sampled = AhA[::sample_rate]
+    AhB_sampled = AhB[::sample_rate]
+
+    # Plot 3D trajectories
+    ax1 = fig.add_subplot(2, 2, 1, projection="3d")
+    ax1.plot(
+        *AhA_sampled[:, 0:].T,
```
nilsvu: (Optional) You can remove this indexing because it selects the full array (delete `[:, 0:]`).

nilsvu (changes requested):
Only a test is missing now as @knelli2 mentioned above. Maybe write some made-up data to a file in a `setUp` function, then run the CLI command to plot it (@knelli2 pointed to some examples). You can't really test that the plot looks correct, but at least this test makes sure that the code runs without error, and keeps working in the future.

You can squash everything into 1 commit.

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -0,0 +1,143 @@
...
+from spectre.Visualization.PlotTrajectories import plot_trajectories_command
+
+
+class TestPlotTrajectories(unittest.TestCase):
+    def setUp(self):
+        self.test_dir = os.path.join(
+            unit_test_build_path(), "Visualization", "PlotTrajectories"
+        )
+        self.h5_filename = os.path.join(
+            unit_test_src_path(),
```
nilsvu: Put this into the `self.test_dir` as well.

More info: The `unit_test_src_path` points to the source code directory. You don't want to create temporary files there, since it can mess with git. It's better to create the data file in the test dir, which gets deleted and recreated every time you run the test.

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -0,0 +1,143 @@
...
+            "Visualization/Python",
+            "TestPlotTrajectoriesReductions.h5",
+        )
+        os.makedirs(self.test_dir, exist_ok=True)
+        self.create_h5_file_if_not_exists()
+
+    def tearDown(self):
+        if TestPlotTrajectories.delete_test_dir:
```
nilsvu: You don't need this `if`. The temporary test directory can always be deleted here.

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -0,0 +1,143 @@
...
+            shutil.rmtree(self.test_dir)
+
+    # if h5 test file does not exist then create one using mock-up inspiral data
+    def create_h5_file_if_not_exists(self):
+        """Create the H5 file if it does not exist."""
+        if not os.path.exists(self.h5_filename):
```
nilsvu: You can delete this `if`. Instead, put `shutil.rmtree(self.work_dir, ignore_errors=True)` just before `os.makedirs` in the `setUp` function above. Then the test always starts from a clean slate.

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -0,0 +1,143 @@
...
+            print(f"Creating HDF5 file: {self.h5_filename}")
```
nilsvu: Switch all `print`s to `logging.info()` (or delete them)

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -0,0 +1,143 @@
...
+            os.makedirs(os.path.dirname(self.h5_filename), exist_ok=True)
```
nilsvu: You can delete this line when the H5 file is put into the test dir.

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -0,0 +1,143 @@
...
+                output_filename,
+            ],
+            catch_exceptions=False,
+        )
+        self.assertEqual(result.exit_code, 0, result.output)
+        self.assertTrue(os.path.exists(output_filename))
+
+
+if __name__ == "__main__":
+    TestPlotTrajectories.delete_test_dir = True
```
nilsvu: You can delete this line

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -0,0 +1,143 @@
...
+
+    # if h5 test file does not exist then create one using mock-up inspiral data
+    def create_h5_file_if_not_exists(self):
+        """Create the H5 file if it does not exist."""
+        if not os.path.exists(self.h5_filename):
+            print(f"Creating HDF5 file: {self.h5_filename}")
+            os.makedirs(os.path.dirname(self.h5_filename), exist_ok=True)
+
+            # Generate mock-up inspiral data
+            nsamples = 25000
```
nilsvu: This might be a bit too much data for a quick test. Reduce to just 100 points?

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -0,0 +1,143 @@
...
+
+import numpy as np
+from click.testing import CliRunner
+
+import spectre.IO.H5 as spectre_h5
+from spectre.Informer import unit_test_build_path, unit_test_src_path
+from spectre.Visualization.PlotTrajectories import plot_trajectories_command
+
+
+class TestPlotTrajectories(unittest.TestCase):
```
nilsvu, line 18: Add this test to the `CMakeLists.txt` file in this directory, so it runs automatically on GitHub.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+def plot_trajectory(AhA, AhB, sample_rate, figsize):
+    """Plot concatenated trajectories in inspiral simulation
+
+    Outputs a 2x2 figure with 4 plots:
+    1st row: A and B trajectories in 3D, coordinate separation in 3D
+    2nd row: A and B trajectories in 2D xy plane, coordinate separation in 2D
+    xy plane. the 3D plots use a subsampled version of the data, with default
+    value of 15, to speed up the plots and avoid memory error
```
nilsvu: Add a description of the expected arguments, like this:

```
Arguments:
  AhA: Array of shape (num_points, 3) with the coordinates of the first object.
  AhB: ...
  sample_rate: ...
  figsize: ...
```

This is important so people know how to use your function.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,179 @@
...
+    "--figsize",
+    nargs=2,
+    type=float,
+    default=(10.0, 10.0),
+    show_default=True,
+    help="Figure size as width and height in inches",
+)
+@apply_stylesheet_command()
+@show_or_save_plot_command()
+def plot_trajectories_command(
```
nilsvu: Move the command below the two functions, just above `main`. That's usually the structure of a Python file so the functions are defined by the time they are used.

src/Visualization/Python/PlotTrajectories.py
```diff
@@ -0,0 +1,179 @@
...
+    "--subfile-name-ahb",
+    "-B",
+    default="ApparentHorizons/ControlSystemAhB_Centers.dat",
+    help=(
+        "Name of subfile containing the apparent horizon centers for object B."
+    ),
+)
+@click.option(
+    "--sample-rate",
+    "-sr",
```
nilsvu: We usually have single-character shortcuts. Maybe just delete this shortcut.

nilsvu (changes requested):

tests/Unit/Visualization/Python/Test_PlotTrajectories.py
```diff
@@ -12,115 +14,110 @@
 from spectre.Informer import unit_test_build_path, unit_test_src_path
 from spectre.Visualization.PlotTrajectories import plot_trajectories_command
 
+# Configure logging
+logging.basicConfig(level=logging.INFO)
```
nilsvu: Move this to `__main__` at the bottom, like this:

```py
if __name__ == "__main__":
    logging.basicConfig(level=logging.DEBUG)
    unittest.main(verbosity=2)
```

nilsvu (changes requested):
Looks great! Move this one thing, then squash everything into a single commit.

nilsvu (approved):

nilsvu:
Yay nice job 👍

PR #5978, opened 2024-05-09
Add CLI to plot basic control system quantities

knelli2:
## Proposed changes

Example of output:

debug

### Upgrade instructions

To use, build the CLI then run
```
spectre plot control-system -h
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/PlotControlSystem.py
+++ b/src/Visualization/Python/PlotControlSystem.py
@@ -1,0 +1,270 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+from typing import Optional, Sequence
+
+import click
+import h5py
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+
+from spectre.Visualization.Plot import (
+    apply_stylesheet_command,
+    show_or_save_plot_command,
+)
+from spectre.Visualization.ReadH5 import available_subfiles, to_dataframe
+
+logger = logging.getLogger(__name__)
+
+
+@click.command(name="control-system")
+@click.argument(
+    "reduction_files",
+    type=click.Path(exists=True, file_okay=True, dir_okay=False, readable=True),
+    nargs=-1,
+)
+@click.option(
+    "--with-shape/--without-shape",
+    default=True,
+    show_default=True,
+    help="Wether or not to plot shape control.",
+)
+@click.option(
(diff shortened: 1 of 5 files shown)


src/Visualization/Python/PlotControlSystem.py
```diff
@@ -0,0 +1,226 @@
...
+    axes[1].set_ylabel("Damping Time")
+    for ax in axes:
+        ax.grid()
+        ax.set_yscale("log")
+    if x_label:
+        # Bottom plot gets x label
+        axes[1].set_xlabel(x_label)
+    if title:
+        plt.title(title)
+    axes[1].legend(loc="center left", bbox_to_anchor=(1.01, 1.1))
```
knelli2: Opinions on how to better represent the shape components are welcome. Just did the easiest thing for now to quickly get it in.

nilsvu: How about an L2 norm over all shape components

knelli2: I think it will be important to be able to distinguish which component is the one that's going bad, so I don't think lumping them together would be best.

nilsvu: You could plot both the L2 norm and the highest mode then

nilsvu: Or one plot where the components of each control system are lumped together so you have only one line each for expansion, rotation, shape, size. This tells you which control system is going bad. Plus one plot for each control system with all the components. This tells you which component is going bad. So you have a figure with 1+num_control_systems rows and 1 col.

nilsdeppe: Maybe do L2 over each `m` for each `l` with an option to show all `m` modes for a specific `l` and nothing else?

knelli2: Having 1+num_control_systems plots sounds a bit crowded and also hard to compare between control systems. I do like doing the L2 over all `m` of a given `l` by default and then having an option to show all `m`.

nilsdeppe (commented):
LGTM, please squash in the `fixup` or rename it depending on how you want to structure the PR

knelli2:
@nilsdeppe Squashed and rebased

knelli2:
Here are the updated plots without/with all of the m modes

no_m

all_m

nilsdeppe (approved):

PR #5988, opened 2024-05-13
Decrease sizes of Time and TimeStepId

wthrowe:
Primarily aiming to decrease checkpoint sizes.

This is the easy optimizations.  Some harder ones that were discussed that I didn't do here are
* Specialize to power-of-two slab fractions instead of generic rationals.  This requires fixing any places that are still using more general slab fractions, and requires deciding how to deal with `TimeDelta` objects with fraction greater than one, which can show up as intermediates, particularly in tests.
* Use some bits from the `TimeStepId` slab number to store the substep.  This complicates things more than any of the changes in this PR, particularly because it involves bit-manipulating signed integers.
* Removing the `Time::value_` field and computing it when needed.  This might have performance implications as I frequently code assuming `Time::value()` is free.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/TimeStepId.hpp
+++ b/src/Time/TimeStepId.hpp
@@ -35,7 +35,7 @@ class TimeStepId {
   TimeStepId(bool time_runs_forward, int64_t slab_number, const Time& step_time,
              uint64_t substep, const TimeDelta& step_size, double substep_time);
 
-  bool time_runs_forward() const { return time_runs_forward_; }
+  bool time_runs_forward() const;
   int64_t slab_number() const { return slab_number_; }
   /// Time at the start of the current step
   const Time& step_time() const { return step_time_; }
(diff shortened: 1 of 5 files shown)


src/Time/TimeStepId.hpp
```diff
@@ -64,7 +64,7 @@ class TimeStepId {
 
   int64_t slab_number_{std::numeric_limits<int64_t>::lowest()};
   Time step_time_{};
-  uint64_t substep_{0};
+  uint8_t substep_{0};
```
nilsdeppe, line 67: Would moving this to after `substep_time_` help with reducing the class size? I assume the issue is that `Rational` is 64- or 32-bit aligned

nilsdeppe (approved):
LGTM! One small question. If you decide to do it, please squash immediately, otherwise feel free to merge this :)

src/Time/TimeStepId.hpp
wthrowe: No.  The class size has to be a multiple of the alignment, so you still get padding even if the small field is at the end.

Issue #5983, opened 2024-05-10
Tests failing when FPEs are reenabled


wthrowe:
When RunTests was split into multiple executables, a chunk of the initialization code was lost, most notably enabling FPEs in most of the unit tests.  I've reenabled then in #5982, but this causes tests to fail.  This issue is to organize getting those fixed so the PR can be merged.  I've attempted to identify the causes of the failing tests and who made the changes.  Apologies if I've misidentified a cause.  You should be able to reproduce the problem by checking out my PR.

@isaaclegred:
* Unit.GrMhd.ValenciaDivClean.PrimitiveFromConservative - overflow in Kastaun

@knelli2:
* Unit.ControlSystem.TimescaleTuner - overflow when `decrease_timescale_threshold` is large

@nilsdeppe:
* Unit.Evolution.Systems.GhValenciaDivClean.Subcell.FixConsAndComputePrims - SIMD
* Unit.Evolution.Systems.ValenciaDivClean.Subcell.FixConsAndComputePrims - SIMD
* Unit.Evolution.GrMhd.ValenciaDivClean.FixConservatives - SIMD
* Unit.Numerical.RootFinding.TOMS748 - presumably related to the SIMD rewrite, although the first failure is not in SIMD code

@Sizheng-Ma:
* Unit.Evolution.Systems.Cce.Actions.KleinGordonH5BoundaryCommunication - sets the final time to NaN

Don't know:
* Unit.Evolution.Systems.ValenciaDivClean.Subcell.TciOnDgGrid - underflow in Kastaun

nilsdeppe:
#6265 I think generally addressed this. There were a couple MC tests I had to disable, but we are not using that code in any evolution and the issues will need to get resolved for that anyway.

@wthrowe are you okay closing this as done?

wthrowe:
Yeah, this is done.

PR #5989, opened 2024-05-14
Fix overflow FPE in TimescaleTuner test

knelli2:
## Proposed changes

Partially addresses #5983.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/ControlSystem/Test_TimescaleTuner.cpp
+++ b/tests/Unit/ControlSystem/Test_TimescaleTuner.cpp
@@ -17,8 +17,10 @@
 namespace {
 template <bool AllowDecrease>
 void test_increase_or_decrease() {
+  // If we don't allow decrease, then just set the threshold to infinity. We
+  // can't set it to max otherwise we'll get an FPE.
   const double decrease_timescale_threshold =
-      AllowDecrease ? 1.0e-2 : std::numeric_limits<double>::max();
+      AllowDecrease ? 1.0e-2 : std::numeric_limits<double>::infinity();
   const double increase_timescale_threshold = 1.0e-4;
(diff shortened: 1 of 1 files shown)


tests/Unit/ControlSystem/Test_TimescaleTuner.cpp
```diff
@@ -189,8 +189,10 @@ void test_increase_or_decrease() {
 
 template <bool AllowDecrease>
 void test_no_change_to_timescale() {
-  const double decrease_timescale_threshold =
-      AllowDecrease ? 1.0e-2 : std::numeric_limits<double>::max();
+  // If we don't allow decrease, then just set the threshold to a super large
+  // value (larger than any test value we'd use). We can't set it to inf or max
+  // otherwise we'll get an FPE.
```
wthrowe: Infinity doesn't FPE.  Don't know whether it does what you want, but the test passes with it.

wthrowe (changes requested):

tests/Unit/ControlSystem/Test_TimescaleTuner.cpp
knelli2: It just needs to be a large number so that it will never decrease. Infinity works fine for this so I changed it to that.

wthrowe: If you're setting it to infinity, remove the comment that says you can't set it to infinity.

knelli2: Oops!

wthrowe (approved):
Looks good.  Thanks!

PR #5996, opened 2024-05-15
Add IntegratedFunctionOfTime

nikwit:
This class is used to adjust the time dependent maps in such a way that the worldtube always perfectly tracks the position of the particle (the position of which is determined by solving an ODE). 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
+++ b/src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
@@ -1,0 +1,159 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp"
+
+#include <algorithm>
+#include <cmath>
+#include <iterator>
+#include <memory>
+#include <ostream>
+
+#include "DataStructures/DataVector.hpp"
+#include "Domain/FunctionsOfTime/ThreadsafeList.hpp"
+#include "Domain/FunctionsOfTime/ThreadsafeList.tpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace domain::FunctionsOfTime {
+
+IntegratedFunctionOfTime::IntegratedFunctionOfTime() = default;
+IntegratedFunctionOfTime::IntegratedFunctionOfTime(IntegratedFunctionOfTime&&) =
+    default;
+IntegratedFunctionOfTime::IntegratedFunctionOfTime(
+    const IntegratedFunctionOfTime&) = default;
+auto IntegratedFunctionOfTime::operator=(IntegratedFunctionOfTime&&)
+    -> IntegratedFunctionOfTime& = default;
+auto IntegratedFunctionOfTime::operator=(const IntegratedFunctionOfTime&)
+    -> IntegratedFunctionOfTime& = default;
+IntegratedFunctionOfTime::~IntegratedFunctionOfTime() = default;
+IntegratedFunctionOfTime::IntegratedFunctionOfTime(
+    CkMigrateMessage* /*unused*/) {}
+
+IntegratedFunctionOfTime::IntegratedFunctionOfTime(
+    const double t, std::array<double, 2> initial_func_and_derivs,
+    const double expiration_time, const bool rotation)
+    : deriv_info_at_update_times_(t), rotation_(rotation) {
+  deriv_info_at_update_times_.insert(t, initial_func_and_derivs,
+                                     expiration_time);
+}
+
+std::array<DataVector, 1> IntegratedFunctionOfTime::func(double t) const {
+  return func_and_derivs<0>(t);
+}
+std::array<DataVector, 2> IntegratedFunctionOfTime::func_and_deriv(
+    double t) const {
+  return func_and_derivs<1>(t);
+}
+[[noreturn]] std::array<DataVector, 3>
+IntegratedFunctionOfTime::func_and_2_derivs(double /*t*/) const {
+  ERROR("Can only return first derivative.");
+}
+
+std::unique_ptr<FunctionOfTime> IntegratedFunctionOfTime::get_clone() const {
+  return std::make_unique<IntegratedFunctionOfTime>(*this);
+}
+
+template <size_t MaxDerivReturned>
+std::array<DataVector, MaxDerivReturned + 1>
+IntegratedFunctionOfTime::func_and_derivs(const double t) const {
+  static_assert(MaxDerivReturned <= 1, "can only return 1 derivative");
+
+  const auto deriv_info_at_t = deriv_info_at_update_times_(t);
+  std::array<DataVector, MaxDerivReturned + 1> result{};
+
+  if (rotation_) {
+    result.at(0) = DataVector{cos(deriv_info_at_t.data[0] / 2.), 0., 0.,
+                              sin(deriv_info_at_t.data[0] / 2.)};
+    if constexpr (MaxDerivReturned == 1) {
+      result.at(1) = DataVector{
+          -sin(deriv_info_at_t.data[0] / 2.) * deriv_info_at_t.data[1] / 2., 0.,
+          0., cos(deriv_info_at_t.data[0] / 2.) * deriv_info_at_t.data[1] / 2.};
+    }
+  } else {
+    for (size_t i = 0; i <= MaxDerivReturned; ++i) {
+      result.at(i) = DataVector{gsl::at(deriv_info_at_t.data, i)};
+    }
+  }
+  return result;
+}
+
+void IntegratedFunctionOfTime::update(const double time_of_update,
+                                      DataVector updated_value_and_derivative,
+                                      const double next_expiration_time) {
+  if (time_of_update < deriv_info_at_update_times_.expiration_time()) {
+    ERROR("Attempted to update from time "
+          << time_of_update << " to time" << next_expiration_time
+          << " but expiration time is "
+          << deriv_info_at_update_times_.expiration_time());
+  }
+
+  ASSERT(updated_value_and_derivative.size() == 2,
+         "The size of the DataVector should be 2: value and derivative");
+  if (time_of_update != deriv_info_at_update_times_.expiration_time()) {
+    update_backlog_[time_of_update] = std::make_pair(
+        std::move(updated_value_and_derivative), next_expiration_time);
+    return;
+  }
+  const std::array<double, 2> func{updated_value_and_derivative[0],
+                                   updated_value_and_derivative[1]};
+  deriv_info_at_update_times_.insert(time_of_update, func,
+                                     next_expiration_time);
+  while (not update_backlog_.empty() and
+         update_backlog_.begin()->first == time_bounds()[1]) {
+    const double current_exp_time = time_bounds()[1];
+    const auto& item = update_backlog_[current_exp_time];
+    const std::array<double, 2> updated_func{item.first[0], item.first[1]};
+    deriv_info_at_update_times_.insert(current_exp_time, updated_func,
+                                       item.second);
+    update_backlog_.erase(current_exp_time);
+  }
+  deriv_info_at_update_times_.truncate_to_length(100);
+}
+
+std::array<double, 2> IntegratedFunctionOfTime::time_bounds() const {
+  return {{deriv_info_at_update_times_.initial_time(),
+           deriv_info_at_update_times_.expiration_time()}};
+}
+
+double IntegratedFunctionOfTime::expiration_after(const double time) const {
+  return deriv_info_at_update_times_.expiration_after(time);
+}
+
+void IntegratedFunctionOfTime::pup(PUP::er& p) {
+  FunctionOfTime::pup(p);
+  size_t version = 0;
+  p | version;
+  if (version >= 0) {
+    p | deriv_info_at_update_times_;
+    p | rotation_;
+    p | update_backlog_;
+  }
+}
+
+bool operator==(const IntegratedFunctionOfTime& lhs,
+                const IntegratedFunctionOfTime& rhs) {
+  return lhs.deriv_info_at_update_times_ == rhs.deriv_info_at_update_times_ and
+         lhs.rotation_ == rhs.rotation_ and
+         lhs.update_backlog_ == rhs.update_backlog_;
+}
+
+bool operator!=(const IntegratedFunctionOfTime& lhs,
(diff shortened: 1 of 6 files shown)


src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/ThreadsafeList.tpp"
+#include "Utilities/Serialization/CharmPupable.hpp"
+#include "Utilities/StdHelpers.hpp"
+
+namespace domain::FunctionsOfTime {
+/*!
+ * \brief A function that is integrated manually.
+ * \details This function only works with global time steppers that have
+ * strictly positively increasing time steps. When evaluated at this global time
```
wthrowe: "time steps" -> "substeps" or "stages" or "substep times" or something.

I can get you a list of these that can be used in the `factory_creation` struct to enforce this at option-parsing time.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
```diff
@@ -0,0 +1,102 @@
...
+  /*!
+   * \brief Returns the function and derivative at the last global time step
+   * before time `t`.
+   */
+  std::array<DataVector, 2> func_and_deriv(double t) const override {
+    return func_and_derivs<1>(t);
+  }
+  std::array<DataVector, 3> func_and_2_derivs(double /*t*/) const override {
+    ERROR("Can only return first derivative.");
+    return {};
```
wthrowe: Remove code after `ERROR`, and move this to the cpp file.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
```diff
@@ -0,0 +1,102 @@
...
+#include <array>
+#include <cstddef>
+#include <map>
+#include <memory>
+#include <ostream>
+#include <pup.h>
+
+#include "DataStructures/DataVector.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/ThreadsafeList.tpp"
```
wthrowe: Include the hpp in the header and the tpp in the cpp.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,143 @@
...
+}
+
+std::unique_ptr<FunctionOfTime> IntegratedFunctionOfTime::get_clone() const {
+  return std::make_unique<IntegratedFunctionOfTime>(*this);
+}
+
+template <size_t MaxDerivReturned>
+std::array<DataVector, MaxDerivReturned + 1>
+IntegratedFunctionOfTime::func_and_derivs(const double t) const {
+  static_assert(MaxDerivReturned <= 1, "can only return 2 derivatives");
```
wthrowe: "1 derivative"

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,143 @@
...
+
+  auto deriv_info_at_t = deriv_info_at_update_times_(t);
```
wthrowe: ERROR if `deriv_info_at_t.expiration != t and t != deriv_info_at_times_.initial_time()`.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,143 @@
...
+                                      const double next_expiration_time) {
+  if (time_of_update < deriv_info_at_update_times_.expiration_time()) {
+    ERROR("Attempted to update from time "
+          << time_of_update << " to time" << next_expiration_time
+          << " but expiration time is "
+          << deriv_info_at_update_times_.expiration_time());
+  }
+
+  ASSERT(updated_value_and_derivative.size() == 2,
+         "The size of the DataVector should be 2: position and velocity");
```
wthrowe: Not position and velocity, generally.

src/Domain/FunctionsOfTime/RegisterDerivedWithCharm.cpp
```diff
@@ -29,6 +30,7 @@ void register_derived_with_charm() {
                               FunctionsOfTime::QuaternionFunctionOfTime<3>,
                               FunctionsOfTime::QuaternionFunctionOfTime<4>,
                               FunctionsOfTime::SettleToConstant,
-                              FunctionsOfTime::SettleToConstantQuaternion>();
+                              FunctionsOfTime::SettleToConstantQuaternion,
+                              FunctionsOfTime::IntegratedFunctionOfTime>();
```
wthrowe: Alphabetize.

tests/Unit/Domain/FunctionsOfTime/Test_IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include <limits>
+#include <memory>
+#include <random>
+#include <vector>
+
+#include "DataStructures/DataVector.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/RegisterDerivedWithCharm.hpp"
+#include "Domain/FunctionsOfTime/ThreadsafeList.tpp"
```
wthrowe: Remove this include.

tests/Unit/Domain/FunctionsOfTime/Test_IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,170 @@
...
+#include "Framework/TestHelpers.hpp"
+#include "Helpers/DataStructures/MakeWithRandomValues.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Serialization/Serialize.hpp"
+
+namespace domain {
+namespace {
+void test(const gsl::not_null<FunctionsOfTime::FunctionOfTime*> f_of_t,
```
wthrowe, line 25: Define this in the function instead of taking it as an argument, since the function won't work if it isn't pointing to the same object as the `f_of_t_derived` argument.  (Or take this one and do the cast to get the other.)

tests/Unit/Domain/FunctionsOfTime/Test_IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,170 @@
...
+        std::array<DataVector, 2>{DataVector{2.5}, DataVector{3.}});
+}
+}  // namespace
+
+SPECTRE_TEST_CASE("Unit.Domain.FunctionsOfTime.IntegratedFunctionOfTime",
+                  "[Domain][Unit]") {
+  FunctionsOfTime::register_derived_with_charm();
+  MAKE_GENERATOR(gen);
+  std::uniform_real_distribution<double> dist(-10., 10.);
+  const DataVector used_for_size(300);
```
wthrowe: Use a `size_t` rather than an uninitialized `DataVector`.  The `make_with_random_values` function will accept either.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,143 @@
...
+    const IntegratedFunctionOfTime&) = default;
+auto IntegratedFunctionOfTime::operator=(IntegratedFunctionOfTime&&)
+    -> IntegratedFunctionOfTime& = default;
+auto IntegratedFunctionOfTime::operator=(const IntegratedFunctionOfTime&)
+    -> IntegratedFunctionOfTime& = default;
+IntegratedFunctionOfTime::~IntegratedFunctionOfTime() = default;
+
+IntegratedFunctionOfTime::IntegratedFunctionOfTime(
+    const double t, std::array<double, 2> initial_func_and_derivs,
+    const double expiration_time, bool rotation)
```
wthrowe: `const bool`

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,143 @@
...
+    : deriv_info_at_update_times_(t), rotation_(rotation) {
+  deriv_info_at_update_times_.insert(t, std::move(initial_func_and_derivs),
```
wthrowe: Several times: `std::array<double, 2>` is trivially copyable, so don't `std::move` it.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,143 @@
...
+template <size_t MaxDerivReturned>
+std::array<DataVector, MaxDerivReturned + 1>
+IntegratedFunctionOfTime::func_and_derivs(const double t) const {
+  static_assert(MaxDerivReturned <= 1, "can only return 2 derivatives");
+
+  auto deriv_info_at_t = deriv_info_at_update_times_(t);
+  std::array<DataVector, MaxDerivReturned + 1> result{};
+
+  if (rotation_) {
+    result.at(0) = DataVector{cos(deriv_info_at_t.data[0] / 2.), 0., 0.,
```
wthrowe, line 67: `#include <cmath>`

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
```diff
@@ -0,0 +1,102 @@
...
+  ~IntegratedFunctionOfTime() override;
+  /*!
+   * \brief Constructs the function using the initial time, initial values,
+   * derivative and expiration time. If `rotation` is true, the function will be
+   * converted to a quaternion before it is output as used by the `Rotation`
+   * map.
+   */
+  IntegratedFunctionOfTime(double t,
+                           std::array<double, 2> initial_func_and_derivs,
+                           double expiration_time, bool rotation);
```
wthrowe, line 43: Test the `rotation = true` code.

wthrowe (changes requested):

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
knelli2: *tpp and hpp both in the cpp?

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
```diff
@@ -0,0 +1,143 @@
...
+#include "DataStructures/DataVector.hpp"
+#include "Domain/FunctionsOfTime/ThreadsafeList.tpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace domain::FunctionsOfTime {
+
+IntegratedFunctionOfTime::IntegratedFunctionOfTime() = default;
+IntegratedFunctionOfTime::IntegratedFunctionOfTime(IntegratedFunctionOfTime&&) =
```
knelli2, line 22: Can these just go in the hpp? Or did clang-tidy yell at you?

wthrowe: They need the tpp definitions visible.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
wthrowe: I don't think we ever decided definitively on a rule for this, but including both in the cpp is certainly fine.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
nikwit: that would be great, thanks

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
nikwit: do you mean greater than expiration or lower than initial? `ThreadSafeList` throws an error but I can add another one here.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
nikwit: The class is not templated so I get a linker error that it cannot find the default constructor of `ThreadsafeList` unless I include `ThreadSafeList.tpp` in `IntegreatedFunctionOfTime.hpp`

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
wthrowe: No, I mean what I wrote.  This class doesn't work correctly with dense output, so you should check that it is not being used for such.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
wthrowe: You need to move the `CkMigrateMessage` constructor definition into the cpp file.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp
wthrowe: Added in #6021 as `increasing_substep_time_steppers`, which can be used in the `factory_creation` struct.  The answer turns out to be `AdamsBashforth` (trivially), `Rk3Owren`, and `Rk4Owren`.

src/Domain/FunctionsOfTime/IntegratedFunctionOfTime.cpp
nikwit: Right, but I will set the update times in between the substep times. This way I dont have to worry about which interval will be selected at the boundary due to roundoff.

wthrowe: OK.  This is fine as-is, then.

knelli2 (commented):
Looks good. @wthrowe want to take another look at this?

wthrowe (approved):
Looks good.  Squash.

nikwit:
I squashed it in. Thanks for the reviews and sorry for the delays, my thesis deadline is coming up.

knelli2 (approved):

PR #5999, opened 2024-05-15
Redesign step/slab size changing

wthrowe:
The biggest improvements are completely avoiding roundoff error in some portions of the algorithm (mainly StepToTimes), and fixing the annoyance of things like control system limits gradually decreasing the time step over time in GTS mode.

### Upgrade instructions

When changing slab sizes, including step size with global time-stepping, some effects now only change the size for a single step, after which the step will revert to the previous value.  These include the StepChoosers documented as limiting the step (as opposed to setting a goal) and the GTS control-system step limit.

The "Increase" StepChooser has been renamed to "LimitIncrease" to reflect this change, and "Constant" has been split into "Constant", which sets a goal, and "Maximum", which does not.  Check the documentation or help text to see which type of change other StepChoosers now apply.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (dismissed):

nilsdeppe:
@wthrowe this has conflicts now that need to be resolved :(

nilsdeppe (approved):

PR #6000, opened 2024-05-16
Reduce the size of Mesh, Basis & Quadrature, Direction & Size, ElementId, FixedHashMap, and combine ElementId and DirectionalId

nilsdeppe:
## Proposed changes

After looking at memory consumption, @kidder, @wthrowe and I were surprised how much these small classes ended up using. This is because we have a lot of them stored per element. Using smaller integer types greatly reduces the memory usage, and then some bit twiddling is done internally in a few classes to further reduce memory usage. This should also help by reducing memory traffic (being able to load an entire Mesh in 1 load instead of 6, etc.). The bit twiddling is just simple bitmasks so nothing too crazy.

One thing I found quite inconvenient and a bit silly is that we version tiny classes like `Direction`. We should instead be versioning the Domain during serialization, not all these small classes. In some cases we used more bytes during serialization for the version than the actual data. That seems unnecessary. In cases where I could, I compressed the version info into unused bits by other parts of the class. This helps, but still seems silly.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/FixedHashMap.hpp
+++ b/src/DataStructures/FixedHashMap.hpp
@@ -181,17 +181,7 @@ class FixedHashMap {
   hasher hash_function() const { return hasher{}; }
 
   // NOLINTNEXTLINE(google-runtime-references)
-  void pup(PUP::er& p) {
-    size_t version = 0;
-    p | version;
-    // Remember to increment the version number when making changes to this
-    // function. Retain support for unpacking data written by previous versions
-    // whenever possible. See `Domain` docs for details.
-    if (version >= 0) {
-      p | data_;
-      p | size_;
-    }
-  }
+  void pup(PUP::er& p);
 
  private:
   template <size_t FMaxSize, class FKey, class FValueType, class FHash,
@@ -233,8 +223,52 @@ class FixedHashMap {
   }
 
   storage_type data_{};
-  size_t size_ = 0;
+  tmpl::conditional_t<(MaxSize <= 255), uint8_t, size_t> size_ = 0;
 };
+
+template <size_t MaxSize, class Key, class ValueType, class Hash,
+          class KeyEqual>
+void FixedHashMap<MaxSize, Key, ValueType, Hash, KeyEqual>::pup(PUP::er& p) {
+  // Remember to increment the version number when making changes to this
+  // function. Retain support for unpacking data written by previous versions
+  // whenever possible. See `Domain` docs for details.
+  const uint8_t current_version = 1;
+  if (p.isUnpacking()) {
+    uint8_t version_read{0};
+    p | version_read;
+    if (LIKELY(version_read > 0)) {
+      p | data_;
+      p | size_;
+    } else {
+      // Read next 3 bytes.
+      p | version_read;
+      p | version_read;
+      p | version_read;
+      // Read "bottom" of version
+      uint32_t old_version{0};
+      p | old_version;
+      if (UNLIKELY(old_version > 0)) {
+        ERROR(
+            "Incompatible version format for Direction. Expected to receive "
+            "version 0 but got "
+            << old_version);
+      }
+      p | data_;
+      if constexpr (MaxSize <= 255) {
+        size_t read_size{0};
+        p | read_size;
+        size_ = static_cast<uint8_t>(read_size);
+      } else {
+        p | size_;
+      }
+    }
+  } else {
+    uint8_t version_write = current_version;
+    p | version_write;
+    p | data_;
+    p | size_;
+  }
+}
 
 /// \cond HIDDEN_SYMBOLS
 template <size_t MaxSize, class Key, class ValueType, class Hash,
--- a/src/Domain/Structure/Direction.cpp
+++ b/src/Domain/Structure/Direction.cpp
@@ -7,56 +7,76 @@
 #include <pup.h>
 #include <pup_stl.h>
 
+#include "Domain/Structure/Side.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+template <size_t VolumeDim>
+Direction<VolumeDim> Direction<VolumeDim>::self() {
+  return Direction<VolumeDim>{Axis::Xi, Side::Self};
+}
+
+template <size_t VolumeDim>
+Direction<VolumeDim>::Direction(Axis axis, Side side)
+    : bit_field_(static_cast<uint8_t>(axis) bitor static_cast<uint8_t>(side)) {
+  ASSERT(bit_field_ < 0b10000, "Direction is too large.");
+}
 
 template <size_t VolumeDim>
 void Direction<VolumeDim>::pup(PUP::er& p) {
-  size_t version = 0;
-  p | version;
   // Remember to increment the version number when making changes to this
   // function. Retain support for unpacking data written by previous versions
   // whenever possible. See `Domain` docs for details.
-  if (version >= 0) {
-    p | axis_;
-    p | side_;
+  const uint8_t current_version = 1 << 6;
+  if (p.isUnpacking()) {
+    uint8_t data = 0;
+    p | data;
+    const uint8_t version_read = (0b11000000 bitand data) >> 6;
+    if (LIKELY(version_read > 0)) {
+      bit_field_ = data bitand 0b00001111;
+    } else {
+      // Read next 3 bytes.
+      p | data;
+      p | data;
+      p | data;
+      // Read "bottom" of version
+      uint32_t old_version{0};
+      p | old_version;
+      if (UNLIKELY(old_version > 0)) {
+        ERROR(
+            "Incompatible version format for Direction. Expected to receive "
+            "version 0 but got "
+            << old_version);
+      }
+      int read_axis{0};
+      p | read_axis;
+      if (UNLIKELY(read_axis < 0)) {
+        ERROR("Expected a non-negative axis value but read " << read_axis);
+      }
+      int read_side{0};
+      p | read_side;
+      if (UNLIKELY(read_side < 0)) {
+        ERROR("Expected a non-negative side value but read " << read_side);
+      }
+      // add 1 to account for the added Uninitialized
+      const auto side = static_cast<uint8_t>(read_side + 1)
+                        << detail::side_shift;
+      bit_field_ = 0;
+      bit_field_ = static_cast<uint8_t>(read_axis) bitor side;
+    }
+  } else {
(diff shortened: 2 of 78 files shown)


src/NumericalAlgorithms/Spectral/Basis.hpp
```diff
@@ -40,16 +43,23 @@ namespace Spectral {
  * a spherical harmonic basis.  By convention, the first dimension represents
  * the polar/zentith angle (or colatitude), while the second dimension
  * represents the azimuthal angle (or longitude)
+ *
+ * \note We store these effectively as a 4-bit integer using the highest 4
+ * bits of a uint8_t, which is why we do the left shift.  We cannot
+ * have more than 16 bases to fit into the 4 bits, including the
+ * `Uninitialized` value. The number of bits to shift is encoded in the variable
+ * `Spectral::detail::basis_shift`.
```
wthrowe, line 53: `detail` stuff should not be part of the public interface.  Either make the constant non-detail or do the shift internally when using the enum instead of incorporating it into the value.

src/NumericalAlgorithms/Spectral/Mesh.hpp
```diff
@@ -250,9 +229,25 @@ class Mesh {
...
+  static Spectral::Quadrature extract_quadrature(uint8_t bits);
+
+  template <size_t LocalDim>
+  friend class Mesh;
+
+  template <size_t LocalDim>
+  friend bool operator==(const Mesh<LocalDim>& lhs, const Mesh<LocalDim>& rhs);
+
+  // - 8 bits per extent = 3 * 8 = 24 bits = 3 bytes
+  //   This limits us to 128 extent per direction. Since FD is 2x DG, this
```
wthrowe: Where does 128 come from?

src/NumericalAlgorithms/Spectral/Mesh.hpp
```diff
@@ -250,9 +229,25 @@ class Mesh {
...
+  //   means at most 64 DG points, which is should be fine.
+  // - We encode the basis & quadrature as two sets of 4 bits in a uint8_t.
+  std::array<uint8_t, 6> bit_field_{
```
wthrowe: I think this would be easier to read if you had two arrays of size 3 instead of one array of size 6 with the first and second halves interpreted differently.

src/NumericalAlgorithms/Spectral/Mesh.hpp
```diff
@@ -250,9 +229,25 @@ class Mesh {
   void pup(PUP::er& p);
 
  private:
-  Index<Dim> extents_{};
-  std::array<Spectral::Basis, Dim> bases_{};
-  std::array<Spectral::Quadrature, Dim> quadratures_{};
+  static Spectral::Basis extract_basis(uint8_t bits);
+  static Spectral::Quadrature extract_quadrature(uint8_t bits);
```
wthrowe: [optional] Bit of a weird asymmetry in that the combining function is in an anonymous namespace but the extracting functions are private statics.  I'd make them them all anonymous.

tests/Unit/NumericalAlgorithms/Spectral/Test_Mesh.cpp
```diff
@@ -95,6 +95,7 @@ void test_explicit_choices_per_dimension() {
   test_extents_basis_and_quadrature(mesh1d, {{2}},
                                     {{Spectral::Basis::Legendre}},
                                     {{Spectral::Quadrature::GaussLobatto}});
+  return;
   CHECK(mesh1d.slice_away(0) == Mesh<0>{});
```
wthrowe: Fix something.

src/Domain/Structure/Direction.cpp
```diff
@@ -7,56 +7,81 @@
...
-template <>
-Direction<1>::Direction(const size_t dimension, const Side side) {
-  ASSERT(
-      0 == dimension,
-      "dim = " << dimension << ", for Direction<1> only dim = 0 is allowed.");
-  axis_ = Axis::Xi;
-  side_ = side;
+template <size_t VolumeDim>
+uint8_t Direction<VolumeDim>::self_bits() {
+  return static_cast<uint8_t>(Axis::Xi) bitor static_cast<uint8_t>(Side::Self);
```
wthrowe: [optional] `return self().bits();` to reduce duplication.  (Or even get rid of this function entirely.)

src/Domain/Structure/ElementId.cpp
```diff
@@ -89,7 +92,23 @@ ElementId<VolumeDim>::ElementId(const size_t block_id,
 }
 
 template <size_t VolumeDim>
-ElementId<VolumeDim>::ElementId(const std::string& grid_name) {
+ElementId<VolumeDim>::ElementId(const Direction<VolumeDim>& direction,
+                                const ElementId<VolumeDim>& element_id)
+    : block_id_(element_id.block_id_),
+      grid_index_(element_id.grid_index_),
+      // NOLINTNEXTLINE(cppcoreguidelines-pro-type-reinterpret-cast)
+      direction_(*reinterpret_cast<const uint8_t*>(&direction)),
```
wthrowe: `direction_(direction.bits())`

wthrowe (changes requested):

src/NumericalAlgorithms/Spectral/Basis.hpp
nilsdeppe: I don't see how the detail is currently part of the interface. If you are adding or removing values from the enum class, then you need to know why it is being shifted and the limitations, as in, you've entered the realm of "details". If you are just using it, you shouldn't be casting an enum class anyway. Should I make this a `\warning` or preface it with `If you are changing the values in this` or something?

src/NumericalAlgorithms/Spectral/Mesh.hpp
nilsdeppe: Heh, probably a typo when computing 2^8. I think the limitation is 255 in practice on FD, which would be a limit of 127 on DG. Does that clarify? (If so, I'll update the comments)

src/NumericalAlgorithms/Spectral/Basis.hpp
wthrowe: You use it in `src/IO/H5/VolumeData.cpp`.

src/NumericalAlgorithms/Spectral/Basis.hpp
```diff
@@ -17,9 +17,11 @@ struct create_from_yaml;
 /// \endcond
 
 namespace Spectral {
-namespace detail {
+/// \brief The amount we left-shift the basis integers by.
+///
+/// We do this to be able to represent the combine Basis and Quadrature as one
```
kidder: combined

src/Domain/Structure/Direction.cpp
```diff
@@ -7,56 +7,81 @@
...
+      bit_field_ = data bitand 0b00001111;
+    } else {
+      // Read next 3 bytes.
+      p | data;
+      p | data;
+      p | data;
+      // Read "bottom" of version
+      uint32_t old_version{0};
+      p | old_version;
+      if (UNLIKELY(version_read > 0)) {
```
kidder: you are in the else clause of the same conditional...  confused...  should be old_version?

src/Domain/Structure/Direction.hpp
```diff
@@ -25,28 +25,35 @@ class Direction {
...
-  Axis axis() const { return axis_; }
+  Axis axis() const { return static_cast<Axis>(bit_field_ bitand 0b0011); }
 
   /// The side of the Direction
-  Side side() const { return side_; }
+  Side side() const { return static_cast<Side>(bit_field_ bitand 0b1100); }
 
   /// The sign for the normal to the Side.
-  double sign() const { return (Side::Lower == side_ ? -1.0 : 1.0); }
+  double sign() const { return (Side::Lower == side() ? -1.0 : 1.0); }
```
kidder: this will give 1.0 for Side::Self and Side::Uninitialized, should we add an ASSERT?

src/Domain/Structure/ElementId.hpp
```diff
@@ -85,6 +98,8 @@ class ElementId {
   /// Create an ElementId from its string representation (see `operator<<`).
   ElementId(const std::string& grid_name);
 
+  /// Create a directioned ElementId from a Direction and an ElementId.
+
```
kidder: huh?

src/DataStructures/FixedHashMap.hpp
```diff
@@ -233,9 +223,53 @@ class FixedHashMap {
...
+      p | size_;
+    } else {
+      // Read next 3 bytes.
+      p | version_read;
+      p | version_read;
+      p | version_read;
+      // Read "bottom" of version
+      uint32_t old_version{0};
+      p | old_version;
+      if (UNLIKELY(version_read > 0)) {
```
kidder: old_version ?

kidder (changes requested):

src/Domain/Structure/Direction.hpp
nilsdeppe: Yea, good idea. I'll add an `ASSERT`

nilsdeppe:
Rebased and pushed fixups. Thanks for the reviews!

kidder (commented):
streaming in the ASSERT is not working on the failing build...

otherwise you have my permission to squash once you have the okay from @wthrowe

src/NumericalAlgorithms/Spectral/Mesh.cpp
```diff
@@ -258,7 +258,9 @@ std::array<Mesh<1>, Dim> Mesh<Dim>::slices() const {
 
 template <size_t Dim>
 void Mesh<Dim>::pup(PUP::er& p) {
-  p | bit_field_;
+  p | extents_;
+  p | quadrature_and_basis_;
+  // p | bit_field_;
```
wthrowe: Remove.

src/NumericalAlgorithms/Spectral/Mesh.cpp
```diff
@@ -274,7 +276,9 @@ bool is_isotropic(const Mesh<Dim>& mesh) {
 
 template <size_t Dim>
 bool operator==(const Mesh<Dim>& lhs, const Mesh<Dim>& rhs) {
-  return lhs.bit_field_ == rhs.bit_field_;
+  // return lhs.bit_field_ == rhs.bit_field_;
```
wthrowe: Remove.

wthrowe (changes requested):
Squash.

nilsdeppe:
Okay, done! I verified that the test failure is fixed locally so 🤞

kidder (approved):

wthrowe (approved):

PR #6004, opened 2024-05-17
Register initial magnetic field in EvolveGhValenciaDivClean with Charm.

ncorsobh:
## Proposed changes

Registers `InitialMagneticField` with Charm in EvolveGhValenciaDivClean. Necessary for evolving a magnetized TOV star with GH.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticData/GrMhd/InitialMagneticFields/InitialMagneticField.hpp
+++ b/src/PointwiseFunctions/AnalyticData/GrMhd/InitialMagneticFields/InitialMagneticField.hpp
@@ -12,10 +12,6 @@
 
 /// \cond
 class DataVector;
-namespace grmhd::AnalyticData::InitialMagneticFields {
-class Poloidal;
-class Toroidal;
-}  // namespace grmhd::AnalyticData::InitialMagneticFields
 
 namespace gsl {
 template <typename T>
@@ -68,8 +64,6 @@ class InitialMagneticField : public PUP::able {
   InitialMagneticField() = default;
 
  public:
-  using creatable_classes = tmpl::list<Poloidal, Toroidal>;
-
   ~InitialMagneticField() override = default;
 
   virtual auto get_clone() const -> std::unique_ptr<InitialMagneticField> = 0;
--- a/tests/Unit/PointwiseFunctions/AnalyticData/GrMhd/InitialMagneticFields/Test_Poloidal.cpp
+++ b/tests/Unit/PointwiseFunctions/AnalyticData/GrMhd/InitialMagneticFields/Test_Poloidal.cpp
@@ -44,7 +44,7 @@ struct PoloidalProxy : Poloidal {
 SPECTRE_TEST_CASE(
     "Unit.PointwiseFunctions.AnalyticData.GrMhd.InitialMagneticFields.Poloidal",
     "[Unit][PointwiseFunctions]") {
-  register_derived_classes_with_charm<InitialMagneticField>();
+  register_classes_with_charm<Poloidal>();
   // test creation
   const auto factory_solution = serialize_and_deserialize(
       TestHelpers::test_factory_creation<InitialMagneticField, Poloidal>(
@@ -70,7 +70,6 @@ SPECTRE_TEST_CASE(
   }
 
   // test derived
-  register_classes_with_charm<Poloidal>();
   const std::unique_ptr<InitialMagneticField> base_ptr =
       std::make_unique<Poloidal>();
   const std::unique_ptr<InitialMagneticField> deserialized_base_ptr =
--- a/tests/Unit/PointwiseFunctions/AnalyticData/GrMhd/Test_MagnetizedTovStar.cpp
+++ b/tests/Unit/PointwiseFunctions/AnalyticData/GrMhd/Test_MagnetizedTovStar.cpp
@@ -12,6 +12,7 @@
 #include "NumericalAlgorithms/Spectral/LogicalCoordinates.hpp"
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
 #include "PointwiseFunctions/AnalyticData/AnalyticData.hpp"
+#include "PointwiseFunctions/AnalyticData/GrMhd/InitialMagneticFields/Factory.hpp"
 #include "PointwiseFunctions/AnalyticData/GrMhd/InitialMagneticFields/InitialMagneticField.hpp"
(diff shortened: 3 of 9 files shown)


wthrowe (changes requested):
Instead of doing this, switch the class to use the `factory_creation` struct in the metavariables instead of the deprecated `creatable_classes` feature.  Look at how it's done for `TimeStepper`, for example.

nilsdeppe:
This was intentionally not using the factory creation. I think it should stay as-is. The factory creation struct is tedious when you know all the derived classes ahead of time. We've discussed that the creatable classes is _not_ deprecated a long time ago and serves a distinct purpose from the factory class approach. There's a lot of code that intentionally doesn't use that and I think should stay that way.

nilsdeppe (dismissed):
Looks great! Thanks for doing this!

wthrowe:
> This was intentionally not using the factory creation. I think it should stay as-is. The factory creation struct is tedious when you know all the derived classes ahead of time. We've discussed that the creatable classes is _not_ deprecated a long time ago and serves a distinct purpose from the factory class approach. There's a lot of code that intentionally doesn't use that and I think should stay that way.

Who's "we"?  I don't recall being involved in any such conversation.  The code clearly marks it as a legacy interface.

The `factory_creation` method requires, depending on how you count, either less code or about the same as `creatable_classes`, so I don't see how it's more tedious.

nilsdeppe:
This was discussed in either the Thursday or Monday call maybe 3-4 years ago. Definitely quite a while ago and you were present. My objection to removing the `using createable_classes` is that there are places in the code where we can currently do explicit instantiations because we know all the derived classes _without_ the metavariables. Requiring the metavariables means more code can only get compiled as part of the executable. I'm fine with using the factory struct being used as `tmpl::pair<Base, typename Base::createable_classes>` but I don't like forcing everything to have typelists outside of the base class. That is useful in some cases where we want different derived classes depending on the executable (e.g. it's fantastic for initial data and events), but things like boundary corrections and boundary conditions are quite universal. Being able to have a `call_with_dynamic_type` over those in an explicit instantiation in the `System` directory is very nice. Keeping the registration with Charm++ through the `factory_creation` struct seems good to me, though.

So if your suggestion is change the `factory_creation` struct in the metavariables to have `tmpl::pair<Base, typename Base::createable_classes>` then I'm in favor. If the suggestion is to move the `createable_classes` typelist into a separate file, then I'm opposed, certainly in the general case, but currently also in the specific case.

wthrowe:
(Your time estimates are presumably off.  This code was only added a bit over 3 years ago.)

You can't explicitly instantiate a factory, and that's the only thing that is directly tied to the `factory_creation` struct.  The argument to `call_with_dynamic_type` doesn't have to come from the `factory_creation` struct if you have a list by some other means, so in those cases having the list available from the base class may make sense, yes.

In any case, that's irrelevant for this PR, since there's no reason to use `call_with_dynamic_type` with `InitialMagneticField`, and nothing is doing anything fancy with option parsing.  What's the advantage of having the list in the base class?  You still need to have the list of headers in a `Factory.hpp`, so it means the same information has to be maintained in two places, and there's very little you can do with the list without including the class's `Factory.hpp`.

nilsdeppe:
I talked about this with @wthrowe on Friday and the conclusion was that in this case moving the list of magnetic fields from the base class to `Factory.hpp` is fine. The plan is to make that the convention whenever possible. I've had trouble finding where the lists are located, particularly with initial data, so trying to standardize that (not in this PR!) would be useful.

ncorsobh:
Force-pushed since the previous version was just a one-liner that got removed anyway. Let me know if this is the idea you're asking for or if I should do something else.

src/PointwiseFunctions/AnalyticData/GrMhd/InitialMagneticFields/InitialMagneticField.hpp
```diff
@@ -68,8 +68,6 @@ class InitialMagneticField : public PUP::able {
   InitialMagneticField() = default;
 
  public:
-  using creatable_classes = tmpl::list<Poloidal, Toroidal>;
```
wthrowe, line 71: Remove the forward declarations from this file.

wthrowe (commented):
Yes, that's what I had in mind.  Tests need fixing, though.

tests/Unit/PointwiseFunctions/AnalyticData/GrMhd/Test_MagnetizedTovStar.cpp
```diff
@@ -181,14 +193,12 @@ void test_equality() {
...
 
 void test_magnetized_tov_star(const TovCoordinates coord_system) {
-  register_derived_classes_with_charm<
-      grmhd::AnalyticData::InitialMagneticFields::InitialMagneticField>();
+  register_factory_classes_with_charm<Metavariables>();
   register_classes_with_charm<grmhd::AnalyticData::MagnetizedTovStar>();
   register_classes_with_charm<EquationsOfState::PolytropicFluid<true>>();
-  const std::unique_ptr<evolution::initial_data::InitialData> option_solution =
-      TestHelpers::test_option_tag_factory_creation<
-          evolution::initial_data::OptionTags::InitialData,
```
wthrowe, line 190: Don't change this so much.  You've lost the test of the `get_clone()` method and possibly the option tag (not sure if that one's important here).  Just switch `test_option_tag_factory_creation` to `test_option_tag` and pass the metavariables.

tests/Unit/PointwiseFunctions/AnalyticData/GrMhd/InitialMagneticFields/Test_Poloidal.cpp
```diff
@@ -44,10 +53,11 @@ struct PoloidalProxy : Poloidal {
 SPECTRE_TEST_CASE(
     "Unit.PointwiseFunctions.AnalyticData.GrMhd.InitialMagneticFields.Poloidal",
     "[Unit][PointwiseFunctions]") {
-  register_derived_classes_with_charm<InitialMagneticField>();
+  register_factory_classes_with_charm<Metavariables>();
```
wthrowe: [optional] I think the only change you need in this file is making this line `register_classes_with_charm<InitialMagneticFields::Poloidal>();`.  The version you've written isn't wrong, but it's a lot more verbose than necessary.

Same for `Test_Toroidal.cpp`.

wthrowe (changes requested):

wthrowe:
OK, looks good.  Squash.

wthrowe (approved):

PR #6011, opened 2024-05-20
Add CLI for plotting CCE output data

knelli2:
## Proposed changes

Also update the CCE tutorial to point to this CLI so we don't just have a large block of python code in our tutorial.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Tutorials/CCE.md
+++ b/docs/Tutorials/CCE.md
@@ -134,61 +134,22 @@ The `EthInertialRetardedTime` is a diagnostic dataset that represents the
 angular derivative of the inertial retarded time, which determines the
 coordinate transformation that is performed at scri+.
 
-The following python script will load data from a successful CCE run and
-construct a plot of all of the physical waveform data.
-```py
-import matplotlib.pyplot as plt
-import numpy as np
-import h5py as h5
+If you'd like to visualize the output of a CCE run, we offer a
+[CLI](py/cli.html) that will produce a plot of all of quantities except
+`EthInertialRetardedTime`. To see how to use this cli, run
 
-
-def spectre_real_mode_index(l, m):
-    return 2 * (l**2 + l + m) + 1
-
-
-def spectre_imag_mode_index(l, m):
-    return 2 * (l**2 + l + m) + 2
-
-
-def get_modes_from_block_output(filename, dataset, modes=[[2, 2], [3, 3]]):
-    with h5.File(filename, "r") as h5_file:
-        timeseries_data = (h5_file[dataset][()][:, [0] + list(
-            np.array([[
-                spectre_real_mode_index(x[0], x[1]),
-                spectre_imag_mode_index(x[0], x[1])
-            ] for x in modes]).flatten())])
-    return timeseries_data
-
-
-plot_quantities = ["Strain", "News", "Psi0", "Psi1", "Psi2", "Psi3", "Psi4"]
-mode_set = [[2, 2], [3, 3]]
-filename = "CharacteristicExtractReduction.h5"
-output_plot_filename = "CCE_plot.pdf"
-
-legend = []
-for (mode_l, mode_m) in mode_set:
-    legend = np.append(legend, [
-        r"Re $Y_{" + str(mode_l) + r"\," + str(mode_m) + r"}$",
-        r"Im $Y_{" + str(mode_l) + r"\," + str(mode_m) + r"}$"
-    ])
-
-plt.figure(figsize=(8, 1.9 * len(plot_quantities)))
-for i in range(len(plot_quantities)):
-    ax = plt.subplot(len(plot_quantities), 1, i + 1)
-    timeseries = np.transpose(
-        get_modes_from_block_output(
-            filename, f"/SpectreR0100.cce/{plot_quantities[i]}", mode_set
-        )
-    )
-    for j in range(len(mode_set)):
-        plt.plot(timeseries[0], timeseries[j + 1], linestyle='--', marker='')
-    ax.set_xlabel("Simulation time (M)")
-    ax.set_ylabel("Mode coefficient")
-    ax.set_title(plot_quantities[i])
-plt.tight_layout()
-plt.savefig(output_plot_filename, dpi=400)
-plt.clf()
 ```
+spectre plot cce -h
+```
+
+If you'd like to do something more complicated than just make a quick plot,
+you'll have to load in the output data yourself using `h5py` or our
+`spectre.IO.H5` bindings.
+
+\note The CLI can also plot the "old" version of CCE output, described above.
+Pass `--cce-group Cce` to the CLI. This option is only for backwards
+compatibility with the old CCE output and is not supported for the current
+version of output. This options is deprecated and will be removed in the future.
 
 ### Input data formats
 
--- a/src/Visualization/Python/Plot.py
+++ b/src/Visualization/Python/Plot.py
@@ -100,6 +100,7 @@ def show_or_save_plot_command():
                         )
                 else:
                     plt.savefig(output)
+                    plt.close()
             else:
                 if not os.environ.get("DISPLAY"):
                     logger.warning(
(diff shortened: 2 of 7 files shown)


src/Visualization/Python/PlotCce.py
```diff
@@ -0,0 +1,278 @@
...
+    nargs=1,
+)
+@click.option(
+    "--modes",
+    "-m",
+    multiple=True,
+    callback=_parse_modes,
+    required=True,
+    help=(
+        "Which mode to plot. Specified as 'l,m'. Will plot both real and"
```
nilsvu, line 49: Use `nargs=2`?

src/Visualization/Python/PlotCce.py
```diff
@@ -0,0 +1,278 @@
...
+    default=False,
+    help="Plot only imaginary modes. Mutually exclusive with '--real'.",
+)
+@click.option(
+    "--extraction-radius",
+    "-r",
+    type=int,
+    help=(
+        "Extraction radius of data to plot as an int. If there is only one Cce"
+        " subfile, that one will be used and this option will be ignored if"
```
nilsvu: Don't silently ignore it. Use an XOR in the check below where you raise a usage error, or assert that that the specified radius matches the single available subfile name.

src/Visualization/Python/PlotCce.py
```diff
@@ -0,0 +1,278 @@
...
+            cycle_idx = (
+                j % len(cycle) if (real or imag) else (j // 2) % len(cycle)
+            )
+            if i == 0:
+                ax.plot(
+                    data.index,
+                    data[f"{quantity}{mode}"],
+                    color=cycle[cycle_idx],
+                    linestyle=linestyle,
+                    label=mode,
```
nilsvu: `label=mode if i == 0 else None,`

Or just always set the label, since you only call `legend()` on the first plot anyway.

tests/Unit/Visualization/Python/Test_PlotCce.py
```diff
@@ -0,0 +1,355 @@
...
+
+
+class TestPlotCce(unittest.TestCase):
+    def setUp(self):
+        self.work_dir = os.path.join(
+            unit_test_build_path(), "Visualization/PlotCce"
+        )
+        self.reduction_file_name_one_subfile = "TestCceReductions1.h5"
+        self.reduction_file_name_two_subfiles = "TestCceReductions2.h5"
+        plot_quantities = [
```
nilsvu, line 24: (Optional) Is it possible to use the C++ routine that writes data in the simulation also to write data here? Would be nice to make sure the code keeps working.

nilsvu (changes requested):

src/Visualization/Python/PlotCce.py
knelli2: I personally don't like using `nargs=2` for click.options (for click.arguments they are good). It feels more natural to me to have comma separated value (maybe it's because there isn't a space so it feels like one option rather than two).

src/Visualization/Python/PlotCce.py
knelli2: This seems a bit redundant especially since this is only meant to give a quick plot. If I have to specify this when there is only one subfile, then I have to `h5ls` the H5 file (or run this with `-l`) first and then copy the extraction radius and (re)run the command.

src/Visualization/Python/PlotCce.py
nilsvu: We're currently not consistent with comma vs space, so either is fine with me

src/Visualization/Python/PlotCce.py
nilsvu: No I mean only if it _is_ specified, assert that the value is consistent with the single existing subfile. Also allow not specifying it if there's only one subfile.

tests/Unit/Visualization/Python/Test_PlotCce.py
knelli2: Not easily I don't think. This data is written from an iterable action that then calls either a threaded action or a local synchronous action to do the writing. The iterable action is pretty complex too

src/Visualization/Python/PlotCce.py
knelli2: Oh my bad, I misunderstood. Yes I agree this would be good

src/Visualization/Python/PlotCce.py
```diff
@@ -0,0 +1,285 @@
...
+    "--extraction-radius",
+    "-r",
+    type=int,
+    help=(
+        "Extraction radius of data to plot as an int. If there is only one Cce"
+        " subfile, that one will be used and this option does not need to be"
+        " specified. If there is only one Cce subfile and this option is"
+        " specified, the value will be checked against the extraction radius in"
+        " the subfile. If there are multiple subfiles, this option must be"
+        " specified otherwise an error will occur. The expected form of the Cce"
```
nilsvu: I think you can remove these two sentences starting with "If".

nilsvu (commented):
Squash 👍

src/Visualization/Python/Plot.py
```diff
@@ -100,6 +100,7 @@ def command(output, **kwargs):
                         )
                 else:
                     plt.savefig(output)
+                    plt.close()
```
nilsvu, line 103: For my information, did you encounter an issue with this?

src/Visualization/Python/PlotCce.py
```diff
@@ -0,0 +1,285 @@
...
+    meant to plot the raw data produced by the SpECTRE CCE module.
+    """
+
+    if real and imag:
+        raise click.UsageError(
+            "Only specify one of '--real'/'--imag'. If you want to plot both"
+            " real and imaginary modes, specify neither."
+        )
+
+    # Filer out modes we aren't plotting
```
nilsvu: Typo "filter"

src/Visualization/Python/Plot.py
knelli2: Yeah when I was running the unit test on my machine, even when I had the `-o` option specified, it would still open the plot interactively causing a timeout. This stopped that from happening.

nilsvu (dismissed):

docs/Tutorials/CCE.md
```diff
@@ -134,61 +134,22 @@ The `EthInertialRetardedTime` is a diagnostic dataset that represents the
...
-filename = "CharacteristicExtractReduction.h5"
-output_plot_filename = "CCE_plot.pdf"
-
-legend = []
-for (mode_l, mode_m) in mode_set:
-    legend = np.append(legend, [
-        r"Re $Y_{" + str(mode_l) + r"\," + str(mode_m) + r"}$",
-        r"Im $Y_{" + str(mode_l) + r"\," + str(mode_m) + r"}$"
-    ])
+If you'd like to do something more complex than just make a quick plot, you'll
```
nilsdeppe: maybe `complex` -> `complicated`? Since CCE deals with complex numbers, just to avoid any possible confusion? Up to you!

docs/Tutorials/CCE.md
```diff
@@ -134,61 +134,22 @@ The `EthInertialRetardedTime` is a diagnostic dataset that represents the
...
-    ax.set_ylabel("Mode coefficient")
-    ax.set_title(plot_quantities[i])
-plt.tight_layout()
-plt.savefig(output_plot_filename, dpi=400)
-plt.clf()
-```
+\note The CLI can also plot the "old" version of CCE output, described above.
+Pass `--cce-group Cce` to the CLI. This option is only for backwards
+compatibility with the old CCE output and is not supported for the current
+version of output.
```
nilsdeppe: Maybe add `This option is deprecated and will be removed in the future.`

nilsdeppe (dismissed):
LGTM! A couple small suggestions. Please squash if you decide to do them, otherwise feel free to merge directly.

knelli2:
@nilsdeppe Squashed in the changes

nilsdeppe (approved):

PR #6014, opened 2024-05-21
Add ErrorIfDataTooBig to BBH pipeline

knelli2:
## Proposed changes

These variables and thresholds were used in SpEC.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/Ringdown.yaml
+++ b/support/Pipelines/Bbh/Ringdown.yaml
@@ -147,6 +147,15 @@ EventsAndTriggers:
   - Trigger:
       Slabs:
         EvenlySpaced:
+          Interval: 10
+          Offset: 0
+    Events:
+      - ErrorIfDataTooBig:
+          Threshold: 100
+          VariablesToCheck: [SpacetimeMetric]
+  - Trigger:
+      Slabs:
+        EvenlySpaced:
           Interval: 100
           Offset: 0
     Events:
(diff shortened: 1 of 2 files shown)


wthrowe (approved):

support/Pipelines/Bbh/Ringdown.yaml
```diff
@@ -144,6 +144,15 @@ EventsAndTriggers:
...
             Components: Sum
+  - Trigger:
+      Slabs:
+        EvenlySpaced:
+          Interval: 10
+          Offset: 0
+    Events:
+      - ErrorIfDataTooBig:
+          Threshold: 100
+          VariablesToCheck: [SpacetimeMetric]
```
nilsdeppe, line 155: Question: why not monitor the constraints during the ringdown? No need to hold up the PR aboutt this, I'm just surprised

knelli2: I don't actually know, but if I were to venture a guess it'd be because the constraint violations at t=Ringdown may be pretty large initially and then settle down quickly, so that initial violation could have been over the threshold? But then why not just delay the trigger? I'm not really sure. We should ask @markscheel at the BBH meeting today.

nilsdeppe: Sounds good! I think SpEC actually monitors the constraints _outside_ the horizons, but that's obviously harder to do.

PR #6019, opened 2024-05-22
Split time-stepper-history cleaning from update

wthrowe:
Reduces the amount of data we need to store in checkpoints, mostly in LTS.

### Upgrade instructions

Evolution executables now need to include the `Actions::CleanHistory` action after each variable update.  See the modifications to the executables in this PR for suggested exact changes.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/Actions/CleanHistory.hpp
+++ b/src/Time/Actions/CleanHistory.hpp
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <optional>
+#include <type_traits>
 
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "Parallel/AlgorithmExecution.hpp"
@@ -13,6 +14,7 @@
 #include "Utilities/TypeTraits/IsA.hpp"
 
 /// \cond
+class LtsTimeStepper;
 class TimeStepper;
 namespace Parallel {
 template <typename Metavariables>
@@ -24,6 +26,10 @@ struct TimeStepper;
 template <typename Tag>
 struct HistoryEvolvedVariables;
 }  // namespace Tags
+namespace evolution::dg::Tags {
+template <size_t Dim, typename CouplingResult>
+struct MortarDataHistory;
+}  // namespace evolution::dg::Tags
 namespace imex::Tags {
 template <typename ImplicitSector>
 struct ImplicitHistory;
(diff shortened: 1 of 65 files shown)


src/Time/Actions/CleanHistory.hpp
```diff
@@ -0,0 +1,74 @@
...
+/// - DataBox:
+///   - Tags::TimeStepper<TimeStepper>
+///
+/// DataBox changes:
+/// - Adds: nothing
+/// - Removes: nothing
+/// - Modifies:
+///   - Tags::HistoryEvolvedVariables<variables_tag>
+template <typename System>
+struct CleanHistory {
```
nilsdeppe, line 63: Just a note while I'm thinking about it: we should revisit all of our actions and see how much we can convert into Mutators, maybe even ones that can take a `db::Access` to move a lot of code into cpp files. This one looks like a possible candidate, which is what made me think of it. I don't think we should change anything in this PR, just noting it so I/we don't forget :)

src/Time/Actions/CleanHistory.hpp
```diff
@@ -68,6 +77,22 @@ struct CleanHistory {
         },
         make_not_null(&box));
 
+    // This action can't depend on Evolution, but most of the time
```
nilsdeppe, line 97: Sorry, what below depends on evolution that the above doesn't?

nilsdeppe (commented):
LGTM, one clarification question is all I have

src/Time/Actions/CleanHistory.hpp
wthrowe: For forward-declared stuff, there's `imex::protocols::ImexSystem` and `imex::Tags::ImplicitHistory` (from Imex, not directly Evolution, I guess), and `evolution::dg::Tags::MortarDataHistory` (actually in the helper above).  For completely not-named stuff, `const auto history` on line 122 has a type including `evolution::dg::MortarData`.

(Really, a lot (all?) of the stuff in the Time directory should probably be merged into Evolution, but that sounds painful.)

src/Time/Actions/CleanHistory.hpp
wthrowe: We might need to template on dimension to get the mortar data tag (currently it searches `DbTags`), but otherwise I don't see any issues with making this a mutator.

src/Time/Actions/CleanHistory.hpp
nilsdeppe: Okay, thanks!

src/Time/Actions/CleanHistory.hpp
nilsdeppe: Sounds good! A future problem, but just wanted to note it :)

nilsdeppe (approved):

nilsdeppe:
Hmm, do you understand the clang-tidy error? Tthis looks like it's telling us to change the STL?

wthrowe:
I think it's just complaining about charm stuff, but the charm class happens to be stored in a `std::optional` so the traces go through there.

I haven't examined the charm code in detail, but it looks like it's doing reference counting, which can cause false positives for this sort of automated analysis, so it's probably fine.

nilsdeppe:
Sounds good! Merging now :)

PR #6021, opened 2024-05-22
Add increasing_substep_time_steppers list

wthrowe:
Necessary property for worldtube particle tracking.

Not a huge fan of the name, so if anyone has a better idea feel free to suggest.  Unfortunately, I've already used "monotonic" for a different property.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/DevGuide/OptionParsing.md
+++ b/docs/DevGuide/OptionParsing.md
@@ -62,8 +62,8 @@ input file as usual, but a warning will be issued if the specified
 value does not match the suggestion.
 
 Examples:
-\snippet Test_Options.cpp options_example_scalar_struct
-\snippet Test_Options.cpp options_example_vector_struct
+\snippet Options/Test_Options.cpp options_example_scalar_struct
+\snippet Options/Test_Options.cpp options_example_vector_struct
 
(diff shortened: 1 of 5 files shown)


knelli2:
What is the other property that "monotonic" is used for? Is it really that bad if you use it again?

wthrowe:
The `AdamsMoultonPcMonotonic` class and `TimeStepper::monotonic()` function refer to the "not deadlocking the control system" property.

knelli2:
Hmm yeah probably shouldn't overload the meaning. However, it seems like using "monotonic" in this PR makes more sense and the other names should be changed to something else

wthrowe:
Do you have a suggestion for the other one, then?

The two properties are, explicitly:
1. Time is non-decreasing for all operations (but substep times can repeat)
2. Substep time is strictly increasing (but things like dense-output time can decrease freely)

knelli2:
Oh wow ok... Yeah I don't have a good idea how to differentiate these. They are so similar in many ways but just slightly different. I guess `increasing_substep_time_steppers` is ok for now.

knelli2 (approved):

PR #6024, opened 2024-05-24
FindHorizon.py: also compute horizon quantities

nilsvu:
## Proposed changes

This allows to get the horizon area, mass, spin, etc from BBH initial data just by calling a Python function.

### Upgrade instructions

The `FindHorizons3D` executable was removed. You can find horizons in 3D volume data in Python like this:

```py
from spectre.Pipelines.Bbh.FindHorizon import find_horizon
from spectre.SphericalHarmonics import Strahlkorper

horizon, quantities = find_horizon(
    h5_files, subfile_name, obs_id, obs_time,
    initial_guess=Strahlkorper(l_max, m_max, radius, center)
)

mass = quantities["ChristodoulouMass"]
spin = quantities["DimensionlessSpinMagnitude"]
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/PointwiseFunctions/GeneralRelativity/Surfaces/Python/Bindings.cpp
```diff
@@ -0,0 +1,64 @@
...
+      [](ylm::Strahlkorper<Frame> horizon,
+         tnsr::ii<DataVector, Dim, Frame> spatial_metric,
+         tnsr::II<DataVector, Dim, Frame> inv_spatial_metric,
+         tnsr::ii<DataVector, Dim, Frame> extrinsic_curvature,
+         tnsr::Ijj<DataVector, Dim, Frame> spatial_christoffel_second_kind,
+         tnsr::ii<DataVector, Dim, Frame> spatial_ricci) -> py::dict {
+        const auto box = db::create<
+            tmpl::append<tmpl::list<ylm::Tags::Strahlkorper<Frame>>,
+                         ::ah::vars_to_interpolate_to_target<Dim, Frame>>,
+            ::ah::compute_items_on_target<Dim, Frame>>(
```
knelli2: We have the dimension**FUL** `AKV` spin tag coded up in `src/PointwiseFunctions/GeneralRelativity/Surfaces/Tags.hpp`, but not the dimension**LESS** spin tag. Want to code it up here and just add it to `compute_items_on_target` that way it'll be available during the evolution as well?

src/PointwiseFunctions/GeneralRelativity/Surfaces/Python/Bindings.cpp
```diff
@@ -0,0 +1,64 @@
...
+namespace gr::surfaces::py_bindings {
+
+template <typename Frame>
+void bind_horizon_quantities_impl(py::module& m) {
+  static constexpr size_t Dim = 3;
+  m.def(
+      "horizon_quantities",
+      [](ylm::Strahlkorper<Frame> horizon,
+         tnsr::ii<DataVector, Dim, Frame> spatial_metric,
+         tnsr::II<DataVector, Dim, Frame> inv_spatial_metric,
```
knelli2, line 28: What about only requiring the `inv_spatial_metric` and just compute the `determinant_and_inverse` here? That way less data needs to be output

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Python/Bindings.cpp
```diff
@@ -0,0 +1,73 @@
...
+                if (requested_quantity == db::tag_name<tag>()) {
+                  result[requested_quantity.c_str()] =
+                      cache.get_var(computer, tag{});
+                  found = true;
+                }
+              });
+              if (not found) {
+                throw std::invalid_argument("Requested quantity '" +
+                                            requested_quantity +
+                                            "' is not available.");
```
knelli2: Print the available quantities.

knelli2 (changes requested):
Add in the upgrade instructions that the `FindHorizon` exec was removed and that the preferred method for finding horizons in data on disk is the CLI now

src/PointwiseFunctions/GeneralRelativity/Surfaces/Python/Bindings.cpp
nilsvu: The spatial metric is in the output anyway because we load it as initial data. I could drop the `inv_spatial_metric` from the volume data and compute it here. But the horizon finder needs the `inv_spatial_metric`, so it would have to compute the inverse in every iteration of the horizon finder. Not sure what's best, so I think just keep as is for now?

knelli2: I'm just worried we are making assumptions about what will be available when we don't necessarily know. Yes for the ID solve we'll have the spatial metric, but what about for a BNS run where somebody wants to see if a horizon formed? Or a supernova bounce? What about taking both as optionals? Then if both are available, you can just use both, but if one or the other are specified then you'll have to compute the inverse (you'd be doing it in volume data anyways)

nilsvu: Can we worry about these applications later? You can always use `spectre transform-vol` to compute the inverse and other quantities. I could do the optionals but that adds quite a lot of logic that I'm not sure is needed and I think is premature optimization.

knelli2: Yeah we can worry about them later. I do think people will want to have the option though.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Python/Bindings.cpp
knelli2: Ping on this.

support/Pipelines/Bbh/FindHorizon.py
```diff
@@ -155,6 +162,38 @@ def find_horizon(
...
+        strahlkorper,
+        spatial_metric=spatial_metric,
+        inv_spatial_metric=inv_spatial_metric,
+        extrinsic_curvature=extrinsic_curvature,
+        spatial_christoffel_second_kind=spatial_christoffel_second_kind,
+        spatial_ricci=spatial_ricci,
+    )
     # Write the horizon to a file and return it
     if output:
         assert output_coeffs_subfile or output_coords_subfile, (
```
knelli2, line 199: Can you add an `.h5` extension to the output if there isn't one already? Ran into this today.

knelli2 (changes requested):

src/PointwiseFunctions/GeneralRelativity/Surfaces/Python/Bindings.cpp
nilsvu: Yes, we want to find horizons in evolution output data. Note that we also have to output ExtrinsicCurvature, spatial Christoffels, and the spatial Ricci tensor, which we probably don't do atm. So we either have to output those (lots of data), or postprocess with `transform-vol` (possible, maybe best solution and works already, see #6071), or build the postprocessing into the horizon finder routines (possible, except for numerical derivatives), or have more flexible horizon finder triggers in the evolution  (hopefully enabled by interpolation framework changes, but doesn't help if you already have the simulation output).

src/ParallelAlgorithms/ApparentHorizonFinder/HorizonAliases.hpp
```diff
@@ -36,21 +36,22 @@ using vars_to_interpolate_to_target =
...
-               ylm::Tags::MinRicciScalarCompute,
-               gr::surfaces::Tags::ChristodoulouMassCompute<Frame>,
-               gr::surfaces::Tags::DimensionlessSpinMagnitudeCompute<Frame>>;
+using tags_for_observing = tmpl::list<
+    gr::surfaces::Tags::AreaCompute<Frame>,
+    gr::surfaces::Tags::IrreducibleMassCompute<Frame>,
+    ylm::Tags::MaxRicciScalarCompute, ylm::Tags::MinRicciScalarCompute,
+    gr::surfaces::Tags::ChristodoulouMassCompute<Frame>,
+    gr::surfaces::Tags::DimensionlessSpinMagnitudeCompute<Frame>,
+    gr::surfaces::Tags::DimensionlessSpinVectorCompute<Frame, Frame>>;
```
knelli2: Damn. This was my bad. I forgot that this tag list is used in the `ObserveTimeSeriesOnSurface` callback, and that requires the types of all tags it observes to be `double`s. I roughly know how to generalize this, but haven't had a chance to implement it. So you can't add the compute tag here, but you can just append it to the list in the pybinding cpp.

knelli2 (changes requested):

knelli2 (approved):
Fixups look good. Squash

PR #6031, opened 2024-05-25
Support SpEC initial data in BBH pipeline

### LINKED ISSUE #6020: Allow BBH inspiral pipeline to start from SpEC ID
We will want to be able to compare inspirals run with SpECTRE ID and SpEC ID so the inspiral pipeline should be able to accept a SpEC ID directory and use that.

nilsvu:
## Proposed changes

Allows to start a BBH evolution from SpEC initial data like this:

```
spectre bbh start-inspiral ID_Params.perl -O Ev
```

Closes #6020.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/Inspiral.py
+++ b/support/Pipelines/Bbh/Inspiral.py
@@ -2,6 +2,7 @@
 # See LICENSE.txt for details.
 
 import logging
+import subprocess
 from pathlib import Path
 from typing import Optional, Union
 
@@ -22,7 +23,7 @@ def inspiral_parameters(
     refinement_level: int,
     polynomial_order: int,
 ) -> dict:
-    """Determine inspiral parameters from initial data.
+    """Determine inspiral parameters from SpECTRE initial data.
 
     These parameters fill the 'INSPIRAL_INPUT_FILE_TEMPLATE'.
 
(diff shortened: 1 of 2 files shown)


support/Pipelines/Bbh/Inspiral.py
```diff
@@ -55,6 +56,77 @@ def inspiral_parameters(
     }
 
 
+def _load_spec_id_params(id_params_file: Path) -> dict:
+    """Load SpEC initial data parameters from 'ID_Params.perl'."""
+    # Here we have to deal with SpEC storing the initial data parameters in a
+    # perl script (yay!). We evaluate the perl script, emit the variables as
+    # YAML, and load that into Python as a dictionary.
```
knelli2: Since we are technically added a dependency on perl now, should we document that somewhere? I'm not sure if it's safe to assume that all systems will have perl in the default PATH (though they should...)

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -89,16 +161,26 @@ def start_inspiral(
     )
```
knelli2, line 148: Can't reach with github comments, but in this docstring, add that the `ID_INPUT_FILE_PATH` can either be to a SpECTRE ID input file or a SpEC `ID_Params.perl` file

knelli2 (changes requested):
Have you tested this out yet?

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -55,6 +56,77 @@ def inspiral_parameters(
...
+        "$ID_MA",
+        "$ID_MB",
+        "\@ID_chiA",
+        "\@ID_chiB",
+        "$ID_chiAMagnitude",
+        "$ID_chiBMagnitude",
+    ]
+    perl_script = [
+        f'do "{id_params_file.resolve()}";',
+        "print YAML::Dump({ "
```
nilsdeppe: Is this part of the core perl install? I don't know how easy it is to install perl modules locally. It may be easier to parse the file ourselves rather than rely on a perl module. We then also don't have to worry about calling into perl from a python script :)

knelli2: I don't think so? I just tried on my local machine a `test.pl` that had only `use YAML` and it complained about a missing module. [This](https://metacpan.org/pod/YAML) also does warn against using the `YAML` module if we were going to use a module.

nilsvu: Ok I wanted to avoid parsing perl syntax with regex because I hate that too :D I'll see what can be done. (this code isn't the most critical in the big picture either, I'd like to delete it again in sooner or later)

knelli2: Could you just read the file in as a string, do some replacements like you're already doing and use a yaml parser in python?

nilsvu: Convert perl syntax to yaml or json? Yea good idea

knelli2: Basically, yeah

nilsdeppe: Just rewrite all of SpEC to use yaml. Easy ;) :D

nilsvu: Ok done (@knelli2's suggestion, not @nilsdeppe's 😄)

nilsvu: > Just rewrite all of SpEC to use yaml. Easy ;) :D

Actually that's kind of what we're doing ;)

knelli2 (approved):

nilsdeppe (approved):

PR #6035, opened 2024-05-29
Improve time step too small error.

nilsdeppe:
## Proposed changes

@deborahferguson got the error message with a CCE run and admittedly it wasn't super clear. I think this is somewhat of an improvement, but could still probably use some work. The tricky bit is that there are a lot of reasons why the code could go unstable and result in a tiny time step size :(

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/Actions/ChangeStepSize.hpp
+++ b/src/Time/Actions/ChangeStepSize.hpp
@@ -97,12 +97,21 @@ bool change_step_size(const gsl::not_null<db::DataBox<DbTags>*> box) {
     desired_step = -desired_step;
   }
 
-  if (abs(desired_step / current_step.slab().duration().value()) < 1.0e-9) {
+  constexpr double smallest_relative_step_size = 1.0e-9;
+  if (abs(desired_step / current_step.slab().duration().value()) <
+      smallest_relative_step_size) {
     ERROR(
-        "Chosen step is extremely small; this can indicate a flaw in the a "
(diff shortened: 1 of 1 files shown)


src/Time/Actions/ChangeStepSize.hpp
```diff
@@ -97,12 +97,21 @@ bool change_step_size(const gsl::not_null<db::DataBox<DbTags>*> box) {
...
         "error-based stepper is naively attempting to resolve. It is unlikely "
-        "that the simulation can proceed");
+        "that the simulation can proceed in a stable and accurate manner. "
+        "A possible issue is an aliasing-driven instability that could be "
+        "cured by more aggressive filtering if you are using DG. The desired "
+        "time step size is "
+        << desired_step << " which, when divided by the slab size ("
+        << current_step.slab().duration().value() << ") is below the tolerance "
+        << smallest_relative_step_size << ". The current time is "
+        << (current_step.slab().start() + current_step).value() << ".");
```
wthrowe: Current time is `time_step_id.substep_time()` (a double).

wthrowe (changes requested):

src/Time/Actions/ChangeStepSize.hpp
nilsdeppe: Done! Thanks for the review :)

wthrowe (approved):

PR #6036, opened 2024-05-29
Fix linker errors when building with shared libs on macOS Apple Silicon

geoffrey4444:
## Proposed changes

I have succeeded in building spectre, with all tests passing, on macOS using shared libraries. (Previously, we required macOS Apple Silicon users to build with static libraries.) But to make this work, I had to fix some linker errors that appeared:

* I added an explicit instantiation for `detail::initialize_coords_and_jacobians()` in `src/Elliptic/DiscontinuousGalerkin/Initialization.cpp`. Otherwise, I get an undefined symbol linker error when trying to build this file.
* I added includes for PartialDerivatives.hpp and PartialDerivatives.tpp to a bunch of our executables, since when building them I was getting linker errors about partial_derivatives being undefined symbols.

A followup PR will update the macOS apple silicon docs.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/DiscontinuousGalerkin/Initialization.cpp
+++ b/src/Elliptic/DiscontinuousGalerkin/Initialization.cpp
@@ -245,6 +245,22 @@ void mortar_jacobian(
       const std::optional<tnsr::II<DataVector, DIM(data)>>&                    \
           inv_metric_on_mortar,                                                \
       const ElementMap<DIM(data), Frame::Inertial>& element_map,               \
+      const domain::FunctionsOfTimeMap& functions_of_time);                    \
+  template void detail::initialize_coords_and_jacobians(                       \
+      gsl::not_null<tnsr::I<DataVector, DIM(data), Frame::ElementLogical>*>    \
+          logical_coords,                                                      \
+      gsl::not_null<tnsr::I<DataVector, DIM(data), Frame::Inertial>*>          \
+          inertial_coords,                                                     \
+      gsl::not_null<InverseJacobian<DataVector, DIM(data),                     \
+                                    Frame::ElementLogical, Frame::Inertial>*>  \
+          inv_jacobian,                                                        \
+      gsl::not_null<Scalar<DataVector>*> det_inv_jacobian,                     \
+      gsl::not_null<Scalar<DataVector>*> det_jacobian,                         \
+      gsl::not_null<InverseJacobian<DataVector, DIM(data),                     \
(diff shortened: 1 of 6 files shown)


nilsdeppe (changes requested):
We shouldn't be including the `tpp` files in any `hpp`s. I would expect there to be an explicit instantiation that handles this. Why is this not an issue on Linux systems? Or phrased differently, why is there no explicit instantiation for the specific case that you need already generated?

geoffrey4444:
> We shouldn't be including the `tpp` files in any `hpp`s. I would expect there to be an explicit instantiation that handles this. Why is this not an issue on Linux systems? Or phrased differently, why is there no explicit instantiation for the specific case that you need already generated?

So I understand, could you please explain the problem with including top files in hpp files?

I don't know why this only seems to break when I try building with shared libraries on Mac and not on Linux. This was just the first thing I learned how to do that let the shared-libraries-on-Apple-Silicon build succeed.

I'd be happy to change this to adding explicit instantiations, but I wasn't sure where to add them or how many different partial derivatives instantiations I might have to add. Do you have any suggestions?

nilsdeppe:
Ah, sorry! So the idea behind `tpp`s is that you can include them in `cpp` files to do explicit instantiations. Then by only including `hpp` files in other places the compiler doesn't have to do implicit instantiations and this should reduce build time. 

So I think it should be only 1 per system. Let's chat on the technical call Thursday if you can make it to see if we can find a way that would make it 1) easy to instantiate and 2) solve the issue and allow you to build shared libs :)

geoffrey4444:
Sounds good, Nils! I should be able to make the technical call, and a discussion to find a path forward for this PR would be great.

By the way, I didn't see this coming because many of our hpp's already include tpp's, including the following:
* TimeDependence.hpp
* DgOperator.hpp
* ComputeTimeDerivative.hpp
* EvolveGhBinaryBlackHole.hpp
* EvolveGhSingleBlackHole.hpp
* EvolveGhValenciaDivCleanWithHorizon.hpp
* EvolveM1Grey.hpp
* EvolveScalarTensorSingleBlackHole.hpp
* ConservativeSystem.hpp

nilsdeppe (approved):
Awesome, thank you!!

geoffrey4444:
You're welcome, @nilsdeppe ! @nilsvu do you still want to take a look at this PR? If so it's ready whenever you get a chance

nilsvu (approved):

PR #6044, opened 2024-05-30
 Restructure ApplyBoundaryCorrections and support AtomicBoundaryInbox 

nilsdeppe:
## Proposed changes

I'll need to squash the two commits later, but I figured this is a bit easier to review.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DiscontinuousGalerkin/Actions/ApplyBoundaryCorrections.hpp
+++ b/src/Evolution/DiscontinuousGalerkin/Actions/ApplyBoundaryCorrections.hpp
@@ -3,9 +3,11 @@
 
 #pragma once
 
+#include <atomic>
 #include <cstddef>
 #include <limits>
 #include <map>
+#include <mutex>
 #include <optional>
 #include <tuple>
 #include <type_traits>
@@ -65,9 +67,7 @@ struct TimeStepId;
 template <typename StepperInterface>
 struct TimeStepper;
 }  // namespace Tags
-/// \endcond
 
-/// \cond
 namespace evolution::dg::subcell {
 // We use a forward declaration instead of including a header file to avoid
 // coupling to the DG-subcell libraries for executables that don't use subcell.
@@ -75,13 +75,13 @@ template <size_t VolumeDim, typename DgComputeSubcellNeighborPackagedData>
 void neighbor_reconstructed_face_solution(
     gsl::not_null<db::Access*> box,
     gsl::not_null<std::pair<
-        const TimeStepId,
+        TimeStepId,
         DirectionalIdMap<VolumeDim, evolution::dg::BoundaryData<VolumeDim>>>*>
         received_temporal_id_and_data);
 template <size_t Dim>
 void neighbor_tci_decision(
     gsl::not_null<db::Access*> box,
-    const std::pair<const TimeStepId,
+    const std::pair<TimeStepId,
                     DirectionalIdMap<Dim, evolution::dg::BoundaryData<Dim>>>&
         received_temporal_id_and_data);
 }  // namespace evolution::dg::subcell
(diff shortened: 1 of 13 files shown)


src/Evolution/DiscontinuousGalerkin/Actions/ApplyBoundaryCorrections.hpp
```diff
@@ -98,6 +98,110 @@ struct TemporaryReference {
...
+          // communications is more important than the size of each
+          // communication, and so a single communication per time/sub
+          // step is preferred.
+          ASSERT(current_ghost_cell_data.has_value(),
+                 "Have not yet received ghost cells at time step "
+                     << time_step_id
+                     << " but the inbox entry already exists. This is a "
+                        "bug in "
+                        "the "
+                        "ordering of the actions.");
```
wthrowe: Formatting.  And a few more times.

src/Evolution/DiscontinuousGalerkin/Actions/ApplyBoundaryCorrections.hpp
```diff
@@ -231,46 +394,74 @@ bool receive_boundary_data_local_time_stepping(
...
-              return false;
-            }
-            const auto received_mortar_data =
-                time_entry->second.find(mortar_id);
-            if (received_mortar_data == time_entry->second.end()) {
-              return false;
-            }
-
+        ASSERT(inbox != nullptr, "The inbox pointer should not be null.");
+        for (auto received_data = inbox->begin();
```
wthrowe: The rest of this function looks like it's based off a version of the code from before #5864.  It needs to incorporate those changes.  I think the intended changes here are mostly changing `inbox` to a pointer, so that shouldn't be too difficult.

wthrowe (changes requested):

nilsdeppe:
@wthrowe this is ready for you to look at again

src/Evolution/DiscontinuousGalerkin/Actions/ApplyBoundaryCorrections.hpp
```diff
@@ -284,9 +474,6 @@ bool receive_boundary_data_local_time_stepping(
                 received_mortar_data->second.interface_mesh,
                 std::move(received_mortar_data->second.boundary_correction_data
                               .value()));
-            // We don't yet communicate the integration order, because
-            // we don't have any variable-order methods.  The
-            // fixed-order methods ignore the field.
```
wthrowe: Restore comment.

src/Evolution/DiscontinuousGalerkin/Actions/ApplyBoundaryCorrections.hpp
```diff
@@ -302,6 +489,23 @@ bool receive_boundary_data_local_time_stepping(
...
+                    typename evolution::dg::Tags::
+                        BoundaryCorrectionAndGhostCellsInbox<Dim>::type>) {
+    // We only decrease the counter if we are done with the current time
+    // and we only decrease it by the number of neighbors at the current
+    // time.
+    tuples::get<evolution::dg::Tags::BoundaryCorrectionAndGhostCellsInbox<Dim>>(
+        *inboxes)
+        .message_count.fetch_sub(
+            db::get<domain::Tags::Element<Dim>>(*box).number_of_neighbors(),
+            std::memory_order_acq_rel);
```
wthrowe, line 511: It looks like this subtraction happens whenever `ApplyBoundaryCorrections::is_ready` returns true in LTS mode, even if no new messages were actually processed.  This happens for every dense-output operation.  Based on the field name, it looks like it's also assuming you get one message from each neighbor, which doesn't sound right if they're stepping at different rates.

If this should be the number of DG boundary messages processed, it's probably easiest to just add a counter in the loop above.

wthrowe (changes requested):

src/Evolution/DiscontinuousGalerkin/Actions/ApplyBoundaryCorrections.hpp
nilsdeppe: Good point. This is code I never got fully working yet and is related to what we'd discussed a couple weeks ago.

Essentially, I think this should subtract the number of neighbors since the neighbors should only increment this when they send data for the next (complete) time step (I need to think about this more based on our previous convo). I don't think this should ever decrement on dense output, only on an actual time step. Is wrapping it in a `not DenseOutput` enough for now? It'll require quite a bit more testing in real use cases and decisions about the algorithm. I can also add an ERROR if you want

wthrowe: What do you do when you don't need any messages from a neighbor?

It sounds like this case still needs a lot of design work.  Making it ERROR for now sounds good.

nilsdeppe: Sounds good! Yes, the LTS stuff isn't working yet. I'm writing tests for the threaded callbacks and once that's done I'll be focusing on LTS support.

nilsdeppe:
I've rebased and squashed in the couple changes. Sorry for accidentally dropping that comment!

wthrowe (approved):

kidder (approved):

PR #6048, opened 2024-05-30
Clarify error in InitializeJ

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/Initialize/InitializeJ.hpp
+++ b/src/Evolution/Systems/Cce/Initialize/InitializeJ.hpp
@@ -201,9 +201,13 @@ double iteratively_adapt_angular_coordinates(
       ERROR(
           "Iterative solve for surface coordinates of initial data failed. The "
           "strain is too large to be fully eliminated by a well-behaved "
-          "alteration of the spherical mesh. For this data, please use an "
-          "alternative initial data generator such as "
-          "`InitializeJInverseCubic`.");
+          "alteration of the spherical mesh. This could be an indication that "
+          "there is an issue with the worldtube data. If you are confident "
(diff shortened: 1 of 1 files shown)


src/Evolution/Systems/Cce/Initialize/InitializeJ.hpp
```diff
@@ -201,9 +201,14 @@ double iteratively_adapt_angular_coordinates(
...
-          "alteration of the spherical mesh. For this data, please use an "
-          "alternative initial data generator such as "
-          "`InitializeJInverseCubic`.");
+          "alteration of the spherical mesh. This could be an indication that "
+          "there is an issue with the worldtube data, such as incorrect "
+          "computation of the spherical harmonic modes. If you are confident "
+          "the worldtube data is correct, then please use an alternative "
+          "initial data generator such as `InverseCubic`. If that fails, "
+          "please double check that your spherical harmonic modes are decaying "
+          "correctly with increasing (l,m).\nError: "
```
knelli2: So this error happens because (physically) the conformal factor is too large and therefore the transformation needed is no longer well-defined. I'm not really sure how to best incorporate that since it's a bit technical and I don't know what to suggest in terms of options or in terms of data to keep the conformal factor reasonable.

I don't really think the part about `incorrect computation of the spherical harmonic modes` will be that useful to most people since they will likely get worldtube data from sources that do their spherical harmonic computation correctly. The part about decaying modes sounds good to keep though.

nilsdeppe: I agree! I just rebased and squashed the removal of that part :) Thanks for the review!

knelli2 (approved):

PR #6052, opened 2024-06-02
Switch to using 4 cores in CI since GitHub supports that now

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/Tests.yaml
+++ b/.github/workflows/Tests.yaml
@@ -254,7 +254,7 @@ jobs:
           -D BUILD_PYTHON_BINDINGS=ON
           $GITHUB_WORKSPACE
 
-          make -j2 module_All
+          make -j4 module_All
       - name: Use new clang-tidy-diff
         # The clang-tidy-diff tool shipped with clang 14 doesn't report exit
         # codes, so we grab the version that does:
@@ -515,6 +515,9 @@ jobs:
         # container which arises from using the "--privileged" flag just below.
         OMPI_ALLOW_RUN_AS_ROOT: 1
         OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
+        # The number of cores to run on. This is given at:
+        # https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories
+        NUMBER_OF_CORES: 4
       # See https://lists.cs.illinois.edu/lists/arc/charm/2018-10/msg00011.html
       # for why we need this
       options: --privileged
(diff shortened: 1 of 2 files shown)


.github/workflows/Tests.yaml
```diff
@@ -714,7 +717,7 @@ ${{ matrix.build_type }}-pch-${{ matrix.use_pch || 'ON' }}"
         if: matrix.build_type == 'Release'
         working-directory: build
         run: |
-          make -j2 Benchmark
+          make -j${NUMBER_OF_CORES} Benchmark
```
nilsvu, line 721: The "Run non-unit tests" step below still uses 2 cores. Switch to 1 core and/or add a comment why it's not the max?

.github/workflows/Tests.yaml
```diff
@@ -822,7 +825,7 @@ ${{ matrix.build_type }}-pch-${{ matrix.use_pch || 'ON' }}"
           -D BUILD_DOCS=OFF
           ..
 
-          make EvolveBurgers -j2
+          make EvolveBurgers -j${NUMBER_OF_CORES}
 
           ctest -j2 -R InputFiles.Burgers.Step.yaml --output-on-failure
```
nilsvu: This also has `-j2`. Delete?

.github/workflows/Tests.yaml
```diff
@@ -515,6 +515,8 @@ jobs:
         # container which arises from using the "--privileged" flag just below.
         OMPI_ALLOW_RUN_AS_ROOT: 1
         OMPI_ALLOW_RUN_AS_ROOT_CONFIRM: 1
+        # The number of cores to run on
```
nilsvu: Mind adding the link to https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners/about-github-hosted-runners#standard-github-hosted-runners-for-public-repositories

nilsvu (changes requested):

.github/workflows/Tests.yaml
nilsdeppe: I've added a comment. That was intentional because I've had them time out sooner than the unit tests on my desktop. Having said that, we could try 4 cores and increase timeouts. That might be better?

Edit: thinking about it more, I will try 4 cores. I think increasing timeouts would be better than under utilizing the resources.

nilsdeppe:
Done! I squashed and rebased. I also added a commit that increases the timeout of the MeanValue test since that timed out on 2 different GCC builds.

knelli2:
Lots of timeouts. I noticed on my local machine that if I used all my cores to run tests, I got a lot of timeouts. Especially for the executables since they need 2 cores to run (1 charm + 1 worker). Maybe for executable tests we only run with 2 cores?

nilsdeppe:
Yea, I'll switch that back to 2 and add a comment :)

nilsdeppe:
Okay, done!

nilsvu (approved):

PR #6054, opened 2024-06-03
Add BnsInitialData System

isaaclegred:
## Proposed changes

This PR adds the system for generating BNS Irrotational Initial data.  To do this, an elliptic system is introduced to solve for the velocity potential using the framework of Baumgarte + Shapiro Ch. 15.  The background consists of a matter profile which is not yet implemented (as it is contained in the initial data).
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/BnsInitialData/BoundaryConditions/Factory.hpp
+++ b/src/Elliptic/Systems/BnsInitialData/BoundaryConditions/Factory.hpp
@@ -1,0 +1,16 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "Elliptic/BoundaryConditions/AnalyticSolution.hpp"
+#include "Elliptic/Systems/BnsInitialData/BoundaryConditions/StarSurface.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace BnsInitialData::BoundaryConditions {
+
+template <typename System>
+using standard_boundary_conditions =
+    tmpl::list<elliptic::BoundaryConditions::AnalyticSolution<System>,
+               StarSurface>;
+}  // namespace BnsInitialData::BoundaryConditions
--- a/src/Elliptic/Systems/BnsInitialData/BoundaryConditions/StarSurface.hpp
+++ b/src/Elliptic/Systems/BnsInitialData/BoundaryConditions/StarSurface.hpp
@@ -1,0 +1,105 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <pup.h>
+#include <string>
+#include <vector>
+
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "DataStructures/Tensor/TypeAliases.hpp"
+#include "Domain/Tags.hpp"
+#include "Domain/Tags/FaceNormal.hpp"
+#include "Elliptic/BoundaryConditions/BoundaryCondition.hpp"
+#include "Elliptic/BoundaryConditions/BoundaryConditionType.hpp"
+#include "Elliptic/Systems//BnsInitialData/Tags.hpp"
+#include "Options/Context.hpp"
+#include "Options/String.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Serialization/CharmPupable.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+/// \endcond
+
+namespace BnsInitialData ::BoundaryConditions {
+
+/*!
+ * Impose StarSurface boundary conditions:
+ * \f[
+ *   n_i F^i = \frac{C}{\alpha^2} B^i n_i.
+ * \f]
+ * The boundary condition results from requiring the conservation
+ * equations be regular at the surface of the neutron star.  See
+ * \cite BaumgarteShapiro 15.79.
+ */
+class StarSurface : public elliptic::BoundaryConditions::BoundaryCondition<3> {
+ private:
+  using Base = elliptic::BoundaryConditions::BoundaryCondition<3>;
+
+ public:
+  static constexpr Options::String help =
+      "StarSurface boundary conditions  n_i F^i = C/square(alpha) B^i n_i .";
+
+  using options = tmpl::list<>;
+
+  StarSurface() = default;
+  StarSurface(const StarSurface&) = default;
+  StarSurface& operator=(const StarSurface&) = default;
+  StarSurface(StarSurface&&) = default;
+  StarSurface& operator=(StarSurface&&) = default;
+  ~StarSurface() override = default;
+
+  /// \cond
+  explicit StarSurface(CkMigrateMessage* m) : Base(m) {}
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(StarSurface);
+  /// \endcond
+
+  std::unique_ptr<domain::BoundaryConditions::BoundaryCondition> get_clone()
+      const override {
+    return std::make_unique<StarSurface>(*this);
+  }
+
+  explicit StarSurface(const Options::Context& context);
+
+  std::vector<elliptic::BoundaryConditionType> boundary_condition_types()
+      const override {
+    return {elliptic::BoundaryConditionType::Neumann};
+  }
+
+  using argument_tags =
+      tmpl::list<gr::Tags::Lapse<DataVector>,
+                 BnsInitialData::Tags::RotationalShift<DataVector>,
+                 BnsInitialData::Tags::EulerEnthalpyConstant,
+                 domain::Tags::FaceNormal<3>>;
+  using volume_tags = tmpl::list<BnsInitialData::Tags::EulerEnthalpyConstant>;
+
+  static void apply(gsl::not_null<Scalar<DataVector>*> velocity_potential,
+                    gsl::not_null<Scalar<DataVector>*> n_dot_flux_for_potential,
+                    const tnsr::i<DataVector, 3>& deriv_velocity_potential,
+                    const Scalar<DataVector>& lapse,
+                    const tnsr::I<DataVector, 3>& rotational_shift,
+                    double euler_enthalpy_constant,
+                    const tnsr::i<DataVector, 3>& normal);
+
+  using argument_tags_linearized = tmpl::list<>;
+  using volume_tags_linearized = tmpl::list<>;
+
+  static void apply_linearized(
+      gsl::not_null<Scalar<DataVector>*> velocity_potential_correction,
+      gsl::not_null<Scalar<DataVector>*> n_dot_flux_for_potential_correction,
+      const tnsr::i<DataVector, 3>& deriv_velocity_potential);
+
+  void pup(PUP::er& p) override;
+};
+
+bool operator==(const StarSurface& lhs, const StarSurface& rhs);
+
+bool operator!=(const StarSurface& lhs, const StarSurface& rhs);
+
+}  // namespace BnsInitialData::BoundaryConditions
--- a/src/Elliptic/Systems/BnsInitialData/Equations.hpp
+++ b/src/Elliptic/Systems/BnsInitialData/Equations.hpp
@@ -1,0 +1,111 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Elliptic/Systems/BnsInitialData/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeWithValue.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+class DataVector;
+namespace PUP {
+class er;
+}  // namespace PUP
+namespace BnsInitialData {
+struct Fluxes;
+struct Sources;
+}  // namespace BnsInitialData
+/// \endcond
+
+namespace BnsInitialData {
+
+/*!
+ * \brief Compute the fluxes
+ * \f[ F^i = \gamma^{ij} n_j \Phi  - n_j \Phi
+ * \frac{B^iB^j}{\alpha^2}  \f] where \f$ n_j\f$ is the
+ * `face_normal`.
+ *
+ * The `face_normal_vector` is \f$ \gamma^{ij} n_j\f$.
+ */
+void fluxes_on_face(
+    gsl::not_null<tnsr::I<DataVector, 3>*> face_flux_for_potential,
+    const tnsr::i<DataVector, 3>& face_normal,
+    const tnsr::I<DataVector, 3>& face_normal_vector,
+    const tnsr::II<DataVector, 3>& rotational_shift_stress,
+    const Scalar<DataVector>& velocity_potential);
+
+/*!
+ * \brief Compute the generic fluxes \f$ F^i = D^i \Phi - B^jD_j\Phi /\alpha^2
+ * B^i \f$ for the Irrotational BNS equation for the velocity potential.
+ */
+void potential_fluxes(
+    gsl::not_null<tnsr::I<DataVector, 3>*> flux_for_potential,
+    const tnsr::II<DataVector, 3>& rotational_shift_stress,
+    const tnsr::II<DataVector, 3>& inverse_spatial_metric,
+    const tnsr::i<DataVector, 3>& velocity_potential_gradient);
+
+/*!
+ * \brief Add the sources \f$S=-\Gamma^i_{ij}F^j - \D_j \left(\alpha \rho /
+ * h\right)  F^j\f$ for the curved-space Irrotational BNS equation on a spatial
+ * metric \f$\gamma_{ij}\f$.
+ *
+ * These sources arise from the non-principal part of the Laplacian on a
+ * non-Euclidean background.
+ */
+void add_potential_sources(
+    gsl::not_null<Scalar<DataVector>*> source_for_potential,
+    const tnsr::i<DataVector, 3>&
+        log_deriv_lapse_times_density_over_specific_enthalpy,
+    const tnsr::i<DataVector, 3>& christoffel_contracted,
+    const tnsr::I<DataVector, 3>& flux_for_potential);
+
+/*!
+ * \brief Compute the fluxes \f$F^i\f$ for the curved-space Irrotatational BNS
+ * equations on a spatial metric \f$\gamma_{ij}\f$.
+ *
+ * \see IrrotationalBns::FirstOrderSystem
+ */
+struct Fluxes {
+  using argument_tags =
+      tmpl::list<gr::Tags::InverseSpatialMetric<DataVector, 3>,
+                 Tags::RotationalShiftStress<DataVector>>;
+  using volume_tags = tmpl::list<>;
+  using const_global_cache_tags = tmpl::list<>;
+  static void apply(gsl::not_null<tnsr::I<DataVector, 3>*> flux_for_potential,
+                    const tnsr::II<DataVector, 3>& inverse_spatial_metric,
+                    const tnsr::II<DataVector, 3>& rotational_shift_stress,
+                    const Scalar<DataVector>& velocity_potential,
+                    const tnsr::i<DataVector, 3>& velocity_potential_gradient);
+  static void apply(gsl::not_null<tnsr::I<DataVector, 3>*> flux_on_face,
+                    const tnsr::II<DataVector, 3>& inverse_spatial_metric,
+                    const tnsr::II<DataVector, 3>& rotational_shift_stress,
+                    const tnsr::i<DataVector, 3>& face_normal,
+                    const tnsr::I<DataVector, 3>& face_normal_vector,
+                    const Scalar<DataVector>& velocity_potential);
+};
+
+/*!
+ * \brief Add the sources \f$S_A\f$ for the curved-space Irrotatioanl BNS
+ * equation on a spatial metric \f$\gamma_{ij}\f$.
+ *
+ * \see IrrotationalBns::FirstOrderSystem
+ */
+struct Sources {
+  using argument_tags = tmpl::list<
+      Tags::DerivLogLapseTimesDensityOverSpecificEnthalpy<DataVector>,
+      gr::Tags::SpatialChristoffelSecondKindContracted<DataVector, 3>>;
+  static void apply(gsl::not_null<Scalar<DataVector>*> equation_for_potential,
+                    const tnsr::i<DataVector, 3>&
+                        log_deriv_lapse_times_density_over_specific_enthalpy,
+                    const tnsr::i<DataVector, 3>& christoffel_contracted,
+                    const Scalar<DataVector>& velocity_potential,
+                    const tnsr::I<DataVector, 3>& flux_for_potential);
+};
+
+}  // namespace BnsInitialData
--- a/tests/Unit/Elliptic/Systems/BnsInitialData/BoundaryConditions/Test_StarSurface.cpp
+++ b/tests/Unit/Elliptic/Systems/BnsInitialData/BoundaryConditions/Test_StarSurface.cpp
@@ -1,0 +1,168 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <cstddef>
+#include <string>
+#include <utility>
+#include <vector>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Elliptic/BoundaryConditions/ApplyBoundaryCondition.hpp"
+#include "Elliptic/BoundaryConditions/BoundaryCondition.hpp"
+#include "Elliptic/BoundaryConditions/BoundaryConditionType.hpp"
+#include "Elliptic/Systems/BnsInitialData/BoundaryConditions/StarSurface.hpp"
+#include "Framework/CheckWithRandomValues.hpp"
+#include "Framework/SetupLocalPythonEnvironment.hpp"
+#include "Framework/TestCreation.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "NumericalAlgorithms/DiscontinuousGalerkin/NormalDotFlux.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace BnsInitialData::BoundaryConditions {
+
+namespace {
+// the velocity potential does not matter
+template <bool Linearized>
+void apply_star_surface_boundary_condition_specific(
+    const gsl::not_null<Scalar<DataVector>*> n_dot_auxilliary_velocity,
+    Scalar<DataVector> velocity_potential = Scalar<DataVector>{
+        DataVector{1.0, 2.0, 3.0}}) {
+  const StarSurface boundary_condition{{}};
+  const tnsr::i<DataVector, 3> velocity_potential_gradient{
+      {DataVector{1.0, 1.0, 1.0}, DataVector{0.0, 0.0, 0.0},
+       DataVector{0.0, 0.0, 0.0}}};
+  const Scalar<DataVector> lapse{DataVector{1.0, 1.5, 1.2}};
+  const tnsr::I<DataVector, 3> rotational_shift{{DataVector{1.0, 1.0, 1.0},
+                                                 DataVector{0.0, 0.0, 0.0},
+                                                 DataVector{0.0, 0.0, 0.0}}};
+  const auto x = tnsr::I<DataVector, 3>{{DataVector{1.0, 1.0, 1.0},
+                                         DataVector{0.0, 0.0, 0.0},
+                                         DataVector{0.0, 0.0, 0.0}}};
+
+  const auto direction = Direction<3>::lower_xi();
+  const tnsr::i<DataVector, 3> normal{{DataVector{1.0, 0.0, 0.0},
+                                       DataVector{0.0, 1.0, 0.0},
+                                       DataVector{0.0, 0.0, 1.0}}};
+  const double euler_enthalpy_constant = 1.0;
+  const auto box = db::create<db::AddSimpleTags<
+      domain::Tags::Faces<3, domain::Tags::Coordinates<3, Frame::Inertial>>,
+      domain::Tags::Faces<3, domain::Tags::FaceNormal<3>>,
+      domain::Tags::Faces<3, gr::Tags::Lapse<DataVector>>,
+      domain::Tags::Faces<3, BnsInitialData::Tags::RotationalShift<DataVector>>,
+      BnsInitialData::Tags::EulerEnthalpyConstant>>(
+      DirectionMap<3, tnsr::I<DataVector, 3>>{{direction, x}},
+      DirectionMap<3, tnsr::i<DataVector, 3>>{{direction, normal}},
+      DirectionMap<3, Scalar<DataVector>>{{direction, lapse}},
+      DirectionMap<3, tnsr::I<DataVector, 3>>{{direction, rotational_shift}},
+      euler_enthalpy_constant);
+  elliptic::apply_boundary_condition<Linearized, void, tmpl::list<StarSurface>>(
+      boundary_condition, box, Direction<3>::lower_xi(),
+      make_not_null(&velocity_potential), n_dot_auxilliary_velocity,
+      velocity_potential_gradient);
+}
+
+void apply_star_surface_boundary_condition_generic(
+    const gsl::not_null<Scalar<DataVector>*> n_dot_auxilliary_velocity,
+    Scalar<DataVector> velocity_potential,
+    const tnsr::i<DataVector, 3>& velocity_potential_gradient,
+    const Scalar<DataVector>& lapse,
+    const tnsr::I<DataVector, 3>& rotational_shift,
+    const double euler_enthalpy_constant,
+    const tnsr::i<DataVector, 3>& normal) {
+  const StarSurface boundary_condition{{}};
+
+  const auto x = tnsr::I<DataVector, 3>{{DataVector{1.0, 1.0, 1.0},
+                                         DataVector{0.0, 0.0, 0.0},
+                                         DataVector{0.0, 0.0, 0.0}}};
+
+  const auto direction = Direction<3>::lower_xi();
+
+  const auto box = db::create<db::AddSimpleTags<
+      domain::Tags::Faces<3, domain::Tags::Coordinates<3, Frame::Inertial>>,
+      domain::Tags::Faces<3, domain::Tags::FaceNormal<3>>,
+      domain::Tags::Faces<3, gr::Tags::Lapse<DataVector>>,
+      domain::Tags::Faces<3, BnsInitialData::Tags::RotationalShift<DataVector>>,
+      BnsInitialData::Tags::EulerEnthalpyConstant>>(
+      DirectionMap<3, tnsr::I<DataVector, 3>>{{direction, x}},
+      DirectionMap<3, tnsr::i<DataVector, 3>>{{direction, normal}},
+      DirectionMap<3, Scalar<DataVector>>{{direction, lapse}},
+      DirectionMap<3, tnsr::I<DataVector, 3>>{{direction, rotational_shift}},
+      euler_enthalpy_constant);
+  elliptic::apply_boundary_condition<false, void, tmpl::list<StarSurface>>(
+      boundary_condition, box, Direction<3>::lower_xi(),
+      make_not_null(&velocity_potential), n_dot_auxilliary_velocity,
+      velocity_potential_gradient);
+}
+}  // namespace
+
+SPECTRE_TEST_CASE("Unit.BnsInitialData.BoundaryConditions.StarSurface",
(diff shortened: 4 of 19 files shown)


isaaclegred:
@nilsvu  whenever you get a chance!

src/Elliptic/Systems/IrrotationalBns/BoundaryConditions/Neumann.cpp
```diff
@@ -0,0 +1,42 @@
...
+#include "Utilities/Gsl.hpp"
+namespace IrrotationalBns::BoundaryConditions {
+
+Neumann::Neumann(const Options::Context& /*context*/){}
+
+void Neumann::apply(
+    gsl::not_null<Scalar<DataVector>*> velocity_potential,
+    gsl::not_null<Scalar<DataVector>*> n_dot_auxiliary_velocity) {
+  set_number_of_grid_points(n_dot_auxiliary_velocity, *velocity_potential);
+  get(*n_dot_auxiliary_velocity) = 0.0;
```
nilsvu: I believe this is missing some terms, based on Baumgarte/Shapiro Eq. (15.79) [please reference this in the docs]

src/Elliptic/Systems/IrrotationalBns/BoundaryConditions/Neumann.hpp
```diff
@@ -0,0 +1,85 @@
...
+/// \cond
+class DataVector;
+/// \endcond
+
+namespace IrrotationalBns ::BoundaryConditions {
+
+/// Impose Neumann boundary conditions \f$a u + b n_i \nabla^i u = c\f$. The
+/// boundary condition is imposed as Neumann-type (i.e. on \f$n_i \nabla^i u\f$)
+/// if \f$|b| > 0\f$ and as Dirichlet-type (i.e. on \f$u\f$) if \f$b = 0\f$.
+class Neumann : public elliptic::BoundaryConditions::BoundaryCondition<3> {
```
nilsvu: Call this `StarSurface` or `RegularityAtStarSurface` or something like that and add docs

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,105 @@
+
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+/// \file
+/// Defines class Poisson::FirstOrderSystem
```
nilsvu: delete

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,105 @@
...
+#include "Utilities/TMPL.hpp"
+
+namespace IrrotationalBns {
+
+/*!
+ * \brief The Irrotational Bns equations From Baumgarte and Shapiro Chapter 15
+ *  formulated as a set of coupled first-order PDEs.
+ *
+ * \details This system formulates the Irrotational Bns HSE equations for the
+ *  velocity potential \f$\Phi\f$. For a background matter distribution (given
```
nilsvu: (optional) note that you can write standard latex `$\Phi$` and `\begin{align}...\end{align}`; you don't need those `\f`.

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,105 @@
...
+ *  by the specific enthalpy h) and a background metric \f$\gamma_{ij}\f$.
+ *
+ * \f{align*}
+ * D_i U^i = -U^i D_i \ln \left(\frac{\alpha}{h}) \\
```
nilsvu: Shall we just use `F` everywhere instead of `U`? At least that's just one extra variable name instead of two :D

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+ * \f}
+ *
+ * where we have chosen the velocity potential gradient as an auxiliary
```
nilsvu: delete this part about auxiliary variables (don't have them anymore)

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,105 @@
...
+ * \brief The Irrotational Bns equations From Baumgarte and Shapiro Chapter 15
+ *  formulated as a set of coupled first-order PDEs.
+ *
+ * \details This system formulates the Irrotational Bns HSE equations for the
+ *  velocity potential \f$\Phi\f$. For a background matter distribution (given
+ *  by the specific enthalpy h) and a background metric \f$\gamma_{ij}\f$.
+ *
+ * \f{align*}
+ * D_i U^i = -U^i D_i \ln \left(\frac{\alpha}{h}) \\
+
```
nilsvu: missing some terms, define F^i, etc

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,105 @@
...
+ * f_u &= f(x) \\
+ * F^i_{v_j} &= u \delta^i_j \\
+ * S_{v_j} &= v_j \\
+ * f_{v_j} &= 0 \text{.}
+ * \f}
+ *
+ * The fluxes and sources simplify significantly when the background metric is
+ * flat and we employ Cartesian coordinates so \f$\gamma_{ij} = \delta_{ij}\f$
+ * and \f$\Gamma^i_{jk} = 0\f$. Set the template parameter `BackgroundGeometry`
+ * to `IrrotationalBns::Geometry::FlatCartesian` to specialise the system for
```
nilsvu: Do you think we'll need the flat case at all or should we just get rid of it and always assume a curved background metric?

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
```diff
@@ -0,0 +1,149 @@
...
+#include "Elliptic/Systems/Poisson/Equations.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/MakeWithValue.hpp"
+
+namespace IrrotationalBns {
+
+void flat_potential_fluxes(
+    const gsl::not_null<tnsr::I<DataVector, 3>*> flux_for_potential,
+    const tnsr::II<DataVector, 3>& rotational_shift_stress,
+    const tnsr::i<DataVector, 3>& auxiliary_velocity) {
```
nilsvu: Maybe call this `velocity_potential_gradient` (or `enthalpy_current`, but that's maybe not so obvious)

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
```diff
@@ -0,0 +1,149 @@
...
+        log_deriv_lapse_over_specific_enthalpy.get(i);
+  }
+}
+
+void add_curved_potential_sources(
+    const gsl::not_null<Scalar<DataVector>*> source_for_potential,
+    const tnsr::i<DataVector, 3>& log_deriv_of_lapse_over_specific_enthalpy,
+    const tnsr::i<DataVector, 3>& christoffel_contracted,
+    const tnsr::I<DataVector, 3>& flux_for_potential) {
+  // The flux for the potential is the auxiliary velocity
```
nilsvu: The flux is the quantity you compute above, Eq. (25) in your notes.

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
```diff
@@ -0,0 +1,149 @@
...
+    gsl::not_null<tnsr::I<DataVector, 3>*> face_flux_for_potential,
+    const tnsr::i<DataVector, 3>& face_normal,
+    const tnsr::I<DataVector, 3>& face_normal_vector,
+    const tnsr::II<DataVector, 3>& rotational_shift_stress,
+    const Scalar<DataVector>& velocity_potential) {
+  ::tenex::evaluate<ti::I>(
+      face_flux_for_potential,
+      velocity_potential() *
+          (face_normal_vector(ti::I) -
+           rotational_shift_stress(ti::I, ti::J) * face_normal(ti::j)));
```
nilsvu: This quantity `B^{ij} n_j` _could_ be precomputed as an optimization. Maybe just add a comment here for the possible performance optimization.

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
```diff
@@ -0,0 +1,149 @@
...
+    const gsl::not_null<Scalar<DataVector>*> source_for_potential,
+    const tnsr::i<DataVector, 3>& log_deriv_of_lapse_over_specific_enthalpy,
+    const tnsr::i<DataVector, 3>& christoffel_contracted,
+    const tnsr::I<DataVector, 3>& flux_for_potential) {
+  // The flux for the potential is the auxiliary velocity
+  // (when the index is in the right place)
+  ::tenex::update(source_for_potential,
+                  (*source_for_potential)() +
+                      flux_for_potential(ti::I) *
+                          (log_deriv_of_lapse_over_specific_enthalpy(ti::i) -
```
nilsvu: Does this include the factor of `\rho` as well in `D_i \ln(\alpha\rho / h)`?

Also I think the Christoffels should be added, not subtracted? The equation is `-\partial_i F^i + S = f` please check.

src/Elliptic/Systems/IrrotationalBns/IrrotationalBns.hpp
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * We solve the equations for the relativistic quasi-equilibrium configuration
+ * assocaited with an object in a quasi-circular orbit around a binary
+ * companion.  We assume the compact object is nonspinning, which allows the
+ * solution to be writen in terms of a "velocity potential", which is a
+ * function \f$\phi\f$ such that \f$D_i \phi = \rho h u_i\f$, with \f$ \rho,
+ * h\f$ the rest mass density and specific enthalpy respectively, and \f$ the
+ * spatial four velocity\f$
+ *
+ */
```
nilsvu: Add the empty namespace here that you want to document

src/Elliptic/Systems/IrrotationalBns/Tags.hpp
```diff
@@ -0,0 +1,89 @@
...
+ *
+ *
+ * \f[\Sigma^i_j = \frac{1}{2}\frac{B^iB_j}{\alpha^2}\f]
+ */
+template <typename DataType>
+struct RotationalShiftStress : db::SimpleTag {
+  using type = tnsr::II<DataType, 3>;
+};
+/*!
+ * \brief  The derivative  \f$D_i \ln (\alpha/h)\f$
```
nilsvu: Shouldn't there be a `\rho_0` in here?

src/Elliptic/Systems/IrrotationalBns/Tags.hpp
```diff
@@ -0,0 +1,89 @@
...
+};
+/*!
+ * \brief The auxiliary velocity variable that represents
+ * the flux of the velocity potential
+ * \f[
+ *  U_i \equiv D_i \Phi  - \frac{C + B^jD_j \Phi}{\alpha^2}B_i
+ * \f]
+ */
+template <typename DataType>
+struct AuxiliaryVelocity : db::SimpleTag {
```
nilsvu: You can delete this tag, and also some or all of the tags below this (at least for now).

nilsvu (changes requested):

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
isaaclegred: Oh huh, that makes things easier.  I'll start doing that from now on

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
isaaclegred: Yes, it does include the factor of `\rho`, I'll change the variable name to make that obvious but maybe we can condense it in the future.

Hm, I would think the Christoffel should be be subtracted,  we have the equation which looks like `-D_i F^i  + S_flat = f ` and then we expand the term `D_i F^i  = \partial_i F^i + Gamma^i_{ij}F^j` so `-Gamma^i_{ij} F^j` gets added to the sources

src/Elliptic/Systems/IrrotationalBns/Tags.hpp
isaaclegred: So basically If I deleted these now they would get added back in when the executable is added?

src/Elliptic/Systems/IrrotationalBns/BoundaryConditions/Neumann.cpp
```diff
@@ -12,31 +12,43 @@
...
   set_number_of_grid_points(n_dot_auxiliary_velocity, *velocity_potential);
-  get(*n_dot_auxiliary_velocity) = 0.0;
+  tenex::evaluate<>(n_dot_auxiliary_velocity,
+                    -euler_enthalpy_constant / square(lapse()) *
+                        rotational_shift(ti::I) * normal(ti::i));
 }
 
-void Neumann::apply_linearized(
+// The constant term on the RHS is nonlinear with respect to the
+// operator, so the linearized form of the RHS is zero.
```
nilsvu: "The term on the RHS is constant w.r.t. the variables, so the linearized form of the RHS is zero."

src/Elliptic/Systems/IrrotationalBns/BoundaryConditions/Neumann.hpp
```diff
@@ -23,63 +28,78 @@ class DataVector;
 
 namespace IrrotationalBns ::BoundaryConditions {
 
-/// Impose Neumann boundary conditions \f$a u + b n_i \nabla^i u = c\f$. The
-/// boundary condition is imposed as Neumann-type (i.e. on \f$n_i \nabla^i u\f$)
-/// if \f$|b| > 0\f$ and as Dirichlet-type (i.e. on \f$u\f$) if \f$b = 0\f$.
-class Neumann : public elliptic::BoundaryConditions::BoundaryCondition<3> {
+/// Impose StarSurface boundary conditions
+/// \f[
+///    n_i F^i = \frac{1}{\sqrt{\partial_i \rho \partial_j \rho \gamma^{ij}}
```
nilsvu: This should be just `n_i F^i = C/square(alpha) B^i n_i`, right?

src/Elliptic/Systems/IrrotationalBns/BoundaryConditions/Neumann.cpp
```diff
@@ -12,31 +12,43 @@
...
+    gsl::not_null<Scalar<DataVector>*> n_dot_auxiliary_velocity,
+    const tnsr::i<DataVector, 3>& /*deriv_velocity_potential*/,
+    const Scalar<DataVector>& lapse,
+    const tnsr::I<DataVector, 3>& rotational_shift,
+    const double euler_enthalpy_constant,
+    const tnsr::i<DataVector, 3>& normal) {
   set_number_of_grid_points(n_dot_auxiliary_velocity, *velocity_potential);
-  get(*n_dot_auxiliary_velocity) = 0.0;
+  tenex::evaluate<>(n_dot_auxiliary_velocity,
+                    -euler_enthalpy_constant / square(lapse()) *
```
nilsvu: I think the minus sign is incorrect?

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
```diff
@@ -47,6 +47,8 @@ void fluxes_on_face(
     const tnsr::I<DataVector, 3>& face_normal_vector,
     const tnsr::II<DataVector, 3>& rotational_shift_stress,
     const Scalar<DataVector>& velocity_potential) {
+  // potential optimization : precompute the produce of the `face_normal`
```
nilsvu: produce -> product

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
```diff
@@ -56,37 +58,38 @@ void fluxes_on_face(
...
-  ::tenex::update(source_for_potential,
-                  (*source_for_potential)() +
-                      flux_for_potential(ti::I) *
-                          (log_deriv_of_lapse_over_specific_enthalpy(ti::i) -
-                           christoffel_contracted(ti::i)));
+  ::tenex::update(
+      source_for_potential,
+      (*source_for_potential)() +
+          flux_for_potential(ti::I) *
+              (log_deriv_lapse_times_density_over_specific_enthalpy(ti::i) -
```
nilsvu: I think both of these terms should have a minus sign. Please check.

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -28,33 +25,39 @@ namespace IrrotationalBns {
...
- * D_i U^i = -U^i D_i \ln \left(\frac{\alpha}{h}) \\
-
- * \f}
- *
- * where we have chosen the velocity potential gradient as an auxiliary
- * variable \f$D_i \Phi = U_i \equiv h u_i\f$ with \f$u_i\f$ the four velocity
+ * velocity potential \f$\Phi\f$. For a background matter distribution (given
+ * by the specific enthalpy h) and a background metric \f$\gamma_{ij}\f$.
+ * The velocity potential is defined by \f$D_i \Phi = h u_i\f$ with \f$u_i\f$
+ (the spatial part of) the four velocity
```
nilsvu: fix this line (and more below that don't start with ` *`)

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -28,33 +25,39 @@ namespace IrrotationalBns {
...
+ * rotational shift is defined as \f$B^i = \beta^i + k^i\f$ which is
+ heuristically the background
+ * motion of the spacetime.
  *
  * The system can be formulated in terms of these fluxes and sources (see
  * `elliptic::protocols::FirstOrderSystem`):
  *
+ *
+ * \f{align*}
+ * D_i F^i + S = f
```
nilsvu: `-\partial_i F^i + S = f` (note the minus sign and `\partial_i`). This means some minus signs in the equations below need to be fixed too.

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -28,33 +25,39 @@ namespace IrrotationalBns {
...
+ * \f}
+ *
  * \f{align*}
- * F^i_u &= \gamma^{ij} v_j(x) \\
- * S_u &= -\Gamma^i_{ij}\gamma^{jk}v_k \\
- * f_u &= f(x) \\
- * F^i_{v_j} &= u \delta^i_j \\
- * S_{v_j} &= v_j \\
- * f_{v_j} &= 0 \text{.}
+ * F^i &=  h u^i - \frac{B^j h u_j}{\alpha^2}B^i  \\
```
nilsvu: `h u^i` -> `\partial_i \phi` to make clear that the flux depends on the gradient of the potential

nilsvu (changes requested):

src/Elliptic/Systems/IrrotationalBns/Tags.hpp
nilsvu: You can keep the ones you surely will need, like the Euler constant. But e.g. the `FixedSources` tag you don't need because there's already one in `DataStructures/DataBox/Prefixes.hpp`.

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
nilsvu: Yes you're right it should be subtracted, but the other term as well if I did the math correctly

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
nilsvu: So I suggest you just delete the special cases for the flat background. It's a performance optimization which we likely won't need. You can always set the background metric to a flat metric, e.g. in a test that compares with a flat-space Python implementation or something like that.

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
nilsvu: ping, fix or delete this comment

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
isaaclegred: 🥦 -> ✖️  👍

isaaclegred:
I think everything should be done except removing the "flat" things.  I'll do that in a separate commit after squashing all of this because I plan on keeping it around on a branch just in case it becomes useful for checking that the executable is working
.

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
nilsvu: still a minus sign to fix (unless I got the math wrong)

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * `elliptic::protocols::FirstOrderSystem`):
+ *
+ *
+ * \f{align*}
+ * -\partial_i F^i + S = f
+ * \f}
+ *
+ * \f{align*}
+ * F^i &=  D_i \phi - \frac{B^j D_j \phi}{\alpha^2}B^i  \\
+ * S &= -F^iD_i \left( \ln \frac{\alpha \rho}{h}\right) -\Gamma^i_{ij}F^j \\
```
nilsvu: Can you add spin terms already? This probably also means we should name the executable something other than "irrotational".

nilsvu (changes requested):
You can squash everything

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
isaaclegred: I'd like to have the irrotational case merged first, at that point it should be straightforward to extend to the spin case but also at that point things stop being "exact".  I'd like to go through the sources and understand what's going on and write up notes before I add that though. 

The minus sign is my bad I have confused myself thoroughly with signs.

src/Elliptic/Systems/IrrotationalBns/BoundaryConditions/CMakeLists.txt
```diff
@@ -0,0 +1,30 @@
...
+# See LICENSE.txt for details.
+
+set(LIBRARY IrrotationalBnsBoundaryConditions)
+
+add_spectre_library(${LIBRARY})
+
+spectre_target_sources(
+  ${LIBRARY}
+  PRIVATE
+  Neumann.cpp
```
nilsvu: Name the files `StarSurface.*pp`

src/Elliptic/Systems/IrrotationalBns/CMakeLists.txt
```diff
@@ -0,0 +1,37 @@
...
+  Equations.cpp
+  )
+
+spectre_target_headers(
+  ${LIBRARY}
+  INCLUDE_DIRECTORY ${CMAKE_SOURCE_DIR}/src
+  HEADERS
+  Equations.hpp
+  FirstOrderSystem.hpp
+  Geometry.hpp
```
nilsvu: remove `Geometry.hpp`

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
nilsvu: ping

src/Elliptic/Systems/IrrotationalBns/Equations.cpp
```diff
@@ -0,0 +1,97 @@
...
+    const tnsr::I<DataVector, 3>& flux_for_potential) {
+  ::tenex::update(
+      source_for_potential,
+      (*source_for_potential)() -
+          flux_for_potential(ti::I) *
+              (log_deriv_lapse_times_density_over_specific_enthalpy(ti::i) +
+               christoffel_contracted(ti::i)));
+}
+
+// Curved-space (generic fluxes)
```
nilsvu: delete comment. Also I don't understand the next comment below. Maybe a leftover from the flat-space specialization?

src/Elliptic/Systems/IrrotationalBns/Equations.hpp
```diff
@@ -0,0 +1,111 @@
...
+ */
+void fluxes_on_face(
+    gsl::not_null<tnsr::I<DataVector, 3>*> face_flux_for_potential,
+    const tnsr::i<DataVector, 3>& face_normal,
+    const tnsr::I<DataVector, 3>& face_normal_vector,
+    const tnsr::II<DataVector, 3>& rotational_shift_stress,
+    const Scalar<DataVector>& velocity_potential);
+
+/*!
+ * \brief Compute the generic fluxes \f$ F^i = \f$ for the Irrotational BNS
```
nilsvu: Equation is missing

src/Elliptic/Systems/IrrotationalBns/Equations.hpp
```diff
@@ -0,0 +1,111 @@
...
+struct Fluxes;
+struct Sources;
+}  // namespace IrrotationalBns
+/// \endcond
+
+namespace IrrotationalBns {
+
+/*!
+ * \brief Compute the fluxes
+ * \f[ F^i = \gamma^{ij} n_j \Phi  + n_i \Phi
```
nilsvu: Another minus sign to check :)

src/Elliptic/Systems/IrrotationalBns/Equations.hpp
```diff
@@ -0,0 +1,111 @@
...
+ * equation for the velocity potential.
+ */
+void potential_fluxes(
+    gsl::not_null<tnsr::I<DataVector, 3>*> flux_for_potential,
+    const tnsr::II<DataVector, 3>& rotational_shift_stress,
+    const tnsr::II<DataVector, 3>& inverse_spatial_metric,
+    const tnsr::i<DataVector, 3>& velocity_potential_gradient);
+
+/*!
+ * \brief Add the sources \f$S=-\Gamma^i_{ij}F^j\f$
```
nilsvu: Missing a term

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,104 @@
...
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace IrrotationalBns {
+
+/*!
+ * \brief The Irrotational Bns equations From Baumgarte and Shapiro Chapter 15
+ *  formulated as a set of coupled first-order PDEs.
+ *
+ * \details This system formulates the Irrotational Bns HSE equations for the
```
nilsvu: what does "HSE" mean? Please define or delete

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,104 @@
...
+ *
+ *
+ * \f{align*}
+ * -\partial_i F^i + S = f
+ * \f}
+ *
+ * \f{align*}
+ * F^i &=  D_i \phi - \frac{B^j D_j \phi}{\alpha^2}B^i  \\
+ * S &= -F^iD_i \left( \ln \frac{\alpha \rho}{h}\right) -\Gamma^i_{ij}F^j \\
+ * f &= -D_i \left(\frac{C B^i}{\alpha^2}\right)  \frac{C}{\alpha^2}B^iD_i\left(
```
nilsvu: Is this missing a minus or plus?

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
nilsvu: Remove paragraph

src/Elliptic/Systems/IrrotationalBns/IrrotationalBns.hpp
```diff
@@ -0,0 +1,23 @@
...
+
+#pragma once
+
+/*!
+ * \ingroup EllipticSystemsGroup
+ * \brief Items related to solving for irrotational Binary Neutron Star initial
+ * data.
+ *
+ * We solve the equations for the relativistic quasi-equilibrium configuration
+ * assocaited with an object in a quasi-circular orbit around a binary
```
nilsvu: "associated"

src/Elliptic/Systems/IrrotationalBns/IrrotationalBns.hpp
```diff
@@ -0,0 +1,23 @@
...
+ * companion.  We assume the compact object is nonspinning, which allows the
+ * solution to be writen in terms of a "velocity potential", which is a
```
nilsvu: "written"

src/Elliptic/Systems/IrrotationalBns/IrrotationalBns.hpp
```diff
@@ -0,0 +1,23 @@
...
+ * function \f$\phi\f$ such that \f$D_i \phi = \rho h u_i\f$, with \f$ \rho,
+ * h\f$ the rest mass density and specific enthalpy respectively, and \f$ the
```
nilsvu: error in definition of "u_i"

src/Elliptic/Systems/IrrotationalBns/Tags.hpp
```diff
@@ -0,0 +1,73 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+/// \file
+/// Defines DataBox tags for the Poisson system
```
nilsvu: delete

tests/Unit/Elliptic/Systems/IrrotationalBns/BoundaryConditions/CMakeLists.txt
```diff
@@ -0,0 +1,20 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+set(LIBRARY "Test_IrrotationalBnsBoundaryConditions")
+
+set(LIBRARY_SOURCES
+  Test_Neumann.cpp
```
nilsvu: Also rename to `Test_StarSurface.cpp`

tests/Unit/Elliptic/Systems/IrrotationalBns/BoundaryConditions/Test_Neumann.cpp
```diff
@@ -0,0 +1,119 @@
...
+        const tnsr::I<DataVector, 3> rotational_shift{
+            {DataVector{1.0, 1.0, 1.0}, DataVector{0.0, 0.0, 0.0},
+             DataVector{0.0, 0.0, 0.0}}};
+        const tnsr::i<DataVector, 3> normal{{DataVector{1.0, 0.0, 0.0},
+                                             DataVector{0.0, 1.0, 0.0},
+                                             DataVector{0.0, 0.0, 1.0}}};
+        const double euler_enthalpy_constant = 1.0;
+        apply_neumann_boundary_condition<false>(
+            make_not_null(&n_dot_auxilliary_velocity));
+        CHECK(get(n_dot_auxilliary_velocity) ==
```
nilsvu: You aren't actually comparing to the Python implementation I think?

tests/Unit/Elliptic/Systems/IrrotationalBns/Test_Tags.cpp
```diff
@@ -0,0 +1,16 @@
...
+#include <string>
+
+#include "Elliptic/Systems/IrrotationalBns/Tags.hpp"
+#include "Helpers/DataStructures/DataBox/TestHelpers.hpp"
+
+SPECTRE_TEST_CASE("Unit.Elliptic.Systems.IrrotationalBns.Tags",
+                  "[Unit][Elliptic]") {
+  TestHelpers::db::test_simple_tag<
+      IrrotationalBns::Tags::VelocityPotential<DataVector>>(
+      std::string{"VelocityPotential"});
```
nilsvu: also test the other tags here

src/Elliptic/Systems/IrrotationalBns/FirstOrderSystem.hpp
nilsvu: Ok that's fine. Probably still we should not name things "irrotational", don't you think? That seems confusing once we add the spin terms (and we don't want to rename everything later).

nilsvu (changes requested):

isaaclegred:
@nilsvu  I renamed everything to `BnsInitialData` in order to make it more general.  Also fixed up comments.

src/Elliptic/Systems/BnsInitialData/BoundaryConditions/Factory.hpp
```diff
@@ -4,13 +4,13 @@
 #pragma once
 
 #include "Elliptic/BoundaryConditions/AnalyticSolution.hpp"
-#include "Elliptic/Systems/IrrotationalBns/BoundaryConditions/Neumann.hpp"
+#include "Elliptic/Systems/BnsInitialdata/BoundaryConditions/StarSurface.hpp"
```
nilsvu: Typo `BnsInitialData`

tests/Unit/Elliptic/Systems/BnsInitialData/BoundaryConditions/Test_StarSurface.cpp
```diff
@@ -64,9 +65,42 @@ void apply_neumann_boundary_condition(
...
+    const tnsr::I<DataVector, 3> rotational_shift,
+    const double euler_enthalpy_constant, const tnsr::i<DataVector, 3> normal) {
+  const StarSurface boundary_condition{{}};
+
+  const auto x = tnsr::I<DataVector, 3>{{DataVector{1.0, 1.0, 1.0},
+                                         DataVector{0.0, 0.0, 0.0},
+                                         DataVector{0.0, 0.0, 0.0}}};
+
+  const auto direction = Direction<3>::lower_xi();
+  ;
```
nilsvu: delete empty line

src/Elliptic/Systems/BnsInitialData/BoundaryConditions/StarSurface.hpp
```diff
@@ -0,0 +1,105 @@
...
+/// \cond
+class DataVector;
+/// \endcond
+
+namespace BnsInitialData ::BoundaryConditions {
+
+/*!
+ * Impose StarSurface boundary conditions:
+ * \f[
+ *   n_i F^i = \frac{C}{\alpha^2} B^i n_i.
```
nilsvu, line 34: Question: Since this is a Neumann boundary condition, it leaves the freedom for a constant offset in `\Phi`. I think the equations also leave this freedom. Have you thought about how to fix this freedom?

src/Elliptic/Systems/BnsInitialData/BoundaryConditions/StarSurface.hpp
```diff
@@ -0,0 +1,105 @@
...
+
+  std::vector<elliptic::BoundaryConditionType> boundary_condition_types()
+      const override {
+    return {elliptic::BoundaryConditionType::Neumann};
+  }
+
+  using argument_tags =
+      tmpl::list<gr::Tags::Lapse<DataVector>,
+                 BnsInitialData::Tags::RotationalShift<DataVector>,
+                 BnsInitialData::Tags::EulerEnthalpyConstant,
```
nilsvu, line 78: Add this tag also to `volume_tags` below so we don't have to store it on every element face.

src/Elliptic/Systems/BnsInitialData/Equations.hpp
```diff
@@ -0,0 +1,111 @@
...
+struct Sources;
+}  // namespace BnsInitialData
+/// \endcond
+
+namespace BnsInitialData {
+
+/*!
+ * \brief Compute the fluxes
+ * \f[ F^i = \gamma^{ij} n_j \Phi  - n_i \Phi
+ * \frac{B^iB^j}{\alpha^2}  \f] where \f$ n_j\f$ is the
```
nilsvu, line 31: swap `i` and `j` in the second term so the indices match the LHS

nilsvu (changes requested):
You can squash everything

src/Elliptic/Systems/BnsInitialData/BoundaryConditions/StarSurface.hpp
isaaclegred: I haven't, I would think a reasonable choice would be that Phi should be zero at the center of the star or something like that, we probably can't set Phi=0 at the surface because I'm almost sure it's not "equipotential"

src/Elliptic/Systems/BnsInitialData/BoundaryConditions/StarSurface.hpp
isaaclegred: So I just add it to `volume_tags` and it will magically not be saved on every element face?

nilsvu: It will not be _retrieved_ from the faces, but just directly from the DataBox. So you don't have to add `domain::Tags::Faces<EulerEnthalpyConstant>` to the DataBox, but just add `EulerEnthalpyConstant` directly.

nilsvu (dismissed):
Just have to appease CI :)

nilsvu (approved):

PR #6056, opened 2024-06-03
Exporter: allow extrapolation into excisions 

nilsvu:
## Proposed changes

This is needed to fill the excision region with (non-constraint-satisfying) data so it can be imported into moving puncture codes.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/Exporter/Exporter.cpp
+++ b/src/IO/Exporter/Exporter.cpp
@@ -8,6 +8,7 @@
 #include <omp.h>
 #endif  // _OPENMP
 
+#include "DataStructures/Tensor/EagerMath/CartesianToSpherical.hpp"
 #include "Domain/BlockLogicalCoordinates.hpp"
 #include "Domain/Creators/RegisterDerivedWithCharm.hpp"
 #include "Domain/Creators/TimeDependence/RegisterDerivedWithCharm.hpp"
@@ -19,6 +20,7 @@
 #include "IO/H5/TensorData.hpp"
 #include "IO/H5/VolumeData.hpp"
 #include "NumericalAlgorithms/Interpolation/IrregularInterpolant.hpp"
+#include "NumericalAlgorithms/Interpolation/PolynomialInterpolation.hpp"
 #include "Utilities/FileSystem.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/GetOutput.hpp"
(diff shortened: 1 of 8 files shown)


src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -125,14 +126,57 @@ block_logical_coordinates_single_point(
...
+    }
+    if (not x_logical.has_value()) {
+      continue;
+    }
+    // Discard block if the point is radially inside the block or on the other
+    // side of the excision sphere
+    const double radial_distance_this_block =
+        x_logical->get(direction.dimension()) * direction.sign();
+    if (radial_distance_this_block < 1.) {
+      continue;
```
wthrowe: If I follow the logic correctly, if we hit this case the point is not inside the excision sphere.  Should that be an error or an early return?

src/Domain/BlockLogicalCoordinates.hpp
```diff
@@ -69,10 +71,30 @@ auto block_logical_coordinates(
...
+    bool allow_extrapolation = false);
 /// @}
+
+/*!
+ * \brief Finds the block-logical coordinates of a point in an excision sphere.
+ *
+ * The point will be reported to belong to one of the blocks that abut the
+ * excision sphere and will have a logical coordinate outside the range [-1, 1]
+ * in the radial direction. This is useful for extrapolating data into excision
+ * regions.
```
wthrowe: What happens if `input_point` is not inside `excision_sphere`?  Either document the result or state it is considered invalid to call the function in that case.

src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -125,14 +126,57 @@ block_logical_coordinates_single_point(
...
+    }
+    // The checks above should leave only 1 valid block, so return that
+    return make_id_pair(domain::BlockId(block_id),
```
wthrowe: Could there be two valid blocks for points on the extrapolation of the boundary between two blocks into the sphere?  The blocks are being pulled from an unordered container, so always choosing the first one isn't deterministic.

src/IO/Exporter/Exporter.cpp
```diff
@@ -123,7 +125,222 @@ void interpolate_to_points(
...
+#pragma omp parallel num_threads(num_threads)
+  {
+    std::vector<BlockLogicalCoords<Dim>> extra_block_logical_coords{};
+    std::vector<ExtrapolationInfo<ExtrapolationOrder>>
+        extra_extrapolation_info{};
+#pragma omp for nowait
+    for (size_t s = 0; s < block_logical_coords->size(); ++s) {
+      auto& block_logical_coord = (*block_logical_coords)[s];
+      // Process only points inside the excision sphere. These points were found
+      // by `block_logical_coordinates` above, have a radial logical coordinate
```
wthrowe: `block_logical_coordinates` doesn't appear anywhere above.

src/IO/Exporter/Exporter.hpp
```diff
@@ -45,6 +45,15 @@ struct ObservationStep {
...
  * Look into the H5 file to see what components are available.
  * \param target_points The points to interpolate to, in inertial coordinates.
+ * \param extrapolate_into_excisions Enables extrapolation into excision regions
+ * of the domain (default is `false`). This can be useful to fill the excision
+ * region with (constraint-violating but smooth) data so it can be imported into
+ * moving puncture codes. Specifically, we implement the strategy used in
+ * \cite Etienne2008re adjusted for distorted excisions: we choose uniformly
+ * spaced radial anchor points spaced as $\Delta r = 0.3 r_\mathrm{AH}$ in the
+ * grid frame (where the excision is spherical), then map the anchor points to
+ * the distorted frame (where we have the target point) and do a 7th order
```
wthrowe, line 55: The code has `constexpr size_t extrapolation_order = 8;`

wthrowe (changes requested):
Exporter test failed on one of the CI jobs.

nilsvu:
Thanks for your review @wthrowe! I pushed a fixup.

nilsvu:
I'm investigating the NaNs in some of the unit tests, but the fixup is ready for your review @wthrowe

src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -153,22 +153,26 @@ BlockLogicalCoords<Dim> block_logical_coordinates_in_excision(
       continue;
     }
     // Discard block if the point has angular logical coordinates outside the
-    // range [-1, 1]
+    // range [-1, 1]. Also discard if the point is on the upper angular face of
+    // the block, to disambiguate which of the two blocks sharing the face the
+    // point is in.
```
wthrowe: This doesn't work.  The blocks don't have aligned frames, so some interfaces are two lower sides or two upper sides.

src/IO/Exporter/Exporter.hpp
```diff
@@ -49,11 +49,11 @@ struct ObservationStep {
...
  * region with (constraint-violating but smooth) data so it can be imported into
  * moving puncture codes. Specifically, we implement the strategy used in
- * \cite Etienne2008re adjusted for distorted excisions: we choose uniformly
+ * \cite Etienne2008re adjusted for distorted excisions: we choose 8 uniformly
  * spaced radial anchor points spaced as $\Delta r = 0.3 r_\mathrm{AH}$ in the
  * grid frame (where the excision is spherical), then map the anchor points to
- * the distorted frame (where we have the target point) and do a 7th order
- * polynomial extrapolation into the excision region.
+ * the distorted frame (where we have the target point) and do a polynomial
+ * extrapolation into the excision region.
```
wthrowe: Is it 8th order or not?

wthrowe (changes requested):

src/IO/Exporter/Exporter.hpp
nilsvu: It's a 7th order polynomial with 8 collocation points. However in the `polynomial_interpolation` function the name `order` is `degree + 1`, so I used that naming here as well:
https://github.com/sxs-collaboration/spectre/blob/bc197745bb08b2397fb31dda402a10b4d17cc856/src/NumericalAlgorithms/Interpolation/PolynomialInterpolation.cpp#L19
To avoid confusion I renamed to `num_extrapolation_anchors`.

nilsvu:
@wthrowe I pushed a new fixup. I dropped the changes to `BlockLogicalCoordinates.*pp` and instead determine if the target point is inside an excision in the distorted frame. I think this is more robust and addresses your comments above.

wthrowe:
OK, looks good.  Squash and fix the clang-tidy complaint.

wthrowe:
@nilsvu I filed a CI failure as #6168 and I'm rerunning it as I'm guessing it's unrelated to this change.  Do you agree with that guess?

Edit: (The gcc-11 Release failure is obviously unrelated so I'm ignoring it.)

nilsvu:
I don't see a connection 🤷‍♂️

wthrowe (approved):
Sounds good.  Let's let the rerun finish, but feel free to merge once that's done (or I'll do it later, but I'm leaving for the day soon).

PR #6058, opened 2024-06-04
AMR: do not enforce 2:1 balance in normal direction

kidder:
## Proposed changes

Remove requirement that h-refinement must have 2:1 balance in the direction normal to faces.

### Upgrade instructions

In input files that use AMR, you will need to add `EnforceTwoToOneBalanceInNormalDirection: true` to the list of Policies

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/InitialData.yaml
+++ b/support/Pipelines/Bbh/InitialData.yaml
@@ -133,6 +133,7 @@ Amr:
   Verbosity: Verbose
   Criteria: []
   Policies:
+    EnforceTwoToOneBalanceInNormalDirection: true
     Isotropy: Anisotropic
     Limits:
       NumGridPoints: Auto
--- a/src/Domain/Amr/UpdateAmrDecision.cpp
+++ b/src/Domain/Amr/UpdateAmrDecision.cpp
@@ -22,7 +22,8 @@ template <size_t VolumeDim>
 bool update_amr_decision(
     const gsl::not_null<std::array<Flag, VolumeDim>*> my_current_amr_flags,
     const Element<VolumeDim>& element, const ElementId<VolumeDim>& neighbor_id,
-    const std::array<Flag, VolumeDim>& neighbor_amr_flags) {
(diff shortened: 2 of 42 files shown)


kidder:
@nilsvu here is the unfinished (and untested) branch for what you asked about

kidder:
@nilsvu this is now ready for review.  It did not cause any problems when I ran the RandomAmr executables.

support/Pipelines/Bbh/InitialData.yaml
```diff
@@ -133,6 +133,7 @@ Amr:
   Verbosity: Verbose
   Criteria: []
   Policies:
+    EnforceTwoToOneBalanceInNormalDirection: true
```
nilsvu, line 136: Maybe set all these options to `false`? To me at least that's the expected behavior of AMR. Do we even have a use case for `true`?

nilsvu (approved):
Looks good! Thanks for adding this.

support/Pipelines/Bbh/InitialData.yaml
kidder: I set them to `true` to reproduce previous behavior.  Until setting them to `false` is more extensively tested, I'd rather keep them `true` for now.

nilsvu: Ok Im just worried that this will remain `true` and propagate to many input files and people will get unexpected behavior when trying AMR

nilsdeppe: I think in evolution problems we probably want it true. Geoffrey had issues if the grid size changed too much form one element to the next in the normal direction to the interface.

nilsvu: That's actually what I'm trying to prevent as well, in a domain that has differently sized blocks. For example, think of the thermal noise problem with thin layers on top of a substrate. You want to h-refine the substrate multiple times in z-direction, but each layer stays at h-refinement level 0.

Anyway, we can merge the PR like this and just have to keep in mind to experiment with this option.

PR #6060, opened 2024-06-04
Docker ubuntu 18.04 support and static execs

nilsdeppe:
## Proposed changes

This adds most of the remaining code needed to CD the CCE executables.

The container is already push to the Dockerhub sxscollaboration image. I previously tested on my fork.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/.github/workflows/BuildDockerContainer.yaml
+++ b/.github/workflows/BuildDockerContainer.yaml
@@ -73,6 +73,37 @@ jobs:
           platforms: linux/amd64
           build-args: PARALLEL_MAKE_ARG=-j4
 
+  build_1804_container:
+    # Note: this container is used for deploying static execs
+    name: Build and deploy Ubuntu 18.04 container
+    runs-on: ubuntu-latest
+    # This environment can be protected to require manual approval before
+    # deployment. On forks where you set your own secrets for authentication
+    # with DockerHub you can choose not to protect this environment.
+    environment: deploy-containers
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Login to DockerHub
+        uses: docker/login-action@v3
+        with:
+          username: ${{ secrets.DOCKERHUB_USERNAME }}
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build dev container for Ubuntu 18.04
+        uses: docker/build-push-action@v5
+        with:
+          push: true
+          context: .
+          file: "./containers/Dockerfile.buildenv"
+          target: dev
+          tags: ${{ inputs.image_name }}:dev1804
+          platforms: linux/amd64
+          build-args: |
+            PARALLEL_MAKE_ARG=-j4
+            UBUNTU_VERSION=18.04
+
   run_tests:
     name: Test
     uses: ./.github/workflows/Tests.yaml
--- a/.github/workflows/DeployStaticExecutables.yaml
+++ b/.github/workflows/DeployStaticExecutables.yaml
@@ -1,0 +1,92 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+name: Deploy static executables
+
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      tag_name:
+        type: string
+        required: true
+        description: The name of the release tag
+
+jobs:
+  deploy_static_execs:
+    name: Deploy static executables and libraries
+    runs-on: ubuntu-latest
+    env:
+      RELEASE_TAG_NAME: ${{ inputs.tag_name || github.event.release.tag_name }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      # Build static executables in a Docker container with an older version
+      # of Ubuntu so that the glibc is compatible with most Linux distros.
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Build static executables
+        uses: docker/build-push-action@v5
+        with:
+          push: false
+          context: .
+          file: "./containers/Dockerfile.buildenv"
+          target: deploy_static_execs_and_libs
+          tags: sxs-collaboration/spectre:deploy_static_execs_and_libs
+          platforms: linux/amd64
+          load: true
+          build-args: |
+            UBUNTU_VERSION=18.04
+            PARALLEL_MAKE_ARG=-j4
+      - name: Run Docker container
+        run: >
+          docker run --name static-execs
+          sxs-collaboration/spectre:deploy_static_execs_and_libs
+      - name: Install python dependencies
+        run: >
+          pip3 install numpy h5py gdown
+      - name: Download extra data to include for CCE
+        run: |
+          gdown \
+          "https://drive.google.com/uc?id=1CmgLLfuLod8stc13EtmjHUvBoIgSVD-Y"
+          gdown \
+          "https://drive.google.com/uc?id=1uaESVdylNmcHwpZwogX4TVrNs8JbWNF1"
+          gdown \
+          "https://drive.google.com/uc?id=1rfzjtXtrP9pQE2az3935uN0z_5aSF1hy"
+      - name: Copy CCE executables from the container
+        run: >
+          mkdir CceExecutables
+
+          cp ./tests/InputFiles/Cce/CharacteristicExtract.yaml
+          ./CceExecutables/CharacteristicExtract.yaml
+
+          for i in CharacteristicExtract ReduceCceWorldtube; do
+            docker cp static-execs:/work/spectre/build/bin/$i ./CceExecutables;
+          done
+      - name: Test CCE executable outside of container
+        run: |
+          sed -i 's/CceR0257/BondiSachsCceR0200/g' \
+          ./CceExecutables/CharacteristicExtract.yaml
+          sed -i 's/H5IsBondiData: False/H5IsBondiData: True/g' \
+          ./CceExecutables/CharacteristicExtract.yaml
+          ./CceExecutables/CharacteristicExtract \
+          --input-file ./CceExecutables/CharacteristicExtract.yaml
+          python ./CheckCceOutput.py
+          rm CharacteristicExtractReduction.h5
+      - name: Create CCE executables release asset
+        # Note: We use xz compression since it's much better than gzip, even
+        # though it's a decent bit slower. Specifically, xz is two thirds the
+        # size of gzip.
+        run:  |
+          mv BondiSachsCceR0200.h5 ./CceExecutables/
+          mv CharacteristicExtractReduction_Expected.h5 ./CceExecutables/
+          mv CheckCceOutput.py ./CceExecutables/
+          tar cJf CceExecutables.tar.xz CceExecutables
+      - name: Upload to release
+        uses: softprops/action-gh-release@v2
+        with:
+          tag_name: ${{ env.RELEASE_TAG_NAME }}
+          fail_on_unmatched_files: true
+          files: |
+            CceExecutables.tar.xz
--- a/containers/Dockerfile.buildenv
+++ b/containers/Dockerfile.buildenv
@@ -4,8 +4,11 @@
 # If you change this file please push the new images to DockerHub. If you do not
 # have permission to push to DockerHub please coordinate with someone who does.
 # Since changes to this image affect our testing infrastructure it is important
-# all changes be carefully reviewed before they are pushed. This is how you use
-# GitHub Actions to build and push the new images:
+# all changes be carefully reviewed before they are pushed.
+#
+# GitHub Actions (recommended method)
+#
+# This is how you use GitHub Actions to build and push the new images:
 #
 # 1. Push your changes to the `develop` branch on your fork of the spectre
 #    repository on GitHub (https://github.com/YOURNAME/spectre).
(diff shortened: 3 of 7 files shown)


nilsdeppe:
I've rebased this and set up auto deployment. I'm doing a final test on my fork to make sure it's working correctly, but I expect anything wrong would be a small typo in the last step of testing the CCE output is correct. This now bundles the `CharacteristicExtract` and `ReduceCceWoldtube` executables along with a `BondiSachsR0200.h5` worldtube file, expected output file, and a python script that checks that things went okay. The last 3 parts are housed in our Google Drive since that was easiest to get working quick.

edit:

The release on my fork is running at https://github.com/nilsdeppe/spectre/actions/runs/9374720262/job/25811397363

nilsdeppe:
This successfully deployed the executables on my fork. https://github.com/nilsdeppe/spectre/releases/tag/v2024.06.04.013

.github/workflows/BuildDockerContainer.yaml
```diff
@@ -73,6 +73,37 @@ jobs:
...
+      - name: Build dev container for Ubuntu 18.04
+        uses: docker/build-push-action@v5
+        with:
+          push: true
+          context: .
+          file: "./containers/Dockerfile.buildenv"
+          target: dev
+          tags: ${{ inputs.image_name }}:dev1804
+          platforms: linux/amd64
+          build-args:
```
nilsvu: I think this needs a `|`: `build-args: |`

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,106 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+name: Deploy static executables
+
+on:
+  release:
+    types: [published]
```
nilsvu, line 8: Could you switch this to:

```yaml
  workflow_call:
    inputs:
      tag_name: ... # same as below
    secrets:
      DOCKERHUB_USERNAME:
        required: true
      DOCKERHUB_TOKEN:
        required: true
```

Then at the bottom of `PostRelease.yaml` add:

```yaml
  static_execs:
    uses: ./.github/workflows/DeployStaticExecutables.yaml
    secrets: inherit
    with:
      tag_name: ${{ inputs.tag_name || github.event.release.tag_name }}
```

Then this runs as a job in the `PostRelease` workflow together with building the other containers.

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,106 @@
...
+        description: The name of the release tag
+
+jobs:
+  deploy_static_execs:
+    name: Deploy static executables and libraries
+    runs-on: ubuntu-latest
+    # Running in a protected environment that provides the necessary secrets
+    environment: release
+    env:
+      RELEASE_TAG_NAME: ${{ inputs.tag_name || github.event.release.tag_name }}
```
nilsvu, line 21: Don't need this anymore when you make the change above. You can use `inputs.tag_name` directly.

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,106 @@
...
+        type: string
+        required: true
+        description: The name of the release tag
+
+jobs:
+  deploy_static_execs:
+    name: Deploy static executables and libraries
+    runs-on: ubuntu-latest
+    # Running in a protected environment that provides the necessary secrets
+    environment: release
```
nilsvu: Also use `deploy-containers` environment? At least that's what DemoContainers.yaml does.

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,106 @@
...
+        with:
+          push: false
+          context: .
+          file: "./containers/Dockerfile.buildenv"
+          target: deploy_static_execs_and_libs
+          tags: sxs-collaboration/spectre:deploy_static_execs_and_libs
+          platforms: linux/amd64
+          load: true
+          build-args:
+            UBUNTU_VERSION=18.04
```
nilsvu, line 40: Add `PARALLEL_MAKE_ARG=-j4`?

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,106 @@
...
+          sed -i 's/CceR0257/BondiSachsCceR0200/g'
+          ./CceExecutables/CharacteristicExtract.yaml
+
+          sed -i 's/H5IsBondiData: False/H5IsBondiData: True/g'
+          ./CceExecutables/CharacteristicExtract.yaml
+
+          ./CceExecutables/CharacteristicExtract
+          --input-file ./CceExecutables/CharacteristicExtract.yaml
+
+          python ./CheckCceOutput.py
```
nilsvu, line 75: Don't you have to point this script to the data?

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,106 @@
...
+          mkdir CceExecutables
+
+          cp ./tests/InputFiles/Cce/CharacteristicExtract.yaml
+          ./CceExecutables/CharacteristicExtract.yaml
+
+          for i in CharacteristicExtract ReduceCceWorldtube; do
+            docker cp static-execs:/work/spectre/build/bin/$i ./CceExecutables;
+          done
+
+          # Download extra data to include for CCE
```
nilsvu: (optional) I'd split this into multiple `run` steps so it's easier to track on CI (where time is spent, what fails, etc). No change to the script, just insert:

```yaml
- name: Download extra data to include for CCE
  run: >
    gdown ...
```

Also helps inserting steps like testing the exporter lib.

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,106 @@
...
+          # Copy CCE data into directory
+
+          mv BondiSachsCceR0200.h5 ./CceExecutables/
+
+          mv CharacteristicExtractReduction_Expected.h5 ./CceExecutables/
+
+          mv CheckCceOutput.py ./CceExecutables/
+
+          tar cJf CceExecutables.tar.xz CceExecutables
+      - name: Create release on GitHub
```
nilsvu: "Upload to release"

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,106 @@
...
+        uses: softprops/action-gh-release@v2
+        with:
+          token: ${{ secrets.GH_TOKEN_RELEASE }}
```
nilsvu: You can delete this line (the PAT is only needed for git pushes etc, not to just upload a file, I believe).

containers/Dockerfile.buildenv
```diff
@@ -31,18 +60,49 @@
 #                 --platform linux/amd64 \
 #                 -f ./containers/Dockerfile.buildenv .
 #
-# 2. `ci` (this should use the `dev` image you just built for most of it)
+#   # To cross compile for arm64 and amd64 use
```
nilsvu: delete this line?

containers/Dockerfile.buildenv
```diff
@@ -110,12 +220,31 @@ RUN apt-get update -y \
 # - We use python-is-python3 because on Ubuntu 20.04 /usr/bin/python was removed
 #   to aid in tracking down anything that depends on python 2. However, many
 #   scripts use `/usr/bin/env python` to find python so restore it.
+# - We install h5py from apt since that's much faster when cross compiling than
```
nilsvu: do you mean "when building the old Ubuntu 18.04 container"? This isn't related to cross-compiling, right? Is the issue that for the old ubuntu no precompiled h5py is available from pip, so it tries to build wheels?
Edit: wait this is in the new container. It should not build wheels (compile h5py from source). But I remember seeing this problem too. Try to delete the h5py version constraint from `support/Python/requirements.txt`.

containers/Dockerfile.buildenv
```diff
@@ -124,24 +253,34 @@ RUN apt-get update -y \
...
 WORKDIR /root
 COPY --from=xx / /
 ARG TARGETPLATFORM
+ARG UBUNTU_VERSION=22.04
 ARG PARALLEL_MAKE_ARG=-j4
+
 # Install build dependencies for the host system
 RUN apt-get update -y \
-    && apt-get install -y cmake clang lld wget
+    && apt-get install -y cmake clang lld flex bison wget git
```
nilsvu, line 257: Can you drop these changes to the `xbuild` stage?

containers/Dockerfile.buildenv
```diff
@@ -163,13 +302,28 @@ RUN wget https://github.com/doxygen/doxygen/archive/Release_1_9_3.tar.gz -O doxy
...
+COPY --from=xbuild /usr/local/bin/cmake /work/
+RUN /work/cmake --version \
+    && if [ ${UBUNTU_VERSION} = 18.04 ]; then \
+    rm /usr/local/bin/ccache /usr/local/bin/doxygen; \
+    else \
+    mv /work/cmake /usr/local/bin; \
+    fi
 
 
+# WARNING: We are now back in the spectre container but we have 2, one for each
```
nilsvu: The docker build doesn't know anything about the other architecture it's building for. I'd delete the "we have 2" part of this comment.

nilsvu (changes requested):

containers/Dockerfile.buildenv
nilsdeppe: Sorry, how do we not have two? Below it says `base-amd64` and `base-arm64`

containers/Dockerfile.buildenv
nilsdeppe: Resolved offline: now does `pip3 --no-cache-dir install --only-binary=h5py...`

containers/Dockerfile.buildenv
nilsdeppe: Reworded to `Cross compiling for AMD64 and ARM64:` to mark it as a "section"

.github/workflows/DeployStaticExecutables.yaml
nilsdeppe: No, the names are hard-coded :shrug:

.github/workflows/DeployStaticExecutables.yaml
nilsdeppe: Sorry, do I need the secrets? This doesn't push to DockerHub so I don't think I need the username or token. This only pushes assets to the release. I had a lot of troubles with the `workflow_call` so I'd rather keep this as-is. Essentially, it wasn't clear to me what was going on and going back and testing it again takes a decent amount of time that I don't really have

.github/workflows/DeployStaticExecutables.yaml
nilsdeppe: I would be a bit surprised by that and don't really have time to test that. Is there a major drawback to having it?

.github/workflows/DeployStaticExecutables.yaml
nilsdeppe: The deploy doesn't have access to the release token.

nilsdeppe (commented):
I've pushed some fixups. Thanks for the review!

.github/workflows/DeployStaticExecutables.yaml
nilsvu: No you're right you don't need the secrets

.github/workflows/DeployStaticExecutables.yaml
nilsvu: You don't need the release token I think. But `deploy-containers` isn't the right environment either because you don't push a container. Question is, do you want to get prompted again for approval before the static execs get uploaded? Then keep the `release` env. Otherwise, I think you can just delete it.

containers/Dockerfile.buildenv
nilsvu: From the two stages `base-amd64` and `base-arm64` only the arch we're actually building is used. The line `FROM base-${TARGETARCH} AS dev` selects it. This is effectively an `if` in the Dockerfile (it's annoying that the Dockerfile syntax doesn't make that easier).

nilsdeppe: Ah okay. What do you think of the wording:

We are now back in the spectre container. We create two base images for the 2 possible target architectures and then select the one we need as the `dev` container below. This is essentially an `if` in the Dockerfile.

nilsvu: Sounds good 👍

nilsvu:
You can squash the fixups 👍

.github/workflows/DeployStaticExecutables.yaml
nilsdeppe: I don't think we need to get prompted before upload? There is a check that the CCE run is consistent with previous results, so the action should fail if something went wrong. I guess then I'll just not specify an env. I'll do this and push. If I misunderstood just let me know and I'll fix it :)

nilsdeppe:
I've squashed, rebased, and squashed in a few of the other changes discussed. I think the bigger thing I'm hesitant to do is the `workflow_call` part. I'd tried something like that previous and must've done something wrong because it didn't work at all as expected.

.github/workflows/DeployStaticExecutables.yaml
```diff
@@ -0,0 +1,92 @@
...
+    env:
+      RELEASE_TAG_NAME: ${{ inputs.tag_name || github.event.release.tag_name }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      # Build static executables in a Docker container with an older version
+      # of Ubuntu so that the glibc is compatible with most Linux distros.
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+      - name: Build and push deploy container
```
nilsvu: "Build static executables"

containers/Dockerfile.buildenv
```diff
@@ -110,12 +203,30 @@ RUN apt-get update -y \
...
+    libssl-dev libreadline-dev libffi-dev libsqlite3-dev wget libbz2-dev \
+    make \
+    && wget https://www.python.org/ftp/python/3.10.1/Python-3.10.1.tgz \
+    && tar -xf Python-3.10.1.tgz && cd ./Python-3.10.1 \
+    && ./configure --enable-optimizations && make ${PARALLEL_MAKE_ARG} \
+    && make altinstall && cd ../ \
+    && rm -rf ./Python-3.10.1.tgz ./Python-3.10.1 \
+    && python3.10 -m pip install --upgrade pip \
+    && pip3.10 --no-cache-dir install pybind11~=2.6.1 \
+    && pip3.10 --no-cache-dir install -r requirements.txt \
```
nilsvu: Also add `--only-binary` here?

containers/Dockerfile.buildenv
```diff
@@ -124,24 +235,34 @@ RUN apt-get update -y \
...
 WORKDIR /root
 COPY --from=xx / /
 ARG TARGETPLATFORM
+ARG UBUNTU_VERSION=22.04
 ARG PARALLEL_MAKE_ARG=-j4
+
 # Install build dependencies for the host system
 RUN apt-get update -y \
-    && apt-get install -y cmake clang lld wget
+    && apt-get install -y cmake clang lld flex bison wget git
```
nilsvu, line 257: Drop the changes until line 303? No changes to the xbuild stage should be necessary

nilsvu (changes requested):
Ok the `on: release` is fine too

containers/Dockerfile.buildenv
nilsdeppe: No, these were necessary. I need bison and flex on the host arch, not the target arch.

containers/Dockerfile.buildenv
nilsdeppe: No, these were necessary to get cross compiling to work.

nilsdeppe:
I squashed the `--only-binary` change

nilsvu (approved):

PR #6066, opened 2024-06-06
Fix notification for new contributors

nilsvu:
This automation hasn't ever really worked because of security restrictions by GitHub. But this change should fix it, see here: https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#pull_request_target

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
Actually I'm just merging this because there's no way to test this anyway and it has no other impact whatsoever

PR #6082, opened 2024-06-11
BBH pipeline: find horizons in initial data, write masses & spins to file

nilsvu:
## Proposed changes

Automatically runs the horizon finder over initial data and writes horizon quantities to the reductions file. This makes important quantities like horizon masses and spins available to later steps in the pipeline.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/FindHorizon.py
+++ b/support/Pipelines/Bbh/FindHorizon.py
@@ -272,7 +272,7 @@ def find_horizon(
 def find_horizon_command(l_max, initial_radius, center, vars, **kwargs):
     """Find an apparent horizon in volume data."""
     initial_guess = Strahlkorper(
-        l_max=l_max, m_max=l_max, radius=initial_radius, center=center
+        l_max=l_max, radius=initial_radius, center=center
     )
     horizon, quantities = find_horizon(
         initial_guess=initial_guess,
--- a/support/Pipelines/Bbh/FindHorizon.py
+++ b/support/Pipelines/Bbh/FindHorizon.py
@@ -50,6 +50,26 @@ def _strahlkorper_vol_data(strahlkorper):
     ]
 
 
+def _horizon_reduction_data(quantities):
+    """Row of horizon quantities, can be written to an H5 dat file."""
+    # Flatten vector quantities into a single row of values. So the spin is
+    # returned as 3 columns with the x, y, and z components.
+    legend = []
+    reduction_data = []
+    for key, value in quantities.items():
+        if isinstance(value, float):
+            legend.append(key)
+            reduction_data.append(value)
+        elif len(value) == 3:
+            legend.extend([f"{key}_{xyz}" for xyz in "xyz"])
+            reduction_data.extend(value)
+        else:
+            raise ValueError(
+                f"Unsupported shape of horizon quantity {key}: {value}"
+            )
+    return legend, reduction_data
+
+
 def find_horizon(
     h5_files: Union[str, Sequence[str]],
     subfile_name: str,
@@ -57,9 +77,11 @@ def find_horizon(
     obs_time: float,
     initial_guess: Strahlkorper,
     fast_flow: Optional[FastFlow] = None,
-    output: Optional[str] = None,
+    output_surfaces_file: Optional[Union[str, Path]] = None,
     output_coeffs_subfile: Optional[str] = None,
     output_coords_subfile: Optional[str] = None,
+    output_reductions_file: Optional[Union[str, Path]] = None,
+    output_quantities_subfile: Optional[str] = None,
     output_l_max: Optional[int] = None,
     tensor_names: Optional[Sequence[str]] = None,
 ):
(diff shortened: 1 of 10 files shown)


support/Pipelines/Bbh/InitialData.py
```diff
@@ -200,6 +205,79 @@ def generate_id(
...
+    for object_label, xcoord, excision_radius in zip(
+        ["AhA", "AhB"], [x_A, x_B], [excision_radius_A, excision_radius_B]
+    ):
+        find_horizon(
+            id_volfiles,
+            subfile_name="VolumeData",
+            obs_id=obs_id,
+            obs_time=0.0,
+            initial_guess=Strahlkorper(
+                l_max=12,
```
nilsdeppe: I don't think we should expose the m_max option. Everything goes unstable if m_max < l_max. Probably the right thing to do is to just not expose the constructor from C++.

nilsdeppe (changes requested):
I only glanced at this, so don't count it as a review...

src/IO/H5/Python/File.cpp
```diff
@@ -6,6 +6,7 @@
 #include <boost/algorithm/string/join.hpp>
 #include <pybind11/pybind11.h>
 #include <pybind11/stl.h>
+#include <pybind11/stl/filesystem.h>
```
knelli2: Hmm. It's not finding this file. Is it dependent on the pybind11 version?

support/Pipelines/Bbh/FindHorizon.py
```diff
@@ -50,16 +50,34 @@ def _strahlkorper_vol_data(strahlkorper):
...
+    # Flatten vector quantities into a single row of values. So the spin is
+    # returned as 3 columns with the x, y, and z components.
+    legend = []
+    reduction_data = []
+    for key, value in quantities.items():
+        if isinstance(value, float):
+            legend.append(key)
+            reduction_data.append(value)
+        else:
+            legend.extend([f"{key}_{i}" for i in range(len(value))])
```
knelli2: `xyz` instead of `012`? I think our tensors do `xyz` suffixes.

support/Pipelines/Bbh/FindHorizon.py
```diff
@@ -78,17 +96,24 @@ def find_horizon(
...
+        specified, or both.
+      output_coeffs_subfile: Optional. Name of the subfile in the
+        'output_surfaces_file' where the horizon Ylm coefficients will be
+        written. These can be used to reconstruct the horizon, e.g. to
+        initialize excisions in domains.
+      output_coords_subfile: Optional. Name of the subfile in the
+        'output_surfaces_file' where the horizon coordinates will be written.
         These can be used for visualization.
+      output_reductions_file: Optional. H5 output file where the reduction
+        quantities on the horizon will be written, e.g. masses and spins.
```
knelli2, line 120: I think we should keep all the horizon stuff in the same H5 file. It reduces the number of files on disk and that we have to keep track of. (like forwarding options through pipeline commands)

support/Pipelines/Bbh/InitialData.py
```diff
@@ -200,6 +205,79 @@ def generate_id(
...
+        _, id_input_file = yaml.safe_load_all(open_input_file)
+    x_B, x_A = id_input_file["Background"]["Binary"]["XCoords"]
+    id_domain = id_input_file["DomainCreator"]["BinaryCompactObject"]
+    excision_radius_A = id_domain[f"ObjectA"]["InnerRadius"]
+    excision_radius_B = id_domain[f"ObjectB"]["InnerRadius"]
+
+    # Find latest observation in output data
+    if id_run_dir is None:
+        id_run_dir = Path(id_input_file_path).resolve().parent
+    id_volfiles = glob.glob(str(Path(id_run_dir) / "BbhVolume*.h5"))
```
knelli2: I think you should try to find the subfile name from the input file like in #5957 so that we keep the ability to run the pipeline steps separately.

support/Pipelines/Bbh/InitialData.py
```diff
@@ -200,6 +205,79 @@ def generate_id(
...
+    obs_id, _ = select_observation(
+        open_volfiles(id_volfiles, "VolumeData"), step=-1
```
knelli2: I'm pretty sure it's guaranteed to always be the last observation, but is that true?

support/Pipelines/Bbh/InitialData.py
knelli2: Yeah probably not, but you're right, this is more an issue with the c++ binding. I think we can do this in a follow-up PR.

We should probably make LMax an option (and default it to something like 12). This won't matter as much once the horizon finder is adaptive, but for now would be nice to have the flexibility.

support/Pipelines/Bbh/InitialData.py
```diff
@@ -200,6 +205,79 @@ def generate_id(
     )
 
 
+def postprocess_id(
```
knelli2: If this is where the ID root find is going to go, thoughts on putting this (and the CLI command) in a separate python file just so things aren't too cluttered? Do we have another example of having two CLIs in the same file?

support/Pipelines/Bbh/InitialData.py
```diff
@@ -369,5 +447,52 @@ def generate_id_command(
     )
 
 
+@click.command(name="postprocess-id", help=postprocess_id.__doc__)
+@click.argument(
+    "id_input_file_path",
+    type=click.Path(
```
knelli2: Help text

support/Pipelines/Bbh/InitialData.yaml
```diff
@@ -5,19 +5,18 @@ Executable: SolveXcts
...
 Next:
-  Run: spectre.Pipelines.Bbh.Inspiral:start_inspiral
+  Run: spectre.Pipelines.Bbh.InitialData:postprocess_id
   With:
     id_input_file_path: __file__
     id_run_dir: ./
-    pipeline_dir: {{ pipeline_dir }}
-    continue_with_ringdown: True
+    pipeline_dir: {{ pipeline_dir | default("None") }}
+    evolve: {{ evolve }}
```
knelli2, line 14: Does this now not have the ability to choose whether to continue with ringdown? I see in `InitialData.py` that `continue_with_ringdown` is always True.

knelli2 (changes requested):

support/Pipelines/Bbh/InitialData.py
nilsvu: I switched the binding to the constructor without `m_max` and added an `l_max` option.

src/IO/H5/Python/File.cpp
nilsvu: I moved this commit to #6090

support/Pipelines/Bbh/FindHorizon.py
nilsvu: The evolutions write to a separate surfaces file though. Don't you want to be consistent with that? For reading in the data I think you can't assume that both the masses+spins and the shape coeffs are in the same file anyway, because they are so different (doubles vs essentially volume data).

Edit: options that I can think of are:
- masses+spins in reductions file, coeffs in new surfaces file
- masses, spins, and coeffs in new horizons file that has no evolution analog

support/Pipelines/Bbh/InitialData.py
nilsvu: Yea that should be fine

support/Pipelines/Bbh/InitialData.py
nilsvu: Ok yea a separate file is probably better. I'll do that when squashing.

support/Pipelines/Bbh/InitialData.py
nilsvu: Positional arguments don't have a `help`, it's part of the main help text. Edited that.

support/Pipelines/Bbh/InitialData.yaml
nilsvu: If you run with `--evolve` it will also just try to continue with ringdown (which isn't working yet). That's the same behavior as before. I could disable that since it's not working yet?

The `start-inspiral` command still has the option whether or not to continue with ringdown.

support/Pipelines/Bbh/FindHorizon.py
knelli2: I like the fixup you did for the postprocess step just writing things into a single `Horizons.h5` file. In the actual horizon find CLI, having the option for separate files is fine.

knelli2 (commented):
Fixups look good. You can squash. We probably have to wait until #6090 is merged first?

nilsdeppe (dismissed):
I don't need to look at the rest. Thanks for making the change to the bindings!

nilsvu:
@knelli2 no I decoupled this from #6090. This is ready when you are :)

knelli2 (approved):

PR #6083, opened 2024-06-12
Add DgElement collection component and CreateElementCollection action

nilsdeppe:
## Proposed changes

This is the last bit of code for GTS nodegroup with no functions of time. I'll submit another PR soon to add support for functions of time. I still need to finish support for LTS.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/ArrayCollection/CreateElementCollection.hpp
+++ b/src/Parallel/ArrayCollection/CreateElementCollection.hpp
@@ -1,0 +1,201 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <optional>
+#include <tuple>
+#include <unordered_set>
+#include <utility>
+#include <vector>
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "Domain/Block.hpp"
+#include "Domain/Creators/Tags/Domain.hpp"
+#include "Domain/Creators/Tags/InitialExtents.hpp"
+#include "Domain/Creators/Tags/InitialRefinementLevels.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Domain/Tags/ElementDistribution.hpp"
+#include "Evolution/DiscontinuousGalerkin/Initialization/QuadratureTag.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/ArrayCollection/SpawnInitializeElementsInCollection.hpp"
+#include "Parallel/ArrayCollection/Tags/ElementCollection.hpp"
+#include "Parallel/ArrayCollection/Tags/ElementLocations.hpp"
+#include "Parallel/ArrayCollection/Tags/NumberOfElementsTerminated.hpp"
+#include "Parallel/CreateElementsUsingDistribution.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Parallel/Info.hpp"
+#include "Parallel/Local.hpp"
+#include "Parallel/NodeLock.hpp"
+#include "Parallel/Reduction.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/Functional.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace Parallel::Actions {
+/*!
+ * \brief Creates the `DgElementArrayMember`s on the (node)group component.
+ *
+ * First the distribution of elements is computed using
+ * `Parallel::create_elements_using_distribution()`, they are inserted on
+ * each (node)group element. A reduction is done over the (node)group before
+ * initializing the `DgElementArrayMember`s themselves, since they are allowed
+ * to communicate with each other during their initialization. The reduction
+ * target is `Parallel::Actions::SpawnInitializeElementsInCollection`.
+ *
+ * Uses:
+ * - DataBox:
+ *   - `domain::Tags::Domain<Dim>`
+ *   - `domain::Tags::InitialRefinementLevels<Dim>`
+ *   - `domain::Tags::InitialExtents<Dim>`
+ *   - `evolution::dg::Tags::Quadrature`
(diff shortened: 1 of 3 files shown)


src/Parallel/ArrayCollection/CreateElementCollection.hpp
```diff
@@ -0,0 +1,201 @@
...
+#include "Utilities/Functional.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+namespace Parallel::Actions {
+/*!
+ * \brief Creates the `DgElementArrayMember`s on the (node)group component.
+ *
+ * First the distribution of elements is computed using
+ * `Parallel::create_elements_using_distribution()`, they they are inserted on
```
knelli2: `they they` -> `they`

src/Parallel/ArrayCollection/CreateElementCollection.hpp
```diff
@@ -0,0 +1,201 @@
...
+        for (const size_t ref_lev : initial_ref_levs) {
+          result += two_to_the(ref_lev);
+        }
+      }
+      return result;
+    }();
+    std::vector<std::pair<ElementId<Dim>, size_t>> my_elements_and_cores{};
+    my_elements_and_cores.reserve(
+        static_cast<size_t>(std::round(static_cast<double>(total_num_elements) /
+                                       static_cast<double>(number_of_nodes))));
```
knelli2: Could this be just `total_num_elements / number_of_nodes + 1`? It's just a reservation so it's ok if it's a bit bigger.

src/Parallel/ArrayCollection/CreateElementCollection.hpp
```diff
@@ -0,0 +1,201 @@
...
+            const ElementId<Dim>& element_id, const size_t target_proc,
+            const size_t target_node) {
+          node_of_elements.insert(std::pair{element_id, target_node});
+          if (target_node == my_node) {
+            my_elements_and_cores.push_back(std::pair{element_id, target_proc});
+          }
+        },
+        element_weight, blocks, initial_extents, initial_refinement_levels,
+        quadrature,
+
```
knelli2: Remove blank line

knelli2 (changes requested):
Also some clang-tidy stuff about structured bindings in lambdas.

@kidder Do you want to look at this?

src/Parallel/ArrayCollection/CreateElementCollection.hpp
nilsdeppe: I personally find the long functions easier to parse when there's a logical split between arguments. In this case everything below is related to parallelization/hardware and the above to the domain.

src/Parallel/ArrayCollection/CreateElementCollection.hpp
nilsdeppe: Sure :shrug:

nilsdeppe:
I've rebased and squashed in the couple changes. I don't know what to do about the clang-tidy errors because I think this is an issue with clang-tidy not compiling the code correctly. I can build the code fine and what clang-tidy is reporting is a compilation error. With C++17 or 20 structured bindings are variables that can be captured in lambdas, whereas prior to that there were local names and thus could not be captured. I think the issue has something to do with how clang-tidy is being used. It's either that we are using an older version or that the standard flag is not being handled correctly in the header file context.

src/Parallel/ArrayCollection/CreateElementCollection.hpp
knelli2: Could you add a comment then instead of a blank line? We do this in some other places and I think it's clearer (a blank line looks like a formatter error to me)

knelli2 (changes requested):
According to the comments of the accepted answer [here](https://stackoverflow.com/questions/46114214/lambda-implicit-capture-fails-with-variable-declared-from-structured-binding), even though the feature is supported in c++17/20, apparently clang (through major version 15) still didn't support this. And I just checked, in the container we use a `clang-tidy` version of 14, so that'd explain why we are getting the errors. Looks like we should update the clang-tidy version we use, but that's beyond this PR.

src/Parallel/ArrayCollection/CreateElementCollection.hpp
```diff
@@ -0,0 +1,199 @@
...
+ * DataBox changes:
+ * - Adds:
+ *   - `Parallel::Tags::ElementCollection`
+ *   - `Parallel::Tags::ElementLocations<Dim>`
+ *   - `Parallel::Tags::NumberOfElementsTerminated`
+ * - Removes: nothing
+ * - Modifies:
+ *   - `Parallel::Tags::ElementCollection`
+ *   - `Parallel::Tags::ElementLocations<Dim>`
+ */
```
kidder, line 68: modifies NumberOfElementsTerminated as well

src/Parallel/ArrayCollection/DgElementCollection.hpp
```diff
@@ -0,0 +1,117 @@
...
+      Parallel::Tags::ElementCollection<Dim, Metavariables, PhaseDepActionList,
+                                        simple_tags_from_options>;
+
+  /// @{
+  /// \brief Returns the derived `Parallel::DgElementArrayMember` from the
+  /// `Parallel::DgElementArrayMemberBase`
+  auto downcast(const Parallel::DgElementArrayMemberBase<Dim>& base)
+      -> const dg_element_array_member& {
+#ifdef SPECTRE_DEBUG
+    return dynamic_cast<const dg_element_array_member&>(base);
```
kidder: what is the reason for doing the cast in this manner?

kidder (changes requested):

src/Parallel/ArrayCollection/DgElementCollection.hpp
nilsdeppe: I haven't needed these (thought I would) so just removing them.

nilsdeppe:
I've rebased and squashed in fixes. Let me know if there's anything else!

kidder (approved):

knelli2 (approved):

knelli2:
Ignoring macOS timeout and outdated clang-tidy errors

PR #6084, opened 2024-06-12
Add threaded callbacks for functions of times

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/Callback.hpp
+++ b/src/Parallel/Callback.hpp
@@ -42,7 +42,7 @@ class SimpleActionCallback : public Callback {
   // NOLINTNEXTLINE(google-explicit-constructor)
   SimpleActionCallback(Proxy proxy, std::decay_t<Args>... args)
       : proxy_(proxy), args_(std::move(args)...) {}
-  SimpleActionCallback(CkMigrateMessage* msg) : Callback(msg) {}
+  explicit SimpleActionCallback(CkMigrateMessage* msg) : Callback(msg) {}
   using PUP::able::register_constructor;
   void invoke() override {
     std::apply(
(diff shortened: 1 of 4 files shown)


src/Parallel/Callback.hpp
```diff
@@ -97,6 +97,71 @@ class SimpleActionCallback<SimpleAction, Proxy> : public Callback {
   std::decay_t<Proxy> proxy_{};
 };
 
+/// Wraps a call to a threaded action and its arguments.
+/// Can be invoked only once.
+template <typename ThreadedAction, typename Proxy, typename... Args>
+class ThreadedActionCallback : public Callback {
```
knelli2, line 103: We have a test for these in `Unit/Parallel/Test_Callback.cpp`. Could you add some there? I know it's a bit more work, but just so we are guaranteed that they are working

src/ParallelAlgorithms/Actions/FunctionsOfTimeAreReady.hpp
```diff
@@ -143,6 +143,32 @@ bool functions_of_time_are_ready_simple_action_callback(
...
 
+/// \ingroup ComputationalDomainGroup
+/// Check that functions of time are up-to-date.
+///
+/// Check that functions of time in \p CacheTag with names in \p
+/// functions_to_check are ready at time \p time.  If  \p functions_to_check is
+/// a `std::nullopt`, checks all functions in \p CacheTag.  If any function is
+/// not ready, schedules a `Parallel::ThreadedActionCallback` with the
+/// GlobalCache which calls the threaded action passed in as a template
+/// parameter. The `Args` are moved into the callback.
```
knelli2: They're forwarded, not moved. There's a difference, right?

knelli2 (changes requested):
Just a couple things

src/ParallelAlgorithms/Actions/FunctionsOfTimeAreReady.hpp
nilsdeppe: Yep, I'll also update the simple action variant to say forwarded as well!

src/Parallel/Callback.hpp
nilsdeppe: Not a problem! Done!

knelli2 (approved):

PR #6087, opened 2024-06-12
Clarify PlotCce modes option

### LINKED ISSUE #6085: spectre plot slice usability improvements
@ctrandrsn and I ran into some confusion when trying to use this feature. Specifically,

- [x] `spectre plot slice ` happily runs with zero h5 files. It should give an error if you give no files.
- [x] If you don't specify a variable to plot it just lists all of them without actually telling you why it's listing them. Having an error would be helpful.
- [ ] `--slice-extent` is `X Y` while everything else is `X,Y`. Having this be consistent would be nice.
- [x] `--subfile-name` seems to have a default subfile but this isn't document. The HDF5 file we ran on only has 1 subfile so I'm not sure if that's why it could do it. In any case, this should be documented in the help text.
- [x] It seems like `--slice-origin` and the other `--slice` options are all required because if we didn't specify them all we get an error that they must be specified. Could these be marked as required in the help text?
- [x] `--var` also isn't marked as required but is.

### LINKED ISSUE #6086: plot cce CLI improvements
@ctrandrsn and I ran into some confusion with what the format for the `l,m` modes should be. Could an example be given in the help text, e.g. that it says `--modes 2,2`? The other advantage is that this gives people very new to waveforms a useful mode to look at since they might instead try to look at the 1,1 mode or something.

knelli2:
## Proposed changes

Fixes #6086 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/PlotCce.py
+++ b/src/Visualization/Python/PlotCce.py
@@ -46,9 +46,9 @@ def _parse_modes(ctx, param, all_modes):
     callback=_parse_modes,
     required=True,
     help=(
-        "Which mode to plot. Specified as 'l,m'. Will plot both real and"
-        " imaginary components unless '--real' or '--imag' are specified. Can"
-        " be specified multiple times."
+        "Which mode to plot. Specified as 'l,m' (e.g. '--modes 2,2'). Will plot"
+        " both real and imaginary components unless '--real' or '--imag' are"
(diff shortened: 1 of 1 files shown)


nilsdeppe (approved):
LGTM, thank you!!

ctrandrsn (approved):

PR #6088, opened 2024-06-12
Improve some CLI errors and docs

### LINKED ISSUE #6085: spectre plot slice usability improvements

nilsvu:
## Proposed changes

Fixes #6085.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/DevGuide/PythonBindings.md
+++ b/docs/DevGuide/PythonBindings.md
@@ -187,11 +187,12 @@ See \ref spectre_using_python "Using SpECTRE's Python"
   subdirectory structures. Passing many files to a script is easy for the user
   by using a glob (note: don't take the glob as a string argument, take the
   expanded list of files directly using `click.argument(..., nargs=-1,
-  type=click.Path(...))`).
+  type=click.Path(...), required=True)`).
+  Note that Click recommends to avoid `required=True` here but to [gracefully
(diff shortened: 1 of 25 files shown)


nilsdeppe (dismissed):
This is awesome! Thank you for doing this so quickly! 🚀

nilsvu:
I extended the last commit a bit further with more error message that benefit from the new `RequiredChoicesError`.

nilsdeppe:
Uh oh, looks like CI is now hitting a check you added :)

nilsvu:
Fixed

knelli2 (approved):

nilsdeppe (approved):

PR #6091, opened 2024-06-13
Change an ASSERT to ERROR in horizon finder

nilsvu:
This can happen dynamically in a simulation (not necessarily a programming error), and just crashed with a
std::bad_optional_access without a backtrace.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/StrahlkorperTransformations.cpp
+++ b/src/Domain/StrahlkorperTransformations.cpp
@@ -25,7 +25,6 @@
 #include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
-#include "Utilities/ErrorHandling/Assert.hpp"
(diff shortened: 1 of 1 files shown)


src/Domain/StrahlkorperTransformations.cpp
```diff
@@ -62,26 +61,26 @@ void coords_to_different_frame(
 
     // If this doesn't have a value, then the point isn't in the domain which is
     // really bad.
-    ASSERT(block_logical_coords[s].has_value(),
-           "Source coordinates are not in the domain.");
+    if (not block_logical_coords[s].has_value()) {
+      ERROR("A point on the Strahlkorper in the "
```
knelli2: Since this will be run over every point of the Strahlkorper, maybe we should wrap all these `if`s in `UNLIKELY` so the compiler can do some optimizations

knelli2 (changes requested):

src/Domain/StrahlkorperTransformations.cpp
nilsvu: I don't know if `UNLIKELY` actually does something, but I can add it

knelli2 (approved):

PR #6096, opened 2024-06-17
Add TaggedVariant, parsable similarly to a factory

wthrowe:
I have mixed feelings on the `visit` interface.  The version here follows `std::visit` and allows multiple TaggedVariants to be passed to a single call, but this makes reporting the tag awkward and, in particular, prevents implicit conversions between reference types in the visitor arguments.  An alternative would be to restrict to the (I assume vastly most common) case of a single TaggedVariant and pass the tag and value as separate arguments, which would fix all the conversion trouble.  Opinions are welcome.

Ping @knelli2 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/TaggedVariant.hpp
+++ b/src/DataStructures/TaggedVariant.hpp
@@ -1,0 +1,448 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <functional>
+#include <type_traits>
+#include <utility>
+#include <variant>
+
+#include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Requires.hpp"
+#include "Utilities/Serialization/PupStlCpp17.hpp"
+#include "Utilities/TMPL.hpp"
+
+/// \cond
+namespace PUP {
+class er;
+}  // namespace PUP
+/// \endcond
+
+/// \ingroup DataStructuresGroup
+///
+/// TaggedVariant and related functionality.
+namespace variants {
+namespace TaggedVariant_detail {
+template <typename R, typename Variant, typename Tag, typename Visitor>
+struct VisitAlternative;
+}  // namespace TaggedVariant_detail
+
+/// \cond
+template <typename... Tags>
+class TaggedVariant;
+template <typename... Tags>
+constexpr bool operator==(const TaggedVariant<Tags...>& a,
+                          const TaggedVariant<Tags...>& b);
+template <typename... Tags>
+constexpr bool operator<(const TaggedVariant<Tags...>& a,
+                         const TaggedVariant<Tags...>& b);
+/// \endcond
+
+/// \ingroup DataStructuresGroup
+///
+/// A class similar to `std::variant`, but indexed by tag structs.
+///
+/// \see variants::get, variants::get_if, variants::holds_alternative,
+/// variants::visit
+template <typename... Tags>
+class TaggedVariant {
+ private:
+  static_assert(sizeof...(Tags) > 0);
+  static_assert(
+      std::is_same_v<tmpl::remove_duplicates<TaggedVariant>, TaggedVariant>,
+      "TaggedVariant cannot have duplicate tags.");
+
+  template <typename Tag>
+  static constexpr size_t data_index =
+      tmpl::index_of<TaggedVariant, Tag>::value;
+
+ public:
+  /// A default constructed instance has the first tag active.
+  TaggedVariant() = default;
+  TaggedVariant(const TaggedVariant&) = default;
+  TaggedVariant(TaggedVariant&&) = default;
+  TaggedVariant& operator=(const TaggedVariant&) = default;
+  TaggedVariant& operator=(TaggedVariant&&) = default;
+  ~TaggedVariant() = default;
+
+  /// Construct with \p Tag active, using \p args to construct the
+  /// contained object.
+  ///
+  /// \snippet DataStructures/Test_TaggedVariant.cpp construct in_place_type
+  template <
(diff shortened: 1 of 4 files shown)


nilsdeppe:
For what it's worth, I added multi-container `get`s https://github.com/sxs-collaboration/spectre/blob/develop/src/DataStructures/TaggedContainers.hpp at some point because, while not the most common case, we did need it

wthrowe:
Apparently the cppreference compiler support tables are wrong or incomplete, so I'll have to figure out what compiler versions constexpr support works in.  Clang 17 and gcc 13 are known working.  gcc 9 was expected to fail, so it does runtime tests instead of static assertions, so I'll switch other versions to that once I figure out what they are.

wthrowe:
The first working gcc version is 12.  I can't easily test clang because clang 15 can't compile DomainCreators and my distro's clang 16 package won't install for some reason I don't feel like debugging.  I'm guessing the changes happened at the same time as constexpr std::vector, which was gcc 12 and clang 15, so that's what I went with there.

src/DataStructures/TaggedVariant.hpp
```diff
@@ -0,0 +1,464 @@
...
+}  // namespace PUP
+/// \endcond
+
+/// \ingroup DataStructuresGroup
+///
+/// TaggedVariant and related functionality.
+namespace variants {
+/// \cond
+template <typename... Tags>
+class TaggedVariant;
```
knelli2: Why the forward declaration? Doesn't look like it's needed

src/DataStructures/TaggedVariant.hpp
```diff
@@ -0,0 +1,464 @@
...
+                b));
+}
+template <typename... Tags>
+constexpr bool operator!=(const TaggedVariant<Tags...>& a,
+                          const TaggedVariant<Tags...>& b) {
+  return not(a == b);
+}
+template <typename... Tags>
+constexpr bool operator<(const TaggedVariant<Tags...>& a,
+                         const TaggedVariant<Tags...>& b) {
```
knelli2, line 391: Just for my understanding, is it common in the standard to implement these comparison operators for a container?

src/DataStructures/TaggedVariant.hpp
```diff
@@ -136,6 +141,15 @@ class TaggedVariant {
 
   void pup(PUP::er& p) { p | data_; }
 
+  /// A TaggedVariant over option tags can be parsed as any of them.
+  /// @{
+  static constexpr Options::String help = "One of multiple options";
```
knelli2, line 152: Is there a way that we can specify this string? Possibly from a template parameter somehow?

I've tested this PR out on some domain creator FoT changes I'm making, and it works really well! I think this will make input files much clearer. The only thing is that when a typo is made in the input file, this help string isn't very helpful. E.g. this was (part of) the error message I got

```
In string:
  While parsing option TestCreationOpt:
  While creating a TranslationMap:
  While parsing option InitialValues:
  While creating a TaggedVariant:
  At line 3 column 5:
  Cannot decide between alternative options.
  
==== Parsing the option string:
  FromVolFil:
    H5Filename: BbhVolume0.h5
    SubfileName: VolumeData
    Time: 0.0
  
==== Description of expected options:
  One of multiple options
  
Options:
    EITHER
      Explicit:
        Something useful here
  
  OR
      FromVolFile:
        Something useful here

```

If under the `Description of expected options` part I was able to use the help string for the actual option this was creating, that'd be much clearer I think (i.e. `Initial values for the translation map, its velocity and acceleration.`)

knelli2 (changes requested):
Thanks for adding this Will! I think this will work very well. As for the `visit` interface. I think you can restrict it to only acting on a single `TaggedVariant`. I imagine this will basically only be used for option stuff in which case we'll only need to visit the tags in one variant. But I don't have super strong feelings either way.

src/DataStructures/TaggedVariant.hpp
wthrowe: Left over from an old attempt.  Will remove.

src/DataStructures/TaggedVariant.hpp
wthrowe: Yes, and now that I think of it these should just forward to the `std::variant` ones.

src/DataStructures/TaggedVariant.hpp
wthrowe: There's nowhere to put a template parameter, unfortunately.

What might be good in general is if there was an error when creating a class it also displayed the help from the tag, but currently the code generating the error message doesn't know that information.

We could add the help test for each option after the "While parsing option" lines in the "backtrace" pretty easily.  It's not the most visible place, but at least it would be there.  Example from ExportCoordinates:
```
In /home/wthrowe/spectre/tests/InputFiles/ExportCoordinates/Input1D.yaml
While parsing option DomainCreator
  The domain to create initially
While creating a unique_ptr
While operating factory for DomainCreator
While creating a Interval
While parsing option TimeDependence
  The time dependence of the moving mesh domain.
While creating a unique_ptr
While operating factory for TimeDependence
While creating a UniformTranslation
At line 48 column 9:
Option 'VelocityX' is not a valid option.

==== Parsing the option string:
InitialTime: 0.0
VelocityX: [0.5]

==== Description of expected options:
A spatially uniform translation initialized with a constant velocity.

Options:
  InitialTime:
    type=double
    The initial time of the functions of time

  Velocity:
    type=[double x1]
    The velocity of the map.
```

knelli2: @wthrowe I'm ok forgoing this change for the time being. I think the benefit to the input file syntax is enough for the moment.

knelli2 (approved):
Fixup looks good. Go ahead and squash

PR #6109, opened 2024-06-24
Add FromVolumeFile time dependent option

knelli2:
## Proposed changes

First in a series of PRs that revamp the time dependent options of the Sphere and BCO domain creators.

Adds an option (not used yet) that can read functions of time that have been serialized in the volume data of an H5 file. The most recent commit of this PR contains the bulk of the changes. The other commits are just some organizational ones and don't contain any real changes.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #6113, opened 2024-06-25
Add several common time dependent options

knelli2:
## Proposed changes

Second PR in the series to revamp the time dependent options of the Sphere and BCO domain creators.

This PR adds common options for the
- Rotation map
- Expansion map
- Translation map

These common options were designed so that different function of time types (`PiecewisePolynomial` or `SettleToConst`) could be chosen at runtime.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #6109~

### DIFF
--- a/src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
+++ b/src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
@@ -1,0 +1,108 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <array>
+#include <cstddef>
+#include <optional>
+#include <string>
+#include <variant>
+
+#include "DataStructures/DataVector.hpp"
+#include "Domain/Creators/TimeDependentOptions/FromVolumeFile.hpp"
+#include "Options/Auto.hpp"
+#include "Options/Context.hpp"
+#include "Options/String.hpp"
+#include "Utilities/TMPL.hpp"
+
+namespace domain::creators::time_dependent_options {
+/*!
+ * \brief Class to be used as an option for initializing expansion map
+ * coefficients.
+ */
+struct ExpansionMapOptions {
+  using type = Options::Auto<ExpansionMapOptions, Options::AutoLabel::None>;
+  static std::string name() { return "ExpansionMap"; }
+  static constexpr Options::String help = {
+      "Options for a time-dependent expansion of the coordinates. Specify "
+      "'None' to not use this map."};
+
+  struct InitialValues {
+    using type =
+        std::variant<std::array<double, 3>, FromVolumeFile<names::Expansion>>;
+    static constexpr Options::String help = {
+        "Initial values for the expansion map, its velocity and "
+        "acceleration."};
+  };
+
+  struct InitialValuesOuterBoundary {
+    using type =
+        std::variant<std::array<double, 3>, FromVolumeFile<names::Expansion>>;
+    static constexpr Options::String help = {
+        "Initial values for the expansion map, its velocity and "
+        "acceleration at the outer boundary. Unless you are starting from a "
+        "checkpoint or continuing an evolution, this option should likely be "
+        "[1.0, 0.0, 0.0] at the start of an evolution"};
+  };
+
+  struct DecayTimescaleOuterBoundary {
+    using type = Options::Auto<double>;
+    static constexpr Options::String help = {
+        "A timescale for how fast the outer boundary expansion approaches its "
+        "asymptotic value. Can optionally specify 'Auto' when reading the "
+        "initial values 'FromVolumeFile' to use the decay timescale from the "
+        "function of time in the volume file. Cannot specify 'Auto' when "
+        "initial values are specified directly."};
+  };
+
+  struct DecayTimescale {
+    using type = Options::Auto<double>;
+    static constexpr Options::String help = {
+        "If specified, a SettleToConstant function of time will be used for "
+        "the expansion map and this number will determine the timescale that "
+        "the expansion approaches its asymptotic value. If 'Auto' is "
+        "specified, a PiecewisePolynomial function of time will be used for "
+        "the expansion map. Note that if you are reading the initial values "
+        "from a volume file, you must specify 'Auto' for this option."};
+  };
+
+  struct AsymptoticVelocityOuterBoundary {
+    using type = Options::Auto<double>;
+    static constexpr Options::String help = {
+        "There are two choices for this option. If a value is specified, a "
+        "FixedSpeedCubic function of time will be used for the expansion map "
+        "at the outer boundary and this number will determine its velocity. If "
+        "'Auto' is specified, the behavior will depend on what is chosen for "
+        "'InitialValuesOuterBoundary'. If values are specified for "
+        "'InitialValuesOuterBoundary', then 'Auto' here means a "
+        "SettleToConstant function of time will be used for the expansion map "
+        "at the outer boundary. If 'FromVolumeFile' is specified for "
+        "'InitialValuesOuterBoundary', then a FixedSpeedCubic function of time "
+        "will be used and the velocity from the function of "
+        "time in the volume file will be used."};
+  };
+
+  using options = tmpl::list<InitialValues, InitialValuesOuterBoundary,
+                             DecayTimescaleOuterBoundary, DecayTimescale,
+                             AsymptoticVelocityOuterBoundary>;
+
+  ExpansionMapOptions() = default;
+  ExpansionMapOptions(
+      const std::variant<std::array<double, 3>,
+                         FromVolumeFile<names::Expansion>>& expansion_values,
+      const std::variant<std::array<double, 3>,
+                         FromVolumeFile<names::Expansion>>&
+          expansion_outer_boundary_values,
+      std::optional<double> decay_timescale_outer_boundary_in,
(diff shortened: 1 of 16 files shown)


src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+  struct InitialValues {
+    using type =
+        std::variant<std::array<double, 3>, FromVolumeFile<names::Expansion>>;
+    static constexpr Options::String help = {
+        "Initial values for the expansion map, its velocity and "
+        "acceleration."};
+  };
+
+  struct InitialValuesOuterBoundary {
```
nilsvu, line 39: Question: why doesn't "FromVolumeFile" just copy the FoT altogether, instead of essentially recreating it from values and derivatives?

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
```diff
@@ -0,0 +1,98 @@
...
+        std::variant<std::array<double, 3>, FromVolumeFile<names::Expansion>>;
+    static constexpr Options::String help = {
+        "Initial values for the expansion map, its velocity and "
+        "acceleration at the outer boundary. Unless you are starting from a "
+        "checkpoint or continuing an evolution, this option should likely be "
+        "[1.0, 0.0, 0.0] at the start of an evolution"};
+  };
+
+  struct DecayTimescaleOuterBoundary {
+    using type = double;
```
nilsvu: Are these options not read from the vol file?

src/Domain/Creators/TimeDependentOptions/TranslationMap.hpp
```diff
@@ -0,0 +1,58 @@
...
+
+  std::array<DataVector, 3> initial_values{};
+};
+
+/*!
+ * \brief Given a `TranslationMapOptions`, return the function of time
+ * coefficients that can be used to initialize a FunctionOfTime.
+ */
+template <size_t Dim>
+std::array<DataVector, 3> initial_translation_func(
```
nilsvu: I don't see a definition of this function. Probably this function is not needed because the options class exposes the values?

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
knelli2: So originally I didn't do this because I was lazy and I just read in the last FoT in the volume file rather than finding the correct observation ID. However, now that I think about it, there's likely a real issue.

Imagine a scenario where I'm writing volume data at simulation time $t_0$. Whenever we write the volume data in human time, that's when we grab the FoTs, serialize them, and write them. There could be a case where, by the time the FoTs were written, a control system has updated the FoTs past $t_0$. This will then cause issues upon restart because the expiration times of the functions of time are all messed up and not what was expected. The control system (at the moment) expects to set all expiration times and to just have FoT values at $t_0$ exactly. So I think for the time being, I think this is how things need to work.

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
knelli2: Currently no. But I could edit the FoT to expose these values so we can copy them in. That's probably the correct thing to do here.

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
nilsvu: Oh so the FoT in the volume data can be wrong? That sounds like a bug that we have to fix. It also affects interpolation of the volume data.

knelli2: No not wrong. If you write volume data at $t_0$, then the FoTs will be valid at $t_0$, which is needed for interpolation and other stuff. That's guaranteed. It's just that the FoTs could also be valid at times after $t_0$ as well. This is because of asynchronous nature of the updates of the FoTs.

nilsvu: We currently retrieve the FoTs from the DataBox when volume data is written by the observer nodegroup, which is is not guaranteed to be close to the time the time step was taken. Is that ok?

knelli2: Yes it's fine. In order for the event to run at all the FoTs must be valid, so grabbing them later doesn't matter, so long as you're only interested in evaluating the FoTs at the simulation time of the event (like reading in/transforming volume data). But for initializing the FoTs, this isn't ok so we can't just copy the serialized FoTs, we have to call them at the initial time.

knelli2:
@nilsvu I think we'll want to get this series of PRs in so we have the flexibility of initializing the FoTs

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
```diff
@@ -0,0 +1,107 @@
...
+    static constexpr Options::String help = {
+        "A timescale for how fast the outer boundary expansion approaches its "
+        "asymptotic value. Can optionally specify 'Auto' when reading the "
+        "initial values 'FromVolumeFile' to use the decay timescale from the "
+        "function of time in the volume file. Cannot specify 'Auto' when "
+        "initial values are specified directly."};
+  };
+
+  struct DecayTimescale {
+    using type = Options::Auto<double, Options::AutoLabel::None>;
```
nilsvu: What should this option be when continuing an evolution from volume data? Generally, there are lots of choices to make for all these options, but when I continue an evolution from volume data I just want to specify "FromVolumeFile" for the whole thing, don't I?

src/Domain/Creators/TimeDependentOptions/RotationMap.hpp
```diff
@@ -0,0 +1,81 @@
...
+        "option. If you are reading the quaternion from a volume file, 'Auto' "
+        "will use the angle values from the volume file. If you are simply "
+        "specifying the quaternion and (optionally) its time derivatives, then "
+        "'Auto' here will set the angle and its time derivatives to zero. If "
+        "values are specified for the angle and its time derivatives, then "
+        "those will override anything computed from the 'InitialQuaternions' "
+        "option."};
+  };
+
+  struct DecayTimescale {
```
nilsvu, line 57: Same for this one

nilsvu (commented):
You can squash

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
knelli2: It should be `None`. Yeah the way it is now probably isn't ideal. Yes just specifying "FromVolumeFile" would be nice, except that'd basically mean moving the "FromVolumeFile" part up a level and adding a separate "ExpansionMap" class for it.

I could change this from `None` to `Auto` so it's at least the same as the other options so when you are specifying that you want to continue from volume data, you just say `Auto` for everything else.

knelli2: @nilsvu Went ahead and made this `Auto` and just added a note in the help string about it

knelli2:
@nilsvu Rebased

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
nilsdeppe: What I understood from a meeting discussion :

Currently we write the serialized functions of time at the end of the observation chain, eg in `ContributeVolumeDataToWriter` right before we call `volume_file.write_volume_data()`. This means the functions of time can have data from later than necessary in them. That is, if we write at $t_0$ and the functions of time expire at $t'>t_0$, then when we actually do the write we could have function of time data for $t>t'$ written to disk. Most of the time this doesn't matter because we only want to evaluate the functions of time at $t_0$. However, in the case where we want to use this data as a "checkpoint" (e.g. starting at a different resolution), the future state of the functions of time is no longer correct. What we need to do is essentially "drop" this future state. This PR provides one implementation for dropping that future state.

nilsvu: Ok that's fine and we can merge it. Would another possible implementation be to just copy the FoT object from the volume data and add a function that drops future data from it? If that has the same effect then it seems a lot more robust to me and avoids a lot of code in this PR (since it just uses the copy constructor). Again, if the decision is that this is more work than to maintain the new code in this PR then fine with me.

nilsdeppe: This feature was historically intentionally omitted from the FoT's because the situation described above is the only case where you should legitimately be dropping times. We figured it would be safer to not allow dropping them because doing so would almost always be a bug.

nilsvu: Ok wouldn't a function like `get_clone_until_time(t)` be a good implementation of this?

nilsvu (approved):

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
nilsdeppe: That could work, but I'd have to think about it more carefully. It does get around the issue for why we originally didn't allow a `drop()` function.

PR #6114, opened 2024-06-25
Add YlmsFromSpEC to shape time dependent options

knelli2:
## Proposed changes

This is tangentially related to the PRs revamping the time dependent options for the Sphere and BCO domain creators.

This adds the ability to read the shape coefficients from a SpEC ID horizons dat file (not a dat file within an H5 file, but an actual dat file). If we are starting from SpEC ID (which we can do), then this is an important feature to have.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments

~Depends on and includes #6109.~

PR #6115, opened 2024-06-25
Use new common time dependent options in Sphere and BCO domains

### LINKED ISSUE #5584: Read in FoT parameters from file in BCO domain creators
This would be useful to have so we could launch a new run with volume data from a previous run as numeric ID. We just need to have the proper functions of time to be able to do this, otherwise things will be discontinuous. There are two main ways I see to do this:

- Write a new option that takes in a volume file (or glob and assert the result is the same for all) and populate all the functions of time with that. Things to consider:
  - The glob is nice because then you don't have to worry about picking a specific volume file (even though 0 will probably always be there). It also has an added benefit that you're guaranteed all your volume files have the same functions of time (so long as you check that). Not strictly necessary, but could be good to have.
  - What to do when reading volume files without specific functions of time that are needed?
    - Error?
    - Set reasonable defaults? (like no rotation, or spherical shape map)
  - What to do with extra functions of time?
    - Error?
    - Ignore?
  - Option in `*BCO.hpp` files would change from the first to the second
```cpp
struct TimeDependentMaps {
  using type =
      Options::Auto<bco::TimeDependentMapOptions, Options::AutoLabel::None>;
  static constexpr Options::String help = bco::TimeDependentMapOptions::help;
};
```
```cpp
struct TimeDependentMaps {
  using type =
      Options::Auto<std::variant<bco::TimeDependentMapOptions, FromFile>,
                    Options::AutoLabel::None>;
  static constexpr Options::String help = bco::TimeDependentMapOptions::help;
};
```
- Same as above, except we do it on a FoT by FoT basis instead of just all at once.
  - More flexibility. Don't have to worry about missing/extra FoTs because it's up to the user to specify which ones should be read in from file.
  - More verbose code/input file than the first option. Would have to change each individual FoT option in `TimeDependentMapOptions` in `BCOHelpers.hpp`

knelli2:
## Proposed changes

Final PR in the series that revamps the time dependent options for the Sphere and BCO domains.

This actually adds all the new common options to the domain creators.

Fixes #5584.

### Upgrade instructions

The options for the Expansion and Rotation map in the BinaryCompactObject and Sphere domains have changed slightly. In the BinaryCompactObject, they are now
```yaml
    TimeDependentMaps:
      ...
      ExpansionMap:
        InitialValues: [1.0, -4.6148457646200002e-05, 0.0] # <-- Must specify all derivs
        AsymptoticVelocityOuterBoundary: -1.0e-6 # <-- Same as before
        DecayTimescaleOuterBoundary: 50.0 # <-- Different name, same value as before
```

In the Sphere domain, you can optionally specify the same options as the BinaryCompactObject like above to use functions of time that expire. Additionally, you can specify options to use SettleToConstant functions of time in the Sphere domain with the options below
```yaml
    TimeDependentMaps:
      ...
      ExpansionMap:
        InitialValues: [1.0, -4.6148457646200002e-05, 0.0] # <-- Must specify all derivs
        InitialValuesOuterBoundary: [1.0, 0.0, 0.0] # <-- Must specify all derivs
        DecayTimescale: 52.0 # <-- Different name
        DecayTimescaleOuterBoundary: 50.0 # <-- Different name
      RotationMap:
        InitialQuaternions: [[1.0, 0.0, 0.0, 0.0]] # <-- Need quat. Can optionally specify derivs
        DecayTimescale: 50 # <-- Different name
```

Additionally, for any of the time dependent map options, you can choose the function of time from a volume H5 file by specifying the following options (not just for Expansion, but for Rotation and Translation as well and shape is slightly different)

```yaml
    TimeDependentMaps:
      ...
      ExpansionMap:
        H5Filename: BbhVolume0.h5
        SubfileName: VolumeData
      ShapeMapA:
        H5Filename: BbhVolume0.h5
        SubfileName: VolumeData
        TransitionEndsAtCube: true
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #6109, #6113, and #6114.~

### DIFF
--- a/src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
+++ b/src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
@@ -16,12 +16,16 @@
 #include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp"
+#include "Domain/Creators/TimeDependentOptions/ExpansionMap.hpp"
+#include "Domain/Creators/TimeDependentOptions/RotationMap.hpp"
 #include "Domain/Creators/TimeDependentOptions/ShapeMap.hpp"
+#include "Domain/Creators/TimeDependentOptions/TranslationMap.hpp"
 #include "Domain/FunctionsOfTime/FixedSpeedCubic.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
 #include "Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp"
 #include "Domain/FunctionsOfTime/PiecewisePolynomial.hpp"
 #include "Domain/FunctionsOfTime/QuaternionFunctionOfTime.hpp"
+#include "Domain/Structure/ObjectLabel.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Spherepack.hpp"
 #include "Options/ParseError.hpp"
 #include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrHorizon.hpp"
@@ -34,19 +38,18 @@
 namespace domain::creators::bco {
 template <bool IsCylindrical>
 TimeDependentMapOptions<IsCylindrical>::TimeDependentMapOptions(
-    double initial_time,
-    std::optional<ExpansionMapOptions> expansion_map_options,
-    std::optional<RotationMapOptions> rotation_map_options,
-    std::optional<TranslationMapOptions> translation_map_options,
-    std::optional<ShapeMapOptions<domain::ObjectLabel::A>> shape_options_A,
-    std::optional<ShapeMapOptions<domain::ObjectLabel::B>> shape_options_B,
+    double initial_time, ExpansionMapOptionType expansion_map_options,
+    RotationMapOptionType rotation_map_options,
+    TranslationMapOptionType translation_map_options,
+    ShapeMapOptionType<domain::ObjectLabel::A> shape_options_A,
+    ShapeMapOptionType<domain::ObjectLabel::B> shape_options_B,
     const Options::Context& context)
     : initial_time_(initial_time),
-      expansion_map_options_(expansion_map_options),
-      rotation_map_options_(rotation_map_options),
-      translation_map_options_(translation_map_options),
-      shape_options_A_(shape_options_A),
-      shape_options_B_(shape_options_B) {
+      expansion_map_options_(std::move(expansion_map_options)),
+      rotation_map_options_(std::move(rotation_map_options)),
+      translation_map_options_(std::move(translation_map_options)),
+      shape_options_A_(std::move(shape_options_A)),
+      shape_options_B_(std::move(shape_options_B)) {
   if (not(expansion_map_options_.has_value() or
           rotation_map_options_.has_value() or
           translation_map_options_.has_value() or
@@ -60,10 +63,15 @@ TimeDependentMapOptions<IsCylindrical>::TimeDependentMapOptions(
 
   const auto check_l_max = [&context](const auto& shape_option,
                                       const domain::ObjectLabel label) {
-    if (shape_option.has_value() and shape_option.value().l_max <= 1) {
-      PARSE_ERROR(context, "Initial LMax for object "
-                               << label << " must be 2 or greater but is "
-                               << shape_option.value().l_max << " instead.");
+    if (shape_option.has_value() and
+        time_dependent_options::l_max_from_shape_options(
+            shape_option.value()) <= 1) {
+      PARSE_ERROR(context,
+                  "Initial LMax for object "
+                      << label << " must be 2 or greater but is "
+                      << time_dependent_options::l_max_from_shape_options(
+                             shape_option.value())
+                      << " instead.");
     }
   };
 
@@ -82,37 +90,42 @@ TimeDependentMapOptions<IsCylindrical>::create_worldtube_functions_of_time()
   std::unordered_map<std::string,
                      std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
       result{};
+
   // The functions of time need to be valid only for the very first time step,
   // after that they need to be updated by the worldtube singleton.
   const double initial_expiration_time = initial_time_ + 1e-10;
   if (not expansion_map_options_.has_value()) {
     ERROR("Initial values for the expansion map need to be provided.");
   }
+
+  const auto& expansion_map_opts =
+      std::get<time_dependent_options::ExpansionMapOptions<false>>(
+          expansion_map_options_.value());
   result[expansion_name] =
       std::make_unique<FunctionsOfTime::IntegratedFunctionOfTime>(
           initial_time_,
-          std::array<double, 2>{
-              {gsl::at(expansion_map_options_.value().initial_values, 0),
-               gsl::at(expansion_map_options_.value().initial_values, 1)}},
+          std::array<double, 2>{expansion_map_opts.initial_values[0][0],
+                                expansion_map_opts.initial_values[1][0]},
           initial_expiration_time, false);
   result[expansion_outer_boundary_name] =
       std::make_unique<FunctionsOfTime::FixedSpeedCubic>(
           1.0, initial_time_,
-          expansion_map_options_.value().outer_boundary_velocity,
-          expansion_map_options_.value().outer_boundary_decay_time);
+          expansion_map_opts.asymptotic_velocity_outer_boundary.value(),
+          expansion_map_opts.decay_timescale_outer_boundary);
+
   if (not rotation_map_options_.has_value()) {
     ERROR(
         "Initial values for the rotation map need to be provided when using "
         "the worldtube.");
   }
+  const auto& rotation_map_opts =
+      std::get<time_dependent_options::RotationMapOptions<false>>(
+          rotation_map_options_.value());
 
   result[rotation_name] =
       std::make_unique<FunctionsOfTime::IntegratedFunctionOfTime>(
           initial_time_,
-          std::array<double, 2>{
-              0.,
-              gsl::at(rotation_map_options_.value().initial_angular_velocity,
-                      2)},
+          std::array<double, 2>{0., rotation_map_opts.angles[1][2]},
           initial_expiration_time, true);
 
   // Size and Shape FunctionOfTime for objects A and B. Only spherical excision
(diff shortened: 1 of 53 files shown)


src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -69,6 +69,19 @@ std::unique_ptr<FunctionOfTime> PiecewisePolynomial<MaxDeriv>::get_clone()
...
 }
 
+template <size_t MaxDeriv>
+std::unique_ptr<FunctionOfTime> PiecewisePolynomial<MaxDeriv>::create_at_time(
+    const double t, const double expiration_time) const {
+  std::vector<DataVector> all_derivs_vec = func_and_all_derivs(t);
+  std::array<DataVector, MaxDeriv + 1> all_derivs{};
+  for (size_t i = 0; i < all_derivs.size(); i++) {
+    gsl::at(all_derivs, i) = std::move(all_derivs_vec[i]);
+  }
```
nilsvu: I think this is `make_array<MaxDeriv+1>(func_and_all_derivs(t))`

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -79,6 +79,21 @@ std::unique_ptr<FunctionOfTime> QuaternionFunctionOfTime<MaxDeriv>::get_clone()
...
 
+template <size_t MaxDeriv>
+std::unique_ptr<FunctionOfTime>
+QuaternionFunctionOfTime<MaxDeriv>::create_at_time(
+    const double t, const double expiration_time) const {
+  std::vector<DataVector> all_angle_derivs_vec = angle_func_and_all_derivs(t);
+  std::array<DataVector, MaxDeriv + 1> all_angle_derivs{};
+  for (size_t i = 0; i < all_angle_derivs.size(); i++) {
+    gsl::at(all_angle_derivs, i) = std::move(all_angle_derivs_vec[i]);
+  }
```
nilsvu: also `make_array`

src/Domain/Creators/TimeDependentOptions/TranslationMap.hpp
```diff
@@ -22,29 +22,42 @@ namespace domain::creators::time_dependent_options {
  * coefficients.
  */
 template <size_t Dim>
-struct TranslationMapOptions {
-  using type = Options::Auto<TranslationMapOptions, Options::AutoLabel::None>;
-  static std::string name() { return "TranslationMap"; }
+struct TranslationMapHardcodedOptions {
```
nilsvu: I always read "hardcore options" 💪

Suggestion: keep this class named `TranslationMapOptions`. And the tag below name just `TranslationMap` because it's already in the `time_dependent_options` namespace.

nilsvu (changes requested):
ok I'll need more coffee for this one

You can squash everything into one commit, except keep the "Allow FoTs to create themselves" commit separate.

tests/Unit/Domain/FunctionsOfTime/Test_IntegratedFunctionOfTime.cpp
```diff
@@ -59,6 +59,21 @@ void test(const gsl::not_null<FunctionsOfTime::FunctionOfTime*> f_of_t,
...
   }
+
+  const size_t index = positions.size() - 8;
+  const auto new_time = static_cast<double>(index) * dt;
+  const auto new_expiration = new_time + dt;
+  const auto copy_at_time = f_of_t->create_at_time(new_time, new_expiration);
+
+  CAPTURE(rotation);
+  //   const double check_time = 0.5 * (new_time + new_expiration);
+  const double check_time = new_time;
```
nilsvu, line 69: Want to check here that the FoT is also the same a bit later?

knelli2:
@nilsvu Pushed my changes. This should be ready for review. Good luck 😉

src/Domain/FunctionsOfTime/PiecewisePolynomial.cpp
```diff
@@ -69,6 +69,14 @@ std::unique_ptr<FunctionOfTime> PiecewisePolynomial<MaxDeriv>::get_clone()
   return std::make_unique<PiecewisePolynomial>(*this);
 }
 
+template <size_t MaxDeriv>
+std::unique_ptr<FunctionOfTime> PiecewisePolynomial<MaxDeriv>::create_at_time(
+    const double t, const double expiration_time) const {
+  return std::make_unique<PiecewisePolynomial>(
+      t, make_array<DataVector, MaxDeriv + 1>(func_and_all_derivs(t)),
```
nilsvu: Can you use `func_and_derivs(t)` so you don't need the `make_array` and elide the array->vector->array transformation?

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
```diff
@@ -79,6 +80,16 @@ std::unique_ptr<FunctionOfTime> QuaternionFunctionOfTime<MaxDeriv>::get_clone()
   return std::make_unique<QuaternionFunctionOfTime>(*this);
 }
 
+template <size_t MaxDeriv>
+std::unique_ptr<FunctionOfTime>
+QuaternionFunctionOfTime<MaxDeriv>::create_at_time(
+    const double t, const double expiration_time) const {
+  return std::make_unique<QuaternionFunctionOfTime>(
+      t, func(t),
+      make_array<DataVector, MaxDeriv + 1>(angle_func_and_all_derivs(t)),
```
nilsvu: Also here `angle_f_of_t_.func_and_derivs(t)`

src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.hpp
```diff
@@ -146,105 +148,30 @@ struct TimeDependentMapOptions {
...
-    static std::string name() { return "RotationMap"; }
-    static constexpr Options::String help = {
-        "Options for a time-dependent rotation map about an arbitrary axis. "
-        "Specify 'None' to not use this map."};
-
-    struct InitialAngularVelocity {
-      using type = std::array<double, 3>;
-      static constexpr Options::String help = {"The initial angular velocity."};
-    };
+  using ExpansionMapOptions =
```
nilsvu, line 151: I got confused here because `ExpansionMapOptions` is also a struct in Expansion.hpp, but not the one you are referring to here.

src/Domain/Creators/TimeDependentOptions/ExpansionMap.hpp
```diff
@@ -21,88 +24,110 @@ namespace domain::creators::time_dependent_options {
  * \brief Class to be used as an option for initializing expansion map
  * coefficients.
  */
+template <bool AllowSettleFoTs>
 struct ExpansionMapOptions {
-  using type = Options::Auto<ExpansionMapOptions, Options::AutoLabel::None>;
-  static std::string name() { return "ExpansionMap"; }
+  using type = ExpansionMapOptions;
```
nilsvu: If you change this line to the `Auto<variant<>>` then you don't need the extra option tag below and the confusion in the other comment is resolved. Or delete this line.

src/Domain/Creators/TimeDependentOptions/ExpansionMap.cpp
```diff
@@ -4,87 +4,206 @@
...
+
+    if (UNLIKELY(pp_volume_fot == nullptr and settle_volume_fot == nullptr)) {
+      ERROR_NO_TRACE(
+          "Expansion function of time read from volume data is not a "
+          "PiecewisePolynomial<2> or a SettleToConstant. Cannot use it to "
+          "initialize the expansion map.");
+    }
+
+    if (pp_volume_fot != nullptr) {
+      result =
```
nilsvu: (optional) you can just return in these code branches

src/Domain/Creators/TimeDependentOptions/ExpansionMap.cpp
```diff
@@ -4,87 +4,206 @@
...
+             "SettleToConstant, but SettleToConstant functions of time aren't "
+             "allowed.");
+      result = volume_fot.at(name)->get_clone();
+    }
+  } else if (std::holds_alternative<ExpansionMapOptions<AllowSettleFoTs>>(
+                 expansion_map_options)) {
+    const auto& hard_coded_options =
+        std::get<ExpansionMapOptions<AllowSettleFoTs>>(expansion_map_options);
+
+    if (hard_coded_options.asymptotic_velocity_outer_boundary.has_value()) {
```
nilsvu, line 114: confused that `asymptotic_velocity_outer_boundary` is not actually used below. Add a comment?

src/Domain/Creators/TimeDependentOptions/ExpansionMap.cpp
```diff
@@ -4,87 +4,206 @@
...
+
+  return result;
+}
+
+template <bool AllowSettleFoTs>
+std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>
+get_expansion_outer_boundary(
+    const std::variant<ExpansionMapOptions<AllowSettleFoTs>, FromVolumeFile>&
+        expansion_map_options,
+    const double initial_time) {
```
nilsvu: (optional) this function could be merged into the one above with a boolean template param or a runtime boolean argument. Maybe that would make it a bit easier to understand (up to you).

src/Domain/Creators/TimeDependentOptions/ExpansionMap.cpp
```diff
@@ -4,87 +4,206 @@
...
+          "not a FixedSpeedCubic or a SettleToConstant. Cannot use it to "
+          "initialize the expansion map.");
+    }
+
+    ASSERT(cubic_volume_fot != nullptr or AllowSettleFoTs,
+           "ExpansionOuterBoundary function of time in the volume file is a "
+           "SettleToConstant, but SettleToConstant functions of time aren't "
+           "allowed.");
+
+    result = volume_fot.at(name)->get_clone();
```
nilsvu: (optional) I think you can always just return `create_at_time()` because it defaults to `get_clone()`, then you don't have to decide if the FoT actually does something special in `create_at_time`.

src/Domain/Creators/TimeDependentOptions/ShapeMap.hpp
```diff
@@ -189,11 +223,9 @@ struct YlmsFromSpEC {
  */
 template <bool IncludeTransitionEndsAtCube, domain::ObjectLabel Object>
 struct ShapeMapOptions {
-  using type = Options::Auto<ShapeMapOptions, Options::AutoLabel::None>;
-  static std::string name() { return "ShapeMap" + get_output(Object); }
   static constexpr Options::String help = {
       "Options for a time-dependent distortion (shape) map about the "
-      "specified object. Specify 'None' to not use this map."};
+      "specified object.."};
```
nilsvu: remove one `.`

src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
```diff
@@ -189,102 +206,56 @@ TimeDependentMapOptions<IsCylindrical>::create_functions_of_time(
...
+        time_dependent_options::get_expansion_outer_boundary(
+            expansion_map_options_.value(), initial_time_);
   }
 
   // RotationMap FunctionOfTime for the rotation angles about each
-  // axis.  The initial rotation angles don't matter as we never
-  // actually use the angles themselves. We only use their derivatives
-  // (omega) to determine map parameters. In theory we could determine
-  // each initial angle from the input axis-angle representation, but
-  // we don't need to.
```
nilsvu, line 215: Could you add some docs to `QuaternionFunctionOfTime` so we don't get confused by this again?

nilsvu (changes requested):

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
knelli2: `func_and_derivs` is private in the PiecewisePolynomial

tests/Unit/Domain/FunctionsOfTime/Test_IntegratedFunctionOfTime.cpp
knelli2: An IntegratedFunctionOfTime is only valid at an update time, so we can't check at a later time. But I will remove the commented out line

src/Domain/FunctionsOfTime/QuaternionFunctionOfTime.cpp
nilsvu: could just make it public

nilsvu (dismissed):

nilsvu (approved):

PR #6118, opened 2024-06-25
Control BBH masses and spins in initial data

iago-mendes:
## Proposed changes

I've implemented a Newton-Raphson iterative scheme with Broyden's method that controls the BBH masses and spins during the Initial Data. This follows a similar approach to what is described in section 2.3 of [arXiv:1506.01689](https://arxiv.org/abs/1506.01689), except that here we control the Kerr masses and spins instead of the excision radii and horizon frequencies.
### Upgrade instructions

If calling `generate_id` or `id_parameters` from `InitialData.py`, you now need to specify the masses `mass_a` and `mass_b` separately instead of the previous `mass_ratio`. Note that `generate_id_command` still takes in `mass_ratio` as an argument.

### Code review checklist
  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/ControlId.py
+++ b/support/Pipelines/Bbh/ControlId.py
@@ -1,0 +1,190 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+from pathlib import Path
+from typing import Optional, Union
+
+import h5py
+import numpy as np
+import yaml
+
+import spectre.IO.H5 as spectre_h5
+from spectre.Pipelines.Bbh.InitialData import generate_id
+from spectre.Visualization.ReadH5 import to_dataframe
+
+logger = logging.getLogger(__name__)
+
+DEFAULT_RESIDUAL_TOLERANCE = 1.0e-6
+DEFAULT_MAX_ITERATIONS = 30
+
+
+def control_id(
+    id_input_file_path: Union[str, Path],
+    id_run_dir: Optional[Union[str, Path]] = None,
+    residual_tolerance: float = DEFAULT_RESIDUAL_TOLERANCE,
+    max_iterations: int = DEFAULT_MAX_ITERATIONS,
+):
+    """Control BBH physical parameters.
+
+    This function is called after initial data has been generated and horizons
+    have been found in 'PostprocessId.py'. It uses an iterative scheme to drive
+    the black hole physical parameters (masses and spins) closer to the desired
+    values.
+
+    For each iteration, this function does the following:
+
+    - Determine new guesses for ID input parameters.
+
+    - Generate initial data using these guesses.
+
+    - Find horizons in the generated initial data.
+
+    - Measure the difference between the horizon quantities and the desired
+      values.
+
+    Arguments:
+      id_input_file_path: Path to the input file of the first initial data run.
+      id_run_dir: Directory of the first initial data run. If not provided, the
+        directory of the input file is used.
+      residual_tolerance: Residual tolerance used for termination condition.
+        (Default: 1.e-6)
+      max_iterations: Maximum of iterations allowed. Note: each iteration is
+        very expensive as it needs to solve an entire initial data problem.
+        (Default: 30)
+    """
+
+    # Read input file
+    if id_run_dir is None:
+        id_run_dir = Path(id_input_file_path).resolve().parent
+    with open(id_input_file_path, "r") as open_input_file:
+        _, id_input_file = yaml.safe_load_all(open_input_file)
+    binary_data = id_input_file["Background"]["Binary"]
+    M_target_A = binary_data["ObjectRight"]["KerrSchild"]["Mass"]
+    M_target_B = binary_data["ObjectLeft"]["KerrSchild"]["Mass"]
+    chi_target_A = binary_data["ObjectRight"]["KerrSchild"]["Spin"]
+    chi_target_B = binary_data["ObjectLeft"]["KerrSchild"]["Spin"]
+    x_B, x_A = binary_data["XCoords"]
+    separation = x_A - x_B
+    orbital_angular_velocity = binary_data["AngularVelocity"]
+    radial_expansion_velocity = binary_data["Expansion"]
+
+    # File to write control diagnostic data
+    data_file = open(f"{id_run_dir}/ControlParamsData.txt", "w")
+
+    iteration = 0
+    control_run_dir = id_run_dir
+
+    # Function to be minimized
+    def Residual(
+        M_Kerr_A: float,
+        M_Kerr_B: float,
+        chi_Kerr_A_x: float,
+        chi_Kerr_A_y: float,
+        chi_Kerr_A_z: float,
+        chi_Kerr_B_x: float,
+        chi_Kerr_B_y: float,
+        chi_Kerr_B_z: float,
+    ):
+        nonlocal iteration
+        nonlocal control_run_dir
+
+        if iteration > 0:
+            logger.info(
+                "\n"
+                "=========================================="
+                f" Control of BBH Parameters ({iteration}) "
+                "=========================================="
+            )
+            control_run_dir = f"{id_run_dir}/ControlParams{iteration:02}"
+
+            # Run ID and find horizons
+            generate_id(
+                mass_a=M_Kerr_A,
+                mass_b=M_Kerr_B,
+                dimensionless_spin_a=[chi_Kerr_A_x, chi_Kerr_A_y, chi_Kerr_A_z],
+                dimensionless_spin_b=[chi_Kerr_B_x, chi_Kerr_B_y, chi_Kerr_B_z],
+                separation=separation,
+                orbital_angular_velocity=orbital_angular_velocity,
+                radial_expansion_velocity=radial_expansion_velocity,
+                run_dir=control_run_dir,
+                control=False,
+                evolve=False,
+                scheduler=None,
+            )
+
+        # Get black hole physical parameters
+        with spectre_h5.H5File(
+            f"{control_run_dir}/Horizons.h5", "r"
+        ) as horizons_file:
+            AhA_quantities = to_dataframe(
+                horizons_file.get_dat("AhA.dat")
+            ).iloc[-1]
+
+            M_A = AhA_quantities["ChristodoulouMass"]
+            chi_A_x = AhA_quantities["DimensionlessSpinVector_x"]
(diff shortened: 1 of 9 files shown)


support/Pipelines/Bbh/InitialData.yaml
```diff
@@ -11,6 +11,9 @@ Next:
     id_input_file_path: __file__
     id_run_dir: ./
     pipeline_dir: {{ pipeline_dir | default("None") }}
+    control: {{ control }}
+    control_residual_tolerance: 1.e-6
+    control_max_iterations: 30
```
nilsvu: I think just delete these two options so the defaults in the function are used.

support/Pipelines/Bbh/InitialData.py
```diff
@@ -50,11 +52,14 @@ def id_parameters(
     """
 
     # Sanity checks
-    assert mass_ratio >= 1.0, "Mass ratio is defined to be >= 1.0."
+    mass_ratio = mass_a / mass_b
```
nilsvu: If you set this to `max(mass_a, mass_b) / min(mass_a, mass_b)` then `q >= 1` will always be true and you can delete the assert below.

support/Pipelines/Bbh/InitialData.py
```diff
@@ -141,6 +149,7 @@ def generate_id(
 
     Scheduling options:
       id_input_file_template: Input file template where parameters are inserted.
+      control: Set to True to control BBH physical parameters (default: False).
```
nilsvu: Please add some more docs to explain this option, maybe in the documentation text above. Note that with this option set to False the horizon masses and spins in the generated data will differ from the input parameters. And with this option set to True a postprocessing control loop will adjust the input parameters to drive the horizon masses and spins to the specified values.

support/Pipelines/Bbh/PostprocessId.py
```diff
@@ -44,6 +52,8 @@ def postprocess_id(
       surface coordinates and coefficients are written to the 'horizons_file' in
       subfiles 'Ah{A,B}/Coordinates' and 'Ah{A,B}/Coefficients'.
 
+    - Alow user to trigger control of BBH parameters.
```
nilsvu: Typo "allow". Add a few more details. Maybe something like "If 'control' is set to True, run a control loop such that masses and spins of the horizons match the input parameters. See ControlId.py for details.".

support/Pipelines/Bbh/PostprocessId.py
```diff
@@ -24,6 +29,9 @@ def postprocess_id(
     id_run_dir: Optional[Union[str, Path]] = None,
     horizon_l_max: int = 12,
     horizons_file: Optional[Union[str, Path]] = None,
+    control: bool = False,
```
nilsvu: I think set this to `True` so it's the same as in the CLI command below.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,173 @@
...
+            generate_id(
+                mass_a=M_Kerr_A,
+                mass_b=M_Kerr_B,
+                dimensionless_spin_a=[chi_Kerr_A_x, chi_Kerr_A_y, chi_Kerr_A_z],
+                dimensionless_spin_b=[chi_Kerr_B_x, chi_Kerr_B_y, chi_Kerr_B_z],
+                separation=separation,
+                orbital_angular_velocity=orbital_angular_velocity,
+                radial_expansion_velocity=radial_expansion_velocity,
+                run_dir=control_run_dir,
+                scheduler=None,
```
nilsvu, line 113: Explicitly set `control=False` and `evolve=False` here, just to be clear. Then the defaults for these parameters set in the other functions don't matter here.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,173 @@
...
+    """Control BBH physical parameters.
+
+    This funciton is called after initial data has been generated and horizons
+    have been found in 'PostprocessId.py'. It uses an iterative scheme to drive
+    the black hole physical parameters (masses and spins) closer to the desired
+    values.
+
+    For each iteration, this function does the following:
+
+    - Determine new guesses for the Superposed Kerr-Schild (SKS).
```
nilsvu: I think this is not specific to SKS. It should also work for other ID gauges if we adjust the relationship between masses/spins and horizon radii/frequencies in `id_parameters`. So reword to something like: "Determine new guesses for ID input parameters"

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,173 @@
...
+        chi_Kerr_A_z: float,
+        chi_Kerr_B_x: float,
+        chi_Kerr_B_y: float,
+        chi_Kerr_B_z: float,
+    ):
+        nonlocal iteration
+        nonlocal control_run_dir
+
+        if iteration > 0:
+            print(
```
nilsvu: Use `logging` so we can control the verbosity of the output. The other files in this directory should have some examples.

Also for the other `print`s in this file.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,173 @@
...
+                separation=separation,
+                orbital_angular_velocity=orbital_angular_velocity,
+                radial_expansion_velocity=radial_expansion_velocity,
+                run_dir=control_run_dir,
+                scheduler=None,
+            )
+
+        # Get black hole physical parameters
+        horizons_file = h5py.File(f"{control_run_dir}/Horizons.h5", "r")
+        M_A = horizons_file["AhA.dat"][0, 4]  # ChristodoulouMass
```
nilsvu: Use `spectre.Visualization.ReadH5.to_dataframe` to read the data as a Pandas DataFrame, then use the names like "ChristodoulouMass" instead of column indices. That will ensure this keeps working when we add more horizon output quantities. There's an example how to do this with our H5 Python bindings in `Inspiral.py`, but `to_dataframe` also works with `h5py`.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,173 @@
...
+                M_B - M_target_B,
+                chi_A_x - chi_target_A[0],
+                chi_A_y - chi_target_A[1],
+                chi_A_z - chi_target_A[2],
+                chi_B_x - chi_target_B[0],
+                chi_B_y - chi_target_B[1],
+                chi_B_z - chi_target_B[2],
+            ]
+        )
+        print("Control Residual =", residual)
```
nilsvu: You can get nice formatting with `f"Control residual = {residual:e}"` (residuals are best displayed in exponential notation)

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,173 @@
...
+
+        data_file.write(
+            f" {iteration},"
```
nilsvu, line 154: (optional) use `", ".join([...])`

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,173 @@
...
+    J = np.identity(len(u))
+
+    while iteration < max_iterations and np.max(np.abs(F)) > residual_tolerance:
+        iteration += 1
+        u_prev = u
+
+        # Update the free parameters using a quasi-Newton-Raphson method
+        u = u_prev - np.dot(np.linalg.inv(J), F)
+
+        Delta_u = u - u_prev
```
nilsvu: I would find it easier to understand if you do this:

- Move this `delta_u` two lines up, and set it to `delta_u = np.dot(np.linalg.inv(J), F)`
- In your `residual` function, add a minus sign so you return `target - measured`.
- Then here you can do `u += delta_u` and you don't need the variable `u_prev` anymore.

nilsvu (changes requested):
Very nice! You can squash all of these commits into one. Please push any changes you make in response to my comments as a new commit. We usually mark such commits with the word "fixup" at the beginning and squash them in once the review is done (the fixup commit helps with review so I can see only the changes you made).

support/Pipelines/Bbh/ControlId.py
iago-mendes: I'll probably change how this is outputted soon, so I think we can leave it like this for now.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,190 @@
...
+
+def control_id(
+    id_input_file_path: Union[str, Path],
+    id_run_dir: Optional[Union[str, Path]] = None,
+    residual_tolerance: float = DEFAULT_RESIDUAL_TOLERANCE,
+    max_iterations: int = DEFAULT_MAX_ITERATIONS,
+):
+    """Control BBH physical parameters.
+
+    This funciton is called after initial data has been generated and horizons
```
nilsvu: Typo "function"

nilsvu (commented):
Looks great! Please squash your commits into one, including this typo fix 👍 Nice job!

nilsvu (dismissed):

support/Pipelines/Bbh/ControlId.py
```diff
@@ -0,0 +1,190 @@
...
+        iteration += 1
+
+        # Update the free parameters using a quasi-Newton-Raphson method
+        Delta_u = np.dot(np.linalg.inv(J), F)
+        u += Delta_u
+
+        F = Residual(*u)
+
+        # Update the Jacobian using Broyden's method
+        J = J - np.dot(F, Delta_u) / np.linalg.norm(Delta_u)
```
nilsvu: In the literature I find always `np.dot(Delta_u, Delta_u)`, i.e. the squared L2 norm (e.g. Numerical Recipes). I think the Ossokine2015 paper might just be missing a square. Could you check this please?

nilsvu (changes requested):

support/Pipelines/Bbh/ControlId.py
iago-mendes: I've fixed the dot product. There was also an issue of a missing outer product (instead of the dot product) between `F` and `Delta_u`. Now the control loop reaches `1.e-6` in only 4 iterations (< 30min).

nilsvu (approved):

support/Pipelines/Bbh/ControlId.py
nilsvu: Awesome!

PR #6120, opened 2024-06-26
Fix unit conversions between SpECTRE and RotNS for non-polytropic EoS

ncorsobh:
## Proposed changes

RotNS uses a special unit normalization when dealing with polytropic equations of state, which is what SpECTRE has been using so far. For any other EoS, however, it uses a different unit normalization. cf. equations 15-23 in [Cook _et al._ (1992)](https://ui.adsabs.harvard.edu/abs/1992ApJ...398..203C/abstract) with equations A1-A10 in [Cook _et al._ (1994)](https://ui.adsabs.harvard.edu/abs/1994ApJ...424..823C/abstract).

This PR allows the user to specify in the input file for the `RotatingStar` executable whether a polytropic EoS was used. The first commit adjusts the structure of `Xcts/WrappedGr` because the new structure for `RotatingStar` introduced in the second commit includes an `Options::Alternatives` that breaks the old implementation.

### Upgrade instructions

If you use `RotatingStar` with polytropic EoSs, nothing changes for you, the input file works the same.

If you use some other EoS, instead of using the `PolytropicConstant` input option, you should use `EquationOfState` followed by the equation of state with which you would like to initialize the data.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -61,6 +63,9 @@ CstSolution::CstSolution(const std::string& filename,
   gamma_.destructive_resize(num_grid_points_);
   omega_.destructive_resize(num_grid_points_);
 
+  const double kappa_norm = pow(hydro::units::cgs::speed_of_light, 2.) /
```
isaaclegred: Is this some kind of normalization constant?  Does it have a reference somewhere in the papers?  Also it can probably be `constexpr`

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -70,12 +75,39 @@ CstSolution::CstSolution(const std::string& filename,
       cst_file >> radius_[index] >> cos_theta_[index] >>
           rest_mass_density_[index] >> alpha_[index] >> rho_[index] >>
           gamma_[index] >> omega_[index] >> fluid_velocity_[index];
-      radius_[index] *= pow(equilibrium_kappa, polytropic_index_ * 0.5);
-      rest_mass_density_[index] *= pow(equilibrium_kappa, -polytropic_index_);
+      if (is_polytrope) {
```
isaaclegred, line 92: Do I understand correctly that what's going on here is that, if the EoS being used is a polytrope, then it is possible to rescale the solution using Eq. 15... in https://articles.adsabs.harvard.edu/pdf/1992ApJ...398..203C, but if it's not then it's not possible?  If so, then even if the EoS is a polytrope it's not absolutely necessary to do the rescaling.  Perhaps the `is_polytrope` variable should be changed to `rescale_with_polytrope`.  I think this would leave the rest of the code structure intact, but would help in the case that someone is trying to figure out what to do if they have a polytropic EoS but don't want to rescale the solution.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -70,12 +75,39 @@ CstSolution::CstSolution(const std::string& filename,
...
+        radius_[index] *= pow(polytropic_constant, 0.5 * polytropic_index_);
+        rest_mass_density_[index] *=
+            pow(polytropic_constant, -polytropic_index_);
+        omega_[index] *= pow(polytropic_constant, -0.5 * polytropic_index_) /
+                         equatorial_radius_;
+        fluid_velocity_[index] *=
+            pow(polytropic_constant, -0.5 * polytropic_index_) /
+            equatorial_radius_;
+      } else {
+        radius_[index] *= sqrt(kappa_norm) / hydro::units::cgs::length_unit;
```
isaaclegred, line 102: Could you add a comment as to broadly speaking what's happening here, this is taking the data in the RotNS format and converting it to geometric units, right?  Perhaps add a pointer to wherever you got these conversion expressions from, I think it's mostly clear modulo the factors of 10^15 (which I think I understand), and the factors of kappa_norm (which I don't think I understand)

isaaclegred (changes requested):
Looks good!  Just a couple comments/requests for additional documentation. Thanks for doing this!

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
ncorsobh: Yep, it's just a normalization constant. It's defined in equation A1 in [Cook _et al._ (1994)](https://ui.adsabs.harvard.edu/abs/1994ApJ...424..823C/abstract). Specifically, it's a length scale. I added comments in the code to address this. Let me know if you think it's sufficient.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
ncorsobh: I think you're onto the right idea. Basically if the EoS in RotNS is set to `PT` (polytrope) then the normalization is done using the method in the 1992 paper, but if any other EoS is used, then it's done according to the method in the 1994 paper.
That being said, if the EoS is `PT`, you must use `is_polytrope=true`. The 1992 paper already uses geometric units, it just lets K (i.e. P=K\rho^\gamma) be a free parameter. RotNS assumes K=1 for its calculations. If the user doesn't want to rescale, they can just set `polytropic_constant=1.` and nothing will change. The 1994 method instead normalizes with CGS, so `is_polytrope=false` converts from CGS into geometric units, so the user would not want to use this for a `PT` EoS, even if they don't want to rescale.

(I specify setting the EoS to `PT` in RotNS because you could in principle import a polytrope as a tabulated EoS into RotNS, and it would use the 1994 normalization instead. But I don't know why someone would do this instead of the built-in polytrope functionality)

I'll still consider renaming the variable if you think it's still necessary, but personally I think it would make more sense to stay as is.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
ncorsobh: Just pushed a change that added a comment clarifying this and probably addressing your other comments a bit. Let me know if it's clearer now.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
pajkosmi: Related to Isaac's below comment, I think a sentence or 2 clarifying the significance of 1e15 and why it's there would help.

ncorsobh: Added a clarification on the 1e15 factor. Let me know if you think it's sufficient.

pajkosmi: I think it's sufficient

pajkosmi (approved):
Looks good.  Make sure @isaaclegred is ok with it too.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
isaaclegred: Yeah I see, makes sense.  The main point is that the settings that are used should have agreement with the settings used in RotNS.  Can this text be added to replace the current doc for `CstSolution` in `RotatingStar.hpp`?

>  if the EoS in RotNS is set to PT (polytrope) then the normalization is done using the method in the 1992 paper, but if any other EoS is used, then it's done according to the method in the 1994 paper.
That being said, if the EoS is PT, you must use is_polytrope=true. The 1992 paper already uses geometric units, it just lets K (i.e. P=K\rho^\gamma) be a free parameter. RotNS assumes K=1 for its calculations. If the user doesn't want to rescale, they can just set polytropic_constant=1. and nothing will change. The 1994 method instead normalizes with CGS, so is_polytrope=false converts from CGS into geometric units, so the user would not want to use this for a PT EoS, even if they don't want to rescale.

isaaclegred:
Looks good to me, please squash everything (including modification to the documentation of `CstSolution`)

ncorsobh:
Before this PR is approved under his nose, @nilsvu should probably give the Xcts changes a look, since he seems to be mainly involved with a lot of the recent work with this section of the code.

nilsvu:
I approve the first commit. Thanks for checking with me. I haven't looked at the second commit.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -61,6 +63,23 @@ CstSolution::CstSolution(const std::string& filename,
...
+  // the user does not want to rescale, they should set `PolytropicConstant` to
+  // be `1.`, which is the value used in RotNS.
+  // If instead `EquationOfState` is specified, `is_polytrope` is set to `false`
+  // and the data is rescaled according to the normalization used in equations
+  // A1-A10 in \cite{Cook1994}. This uses CGS units, defined by a fundamental
+  // length scale $\kappa^{1/2}$, defined as $\kappa = \frac{c^2}{G
+  // \epsilon_0}$, where $\epsilon_0 = 10^{15}$ g cm^-3 is an arbitrarily
+  // defined density scale.
+
+  const double kappa_norm = pow(hydro::units::cgs::speed_of_light, 2.) /
```
nilsdeppe: you can just use `square(hydro::units::cgs::speed_of_light)`. `pow` is notoriously slow and so if you know the integer exponent, using `square`, `cube`, or `pow<INTEGER_EXPONENT_VALUE>` is better

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -636,7 +714,9 @@ tuples::TaggedTuple<hydro::Tags::Pressure<DataType>> RotatingStar::variables(
     tmpl::list<hydro::Tags::Pressure<DataType>> /*meta*/) const {
   const auto rest_mass_density = get<hydro::Tags::RestMassDensity<DataType>>(
       variables(vars, x, tmpl::list<hydro::Tags::RestMassDensity<DataType>>{}));
-  return {equation_of_state_.pressure_from_density(rest_mass_density)};
+  using std::max;
+  return {equation_of_state_->pressure_from_density(
+      Scalar<DataType>{DataType{max(1.0e-300, get(rest_mass_density))}})};
```
nilsdeppe: Why do you have these magic numbers now when we didn't need them before? If you really do need them, could you add a `static constexpr double` private member variable with this value so it can be documented and is only in one place?

The difficulty with hard coding values like this is that other people will have no idea why you choose this value or why it's necessary.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -614,9 +691,10 @@ RotatingStar::variables(
   using std::max;
   return {hydro::relativistic_specific_enthalpy(
       Scalar<DataType>{DataType{max(1.0e-300, get(rest_mass_density))}},
-      equation_of_state_.specific_internal_energy_from_density(
-          rest_mass_density),
-      equation_of_state_.pressure_from_density(rest_mass_density))};
+      equation_of_state_->specific_internal_energy_from_density(
+          Scalar<DataType>{DataType{max(1.0e-300, get(rest_mass_density))}}),
+      equation_of_state_->pressure_from_density(
```
nilsdeppe: I think the cleaner approach would be to retrieve the pressure using the `get(get<hydro::Tags::Pressure<DataType>>(variables(vars, x, tmpl::list<hydro::Tags::Pressure<DataType>>{})))`, the reason being that then you only need the number `1e-300` for the pressure once there.

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -614,9 +691,10 @@ RotatingStar::variables(
   using std::max;
   return {hydro::relativistic_specific_enthalpy(
       Scalar<DataType>{DataType{max(1.0e-300, get(rest_mass_density))}},
-      equation_of_state_.specific_internal_energy_from_density(
-          rest_mass_density),
-      equation_of_state_.pressure_from_density(rest_mass_density))};
+      equation_of_state_->specific_internal_energy_from_density(
```
nilsdeppe: Same as the pressure, call the other function so you don't have multiple copies of the magic number floating around.

nilsdeppe (changes requested):
You can squash immediately

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
ncorsobh: Some EoS's run into FPEs when fed in 0 for these kinds of state variable functions (in my case it was `Spectral`). Likely it wasn't needed before because people were only running with polytropes.
I'm happy to make it neater as you suggest, but I should note that I just took this approach from elsewhere in the file (see line 693). I'm not sure why it was there before, but I'm assuming for similar reasons? It also shows up in line 507 of `src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp`. So this may speak to a broader issue, e.g. maybe we should put protection from FPEs in the case of 0 density directly in these EoS functions instead of what I'm doing here to compensate for it externally

nilsdeppe: I don't like the idea of protecting against 0 in the EOS because practically speaking we need an atmosphere anyway since the evolution equations assume a large number of particles, and, more importantly, because this adds checks that are done every step of the evolution where we have other code that guarantees positive definite numbers. I think unfortunately the best approach is what you're doing here. We override the small numbers later anyway with our atmosphere fix :shrug: I do wonder if 1e-50 would be better since in some cases (e.g. polytrope) we might square the number and 1e-300^2=underflow and that can cause problems

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -600,7 +677,9 @@ RotatingStar::variables(
     const tnsr::I<DataType, 3>& /*x*/,
     tmpl::list<hydro::Tags::RestMassDensity<DataType>> /*meta*/) const {
   interpolate_vars_if_necessary(vars);
-  return {Scalar<DataType>{vars->rest_mass_density.value()}};
+  using std::max;
+  return {Scalar<DataType>{
+      DataType{max(atmosphere_floor_, vars->rest_mass_density.value())}}};
```
ncorsobh, line 682: @nilsdeppe The problem has been fixed here. `RotatingStar` only uses 1D EoS to initialize currently, so any FPE issues can be taken care of by setting the floor on the density when it's first loaded in. Thus, the floor is only applied once.
(We discussed in person but I'm noting it here for the good of the cause)

nilsdeppe (approved):

PR #6123, opened 2024-06-26
Stop storing volume time-stepper errors

wthrowe:
## Proposed changes

Instead, calculate the error norm we use immediately and store that (a double).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/StepChoosers/ErrorControl.hpp
+++ b/src/Time/StepChoosers/ErrorControl.hpp
@@ -7,6 +7,7 @@
 #include <cmath>
 #include <limits>
 #include <memory>
+#include <optional>
 #include <pup.h>
 #include <string>
 #include <type_traits>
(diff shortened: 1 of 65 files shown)


src/Time/StepChoosers/ErrorControl.hpp
```diff
@@ -311,4 +326,70 @@ struct IsUsingTimeSteppingErrorControlCompute
...
+    if (const auto* const error_control =
+            dynamic_cast<const ::StepChoosers::ErrorControl_detail::
+                             ErrorControlBase<EvolvedVariableTag>*>(
+                &step_chooser);
+        error_control != nullptr) {
+      const auto this_tolerances = error_control->tolerances();
+      if (tolerances->has_value() and tolerances->value() != this_tolerances) {
+        ERROR_NO_TRACE(
+            "All ErrorControl events for the same variables must use the same "
+            "tolerances.");
```
nilsdeppe: Would it be helpful to print the variables tag name at the end of the error? that way users can at least know what to look at in the input file a bit better.

tests/Unit/Time/Tags/Test_StepperErrorTolerances.cpp
```diff
@@ -0,0 +1,19 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Framework/TestingFramework.hpp"
+
+#include <string>
+
+#include "Helpers/DataStructures/DataBox/TestHelpers.hpp"
+#include "Helpers/DataStructures/TestTags.hpp"
```
kidder: not needed

kidder (changes requested):
you can squash the change

wthrowe:
Squashed both directly.

kidder (approved):

PR #6127, opened 2024-06-26
Set shape of initial BBH inspiral excisions to ID excisions

### LINKED ISSUE #6126: Revert BBH inspiral initial excisions to ID excision
The recent #5957 changed the initial size and shape of the evolution excisions. This seems to break the control system latching on to the horizons initially in some cases. We should make the evolution excisions the same as the ID excisions again. Then later we can investigate if it helps the control systems to deform the initial evolution excisions to conform to the ID horizon shapes.

nilsvu:
## Proposed changes

Fixes #6126. This makes sure that the ID can always be interpolated to the Ev grid. Leaves #6128 open to do later.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -116,7 +116,7 @@ DomainCreator:
             - {{ ExcisionAShapeSpin_y }}
             - {{ ExcisionAShapeSpin_z }}
 {% endif %}
-        SizeInitialValues: [0.0, 0.0, 0.0]
+        SizeInitialValues: [0.0, -0.1, 0.0]
```
nilsvu: @knelli2 I had to set this as well to get the control systems to start. Do you have a better idea how to do this; and do you know what SpEC does here? @markscheel do you know?

geoffrey4444: Sometimes you have to do this in spec for high spins (make the excision surface move outward to avoid initial incoming char speeds), but I did not have to do this for the spec initial data I evolved

nilsvu: I dropped this change and instead made the ID excisions bigger to match the value SpEC uses for negative-expansion boundary conditions. That seems to help.

knelli2 (approved):

PR #6140, opened 2024-07-03
Compute total ADM linear momentum integrands.

iago-mendes:
## Proposed changes

Add pointwise functions for the integrands needed to compute the total ADM linear momentum. The formulas were taken from section 2.4 of [arXiv:1506.01689](https://arxiv.org/abs/1506.01689).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/Xcts/AdmLinearMomentum.cpp
+++ b/src/PointwiseFunctions/Xcts/AdmLinearMomentum.cpp
@@ -1,0 +1,82 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "PointwiseFunctions/Xcts/AdmLinearMomentum.hpp"
+
+#include <cmath>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Slice.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/IndexToSliceAt.hpp"
+#include "NumericalAlgorithms/LinearOperators/Divergence.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace Xcts {
+
+void adm_linear_momentum_surface_integrand(
+    gsl::not_null<tnsr::II<DataVector, 3>*> result,
+    const Scalar<DataVector>& conformal_factor,
+    const tnsr::II<DataVector, 3>& inv_spatial_metric,
+    const tnsr::II<DataVector, 3>& inv_extrinsic_curvature,
+    const Scalar<DataVector>& trace_extrinsic_curvature) {
+  tenex::evaluate<ti::I, ti::J>(
+      result,
+      1. / (8. * M_PI) * pow<10>(conformal_factor()) *
+          (inv_extrinsic_curvature(ti::I, ti::J) -
+           trace_extrinsic_curvature() * inv_spatial_metric(ti::I, ti::J)));
+}
+
+tnsr::II<DataVector, 3> adm_linear_momentum_surface_integrand(
+    const Scalar<DataVector>& conformal_factor,
+    const tnsr::II<DataVector, 3>& inv_spatial_metric,
+    const tnsr::II<DataVector, 3>& inv_extrinsic_curvature,
+    const Scalar<DataVector>& trace_extrinsic_curvature) {
+  tnsr::II<DataVector, 3> result;
+  adm_linear_momentum_surface_integrand(
+      make_not_null(&result), conformal_factor, inv_spatial_metric,
+      inv_extrinsic_curvature, trace_extrinsic_curvature);
+  return result;
+}
+
+void adm_linear_momentum_volume_integrand(
+    gsl::not_null<tnsr::I<DataVector, 3>*> result,
+    const tnsr::II<DataVector, 3>& surface_integrand,
+    const Scalar<DataVector>& conformal_factor,
+    const tnsr::i<DataVector, 3>& conformal_factor_deriv,
+    const tnsr::ii<DataVector, 3>& conformal_metric,
+    const tnsr::II<DataVector, 3>& inv_conformal_metric,
+    const tnsr::Ijj<DataVector, 3>& conformal_christoffel_second_kind,
+    const tnsr::i<DataVector, 3>& conformal_christoffel_contracted) {
+  tenex::evaluate<ti::I>(
+      result,
+      -(conformal_christoffel_second_kind(ti::I, ti::j, ti::k) *
+            surface_integrand(ti::J, ti::K) +
+        conformal_christoffel_contracted(ti::k) *
+            surface_integrand(ti::I, ti::K) -
+        2. * conformal_metric(ti::j, ti::k) * surface_integrand(ti::J, ti::K) *
+            inv_conformal_metric(ti::I, ti::L) * conformal_factor_deriv(ti::l) /
+            conformal_factor()));
+}
+
+tnsr::I<DataVector, 3> adm_linear_momentum_volume_integrand(
+    const tnsr::II<DataVector, 3>& surface_integrand,
+    const Scalar<DataVector>& conformal_factor,
+    const tnsr::i<DataVector, 3>& conformal_factor_deriv,
+    const tnsr::ii<DataVector, 3>& conformal_metric,
+    const tnsr::II<DataVector, 3>& inv_conformal_metric,
+    const tnsr::Ijj<DataVector, 3>& conformal_christoffel_second_kind,
+    const tnsr::i<DataVector, 3>& conformal_christoffel_contracted) {
+  tnsr::I<DataVector, 3> result;
+  adm_linear_momentum_volume_integrand(
+      make_not_null(&result), surface_integrand, conformal_factor,
+      conformal_factor_deriv, conformal_metric, inv_conformal_metric,
+      conformal_christoffel_second_kind, conformal_christoffel_contracted);
+  return result;
+}
+
+}  // namespace Xcts
--- a/src/PointwiseFunctions/Xcts/AdmLinearMomentum.hpp
+++ b/src/PointwiseFunctions/Xcts/AdmLinearMomentum.hpp
@@ -1,0 +1,103 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Slice.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/IndexToSliceAt.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/Gsl.hpp"
+
+#include <cmath>
+
+#include "NumericalAlgorithms/LinearOperators/Divergence.hpp"
+#include "NumericalAlgorithms/LinearOperators/PartialDerivatives.hpp"
+
+namespace Xcts {
+
+/// @{
+/*!
+ * \brief Surface integrand for ADM linear momentum calculation defined as (see
+ * Eq. 20 in \cite Ossokine2015yla):
+ *
+ * \begin{equation}
+ *   \frac{1}{8\pi} \psi^{10} (K^{ij} - K \gamma^{ij})
+ * \end{equation}
+ *
+ * \param result output buffer for the surface integrand
+ * \param conformal_factor the conformal factor $\psi$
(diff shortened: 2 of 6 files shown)


src/PointwiseFunctions/Xcts/AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,127 @@
...
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Slice.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/IndexToSliceAt.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+
+#include <cmath>
```
nilsvu: Please run clang format over your files, it will sort includes for you etc. An easy way to do this is to run `git-clang-format` in the repo directory. It will format all changed files.

FYI, our style guide is here: https://spectre-code.org/code_review_guide.html

Specifically, please follow the "header order" specified there (move system includes up).

src/PointwiseFunctions/Xcts/AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,127 @@
...
+    const tnsr::II<DataVector, 3>& intermediate_P,
+    const tnsr::ii<DataVector, 3>& conformal_metric,
+    const tnsr::II<DataVector, 3>& inv_conformal_metric,
+    const tnsr::Ijj<DataVector, 3>& conformal_christoffel_second_kind) {
+  const auto first_term = tenex::evaluate<ti::I>(
+      conformal_christoffel_second_kind(ti::I, ti::j, ti::k) *
+      intermediate_P(ti::J, ti::K));
+
+  const auto second_term = tenex::evaluate<ti::I>(
+      conformal_christoffel_second_kind(ti::J, ti::j, ti::k) *
```
nilsvu: Pass this quantity in as `conformal_christoffel_contracted`. It's already available in the XCTS system.

src/PointwiseFunctions/Xcts/AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,127 @@
...
+
+void adm_intermediate_G(
+    gsl::not_null<tnsr::I<DataVector, 3>*> result,
+    const Scalar<DataVector>& conformal_factor,
+    const tnsr::i<DataVector, 3>& conformal_factor_deriv,
+    const tnsr::II<DataVector, 3>& intermediate_P,
+    const tnsr::ii<DataVector, 3>& conformal_metric,
+    const tnsr::II<DataVector, 3>& inv_conformal_metric,
+    const tnsr::Ijj<DataVector, 3>& conformal_christoffel_second_kind) {
+  const auto first_term = tenex::evaluate<ti::I>(
```
nilsvu: I think just inline these three terms in the sum below. That avoids allocating memory for each term. I understand the idea of organizing the code into these three terms, but I think it's also easy enough to understand if the three terms are summed directly and it avoids the memory allocations.

src/PointwiseFunctions/Xcts/AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,127 @@
...
+    const tnsr::II<DataVector, 3>& inv_spatial_metric,
+    const tnsr::II<DataVector, 3>& inv_extrinsic_curvature,
+    const Scalar<DataVector>& trace_extrinsic_curvature) {
+  tnsr::II<DataVector, 3> intermediate_P;
+  adm_intermediate_P(make_not_null(&intermediate_P), conformal_factor,
+                     inv_extrinsic_curvature, inv_spatial_metric,
+                     trace_extrinsic_curvature);
+
+  tenex::evaluate<ti::I, ti::J>(
+      result, 1. / (8. * M_PI) * intermediate_P(ti::I, ti::J));
```
nilsvu: All this does is multiply by 1/8pi. I suggest you eliminate the `adm_intermediate_P` function entirely and just compute the surface integrand including the 1/8pi here. Then you can use the surface integrand also in the computation of the volume integrand (see below).

src/PointwiseFunctions/Xcts/AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,127 @@
...
+    gsl::not_null<tnsr::I<DataVector, 3>*> result,
+    const Scalar<DataVector>& conformal_factor,
+    const tnsr::i<DataVector, 3>& conformal_factor_deriv,
+    const tnsr::ii<DataVector, 3>& conformal_metric,
+    const tnsr::II<DataVector, 3>& inv_conformal_metric,
+    const tnsr::II<DataVector, 3>& inv_spatial_metric,
+    const tnsr::II<DataVector, 3>& inv_extrinsic_curvature,
+    const Scalar<DataVector>& trace_extrinsic_curvature,
+    const tnsr::Ijj<DataVector, 3>& conformal_christoffel_second_kind) {
+  tnsr::II<DataVector, 3> intermediate_P;
```
nilsvu: Take the surface integrand as an argument to this function. Then you don't need the extra `adm_intermediate_G` function but you just compute the volume integrand directly here.

tests/Unit/PointwiseFunctions/Xcts/Test_AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,233 @@
...
+                     Spectral::Quadrature::GaussLobatto};
+
+  // Initialize surface integral
+  tnsr::I<double, 3> surface_integral;
+  for (int I = 0; I < 3; I++) {
+    surface_integral.get(I) = 0.;
+  }
+
+  // Compute integrals by summing over each element
+  for (const auto& element_id : element_ids) {
```
nilsvu, line 88: Iterate only over elements at the outer boundary to avoid unnecessary work that probably slows down this test a lot (at least for high h-refinement). One way to do this is:

```cpp
const size_t radial_dimension = 2;
const auto radial_segment_id = element_id.segment_id(radial_dimension);
if (radial_segment_id.index() != two_to_the(radial_segment_id.refinement_level()) - 1) {
  continue;
}
```

tests/Unit/PointwiseFunctions/Xcts/Test_AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    // Get coordinates and map
+    const auto logical_coords = logical_coordinates(mesh);
+    const ElementMap<3, Frame::Inertial> logical_to_inertial_map(
+        element_id, current_block.stationary_map().get_clone());
+    const tnsr::I<DataVector, 3> inertial_coords =
+        logical_to_inertial_map(logical_coords);
+    const auto inv_jacobian =
+        logical_to_inertial_map.inv_jacobian(logical_coords);
+    const auto det_inv_jacobian = determinant(inv_jacobian);
```
nilsvu: I think you don't need these quantities (except for the ElementMap) because you only need fields on the element faces, and you don't need to take derivatives in the volume.

tests/Unit/PointwiseFunctions/Xcts/Test_AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,233 @@
...
+
+    // Get required fields
+    const auto background_fields = solution.variables(
```
nilsvu: You can move this into the loop over faces and only evaluate it on the face. You can remove the `mesh` and `inv_jacobian` arguments because you don't need to take derivatives (only second derivatives of the metric like the Ricci tensor are computed numerically because they are so complicated analytically).

tests/Unit/PointwiseFunctions/Xcts/Test_AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,233 @@
...
+    tenex::evaluate<ti::I, ti::J>(make_not_null(&inv_extrinsic_curvature),
+                                  inv_spatial_metric(ti::I, ti::K) *
+                                      inv_spatial_metric(ti::J, ti::L) *
+                                      extrinsic_curvature(ti::k, ti::l));
+
+    for (auto boundary_direction : current_element.external_boundaries()) {
+      const auto face_logical_coords =
+          interface_logical_coordinates(face_mesh, boundary_direction);
+      const auto direction = Direction<3>::upper_zeta();
+      const size_t slice_index = index_to_slice_at(mesh.extents(), direction);
```
nilsvu: As noted above you don't need to slice, but you can evaluate the solution on the face directly and not in the volume.

tests/Unit/PointwiseFunctions/Xcts/Test_AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,233 @@
...
+    tnsr::II<DataVector, 3> inv_extrinsic_curvature;
+    tenex::evaluate<ti::I, ti::J>(make_not_null(&inv_extrinsic_curvature),
+                                  inv_spatial_metric(ti::I, ti::K) *
+                                      inv_spatial_metric(ti::J, ti::L) *
+                                      extrinsic_curvature(ti::k, ti::l));
+
+    for (auto boundary_direction : current_element.external_boundaries()) {
+      const auto face_logical_coords =
+          interface_logical_coordinates(face_mesh, boundary_direction);
+      const auto direction = Direction<3>::upper_zeta();
```
nilsvu: This should be `boundary_direction`. Probably best to add a `CHECK(boundary_direction == Direction<3>::upper_zeta())` and then use only one variable.

tests/Unit/PointwiseFunctions/Xcts/Test_AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,233 @@
...
+        face_normal.get(d) /= get(face_normal_magnitude);
+      }
+
+      // Compute surface integrand
+      const auto surface_integrand =
+          Xcts::adm_linear_momentum_surface_integrand(
+              face_conformal_factor, face_inv_spatial_metric,
+              face_inv_extrinsic_curvature, face_trace_extrinsic_curvature);
+      // We need the negative sign because the face normal points out of
+      // the element (into the excision)
```
nilsvu: Delete comment

nilsvu (changes requested):

src/PointwiseFunctions/Xcts/AdmLinearMomentum.hpp
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * Note that we are including the negative sign in the integrand.
+ *
+ * \param result output buffer for the surface integrand
+ * \param surface_integrand the result of
+ * `adm_linear_momentum_surface_integrand`
+ * \param conformal_factor the conformal factor $\psi$
+ * \param conformal_factor_deriv the derivative of the conformal factor
+ * $\partial_i\psi$
+ * \param conformal_metric the conformal metric $\tilde\gamma_{ij}$
```
nilsvu: We denote this `\bar` not tilde

src/PointwiseFunctions/Xcts/AdmLinearMomentum.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * \param inv_conformal_metric the inverse conformal metric curvature $K$
```
nilsvu: Fix this sentence

nilsvu (changes requested):
Looks good! Please squash, including these small changes

src/PointwiseFunctions/Xcts/AdmLinearMomentum.hpp
```diff
@@ -0,0 +1,103 @@
...
+ *
+ * \begin{equation}
+ *   - \frac{1}{8\pi} (
+ *       \tilde\Gamma^i_{jk} P^{jk}
+ *       + \tilde\Gamma^j_{jk} P^{jk}
+ *       - 2 \tilde\gamma_{jk} P^{jk} \tilde\gamma^{il} \partial_l(\ln\psi)
+ *   ),
+ * \end{equation}
+ *
+ * where \frac{1}{8\pi} P^{jk} is the result from
```
nilsvu: Missing $

nilsvu (changes requested):

src/PointwiseFunctions/Xcts/AdmLinearMomentum.cpp
```diff
@@ -0,0 +1,82 @@
...
+
+#include "PointwiseFunctions/Xcts/AdmLinearMomentum.hpp"
+
+#include <cmath>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Slice.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/IndexToSliceAt.hpp"
```
nilsvu, line 12: You can delete a bunch of these includes next time you touch this file

nilsvu (approved):

PR #6147, opened 2024-07-10
Make MortarData a simple struct

kidder:
## Proposed changes

- Introduce class MortarDataHolder that holds two MortarData to make GTS and LTS code more similar. (For LTS BoundaryHistory already holds MortarData on each side of the mortar.)
- Change MortarData to hold data on only one side of the mortar.  
- Turn MortarData into a simple struct in order to make it easier to use with AMR

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DiscontinuousGalerkin/MortarData.hpp
+++ b/src/Evolution/DiscontinuousGalerkin/MortarData.hpp
@@ -21,14 +21,15 @@ namespace evolution::dg {
  * \brief Data on the mortar used to compute the boundary correction for the
  * DG scheme.
  *
- * The class holds the local data that has been projected to the mortar as well
- * as the neighbor data that has been projected to the mortar. The local and
- * neighbor data is later used to compute the same unique boundary correction on
- * the mortar for both elements. That is, the final boundary correction
+ * The class holds the data that has been projected to one side of the mortar.
+ * It is meant to be used in a container (either MortarDataHolder or
+ * TimeSteppers::BoundaryHistory) that holds MortarData on each side of the
+ * mortar. The data is later used to compute the same unique boundary correction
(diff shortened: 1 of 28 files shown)


src/Evolution/DiscontinuousGalerkin/MortarData.hpp
```diff
@@ -29,117 +29,48 @@ namespace evolution::dg {
...
-
-  auto mortar_data() -> std::optional<std::pair<Mesh<Dim - 1>, DataVector>>& {
-    return mortar_data_;
-  }
-  /// @}
+struct MortarData {
+  std::optional<DataVector> mortar_data{std::nullopt};
+  std::optional<Scalar<DataVector>> face_normal_magnitude{std::nullopt};
+  std::optional<Scalar<DataVector>> face_det_jacobian{std::nullopt};
+  std::optional<Scalar<DataVector>> volume_det_inv_jacobian{std::nullopt};
```
nilsdeppe, line 64: These were intentionally 1 allocation because doing allocations on the faces is extremely expensive. They scale with the number of faces and this is really bad. We need some way of keeping it to 1 allocation. Ideally it would be 1 allocation per process per element face, but that'll be harder. I'm not sure what to recommend in terms of design because for nodegroups another optimization is sharing the computation of the Riemann problem between elements on the same node and/or sharing the input in a single memory allocation. From what I've heard this will be even more important on GPUs... I'm not sure if it would be easier to discuss on a call?

Having said that, this impacts Gauss, not GL, though I'm still not convinced that GL is better than G.

nilsdeppe (changes requested):

nilsdeppe (approved):

PR #6149, opened 2024-07-11
Add some pybindings

nilsdeppe:
## Proposed changes

I've been using these to look at some BBH runs to compute quantities without rerunning.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/Tensor/EagerMath/Python/Determinant.cpp
+++ b/src/DataStructures/Tensor/EagerMath/Python/Determinant.cpp
@@ -8,6 +8,7 @@
 
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/EagerMath/Determinant.hpp"
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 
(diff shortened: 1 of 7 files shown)


src/Visualization/Python/TransformVolumeData.py
```diff
@@ -218,6 +220,8 @@ def one_index_constraint(
 
     if arg.annotation in Mesh.values():
         return ElementArg("mesh")
+    elif arg.name == "element":
```
nilsvu: For my understanding, what are you trying to do with this? Note that this returns the Python-representation of an element in a volume data file (`IterElements.Element`), not the C++ class `Element` as you might expect. If you want to expose another property of the `IterElements.Element` class, maybe just add that specific one here, as is already done with coords, jacobians, etc?

Either way, please add this to the docstring.

Edit: I didn't read the commit message before. I understand now that you're trying to filter on the element ID. I think this can be accomplished by adding the `element_patterns` argument to the two calls of `iter_elements` in `TransformVolumeData.py`, but it also requires adding support for writing to a new volfile (with fewer elements) to `TransformVolumeData.py`, because I think currently it can only add tensor components of the same length to the existing volume data. So I'm curious how you accomplished this by filtering on the Element ID within the kernel function.

nilsdeppe: Thanks for the info! I ended up not really needing this (yet). I'll leave it on a branch and we can figure out a good way to handle some filtering. The other possible use-case I had in mind is to get use the element->inertial maps, but then I realized that also needs the domain, etc. and just didn't/don't have time to figure that out right now. I'll drop it and bring it back once I actually have time to get it into a clean state :)

nilsdeppe:
I've rebased, dropped the element commit, and added a commit that allows `print(mesh)` to call C++ printing.

nilsvu (approved):

PR #6151, opened 2024-07-14
Observe ADM linear momentum in SolveXcts executable.

iago-mendes:
## Proposed changes

Compute the total ADM linear momentum integral after an XCTS solve. This creates a new dataset `/AdmIntegrals.dat` inside `BbhReductions.h5`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
+++ b/src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
@@ -1,0 +1,110 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.hpp"
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/AreaElement.hpp"
+#include "Domain/Tags.hpp"
+#include "NumericalAlgorithms/DiscontinuousGalerkin/ProjectToBoundary.hpp"
+#include "NumericalAlgorithms/LinearOperators/DefiniteIntegral.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "PointwiseFunctions/Xcts/AdmLinearMomentum.hpp"
+
+namespace Events {
+
+void local_adm_integrals(
+    gsl::not_null<tnsr::I<double, 3>*> result,
+    const Scalar<DataVector>& conformal_factor,
+    const tnsr::ii<DataVector, 3>& spatial_metric,
+    const tnsr::II<DataVector, 3>& inv_spatial_metric,
+    const tnsr::ii<DataVector, 3>& extrinsic_curvature,
+    const Scalar<DataVector>& trace_extrinsic_curvature,
+    const InverseJacobian<DataVector, 3, Frame::ElementLogical,
+                          Frame::Inertial>& inv_jacobian,
+    const Mesh<3>& mesh, const Element<3>& element,
+    const DirectionMap<3, tnsr::i<DataVector, 3>>& conformal_face_normals) {
+  // Initialize integral to 0
+  for (int I = 0; I < 3; I++) {
+    result->get(I) = 0;
+  }
+
+  // Skip elements not at the outer boundary
+  const size_t radial_dimension = 2;
+  const auto radial_segment_id = element.id().segment_id(radial_dimension);
+  if (radial_segment_id.index() !=
+      two_to_the(radial_segment_id.refinement_level()) - 1) {
+    return;
+  }
+
+  // Loop over external boundaries
+  for (const auto boundary_direction : element.external_boundaries()) {
+    // Skip interfaces not at the outer boundary
+    if (boundary_direction != Direction<3>::upper_zeta()) {
+      continue;
+    }
+
+    // Project fields to the boundary
+    const auto face_conformal_factor = dg::project_tensor_to_boundary(
+        conformal_factor, mesh, boundary_direction);
+    const auto face_spatial_metric = dg::project_tensor_to_boundary(
+        spatial_metric, mesh, boundary_direction);
+    const auto face_inv_spatial_metric = dg::project_tensor_to_boundary(
+        inv_spatial_metric, mesh, boundary_direction);
+    const auto face_extrinsic_curvature = dg::project_tensor_to_boundary(
+        extrinsic_curvature, mesh, boundary_direction);
+    const auto face_trace_extrinsic_curvature = dg::project_tensor_to_boundary(
+        trace_extrinsic_curvature, mesh, boundary_direction);
+    // This projection could be avoided by using
+    // domain::Tags::DetSurfaceJacobian from the DataBox, which is computed
+    // directly on the face (not projected). That would be better on Gauss
+    // meshes that have no grid point at the boundary. The DetSurfaceJacobian
+    // could then be multiplied by the (conformal) metric determinant to form
+    // the area element. Note that the DetSurfaceJacobian is computed using the
+    // conformal metric.
+    const auto face_inv_jacobian =
+        dg::project_tensor_to_boundary(inv_jacobian, mesh, boundary_direction);
+
+    // Compute the inverse extrinsic curvature
+    tnsr::II<DataVector, 3> face_inv_extrinsic_curvature;
+    tenex::evaluate<ti::I, ti::J>(make_not_null(&face_inv_extrinsic_curvature),
+                                  face_inv_spatial_metric(ti::I, ti::K) *
+                                      face_inv_spatial_metric(ti::J, ti::L) *
+                                      face_extrinsic_curvature(ti::k, ti::l));
+
+    // Compute curved area element
+    const auto face_sqrt_det_spatial_metric =
+        Scalar<DataVector>(sqrt(get(determinant(face_spatial_metric))));
+    const auto curved_area_element =
+        area_element(face_inv_jacobian, boundary_direction,
+                     face_inv_spatial_metric, face_sqrt_det_spatial_metric);
+
+    // Get face mesh and face normal from data box
+    const auto& face_mesh = mesh.slice_away(boundary_direction.dimension());
+    // Note that this face normal is related to the conformal metric. To get
+    // the non-conformal face normal, we multiply by the square of the
+    // conformal factor in the computation of 'contracted_integrand'.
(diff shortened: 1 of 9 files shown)


src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -0,0 +1,102 @@
...
+#include "Domain/AreaElement.hpp"
+#include "Domain/Tags.hpp"
+#include "NumericalAlgorithms/DiscontinuousGalerkin/ProjectToBoundary.hpp"
+#include "NumericalAlgorithms/LinearOperators/DefiniteIntegral.hpp"
+#include "NumericalAlgorithms/Spectral/Mesh.hpp"
+#include "PointwiseFunctions/Xcts/AdmLinearMomentum.hpp"
+
+namespace Events {
+
+void compute_linear_momentum(
```
nilsvu: We don't generally prefix function names with `compute_` because they better compute something :D So we don't add this prefix to every function. I suggest you name this `local_adm_integrals` and add the other integrals here as well once they're implemented. The "local" refers to the fact that it's computed only over this element, and needs to be summed over in a reduction. Please mention this in the docs as well.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -0,0 +1,102 @@
...
+    gsl::not_null<tnsr::I<double, 3>*> result,
+    const Scalar<DataVector>& conformal_factor,
+    const tnsr::ii<DataVector, 3>& spatial_metric,
+    const tnsr::II<DataVector, 3>& inv_spatial_metric,
+    const tnsr::ii<DataVector, 3>& extrinsic_curvature,
+    const Scalar<DataVector>& trace_extrinsic_curvature,
+    const InverseJacobian<DataVector, 3, Frame::ElementLogical,
+                          Frame::Inertial>& inv_jacobian,
+    const Mesh<3>& mesh, const Element<3>& element,
+    const DirectionMap<3, tnsr::i<DataVector, 3>>& faces_map_normal) {
```
nilsvu: I suggest you name this argument `conformal_face_normals` to be more clear.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -0,0 +1,102 @@
...
+  // Skip elements not at the outer boundary
+  const size_t radial_dimension = 2;
+  const auto radial_segment_id = element.id().segment_id(radial_dimension);
+  if (radial_segment_id.index() !=
+      two_to_the(radial_segment_id.refinement_level()) - 1) {
+    return;
+  }
+
+  // Loop over external boundaries
+  for (auto boundary_direction : element.external_boundaries()) {
```
nilsvu: `const auto`

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -0,0 +1,102 @@
...
+    // Skip interfaces not at the outer boundary
+    if (boundary_direction != Direction<3>::upper_zeta()) {
+      continue;
+    }
+
+    // Project fields to the boundary
+    const auto& face_conformal_factor = dg::project_tensor_to_boundary(
```
nilsvu: `const auto` (no `&`) for all of these. Not sure if it actually makes a difference for the compiler here, but it makes it clear to the reader that these are all memory allocations. I think that's fine because this is only for observation and not performance critical.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -0,0 +1,102 @@
...
+    const auto& face_spatial_metric = dg::project_tensor_to_boundary(
+        spatial_metric, mesh, boundary_direction);
+    const auto& face_inv_spatial_metric = dg::project_tensor_to_boundary(
+        inv_spatial_metric, mesh, boundary_direction);
+    const auto& face_extrinsic_curvature = dg::project_tensor_to_boundary(
+        extrinsic_curvature, mesh, boundary_direction);
+    const auto& face_trace_extrinsic_curvature = dg::project_tensor_to_boundary(
+        trace_extrinsic_curvature, mesh, boundary_direction);
+    const auto& face_inv_jacobian =
+        dg::project_tensor_to_boundary(inv_jacobian, mesh, boundary_direction);
```
nilsvu, line 68: Please add a note here that this projection could be avoided by using the `domain::Tags::DetSurfaceJacobian` that's in the DataBox, which is computed directly on the face (not projected). That would be better on Gauss meshes that have no grid point at the boundary.
The `DetSurfaceJacobian` can then be multiplied by the (conformal) metric determinant to form the area element. Note that the `DetSurfaceJacobian` is computed using the _conformal_ metric.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.hpp
```diff
@@ -0,0 +1,192 @@
...
+    for (int I = 0; I < 3; I++) {
+      linear_momentum.get(I) = 0;
+    }
+    compute_linear_momentum(make_not_null(&linear_momentum), conformal_factor,
+                            spatial_metric, inv_spatial_metric,
+                            extrinsic_curvature, trace_extrinsic_curvature,
+                            inv_jacobian, mesh, element, faces_map_normal);
+
+    // Save components of linear momentum as reduction data
+    ReductionData reduction_data{linear_momentum.get(0), linear_momentum.get(1),
```
nilsvu: Use the version `get<0>(linear_momentum)` when you know the index at compile time.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.hpp
```diff
@@ -0,0 +1,192 @@
...
+                                 linear_momentum.get(2)};
+    std::vector<std::string> legend{
+        "ADMLinearMomentum_x", "ADMLinearMomentum_y", "ADMLinearMomentum_z"};
```
nilsvu: Do full CamelCase: `AdmLinearMomentum_x`

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
```diff
@@ -0,0 +1,163 @@
...
+                                  const size_t polynomial_order) {
+  // Define black hole parameters
+  const double mass = 1;
+  const double boost_speed = 0.5;
+  const double lorentz_factor = 1. / sqrt(1. - square(boost_speed));
+
+  // Get Kerr-Schild solution
+  const std::array<double, 3> dimensionless_spin{{0., 0., 0.}};
+  const std::array<double, 3> boost_velocity{{0., 0., boost_speed}};
+  const KerrSchild solution{mass, dimensionless_spin,
```
nilsvu: Please also test this with a nontrivial conformal factor. `WrappedGr` sets $\psi=1$ (see its docs), so any dependence on the conformal decomposition is not yet tested here and the powers of the conformal factor are very easy to get wrong.

One way to test this is by using a Schwarzschild solution in, e.g., isotropic coordinates (see `src/PointwiseFunctions/AnalyticSolutions/Xcts/Schwarzschild.hpp`) and apply a Lorentz boost to all quantities (see e.g. `test_boosted_spacetime_metric` in `tests/Unit/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/Test_KerrSchild.cpp`).

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
```diff
@@ -0,0 +1,163 @@
...
+        get<Xcts::Tags::InverseConformalMetric<DataVector, 3, Frame::Inertial>>(
+            background_fields);
+    const auto& extrinsic_curvature =
+        get<gr::Tags::ExtrinsicCurvature<DataVector, 3, Frame::Inertial>>(
+            background_fields);
+    const auto& trace_extrinsic_curvature =
+        get<gr::Tags::TraceExtrinsicCurvature<DataVector>>(background_fields);
+
+    // Compute face normal (related to the conformal metric)
+    auto direction = Direction<3>::upper_zeta();
```
nilsvu, line 122: I think you're also integrating over internal element boundaries here. Shouldn't the test fail because of that? I may be missing something.

nilsvu (changes requested):

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
iago-mendes: It's my understanding that the reduction will call `Events::local_adm_integrals` on every element (even internal elements). Then, within that function, I handle which elements we should skip. Here, my intention was to simulate how the reduction will call that function (on every element). Am I understanding it wrong?

nilsvu: Yes that sounds good, thanks for clarifying 👍

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
iago-mendes: I believe it is a good idea to work on this test in a separate PR. I've added my partial implementation in this branch: https://github.com/iago-mendes/spectre/tree/boosted-schwarzschild

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.hpp
```diff
@@ -0,0 +1,194 @@
...
+                          Frame::Inertial>& inv_jacobian,
+    const Mesh<3>& mesh, const Element<3>& element,
+    const DirectionMap<3, tnsr::i<DataVector, 3>>& conformal_face_normals);
+/// @}
+
+/// @{
+/*!
+ * \brief Observe ADM integrals after the XCTS solve.
+ *
+ * Writes reduction quantities:
```
nilsvu, line 65: Please add that this event takes a surface integral over the outer boundary, which is defined as the domain boundary in upper logical zeta direction.

nilsvu (changes requested):
Looks good, please squash including this

nilsvu (approved):

PR #6170, opened 2024-07-19
Remove some old clusters

knelli2:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


tests/Unit/Parallel/Test_AlgorithmReduction.cpp
```diff
@@ -42,11 +42,11 @@ class DataBox;
 }  // namespace db
 struct TestMetavariables;
 
-// The reason we use a 46 element array is that on Wheeler, the SXS
-// supercomputer at Caltech, there are 23 worker threads per node and we want to
-// be able to test on two nodes to make sure multinode communication is working
-// correctly.
-static constexpr int number_of_1d_array_elements = 46;
+// The reason we use a 54 element array is that on Caltech HPC, the
```
nilsdeppe: I would just leave the number as-is and change the comment to be something system agnostic or change the number to 62 since 32 cores per node is more common than 28.

nilsdeppe (changes requested):

nilsdeppe (approved):

PR #6172, opened 2024-07-20
Add nodegroup support to control system and interpolation framework

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Actions/FunctionsOfTimeAreReady.hpp
+++ b/src/ParallelAlgorithms/Actions/FunctionsOfTimeAreReady.hpp
@@ -16,10 +16,14 @@
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
 #include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/ArrayCollection/IsDgElementCollection.hpp"
+#include "Parallel/ArrayCollection/PerformAlgorithmOnElement.hpp"
+#include "Parallel/ArrayCollection/Tags/ElementLocations.hpp"
 #include "Parallel/ArrayComponentId.hpp"
 #include "Parallel/Callback.hpp"
 #include "Parallel/GlobalCache.hpp"
 #include "Parallel/ParallelComponentHelpers.hpp"
+#include "ParallelAlgorithms/Actions/GetItemFromDistributedObject.hpp"
 #include "Utilities/Algorithm.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/StdHelpers.hpp"
@@ -50,7 +54,27 @@ bool functions_of_time_are_ready_impl(
     const auto& proxy =
         ::Parallel::get_parallel_component<Component>(cache)[array_index];
     const Parallel::ArrayComponentId array_component_id =
-        Parallel::make_array_component_id<Component>(array_index);
+        [&]() -> Parallel::ArrayComponentId {
+      if constexpr (Parallel::is_dg_element_collection_v<Component>) {
+        static_assert(
+            sizeof...(args) == 1,
+            "This currently assumes the only argument is the ElementId. If you "
+            "need extra arguments, this can be generalized. We need the "
+            "ElementId instead of the array_index (which is the nodegroup ID) "
+            "since each ArrayComponentId is only allowed to register one "
+            "callback, additional ones are ignored. This means of a "
+            "DgElementCollection only 1 callback _per node_ would be "
+            "registered, while we need 1 callback for each element. An "
+            "alternative approach would be to have the callback be a broadcast "
+            "to all elements instead of one specific one.");
+        static_assert((std::is_same_v<ElementId<Metavariables::volume_dim>,
+                                      std::decay_t<Args>> and
+                       ...));
+        return Parallel::make_array_component_id<Component>(args...);
+      } else {
+        return Parallel::make_array_component_id<Component>(array_index);
+      }
+    }();
 
     return Parallel::mutable_cache_item_is_ready<CacheTag>(
         cache, array_component_id,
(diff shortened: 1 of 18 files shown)


src/ParallelAlgorithms/Actions/FunctionsOfTimeAreReady.hpp
```diff
@@ -50,7 +54,24 @@ bool functions_of_time_are_ready_impl(
     const auto& proxy =
         ::Parallel::get_parallel_component<Component>(cache)[array_index];
     const Parallel::ArrayComponentId array_component_id =
-        Parallel::make_array_component_id<Component>(array_index);
+        [&]() -> Parallel::ArrayComponentId {
+      if constexpr (Parallel::is_dg_element_collection_v<Component>) {
+        static_assert(
+            sizeof...(args) == 1,
```
knelli2, line 60: I think we should also `static_assert` after this that the single arg is in fact an `ElementId`.

src/ParallelAlgorithms/Interpolation/Actions/ElementReceiveInterpPoints.hpp
```diff
@@ -31,12 +31,12 @@ namespace Actions {
...
 struct ElementReceiveInterpPoints {
   template <typename ParallelComponent, typename DbTags, typename Metavariables,
-            typename ArrayIndex>
+            typename ArrayIndex, size_t Dim>
   static void apply(
       db::DataBox<DbTags>& box,
       const Parallel::GlobalCache<Metavariables>& /*cache*/,
       const ArrayIndex& /*array_index*/,
-      tnsr::I<DataVector, Metavariables::volume_dim,
+      tnsr::I<DataVector, Dim,
```
knelli2, line 39: I had added comments for the previous commits asking to do the opposite of this. Could you give a brief explanation? Was this necessary to compile?

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorRegisterElement.hpp
```diff
@@ -111,11 +113,26 @@ struct RegisterElementWithInterpolator
...
   static void register_or_deregister_impl(
       Parallel::GlobalCache<Metavariables>& cache,
-      const ArrayIndex& /*array_index*/) {
-    auto& interpolator = *Parallel::local_branch(
-        Parallel::get_parallel_component<::intrp::Interpolator<Metavariables>>(
-            cache));
-    Parallel::simple_action<RegisterOrDeregisterAction>(interpolator);
+      const ArrayIndex& array_index) {
+    // ERROR("We need to get the correct CoreID if doing nodegroups. "
+    //       << array_index);
```
knelli2: Remove?

src/ParallelAlgorithms/Interpolation/Interpolate.hpp
```diff
@@ -51,11 +52,20 @@ void interpolate(
       });
 
   // Send volume data to the Interpolator, to trigger interpolation.
-  auto& interpolator = *Parallel::local_branch(
-      ::Parallel::get_parallel_component<Interpolator<Metavariables>>(cache));
-  Parallel::simple_action<Actions::InterpolatorReceiveVolumeData<
-      typename InterpolationTargetTag::temporal_id>>(
-      interpolator, temporal_id, array_index, mesh, interpolator_source_vars);
+  if (interpolator_id.has_value()) {
+    auto interpolator =
```
knelli2, line 62: Could you add a comment about what the `interpolator_id` is/represents? I think that'll be helpful for future people (me) editing this.

knelli2 (changes requested):

src/ParallelAlgorithms/Interpolation/Actions/ElementReceiveInterpPoints.hpp
nilsdeppe: We've generally found "reaching up" into the metavariables to be fairly brittle and tricky. Considering we deduce the dim pretty much everywhere in the code I saw no reason to avoid doing that here. I've also sometimes had issues with GCC not being able to retrieve `Metavaraibles::volume_dim` and those were extremely difficult to debug, so I feel like suggesting that as the "default" is not really a good thing.

knelli2: I agree in general we shouldn't reach up, but I thought the `volume_dim` was an exception since it's so ubiquitous in our metavars. We also do this a lot throughout the code for the dim, so we aren't really consistent. But if it's causing issues with GCC then passing in the Dim explicitly is ok

nilsdeppe:
@knelli2 I pushed fixups. Thanks for the review :)

src/ParallelAlgorithms/Interpolation/Events/Interpolate.hpp
```diff
@@ -89,8 +94,21 @@ class Interpolate<VolumeDim, InterpolationTargetTag,
     static_assert(
         std::is_same_v<typename Metavariables::interpolator_source_vars,
                        tmpl::list<InterpolatorSourceVarTags...>>);
-    interpolate<InterpolationTargetTag>(temporal_id, mesh, cache, array_index,
-                                        interpolator_source_vars...);
+    if constexpr (Parallel::is_dg_element_collection_v<ParallelComponent>) {
+      const auto core_id = static_cast<int>(
+          Parallel::local_synchronous_action<
```
knelli2, line 99: Not a change request, but will this impact performance? This runs on every element any time an interpolation is triggered which means it'll keep locking the node group box to get the element collection tag.

knelli2 (approved):
Looks good. You can squash

src/ParallelAlgorithms/Interpolation/Events/Interpolate.hpp
nilsdeppe: local synch actions don't lock

nilsdeppe:
Rebased and squashed. Thanks for the reviews!

src/ParallelAlgorithms/Interpolation/Events/Interpolate.hpp
knelli2: Oh right it just passes the node lock along, and then the action can decide to actually lock it. Forgot about that

PR #6174, opened 2024-07-22
Add FromEos Option to ElectronFraction in Spec Initial Data

ctrandrsn:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticData/GrMhd/SpecInitialData.cpp
+++ b/src/PointwiseFunctions/AnalyticData/GrMhd/SpecInitialData.cpp
@@ -5,6 +5,7 @@
 
 #include <Exporter.hpp>  // The SpEC Exporter
 #include <memory>
+#include <optional>
 #include <pup.h>
 #include <string>
 #include <utility>
@@ -21,6 +22,7 @@
 #include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Serialization/PupStlCpp17.hpp"
 #include "Utilities/System/ParallelInfo.hpp"
 #include "Utilities/TaggedTuple.hpp"
 
@@ -30,7 +32,7 @@ template <size_t ThermodynamicDim>
 SpecInitialData<ThermodynamicDim>::SpecInitialData(
     std::string data_directory,
     std::unique_ptr<equation_of_state_type> equation_of_state,
-    const double density_cutoff, const double electron_fraction)
+    const double density_cutoff, const std::optional<double> electron_fraction)
     : data_directory_(std::move(data_directory)),
       equation_of_state_(std::move(equation_of_state)),
       density_cutoff_(density_cutoff),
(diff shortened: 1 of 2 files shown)


src/PointwiseFunctions/AnalyticData/GrMhd/SpecInitialData.cpp
```diff
@@ -30,7 +32,7 @@ template <size_t ThermodynamicDim>
 SpecInitialData<ThermodynamicDim>::SpecInitialData(
     std::string data_directory,
     std::unique_ptr<equation_of_state_type> equation_of_state,
-    const double density_cutoff, const double electron_fraction)
+    const double density_cutoff, std::optional<double> electron_fraction)
```
nilsdeppe: `const std::optional`?

src/PointwiseFunctions/AnalyticData/GrMhd/SpecInitialData.cpp
```diff
@@ -280,10 +282,27 @@ template <size_t ThermodynamicDim>
...
     const gsl::not_null<Scalar<DataType>*> electron_fraction,
-    const gsl::not_null<Cache*> /*cache*/,
+    const gsl::not_null<Cache*> cache,
     hydro::Tags::ElectronFraction<DataType> /*meta*/) const {
-  std::fill(electron_fraction->begin(), electron_fraction->end(),
-            constant_electron_fraction);
+  const auto& temperature =
+      cache->get_var(*this, hydro::Tags::Temperature<DataType>{});
+  const auto& rest_mass_density =
+      get<hydro::Tags::RestMassDensity<DataType>>(interpolated_data);
```
nilsdeppe, line 290: I wonder if all these calls should also instead grab tho the cache? That way if the function that computes rho does some atmosphere flooring in the future everywhere else has that applied too.

src/PointwiseFunctions/AnalyticData/GrMhd/SpecInitialData.cpp
```diff
@@ -280,10 +282,27 @@ template <size_t ThermodynamicDim>
...
+  const size_t num_points = get_size(get(rest_mass_density));
+  for (size_t i = 0; i < num_points; ++i) {
+     if (electron_fraction_value.has_value()) {
```
nilsdeppe: Why do this check inside the for loop? Here you are setting all grid points to the constant value `num_points` times. If you put the `if` outside the for loop and have the for look just in the `else` then you don't do the redundant work.

nilsdeppe (changes requested):

src/PointwiseFunctions/AnalyticData/GrMhd/SpecInitialData.cpp
```diff
@@ -280,10 +282,27 @@ template <size_t ThermodynamicDim>
...
-            constant_electron_fraction);
+  const auto& temperature =
+      cache->get_var(*this, hydro::Tags::Temperature<DataType>{});
+  const auto& rest_mass_density =
+      get<hydro::Tags::RestMassDensity<DataType>>(interpolated_data);
+  const size_t num_points = get_size(get(rest_mass_density));
+  if (electron_fraction_value.has_value()) {
+    std::fill(electron_fraction->begin(), electron_fraction->end(),
+              electron_fraction_value.value());
+    } else {
```
nilsdeppe: This looks over-indented. Could you run clangformat over this section?

nilsdeppe (approved):

PR #6177, opened 2024-07-25
Test ADM integrals with a boosted Schwarzschild solution.

iago-mendes:
## Proposed changes

This PR tests the ADM mass and ADM linear momentum calculations with a spacetime solution that has a conformal factor different than 1. 

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
+++ b/tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
@@ -6,71 +6,89 @@
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/EagerMath/Determinant.hpp"
 #include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
-#include "DataStructures/Tensor/Slice.hpp"
+#include "DataStructures/Tensor/EagerMath/Trace.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Domain/AreaElement.hpp"
-#include "Domain/CoordinateMaps/Composition.hpp"
 #include "Domain/CoordinateMaps/CoordinateMap.hpp"
-#include "Domain/CoordinateMaps/CoordinateMap.tpp"
-#include "Domain/CoordinateMaps/Wedge.hpp"
 #include "Domain/CreateInitialElement.hpp"
 #include "Domain/Creators/Sphere.hpp"
 #include "Domain/ElementMap.hpp"
 #include "Domain/ElementToBlockLogicalMap.hpp"
 #include "Domain/FaceNormal.hpp"
-#include "Domain/InterfaceLogicalCoordinates.hpp"
-#include "Domain/Structure/IndexToSliceAt.hpp"
 #include "Domain/Structure/InitialElementIds.hpp"
 #include "Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.hpp"
 #include "NumericalAlgorithms/DiscontinuousGalerkin/ProjectToBoundary.hpp"
 #include "NumericalAlgorithms/LinearOperators/DefiniteIntegral.hpp"
 #include "NumericalAlgorithms/Spectral/LogicalCoordinates.hpp"
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
-#include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrHorizon.hpp"
-#include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp"
-#include "PointwiseFunctions/AnalyticSolutions/Xcts/WrappedGr.hpp"
+#include "PointwiseFunctions/AnalyticSolutions/Xcts/Schwarzschild.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Christoffel.hpp"
+#include "PointwiseFunctions/GeneralRelativity/ExtrinsicCurvature.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Lapse.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Shift.hpp"
+#include "PointwiseFunctions/GeneralRelativity/SpacetimeMetric.hpp"
+#include "PointwiseFunctions/GeneralRelativity/SpatialMetric.hpp"
+#include "PointwiseFunctions/SpecialRelativity/LorentzBoostMatrix.hpp"
+#include "PointwiseFunctions/Xcts/ExtrinsicCurvature.hpp"
 
 namespace {
 
-using KerrSchild = Xcts::Solutions::WrappedGr<gr::Solutions::KerrSchild>;
-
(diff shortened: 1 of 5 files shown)


tests/Unit/PointwiseFunctions/Xcts/Test_AdmMass.cpp
```diff
@@ -45,44 +29,48 @@
...
+  }
+  if (auto* kerrschild = std::get_if<KerrSchild>(&solution)) {
+    return kerrschild->variables(coords, tags);
+  }
+  return tuples::TaggedTuple<RequestedTags...>();
+}
 
+void test_mass_surface_integral(
+    const double distance, const double mass, const double boost_speed,
+    const std::variant<Schwarzschild, KerrSchild> solution) {
```
nilsvu: Instead of the variant and the helper function above, just template this function on `Solution` and take `const Solution& solution` as argument. Then the class is inferred at compile time and you can just do `solution.variables(coords, tags)`.

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMass.cpp
```diff
@@ -45,44 +29,48 @@
...
   // Set up domain
+  const size_t h_refinement = 1;
+  const size_t p_refinement = 6;
   domain::creators::Sphere shell{
-      /* inner_radius */ 2 * mass,
+      /* inner_radius */ 0.5 * mass,
```
nilsvu: You can take the horizon radius or inner radius as an argument to this function (though currently it does not matter because you don't do a volume integral)

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMass.cpp
```diff
@@ -171,23 +159,46 @@ void test_mass_surface_integral(const double distance,
...
 
 }  // namespace
 
 SPECTRE_TEST_CASE("Unit.PointwiseFunctions.Xcts.AdmMass",
                   "[Unit][PointwiseFunctions]") {
-  const size_t P = 6;
-  const size_t L = 1;
-  for (double distance : std::array<double, 3>{{1.e3, 1.e5, 1.e10}}) {
-    test_mass_surface_integral(distance, L, P);
+  // Define black hole parameters.
```
nilsvu: Use scopes like this:

```cpp
{
  INFO("Boosted Kerr-Schild");
  const double mass = 1;
  // ...
}
{
  INFO("Isotropic Schwarzschild");
  // ...
}
```

Then you can use the same variable names in both sections, and you get more information when a test fails (the `INFO` is printed).

tests/Unit/PointwiseFunctions/Xcts/Test_AdmLinearMomentum.cpp
```diff
@@ -3,87 +3,74 @@
...
+  }
+  if (auto* kerrschild = std::get_if<KerrSchild>(&solution)) {
+    return kerrschild->variables(coords, tags);
+  }
+  return tuples::TaggedTuple<RequestedTags...>();
+}
 
+void test_linear_momentum_surface_integral(
+    const double distance, const double mass, const double boost_speed,
+    const std::variant<Schwarzschild, KerrSchild> solution) {
```
nilsvu: Same comments as in `Test_AdmMass.cpp`. If it makes things easier you could combine the two files (up to you).

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
```diff
@@ -6,156 +6,245 @@
...
-    // Compute face normal (related to the conformal metric)
+        determinant_and_inverse(spatial_metric).second;
+    const auto shift = gr::shift(spacetime_metric, inv_spatial_metric);
+    const auto lapse = gr::lapse(shift, spacetime_metric);
+
+    // Do the conformal decomposition.
+    // Note that we choose the same conformal factor as the one used for the
+    // Schwarzschild solution in isotropic coordinates. Other conformal factors
+    // could also work.
+    const auto solution_conformal_factor = solution.variables(inertial_coords,
```
nilsvu: Shouldn't these be the barred coords?

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
```diff
@@ -6,156 +6,245 @@
...
+        partial_derivative(conformal_metric, mesh, inv_jacobian);
+
+    // Compute conformal Christoffel symbols.
+    const auto conformal_phi =
+        gh::phi(lapse, deriv_lapse, shift, deriv_shift, conformal_metric,
+                deriv_conformal_metric);
+    const auto conformal_christoffel_second_kind =
+        gh::christoffel_second_kind(conformal_phi, inv_conformal_metric);
+    const auto conformal_christoffel_contracted = tenex::evaluate<ti::i>(
+        conformal_christoffel_second_kind(ti::J, ti::i, ti::j));
```
nilsvu: Use the Christoffel routines from `src/PointwiseFunctions/GeneralRelativity/Christoffel.hpp`, not from `GeneralizedHarmonic`

nilsvu (changes requested):

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
iago-mendes: I don't think so. This is where we start to do the conformal decomposition. We could choose any conformal factor, but I chose to keep the same one as the (unboosted) Schwarzschild solution. I believe the Lorentz boost is only going to stretch spacetime, not curve it. Then, the previous conformal factor should still work.

Perhaps a better conformal decomposition would be to make the conformal metric flat (like in isotropic coordinates) and then find the proper conformal factor. Here we were just interested in testing a boosted solution with a non-unity conformal factor, so I thought it was good enough to keep the previous one. At least this means that we already know the conformal factor and that it falls off as 1/r.

nilsvu: I believe evaluating the solution at the unbarred coords will fail if they are close to the horizon, which in the unbarred frame is Lorentz-contracted. Schwarzschild in isotropic coords is not horizon-penetrating, so evaluating it at coords within the horizon is problematic. You currently only use values at the outer boundary, so it's probably fine at the moment, but breaks if you try a volume integral (that's my guess).

nilsvu (commented):
You can squash. Best fix the clang-tidy issues while you're at it.

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
iago-mendes: Hmm, that's a good point. I had not realized this. Since we can choose any conformal factor here, let's go with the one that avoids this issue. I've changed it to the barred coordinates.

nilsvu (approved):

Issue #6183, opened 2024-07-29
Scheduler requires mpi use on head node


guilara:
# Bug reports:

The scheduler `(spectre schedule)` command checks that the executable can parse correctly the input file on the head node (this requires to use MPI on the head node). Once this is done, the submit script is generated and submission to the queue proceeds.

On our cluster Urania, however, we need to load an interactive mpi module to run it on the head node (and thus to allow the scheduler to validate the input file). However, this module prevents sbatch to submit jobs (which is the next step):

```
(env) guilara@urania02:/urania/.../RunDir> sbatch Submit.sh 
You cannot use 'sbatch' while the module impi-interactive is loaded!
```

We are not sure of other clusters will have similar problems.

I suggest to remove the lines where the input file is validated (https://github.com/sxs-collaboration/spectre/blob/a6a8ee404306bec9d92da8ab89f636b037aefc25/support/Python/Schedule.py#L544). Or perhaps validate directly on the compute nodes before execution?

### Environment:

Add as an attachment `$SPECTRE_BUILD_DIR/BuildInfo.txt` or
add its contents here.

# Feature request:

### Component:

- [ ] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] Detail 1
- [ ] Detail 2
- [ ] Detail 3

nilsvu:
As a quick workaround, let's add a `--no-validate` flag to skip validation. You can even use a config file to always pass this flag to the CLI automatically on the cluster (see `spectre --help`).

Running the validation on the compute node defeats its purpose, because you want to validate the input file _at job submission_ so you don't have to wait until the job has gone through the queue only to have it fail with a syntax error in the input file. If you can't run an executable and sbatch with the same modules on the head node I'm not sure what to do. Can you run just the executable without `mpirun` on the headnode?

guilara:
@nilsvu I like the idea of the flag. Unfortunately, the executable doesn't run on the head node even without mpirun (unless I load the interactive module ofc). Right, I understand that it defeats the purpose of validation if its done on the compute node. But not sure at the moment what could be another fix.

nilsvu:
Ok can you add the `--no-validate` flag to work around this then? You can add it to `Schedule.py` and the `scheduler_options`.

guilara:
Will do

PR #6184, opened 2024-07-30
Add Urania machine and validate flag to scheduler

guilara:
## Proposed changes

Add submit script information about Urania. Add no-validate flag to scheduler to prevent it from attempting to run mpi on the head node ( #6183).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Environments/urania.sh
+++ b/support/Environments/urania.sh
@@ -1,0 +1,60 @@
+#!/bin/env sh
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+spectre_load_modules() {
+    module load gcc/11
+    module load impi/2021.7
+    module load boost/1.79
+    module load gsl/1.16
+    module load cmake/3.26
+    module load hdf5-serial/1.12.2
+    module load anaconda/3/2021.11
+    module load paraview/5.10
+    # Load Spack environment
+    source /u/guilara/repos/spack/share/spack/setup-env.sh
+    spack env activate env3_spectre_impi
+    # Load python environment
+    source /u/guilara/envs/spectre_env/bin/activate
+    # Define Charm paths
+    export CHARM_ROOT=/u/guilara/charm_impi_3/mpi-linux-x86_64-smp
+    export PATH=$PATH:/u/guilara/charm_impi_3/mpi-linux-x86_64-smp/bin
+}
+
+spectre_unload_modules() {
+    module unload gcc/11
+    module unload impi/2021.7
+    module unload boost/1.79
+    module unload gsl/1.16
+    module unload cmake/3.26
+    module unload hdf5-serial/1.12.2
+    module unload anaconda/3/2021.11
+    module unload paraview/5.10
+    # Unload Spack environment
+    spack env deactivate
+    # Unload python environment
+    deactivate
+}
+
+spectre_run_cmake() {
+    if [ -z ${SPECTRE_HOME} ]; then
+        echo "You must set SPECTRE_HOME to the cloned SpECTRE directory"
+        return 1
+    fi
+    spectre_load_modules
+
+    cmake -D CMAKE_C_COMPILER=gcc \
+          -D CMAKE_CXX_COMPILER=g++ \
+          -D CMAKE_Fortran_COMPILER=gfortran \
+          -D CHARM_ROOT=$CHARM_ROOT \
+          -D CMAKE_BUILD_TYPE=Release \
+          -D DEBUG_SYMBOLS=OFF \
+          -D BUILD_SHARED_LIBS=ON \
+          -D MEMORY_ALLOCATOR=JEMALLOC \
+          -D BUILD_PYTHON_BINDINGS=ON \
+          -D MACHINE=Urania \
+          -D SPEC_ROOT=/u/guilara/repos/spec \
+          -D Catch2_DIR=/u/guilara/repos/Catch2/install_dir/lib64/cmake/Catch2 \
+          "$@" $SPECTRE_HOME
+}
--- a/support/Python/Schedule.py
+++ b/support/Python/Schedule.py
@@ -137,6 +137,7 @@ def schedule(
     submit: Optional[bool] = None,
     clean_output: bool = False,
     force: bool = False,
+    validate: Optional[bool] = True,
     extra_params: dict = {},
     **kwargs,
 ) -> Optional[subprocess.CompletedProcess]:
@@ -307,6 +308,8 @@ def schedule(
         files in the 'run_dir' before scheduling the run. (Default: 'False')
       force: Optional. When 'True', overwrite input file and submit script
         in the 'run_dir' instead of raising an error when they already exist.
+      validate: Optional. When 'True', validate that the input file is parsed
+        correctly. When 'False' skip this step.
       extra_params: Optional. Dictionary of extra parameters passed to input
         file and submit script templates. Parameters can also be passed as
         keyword arguments to this function instead.
@@ -538,9 +541,11 @@ def schedule(
     )
 
     # Validate input file
-    validate_input_file(
-        input_file_path.resolve(), executable=executable, work_dir=run_dir
-    )
+    if validate:
+        validate_input_file(
+            input_file_path.resolve(), executable=executable, work_dir=run_dir
+        )
+
     # - If the input file may request resubmissions, make sure we have a
     #   segments directory
     metadata, input_file = yaml.safe_load_all(rendered_input_file)
@@ -846,6 +851,11 @@ def scheduler_options(f):
             "Overwrite existing files in the '--run-dir' / '-o'. "
             "You may also want to use '--clean-output'."
         ),
+    )
+    @click.option(
+        "--validate/--no-validate",
+        default=True,
+        help="Validate or skip the validation of the input file.",
     )
     # Scheduling options
     @click.option(
--- a/support/SubmitScripts/Urania.sh
+++ b/support/SubmitScripts/Urania.sh
@@ -1,0 +1,37 @@
+{% extends "SubmitTemplateBase.sh" %}
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+# Urania -- HPC cluster of ACR division of MPI for Grav Physics, housed at the
+# Max Planck Computing & Data Facility.
+# https://docs.mpcdf.mpg.de/doc/computing/clusters/systems/Gravitational_Physics_ACR.html
+
+{% block head %}
+{{ super() -}}
+#SBATCH --nodes {{ num_nodes | default(1) }}
+#SBATCH --ntasks-per-node=1
+#SBATCH --ntasks-per-core=1
+#SBATCH --cpus-per-task=72
+#SBATCH -t {{ time_limit | default("1-00:00:00") }}
+#SBATCH -p {{ queue | default("p.urania") }}
+{% endblock %}
+
+{% block charm_ppn %}
+# Two thread for communication
+CHARM_PPN=$(expr ${SLURM_CPUS_PER_TASK} - 2)
+{% endblock %}
+
+{% block list_modules %}
+# Load compiler and MPI modules with explicit version specifications,
+# consistently with the versions used to build the executable.
+source ${SPECTRE_HOME}/support/Environments/urania.sh
+spectre_load_modules
+{% endblock %}
+
+{% block run_command %}
+srun -n ${SLURM_NTASKS} ${SPECTRE_EXECUTABLE} \
+    --input-file ${SPECTRE_INPUT_FILE} \
+    ++ppn ${CHARM_PPN} +pemap 0-34,36-70 +commap 35,71 \
+    ${SPECTRE_CHECKPOINT:+ +restart "${SPECTRE_CHECKPOINT}"}
+{% endblock %}
--- a/support/Machines/Urania.yaml
+++ b/support/Machines/Urania.yaml
@@ -1,0 +1,11 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+Machine:
+  Name: Urania
+  Description: |
+    Supercomputer at the Max Planck Computing Data Facilty.
+  DefaultProcsPerNode: 72
+  DefaultQueue: "p.urania"
+  DefaultTimeLimit: "1-00:00:00"
+  LaunchCommandSingleNode: ["srun", "-n", "1"]
--- a/tests/support/Python/Test_Schedule.py
+++ b/tests/support/Python/Test_Schedule.py
@@ -230,6 +230,7 @@ NUM_NODES={{ num_nodes | default(1) }}
                 extra_option="TestOpt",
                 metadata_option="MetaOpt",
                 force=False,
+                validate=True,
                 input_file="InputFile.yaml",
                 input_file_name="InputFile.yaml",
                 input_file_template=str(self.test_dir / "InputFile.yaml"),


support/Python/Schedule.py
```diff
@@ -854,6 +857,11 @@ def scheduler_options(f):
             "You may also want to use '--clean-output'."
         ),
     )
+    @click.option(
+        "--no-validate",
```
nilsvu: Make this `--validate/--no-validate`, and set `default=True`. Then the variable is positive, which is much easier to understand (you have to change the variable name `no_validate` to `validate`)

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,56 @@
...
+
+# Supercomputer at the Max Planck Computing Data Facility.
+# More information:
+# https://docs.mpcdf.mpg.de/doc/computing/clusters/systems/Gravitational_Physics_ACR.html
+
+{% block head %}
+{{ super() -}}
+#SBATCH -D ./
+#SBATCH --nodes {{ num_nodes | default(1) }}
+#SBATCH --ntasks-per-node=1
```
nilsvu, line 13: We have found that 2 tasks per node work well on other clusters of this size, maybe even 3. I see you currently do 1 task but 2 comm threads. Have you tried 2 tasks with 1 comm thread each? Not sure if there's a difference @knelli2 @nilsdeppe do you know?

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,56 @@
...
+#SBATCH --ntasks-per-core=1
+#SBATCH --cpus-per-task=72
+#SBATCH --mem=240000
+#SBATCH -t {{ time_limit | default("1-00:00:00") }}
+#SBATCH -p {{ queue | default("p.debug") }}
```
nilsvu: The default in the yaml is `p.urania`. Make it the same here?

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,56 @@
...
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+# Supercomputer at the Max Planck Computing Data Facility.
+# More information:
+# https://docs.mpcdf.mpg.de/doc/computing/clusters/systems/Gravitational_Physics_ACR.html
+
+{% block head %}
+{{ super() -}}
+#SBATCH -D ./
```
nilsvu: This isn't needed I think

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,56 @@
...
+#SBATCH --nodes {{ num_nodes | default(1) }}
+#SBATCH --ntasks-per-node=1
+#SBATCH --ntasks-per-core=1
```
nilsvu, line 14: Doesn't this conflict with `--ntasks-per-node`?

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,56 @@
...
+# Load compiler and MPI modules with explicit version specifications,
+# consistently with the versions used to build the executable.
+module purge
+module load gcc/11
+module load impi/2021.7
+module load boost/1.79
+module load gsl/1.16
+module load cmake/3.26
+module load hdf5-serial/1.12.2
+module load anaconda/3/2021.11
```
nilsvu: For the other clusters we have a file like `support/Environments/caltech_hpc_gcc.sh` that loads the modules. That enables you to load the modules to configure a build directory. Then you don't need to load the modules in the submit script, or rather, you would load the modules before submitting a job. I also load the modules in my `~/.bashrc` sometimes (that's personal preference).

nilsvu (changes requested):

support/SubmitScripts/Urania.sh
knelli2: Typically each task = an MPI rank = a comm core

support/Python/Schedule.py
guilara: Changed the name.

support/SubmitScripts/Urania.sh
guilara: I haven't really experimented with this. So I don't know.

support/SubmitScripts/Urania.sh
guilara: Changed. Had left it when testing.

support/SubmitScripts/Urania.sh
guilara: Removed

support/SubmitScripts/Urania.sh
guilara: --ntasks-per-core conflicts with --ntasks-per-node? is that the question? 
--ntasks-per-core I think is there to enable/disable hyperthreading

support/SubmitScripts/Urania.sh
guilara: I added an environment file. I tried to remove the module loads from the submit script but it printed there where no modules loaded even if I had them loaded on my terminal when submitting. So I put them back in

support/Environments/urania.sh
```diff
@@ -0,0 +1,69 @@
...
+    module load impi/2021.7
+    module load boost/1.79
+    module load gsl/1.16
+    module load cmake/3.26
+    module load hdf5-serial/1.12.2
+    module load anaconda/3/2021.11
+    module load paraview/5.10
+    # Load Spack environment
+    source /u/guilara/repos/spack/share/spack/setup-env.sh
+    spack env activate env3_spectre_impi
```
nilsvu, line 17: Note that you can have Spack generate a load file for you so you don't have to source `setup-env.sh` (which can take a long time): https://spack.readthedocs.io/en/latest/environments.html#loading

support/Environments/urania.sh
```diff
@@ -0,0 +1,69 @@
...
+    # Load python environment
+    source $SPECTRE_HOME/env/bin/activate
```
nilsvu: I suggest you put this Python environment in some fixed location so others can use it, e.g. `/u/guilara/envs/spectre_python`. Otherwise everyone has to create a new Py env in every spectre checkout.

support/Environments/urania.sh
```diff
@@ -0,0 +1,69 @@
+#!/bin/env sh
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+spectre_setup_charm_paths() {
+    # Define Charm paths
+    export CHARM_ROOT=/u/guilara/charm_impi_2/mpi-linux-x86_64-smp
+    export PATH=$PATH:/u/guilara/charm_impi_2/mpi-linux-x86_64-smp/bin
```
nilsvu: I suggest you create a module file like `/u/guilara/modules/charm/7.0.0-impi-2` that sets these paths. Then you can `module load` it below and don't have to call this extra `spectre_setup_charm_paths` function.

support/Environments/urania.sh
```diff
@@ -0,0 +1,69 @@
...
+          -D CMAKE_CXX_COMPILER=g++ \
+          -D CMAKE_Fortran_COMPILER=gfortran \
+          -D CHARM_ROOT=$CHARM_ROOT \
+          -D CMAKE_BUILD_TYPE=Release \
+          -D DEBUG_SYMBOLS=OFF \
+          -D BUILD_SHARED_LIBS=ON \
+          -D MEMORY_ALLOCATOR=JEMALLOC \
+          -D BUILD_PYTHON_BINDINGS=ON \
+          -D MACHINE=Urania \
+          -D Python_EXECUTABLE=${SPECTRE_HOME}/env/bin/python \
```
nilsvu: You can delete this line since the Py env is sourced, so cmake should find the correct `python`.

support/Environments/urania.sh
```diff
@@ -0,0 +1,69 @@
...
+          -D Catch2_DIR=/u/guilara/repos/Catch2/install_dir/lib64/cmake/Catch2 \
+          -D MPI_C_COMPILER=/mpcdf/soft/SLE_15/packages/skylake\
+/impi/gcc_11-11.2.0/2021.7.1/bin/mpigcc \
```
nilsvu: So cmake doesn't find these without help even though the `impi/2021.7` module is loaded?

support/Python/Schedule.py
```diff
@@ -149,6 +149,7 @@ def schedule(
     submit: Optional[bool] = None,
     clean_output: bool = False,
     force: bool = False,
+    validate: Optional[bool] = None,
```
nilsvu: `validate: bool = true` (also update docstring)

support/Python/Schedule.py
```diff
@@ -548,10 +551,12 @@ def schedule(
         force=force,
     )
 
-    # Validate input file
-    validate_input_file(
-        input_file_path.resolve(), executable=executable, work_dir=run_dir
-    )
+    if validate or validate is None:
```
nilsvu: `if validate`

support/Python/Schedule.py
```diff
@@ -861,6 +866,11 @@ def scheduler_options(f):
             "You may also want to use '--clean-output'."
         ),
     )
+    @click.option(
+        "--validate/--no-validate",
+        default=None,
```
nilsvu: `default=True`

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,55 @@
...
+#SBATCH --ntasks-per-node=1
+#SBATCH --ntasks-per-core=1
+#SBATCH --cpus-per-task=72
+#SBATCH --mem=240000
+#SBATCH -t {{ time_limit | default("1-00:00:00") }}
+#SBATCH -p {{ queue | default("p.urania") }}
+{% endblock %}
+
+{% block charm_ppn %}
+# One thread for communication
```
nilsvu: <s>One</s> Two threads

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,55 @@
...
+# More information:
+# https://docs.mpcdf.mpg.de/doc/computing/clusters/systems/Gravitational_Physics_ACR.html
+
+{% block head %}
+{{ super() -}}
+#SBATCH --nodes {{ num_nodes | default(1) }}
+#SBATCH --ntasks-per-node=1
+#SBATCH --ntasks-per-core=1
+#SBATCH --cpus-per-task=72
+#SBATCH --mem=240000
```
nilsvu: `--mem=240G` (is this needed?)

support/SubmitScripts/Urania.sh
nilsvu: You really don't want to duplicate the list of modules in these two places. Replace it here with this:

```sh
source ${SPECTRE_HOME}/support/Environments/urania.sh
spectre_load_modules
```

nilsvu (changes requested):
You can squash

support/Environments/urania.sh
guilara: I just tried removing them. I must have added them because it couldn't find it.

```
-- Could NOT find MPI_C (missing: MPI_C_LIB_NAMES MPI_C_HEADER_DIR MPI_C_WORKS) 
-- Could NOT find MPI (missing: MPI_C_FOUND C)
```

nilsvu: Is MPI needed? If everything works without explicitly linking MPI then just remove these lines.

support/SubmitScripts/Urania.sh
guilara: Removed

support/SubmitScripts/Urania.sh
guilara: Corrected

support/Python/Schedule.py
guilara: Changed

support/Python/Schedule.py
guilara: Changed

support/Python/Schedule.py
guilara: Changed

support/Environments/urania.sh
guilara: Removed

support/Environments/urania.sh
guilara: I created this

support/Environments/urania.sh
guilara: I don't think I want to tackle this right now. Installing charm was a bit tricky

support/Environments/urania.sh
guilara: I think we do need it, no?

support/Environments/urania.sh
guilara: changed

support/SubmitScripts/Urania.sh
guilara: Done

support/Environments/urania.sh
```diff
@@ -0,0 +1,69 @@
...
+    module load cmake/3.26
+    module load hdf5-serial/1.12.2
+    module load anaconda/3/2021.11
+    module load paraview/5.10
+    # Load Spack environment
+    source /u/guilara/repos/spack/share/spack/setup-env.sh
+    source /urania/u/guilara/repos/spack/var/spack/environments\
+/env3_spectre_impi/loads
+    # Load python environment
+    source /u/guilara/envs/spectre_env
```
nikwit: missing "bin/activate" ?

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,39 @@
...
+# Two thread for communication
+CHARM_PPN=$(expr ${SLURM_CPUS_PER_TASK} - 2)
+{% endblock %}
+
+{% block list_modules %}
+# Load compiler and MPI modules with explicit version specifications,
+# consistently with the versions used to build the executable.
+source ${SPECTRE_HOME}/support/Environments/urania.sh
+spectre_load_modules
+spectre_setup_charm_paths
```
knelli2: [optional] Might also want to `module list` after this so you can see what's loaded

knelli2 (commented):
I think this looks good for now. If you want to change anything later, you can.

Please squash this all into two commits (after @nilsvu has looked at this again). 1 for the new validate option in python, and 2 for adding the Urania env, machine, and submit files.

support/Environments/urania.sh
nilsvu: Then better set these paths in `spectre_load_modules`, or else you always have to call this extra function?

support/Environments/urania.sh
```diff
@@ -0,0 +1,69 @@
...
+    module load gcc/11
+    module load impi/2021.7
+    module load boost/1.79
+    module load gsl/1.16
+    module load cmake/3.26
+    module load hdf5-serial/1.12.2
+    module load anaconda/3/2021.11
+    module load paraview/5.10
+    # Load Spack environment
+    source /u/guilara/repos/spack/share/spack/setup-env.sh
```
nilsvu, line 16: You can delete this line

support/Environments/urania.sh
nilsvu: No we don't link to it explicitly, only charm is built with it.

nilsvu (changes requested):

support/Environments/urania.sh
```diff
@@ -0,0 +1,69 @@
...
+          -D CMAKE_CXX_COMPILER=g++ \
+          -D CMAKE_Fortran_COMPILER=gfortran \
+          -D CHARM_ROOT=$CHARM_ROOT \
+          -D CMAKE_BUILD_TYPE=Release \
+          -D DEBUG_SYMBOLS=OFF \
+          -D BUILD_SHARED_LIBS=ON \
+          -D MEMORY_ALLOCATOR=JEMALLOC \
+          -D BUILD_PYTHON_BINDINGS=ON \
+          -D MACHINE=Urania \
+          -D Catch2_DIR=/u/guilara/repos/Catch2/install_dir/lib64/cmake/Catch2
```
nilsvu: Backslash at end of line missing

nilsvu (changes requested):

guilara:
@nilsvu I know its been a while. I've rebased this. Hopefully we can get this merged. Unfortunately, spack loads doesn't work for me on either Urania nor Viper so I've kept the "slow" way of activating the spack environment

nilsvu (dismissed):

guilara:
@nilsvu I've just removed here the function to set the charm path

support/SubmitScripts/Urania.sh
```diff
@@ -0,0 +1,39 @@
...
```
nilsvu: Delete here too

nilsvu (changes requested):

support/SubmitScripts/Urania.sh
guilara: Sorry, thank you for catching that

nilsvu (approved):

PR #6187, opened 2024-07-31
Pass spec pythonpath directly to spectre

guilara:
## Proposed changes

Since the spec exporter might be a bit difficult to set up, I would like to use the spec python scripts independently of it (still useful for using the pipeline). Could we pass the path directly if it fails to find spec like this?

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupSpec.cmake
+++ b/cmake/SetupSpec.cmake
@@ -3,13 +3,15 @@
 
 find_package(SpEC)
 
+# Make SpEC scripts available in Python. These can be used until we have ported
+# them to SpECTRE.
+if (SPEC_ROOT)
+  set(PYTHONPATH "${SPEC_ROOT}/Support/Python:${PYTHONPATH}")
+endif()
+
 if (NOT SpEC_FOUND)
   return()
 endif()
-
-# Make SpEC scripts available in Python. These can be used until we have ported
-# them to SpECTRE.
-set(PYTHONPATH "${SPEC_ROOT}/Support/Python:${PYTHONPATH}")
 
 file(APPEND
   "${CMAKE_BINARY_DIR}/BuildInfo.txt"


cmake/SetupSpec.cmake
```diff
@@ -4,6 +4,12 @@
 find_package(SpEC)
 
 if (NOT SpEC_FOUND)
+  # Pass `SPEC_PYTHONPATH` to cmake to make SpEC scripts available in Python
+  # independently of the SpEC exporter. These can be used until we have ported
+  # them to SpECTRE.
+  if (SPEC_PYTHONPATH)
+    set(PYTHONPATH "${SPEC_PYTHONPATH}:${PYTHONPATH}")
```
nilsvu: I suggest you just move up the line that sets the pythonpath so it's always set, like this:

```cmake
find_package(SpEC)

# Make SpEC scripts available in Python. These can be used until we have ported
# them to SpECTRE.
if (SPEC_ROOT)
  set(PYTHONPATH "${SPEC_ROOT}/Support/Python:${PYTHONPATH}")
endif()

# That's it. No need to do anything else in this file.
```

nilsvu (changes requested):

cmake/SetupSpec.cmake
guilara: Oh, right. That's simpler. Done.

cmake/SetupSpec.cmake
```diff
@@ -3,14 +3,14 @@
...
 
-if (NOT SpEC_FOUND)
-  return()
-endif()
-
 # Make SpEC scripts available in Python. These can be used until we have ported
 # them to SpECTRE.
 set(PYTHONPATH "${SPEC_ROOT}/Support/Python:${PYTHONPATH}")
 
+if (NOT SpEC_FOUND)
```
nilsvu, line 12: You have to wrap the line above in `if (SPEC_ROOT)` though

nilsvu (changes requested):

cmake/SetupSpec.cmake
guilara: Oops. Done

nilsvu (approved):

PR #6192, opened 2024-08-01
Measure center of mass in XCTS system.

iago-mendes:
## Proposed changes

This PR computes the center of mass as a surface integral over the domain outer boundary.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
+++ b/src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
@@ -13,12 +13,14 @@
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
 #include "PointwiseFunctions/Xcts/AdmLinearMomentum.hpp"
 #include "PointwiseFunctions/Xcts/AdmMass.hpp"
+#include "PointwiseFunctions/Xcts/CenterOfMass.hpp"
 
 namespace Events {
 
 void local_adm_integrals(
     gsl::not_null<Scalar<double>*> adm_mass,
     gsl::not_null<tnsr::I<double, 3>*> adm_linear_momentum,
+    gsl::not_null<tnsr::I<double, 3>*> center_of_mass,
     const Scalar<DataVector>& conformal_factor,
     const tnsr::i<DataVector, 3>& deriv_conformal_factor,
     const tnsr::ii<DataVector, 3>& conformal_metric,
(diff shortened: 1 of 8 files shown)


src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -77,6 +81,8 @@ void local_adm_integrals(
         extrinsic_curvature, mesh, boundary_direction);
     const auto face_trace_extrinsic_curvature = dg::project_tensor_to_boundary(
         trace_extrinsic_curvature, mesh, boundary_direction);
+    const auto face_inertial_coords = dg::project_tensor_to_boundary(
+        inertial_coords, mesh, boundary_direction);
```
nilsvu: The face coords are available from the DataBox. Use `domain::Tags::Faces<3, domain::Tags::Coordinates<3, Frame::Inertial>>` instead of projecting to the boundary.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -94,6 +100,11 @@ void local_adm_integrals(
                                       face_inv_spatial_metric(ti::J, ti::L) *
                                       face_extrinsic_curvature(ti::k, ti::l));
 
+    // Compute outward-pointing unit normal.
+    const auto face_r = magnitude(face_inertial_coords);
+    const auto face_unit_normal =
+        tenex::evaluate<ti::I>(face_inertial_coords(ti::I) / face_r());
```
nilsvu: You already have the `conformal_face_normal` below. That one has unit magnitude w.r.t. the conformal metric. I see that Ossokine2015 uses `x / r` here. In your test the two are the same because isotropic Schwarzschild has a flat conformal metric. For the binary the two are also the same because we are conformally flat at large distances. So I'd say just use the conformal face normal, or understand what difference this choice makes here.

For the volume integral / Gauss law you do have to be careful which normal you are working with here vs the one consistent with the proper volume element.

nilsvu (changes requested):

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
iago-mendes: For now, I'll use the `conformal_face_normal` one-form as you suggested (though I still have to transform it into a vector). It's not clear to me which one we want to use for Gauss' law, but we might not even need it if we go with the 1/r expansion approach. I think it's fine to leave it like this for now, and we'll come back to it if/when we need.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
iago-mendes: We don't need the coordinates anymore because we're now using `conformal_face_normal` to get the unit normal vector.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -100,10 +97,16 @@ void local_adm_integrals(
...
+        conformal_face_normals.at(boundary_direction);
+
     // Compute outward-pointing unit normal.
-    const auto face_r = magnitude(face_inertial_coords);
-    const auto face_unit_normal =
-        tenex::evaluate<ti::I>(face_inertial_coords(ti::I) / face_r());
+    // Note that `conformal_face_normal` is a one-form, so here we're simply
+    // transforming it into a vector.
+    const auto conformal_unit_normal = tenex::evaluate<ti::I>(
+        face_inv_conformal_metric(ti::I, ti::J) * conformal_face_normal(ti::j));
```
nilsvu: Oh these are actually also available in the DataBox as `domain::Tags::Faces<3, domain::Tags::FaceNormalVector<3>>`.

nilsvu (changes requested):
You can squash everything

nilsvu (approved):

PR #6201, opened 2024-08-08
Add link to logos

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (dismissed):

docs/Main.md
```diff
@@ -37,6 +37,8 @@ style="max-width:100%;"></a>
 </p>
 \endhtmlonly
 
+![banner](docs/Images/banner.png)
```
knelli2: This actually prints the word "banner" under the image (and is probably the same in the README)

knelli2: Should the word "spectre" be a bit larger? It looks a tad small in the space that it's in.

knelli2 (changes requested):

docs/Main.md
nilsvu: 🤷‍♂️ I think it's fine, it's personal preference I suppose. Try this size and adjust later if we find it's too small?

FYI, I intend to add a gravitational wave visualization to this banner when I find the time, to have the "s" radiate a burst of GWs in 3D.

nilsvu: Fixed, thanks for catching

knelli2 (approved):
I think we'll want to change how this looks for dark-mode. The random white rectangle looks a bit odd. But we can do that later.

image

PR #6203, opened 2024-08-09
Start adding support for Kokkos and the Nvidia compiler

nilsdeppe:
## Proposed changes

My plan is to first get support for nvcc (the Nvidia compiler) to compile our code. nvcc uses the Edison Design Group (EDG) compiler front end, and so will parse code differently from GCC and Clang. I plan on first getting the code to compile with nvcc without target GPUs. Once that's working I'll continue getting the necessary bits to run on a GPU. The functions with `KOKKOS_FUNCTION` in front are ones that are compiled with nvcc to run on the GPU.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Installation/BuildSystem.md
+++ b/docs/Installation/BuildSystem.md
@@ -190,6 +190,8 @@ alphabetical order):
   - Minimum priority of input file tests to run. Possible values are: `low` (not
     usually run on CI), `normal` (run at least once on CI), `high` (run always
     on CI). (default is `normal`)
+- SPECTRE_KOKKOS (default: `OFF`)
+  - Enable Kokkos support. You must have Kokkos installed and detectable.
 - SPECTRE_LTO
   - Enable link-time optimization if the compiler supports it.
 - SPECTRE_LTO_CORES
@@ -527,3 +529,22 @@ To further aid in reproducibility, the `printenv` output and
 ```
 h5dump -d /src.tar.gz -b LE -o src.tar.gz /path/to/hdf5/file.h5
 ```
+
+# Using Kokkos
+
+Kokkos support is still extremely experimental, but can be enabled by passing
+`-D SPECTRE_KOKKOS=ON` to CMake. You must pass additional CMake flags that
+Kokkos will use to configure itself. For example, to enable CUDA support you
+must pass `-D Kokkos_ENABLE_CUDA=ON`. See the
+[Kokkos documentation](https://kokkos.org/kokkos-core-wiki/keywords.html) for
+details.
+
+## Nvidia Compiler
+
+If you are using CUDA to compile for Nvidia GPUs but do not have the target GPU
+on the system you are compiling on then you must also tell CMake what CUDA
+architecture to use. You can do this by passing `-D CMAKE_CUDA_ARCHITECTURES=80`
+to CMake. You must choose the architecture that will be compatible with the GPU
+you plan to use. Note that this is actually called the compute capability
+version by Nvidia and can be viewed
+[here](https://developer.nvidia.com/cuda-gpus).
--- a/src/DataStructures/Tensor/EagerMath/CartesianToSpherical.cpp
+++ b/src/DataStructures/Tensor/EagerMath/CartesianToSpherical.cpp
@@ -5,6 +5,7 @@
 
 #include <cstddef>
 
+#include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
--- a/src/Utilities/ErrorHandling/Error.hpp
+++ b/src/Utilities/ErrorHandling/Error.hpp
@@ -63,6 +63,11 @@ template <typename ExceptionTypeToThrow, typename F>
 // 20160415) can't figure out that the else branch and everything
 // after it is unreachable, causing warnings (and possibly suboptimal
 // code generation).
+#ifdef __CUDA_ARCH__
+#define ERROR(m)                                              \
+  printf("Error in file %s on line %d.", __FILE__, __LINE__); \
+  __trap();
+#else
 #define ERROR(m)                                                             \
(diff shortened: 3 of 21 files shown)


tools/BlazeExceptions.hpp
```diff
@@ -0,0 +1,19 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#ifndef SPECTRE_BLAZE_EXCEPTIONS_HPP
+#define SPECTRE_BLAZE_EXCEPTIONS_HPP
+
+#ifdef SPECTRE_DEBUG
+#define BLAZE_THROW(EXCEPTION)           \
```
nilsvu, line 10: Can you just include `csignal` here so it doesn't need special handling in cmake?

src/Utilities/Simd/Simd.hpp
```diff
@@ -242,17 +242,25 @@ T select(const bool cond, const T true_branch, const T false_branch) {
 }
 
 inline std::pair<float, float> sincos(const float val) {
+#if (defined(__CUDACC__) && defined(__CUDA_ARCH__)) or (not defined(__CUDACC__))
```
nilsvu, line 249: Can you add a comment when this needs to be disabled? What does it mean if `__CUDA_ARCH__` is undefined?

src/DataStructures/Tensor/EagerMath/CartesianToSpherical.cpp
```diff
@@ -5,6 +5,7 @@
 
 #include <cstddef>
 
+#include "DataStructures/DataVector.hpp"
```
nilsvu, line 8: Question: can we limit `nvcc` support to only part of the code (essentially PointwiseFunctions & NumericalAlgorithms) so we don't have to keep the full code compatible with nvcc? I.e., can we compile only GPU kernels with `nvcc` and the rest with clang?

src/Utilities/ErrorHandling/Error.hpp
```diff
@@ -85,6 +91,11 @@ template <typename ExceptionTypeToThrow, typename F>
  * \note Any exception types used must have an associated explicit
  * instantiation in `Utilities/ErrorHandling/AbortWithErrorMessage.cpp`
  */
+#ifdef __CUDA_ARCH__
+#define ERROR_AS(m)                                           \
```
nilsvu: Needs additional `EXCEPTION_TYPE` arg, or will it just be discarded?

src/DataStructures/Tensor/EagerMath/CartesianToSpherical.cpp
nilsdeppe: Good question! I think practically speaking the answer is "no." Here are a list of things I can immediately think of that I think would be challenging to overcome (in no particular order):
1. We would need to figure out how to have CMake use a different compiler for different libraries.
2. We would need to filter out the compiler arguments that Kokkos adds for nvcc that GCC/Clang can't handle. I'm already running into this with our linker wrapper, but I haven't dug into what's going on there.
3. There may be ABI incompatibilities now or in the future because things like `__host__` and `__device__` may or may not be included in the function signature. This is implementation defined and Clang and nvcc currently behave differently.
4. We need essentially all header files to be compatible with nvcc anyway since they get included in PointwiseFunctions, NumericalAlgorithms, ParallelAlgorithms, and the final executable (I think we need to compile the final executable with nvcc, I don't see a way around that). This means we would only benefit from not having to make the cpp files compatible, which is definitely not negligible, but I worry that this will be very confusing in the long term. I've found that it is also easier to make one library at a time work with nvcc rather than trying to deal with all the possible headers files when building at the end. Essentially, incremental conversion, even of full libraries, seems easier than many libraries all at once from my experience so far.

nilsdeppe:
I've rebased and made the changes. I squashed most since they were 1 line, and added a commit for `csignal` since it's logically different from the previous commit. Thanks for the review!

src/DataStructures/Tensor/EagerMath/CartesianToSpherical.cpp
nilsvu: Looks like CMake supports separate host and device compilers with [CUDA_SEPARABLE_COMPILATION](https://cmake.org/cmake/help/latest/prop_tgt/CUDA_SEPARABLE_COMPILATION.html). I haven't seen support for this in Kokkos though.

We can also try `-DCMAKE_CUDA_COMPILER=clang++` to avoid having to keep compatibility with `nvcc` for the whole code.

docs/Installation/BuildSystem.md
```diff
@@ -527,3 +529,22 @@ To further aid in reproducibility, the `printenv` output and
 ```
 h5dump -d /src.tar.gz -b LE -o src.tar.gz /path/to/hdf5/file.h5
 ```
+
+# Using Kokkos
```
nilsvu, line 533: Should we install Kokkos in the container and/or in a CI build? I haven't checked how well we can test (emulated) GPUs on CI, but we can certainly test compiling with Kokkos on CPU. Also compatibility with nvcc would be important to test on CI so we don't accidentally break it.

nilsvu (approved):

src/DataStructures/Tensor/EagerMath/CartesianToSpherical.cpp
nilsdeppe: That might be worth trying. I just don't know how it would work with Kokkos because kokkos actually has a wrapper around the Nvidia compiler.

docs/Installation/BuildSystem.md
nilsdeppe: Short answer: yes. What I've been doing is compiling on mbot for the A100 GPUs. I don't need the GPU to compile the code, but presumably I would to run it. I think in CI what we can do is install CUDA, Kokkos, and then at least we can compile with nvcc to make sure we don't break anything. We'll want to do the same for the ROCm and DPC++ compilers too, but that's a more future problem :)

PR #6204, opened 2024-08-09
Observe DimensionfulSpinVector on horizons

AzerAfram:
## Proposed changes
With these changes black hole and binary black hole simulations will output the spin of each black hole. This allows constraint energy control to be done just as in SpEC.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

In the make_legend() function a special case is set aside for when the tag is included, this is a terrible solution for obvious reasons so I slightly editted this version to query the return_type variable in every tag and will accommodate the legend appropriately but that may not be the most efficient solution. Any ideas welcome.

### DIFF
--- a/src/IO/Observer/ReductionActions.hpp
+++ b/src/IO/Observer/ReductionActions.hpp
@@ -229,6 +229,10 @@ void append_to_reduction_data(
     const gsl::not_null<std::vector<double>*> all_reduction_data,
     const std::vector<double>& t);
 
+void append_to_reduction_data(
+    gsl::not_null<std::vector<double>*> all_reduction_data,
+    const std::array<double, 3>& t);
+
 template <typename... Ts, size_t... Is>
 void write_data(const std::string& subfile_name,
                 const std::string& input_source,
--- a/src/ParallelAlgorithms/ApparentHorizonFinder/HorizonAliases.hpp
+++ b/src/ParallelAlgorithms/ApparentHorizonFinder/HorizonAliases.hpp
@@ -41,9 +41,8 @@ using tags_for_observing = tmpl::list<
     gr::surfaces::Tags::IrreducibleMassCompute<Frame>,
     ylm::Tags::MaxRicciScalarCompute, ylm::Tags::MinRicciScalarCompute,
     gr::surfaces::Tags::ChristodoulouMassCompute<Frame>,
-    gr::surfaces::Tags::DimensionlessSpinMagnitudeCompute<Frame>
-    // Needs `ObserveTimeSeriesOnSurface` to be able to write a `std::array`
-    // gr::surfaces::Tags::DimensionlessSpinVectorCompute<Frame, Frame>
+    gr::surfaces::Tags::DimensionlessSpinMagnitudeCompute<Frame>,
+    gr::surfaces::Tags::DimensionfulSpinVectorCompute<Frame, Frame>
     >;
 
 using surface_tags_for_observing = tmpl::list<ylm::Tags::RicciScalar>;
--- a/src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
@@ -40,9 +40,35 @@ namespace callbacks {
 
 namespace detail {
 
+template<typename T>
+struct is_array_of_double : std::false_type {};
+
+template<std::size_t N>
+struct is_array_of_double<std::array<double, N>> : std::true_type {};
+
 template <typename... Ts>
 auto make_legend(tmpl::list<Ts...> /* meta */) {
-  return std::vector<std::string>{"Time", db::tag_name<Ts>()...};
+    std::vector<std::string> legend = {"Time"};
+
+    auto append_tags = [&legend](auto tag) {
+        using TagType = decltype(tag);
+        using ReturnType = typename TagType::type;
+
+        if constexpr (is_array_of_double<ReturnType>::value) {
+            constexpr std::array<const char*, 3> suffix = {"_x", "_y", "_z"};
+            for (size_t i = 0; i < std::tuple_size<ReturnType>::value; ++i) {
+                legend.push_back(db::tag_name<TagType>() + gsl::at(suffix, i));
+            }
+        }
+        else {
+            legend.push_back(db::tag_name<TagType>());
+        }
+
+    };
+
+    (append_tags(Ts{}), ...);
+
+    return legend;
 }
 
 template <typename DbTags, typename... Ts>
--- a/src/IO/Observer/ReductionActions.cpp
+++ b/src/IO/Observer/ReductionActions.cpp
@@ -21,4 +21,10 @@ void append_to_reduction_data(
   all_reduction_data->insert(all_reduction_data->end(), t.begin(), t.end());
 }
 
+void append_to_reduction_data(
+    const gsl::not_null<std::vector<double>*> all_reduction_data,
+    const std::array<double, 3>& t) {
+    all_reduction_data->insert(all_reduction_data->end(), t.begin(), t.end());
+}
+
 }  // namespace observers::ThreadedActions::ReductionActions_detail
--- a/src/PointwiseFunctions/GeneralRelativity/Surfaces/Spin.cpp
+++ b/src/PointwiseFunctions/GeneralRelativity/Surfaces/Spin.cpp
@@ -636,7 +636,7 @@ GENERATE_INSTANTIATIONS(INSTANTIATE,
 
 GENERATE_INSTANTIATIONS(INSTANTIATE,
                         (Frame::Grid, Frame::Distorted, Frame::Inertial),
-                        (Frame::Inertial))
+                        (Frame::Inertial, Frame::Distorted))
 #undef INSTANTIATE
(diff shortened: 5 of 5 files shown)


src/ParallelAlgorithms/ApparentHorizonFinder/HorizonAliases.hpp
```diff
@@ -41,7 +41,8 @@ using tags_for_observing = tmpl::list<
     gr::surfaces::Tags::IrreducibleMassCompute<Frame>,
     ylm::Tags::MaxRicciScalarCompute, ylm::Tags::MinRicciScalarCompute,
     gr::surfaces::Tags::ChristodoulouMassCompute<Frame>,
-    gr::surfaces::Tags::DimensionlessSpinMagnitudeCompute<Frame>
+    gr::surfaces::Tags::DimensionlessSpinMagnitudeCompute<Frame>,
+    gr::surfaces::Tags::DimensionfulSpinVectorCompute<Frame, Frame>
```
nilsvu, line 45: Remove comment below

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
```diff
@@ -41,8 +41,24 @@ namespace callbacks {
...
 template <typename... Ts>
-auto make_legend(tmpl::list<Ts...> /* meta */) {
-  return std::vector<std::string>{"Time", db::tag_name<Ts>()...};
+auto make_legend(tmpl::list<Ts...>) {
+    std::vector<std::string> legend = {"Time"};
+
+    // Helper lambda to append tag names and print
+    auto append_and_print = [&legend](auto tag_name) {
+        std::string name = db::tag_name<decltype(tag_name)>();
+        if (name == "DimensionfulSpinVector") {
```
nilsvu: Don't hardcode this to "DimensionalfulSpinVector" please. Can you perhaps check if the type of the tag is a sequence and if so, expand the components?

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
```diff
@@ -41,8 +41,24 @@ namespace callbacks {
 namespace detail {
 
 template <typename... Ts>
-auto make_legend(tmpl::list<Ts...> /* meta */) {
-  return std::vector<std::string>{"Time", db::tag_name<Ts>()...};
+auto make_legend(tmpl::list<Ts...>) {
```
nilsvu: Please extend Test_ObserveTimeSeriesOnSurface.cpp so your new additions are tested.

nilsvu (changes requested):

AzerAfram:
Any return types of the form `std::array<double, size_t>` are supported for any valid size_t now. This of course relies on all of the doubles being appended to the data vector otherwise there will be a mismatch between size of data and size of legend which would cause a runtime error.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
AzerAfram: Did you mean Test_ObserveTimeSeriesAndSurfaceData.cpp? If so, should I add a test for all of the non-ylm tags; the tags found in /ApparentHorizon.dat in the .h5 file, because I only see tests for the ylm tags in /ApparentHorizon_Ylm.dat

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
```diff
@@ -40,9 +40,34 @@ namespace callbacks {
...
+struct is_array_of_double<std::array<double, N>> : std::true_type {};
+
 template <typename... Ts>
 auto make_legend(tmpl::list<Ts...> /* meta */) {
-  return std::vector<std::string>{"Time", db::tag_name<Ts>()...};
+    std::vector<std::string> legend = {"Time"};
+    
+    auto append_tags = [&legend](auto tag) {
+        using TagType = decltype(tag);
+        using ReturnType = typename TagType::return_type;
```
nilsvu: Can you make this just `typename TagType::type`? Only compute tags have `::return_type`, but all tags have `::type`.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
```diff
@@ -40,9 +40,34 @@ namespace callbacks {
...
+
+        if constexpr (is_array_of_double<ReturnType>::value) {
+            for (size_t i = 0; i < std::tuple_size<ReturnType>::value; ++i) {
+                legend.push_back(db::tag_name<TagType>() + std::to_string(i));
```
nilsvu: Please output these as `TagName_x`, `TagName_y`, `TagName_z`. I understand that the code is generic for any number of array elements and they don't necessarily mean dimensions, but for current use cases they do and it's important to have the "xyz" suffix in the output I think for usability of the output data.

(In the future I think we should change the type of the `DimensionfulSpinVector` tag to `tnsr::I<double, 3, Frame::Inertial>` so it's unambiguous what the numbers mean and in which coordinate frame they are, but that's out of scope for this PR)

nilsvu (changes requested):

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
nilsvu: Hm no that file is testing something different. Unfortunately I don't find a good existing test of the other quantities, which seems like an oversight. Writing a whole new test is out of scope for this PR, so we'll just merge it like this.

nilsvu:
Please fix the errors reported by the checks. Also run `git-clang-format` over your changes, and squash all of your commits into a single one.

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
```diff
@@ -40,9 +40,35 @@ namespace callbacks {
...
 template <typename... Ts>
 auto make_legend(tmpl::list<Ts...> /* meta */) {
-  return std::vector<std::string>{"Time", db::tag_name<Ts>()...};
+    std::vector<std::string> legend = {"Time"};
+
+    auto append_tags = [&legend](auto tag) {
+        using TagType = decltype(tag);
+        using ReturnType = typename TagType::type;
+
+        constexpr std::array<const char*, 3> pattern = {"_x", "_y", "_z"};
```
nilsvu: GCC 9 wants you to move this line into the `if constexpr` (see the failed check)

nilsvu (dismissed):
Looks good to me 👍 Thanks for adding this. Please fix the failed checks and squash all commits into a single one.

nilsvu:
We can merge this as soon as you squash all the commits into one. Here's how I would do it:
1. Rebase on latest `develop` branch: `git pull -r upstream develop`
2. Reset your long list of commits: `git reset upstream/develop`
3. Commit your changes as a single commit: `git commit -a -m"Observe DimensionfulSpinVector on horizons"

nilsvu (dismissed):

kidder (changes requested):
There are still clang-tidy errors that need to be fixed

src/ParallelAlgorithms/Interpolation/Callbacks/ObserveTimeSeriesOnSurface.hpp
```diff
@@ -40,9 +40,35 @@ namespace callbacks {
...
+    std::vector<std::string> legend = {"Time"};
+
+    auto append_tags = [&legend](auto tag) {
+        using TagType = decltype(tag);
+        using ReturnType = typename TagType::type;
+
+        if constexpr (is_array_of_double<ReturnType>::value) {
+            constexpr std::array<const char*, 3> pattern = {"_x", "_y", "_z"};
+            for (size_t i = 0; i < std::tuple_size<ReturnType>::value; ++i) {
+                legend.push_back(db::tag_name<TagType>() + pattern[i]);
```
nilsvu: clang-tidy wants you to use `gsl::at(pattern, i)` instead of `pattern[i]` so the array bounds are checked.

src/IO/Observer/ReductionActions.hpp
```diff
@@ -229,6 +229,10 @@ void append_to_reduction_data(
     const gsl::not_null<std::vector<double>*> all_reduction_data,
     const std::vector<double>& t);
 
+void append_to_reduction_data(
+    const gsl::not_null<std::vector<double>*> all_reduction_data,
```
nilsvu: clang-tidy wants you to delete `const` for this argument (and only this argument, because it has no effect in this function declaration as it's not part of a `const &` type declaration)

nilsvu (commented):
This is what @kidder meant with fixing the clang-tidy issues.

Also, do a `git rebase -r upstream develop`. Then, the clang-tidy issues should show up as failed checks correctly.

AzerAfram:
Ok thanks, all done now I think ...

nilsvu (approved):

PR #6212, opened 2024-08-14
Control center of mass and linear momentum in initial data.

### LINKED ISSUE #5459: Off-axis excisions in BBH domain
To control the center of mass in BBH initial data, SpEC offsets the excision spheres slightly from the x-axis. Currently, our `BinaryCompactObject` domain only allows to specify the x coordinate of the excisions. See Sec. 2.5 in https://arxiv.org/abs/1506.01689.

iago-mendes:
## Proposed changes

This PR extends the control loop in initial data to adjust the center of mass and ADM linear momentum.

### Upgrade instructions

This PR changes `BinaryCompactObject`, which is used in multiple parts of the code. Now, the constructor expects the argument `std::array<double, 2> center_of_mass_offset`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/BinaryCompactObject.hpp
+++ b/src/Domain/Creators/BinaryCompactObject.hpp
@@ -151,6 +151,8 @@ class BinaryCompactObject : public DomainCreator<3> {
   using Affine = CoordinateMaps::Affine;
   using Affine3D = CoordinateMaps::ProductOf3Maps<Affine, Affine, Affine>;
   using Identity2D = CoordinateMaps::Identity<2>;
+  // The Translation type is no longer needed, but it is kept here for backwards
+  // compatibility with old domains.
   using Translation = CoordinateMaps::ProductOf2Maps<Affine, Identity2D>;
   using Equiangular = CoordinateMaps::Equiangular;
   using Equiangular3D =
@@ -175,6 +177,12 @@ class BinaryCompactObject : public DomainCreator<3> {
                             CoordinateMaps::Wedge<3>>,
       domain::CoordinateMap<Frame::BlockLogical, Frame::Inertial,
                             CoordinateMaps::Wedge<3>, Translation>,
+      domain::CoordinateMap<Frame::BlockLogical, Frame::Inertial, Affine3D,
+                            Affine3D>,
+      domain::CoordinateMap<Frame::BlockLogical, Frame::Inertial, Equiangular3D,
+                            Affine3D>,
+      domain::CoordinateMap<Frame::BlockLogical, Frame::Inertial,
+                            CoordinateMaps::Wedge<3>, Affine3D>,
       bco::TimeDependentMapOptions<false>::maps_list>>;
 
   /// Options for an excision region in the domain
@@ -311,6 +319,15 @@ class BinaryCompactObject : public DomainCreator<3> {
         "x-axis)."};
   };
 
+  struct CenterOfMassOffset {
+    using type = std::array<double, 2>;
+    static constexpr Options::String help = {
+        "Offset in the y and z axes applied to both object A and B in order to "
+        "control the center of mass. This moves the location of the two objects"
+        " in the grid frame but keeps the Envelope and OuterShell centered on "
+        "the origin in the grid frame."};
+  };
+
   struct Envelope {
     static constexpr Options::String help = {
         "Options for the sphere enveloping the two objects."};
@@ -410,10 +427,10 @@ class BinaryCompactObject : public DomainCreator<3> {
 
   template <typename Metavariables>
   using options = tmpl::append<
-      tmpl::list<ObjectA, ObjectB, EnvelopeRadius, OuterRadius,
-                 InitialRefinement, InitialGridPoints, UseEquiangularMap,
-                 RadialDistributionEnvelope, RadialDistributionOuterShell,
-                 OpeningAngle, TimeDependentMaps>,
+      tmpl::list<ObjectA, ObjectB, CenterOfMassOffset, EnvelopeRadius,
+                 OuterRadius, InitialRefinement, InitialGridPoints,
+                 UseEquiangularMap, RadialDistributionEnvelope,
+                 RadialDistributionOuterShell, OpeningAngle, TimeDependentMaps>,
       tmpl::conditional_t<
           domain::BoundaryConditions::has_boundary_conditions_base_v<
               typename Metavariables::system>,
@@ -449,7 +466,8 @@ class BinaryCompactObject : public DomainCreator<3> {
 
   BinaryCompactObject(
       typename ObjectA::type object_A, typename ObjectB::type object_B,
-      double envelope_radius, double outer_radius,
+      std::array<double, 2> center_of_mass_offset, double envelope_radius,
+      double outer_radius,
       const typename InitialRefinement::type& initial_refinement,
       const typename InitialGridPoints::type& initial_number_of_grid_points,
       bool use_equiangular_map = true,
@@ -507,6 +525,7 @@ class BinaryCompactObject : public DomainCreator<3> {
  private:
   typename ObjectA::type object_A_{};
   typename ObjectB::type object_B_{};
+  std::array<double, 2> center_of_mass_offset_{};
   double envelope_radius_ = std::numeric_limits<double>::signaling_NaN();
   double outer_radius_ = std::numeric_limits<double>::signaling_NaN();
   std::vector<std::array<size_t, 3>> initial_refinement_{};
--- a/src/PointwiseFunctions/AnalyticData/Xcts/Binary.hpp
+++ b/src/PointwiseFunctions/AnalyticData/Xcts/Binary.hpp
@@ -321,6 +321,12 @@ class Binary : public elliptic::analytic_data::Background,
         "The coordinates on the x-axis where the two objects are placed";
     using type = std::array<double, 2>;
   };
+  struct CenterOfMassOffset {
+    static constexpr Options::String help = {
+        "Offset in the y and z axes applied to both objects in order to "
+        "control the center of mass."};
+    using type = std::array<double, 2>;
+  };
   struct ObjectLeft {
     static constexpr Options::String help =
         "The object placed on the negative x-axis";
@@ -355,8 +361,9 @@ class Binary : public elliptic::analytic_data::Background,
         "to disable the Gaussian falloff.";
     using type = Options::Auto<std::array<double, 2>, Options::AutoLabel::None>;
   };
-  using options = tmpl::list<XCoords, ObjectLeft, ObjectRight, AngularVelocity,
-                             Expansion, LinearVelocity, FalloffWidths>;
+  using options =
+      tmpl::list<XCoords, CenterOfMassOffset, ObjectLeft, ObjectRight,
+                 AngularVelocity, Expansion, LinearVelocity, FalloffWidths>;
   static constexpr Options::String help =
       "Binary compact-object data in general relativity, constructed from "
       "superpositions of two isolated objects.";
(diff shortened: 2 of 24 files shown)


src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -311,6 +311,13 @@ class BinaryCompactObject : public DomainCreator<3> {
         "x-axis)."};
   };
 
+  struct ComOffset {
```
kidder: Please spell out CenterOfMassOffset

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -449,7 +456,8 @@ class BinaryCompactObject : public DomainCreator<3> {
 
   BinaryCompactObject(
       typename ObjectA::type object_A, typename ObjectB::type object_B,
-      double envelope_radius, double outer_radius,
+      std::array<double, 2> com_offset, double envelope_radius,
```
kidder: also spell out center_of_mass_offset

kidder (commented):
Not a complete review, but please don't abbreviate "center of mass" as com
I don't need to approve before the PR is merged

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -151,7 +151,7 @@ class BinaryCompactObject : public DomainCreator<3> {
   using Affine = CoordinateMaps::Affine;
   using Affine3D = CoordinateMaps::ProductOf3Maps<Affine, Affine, Affine>;
   using Identity2D = CoordinateMaps::Identity<2>;
-  using Translation = CoordinateMaps::ProductOf2Maps<Affine, Identity2D>;
+  using Translation = CoordinateMaps::ProductOf3Maps<Affine, Affine, Affine>;
```
nilsvu: For backwards compatibility with existing data files you may only _add_ to the `maps_list` below, never remove or change any maps. So, define a new translation map alias here and add the new maps to the list below.

src/PointwiseFunctions/AnalyticData/Xcts/Binary.hpp
```diff
@@ -456,6 +473,9 @@ class Binary : public elliptic::analytic_data::Background,
           inv_jacobian,
       tmpl::list<RequestedTags...> /*meta*/) const {
     std::array<tnsr::I<DataVector, 3>, 2> x_isolated{{x, x}};
+    std::array<std::array<double, 3>, 2> coords_isolated{
```
nilsvu: Const

support/Pipelines/Bbh/ControlId.py
```diff
@@ -89,6 +91,12 @@ def Residual(
         chi_Kerr_B_x: float,
         chi_Kerr_B_y: float,
         chi_Kerr_B_z: float,
+        center_A_x: float,
+        center_A_y: float,
+        center_A_z: float,
```
nilsvu: Question: would it make sense to control the offset in xyz from the Newtonian COM rather than center_A? Seems nicer because the initial guess is 0, but I haven't thought carefully about how this would impact the control loop Jacobian.

nilsvu (changes requested):
Not yet a complete review, I'll look at the control loop changes tomorrow.

support/Pipelines/Bbh/ControlId.py
iago-mendes: That's a good idea. I've tested it, and I believe it leads to better results. CoM and Padm still don't converge in some cases, which I think might be related to the low accuracy in their computation. Until this is resolved, I've disabled their control by default.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -18,8 +18,18 @@
 DEFAULT_RESIDUAL_TOLERANCE = 1.0e-6
 DEFAULT_MAX_ITERATIONS = 30
 
+SuportedParams = Literal[
```
nilsvu: typo `SupportedParams`

support/Pipelines/Bbh/ControlId.py
```diff
@@ -45,7 +55,31 @@ def control_id(
...
+      spin_A: Dimensionless spin of the larger black hole.
+      spin_B: Dimensionless spin of the smaller black hole.
+      center_of_mass: Center of mass integral in general relativity.
+      linear_momentum: ADM linear momentum.
+
+    Example of control_params for an equal-mass non-spinning run with minimal
+    drift of the center of mass:
+      ```py
+        control_params = {
+          mass_A: 0.5,
```
nilsvu: in this syntax you need quotes around `"mass_A"`. Another nice way to write this is `dict(mass_A=0.5, ...)` so you need no quotes.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -45,7 +55,31 @@ def control_id(
...
+          spin_A: [0., 0., 0.],
+          spin_B: [0., 0., 0.],
+          center_of_mass: [0., 0., 0.],
+          linear_momentum: [0., 0., 0.],
+        }
+      ```
+
     Arguments:
+      control_params: Dictionary used to customize control. The keys determine
+        which physical parameters are controled and the values determine their
```
nilsvu: typo `controlled`

support/Pipelines/Bbh/ControlId.py
```diff
@@ -18,8 +18,18 @@
...
+    "mass_B",
+    "spin_A",
+    "spin_B",
+    "center_of_mass",
+    "linear_momentum",
+]
+
 
 def control_id(
+    control_params: Dict[SuportedParams, Union[float, Sequence[float]]],
```
nilsvu: (nitpick) I'd move this argument down 1 line for consistency with other pipeline functions that usually take the input file path that they operate on as first argument, then other required arguments, then optional arguments.

src/Domain/Creators/BinaryCompactObject.hpp
iago-mendes: I'm using `Affine3D` instead of changing `Translation` now. I still have to fix some failed domain tests.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -58,22 +92,45 @@ def control_id(
...
-    M_target_A = binary_data["ObjectRight"]["KerrSchild"]["Mass"]
-    M_target_B = binary_data["ObjectLeft"]["KerrSchild"]["Mass"]
-    chi_target_A = binary_data["ObjectRight"]["KerrSchild"]["Spin"]
-    chi_target_B = binary_data["ObjectLeft"]["KerrSchild"]["Spin"]
-    x_B, x_A = binary_data["XCoords"]
-    y_offset, z_offset = binary_data["CenterOfMassOffset"]
-    separation = x_A - x_B
+
+    # Get initial Kerr masses and spins
+    initial_mass_Kerr_A = binary_data["ObjectRight"]["KerrSchild"]["Mass"]
```
nilsvu: I suggest you build a dict from the initial params:

```py
initial_params = dict(
    mass_A=binary_data["ObjectRight"]["KerrSchild"]["Mass"],
    # ...
)
```

Then update it with the controlled params in the residual function like this (needs some care to handle lists of vector components but you get the idea):

```py
params = initial_params.copy()
params.update(zip(control_params, u))
```

Then you don't need the `current_index` thing below, and don't have to be so explicit about the specific parameters. You can just do this to pass the params along:

```py
generate_id(**params, ...)
```

And to return the residual maybe something like this:

```py
if "mass_A" in control_params:
    measured_params["mass_A"] = AhA["ChristodoulouMass"]
# ...

return [measured_params[key] - target for key, target in control_params.items()]
```

nilsvu (changes requested):

support/Pipelines/Bbh/ControlId.py
iago-mendes: I've changed the control logic using your tips. I still had to iterate through `u` in order to handle the vector components properly, but I think it's still better now. What do you think?

tests/Unit/Domain/Test_Domain.cpp
```diff
@@ -1161,9 +1162,13 @@ Domain<3> create_serialized_domain() {
   Maps maps{};
 
   const Translation translation_A{
-      Affine{-1.0, 1.0, -1.0 + x_coord_a, 1.0 + x_coord_a}, Identity2D{}};
+      Affine{-1.0, 1.0, -1.0 + x_coord_a, 1.0 + x_coord_a},
+      Affine{-1.0, 1.0, -1.0 + y_offset, 1.0 + y_offset},
+      Affine{-1.0, 1.0, -1.0 + z_offset, 1.0 + z_offset}};
```
nilsvu: I think for the test to succeed you have to drop this change, and the one for `translation_B` below. Because these are the maps actually written to disk in the old dataset.

nilsvu (dismissed):
LGTM please squash 👍 It would help @knelli2 if you split the changes into two commits: one for the domain creator and the other for the control loop.

tests/Unit/Domain/Test_Domain.cpp
iago-mendes: That worked, thanks!

support/Pipelines/Bbh/ControlId.py
```diff
@@ -18,9 +18,37 @@
 DEFAULT_RESIDUAL_TOLERANCE = 1.0e-6
 DEFAULT_MAX_ITERATIONS = 30
 
+# Initial data physical parameters that are supported in this control scheme
+SupportedParams = Literal[
```
nilsvu, line 22: (optional) You can do this to avoid duplicating this list:

```py
SupportedParams = Literal[*FreeDataFromParams.keys()]
```

nilsvu (dismissed):

support/Pipelines/Bbh/ControlId.py
iago-mendes: It seems like `Literal` doesn't take "unpacked arguments", so this would disable the type checking. Since these parameters are specified by users in the input file, I think it would be better to keep it the way it is.

nilsvu: Oh ok it worked on my machine with Py 3.12, but maybe this was a more recent addition to Python. Fine with me.

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -311,6 +319,13 @@ class BinaryCompactObject : public DomainCreator<3> {
         "x-axis)."};
   };
 
+  struct CenterOfMassOffset {
+    using type = std::array<double, 2>;
+    static constexpr Options::String help = {
+        "Offset in the y and z axes applied to both object A and B in order to "
+        "control the center of mass."};
```
knelli2: Can you add a note that this moves the location of the two objects in the grid frame but keeps the Envelope and OuterShell centered on the origin in the grid frame

src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -516,7 +532,9 @@ class BinaryCompactObject : public DomainCreator<3> {
       CoordinateMaps::Distribution::Projective;
   CoordinateMaps::Distribution radial_distribution_outer_shell_ =
       CoordinateMaps::Distribution::Linear;
-  double translation_{};
+  double x_offset_{};
+  double y_offset_{};
+  double z_offset_{};
```
knelli2: Make this a `std::array` named `center_of_mass_offset_`. With the excision offset that will be added in soon, we'll want to be very explicit about what offset is moving what coordinates.

src/PointwiseFunctions/AnalyticData/Xcts/Binary.hpp
```diff
@@ -440,6 +455,8 @@ class Binary : public elliptic::analytic_data::Background,
 
  private:
   std::array<double, 2> xcoords_{};
+  double y_offset_{};
+  double z_offset_{};
```
knelli2, line 459: These are probably ok as is

knelli2 (changes requested):

knelli2 (approved):
Looks good! The clang-tidy stuff is unrelated to your changes

nilsvu (approved):

Issue #6092, opened 2024-06-13
BBH domain: translate inner spheres to increase size of cubes


nilsvu:
We currently have the problem that the excisions grow larger than the inner cubes of the BBH domain, which crashes simulations because only the inner cubes can be deformed to conform to horizons.

Here's the domain at the start of the simulation:

![Bildschirmfoto 2024-06-13 um 11 39 24](https://github.com/sxs-collaboration/spectre/assets/746230/f51fce9a-367b-4efb-9f29-223f97b8e9cf)

And here it is close to merger, right before it crashes:

![image (1)](https://github.com/sxs-collaboration/spectre/assets/746230/a36be43e-098b-4d1d-845e-bb76a61edd91)

So to fix this, we can increase the size of the inner cubes, which we can do by translating the inner spheres within the inner cubes slightly towards the cutting plane at the center of the domain. We have to make sure to retain the equiangular map in the spheres and adjust the shape transition functions in the inner cubes.

knelli2:
Related to #5459

macedo22:
Addressed by #6214, #6234, and #6243

PR #6214, opened 2024-08-14
Enable Wedges to have a focal offset

macedo22:
## Proposed changes
Addresses #6092

This code implementation is based off of @mar-celine's mathematical derivation of the offset wedge and was made possible by her extensive and meticulous efforts both in deriving it and mentoring myself and @AlexCarpenter46 in domain concepts

This PR enables `Wedge`s to have an off-center focus from which they are focally lifted. Some missing documentation and cleanup is first done partially to make the subsequent focal offset changes make more sense.

This is a precursor to a future follow-up PR that will make use of this feature by enabling the `BinaryCompactObject` domain to be constructed such that the length of the cubes around the excision surfaces can be larger than the separation between the excision surfaces. This future follow-up PR that will enable users to create a BCO with this feature will come after @knelli2 's upcoming PR that will update the `ShapeMapTransitionFunction` to first work with offset `Wedge`s.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Domain/CoordinateMaps/Wedge.hpp
```diff
@@ -650,15 +1099,29 @@ class Wedge {
...
+  /// `focal_offset` is nonzero, `sphericity_inner` must be `1.0`.
   double sphericity_inner_{std::numeric_limits<double>::signaling_NaN()};
   /// Value between 0 and 1 which determines whether the outer surface is flat
-  /// (value of 0), spherical (value of 1) or somewhere in between.
+  /// (value of 0), spherical (value of 1) or somewhere in between. If
+  /// `focal_offset` is nonzero, `sphericity_outer` must be `0.0` or `1.0`.
   double sphericity_outer_{std::numeric_limits<double>::signaling_NaN()};
+  /// Half the length of the parent surface (see Wedge documentation for more
+  /// details). This parameter has no effect when `focal_offset` is zero.
+  double cube_half_length_{std::numeric_limits<double>::signaling_NaN()};
```
macedo22: One aspect of the code that we'd like to ask the reviewers about is the interface of the `Wedge` constructor and the member data. In wrapping up this PR, we realized that there is a lot of complexity introduced in the constructor parameters and member data, e.g. when you have a centered `Wedge`, the `cube_length` argument and `cube_length_` members are meaningless and changing them doesn't do anything. Likewise, when you have an offset `Wedge` with `outer_sphericity == 0.0`, `radius_outer_` doesn't correspond to the radial size of the `Wedge` and changing it doesn't  do anything, either. Moreover, `opening_angles` don't make sense with an offset `Wedge` because the angular size is set by the offset, itself.

There is enough different behavior and things that do or don't make sense to pass into a constructor and store as member data depending on whether you have an offset wedge or centered wedge that we considered creating an `OffsetWedge` subclass or instead two separate `CubicalOffsetWedge` and `SphericalOffsetWedge` subclasses that inherited from `Wedge` to separate the constructor interface and member data or the centered and offset cases. For example, instead of the common class and constructor in this PR:
```
Wedge(double radius_inner, double radius_outer, double sphericity_inner,
        double sphericity_outer, double cube_half_length,
        std::array<double, Dim> focal_offset,
        OrientationMap<Dim> orientation_of_wedge, bool with_equiangular_map,
        WedgeHalves halves_to_use = WedgeHalves::Both,
        Distribution radial_distribution = Distribution::Linear,
        const std::array<double, Dim - 1>& opening_angles =
            make_array<Dim - 1>(M_PI_2),
        bool with_adapted_equiangular_map = true);
```

we considered something like keeping the existing `Wedge` constructor on `develop` (no `cube_length` or `focal_offset`):
```
Wedge(double radius_inner, double radius_outer, double sphericity_inner,
        double sphericity_outer,
        OrientationMap<Dim> orientation_of_wedge, bool with_equiangular_map,
        WedgeHalves halves_to_use = WedgeHalves::Both,
        Distribution radial_distribution = Distribution::Linear,
        const std::array<double, Dim - 1>& opening_angles =
            make_array<Dim - 1>(M_PI_2),
        bool with_adapted_equiangular_map = true);
```

and then perhaps:
```
/// For a Wedge with a flat outer surface, we don't take a radius_outer because it's meaningless
/// in this case, we don't take a sphericity_outer because it's implicitly 0.0, we don't
/// take sphericity_inner because it has to be 1.0, and we don't take any opening_angles
/// or with_adapted_equiangular_map because those don't make sense with an offset
CubicalOffsetWedge(double radius_inner, double cube_half_length,
        std::array<double, Dim> focal_offset,
        OrientationMap<Dim> orientation_of_wedge, bool with_equiangular_map,
        WedgeHalves halves_to_use = WedgeHalves::Both,
        Distribution radial_distribution = Distribution::Linear;

/// For a Wedge with a spherical outer surface,  it's the same as the above case but instead the
/// outer_sphericity is implicitly 1.0 and we additionally take a radius_outer because it *is* used
/// when outer_sphericity = 1.0
SphericalOffsetWedge(double radius_inner, double radius_outer, double cube_half_length,
        std::array<double, Dim> focal_offset,
        OrientationMap<Dim> orientation_of_wedge, bool with_equiangular_map,
        WedgeHalves halves_to_use = WedgeHalves::Both,
        Distribution radial_distribution = Distribution::Linear;
```

We didn't want to make any big changes like this without first asking reviewers for possible better solutions and feedback, because we know we want this feature to get merged soon. The complexity of the interface and member data doesn't sit well with us but we also don't want to unnecessarily delay anything

src/Domain/CoordinateMaps/Wedge.hpp
```diff
@@ -47,209 +47,482 @@ struct WedgeCoordOrientation<3> {
...
+ *             \Xi \\
+ *             \mathrm{H} \\
+ *             1 \\
+ *           \end{bmatrix}
+ * \end{align}
+ *
+ * We constrain $L$ by demanding that the spherical face circumscribe the cube.
+ * With this condition, we have \f$L = R/\sqrt3\f$.
+ * \note This differs from the choice in SpEC where it is demanded that the
+ * surfaces touch at the center, which leads to \f$L = R\f$.
```
knelli2: `touch at the cube face centers`

src/Domain/CoordinateMaps/Wedge.hpp
```diff
@@ -47,209 +47,482 @@ struct WedgeCoordOrientation<3> {
...
+ *       = \frac{1}{1+\Xi^2}\frac{\mathrm{d}\Xi}{\mathrm{d}\xi}
+ *       = \frac{\pi}{4}.
+ * \end{align}
+ *
+ * In other words, this parameterization has the property that the logical
+ * coordinate $\xi$ subtends the angle $\theta$ at a constant rate. In general,
+ * we say that a curve $\vec{\Gamma}(\xi)$ is parameterized *equiangularly* if
+ *
+ * \begin{align}
+ *   \frac{\mathrm{d}\theta}{\mathrm{d}\xi} = C.
```
knelli2: `C` is some constant, yes? It doesn't have to be $\pi/4$?

src/Domain/CoordinateMaps/Wedge.hpp
```diff
@@ -523,6 +525,264 @@ struct WedgeCoordOrientation<3> {
...
+ * particular, when the the offset is set to some non-zero value, the
+ * `opening_angles_` member variable is set to $\pi/2$. Note that the
+ * `opening_angles_` member being set to $\pi/2$ does not imply the
+ * resulting Wedge will have an angular size of $\pi/2$. On the contrary, the
+ * Wedge will have the angular size that is determined by the application of
+ * the focal lifting method on the parent surface, which is the upper $+z$ face
+ * of a cube that is centered at the origin.
+ *
+ * Because `opening_angles_` is set to $\pi/2$ when there is a non-zero focal
+ * offset, when there is a non-zero focal offset and `with_equiangular_map_` is
```
knelli2, line 794: Probably don't need both "when there is a non-zero focal offset"

src/Domain/CoordinateMaps/Wedge.hpp
```diff
@@ -523,6 +525,264 @@ struct WedgeCoordOrientation<3> {
  * $-x$ directions have angular dimensions and gridpoint distributions
  * determined by the other four `Wedge`s, as the six `Wedge`s must conforming
  * have gridpoint distributions at the $\xi = \pm1$, $\eta = \pm 1$ boundaries.
+ *
+ * ### Wedge with a Focal Offset
+ * \image html FocalOffset.jpg "Wedges without and with a focal offset"
```
knelli2, line 531: I don't know if you explicitly state it, but this requires the inner surface to be spherical, correct? But the outer surface can be whatever you like?

src/Domain/CoordinateMaps/Wedge.cpp
```diff
@@ -63,21 +65,81 @@ Wedge<Dim>::Wedge(const double radius_inner, const double radius_outer,
       "Wedge rotations must be done in such a manner that the sign of "
       "the determinant of the discrete rotation is positive. This is to "
       "preserve handedness of the coordinates.");
+  ASSERT(approx_zero_offset or (opening_angles_ == make_array<Dim - 1>(M_PI_2)),
+         "Cannot use both a non-zero focal offset and opening angles not equal "
+         "to pi/2.");
```
knelli2: The double negative is a bit confusing. Maybe
```
The focal offset is non-zero, but the opening angles are not pi/2.
```

src/Domain/CoordinateMaps/Wedge.cpp
```diff
@@ -32,12 +37,16 @@ Wedge<Dim>::Wedge(const double radius_inner, const double radius_outer,
...
+      cube_half_length_(cube_half_length),
+      focal_offset_(focal_offset),
       orientation_of_wedge_(std::move(orientation_of_wedge)),
       with_equiangular_map_(with_equiangular_map),
       halves_to_use_(halves_to_use),
       radial_distribution_(radial_distribution),
       opening_angles_(opening_angles) {
   const double sqrt_dim = sqrt(double{Dim});
+  const bool approx_zero_offset =
+      equal_within_roundoff(magnitude(focal_offset_), 0.0);
```
knelli2: Why not exactly zero offset?

src/Domain/CoordinateMaps/Wedge.hpp
```diff
@@ -541,18 +801,30 @@ class Wedge {
...
+   * somewhere in between. If `focal_offset` is nonzero, `sphericity_inner` must
+   * be `1.0`.
    * \param sphericity_outer Value between 0 and 1 which determines
    * whether the outer surface is flat (value of 0), spherical (value of 1) or
-   * somewhere in between.
+   * somewhere in between. If `focal_offset` is nonzero, `sphericity_outer` must
+   * be `0.0` or `1.0`.
+   * \param cube_half_length Half the length of the parent surface (see Wedge
+   * documentation for more details). This parameter has no effect when
+   * `focal_offset` is zero.
```
knelli2: We usually don't like having parameters that are completely ignored based on some other parameter. It can make debugging hard if you have to track down a number just to realize it didn't matter in the first place. This seems like a good example to use a `std::optional` with  `std::pair` since the cube half length and focal offset are intimately tied. So the option in the constructor would be

```cpp
Wedge(...
      const std::optional<std::pair<std::array<double, 3>, double>>&
             focal_offset_and_half_cube_length,
      ...);
```

This way nothing is ignored and as a bonus, you can know exactly (not approximately) if you have an offset.

knelli2 (changes requested):
This looks really good! Mostly docs changes with one more substantial code change, but nothing to the implementation itself. Though I read through the docs and checked the code, I didn't attempt to check the math. I have, however, run several successful BBH simulations using these changes, so I'm confident the math is correct.

src/Domain/CoordinateMaps/Wedge.hpp
macedo22: Totally on board with not having unused parameters (and member variables). It was very upsetting to me 😂 Not sure if you saw my long comment above (for some reason I can't figure out how to get a link to it) that includes this issue + member data issues and more.

If we go with something like what you're proposing here, then I like the `std::optional<std::pair<___, ___>>` as a constructor parameter, but then my only question is what happens with the member data? I can imagine the member `focal_offset_` still just being an array that either holds zeros or non-zeros, but what about `cube_length_`? Would we have it be a `double` that stores `NaN` or have its type be `std::optional<double>` or should we have a combined member variable that is likewise a `std::optional<std::pair<___, ___>>` for both?

Regardless of whether or not we go with this `std::optional<std::pair<>`> idea, I still have more concerns about other unused constructor parameters and member data (as mentioned in my long comment). If we do this change, we still have the case where `radius_outer_` is unused and for any offset wedge, all of the opening angle stuff is unused, and don't make sense as parameters nor member variables of the class. See my long comment above and let me know your thoughts!

macedo22: Reasons I might prefer an offset subclass (or two) over just using one class for both centered and offset wedges are:
- the member data that should be stored is different
- the logic for the special handling is siloed, so cognitive complexity will be better thus maybe also more maintainable. It's easier to add features to, say, just the centered wedge if we only have some feature solved for the centered case.
- I imagine if we used  `std::optional` for cube_length and focal_offset, then when we have a centered wedge (the optional has no value), then in the cpp where we do the math, we would either need to do `if not opt.has_value(){} else {}` in a bunch of places and write different code in each case, or we would need to just give `cube_length` some real value that is meaningless so that we could still use the current math code in the cpp (since we can't do math on an optional that doesn't have a value).

Whatever you and/or other core devs think is best, we should do. But because there is a lot of complexity here and big differences to the centered and offset wedge, I wanted to make sure that I mentioned these things that may be important considerations, e.g. in a combined class, for the offset wedge's with outside sphericity = 0.0, `radius_outer_` is meaningless

macedo22: Another solution would simply be to have one class but two (or three) different constructors (interfaces). This would basically be like what I suggested in my long comment at the top of the PR but just have those 3 constructors in the same class. That would at least handle things like the fact that offset wedges shouldn't take an opening angles parameters, and then if it's possible to have these opening angles be an optional or NaN or something, then that handles the unused member data

src/Domain/CoordinateMaps/Wedge.hpp
macedo22: This line is a general statement about equiangular parametrization to follow the finding that our $\frac{\mathrm{d}\theta}{\mathrm{d}\xi} = \pi/4$ and why that means equiangular. The mention of this general property after that finding is meant to substantiate/explain why it means the $\pi/4$ means equiangular. Let us know if we can change any wording to be more clear!

src/Domain/CoordinateMaps/Wedge.hpp
macedo22: The inner surface must be spherical but the outer surface must either be cubical with the outer surface being the parent surface or spherical with any outer radius. So the outer surface cannot be anything but 0.0 or 1.0 and if it's 0.0, the outer surface must be the parent surface

macedo22: This is said in the documentation for the `radius_outer`, `sphericity_outer`, and `sphericity_inner`, but I think it's worth stating explicitly in the class docs, too, so we'll add that. Thanks!

src/Domain/CoordinateMaps/Wedge.cpp
macedo22: Good question! @AlexCarpenter46 @mar-celine Did we have a specific reason for checking the offset approximately == 0 instead of exactly == 0?

macedo22 (commented):
Accidentally started a review instead of leaving a comment. Submitting my "review" just so I can close it without losing the comments

src/Domain/CoordinateMaps/Wedge.hpp
knelli2: Probably just replace `C` with `\text{const}` so it's clear that it's const, but the value of the constant is less important

src/Domain/CoordinateMaps/Wedge.cpp
knelli2: Discussion on 8/21/24: Make this check for exactly zero. If it is, initialize the class exactly like if there was no offset with default arguments (e.g. for opening angles).

src/Domain/CoordinateMaps/Wedge.hpp
knelli2: Discussion on 8/21/24: Use two constructors for this class. First constructor is the same as before and assumes zero offset (allows fewer changes for existing code). Second takes the offset but omits values that no longer make sense to specify. To combine the cubical and spherical offset part, take `radius_outer` as a `std::optional` (`std::nullopt` means cubical).

src/Domain/CoordinateMaps/Wedge.hpp
knelli2: Discussion on 8/21/24: The focal offset member should probably be a `std::optional`. This can be the sentinel for other things. The other variables that won't be used in one case vs the other can just be initialized to signaling NaN.

macedo22:
@knelli2 sorry I forgot I wasn't on a working branch, I didn't mean to push to the PR. Fixups are still in progress, so just ignore these fixups for right now and I'll ping you when they're ready for review

src/Domain/CoordinateMaps/Wedge.hpp
macedo22: I ended up opting to leave as is since we use language like this in other places since someone might pass a focal offset with zeros, which reduces to a centered Wedge

macedo22:
@knelli2 Fixups addressed, but one unit test for serialization is failing and still trying to fix

src/Domain/CoordinateMaps/Wedge.hpp
```diff
@@ -841,30 +842,92 @@ class Wedge {
...
+   * the Wedge, opening angles cannot be used with offset Wedges.
+   *
+   * In the event that `focal_offset` happens to be zero, the Wedge's member
+   * variables and behavior will be set up to be equivalent to that of a
+   * centered Wedge:
+   * - `cube_half_length` will be discarded and `cube_half_length_` will be set
+   * to `std::nullopt`
+   * - if `radius_outer` is `std::nullopt`, `radius_outer_` will be given the
+   * value $\sqrt{\mathrm{Dim}}L$, where $L$ is the `cube_half_length`
+   * - `opening_angles_` and `opening_angles_distribution_` will be set to
```
knelli2: No need to mention member variables here. Those are an implementation detail that the user doesn't (shouldn't) care about. Only need to mention the constructor args.

Same goes for the rest of the docs for this constructor

src/Domain/CoordinateMaps/Wedge.cpp
```diff
@@ -37,50 +77,100 @@ Wedge<Dim>::Wedge(const double radius_inner, const double radius_outer,
...
+Wedge<Dim>::Wedge(
+    const double radius_inner, const std::optional<double> radius_outer,
+    const double cube_half_length, const std::array<double, Dim> focal_offset,
+    OrientationMap<Dim> orientation_of_wedge, const bool with_equiangular_map,
+    const WedgeHalves halves_to_use, const Distribution radial_distribution)
+    : radius_inner_(radius_inner),
+      radius_outer_((focal_offset == make_array<Dim>(0.0))
+                        ? (radius_outer.has_value()
+                               ? radius_outer.value()
+                               : sqrt(Dim) * cube_half_length)
```
knelli2: `radius_outer.value_or(sqrt(Dim) * cube_half_length)`

macedo22:
@knelli2 Ready for rereview! I think the test failures are unrelated

knelli2 (changes requested):
Looks good. Please squash in these last couple changes

macedo22:
Squashed and rebased 🙌🏻

knelli2 (changes requested):
It seems the CylindricalBCO test is timing out consistently for some CI runs. Probably need to increase its timeout. Should these changes have affected the CylindricalBCO?

macedo22:
It appears so. CylindricalBCO uses Wedge2D

tests/Unit/Domain/Creators/Test_CylindricalBinaryCompactObject.cpp
```diff
@@ -661,7 +661,7 @@ void test_cylindrical_bbh() {
 }
 }  // namespace
 
-// [[TimeOut, 45]]
+// [[TimeOut, 48]]
```
knelli2: Just go ahead and make it 60. You can squash

knelli2 (changes requested):

knelli2 (approved):

PR #6221, opened 2024-08-15
Add protection against negative pressure in `HybridEos`

ncorsobh:
## Proposed changes

Floating point issues can lead to negative pressures in `HybridEos`. This quick PR puts a safeguard against this.

Edit: delving into fixing the unit-tests has revealed another related issue, so I've split the PR into two commits. For the first commit, the `specific_internal_energy_lower_bound` and `specific_internal_energy_upper_bound` functions in 1D EoSs are supposed to provide these bounds as a function of rest mass density. Since in the case of a 1D EoS the internal energy is not a free parameter when density is known, these bounds can only be the exact specific internal energy returned by the 1D EoS.
The second commit deals with the original floating point problem with `HybridEos`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/Hydro/EquationsOfState/EquationOfState.hpp
+++ b/src/PointwiseFunctions/Hydro/EquationsOfState/EquationOfState.hpp
@@ -308,15 +308,14 @@ class EquationOfState<IsRelativistic, 1> : public PUP::able {
   virtual double temperature_upper_bound() const {
     return std::numeric_limits<double>::max();
   };
+
   /// The lower bound of the specific internal energy that is valid for this EOS
-  /// at the given rest mass density \f$\rho\f$
-  virtual double specific_internal_energy_lower_bound(
-      double rest_mass_density) const = 0;
+  virtual double specific_internal_energy_lower_bound() const { return 0.0; };
 
   /// The upper bound of the specific internal energy that is valid for this EOS
-  /// at the given rest mass density \f$\rho\f$
-  virtual double specific_internal_energy_upper_bound(
-      double rest_mass_density) const = 0;
+  virtual double specific_internal_energy_upper_bound() const {
+    return std::numeric_limits<double>::max();
+  };
 
   /// The lower bound of the specific enthalpy that is valid for this EOS
   virtual double specific_enthalpy_lower_bound() const = 0;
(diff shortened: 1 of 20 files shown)


nilsdeppe (dismissed):

nilsdeppe:
Uh oh, tests are unhappy :(

ncorsobh:
@nilsdeppe updated PR, everything should be fixed now. Let me know if the split into two commits was unorthodox and whether a separate PR would be more appropriate.
Also @isaaclegred maybe you could take a look too since this is digging up some EoS changes?

src/PointwiseFunctions/Hydro/EquationsOfState/EquationOfState.hpp
```diff
@@ -311,12 +311,18 @@ class EquationOfState<IsRelativistic, 1> : public PUP::able {
   /// The lower bound of the specific internal energy that is valid for this EOS
   /// at the given rest mass density \f$\rho\f$
   virtual double specific_internal_energy_lower_bound(
-      double rest_mass_density) const = 0;
+      double rest_mass_density) const {
```
nilsdeppe: I think this might break @jyoo1042 's simulations with a 1d EOS. I believe for primitive recovery we need the thermal part effectively unrestricted and trying to restrict it caused the simulations to fail. Isn't what you would like for the 2d EOS to have a density-dependent bound, not the 1d EOS? For a 1d EOS the thermal part doesn't mattr

ncorsobh:
@nilsdeppe PR updated to reflect our last discussion. For posterity's sake, we had agreed that the 1D specific internal energy bounds should lose their density dependence entirely, and that `HybridEos` should set its lower bound for specific internal energy using `cold_eos_.specific_internal_energy_from_density`.
I overwrote the first commit rather than pushing a `fixup` commit since it was pretty much completely overhauled anyway.
For `PolytropicFluid` and `PiecewisePolytropicFluid`, the special upper bounds were chosen to be consistent with the upper bounds set on rest mass density.
@jyoo1042, if the PR is merged in its current state, the main (likely only) thing you'll need to change is to remove density from the function call `specific_internal_energy_lower_bound` (and upper bound) if you're using a 1D EoS.

nilsdeppe (approved):

PR #6223, opened 2024-08-15
Numerically invert Jacobian in combined call

nikwit:
In some maps like `Shape` and `RotScaleTrans`, the inverse jacobian is compute by computing the jacobian and inverting it numerically. This means that for BBH simulations, we currently compute these jacobians twice,  invert them and multiply them together. In this PR, this is changed so that only the jacobian is computed for all maps and the final jacobian is inverted. 

For BBH simulations, I found this speeds up the simulation between 10% and 18%, depending on the number of l-modes.

image

### DIFF
--- a/src/Domain/CoordinateMaps/CoordinateMap.tpp
+++ b/src/Domain/CoordinateMaps/CoordinateMap.tpp
@@ -21,6 +21,7 @@
 #include <utility>
 #include <vector>
 
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
 #include "DataStructures/Tensor/Identity.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Domain/CoordinateMaps/CoordinateMapHelpers.hpp"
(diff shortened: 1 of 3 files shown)


src/Domain/CoordinateMaps/CoordinateMap.tpp
```diff
@@ -557,8 +541,9 @@ auto CoordinateMap<SourceFrame, TargetFrame, Maps...>::
                     InverseJacobian<T, dim, SourceFrame, TargetFrame>,
                     Jacobian<T, dim, SourceFrame, TargetFrame>,
                     tnsr::I<T, dim, TargetFrame>>{
-      tnsr::I<T, dim, TargetFrame>(std::move(mapped_point)), std::move(inv_jac),
-      std::move(jac), std::move(frame_velocity)};
+      tnsr::I<T, dim, TargetFrame>(std::move(mapped_point)),
+      determinant_and_inverse(jac).second, std::move(jac),
+      std::move(frame_velocity)};
```
knelli2, line 546: Add in the `CoordinateMap.hpp` docs for this function that we don't call the individual `inverse_jacobian` functions and multiply them together, but rather just compute the jacobian and numerically invert it because most of our production time-dependent maps do this already.

knelli2 (changes requested):

src/Domain/CoordinateMaps/CoordinateMap.tpp
wthrowe: Is there a reason to keep the `inverse_jacobian` implementations on the individual maps at all?  Doesn't need to be done in this PR, but removing them would be nicer than documenting that they are unused.

knelli2: Well they are still used in the individual `inverse_jacobian` function of `CoordinateMap{Base}`, correct? Whether or not that function is actually used in practice, I don't know. If not, then yes we can probably get rid of it

wthrowe: Or we could make it do the numerical inversion like this function now does.

knelli2: Yeah we could! Let's discuss on the Monday call

nilsdeppe: Another thing for a separate PR: this function could also return the Jacobian determinant. We need this a bit in the hydro cases.

nilsdeppe (commented):
I'm happy with the code. Thanks for doing this!

src/Domain/CoordinateMaps/CoordinateMap.tpp
```diff
@@ -557,8 +541,9 @@ auto CoordinateMap<SourceFrame, TargetFrame, Maps...>::
                     InverseJacobian<T, dim, SourceFrame, TargetFrame>,
                     Jacobian<T, dim, SourceFrame, TargetFrame>,
                     tnsr::I<T, dim, TargetFrame>>{
-      tnsr::I<T, dim, TargetFrame>(std::move(mapped_point)), std::move(inv_jac),
-      std::move(jac), std::move(frame_velocity)};
+      tnsr::I<T, dim, TargetFrame>(std::move(mapped_point)),
+      determinant_and_inverse(jac).second, std::move(jac),
```
nilsvu, line 545: Make sure to add a comment that this done numerically on purpose and was profiled.

nikwit:
I added a comment about the inverse Jacobian being inverted numerically in `CoordinateMap.hpp`. @nilsdeppe the function can be altered to return the determinant (this is free now as you get it automatically with the inversion). I am not sure how much of a pain it is but probably not too bad, I can look into it later this week.

nilsdeppe:
Sounds good. Probably would require updating some tags. I think it's low priority :) Thank you!

knelli2 (approved):

src/Domain/CoordinateMaps/CoordinateMap.tpp
knelli2: From the Monday call: We can just numerically invert the jacobian in the individual function as well which means we don't need the individual `inverse_jacobian` functions. See #6233

PR #6224, opened 2024-08-15
Use PostNewtonian.jl for initial orbital parameters

nilsvu:
## Proposed changes

Replaces the dependence on SpEC to compute low-eccentricity initial orbital parameters with the `sxs` package. It provides the PN approximations at higher PN order than SpEC, is much faster, and avoids spurious output from old Fortran code (LSODA) that was used in SpEC.
The PN approximations are provided through the `PostNewtonian.jl` Julia package, so Julia will be downloaded on first use, which may take a few minutes.

See also https://github.com/moble/PostNewtonian.jl/issues/41, which made this possible.

### Upgrade instructions

To use low-eccentricity initial orbital parameters for BBH initial data, install the `sxs` package in your Python environment:

```sh
# In the build directory:
./bin/python-spectre -m pip install sxs
```

Now you can use `./bin/spectre bbh generate-id` with the `--eccentricity=0` option. You don't need SpEC for this anymore.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/EccentricityControl/InitialOrbitalParameters.py
+++ b/support/Pipelines/EccentricityControl/InitialOrbitalParameters.py
@@ -7,9 +7,52 @@ from typing import Optional, Sequence, Tuple
 import numpy as np
 from scipy.optimize import minimize
 
-from spectre.support.CheckSpecImport import check_spec_import
-
 logger = logging.getLogger(__name__)
+
+
+# The following two functions are modernized versions of those in SpEC's
+# `ZeroEccParamsFromPN.py`. They use higher PN orders (whichever are implemented
+# in the PostNewtonian module), are much faster, and avoid spurious output from
+# old Fortran code (LSODA) that was used in SpEC's `ZeroEccParamsFromPN.py`.
+# They are consistent with SpEC up to 2.5 PN order, as tested by Mike Boyle (see
+# https://github.com/moble/PostNewtonian.jl/issues/41).
+#
+# Since these functions use Julia through Python bindings, they will download
+# Julia and precompile the packages on first use, which may take a few minutes
+# (see https://moble.github.io/PostNewtonian.jl/dev/interface/python/).
+
+
+def omega_and_adot(r, q, chiA, chiB):
+    from sxs.julia import PostNewtonian
+
+    # The BBH system state vector is documented here:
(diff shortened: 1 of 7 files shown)


support/Pipelines/EccentricityControl/InitialOrbitalParameters.py
```diff
@@ -10,6 +10,43 @@
...
+# Julia and precompile the packages on first use, which may take a few minutes
+# (see https://moble.github.io/PostNewtonian.jl/dev/interface/python/).
+
+
+def omega_and_adot(r, q, chiA, chiB):
+    from sxs.julia import PostNewtonian
+
+    pn = PostNewtonian.BBH(
+        np.array(
+            [1.0 / (1.0 + q), q / (1.0 + q), *chiA, *chiB, 1, 0, 0, 0, 1, 0]
```
knelli2: Please add a note what these magic 1s and 0s are. I believe they are the following, but the docs/code of the julia implementation don't match so I'm unsure...

1 = v (velocity???)
0, 0, 0, 1 = rotor (quaternion??)
0 = phi

support/Pipelines/EccentricityControl/InitialOrbitalParameters.py
```diff
@@ -10,6 +10,43 @@
...
+        )
+    )
+    pn.state[12] = PostNewtonian.separation_inverse(r, pn)
```
knelli2: Why replace? Please add some docs

tests/support/Pipelines/EccentricityControl/CMakeLists.txt
```diff
@@ -9,7 +9,9 @@ spectre_add_python_bindings_test(
   None
   TIMEOUT 60)
 
-if (SpEC_FOUND)
+# Disable this test if using jemalloc as a shared library, because Julia
+# doesn't like the 'LD_PRELOAD' trick to load jemalloc for Pybindings.
+if (NOT "${JEMALLOC_LIB_TYPE}" STREQUAL SHARED)
```
knelli2: Will the source code work in the CLI if we are using JEMALLOC?

knelli2 (changes requested):
Currently only HPC has a sufficient version of `sxs`. Ocean and mbot will need to be updated

knelli2:
Also needs a rebase to fix conflicts

nilsvu:
This is now rebased and ready for review

PR #6227, opened 2024-08-16
Combine Shape map calls

nikwit:
Currently, the shape map computes the coordinates, frame velocity and jacobians separately which duplicates a lot of work such as computing the spherical harmonic expansion.

This PR adds a combined call to these three quantities. I found that this speeds up simulations between 16 % and 27% (when combined with #6223)

depends on #6223 

image

### DIFF
--- a/src/Domain/CoordinateMaps/CoordinateMap.tpp
+++ b/src/Domain/CoordinateMaps/CoordinateMap.tpp
@@ -42,6 +42,30 @@ namespace CoordinateMap_detail {
 CREATE_IS_CALLABLE(function_of_time_names)
 CREATE_IS_CALLABLE_V(function_of_time_names)
 
+template <size_t Dim, typename T>
+using combined_coords_frame_velocity_jacs_t =
+    decltype(std::declval<T>().coords_frame_velocity_jacobian(
+        std::declval<gsl::not_null<std::array<DataVector, Dim>*>>(),
+        std::declval<gsl::not_null<std::array<DataVector, Dim>*>>(),
+        std::declval<
+            gsl::not_null<tnsr::Ij<DataVector, Dim, Frame::NoFrame>*>>(),
+        std::declval<const double>(),
+        std::declval<const std::unordered_map<
+            std::string,
+            std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&>()));
+
+template <size_t Dim, typename T, typename = std::void_t<>>
+struct has_combined_coords_frame_velocity_jacs : std::false_type {};
+
+template <size_t Dim, typename T>
+struct has_combined_coords_frame_velocity_jacs<
+    Dim, T, std::void_t<combined_coords_frame_velocity_jacs_t<Dim, T>>>
+    : std::true_type {};
+
+template <size_t Dim, typename T>
+inline constexpr bool has_combined_coords_frame_velocity_jacs_v =
+    has_combined_coords_frame_velocity_jacs<Dim, T>::value;
+
 template <typename T>
 struct map_type {
   using type = T;
@@ -449,36 +473,56 @@ auto CoordinateMap<SourceFrame, TargetFrame, Maps...>::
   tuple_transform(
       maps_,
       [&frame_velocity, &jac, &mapped_point, time, &functions_of_time](
-          const auto& map, auto index, const std::tuple<Maps...>& maps) {
+          const auto& map, auto index, const std::tuple<Maps...>& /*maps*/) {
         constexpr size_t count = decltype(index)::value;
         using Map = std::decay_t<decltype(map)>;
+        static constexpr bool is_time_dependent =
+            domain::is_map_time_dependent_v<Map>;
+        static constexpr bool use_combined_call =
+            CoordinateMap_detail::has_combined_coords_frame_velocity_jacs_v<
+                dim, Map> and
+            std::is_same_v<T, DataVector>;
+
+        [[maybe_unused]] std::array<T, dim> noframe_frame_velocity{};
         tnsr::Ij<T, dim, Frame::NoFrame> noframe_jac{};
+        if constexpr (use_combined_call) {
+          map.coords_frame_velocity_jacobian(
+              make_not_null(&mapped_point),
+              make_not_null(&noframe_frame_velocity),
+              make_not_null(&noframe_jac), time, functions_of_time);
+        } else {
+          // if the map is the identity we do not compute it unless it is the
+          // first map, then it is used for initialization
+          if (not(map.is_identity() and count != 0)) {
+            ::domain::detail::get_jacobian(
+                make_not_null(&noframe_jac), map, mapped_point, time,
+                functions_of_time,
+                domain::is_jacobian_time_dependent_t<Map, T>{});
+            if constexpr (is_time_dependent) {
+              noframe_frame_velocity = domain::detail::get_frame_velocity(
+                  map, mapped_point, time, functions_of_time);
+            }
+            CoordinateMap_detail::apply_map(
+                make_not_null(&mapped_point), map, time, functions_of_time,
+                domain::is_map_time_dependent_t<decltype(map)>{});
+          }
+        }
         if constexpr (count == 0) {
-          // Set Jacobian
-          ::domain::detail::get_jacobian(
-              make_not_null(&noframe_jac), map, mapped_point, time,
-              functions_of_time,
-              domain::is_jacobian_time_dependent_t<Map, T>{});
           for (size_t target = 0; target < dim; ++target) {
             for (size_t source = 0; source < dim; ++source) {
               jac.get(target, source) =
                   std::move(noframe_jac.get(target, source));
             }
           }
-
           // Set frame velocity
-          if constexpr (domain::is_map_time_dependent_v<std::tuple_element_t<
-                            0, std::decay_t<decltype(maps)>>>) {
-            std::array<T, dim> noframe_frame_velocity =
-                ::domain::detail::get_frame_velocity(map, mapped_point, time,
-                                                     functions_of_time);
+          if constexpr (is_time_dependent) {
             for (size_t i = 0; i < dim; ++i) {
               frame_velocity.get(i) =
                   std::move(gsl::at(noframe_frame_velocity, i));
             }
           } else {
-            // If the first map is time-independent the velocity is initialized
-            // to zero
+            // If the first map is time-independent the velocity is
+            // initialized to zero
             for (size_t i = 0; i < frame_velocity.size(); ++i) {
               frame_velocity[i] = make_with_value<T>(get<0, 0>(jac), 0.0);
             }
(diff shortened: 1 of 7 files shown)


src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -335,6 +336,218 @@ tnsr::Ij<tt::remove_cvref_wrap_t<T>, 3, Frame::NoFrame> Shape::jacobian(
   return result;
 }
 
+void Shape::coords_frame_velocity_jacobian(
+    gsl::not_null<std::array<DataVector, 3>*> target_coords,
```
knelli2: To avoid duplicating code and potentially introducing a bug in the future when only one function is updated but another isn't, I think you should factor out the bulk of the `operator()`, frame velocity, and jacobian functions into free functions that take some common arguments (like centered coords, theta_phis, functions of time)

src/Domain/CoordinateMaps/CoordinateMap.tpp
```diff
@@ -448,35 +449,56 @@ auto CoordinateMap<SourceFrame, TargetFrame, Maps...>::
...
       [&frame_velocity, &jac, &mapped_point, time, &functions_of_time](
-          const auto& map, auto index, const std::tuple<Maps...>& maps) {
+          const auto& map, auto index, const std::tuple<Maps...>& /*maps*/) {
         constexpr size_t count = decltype(index)::value;
         using Map = std::decay_t<decltype(map)>;
+        static constexpr bool is_time_dependent =
+            domain::is_map_time_dependent_v<Map>;
+        static constexpr bool use_combined_shape_call =
+            std::is_same_v<CoordinateMaps::TimeDependent::Shape, Map> and
+            std::is_same_v<T, DataVector>;
```
knelli2: Instead of checking if this is a shape map explicitly (and thus adding an strange dependency for the general coord map on the time dependent maps), add a meta function that checks if the `Map` object has a function named `coords_frame_velocity_jacobian` that can be called with the given arguments.

Then this easily allows us to edit the other time dependent maps if we want to to add this feature to those as well

knelli2 (changes requested):
Thank you for investigating this!

src/Domain/CoordinateMaps/TimeDependent/Shape.hpp
```diff
@@ -258,6 +258,7 @@ class Shape {
   size_t l_max_ = 2;
   size_t m_max_ = 2;
   ylm::Spherepack ylm_{2, 2};
+  ylm::Spherepack extended_ylm_{2, 2};
```
nilsdeppe: `3, 3`?

src/Domain/CoordinateMaps/CoordinateMap.tpp
nilsdeppe: Note: I think we could get away with a forward declaration of the map to avoid the coupling. Just providing another option.

nilsdeppe (commented):
LGTM, no need for my approval to merge.

nikwit:
@knelli2 I have factored out large chunks of the jacobian, I do not think this is worth doing for the operator() or frame velocity as it is just two statements. I also avoid duplicate calls to the transition_function in the combined call which I had overlooked before.

src/Domain/CoordinateMaps/TimeDependent/Shape.hpp
```diff
@@ -251,7 +251,7 @@ class Shape {
       gsl::not_null<std::array<DataVector, 3>*> target_coords,
       gsl::not_null<std::array<DataVector, 3>*> frame_vel,
       gsl::not_null<tnsr::Ij<DataVector, 3, Frame::NoFrame>*> jac,
-      const std::array<DataVector, 3> source_coords, double time,
+      const std::array<DataVector, 3> source_coords, const double time,
```
knelli2: Make `source_coords` a reference (you'll have to edit the metafunction as well). Also no `const` for `time` in the declaration

knelli2 (changes requested):
Looks good! You can squash this last change in with everything else

src/Domain/CoordinateMaps/TimeDependent/Shape.hpp
nikwit: Yes, I was unsure about this. The coords_frame_velocity_jac function in `CoordinateMap.tpp` currently uses this method with the reference return coords object and this const argument having the same address (the source coords are overwritten with the target coords). This  is why I copied it to be safe. It should be fine either way as the arg is used only before the reference return object is touched by it still feels uncomfortable to have it this way. What is the usual C++ policy on this?

knelli2: Ohh I see now. In `src/Domain/CoordinateMaps/CoordinateMapHelpers.hpp` for the `apply_map` function, we just pass the coords through the `gsl::not_null` part and use that as the source coords. We can probably do the same thing here

knelli2 (changes requested):
Also fix the compilation errors from gcc-9 and the clang-tidy issues. The one about using `l` as a variable you can ignore or add it to our `.clang-tidy` file because it's pretty silly

nikwit:
Thanks for your reviews. I changed everything but the long link `domain::CoordinateMaps::ShapeMapTransitionFunctions::ShapeMapTransitionFunction::original_radius_over_radius` in the documentation Shape.hpp:162 is over 80 characters and not resolved if I break it up over two lines. Does someone know how this can be resolved?

src/Domain/CoordinateMaps/TimeDependent/Shape.hpp
```diff
@@ -241,6 +243,16 @@ class Shape {
       const std::array<T, 3>& source_coords, double time,
       const FunctionsOfTimeMap& functions_of_time) const;
 
+  /*!
+   * \brief An optimized call that computes the target coordinates, frame
+   * velocity and jacobian at once to avoid duplicate calculations.
+   */
+  void coords_frame_velocity_jacobian(
+      gsl::not_null<std::array<DataVector, 3>*> source_and_target_coords,
```
knelli2: Sorry one last thing since I know I'll get confused in the future. Can you add `\details` about how the source coords are expected to be passed in to `source_and_target_coords`, but the resulting mapped coords are then also placed in `source_and_target_coords` at the end of the function once the source coords have been used.

knelli2 (changes requested):
You can squash directly. As for the long link, our commit hooks check for linked that contain either `\link` or `\endlink` and ignore those, so just make sure one of those is on the same line as the link.

nikwit:
all done and squashed in, thank you for the review!

knelli2 (dismissed):
Thanks for optimizing this!

knelli2:
This test failure is concerning me since I (and @nilsvu) have never seen it before and I want to make sure it's not related to the shape map changes. The test failure happened twice in a row with the same error

nikwit:
oh, that is weird. So it only fails with clang-13 in Release mode. I don't know the executable at all. Would this executable call the `coords_framevel_jac` function of `CoordinateMap`? If so, do you know which compile time and runtime path it would take in that function? @nilsvu

nikwit:
Looks like reducing the `LMax` reduced the weird test failure of clang-13 in the Xcts executable @knelli2

nikwit:
@knelli2 @nilsvu @nilsdeppe I looked into the test failure a bit more. It seems that the test failure is unrelated to combining the shape map calls. I can "fix" the failure by simply adding or removing some temporaries in the shape map that change the evaluation order of operations (all additions and multiplications). One change that causes the test to fail is e.g. here, where I just introduce a temporary that should definitely be safe: https://github.com/nikwit/spectre/commit/6ecb7300cc05bcb01d4e95334616fcb83f1612c2

Also, simply inlining the `jacobian_helper` function causes the test to work.

I am therefore convinced that this PR simply changes some floating point precision that propagates through the elliptic solver and changes the residuals that are computed (which are still convergent). IMO, we should simply raise the tolerance of that test.

knelli2:
@nilsvu I think it is up to you what to do here since this only affects the elliptic solver.

Also @nikwit this has a conflict now and needs a rebase

nilsvu:
I ran this branch on my machine and the test passes. It looks like specifically clang-13 Release fails, and specifically the first GMRES iteration after AMR iteration is different. The shape map gets evaluated on the new refined grid points, so at least it's plausible that an issue in the shape map can cause this failure. I worries me enough that I wouldn't just reduce the Lmax to "fix" it without checking that BBH evolutions aren't affected. Suggestions:

- How about adding a regression test to GH/KerrSchild.yaml like the one in Xcts/KerrSchild.yaml, to check that the first couple of time steps remain identical? This should use the same Kerr spin and shape map Lmax that caused issues in XCTS.
- `ylm::Spherepack` does some fancy caching in a `static thread_local` memory pool, maybe that's causing issues somehow? Just speculating. I guess you could try making it a normal member variable by removing the `static thread_local` just to see if that affects the issue.
- You could also run a BBH on clang-13 Release to make sure the results are the same before and after this PR (to numerical roundoff error, not just truncation error).

It's a very mysterious issue, and I just want to ensure that it doesn't affect our BBH simulations in mysterious ways 😅

nikwit:
@nilsvu I think my analysis above shows that the issue is related simply to the order in which expressions are evaluated in the shape map  (https://github.com/nikwit/spectre/commit/6ecb7300cc05bcb01d4e95334616fcb83f1612c2 fixes it!). The test appears to blow up a floating point difference to a difference of 1e-9 somehow. I do not see how this can be related to `ylm::Spherepack`. 

My suspicion is that the test is not a fair test. Can you describe where the residuals that are compared in the test come from? It looks like the test uses 4 grid points per element and no h-refinement. How can this lead to a residual of 1e-11?

nilsvu:
> Can you describe where the residuals that are compared in the test come from? It looks like the test uses 4 grid points per element and no h-refinement. How can this lead to a residual of 1e-11?

The refinement of the domain and the DG scheme defines a discretized problem $A(u_q)_p = b_p$ over the grid points (it's a matrix equation if the problem is linear). The residual is the $L_2$ norm $|| A(u_q)_p - b_p||_2$ over the grid points. This discretized problem can be solved for the values $u_p$ down to machine precision (or to 1e-11 in this case), independent of resolution. If the problem is linear it's just a matrix inversion, which can be done to machine precision. This accuracy is only the numerical solution to the discretized problem, it doesn't mean the _truncation error_ is so small as well. The truncation error is limited by how well the discretized problem represents the true problem, and is ~1e-3 in this case (see the "Discretization error" check in KerrSchild.yaml, which compares the solution $u_p$ to the analytic solution at every grid point).

Now I don't see how the numerical algorithm that solves the discretized problem should blow up roundoff error in the shape map to truncation error level differences, though I might be missing something of course.

nikwit:
alright,  I see.

At this point I would argue that since changing the order in which `DataVector`s are evaluated fixes the test, this must mean  that any issue that may or may not be in the code is unrelated to this PR. If anything, the failing test is a symptom of an issue already on develop (though I do not believe there is one).

nilsvu:
I have now seen two other cases of this test failing on CI, so I agree that it is not directly related to the changes in this PR and I'm happy with this PR getting merged. I still have no idea what causes the issue though :(

knelli2:
@nikwit Just needs another rebase to fix the conflict and then I'll merge this. @nilsvu Thanks for looking into this!

nikwit:
rebased. Thank you all for your due diligence!

knelli2 (approved):

PR #6230, opened 2024-08-18
Project mortar data

kidder:
## Proposed changes

This projects the Boundary History for p-refinement.  This is needed for AMR with LTS.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
+++ b/src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
@@ -72,6 +72,98 @@ mortars_apply_impl(const std::vector<std::array<size_t, Dim>>& initial_extents,
                    Spectral::Quadrature quadrature, const Element<Dim>& element,
                    const TimeStepId& next_temporal_id,
                    const Mesh<Dim>& volume_mesh);
+
+template <size_t Dim, typename MortarDataHistoryType>
+void p_project(
+    const gsl::not_null<
+        ::dg::MortarMap<Dim, evolution::dg::MortarDataHolder<Dim>>*>
+    /* mortar_data */,
+    const gsl::not_null<::dg::MortarMap<Dim, Mesh<Dim - 1>>*> mortar_mesh,
+    const gsl::not_null<
+        ::dg::MortarMap<Dim, std::array<Spectral::MortarSize, Dim - 1>>*>
+    /* mortar_size */,
+    const gsl::not_null<::dg::MortarMap<Dim, TimeStepId>*>
+    /* mortar_next_temporal_id */,
+    const gsl::not_null<
+        DirectionMap<Dim, std::optional<Variables<tmpl::list<
+                              evolution::dg::Tags::MagnitudeOfNormal,
+                              evolution::dg::Tags::NormalCovector<Dim>>>>>*>
+        normal_covector_and_magnitude,
+    const gsl::not_null<MortarDataHistoryType*> mortar_data_history,
+    const Mesh<Dim>& new_mesh, const Element<Dim>& new_element,
+    const std::unordered_map<ElementId<Dim>, amr::Info<Dim>>& neighbor_info,
+    const std::pair<Mesh<Dim>, Element<Dim>>& old_mesh_and_element) {
+  const auto& [old_mesh, old_element] = old_mesh_and_element;
+  ASSERT(old_element.id() == new_element.id(),
+         "p-refinement should not have changed the element id");
+
+  const bool mesh_changed = old_mesh != new_mesh;
+
+  for (const auto& [direction, neighbors] : new_element.neighbors()) {
+    const auto sliced_away_dimension = direction.dimension();
+    const auto old_face_mesh = old_mesh.slice_away(sliced_away_dimension);
+    const auto new_face_mesh = new_mesh.slice_away(sliced_away_dimension);
+    const bool face_mesh_changed = old_face_mesh != new_face_mesh;
+    if (face_mesh_changed) {
+      (*normal_covector_and_magnitude)[direction] = std::nullopt;
+    }
+    for (const auto& neighbor : neighbors) {
+      const DirectionalId<Dim> mortar_id{direction, neighbor};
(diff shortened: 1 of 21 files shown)


tests/Unit/Domain/Structure/Test_OrientationMapHelpers.cpp
```diff
@@ -567,7 +585,7 @@ void test_2d_slice_with_orientation(const OrientationMap<3>& orientation_map) {
     get(get<ScalarTensor>(expected_vars)) = oriented_mapped_coords[0];
     get<0>(get<Coords<2>>(expected_vars)) = oriented_mapped_coords[0];
     get<1>(get<Coords<2>>(expected_vars)) = oriented_mapped_coords[1];
-    CHECK(oriented_vars == expected_vars);
+    CHECK_VARIABLES_APPROX(oriented_vars, expected_vars);
```
wthrowe, line 588: What's this change about?  Seems unrelated to the rest of the commit.

tests/Unit/Evolution/DiscontinuousGalerkin/Test_MortarDataHolder.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "Framework/TestHelpers.hpp"
+#include "Utilities/GetOutput.hpp"
+#include "Utilities/MakeString.hpp"
+
+namespace evolution::dg {
+namespace {
+template <size_t Dim>
+void test() {
+  MortarDataHolder<Dim> holder{};
+  std::string expected_output = MakeString{} << "Local mortar data:\n"
```
wthrowe: This could just be a literal instead of using `MakeString`.

tests/Unit/Evolution/DiscontinuousGalerkin/Test_MortarDataHolder.cpp
```diff
@@ -0,0 +1,47 @@
...
+#include "Evolution/DiscontinuousGalerkin/MortarDataHolder.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "Utilities/GetOutput.hpp"
+#include "Utilities/MakeString.hpp"
+
+namespace evolution::dg {
+namespace {
+template <size_t Dim>
+void test() {
+  MortarDataHolder<Dim> holder{};
```
wthrowe, line 38: [optional] It would be nice if this test tested something beyond an empty state, but since this is adding a test for a previously untested existing class I won't require it.

src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
```diff
@@ -72,6 +72,96 @@ mortars_apply_impl(const std::vector<std::array<size_t, Dim>>& initial_extents,
...
+        normal_covector_and_magnitude,
+    const gsl::not_null<MortarDataHistoryType*> mortar_data_history,
+    const Mesh<Dim>& new_mesh, const Element<Dim>& new_element,
+    const std::unordered_map<ElementId<Dim>, amr::Info<Dim>>& neighbor_info,
+    const std::pair<Mesh<Dim>, Element<Dim>>& old_mesh_and_element) {
+  const auto& [old_mesh, old_element] = old_mesh_and_element;
+  ASSERT(old_element.id() == new_element.id(),
+         "p-refinement should not have changed the element id");
+
+  std::unordered_set<DirectionalId<Dim>> old_mortar_ids;
```
wthrowe: No need to copy the hash table to another hash table.  The `old_mortar_ids.contains(mortar_id)` check below is the same as `mortar_mesh->contains(mortar_id)`.

src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
```diff
@@ -72,6 +72,96 @@ mortars_apply_impl(const std::vector<std::array<size_t, Dim>>& initial_extents,
...
+            const auto project_boundary_data =
+                [&new_mortar_mesh, &new_face_mesh, &new_mesh](
+                    const TimeStepId& /* id */,
+                    const gsl::not_null<::evolution::dg::MortarData<Dim>*>
+                        mortar_data) {
+                  p_project(mortar_data, new_mortar_mesh, new_face_mesh,
+                            new_mesh);
+                };
+            local_history.for_each(project_boundary_data);
+            remote_history.for_each(project_boundary_data);
```
wthrowe: This passes the wrong face and volume meshes to `p_project` for the remote side, but I think either the associated remote fields are always unset or the values are never used.  Assuming that's true, add a comment noting that that's why it's OK.

wthrowe (changes requested):

tests/Unit/Domain/Structure/Test_OrientationMapHelpers.cpp
kidder: When I adapted the existing test to also test my new function, it changed the extents of the mesh used to fill oriented_vars which are set based on the logical coords.  This led to a roundoff difference instead of exact equality.

tests/Unit/Evolution/DiscontinuousGalerkin/Test_MortarDataHolder.cpp
kidder: ok, updated

src/Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp
kidder: I added a new project function that only projects the mortar data to make this clear.  (I am adding the new function for the event-based refinement PR, so just seemed easier to add it now.)

tests/Unit/Domain/Structure/Test_OrientationMapHelpers.cpp
wthrowe: OK.

wthrowe (commented):
You have some warnings and clang-tidy problems, otherwise looks good.  Squash.

PR #6232, opened 2024-08-19
Fix compatibility with Numpy 2.1

nilsvu:
## Proposed changes

Something about symbol visibility changed in Numpy 2.1: https://numpy.org/doc/stable/reference/c-api/array.html#c.NPY_API_SYMBOL_ATTRIBUTE

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2 (approved):

knelli2 (changes requested):
The boost from Homebrew was updated from 1.85.0 to 1.86.0 and this is causing things to fail in macOS CI. Can you fix that as well here and update the installation page if we have to keep boost below 1.86.0?

nilsvu:
Added a fix for Boost v1.86.

knelli2 (approved):

PR #6239, opened 2024-08-22
Make FixToAtmosphere fix the electron fraction in atmosphere

isaaclegred:
## Proposed changes

Currently we do not perform atmosphere fixing in the hydro evolution after Finite Difference reconstruction, which can result in unphysical values of primitive variables being passed to equation of state calls leading to floating point exceptions.  

Additionally, atmosphere fixing currently doesn't fix the electron fraction, which is necessary in simulations using a 3D EoS.  

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

The reason why these problems are apparently coupled is because in primitive recovery unphysical electron fraction values are limited to the max/min of the EoS.  However, in reconstruction the primitives are not the direct output of a con2prim call, and therefore the electron fraction is not generally snapped back to EoS bounds, and this needs to be done explicitly.

### DIFF
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/ReconstructWork.tpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/ReconstructWork.tpp
@@ -17,14 +17,15 @@
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "DataStructures/Variables.hpp"
 #include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/DirectionMap.hpp"
 #include "Domain/Structure/DirectionalId.hpp"
 #include "Domain/Structure/DirectionalIdMap.hpp"
-#include "Domain/Structure/DirectionMap.hpp"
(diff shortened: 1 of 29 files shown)


src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/ReconstructWork.tpp
```diff
@@ -71,16 +76,31 @@ void compute_conservatives_for_reconstruction(
...
+      make_not_null(&specific_internal_energy),
+      make_not_null(&spatial_velocity),
+      make_not_null(&lorentz_factor),
+      make_not_null(&pressure),
+      make_not_null(&temperature),
+      make_not_null(&electron_fraction),
+      spatial_metric,
+      eos);
+    ASSERT(min(rest_mass_density.get()) >= 0.0, "rest mass density should be positive");
+   ASSERT(min(electron_fraction.get()) >= 0.0, "rest mass density should be positive");
```
ncorsobh: I think the statement should be "electron fraction should be positive" here.

PR #6234, opened 2024-08-19
Allow offset in Wedge transition function

knelli2:
## Proposed changes

Complements #6214. In order to have offset excisions in the BCO domain, the Wedge shape map transition function also needed to be updated to allow non-centered excisions.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #6243, opened 2024-08-22
Add Wedge Offset Option to BCO

### LINKED ISSUE #6092: BBH domain: translate inner spheres to increase size of cubes

AlexCarpenter46:
## Proposed changes

This is 3rd PR in the chain that adds the ability to offset the wedges. This PR adds the CubeScale option to BCO domain which will offset the wedge if your CubeScale is larger than 1.0. Specifying CubeScale as 1.0 makes the cube size the initial separation between the two objects.

### Upgrade instructions

Add the CubeScale option in the DomainCreator->BinaryCompactObject section like this.
```
DomainCreator:
  BinaryCompactObject:
    CubeScale: 1.0
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

This PR depends on #6234 and #6214


src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -111,10 +112,26 @@ BinaryCompactObject::BinaryCompactObject(
   // Determination of parameters for domain construction:
   const double tan_half_opening_angle = tan(0.5 * opening_angle_);
   translation_ = 0.5 * (x_coord_a_ + x_coord_b_);
-  length_inner_cube_ = abs(x_coord_a_ - x_coord_b_);
+  if (cube_length.has_value()) {
+    length_inner_cube_ = cube_length.value();
```
knelli2: Use `cube_length.value_or(x_coord_a_ - x_coord_b_)`

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -111,10 +112,26 @@ BinaryCompactObject::BinaryCompactObject(
...
+  } else {
+    length_inner_cube_ = x_coord_a_ - x_coord_b_;
+  }
   length_outer_cube_ =
       2.0 * envelope_radius_ / sqrt(2.0 + square(tan_half_opening_angle));
 
+  if (equal_within_roundoff(length_inner_cube_, (x_coord_a_ - x_coord_b_),
+                            std::numeric_limits<double>::epsilon() * 100.0,
+                            length_inner_cube_)) {
```
knelli2, line 132: Please add a comment here explaining that we choose equal within roundoff instead of exactly equal because the `Wedge` map expects exact equality when checking if there is a zero offset. Therefore we handle roundoff differences here, rather than there

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -153,6 +170,12 @@ BinaryCompactObject::BinaryCompactObject(
         "malformed. A recommended radius is:\n"
             << suggested_value);
   }
+  if (length_inner_cube_ < (x_coord_a_ - x_coord_b_)) {
+    PARSE_ERROR(
```
knelli2: This check should probably go right after you set `length_inner_cube_`

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -200,6 +223,18 @@ BinaryCompactObject::BinaryCompactObject(
           "or neither.");
     }
   }
+  const bool filled_excision_a =
+      (not use_single_block_a_ and not is_excised_a_);
```
knelli2: This is just

```cpp
const bool filled_excision_a = not (use_single_block_a_ or is_excised_a_);
```

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -395,17 +446,19 @@ Domain<3> BinaryCompactObject::create_domain() const {
     Maps maps_center_A =
         domain::make_vector_coordinate_map_base<Frame::BlockLogical,
                                                 Frame::Inertial, 3>(
-            sph_wedge_coordinate_maps(object_a.inner_radius,
-                                      object_a.outer_radius, inner_sphericity_A,
-                                      1.0, use_equiangular_map_, false, {},
-                                      object_A_radial_distribution),
+            sph_wedge_coordinate_maps(
+                object_a.inner_radius, object_a.outer_radius,
+                inner_sphericity_A, 1.0, length_inner_cube_ / 2.0,
```
knelli2: Multiply by `0.5` instead of dividing by `2.0`. Same for the rest of the changes in this file

src/Domain/DomainHelpers.hpp
```diff
@@ -178,7 +181,8 @@ size_t which_wedge_index(const ShellWedges& which_wedges);
  */
 std::vector<domain::CoordinateMaps::Wedge<3>> sph_wedge_coordinate_maps(
     double inner_radius, double outer_radius, double inner_sphericity,
-    double outer_sphericity, bool use_equiangular_map,
+    double outer_sphericity, double cube_half_length,
+    std::array<double, 3> focal_offset, bool use_equiangular_map,
```
knelli2: Combine these two parameters into `std::optional<std::pair<double, std::array<double, 3>>>`, move the argument after `use_equiangular_map`, and make it default to `std::nullopt`. That way you don't have to make up values when it doesn't make sense to specify these arguments. And there's less changes to existing function calls for this.

tests/Unit/Domain/Creators/TimeDependentOptions/Test_BinaryCompactObject.cpp
```diff
@@ -599,7 +606,7 @@ void test_errors() {
             std::nullopt};
         time_dep_opts.build_maps(
             std::array{std::array{0.0, 0.0, 0.0}, std::array{0.0, 0.0, 0.0}},
-            radii, radii, 25.0, 100.0);
+            std::nullopt, std::nullopt, radii, radii, 25.0, 100.0);
       }()),
       Catch::Matchers::ContainsSubstring(
```
knelli2, line 612: Can you add an error test for specifying an offset with the cylindrical BCO?

tests/Unit/Domain/Test_DomainHelpers.cpp
```diff
@@ -94,12 +94,13 @@ void test_periodic_different_blocks() {
 }
 
 std::vector<CoordinateMaps::Wedge<3>> test_wedge_map_generation(
-    double inner_radius, double outer_radius, double inner_sphericity,
+    double inner_radius, double outer_radius, double cube_half_length,
+    std::array<double, 3> focal_offset, double inner_sphericity,
```
knelli2: Same here about function arguments

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_InitializeElementFacesGridCoordinates.cpp
```diff
@@ -125,6 +125,7 @@ SPECTRE_TEST_CASE(
         std::array<double, 2>{{0.1, 0.2}},
         30.,
         50.,
+        13.0,
```
knelli2: Make this `std::nullopt` so it's guaranteed to be the separation even if the x-coords of this test change. Same for the rest of these tests that only use the BCO domain

knelli2 (changes requested):
Thanks for doing this! Please add a fixup with these changes

src/Domain/DomainHelpers.hpp
```diff
@@ -181,8 +181,9 @@ size_t which_wedge_index(const ShellWedges& which_wedges);
  */
 std::vector<domain::CoordinateMaps::Wedge<3>> sph_wedge_coordinate_maps(
     double inner_radius, double outer_radius, double inner_sphericity,
-    double outer_sphericity, double cube_half_length,
-    std::array<double, 3> focal_offset, bool use_equiangular_map,
+    double outer_sphericity, bool use_equiangular_map,
+    std::optional<std::pair<double, std::array<double, 3>>> offset_options =
```
knelli2: Make `offset_options` a reference

src/Domain/Creators/BinaryCompactObject.cpp
```diff
@@ -445,24 +442,42 @@ Domain<3> BinaryCompactObject<UseWorldtube>::create_domain() const {
...
-        domain::make_vector_coordinate_map_base<Frame::BlockLogical,
-                                                Frame::Inertial, 3>(
-            sph_wedge_coordinate_maps(
-                object_a.outer_radius, sqrt(3.0) * 0.5 * length_inner_cube_,
-                1.0, 0.0, length_inner_cube_ / 2.0,
-                {{offset_x_coord_a_, 0.0, 0.0}}, use_equiangular_map_),
-            translation_A);
+    Maps maps_center_A{};
+    Maps maps_cube_A{};
+    if (offset_x_coord_a_ == 0.0) {
```
knelli2: Why the extra `if/else` here? I think you can keep the previous way, but use

```cpp
offset_x_coord_a_ == 0.0 ? std::nullopt :
  std::make_optional(std::make_pair(...))
```

for the offset options argument. That way we don't duplicate the majority of these function calls.

knelli2 (changes requested):
Looks good. You can squash these last couple things in all together

src/Domain/Creators/BinaryCompactObject.cpp
AlexCarpenter46: oh I didn't think of that, I'll switch it to that

AlexCarpenter46:
Squashed @knelli2, thanks for the reviews :)

knelli2 (dismissed):

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
```diff
@@ -59,6 +59,7 @@ DomainCreator:
         ExciseWithBoundaryCondition: Worldtube
       UseLogarithmicMap: false
     CenterOfMassOffset: [0., 0.]
+    CubeLength: Auto
```
nilsvu: What do you think about making this `CubeScale: 1` instead so we can control the cube length relative to the separation, @AlexCarpenter46 @knelli2 ?

knelli2: I'd like to do whatever is more intuitive for a user who isn't super familiar with our BCO domains. I think my vote for which is clearer would be `CubeLength` because it's exactly what it says it is vs `CubeScale` where you have to know *a priori* that the scale is with respect to the separation. (of course you could write this in the docs, but I was going based off the option name)

AlexCarpenter46: I had a version with that, but after discussing with @knelli2 we agreed this might be more intuitive and less easily confused with other things in the BCO that use the word Scale

AlexCarpenter46: Oh I didn't refresh github before sending my comment

wthrowe: Just a drive-by opinion, so don't let me block useful progress, but my vote would be for a "scale"-type option, because then there would be no need for an "Auto" value.  Having fewer default-like things is better.

knelli2: @nilsvu and I decided to go with `CubeScale`. @AlexCarpenter46 could you please switch this?

AlexCarpenter46: Will do!

knelli2 (approved):
Looks good! Squash

PR #6244, opened 2024-08-23
Add ScalarGaussBonnet system and functions.

PunJustice:
## Proposed changes

Implements the system (including equations and boundary conditions) for solving the stationary scalar equation in scalar Gauss-Bonnet gravity. Also implements some functions for completing the initial data set for evolution.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Elliptic/Systems/ScalarGaussBonnet/Equations.cpp
```diff
@@ -0,0 +1,147 @@
...
+                   const tnsr::II<DataVector, 3>& inv_conformal_metric,
+                   const tnsr::I<DataVector, 3>& shift,
+                   const Scalar<DataVector>& lapse,
+                   const Scalar<DataVector>& conformal_factor,
+                   const tnsr::i<DataVector, 3>& field_gradient) {
+  raise_or_lower_index(flux_for_field, field_gradient, inv_conformal_metric);
+
+  for (size_t i = 0; i < 3; i++) {
+    flux_for_field->get(i) /= pow<4>(get(conformal_factor));
+    flux_for_field->get(i) -= get(dot_product(shift, field_gradient)) *
```
nilsvu: Move the dot product divided by lapse out of the loop, so it isn't repeated three times. Or add a note that it's a possible performance optimization.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.cpp
```diff
@@ -0,0 +1,147 @@
...
+                 const tnsr::I<DataVector, 3>& shift,
+                 const Scalar<DataVector>& lapse,
+                 const Scalar<DataVector>& conformal_factor,
+                 const tnsr::i<DataVector, 3>& face_normal,
+                 const Scalar<DataVector>& field) {
+  raise_or_lower_index(flux_for_field, face_normal, inv_conformal_metric);
+
+  for (size_t i = 0; i < 3; i++) {
+    flux_for_field->get(i) /= pow<4>(get(conformal_factor));
+    flux_for_field->get(i) -= get(dot_product(shift, face_normal)) *
```
nilsvu: Same here. Also add a note that this whole term could actually be precomputed once at the start of the simulation as a possible performance optimization.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.cpp
```diff
@@ -0,0 +1,147 @@
...
+}
+
+void add_GB_terms(gsl::not_null<Scalar<DataVector>*> scalar_tensor_equation,
+                  const double eps2, const double eps4,
+                  const Scalar<DataVector>& weyl_electric,
+                  const Scalar<DataVector>& weyl_magnetic,
+                  const Scalar<DataVector>& field) {
+  get(*scalar_tensor_equation) -=
+      2. * (weyl_electric.get() - weyl_magnetic.get()) *
+      (eps2 * field.get() + eps4 * field.get() * field.get() * field.get());
```
nilsvu: cube(get(field))

src/Elliptic/Systems/ScalarGaussBonnet/Equations.cpp
```diff
@@ -0,0 +1,147 @@
...
+void add_linearized_GB_terms(
+    gsl::not_null<Scalar<DataVector>*> linearized_scalar_tensor_equation,
+    const double eps2, const double eps4,
+    const Scalar<DataVector>& weyl_electric,
+    const Scalar<DataVector>& weyl_magnetic, const Scalar<DataVector>& field,
+    const Scalar<DataVector>& field_correction) {
+  get(*linearized_scalar_tensor_equation) -=
+      2. * (weyl_electric.get() - weyl_magnetic.get()) *
+      (eps2 * field_correction.get() +
+       3. * eps4 * field.get() * field.get() * field_correction.get());
```
nilsvu: square(get(field))

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
```diff
@@ -0,0 +1,172 @@
...
+namespace sgb {
+struct Fluxes;
+struct Sources;
+}  // namespace sgb
+/// \endcond
+
+namespace sgb {
+
+/*!
+ * \brief Compute the fluxes \f$F^i=\gamma^{ij}\partial_j u(x)\f$ for the scalar
```
nilsvu: FYI, you don't need the `\f`, standard `$` latex syntax works as well (I'm not requesting you change anything).

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
```diff
@@ -0,0 +1,172 @@
...
+                 const tnsr::II<DataVector, 3>& inv_conformal_metric,
+                 const tnsr::I<DataVector, 3>& shift,
+                 const Scalar<DataVector>& lapse,
+                 const Scalar<DataVector>& conformal_factor,
+                 const tnsr::i<DataVector, 3>& face_normal,
+                 const Scalar<DataVector>& field);
+
+/*!
+ * \brief Add the source terms \f$S=-\Gamma^i_{ij}v^j\f$ for the scalar equation
+ * in sGB gravity arising from the covariant derivative with respect ot the
```
nilsvu: typo "to"

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
```diff
@@ -0,0 +1,172 @@
...
+void face_fluxes(gsl::not_null<tnsr::I<DataVector, 3>*> flux_for_field,
+                 const tnsr::II<DataVector, 3>& inv_conformal_metric,
+                 const tnsr::I<DataVector, 3>& shift,
+                 const Scalar<DataVector>& lapse,
+                 const Scalar<DataVector>& conformal_factor,
+                 const tnsr::i<DataVector, 3>& face_normal,
+                 const Scalar<DataVector>& field);
+
+/*!
+ * \brief Add the source terms \f$S=-\Gamma^i_{ij}v^j\f$ for the scalar equation
```
nilsvu: This also adds the other linear terms. Please add the full latex expression.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
```diff
@@ -0,0 +1,172 @@
...
+void add_curved_sources(gsl::not_null<Scalar<DataVector>*> source_for_field,
+                        const tnsr::i<DataVector, 3>& christoffel_contracted,
+                        const tnsr::I<DataVector, 3>& flux_for_field,
+                        const tnsr::i<DataVector, 3>& deriv_lapse,
+                        const Scalar<DataVector>& lapse,
+                        const Scalar<DataVector>& conformal_factor,
+                        const tnsr::i<DataVector, 3>& conformal_factor_deriv);
+
+/*!
+ * \brief Add the sGB coupling term.
```
nilsvu: Also please add the full latex expression here

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
```diff
@@ -0,0 +1,172 @@
...
+    const Scalar<DataVector>& field_correction);
+
+/*!
+ * \brief Compute the fluxes \f$F^i\f$ for the scalar equation in sGB gravity on
+ * a spatial metric \f$\gamma_{ij}\f$.
+ */
+struct Fluxes {
+  using argument_tags = tmpl::list<
+      Xcts::Tags::InverseConformalMetric<DataVector, 3, Frame::Inertial>,
+      sgb::Tags::RolledOffShift, gr::Tags::Lapse<DataVector>,
```
nilsvu, line 93: Can you just use `gr::Tags::Shift` and decide at initialization if you want to apply a rolloff or other things?

src/Elliptic/Systems/ScalarGaussBonnet/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,51 @@
...
+
+/*!
+ * \brief The scalar equation in sGB theories of gravity in the decoupled limit,
+ * assuming quasi-stationarity. See \ref sgb for details on the
+ * explicit equation.
+ */
+struct FirstOrderSystem
+    : tt::ConformsTo<elliptic::protocols::FirstOrderSystem> {
+ private:
+  using field = ::CurvedScalarWave::Tags::Psi;
```
nilsvu: Is this tag the only dependence of the whole system on `CurvedScalarWave`? Then you could just define your own `sgb::Tags::Field` and don't worry about using `CurvedScalarWave` things. Then the `CurvedScalarWave` system isn't a dependency anymore and doesn't need to be compiled.

src/Elliptic/Systems/ScalarGaussBonnet/ScalarGaussBonnet.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * the action with respect ot $\Psi$, one obtains the wave-like equation
+ *
+ * \begin{equation}
+ * \Box \Psi = - \ell^2 f'(\Psi) \mathcal{G}
+ * \end{equation}
+ *
+ * In the spirit of quasi-stationarity, we set $\partial_t \Psi = \partial_t^2
+ * \Psi = \partial_t \alpha = \partial_t \beta^{i} = 0$, where $\alpha$ and
+ * $\beta^i$ are the lapse and shift respectively. This yields ($\ref{eq:
+ * sGB}$), with $\gamma^{ij}$ being the spatial metric, and $\Gamma_i$ it's
```
nilsvu: typo "its"

src/Elliptic/Systems/ScalarGaussBonnet/Tags.hpp
```diff
@@ -0,0 +1,137 @@
...
+ */
+struct PiWithRolledOffShift : db::SimpleTag {
+  using type = Scalar<DataVector>;
+};
+
+/*!
+ * \brief Action that moves the derivative of the scalar field to a different
+ * tag, for importing into the evolution code.
+ */
+struct MoveDerivToPhi : ::CurvedScalarWave::Tags::Phi<3>, db::ComputeTag {
```
nilsvu: Is this only to "rename" the output dataset? No need to do that, this can and should be done on the importing side. Don't even need to output deriv(psi) necessarily because the derivative can be taken on the evolution grid after importing as well (that's what we do for GH).

src/PointwiseFunctions/ScalarGaussBonnet/ScalarMomentum.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/Tensor/EagerMath/DotProduct.hpp"
+#include "Elliptic/Systems/ScalarGaussBonnet/Tags.hpp"
+#include "Elliptic/Systems/Xcts/Tags.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+
+namespace sgb {
+
+void compute_pi(const gsl::not_null<Scalar<DataVector>*> result,
```
nilsvu: Remove `compute_` prefix. Name `scalar_momentum` if you think `pi` is too short.

nilsvu (changes requested):

src/Elliptic/Systems/ScalarGaussBonnet/Equations.cpp
PunJustice: I have also moved computing (conformal_factor)^4 outside of the loop.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
PunJustice: Even if we do make this type of decision at initialization, I still think we want to keep whatever shift we use for the fluxes in a separate tag. We will want to be able to output all of the gh variables from the xcts solve with the scalar field in one set of volume files, and so I would want to keep a copy of `gr::Tags::Shift` in the databox anyway.

An alternative would be to do the rolling off explictly in `curved_fluxes` and `face_fluxes` instead, and just get rid of `sgb::Tags::RolledOffShift`. This is a bit of a pain since it means I would need `sgb::Tags::RolloffLocation` and `sgb::Tags::RolloffRate` (and these tags wrapped in subdomain tags, faces tags, and both subdomain and faces tags) available in the databox. Happy to do this instead though.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
PunJustice: I've adjusted the comment to be a little more descriptive as well.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
PunJustice: I've removed them anyway for cleanliness.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.cpp
nilsvu: Ok it's not obvious that moving psi^4 out of the loop is faster because it does an extra memory allocation now. The dot product does that anyway, so moving it out of the loop is clearly an optimization. I'd say don't worry about it until you do a profiling + performance optimization pass over the solver.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
```diff
@@ -0,0 +1,172 @@
...
+namespace sgb {
+struct Fluxes;
+struct Sources;
+}  // namespace sgb
+/// \endcond
+
+namespace sgb {
+
+/*!
+ * \brief Compute the fluxes $F^i=\gamma^{ij}\partial_j u(x)$ for the scalar
```
nilsvu: Oh this is missing the shift term

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
```diff
@@ -0,0 +1,172 @@
...
+ */
+void curved_fluxes(gsl::not_null<tnsr::I<DataVector, 3>*> flux_for_field,
+                   const tnsr::II<DataVector, 3>& inv_conformal_metric,
+                   const tnsr::I<DataVector, 3>& shift,
+                   const Scalar<DataVector>& lapse,
+                   const Scalar<DataVector>& conformal_factor,
+                   const tnsr::i<DataVector, 3>& field_gradient);
+
+/*!
+ * \brief Compute the fluxes $F^i=\gamma^{ij}\partial_j u(x)$ for the scalar
```
nilsvu: this too

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
```diff
@@ -0,0 +1,172 @@
...
+                 const tnsr::I<DataVector, 3>& shift,
+                 const Scalar<DataVector>& lapse,
+                 const Scalar<DataVector>& conformal_factor,
+                 const tnsr::i<DataVector, 3>& face_normal,
+                 const Scalar<DataVector>& field);
+
+/*!
+ * \brief Adds the source terms arising from the $\box \Psi$ term in the
+ * equation of motion for the scalar field: $S=-\Gamma^i_{ij}v^j+v^j
+ * \partial_j \ln (\alpha)$.*/
```
nilsvu: Please put this into the form you actually implemented. There should be a grad(conformal_factor) for example. Also define `v^j` or just write `F^j`. Generally it's very helpful to have the docs for a function include the expression that's implemented and ideally some reference to where it comes from, ideally an equation in a paper.

src/Elliptic/Systems/ScalarGaussBonnet/Equations.hpp
nilsvu: Ok let's keep `sgb::Tags::RolledOffShift` for now then

src/Elliptic/Systems/ScalarGaussBonnet/FirstOrderSystem.hpp
```diff
@@ -0,0 +1,51 @@
...
+  using field = ::sgb::Tags::Psi;
+
+ public:
+  static constexpr size_t volume_dim = 3;
+
+  using primal_fields = tmpl::list<field>;
+  using primal_fluxes =
+      tmpl::list<::Tags::Flux<field, tmpl::size_t<3>, Frame::Inertial>>;
+
+  using background_fields = tmpl::list<
```
nilsvu, line 37: Please add a comment that numerically imported background fields are not listed here

src/Elliptic/Systems/ScalarGaussBonnet/Tags.hpp
```diff
@@ -0,0 +1,125 @@
...
+
+/// \cond
+class DataVector;
+/// \endcond
+
+namespace sgb {
+/// Tags related to solving the scalar equation in sGB gravity.
+namespace OptionTags {
+
+struct Epsilon2 {
```
nilsvu, line 35: Put these option tags into a group so they appear together in the input file. To do this, just add a `struct OptionGroup` with `static std::string name() { return "ScalarGaussBonnet"; }` and a help string. Then add `using group = OptionGroup` to the tags.

src/PointwiseFunctions/ScalarGaussBonnet/ScalarMomentum.cpp
```diff
@@ -0,0 +1,30 @@
...
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+
+namespace sgb {
+
+void scalar_momentum(const gsl::not_null<Scalar<DataVector>*> result,
+                     const tnsr::i<DataVector, 3, Frame::Inertial>& deriv,
+                     const tnsr::I<DataVector, 3>& shift,
+                     const Scalar<DataVector>& lapse) {
+  get(*result) = get(dot_product(shift, deriv)) / get(lapse);
```
nilsvu:
```cpp
dot_product(result, shift, deriv);
get(*result) /= get(lapse);
```

(easy to do and avoids an allocation)

src/PointwiseFunctions/ScalarGaussBonnet/ScalarMomentum.hpp
```diff
@@ -0,0 +1,77 @@
...
+#include "DataStructures/DataBox/Tag.hpp"
+#include "Elliptic/Systems/ScalarGaussBonnet/Tags.hpp"
+#include "Elliptic/Systems/Xcts/Tags.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
+
+namespace sgb {
+/// @{
+/*!
+ * \brief Compute the momentum$\Pi$ assuming quasi-stationarity in sGB gravity.
```
nilsvu: missing space after "momentum"

src/PointwiseFunctions/ScalarGaussBonnet/ScalarMomentum.hpp
```diff
@@ -0,0 +1,77 @@
...
+    const tnsr::i<DataVector, 3, Frame::Inertial>& deriv,
+    const tnsr::I<DataVector, 3>& shift, const Scalar<DataVector>& lapse);
+/// @{
+
+namespace Tags {
+
+/*!
+ * \brief Re-compute the momentum Pi, assuming quasi-stationarity.
+ */
+struct PiCompute : ::CurvedScalarWave::Tags::Pi, db::ComputeTag {
```
nilsvu: (optional) If you want you can avoid the two nearly identical structs by templating on `ShiftTag` and `OutputTag`:

```cpp
template <typename ShiftTag, typename OutputTag>
struct PiCompute : OutputTag, db::ComputeTag {
  // ...
  using argument_tags = tmpl::list<ShiftTag, ...>;
```

nilsvu (changes requested):
Please push changes as fixup commits, it helps a lot with reviewing

nilsvu (approved):
LGTM, please rebase and squash

PR #6245, opened 2024-08-24
Add logical_divergence function, elliptic StrongLogical DG formulation

nilsvu:
## Proposed changes

This variant of the elliptic DG formulation moves the Jacobian into the divergence like this:

$$
\partial_i F^i = \frac{1}{J} \partial_\hat{\imath} J {J^{-1}}^{\hat{\imath}}_j F^j
$$

This formulation makes the strong scheme work as well as the weak scheme for extremely stretched grids (initial data with large outer boundaries).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/DiscontinuousGalerkin/DgOperator.hpp
+++ b/src/Elliptic/DiscontinuousGalerkin/DgOperator.hpp
@@ -626,6 +626,17 @@ struct DgOperatorImpl<System, Linearized, tmpl::list<PrimalFields...>,
         // This is the sign flip that makes the operator _minus_ the Laplacian
         // for a Poisson system
         *operator_applied_to_vars *= -1.;
+      } else if (formulation == ::dg::Formulation::StrongLogical) {
+        // Strong divergence but with the Jacobian moved into the divergence:
+        //   div(F) = 1/J_p \sum_q (D_\hat{i})_pq J_q (J^\hat{i}_i)_q (F^i)_q.
+        const auto logical_fluxes = transform::first_index_to_different_frame(
+            primal_fluxes, det_times_inv_jacobian);
+        logical_divergence(operator_applied_to_vars, logical_fluxes, mesh);
+        if (massive) {
+          *operator_applied_to_vars *= -1.;
+        } else {
+          *operator_applied_to_vars *= -1. * get(det_inv_jacobian);
+        }
       } else if (formulation == ::dg::Formulation::WeakInertial) {
         // Compute weak divergence:
         //   F^i \partial_i \phi = 1/w_p \sum_q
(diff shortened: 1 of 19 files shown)


tests/Unit/NumericalAlgorithms/LinearOperators/Test_Divergence.cpp
```diff
@@ -166,6 +169,12 @@ void test_divergence_impl(
   const auto& expected =
       get<Tags::div<Flux1<DataType, Dim, Frame>>>(div_fluxes);
   CHECK_ITERABLE_CUSTOM_APPROX(expected, div_vector, local_approx);
+
+  // Test logical divergence
+  auto logical_fluxes =
+      transform::first_index_to_different_frame(fluxes, inv_jacobian);
+  logical_divergence(make_not_null(&div_fluxes), logical_fluxes, mesh);
```
nilsdeppe: I would create a `logical_div_fluxes` variable since the current code I think would pass if `logical_div_fluxes` was a no-op

src/Elliptic/DiscontinuousGalerkin/DgOperator.hpp
```diff
@@ -626,6 +626,16 @@ struct DgOperatorImpl<System, Linearized, tmpl::list<PrimalFields...>,
         // This is the sign flip that makes the operator _minus_ the Laplacian
         // for a Poisson system
         *operator_applied_to_vars *= -1.;
+      } else if (formulation == ::dg::Formulation::StrongLogical) {
+        // Strong divergence but with the Jacobian moved into the divergence:
+        //   div(F) = 1/J_p \sum_q (D_\hat{i})_pq J_q (J^\hat{i}_i)_q (F^i)_q.
+        const auto logical_fluxes = transform::first_index_to_different_frame(
+            primal_fluxes, det_times_inv_jacobian);
+        logical_divergence(operator_applied_to_vars, logical_fluxes, mesh);
+        *operator_applied_to_vars *= -1.;
```
nilsdeppe: I'd combine the negation and the multiply when possible so we only run through the memory once. Even better would be having a bool template parameter on `logical_divergence` to `-=` instead of `+=` but that's a bigger change so optional.

nilsdeppe (changes requested):
LGTM, a couple suggestions you can squash immediately.

src/Elliptic/DiscontinuousGalerkin/DgOperator.hpp
nilsvu: Combined the negation with the Jacobian multiply. The `-=` can be done too but I didn't decide immediately how best to do it.

nilsdeppe (approved):

PR #6247, opened 2024-08-24
Add single BH control system

knelli2:
## Proposed changes

Controls object to the origin of the domain.

### Upgrade instructions

Must now specify a `Translation:` control system in `EvolveGhSingleBlackHole` input files.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/ControlSystem/ControlErrors/Translation.cpp
```diff
@@ -0,0 +1,11 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ControlSystem/ControlErrors/Translation.hpp"
+
+#include <cstddef>
+
+namespace control_system::ControlErrors {
+template struct Translation<1_st>;
+template struct Translation<2_st>;
```
wthrowe: These don't do anything useful because the class is entirely defined in the hpp.

support/Pipelines/Bbh/Ringdown.yaml
```diff
@@ -246,6 +246,22 @@ ControlSystems:
   WriteDataToDisk: false
   MeasurementsPerUpdate: 4
   Verbosity: Silent
+  Translation:
+    IsActive: false
```
wthrowe: Did you want to turn this on?

wthrowe (changes requested):

knelli2:
Squashed everything in

wthrowe (dismissed):
clang-tidy timed out for no obvious reason.  Rerunning to see if it was some sort of glitch.  If it still times out I'm not going to worry about it.

knelli2:
@wthrowe I've seen clang-tidy time out on several PRs lately for what seems like no good reason (not that many files changed). I haven't tried to figure out why though

wthrowe:
Sounds like you fixed it in #6282.  Want to rebase to pick that up?

knelli2:
I don't think #6282 was the issue as I'm pretty sure the clang-tidy timeouts were happening before I introduced that, but I rebased just in case.

wthrowe:
Actually, looks like your fix broke it more.  CI is spewing errors.

wthrowe:
And then reporting itself as passed.

wthrowe (approved):
Looks like that fixed clang-tidy.  One of the errors there is legitimate, but it wasn't introduced in this PR and we should get the fix in quickly, so I'm not going to worry about it.

PR #6253, opened 2024-08-27
Fix communication race bug in interpolation framework

knelli2:
## Proposed changes

We were not checking the order we received temporal IDs before sending points to the interpolator and this made some horizon finds happen out of order. I don't think this had a significant impact, but still best to fix it. In order to make some of the logic easier, I did a couple things

- Explicitly only allow sequential targets in the interpolator. This was already implicitly done because horizon finds are sequential and no other interpolation used the Interpolator. But best to make it explicit. (commit 2)
- There were multiple times I found where an action running on the target singleton would schedule another action on itself when it could just call it directly. So I did that. (commit 3)

I tested this branch on a BBH test case that I know to fail, and it has gotten significantly past the point of failure

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Actions/AddTemporalIdsToInterpolationTarget.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Actions/AddTemporalIdsToInterpolationTarget.hpp
@@ -4,8 +4,10 @@
 #pragma once
 
 #include <algorithm>
+#include <type_traits>
 
 #include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/LinkedMessageId.hpp"
 #include "DataStructures/Variables.hpp"
 #include "Domain/Tags.hpp"
 #include "Parallel/GlobalCache.hpp"
@@ -16,36 +18,33 @@
 #include "ParallelAlgorithms/Interpolation/Tags.hpp"
 #include "Utilities/TaggedTuple.hpp"
 
-namespace intrp {
-
-namespace Actions {
+namespace intrp::Actions {
 /// \ingroup ActionsGroup
-/// \brief Adds `temporal_id`s on which this InterpolationTarget
+/// \brief Adds a `temporal_id` on which this InterpolationTarget
 /// should be triggered.
 ///
 /// Invoked on an InterpolationTarget.
 ///
 /// In more detail, does the following:
-/// - Adds the passed-in `temporal_id`s to PendingTemporalIds.
-/// - If the InterpolationTarget is sequential
-///   - If TemporalIds and PendingTemporalIds were both initially empty,
-///     invokes Actions::VerifyTemporalIdsAndSendPoints. (Otherwise there
-///     is an interpolation in progress and nothing needs to be done.)
-/// - If the InterpolationTarget is not sequential
-///   - Invokes Actions::SendPointsToInterpolator on all TemporalIds
-///   - Invokes Actions::VerifyTemporalIdsAndSendPoints if PendingTemporalIds
-///     is non-empty.
+/// - Adds the passed-in `temporal_id` to PendingTemporalIds.
+/// - If all the following conditions are met:
+///    - CurrentTemporalId doesn't have a value
+///    - PendingTemporalIds is not empty,
+///    - The temporal id is a LinkedMessageId or PendingTemporalIds was empty
+///      upon entry
+///   call Actions::VerifyTemporalIdsAndSendPoints. (Otherwise there
+///   is an interpolation in progress and nothing needs to be done.)
 ///
 /// Uses:
 /// - DataBox:
-///   - `Tags::TemporalIds<TemporalId>`
+///   - `Tags::CurrentTemporalId<TemporalId>`
 ///   - `Tags::CompletedTemporalIds<TemporalId>`
 ///
 /// DataBox changes:
 /// - Adds: nothing
 /// - Removes: nothing
 /// - Modifies:
-///   - `Tags::TemporalIds<TemporalId>`
+///   - `Tags::PendingTemporalIds<TemporalId>`
 template <typename InterpolationTargetTag>
 struct AddTemporalIdsToInterpolationTarget {
   template <typename ParallelComponent, typename DbTags, typename Metavariables,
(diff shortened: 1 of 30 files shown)


src/ParallelAlgorithms/Interpolation/Actions/AddTemporalIdsToInterpolationTarget.hpp
```diff
@@ -54,56 +54,40 @@ struct AddTemporalIdsToInterpolationTarget {
                     Parallel::GlobalCache<Metavariables>& cache,
```
wthrowe, line 53: Update this class's documentation.

src/ParallelAlgorithms/Interpolation/Actions/VerifyTemporalIdsAndSendPoints.hpp
```diff
@@ -40,23 +47,6 @@ void verify_temporal_ids_and_send_points_time_independent(
...
-        // there's nothing we could do and the later interpolation would happen
-        // before the earlier one (even for sequential targets). However, there
-        // is a scenario where this happens, except there is an interpolation in
-        // progress so it doesn't send points to the interpolator when it
-        // receives the later time first. In that case, we then receive the
-        // earlier time second, and sort them here so they are in temporal
-        // order. Note that this isn't a permanent actual fix, but so far works
-        // in practice.
-        alg::sort(*ids);
-        alg::sort(*pending_ids);
```
wthrowe, line 59: Can you confirm that the changes in this commit fix this issue?  The comment here says it's a race related to receiving messages from the elements, which doesn't sound related to the changes here.  Same thing again later.

Actually, this is probably just in the wrong commit.

tests/Unit/ParallelAlgorithms/Interpolation/Test_AddTemporalIdsToInterpolationTarget.cpp
```diff
@@ -150,12 +183,19 @@ void test_add_temporal_ids() {
...
   for (int i = 0; i < 2; ++i) {
     ActionTesting::next_action<target_component>(make_not_null(&runner), 0);
   }
+  ActionTesting::emplace_array_component<interpolator_component>(
+      make_not_null(&runner), ActionTesting::NodeId{0},
+      ActionTesting::LocalCoreId{0}, 0);
   ActionTesting::set_phase(make_not_null(&runner), Parallel::Phase::Testing);
+  auto& target_box =
+      ActionTesting::get_databox<target_component>(make_not_null(&runner), 0);
+  (void)target_box;
```
wthrowe: Why this line?  It appears to be used unconditionally later.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -411,18 +411,94 @@ void flag_temporal_id_as_pending(const gsl::not_null<db::DataBox<DbTags>*> box,
...
+          if constexpr (std::is_same_v<TemporalId, LinkedMessageId<double>>) {
+            if (alg::found(*pending_pending_ids, temporal_id)) {
+              // Do nothing
+              return;
+            } else {
+              const auto print_pending = [&](const std::string& pending_str) {
+                Parallel::printf(
+                    "%s: Adding %s to %s\n"
+                    " Pending: %s\n"
+                    " Pending pending: %s\n",
```
wthrowe: This seems like it should be controlled by some verbosity setting.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -411,18 +411,94 @@ void flag_temporal_id_as_pending(const gsl::not_null<db::DataBox<DbTags>*> box,
...
+                  }
+                } else {
+                  check_and_place_id(ids);
+                }
+              } else {
+                check_and_place_id(*pending_ids);
+              }
+            }
+
+            alg::sort(*pending_ids);
```
wthrowe: This one should always be in order, no?

src/ParallelAlgorithms/Interpolation/Actions/InterpolationTargetReceiveVars.hpp
```diff
@@ -182,6 +185,38 @@ struct InterpolationTargetReceiveVars {
...
+                        pending_pending_ids) {
+                  const auto& most_recent_completed_id = completed_ids.back();
+                  const auto& next_pending_pending_id =
+                      pending_pending_ids->front();
+
+                  if (next_pending_pending_id.previous.value() ==
+                      most_recent_completed_id.id) {
+                    Parallel::printf(
+                        "%s: Moving %s from pending pending to pending and "
+                        "sending points\n",
```
wthrowe: Again, this should probably have some verbosity control.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -411,18 +411,94 @@ void flag_temporal_id_as_pending(const gsl::not_null<db::DataBox<DbTags>*> box,
...
+                    if (temporal_id.previous.value() ==
+                        ids_to_check.back().id) {
+                      pending_ids->push_back(temporal_id);
+                      print_pending("pending (pending not empty)");
+                    } else {
+                      pending_pending_ids->push_back(temporal_id);
+                      print_pending("pending pending (pending not empty)");
+                    }
+                  };
+              // Cases: (1p = pending, 2p = pending pending)
```
wthrowe: [Arbitrary line] This seems overcomplicated.  Instead of multiple levels of pending, could you just make the pending list a priority queue and check if the first element's previous link is pointing to the last completed id before running it?

wthrowe (changes requested):

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
knelli2: Hmm possibly. I think in order to make this change, the bulk of this logic would move into `VerifyTemporalIdsAndSendPoints` because there it loops over pending ids to see if they are ready, and then this function would basically remain unchanged (except now the sort of pending ids would always be needed to ensure the earliest id is first).

I also think with all the changes in this PR that there can only be a single id in `Tags::TemporalIds` at a time (since only one horizon find can be happening at once). That makes more things easier I think.

wthrowe: If you're not familiar with the [priority queue](https://en.cppreference.com/w/cpp/container/priority_queue) I suggest you read up on that data structure.

src/ParallelAlgorithms/Interpolation/Actions/VerifyTemporalIdsAndSendPoints.hpp
```diff
@@ -266,41 +84,142 @@ struct VerifyTemporalIdsAndSendPoints {
...
+        auto& this_proxy =
+            Parallel::get_parallel_component<ParallelComponent>(cache);
+        double min_expiration_time = std::numeric_limits<double>::max();
+        const Parallel::ArrayComponentId array_component_id =
+            Parallel::make_array_component_id<ParallelComponent>(array_index);
+        const bool next_pending_id_is_ready = [&cache, &array_component_id,
+                                               &this_proxy, &next_pending_id,
+                                               &min_expiration_time]() {
+          if constexpr (Parallel::is_in_mutable_global_cache<
+                            Metavariables, domain::Tags::FunctionsOfTime>) {
```
wthrowe: Under what circumstances is FunctionsOfTime in the cache but not the mutable cache?

tests/Unit/ParallelAlgorithms/Interpolation/Test_AddTemporalIdsToInterpolationTarget.cpp
```diff
@@ -312,6 +348,216 @@ void test_add_temporal_ids() {
...
+          completed_ids->clear();
+          pending_ids->clear();
+          ids->clear();
+        },
+        make_not_null(&target_box));
+  };
+
+  const auto insert_id_into = [&]<template <typename> typename Tag>(
+                                  const LinkedMessageId<double>& id,
+                                  const bool insert_at_back = true) {
```
wthrowe: Last argument isn't used.

wthrowe (changes requested):
Yes, this seems cleaner.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
knelli2: I looked over the priority queue and found that it won't be suitable for our case. Within in the source code (and tests) we need to search through the queue to see if an id is already there and the priority queue does not offer iterators or really any way to query what is in the queue besides the `top` value and size. Therefore, I'm just going to stick with a `deque` and sort it.

src/ParallelAlgorithms/Interpolation/Actions/VerifyTemporalIdsAndSendPoints.hpp
knelli2: Oops, this was just copied from one of the other functions I removed. This `if constexpr` can be removed I believe

knelli2:
@wthrowe I implemented your suggestions (except for the priority queue), updated dox, and changed the current temporal id tag type to optional if the target is sequential (since only one interpolation can happen at a time for sequential targets). I think this should be good to go. I tested the first fixup and that worked on a BBH case that was failing before and I'm testing the second fixup now.

src/ParallelAlgorithms/Interpolation/Actions/AddTemporalIdsToInterpolationTarget.hpp
```diff
@@ -18,36 +18,33 @@
...
-/// - If the InterpolationTarget is not sequential
-///   - Invokes Actions::SendPointsToInterpolator on all TemporalIds
-///   - Invokes Actions::VerifyTemporalIdsAndSendPoints if PendingTemporalIds
-///     is non-empty.
+/// - Adds the passed-in `temporal_id` to PendingTemporalIds.
+/// - If:
+///    - CurrentTemporalId doesn't have a value
+///    - PendingTemporalIds is not empty,
+///    - The temporal id is a LinkedMessageId or PendingTemporalIds was empty
+///      upon entry
```
wthrowe, line 34: Are these conditions ORed or ANDed?

wthrowe (changes requested):
OK.  Go ahead and squash once you're satisfied with your testing.

knelli2:
Second fixup also passed my test. Squashed everything in

wthrowe (approved):

knelli2:
@nilsdeppe Do you want to look at this? I believe it will fix the issue you ran into testing nodegroups

nilsdeppe:
I'll take a quick glance today, but won't do a detailed review since Will did that :)

nilsdeppe (approved):

PR #6254, opened 2024-08-27
Error if the time step becomes too small

wthrowe:
### Upgrade instructions

Input files for evolution executables now require a new `MinimumTimeStep` option in the `Evolution` section.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DiscontinuousGalerkin/Actions/ComputeTimeDerivative.hpp
+++ b/src/Evolution/DiscontinuousGalerkin/Actions/ComputeTimeDerivative.hpp
@@ -56,6 +56,7 @@
 #include "Time/Actions/SelfStartActions.hpp"
 #include "Time/BoundaryHistory.hpp"
 #include "Time/Tags/HistoryEvolvedVariables.hpp"
+#include "Time/Tags/MinimumTimeStep.hpp"
 #include "Time/TakeStep.hpp"
 #include "Utilities/Algorithm.hpp"
 #include "Utilities/Gsl.hpp"
@@ -353,7 +354,9 @@ struct ComputeTimeDerivative {
   using const_global_cache_tags = tmpl::append<
       tmpl::list<::dg::Tags::Formulation,
                  evolution::Tags::BoundaryCorrection<EvolutionSystem>,
-                 domain::Tags::ExternalBoundaryConditions<Dim>>>;
+                 domain::Tags::ExternalBoundaryConditions<Dim>>,
+      tmpl::conditional_t<LocalTimeStepping,
+                          tmpl::list<::Tags::MinimumTimeStep>, tmpl::list<>>>;
 
   template <typename DbTagsList, typename... InboxTags, typename ArrayIndex,
             typename ActionList, typename ParallelComponent,
--- a/support/Pipelines/Bbh/Inspiral.yaml
+++ b/support/Pipelines/Bbh/Inspiral.yaml
@@ -139,6 +139,7 @@ DomainCreator:
 Evolution:
   InitialTime: *InitialTime
   InitialTimeStep: 0.0002
+  MinimumTimeStep: 1e-7
   # This is the smallest interval we'd need to observe time step/constraints. If
   # you need it smaller you can edit it, but make sure to change the slab
   # intervals in the EventsAndTriggers
--- a/support/Pipelines/Bbh/Ringdown.yaml
+++ b/support/Pipelines/Bbh/Ringdown.yaml
@@ -46,6 +46,7 @@ DomainCreator:
 Evolution:
   InitialTime: 0.0
   InitialTimeStep: 0.0002
+  MinimumTimeStep: 1e-7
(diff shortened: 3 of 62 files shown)


src/Evolution/DiscontinuousGalerkin/Actions/ComputeTimeDerivative.hpp
```diff
@@ -353,7 +354,9 @@ struct ComputeTimeDerivative {
   using const_global_cache_tags = tmpl::append<
       tmpl::list<::dg::Tags::Formulation,
                  evolution::Tags::BoundaryCorrection<EvolutionSystem>,
-                 domain::Tags::ExternalBoundaryConditions<Dim>>>;
+                 domain::Tags::ExternalBoundaryConditions<Dim>>,
+      tmpl::conditional_t<LocalTimeStepping,
+                          tmpl::list<::Tags::MinimumTimeStep>, tmpl::list<>>>;
```
knelli2, line 359: Why add this is more than one place? Once in a LTS action seems like it should be enough

support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -139,6 +139,7 @@ DomainCreator:
 Evolution:
   InitialTime: *InitialTime
   InitialTimeStep: 0.0002
+  MinimumTimeStep: 1e-6
```
knelli2: Was there a reason to change this from the hardcoded value 1e-9?

src/Evolution/DiscontinuousGalerkin/Actions/ComputeTimeDerivative.hpp
wthrowe: This one is for LTS, the one in `src/Time/ChangeSlabSize/Action.hpp` is for GTS (which will usually also be used for LTS), and the one in `src/Time/Actions/ChangeStepSize.hpp` is for CCE.

support/Pipelines/Bbh/Inspiral.yaml
wthrowe: I don't know what a good value is.  SpEC uses 1e-5, but because of different time stepper choices we expect to be a bit smaller than SpEC in some cases.  The 1e-9 was chosen as "about 2^-31", which is still there except changed to exactly 2^-31, and is a relative limit rather than an absolute one.

What minimum time step do we typically see in BBH runs?

knelli2: Here's the min/max time step for a high resolution BBH from merger to ringdown. It reaches down to 1e-5. This run was with relative tolerance 1e-6 and absolute tolerance 1e-8, but @geoffrey4444 is trying runs with 100x tighter tolerances. I'd be worried that 1e-6 was too large near merger

image

nilsdeppe: I'm not sure if spectre is doing this, but in spec the time step near merger gets small because of the control system, not the error stepping. I guess we don't have that coupling here?  I don't know what to expect with 100x tighter tolerance, but assuming the solution is reasonably well-behaved then I would expect only a ~2-3 smaller time step for a 4th-order method. If we are doing higher order AMPC then it could be even less than that.

knelli2: No we don't have a coupling between the time step and control system. @wthrowe @nilsdeppe What do you think about 1e-7 then as the value we have for the BBH/SingleBH LTS executables? 1e-6 seems ok for other systems (especially ones using GTS)

wthrowe: After seeing the above plot, my inclination is to drop the minimum to 1e-7 (just haven't gotten to it yet).  I also don't expect a very large change from the higher tolerance, but if we're measuring time steps of ~1e-5 on an easy case we'll probably get a bit lower on more extreme systems.  There may also be smaller steps that we don't see in the plot, since we don't output every time step.

nilsdeppe: That sounds good to me!

wthrowe: Changed.

support/Pipelines/Bbh/Ringdown.yaml
```diff
@@ -46,6 +46,7 @@ DomainCreator:
 Evolution:
   InitialTime: 0.0
   InitialTimeStep: 0.0002
+  MinimumTimeStep: 1e-6
```
knelli2: I think for the ringdown we want this to also be `1e-7`

knelli2 (changes requested):
Maybe also give this a rebase since some nvcc stuff was just merged

support/Pipelines/Bbh/Ringdown.yaml
wthrowe: Oops.  Missed the pipeline files.  Fixed.

knelli2 (approved):

PR #6255, opened 2024-08-28
Adds worldtube functions of time to `BinaryCompactObject`

nikwit:
This PR allows the `BinaryCompactObject` domain creator to return the functions of time needed by the worldtube evolution.

This required either adding a Worldtube argument that could be determined in the input file at runtime or templating the class on a bool. @knelli2 and I decided it is best to use the latter approach, as a `UseWorldtube` option in the input file might be confusing to 99 % of users who would not need it. The template parameter is defaulted to `false` and only this instantiation is added to the Factory so it should be very hard for a user to use the wrong one by accident.

### DIFF
--- a/src/Domain/Creators/Factory3D.hpp
+++ b/src/Domain/Creators/Factory3D.hpp
@@ -21,8 +21,8 @@ template <>
 struct domain_creators<3> {
   using type =
       tmpl::list<domain::creators::AlignedLattice<3>,
-                 domain::creators::BinaryCompactObject, domain::creators::Brick,
-                 domain::creators::Cylinder,
+                 domain::creators::BinaryCompactObject<false>,
+                 domain::creators::Brick, domain::creators::Cylinder,
                  domain::creators::CylindricalBinaryCompactObject,
                  domain::creators::FrustalCloak,
                  domain::creators::RotatedBricks, domain::creators::Sphere>;
--- a/src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
+++ b/src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
@@ -19,6 +19,7 @@
 #include "Domain/Creators/TimeDependentOptions/ShapeMap.hpp"
 #include "Domain/FunctionsOfTime/FixedSpeedCubic.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/IntegratedFunctionOfTime.hpp"
 #include "Domain/FunctionsOfTime/PiecewisePolynomial.hpp"
 #include "Domain/FunctionsOfTime/QuaternionFunctionOfTime.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Spherepack.hpp"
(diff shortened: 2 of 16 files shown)


src/Domain/Creators/Factory3D.hpp
```diff
@@ -21,8 +21,8 @@ template <>
 struct domain_creators<3> {
   using type =
       tmpl::list<domain::creators::AlignedLattice<3>,
-                 domain::creators::BinaryCompactObject, domain::creators::Brick,
-                 domain::creators::Cylinder,
+                 domain::creators::BinaryCompactObject<false>,
+                 domain::creators::Brick, domain::creators::Cylinder,
```
knelli2, line 25: In the BBH executable, you also need to have `BCO<false>` since we use a different list there

src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
```diff
@@ -71,11 +72,91 @@ TimeDependentMapOptions<IsCylindrical>::TimeDependentMapOptions(
 }
 
 template <bool IsCylindrical>
+std::unordered_map<std::string,
+                   std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
+TimeDependentMapOptions<IsCylindrical>::create_worldtube_functions_of_time()
+    const {
+  ASSERT(not translation_map_options_.has_value(),
+         "Translation map is not implemented for worldtube evolutions.");
```
knelli2: Make these `ERROR`s. If the user specifies an incorrect option, we should get an error and not be able to proceed

src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
```diff
@@ -208,7 +208,8 @@ struct EvolutionMetavars {
                     standard_boundary_conditions<volume_dim>,
                 CurvedScalarWave::BoundaryConditions::Worldtube<volume_dim>>>>,
         tmpl::pair<DenseTrigger, DenseTriggers::standard_dense_triggers>,
-        tmpl::pair<DomainCreator<volume_dim>, domain_creators<volume_dim>>,
+        tmpl::pair<DomainCreator<volume_dim>,
+                   tmpl::list<domain::creators::BinaryCompactObject<false>>>,
```
knelli2, line 212: Should this be `<true>`? Or are you not enabling the worldtube FoTs yet?

src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
```diff
@@ -71,11 +72,91 @@ TimeDependentMapOptions<IsCylindrical>::TimeDependentMapOptions(
...
+template <bool UseWorldtube>
 std::unordered_map<std::string,
                    std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
 TimeDependentMapOptions<IsCylindrical>::create_functions_of_time(
     const std::unordered_map<std::string, double>& initial_expiration_times)
     const {
+  if constexpr (UseWorldtube) {
+    static_assert(not IsCylindrical,
+                  "Cylindrical map not supported with worldtube");
+    return create_worldtube_functions_of_time();
```
knelli2: Also `(void)initial_expiration_times`. Or you could optionally check that the initial expiration times are empty because there shouldn't be any control systems for the CSW system.

knelli2 (changes requested):
And then also fix the `clang-tidy` errors

src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
nikwit: I am not enabling them yet but will do so in a follow-up PR (hopefully tomorrow). Currently, the worldtube fots are not being updated, so the executable would not be usable.

nikwit:
@knelli2 thanks, I included your suggestions. test failures look unrelated to me and I am not sure why clang-tidy got stuck

knelli2 (approved):
Looks good! You can squash. Yeah I'm not sure why clang-tidy is being difficult. This has happened on other PRs as well.

nikwit:
Squashed in, thank you for your review!

knelli2:
Ignoring unrelated test failures

PR #6260, opened 2024-08-30
Fix interpolation stencil misalignment in `RotatingStar`

ncorsobh:
## Proposed changes

In `RotatingStar`, the setup of the stencils for the interpolation of the `RotNS` data was slightly offset. This resulted in rare edge cases where the target point could fall outside of the stencil if the rounding behaved unfavorably, especially at high resolutions. However, even when the target point fell inside the stencil, it would likely be closer to its upper edge.

This PR both fixes the rounding issue that resulted in the target point falling outside of the stencil and improves on the stencil construction overall. Now, if the stencil size is an even number, the target point will always be placed between the middle two points of the stencil, and if the stencil size is an odd number, the target point will be closest to the central point in the stencil.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
@@ -198,13 +198,12 @@ std::array<double, 6> CstSolution::interpolate(
   }
   using std::abs;
   const double target_abs_cos_theta = abs(target_cos_theta);
-  const auto angular_index = static_cast<size_t>(
-      target_abs_cos_theta * static_cast<double>(num_angular_points_ - 1));
-  // radius_index differs from SpEC by +1 to get the stencils to be centered.
-  size_t radius_index =
-      static_cast<size_t>(num_radial_points_ * target_radius /
-                          (target_radius + equatorial_radius_)) +
-      1;
+  // The `lround` centers the stencils and avoids rounding problems.
+  const auto angular_index = static_cast<size_t>(lround(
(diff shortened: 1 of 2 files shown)


src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
```diff
@@ -232,12 +231,27 @@ std::array<double, 6> CstSolution::interpolate(
...
+  // of the stencil, i.e. between the middle 2 points.
+  int radial_stencil_adjust = 0;
+  int angular_stencil_adjust = 0;
+  if constexpr (stencil_size % 2 == 0) {
+    if (target_radius > radius_[radius_index * num_angular_points_]) {
+      radial_stencil_adjust++;
+    };
+    if (target_abs_cos_theta > cos_theta_[angular_index]) {
+      angular_stencil_adjust++;
+    };
```
wthrowe: [optional] Two extra semicolons.

wthrowe (changes requested):
The changes look fine.  Is there anything here that it is possible to test?

src/PointwiseFunctions/AnalyticSolutions/RelativisticEuler/RotatingStar.cpp
ncorsobh: Squashed the semicolons immediately. Added the unit test in a fixup commit. The test attempts to interpolate at a point which would've caused the original stencil construction to fail (I tested it with the old method and it did indeed fail).

wthrowe:
Looks good.  Squash.

wthrowe (approved):

PR #6261, opened 2024-08-30
Add smoothly broken power law to adjust worldtube radius

nikwit:


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -102,6 +102,72 @@ struct SelfForceOptions {
...
+    using type = double;
+    static constexpr Options::String help{
+        "The width delta of the transition region."};
+    static double lower_bound() { return 1e-3; }
+  };
+
+  struct Exponent {
+    using type = double;
+    static constexpr Options::String help{
+        "The exponent according to which the excision sphere grows with "
```
PunJustice: Can you specify this is \alpha like you do for the TransitionWidth.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -102,6 +102,72 @@ struct SelfForceOptions {
...
+    static constexpr Options::String help{
+        "The amplitude of the smoothly broken power law."};
+    static double lower_bound() { return 0.; }
+  };
+
+  struct TransitionRadius {
+    using type = double;
+    static constexpr Options::String help{
+        "The transition radius rb of the smoothly broken power law. At this "
+        "point the transitions to a constant value."};
```
PunJustice: the radius transitions to a constant value.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -102,6 +102,72 @@ struct SelfForceOptions {
...
+    if constexpr (IsWorldtube) {
+      return "WorldtubeRadiusOptions";
+    } else {
+      return "BlackHoleRadiusOptions";
+    }
+  }
+  struct Amplitude {
+    using type = double;
+    static constexpr Options::String help{
+        "The amplitude of the smoothly broken power law."};
```
PunJustice: The amplitude A

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_RadiusFunctions.cpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "Framework/TestingFramework.hpp"
+
+#include <cmath>
+
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/RadiusFunctions.hpp"
+#include "Framework/TestHelpers.hpp"
+
+namespace CurvedScalarWave::Worldtube {
+namespace {
```
PunJustice, line 12: If you wanted one more test you could reimplement these in python and test against them, but I think these tests are already quite good.

src/Evolution/Systems/CurvedScalarWave/Worldtube/RadiusFunctions.hpp
```diff
@@ -0,0 +1,40 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+namespace CurvedScalarWave::Worldtube {
+
+/*!
+ * \brief A smoothly broken power law, that grows falls off to a constant value
```
PunJustice: that grows->thats growth, or reword.

src/Evolution/Systems/CurvedScalarWave/Worldtube/RadiusFunctions.cpp
```diff
@@ -0,0 +1,57 @@
...
+
+#include "Utilities/ErrorHandling/Error.hpp"
+
+namespace CurvedScalarWave::Worldtube {
+
+double smooth_broken_power_law(double orbit_radius, double alpha, double amp,
+                               double rb, double delta) {
+  if (alpha < 0. or alpha > 4.) {
+    ERROR(
+        "Only exponents between 0 and 4 have been tested for the broken law. "
```
PunJustice: broken power law

PunJustice (changes requested):

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_RadiusFunctions.cpp
nikwit: yes, the python tests are usually a good idea if there is no other test available but I think the ones there cover all regions quite well.

nikwit:
thanks, I incorporated your changes and changed the order of the parameters to something more natural. 

I also increased the minimum value of delta from 1e-3 to 1e-2. This seemed to cause an FPE in the CI for all compilers which I was not able to reproduce with clang or gcc locally or on a cluster. Probably something hardware specific.

PunJustice (approved):

src/Evolution/Systems/CurvedScalarWave/Worldtube/RadiusFunctions.hpp
```diff
@@ -0,0 +1,40 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+namespace CurvedScalarWave::Worldtube {
+
```
knelli2, line 7: Remove extra line breaks

src/Evolution/Systems/CurvedScalarWave/Worldtube/RadiusFunctions.cpp
```diff
@@ -0,0 +1,56 @@
...
+
+#include <cmath>
+#include <cstddef>
+
+#include "Utilities/ErrorHandling/Error.hpp"
+
+namespace CurvedScalarWave::Worldtube {
+
+double smooth_broken_power_law(double orbit_radius, double alpha, double amp,
+                               double rb, double delta) {
```
knelli2: [optional] Make these all `const` since you don't change them. Also can remove the extra space above

src/Evolution/Systems/CurvedScalarWave/Worldtube/RadiusFunctions.cpp
```diff
@@ -0,0 +1,56 @@
...
+  if (alpha < 0. or alpha > 4.) {
+    ERROR(
```
knelli2, line 17: Factor out these two checks into a `detail` namespace so they share the same wording for both this function and the deriv

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_RadiusFunctions.cpp
```diff
@@ -0,0 +1,78 @@
...
+          168. * func(x - 2. * h) - 672. * func(x - h) + 672. * func(x + h) +
+          -168. * func(x + 2. * h) + 32. * func(x + 3. * h) -
+          3. * func(x + 4. * h)) /
+         (840. * h);
+}
+
+void test_smooth_broken_power_law_derivative() {
+  const double amp = 1.234;
+  const double rb = 10.;
+  const Approx custom_approx = Approx::custom().epsilon(1.e-10).scale(1.0);
```
knelli2: Is this larger approx because of the wide range of radii tested below? Whatever the reason, maybe add a small comment explaining.

knelli2 (changes requested):
Squash directly along with the current fixup for @PunJustice's changes

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_RadiusFunctions.cpp
nikwit: The comparison is done with a finite difference method which I found to be accurate to about 1e-10. I added a small comment.

nikwit:
Sorry for the delay. I squashed your suggestions in, thank you for your reviews!

knelli2 (approved):

PR #6262, opened 2024-08-31
Fix printing tensors in GDB on mbot

nilsdeppe:
## Proposed changes

There were a number of subtleties, and unfortunately the GDB documentation is essentially non-existent, but this resolves the issue. It does make the output not as pretty since I need to print the storage index instead of the component suffix. I don't know of an easy way around this, though.

@wthrowe would you mind testing this on your machine in case there's some GDB version-dependent behavior going on?

I'm leaving the evaluation string so we can revisit this in the future without having to learn all that again.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


wthrowe (approved):
At least basic printing tensors works.

PR #6264, opened 2024-09-03
Add `UpdateFunctionsOfTime` action

nikwit:
Adds the action that adjusts the functions of time according to the motion of the particle.

~depends on #6255 #6261~

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeEvolvedVariables.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeEvolvedVariables.hpp
@@ -33,6 +33,7 @@ struct InitializeEvolvedVariables {
 
   using simple_tags =
       tmpl::list<variables_tag, dt_variables_tag, Tags::CurrentIteration,
+                 Tags::ExpirationTime, Tags::WorldtubeRadius,
                  ::Tags::HistoryEvolvedVariables<variables_tag>>;
   using return_tags = simple_tags;
 
@@ -41,7 +42,8 @@ struct InitializeEvolvedVariables {
   using mutable_global_cache_tags = tmpl::list<>;
   using simple_tags_from_options = tmpl::list<Tags::InitialPositionAndVelocity>;
   using argument_tags = tmpl::list<::Tags::TimeStepper<TimeStepper>,
-                                   Tags::InitialPositionAndVelocity>;
+                                   Tags::InitialPositionAndVelocity,
+                                   ::Tags::Time, Tags::ExcisionSphere<Dim>>;
   static void apply(
       const gsl::not_null<Variables<
           tmpl::list<Tags::EvolvedPosition<Dim>, Tags::EvolvedVelocity<Dim>>>*>
@@ -51,11 +53,20 @@ struct InitializeEvolvedVariables {
                                ::Tags::dt<Tags::EvolvedVelocity<Dim>>>>*>
           dt_evolved_vars,
       const gsl::not_null<size_t*> current_iteration,
+      const gsl::not_null<double*> expiration_time,
+      const gsl::not_null<double*> worldtube_radius,
       const gsl::not_null<::Tags::HistoryEvolvedVariables<variables_tag>::type*>
           time_stepper_history,
       const TimeStepper& time_stepper,
-      const std::array<tnsr::I<double, Dim>, 2>& initial_pos_and_vel) {
+      const std::array<tnsr::I<double, Dim>, 2>& initial_pos_and_vel,
+      const double initial_time, const ExcisionSphere<Dim>& excision_sphere) {
     *current_iteration = 0;
+
+    // the functions of time should be ready during the first step but not the
+    // second. We choose the arbitrary value of 1e-10 here to ensure this.
+    *expiration_time = initial_time + 1e-10;
+    *worldtube_radius = excision_sphere.radius();
+
     const size_t starting_order =
         time_stepper.number_of_past_steps() == 0 ? time_stepper.order() : 1;
     *time_stepper_history =
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/UpdateFunctionsOfTime.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/UpdateFunctionsOfTime.hpp
@@ -1,0 +1,153 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <memory>
+#include <unordered_map>
+#include <utility>
+
+#include "ControlSystem/UpdateFunctionOfTime.hpp"
+#include "DataStructures/DataBox/DataBox.hpp"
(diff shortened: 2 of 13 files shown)


tests/Unit/Helpers/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelpers.cpp
```diff
@@ -70,21 +72,25 @@ std::unique_ptr<DomainCreator<3>> worldtube_binary_compact_object(
       "      InitialAngularVelocity: [0.0, 0.0," +
       angular_velocity_stream.str() +
       "]\n"
-      "    TranslationMap:\n"
-      "      InitialValues: [[0.0, 0.0, 0.0], [0.0, 0.0, 0.], [0.0, 0.0, "
-      "0.0]]\n"
+      "    TranslationMap: None\n"
       "    ShapeMapA:\n"
-      "      LMax: 8\n"
+      "      LMax: 2\n"
```
knelli2, line 77: Why reduce l_max? For speed?

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/Test_Tags.cpp
```diff
@@ -71,8 +71,8 @@ void test_initial_position_velocity_tag() {
   const double orbital_radius = 7.;
   const double angular_vel = 0.1;
```
knelli2, line 72: [random line] Capitalize "Template" in commit message

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/InitializeEvolvedVariables.hpp
```diff
@@ -51,11 +53,15 @@ struct InitializeEvolvedVariables {
...
       const gsl::not_null<size_t*> current_iteration,
+      const gsl::not_null<double*> expiration_time,
       const gsl::not_null<::Tags::HistoryEvolvedVariables<variables_tag>::type*>
           time_stepper_history,
       const TimeStepper& time_stepper,
-      const std::array<tnsr::I<double, Dim>, 2>& initial_pos_and_vel) {
+      const std::array<tnsr::I<double, Dim>, 2>& initial_pos_and_vel,
+      const double initial_time) {
     *current_iteration = 0;
+    *expiration_time = initial_time + 1e-10;
```
knelli2: Please add a comment why you chose this value.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -430,6 +430,21 @@ struct CurrentIteration : db::SimpleTag {
...
+ * controlled by the worldtube singleton.
+ */
+struct ExpirationTime : db::SimpleTag {
+  using type = double;
+};
+
+/*!
+ * \brief The current worldtube radius held by the singleton.
+ */
+struct WorldtubeRadius : db::SimpleTag {
```
knelli2, line 444: Is this tag only used in a test? Or will it also be used in source code later?

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/UpdateFunctionsOfTime.hpp
```diff
@@ -0,0 +1,155 @@
...
+    all_updates["Rotation"] =
+        std::make_pair(std::move(angular_update), new_expiration_time);
+    all_updates["Expansion"] =
+        std::make_pair(std::move(expansion_update), new_expiration_time);
+    all_updates["SizeA"] = std::make_pair(size_a_update, new_expiration_time);
+    all_updates["SizeB"] =
+        std::make_pair(std::move(size_b_update), new_expiration_time);
+
+    Parallel::mutate<::domain::Tags::FunctionsOfTime,
+                     control_system::UpdateMultipleFunctionsOfTime>(
```
knelli2, line 147: If you want to avoid adding a dependency of the CSW system on the ControlSystem, you can just forward declare `control_system::UpdateMultipleFunctionsOfTime`. Then you can just add the dependency in the final exec (or test).

Probably we actually want to move `UpdateMultipleFunctionsOfTime` to some other common place like `ParallelAlgorithms`, but that's beyond this PR

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/UpdateFunctionsOfTime.hpp
```diff
@@ -0,0 +1,155 @@
...
+
+    DataVector angular_update(2);
+    DataVector expansion_update(2);
+    DataVector size_a_update(2);
+    DataVector size_b_update(2);
+
+    angular_update[0] = atan2(y, x);
+    angular_update[1] = (x * ydot - y * xdot) / square(r);
+    expansion_update[0] = r / grid_radius_particle;
+    expansion_update[1] = radial_vel / grid_radius_particle;
```
knelli2: Rather than initializing a non-const DV, and then assigning afterwards, do
```cpp
const DataVector angular_update{atan2(y, x), (x * ydot - y * xdot) / square(r)};
```
and same for the other variables

tests/Unit/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/Test_UpdateFunctionsOfTime.cpp
```diff
@@ -0,0 +1,205 @@
...
+struct MockMetavariables {
+  static constexpr size_t volume_dim = Dim;
+  using component_list = tmpl::list<MockWorldtubeSingleton<MockMetavariables>>;
+  using const_global_cache_tags =
+      tmpl::list<Tags::ExcisionSphere<Dim>, domain::Tags::Domain<3>,
+                 Tags::WorldtubeRadiusParameters,
+                 Tags::BlackHoleRadiusParameters>;
+};
+
+void test_particle_motion() {
```
knelli2: This test could do with a couple more comments explaining what's happening

knelli2 (changes requested):

tests/Unit/Helpers/Evolution/Systems/CurvedScalarWave/Worldtube/TestHelpers.cpp
nikwit: yes, the worldtube always has to be spherical so the shape map is never used. This is the minimum l_max allowed by the shape map.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: I included a use case in the fixup which uses the correct worldtube radius to compute the coefficients.

knelli2 (commented):
Except for the clang-tidy errors and some missing tags in the tests, looks good.

nikwit:
My bad, I accidentally left out some of the changes. I pushed a new fixup.

knelli2 (approved):
Looks good, squash

nikwit:
done, thanks for the review!

PR #6265, opened 2024-09-03
Fix FPEs in tests

nilsdeppe:
## Proposed changes

There are a couple MC neutrino tests I commented out because they're too hard for me to fix. That code isn't currently being used in simulations and will need a lot more debugging and testing in general. I'll talk with Francois Foucart about getting that fixed separately. My main goal is to just get FPEs on again in tests.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/FixConservatives.cpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/FixConservatives.cpp
@@ -6,6 +6,7 @@
 #include <algorithm>
 #include <cmath>
 #include <cstddef>
+#include <limits>
 #include <ostream>
 #include <pup.h>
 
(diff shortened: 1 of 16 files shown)


src/NumericalAlgorithms/RootFinding/TOMS748.hpp
```diff
@@ -345,10 +360,14 @@ std::pair<T, T> toms748_solve(F f, const T& ax, const T& bx, const T& fax,
   T e(static_cast<T>(1e5F));
   T fd(static_cast<T>(1e5F));
 
-  T c(std::numeric_limits<T>::signaling_NaN());
-  T d(std::numeric_limits<T>::signaling_NaN());
+  // T c(std::numeric_limits<T>::quiet_NaN());
+  // T d(std::numeric_limits<T>::quiet_NaN());
```
wthrowe: Remove.

src/Evolution/Systems/GrMhd/ValenciaDivClean/FixConservatives.cpp
```diff
@@ -350,13 +351,21 @@ bool FixConservatives::operator()(
                                          : static_cast<SimdType>(0);
 
     // Equation B.27 of Foucart
+    //
+    // We avoid division by zero in the SIMD case using a mask and effectively
+    // set the result to zero by dividing by the max double.
```
wthrowe: It looks like you decided on a different implementation than this, without the max double part.

src/Time/ChooseLtsStepSize.cpp
```diff
@@ -22,7 +22,10 @@ TimeDelta choose_lts_step_size(const Time& time, const double desired_step) {
   // n because of floating-point truncation.  The inner ceil call
   // avoids this problem.
   const auto desired_step_power =
-      static_cast<size_t>(std::ceil(std::log2(std::ceil(desired_step_count))));
+      desired_step == std::numeric_limits<double>::infinity()
```
wthrowe: This needs an `abs()`, since the step can be negative, or check for `desired_step_count == 0.0` instead.

wthrowe (changes requested):
Thanks for going through all these.  I agree we can deal with the MC problems later.

nilsdeppe:
I made the changes. Since they are pretty small I just went ahead and squashed it. Partly to get this in before more FPEs appear 🙃

wthrowe (approved):
Looks good.  I would have requested a direct squash if I'd thought of it.

PR #6267, opened 2024-09-03
Add Perlmutter machine

nilsvu:
## Proposed changes

Support for [Perlmutter](https://docs.nersc.gov/systems/perlmutter/).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupCCache.cmake
+++ b/cmake/SetupCCache.cmake
@@ -24,10 +24,9 @@ if(NOT USE_CCACHE)
   return()
 endif()
 
-find_program(CCACHE_EXEC ccache)
+find_package(ccache)
 
 if (NOT CCACHE_EXEC)
-  message(STATUS "Could not find ccache")
   return()
 endif()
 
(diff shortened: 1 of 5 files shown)


support/Machines/Perlmutter.yaml
```diff
@@ -0,0 +1,13 @@
...
+# See LICENSE.txt for details.
+
+Machine:
+  Name: Perlmutter
+  Description: |
+    DOE supercomputer at Lawrence Berkely National Laboratory
+    More information:
+    https://docs.nersc.gov/systems/perlmutter/
+  DefaultTasksPerNode: 8
+  DefaultProcsPerTasks: 32
```
knelli2, line 11: The [docs](https://docs.nersc.gov/systems/perlmutter/architecture/#cpu-nodes) say that the CPU nodes have 128 cores. Is there some threading you're accounting for here?

support/Machines/Perlmutter.yaml
```diff
@@ -0,0 +1,13 @@
...
+  DefaultQueue: "regular"
+  DefaultTimeLimit: "1-00:00:00"
```
knelli2, line 13: Not sure, but [docs](https://docs.nersc.gov/jobs/best-practices/#time-limits) say shorter jobs are better on this machine. But that could be more for GPU jobs. @ermost any insight for this?

cmake/SetupCCache.cmake
```diff
@@ -24,24 +24,29 @@ if(NOT USE_CCACHE)
...
+set(_CCACHE_LAUNCHER_ENV_VARS
   ${CCACHE_LAUNCHER_EXTRA_ENV_VARS}
   "CCACHE_SLOPPINESS=pch_defines,time_macros,include_file_mtime,\
-include_file_ctime,locale"
-  ${REAL_CCACHE_EXEC}
-)
+include_file_ctime")
+# The `locale` sloppiness is only available in ccache 3.6 and later
+if (CCACHE_VERSION VERSION_GREATER_EQUAL 3.6)
+  string(APPEND _CCACHE_LAUNCHER_ENV_VARS ",locale")
```
knelli2, line 44: Why not just install a newer ccache on perlmutter for us to use? I think your config uses the system ccache in `/usr/bin/ccache`

knelli2 (changes requested):

support/Machines/Perlmutter.yaml
nilsvu: Yep they are hyperthreaded. They document what to do for job submissions here: https://docs.nersc.gov/systems/perlmutter/running-jobs/

support/Machines/Perlmutter.yaml
nilsvu: I think it's like every other cluster where short jobs can start quicker because they can be squeezed in by the scheduler.

cmake/SetupCCache.cmake
nilsvu: It works 🤷‍♂️

support/Machines/Perlmutter.yaml
ermost: There are some quirks about 2hr jobs being charged less. but by and large Perlmutter behaves like a normal cluster (unlike other DOE machines).

knelli2 (approved):

PR #6269, opened 2024-09-04
Test complex elliptic DG operator

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticSolutions/Poisson/Lorentzian.hpp
+++ b/src/PointwiseFunctions/AnalyticSolutions/Poisson/Lorentzian.hpp
@@ -24,27 +24,28 @@ namespace detail {
 template <typename DataType, size_t Dim>
 struct LorentzianVariables {
   using Cache = CachedTempBuffer<
-      Tags::Field,
-      ::Tags::deriv<Tags::Field, tmpl::size_t<Dim>, Frame::Inertial>,
-      ::Tags::Flux<Tags::Field, tmpl::size_t<Dim>, Frame::Inertial>,
-      ::Tags::FixedSource<Tags::Field>>;
+      Tags::Field<DataType>,
+      ::Tags::deriv<Tags::Field<DataType>, tmpl::size_t<Dim>, Frame::Inertial>,
+      ::Tags::Flux<Tags::Field<DataType>, tmpl::size_t<Dim>, Frame::Inertial>,
+      ::Tags::FixedSource<Tags::Field<DataType>>>;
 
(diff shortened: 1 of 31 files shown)


nilsvu:
@wthrowe would you be able to review this one as well? (continuing the quest of DG with complex numbers)

src/PointwiseFunctions/AnalyticSolutions/Poisson/Lorentzian.hpp
```diff
@@ -31,8 +31,11 @@ struct LorentzianVariables {
       ::Tags::Flux<Tags::Field<DataType>, tmpl::size_t<Dim>, Frame::Inertial>,
       ::Tags::FixedSource<Tags::Field<DataType>>>;
 
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
   const tnsr::I<DataVector, Dim>& x;
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
   const double constant;
+  // NOLINTNEXTLINE(cppcoreguidelines-avoid-const-or-ref-data-members)
   const double complex_phase;
```
wthrowe: The clang-tidy warnings look right, at least for these last two.  Why should they be ignored?  Similar comments for the other file.

wthrowe (changes requested):

src/PointwiseFunctions/AnalyticSolutions/Poisson/Lorentzian.hpp
nilsvu: I don't want functions in this class be able to modify this value. It's the user-defined value from options that gets passed down to this struct. I don't understand why clang-tidy want's me to remove the `const` when really I want it to be const.

wthrowe: The [documentation](https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/avoid-const-or-ref-data-members.html) links to the core guidelines rule and some extra reading.  I don't agree with everything in there, but that section seems pretty good.  The short answer for your class is that the functions already can't modify the value because they are const member functions.

nilsvu: Ok makes sense, I removed the `const`

wthrowe (approved):
Looks good.

PR #6273, opened 2024-09-05
Add tags used to smoothly turn on the self force

nikwit:
The tags will be used in a future PR.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
@@ -29,12 +29,20 @@
 
 namespace CurvedScalarWave::Worldtube::OptionTags {
 SelfForceOptions::SelfForceOptions() = default;
-SelfForceOptions::SelfForceOptions(double mass_in, size_t iterations_in)
-    : mass(mass_in), iterations(iterations_in) {}
+SelfForceOptions::SelfForceOptions(const double mass_in,
+                                   const size_t iterations_in,
+                                   const double turn_on_time_in,
+                                   const double turn_on_interval_in)
+    : mass(mass_in),
+      iterations(iterations_in),
+      turn_on_time(turn_on_time_in),
+      turn_on_interval(turn_on_interval_in) {}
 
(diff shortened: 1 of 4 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -29,12 +29,20 @@
...
-SelfForceOptions::SelfForceOptions(double mass_in, size_t iterations_in)
-    : mass(mass_in), iterations(iterations_in) {}
+SelfForceOptions::SelfForceOptions(const double mass_in,
+                                   const size_t iterations_in,
+                                   const double turn_on_time_in,
+                                   const double turn_on_interval_in)
+    : mass(mass_in),
+      iterations(iterations_in),
+      turn_on_time(turn_on_time_in),
+      turn_on_interval(turn_on_interval_in) {}
```
knelli2, line 39: Is it ok if the turn on interval is zero? I believe it's in the denominator of the exponent

nikwit: Yes, good catch. I limited it to 1M now which is still very low but can not cause any numerical issues.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.cpp
```diff
@@ -36,7 +36,11 @@ SelfForceOptions::SelfForceOptions(const double mass_in,
     : mass(mass_in),
       iterations(iterations_in),
       turn_on_time(turn_on_time_in),
-      turn_on_interval(turn_on_interval_in) {}
+      turn_on_interval(turn_on_interval_in) {
+  if (turn_on_interval_in < 1.) {
+    ERROR("The turn on interval must be at least 1 M.");
+  }
```
knelli2: The `lower_bound()` in the option in the `hpp` already takes care of this. No need to check here.

knelli2 (commented):
Looks good, squash

nikwit:
Done, thanks for the review!

knelli2 (approved):

PR #6276, opened 2024-09-06
Add volume integrands for XCTS asymptotic quantities.

iago-mendes:
## Proposed changes

This PR adds volume integrands for the calculation of ADM mass, ADM linear momentum, and center of mass in the XCTS system.

The primary goal of using infinite volume integrals instead of infinite surface integrals is to improve numerical accuracy. We haven't seen significant accuracy improvements yet, and we have found some analytic test cases in which the volume integrals do not give the expected results (which could be a bug in the method or an assumption being broken somewhere in these cases). Then, we'll keep using the only surface integrals in the ID parameter control for now.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
+++ b/src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
@@ -31,12 +31,11 @@ void local_adm_integrals(
     const tnsr::II<DataVector, 3>& inv_spatial_metric,
     const tnsr::ii<DataVector, 3>& extrinsic_curvature,
     const Scalar<DataVector>& trace_extrinsic_curvature,
+    const tnsr::I<DataVector, 3, Frame::Inertial>& inertial_coords,
     const InverseJacobian<DataVector, 3, Frame::ElementLogical,
                           Frame::Inertial>& inv_jacobian,
     const Mesh<3>& mesh, const Element<3>& element,
-    const DirectionMap<3, tnsr::i<DataVector, 3>>& conformal_face_normals,
-    const DirectionMap<3, tnsr::I<DataVector, 3>>&
-        conformal_face_normal_vectors) {
+    const DirectionMap<3, tnsr::i<DataVector, 3>>& conformal_face_normals) {
   // Initialize integrals to 0
   adm_mass->get() = 0;
   for (int I = 0; I < 3; I++) {
@@ -82,6 +81,8 @@ void local_adm_integrals(
         extrinsic_curvature, mesh, boundary_direction);
     const auto face_trace_extrinsic_curvature = dg::project_tensor_to_boundary(
         trace_extrinsic_curvature, mesh, boundary_direction);
+    const auto face_inertial_coords = dg::project_tensor_to_boundary(
+        inertial_coords, mesh, boundary_direction);
     // This projection could be avoided by using
     // domain::Tags::DetSurfaceJacobian from the DataBox, which is computed
     // directly on the face (not projected). That would be better on Gauss
@@ -103,15 +104,18 @@ void local_adm_integrals(
     const auto& face_mesh = mesh.slice_away(boundary_direction.dimension());
     const auto& conformal_face_normal =
         conformal_face_normals.at(boundary_direction);
-    const auto& conformal_face_normal_vector =
-        conformal_face_normal_vectors.at(boundary_direction);
+    const auto face_normal_magnitude = magnitude(conformal_face_normal);
+    const auto flat_face_normal = tenex::evaluate<ti::i>(
+        conformal_face_normal(ti::i) / face_normal_magnitude());
 
-    // Compute curved area element
+    // Compute curved and flat area elements
     const auto face_sqrt_det_conformal_metric =
         Scalar<DataVector>(sqrt(get(determinant(face_conformal_metric))));
     const auto conformal_area_element =
         area_element(face_inv_jacobian, boundary_direction,
                      face_inv_conformal_metric, face_sqrt_det_conformal_metric);
+    const auto flat_area_element =
+        euclidean_area_element(face_inv_jacobian, boundary_direction);
 
     // Compute surface integrands
     const auto mass_integrand = Xcts::adm_mass_surface_integrand(
@@ -124,26 +128,24 @@ void local_adm_integrals(
             face_inv_extrinsic_curvature, face_trace_extrinsic_curvature);
     const auto center_of_mass_integrand =
         Xcts::center_of_mass_surface_integrand(face_conformal_factor,
-                                               conformal_face_normal_vector);
+                                               face_inertial_coords);
 
     // Contract surface integrands with face normal
     const auto contracted_mass_integrand =
         tenex::evaluate(mass_integrand(ti::I) * conformal_face_normal(ti::i));
     const auto contracted_linear_momentum_integrand = tenex::evaluate<ti::I>(
-        linear_momentum_integrand(ti::I, ti::J) * conformal_face_normal(ti::j));
+        linear_momentum_integrand(ti::I, ti::J) * flat_face_normal(ti::j));
 
     // Take integrals
     adm_mass->get() += definite_integral(
         get(contracted_mass_integrand) * get(conformal_area_element),
         face_mesh);
     for (int I = 0; I < 3; I++) {
-      adm_linear_momentum->get(I) +=
-          definite_integral(contracted_linear_momentum_integrand.get(I) *
-                                get(conformal_area_element),
-                            face_mesh);
-      center_of_mass->get(I) += definite_integral(
-          center_of_mass_integrand.get(I) * get(conformal_area_element),
+      adm_linear_momentum->get(I) += definite_integral(
+          contracted_linear_momentum_integrand.get(I) * get(flat_area_element),
           face_mesh);
+      center_of_mass->get(I) += definite_integral(
+          center_of_mass_integrand.get(I) * get(flat_area_element), face_mesh);
     }
   }
 }
--- a/src/PointwiseFunctions/Xcts/AdmMass.cpp
+++ b/src/PointwiseFunctions/Xcts/AdmMass.cpp
@@ -33,4 +33,67 @@ tnsr::I<DataVector, 3> adm_mass_surface_integrand(
   return result;
 }
 
+void adm_mass_volume_integrand(
+    gsl::not_null<Scalar<DataVector>*> result,
+    const Scalar<DataVector>& conformal_factor,
+    const Scalar<DataVector>& conformal_ricci_scalar,
+    const Scalar<DataVector>& trace_extrinsic_curvature,
+    const Scalar<DataVector>&
+        longitudinal_shift_minus_dt_conformal_metric_over_lapse_square,
+    const Scalar<DataVector>& energy_density,
+    const tnsr::II<DataVector, 3>& inv_conformal_metric,
+    const tnsr::iJK<DataVector, 3>& deriv_inv_conformal_metric,
+    const tnsr::Ijj<DataVector, 3>& conformal_christoffel_second_kind,
+    const tnsr::i<DataVector, 3>& conformal_christoffel_contracted,
+    const tnsr::iJkk<DataVector, 3>& deriv_conformal_christoffel_second_kind) {
+  tenex::evaluate(
+      result,
+      1. / (16. * M_PI) *
+          (deriv_inv_conformal_metric(ti::i, ti::J, ti::K) *
+               conformal_christoffel_second_kind(ti::I, ti::j, ti::k) +
+           inv_conformal_metric(ti::J, ti::K) *
+               deriv_conformal_christoffel_second_kind(ti::i, ti::I, ti::j,
+                                                       ti::k) +
+           conformal_christoffel_contracted(ti::l) *
+               inv_conformal_metric(ti::J, ti::K) *
+               conformal_christoffel_second_kind(ti::L, ti::j, ti::k) -
+           deriv_inv_conformal_metric(ti::i, ti::I, ti::J) *
+               conformal_christoffel_contracted(ti::j) -
+           inv_conformal_metric(ti::I, ti::J) *
+               deriv_conformal_christoffel_second_kind(ti::i, ti::K, ti::k,
+                                                       ti::j) -
+           conformal_christoffel_contracted(ti::l) *
+               inv_conformal_metric(ti::L, ti::J) *
+               conformal_christoffel_contracted(ti::j) -
+           conformal_factor() * conformal_ricci_scalar() -
+           2. / 3. * pow<5>(conformal_factor()) *
+               square(trace_extrinsic_curvature()) +
+           pow<5>(conformal_factor()) / 4. *
+              longitudinal_shift_minus_dt_conformal_metric_over_lapse_square() +
+           16. * M_PI * pow<5>(conformal_factor()) * energy_density()));
+}
+
+Scalar<DataVector> adm_mass_volume_integrand(
+    const Scalar<DataVector>& conformal_factor,
+    const Scalar<DataVector>& conformal_ricci_scalar,
+    const Scalar<DataVector>& trace_extrinsic_curvature,
+    const Scalar<DataVector>&
+        longitudinal_shift_minus_dt_conformal_metric_over_lapse_square,
+    const Scalar<DataVector>& energy_density,
+    const tnsr::II<DataVector, 3>& inv_conformal_metric,
+    const tnsr::iJK<DataVector, 3>& deriv_inv_conformal_metric,
+    const tnsr::Ijj<DataVector, 3>& conformal_christoffel_second_kind,
+    const tnsr::i<DataVector, 3>& conformal_christoffel_contracted,
+    const tnsr::iJkk<DataVector, 3>& deriv_conformal_christoffel_second_kind) {
+  Scalar<DataVector> result;
+  adm_mass_volume_integrand(
+      make_not_null(&result), conformal_factor, conformal_ricci_scalar,
+      trace_extrinsic_curvature,
+      longitudinal_shift_minus_dt_conformal_metric_over_lapse_square,
+      energy_density, inv_conformal_metric, deriv_inv_conformal_metric,
+      conformal_christoffel_second_kind, conformal_christoffel_contracted,
+      deriv_conformal_christoffel_second_kind);
+  return result;
+}
+
 }  // namespace Xcts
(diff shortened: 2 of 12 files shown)


src/PointwiseFunctions/Xcts/AdmMass.hpp
```diff
@@ -52,4 +52,74 @@ tnsr::I<DataVector, 3> adm_mass_surface_integrand(
...
+ * We cast the ADM mass as an infinite volume integral by applying Gauss' law on
+ * the surface integral defined in `adm_mass_surface_integrand`:
+ *
+ * \begin{equation}
+ *   M_\text{ADM} = \frac{1}{16\pi}
+ *                  \int_{V_\infty} \Big(
+ *                    \bar\gamma^{jk} \partial_i \bar\Gamma^i_{jk}
+ *                    + \bar\Gamma^i_{jk} \partial_i \bar\gamma^{jk}
+ *                    - \bar\gamma^{ij} \partial_i \bar\Gamma_j
+ *                    - \bar\Gamma_j \partial_i \bar\gamma^{ij}
```
nilsvu: I think this is missing 2 terms: $+\gamma^{jk} \Gamma_l \Gamma^l_{jk} - \gamma^{lj} \Gamma_l \Gamma_j$ where $\Gamma_l = \Gamma^k_{kl}$.
I don't think these cancel since $\gamma^{jk}\Gamma^l_{jk} \neq \gamma^{lj}\Gamma^k_{jk}$ (or do they somehow?)

src/PointwiseFunctions/Xcts/AdmMass.hpp
```diff
@@ -52,4 +52,74 @@ tnsr::I<DataVector, 3> adm_mass_surface_integrand(
...
+ *                    - 8 \bar D^2 \psi
+ *                  \Big) dV,
+ * \end{equation}
+ *
+ * where we can use the momentum constraint to replace $8 \bar D^2 \psi$ with
```
nilsvu: <s>momentum</s> Hamiltonian constraint (Eq. 3.110 in \cite BaumgarteShapiro)

src/PointwiseFunctions/Xcts/AdmMass.hpp
```diff
@@ -52,4 +52,74 @@ tnsr::I<DataVector, 3> adm_mass_surface_integrand(
...
+ *
+ * \begin{equation}
+ *   8 \bar D^2 \psi = \psi \bar R + \frac{2}{3} \psi^5 K^2
+ *                     - \frac{1}{4} \psi^5 \frac{1}{\alpha^2}
+ *                         \Big[ (\bar L \beta)_{ij} - \bar u_{ij} \Big]
+ *                         \Big[ (\bar L \beta)^{ij} - \bar u^{ij} \Big].
```
nilsvu: Add matter term please: $-16\pi \psi^5\rho$

src/PointwiseFunctions/Xcts/AdmMass.cpp
```diff
@@ -33,4 +33,56 @@ tnsr::I<DataVector, 3> adm_mass_surface_integrand(
...
+      1. / (16. * M_PI) *
+          (inv_conformal_metric(ti::J, ti::K) *
+               deriv_conformal_christoffel_second_kind(ti::i, ti::I, ti::j,
+                                                       ti::k) +
+           conformal_christoffel_second_kind(ti::I, ti::j, ti::k) *
+               deriv_inv_conformal_metric(ti::i, ti::J, ti::K) -
+           inv_conformal_metric(ti::I, ti::J) *
+               deriv_conformal_christoffel_second_kind(ti::i, ti::K, ti::k,
+                                                       ti::j) -
+           conformal_christoffel_second_kind(ti::K, ti::k, ti::j) *
```
nilsvu: Take `conformal_christoffel_contracted` as input argument.

src/PointwiseFunctions/Xcts/CenterOfMass.hpp
```diff
@@ -17,33 +17,66 @@ namespace Xcts {
...
+/// @{
+/*!
+ * \brief Volume integrand for the center of mass calculation.
+ *
+ * We cast the center of mass as an infinite volume integral by applying Gauss'
+ * law on the surface integral defined in `center_of_mass_surface_integrand`:
+ *
+ * \begin{equation}
+ *   C_\text{CoM}^i = \frac{3}{8 \pi M_\text{ADM}}
+ *               \int_{V_\infty} 4 \psi^3 \partial_j \psi n^i n^j \, dV.
```
nilsvu: I think this is missing a term: $+ \frac{2}{3} n^i \psi^4$
This is comes from the term $\partial_i n^i =2/r$ assuming a flat metric and $n^i=x^i/r$.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -141,9 +143,10 @@ void local_adm_integrals(
           definite_integral(contracted_linear_momentum_integrand.get(I) *
                                 get(conformal_area_element),
                             face_mesh);
-      center_of_mass->get(I) += definite_integral(
-          center_of_mass_integrand.get(I) * get(conformal_area_element),
-          face_mesh);
+      center_of_mass->get(I) +=
+          definite_integral(contracted_center_of_mass_integrand.get(I) *
+                                get(conformal_area_element),
```
nilsvu: I think for Gauss' law you have to either compute derivatives of the conformal unit normal correctly ($\partial_i \bar{n}^j$), which needs second derivatives of the coordinate maps (seems complicated). Or you compute this integral assuming flat space and $n^i=x^i/r$ (seems easier). To do Gauss' law for this, use $n^i=x^i/r$ when computing the integrands and then integrate both the surface and volume with the Euclidean area/volume element. I think I was wrong suggesting to use the conformal face normal here before.

src/PointwiseFunctions/Xcts/CenterOfMass.cpp
```diff
@@ -6,20 +6,42 @@
 namespace Xcts {
 
 void center_of_mass_surface_integrand(
-    gsl::not_null<tnsr::I<DataVector, 3>*> result,
+    gsl::not_null<tnsr::II<DataVector, 3>*> result,
```
nilsvu: I suggest you don't do this change, since you'll just contract with the unit normal again for the surface integral.

src/PointwiseFunctions/Xcts/CenterOfMass.hpp
```diff
@@ -17,33 +17,66 @@ namespace Xcts {
...
+ * \param conformal_factor the conformal factor $\psi$
+ * \param deriv_conformal_factor the gradient of the conformal factor
+ * $\partial_i \psi$
+ * \param unit_normal the outward-pointing unit normal $n^i = x^i / r$
+ */
+void center_of_mass_volume_integrand(
     gsl::not_null<tnsr::I<DataVector, 3>*> result,
     const Scalar<DataVector>& conformal_factor,
+    const tnsr::i<DataVector, 3, Frame::Inertial>& deriv_conformal_factor,
     const tnsr::I<DataVector, 3>& unit_normal);
```
nilsvu: I suggest you pass the coordinates $x^i$ and the Euclidean radius $r$ to this function and to the surface integrand. That seems easiest, because then you don't have to compute the Euclidean normal outside this function and you can include the correct powers of $r$ within this function.

nilsvu (changes requested):
I haven't reviewed the tests yet. Please push changes as fixup commits.

src/PointwiseFunctions/Xcts/CenterOfMass.hpp
iago-mendes: I had an algebraic error that made me believe this term would be zero. Thanks for pointing it out!

src/PointwiseFunctions/Xcts/AdmMass.hpp
iago-mendes: If I understand it right, these extra terms would only show up if we take the covariant derivative $\bar D_i$ when applying Gauss' law instead of a partial derivative $\partial_i$ (this is what I'm currently doing). I remember we briefly talked about these two options (referencing Saul's recent DG paper), and we decided that $\partial_i$ was the way to go.

That said, after redoing the calculations, I found an extra term involving the conformal factor first derivative:
$\partial_i ( - 8 \bar\gamma^{ij} \partial_j \psi ) = - 8 \bar\gamma^{ij} \partial_i \partial_j \psi - 8 \partial_i \bar\gamma^{ij} \partial_j \psi = - 8 \bar D^2 \psi - 8 \partial_i \bar\gamma^{ij} \partial_j \psi $.
Again, this is only if we use $\partial_i$ in Gauss' Law. Looking at Baumgarte-Shapiro's equation (3.146), they seem to use $\bar D_i$ in Gauss' Law.

I'm not sure what approach we should take here, but both of them require updates to these integrands.

src/PointwiseFunctions/Xcts/CenterOfMass.hpp
iago-mendes: Btw, I'm assuming the 3 was a typo. Did you mean $+ \frac{2}{r} \psi^4 n^i$?

src/PointwiseFunctions/Xcts/CenterOfMass.cpp
iago-mendes: I was thinking that it would be nice to keep the surface integrands consistent between each other. That is, wouldn't it be confusing if some integrands need to be contracted before taking the surface integral ($P_{ADM}$ and $M_{ADM}$) and this one doesn't? Or maybe it doesn't matter much because these integrands are probably not going to be used by other parts of the code?

src/PointwiseFunctions/Xcts/CenterOfMass.hpp
iago-mendes: Since $r$ is just `magnitude(coords)`, do you think we could just pass the coordinates $x^i$? Should we do the same with the surface integrand?

nilsvu: You can also just pass in the coords, but then you may have to repeat the magnitude computation in multiple places. Not a big deal, but can be avoided.

src/PointwiseFunctions/Xcts/CenterOfMass.cpp
```diff
@@ -27,20 +29,22 @@ void center_of_mass_volume_integrand(
...
     const tnsr::i<DataVector, 3, Frame::Inertial>& deriv_conformal_factor,
-    const tnsr::I<DataVector, 3>& unit_normal) {
-  tenex::evaluate<ti::I>(result, 3. / (8. * M_PI) * 4. *
-                                     pow<3>(conformal_factor()) *
-                                     deriv_conformal_factor(ti::j) *
-                                     unit_normal(ti::I) * unit_normal(ti::J));
+    const tnsr::I<DataVector, 3>& coords) {
+  tenex::evaluate<ti::I>(
+      result,
+      3. / (4. * M_PI * pow<2>(magnitude(coords)())) *
```
nilsvu: Move `magnitude(coords)` out of this expression into a separate line above. It allocates a DataVector, so it doesn't help to have it as part of this tensor expression.

nilsvu (changes requested):
You can squash these two commits (including this change)

src/PointwiseFunctions/Xcts/CenterOfMass.cpp
nilsvu: Nah, rather don't do unnecessary computation. You can write in the docs how the integrand is supposed to be integrated.

src/PointwiseFunctions/Xcts/AdmMass.hpp
nilsvu: You need the covariant Laplacian of $\psi$ to replace it with the Hamiltonian constraint, so I would go with the covariant derivatives and the Christoffel terms I wrote above. As you say, either way you have to account for the covariant derivatives with Christoffel terms.

src/PointwiseFunctions/Xcts/CenterOfMass.hpp
nilsvu: Yes good catch 👍

nilsvu:
Looks good, you can squash the fixup.

src/PointwiseFunctions/Xcts/AdmMass.hpp
iago-mendes: Ok, that makes sense. This raises another point. If we're using $\bar D_i$ when applying Gauss' law, then can we still assume a flat space in this transformation?

nilsvu: For Gauss' law you transform $\oint n_i \sqrt{\gamma} F^i = \int \partial_i (\sqrt{\gamma} F^i) = \int \sqrt{\gamma} \nabla_i F^i$ (see Eqs. (7-10) in https://arxiv.org/abs/physics/9802027v1). So you get the covariant derivative from moving the metric determinant through the partial derivative.
So, you integrate with the curved area/volume element, get a covariant derivative from Gauss' law, and expand it into Christoffel terms where needed.

iago-mendes: This is helpful, thanks! The test that I had in `Test_AdmMass` was using flat area/volume elements and started failing after these changes. Once I updated this test to use curved area/volume elements, I started getting the correct values again. So, things seem consistent.

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
iago-mendes: Sounds good. For now, I'm only taking the infinite surface integral, but I'm doing it with an Euclidean area element to indicate what we'll do later.

nilsvu:
New fixups look good too, you can squash

src/PointwiseFunctions/Xcts/AdmMass.hpp
nilsvu: Ok great! Both approaches work, we just have to be consistent. Please note in the docs for the integrand function if it needs to be integrated with the flat or conformal area/volume element.

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
```diff
@@ -195,6 +196,15 @@ void test_local_adm_integrals(const double& distance,
         deriv_conformal_metric, inv_conformal_metric);
     const auto conformal_christoffel_contracted = tenex::evaluate<ti::i>(
         conformal_christoffel_second_kind(ti::J, ti::i, ti::j));
+    const auto deriv_conformal_christoffel_second_kind = partial_derivative(
```
nilsvu: At the moment you don't use this yet, right? Same for Ricci scalar/tensor and longitudinal shift below. Maybe hold off with adding these until you use them in this test.

tests/Unit/PointwiseFunctions/Xcts/Test_CenterOfMass.cpp
```diff
@@ -99,55 +106,156 @@ void test_center_of_mass_surface_integral(const double distance) {
...
         surface_integral.get(I) += definite_integral(
-            surface_integrand.get(I) * get(conformal_area_element), face_mesh);
+            surface_integrand.get(I) * get(flat_area_element), face_mesh);
       }
     }
   }
 
   // Check result
-  auto custom_approx = Approx::custom().epsilon(10. / distance).scale(1.0);
+  auto custom_approx = Approx::custom().epsilon(TOLERANCE).scale(1.0);
```
nilsvu, line 128: Can you keep this at O(1/distance)?

tests/Unit/PointwiseFunctions/Xcts/Test_CenterOfMass.cpp
```diff
@@ -99,55 +106,156 @@ void test_center_of_mass_surface_integral(const double distance) {
...
+                                   const double z_shift) {
+  // Get Schwarzschild solution in isotropic coordinates.
+  const double mass = 1;
+  const Xcts::Solutions::Schwarzschild solution(
+      mass, Xcts::Solutions::SchwarzschildCoordinates::Isotropic);
+  const double horizon_radius = 0.5 * mass;
+
+  // Set up domain
+  const size_t h_refinement = 1;
+  const size_t p_refinement = 11;
```
nilsvu, line 144: This is very high refinement. My understanding is that the accuracy of the integral is currently not limited by resolution, so can you lower this?

nilsvu (changes requested):
Looks good. You can squash everything, including this.

tests/Unit/PointwiseFunctions/Xcts/Test_CenterOfMass.cpp
iago-mendes: It doesn't work for this test because the CoM integral "diverges" as we increase R. I believe this happens due to round-off errors when canceling larger and larger terms. I describe a bit of this at the top of this file.

tests/Unit/PointwiseFunctions/Xcts/Test_CenterOfMass.cpp
iago-mendes: This was the lowest resolution that I was able to find in order to keep the tolerance at $10^{-3}$. Figure 3.2.1 in these [convergence plots](https://iago-mendes.notion.site/Convergence-plots-90906819ac0742bcb8cdb8cb35292505) shows that the volume integral residual gets slightly better for higher resolution.

I think there must be some change that we can make in the calculation of the CoM integrals to get around these round-off errors, but this is probably something for a future PR.

nilsvu (dismissed):

nilsvu:
ADM mass test times out, probably just relax the timeout

iago-mendes:
> ADM mass test times out, probably just relax the timeout

How do I relax the timeout?

nilsvu:
Add `// [[TimeOut, 10]]` above the `SPECTRE_TEST_CASE`

iago-mendes:
It seems like it's still timing out for some compilers. Should I relax it even further?

nilsvu:
Yes just set a reasonable timeout, and/or speed up the test

iago-mendes:
> Yes just set a reasonable timeout, and/or speed up the test

Done. The ADM mass test now succeeds after <20s for the compilers that were failing before (gcc-9, gcc-11). Now the fails are from `PartialDerivs`, so it's unrelated to my changes.

nilsvu (approved):

PR #6280, opened 2024-09-07
Fix yaml issues with BBH ID example nb

vaishakp:
## Proposed changes
1. Fix loading and writing of yaml files with multiple yaml documents in the same stream. Previously, the existing tutorial presumably worked with a single document yaml stream. The latest `BinaryBlackHole.yaml` has multiple docs. This change allows such streams to be read and edited in this example notebook.

1. Make minor changes in path names to make it more informative.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
Hey Vaishak! Thanks for looking at this. Indeed this tutorial is very out of date. I was going to replace it by the one from SXScon.

vaishakp:
> Hey Vaishak! Thanks for looking at this. Indeed this tutorial is very out of date. I was going to replace it by the one from SXScon.

The one from SXSCon is nice and easy for BBH ID+evolution at the highest level. I however find that this one is informative and could be preserved as an ID example that works at a slightly lower level, working directly with the input files. In my opinion, it will be useful to keep both.

PR #6283, opened 2024-09-11
Change non-monotonic AM predictor coefficients

wthrowe:
Either version works, but this version makes the analytic expressions nicer and is what will be in the paper.

Note that this does not affect the BBH paper, because that uses the monotonic method.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/TimeSteppers/AdamsMoultonPc.cpp
+++ b/src/Time/TimeSteppers/AdamsMoultonPc.cpp
@@ -396,26 +396,17 @@ void AdamsMoultonPc<Monotonic>::add_boundary_delta_impl(
   } else {
     adams_lts::AdamsScheme scheme{adams_lts::SchemeType::Implicit,
                                   current_order};
-    auto remote_scheme = scheme;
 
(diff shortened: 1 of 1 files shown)


nilsdeppe (dismissed):

nilsdeppe:
@wthrowe I don't know if this relates at all to your group meeting update, but in any case, I'll let you merge it when you are ready :)

wthrowe:
That's a good question, actually.  Maybe I should hold off on this until I figure out exactly what we want.

knelli2:
@nilsdeppe @wthrowe Maybe post a quick synopsis of what was brought up so we have a record of it?

wthrowe:
Basically, what's on develop, what's in this PR, and what I wrote in my paper draft are all different.  They all satisfy the requirements on the method, so it's not clear which one to choose.  (And I'm now also considering a fourth version.)

wthrowe:
Decided to go with a simpler version without the conditional on alignment.  This should be ready now.

nilsdeppe (approved):

Issue #6246, opened 2024-08-24
ICERM CCE Feedback


nilsdeppe:
Here are notes I took at ICERM from feedback I got from various folks on how we could improve CCE usability.

- [x] A grid point-based approach would be helpful where we give people the grid point locations to dump and they can dump the metric and derivatives there. (#6398 )
   * Zach would like radial derivatives
   * ETK would like Cartesian derivatives.
- [x] We should have a glob-approach and then combine the HDF5 file on the fly so that people don’t need to do `combine-h5` or something to get a single file. One option would be to make the `combine-h5` distributed, but it would be even easier for users if we just do the combine before starting the CCE computation.
   1. Add HDF5 combining to `CceReduceWorldtube` by giving folks the ability to specify a glob for the file names. E.g. `Lev1_*/Run/CceOutputR0215.h5` (#6398 )
   2. There may be overlapping times in the files. We need to remove these. We should have an option to select times from either the earlier or later file. Ideally the data should be the same, but I don't think we need to check that. ETK uses first file. (partially #6298. This uses later file.)
- [ ] We need to document reasonable temporal sampling for the worldtube input. That is, how often should people dump worldtube data? We should provide some heuristic advice and comment on inspiral vs. merger since the time scales change. I would guess in SpEC and SpECTRE we get around this by dump some every `N` steps, but I don't know. @keefemitman @markscheel could you comment on how we decide how frequently to dump the worldtube?
- [ ] Could we compress the files? E.g. RPDMB, spectral fit in time? This is lower priority compared to the other points.
- [x] Document the Ylm convention we assume since there are different ones. This probably is just pointing to spherepack docs, roughly speaking. @nilsdeppe can add this. (#6271 )
- [x] We should provide an executable that dumps a text file of the theta & phi (or maybe 2 files) coordinates to use for a given ell. @knelli2 is working on this. (#6284 )
- [x] We should provide a C++ routine that takes in a given ell and returns the theta and phi points that need to be dumped to. @nilsdeppe will do this. The idea is to have a function in a test file that computes the points and then we have a snippet in the documentation. We can then test that the function returns the correct values for different ell. (#6271 )
- [ ] Rename `CharacteristicExtract` to `CharacteristicEvolution`
- [x] Rename `ReduceCceWorldtube` to `PreprocessCceWorldtube` (or something like that) (#6398 )
[...]

PR #6284, opened 2024-09-12
Add executable to write CCE worldtube coords to a file

knelli2:
## Proposed changes

Part of #6246. A C++ executable was explicitly chosen for this even though it is very simple so it can be compiled and distributed like our other CCE executables. A python version would have to have some kind of PyBindings packaged in and that didn't seem like the best approach.

I added a function that does this, bound it in python just in case we want to use it, and then made the executable.

I currently don't add this to the tarball that we create at each release because there is still work that needs to be done in the CCE executable and the ReduceCceWorldtube executable. Once that is done, then I will add this executable.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Executables/WriteCceWorldtubeCoordsToFile/WriteCceWorldtubeCoordsToFile.cpp
```diff
@@ -0,0 +1,79 @@
...
+  const std::string output_file = vars["output_file"].as<std::string>();
+  const bool overwrite = vars["force"].as<bool>();
+
+  try {
+    ylm::write_strahlkorper_coords_to_text_file(
+        radius, l_max, center, output_file, ylm::AngularOrdering::Cce,
+        overwrite);
+  } catch (const std::exception& exception) {
+    Parallel::printf("%s\n", exception.what());
+    return 1;
```
knelli2, line 77: I didn't know the best way to handle this. If there's an error like if the file already exists, the wallclock time that's printed is wrong (this isn't a charm exec), it prinds Node and Core = 0 which doesn't really make sense in this context, and it prints the file where the error happened (not super useful to outside users).

I debated having the `write_strahlkorper_coords_to_text_file` function return a `std::optional<std::string>` that is the error message and then handle the error printing here. Thoughts are welcome.

src/NumericalAlgorithms/SphericalHarmonics/IO/StrahlkorperCoordsToTextFile.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * be overwritten with the \p overwrite option.
+ *
+ * The second overload will construct a spherical `ylm::Strahlkorper` with the
+ * given \p radius, \p l_max, and \p center.
+ */
+template <typename Frame>
+void write_strahlkorper_coords_to_text_file(
+    const Strahlkorper<Frame>& strahlkorper,
+    const std::string& output_file_name, AngularOrdering ordering,
+    bool overwrite = false);
```
nilsdeppe: [optional] name this `overwrite_file` to be explicit?

src/NumericalAlgorithms/SphericalHarmonics/IO/StrahlkorperCoordsToTextFile.hpp
```diff
@@ -0,0 +1,39 @@
...
+ * \brief Writes the collocation points of a `ylm::Strahlkorper` to an output
+ * text file.
+ *
+ * \details The ordering of the points can be either the typical
+ * `ylm::Spherepack` ordering or CCE ordering that works with libsharp. Also, an
+ * error will occur if the output file already exists, but the output file can
+ * be overwritten with the \p overwrite option.
+ *
+ * The second overload will construct a spherical `ylm::Strahlkorper` with the
+ * given \p radius, \p l_max, and \p center.
```
nilsdeppe, line 25: It would be useful to specify the output file format:
```
x0 y0 z0
x1 y1 z1
```

src/Executables/WriteCceWorldtubeCoordsToFile/WriteCceWorldtubeCoordsToFile.cpp
nilsdeppe: I wouldn't worry about it. The main message is that the file already exists and as long as people can figure that out easily I think it's fine.

nilsdeppe (changes requested):
Thank you for doing this! LGTM, just a couple small suggestions. Please squash right away :)

knelli2:
@nilsdeppe Squashed, thanks!

nilsdeppe (approved):

PR #6285, opened 2024-09-12
Fixes for nodegroups running on multiple nodes

nilsdeppe:
## Proposed changes

There's still some issue I need to resolve, but these were the easy ones to find.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/ArrayCollection/ReceiveDataForElement.hpp
+++ b/src/Parallel/ArrayCollection/ReceiveDataForElement.hpp
@@ -38,14 +38,14 @@ struct ReceiveDataForElement {
                     const ElementId<Dim>& element_to_execute_on,
                     typename ReceiveTag::temporal_id instance,
                     ReceiveData receive_data) {
-    ERROR(
-        "The multi-node code hasn't been tested. It should work, but be aware "
-        "that I haven't tried yet.");
+    [[maybe_unused]] const size_t my_node = Parallel::my_node<size_t>(cache);
     auto& element_collection = db::get_mutable_reference<
         typename ParallelComponent::element_collection_tag>(
         make_not_null(&box));
-    // Note: We'll be able to do a counter-based check here too once that
-    // works for LTS in `SendDataToElement`
+
+    ASSERT(
+        element_collection.count(element_to_execute_on) == 1,
+        "ElementId " << element_to_execute_on << " is not on node " << my_node);
     ReceiveTag::insert_into_inbox(
         make_not_null(&tuples::get<ReceiveTag>(
             element_collection.at(element_to_execute_on).inboxes())),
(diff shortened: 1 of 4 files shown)


src/Parallel/ArrayCollection/ReceiveDataForElement.hpp
```diff
@@ -38,14 +38,14 @@ struct ReceiveDataForElement {
...
+    [[maybe_unused]] const size_t my_node = Parallel::my_node<size_t>(cache);
     auto& element_collection = db::get_mutable_reference<
         typename ParallelComponent::element_collection_tag>(
         make_not_null(&box));
-    // Note: We'll be able to do a counter-based check here too once that
-    // works for LTS in `SendDataToElement`
+
+    ASSERT(
+        element_collection.count(element_to_execute_on) == 1,
+        "ElementId " << element_to_execute_on << " is not on node " << my_node);
```
knelli2, line 48: [optional] You could put the `Parallel::my_node<size_t>(cache)` here to avoid the `[[maybe_unused]]` temporary

src/Parallel/ArrayCollection/SpawnInitializeElementsInCollection.hpp
```diff
@@ -24,7 +24,8 @@ struct SpawnInitializeElementsInCollection {
                     Parallel::GlobalCache<Metavariables>& cache,
                     const ArrayIndex& /*array_index*/,
                     const double /*unused_but_we_needed_to_reduce_something*/) {
-    auto my_proxy = Parallel::get_parallel_component<ParallelComponent>(cache);
+    auto my_proxy = Parallel::get_parallel_component<ParallelComponent>(
+        cache)[Parallel::my_node<size_t>(cache)];
```
knelli2, line 28: Just for my understanding, this used to do a broadcast to the whole collection for every element on this node? But that didn't matter because you only ran on a single node? And now you only send the messages to the local node?

knelli2 (commented):
Squash directly if you decide to change anything

src/Parallel/ArrayCollection/SpawnInitializeElementsInCollection.hpp
nilsdeppe: Right, and the issue was that it would try to run an element that wasn't on its node and would then fail with "I can't find this element!"

nilsdeppe:
I squashed the move of `my_node`. I agree, that's cleaner. I also rebased on develop. Thank you for the review!!

src/Parallel/ArrayCollection/SpawnInitializeElementsInCollection.hpp
knelli2: Thanks for the explanation!

knelli2 (approved):

PR #6291, opened 2024-09-16
Update Inspiral.yaml 

AlexCarpenter46:
## Proposed changes

Updates the Inspiral.yaml to be what worked well in @geoffrey4444's BBH runs for the paper. This also starts to set up the volume data needed for the transition to ringdown.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

Do we want to cleanly terminate the run at a certain separation?

### DIFF
--- a/support/Pipelines/Bbh/Inspiral.yaml
+++ b/support/Pipelines/Bbh/Inspiral.yaml
@@ -82,23 +82,25 @@ DomainCreator:
         ConstraintPreservingBjorhus:
           Type: ConstraintPreservingPhysical
     UseEquiangularMap: True
-    CubeScale: 1.0
+    # Found that a CubeScale of 1.2 worked well for equal mass non-spinning
+    # inspirals and allowed for more common horizon finds.
+    CubeScale: 1.2
     # These values have been tested to be stable for full equal mass mergers
     # with constraints ~10^-7 for L=1, P=10
     InitialRefinement:
-      ObjectAShell:   [{{ L     }}, {{ L     }}, {{ L + 1 }}]
-      ObjectACube:    [{{ L + 1 }}, {{ L + 1 }}, {{ L     }}]
-      ObjectBShell:   [{{ L     }}, {{ L     }}, {{ L + 1 }}]
-      ObjectBCube:    [{{ L + 1 }}, {{ L + 1 }}, {{ L     }}]
-      Envelope:       [{{ L     }}, {{ L     }}, {{ L     }}]
-      OuterShell:     [{{ L     }}, {{ L     }}, {{ L + 1 }}]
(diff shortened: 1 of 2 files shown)


support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -327,12 +329,33 @@ EventsAndTriggers:
...
+      - ChangeSlabSize:
+          DelayChange: 0
+          StepChoosers:
+            - Constant: 0.01
       - ObservationAhC
+      - ObserveFields:
+          SubfileName: ForContinuation
+          VariablesToObserve:
+            - Lapse
+            - ConstraintEnergy
```
nilsdeppe: I'm a bit surprised we write out the constraint energy and lapse. I think these could be computed from the 3 evolved variables. It would be good to have a comment why they are observed instead of computed when needed.

nilsdeppe (commented):
LGTM, just one small requset for clarification :)

support/Pipelines/Bbh/Inspiral.yaml
AlexCarpenter46: Oh, we don't think we need to output those for the transition to ringdown script to work, they were originally there to see where the constraints were blowing up or getting large. I can take that out

support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -327,12 +329,33 @@ EventsAndTriggers:
...
   - Trigger:
-      SeparationLessThan:
-        Value: 2.5
+      And:
+        - SeparationLessThan:
+            Value: 2.38 # for q=1, spin=0, quasicircular
+        - Slabs:
+            EvenlySpaced:
+              Interval: 1
+              Offset: 0
```
knelli2: I believe this trigger (and the `And:`) are unnecessary because `SeparationLessThan` is checked every slab

support/Pipelines/Bbh/Inspiral.yaml
knelli2: For the pipeline and transition to ringdown, yeah we probably don't need these.

knelli2 (changes requested):

AlexCarpenter46:
Squashed those changes in @nilsdeppe and @knelli2 :)

nilsdeppe (dismissed):

knelli2 (dismissed):

knelli2:
@AlexCarpenter46 Could you rebase on latest develop? There's a conflict now

AlexCarpenter46:
Rebased :) @knelli2

knelli2 (approved):

PR #6298, opened 2024-09-19
Add function to combine H5 dat files

knelli2:
## Proposed changes

Towards #6246.

This function is distinct from the existing CLI function `spectre combine-h5 dat` because this takes care of overlapping times in the dat subfiles (we could optionally replace the python function with this one if we bind this function). Also, this is a C++ function and not a python function because it will be used in the `ReduceCceWorldtube` executable which must be statically linked.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/H5/CombineH5.cpp
+++ b/src/IO/H5/CombineH5.cpp
@@ -3,26 +3,34 @@
 
 #include "IO/H5/CombineH5.hpp"
 
-#include <boost/program_options.hpp>
+#include <array>
 #include <cstddef>
 #include <cstdlib>
 #include <iterator>
 #include <limits>
+#include <optional>
+#include <sstream>
 #include <string>
+#include <unordered_map>
 #include <vector>
 
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/Matrix.hpp"
 #include "Domain/Domain.hpp"
 #include "Domain/Structure/BlockGroups.hpp"
 #include "IO/H5/AccessType.hpp"
 #include "IO/H5/CheckH5PropertiesMatch.hpp"
+#include "IO/H5/Dat.hpp"
 #include "IO/H5/File.hpp"
 #include "IO/H5/SourceArchive.hpp"
 #include "IO/H5/TensorData.hpp"
 #include "IO/H5/VolumeData.hpp"
+#include "IO/Logging/Verbosity.hpp"
 #include "Parallel/Printf/Printf.hpp"
 #include "Utilities/Algorithm.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/FileSystem.hpp"
+#include "Utilities/Gsl.hpp"
 #include "Utilities/MakeString.hpp"
 #include "Utilities/Serialization/Serialize.hpp"
 #include "Utilities/StdHelpers.hpp"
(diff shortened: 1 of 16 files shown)


src/IO/H5/CombineH5.cpp
```diff
@@ -149,4 +157,186 @@ void combine_h5_vol(const std::vector<std::string>& file_names,
...
+        num_time_map.at(dat_filename)[index] = dimensions[0];
+        earliest_time = times(0, 0);
+        continue;
+      }
+
+      // Determine if the earliest time of the previous file (previous = later
+      // in the sequence of files since we are looping backward) is before
+      // any of the times in this file. If so, don't include those times.
+      std::optional<size_t> row = times.rows() - 1;
+      while (times(row.value(), 0) >= earliest_time.value()) {
```
nilsdeppe: I'm not sure that we guarantee that our HDF5 Dat output is actually sorted in time. I think you need to first sort the matrix by the first column before combining. I think what you need to do is store the index of the sorted matrix, and then sort again in the next for loop below.

src/IO/H5/CombineH5.cpp
```diff
@@ -149,4 +157,186 @@ void combine_h5_vol(const std::vector<std::string>& file_names,
...
+        ERROR_NO_TRACE("Version of dat file "
+                       << dat_filename << " in H5 file " << filename
+                       << " doesn't match other H5 files.");
+      }
+
+      // This is the first file. We can't make any decisions here so just store
+      // the number of times and the earliest time of this file
+      if (not earliest_time.has_value()) {
+        num_time_map.at(dat_filename)[index] = dimensions[0];
+        earliest_time = times(0, 0);
```
nilsdeppe: This is not guarantee to be the earliest time. It's the first time written, but that doesn't have to be the earliest time, I think. I believe our observers do not enforce ordering on write.

src/IO/H5/CombineH5.cpp
```diff
@@ -149,4 +157,186 @@ void combine_h5_vol(const std::vector<std::string>& file_names,
     new_file.close_current_object();
   }
 }
+
+void combine_h5_dat(const std::vector<std::string>& h5_files_to_combine,
```
nilsdeppe, line 292: I think you assume that the files are in increasing order in time. It would be good to add a check for that since otherwise users may be very surprised. Another option would be to just sort the list in increasing order for the subfile(s).

src/IO/H5/CombineH5.cpp
```diff
@@ -149,4 +157,186 @@ void combine_h5_vol(const std::vector<std::string>& file_names,
...
+        output_dat_file =
+            &output_h5_file.insert<h5::Dat>(dat_filename, legend, version);
+      }
+
+      const std::optional<size_t>& num_times = num_time_map.at(dat_filename)[i];
+
+      // Only append data if we include data from this file
+      if (num_times.has_value()) {
+        // Always start with row 0
+        const Matrix data_to_append = input_dat_file.get_data_subset(
```
nilsdeppe: This will need to load the entire matrix, sort it, and then trim it before write.

src/IO/H5/CombineH5.hpp
```diff
@@ -6,8 +6,44 @@
...
+ *
+ * \details If there are overlapping times, the "latest" one is always used;
+ * meaning if you have data in `File1.h5` and `File2.h5` and if the first time
+ * in `File2.h5` is before some times in `File1.h5`, those times in `File1.h5`
+ * will be discarded and won't appear in the combined H5 file.
+ *
+ * If the H5 files in \p h5_files_to_combine have other types of subfiles, those
+ * will be ignored and will not appear in \p output_h5_filename. This function
+ * also assumes that the times in each of the \p h5_files_to_combine are already
+ * sorted.
```
nilsdeppe: I think this second assumption is generally not valid for spectre, unfortunately. If you would like to keep it, I think you should add a check that it's true because someone will inevitably run this over a file where it's not true.

src/IO/H5/CombineH5.hpp
```diff
@@ -6,8 +6,44 @@
...
                     bool check_src = true);
+
+/*!
+ * \brief Combine the `h5::Dat` subfiles of multiple `h5::H5File`s into a single
+ * H5 file.
+ *
+ * \details If there are overlapping times, the "latest" one is always used;
+ * meaning if you have data in `File1.h5` and `File2.h5` and if the first time
+ * in `File2.h5` is before some times in `File1.h5`, those times in `File1.h5`
+ * will be discarded and won't appear in the combined H5 file.
```
nilsdeppe: I think the constraint is quite strong and should be explicit: the files past must be in increasing time order.

tests/Unit/IO/H5/Test_CombineH5.cpp
```diff
@@ -0,0 +1,226 @@
...
+
+  if (file_system::check_if_file_exists(combined_filename)) {
+    file_system::rm(combined_filename, true);
+  }
+
+  for (const std::string& filename : individual_filenames) {
+    if (file_system::check_if_file_exists(filename)) {
+      file_system::rm(filename, true);
+    }
+  }
```
nilsdeppe, line 123: It would be good to add tests for catching unordered files passed in, and for unordered data in the subfiles. I worry that these will both be common mistakes.

nilsdeppe (changes requested):
Please read all the comments first. I give a few different options that I think would all resolve my concerns and I don't have a preference for which to do :)

knelli2:
I decided to simply assert that the H5 files were monotonically increasing in their earliest time in the files. I think something else should be responsible for actually putting the H5 files in order properly. This also allows for the times in each dat file to be unordered. One downside is that we have to read all data in from a dat file first, sort it, then trim the overlapping times we don't need. But that shouldn't be too bad.

To facilitate this sorting (because a `Matrix` can't be sorted easily) I just added overloads to the `dat.get_data()` to return the data as `std::vector<std::vector<double>>` which can be sorted very easily. These are the two new commits before the `fixup`

src/IO/H5/CombineH5.cpp
```diff
@@ -355,10 +353,14 @@ void combine_h5_dat(const std::vector<std::string>& h5_files_to_combine,
...
-      const Matrix times = dat_file.get_data_subset({0}, 0, dimensions[0]);
+      auto times = dat_file.get_data_subset<std::vector<std::vector<double>>>(
+          {0}, 0, dimensions[0]);
+      alg::sort(times,
+                [](const std::vector<double>& v1,
+                   const std::vector<double>& v2) { return v1[0] < v2[0]; });
 
       // Makes things easier below.
-      if (UNLIKELY(times.rows() == 0)) {
+      if (UNLIKELY(times.size() == 0)) {
```
nilsdeppe: clang-tidy: `empty()` when you squash

src/IO/H5/CombineH5.cpp
```diff
@@ -373,35 +375,43 @@ void combine_h5_dat(const std::vector<std::string>& h5_files_to_combine,
...
+      // Check that the earliest time of the previous file (previous = later
+      // in the sequence of files since we are looping backward) is after the
+      // first time of this file. We require the files to be passed in
+      // increasing order by their first time.
+      if (UNLIKELY(times[0][0] >= earliest_time.value())) {
+        ERROR_NO_TRACE("The H5 files passed in "
+                       << h5_files_to_combine
+                       << " are not monotonically increasing in their first "
+                          "times for dat file "
+                       << dat_filename << "");
```
nilsdeppe: remove `<< ""`?

src/IO/H5/CombineH5.cpp
```diff
@@ -373,35 +375,43 @@ void combine_h5_dat(const std::vector<std::string>& h5_files_to_combine,
...
+      }
+
+      // Determine if the earliest time of the previous file is before any of
+      // the times in this file. If so, don't include those times.
+      size_t row = times.size() - 1;
+      while (times[row][0] >= earliest_time.value()) {
+        // This should have been taken care of before so we should never get
```
nilsdeppe: I had to think about this (terrible, I know), but maybe change to ~ `Make sure we don't reach row 0 since that would mean the files are not ordered. We should've checked for this above, so this is an additional safety check.`

nilsdeppe (commented):
LGTM, a few minor suggestions you can make when you squash. Thanks for doing this!

nilsdeppe (approved):

PR #6303, opened 2024-09-24
Add support for tetrahedral connectivity and use scipy to generate it

nilsdeppe:
## Proposed changes

This provides an easy way to fill gaps in for domains with Gauss points. There are several optimizations possible to make it run faster if needed.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Visualization/Python/GenerateXdmf.py
```diff
@@ -219,10 +223,16 @@ def _xmf_grid(
             )
     else:
         # Cover volume
+        topology_type = {3: "Hexahedron", 2: "Quadrilateral"}[topo_dim]
+        connectivity_name = "connectivity"
+        if use_tetrahedral_connectivity:
```
nilsvu, line 226: In Python it's easier to read if you move the code above into an `else` clause (scopes are more relaxed in Python so the variable definitions are available outside the `else`)

src/Visualization/Python/GenerateXdmf.py
```diff
@@ -501,6 +518,15 @@ def generate_xdmf(
     show_default=True,
     help="The coordinates to use for visualization",
 )
+@click.option(
+    "--use-tetrahedral-connectivity",
+    is_flag=True,
+    default=False,
+    help=(
+        "Use a tetrahedral connectivity called tetrahedral_connectivity in "
+        "the HDF5 file"
```
nilsvu: Can you add here and/or in other docs what effect this will have, why you might want this, etc? Just looking for some context to tell the user why this exists and how to use it. Can also just point to the docs of GenerateTetrahedralConnectivity.py

tests/Unit/Visualization/Python/Test_GenerateXdmf.py
```diff
@@ -72,6 +72,46 @@ def test_generate_xdmf(self):
             ),
         )
 
+    def test_tetrahedral_generate_xdmf(self):
```
nilsvu: Is this test the same as above? Then you can move the two tests into a for loop

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+    connectivity may even fail in qhull, which is unfortunately difficult
+    to fix.
+
+    After the tetrahedral connectivity has been written you can run
+    'generate-xdmf' with the flag '--use-tetrahedral-connectivity'. ParaView
+    volume renderings are sometimes nicer with tetrahedral connectivity and this
+    can be used to fill gaps between finite-difference or Gauss elements.
+
+    If this algorithm is too slow, one possible improvement is to apply
+    qhull to each block of the domain and then connect the blocks to each
```
nilsvu, line 58: Note that this could also be done in parallel for each block or H5 file? I think parallelization over H5 files with `multiprocessing.Pool` would be the easiest way to speed this up, since it probably isn't IO bound.

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+      h5files: List of H5 volume data files.
+      subfile_name: Volume data subfile in the H5 files.
+      start_time: Optional. The earliest time at which to start visualizing. The
+        start-time value is included.
+      stop_time: Optional. The time at which to stop visualizing. The stop-time
+        value is not included.
+      stride: Optional. View only every stride'th time step.
+      coordinates: Optional. Name of coordinates dataset. Default:
+        "InertialCoordinates".
+      delete_tetrahedral_connectivity: Optional. Delete the dataset and return.
```
nilsvu: What's the purpose of this option? If the idea is to overwrite the dataset then for consistency with other tools I suggest you rename the option to `force`.

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+        Default: False
+    """
+    h5files = [(h5py.File(filename, "a"), filename) for filename in h5files]
+
+    if not subfile_name:
+        subfiles = available_subfiles(
+            (h5file for h5file, _ in h5files), extension=".vol"
+        )
+        raise RequiredChoiceError(
```
nilsvu: wrap in:

```py
        if len(subfiles) == 1:
            subfile_name = subfiles[0]
            logger.info(
                f"Selected subfile {subfile_name} (the only available one)."
            )
        else:
            # RequiredChoiceError
```

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+
+    Given the coordinates, this generates a tetrahedral connectivity that can
+    be read in by ParaView or other visualization software (e.g. VisIt or yt).
+    It uses the scipy.spatial.Delaunay class to generate the connectivity, which
+    uses the Qhull library (github.com/qhull/qhull/), which uses the quickhull
+    algorithm and scales as O(n^2) in the worst case. Thus, generating the
+    connectivity can take several minutes per temporal ID. Unfortunately,
+    depending on the particular grid point distribution, generating the
+    connectivity may even fail in qhull, which is unfortunately difficult
+    to fix.
```
nilsvu, line 42: Add a note that this is faster than ParaView's Delaunay filter and only has to be done once, since it's stored in the file

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+            if start_time is not None and time < start_time:
+                continue
+            if stop_time is not None and time > stop_time:
+                break
+
+            data_at_id = vol_subfile[temporal_id]
+
+            if delete_tetrahedral_connectivity:
+                if "tetrahedral_connectivity" in data_at_id:
+                    del data_at_id["tetrahedral_connectivity"]
```
nilsvu: Note that to free up the memory in the h5 file for this deleted dataset you actually have to repack the h5 file (depending on the [file space strategy](https://docs.hdfgroup.org/hdf5/rfc/FileSpaceManagement.pdf)). This is done in `DeleteSubfiles.py`. So I suggest you name this option `force` and just delete the `continue` in the next line. And then later we can extend `DeleteSubfiles.py` to take a glob pattern of dataset names so it can clean up datasets in multiple observations.

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+                coords = np.column_stack((x_coords, y_coords, z_coords))
+            else:
+                coords = np.column_stack((x_coords, y_coords))
+
+            logger.info(
+                f"Generating tetrahedral connectivity at {temporal_id}/{time}. "
+                "This may take a few minutes and may even fail depending on "
+                "the grid structure."
+            )
+            delaunay = spatial.Delaunay(coords)
```
nilsvu: This won't connect points across H5 files, right? Does that lead to visualization artifacts? `Delaunay` supports [iteratively adding points](https://docs.scipy.org/doc/scipy/reference/generated/scipy.spatial.Delaunay.add_points.html#scipy.spatial.Delaunay.add_points) so you could triangulate across files, though I'm not sure where to store this global triangulation.

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+            data_at_id.create_dataset(
+                "tetrahedral_connectivity", data=delaunay.simplices.flatten()
+            )
+
+    for h5file in h5files:
+        h5file[0].close()
+
+
+@click.command(
+    name="generate-tetrahedral-connectivity",
```
nilsvu, line 156: (optional) Maybe just name `triangulate`?

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+        "The earliest time at which to start visualizing. The start-time "
+        "value is included."
+    ),
+)
+@click.option(
+    "--stop-time",
+    type=float,
+    help=(
+        "The time at which to stop visualizing. The stop-time value is "
+        "not included."
```
nilsvu: Actually I think it _is_ included, no? This might be a documentation bug in GenerateXdmf.py as well.

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+                "Specify '--subfile-name' / '-d' to select a"
+                " subfile containing volume data."
+            ),
+            choices=subfiles,
+        )
+
+    if not subfile_name.endswith(".vol"):
+        subfile_name += ".vol"
+
+    for h5file, filename in h5files:
```
nilsvu: You can relatively easily show a progress bar here so the script doesn't appear "frozen" without any indication that it is doing anything. Here's how:

```py
import rich.progress

progress = rich.progress.Progress(
        rich.progress.TextColumn("[progress.description]{task.description}"),
        rich.progress.BarColumn(),
        rich.progress.MofNCompleteColumn(),
        rich.progress.TimeRemainingColumn(),
        disable=(len(h5files) == 1),
    )
task_id = progress.add_task("Processing files", total=len(h5files))

with progress:
    for h5file, filename in progress.track(h5files, task_id=task_id):
        # ...
        progress.update(task_id, advance=1) # don't remember if this is needed
```

Docs are here: https://rich.readthedocs.io/en/stable/progress.html

src/Visualization/Python/GenerateTetrahedralConnectivity.py
```diff
@@ -0,0 +1,208 @@
...
+    h5files,
+    subfile_name: str,
+    relative_paths: bool = True,
+    start_time: Optional[float] = None,
+    stop_time: Optional[float] = None,
+    stride: int = 1,
+    coordinates: str = "InertialCoordinates",
+    delete_tetrahedral_connectivity: bool = False,
+):
+    """Generate tetrahedral connectivity using scipy.spatial.Delaunay
```
nilsvu, line 32: Question: how does this handle shared grid points on LGL element boundaries? The tetrahedra connecting these points would be degenerate.

nilsvu (changes requested):

src/Visualization/Python/GenerateTetrahedralConnectivity.py
nilsdeppe: I don't feel too strongly, I used this so it matches up with GenerateXdmf `use-tetrahedral-connectivity` but I agree it's very long

src/Visualization/Python/GenerateTetrahedralConnectivity.py
nilsdeppe: That's right, so you need to first combine HDF5 files. I think it is doable to generate across HDF5 files, but considering the cost scales as N^2, I don't expect it to be practically useful to want to do that a lot... The bookkeeping is tricky because you need to get all the right indices in the end.

src/Visualization/Python/GenerateTetrahedralConnectivity.py
nilsdeppe: I haven't read the paper that describes qhull, but as far as I can tell it works fine and doesn't produce any double connectivity.

src/Visualization/Python/GenerateTetrahedralConnectivity.py
nilsdeppe: Yes, but then you wouldn't connect between blocks, which is one of the things I'm looking for this to do... However, we could still parallelize over blocks and then just connect the blocks afterward.

src/Visualization/Python/GenerateTetrahedralConnectivity.py
nilsvu: Ah ok maybe mention in the docs that you should run `combine-h5` first to avoid gaps in the rendering? And/or make this script take only a single H5 file rather than a list

src/Visualization/Python/GenerateTetrahedralConnectivity.py
nilsvu: Up to you

src/Visualization/Python/GenerateTetrahedralConnectivity.py
nilsvu: (optional) showing a progress bar (over timesteps rather than files as suggested before) instead of the log messages would be nicer

nilsvu:
LGTM please squash.

nilsvu (approved):

PR #6304, opened 2024-09-24
Hitchhiker’s Guide Update

SunPeike:
Updated the Hitchhiker’s Guide for SpECTRE. 
Added necessary steps for using ParaView from clusters

### DIFF
--- a/docs/Tutorials/BeginnersTutorial.md
+++ b/docs/Tutorials/BeginnersTutorial.md
@@ -204,10 +204,14 @@ the Paraview server, run
 pvserver &
 ```
 
-\note on compute clusters you should use the cluster-specific instructions for
-using ParaView since you may require MPI and specifying which port to use. For
-example, `mpirun -n 1 pvserver -p 11111` is likely more appropriate on a
-cluster.
+\note For cluster users, check cluster-specific instructions for ParaView.
+You may require MPI and specify which port to use.
+You can run `mpirun -np 1 pvserver -p PARAVIEW_REMOTE_PORT` on the remote
+cluster. Here, `PARAVIEW_REMOTE_PORT` is the port that the ParaView server
+will use on the cluster. You may choose it to be, for example, 11112.
+With 'pvserver' running on the remote machine, open a new local terminal.
+Locally, run
+`ssh -L11111:YOUR_CLUSTER:PARAVIEW_REMOTE_PORT USERNAME@YOUR_CLUSTER`.
 
 The `&` is so that the server runs in the background. If you hit `Enter`
 a couple times you'll get back to being able to type commands. You should see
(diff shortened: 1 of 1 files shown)


docs/Tutorials/BeginnersTutorial.md
```diff
@@ -204,10 +204,8 @@ the Paraview server, run
 pvserver &
 ```
 
-\note on compute clusters you should use the cluster-specific instructions for
-using ParaView since you may require MPI and specifying which port to use. For
-example, `mpirun -n 1 pvserver -p 11111` is likely more appropriate on a
-cluster.
+\note On compute clusters you should use the cluster-specific instructions for
+using ParaView since you may require MPI and specifying which port to use. You can run `mpirun -np 1 pvserver -p PARAVIEW_REMOTE_PORT` on the remote cluster, where the `PARAVIEW_REMOTE_PORT` is the port that the paraView server will use on the cluster, you may choose it to be, for example, 11112. With 'pvserver' running on the remote machine, open a new local terminal and run `ssh -L11111:YOUR_CLUSTER:PARAVIEW_REMOTE_PORT USERNAME@YOUR_CLUSTER`.
```
knelli2: Change "on the cluster, you may..." to "on the cluster. You may...`

docs/Tutorials/BeginnersTutorial.md
```diff
@@ -221,8 +219,12 @@ Accepting connection(s): 92bbb69f2af2:11111
 
 This means it's waiting for you to connect some external Paraview session (the
 "client") to the server. Now, ***outside*** the container, start a session of
-Paraview 5.10.1. (Again, you must use this version otherwise it won't work
-properly.) Go to `File > Connect`. Click `Add Server`. Name it whatever you
+Paraview. 
+
+\note To figure out the version of ParaView to use, run `mpirun -n 1 pvserver --version`, then download ***exactly the same*** version of Paraview. At the time when writing this tutorial, we are using ParaView 5.11.1
```
knelli2: I like the placement of the note here, but I'd keep the original wording all the way through "won't work properly.)" in the first paragraph. Then have your `\note` but clarify that that command you wrote is only if you are using a cluster. Also, since we sometimes have different versions of Paraview on different clusters, we should avoid putting a version in this `\note`.

Then the next paragraph starts exactly like you have it with "Go to..."

knelli2 (changes requested):
Thank you for doing this! It's greatly appreciated when people take the time to update our tutorials :)

Couple overall things:

- Both in our code and in our docs, we limit the length of a line to 80 characters. Please split the lines you added to adhere to this.
- You can squash these changes into your original changes, and remove the merge commits on your branch. If you're not sure how to do this, we can go over how to do it on Slack.

knelli2 (approved):

knelli2:
@SunPeike Could you update the PR description? You don't say anything about Paraview 5.11.1 anymore.

knelli2:
Ignoring unrelated errors

PR #6308, opened 2024-09-25
Remove zeros after LtsCoefficients arithmetic

wthrowe:
## Proposed changes

Remove zeros after LtsCoefficients arithmetic.  These are the coefficients in a linear combination of Variables objects, so avoiding terms with zero coefficients is a worthwhile minor optimization.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/TimeSteppers/AdamsLts.cpp
+++ b/src/Time/TimeSteppers/AdamsLts.cpp
@@ -4,8 +4,10 @@
 #include "Time/TimeSteppers/AdamsLts.hpp"
 
 #include <algorithm>
+#include <cmath>
 #include <cstddef>
 #include <iterator>
+#include <limits>
 #include <type_traits>
 #include <utility>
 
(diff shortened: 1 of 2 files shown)


src/Time/TimeSteppers/AdamsLts.cpp
```diff
@@ -87,6 +88,15 @@ LtsCoefficients& add_assign_impl(LtsCoefficients& a, const LtsCoefficients& b,
       ++write;
     }
   }
+  {
+    auto write = a.begin();
+    for (const auto& entry : a) {
+      if (get<2>(entry) != 0.0) {
```
nilsdeppe, line 103: Is there any reason to think that these numbers might be approximately zero to the level we don't care about?

nilsdeppe (dismissed):

src/Time/TimeSteppers/AdamsLts.cpp
wthrowe: You're right, these aren't exact.  This should have a roundoff check of some sort.

nilsdeppe:
fixup LGTM. Please squash. I should've said you can squash directly on this :)

wthrowe:
I almost did, but decided since the change I made was nowhere near the PR comment it was better to make it obvious.

nilsdeppe (approved):

PR #6311, opened 2024-09-26
Add Sonic HPC

vaishakp:
## Proposed changes

Adds the Sonic HPC at ICTS-TIFR

### Code review checklist

- [ x] The code is documented and the documentation renders correctly. Run
  Then open `index.html`.
- [ x] The code follows the stylistic and code quality guidelines listed in the
- [ x] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### DIFF
--- a/docs/Installation/InstallationOnClusters.md
+++ b/docs/Installation/InstallationOnClusters.md
@@ -116,3 +116,24 @@ The second command with `SPECTRE_DEBUG=ON` enables sanity checks and
 optimizations. This means it can be used in production-level runs to ensure
 there aren't any subtle bugs that might only arise after a decently
 long simulation.
+
+## Sonic at ICTS-TIFR
+
+Cluster documentation: https://it.icts.res.in/docs/sonic-cluster
+
+1. Clone the spectre source code and set `SPECTRE_HOME` to point it.
+2. Create a directory for building SpECTRE (`SPECTRE_BUILD`).
+3. Setup the software environment
+    ```
+    . /mnt/pfs/vaishak.p/soft/modules-5.2.0/init/bash
+    ```
+    Add these lines to your `.bashrc` and source it
+    if you would like to make it persistent.
+4. Source the `$SPECTRE_HOME/support/Environments/Sonic.sh` script.
+5. Build SpECTRE by changing to `$SPECTRE_BUILD` and using
+   one of the variants of the `spectre_run_cmake` commands.
+   This by default compiles SpECTRE in Release mode,
+   without debug symbols and LTO, and with system malloc.
+   If you wish to compile in debug mode, with LTO/JEMALLOC, or other options,
+   simply suffix the additional cmake flags to the command.
+6. Use `make` to build your targets.
--- a/support/Environments/sonic.sh
+++ b/support/Environments/sonic.sh
@@ -1,0 +1,64 @@
+#!/bin/env sh
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+spectre_setup_modules() {
+    echo "All modules on Sonic are provided by the system"
+}
+
+spectre_load_modules() {
+    # The order here is important
+    module load sxs
+    module load spectre-env > /dev/null 2>&1
+}
+
+spectre_unload_modules() {
+    # The order here is important
+    module unload spectre-env > /dev/null 2>&1
+    module unload sxs
+}
+
+spectre_run_cmake_clang() {
+    if [ -z ${SPECTRE_HOME} ]; then
+        echo "You must set SPECTRE_HOME to the cloned SpECTRE directory"
+        return 1
+    fi
+
+    spectre_load_modules > /dev/null 2>&1
+    cmake -D CMAKE_C_COMPILER=clang \
+          -D CMAKE_CXX_COMPILER=clang++ \
+          -D CMAKE_Fortran_COMPILER=flang \
+          -D CHARM_ROOT=$CHARM_ROOT \
+          -D MEMORY_ALLOCATOR=SYSTEM \
+          -D CMAKE_BUILD_TYPE=Release \
+          -D BUILD_PYTHON_BINDINGS=ON \
+          -D ENABLE_PARAVIEW=ON \
+          -D MACHINE=Sonic \
+          -D USE_XSIMD=yes \
+          -D DEBUG_SYMBOLS=OFF \
+          "$@" \
+          $SPECTRE_HOME
+}
+
+spectre_run_cmake_gcc() {
+    if [ -z ${SPECTRE_HOME} ]; then
+        echo "You must set SPECTRE_HOME to the cloned SpECTRE directory"
+        return 1
+    fi
+
+    spectre_load_modules > /dev/null 2>&1
+    cmake -D CMAKE_C_COMPILER=gcc \
+          -D CMAKE_CXX_COMPILER=g++ \
+          -D CMAKE_Fortran_COMPILER=gfortran \
+          -D CHARM_ROOT=$CHARM_ROOT \
+          -D MEMORY_ALLOCATOR=SYSTEM \
+          -D CMAKE_BUILD_TYPE=Release \
+          -D BUILD_PYTHON_BINDINGS=ON \
+          -D ENABLE_PARAVIEW=ON \
+          -D MACHINE=Sonic \
+          -D USE_XSIMD=yes \
+          -D DEBUG_SYMBOLS=OFF \
+          "$@" \
+          $SPECTRE_HOME
+}
--- a/support/Machines/Sonic.yaml
+++ b/support/Machines/Sonic.yaml
@@ -1,0 +1,14 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+Machine:
+  Name: Sonic
+  Description: |
+    HPC at ICTS-TIFR hosted by AstroRel.
+    More information:
+    https://it.icts.res.in/docs/sonic-cluster/
+  DefaultTasksPerNode: 3
+  DefaultProcsPerTasks: 32
+  DefaultQueue: "long"
+  DefaultTimeLimit: "1-00:00:00"
+  LaunchCommandSingleNode: ["mpirun", "-n", "1"]
--- a/support/SubmitScripts/Sonic.sh
+++ b/support/SubmitScripts/Sonic.sh
@@ -1,0 +1,17 @@
+{% extends "SubmitTemplateBase.sh" %}
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+# Sonic is an HPC at ICTS-TIFR hosted by its AstroRel group.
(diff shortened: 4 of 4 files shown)


support/Environments/sonic.sh
```diff
@@ -0,0 +1,50 @@
...
+}
+
+spectre_unload_modules() {
+    # The order here is important
+    module unload spectre-env > /dev/null 2>&1
+    # module unload gcc/11.1.0
+    module unload use.own
+}
+
+spectre_run_cmake_gcc_release() {
```
knelli2: To keep this consistent with other clusters, name this `spectre_run_cmake()`. If in the future you add a clang version, then you can add the `_gcc` and `_clang` suffixes to the function names

support/Environments/sonic.sh
```diff
@@ -0,0 +1,50 @@
...
+    if [ -z ${SPECTRE_SOURCE} ]; then
+        echo "You must set SPECTRE_SOURCE to the cloned SpECTRE directory"
```
knelli2: Also to keep this consistent with other clusters, we require that `SPECTRE_HOME` be set

support/Environments/sonic.sh
```diff
@@ -0,0 +1,50 @@
...
+        return 1
+    fi
+
+    if [ -z ${SPECTRE_INSTALL} ]; then
+        echo "You must set SPECTRE_INSTALL to the target directory"
+        return 1
```
knelli2: We typically don't require this to be set. You should just run `spectre_run_cmake` directly in your build directory. Is there a particular reason you'd like this to be set on Sonic?

support/Machines/Sonic.yaml
```diff
@@ -0,0 +1,14 @@
...
+# See LICENSE.txt for details.
+
+Machine:
+  Name: Sonic
+  Description: |
+    HPC at ICTS-TIFR hosted by AstroRel.
+    More information:
+    https://it.icts.res.in/docs/sonic-cluster/
+  DefaultTasksPerNode: 6
+  DefaultProcsPerTasks: 16
```
knelli2: Please make these consistent between the machine file and the submit script.

knelli2 (changes requested):
Please also add Sonic to the `docs/Installation/InstallationOnClusters.md` page with any notes/tips that would be helpful for specifically running on Sonic.

I have no way to test this, so I'll take your word that it works.

support/Environments/sonic.sh
vaishakp: I would really like to keep it consistent. However I noticed the usage of `SPECTRE_ROOT` in some places, and `SPECTRE_HOME`, making it confusing. Further, we run SpEC on Sonic and well and there `SPEC_HOME` is set to the directory where SpEC is built and contains the supporting executables. We therefore decided to keep the environment variables self-explanatory, SPECTRE_SOURCE for the source code, SPECTRE_BUILD for the build directory, and SPECTRE_INSTALL for the install directory.  I also checked that this does not have any bearing on the running of SpECTRE itself, as it is not being used in anywhere in the source code, apart from some examples and documentation.

Please let me know if you have further concerns.

support/Environments/sonic.sh
vaishakp: First of all, thanks for your comments and suggestions. 

> To keep this consistent with other clusters, name this `spectre_run_cmake()`. If in the future you add a clang version, then you can add the `_gcc` and `_clang` suffixes to the function names

Sure

support/Environments/sonic.sh
vaishakp: We would like to keep the build directory separate from the directory from which SpECTRE is run on Sonic. This, e.g. avoids creation of multiple build directories, one for each type of build, cluttering with the build files, etc.

Please let me know if you have further concerns.

support/Machines/Sonic.yaml
vaishakp: Thanks for catching this.

vaishakp:
I should also add that Sonic is a very small HPC at ICTS-TIFR, and we would be a few people using it, atleast for now.

vaishakp:
> Please also add Sonic to the `docs/Installation/InstallationOnClusters.md` page with any notes/tips that would be helpful for specifically running on Sonic.
> 
> I have no way to test this, so I'll take your word that it works.

This is currently being also tested by one of my colleagues at ICTS

support/Environments/sonic.sh
knelli2: Yes you are right that we aren't consistent with `SPECTRE_ROOT` vs `SPECTRE_HOME` throughout the code. We'll have to fix that eventually. And I do understand your reasoning for wanting to use different names. However, we *are* consistent with `SPECTRE_HOME` in all of our cluster environment scripts and I think we should keep the consistency. Otherwise, when adding new clusters, people will be confused why Sonic is different from the others. We want to make the experience of using spectre on different clusters as uniform as possible.

support/Environments/sonic.sh
knelli2: Sorry, there’s something I’m not understanding about the workflow you would like to use and what we are currently doing. Currently our workflow on all clusters is

1.  Clone spectre into `SPECTRE_HOME` and then choose a build directory anywhere else, doesn’t have to be inside the source tree.
2. Do `spectre_run_cmake` inside the build directory, which builds against the `SPECTRE_HOME` source tree and all output goes into the build directory.
3. Then, when we run an executable, we typically copy the executable to the run directory, which can be anywhere and doesn’t have to be inside the build directory (we typically don’t run in the build directory!).

Now, I think we have support for an install directory, but this CMake feature is really geared at end-users who don’t want to develop or change anything in the code. This essentially adds the `make install` step to the build process. For most use cases, I don’t envision this being the right way to go, since inevitably (at least right now) some sort of debugging and code development will need to happen.

docs/Installation/InstallationOnClusters.md
```diff
@@ -116,3 +116,27 @@ The second command with `SPECTRE_DEBUG=ON` enables sanity checks and
 optimizations. This means it can be used in production-level runs to ensure
 there aren't any subtle bugs that might only arise after a decently
 long simulation.
+
+## Sonic at ICTS-TIFR
+
+Cluster documentation: https://it.icts.res.in/docs/sonic-cluster
+
+1. Clone the spectre source code and set `SPECTRE_SOURCE` to point it.
```
knelli2: Please update this list with the workflow I outlined below

docs/Installation/InstallationOnClusters.md
```diff
@@ -116,3 +116,27 @@ The second command with `SPECTRE_DEBUG=ON` enables sanity checks and
...
+2. Create a build directory and set `SPECTRE_BUILD` to it.
+3. Set `SPECTRE_INSTALL` to the destination location.
+4. Setup the software environment
+    1. Set the location of optimized libraries and load environment modules:
+    ```
+    export SOFT_DIR=/mnt/pfs/vaishak.p/soft
+    . $SOFT_DIR/modules-5.2.0/init/bash
```
knelli2: Could this just go in the environment script in the `spectre_load_modules` function? Then there's less that has to go in people's `.bashrc` file

docs/Installation/InstallationOnClusters.md
```diff
@@ -116,3 +116,27 @@ The second command with `SPECTRE_DEBUG=ON` enables sanity checks and
...
+    ```
+    Add these lines to your `.bashrc` and source it.
+    2. Copy the `privatemodules` dir from `$SOFT_DIR` to your home directory
```
knelli2: Is this really necessary? Why can't people point into your `privatemodules` directory?

docs/Installation/InstallationOnClusters.md
```diff
@@ -116,3 +116,27 @@ The second command with `SPECTRE_DEBUG=ON` enables sanity checks and
...
+5. Source the `$SPECTRE_SOURCE/support/Environments/Sonic.sh` script.
+6. Build SpECTRE by changing to `$SPECTRE_BUILD` and using one of the commands
```
knelli2: There's only one command

knelli2 (changes requested):

support/Environments/sonic.sh
vaishakp: > Sorry, there’s something I’m not understanding about the workflow you would like to use and what we are currently doing. Currently our workflow on all clusters is
> 
> 1. Clone spectre into `SPECTRE_HOME` and then choose a build directory anywhere else, doesn’t have to be inside the source tree.
> 2. Do `spectre_run_cmake` inside the build directory, which builds against the `SPECTRE_HOME` source tree and all output goes into the build directory.
> 3. Then, when we run an executable, we typically copy the executable to the run directory, which can be anywhere and doesn’t have to be inside the build directory (we typically don’t run in the build directory!).
> 
> Now, I think we have support for an install directory, but this CMake feature is really geared at end-users who don’t want to develop or change anything in the code. This essentially adds the `make install` step to the build process. For most use cases, I don’t envision this being the right way to go, since inevitably (at least right now) some sort of debugging and code development will need to happen.

As I understand, this is this default workflow when submitting jobs to queues. Does SpECTRE also copy executables to another location auomatically when we are testing code by directing calling the executables, say e.g. in a jupyter notebook? This can be seen e.g. in [example notebooks](https://github.com/sxs-collaboration/spectre/blob/develop/docs/Examples/BbhInitialData/BbhInitialData.ipynb) in SpECTRE. I find myself doing this for debugging using pythonic interface.

support/Environments/sonic.sh
vaishakp: > Yes you are right that we aren't consistent with `SPECTRE_ROOT` vs `SPECTRE_HOME` throughout the code. We'll have to fix that eventually. And I do understand your reasoning for wanting to use different names. However, we _are_ consistent with `SPECTRE_HOME` in all of our cluster environment scripts and I think we should keep the consistency. Otherwise, when adding new clusters, people will be confused why Sonic is different from the others. We want to make the experience of using spectre on different clusters as uniform as possible.

Just trying to adopt good build practices. If SpECTRE code guidelines require this, we will be happy to adhere to this. I am sorry for assuming that we had liberty to choose how SpECTRE is setup and run on our cluster. It would be nice to add this somewhere in guidelines in case I have not missed it.

docs/Installation/InstallationOnClusters.md
vaishakp: Thanks for the question. 
This has been thought over. It was decided not to push these hard links to SpECTRE repo as we wouldn't have to push commits to SpECTRE if we decide to change the location of the softwares.

docs/Installation/InstallationOnClusters.md
vaishakp: Sure they can point to it. This was done because currently we are experiencing latency in loading modules, possibly resolving paths in the symlinks.

knelli2 (approved):
Looks great! Thanks for adding this

PR #6315, opened 2024-09-30
Enable BBH ID control of CoM and Padm.

iago-mendes:
## Proposed changes

With recent changes to the calculation of the center of mass and linear momentum integrals, we are now ready to enable their control by default. We also take this opportunity to set the control residual tolerance to `1.e-4` instead of `1.e-6`, which makes more sense given that junk radiation effects are on the order of `1.e-3`.

For an equal-mass non-spinning case with default settings, the control loop currently reaches the residual tolerance in just 3 iterations. In human time, this takes ~17 minutes on 1 node of CaltechHPC.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/ControlId.py
+++ b/support/Pipelines/Bbh/ControlId.py
@@ -15,7 +15,9 @@ from spectre.Visualization.ReadH5 import to_dataframe
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_RESIDUAL_TOLERANCE = 1.0e-6
+# The default tolerance is just below the effect of junk radiation on the
+# controlled paramaters, which  is about 1.0e-3.
+DEFAULT_RESIDUAL_TOLERANCE = 1.0e-4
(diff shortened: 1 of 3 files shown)


support/Pipelines/Bbh/ControlId.py
```diff
@@ -15,7 +15,7 @@
 
 logger = logging.getLogger(__name__)
 
-DEFAULT_RESIDUAL_TOLERANCE = 1.0e-6
+DEFAULT_RESIDUAL_TOLERANCE = 1.0e-4
```
nilsvu, line 20: Please add a comment here that this should be just below junk radiation effects on the controlled parameters

nilsvu (changes requested):

support/Pipelines/Bbh/ControlId.py
iago-mendes: Done!

nilsvu (approved):

PR #6329, opened 2024-10-10
Conform BNS ID domain to star surfaces

nilsvu:
## Proposed changes

Allows to deform the BNS initial data domain so a block boundary conforms to the star surfaces:

domain-5

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp
@@ -14,12 +14,10 @@ namespace domain::CoordinateMaps::ShapeMapTransitionFunctions {
 
 /*!
  * \ingroup CoordMapsTimeDependentGroup
- * \brief A transition function that falls off as $f(r) = g(r) / r$ where $g(r)
- * = ar + b$.
+ * \brief A transition function that falls off as $f(r) = ar + b$.
  *
- * \details The coefficients $a$ and $b$ are chosen so that the function $g(r) =
- * ar + b$ falls off linearly from 1 at `r_min` to 0 at `r_max`. This means that
- * $f(r)$ falls off from $1/r_{\text{min}}$ at `r_min` to 0 at `r_max`. The
+ * \details The coefficients $a$ and $b$ are chosen so that the function $f(r) =
+ * ar + b$ falls off linearly from 1 at `r_min` to 0 at `r_max`. The
  * coefficients are
  *
  * \f{align}{
--- a/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp
@@ -25,11 +25,15 @@ namespace domain::CoordinateMaps::ShapeMapTransitionFunctions {
  * b &= \frac{r_{\text{max}}}{r_{\text{max}} - r_{\text{min}}} = -a
  * r_{\text{max}}
  * \f}
+ *
+ * If the `reverse` flag is set to `true`, then the function falls off from 0 at
+ * `r_min` to 1 at `r_max`. To do this, the coefficients are modified as
+ * $a \rightarrow -a$ and $b \rightarrow 1-b$.
  */
 class SphereTransition final : public ShapeMapTransitionFunction {
  public:
   explicit SphereTransition() = default;
-  SphereTransition(double r_min, double r_max);
+  SphereTransition(double r_min, double r_max, bool reverse = false);
 
   double operator()(const std::array<double, 3>& source_coords) const override;
   DataVector operator()(
--- a/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp
@@ -41,7 +41,7 @@ class SphereTransition final : public ShapeMapTransitionFunction {
 
   std::optional<double> original_radius_over_radius(
       const std::array<double, 3>& target_coords,
-      double distorted_radius) const override;
+      double radial_distortion) const override;
 
   std::array<double, 3> gradient(
       const std::array<double, 3>& source_coords) const override;
--- a/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
@@ -62,11 +62,12 @@ Wedge::Wedge(const std::array<double, 3>& inner_center,
              const double inner_radius, const double inner_sphericity,
              const std::array<double, 3>& outer_center,
              const double outer_radius, const double outer_sphericity,
-             const Axis axis)
+             const Axis axis, const bool reverse)
     : inner_surface_(inner_center, inner_radius, inner_sphericity),
       outer_surface_(outer_center, outer_radius, outer_sphericity),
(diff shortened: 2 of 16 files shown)


src/ParallelAlgorithms/SurfaceFinder/Python/FindRadialSurface.py
```diff
@@ -169,6 +169,7 @@ def find_radial_surface(
                         functions_of_time,
                     )
                 )[:, 0]
+                - initial_guess.expansion_center
```
knelli2, line 172: Update the test to be off-center to catch this?

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp
```diff
@@ -29,7 +29,7 @@ namespace domain::CoordinateMaps::ShapeMapTransitionFunctions {
 class SphereTransition final : public ShapeMapTransitionFunction {
  public:
   explicit SphereTransition() = default;
-  SphereTransition(double r_min, double r_max);
+  SphereTransition(double r_min, double r_max, bool reverse = false);
```
knelli2, line 36: Add docs above about what reverse does and the new coefficient values for $a$ and $b$

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -376,7 +382,8 @@ std::optional<double> Wedge::original_radius_over_radius(
   // can check the outermost distance because its surface doesn't move. We can't
   // check the innermost surface because the inner bound is the origin which we
   // already checked above.
-  if (centered_coords_magnitude - eps_ > outer_distance) {
+  if ((not reverse_ and (centered_coords_magnitude > outer_distance + eps_)) or
```
knelli2: Update the comment

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -393,19 +400,24 @@ std::optional<double> Wedge::original_radius_over_radius(
   // so the map is again the identity so the radius and original radius are
   // equal. We can't check the overall inner distance because that has been
   // distorted so we don't know where the mapped inner distance is.
-  if (equal_within_roundoff(centered_coords_magnitude, outer_distance)) {
+  if ((not reverse_ and
```
knelli2: Update the comment

src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.hpp
```diff
@@ -316,7 +316,8 @@ struct TimeDependentMapOptions {
           object_A_radii,
       const std::optional<std::array<double, IsCylindrical ? 2 : 3>>&
           object_B_radii,
-      double envelope_radius, double domain_outer_radius);
+      bool object_A_filled, bool object_B_filled, double envelope_radius,
+      double domain_outer_radius);
```
knelli2, line 331: Update dox for what the filled parameters will do

src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
```diff
@@ -329,102 +330,113 @@ void TimeDependentMapOptions<IsCylindrical>::build_maps(
...
+      const std::optional<std::array<double, 3>>& cube_center =
+          i == 0 ? cube_A_center : cube_B_center;
+      const std::array<double, 3>& inner_center = gsl::at(object_centers, i);
+      const std::array<double, 3>& outer_center =
+          transition_ends_at_cube
+              ? cube_center.value_or(gsl::at(object_centers, i))
+              : gsl::at(object_centers, i);
+
+      const double inner_sphericity = 1.0;
+      const double outer_sphericity = transition_ends_at_cube ? 0.0 : 1.0;
```
knelli2, line 413: Can you add a comment here and/or below by the `inner_radius` about how if `filled` is True, then these values are only used for the outer shell of wedges and different values will be used for the inner wedges? I think this will help in the future when we (I) look at this again

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -417,8 +417,8 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
         std::array{rotate_from_z_to_x_axis(center_A_),
                    rotate_from_z_to_x_axis(center_B_)},
         std::nullopt, std::nullopt, std::array{radius_A_, outer_radius_A_},
-        std::array{radius_B_, outer_radius_B_}, inner_common_radius,
-        outer_radius_);
+        std::array{radius_B_, outer_radius_B_}, include_inner_sphere_A,
+        include_inner_sphere_B, inner_common_radius, outer_radius_);
```
knelli2: I think `include_inner_sphere_{A/B}` should be replaced with just `False` here, right? We don't allow central object cubes in the cylindrical domain. This inner sphere is just another set of wedges, but isn't filled

src/Domain/Creators/CylindricalBinaryCompactObject.cpp
```diff
@@ -417,8 +417,8 @@ CylindricalBinaryCompactObject::CylindricalBinaryCompactObject(
         std::array{rotate_from_z_to_x_axis(center_A_),
                    rotate_from_z_to_x_axis(center_B_)},
```
knelli2, line 418: Can't reach it with GitHub comments, but I believe there's a bug above when checking if both inner spheres and the outer sphere are specified. I think it should be

```cpp
  if (time_dependent_options_.has_value() and
      not(include_inner_sphere_A and include_inner_sphere_B
                                 and include_outer_sphere)) {
```

tests/Unit/Domain/Creators/TimeDependentOptions/Test_BinaryCompactObject.cpp
```diff
@@ -290,129 +294,127 @@ void test(const bool include_expansion, const bool include_rotation,
...
+                            Catch::Matchers::ContainsSubstring(
+                                "If the object is filled, the transition must "
+                                "end at the cube."));
+          continue;
+        }
+        if (include_shape_a and not excise_A) {
+          CHECK_THROWS_WITH(
+              build_maps(),
+              Catch::Matchers::ContainsSubstring(
+                  "the centers of the inner and outer object are different"));
```
knelli2, line 339: Where is this error from? I wasn't able to find it in the code base or this PR

knelli2 (changes requested):

src/ParallelAlgorithms/SurfaceFinder/Python/FindRadialSurface.py
nilsvu: Didn't do it because the `Sphere` domain creator can't do off-center coords at the moment (would be nice though)

tests/Unit/Domain/Creators/TimeDependentOptions/Test_BinaryCompactObject.cpp
nilsvu: `src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp:75`

knelli2: Ugh that line break messed up my search.

Should this be caught one level up in the TimeDepBCO? Seems a bit odd to be getting an error from the wedge transition here

nilsvu: Dunno, the logic in these tests is pretty complicated and took a while to work out because there are so many different configurations of the BCO domain so I don't feel particularly motivated to move things around

knelli2 (approved):
Squash

PR #6330, opened 2024-10-10
Update installation instructions

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupYamlCpp.cmake
+++ b/cmake/SetupYamlCpp.cmake
@@ -12,7 +12,7 @@ if (NOT yaml-cpp_FOUND)
   include(FetchContent)
   FetchContent_Declare(yaml-cpp
     GIT_REPOSITORY https://github.com/jbeder/yaml-cpp.git
-    GIT_TAG yaml-cpp-0.7.0
+    GIT_TAG yaml-cpp-0.8.0
     GIT_SHALLOW TRUE
     ${SPECTRE_FETCHCONTENT_BASE_ARGS}
   )
--- a/docs/Installation/Installation.md
+++ b/docs/Installation/Installation.md
@@ -6,19 +6,146 @@ See LICENSE.txt for details.
 
 \tableofcontents
 
-This page details the installation procedure for SpECTRE on personal computers
-using x86-64 processors. For configuring SpECTRE, please refer to the
-\subpage spectre_build_system page. For instructions on installing SpECTRE on
-clusters please refer to the \subpage installation_on_clusters page. For
-instructions on installing SpECTRE on Apple Silicon Macs, please refer to the
-\subpage installation_on_apple_silicon page. Refer to the
-\subpage versioning_and_releases page for information on specific versions
-to install.
+This page details how to install SpECTRE on personal machines and on clusters
+that have no official support (yet).
 
-### Install a specific version with Spack
+- For details on installing SpECTRE on a number of clusters that we support
+  please refer to:
+  \subpage installation_on_clusters
+- For configuring SpECTRE please refer to:
+  \subpage spectre_build_system
+- For instructions on installing SpECTRE on Apple Silicon Macs please refer to:
+  \subpage installation_on_apple_silicon
+- For information on our versioning scheme and public releases please refer to:
+  \subpage versioning_and_releases
 
-You can install SpECTRE with the [Spack](https://github.com/spack/spack) package
-manager:
+### Running containerized releases
+
+A quick way to run the code without installing anything at all is with our
+containerized releases:
+
+```
+docker run sxscollaboration/spectre --help
+```
+
+You can also use [Apptainer/Singularity](https://apptainer.org) instead of
+Docker, which works better on computing clusters and is more convenient because
+it shares the host's file system:
+
+```
+apptainer run docker://sxscollaboration/spectre --help
+```
+
+The entrypoint to this container is the SpECTRE
+\ref tutorial_cli "command-line interface (CLI)".
+For example, you can generate initial data for a simulation of merging black
+holes and plot the result like this:
+
+```
+apptainer run docker://sxscollaboration/spectre bbh generate-id \
+  -q 1 --chi-A 0 0 0 --chi-B 0 0 0 -D 16 -w 0.015 -a 0 -o ./bbh_id
+apptainer run docker://sxscollaboration/spectre plot slice \
+  bbh_id/BbhVolume*.h5 -C 0,0,0 -n 0,0,1 -u 0,1,0 -X 24 24 \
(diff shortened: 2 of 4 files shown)


cmake/SetupYamlCpp.cmake
```diff
@@ -12,7 +12,7 @@ if (NOT yaml-cpp_FOUND)
   include(FetchContent)
   FetchContent_Declare(yaml-cpp
     GIT_REPOSITORY https://github.com/jbeder/yaml-cpp.git
-    GIT_TAG yaml-cpp-0.7.0
+    GIT_TAG yaml-cpp-0.8.0
```
knelli2, line 15: Update installation docs for this as well. Also I think the container version needs be updated as well

docs/Installation/Installation.md
```diff
@@ -6,19 +6,121 @@ See LICENSE.txt for details.
 
 \tableofcontents
 
-This page details the installation procedure for SpECTRE on personal computers
-using x86-64 processors. For configuring SpECTRE, please refer to the
```
knelli2, line 10: I think we should keep a blurb of some kind here at the top before going in to details. The import links I think that should be highlighted are

- installation on clusters
- versioning and releases

docs/Installation/Installation.md
```diff
@@ -6,19 +6,121 @@ See LICENSE.txt for details.
...
+
+```
+apptainer run docker://sxscollaboration/spectre bbh generate-id \
+  -q 1 --chi-A 0 0 0 --chi-B 0 0 0 -D 16 -w 0.015 -a 0 -o ./bbh_id
+apptainer run docker://sxscollaboration/spectre plot slice \
+  bbh_id/BbhVolume*.h5 -C 0,0,0 -n 0,0,1 -u 0,1,0 -X 24 24 \
+  -y ConformalFactor -o plot.pdf
+```
+
+The containers currently have precompiled code only for Linux x86 platforms.
```
knelli2: Make this a `\note` so it's more prominent, and then also link to the Apply Silicon instructions so mac users have somewhere to look

docs/Installation/Installation.md
```diff
@@ -6,19 +6,121 @@ See LICENSE.txt for details.
...
-\subpage spectre_build_system page. For instructions on installing SpECTRE on
-clusters please refer to the \subpage installation_on_clusters page. For
-instructions on installing SpECTRE on Apple Silicon Macs, please refer to the
-\subpage installation_on_apple_silicon page. Refer to the
-\subpage versioning_and_releases page for information on specific versions
-to install.
+### Running containerized releases
+
+A quick way to run the code without installing anything at all is with our
+containerized releases:
```
knelli2, line 25: I think we need to emphasize here that the entry point to this container is the CLI (and link to the docs) and that only very specific executables are built, namely SolveXcts, BBH, SingleBH. We don't want to give people the impression that they can just download this container and run anything from spectre.

docs/Installation/Installation.md
```diff
@@ -6,19 +6,121 @@ See LICENSE.txt for details.
...
+```
+apptainer run docker://sxscollaboration/spectre --help
+```
+
+For example, you can generate initial data for a simulation of merging black
+holes and plot the result like this:
+
+```
+apptainer run docker://sxscollaboration/spectre bbh generate-id \
+  -q 1 --chi-A 0 0 0 --chi-B 0 0 0 -D 16 -w 0.015 -a 0 -o ./bbh_id
```
knelli2, line 46: [optional] I'd advise against putting these examples in since we may change arguments/names/anything at some point, and then forget to update this, and then the very first easy thing we tell people to do won't work.

docs/Installation/Installation.md
```diff
@@ -6,19 +6,121 @@ See LICENSE.txt for details.
...
+  See \ref building-spectre for details and \ref common_cmake_flags for a list
+  of possible configuration options. For example, set `-D ENABLE_OPENMP=ON`
+  to enable OpenMP-parallelization for the exporter library.
+
+  Now you can compile executables (again, see \ref building-spectre for
+  details). For example:
+
+  ```sh
+  make -j12 cli
+  make -j12 BundledExporter
```
knelli2, line 142: Probably do `-j4` just to lessen the likelihood that people will run out of memory

docs/Installation/Installation.md
```diff
@@ -6,19 +6,121 @@ See LICENSE.txt for details.
...
+  ```
+
+### Installation with Spack
+
+You can also install SpECTRE with the [Spack](https://github.com/spack/spack)
+package manager:
```
knelli2, line 148: Do we want to warn people that this may not work for a number of reasons? (local env, spack upstream, etc.)

knelli2 (changes requested):
You can squash these in directly

docs/Installation/Installation.md
wthrowe: Drive-by comment, no need for further input from me: To many Linux users "x86" suggests 32-bit.  Using "x86_64" is unambiguous.

cmake/SetupYamlCpp.cmake
nilsvu: No v0.7.0 is still fine, this just bumps the version that is fetched automatically.

docs/Installation/Installation.md
nilsvu: Currently only CLI, CCE, and SolveXcts are built

docs/Installation/Installation.md
nilsvu: I thought it would be nice to give an example that actually does something, but I agree with your concern. I'll revisit this when I put the initial data tutorial online and maybe just link to it.

cmake/SetupYamlCpp.cmake
knelli2: Oh right, my bad

docs/Installation/Installation.md
```diff
@@ -6,19 +6,142 @@ See LICENSE.txt for details.
...
+
+The containers currently have precompiled code only for Linux x86_64 platforms,
+and only have a limited set of executables precompiled. The supported features
+available in the precompiled containers are:
+
+- Generating initial data
+- Running CCE (see \ref tutorial_cce)
+- Running Python support code with the SpECTRE CLI (see \ref tutorial_cli)
+
+### Running precompiled executables
```
knelli2: Given the wording in the previous section, this title is a bit confusing since this section is completely unrelated to the docker container. Maybe something like "Running with static binaries" ?

knelli2 (changes requested):
Looks good, just one last thing. You can squash everything together

knelli2 (approved):

PR #6333, opened 2024-10-15
Use SpEC's eccentricity control

nilsvu:
## Proposed changes

Call into SpEC to get eccentricity reduction updates. Replaces the port of an old SpEC script.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
@sxs-collaboration/spectre-core-devs could one of you take this one?

knelli2 (commented):
[not a full review] Seems like some modules need to be fixed. Am I understanding this correctly that we have to have SpEC installed in order to do eccentricity control currently? Should we add an error/warning if somebody tries to use it but their build wasn't configured with SpEC?

nilsvu:
Yes that's correct. I'll add some more checks for that (edit: done). For independence from SpEC we'll have to port the code over at some point.

support/Pipelines/EccentricityControl/EccentricityControlParams.py
```diff
@@ -0,0 +1,298 @@
...
+    target_eccentricity: float = 0.0,
+    target_mean_anomaly_fraction: Optional[float] = None,
+    plot_output_dir: Optional[Union[str, Path]] = None,
+) -> Tuple[float, float, Dict[OrbitalParams, float]]:
+    """Get new orbital parameters for a binary system to control eccentricity.
+
+    The eccentricity is estimated from the trajectories of the binary objects
+    and updates to the orbital parameters are suggested to drive the orbit to
+    the target eccentricity, using SpEC's OmegaDotEccRemoval.py. Currently
+    supports only circular target orbits (target eccentricity = 0).
```
knelli2, line 51: Add a check that the eccentricity is in fact 0 until we add support for non-zero ecc

support/Pipelines/Bbh/EccentricityControl.py
```diff
@@ -2,118 +2,71 @@
...
-def eccentricity_control(
-    h5_file,
-    id_input_file_path,
-    subfile_name_aha="ApparentHorizons/ControlSystemAhA_Centers.dat",
-    subfile_name_ahb="ApparentHorizons/ControlSystemAhB_Centers.dat",
-    tmin=500,
-    tmax=None,  # use all available data
-    output=None,  # reads from Inspiral.yaml
-):
+def eccentricity_control(h5_files, id_input_file_path, **kwargs):
```
knelli2: Type the arguments

support/Pipelines/EccentricityControl/EccentricityControlParams.py
```diff
@@ -0,0 +1,298 @@
...
+from spectre.Visualization.PlotTrajectories import import_A_and_B
+from spectre.Visualization.ReadH5 import to_dataframe
+
+logger = logging.getLogger(__name__)
+
+# Input orbital parameters that can be controlled
+OrbitalParams = Literal[
+    "Omega0",
+    "adot0",
+    "D0",
```
knelli2, line 26: Do we want to take this opportunity and move away from the names in SpEC to better names? Like `initial_angular_velocity`, `initial_radial_velocity`, `initial_separation`

support/Pipelines/EccentricityControl/EccentricityControlParams.py
```diff
@@ -0,0 +1,298 @@
...
+            sB = [
+                horizon_params[f"AhB DimensionfulSpinVector_{xyz}"]
+                for xyz in "xyz"
+            ]
+        else:
+            logger.warning("No horizon spins found in data, ignoring spins.")
+            sA = sB = None
+
+    # Call into SpEC's OmegaDotEccRemoval.py
+    t, Omega, dOmegadt, OmegaVec = ComputeOmegaAndDerivsFromFile(traj_A, traj_B)
```
knelli2: Can you also add a comment that despite the function name, this expects arrays of data, and not filenames

support/Pipelines/EccentricityControl/EccentricityControlParams.py
```diff
@@ -0,0 +1,298 @@
...
+DEFAULT_AHB_TRAJECTORIES = "ApparentHorizons/ControlSystemAhB_Centers.dat"
+
+
+def eccentricity_control_params(
+    h5_files: Union[Union[str, Path], Sequence[Union[str, Path]]],
+    id_input_file_path: Union[str, Path],
+    subfile_name_aha_trajectories: str = DEFAULT_AHA_TRAJECTORIES,
+    subfile_name_ahb_trajectories: str = DEFAULT_AHB_TRAJECTORIES,
+    subfile_name_aha_quantities: str = "ObservationAhA.dat",
+    subfile_name_ahb_quantities: str = "ObservationAhB.dat",
```
knelli2, line 39: Why not factor these defaults out as well?

tests/support/Pipelines/MockBinaryData.py
```diff
@@ -0,0 +1,54 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import numpy as np
+
+import spectre.IO.H5 as spectre_h5
+
+
+def write_mock_trajectory_data(
```
knelli2: We already have a class `tests/Unit/Helpers/PointwiseFunctions/PostNewtonian/BinaryTrajectories.hpp` that can compute low order PN (and Newtonian) trajectories that we use in control system testing. Maybe we should move it to a common location, extend it a bit, and bind it in python to use here (that way we don't rewrite things)

knelli2 (changes requested):

support/Pipelines/EccentricityControl/EccentricityControlParams.py
nilsvu: These names are actually used a lot, so I tend towards keeping them.

support/Pipelines/EccentricityControl/EccentricityControlParams.py
nilsvu: I only factored out the defaults because the lines were too long

tests/support/Pipelines/MockBinaryData.py
nilsvu: Done in #6366

support/Pipelines/EccentricityControl/EccentricityControlParams.py
```diff
@@ -94,6 +94,9 @@ def eccentricity_control_params(
             "'-D SPEC_ROOT' to a SpEC installation when configuring the build "
             "with CMake."
         )
+    assert (
+        target_eccentricity == 0.0
+    ), "Only circular orbits are supported for eccentricity control yet."
```
knelli2: `yet` -> `currently`

support/Pipelines/EccentricityControl/EccentricityControlParams.py
knelli2: Ok I'm fine with this, but we should be aware that if we want to change the names it's better to do it earlier rather than later while we're still building everything. We'll be even more resistant to change in the future once this is being used

knelli2 (changes requested):
You can squash in this last change

knelli2 (approved):

PR #6334, opened 2024-10-15
Allow multiple callbacks on same component

knelli2:
## Proposed changes

It is possible for multiple unique callbacks to be registered to the same parallel component (size control system does this). Previously, we only took the first callback and discarded the rest (because the elements would register multiple identical callbacks and we only needed one). Now we still discard identical callbacks, but keep unique callbacks on the same component.

This was found while debugging the nodegroup implementation of our algorithm, but I believe this is also a bug on develop. I guess we just have been lucky and haven't hit this so far.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/Callback.hpp
+++ b/src/Parallel/Callback.hpp
@@ -6,6 +6,7 @@
 
 #pragma once
 
+#include <memory>
 #include <pup.h>
 #include <tuple>
 #include <utility>
@@ -64,6 +65,7 @@ class Callback : public PUP::able {
    */
   virtual bool is_equal_to(const Callback& rhs) const = 0;
   virtual std::string name() const = 0;
+  virtual std::unique_ptr<Callback> get_clone() = 0;
 };
 
 /// Wraps a call to a simple action and its arguments.
@@ -115,6 +117,11 @@ class SimpleActionCallback : public Callback {
            "," + pretty_type::name<Proxy>() + ")";
   }
 
+  std::unique_ptr<Callback> get_clone() override {
+    return std::make_unique<SimpleActionCallback<SimpleAction, Proxy, Args...>>(
+        *this);
+  }
+
  private:
   std::decay_t<Proxy> proxy_{};
   std::tuple<std::decay_t<Args>...> args_{};
@@ -154,6 +161,10 @@ class SimpleActionCallback<SimpleAction, Proxy> : public Callback {
     // likely be really long with the template parameters which is unnecessary
     return "SimpleActionCallback(" + pretty_type::get_name<SimpleAction>() +
            "," + pretty_type::name<Proxy>() + ")";
+  }
+
+  std::unique_ptr<Callback> get_clone() override {
+    return std::make_unique<SimpleActionCallback<SimpleAction, Proxy>>(*this);
   }
 
  private:
@@ -210,6 +221,11 @@ class ThreadedActionCallback : public Callback {
            "," + pretty_type::name<Proxy>() + ")";
   }
 
+  std::unique_ptr<Callback> get_clone() override {
+    return std::make_unique<
+        ThreadedActionCallback<ThreadedAction, Proxy, Args...>>(*this);
+  }
+
  private:
   std::decay_t<Proxy> proxy_{};
   std::tuple<std::decay_t<Args>...> args_{};
@@ -252,6 +268,11 @@ class ThreadedActionCallback<ThreadedAction, Proxy> : public Callback {
            "," + pretty_type::name<Proxy>() + ")";
   }
 
+  std::unique_ptr<Callback> get_clone() override {
+    return std::make_unique<ThreadedActionCallback<ThreadedAction, Proxy>>(
+        *this);
+  }
+
  private:
   std::decay_t<Proxy> proxy_{};
 };
@@ -288,6 +309,10 @@ class PerformAlgorithmCallback : public Callback {
     // Only use pretty_type::name for proxy because it'll likely be really long
     // with the template parameters which is unnecessary
     return "PerformAlgorithmCallback(" + pretty_type::name<Proxy>() + ")";
+  }
+
+  std::unique_ptr<Callback> get_clone() override {
+    return std::make_unique<PerformAlgorithmCallback<Proxy>>(*this);
   }
 
  private:
--- a/src/Parallel/Callback.hpp
+++ b/src/Parallel/Callback.hpp
@@ -11,10 +11,39 @@
 #include <utility>
 
 #include "Parallel/Invoke.hpp"
+#include "Utilities/PrettyType.hpp"
 #include "Utilities/Serialization/CharmPupable.hpp"
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
+#include "Utilities/TypeTraits/HasEquivalence.hpp"
 
 namespace Parallel {
+namespace detail {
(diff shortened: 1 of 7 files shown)


src/Parallel/Callback.hpp
```diff
@@ -30,6 +60,16 @@ class Callback : public PUP::able {
   explicit Callback(CkMigrateMessage* msg) : PUP::able(msg) {}
   virtual void invoke() = 0;
   virtual void register_with_charm() = 0;
+  /*!
+   * \brief Returns if this callback is equal to the one passed in.
+   *
+   * \details If the `std::optional` doesn't have a value, then the pointer
+   * cast between the two callbacks failed. If it does have a value, then the
```
wthrowe: There's some sort of context missing for this documentation.  There doesn't seem to be any mention of pointers or casts anywhere else.

src/Parallel/Callback.hpp
```diff
@@ -65,6 +105,22 @@ class SimpleActionCallback : public Callback {
     register_classes_with_charm<SimpleActionCallback>();
   }
 
+  std::optional<bool> is_equal_to(const Callback& rhs) const override {
+    const auto* downcast_ptr = dynamic_cast<const SimpleActionCallback*>(&rhs);
+    if (downcast_ptr == nullptr) {
+      return std::nullopt;
```
wthrowe: Why this distinction instead of just returning false?

tests/Unit/Parallel/Test_Callback.cpp
```diff
@@ -267,6 +285,10 @@ struct RunCallbacks {
     callbacks.emplace_back(
         std::make_unique<Parallel::SimpleActionCallback<
             MultiplyValueByFactor, decltype(proxy_2), double>>(proxy_2, 2.0));
+    const std::optional<bool> callback_2_equal_to_vec_end =
+        callback_2.is_equal_to(*callbacks.back());
+    SPECTRE_PARALLEL_REQUIRE(callback_2_equal_to_vec_end.has_value());
+    SPECTRE_PARALLEL_REQUIRE_FALSE(callback_2_equal_to_vec_end.value());
```
wthrowe: [optional] `callback_2_equal_to_vec_end == std::optional{false}` seems clearer to me than these two lines.

src/Parallel/Callback.hpp
```diff
@@ -6,15 +6,45 @@
...
 #include "Parallel/Invoke.hpp"
+#include "Utilities/PrettyType.hpp"
 #include "Utilities/Serialization/CharmPupable.hpp"
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
+#include "Utilities/TypeTraits/HasEquivalence.hpp"
 
 namespace Parallel {
+namespace detail {
+// Not all tuple arguments are guaranteed to have operator==, so we check the
+// ones we can.
```
wthrowe, line 23: I'm a bit worried about silently ignoring entries.  What types typically occur that cause problems?

wthrowe (changes requested):

src/Parallel/Callback.hpp
knelli2: The problematic types were proxies and I didn't want to deal with how to evaluate equivalence of those

src/Parallel/Callback.hpp
knelli2: It was particularly useful in debugging to know if the callbacks were different types, or if they were the same but had different arguments. At some point, I would like to put debugging info/prints into the cache and this kind of info would be useful, but that's beyond this PR so I just left the return types as `optional<bool>` for now

wthrowe: Just return a bool.  If you want to know if the types are the same for debugging you can check that directly without complicating the public interface.

src/Parallel/Callback.hpp
wthrowe: OK.  https://github.com/charmplusplus/charm/pull/3848 is probably the real fix, for what it's worth, but we can't wait for that.

wthrowe:
Looks good.  Squash.

knelli2:
@wthrowe I realized I inadvertently added a bug related to removing an unnecessary callback if the cache item was mutated and is now ready. I added a new commit and a new fixup, so let me know if you need more of a explanation, or if the code is enough.

src/Parallel/GlobalCache.hpp
```diff
@@ -554,7 +556,25 @@ bool GlobalCache<Metavariables>::mutable_cache_item_is_ready(
                          std::vector<std::unique_ptr<Callback>>>& callbacks =
           std::get<1>(tuples::get<tag>(mutable_global_cache_));
 
-      callbacks.erase(array_component_id);
+      // It's possible that no new callbacks were registered, so make sure this
+      // array component id still has callbacks before trying to remove them.
+      if (callbacks.contains(array_component_id)) {
+        if (callbacks.at(array_component_id).empty()) {
+          callbacks.erase(array_component_id);
```
wthrowe, line 572: Check if the vector is empty after removing the element, instead of before.

src/Parallel/Callback.hpp
```diff
@@ -115,6 +119,11 @@ class SimpleActionCallback : public Callback {
            "," + pretty_type::name<Proxy>() + ")";
   }
 
+  std::unique_ptr<Callback> get_clone() override {
+    return std::make_unique<SimpleActionCallback<SimpleAction, Proxy, Args...>>(
+        proxy_, args_);
```
wthrowe: [optional] Using the copy constructor is simpler.  (Also several other places.)

wthrowe (changes requested):
Good catch.  Test isn't compiling, though.

src/Parallel/GlobalCache.hpp
```diff
@@ -554,7 +556,25 @@ bool GlobalCache<Metavariables>::mutable_cache_item_is_ready(
...
+        } else {
+          // If this callback was a duplicate, we'll have to search through all
+          // callbacks to determine which to remove. If it wasn't a dupliate,
```
nilsdeppe: `dupliate` -> `duplicate`

src/Parallel/GlobalCache.hpp
```diff
@@ -554,7 +556,25 @@ bool GlobalCache<Metavariables>::mutable_cache_item_is_ready(
...
+          // then it'll just be the last callback in the vector.
+          auto& vec_callbacks = callbacks.at(array_component_id);
+          auto it_to_erase = vec_callbacks.begin();
+          for (; it_to_erase != vec_callbacks.end(); it_to_erase++) {
+            if ((*it_to_erase)->is_equal_to(*clone_of_optional_callback)) {
+              vec_callbacks.erase(it_to_erase);
+              break;
+            }
+          }
```
nilsdeppe: I think this can be done with
```cpp
vec_callbacks.erase(std::find(vec_callbacks.begin(), vec_callbacks.end(),
[&clone_of_optional_callback](const auto& t) {
  return t.is_equal_to(*clone_of_optional_callback);
});
```

nilsdeppe (changes requested):
Just a couple small things you can squash with everything else.

wthrowe:
Looks good.  Squash.

wthrowe (approved):

nilsdeppe (approved):

PR #6336, opened 2024-10-16
Add PiecewisePolynomial, QuaternionFunctionofTime, and Serialize Pybindings

AlexCarpenter46:
## Proposed changes

Adds some pybindings necessary for creating volume data with functions of time in python.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Python/FunctionsOfTime.cpp
+++ b/src/Domain/Python/FunctionsOfTime.cpp
@@ -12,6 +12,8 @@
 #include <unordered_map>
 
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/PiecewisePolynomial.hpp"
+#include "Domain/FunctionsOfTime/QuaternionFunctionOfTime.hpp"
 #include "Domain/FunctionsOfTime/RegisterDerivedWithCharm.hpp"
 #include "Utilities/Serialization/Serialize.hpp"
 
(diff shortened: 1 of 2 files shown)


src/Domain/Python/FunctionsOfTime.cpp
```diff
@@ -21,22 +23,99 @@ namespace domain::py_bindings {
...
       .def("func_and_deriv", &FunctionsOfTime::FunctionOfTime::func_and_deriv,
            py::arg("t"))
       .def("func_and_2_derivs",
            &FunctionsOfTime::FunctionOfTime::func_and_2_derivs, py::arg("t"));
+  m.def(
+      "serialize_functions_of_time",
+      [](const std::unordered_map<
+          std::string,
+          std::shared_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
+             shared_deserialized_functions_of_time) {
```
knelli2: You can just call this argument `functions_of_time`. The fact that it's shared is just a python thing, and they are clearly deserialized (since it's not a `vector<char>`)

src/Domain/Python/FunctionsOfTime.cpp
```diff
@@ -21,22 +23,99 @@ namespace domain::py_bindings {
...
+        std::unordered_map<
+            std::string,
+            std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
+            unique_deserialized_functions_of_time{};
+        if (shared_deserialized_functions_of_time.contains("Expansion")) {
```
knelli2: Just loop over all keys in the input map and call `get_clone`. No need to get specific names

src/Domain/Python/FunctionsOfTime.cpp
```diff
@@ -21,22 +23,99 @@ namespace domain::py_bindings {
...
       "deserialize_functions_of_time",
       [](const std::vector<char>& serialized_functions_of_time) {
         return deserialize<std::unordered_map<
             std::string,
-            std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>>(
+            std::shared_ptr<domain::FunctionsOfTime::FunctionOfTime>>>(
             serialized_functions_of_time.data());
       },
       py::arg("serialized_functions_of_time"));
+  py::class_<FunctionsOfTime::PiecewisePolynomial<3>,
```
knelli2: Just for organization, move these classes right after the `FunctionsOfTime::FunctionOfTime` class.

src/Domain/Python/FunctionsOfTime.cpp
```diff
@@ -21,22 +23,99 @@ namespace domain::py_bindings {
...
+             FunctionsOfTime::FunctionOfTime,
+             std::shared_ptr<FunctionsOfTime::PiecewisePolynomial<3>>>(
```
knelli2: For the PiecewisePolynomial, let's bind both order 2 and 3. To avoid duplicating code, write a lambda that's templated and call that for 2 and 3 like so

```cpp
const auto bind_piecewise_polynomial = [&]<size_t Order>() {
  // code that you have now using Order
  // Have each name be "PiecewisePolynomial" + std::to_string(Order)
};

bind_piecewise_polynomial.template operator()<2>();
bind_piecewise_polynomial.template operator()<3>();
```

(no need to do this for the Quat)

knelli2 (changes requested):
You can squash everything in directly. No need for fixups

knelli2 (dismissed):

src/Domain/Python/FunctionsOfTime.cpp
```diff
@@ -21,22 +23,88 @@ namespace domain::py_bindings {
...
+               FunctionsOfTime::FunctionOfTime,
+               std::shared_ptr<FunctionsOfTime::PiecewisePolynomial<Order>>>(
+        m, class_name.c_str())
+        .def(py::init(
+            [](double time,
+               std::array<DataVector, Order + 1>& initial_func_and_derivs,
+               double expiration_time) {
+              return FunctionsOfTime::PiecewisePolynomial<Order>(
+                  time, initial_func_and_derivs, expiration_time);
+            }))
```
nilsvu: Replace this with `.def(py::init<double, std::array<DataVector, Order + 1>, double>())` (this just uses the constructor with these types, no need for an extra lambda). Also add `py::arg()`s to name the arguments.

src/Domain/Python/FunctionsOfTime.cpp
```diff
@@ -21,22 +23,88 @@ namespace domain::py_bindings {
...
+        .def("time_bounds", &FunctionsOfTime::FunctionOfTime::time_bounds)
```
nilsvu: Doesn't pybind11 expose the superclass' bound functions automatically?

tests/Unit/Domain/Python/Test_FunctionsOfTime.py
```diff
@@ -33,6 +39,54 @@ def test_deserialize(self):
             3 * [DataVector(size=3, fill=0.0)],
         )
 
+    def test_serialize(self):
+        translation_fot = PiecewisePolynomial2(
+            0.0, 3 * [DataVector(size=3, fill=0.0)], math.inf
```
nilsvu: In Python always use named arguments when it makes your code clearer (almost always).

nilsvu (changes requested):

nilsvu (approved):
LGTM 👍

knelli2 (approved):

PR #6339, opened 2024-10-17
Numeric ID importer: support p-refinement

nilsvu:
## Proposed changes

When importing data from a domain that differs only by p-refinement, the data will be interpolated only within the matching elements. This eliminates errors from interpolating boundary points in the wrong neighbor. Important for PB&J.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/IO/Importers/Actions/ReadVolumeData.hpp
```diff
@@ -153,49 +155,34 @@ tuples::tagged_tuple_from_typelist<FieldTagsList> extract_element_data(
...
+    const tnsr::I<DataVector, Dim, Frame::Inertial>& inertial_coords) {
+  const auto logical_coords = logical_coordinates(mesh);
+  ElementMap<Dim, Frame::Inertial> element_map{
+      element_id, domain.blocks()[element_id.block_id()]};
+  const auto mapped_inertial_coords =
+      element_map(logical_coords, time, functions_of_time);
+  if (not equal_within_roundoff(mapped_inertial_coords, inertial_coords)) {
+    ERROR_NO_TRACE(
+        "The source and target domain don't match on grid "
+        << element_id
```
knelli2: `"don't match in block " << element_id.block_id() << " within element " << element id`

src/IO/Importers/Actions/ReadVolumeData.hpp
```diff
@@ -533,45 +545,57 @@ struct ReadAllVolumeDataAndDistribute {
...
-          source_domain_functions_of_time = deserialize<std::unordered_map<
-              std::string,
-              std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>>(
-              serialized_functions_of_time->data());
+      } else {
+        if (elements_match) {
+          Parallel::printf(
+              "WARNING: No serialized domain found in file. "
+              "Verification that elements in the source and target domain "
+              "match will be skipped.\n");
```
knelli2: I'm not sure we should allow this. Is there any scenario where somebody has an h5 file with volume data in the exact configuration of domain, but without the serialized domain?

src/IO/Importers/Tags.hpp
```diff
@@ -68,15 +69,19 @@ struct ObservationValueEpsilon {
 /*!
  * \brief Toggle interpolation of numeric data to the target domain
  */
-struct EnableInterpolation {
-  static std::string name() { return "Interpolate"; }
+struct ElementsMatch {
```
knelli2: This name is a bit obscure and I'm not sure users will have a sense of what this option does just from the name. A good name doesn't immediately stand out to me (and I see why you removed the word "interpolate" from the name since we can still do interpolation for matching grids). Here are some ideas but I'm open to more
- `InterpolateAcrossBlocks{Elements}` (or the opposite `InterpolateWithinElements`)
- `IdenticalVolumeFileDomain` ( `IdenticalH5Domain`, `IdenticalSerializedDomain`)
- `IdenticalDomains` ( `IdenticalDomainStructure`)

tests/InputFiles/GeneralizedHarmonic/KerrSchild.yaml
```diff
@@ -60,15 +60,6 @@ Amr:
   Verbosity: Quiet
 
 PhaseChangeAndTriggers:
-  - Trigger:
```
knelli2: Why remove this? If you want to remove it, do so in a separate commit since it's unrelated to volume data reading

tests/Unit/IO/Importers/Test_VolumeDataReaderActions.cpp
```diff
@@ -286,16 +302,18 @@ void test_actions(const std::variant<double, importers::ObservationSelector>&
...
+  CHECK_THROWS_WITH(test_actions<false>(0., false, false, true),
                     Catch::Matchers::ContainsSubstring("not supported"));
 
   for (const bool subcell_is_active : {false, true}) {
-    test_actions<true>(0., subcell_is_active);
-    test_actions<true>(importers::ObservationSelector::First,
+    test_actions<true>(0., false, subcell_is_active);
+    test_actions<true>(importers::ObservationSelector::First, false,
+                       subcell_is_active);
+    test_actions<true>(importers::ObservationSelector::Last, false,
```
knelli2, line 318: If this feature isn't supported for subcell or FD grids, does it error in a clear way? Meaning do we print a message somewhere like "this isn't supported for FD grids" or will it error in an obscure place like when we compare the inertial coordinates?

knelli2 (changes requested):

src/IO/Importers/Actions/ReadVolumeData.hpp
nilsvu: Yes if they generate the file by providing data at given coordinates. It's important not to babysit users too much, i.e. allow the "I know what I'm doing is correct, just let me do it" scenario. Serializing a domain into the H5 file just to enable a check is too strict.

src/IO/Importers/Actions/ReadVolumeData.hpp
nilsvu: The block ID is already printed as part of the element ID

tests/InputFiles/GeneralizedHarmonic/KerrSchild.yaml
nilsvu: Oh this was just to run the input file without segments, dropped this change

src/IO/Importers/Tags.hpp
nilsvu: changed to `ElementsAreIdentical`

knelli2 (dismissed):
Looks good. Squash in the clang-tidy fix as well

knelli2 (approved):

PR #6344, opened 2024-10-19
All cocal implementation

MaiLi-UIUC:
## Proposed changes

This pull request introduces a COCAL initial data reader into SpECTRE, enabling simulations to use initial data generated by the COCAL code.

### Upgrade instructions

No upgrade instructions are necessary.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

The procedure to use the COCAL initial data is:
1. Run `export COCAL_ROOT=/Path/To/Main/SpECTRE/Directory/src/PointwiseFunctions/AnalyticData/GrMhd/COCAL_reader`
2. Build SpECTRE and executable as always.
3. In the Input file, specify these three options:
InitialData:
  CocalInitialData:
    DataDirectory: "/Path/To/The/Cocal/Initial/Data"
    IdType: (Choose from "Ir", "Co" and "Sp")  
    ElectronFraction: 
4. In the submit.sh file, change ++ppn to +ppn and submit.


CMakeLists.txt
```diff
@@ -49,7 +49,6 @@ endif()
 if (CMAKE_VERSION VERSION_GREATER_EQUAL 3.30)
   cmake_policy(SET CMP0167 NEW)
 endif ()
-
 # FetchContent can be used to download and build some dependencies. We can
```
rhaas80, line 52: why this whitespace change?

CMakeLists.txt
```diff
@@ -76,6 +75,7 @@ include(SpectreSetSiteName)
 # We need Python in the build system and throughout the code. Setting it up
 # early so we can rely on a consistent Python version in the build system.
 find_package(Python 3.8 REQUIRED)
+# find_package(COCAL REQUIRED)
```
rhaas80, line 78: The docs do not say anything about having to uncomment cmake statements. So is this superfluous code that can be removed?

CMakeLists.txt
```diff
@@ -112,6 +112,7 @@ include(ProhibitInSourceBuild)
 include(SpectreSetupFlagsTarget)
 include(SetupFortran)
 include(SetupNinjaColors)
+include(SetupCocal)  # Add this line
```
rhaas80, line 115: why here? And not eg with `Fuka` where it would thematically fit? Ideally these would be sorted alphabetically...

CMakeLists.txt
```diff
@@ -190,6 +189,7 @@ include(UpdateAddExecutables)
 # dependencies on the PCH if necessary.
 include(SetupClangFormat)
 include(SetupClangTidy)
+#include(SetupIwyu) // This somehow doesn't work for COCAL reader
```
rhaas80, line 192: remove dead code.

CMakeLists.txt
```diff
@@ -216,6 +216,8 @@ include_directories(${CMAKE_SOURCE_DIR}/src)
 include_directories(SYSTEM ${CMAKE_BINARY_DIR}/src)
 
 add_subdirectory(src)
+# add_subdirectory(src/PointwiseFunctions/AnalyticData/GrMhd/COCAL_reader)
+# target_link_libraries(EvolveGhValenciaDivCleanBns PRIVATE COCALReader)
```
rhaas80, line 220: remove unused comments.

cmake/FindCOCAL.cmake
```diff
@@ -0,0 +1,53 @@
...
+
+# Find the COCAL initial data code (assumed repository or documentation link)
+#
+# Pass `COCAL_ROOT` to the CMake build configuration to set up the following
+# targets:
+#
+# - COCAL::Exporter: Functionality to load COCAL volume data and interpolate to
+#   arbitrary points.
+
+if(NOT DEFINED COCAL_ROOT OR COCAL_ROOT STREQUAL "")
```
rhaas80, line 12: why treat the empty string special? None of the other `FindFOO.cmake` seem to do that (see eg `FindFFTW.cmake`).

rhaas80: admittedly they mostly use `if(NOT COCAL_ROOT)` which is probably equivalent as long as empty strings are false.

rhaas80: note the comments in FFTW:

```
if(NOT FFTW_ROOT)
  # Need to set to empty to avoid warnings with --warn-uninitialized
  set(FFTW_ROOT "")
  set(FFTW_ROOT $ENV{FFTW_ROOT})
endif()
```

cmake/FindCOCAL.cmake
```diff
@@ -0,0 +1,53 @@
...
+  set(COCAL_ROOT $ENV{COCAL_ROOT})
+endif()
+
+if (DEFINED COCAL_ROOT AND NOT COCAL_ROOT STREQUAL "")
+  
+  # Find the COCAL library
+  find_library(
+    COCAL_LIB
+    NAMES libcocal.a
```
rhaas80, line 21: this should not contain the `lib` or the `.a` (at least the latter is system dependent).

cmake/FindCOCAL.cmake
```diff
@@ -0,0 +1,53 @@
...
+    PATHS ${COCAL_ROOT}/lib
```
rhaas80, line 22: presumably `HINTS` would be closer to what the others do. Note that `lib` is system specific (could be `lib` or `lib64` or `lib/x86_64-linux-gnu` to name a few).

cmake/FindCOCAL.cmake
```diff
@@ -0,0 +1,53 @@
...
+    NO_DEFAULT_PATHS
```
rhaas80, line 23: why not? One could have a COCAL one actually uses.

cmake/FindCOCAL.cmake
```diff
@@ -0,0 +1,53 @@
...
+  )
+
+  # Find the COCAL include directory
+  find_path(
+    COCAL_INCLUDE_DIR
+    NAMES coc2cac_bns.f90
```
rhaas80, line 29: a `.f90` file in an include directory? Why?

cmake/SetupCocal.cmake
```diff
@@ -0,0 +1,15 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+
+find_package(COCAL)
+
+if (COCAL_FOUND)
+  message("-- Found Cocal at ${COCAL_ROOT}")
```
rhaas80, line 8: should use `STATUS`, should output actually useful information, see eg `SetupGsl.cmake`.

src/IO/External/InterpolateFromCocal.cpp
```diff
@@ -0,0 +1,204 @@
...
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+extern "C" {
+// Define the COCAL Fortran reader functions here for different id_types.
+void coc2cac_co(const int& npoints, const double* x, const double* y,
```
rhaas80, line 18: since this is provided by COCAL itself, shouldn't there be a header file it provides that one can use to get this?

rhaas80: there is not really much of a difference between a `&` and a `*`. Are you trying to imply that the `*` ones are actually arrays?

rhaas80: a `const int &` make not much sense, I'd just pass by value. Unless you somehow are actually calling out to a Fortran routine, in which case this is all bad anyway since different Fortran compilers use different name mangling schemes.

src/IO/External/InterpolateFromCocal.cpp
```diff
@@ -0,0 +1,204 @@
...
+                double* spatial_metric_yy, double* spatial_metric_yz,
+                double* spatial_metric_zz, double* extrinsic_curvature_xx,
+                double* extrinsic_curvature_xy, double* extrinsic_curvature_xz,
+                double* extrinsic_curvature_yy, double* extrinsic_curvature_yz,
+                double* extrinsic_curvature_zz, double* rest_mass_density,
+                double* spatial_velocity_x, double* spatial_velocity_y,
+                double* spatial_velocity_z, double* pressure,
+                double* specific_internal_energy);
+
+void coc2cac_ir(const int& npoints, const double* x, const double* y,
```
rhaas80, line 32: I assume `_co` and `_ir` stand for "corotating" and "irrotating" respectively? Why would this code here care? Shouldn't COCAL provide a single reader routine that takes as input a file name and produces output? Ie Spectre should not need to know the details of what sort of method was used to produce this ID?

src/IO/External/InterpolateFromCocal.cpp
```diff
@@ -0,0 +1,204 @@
...
+                double* spatial_metric_yy, double* spatial_metric_yz,
+                double* spatial_metric_zz, double* extrinsic_curvature_xx,
+                double* extrinsic_curvature_xy, double* extrinsic_curvature_xz,
+                double* extrinsic_curvature_yy, double* extrinsic_curvature_yz,
+                double* extrinsic_curvature_zz, double* rest_mass_density,
+                double* spatial_velocity_x, double* spatial_velocity_y,
+                double* spatial_velocity_z, double* pressure,
+                double* specific_internal_energy);
+
+void coc2cac_sp(const int& npoints, const double* x, const double* y,
```
rhaas80, line 46: spinning?

src/IO/External/InterpolateFromCocal.cpp
```diff
@@ -0,0 +1,204 @@
...
+                double* spatial_metric_xy, double* spatial_metric_xz,
+                double* spatial_metric_yy, double* spatial_metric_yz,
+                double* spatial_metric_zz, double* extrinsic_curvature_xx,
+                double* extrinsic_curvature_xy, double* extrinsic_curvature_xz,
+                double* extrinsic_curvature_yy, double* extrinsic_curvature_yz,
+                double* extrinsic_curvature_zz, double* rest_mass_density,
+                double* spatial_velocity_x, double* spatial_velocity_y,
+                double* spatial_velocity_z, double* pressure,
+                double* specific_internal_energy);
+}
```
rhaas80, line 59: this is still within the `extern "C"` isn't it?

src/IO/External/InterpolateFromCocal.cpp
```diff
@@ -0,0 +1,204 @@
...
+  std::vector<double> specific_internal_energy(num_points);
+
+  // Pass the string and its length to Fortran
+  const char* dir_cstr =
+      data_directory.c_str();  // Convert std::string to C string
+  int len_dir = static_cast<int>(
+      data_directory.length());  // Get the length of the string
+
+  if (id_type == CocalIdType::Co) {
+    coc2cac_co(num_points, x_coords.data(), y_coords.data(), z_coords.data(),
```
rhaas80, line 118: since you break this down to a plain C pointer anyway, why not use the `DataVector`'s `data()` member (inherited from Blaze https://bitbucket.org/blaze-lib/blaze/wiki/Vector%20Operations#!data-data)  instead?

src/IO/External/InterpolateFromCocal.cpp
```diff
@@ -0,0 +1,204 @@
...
+               spatial_metric_zz.data(), extrinsic_curvature_xx.data(),
+               extrinsic_curvature_xy.data(), extrinsic_curvature_xz.data(),
+               extrinsic_curvature_yy.data(), extrinsic_curvature_yz.data(),
+               extrinsic_curvature_zz.data(), rest_mass_density.data(),
+               spatial_velocity_x.data(), spatial_velocity_y.data(),
+               spatial_velocity_z.data(), pressure.data(),
+               specific_internal_energy.data());
+  }
+
+  get(get<gr::Tags::Lapse<DataVector>>(result)) = to_datavector(lapse);
```
rhaas80, line 158: I'd just create `DataVector`s above then pass their `.data()` member's results pointer to COCAL then or even use `get<gr::Tags::Lapse<DataVector>>(result).data()` in the COCAL call (though on will have to check that this already has the correct size).

src/IO/External/InterpolateFromCocal.hpp
```diff
@@ -0,0 +1,46 @@
...
+               gr::Tags::ExtrinsicCurvature<DataVector, 3>,
+               hydro::Tags::RestMassDensity<DataVector>,
+               hydro::Tags::SpecificInternalEnergy<DataVector>,
+               hydro::Tags::Pressure<DataVector>,
+               hydro::Tags::LorentzFactor<DataVector>,
+               hydro::Tags::SpatialVelocity<DataVector, 3>>;
+
+/*!
+ * \brief Interpolate numerical COCAL initial data to arbitrary points
+ * \param cocal_lock A simple lock to lock the thread
```
rhaas80, line 35: required b/c COCAL itself is not thread safe

src/PointwiseFunctions/AnalyticData/GrMhd/CMakeLists.txt
```diff
@@ -93,4 +93,37 @@ if (TARGET FUKA::Exporter)
     )
 endif()
 
+if (TARGET COCAL::Exporter)
+add_custom_command(
+  OUTPUT ${CMAKE_SOURCE_DIR}/src/PointwiseFunctions/AnalyticData/GrMhd/COCAL_reader/libcocal.a
```
rhaas80, line 98: `lib` and `.a` are system specific.

src/PointwiseFunctions/AnalyticData/GrMhd/CMakeLists.txt
```diff
@@ -93,4 +93,37 @@ if (TARGET FUKA::Exporter)
...
+  COMMAND make -C ${CMAKE_SOURCE_DIR}/src/PointwiseFunctions/AnalyticData/GrMhd/COCAL_reader clean
```
rhaas80, line 99: `make` may not be available (could be `ninja` only)

src/PointwiseFunctions/AnalyticData/GrMhd/CMakeLists.txt
```diff
@@ -93,4 +93,37 @@ if (TARGET FUKA::Exporter)
...
+  COMMAND make -C ${CMAKE_SOURCE_DIR}/src/PointwiseFunctions/AnalyticData/GrMhd/COCAL_reader
```
rhaas80, line 100: so why require me to point to a COCAL library above? Does it take the source code? That is somewhat strange, usually one would point to the compiled version.

src/PointwiseFunctions/AnalyticData/GrMhd/COCAL_reader/IO_input_CF_flco_export.f90
```diff
@@ -0,0 +1,22 @@
+subroutine IO_input_CF_flco_export(filenm,emd,ome,ber,radi)
```
rhaas80, line 1: huh? What does this routine do? If for COCAL why not put it into a COCAL F90 module to avoid namespace pollution?

rhaas80: ok, this seems to be a mostly full copy of COCAL. Why then did I provide a `COCAL_ROOT` above? Shouldn't all these files be removed?

src/PointwiseFunctions/AnalyticData/GrMhd/COCAL_reader/coc2cac_rs_WL_v1.f90
```diff
@@ -0,0 +1,300 @@
+!_____________________________________________________________________________
+!
+!    CACTUS READER OF COCAL ROTATING STAR IN WAVELESS FORMALISM
```
rhaas80, line 3: reasonably sure that at Cactus reader for COCAL is not required for Spectre.

src/PointwiseFunctions/AnalyticData/GrMhd/CocalInitialData.cpp
```diff
@@ -0,0 +1,59 @@
...
+
+#include "PointwiseFunctions/AnalyticData/GrMhd/CocalInitialData.hpp"
+
+#include <memory>
+#include <pup.h>
+
+namespace grmhd::AnalyticData {
+
+CocalInitialData::CocalInitialData(std::string data_directory,
+                                   const double electron_fraction,
```
rhaas80, line 12: a bit strange that `electron_fraction` has to be input. Does the code actually do anything with it? Otherwise is there any generic code that can handle this (eg how is this done for an run-of-the mill polytropic TOV star)?

src/PointwiseFunctions/AnalyticData/GrMhd/CocalInitialData.hpp
```diff
@@ -0,0 +1,216 @@
...
+
+  struct ElectronFraction {
+    using type = double;
+    static constexpr Options::String help = {"Constant electron fraction"};
+  };
+
+  struct IdType {
+    using type = CocalIdType;
+    static constexpr Options::String help = {
+        "The ID type of COCAL data (Co, Ir, Sp)"};
```
rhaas80, line 62: what do these two letter abbreviations mean? Why not use actually telling names?

src/PointwiseFunctions/AnalyticData/GrMhd/CocalInitialData.hpp
```diff
@@ -0,0 +1,216 @@
...
+  explicit CocalInitialData(CkMigrateMessage* msg);
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(CocalInitialData);
+  /// \endcond
+
+  template <typename DataType>
+  using tags = tmpl::append<
+      tmpl::list<gr::Tags::SpatialMetric<DataType, 3>,
+                 gr::Tags::ExtrinsicCurvature<DataType, 3>,
+                 gr::Tags::Lapse<DataType>, gr::Tags::Shift<DataType, 3>>,
```
rhaas80, line 91: no `dtlapse` or `dtshift`?

src/PointwiseFunctions/AnalyticData/GrMhd/CocalInitialData.hpp
```diff
@@ -0,0 +1,216 @@
...
+    // for (size_t i = 0; i < num_points; ++i) {
+    //   const double local_rest_mass_density = get(rest_mass_density)[i];
+    //   if (equal_within_roundoff(local_rest_mass_density, 0.)) {
+    //     get(specific_enthalpy)[i] = 1.;
+    //   } else {
+    //     get(specific_enthalpy)[i] =
+    //     get(hydro::relativistic_specific_enthalpy(
+    //         Scalar<double>(local_rest_mass_density),
+    //         Scalar<double>(get(specific_internal_energy)[i]),
+    //         Scalar<double>(get(pressure)[i])));
```
rhaas80, line 144: if these are not used, I'd remove the commented out code.

src/PointwiseFunctions/AnalyticData/GrMhd/CocalInitialData.hpp
```diff
@@ -0,0 +1,216 @@
...
+                   hydro::Tags::LorentzFactor<DataVector>,
+                   hydro::Tags::SpatialVelocity<DataVector, 3>>>(
+        [&result, &interpolated_vars](const auto tag_v) {
+          using tag = tmpl::type_from<std::decay_t<decltype(tag_v)>>;
+          get<tag>(result) = std::move(get<tag>(interpolated_vars));
+        });
+    // Compute derived quantities from interpolated data
+    // const size_t num_points = x.begin()->size();
+    // const auto& rest_mass_density =
+    //     get<hydro::Tags::RestMassDensity<DataVector>>(result);
```
rhaas80, line 123: remove commented out code.

src/PointwiseFunctions/AnalyticData/GrMhd/CocalInitialData.hpp
```diff
@@ -0,0 +1,216 @@
...
+    // Constant electron fraction specified by input file
+    auto& electron_fraction =
+        get<hydro::Tags::ElectronFraction<DataVector>>(result);
+    get(electron_fraction) = DataVector(num_points, electron_fraction_);
+    // Zero magnetic field and divergence cleaning field
+    auto& magnetic_field =
+        get<hydro::Tags::MagneticField<DataVector, 3>>(result);
+    get<0>(magnetic_field) = DataVector(num_points, 0.);
+    get<1>(magnetic_field) = DataVector(num_points, 0.);
+    get<2>(magnetic_field) = DataVector(num_points, 0.);
```
rhaas80, line 155: no support for magnetic fields in COCAL yet?

src/PointwiseFunctions/AnalyticData/GrMhd/CocalInitialData.hpp
```diff
@@ -0,0 +1,216 @@
...
+  void pup(PUP::er& /*p*/) override;
+
+ private:
+  std::string data_directory_{};
+  CocalIdType id_type_;
+  double electron_fraction_ = std::numeric_limits<double>::signaling_NaN();
+
+  // This lock is used to ensure that only one thread at a time is calling the
+  // COCAL interpolation routines. We make some assumptions here to guarantee
+  // thread-safety:
```
rhaas80, line 183: assumptions won't guarantee anything. Are these assumptions actually true? If not then this needs to be checked at runtime (and the assumptions *documented* like they are done here).

rhaas80 (changes requested):
dead code, contains complete copy of COCAL, some suggestion on how to make things easier on the end user.

Have not checked if this violates any SPECTRE code convention. Has the C++ code been run through the mandated version of clang-format (if any)?

PR #6316, opened 2024-09-30
Add Transition to Ringdown Script

AlexCarpenter46:
## Proposed changes

This is the first iteration of the transition to ringdown script for the BBH's. To use this script, you have to have a BBH that has reached common horizon and output reduction and volume data at the time of common horizon finds using the pipeline, then you can do something like

`spectre bbh start-ringdown path/to/final/segment/directory/from/inspiral --path-to-output-h5 Ringdown_Coefs.h5 --match-time 5000 --number-of-steps 10 --settling-timescale 10.0 -L 2 -P 9 -O Ringdown_9 --no-submit` 

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

### Further comments
This script is still very experimental, if you try it out and things don't work or break, please let me know :)

PR #6346, opened 2024-10-21
Update Inspiral.yaml

### LINKED ISSUE #6124: Updates to BBH evolution pipeline YAML
- Use AM not AB
- Remove ElementSizeCFL step chooser (just use error based stepping)
- Near merger, write volume data very often

Partially addressed in #6291.

Still need to update constraint damping parameters

AlexCarpenter46:
## Proposed changes

This update produces a yaml very close to what was done in the bbh paper. Some widths, amplitudes and timescales are slightly different (off by 1e-6) and the filtering is different, but produces the same result.
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/Inspiral.py
+++ b/support/Pipelines/Bbh/Inspiral.py
@@ -64,12 +64,16 @@ def _constraint_damping_params(
 ) -> dict:
     total_mass = mass_left + mass_right
     return {
-        "Gamma0Constant": 0.001 / total_mass,
+        # SpEC chooses a Gamma0Constant of 0.001, but we found 0.01 produces
+        # smaller constraints violations in the envelope/outer shell region.
+        "Gamma0Constant": 0.01 / total_mass,
         "Gamma0LeftAmplitude": 4.0 / mass_left,
         "Gamma0LeftWidth": 7.0 * mass_left,
         "Gamma0RightAmplitude": 4.0 / mass_right,
         "Gamma0RightWidth": 7.0 * mass_right,
-        "Gamma0OriginAmplitude": 0.075 / total_mass,
+        # SpEC chooses a Gamm0OriginAmplitude of 0.075, but we found that 0.75
+        # produces a smaller burst of constraints from junk radiation.
+        "Gamma0OriginAmplitude": 0.75 / total_mass,
         "Gamma0OriginWidth": 2.5 * initial_separation,
         "Gamma1Width": 10.0 * initial_separation,
     }
@@ -147,8 +151,12 @@ def inspiral_parameters(
             / (id_input_file["Observers"]["VolumeFileName"] + "*.h5")
         ),
         # Domain geometry
-        "ExcisionRadiusA": id_domain_creator["ObjectA"]["InnerRadius"],
-        "ExcisionRadiusB": id_domain_creator["ObjectB"]["InnerRadius"],
+        # This factor is to account for the ID excision not being the same shape
+        # as the final horizon found after the last iteration. Found through
+        # trial and error that increasing the excision size by this factor
+        # allowed the runs to evolve without early incoming char speeds.
+        "ExcisionRadiusA": id_domain_creator["ObjectA"]["InnerRadius"] * 1.0385,
+        "ExcisionRadiusB": id_domain_creator["ObjectB"]["InnerRadius"] * 1.0385,
         "XCoordA": id_domain_creator["ObjectA"]["XCoord"],
         "XCoordB": id_domain_creator["ObjectB"]["XCoord"],
         "CenterOfMassOffset_y": id_domain_creator["CenterOfMassOffset"][0],
@@ -243,8 +251,11 @@ def inspiral_parameters_spec(
         # Initial data files
         "SpecDataDirectory": str(Path(id_run_dir).resolve()),
         # Domain geometry
-        "ExcisionRadiusA": id_params["ID_rExcA"],
-        "ExcisionRadiusB": id_params["ID_rExcB"],
+        # SpEC excision in ID_Params.perl is 0.89 * horizon radius, but
+        # usually you want to excise less than the maximum. Here use 6% larger,
+        # or about 0.9434 * horizon radius.
+        "ExcisionRadiusA": id_params["ID_rExcA"] * 1.06,
+        "ExcisionRadiusB": id_params["ID_rExcB"] * 1.06,
         "XCoordA": id_params["ID_cA"][0],
         "XCoordB": id_params["ID_cB"][0],
         # COM offset in y and z is the same for both objects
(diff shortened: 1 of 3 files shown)


support/Pipelines/Bbh/Inspiral.py
```diff
@@ -147,8 +147,8 @@ def inspiral_parameters(
             / (id_input_file["Observers"]["VolumeFileName"] + "*.h5")
         ),
         # Domain geometry
-        "ExcisionRadiusA": id_domain_creator["ObjectA"]["InnerRadius"],
-        "ExcisionRadiusB": id_domain_creator["ObjectB"]["InnerRadius"],
+        "ExcisionRadiusA": id_domain_creator["ObjectA"]["InnerRadius"] * 1.0385,
+        "ExcisionRadiusB": id_domain_creator["ObjectB"]["InnerRadius"] * 1.0385,
```
knelli2, line 159: Just need to add the factor for the spec ID from #6316

knelli2 (changes requested):

support/Pipelines/Bbh/Inspiral.py
AlexCarpenter46: squashed that in :)

AlexCarpenter46:
Have tested these changes with the initial data from the bbh paper and the run seems stable and is at t=338M currently.

knelli2 (dismissed):

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -64,12 +64,12 @@ def _constraint_damping_params(
 ) -> dict:
     total_mass = mass_left + mass_right
     return {
-        "Gamma0Constant": 0.001 / total_mass,
+        "Gamma0Constant": 0.01 / total_mass,
         "Gamma0LeftAmplitude": 4.0 / mass_left,
         "Gamma0LeftWidth": 7.0 * mass_left,
         "Gamma0RightAmplitude": 4.0 / mass_right,
         "Gamma0RightWidth": 7.0 * mass_right,
-        "Gamma0OriginAmplitude": 0.075 / total_mass,
```
nilsvu, line 72: Please add a comment here how this differs from SpEC and why (please always add comments for things like this that changed for a reason but we will forget the reason next week)

support/Pipelines/Bbh/Inspiral.py
nilsvu: I don't understand this factor. Please add a comment. Why not just increase the initial data excision by this factor?

support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -147,14 +147,14 @@ Evolution:
...
+  InitialSlabSize: 0.1
   StepChoosers:
     - LimitIncrease:
         Factor: 2
     - PreventRapidIncrease
     - ErrorControl:
-        AbsoluteTolerance: 1e-8
-        RelativeTolerance: 1e-6
+        AbsoluteTolerance: 1e-10
+        RelativeTolerance: 1e-8
```
nilsvu, line 159: Again, please add comments if this changed for a reason

nilsvu (changes requested):

support/Pipelines/Bbh/Inspiral.py
AlexCarpenter46: I'm not sure why this differs from SpEC, I believe it should be 0.075, and I pushed a fixup for this maybe a year ago now. I changed it back here since this is what was done in the paper, I'm not sure why it was done or if it changes the result in a significant way. @geoffrey4444 do you know if this impacts the bbh runs?

support/Pipelines/Bbh/Inspiral.py
AlexCarpenter46: This factor comes from dividing the excision radius used in the paper by what develop outputs for the excision radius when using the same initial data. I guess I could increase the initial data, I didn't think about that, that way there would only be one number floating around changing the excision radius.

knelli2: I think this is because the ID excision in the last iteration of the control loop is not the same shape as the final horizon found after the last iteration. But then the inspiral excision is set based off this final horizon mass/spin. Therefore it's possible they don't line up properly, so making the inspiral excision slightly bigger accounts for this.

support/Pipelines/Bbh/Inspiral.py
knelli2: Yes I remember this being changed from 0.75 down to 0.075, but I believe this was also when we were running with too low of resolution and we weren't convergent. Once our resolution was convergent, this new value seemed to decrease the constraint violations in problem areas . But @geoffrey4444 please confirm

support/Pipelines/Bbh/Inspiral.py
nilsvu: Hm can we just start from the ID excision, using the same initial shape for the excision as the ID? Not just the same inner radius, but set the initial shape to the same Kerr-Schild parameters as well (instead of using the AH shape). This would avoid empirically found and potentially fragile factors like this one.

knelli2: We could but then this would cause shape/size control to take longer (or potentially many more iterations) to settle to the correct shape/size since we're controlling to the horizon.

I'd like to get what we did for the paper onto develop so we at least have it, and then we can decide what's best to do after. There's quite a lot more automation that has to happen and decisions to be made about what's best/robust. Even if we use the ID excision shape/size, we'll likely change that anyways.

support/Pipelines/Bbh/Inspiral.py
geoffrey4444: I used 0.75 instead of 0.075 as in SpEC, in the bbh inspirals for the paper, because I found that this helped with reducing the big burst of constraints from junk radiation. I also set the constant to 0.01 instead of 0.001 as in SpEC, e.g. https://arxiv.org/src/2410.00265v1/anc/Bbh_Lev1_Inspiral.yaml

I'm not saying those are the perfect numbers, but they worked well for the runs in the paper. I just multiplied these two numbers by 10 and the constraints decayed better after the initial junk burst. In ParaView I saw large constraints far from the inner shells, around the envelope part of the domain, so I focused on these terms, rather than the terms controlling constraints close to the excision surfaces

support/Pipelines/Bbh/Inspiral.yaml
geoffrey4444: I adjusted the excision radii in the evolution by trial and error to avoid incoming char speeds when starting the evolution. I found that not excising as far in as the initial data did worked better in my runs.

nilsvu (approved):

knelli2 (approved):

PR #6347, opened 2024-10-21
Allow three extra refinement levels for an ElementId

kidder:
This is done by using a compact representation of a SegmentId. Also fix some clang-tidy errors.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Structure/ElementId.cpp
+++ b/src/Domain/Structure/ElementId.cpp
@@ -3,29 +3,32 @@
 
 #include "Domain/Structure/ElementId.hpp"
 
-#include <boost/functional/hash.hpp>
-#include <exception>
-#include <limits>
+#include <array>
+#include <cstddef>
+#include <cstdint>
+#include <functional>
+#include <optional>
 #include <ostream>
-#include <pup.h>
-#include <pup_stl.h>
 #include <regex>
 #include <string>
+#include <type_traits>
 
+#include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/SegmentId.hpp"
+#include "Domain/Structure/Side.hpp"
 #include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
-#include "Utilities/Gsl.hpp"
 #include "Utilities/Literals.hpp"
-#include "Utilities/MakeArray.hpp"
-#include "Utilities/Numeric.hpp"
 #include "Utilities/StdHelpers.hpp"
+#include "Utilities/StdHelpers/Bit.hpp"
 
 // The `static_assert`s verify that `ElementId` satisfies the constraints
 // imposed by Charm++ to make `ElementId` able to act as an index into Charm++'s
 // arrays. These constraints are:
 // - `ElementId` must satisfy `std::is_standard_layout` and `std::is_trivial`
-// - `ElementId` must not be larger than the size of three `int`s, i.e.
-//   `sizeof(ElementId) <= 3 * sizeof(int)`
+// - `ElementId` is the size of two `int`s`
 static_assert(std::is_standard_layout_v<ElementId<1>> and
               std::is_trivial_v<ElementId<1>>);
 static_assert(std::is_standard_layout_v<ElementId<2>> and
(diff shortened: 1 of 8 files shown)


src/Utilities/StdHelpers/Bit.hpp
```diff
@@ -0,0 +1,34 @@
...
+
+template <typename T,
+          Requires<std::is_same_v<T, uint8_t> or std::is_same_v<T, uint16_t> or
+                   std::is_same_v<T, uint32_t> or std::is_same_v<T, uint64_t> or
+                   std::is_same_v<T, size_t>> = nullptr>
+// NOLINTNEXTLINE(cert-dcl58-cpp)
+constexpr bool has_single_bit(T x) noexcept {
+  return std::popcount(x) == 1;
+}
+}  // namespace std
```
nilsdeppe: We should only be adding this into the std namespace if it's not defined in the standard, otherwise we hit UB. There's is identically 0 chance that people will correctly place the if-else checks for the includes in all the right places.

Also, please make adding this a separate commit.

Make sure to add a code comment into this file explaining why this was necessary.

src/Domain/Structure/SegmentId.cpp
```diff
@@ -12,7 +14,7 @@
 
 SegmentId::SegmentId(const size_t refinement_level, const size_t index)
     : refinement_level_(refinement_level), index_(index) {
-  ASSERT(refinement_level < max_refinement_level,
+  ASSERT(refinement_level <= max_refinement_level,
```
nilsdeppe: Just remove this check. It's not a constraint on `SegmentId` but a constraint on `ElementId`

src/Domain/Structure/ElementId.cpp
```diff
@@ -3,29 +3,37 @@
...
-#include <pup_stl.h>
 #include <regex>
 #include <string>
+#include <type_traits>
 
+#if defined(__cpp_lib_int_pow2) && __cpp_lib_int_pow2 >= 202002L
+#include <bit>
+#else
+#include "Utilities/StdHelpers/Bit.hpp"
+#endif
```
nilsdeppe: This check should be in the `Bit.hpp` file. Nobody will know to do this magic when including that file.

src/Domain/Structure/ElementId.cpp
```diff
@@ -40,17 +48,88 @@ static_assert(sizeof(ElementId<2>) == 2 * sizeof(int),
...
+// 0000000000000011 L1I1
+// 0000000000000100 L2I0
+// ...
+// 1111111111111111 L15I32767
+static_assert(ElementId<1>::max_refinement_level ==
+              SegmentId::max_refinement_level);
+static_assert(ElementId<2>::max_refinement_level ==
+              SegmentId::max_refinement_level);
+static_assert(ElementId<3>::max_refinement_level ==
+              SegmentId::max_refinement_level);
```
nilsdeppe: I think we can remove these checks since there's no actual restriction on `SegmentId`

src/Domain/Structure/ElementId.cpp
```diff
@@ -40,17 +48,88 @@ static_assert(sizeof(ElementId<2>) == 2 * sizeof(int),
...
+}
+
+size_t get_index(uint16_t compact_segment_id) {
+  ASSERT(compact_segment_id > 0, "Undefined compact segment id");
+  return static_cast<size_t>(compact_segment_id -
+                             std::bit_floor(compact_segment_id));
+}
+
+SegmentId make_segment_id(const uint16_t compact_segment_id) {
+  return SegmentId{get_refinement_level(compact_segment_id),
```
nilsdeppe, line 83: Add the `ASSERT` here too about  undefined compact segment id?

src/Domain/Structure/ElementId.cpp
```diff
@@ -40,17 +48,88 @@ static_assert(sizeof(ElementId<2>) == 2 * sizeof(int),
...
+         static_cast<uint16_t>(segment_id.index());
+}
+
+size_t get_refinement_level(uint16_t compact_segment_id) {
+  ASSERT(compact_segment_id > 0, "Undefined compact segment id");
+  return SegmentId::max_refinement_level -
+         static_cast<size_t>(std::countl_zero(compact_segment_id));
+}
+
+size_t get_index(uint16_t compact_segment_id) {
```
nilsdeppe: `const uint16_t`?

src/Domain/Structure/ElementId.cpp
```diff
@@ -40,17 +48,88 @@ static_assert(sizeof(ElementId<2>) == 2 * sizeof(int),
...
+              SegmentId::max_refinement_level);
+static_assert(ElementId<3>::max_refinement_level ==
+              SegmentId::max_refinement_level);
+namespace {
+uint16_t make_compact_segment_id(const SegmentId& segment_id) {
+  return (uint16_t{1} << segment_id.refinement_level()) +
+         static_cast<uint16_t>(segment_id.index());
+}
+
+size_t get_refinement_level(uint16_t compact_segment_id) {
```
nilsdeppe: `const uint16_t`?

src/Domain/Structure/ElementId.cpp
```diff
@@ -40,17 +48,88 @@ static_assert(sizeof(ElementId<2>) == 2 * sizeof(int),
...
+bool is_on_lower_block_boundary(const uint16_t compact_segment_id) {
+  ASSERT(compact_segment_id > 0, "Undefined compact segment id");
+  // true if index is zero, which means only a single bit is a one
+  return std::has_single_bit(compact_segment_id);
+}
+
+bool is_on_upper_block_boundary(const uint16_t compact_segment_id) {
+  ASSERT(compact_segment_id > 0, "Undefined compact segment id");
+  // true if all bits after the leading one are also one
+  return 16 == std::countl_zero(compact_segment_id) +
```
nilsdeppe: I would find this easier to read as `16 == (...)`

src/Domain/Structure/ElementId.cpp
```diff
@@ -40,17 +48,88 @@ static_assert(sizeof(ElementId<2>) == 2 * sizeof(int),
...
+}
+}  // namespace
+
 template <size_t VolumeDim>
-ElementId<VolumeDim>::ElementId(const size_t block_id, const size_t grid_index)
+ElementId<VolumeDim>::ElementId(uint8_t block_id, uint8_t grid_index,
+                                uint8_t direction,
+                                uint16_t compact_segment_id_xi,
+                                uint16_t compact_segment_id_eta,
+                                uint16_t compact_segment_id_zeta)
```
nilsdeppe: `const uint?_t`

nilsdeppe (changes requested):

nilsvu (dismissed):
This is pretty cool 😎 thanks for doing this.

nilsdeppe (commented):
LGTM, could you move the adding of `std::bit_*` into a separate commit as you squash?

nilsdeppe (approved):

PR #6359, opened 2024-11-01
Extend sphere and wedge transition

knelli2:
## Proposed changes

Allows the sphere and wedge shape map transition function to be extended outside the inner and outer radii/surfaces. This will be needed for BBH ringdown transition and also for extrapolating into excision spheres.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
@@ -12,6 +12,7 @@
 #include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/MakeString.hpp"
 
 namespace domain::CoordinateMaps::ShapeMapTransitionFunctions {
 
@@ -55,10 +56,23 @@ std::optional<double> SphereTransition::original_radius_over_radius(
   }
   const double original_radius = (mag + radial_distortion * b_) / denom;
 
-  return (original_radius + eps_) >= r_min_ and
-                 (original_radius - eps_) <= r_max_
-             ? std::optional<double>{original_radius / mag}
-             : std::nullopt;
+  // If we are within r_min and r_max, doesn't matter if we are reverse, we just
+  // return the value we calculated. If we are reverse and the point is outside
+  // r_max or we aren't reversed and the point is inside r_min, then we return a
+  // simplified formula since the transition function is 1 in this region. If
+  // the above conditions aren't true, then we are in the region where the
+  // transition function is 0, so we return 1.0. Otherwise we return nullopt.
+  if ((original_radius + eps_) >= r_min_ and
+      (original_radius - eps_) <= r_max_) {
+    return std::optional<double>{original_radius / mag};
+  } else if ((a_ > 0.0 and mag > r_max_) or (a_ < 0.0 and mag < r_min_)) {
+    // a_ being positive is a sentinel for reverse (see constructor)
+    return {1.0 + radial_distortion / mag};
+  } else if ((a_ < 0.0 and mag > r_max_) or (a_ > 0.0 and mag < r_min_)) {
+    return {1.0};
+  } else {
+    return std::nullopt;
+  }
 }
 
 std::array<double, 3> SphereTransition::gradient(
@@ -73,15 +87,17 @@ std::array<DataVector, 3> SphereTransition::gradient(
 template <typename T>
 T SphereTransition::call_impl(const std::array<T, 3>& source_coords) const {
   const T mag = magnitude(source_coords);
-  check_magnitudes(mag);
-  return a_ * mag + b_;
+  check_magnitudes(mag, false);
+  // See https://github.com/sxs-collaboration/spectre/issues/6376 for why the
+  // T{} is necessary inside the clamp.
+  return blaze::clamp(T{a_ * mag + b_}, 0.0, 1.0);
 }
 
 template <typename T>
 std::array<T, 3> SphereTransition::gradient_impl(
     const std::array<T, 3>& source_coords) const {
   const T mag = magnitude(source_coords);
-  check_magnitudes(mag);
+  check_magnitudes(mag, true);
   return a_ * source_coords / mag;
 }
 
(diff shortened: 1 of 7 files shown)


knelli2:
@nilsvu Pushed a fixup for the Wedge transition as well

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
```diff
@@ -12,12 +12,13 @@
...
 #include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
+#include "Utilities/MakeString.hpp"
 
 namespace domain::CoordinateMaps::ShapeMapTransitionFunctions {
 
 SphereTransition::SphereTransition(const double r_min, const double r_max,
-                                   const bool reverse)
-    : r_min_(r_min), r_max_(r_max) {
+                                   const bool reverse, const bool extend)
```
nilsvu: Just enable this always, can't hurt. That's what other maps do.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
```diff
@@ -72,14 +89,21 @@ std::array<DataVector, 3> SphereTransition::gradient(
...
-  return a_ * mag + b_;
+  auto& result = mag;
+  result = a_ * mag + b_;
+  return extend_ ? blaze::clamp(result, 0.0, 1.0) : result;
 }
 
 template <typename T>
 std::array<T, 3> SphereTransition::gradient_impl(
     const std::array<T, 3>& source_coords) const {
+  if (UNLIKELY(extend_)) {
```
nilsvu: Only error here if you are evaluating outside of the bounds

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -62,12 +63,13 @@ Wedge::Wedge(const std::array<double, 3>& inner_center,
              const double inner_radius, const double inner_sphericity,
              const std::array<double, 3>& outer_center,
              const double outer_radius, const double outer_sphericity,
-             const Axis axis, const bool reverse)
+             const Axis axis, const bool reverse, const bool extend)
```
nilsvu: Also here, no need to disable

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -76,6 +78,14 @@ Wedge::Wedge(const std::array<double, 3>& inner_center,
         << inner_surface_.center
         << "\nouter center: " << outer_surface_.center);
   }
+  if (projection_center_ != std::array{0.0, 0.0, 0.0} and extend_) {
```
nilsvu: Only error if called outside bounds

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -441,6 +473,11 @@ std::array<DataVector, 3> Wedge::gradient(
 template <typename T>
 std::array<T, 3> Wedge::gradient_impl(
     const std::array<T, 3>& source_coords) const {
+  if (UNLIKELY(extend_)) {
```
nilsvu: Same here

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
```diff
@@ -55,10 +56,26 @@ std::optional<double> SphereTransition::original_radius_over_radius(
...
+  // the transition function is 1 in this region. If we are extending, but the
+  // above conditions aren't true, then we are in the region where the
+  // transition function is 0, so we return 1.0. Otherwise we return nullopt.
+  if ((original_radius + eps_) >= r_min_ and
+      (original_radius - eps_) <= r_max_) {
+    return std::optional<double>{original_radius / mag};
+  } else if (extend_) {
+    // a_ being positive is a sentinel for reverse (see constructor)
+    if ((a_ > 0.0 and mag > r_max_) or (a_ < 0.0 and mag < r_min_)) {
+      return {1.0 + radial_distortion / mag};
```
nilsvu: Do you have a test for the inverse in this regime, which maps the point back and forth through the wedge? We should have that explicitly somewhere to guarantee that the wedge+shape map and its inverse work outside of the wedge

nilsvu (changes requested):

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
knelli2: The reason I have the choice and explicitly don't enable it in most cases is because this does impact control system horizon finds. This is different from other maps because the shape map acts in the distorted frame. If we always allow this to succeed, then when finding the horizon in the distorted frame and a point moves outside a block with a distorted frame, the inverse (of a different block) would still find the point and we'd never know that the horizon has actually moved out of a region with a distorted frame.

I think it's better for keeping track of things to only allow an extended shape map if we actually need it

nilsvu: I thought we find horizons only in blocks that have a distorted frame, meaning we don't even try the inverse when there is no distorted frame. Also in the logical frame we usually reject points outside [-1, 1].

To allow extrapolation into excisions we need the map evaluation in excisions to be possible with domains written to H5 files. I think the solution is to decide whether to allow or reject logical coords outside [-1, 1] after calling the map, not make this decision at construction of the map.

nilsvu (dismissed):

knelli2:
@nilsvu Ok squashed in my fix and added a commit that will actually print a full numpy array if an input file test fails again.

tools/CheckOutputFiles.py
```diff
@@ -115,22 +115,55 @@ def check_h5_file(self, h5_file, test_entity, expected_entity):
...
+                            print(
+                                "DESIRED:"
+                                f" {np.array(expected_data[:, column_mask])}"
+                            )
+                            print(
+                                f"ACTUAL: {np.array(test_data[:, column_mask])}"
+                            )
+                            raise AssertionError(
+                                "Test data is not equal to the expected data"
+                            )
```
nilsvu: Raise the original exception

nilsvu (approved):

PR #6364, opened 2024-11-06
Support filled sphere domain with shape map

nilsvu:
## Proposed changes

Allows to deform a filled sphere domain with a shape map. Used for BNS initial data where we want to deform the sphere to conform to the star surface.

Bildschirmfoto 2024-11-06 um 14 13 35

Here's another sphere domain that's possible now (deformed filled sphere, translated, enveloped by a region where the translation rolls off):

Bildschirmfoto 2024-11-06 um 14 16 44

### Upgrade instructions

When constructing a Sphere domain with time-dependent maps, add the new option `TransitionRotScaleTrans` (boolean). Set it to `True` to keep the old behavior of transitioning the rotation, scaling, and translation maps to zero at the outer boundary (see second image in the PR description, used e.g. to impose boundary conditions in the BBH ringdown domain). Set it to `False` to disable the transition, e.g. to rigidly translate the sphere.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/TimeDependentOptions/Sphere.cpp
+++ b/src/Domain/Creators/TimeDependentOptions/Sphere.cpp
@@ -17,6 +17,7 @@
 #include "Domain/CoordinateMaps/CoordinateMap.tpp"
 #include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp"
 #include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.hpp"
+#include "Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp"
 #include "Domain/Creators/TimeDependentOptions/ShapeMap.hpp"
 #include "Domain/FunctionsOfTime/FixedSpeedCubic.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
@@ -57,17 +58,18 @@ TimeDependentMapOptions::TimeDependentMapOptions(
     const double initial_time, std::optional<ShapeMapOptions> shape_map_options,
     std::optional<RotationMapOptions> rotation_map_options,
     std::optional<ExpansionMapOptions> expansion_map_options,
-    std::optional<TranslationMapOptions> translation_map_options)
+    std::optional<TranslationMapOptions> translation_map_options,
+    const bool transition_rot_scale_trans)
     : initial_time_(initial_time),
       shape_map_options_(std::move(shape_map_options)),
       rotation_map_options_(std::move(rotation_map_options)),
       expansion_map_options_(std::move(expansion_map_options)),
-      translation_map_options_(std::move(translation_map_options)) {}
+      translation_map_options_(std::move(translation_map_options)),
+      transition_rot_scale_trans_(transition_rot_scale_trans) {}
 
 std::unordered_map<std::string,
                    std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>
 TimeDependentMapOptions::create_functions_of_time(
-    const double inner_radius,
     const std::unordered_map<std::string, double>& initial_expiration_times)
     const {
   std::unordered_map<std::string,
@@ -90,7 +92,7 @@ TimeDependentMapOptions::create_functions_of_time(
   if (shape_map_options_.has_value()) {
     auto [shape_funcs, size_funcs] =
         time_dependent_options::initial_shape_and_size_funcs(
-            shape_map_options_.value(), inner_radius);
+            shape_map_options_.value(), deformed_radius_);
 
     // ShapeMap FunctionOfTime
     result[shape_name] =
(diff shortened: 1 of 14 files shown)


tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -308,121 +307,30 @@ void test_sphere_construction(
...
+    const ElementMap<3, Frame::Inertial> element_map{ElementId<3>{block_id},
+                                                     block};
     {
       INFO("Block boundaries are spherical");
       // This section tests if the logical coordinates of corners from all
       // blocks (and points on upper wedge faces) lie on spherical shells
       // specified by inner radius, radial partitions, or outer radius
-      const auto coords_on_spherical_partition = logical_coords(
+      //
+      // First, get logical coordinates of a random block corner
```
knelli2: "get the element logical coordinates"

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
```diff
@@ -183,24 +185,79 @@ TimeDependentMapOptions::create_functions_of_time(
...
-                          shape_name,
-                          size_name};
+        transition_func;
+    if (filled_) {
+      if (transition_rot_scale_trans_ and radial_partitions.size() < 2) {
+        ERROR(
+            "Currently at least two radial partitions are required to "
+            "transition the RotScaleTrans map to zero in the outer shell "
+            "when a shape map is present.");
+      }
```
knelli2: "in the outer shell while a shape map is present and the inner cube is filled". Could also remove "Currently" unless you have plans to change this later?

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
```diff
@@ -183,24 +185,79 @@ TimeDependentMapOptions::create_functions_of_time(
...
+      using WedgeTransition =
+          domain::CoordinateMaps::ShapeMapTransitionFunctions::Wedge;
+      // Shape map transitions from 0 to 1 from the inner cube to this surface
+      deformed_radius_ =
+          radial_partitions.empty() ? outer_radius : radial_partitions.front();
+      // Shape map transitions from 1 to 0 from the deformed surface to the next
+      // radial partition or to the outer radius
+      const bool has_shape_rolloff = not radial_partitions.empty();
+      const double shape_outer_radius =
+          radial_partitions.size() > 1 ? radial_partitions[1] : outer_radius;
```
knelli2, line 212: [optional] Combine these into a `std::optional<double>`

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
```diff
@@ -250,36 +316,35 @@ TimeDependentMapOptions::distorted_to_inertial_map(
...
-    }
-    return std::make_unique<GridToDistortedComposition>(shape_map_.value());
+TimeDependentMapOptions::grid_to_distorted_map(const size_t block_number,
+                                               const bool is_inner_cube) const {
+  const bool block_has_shape_map = shape_map_options_.has_value() and
+                                   block_number < (filled_ ? 12 : 6) and
+                                   not is_inner_cube;
+  if (block_has_shape_map) {
+    return std::make_unique<GridToDistortedComposition>(
+        gsl::at(shape_maps_, filled_ ? block_number : 0));
```
knelli2, line 331: Please add a comment here that when the inner cube isn't filled, then we use the sphere transition which is the same for all wedges and not the wedge transition. Hence, why we grab only the 0th shape map regardless of block. Also below in the grid to inertial map

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
```diff
@@ -213,35 +270,44 @@ void TimeDependentMapOptions::build_maps(
...
-      outer_shell_radii.first,
-      outer_shell_radii.second,
-      domain::CoordinateMaps::TimeDependent::RotScaleTrans<
-          3>::BlockRegion::Transition};
+  if (transition_rot_scale_trans_) {
+    if (radial_partitions.empty()) {
+      ERROR(
+          "Currently at least one radial partition is required to transition "
+          "the RotScaleTrans map to zero in the outer shell.");
+    }
```
knelli2, line 282: There's a combination of options here that is likely to lead to confusion because of boundary conditions:
- Has an excision
- Has a shape map
- Has an expansion or translation map
- `transition_rot_scale_trans_` is false

This combination creates a scenario where the outer boundary is moving significantly and thus the only BCs that will work are analytic ones. Our constraint preserving physical BCs won't work. Not sure we should do anything about this since if we have a filled cube and a shape map, the outer boundary can be distorted and there also you have to have special BCs.

tests/Unit/Domain/Creators/Test_Sphere.cpp
```diff
@@ -577,20 +562,23 @@ void test_sphere(const gsl::not_null<Generator*> gen) {
...
-             : index);
+    auto array_index = (use_hard_coded_time_dep_options and
+                                gsl::at(radial_partitioning, index).empty()
+                            ? index + 1
+                            : index);
+    // If the sphere is filled we need 2 radial partitions to transition the
+    // RotScaleTrans map to zero in the outer shell
+    if (fill_interior) {
+      array_index = 2;
+    }
```
knelli2: But you set the transition to false below, so do you need this?

knelli2 (changes requested):

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
nilsvu: Yea we could change this later if someone needs it

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
nilsvu: Not sure what you think could be combined into an optional

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
nilsvu: Yea I don't think we can really prevent these things. Boundary condition classes could try to warn if they assume a static boundary but it is moving or something like that, not sure.

nilsvu:
Pushed two fixups @knelli2

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
knelli2:
```cpp
const std::optional<double> shape_outer_radius = 
    radial_partitions.empty() ? std::nullopt : 
    (radial_partitions.size() > 1 ? radial_partitions[1] : outer_radius);


for (size_t j=0; j < (shape_outer_radius.has_value() ? 12 : 6); ++j) {
  if (j < 6) {
    ...
  } else {
    // use shape_outer_radius.value()
  }
}
```

knelli2 (commented):
Looks good. If you decide to make this change, you can squash it also

src/Domain/Creators/TimeDependentOptions/Sphere.cpp
nilsvu: Oh I see now, thanks for explaining. I don't think that makes the code much cleaner or more readable, so I'll just keep it as is

knelli2 (approved):
@nilsvu Please add Upgrade Instructions since you added a new option

nilsvu:
Done

Issue #6328, opened 2024-10-09
ID control should run first iteration in a subdirectory


wthrowe:
Having the later iterations as subdirectories of the first iteration is confusing.  It looks like iteration 0 is the final result.

nilsvu:
Yes absolutely. We haven't decided on a directory structure yet.

iago-mendes:
Should this issue be closed because of #6379?

nilsvu:
Yes, thanks for doing this @iago-mendes !

PR #6379, opened 2024-11-15
Update directory structure of ID parameter control

iago-mendes:
## Proposed changes

This PR addresses issue #6328 by moving the 0th initial data run into the subdirectory `ControlParams00`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/ControlId.py
+++ b/support/Pipelines/Bbh/ControlId.py
@@ -149,7 +149,7 @@ def control_id(
     radial_expansion_velocity = binary_data["Expansion"]
 
     # File to write control diagnostic data
-    data_file = open(f"{id_run_dir}/ControlParamsData.txt", "w")
+    data_file = open(f"{id_run_dir}/../ControlParamsData.txt", "w")
 
     iteration = 0
     control_run_dir = id_run_dir
@@ -166,7 +166,7 @@ def control_id(
                 f" Control of BBH Parameters ({iteration}) "
                 "=========================================="
             )
-            control_run_dir = f"{id_run_dir}/ControlParams{iteration:02}"
+            control_run_dir = f"{id_run_dir}/../ControlParams_{iteration:03}"
 
             # Start with initial free data choices and update the ones being
             # controlled in `control_params` with the numeric value from `u`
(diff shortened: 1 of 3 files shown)


support/Pipelines/Bbh/InitialData.py
```diff
@@ -196,6 +198,9 @@ def generate_id(
         )
     if pipeline_dir and not run_dir:
         run_dir = pipeline_dir / "001_InitialData"
+    if control:
+        run_dir = f"{run_dir}/ControlParams00"
```
nilsvu: Please add an underscore for consistency with pipeline steps and segments directory names: `ControlParams_00`

support/Pipelines/Bbh/ControlId.py
```diff
@@ -166,7 +166,7 @@ def Residual(u):
                 f" Control of BBH Parameters ({iteration}) "
                 "=========================================="
             )
-            control_run_dir = f"{id_run_dir}/ControlParams{iteration:02}"
+            control_run_dir = f"{id_run_dir}/../ControlParams{iteration:02}"
```
nilsvu: Also add the underscore here

nilsvu (changes requested):

support/Pipelines/Bbh/ControlId.py
```diff
@@ -149,7 +149,7 @@ def control_id(
     radial_expansion_velocity = binary_data["Expansion"]
 
     # File to write control diagnostic data
-    data_file = open(f"{id_run_dir}/ControlParamsData.txt", "w")
+    data_file = open(f"{id_run_dir}/../ControlParamsData.txt", "w")
```
knelli2, line 152: @nilsvu Should we avoid relative paths and just pass two arguments? 1) The control params directory, 2) the directory that holds all the control params dirs

nilsvu: I think this is fine, it's confined to the control loop with no public interface that can go wrong

support/Pipelines/Bbh/InitialData.py
iago-mendes: I've also changed to 3 0-padding for the same reason.

nilsvu (approved):

PR #6380, opened 2024-11-18
TransformVol.py: add start, stop, stride

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/H5/Python/IterElements.py
+++ b/src/IO/H5/Python/IterElements.py
@@ -138,7 +138,7 @@ def iter_elements(
     """
     if isinstance(volfiles, spectre_h5.H5Vol):
         volfiles = [volfiles]
-    if isinstance(obs_ids, int):
+    if isinstance(obs_ids, (int, np.integer)):
         obs_ids = [obs_ids]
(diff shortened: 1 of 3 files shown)


knelli2 (dismissed):

src/IO/H5/Python/IterElements.py
```diff
@@ -138,7 +138,7 @@ def iter_elements(
     """
     if isinstance(volfiles, spectre_h5.H5Vol):
         volfiles = [volfiles]
-    if isinstance(obs_ids, int):
+    if isinstance(obs_ids, (int, np.integer)):
         obs_ids = [obs_ids]
```
knelli2, line 142: Maybe since you're allowing numpy ints, this should now be
```py
    if isinstance(obs_ids, (int, np.integer)):
        obs_ids = [int(obs_ids)]
```

nilsvu: Should be fixed now. The cast to Python int is not necessary here.

knelli2 (approved):

PR #6381, opened 2024-11-19
Fix machine config files

nilsvu:
## Proposed changes

Make them consistent. These attributes are currently not used.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Python/Machines.py
+++ b/support/Python/Machines.py
@@ -41,10 +41,12 @@ class Machine(yaml.YAMLObject):
         any information that may help people get started using the machine.
         Provide links to wiki pages, signup pages, etc., for additional
         information.
-      DefaultProcsPerNode: Default number of worker threads spawned per node.
+      DefaultTasksPerNode: Default number of tasks per node (MPI ranks).
+        Often chosen to be the number of sockets on a node.
+      DefaultProcsPerTask: Default number of worker threads spawned per task.
         It is often advised to leave one core per node or socket free for
         communication, so this might be the number of cores or hyperthreads
-        per node minus one.
+        per node or socket minus one.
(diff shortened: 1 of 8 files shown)


support/Python/Machines.py
```diff
@@ -41,10 +41,12 @@ class Machine(yaml.YAMLObject):
...
         Provide links to wiki pages, signup pages, etc., for additional
         information.
-      DefaultProcsPerNode: Default number of worker threads spawned per node.
+      DefaultTasksPerNode: Default number of tasks per node (MPI ranks).
+        Often chosen to be the number of sockets on a node.
+      DefaultProcsPerTask: Default number of worker threads spawned per task.
         It is often advised to leave one core per node or socket free for
         communication, so this might be the number of cores or hyperthreads
-        per node minus one.
+        per node or socket minus one.
```
knelli2, line 49: I think we should change this to include the comm core because I envision these variables will be used to represent SLURM values which include *all* cores. We take care of the -1 in the submit script base with `CHARM_PPN=$(expr ${SLURM_CPUS_PER_TASK} - 1)`. Is it reasonable for the average user to be expected to know about the comm core and take it into account?

wthrowe (approved):

support/Python/Machines.py
nilsvu: We don't use these variables at all at the moment, in this PR I'm just fixing an inconsistency. We can decide this when we actually use these variables.

wthrowe: I agree with @knelli2 that we should change it, and with @nilsvu that we shouldn't do it in this PR.

knelli2: Ok we can address this in a follow up when we actually use it

knelli2 (approved):

PR #6387, opened 2024-11-27
Change the error condition for bounds on TOMS748

jyoo1042:
Change from greater than or equal to to strictly greater than.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


wthrowe (approved):
For context, since I was involved in figuring this out: This only matters if the upper and lower bounds are equal and the function evaluated at that value is exactly zero.  We generally avoid calling the root-finder in that case, since we already know the answer, but with the switch to vectorized root-finding it sometimes happens for some values in the vector but not others.

PR #6389, opened 2024-11-28
Logical coordinates for spherical harmonic mesh

kidder:
## Proposed changes

Computes the logical coordinates for a spherical harmonic basis

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/Spectral/CMakeLists.txt
+++ b/src/NumericalAlgorithms/Spectral/CMakeLists.txt
@@ -40,10 +40,12 @@ target_link_libraries(
   DataStructures
   ErrorHandling
   Options
+  Utilities
   PRIVATE
   BLAS::BLAS
   LAPACK::LAPACK
   RootFinding
+  SPHEREPACK
   )
 
 add_subdirectory(Python)
--- a/src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
+++ b/src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
@@ -14,6 +14,7 @@
 #include "NumericalAlgorithms/Spectral/Spectral.hpp"
 #include "Utilities/Gsl.hpp"
 #include "Utilities/SetNumberOfGridPoints.hpp"
+#include "Utilities/Spherepack.hpp"
 
 template <size_t VolumeDim>
 void logical_coordinates(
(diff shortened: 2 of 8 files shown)


src/NumericalAlgorithms/Spectral/CMakeLists.txt
```diff
@@ -40,6 +40,8 @@ target_link_libraries(
   DataStructures
   ErrorHandling
   Options
+  SPHEREPACK
```
nilsvu: Move spherepack to `PRIVATE`

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
```diff
@@ -22,11 +23,63 @@ void logical_coordinates(
...
   for (size_t d = 0; d < VolumeDim; ++d) {
-    const auto& collocation_points_in_this_dim =
-        Spectral::collocation_points(mesh.slice_through(d));
-    for (IndexIterator<VolumeDim> index(mesh.extents()); index; ++index) {
-      logical_coords->get(d)[index.collapsed_index()] =
-          collocation_points_in_this_dim[index()[d]];
+    switch (mesh.basis(d)) {
+      case Spectral::Basis::SphericalHarmonic: {
+        switch (mesh.quadrature(d)) {
+          case Spectral::Quadrature::Gauss: {
```
nilsvu, line 29: Add a comment and a note in the docs that the theta direction is identified by Gauss quadrature and the phi direction is identified by equiangular.

Also add notes that logical coordinates for the spherical harmonic basis will be already in spherical coordinates, not in [-1, 1], presumably because we never need the [-1, 1] coordinates and spherepack works directly with the collocation points in spherical coordinates.

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
```diff
@@ -22,11 +23,63 @@ void logical_coordinates(
...
+        }
+        break;
+      }
+      case Spectral::Basis::Chebyshev:
+        // fall through
+      case Spectral::Basis::Legendre:
+        // fall through
+      case Spectral::Basis::FiniteDifference: {
+        const auto& collocation_points_in_this_dim =
+            Spectral::collocation_points(mesh.slice_through(d));
```
nilsvu, line 75: Move the computation of the collocation points one level deeper if possible, into `Spectral::collocation_points`. That way you also get the static caching for free.

nilsvu (changes requested):

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
```diff
@@ -22,11 +23,63 @@ void logical_coordinates(
...
+            ERROR(
+                "Quadrature must be Gauss or Equiangular for Basis "
+                "SphericalHarmonic");
+        }
+        break;
+      }
+      case Spectral::Basis::Chebyshev:
+        // fall through
+      case Spectral::Basis::Legendre:
+        // fall through
```
nilsdeppe: https://en.cppreference.com/w/cpp/language/attributes/fallthrough

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
nilsdeppe: The basis functions are not defined on -1 to 1. It's only the standard Jacobi polynomials that are defined on this interval. There are basis functions defined on `0, 1`, `0, infty`, `-infty, 0`, `-infty, infty`, `0, pi`, etc. It's just that Jacobi polynomials happen to be orthonormal on `[-1, 1]`, that's all. There's nothing generally special about those two numbers. This is just how basis functions work. I have nothing against stating the domain of the basis functions (we should do so for all of them, then!), but we shouldn't add any language implying there's something "weird" because they aren't on `-1, 1` and that we magically "hid" the `-1, 1` coordinates. That's not what's happening, there are no `-1, 1` coordinates for all basis functions.

nilsvu: Sounds good, there is a relation to Jacobi polynomials though since we use Legendre-Gauss points in cos(theta). Anyway, just stating the domain of the logical coords seems useful, and stating what collocation points are used for the ylm basis.

src/NumericalAlgorithms/Spectral/CMakeLists.txt
kidder: done

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
kidder: I expanded the documentation.

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
kidder: While it is possible, I don't think it makes sense to do so.  There are currently ten functions that use the macro `SPECTRAL_QUANTITY_FOR_MESH` and only `collocation_points` would make sense with Basis::SphericalHarmonic.  So since the branching is unique, why bury it and then require error handling for all the other functions?  Furthermore `collocation_points` calls `compute_collocation_points_and_weights` to get the collcation points, and I don't think there are any useful weights

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
kidder: done

nilsdeppe (dismissed):

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
nilsvu: Ok, just two concerns: (1) it would be nice to get the static caching (not sure how important it actually is in practice for collocation points, certainly more important for diff matrices). (2) When calling `Spectral::collocation_points` it gives a "missing basis" error, but `logical_coordinates` with a 1D Ylm mesh will happily return the points. Moving the collocations points a level deeper would fix this inconsistency (I don't think it's a problem to move the missing-basis error from `get_spectral_quantity_for_mesh` a level deeper into the diff matrices etc).

nilsdeppe: But then, as Larry said, all the other functions would still be inconsistent. I don't see any benefit to having the inconsistencies be in function defined in `Spectral.hpp`, I find it easier to have the difference in behavior at the `logical_coordinates` (which is not in `Spectral.hpp`). We should probably rename `Spectral.hpp` to `JacobiPolynomials.hpp` or `Basis1d.hpp` or something. `Spectral.hpp` is not a very clear name and is now clearly different from our 2d basis functions. I think the current implementation is the right avenue.
I'm not worried about caching of the logical coordinates at this point, I'm more concerned about getting the benefits of spherical shells into the code.

nilsvu: Ok, I don't think it would be "inconsistent" to have the subset of functions in Spectral.hpp without Ylm implementation error out if a Ylm basis is requested. They already error when a Ylm basis is requested. Only `Spectral::collocation_points` is inconsistent because it errors with "missing basis" for a Ylm basis although `logical_coordinates` will return the collocation points just fine.

Anyway, the current implementation is fine with me. Maybe add a better error message to `Spectral::collocation_points` when a Ylm basis is requested or something like that.

nilsvu (dismissed):

src/NumericalAlgorithms/Spectral/LogicalCoordinates.cpp
nilsdeppe: Ah, okay. I see. Yes, a better error message seems very reasonable. @kidder would you mind doing that?

nilsvu (approved):

PR #6392, opened 2024-12-02
Support spherical shell blocks, create Sphere domain

nilsvu:
## Proposed changes

Add an enum to `Block` that indicates whether the block represents a `Cube` or a `SphericalShell`. Cubes can only have one block neighbor per dimension, whereas spherical shells can have multiple block neighbors (another shell, or 4 wedges in 2D, or 6 wedges in 3D).
Also adds the code to connect element neighbors across spherical and cubed shells, to create the Ylm meshes, and to create a Sphere domain with spherical shells.
Only the spherical-to-cartesian coordinate map is missing to create spherical shells in ExportCoordinates3D. The coordinate maps are in #5588, but need to be adjusted a bit because we use the spherical coordinates directly as logical coordinates, so we don't need the maps from [-1, 1] to angles.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
@kidder this is to share this code with you. I have cleaned up the first 3 commits. The others need some more cleanup. I can do that and add some more tests. <s>The last commit e4c2593 will be superseded by your #6389 .</s>

nilsdeppe:
@nilsvu can we close this?

PR #6393, opened 2024-12-02
Add transformation to tortoise radius

nilsvu:
## Proposed changes

Bundles some root-finding code from https://github.com/tosburn3/KerrGeodesicsC to transform between Boyer-Lindquist and tortoise radius. The bundled code has figured out good initial guesses and a robust method for the root find.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/RootFinding/TOMS748.hpp
+++ b/src/NumericalAlgorithms/RootFinding/TOMS748.hpp
@@ -14,6 +14,7 @@
 #include <type_traits>
 
 #include "DataStructures/DataVector.hpp"
+#include "Utilities/ErrorHandling/CaptureForError.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/ErrorHandling/Exceptions.hpp"
 #include "Utilities/GetOutput.hpp"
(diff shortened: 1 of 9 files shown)


wthrowe:
I think this is a license violation.  The GPL would let us distribute a verbatim copy of the upstream package, but I don't think we can distribute a version that has been modified like this as part of SpECTRE.

nilsvu:
My understanding is that this is a redistribution of the GPL-licensed code with some changes still under the GPL license, which is ok. The changes are: Deleted some functions to avoid dependence on FFTW, added a CMakeLists.txt to compile the code with CMake, added a C++ compatibility header, and auto-formatted the code. These are listed in the README, and the license is also included.

wthrowe:
The GPL does not give you permission to do that.  You must license the entire work as GPL if you include modified GPL code, not just your changes.  Since we're not relicensing SpECTRE, your options are:
1. Use the unmodified GPL code, so the modification clauses do not come into effect,
2. Convince the author to relicense, or
3. Use TOMS748 instead of including an external package just to do a 1d root find.

wthrowe (approved):
Looks good.

Noting in case someone later wants to do bounds optimization: For large radius, good bounds look to be `0 < r - (r* - 2 M log(r*/(2 M))) < const ~ 1.5 M`.  I didn't experiment with small radius.

PR #6395, opened 2024-12-04
Fix shape map loaded from file

nilsvu:
## Proposed changes

The size of the domain didn't match the Strahlkorper in the file because the loaded l=0, m=0 coefficient was interpreted as a deformation/scale. Fixed by subtracting the size of the original sphere.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/TimeDependentOptions/ShapeMap.hpp
+++ b/src/Domain/Creators/TimeDependentOptions/ShapeMap.hpp
@@ -217,7 +217,9 @@ struct ShapeMapOptions {
     static constexpr Options::String help = {
         "Initial value and two derivatives of the 00 coefficient. Specify "
         "'Auto' to use the 00 coefficient specified in the 'InitialValues' "
-        "option."};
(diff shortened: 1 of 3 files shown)


src/Domain/Creators/TimeDependentOptions/ShapeMap.hpp
```diff
@@ -262,5 +262,5 @@ template <bool IncludeTransitionEndsAtCube, domain::ObjectLabel Object>
 std::pair<std::array<DataVector, 3>, std::array<DataVector, 4>>
 initial_shape_and_size_funcs(
```
knelli2, line 265: Can't reach with GitHub comments, but we should note somewhere in some help text that with this change, the deformed radius will be exactly what is in the YlmsFromFile, and not what is set for the inner/outer radius in the input file.

knelli2 (changes requested):
Tests need an iterable approx. You can squash everything

nilsvu:
done

knelli2 (approved):

PR #6399, opened 2024-12-05
Compute ADM angular momentum integral in BBH ID

iago-mendes:
## Proposed changes

This PR implements the $J_{ADM}$ integral based on [arXiv:1506.01689](https://arxiv.org/abs/1506.01689). We only compute the z-component because that's the quantity of interest to control hyperbolic encounters. Since the $J_{ADM}$ integrand is closely related to the $P_{ADM}$ integrand, I've put both pointwise functions together, renaming `AdmLinearMomentum` to `AdmMomentum`.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
+++ b/tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
@@ -24,7 +24,7 @@
 #include "PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp"
 #include "PointwiseFunctions/AnalyticSolutions/Xcts/Schwarzschild.hpp"
 #include "PointwiseFunctions/AnalyticSolutions/Xcts/WrappedGr.hpp"
-#include "PointwiseFunctions/Xcts/AdmLinearMomentum.hpp"
+#include "PointwiseFunctions/Xcts/AdmMomentum.hpp"
 
 namespace {
 
@@ -32,15 +32,18 @@ using Schwarzschild = Xcts::Solutions::Schwarzschild;
 using KerrSchild = Xcts::Solutions::WrappedGr<gr::Solutions::KerrSchild>;
 
 template <typename Solution>
-void test_infinite_surface_integral(const double distance, const double mass,
-                                    const double horizon_radius,
-                                    const double boost_speed,
-                                    const Solution& solution) {
+void test_infinite_surface_integrals(const double distance, const double mass,
+                                     const double horizon_radius,
+                                     const double boost_speed,
+                                     const Solution& solution) {
+  // Define x-shift in the domain
+  const double x_shift = 1.;
+
   // Set up domain
   const size_t h_refinement = 1;
   const size_t p_refinement = 6;
   const domain::creators::Sphere shell{
-      /* inner_radius */ horizon_radius,
+      /* inner_radius */ horizon_radius + x_shift,
       /* outer_radius */ distance,
       /* interior */ domain::creators::Sphere::Excision{},
       /* initial_refinement */ h_refinement,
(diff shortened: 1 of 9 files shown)


tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
```diff
@@ -32,15 +32,18 @@ using Schwarzschild = Xcts::Solutions::Schwarzschild;
...
+                                     const Solution& solution) {
+  // Define x-shift in the domain
+  const double x_shift = 1.;
+
   // Set up domain
   const size_t h_refinement = 1;
   const size_t p_refinement = 6;
   const domain::creators::Sphere shell{
-      /* inner_radius */ horizon_radius,
+      /* inner_radius */ horizon_radius + x_shift,
```
nilsvu, line 46: This will only increase the inner radius, not shift coordinates

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
```diff
@@ -128,39 +136,57 @@ void test_infinite_surface_integral(const double distance, const double mass,
...
                                    const Solution& solution) {
+  // Define x-shift in the domain
+  const double x_shift = 1.;
+
   // Set up domain
   const size_t h_refinement = 1;
   const size_t p_refinement = 6;
   const domain::creators::Sphere shell{
-      /* inner_radius */ 2 * horizon_radius,
+      /* inner_radius */ horizon_radius + x_shift,
```
nilsvu, line 189: same here

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
```diff
@@ -249,37 +280,56 @@ void test_infinite_volume_integral(const double distance, const double mass,
...
 
       // Get interface coordinates.
-      const auto face_logical_coords =
+      auto face_logical_coords =
           interface_logical_coordinates(face_mesh, boundary_direction);
       const auto face_inv_jacobian =
           logical_to_inertial_map.inv_jacobian(face_logical_coords);
 
+      // Shift x-coordinate
+      face_logical_coords.get(0) -= x_shift;
```
nilsvu: Logical coordinates are [-1, 1] on the reference cube, you shouldn't shift them. The shift is part of the map from logical to inertial coords. I think you can just delete this line, because you don't need the face coordinates below.

nilsvu (changes requested):

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
iago-mendes: This is increasing the domain inner radius so that we don't use an isotropic Schwarzschild solution inside the horizon. Line 99 below is responsible for shifting the coordinates.

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
iago-mendes: Same as above. Line 230 is responsible for shifting the coordinates.

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
iago-mendes: Indeed, this was a typo and shouldn't be here. Thanks for catching that!

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
nilsvu: Ok I see now how you need this to evaluate the solution at the shifted coordinates.

nilsvu:
You can squash

tests/Unit/PointwiseFunctions/Xcts/Test_AdmMomentum.cpp
nilsvu: Just noting here that you could also do the reverse: evaluate the solution at the unshifted coordinates and use the shifted coords to compute J_ADM.

nilsvu (approved):

PR #6400, opened 2024-12-09
Add EffectiveSource code

nilsvu:
## Proposed changes

Bundles the EffectiveSource (https://github.com/barrywardell/EffectiveSource) code in external/ (<s>GPL</s> MIT licensed). This is a widely used code in the self-force community to compute the effective source of a scalar point charge in Kerr spacetime.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


kidder:
why are we putting this into spectre and not just using it as an optional external dependency?

nilsvu:
It won't ever change and is easier to bundle than to install as an extra dependency (in a recent effort to make the code easier to use we have been bundling other dependencies like libsharp as well). It's one .c file, and the original repo doesn't even have a build system, so it's easier to bundle the .c file.

nilsdeppe:
Can you ask Barry if he'd be willing to relicense to something more permissive so we don't have to worry about GPL things again?

nilsvu:
Ok I'll ask!

nilsvu:
The code is licensed under MIT now 👍 I updated this PR

external/EffectiveSource/CMakeLists.txt
```diff
@@ -0,0 +1,11 @@
+set(LIBRARY EffectiveSource)
+
+add_library(${LIBRARY} kerr-circular.c)
+
+target_link_libraries(
+  ${LIBRARY}
+  PRIVATE
+  GSL::gsl
+  )
```
nilsdeppe, line 9: Won't matter until we get other dependencies worked out so tracking doesn't tell us we are broken. We could add the header files to track those as well. Up to you :)

nilsdeppe (dismissed):
LGTM! Thanks for working with Barry to get the license changed!

external/EffectiveSource/kerr-circular.c
```diff
@@ -0,0 +1,848 @@
...
+ ******************************************************************************/
+
+#include <math.h>
+#include <assert.h>
+#include "effsource.h"
+#include <stdio.h>
+#include <gsl/gsl_sf_ellint.h>
+
+/* The particle's coordinate location and 4-velocity */
+static struct coordinate xp;
```
nilsdeppe, line 12: Just an FYI, I'm worried all of this code isn't threadsafe., but I haven't looked in detail and I don't know if it matters :)

nilsvu: yep probably not threadsafe, I've been working around that when calling the functions

external/EffectiveSource/CMakeLists.txt
nilsvu: ok sure added the `add_interface_lib_headers`

nilsdeppe: Oh, I think there are still some bugs in how to do this (I have another repo where I fixed it and need to backport). Just remove it for now, sorry!

nilsdeppe (approved):

PR #6401, opened 2024-12-10
Allow Singletons to use `ExecutePhaseChange`

nikwit:
## Proposed changes

This PR allows singletons to contribute to reduction data by calling `ExecutePhaseChange`. I needed this to checkpoint the worldtube singleton.

I duplicate the algorithm unit test for this action it to show that it works with singletons now as well.

### DIFF
--- a/src/Parallel/PhaseControl/ContributeToPhaseChangeReduction.hpp
+++ b/src/Parallel/PhaseControl/ContributeToPhaseChangeReduction.hpp
@@ -72,23 +72,27 @@ void contribute_to_phase_change_reduction(
                               PhaseControl::TaggedTupleCombine, Ts...>(nullptr),
                       cache.get_main_proxy().value());
   reduction_data_type reduction_data{data_for_reduction};
-  // Note that Singletons could be supported by directly calling the main
-  // entry function, but due to this and other peculiarities with
-  // Singletons, it is best to discourage their use.
-  static_assert(
-      not std::is_same_v<typename SenderComponent::chare_type,
-                         Parallel::Algorithms::Singleton>,
-      "Phase change reduction is not supported for singleton chares. "
-      "Consider constructing your chare as a length-1 array chare if you "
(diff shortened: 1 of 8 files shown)


tests/Unit/Parallel/Test_AlgorithmPhaseControlSingleton.yaml
```diff
@@ -0,0 +1,19 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+---
+---
+
+# note that this is a vector of pairs, so has the peculiar '- -' for the
+# elements
```
knelli2: This comment is outdated from an old syntax for PhaseChangeAndTriggers.

src/Parallel/PhaseControl/ContributeToPhaseChangeReduction.hpp
```diff
@@ -72,23 +72,27 @@ void contribute_to_phase_change_reduction(
...
-                   reduction_data.packed().get(),
-                   Parallel::charmxx::charm_reducer_functions.at(
-                       std::hash<Parallel::charmxx::ReducerFunctions>{}(
-                           &reduction_data_type::combine)),
-                   callback);
+
+  if constexpr (std::is_same_v<typename SenderComponent::chare_type,
+                               Parallel::Algorithms::Singleton>) {
+    Parallel::local(Parallel::get_parallel_component<SenderComponent>(cache))
+        ->contribute(static_cast<int>(reduction_data.size()),
```
knelli2, line 79: Note that technically this could have been put into the other `contribute_to_phase_change_reduction` function that already uses `Parallel::local`, but then every phase change would have to be updated to call the other function for both arrays and singletons. I think this current way is cleaner.

knelli2 (changes requested):
You can squash everything into just one commit

knelli2 (approved):

PR #6404, opened 2024-12-10
Add compute tags for constraint gammas in worldtube evolution

nikwit:
## Proposed changes

A stable worldtube evolution requires the profile of gamma2 to track the position of the particle. This PR includes this with a compute tag.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
@@ -839,5 +839,40 @@ struct dtPsi0 : db::SimpleTag {
   using type = Scalar<DataVector>;
 };
 
+/*!
+ * \brief Sets Gamma1 to zero throughout the domain. The equations are given in
+ * Initialization::InitializeConstraintDampingGammas.
+ */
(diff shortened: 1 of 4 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -839,5 +839,38 @@ struct dtPsi0 : db::SimpleTag {
...
+  static void function(gsl::not_null<Scalar<DataVector>*> gamma1,
+                       const tnsr::I<DataVector, Dim, Frame::Inertial>& coords);
+};
+
+/*!
+ * \brief Sets Gamma2 to a Gaussian that falls off to a constant value centered
+ * on the position of the particle. This was found to be necessary for a stable
+ * evolution.
+ */
+struct ConstraintGamma2Compute : CurvedScalarWave::Tags::ConstraintGamma2,
```
knelli2: Please update the dox for both these compute tags to point to the `InitializeConstraintDampingGammas` mutator which has the exact equations and values that are used.

knelli2 (changes requested):
Squash immediately

knelli2 (approved):

PR #6407, opened 2024-12-11
Drop modifications to libsharp

wthrowe:
I do not believe we are allowed to integrate modified GPL code into
the SpECTRE source.  There may be an exception if the modified code
could still be used separately, but the changes we made to libsharp
did not allow that since we replaced its build system with SpECTRE's.

This replaces the contents of the external/libsharp directory with the
unmodified source, and builds it as an external project.

Note to reviewers: in the last commit, the non-libsharp files modified are `docs/Installation/Installation.md` and `external/CMakeLists.txt`.

### Upgrade instructions

GNU Make is now a required dependency.  Using Ninja to build SpECTRE will continue to work, but it will internally use make to build the bundled libsharp libraries.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (approved):
LGTM! Thanks for doing this!!

PR #6409, opened 2024-12-11
Add `InsideHorizon` trigger for worldtube excision

nikwit:
## Proposed changes

Add the `InsideHorizon` trigger, that indicates when the worldtube is entirely contained within the event horizon of the central black hole.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.cpp
@@ -1,0 +1,31 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.hpp"
+
+#include <array>
+
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/RadiusFunctions.hpp"
+
+namespace Triggers {
+
+bool InsideHorizon::operator()(
+    const std::array<tnsr::I<double, Dim, Frame::Inertial>, 2>&
+        position_and_velocity,
+    const std::array<double, 4>& worldtube_radius_params) const {
+  const double orbit_radius = get(magnitude(position_and_velocity[0]));
+  // optimization that will almost always be true
+  if (LIKELY(orbit_radius > 1.99)) {
+    return false;
+  }
+  const double wt_radius_inertial =
+      CurvedScalarWave::Worldtube::smooth_broken_power_law(
+          orbit_radius, worldtube_radius_params[0], worldtube_radius_params[1],
+          worldtube_radius_params[2], worldtube_radius_params[3]);
+  return orbit_radius + wt_radius_inertial < 1.99;
+}
+
+PUP::able::PUP_ID InsideHorizon::my_PUP_ID = 0;  // NOLINT
+}  // namespace Triggers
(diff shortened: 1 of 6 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.hpp
```diff
@@ -0,0 +1,44 @@
...
+
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "ParallelAlgorithms/EventsAndTriggers/Trigger.hpp"
+#include "Utilities/Serialization/CharmPupable.hpp"
+
+namespace Triggers {
+
+/*!
+ * \brief This trigger is true when the worldtube is entirely within a
+ * coordinate sphere of radius 1.99 M.
```
knelli2: Add "centered on the origin in the inertial frame"

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.hpp
```diff
@@ -0,0 +1,44 @@
...
+  /// \cond
+  InsideHorizon() = default;
+  explicit InsideHorizon(CkMigrateMessage* /*unused*/) {}
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(InsideHorizon);  // NOLINT
+  /// \endcond
+
+  static constexpr Options::String help =
+      "Triggers if the worldtube is entirely inside a coordinate sphere with "
+      "radius 1.99 M.";
```
knelli2: Add "centered on the origin in the inertial frame".

Also, should 1.99 be an option in the input file? Does this depend on the mass of the larger black hole always being 1M?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.cpp
```diff
@@ -0,0 +1,31 @@
...
+
+bool InsideHorizon::operator()(
+    const std::array<tnsr::I<double, Dim, Frame::Inertial>, 2>&
+        position_and_velocity,
+    const std::array<double, 4>& worldtube_radius_params) const {
+  const double orbit_radius = get(magnitude(position_and_velocity[0]));
+  // optimization that will almost always be true
+  if (orbit_radius > 1.99) {
+    return false;
+  }
```
knelli2:
```cpp
if (LIKELY(orbit_radius > 1.99)) {
  return false;
}
```

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.cpp
```diff
@@ -0,0 +1,31 @@
...
+  const double wt_radius_inertial =
+      CurvedScalarWave::Worldtube::smooth_broken_power_law(
+          orbit_radius, worldtube_radius_params[0], worldtube_radius_params[1],
+          worldtube_radius_params[2], worldtube_radius_params[3]);
+  return orbit_radius + wt_radius_inertial < 1.99 ? true : false;
```
knelli2:
```cpp
return orbit_radius + wt_radius_inertial < 1.99
```

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.cpp
```diff
@@ -0,0 +1,31 @@
...
+}
+
+PUP::able::PUP_ID InsideHorizon::my_PUP_ID = 0;
```
knelli2: `//  NOLINT`

knelli2 (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/InsideHorizon.hpp
nikwit: This does assume that the central black hole has mass 1 but that is already assumed in a few places in the code. I am hesitant to make this an input file choice as only a very small range of radii make sense here and I found 1.99 M to be a very robust choice.

knelli2 (approved):
Looks good. Squash

PR #6412, opened 2024-12-12
Add pybindings for Spherepack

nikwit:
## Proposed changes

Adds some pybindings for `Spherepack`, mostly `phys_to_spec` and `spec_to_phys`. I also added bindings for the `SpherpackIterator` to write a sensible unit test comparing to scipy.

### DIFF
--- a/src/NumericalAlgorithms/SphericalHarmonics/Python/Spherepack.cpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/Python/Spherepack.cpp
@@ -1,0 +1,65 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "NumericalAlgorithms/SphericalHarmonics/Python/Spherepack.hpp"
+
+#include <array>
+#include <cstddef>
+#include <pybind11/operators.h>
+#include <pybind11/pybind11.h>
+#include <pybind11/stl.h>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/ModalVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/Spherepack.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/SpherepackIterator.hpp"
+
+namespace py = pybind11;
+
+namespace ylm::py_bindings {
+void bind_spherepack(pybind11::module& m) {  // NOLINT
+  py::class_<ylm::Spherepack>(m, "Spherepack")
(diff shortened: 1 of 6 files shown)


src/NumericalAlgorithms/SphericalHarmonics/Python/Spherepack.cpp
```diff
@@ -0,0 +1,65 @@
...
+                           const>(&ylm::Spherepack::phys_to_spec),
+           py::arg("collocation_values"), py::arg("physical_stride") = 1,
+           py::arg("physical_offset") = 0)
+      .def("spec_to_phys",
+           static_cast<DataVector (ylm::Spherepack::*)(
+               const DataVector& spectral_coefs, const size_t spectral_stride,
+               const size_t spectral_offset) const>(
+               &ylm::Spherepack::spec_to_phys),
+           py::arg("collocation_values"), py::arg("physical_stride") = 1,
+           py::arg("physical_offset") = 0)
```
nilsvu: fix argument names

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Python/Test_Spherepack.py
```diff
@@ -0,0 +1,75 @@
...
+import numpy.testing as npt
+from scipy.special import sph_harm
+
+from spectre.SphericalHarmonics import Spherepack, SpherepackIterator
+
+
+def convert_coef_to_spherepack(complex_coef, l, m):
+    sign = 1.0 if m % 2 == 0 else -1.0
+    part = np.real(complex_coef) if m >= 0 else np.imag(complex_coef)
+    return 0.5 * sign * np.sqrt(2.0 / np.pi) * part
```
nilsvu: Is this conversion documented somewhere? If so, could you link to those docs here?

Edit: Looks like this is documented at https://spectre-code.org/classylm_1_1Spherepack.html, except I don't see the factor of 0.5.

nilsvu (changes requested):
nice!

tests/Unit/NumericalAlgorithms/SphericalHarmonics/Python/Test_Spherepack.py
nikwit: yes, I suspect I am counting double by just taking the real part of the complex harmonic. I will check that and update.

nikwit: I was indeed and corrected it in a fixup

nilsvu (approved):

PR #6415, opened 2024-12-13
Use git-clang-format corresponding to clang-format

wthrowe:
Might not be compatible with git-clang-format version.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tools/Hooks/pre-commit.sh
+++ b/tools/Hooks/pre-commit.sh
@@ -33,10 +33,10 @@ printf '%s\0' "${commit_files[@]}" | run_checks "${standard_checks[@]}"
 
 ###############################################################################
 # Use git-clang-format to check for any suspicious formatting of code.
-@CLANG_FORMAT_BIN@ --version > /dev/null
+@GIT_CLANG_FORMAT_BIN@ --help > /dev/null
 if [ $? -eq 0 ]; then
-    clang_format_diff=$(@GIT_EXECUTABLE@ --no-pager \
-        clang-format --binary @CLANG_FORMAT_BIN@ --diff --quiet)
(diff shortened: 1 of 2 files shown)


tools/Hooks/pre-commit.sh
```diff
@@ -33,10 +33,9 @@ printf '%s\0' "${commit_files[@]}" | run_checks "${standard_checks[@]}"
 
 ###############################################################################
 # Use git-clang-format to check for any suspicious formatting of code.
-@CLANG_FORMAT_BIN@ --version > /dev/null
+@GIT_EXECUTABLE@ clang-format -h > /dev/null
 if [ $? -eq 0 ]; then
-    clang_format_diff=$(@GIT_EXECUTABLE@ --no-pager \
-        clang-format --binary @CLANG_FORMAT_BIN@ --diff --quiet)
+    clang_format_diff=$(@GIT_EXECUTABLE@ --no-pager clang-format --diff --quiet)
```
nilsvu: This will choose whichever clang-format binary is in the `PATH`, which can change depending on the user's environment. Instead, the current code finds clang-format at cmake config time and always uses that. If the problem is that `git clang-format` will find whichever `git-clang-format` is in the user environment's `PATH`, then I think the more robust solution is to find `git-clang-format` alongside `clang-format` in `FindClangFormat.cmake` and use it here, to make sure the two are compatible.

wthrowe: Good idea.  Done.  (Complete rewrite of the PR, so no fixups.)

nilsvu (approved):

PR #6425, opened 2024-12-28
Improve NVCC & Kokkos support, test on CI

nilsvu:
## Proposed changes

A few things that allowed me to compile with Kokkos.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


cmake/SetupKokkos.cmake
```diff
@@ -43,24 +43,26 @@ if(SPECTRE_KOKKOS)
     FetchContent_MakeAvailable(Kokkos)
   endif()
 
-  if (TARGET Kokkos::kokkos
-      AND Kokkos_ENABLE_CUDA
+  if (Kokkos_ENABLE_CUDA
       AND CMAKE_CXX_COMPILER_ID STREQUAL "GNU")
-    set_property(TARGET Kokkos::kokkos
+    # Not using the 'Kokkos::kokkos' target because that's an alias to 'kokkos',
+    # and we can't set properties on alias targets
```
nilsdeppe: Do you know if this is kokkos version dependent? Because this had worked for me previously. We might need to pin to kokkos versions like we do for blaze anyway. I'm not asking you to change anything, just curious if you know. "No" is totally fine :)

tests/Unit/Utilities/Test_TMPLDocumentation.cpp
```diff
@@ -575,41 +575,45 @@ assert_same<tmpl::reverse_range<size_t, 7, 7>, tmpl::list<>>();
...
+static_assert(tmpl::all<
+  List1<tmpl::size_t<0>, tmpl::size_t<1>, tmpl::size_t<0>>,
+  tmpl::less<tmpl::_1, tmpl::size_t<2>>>::value);
+static_assert(tmpl::all<List1<>>::value);
 // [tmpl::all]
 
 // [tmpl::all:inhomogeneous]
-assert_same<tmpl::all<List1<std::true_type, tmpl::true_type>, tmpl::_1>,
-            tmpl::false_type>();
+#ifndef __CUDACC__
```
nilsdeppe, line 592: Do you know why some of these fail with nvcc? It looks like a lot of them work from the any section. I'm worried we use these features and nvcc fails.

Would it make sense to move the ifndef outside the doxygen snippet statements?

cmake/SetupKokkos.cmake
```diff
@@ -62,4 +62,18 @@ if(SPECTRE_KOKKOS)
       )
```
nilsdeppe: Remove "fixup disable another test" from commit message?

tests/Unit/DataStructures/CMakeLists.txt
```diff
@@ -47,6 +46,11 @@ set(LIBRARY_SOURCES
   Test_VectorImplTestHelper.cpp
   )
 
+# The following test is disabled for nvcc because nvcc can't handle it.
```
nilsdeppe: maybe specify what fails? E.g. it just runs out of RAM?

tests/Unit/Evolution/Systems/GrMhd/GhValenciaDivClean/CMakeLists.txt
```diff
@@ -30,6 +29,11 @@ set(LIBRARY_SOURCES
   Test_TimeDerivativeTerms.cpp
   )
 
+# The following test is disabled for nvcc because nvcc can't handle it.
```
nilsdeppe: maybe specify what fails? E.g. it just runs out of RAM?

nilsdeppe (changes requested):
LGTM, just a few questions!

cmake/SetupKokkos.cmake
nilsvu: Looks like the `Kokkos::kokkos` target can either be an alias or not, depending on how it was installed (pointing to source tree vs installed on system), so I think we have to support both. I'm thinking, can we just remove these two `set_property` calls? The first I already moved to `EnableWarnings.cmake` anyway, and the second (the one with `-Wno-unknown-cuda-version`) we could also either move to `EnableWarnings.cmake` or delete altogether. The mapping of clang to cuda versions is not so straightforward anyway, so we might as well emit the warning if the versions are marked incompatible. What do you think?

tests/Unit/Utilities/Test_TMPLDocumentation.cpp
nilsvu: Yes I know what's happening here. Brigand has a special code branch for nvcc (here: https://github.com/sxs-collaboration/spectre/blob/develop/external/brigand/include/brigand/brigand.hpp#L951). Unfortunately it behaves slightly different to the standard code branch, in that it returns `std::true_type` where the standard branch returns `tmpl::true_type`. Therefore I put a warning in the docs saying the behavior may depend on the compiler. As long as we always use `::value` this should not matter, and I checked and found that we never use `tmpl::true_type` or `tmpl::false_type` in the code, so we should be fine.

nilsvu: Note that I also tried to remove the special code branch in Brigand for nvcc and it compiled fine, so maybe this was only necessary in Brigand for an old version of nvcc. But it was easier to modify our tests since we never rely on `std::true_type` vs `tmpl::true_type` in our code anyway, than to change upstream Brigand or patch it.

nilsvu: I kept the ifndef inside the snippet to emphasize in the docs that it can be compiler-dependent

nilsdeppe (dismissed):
LGTM! Please squash. Thanks for doing this!!

nilsdeppe (approved):

PR #6429, opened 2025-01-04
Better domain errors, atmosphere density in ID, and velocity limiting in atmosphere.

nilsdeppe:
## Proposed changes

Various changes that make BNS simulations more robust. More will be added in future PRs as I write tests for the code.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Evolution/Systems/GrMhd/ValenciaDivClean/KastaunEtAlHydro.tpp
```diff
@@ -39,9 +39,8 @@ class FunctionOfZ {
               const double electron_fraction, const EosType& equation_of_state,
               const double lorentz_max)
       : q_(tau / rest_mass_density_times_lorentz_factor),
-        r_squared_(momentum_density_squared /
-                   square(rest_mass_density_times_lorentz_factor)),
-        r_(std::sqrt(r_squared_)),
+        r_(std::sqrt(momentum_density_squared /
```
kidder, line 42: why sqrt a square?

src/Domain/DomainHelpers.cpp
```diff
@@ -349,8 +349,9 @@ void set_cartesian_periodic_boundaries(
     const gsl::not_null<
         std::vector<DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>>>*>
         neighbors_of_all_blocks) {
-  ASSERT(orientations_of_all_blocks.size() == corners_of_all_blocks.size(),
-         "Each block must have an OrientationMap relative to an edifice.");
+  if (orientations_of_all_blocks.size() != corners_of_all_blocks.size()) {
+    ERROR("Each block must have an OrientationMap relative to an edifice.");
```
kidder, line 353: what is an edifice?

kidder (changes requested):
you can squash any changes

src/Evolution/Systems/GrMhd/ValenciaDivClean/KastaunEtAlHydro.tpp
nilsdeppe: I agree, it shouldn't be needed, but might help with robustness against negative evolved vars, so I would prefer to keep it to match the previous algorithm...

src/Domain/DomainHelpers.cpp
nilsdeppe: I have no idea, sorry. You'll have to ask the original author, which looks like is Marcie based on the git log.

kidder (approved):

PR #6430, opened 2025-01-07
Allow choice of reading Pi and Phi from numeric ID

knelli2:
Or setting them based on the 1- and 3-index constraints. Only works for numeric ID from GH vars. Numeric ID from ADM vars still computes phi with numerical derivatives.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/Actions/SetInitialData.hpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/Actions/SetInitialData.hpp
@@ -9,11 +9,13 @@
 #include <variant>
 
 #include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/Tag.hpp"
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Domain/Structure/ElementId.hpp"
 #include "Domain/Tags.hpp"
 #include "Evolution/Initialization/InitialData.hpp"
+#include "Evolution/Systems/GeneralizedHarmonic/GaugeSourceFunctions/SetPiAndPhiFromConstraints.hpp"
 #include "Evolution/Systems/GeneralizedHarmonic/Tags.hpp"
 #include "IO/Importers/Actions/ReadVolumeData.hpp"
 #include "IO/Importers/ElementDataReader.hpp"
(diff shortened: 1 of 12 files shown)


support/Pipelines/Bbh/Ringdown.yaml
```diff
@@ -19,9 +19,11 @@ InitialData:
     ObservationValue: &InitialTime "{{ MatchTime }}"
     ObservationValueEpsilon: Auto
     ElementsAreIdentical: False
+    SetPiAndPhiFromConstraints: False
     Variables:
       SpacetimeMetric: SpacetimeMetric
       Pi: Pi
+      Phi: Phi
```
nilsvu, line 25: If `SetPiAndPhiFromConstraints` is true then a Phi dataset must be present in the file but will be unused. Also, `NumericInitialData` doesn't even need to know if we want to adjust the gauge later (which is the main thing that `SetPiAndPhiFromConstraints` does in my understanding). I think here's a way to fix this:

1. Add a variant to `NumericInitialData::Variables` that contains g, Pi, and Phi (in addition to the one with only g and Pi). Then set phi=deriv(g) for the one where phi is missing, and use the numeric phi when it's given.
2. Remove the option `SetPiAndPhiFromConstraints` from the numeric ID and put it into the `const_global_cache_tags` for the `SetPiAndPhiFromConstraints` action instead, just to enable or disable the action.

src/Evolution/Systems/GeneralizedHarmonic/Actions/SetInitialData.hpp
```diff
@@ -103,14 +105,16 @@ class NumericInitialData : public evolution::initial_data::InitialData {
...
     static constexpr Options::String help =
-        "GH variables: 'SpacetimeMetric' and 'Pi'. These variables are "
-        "used to set the initial data directly; Phi is then set to the "
+        "GH variables: 'SpacetimeMetric', 'Pi', and 'Phi'. These variables are "
+        "used to set the initial data directly; If the option "
+        "'SetPiAndPhiFromConstraints' is true, Phi is then set to the "
         "numerical derivative of SpacetimeMetric, to enforce the 3-index "
-        "constraint.";
+        "constraint and Pi is set using the 1-index constraint. Otherwise, "
+        "they are read in from the numeric ID.";
```
nilsvu: Note that Pi is always read from numeric ID, then modified to transform to the evolution gauge if `SetPiAndPhiFromConstraints` is true

nilsvu (changes requested):

support/Pipelines/Bbh/Ringdown.yaml
knelli2: For 2, though this makes the code a bit simpler, it makes the options in the input file look ugly. The `InitialData:` in the input file isn't a group, but just a tag that holds a unique ptr to ID, so if I move `SetPiAndPhiFromConstraints` to the cache, it will just be at the root yaml level and not under `InitialData:`. It works, but isn't very organized. Any suggestions?

nilsdeppe (dismissed):
LGTM!!! Thanks for doing this!!

The only possible thing we may want is the ability to set only pi, only phi, both, or none. However, I think that's beyond  the scope of this PR and can be delayed until we need it. Probably the easiest way to generalize is to change the tag type from a bool to an enum with different states.

support/Pipelines/Bbh/Ringdown.yaml
nilsvu: Yea you can put the tag in the evolution system group, next to the gauge condition and damping functions

geoffrey4444: I prefer keeping the option in the InitialData block

knelli2:
Updated after discussions with @nilsvu. Decided that if we are reading GH numeric data, then we won't need to set Pi and Phi at all. Examples being PB&J and Ringdown. And if we are using ADM variables, we'll need to set Pi to work with the evolution gauge and set Phi to satisfy constraints. So I removed the option in the input file and now we just always set Pi and Phi from constraints unless we are using GH numeric ID

src/Evolution/Systems/GeneralizedHarmonic/Actions/SetInitialData.hpp
```diff
@@ -96,21 +98,21 @@ class NumericInitialData : public evolution::initial_data::InitialData {
         "ADM variables: 'Lapse', 'Shift', 'SpatialMetric' and "
         "'ExtrinsicCurvature'. The initial GH variables will be computed "
         "from these numeric fields, as well as their numeric spatial "
-        "derivatives on the computational grid.";
+        "derivatives on the computational grid. The GH variable Pi will be set "
+        "to satisfy the gauge constaint using the evolution gauge. The GH "
```
nilsvu: typo "constraint"

nilsvu (dismissed):

nilsvu (approved):

PR #6431, opened 2025-01-07
Clean up ringdown script

AlexCarpenter46:
## Proposed changes

Cleaning and setting up the ringdown script for future updates.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/IO/H5/Python/FunctionsOfTimeFromVolume.py
```diff
@@ -0,0 +1,82 @@
...
+        logger.info("Selected ObservationID: " + str(which_obs_id))
+        logger.info("Selected match time: " + str(fot_times[which_obs_id]))
+
+        functions_of_time_at_match_time_dict["MatchTime"] = fot_times[
+            which_obs_id
+        ]
+
+        for fot_name, fot in functions_of_time.items():
+            if fot_name == "Rotation":
+                rot_func_and_2_derivs_tuple = fot.func_and_2_derivs(
```
knelli2: To keep this general and so we don't have to keep adding `if`s for future FoTs, just get the func and 2 deriv tuple for the FoT and convert to either a list of lists or 2d numpy array and return that. Doesn't matter how many components the FoT has. Whoever called this function should be the one to deal with it

support/Pipelines/Bbh/Ringdown.py
```diff
@@ -163,37 +163,43 @@ def start_ringdown(
...
+    match_time = evaluated_fot_dict["MatchTime"]
+
+    # This section checks for functions of time in the dictionary and alters
+    # some values in the evaluated functions of time dictionary. The ringdown
+    # only requires the outer boundary expansion map so here we set the inner
+    # expansion map to the identity. We also set the translation map to 0 since
+    # translation is not supported in the transition to ringdown yet.
+    if (
+        evaluated_fot_dict.get("Expansion") == None
+        or evaluated_fot_dict.get("ExpansionOuterBoundary") == None
```
knelli2: You just overwrite the expansion below and never use the value in the fot_dict, so you probably only need to check the outer boundary

support/Pipelines/Bbh/Ringdown.py
```diff
@@ -242,27 +249,37 @@ def start_ringdown(
...
-    ringdown_params["dtExpansionOuterBdry"] = (
-        expansion_func_outer_boundary_with_2_derivs[1]
-    )
-    ringdown_params["dt2ExpansionOuterBdry"] = (
-        expansion_func_outer_boundary_with_2_derivs[2]
-    )
+    ringdown_params["Rotation0"] = evaluated_fot_dict["Rotation"][0][0]
+    ringdown_params["Rotation1"] = evaluated_fot_dict["Rotation"][0][1]
+    ringdown_params["Rotation2"] = evaluated_fot_dict["Rotation"][0][2]
+    ringdown_params["Rotation3"] = evaluated_fot_dict["Rotation"][0][3]
```
knelli2: Let's try to figure out how to write lists of lists to yaml so we can avoid this. If it turns out that we can't, then we can just do this except using for loops like
```py
for i in range(4):
  for j in range(3):
    prefix = "" if j == 0 else ("dt" if j == 1 else "dt2")
    ringdown_params[f"{prefix}Rotation{i}"] = evaluated_fot_dict["Rotation"][j][i]
    if i == 4:
      continue
    # same for expansion and translation

**Review by knelli2 — CHANGES_REQUESTED** [review]
Thank you for doing this!

File: support/Pipelines/Bbh/Ringdown.py [hunk 4]
```diff
@@ -163,37 +163,43 @@ def start_ringdown(
...
     )
+    match_time = evaluated_fot_dict["MatchTime"]
+
+    # This section checks for functions of time in the dictionary and alters
+    # some values in the evaluated functions of time dictionary. The ringdown
+    # only requires the outer boundary expansion map so here we set the inner
+    # expansion map to the identity. We also set the translation map to 0 since
+    # translation is not supported in the transition to ringdown yet.
+    if (
+        evaluated_fot_dict.get("Expansion") == None
```
nilsvu:
```py
if not ("Expansion" in evaluated_fot_dict and "ExpansionOuterBoundary" in evaluated_fot_dict)`
```

support/Pipelines/Bbh/Ringdown.py
```diff
@@ -163,37 +163,43 @@ def start_ringdown(
...
+        or evaluated_fot_dict.get("ExpansionOuterBoundary") == None
+    ):
+        raise ValueError(
+            f"The transition to ringdown script requires the Expansion and"
+            f" ExpansionOuterBoundary functions of time to be valid at the"
+            f" match time. "
+        )
+    if evaluated_fot_dict.get("Rotation") == None:
```
nilsvu: also here `if "Rotation" not in evaluated_fot_dict`

support/Pipelines/Bbh/Ringdown.py
nilsvu: Should be possible either by passing the list through to the `Jinja2` rendering of the input file template, or by having the `yaml` module dump a yaml-formatted string and then passing that on to the input file template.

support/Pipelines/Bbh/Ringdown.py
```diff
@@ -229,11 +235,12 @@ def start_ringdown(
         ahc_dt2_datfile.append(ringdown_ylm_coefs[2])
     logger.info("Obtained ringdown coefs")
     # Print out coefficients for insertion into BBH domain
-    logger.info("Expansion: " + str(expansion_func_with_2_derivs))
+    logger.info("Expansion: " + str(evaluated_fot_dict["Expansion"]))
```
nilsvu: are all of these logging calls really needed? If they're for debugging then make them `logger.debug`, or delete because all params are printed out anyway when an error occurs.

nilsvu (changes requested):
Just commenting on some Python things

support/Pipelines/Bbh/Ringdown.py
AlexCarpenter46: I'll change them to logger.debug

support/Pipelines/Bbh/Ringdown.py
```diff
@@ -168,21 +168,28 @@ def start_ringdown(
...
-            f"The transition to ringdown script requires the Expansion and"
+            f"The transition to ringdown script requires the "
             f" ExpansionOuterBoundary functions of time to be valid at the"
             f" match time. "
         )
-    if evaluated_fot_dict.get("Rotation") == None:
+    if "Rotation" in evaluated_fot_dict:
+        evaluated_fot_dict["Rotation"] = [
+            [coef for coef in x] for x in evaluated_fot_dict["Rotation"]
+        ]
```
knelli2: Is this necessary? I forget what the python return type from `func_and_two_derivs` is.

support/Pipelines/Bbh/Ringdown.py
```diff
@@ -233,52 +240,33 @@ def start_ringdown(
...
-    ringdown_params["Translation2"] = evaluated_fot_dict["Translation"][0][2]
-    ringdown_params["dtTranslation0"] = evaluated_fot_dict["Translation"][1][0]
-    ringdown_params["dtTranslation1"] = evaluated_fot_dict["Translation"][1][1]
-    ringdown_params["dtTranslation2"] = evaluated_fot_dict["Translation"][1][2]
-    ringdown_params["dt2Translation0"] = evaluated_fot_dict["Translation"][2][0]
-    ringdown_params["dt2Translation1"] = evaluated_fot_dict["Translation"][2][1]
-    ringdown_params["dt2Translation2"] = evaluated_fot_dict["Translation"][2][2]
+    ringdown_params["Rotation"] = str(evaluated_fot_dict["Rotation"][0])
+    ringdown_params["dtRotation"] = str(evaluated_fot_dict["Rotation"][1])
+    ringdown_params["dt2Rotation"] = str(evaluated_fot_dict["Rotation"][2])
```
knelli2: What happens if you just have a Jinja template in the input file
```yaml
  Rotation: {{ Rotation }}
```
and just do
```py
ringdown_params["Rotation"] = evaluated_fot_dict["Rotation"]
```
The resulting input file doesn't need to look super pretty, just needs to parse.

knelli2 (changes requested):
And need to fix conflicts caused by #6115. Sorry, my bad!

support/Pipelines/Bbh/Ringdown.py
nilsvu: Currently the `schedule` function will recurse over lists, scheduling one run for every item in the list (e.g. to do convergence tests where you give it a list of resolutions or outer radii).

The `str` here is not robust enough because it gives Python's string representation of the object, which may or may not coincide with YAML syntax. Try `yaml.safe_dump(obj)`. This will give you the YAML representation of the object. If needed you can give it options, e.g. so a list is returned as `[1, 2]` instead of `- 1\n- 2\n`.

support/Pipelines/Bbh/Ringdown.py
AlexCarpenter46: I wondered the same, but it is needed. It returns a list of DataVectors and the StrahlkorperCoefsInRingdownDistortedFrame call requires the FoTs to be a list of doubles (for expansion/expansion outer boundary) or list of list of doubles for all the other FoTs.

knelli2: You could covert everything to a list of lists inside `functions_of_time_from_volume`, then you don't have to worry about it here.

knelli2 (commented):
Looks good, you can squash

nilsvu:
No need to wait for my review

knelli2 (approved):

PR #6433, opened 2025-01-09
Compute ADM momenta as volume integrals

iago-mendes:
## Proposed changes

This PR improves the accuracy of P_ADM and J_ADM by computing them as a sum of infinite volume integrals and finite surface integrals, which were derived from the standard ADM equations using Gauss' Law. We have attempted to use this same trick for other asymptotic quantities (ADM mass/energy and center of mass), but the volume version didn't improve accuracy. This is evident from the convergence plot below, in which the CoM volume results have been omitted because of blow-ups.

resolution_convergence_with_Jadm

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
+++ b/src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
@@ -45,19 +45,42 @@ void local_adm_integrals(
     center_of_mass->get(I) = 0;
   }
 
-  // Skip elements not at the outer boundary
-  const size_t radial_dimension = 2;
-  const auto radial_segment_id = element.id().segment_id(radial_dimension);
-  if (radial_segment_id.index() !=
-      two_to_the(radial_segment_id.refinement_level()) - 1) {
-    return;
+  // Compute quantities not in the data box
+  tnsr::II<DataVector, 3> inv_extrinsic_curvature;
+  tenex::evaluate<ti::I, ti::J>(make_not_null(&inv_extrinsic_curvature),
+                                inv_spatial_metric(ti::I, ti::K) *
+                                    inv_spatial_metric(ti::J, ti::L) *
+                                    extrinsic_curvature(ti::k, ti::l));
+
+  // Get integrands
+  const auto linear_momentum_surface_integrand =
+      Xcts::adm_linear_momentum_surface_integrand(
+          conformal_factor, inv_spatial_metric, inv_extrinsic_curvature,
+          trace_extrinsic_curvature);
+  const auto linear_momentum_volume_integrand =
+      Xcts::adm_linear_momentum_volume_integrand(
+          linear_momentum_surface_integrand, conformal_factor,
+          deriv_conformal_factor, conformal_metric, inv_conformal_metric,
+          conformal_christoffel_second_kind, conformal_christoffel_contracted);
+  const auto angular_momentum_z_volume_integrand =
+      Xcts::adm_angular_momentum_z_volume_integrand(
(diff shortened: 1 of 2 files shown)


src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
```diff
@@ -45,117 +45,199 @@ void local_adm_integrals(
...
+    adm_linear_momentum->get(I) += definite_integral(
+        linear_momentum_volume_integrand.get(I) * get(det_jacobian), mesh);
   }
 
   // Loop over external boundaries
   for (const auto boundary_direction : element.external_boundaries()) {
-    // Skip interfaces not at the outer boundary
-    if (boundary_direction != Direction<3>::upper_zeta()) {
-      continue;
+    // Interfaces at the inner boundary
```
nilsvu: You can do this to avoid duplicating all these projections:

```cpp
if (boundary_direction.dimension() != 2) {
  continue;
}
// Do projections
```

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
```diff
@@ -54,7 +54,7 @@ void test_local_adm_integrals(const double& distance,
                               const std::vector<double>& prev_distances) {
   // Define black hole parameters.
   const double mass = 1;
-  const double boost_speed = 0.5;
+  const double boost_speed = 0.;
```
nilsvu, line 55: Is this change needed?

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
```diff
@@ -198,6 +198,8 @@ void test_local_adm_integrals(const double& distance,
         deriv_conformal_metric, inv_conformal_metric);
     const auto conformal_christoffel_contracted = tenex::evaluate<ti::i>(
         conformal_christoffel_second_kind(ti::J, ti::i, ti::j));
+    const auto deriv_conformal_christoffel_second_kind = partial_derivative(
```
nilsvu: Do you need this and the longitudinal shift, Ricci additions below?

nilsvu (changes requested):

src/Elliptic/Systems/Xcts/Events/ObserveAdmIntegrals.cpp
iago-mendes: Thanks for pointing it out! I've also removed the duplicated code for the face normals and area elements.

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
iago-mendes: Yes, because the volume integrals give incorrect results for a boosted Schwarzschild case. In the past, we've wondered if this is due to some falloff assumption from Gauss Law being broken.

tests/Unit/Elliptic/Systems/Xcts/Events/Test_ObserveAdmIntegrals.cpp
iago-mendes: No, these quantities were only needed for the ADM mass volume integral, which isn't being computed here. Thanks for catching that!

nilsvu (approved):

PR #6436, opened 2025-01-14
Add puncture field as initial data for worldtube simulations

nikwit:
## Proposed changes

Adds the zeroth order puncture field to be used as initial data for the worldtube simulations.

### DIFF
--- a/src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.cpp
+++ b/src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.cpp
@@ -9,11 +9,13 @@
 #include "Domain/Creators/TimeDependence/RegisterDerivedWithCharm.hpp"
 #include "Domain/FunctionsOfTime/RegisterDerivedWithCharm.hpp"
 #include "Evolution/Systems/CurvedScalarWave/BoundaryCorrections/RegisterDerived.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/InitialData/ZerothOrderPuncture.hpp"
 #include "Parallel/CharmMain.tpp"
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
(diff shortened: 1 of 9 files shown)


src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.cpp
```diff
@@ -9,11 +9,13 @@
...
 #include "Evolution/Systems/CurvedScalarWave/BoundaryCorrections/RegisterDerived.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/InitialData/ZerothOrderPuncture.hpp"
 #include "Parallel/CharmMain.tpp"
 #include "Utilities/Serialization/RegisterDerivedClassesWithCharm.hpp"
 
-using metavariables = EvolutionMetavars<gr::Solutions::KerrSchild,
-                                        ScalarWave::Solutions::PlaneWave<3>>;
+using metavariables =
+    EvolutionMetavars<gr::Solutions::KerrSchild,
+                      CurvedScalarWave::AnalyticData::ZerothOrderPuncture>;
```
knelli2, line 18: Question: Does it make sense to keep the ability to use PlaneWave ID at all?

knelli2:
Maybe also give this a rebase since I believe the error from the nvcc CI was fixed.

src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.cpp
nikwit: not really. I always used a wave with amplitude 0. to get 0 initial data. But in the cases where I still need this, I can just set `Charge` to 0. Any plane wave solution is not going to be good physical solution with a scalar charge present that is not at infinity.

nikwit:
Done. Thanks for the review!

knelli2 (approved):

PR #6437, opened 2025-01-16
Revamp some deadlock analysis for GH

knelli2:
## Proposed changes

While debugging nodegroup deadlocks, these changes were helpful giving more, better organized, information.

This last commit was useful because sometimes after a deadlock, I'd want to run an event to get some extra information and we have a phase for doing that (`PostFailureCleanup`), but it wasn't run after a deadlock. If the method I have for implementing this needs a lot of discussion, I'm ok leaving it out for now. The other commits are more important I think.

### Upgrade instructions

For BBH and SingleBH executables, must add a block to the input file
```yaml
EventsRunAtCleanup:
  ObservationValue: -1000.0
  Events:
    - Event1
    - Event2
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/DistributedObject.hpp
+++ b/src/Parallel/DistributedObject.hpp
@@ -321,7 +321,13 @@ class DistributedObject<ParallelComponent,
   /// Start execution of the phase-dependent action list in `next_phase`. If
   /// `next_phase` has already been visited, execution will resume at the point
   /// where the previous execution of the same phase left off.
-  void start_phase(const Parallel::Phase next_phase);
+  ///
+  /// If \p force is true, then regardless of how this component terminated
+  /// (error or deadlock), it will resume.
+  ///
+  /// \warning Don't set \p force to true unless you are absolutely sure you
+  /// want to. This can have very unintended consequences if used incorrectly.
+  void start_phase(Parallel::Phase next_phase, bool force = false);
 
   /// Get the current phase
   Phase phase() const { return phase_; }
@@ -968,8 +974,15 @@ void DistributedObject<ParallelComponent,
 template <typename ParallelComponent, typename... PhaseDepActionListsPack>
 void DistributedObject<ParallelComponent,
                        tmpl::list<PhaseDepActionListsPack...>>::
-    start_phase(const Parallel::Phase next_phase) {
+    start_phase(const Parallel::Phase next_phase, const bool force) {
   try {
+    // An algorithm must always be set to terminate at the beginning of a phase,
+    // otherwise it's an error (see below). Therefore if we want to force a
+    // phase start, we must set terminate to true regardless of what it was
+    // before.
+    if (force) {
+      set_terminate(true);
+    }
     // terminate should be true since we exited a phase previously.
     if (not get_terminate() and not halt_algorithm_until_next_phase_) {
       ERROR(
--- a/src/Parallel/GlobalCache.hpp
+++ b/src/Parallel/GlobalCache.hpp
@@ -344,6 +344,17 @@ class GlobalCache
   void set_resource_info(
       const Parallel::ResourceInfo<Metavariables>& resource_info);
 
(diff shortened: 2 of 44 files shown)


src/Parallel/GlobalCache.hpp
```diff
@@ -698,6 +709,38 @@ void GlobalCache<Metavariables>::set_resource_info(
...
+           << "):\n";
+
+        for (const auto& [array_component_id, vec_callbacks] : callbacks) {
+          for (const auto& callback : vec_callbacks) {
+            ss << "  ArrayComponentId " << array_component_id << ": "
+               << callback->name() << "\n";
+          }
+        }
+
+        Parallel::fprintf(file_name, "%s\n", ss.str());
```
nilsdeppe: It might be easier to read the output file if you do this print _outside_ the `for_each`. Then you could do something like
```
============== BEGIN CALLBACKS ON NODE # =====================
...
============== END CALLBACKS ON NODE # =====================
```
to make the file easier to parse.

src/ParallelAlgorithms/Interpolation/Actions/PrintInterpolationTargetForDeadlock.hpp
```diff
@@ -0,0 +1,110 @@
...
+      }
+
+      for (const auto& temporal_id : temporal_ids) {
+        ss.str("");
+        ss << pretty_type::name<TargetTag>() << ", temporal id " << temporal_id
+           << ", ";
+
+        stream_points(temporal_id);
+
+        Parallel::fprintf(file_name, "%s\n", ss.str());
```
nilsdeppe: I would move the stream to outside the for loop. Then the block of text is guaranteed to be contiguous. You can then also delineate it with blocks like
```
=========== BEGIN INTERPOLATION TARGET ========
```

src/ParallelAlgorithms/Interpolation/Actions/PrintInterpolatorForDeadlock.hpp
```diff
@@ -0,0 +1,101 @@
...
+          for (const auto& element : num_elements) {
+            if (not info.interpolation_is_done_for_these_elements.contains(
+                    element)) {
+              difference.insert(element);
+            }
+          }
+
+          ss << difference;
+
+          Parallel::fprintf(file_name, "%s\n", ss.str());
```
nilsdeppe: I would move the stream to outside the loops. Then the block of text is guaranteed to be contiguous. You can then also delineate it with blocks like
```
=========== BEGIN INTERPOLATOR =============
```

src/Parallel/DistributedObject.hpp
```diff
@@ -968,8 +974,11 @@ void DistributedObject<ParallelComponent,
 template <typename ParallelComponent, typename... PhaseDepActionListsPack>
 void DistributedObject<ParallelComponent,
                        tmpl::list<PhaseDepActionListsPack...>>::
-    start_phase(const Parallel::Phase next_phase) {
+    start_phase(const Parallel::Phase next_phase, const bool force) {
   try {
+    if (force) {
+      set_terminate(true);
```
nilsdeppe, line 984: It's not obvious why terminate should be true. Could you add a code comment?

nilsdeppe (changes requested):
LGTM! Thanks for doing this! A few small suggestions on printing. Please squash immediately :)

src/ParallelAlgorithms/Interpolation/Actions/PrintInterpolatorForDeadlock.hpp
knelli2: Moving outside the `for` loop is good, but can't move outside the `for_each` because I write a new file for each target tag (the output can be long).

knelli2:
Squashed!

nilsdeppe (approved):

PR #6445, opened 2025-01-20
BBH pipeline: store target params, continue from evolution data for PBJ

nilsvu:
## Proposed changes

This enables PBJ by allowing to start from evolution data. To make sure all the parameters are consistent between the stopped and the continued evolution, the target parameters (like mass ratio and spins) are stored in input files and used to compute things like constraint damping parameters. The target parameters are also used in ID control and ecc-control.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
@iago-mendes this PR promotes target params for ID control and ecc-control to input file metadata, could you check that I didn't break anything in the control loop?

support/Pipelines/Bbh/ControlId.py
```diff
@@ -20,27 +20,17 @@
...
-FreeDataFromParams: Dict[SupportedParams, str] = {
-    "mass_A": "mass_a",
-    "mass_B": "mass_b",
-    "spin_A": "dimensionless_spin_a",
-    "spin_B": "dimensionless_spin_b",
-    "center_of_mass": "center_of_mass_offset",
-    "linear_momentum": "linear_velocity",
+# Note 2: conformal_mass_a/b and conformal_spin_a/b refer to the Kerr masses and
+#         spins used in the background.
+FreeDataFromParams: Dict[TargetParams, str] = {
```
iago-mendes, line 27: We also need to change `ScalarQuantities` below to match these new variables:

```
ScalarQuantities = ["conformal_mass_a", "conformal_mass_b", "MassA", "MassB"]
```

support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -427,7 +458,7 @@ ApparentHorizons:
     Verbosity: Quiet
   ObservationAhB: &AhB
     InitialGuess:
-      LMax: *LMax
+      LMax: *ObsLMax
```
knelli2, line 461: Until we actually have the ability to have the shape map LMax be different from the observation LMax, keep them the same anchor. Otherwise this could cause a lot of confusion and issues.

tests/support/Pipelines/Bbh/Test_Inspiral.py
```diff
@@ -98,23 +98,23 @@ def test_inspiral_parameters(self):
...
         # Control system
         self.assertEqual(params["MaxDampingTimescale"], 20.0)
-        self.assertEqual(params["KinematicTimescale"], 0.4)
-        self.assertAlmostEqual(params["SizeATimescale"], 0.04)
-        self.assertAlmostEqual(params["SizeBTimescale"], 0.04)
-        self.assertAlmostEqual(params["ShapeATimescale"], 2.0)
-        self.assertAlmostEqual(params["ShapeBTimescale"], 2.0)
+        self.assertEqual(params["KinematicTimescale"], 0.2)
+        self.assertAlmostEqual(params["SizeATimescale"], 0.024)
+        self.assertAlmostEqual(params["SizeBTimescale"], 0.016)
```
knelli2, line 103: Why are all these different now?

knelli2 (changes requested):

tests/support/Pipelines/Bbh/Test_Inspiral.py
nilsvu: they used the horizon masses before, and the target masses now.

support/Pipelines/Bbh/Inspiral.yaml
nilsvu: I had to do this because the shape map lmax isn't in the input file anymore when loading from file

nilsvu:
I added 3 small commits with improvements to the pipeline

tests/support/Pipelines/Bbh/Test_Inspiral.py
knelli2: But these are a factor of 2 different. Surely the horizon masses and target masses can't be that different?

nilsvu: These were just the numbers written in the test (horizon masses were 1 and target masses 0.5).

knelli2: Looks like the target masses are 0.6 and 0.4 actually, but yes I see now where the horizon masses were just arbitrarily chosen.

knelli2 (approved):

Issue #4603, opened 2023-01-12
Add CutX and Skew maps


geoffrey4444:
# Feature request:

### Component:

- [x] Code
- [ ] Documentation
- [ ] Build system
- [ ] Continuous integration

### Desired feature:

- [ ] At some point, unequal-mass BBHs will need the CutX and skew maps from SpEC, so add these to spectre.

PR #6446, opened 2025-01-22
Add Skew map

knelli2:
## Proposed changes

Part of #4603.

image

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
@@ -1,0 +1,408 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/CoordinateMaps/TimeDependent/Skew.hpp"
+
+#include <array>
+#include <cmath>
+#include <limits>
+#include <ostream>
+#include <pup.h>
+#include <pup_stl.h>
+#include <type_traits>
+#include <unordered_set>
+#include <utility>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/DeterminantAndInverse.hpp"
+#include "DataStructures/Tensor/Identity.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "NumericalAlgorithms/RootFinding/TOMS748.hpp"
+#include "Utilities/ContainerHelpers.hpp"
+#include "Utilities/DereferenceWrapper.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/MakeArray.hpp"
+#include "Utilities/MakeWithValue.hpp"
+#include "Utilities/StdArrayHelpers.hpp"
+#include "Utilities/StdHelpers.hpp"
+#include "Utilities/StlStreamDeclarations.hpp"
+#include "Utilities/TypeTraits/RemoveReferenceWrapper.hpp"
+
+namespace domain::CoordinateMaps::TimeDependent {
+
+Skew::Skew(std::string function_of_time_name,
+           const std::array<double, 3>& center, const double outer_radius)
+    : f_of_t_name_(std::move(function_of_time_name)),
+      center_(center),
+      outer_radius_(outer_radius),
+      outer_radius_squared_(square(outer_radius_)),
+      f_of_t_names_({f_of_t_name_}) {
+  if (magnitude(center_) >= outer_radius) {
+    ERROR("Center of Skew map "
+          << center_ << " with radius " << magnitude(center_)
+          << " must be within the outer radius " << outer_radius_);
+  }
+}
+
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, 3> Skew::operator()(
+    const std::array<T, 3>& source_coords, const double time,
+    const domain::FunctionsOfTimeMap& functions_of_time) const {
+  return map_and_velocity_helper(source_coords, time, functions_of_time, false);
+}
+
+std::optional<std::array<double, 3>> Skew::inverse(
+    const std::array<double, 3>& target_coords, const double time,
+    const domain::FunctionsOfTimeMap& functions_of_time) const {
+  const auto& function_of_time = functions_of_time.at(f_of_t_name_);
+  const DataVector func = function_of_time->func_and_deriv(time)[0];
+  ASSERT(func.size() == 2, "Expected a function of time with size 2, not "
+                               << func.size() << " in the Skew map.");
+
+  // Short circuit if there's no function of time
+  if (equal_within_roundoff(func[0], 0.0) and
+      equal_within_roundoff(func[1], 0.0)) {
+    return target_coords;
+  }
+
+  // Short circuit if we're at the outer radius
+  const double target_radius = magnitude(target_coords);
+  if (equal_within_roundoff(target_radius, outer_radius_)) {
+    return target_coords;
+  } else if (target_radius >= outer_radius_) {
+    return std::nullopt;
+  }
+
+  // Another short circuit. Target y & z are the same as source y & z
+  const double tan_sum = tan(func[0]) * (target_coords[1] - center_[1]) +
+                         tan(func[1]) * (target_coords[2] - center_[2]);
+  if (equal_within_roundoff(tan_sum, 0.0)) {
+    return target_coords;
+  }
+
+  std::array<double, 3> temporary_source_coord = target_coords;
+
+  const auto root_func = [&](const double source_coord_x) -> double {
+    temporary_source_coord[0] = source_coord_x;
+    const double width = get_width(temporary_source_coord);
+    return source_coord_x + width * tan_sum - target_coords[0];
+  };
+
+  // \bar{x} -> x + w * (tan(f_y)*(y-y_C) + tan(f_z)*(z-z_C)) and 0<=w<=1, then
+  // x <= \bar{x} <= x + (tan(f_y)*(y-y_C) + tan(f_z)*(z-z_C)), ==>
+  // 0 <= \bar{x} - x <= (tan(f_y)*(y-y_C) + tan(f_z)*(z-z_C)), ==>
+  // -\bar{x} <= - x <= -\bar{x} + (tan(f_y)*(y-y_C) + tan(f_z)*(z-z_C)), ==>
+  // \bar{x} >= x >= \bar{x} - (tan(f_y)*(y-y_C) + tan(f_z)*(z-z_C))
+  // This last line is our bracket for the root
+
+  // We give a small epsilon to give the root find a little buffer. It is highly
+  // unlikely there will be another root within these bounds
+  const double eps = std::numeric_limits<double>::epsilon() * 100.0;
+
+  const double x_0 = target_coords[0] - tan_sum - eps;
+  const double x_1 = target_coords[0] + eps;
+  const double lower_x = std::min(x_0, x_1);
+  const double upper_x = std::max(x_0, x_1);
+  const double lower_func = root_func(lower_x);
+  const double upper_func = root_func(upper_x);
+
+  // Since we adjusted the bounds by eps above, we have a slightly larger eps
+  // here to account for the arithmetic of the root func
+  if (equal_within_roundoff(lower_func, 0.0, 5 * eps)) {
+    temporary_source_coord[0] = lower_x;
+    return temporary_source_coord;
+  }
+
+  if (equal_within_roundoff(upper_func, 0.0, 5 * eps)) {
+    temporary_source_coord[0] = upper_x;
+    return temporary_source_coord;
+  }
+
+  if (lower_func * upper_func > 0.0) {
+    ERROR("No root in Skew map inverse!. Target coord = "
+          << target_coords << ", lower_x = " << lower_x
+          << ", lower_func = " << lower_func << ", upper_x = " << upper_x
+          << ", upper_func = " << upper_func << ", tan_sum = " << tan_sum);
+  }
+
+  temporary_source_coord[0] = RootFinder::toms748<true>(
+      root_func, lower_x, upper_x, lower_func, upper_func, eps, eps);
+
+  return temporary_source_coord;
+}
+
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, 3> Skew::frame_velocity(
+    const std::array<T, 3>& source_coords, const double time,
+    const domain::FunctionsOfTimeMap& functions_of_time) const {
+  return map_and_velocity_helper(source_coords, time, functions_of_time, true);
+}
+
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, 3, Frame::NoFrame> Skew::jacobian(
+    const std::array<T, 3>& source_coords, const double time,
+    const domain::FunctionsOfTimeMap& functions_of_time) const {
+  using ResultT = tt::remove_cvref_wrap_t<T>;
+
+  check_for_singular_map(source_coords, time, functions_of_time);
+
+  const ResultT width = get_width(source_coords);
+  const std::array<ResultT, 3> width_deriv = get_width_deriv(source_coords);
+
+  const auto& function_of_time = functions_of_time.at(f_of_t_name_);
+  const DataVector func = function_of_time->func_and_deriv(time)[0];
+  ASSERT(func.size() == 2, "Expected a function of time with size 2, not "
+                               << func.size() << " in the Skew map.");
+
+  auto result = identity<3>(dereference_wrapper(source_coords[0]));
+
+  const DataVector tan_func = tan(func);
+  // Temporarily use component to avoid allocation
+  auto& tan_sum = get<0, 2>(result);
+  tan_sum = tan_func[0] * (source_coords[1] - center_[1]) +
+            tan_func[1] * (source_coords[2] - center_[2]);
+
+  get<0, 0>(result) += width_deriv[0] * tan_sum;
+  get<0, 1>(result) = width_deriv[1] * tan_sum + width * tan_func[0];
+  get<0, 2>(result) = width_deriv[2] * tan_sum + width * tan_func[1];
+
+  return result;
+}
+
+template <typename T>
+tnsr::Ij<tt::remove_cvref_wrap_t<T>, 3, Frame::NoFrame> Skew::inv_jacobian(
+    const std::array<T, 3>& source_coords, const double time,
+    const domain::FunctionsOfTimeMap& functions_of_time) const {
+  return determinant_and_inverse(
+             jacobian(source_coords, time, functions_of_time))
+      .second;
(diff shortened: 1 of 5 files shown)


src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
```diff
@@ -0,0 +1,292 @@
...
+ * When the $\cos$ term is $-1$, then $W = 0$, and when the $\cos$ term is 1,
+ * then $W = 1$. Therefore, the function $\lambda(\vec{x})$ must go from 0 at
+ * the \p center, to 1 at $R$. In order to derive
+ * $\lambda(\vec{x})$, we introduce three vectors, shown in the figure, all
+ * originating from the origin: $\vec{x}$ is our source coordinate, $\vec{x}_0$
+ * is the \p center, and $\vec{x}_P$ is the vector to the intersection point
+ * between the \p outer_radius and the ray connecting $\vec{x}_0$ and $\vec{x}$.
+ * Therefore, $|\vec{x}_P| = R$.
+ *
+ * \image html Skew.svg "Vectors for Skew map"
```
nilsvu: This file is relatively large for what it does with 111KB (our limit is 200KB). It's the handwriting as vector graphic that makes it large I think. I expect it would be much smaller as a PNG, could you try that?

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
```diff
@@ -0,0 +1,292 @@
...
+ * \brief %Time dependent coordinate map that keeps the $y$ and $z$ coordinates
+ * unchanged, but will distort (or skew) the $x$ coordinate based on functions
+ * of time and radial distance to the origin.
+ *
+ * \details This coordinate map is only available in 3 dimensions and assumes
+ * all coordinates are given with respect to the origin of the coordinate
+ * system.
+ *
+ * ### Mapped Coordinates
+ * The Skew coordinte map is given by the mapping
```
nilsvu: typo "coordinate"

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
```diff
@@ -0,0 +1,292 @@
...
+ * the coordinate. The inverse of the $\bar{y}$ and $\bar{z}$ coordinates are
+ * trivial because there was no mapping. The equation we need to find the root
+ * of is
+ *
+ * \begin{equation}
+ *   0 = x + W(\vec{x})\left(\tan(F_y(t))y + \tan(F_z(t))z\right) - \bar{x}
+ * \end{equation}
+ *
+ * We can bound the root by noticing that in $\ref{eq:map}$, if we substitute
+ * the extramal values of $W = 0$ and $W = 1$, we get bounds on $\bar{x}$ which
```
nilsvu: typo "extremal"

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
```diff
@@ -0,0 +1,292 @@
...
+ *
+ * \begin{align}
+ *   \label{eq:map}
+ *   \bar{x} &= x + W(\vec{x})\left(\tan(F_y(t))y + \tan(F_z(t))z\right) \\
+ *   \bar{y} &= y \\
+ *   \bar{z} &= z
+ * \end{align}
+ *
+ * where $F_y(t)$ and $F_z(t)$ are the angles within the $(x,y)$ plane between
+ * the undistored $x$-axis and the skewed $\bar{y}$-axis at the \p center,
```
nilsvu: typo "undistorted"

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
```diff
@@ -0,0 +1,292 @@
...
+ *   \vec{x}_0\right)\cdot\vec{x}_0\right] + 2(R^2 - |\vec{x}_0|^2)(x^i -
+ * x_0^i). \end{equation}
+ *
+ * The inverse jacobian is computed by numerically inverting the jacobian.
+ */
+class Skew {
+ public:
+  static constexpr size_t dim = 3;
+
+  Skew(std::string function_of_time_name, const std::array<double, 3>& center,
```
nilsvu, line 173: Please add to the docs what expectations you have for the FoT. This info is also missing from other maps, but it's important because otherwise I don't know what numbers to put into the FoT to control the map.

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -0,0 +1,423 @@
...
+      center_(center),
+      outer_radius_(outer_radius),
+      outer_radius_squared_minus_center_radius_squared_(square(outer_radius) -
+                                                        dot(center_, center_)),
+      f_of_t_names_({f_of_t_name_}) {}
+
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, 3> Skew::operator()(
+    const std::array<T, 3>& source_coords, const double time,
+    const std::unordered_map<
```
nilsvu: (optional) `domain::FunctionsOfTimeMap` here and below

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -0,0 +1,423 @@
...
+
+  // Short circuit if there's no function of time
+  if (equal_within_roundoff(func[0], 0.0) and
+      equal_within_roundoff(func[1], 0.0)) {
+    return target_coords;
+  }
+
+  // Short ciruit if we're at the outer radius
+  const double target_radius = magnitude(target_coords);
+  if (target_radius >= outer_radius_ and
```
nilsvu: why this first condition? or do you mean `or`?

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -0,0 +1,423 @@
...
+  }
+  // Will be reused for result
+  ResultT lambda = (dot(centered_source_coords, center_) +
+                    sqrt(get_delta(centered_source_coords))) /
+                   outer_radius_squared_minus_center_radius_squared_;
+
+  ResultT& result = lambda;
+
+  for (size_t i = 0; i < get_size(result); i++) {
+    if (get_element(lambda, i) <= 0.0 and
```
nilsvu: Also here did you mean `or`?

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -0,0 +1,423 @@
...
+        equal_within_roundoff(get_element(lambda, i), 0.0)) {
+      get_element(result, i) = 1.0;
+    } else if (get_element(lambda, i) >= 1.0 and
```
nilsvu: Also here

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -0,0 +1,423 @@
...
+                         outer_radius_squared_minus_center_radius_squared_;
+
+  const ResultT centered_source_coords_dot_center =
+      dot(centered_source_coords, center_);
+  std::array<ResultT, 3> grad_delta{
+      center_[0] * centered_source_coords_dot_center,
+      center_[1] * centered_source_coords_dot_center,
+      center_[2] * centered_source_coords_dot_center};
+  // We don't multiply grad_delta by 2 like in the dox because it will just
+  // canel with the factor of 1/2 in grad_width
```
nilsvu: typo "cancel"

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -0,0 +1,423 @@
...
+                outer_radius_squared_minus_center_radius_squared_;
+
+  // Start off with grad_width = grad_lambda. We ignore the factor of 1/2 on the
+  // second term because it will cancel with the factor of 2 from grad_delta
+  std::array<ResultT, 3> grad_width =
+      (center_ + grad_delta / (sqrt(delta))) /
+      outer_radius_squared_minus_center_radius_squared_;
+  for (size_t i = 0; i < get_size(lambda); i++) {
+    // sin(lambda * M_PI) is zero at both lambda=0 and lambda=1
+    if ((get_element(lambda, i) <= 0.0 and
```
nilsvu: Also here `or` and below?

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -0,0 +1,423 @@
...
+  }
+
+  return result;
+}
+
+template <typename T>
+std::array<tt::remove_cvref_wrap_t<T>, 3> Skew::get_width_deriv(
+    const std::array<T, 3>& source_coords_cvref) const {
+  using ResultT = tt::remove_cvref_wrap_t<T>;
+  std::array<ResultT, 3> source_coords{};
```
nilsvu: Why do you need this non-const variable? Seems dangerous. Please add a comment if this is really necessary.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Skew.cpp
```diff
@@ -0,0 +1,123 @@
...
+      test_frame_velocity(skew_map_deserialized, points, t, f_of_t_list);
+    };
+
+    run_checks(point_xi);
+    test_coordinate_map_argument_types(skew_map, point_xi, t, f_of_t_list);
+    test_coordinate_map_argument_types(skew_map_deserialized, point_xi, t,
+                                       f_of_t_list);
+    test_inverse_map(skew_map, point_xi, t, f_of_t_list);
+    test_inverse_map(skew_map_deserialized, point_xi, t, f_of_t_list);
+    run_checks(dv_point_xi);
```
nilsvu, line 104: Maybe check that the map gives expected results at some points that you know, like at the center and at the outer radius

nilsvu (changes requested):

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
knelli2: No, I'm checking if the radius is within roundoff of the outer radius, but strictly *outside* the outer radius. If it's epsilon inside the outer radius everything is fine and the point can just go through the map/inverse.

nilsvu: If it's epsilon within the outer radius you can also short-circuit, no need to restrict the short-circuit to just one direction of the epsilon.

knelli2:
@nilsvu Pushed fixups and compressed the image some more. Wanted to keep it an svg so it rendered better online.

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -232,18 +229,22 @@ template <typename T>
...
-      // NOLINTBEGIN
-      gsl::at(source_coords, i)
-          .set_data_ref(make_not_null(&const_cast<DataVector&>(
-              dereference_wrapper(gsl::at(source_coords_cvref, i)))));
-      // NOLINTEND
-    } else {
-      gsl::at(source_coords, i) = gsl::at(source_coords_cvref, i);
+  // Can't do math with the cvref typed data so we convert it to the result
+  // type. If it's a DataVector then we use views. If it's doubles, just a copy
+  const std::array<ResultT, 3> source_coords = [&source_coords_cvref]() {
```
nilsvu: Can you just do this:

```cpp
const auto& x = dereference_wrapper(source_coords_cvref[0]);
const auto& y = dereference_wrapper(source_coords_cvref[1]);
const auto& z = dereference_wrapper(source_coords_cvref[2]);
```

Then you don't need any of these const-casts etc.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Skew.cpp
```diff
@@ -113,11 +115,74 @@ void test(const gsl::not_null<Generator*> generator) {
...
+    test_point = center;
+    CAPTURE(test_point);
+
+    const auto mapped_point = skew_map(test_point, time, f_of_t_list);
+    CHECK_ITERABLE_APPROX(mapped_point, (std::array{0.0, 1.0, 0.0}));
+    const auto inverse_point =
+        skew_map.inverse(mapped_point, time, f_of_t_list);
+    CHECK(inverse_point.has_value());
+    CHECK_ITERABLE_APPROX(inverse_point.value(), test_point);
+    // Jacobian not defined at the center
```
nilsvu: do we not need it?

nilsvu (changes requested):

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
nilsvu: Edit: I think in other maps we just do math with `cref(DataVector)` and that seems to work, see Wedge.cpp. Haven't checked the details.

knelli2: No I want to do math with the entire `std::array`, not the individual components. However, that math won't work between a `std::array<DataVector>` and `std::array<std::reference_wrapper<const DataVector>>`. That's why I const cast.

tests/Unit/Domain/CoordinateMaps/TimeDependent/Test_Skew.cpp
knelli2: Unclear. SpEC didn't handle this specially (and thus would get a division by zero error if the center was passed to the jacobian).

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
nilsvu: ok const-casts are pretty dangerous so maybe sprinkle some dereferencing in `StdArrayHelpers.hpp` instead?

knelli2: We use `const_cast` in a lot of places for this exact purpose (to set data references). Quick VSCode/git search shows 136 occurrences in 64 files

knelli2:
Ok @nilsvu I updated this to address the concerns we had about undefined points and smoothness. I squashed all previous fixups together into one to make it easier to see what I changed

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
```diff
@@ -59,76 +61,22 @@ namespace domain::CoordinateMaps::TimeDependent {
...
- *
- * We choose $\lambda(\vec{x})$ to linearly go from 0 at $\vec{x}_0$ to 1 at
- * $\vec{x}_P$ with the form
+ * the origin, to 1 at $R$. We choose $\lambda(\vec{x})$ to quadratically go
+ * from 0 at the origin to 1 at $R$ with the form
  *
  * \begin{equation}
  *   \label{eq:lambda}
- *   \lambda(\vec{x}) = \frac{|\vec{x} - \vec{x}_0|}{|\vec{x}_P - \vec{x}_0|}
+ *   \lambda(\vec{x}) = \frac{|\vec{x}|^2}{R^2}.
```
nilsvu, line 79: My understanding is that you added the square to make this smooth - if that is the case, I don't see the reasoning there. The square only changes the degree of differentiability by 1, doesn't it? (Edit: this was wrong. $|\vec{x}|^2$ is smooth even though $|\vec{x}|$ is not.) The issue remains that $\frac{dr}{dx^i} = \frac{x^i}{r}$ is not smooth at the origin (e.g. along the x-axis it is -1 when approached from the left and +1 when approached from the right). I think this is a problem, unless we can guarantee that the origin is always on an element boundary.

nilsvu (changes requested):

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
knelli2: $W$ ***IS*** smooth though so it's fine. Also I've checked first 3 derivatives of $W$ are smooth as well (it's just sine terms multiplied by powers of $x$). Whether $\lambda$ is smooth is irrelevant. Only the derivatives of $W$ are the ones that mater.

nilsvu: Ok I'm convinced now that the map is smooth. The square is actually not relevant because the cos is symmetric, so $\lambda(\vec{x}) = \frac{|\vec{x}|}{R}$ would also work.

knelli2:
@nilsvu I incorporated the change found in https://github.com/sxs-collaboration/spec/pull/2830 of subtracting off the center of the skew map in certain places, which was discussed with @markscheel. Now this should be smooth and correct at all points.

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
```diff
@@ -0,0 +1,246 @@
...
+namespace domain::CoordinateMaps::TimeDependent {
+/*!
+ * \ingroup CoordMapsTimeDependentGroup
+ * \brief %Time dependent coordinate map that keeps the $y$ and $z$ coordinates
+ * unchanged, but will distort (or skew) the $x$ coordinate based on functions
+ * of time and radial distance to the origin.
+ *
+ * \details This coordinate map is only available in 3 dimensions and assumes
+ * all coordinates are given with respect to the origin of the coordinate
+ * system.
```
nilsdeppe: I don't understand what this means. What else would I give the coordinates relative to?

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
markscheel: I don't think `|x|/R` works, despite the cosine being symmetric.  Because when you differentiate, you get sine terms, and even worse, you get derivs of `|x|/R`).  `|x|^2/R^2` is fine though.

src/Domain/CoordinateMaps/TimeDependent/Skew.hpp
knelli2: This was to try and disambiguate the origin from the center of the skew map, but this does make it confusing. I can remove it when I squash.

nilsdeppe: Ah, okay. That's what I assumed but was only able to figure it out because I reviewed Mark's Spec PR :) Thanks for removing it!

src/Domain/CoordinateMaps/TimeDependent/Skew.cpp
```diff
@@ -0,0 +1,424 @@
...
+std::array<tt::remove_cvref_wrap_t<T>, 3> Skew::operator()(
+    const std::array<T, 3>& source_coords, const double time,
+    const domain::FunctionsOfTimeMap& functions_of_time) const {
+  return map_and_velocity_helper(source_coords, time, functions_of_time, false);
+}
+
+std::optional<std::array<double, 3>> Skew::inverse(
+    const std::array<double, 3>& target_coords, const double time,
+    const std::unordered_map<
+        std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>&
```
nilsvu: (optional) Can use `domain::FunctionsOfTimeMap` here as well

nilsvu (dismissed):
LGTM please squash

nilsvu (approved):

PR #6447, opened 2025-01-22
Add Skew map to BCO

knelli2:
## Proposed changes

Adds the Skew map to the BinaryCompactObject domain. Sets it to None in all input files for now until a control system is written for it.

Part of #4603.

### Upgrade instructions

When using the BinaryCompactObject domain creator, you must add the following option to the time depndent options:
```yaml
  TimeDependentOptions:
    SkewMap:
      InitialValuesY: [0.0, 0.01, 0.0]
      InitialValuesZ: [0.0, 0.01, 0.0]
```
It has the same capabilities to be `None` or from a volume file as the other time dependent map options.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #6446.~

### DIFF
--- a/src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
+++ b/src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
@@ -19,6 +19,7 @@
 #include "Domain/Creators/TimeDependentOptions/ExpansionMap.hpp"
 #include "Domain/Creators/TimeDependentOptions/RotationMap.hpp"
 #include "Domain/Creators/TimeDependentOptions/ShapeMap.hpp"
+#include "Domain/Creators/TimeDependentOptions/SkewMap.hpp"
 #include "Domain/Creators/TimeDependentOptions/TranslationMap.hpp"
 #include "Domain/FunctionsOfTime/FixedSpeedCubic.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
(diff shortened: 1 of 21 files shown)


knelli2:
Rebased and no longer dependent

src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
```diff
@@ -469,8 +488,14 @@ TimeDependentMapOptions<IsCylindrical>::distorted_to_inertial_map(
                                     : RotScaleTrans{};
 
   if (block_has_shape_map) {
-    if (rot_scale_trans_map_.has_value()) {
+    if (rot_scale_trans_map_.has_value() and
+        (use_rigid_map and skew_map_.has_value())) {
```
nilsvu: If `use_rigid_map=false` but both `rot_scale_trans` and `skew` maps are given, this will ignore the skew map. Is that what you want? Could you make this clearer, maybe with a comment or by throwing errors or something

nilsvu (changes requested):

src/Domain/Creators/TimeDependentOptions/BinaryCompactObject.cpp
knelli2: Yes this is correct. I'll add a comment

nilsvu: So when `use_rigid_map=false` the skew map can't be applied? Better throw an error then if the skew map was requested.

knelli2: Not "can't" be applied, but "won't" be applied. If there is a skew map, but the block we are in is in the outer shell, then `use_rigid_map=false` and `skew_map_.has_value()=true`, yet we don't want to apply the skew map

nilsvu: Ok that clarifies it, thanks 👍

I think in this code branch and also in the `block_has_shape_map` branch in `grid_to_distorted_map` below we're always within the inner cubes, so `use_rigid_map` should be always `true`. Maybe assert that instead of the sentinel.

knelli2: Added some errors

nilsvu (approved):

PR #6449, opened 2025-01-22
Remove GH dependence on ScalarTensor and Hydro

nilsdeppe:
## Proposed changes

GH shouldn't depend on other evolution systems. The MHD solutions do not depend on the evolution systems and so this is would've been fine, but I think made it too easy to misunderstand the dependencies. I think the tags for ScalarTensor (and probably most of our systems) should be moved to a PointwiseFunctions/ScalarTensor directory. That's likely where most of the stuff currently in the system directory should live, with only the assembly for the executable being in the system. That would include actions, instantiations, and most of the FD/subcell code since it's not pointwise.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


kidder (approved):

nilsvu:
We can even put instantiations, actions etc next to the executable so we eliminate one directory tree altogether (either Systems/ or Executables/)

nilsdeppe:
Oh yea, that's a good idea! And we can change a lot of things to use `db::Access` and move them to cpp files too, I think.

PR #6454, opened 2025-01-24
BBH pipeline: prompt to edit README.md

nilsvu:
## Proposed changes

When starting runs, prompt to edit the README.md file. I hope this helps keeping better track of notes for each run. Can be disabled on the command line.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/InitialData.py
+++ b/support/Pipelines/Bbh/InitialData.py
@@ -130,6 +130,7 @@ def generate_id(
     pipeline_dir: Optional[Union[str, Path]] = None,
     run_dir: Optional[Union[str, Path]] = None,
     segments_dir: Optional[Union[str, Path]] = None,
+    edit_readme: Optional[bool] = False,
     out_file_name: str = "spectre.out",
     **scheduler_kwargs,
 ):
@@ -176,6 +177,9 @@ def generate_id(
         with 'pipeline_dir'.
       segments_dir: Directory where the evolution data is generated. Mutually
         exclusive with 'pipeline_dir' and 'run_dir'.
+      edit_readme: Edit the README.md file in the pipeline directory or run
+        directory. If not specified, a prompt will ask whether to edit the
+        README.
       out_file_name: Optional. Name of the log file. (Default: "spectre.out")
     """
     logger.warning(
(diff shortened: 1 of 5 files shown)


support/Pipelines/Bbh/Inspiral.yaml
```diff
@@ -423,21 +423,21 @@ ApparentHorizons:
       DivergenceTol: 1.2
       DivergenceIter: 5
       MaxIts: 100
-    Verbosity: Quiet
+    Verbosity: Silent
```
knelli2, line 426: Leave these as quiet. We don't write this information out anywhere else (even though maybe we should) and it's useful to know. Also, it's not affecting performance (that I know of)

support/Pipelines/Bbh/InitialData.py
```diff
@@ -393,6 +405,14 @@ def generate_id(
     ),
     help="Directory where steps in the pipeline are created.",
 )
+@click.option(
+    "--edit-readme/--no-edit-readme",
+    default=None,
+    help=(
+        "Edit the README file in the pipeline directory or run directory. "
+        "If not specified, a prompt will ask whether to edit the README."
```
knelli2, line 413: I'm not a huge fan of always being prompted for this. I'd like the option, but just not being forced to do it by default. I think this should just be a bool that defaults to false.

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -404,6 +405,17 @@ def start_inspiral(
...
 
+    # Edit README
+    readme = (
+        Path(pipeline_dir or segments_dir or run_dir or Path.cwd())
+        / "README.md"
+    )
+    if edit_readme or (
+        edit_readme is None and click.confirm(f"Edit {readme}?", default=True)
+    ):
+        click.edit(filename=readme)
```
knelli2, line 416: Is there a way to not repeat this or the click option multiple times?

knelli2 (changes requested):
Tests are failing because everything is being p

support/Pipelines/Bbh/InitialData.py
nilsvu: Without the prompt there's little point to have the option. The idea is to encourage writing notes into a readme file.

nilsvu:
We decided not to do this for now

PR #6455, opened 2025-01-24
Allow skipping GRMHD RHS computation in atmosphere/wavezone

nilsdeppe:
## Proposed changes

No need to do all the extra work. This is currently only on DG since FD will mostly happen there where we are not in atmosphere.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DiscontinuousGalerkin/Actions/VolumeTermsImpl.tpp
+++ b/src/Evolution/DiscontinuousGalerkin/Actions/VolumeTermsImpl.tpp
@@ -4,6 +4,7 @@
 #pragma once
 
 #include <cstddef>
+#include <cstdlib>
 #include <optional>
 #include <ostream>
 #include <type_traits>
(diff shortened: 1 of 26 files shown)


src/Evolution/DiscontinuousGalerkin/Actions/VolumeTermsImpl.tpp
```diff
@@ -289,6 +298,8 @@ void volume_terms(
             }
           }
         });
+  end_of_flux_div_calculation:
+    (void)5;
```
kidder: replace with `; // null statement required after label at end of block until C++23`

kidder (changes requested):
you can squash

src/Evolution/DiscontinuousGalerkin/Actions/VolumeTermsImpl.tpp
nilsdeppe: Done! Thanks for the review!

kidder (approved):
optional:  I don't think you need the '(void)5' but the comment was the important part...

PR #6457, opened 2025-01-24
Add StressEnergyTensor free function

jyoo1042:
## Proposed changes

Adding a function to evaluate the stress-energy tensor, which currently does not exist in StressEnergy.cpp
This will be mainly used for evaluating diagnostics for accretion disk simulations.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


teukolsky:
Can you please put in a comment about the purpose of this PR?  What's new? Thanks.

src/PointwiseFunctions/Hydro/StressEnergy.cpp
```diff
@@ -70,23 +99,78 @@ void stress_trace(gsl::not_null<Scalar<DataType>*> result,
...
+  const auto four_velocity_v =
+      four_velocity(spatial_velocity, shift, lorentz_factor, lapse);
+
+  const auto rho_h_star =
+      (get(rest_mass_density) +
+       get(rest_mass_density) * get(specific_internal_energy)) +
+      get(pressure) + square(get(comoving_magnetic_field_magnitude));
+
+  const auto p_star =
+      get(pressure) + square(get(comoving_magnetic_field_magnitude)) / 2.;
```
nilsdeppe: You don't want auto here because this creates an expression tree that gets evaluated many times in the loop below. Specify the type

src/PointwiseFunctions/Hydro/ComovingMagneticField.hpp
```diff
@@ -37,8 +37,25 @@ namespace hydro {
...
+    const Scalar<DataType>& lapse);
+
+template <typename DataType>
+tnsr::A<DataType, 3> comoving_magnetic_field(
+    const tnsr::I<DataType, 3>& spatial_velocity,
+    const tnsr::I<DataType, 3>& magnetic_field,
+    const Scalar<DataType>& magnetic_field_dot_spatial_velocity,
+    const Scalar<DataType>& lorentz_factor, const tnsr::I<DataType, 3>& shift,
+    const Scalar<DataType>& lapse);
+
```
nilsdeppe, line 57: Factor this function into its own commit and add a test

src/PointwiseFunctions/Hydro/StressEnergy.cpp
```diff
@@ -70,23 +99,78 @@ void stress_trace(gsl::not_null<Scalar<DataType>*> result,
           (square(get(lorentz_factor)) * get(spatial_velocity_squared) + 1.);
 }
 
+template <typename DataType>
+void stress_energy_tensor(
+    gsl::not_null<tnsr::AA<DataType, 3>*> result,
```
nilsdeppe: const

src/PointwiseFunctions/Hydro/StressEnergy.cpp
nilsdeppe: don't divide by two, multiply by 0.5 on the left

src/PointwiseFunctions/Hydro/StressEnergy.cpp
```diff
@@ -70,23 +99,78 @@ void stress_trace(gsl::not_null<Scalar<DataType>*> result,
...
+    const Scalar<DataType>& specific_internal_energy,
+    const Scalar<DataType>& pressure, const Scalar<DataType>& lorentz_factor,
+    const Scalar<DataType>& lapse,
+    const Scalar<DataType>& comoving_magnetic_field_magnitude,
+    const tnsr::I<DataType, 3>& spatial_velocity,
+    const tnsr::I<DataType, 3>& shift,
+    const tnsr::I<DataType, 3>& magnetic_field,
+    const tnsr::ii<DataType, 3>& spatial_metric,
+    const tnsr::II<DataType, 3>& inverse_spatial_metric) {
+  const auto inverse_spacetime_metric =
```
nilsdeppe: I'm generally unhappy about the large number of allocations in this function. It would be good to switch to the not_null versions and preallocate in a Variables

nilsdeppe (changes requested):

src/PointwiseFunctions/Hydro/ComovingMagneticField.hpp
jyoo1042: done.

src/PointwiseFunctions/Hydro/StressEnergy.cpp
jyoo1042: done.

src/PointwiseFunctions/Hydro/StressEnergy.cpp
jyoo1042: done.

src/PointwiseFunctions/Hydro/StressEnergy.cpp
jyoo1042: done.

tests/Unit/PointwiseFunctions/Hydro/Test_ComovingMagneticField.cpp
```diff
@@ -20,6 +20,14 @@ SPECTRE_TEST_CASE("Unit.PointwiseFunctions.Hydro.ComovingMagneticField",
...
       "PointwiseFunctions/Hydro/");
   const DataVector used_for_size(5);
+  pypp::check_with_random_values<1>(
+      static_cast<tnsr::A<DataVector, 3> (*)(
+          const tnsr::I<DataVector, 3>&, const tnsr::I<DataVector, 3>&,
+          const Scalar<DataVector>&, const Scalar<DataVector>&,
+          const tnsr::I<DataVector, 3>&, const Scalar<DataVector>&)>(
+          &comoving_magnetic_field<DataVector>),
+      "TestFunctions", "comoving_magnetic_field", {{{0.0, 1.0}}},
+      used_for_size);
```
nilsdeppe, line 129: Can you test that `b^2=b^ab_a` from the 3 functions?

src/PointwiseFunctions/Hydro/ComovingMagneticField.cpp
```diff
@@ -10,6 +10,39 @@
...
+    const Scalar<DataType>& lorentz_factor, const tnsr::I<DataType, 3>& shift,
+    const Scalar<DataType>& lapse) {
+  get<0>(*result) = get(lorentz_factor) *
+                    get(magnetic_field_dot_spatial_velocity) / get(lapse);
+
+  for (size_t i = 0; i < 3; ++i) {
+    result->get(i + 1) =
+        (magnetic_field.get(i) / get(lorentz_factor)) +
+        get(lapse) * get<0>(*result) *
+            (spatial_velocity.get(i) - shift.get(i) / get(lapse));
```
nilsdeppe: why not analytically multiply through the lapse? Then you save yourself a division, which is expensive.

src/PointwiseFunctions/Hydro/StressEnergy.cpp
```diff
@@ -70,23 +101,99 @@ void stress_trace(gsl::not_null<Scalar<DataType>*> result,
...
+    result->get(i, i) = get(rho_h_star) * square(four_velocity_v.get(i)) +
+                        get(p_star) * inverse_spacetime_metric_v.get(i, i) -
+                        square(comoving_magnetic_field_v.get(i));
+
+    for (size_t j = i + 1; j < 4; ++j) {
+      result->get(i, j) =
+          get(rho_h_star) * four_velocity_v.get(i) * four_velocity_v.get(j) +
+          get(p_star) * inverse_spacetime_metric_v.get(i, j) -
+          comoving_magnetic_field_v.get(i) * comoving_magnetic_field_v.get(j);
+    }
```
nilsdeppe: why do you have these two separate? Isn't it the same equation for j=i?

nilsdeppe (changes requested):

tests/Unit/PointwiseFunctions/Hydro/Test_ComovingMagneticField.cpp
jyoo1042: done.

src/PointwiseFunctions/Hydro/ComovingMagneticField.cpp
jyoo1042: done.

src/PointwiseFunctions/Hydro/StressEnergy.cpp
jyoo1042: done.

tests/Unit/PointwiseFunctions/Hydro/Test_ComovingMagneticField.cpp
```diff
@@ -8,10 +8,111 @@
...
+      make_with_value<Scalar<DataType>>(used_for_size, 0.0);
+  for (size_t i = 0; i < 4; ++i) {
+    get(comoving_magnetic_field_squared_calc) +=
+        comoving_magnetic_field_v.get(i) *
+        comoving_magnetic_field_one_form_v.get(i);
+  }
+  const double epsilon_approx = 1.e-15;
+  CHECK_ITERABLE_CUSTOM_APPROX(
+      comoving_magnetic_field_squared_calc, comoving_magnetic_field_squared_v,
+      Approx::custom().epsilon(epsilon_approx).scale(1.0));
```
nilsdeppe: Why the custom approx? We default to epsilon of 1e-14, but then you set the scale relative to 1.0. Would you instead want to also compare a specific number of digits, so not scale relative to 1?

src/PointwiseFunctions/Hydro/StressEnergy.cpp
```diff
@@ -4,11 +4,42 @@
...
 #include "Utilities/Gsl.hpp"
 
+namespace {
+template <typename DataType>
+void four_velocity(const gsl::not_null<tnsr::A<DataType, 3>*> result,
+                   const tnsr::I<DataType, 3>& spatial_velocity,
+                   const tnsr::I<DataType, 3>& shift,
+                   const Scalar<DataType>& lorentz_factor,
+                   const Scalar<DataType>& lapse) {
+  get<0>(*result) = get(lorentz_factor) / get(lapse);
```
nilsdeppe, line 24: To reduce the number of divisions, you can set this to `1 / lapse` and then after the for loop multiple the W. Then you can not divide by the lapse in the for loop and multiply by the inverse lapse instead.

tests/Unit/PointwiseFunctions/Hydro/Test_StressEnergy.cpp
```diff
@@ -17,17 +17,20 @@ namespace hydro {
...
-  pypp::check_with_random_values<1>(&stress_trace<DataVector>, "TestFunctions",
-                                    {"stress_trace"}, {{{0.0, 1.0}}},
-                                    used_for_size);
+  pypp::check_with_random_values<1>(&stress_trace<DataVector>,
+                                    "Test_StressEnergy", {"stress_trace"},
+                                    {{{0.0, 1.0}}}, used_for_size);
+  pypp::check_with_random_values<1>(
+      &stress_energy_tensor<DataVector>, "Test_StressEnergy",
+      {"stress_energy_tensor"}, {{{0.0, 1.0}}}, used_for_size);
 }
```
nilsdeppe, line 202: Can you test that what you get agrees with the implementation in GRMHD? There we compute the trace-reversed stress tensor with downstairs indices.

nilsdeppe (changes requested):

tests/Unit/PointwiseFunctions/Hydro/Test_ComovingMagneticField.cpp
jyoo1042: changed to CHECK_ITERABLE_APPROX

src/PointwiseFunctions/Hydro/StressEnergy.cpp
jyoo1042: done.

tests/Unit/PointwiseFunctions/Hydro/Test_StressEnergy.cpp
jyoo1042: @nilsdeppe, I added a new test for this: compute the $T^{\mu \nu}$, lower it, compute $T_{\mu \nu} - \frac{1}{2} g_{\mu \nu} g^{\sigma \rho} T_{\sigma \rho}$ and compare to the result of trace_reversed_stress_energy function in GRMHD. However, I am hitting some cases where the test fails with the default value epsilon. The test passes reliably with an epsilon of 1e-10.

nilsdeppe: I would just set the epsilon to 1e-10 and add a comment. I think that's fine. There is catastrophic cancellation in this computation, which is why we implemented it directly in the GH system.

tests/Unit/PointwiseFunctions/Hydro/Test_StressEnergy.cpp
```diff
@@ -7,27 +7,195 @@
...
+                                    spacetime_metric_v(ti::c, ti::a) *
+                                    spacetime_metric_v(ti::d, ti::b));
+  tenex::evaluate<ti::a, ti::b>(
+      make_not_null(&trace_reversed_stress_energy_tensor_calc),
+      stress_energy_tensor_lowered(ti::a, ti::b) -
+          0.5 * spacetime_metric_v(ti::a, ti::b) *
+              inverse_spacetime_metric_v(ti::C, ti::D) *
+              stress_energy_tensor_lowered(ti::d, ti::c));
+
+  Approx approx = Approx::custom().epsilon(1e-10);
```
nilsdeppe: `1.e-10` (see second last bullet here https://spectre-code.org/code_review_guide.html )

jyoo1042: okay. changed it.

tests/Unit/PointwiseFunctions/Hydro/Test_StressEnergy.cpp
jyoo1042: okay.

nilsdeppe (approved):

PR #6459, opened 2025-01-26
Fix bug in PreprocessCceWorldtube

### LINKED ISSUE #6450: A bug in PreprocessCceWorldtube.cpp
There is a bug in `src/Executables/PreprocessCceWorldtube/PreprocessCceWorldtube.cpp` at

https://github.com/sxs-collaboration/spectre/blob/f46bc1332ed5651e749041d49f12adf6ebe4c9fe/src/Executables/PreprocessCceWorldtube/PreprocessCceWorldtube.cpp#L166

The bool parameter `fix_spec_normalization` is wrongly passed when instantiating `Cce::MetricWorldtubeH5BufferUpdater<ComplexModalVector> `. This is because the last argument of the `MetricWorldtubeH5BufferUpdater` constructor, `file_is_from_spec`, has a different meaning than `fix_spec_normalization`.

Here `file_is_from_spec=True` means that the worldtube data are stored in descending m, see https://github.com/sxs-collaboration/spectre/blob/f46bc1332ed5651e749041d49f12adf6ebe4c9fe/src/Evolution/Systems/Cce/WorldtubeBufferUpdater.cpp#L503
whereas `fix_spec_normalization=True` tries to fix incorrect normalized radial derivatives from an old version of SpEC.

To fix the issue, one could make the final argument of the instantiation as a customized option, and comment that it should be `True` while using SpEC data.

Also, the name `file_is_from_spec` is misleading. I'd suggest renaming it to something more descriptive, such as `m_descending`.

knelli2:
## Proposed changes

Fixes #6450.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Tutorials/CCE.md
+++ b/docs/Tutorials/CCE.md
@@ -292,6 +292,9 @@ Here are some notes about the different options in the YAML input file:
   simulation that provided the boundary dataset.
 - `FixSpecNormalization` should always be `False` unless you are using a
   particualy old version of SpEC
+- `DescendingM` should always be `False`. This option is to support metric modal
+  data from SpEC. Our [data format](#input_worldtube_data_format) is always in
+  *ascending* m.
 - `BufferDepth` is an advanced option that lets you load more data into RAM at
   once so there are fewer filesystem accesses.
 
--- a/src/Evolution/Systems/Cce/WorldtubeBufferUpdater.cpp
+++ b/src/Evolution/Systems/Cce/WorldtubeBufferUpdater.cpp
@@ -307,10 +307,10 @@ double update_buffers_for_time(
 template <typename T>
 MetricWorldtubeH5BufferUpdater<T>::MetricWorldtubeH5BufferUpdater(
     const std::string& cce_data_filename,
-    const std::optional<double> extraction_radius, const bool file_is_from_spec)
+    const std::optional<double> extraction_radius, const bool descending_m)
     : cce_data_file_{cce_data_filename},
       filename_{cce_data_filename},
-      file_is_from_spec_(file_is_from_spec) {
+      descending_m_(descending_m) {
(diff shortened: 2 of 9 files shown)


docs/Tutorials/CCE.md
```diff
@@ -292,6 +292,9 @@ Here are some notes about the different options in the YAML input file:
   simulation that provided the boundary dataset.
 - `FixSpecNormalization` should always be `False` unless you are using a
   particualy old version of SpEC
+- `DescendingM` should always be `False`. This option is to maintain
```
Sizheng-Ma: This sentence could be confusing because `DescendingM` depends on how people provide their metric modal data. It can be `True`. I guess you are referring to the output Bondi modal data, which always have ascending m.

docs/Tutorials/CCE.md
```diff
@@ -292,6 +292,9 @@ Here are some notes about the different options in the YAML input file:
...
+  compatibility with older versions of SpEC. Our
```
Sizheng-Ma: Current SpEC still dumps metric modal data with descending m (even though this option has been disabled in a default run). I would be more explicit here:

`This option is to support metric modal data from SpEC.`

Sizheng-Ma:
This executable can run properly.

docs/Tutorials/CCE.md
knelli2: Yes technically it can be true for metric modal data, but I think we want to encourage people to always have their data in ascending m order (unless it's from spec). Hence the strong language that it should always be False.

Sizheng-Ma:
Looks good.

Sizheng-Ma (approved):

nilsdeppe (approved):

PR #6462, opened 2025-01-27
Fix clang-tidy on things including libsharp

wthrowe:
Also disable a new check with a lot of false positives.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


.clang-tidy
```diff
@@ -59,6 +59,8 @@ Checks: '*,
 -misc-confusable-identifiers,
 # thinks constexpr variables in header files cause ODR violations,
 -misc-definitions-in-headers,
+# Insufficiently configurable, does not handle system headers well,
+-misc-include-cleaner,
```
kidder, line 63: - I've been using it some to check for missing/extra includes and agree it has some promise, but also agree it definitely needs some control over its behavior before enabling it on CI.
 - in addition to the private system includes,  I also got errors I'd like to ignore when including macros from other files

nilsdeppe (approved):

PR #6463, opened 2025-01-28
Status CLI / dashboard: handle pending jobs better, plot simulation time over calendar time

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Parallel/PhaseControl/CheckpointAndExitAfterWallclock.hpp
+++ b/src/Parallel/PhaseControl/CheckpointAndExitAfterWallclock.hpp
@@ -24,6 +24,7 @@
 #include "Utilities/System/ParallelInfo.hpp"
 #include "Utilities/TMPL.hpp"
 #include "Utilities/TaggedTuple.hpp"
+#include "Utilities/UtcTime.hpp"
 
 namespace PhaseControl {
 
@@ -235,6 +236,8 @@ CheckpointAndExitAfterWallclock::arbitrate_phase_change_impl(
     } else {
       // if current_phase is WriteCheckpoint, we follow with updating options
       if (current_phase == Parallel::Phase::WriteCheckpoint) {
+        Parallel::printf("Restarting from checkpoint. Date and time: %s\n",
+                         utc_time());
         return std::make_pair(
             Parallel::Phase::UpdateOptionsAtRestartFromCheckpoint,
             ArbitrationStrategy::PermitAdditionalJumps);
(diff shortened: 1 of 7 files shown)


src/Parallel/PhaseControl/CheckpointAndExitAfterWallclock.hpp
```diff
@@ -235,6 +235,10 @@ CheckpointAndExitAfterWallclock::arbitrate_phase_change_impl(
     } else {
       // if current_phase is WriteCheckpoint, we follow with updating options
       if (current_phase == Parallel::Phase::WriteCheckpoint) {
+        const auto now = std::chrono::system_clock::now();
+        const std::time_t now_c = std::chrono::system_clock::to_time_t(now);
+        Parallel::printf("Restarting from checkpoint. Date and time: %s\n",
+                         std::ctime(&now_c));
```
nilsdeppe: Please print time in UTC. Dealing with time zones has been very challenging with SpEC output.

nilsdeppe (changes requested):

src/Parallel/PhaseControl/CheckpointAndExitAfterWallclock.hpp
nilsvu: ok. even better would be (also) storing the timing in some data file, rather than the log file. not sure where yet though (I was thinking reductions file attributes).

nilsdeppe: Yea, that's a good idea, but I'm also not sure where. We'll need to update the startup info we print (usually into spectre.out) anyway to include things like Slurm info. I  did this for spec and will port it to spectre

nilsvu: Maybe we just parse timing etc out of the log file in a postprocessing step when a simulation is complete, and store in a yaml or json metadata file that will then interface with the SXS catalog.

nilsvu: Btw, Charm currently outputs date and time at startup in local time. So when you update the startup info, also add the UTC date and time please.

nilsvu:
@nilsdeppe I added a function using Boost to get the UTC time as nicely formatted string because C++ is sooo needlessly complicated sometimes... this would be easier in C++20 but the features are not available yet in all compilers.

nilsdeppe (approved):

PR #6467, opened 2025-02-01
Check SpEC version when importing

nilsvu:
## Proposed changes

Check the SpEC version when import Python modules to make sure it's compatible (of course the real solution is to port SpEC modules into SpECTRE when we need them or factor the code into a separate repository, but that requires some work and comes with a maintenance burden keeping the two codes in sync).

### Upgrade instructions

To use eccentricity control, make sure you're pointing to a recent enough SpEC version (Aug 2024).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Python/CheckSpecImport.py
+++ b/support/Python/CheckSpecImport.py
@@ -1,0 +1,40 @@
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+from pathlib import Path
+from typing import Optional
+
+
+def check_spec_import(contains_commit: Optional[str] = None):
+    """Check that the SpEC Python modules can be imported.
+
+    Arguments:
+      contains_commit: If specified, check that the SpEC repository contains
+        the specified commit. This can be used as a version check to ensure
+        that the installed SpEC version is compatible with the code that is
+        importing it.
+    """
+    try:
+        import Utils as spec_utils
+    except ImportError:
+        raise ImportError(
+            "Importing from SpEC failed. Make sure you have pointed "
+            "'-D SPEC_ROOT' to a SpEC installation when configuring the build "
+            "with CMake."
+        )
+    if contains_commit:
+        import subprocess
+
+        spec_repo_dir = Path(spec_utils.__file__).parent.parent.parent
+        result = subprocess.run(
+            ["git", "merge-base", "--is-ancestor", contains_commit, "HEAD"],
+            cwd=spec_repo_dir,
+            capture_output=True,
+            text=True,
+        )
+        if result.returncode != 0:
+            raise ImportError(
+                "SpEC version mismatch: requested revision"
+                f" {contains_commit} is not in the SpEC repository at"
+                f" {spec_repo_dir}."
+            )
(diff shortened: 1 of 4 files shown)


knelli2:
Will this be superceded by #6471? @nilsvu @nilsdeppe

nilsvu:
no I don't think so. rebased.

knelli2 (changes requested):
CI very unhappy

knelli2 (dismissed):

knelli2:
@nilsdeppe Did you want to take a look at this? Check that it works on mbot?

nilsdeppe:
Yes, please don't merge this because it'll break mbot and I don't have time
to upgrade spec right now.

nilsvu:
It shouldn't break mbot, only the unit tests that try to import spec will fail (ecc control and initial orbital params).

support/Python/requirements.txt
```diff
@@ -16,6 +16,8 @@
 # Click: to compose our Python CLI, load subcommands lazily, and for
 # autocompletion
 click
+# GitPython: to interact with Git repositories, e.g. SpEC version check
+GitPython
```
nilsdeppe: Let's not continue ballooning our dependencies. It's just a git call. See my comment in the python file for how to do this.

support/Python/CheckSpecImport.py
```diff
@@ -0,0 +1,36 @@
...
+    if contains_commit:
+        spec_repo = git.Repo(Path(spec_utils.__file__).parent.parent.parent)
+        try:
+            assert spec_repo.is_ancestor(contains_commit, spec_repo.head.commit)
+        except (AssertionError, git.exc.GitCommandError):
+            raise ImportError(
+                "SpEC version mismatch: requested revision"
+                f" {contains_commit} is not in the SpEC repository at"
+                f" {spec_repo.working_dir}."
+            )
```
nilsdeppe:
```python
-    try:
         import Utils as spec_utils
     except ImportError:
         raise ImportError(
             "Importing from SpEC failed. Make sure you have pointed "
             "'-D SPEC_ROOT' to a SpEC installation when configuring the build "
             "with CMake."
         )
     if contains_commit:
!        import subprocess
!
!        result = subprocess.run(
!            [
!                "git",
!                "merge-base",
!                "--is-ancestor",
!                str(contains_commit),
!                "HEAD",
!            ],
!            cwd=str(Path(spec_utils.__file__).parent.parent.parent),
!            capture_output=True,
!            text=True,
!        )
!        if result.returncode != 0:
             raise ImportError(
                 "SpEC version mismatch: requested revision"
                 f" {contains_commit} is not in the SpEC repository at"
!                f" {Path(spec_utils.__file__).parent.parent.parent}."
             )
```
Just do this. It's really annoying having the dependency list continually grow for trivial features.

nilsdeppe (changes requested):
I've updated mbot to work with this as long as it doesn't use the git python package.

support/Python/CheckSpecImport.py
nilsvu: Ok done

nilsdeppe (approved):
This is working as-is on mbot. Thanks :)

PR #6468, opened 2025-02-03
BBH pipeline: write eccentricity params to file, add CLI to compute params & plot

nilsvu:
## Proposed changes

Creates plots of the path through orbital parameter space like this:
ecc

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/PlotEccentricityControl.py
+++ b/src/Visualization/Python/PlotEccentricityControl.py
@@ -1,0 +1,159 @@
+#!/usr/bin/env python
+
+# Distributed under the MIT License.
+# See LICENSE.txt for details.
+
+import logging
+from pathlib import Path
+from typing import Sequence, Union
+
+import click
+import matplotlib.pyplot as plt
+import numpy as np
+import pandas as pd
+import yaml
+from matplotlib.colors import LogNorm
+from scipy.interpolate import CloughTocher2DInterpolator
+
+from spectre.Visualization.Plot import (
+    apply_stylesheet_command,
+    show_or_save_plot_command,
+)
+
+logger = logging.getLogger(__name__)
+
+
+def plot_eccentricity_control(
+    ecc_params_files: Sequence[Union[str, Path]],
+):
+    """Plot eccentricity control iterations.
+
+    \f
+    Arguments:
+      ecc_params_files: YAML files containing the output of
+        'eccentricity_control_params()'.
+
+    Returns:
+      The figure containing the plot.
+    """
+    # Load the eccentricity parameters
+    ecc_params = []
+    for ecc_params_file in ecc_params_files:
+        with open(ecc_params_file, "r") as open_file:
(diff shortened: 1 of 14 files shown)


support/Pipelines/Bbh/EccentricityControl.py
```diff
@@ -82,42 +81,17 @@ def eccentricity_control(
...
-    df = pd.DataFrame(data)
-    # Print header line
-    print("=" * 40)
-    # Display table
-    print(df.to_string(index=False))
-    print("=" * 40)
-
     # Stop eccentricity control if eccentricity is below threshold
-    if eccentricity < 0.001:
+    if ecc_params["Eccentricity"] < 0.001:
```
nilsdeppe, line 89: just a quick glance because I'll adapt a lot of this to BNS at some point. Should the threshold be specifiable somewhere? In SpEC we've had a harder time getting lower eccs for high q & chi runs. So much so that sometimes in BFI we just override the BFI-based target ecc.

nilsdeppe (commented):
Not a review, just small question/idea :)

support/Pipelines/Bbh/EccentricityControl.py
nilsvu: Yes definitely, we can just make it a parameter, that's very easy to do

knelli2:
@nilsvu I'll review after conflicts are fixed

nilsvu:
rebased

src/Visualization/Python/PlotEccentricityControl.py
```diff
@@ -0,0 +1,159 @@
...
+    )
+    plt.scatter(ecc_params["Omega0"], ecc_params["Adot0"], **scatter_kwargs)
+    cbar_ax.scatter(
+        np.repeat(0.5, len(ecc_params)),
+        ecc_params["Eccentricity"],
+        **scatter_kwargs,
+    )
+    for i in range(len(ecc_params)):
+        plt.annotate(
+            f"{i + 1}",
```
knelli2: In the directories, are these zero-indexed or one-indexed? If it's zero, this could be confusing.

nilsvu: The directories aren't currently indexed with `EccN` at all, I haven't gotten to that yet. I'm considering to switch to 1-indexing for pipeline steps and segments as well though, because it seems easier understand.

knelli2: To be honest, I think 1-indexing will be more confusing. Basically everything we do in code is 0-indexed and having to convert back and forth just seems problematic.

nilsvu: yea maybe. I switched to 0-indexing for consistency with the rest for now, and will reconsider later.

knelli2 (dismissed):

knelli2 (approved):

PR #6473, opened 2025-02-05
Add ability to list files to copy in input file tests

nilsdeppe:
## Proposed changes

This allows us to copy files from the source tree into the run directory of an input file tests. We probably won't need this a lot, but it can be helpful for testing reading parameters from additional files (e.g. SpEC ID type situations)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/RunInputFileTest.sh
+++ b/cmake/RunInputFileTest.sh
@@ -9,7 +9,9 @@
 # - $3: directory name
 # - $4: space-separated list of expected exit codes
 # - $5: "true" to check output files or "false" to skip the check
-# - $6: additional command-line arguments forwarded to the executable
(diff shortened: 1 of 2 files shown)


cmake/RunInputFileTest.sh
```diff
@@ -39,13 +52,17 @@ for expected_code in $4 ; do
     fi
 done
 
+echo "$5"
```
knelli2: Remove? Or label what the value is

nilsdeppe: Yea, removed. That was just debugging stuff :)

knelli2 (approved):

PR #6475, opened 2025-02-06
Add phase for disable rotation in BNS

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/BinaryCompactObject.hpp
+++ b/src/Domain/Creators/BinaryCompactObject.hpp
@@ -115,13 +115,13 @@ create_grid_anchors(const std::array<double, 3>& center_a,
  *
  * \par Notes:
  * - Object A is located to the right of the origin (along the positive x-axis)
- *   and Object B is located to the left of the origin.
+ *   and Object B is located to the left of the origin in the Grid frame.
  * - This domain offers some grid anchors. See
  *   `domain::creators::bco::create_grid_anchors` for which ones are offered.
  * - "Cutting plane" refers to the plane along which the domain divides into two
  *   hemispheres. The cutting plane always intersects the x-axis at the origin.
  * - The x-coordinate locations of the two objects should be chosen such that
- *   the center of mass is located at x=0.
+ *   the center of mass is located at x=0 at the initial time (typically t=0).
  * - The cubes are first constructed at the origin. Then, they are translated
  *   left/right by their Object's x-coordinate and offset depending on the cube
  *   length.
--- a/src/Parallel/Phase.cpp
+++ b/src/Parallel/Phase.cpp
@@ -19,6 +19,7 @@ std::vector<Phase> known_phases() {
           Phase::BuildMatrix,
           Phase::CheckDomain,
           Phase::Cleanup,
+          Phase::DisableRotationControl,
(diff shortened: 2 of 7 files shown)


src/Domain/Creators/BinaryCompactObject.hpp
```diff
@@ -115,13 +115,13 @@ create_grid_anchors(const std::array<double, 3>& center_a,
...
  * - Object A is located to the right of the origin (along the positive x-axis)
- *   and Object B is located to the left of the origin.
+ *   and Object B is located to the left of the origin in the Grid frame.
  * - This domain offers some grid anchors. See
  *   `domain::creators::bco::create_grid_anchors` for which ones are offered.
  * - "Cutting plane" refers to the plane along which the domain divides into two
  *   hemispheres. The cutting plane always intersects the x-axis at the origin.
  * - The x-coordinate locations of the two objects should be chosen such that
- *   the center of mass is located at x=0.
+ *   the center of mass is located at x=0 at t=0.
```
knelli2: Say `at initial time` because it doesn't always have to be t=0

src/Parallel/Phase.cpp
```diff
@@ -19,6 +19,7 @@ std::vector<Phase> known_phases() {
           Phase::BuildMatrix,
           Phase::CheckDomain,
           Phase::Cleanup,
+          Phase::DisableRotationControl,
```
knelli2, line 22: What do you think about naming this phase `DisableControlSystems` instead? That way we can use it in the future if any of our systems need to disable any control system

knelli2 (changes requested):

src/Parallel/Phase.cpp
nilsdeppe: So I thought about this a decent amount before submitting the PR and decided to keep it specific because I don't know yet if there will be multiple control systems that need to be disabled simultaneously. If not, then you need a different phase name for them anyway, or some way of tracking state of which one was and should be disabled. My take is that it's less likely to be confusing having the different name. As not-ideal as it is in terms of scaling the code, I think clarity is important since the control system interactions are quite complicated as-is.

src/Domain/Creators/BinaryCompactObject.hpp
nilsdeppe: Okay, updated!

src/Parallel/Phase.cpp
knelli2: Ok yes, this makes sense. I think if we do need to add more control systems to be disabled, we can probably do it in such a way that we'd only need one phase, but I haven't thought too much about it. We can cross that bridge when we get to it.

knelli2 (approved):

src/Parallel/Phase.cpp
nilsdeppe: Sounds good! Yea, I think it'll be clearer if/when we get there. I don't think for BNS we'll have anything else until we can do a BH+matter in the post-merger. Translation and can just track the global COM instead of the average of the two halves. And for BNS I think we only need Rotation+Translation 🤞

PR #6477, opened 2025-02-07
Properly extend shape map into interior region

### LINKED ISSUE #6451: Shape map incorrect inside excision
The shape map needs to be valid inside the excision so that we can support the transition to ringdown code. A previous PR (#6359) extended the transition functions within the excisions, but this was not everything needed. There are a couple problems

1. The shape map doesn't allow points at the center (https://github.com/sxs-collaboration/spectre/blob/f46bc1332ed5651e749041d49f12adf6ebe4c9fe/src/Domain/CoordinateMaps/TimeDependent/Shape.cpp#L485-L487) This needs to be relaxed.
2. The shape map incorrectly divides by the radius if the point is within the excison (https://github.com/sxs-collaboration/spectre/blob/f46bc1332ed5651e749041d49f12adf6ebe4c9fe/src/Domain/CoordinateMaps/TimeDependent/Shape.cpp#L244-L245). Instead, if the point is within the excision, the shape map should divide by the excision radius.
3. The shape map has no knowledge of what the excision radius is, and therefore can't currently address 2. However the transition functions do. I believe the changes in #5636 related to having the shape map handle the division by radius were a mistake and that the transition function should be the one to handle the division of the radius.
4. Once this is done, we can add the full grid to inertial map to the excision spheres.

knelli2:
This is only allowed when the inner surface is spherical and when we are using the "regular" transition (i.e. not reversed).

## Proposed changes

Fixes #6451.

A previous PR (#6359) attempted to extend the shape map into the excision, but did so incorrectly. This PR implements this correctly. In order to do this, the meaning of the transition function had to be changed slightly. Previously, the shape map was

$$
\tilde{r} = r \left(1 - \frac{f(r, \theta, \phi)}{r} \sum_{\ell,m} \lambda_{\ell m}Y_{\ell m}(\theta, \phi)\right)
$$

and the transition function was $f(r, \theta, \phi)$. The functional form of the shape map hasn't changed, but now it reads

$$
\tilde{r} = r \left(1 - G(r, \theta, \phi) \sum_{\ell,m} \lambda_{\ell m}Y_{\ell m}(\theta, \phi)\right)
$$
$$
G(r, \theta, \phi) = \frac{f(r, \theta, \phi)}{r}
$$

where $G(r, \theta, \phi)$ is the transition. The $1/r$ factor was just absorbed into the transition function because the full $G$ is different inside the excision, not just $f$.

Also, the transition function inside the excisions is

$$
G(r, \theta, \phi) = \frac{r^2}{r_{\text{min}}^3}
$$

which is different than it is in SpEC. The transition defined in SpEC causes the jacobian to be undefined at the center of the shape map, which means the full frame velocity of all the maps can't be computed at the center of the shape map. This new transition inside the excision allows for this and has the correct behavior.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
+++ b/src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
@@ -11,6 +11,7 @@
 #include <pup.h>
 #include <pup_stl.h>
 #include <string>
+#include <type_traits>
 #include <unordered_set>
 
 #include "DataStructures/DataVector.hpp"
@@ -21,10 +22,12 @@
 #include "NumericalAlgorithms/SphericalHarmonics/SpherepackIterator.hpp"
 #include "Utilities/ContainerHelpers.hpp"
 #include "Utilities/DereferenceWrapper.hpp"
+#include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/Gsl.hpp"
 #include "Utilities/Serialization/PupStlCpp17.hpp"
+#include "Utilities/StdArrayHelpers.hpp"
 #include "Utilities/StdHelpers.hpp"
 
 namespace domain::CoordinateMaps::TimeDependent {
@@ -47,8 +50,7 @@ void Shape::jacobian_helper(
     gsl::not_null<tnsr::Ij<T, 3, Frame::NoFrame>*> result,
     const ylm::Spherepack::InterpolationInfo<T>& interpolation_info,
     const DataVector& extended_coefs, const std::array<T, 3>& centered_coords,
-    const T& radial_distortion, const T& one_over_radius,
-    const T& transition_func_over_radius) const {
+    const T& radial_distortion, const T& transition_func) const {
   const auto angular_gradient =
       extended_ylm_.gradient_from_coefs(extended_coefs);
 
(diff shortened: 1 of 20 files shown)


knelli2:
Ok this is ready for review (maybe @nilsvu ?). Just an FYI for @nikwit that this adds the shape map to the ExcisionSpheres, but if you're only evaluating it at the center of the excision, then this shouldn't behave any differently than before when the shape map wasn't included.

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -77,92 +79,42 @@ void Shape::jacobian_helper(
...
-  get<1, 1>(*result) += 1. - radial_distortion * transition_func_over_radius;
-  get<2, 2>(*result) += 1. - radial_distortion * transition_func_over_radius;
+    // interpolate the cartesian gradient to the thetas and phis of the
+    // `source_coords`
+    extended_ylm_.interpolate(make_not_null(&gsl::at(target_gradient, i)),
+                              cartesian_gradient.get(i).data(),
+                              interpolation_info);
+  }
+
+  auto transition_func_over_radius =
```
nilsdeppe, line 101: So this is `G` or `G/r`? Because my understanding from the dox change is this is `G/r`

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.hpp
```diff
@@ -243,26 +276,21 @@ namespace domain::CoordinateMaps::ShapeMapTransitionFunctions {
  * \parblock
  *
  * \note If \p reverse is true, then the value multiplying $\Sigma$ in the
- * numerator is now $-|\vec x_0 - \vec P|$ and in the denomintor $\Sigma$ picks
+ * numerator is now $-|\vec x_0 - \vec P|$ and in the denominator $\Sigma$ picks
```
nilsdeppe, line 279: demonator? 😅 (JK I'm just teasing!)

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
```diff
@@ -37,68 +47,197 @@ SphereTransition::SphereTransition(const double r_min, const double r_max,
...
+
+      // Numerical Recipes (3rd edition) pg 228, eq 5.6.10
+      const double R = 0.5 * factor;
+      const double Q = factor / 3.0;
+      const bool multiple_real_roots = square(R) < cube(Q);
+
+      if (multiple_real_roots) {
+        // Numerical Recipes (3rd edition) pg 228, eqs 5.6.11 - 5.6.12
+        const double theta = acos(R / sqrt(cube(Q)));
+        std::vector<double> roots{
```
nilsdeppe, line 96: why not use an array? or a static vector? `remove_if` returns an iterator to the end, so you can use it on an array.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -385,36 +453,91 @@ std::optional<double> Wedge::original_radius_over_radius(
...
+
+      // Numerical Recipes (3rd edition) pg 228, eq 5.6.10
+      const double R = 0.5 * factor;
+      const double Q = factor / 3.0;
+      const bool multiple_real_roots = square(R) < cube(Q);
+
+      if (multiple_real_roots) {
+        // Numerical Recipes (3rd edition) pg 228, eqs 5.6.11 - 5.6.12
+        const double theta = acos(R / sqrt(cube(Q)));
+        std::vector<double> roots{
```
nilsdeppe, line 512: same question about array

nilsdeppe (changes requested):

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
knelli2: You are correct, this is `G/r`. I'll add a small comment here to clarify

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
knelli2: I had assumed `remove_if` would resize the container, but I see I was incorrect. I think what I actually want is to use `std::erase_if` so the vector actually contains the positive roots.

src/Domain/CoordinateMaps/TimeDependent/Shape.cpp
```diff
@@ -77,92 +79,43 @@ void Shape::jacobian_helper(
...
+    // interpolate the cartesian gradient to the thetas and phis of the
+    // `source_coords`
+    extended_ylm_.interpolate(make_not_null(&gsl::at(target_gradient, i)),
+                              cartesian_gradient.get(i).data(),
+                              interpolation_info);
+  }
+
+  // G / r
+  auto transition_func_over_radius =
+      transition_func_->operator()(centered_coords, {1});
```
nilsvu, line 102: (optional) don't need the curly braces I think

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp
```diff
@@ -68,13 +73,20 @@ class ShapeMapTransitionFunction : public PUP::able {
...
+   * $r$.
+   *
+   * \details If \p one_over_radius_power has a value, then divide $G$ by $r$ to
+   * that power. This is done here because the transition function knows how to
+   * handle points in various regions of the map.
    */
   virtual double operator()(
-      const std::array<double, 3>& source_coords) const = 0;
+      const std::array<double, 3>& source_coords,
+      const std::optional<size_t>& one_over_radius_power) const = 0;
```
nilsvu, line 86: (optional) default to std::nullopt

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
```diff
@@ -37,68 +48,197 @@ SphereTransition::SphereTransition(const double r_min, const double r_max,
...
 }
+
 DataVector SphereTransition::operator()(
-    const std::array<DataVector, 3>& source_coords) const {
-  return call_impl<DataVector>(source_coords);
+    const std::array<DataVector, 3>& source_coords,
+    const std::optional<size_t>& one_over_radius_power) const {
+  DataVector result = source_coords[0];
+  // Go point by point
+  for (size_t i = 0; i < source_coords[0].size(); i++) {
```
nilsvu: The performance of this might actually be relevant, so if it's easy enough I'd stick with Blaze for this computation instead of doing the loop yourself.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
```diff
@@ -37,68 +48,197 @@ SphereTransition::SphereTransition(const double r_min, const double r_max,
...
   const double mag = magnitude(target_coords);
+  // If we are at the center, the radius is the same
+  if (UNLIKELY(equal_within_roundoff(mag, 0.0))) {
+    return interior_ ? std::optional{1.0} : std::nullopt;
+  }
+
+  // a_ being positive is a sentinel for reversed.
+  // If we aren't reversed, check near or within r_min_.
+  if (a_ < 0.0) {
+    if (mag + radial_distortion < (1.0 - eps_) * r_min_) {
```
nilsvu, line 76: Why the `+ radial_distortion` here?

I think you have to be careful that you compare r_min and r_max only to the original radius, not the target radius. E.g. in line 157 there's a comparison to the target radius, which I think could be a problem.

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
```diff
@@ -37,68 +48,197 @@ SphereTransition::SphereTransition(const double r_min, const double r_max,
...
-}
-
-template <typename T>
-std::array<T, 3> SphereTransition::gradient_impl(
-    const std::array<T, 3>& source_coords) const {
-  const T mag = magnitude(source_coords);
-  check_magnitudes(mag, true);
-  return a_ * source_coords / mag;
+  auto result = source_coords;
+  for (size_t i = 0; i < source_coords[0].size(); i++) {
```
nilsvu: Also here, probably better to stick with Blaze

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -313,55 +332,95 @@ std::array<T, 3> Wedge::compute_lambda_gradient(
...
+             centered_coords_magnitude,
+             static_cast<int>(1 + one_over_radius_power.value_or(0_st)));
+}
+
+DataVector Wedge::operator()(
+    const std::array<DataVector, 3>& source_coords,
+    const std::optional<size_t>& one_over_radius_power) const {
+  DataVector result = source_coords[0];
+  // Go point by point to avoid the center point
+  for (size_t i = 0; i < source_coords[0].size(); i++) {
```
nilsvu: Blaze?

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/Wedge.cpp
```diff
@@ -544,17 +675,34 @@ std::array<T, 3> Wedge::gradient_impl(
...
+               one_over_distance_difference *
+               one_over_centered_coords_magnitude * (reverse_ ? -1.0 : 1.0) -
+           source_coords * (*this)(source_coords, {2});
+  }
+}
+
+std::array<DataVector, 3> Wedge::gradient(
+    const std::array<DataVector, 3>& source_coords) const {
+  auto result = source_coords;
+  for (size_t i = 0; i < source_coords[0].size(); i++) {
```
nilsvu: Blaze?

nilsvu (changes requested):

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
knelli2: Well for each point, I'd still need to do the checks I do in the `double` implementation since this can be called with any set of points and to ensure everything is consistent which would require this to still be a loop.

nilsvu: Oh hide the checks behind `#ifdef SPECTRE_DEBUG` then their performance doesn't matter and you can do loops in them

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
knelli2: It's undoing the distortion assuming we are at the f=1 surface. Therefore if this condition is true, we are guaranteed to be within the excision and if not, we're outside of it. I ported this check over from SpEC.

Below on 157 should still be the target radius, but I forgot to add the radial distortion since I'm trying to see if we are exactly at the excision (because then the formula is simpler). In the Wedge it is correct

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp
knelli2: Quick google search says having default parameters in pure virtual functions is not recommended so I'll leave it as is.

knelli2:
@nilsvu Pushed fixups

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
nilsvu: Please do these like the `gradient_impl` function you did below, that's also how these functions were implemented before you split them into separate `double` and `DataVector` overloads in this PR. Otherwise the code can do something different in debug vs release, which seems very hard to debug 😅

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/ShapeMapTransitionFunction.hpp
nilsvu: ok

src/Domain/CoordinateMaps/TimeDependent/ShapeMapTransitionFunctions/SphereTransition.cpp
knelli2: Ok done

nilsvu (dismissed):

knelli2:
@nilsvu squashed

nilsvu (approved):

nilsvu:
@nilsdeppe did you want to look at this again before we merge?

nilsdeppe:
No go ahead without me

PR #6480, opened 2025-02-14
Merge stepping actions in CCE

wthrowe:
Simplifies control flow by getting rid of jumping around the action list.  Will simplify some changes to the step-size logic in an upcoming PR.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (dismissed):

wthrowe:
CI appears to have run out of memory.  I don't think there's any reason this PR would significantly increase compiler memory use, but we may have been on the edge before and a small increase tipped it over or something.  Although it's odd that it happened on all three compiler versions.

I don't see any obvious easy savings in `ComputeTimeDerivativeImpl.tpp`.  (Unsurprising, since I'm not the first to look for them.)  I may try splitting the instantiations across more files.

nilsdeppe (approved):
The splitting up of the file more seems to have worked. Thanks!

PR #6482, opened 2025-02-15
Allow specifying 'None' for control sys in input file

knelli2:
## Proposed changes

If a control system is inactive, it doesn't make a ton of sense to have to specify parameters that will never be used. This implements a similar system to the time dependent options for the BCO and Sphere where you can specify `None` for a control system to disable it. This also decreases the length in input files like BNS which won't use most control systems.

There was no functionality change in how the control system works. This was purely cosmetic.

### Upgrade instructions

To turn a control system off in the input file, use
```yaml
ControlSystems:
  Expansion: None
```
When keeping a control system active, keep all the same previous options, except remove the `IsActive:` option. I.e.
```yaml
ControlSystems:
  Expansion:
    Averager:
      AverageTimescaleFraction: 0.25
      Average0thDeriv: false
    Controller:
      UpdateFraction: 0.03
    TimescaleTuner:
      InitialTimescales: [0.2]
      MinTimescale: 1.0e-2
      MaxTimescale: 10.0
      IncreaseThreshold: 2.5e-4
      DecreaseThreshold: 1.0e-3
      IncreaseFactor: 1.01
      DecreaseFactor: 0.98
    ControlError:
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Tags/FunctionsOfTimeInitialize.hpp
+++ b/src/ControlSystem/Tags/FunctionsOfTimeInitialize.hpp
@@ -17,6 +17,7 @@
 #include "Domain/Creators/Tags/FunctionsOfTime.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
 #include "Domain/FunctionsOfTime/Tags.hpp"
+#include "Options/Auto.hpp"
 #include "Time/OptionTags/InitialTime.hpp"
 #include "Time/OptionTags/InitialTimeStep.hpp"
 #include "Utilities/TMPL.hpp"
@@ -74,14 +75,27 @@ struct FunctionsOfTimeInitialize : domain::Tags::FunctionsOfTime,
           tmpl::list<>>,
       domain::OptionTags::DomainCreator<Metavariables::volume_dim>>;
 
+  /// @{
   /// This version of create_from_options is used if the metavariables did
   /// define control systems
   template <typename Metavariables, typename... OptionHolders>
   static type create_from_options(
       const std::unique_ptr<::DomainCreator<Metavariables::volume_dim>>&
           domain_creator,
-      const int measurements_per_update,
-     const double initial_time, const OptionHolders&... option_holders) {
+      const int measurements_per_update, const double initial_time,
+      const Options::Auto<OptionHolders,
+                          Options::AutoLabel::None>&... option_holders) {
+    return create_from_options<Metavariables>(
+        domain_creator, measurements_per_update, initial_time,
+        static_cast<const std::optional<OptionHolders>&>(option_holders)...);
+  }
+
+  template <typename Metavariables, typename... OptionHolders>
+  static type create_from_options(
+      const std::unique_ptr<::DomainCreator<Metavariables::volume_dim>>&
+          domain_creator,
+      const int measurements_per_update, const double initial_time,
+      const std::optional<OptionHolders>&... option_holders) {
     const auto initial_expiration_times =
         control_system::initial_expiration_times(
             initial_time, measurements_per_update, domain_creator,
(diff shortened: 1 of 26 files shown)


src/Parallel/CreateFromOptions.hpp
```diff
@@ -3,19 +3,31 @@
...
+
 template <typename Metavariables, typename Tag, typename... OptionTags,
           Requires<Tag::pass_metavariables> = nullptr>
 typename Tag::type create_initialization_item_from_options(
     const tuples::TaggedTuple<OptionTags...>& options) {
   return tuples::apply<typename Tag::template option_tags<Metavariables>>(
       [](const auto&... option) {
-        return Tag::template create_from_options<Metavariables>(option...);
+        return Tag::template create_from_options<Metavariables>(
+            get_option(option)...);
```
wthrowe: I don't like special-casing `Auto` here.  It isn't explicitly converted in the more common option parsing locations, and this should be kept consistent with those.

It is unfortunate that C++ can't figure out to do the implicit conversion in some cases, but I think it is the job of the class requesting an `Auto` to be able to handle receiving one.

src/ControlSystem/TimescaleTuner.hpp
```diff
@@ -151,7 +151,7 @@ class TimescaleTuner {
 
   DataVector timescale_;
   bool timescales_have_been_set_{false};
-  double initial_timescale_{std::numeric_limits<double>::signaling_NaN()};
+  double initial_timescale_;
```
wthrowe: I don't understand what this commit is doing.

wthrowe (changes requested):

src/ControlSystem/TimescaleTuner.hpp
knelli2: If you default constructed a TimescaleTuner and then compared it with `==`, you'd get an FPE because of the signaling NaN.

src/Parallel/CreateFromOptions.hpp
knelli2: Yeah this was very unfortunate that it couldn't figure it out. I'll remove this commit and try some other magic in the tags themselves.

src/ControlSystem/TimescaleTuner.hpp
wthrowe: So now you're comparing uninintialized memory instead?

knelli2: I'll set it to 0

src/ControlSystem/Tags/FunctionsOfTimeInitialize.hpp
```diff
@@ -74,8 +74,23 @@ struct FunctionsOfTimeInitialize : domain::Tags::FunctionsOfTime,
...
+      Requires<tmpl2::flat_all_v<tt::is_a_v<Options::Auto, OptionHolders>...>> =
+          nullptr>
+  static type create_from_options(
+      const std::unique_ptr<::DomainCreator<Metavariables::volume_dim>>&
+          domain_creator,
+      const int measurements_per_update, const double initial_time,
+      const OptionHolders&... option_holders) {
+    return create_from_options<Metavariables>(
+        domain_creator, measurements_per_update, initial_time,
+        detail::get_option_holder(option_holders)...);
```
wthrowe: Are these ever going to be anything except `Auto`s?  If not, just cast them to optionals, or add a method to `Auto` to do the cast so you don't have to deal with the template argument.

wthrowe (changes requested):

wthrowe:
Looks good.  Squash.

wthrowe (approved):

PR #6486, opened 2025-02-19
Always redo step if size change is desired

wthrowe:
The reason for accepting steps and changing the next is to avoid
discarding the work involved in taking steps, but we only change the
size at the start of the step, so the only work that has been done is
a single update of the variables.  Taking a larger step to avoid an
extra RHS evaluation is better.

Only allow the size to change once per step to avoid potential
infinite loops.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/Time/TakeStep.hpp
```diff
@@ -37,16 +37,18 @@ void take_step(const gsl::not_null<db::DataBox<DbTags>*> box) {
...
         update_u<System>(box);
-      } while (not change_step_size<StepChoosersToUse>(box));
+      }
       db::mutate<Tags::AdaptiveSteppingDiagnostics>(
           [&](const gsl::not_null<AdaptiveSteppingDiagnostics*> diags,
               const TimeDelta& new_step) {
-            diags->number_of_step_rejections += step_attempts - 1;
+            if (not accepted) {
+              ++diags->number_of_step_rejections;
+            }
```
nilsdeppe, line 47: Do I understand correctly there's a subtle logic difference here where now we will only change the step size once? That means if our estimate of the step size change to get to a target error was incorrect, we would be above our target error, right? The documentation I think needs to be updated to reflect this. It looks like the last bit says it loops until accepted.

nilsdeppe (changes requested):
You can squash immediately

teukolsky:
I'm confused by the explanation: If I'm at time t and take a step of size delta_1, I get to t+delta_1. If I accept the step and take a bigger step of size delta_2, I get to t+delta_1+delta_2. I've made a total of 2 rhs evaluations in Adams-Bashforth. If I discard the first step, I get to t+delta_2 instead, with still 2 rhs evaluations. Obviously I'm not understanding something. Can you please clarify the explanation? Thanks!

src/Time/TakeStep.hpp
wthrowe: Yes, the logic has changed (as mentioned in the commit message and PR description).  I've pushed a fixup updating the documentation by removing the second paragraph, since there didn't seem to be anything interesting left there after the looping discussion was removed.

And correct, we will now ignore an estimate saying the error is too large after a change.  Note that AdamsMoultonPc doesn't supply an error estimate at the start of the step, so this only affects AdamsBashforth.

wthrowe:
@teukolsky The RHS evaluations for Adams-Bashforth are independent of the step size.  The step from t to t+delta_2 reuses the evaluation from the discarded step to t+delta_1.

nilsdeppe (commented):
LGTM, please squash!

nilsdeppe (approved):

PR #6487, opened 2025-02-19
Add Skew control system to BBH

knelli2:
## Proposed changes

Part of #4603.

~Depends on (but doesn't include) #6482 .~

### Upgrade instructions

Any input file for the BBH executable must now also have a `Skew:` block under the `ControlSystems:` group with all the same options as the other control systems.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


knelli2:
Still dependent, but now on #6493 (not included)

knelli2:
No longer dependent and ready for review

src/ControlSystem/UpdateTimescaleTuner.hpp
```diff
@@ -0,0 +1,78 @@
...
+          function_of_time_name, time, old_timescale,
+          MakeString{} << suggested_timescale, min(tuner->current_timescale()));
+    }
+
+    // This reset call is ok because if there was a discontinuous change
+    // above after calculating the control error, the control error class was
+    // already reset so this reset won't do anything. If there wasn't a
+    // discontinuous change, then this will only affect the suggested
+    // timescale, which we always want to reset.
+    control_error->reset();
```
wthrowe: I find it surprising that `get_suggested_timescale() { return nullopt; }` is different from not implementing the function at all, and the ControlError protocol doesn't contain `reset`, so it's not clear to me what this actually does, in general.

src/ControlSystem/ControlErrors/Skew.hpp
```diff
@@ -0,0 +1,277 @@
...
+                          (weight_a + weight_b) -
+                      (1.0 - temporal_transition_function) * func;
+    }
+
+    return control_error;
+  }
+
+ private:
+  std::optional<double> suggested_timescale_;
+  // not pupped
```
wthrowe: I think I know why these member variables exist, but please explain in the comment.

src/ControlSystem/ControlErrors/Skew.hpp
```diff
@@ -0,0 +1,277 @@
...
+
+    // Hardcoded value used in SpEC
+    constexpr double activation_threshold = 0.0025;
+
+    const auto& functions_of_time =
+        Parallel::get<domain::Tags::FunctionsOfTime>(cache);
+    const DataVector func =
+        functions_of_time.at(function_of_time_name)->func(time)[0];
+
+    DataVector control_error = -func;
```
wthrowe: I can't tell whether you care about memory allocations in this function or not.  I assumed that's why you're caching things between calls, but then you allocate three extra size-2 DataVectors.

src/ControlSystem/Systems/Skew.hpp
```diff
@@ -0,0 +1,121 @@
...
+ */
+template <size_t DerivOrder, typename Measurement>
+struct Skew : tt::ConformsTo<protocols::ControlSystem> {
+  static constexpr size_t deriv_order = DerivOrder;
+
+  static std::string name() {
+    return pretty_type::short_name<Skew<DerivOrder, Measurement>>();
+  }
+
+  // Skew only has one component so just make it "Skew"
```
wthrowe: It appears to have two named "Y" and "Z".

src/ControlSystem/ControlErrors/Skew.hpp
```diff
@@ -0,0 +1,277 @@
...
+          }
+        };
+
+    // These are the values we need at the theta/phi described above
+    DataVector inclination_angle_a{2, 0.0};
+    DataVector inclination_angle_b{2, 0.0};
+    std::array<double, 3> intersection_coord_a{};
+    std::array<double, 3> intersection_coord_b{};
+    std::array<double, 3> intersection_normal_one_form_a{};
+    std::array<double, 3> intersection_normal_one_form_b{};
```
wthrowe: These are only used as temporaries inside `set_intersection`.  Move them there.

wthrowe (changes requested):

knelli2:
@wthrowe pushed fixups

src/ControlSystem/Protocols/ControlError.hpp
```diff
@@ -45,6 +50,12 @@ struct has_signature
 ///   `domain::ObjectLabel`s. These are the objects that will require the
 ///   `domain::Tags::ObjectCenter`s tags to be in the GlobalCache for this
 ///   control system to work.
+/// - a function with signature `const std::optonal<double>&
+///   get_suggested_timescal() const;` which returns a potential suggested
```
wthrowe: Is there a reason for return-by-reference here?  An `optional<double>` isn't large, and returning by value would eliminate all the "unused" variables added here.

Also: typo "timescal".

src/ControlSystem/ControlErrors/Skew.hpp
```diff
@@ -252,19 +244,23 @@ struct Skew : tt::ConformsTo<protocols::ControlError> {
...
-                          (weight_a + weight_b) -
-                      (1.0 - temporal_transition_function) * func;
+      return temporal_transition_function *
+                 (weight_a * inclination_angle_a_ +
+                  weight_b * inclination_angle_b_) /
+                 (weight_a + weight_b) -
+             (1.0 - temporal_transition_function) *
+                 function_of_time->func(time)[0];
+    } else {
+      return -function_of_time->func(time)[0];
```
wthrowe: [optional] We're getting into fairly high complication for minimal returns, but I think you save another allocation if you do
```c++
DataVector control_error = std::move(function_of_time->func(time)[0]);
if (temporal_transition_function > activation_threshold) {
  ...;
  control_error = stuff - (1.0 - temporal_transition_function) * control_error;
} else {
  control_error *= -1.0;
}
return control_error;
```
(I am kind of disappointed with the STL that that `std::move` is necessary.  You wouldn't need it with a tuple.)

wthrowe (changes requested):

knelli2:
@wthrowe pushed fixups

src/ControlSystem/ControlErrors/Skew.hpp
```diff
@@ -244,15 +246,16 @@ struct Skew : tt::ConformsTo<protocols::ControlError> {
...
+                          (weight_a * inclination_angle_a_ +
+                           weight_b * inclination_angle_b_) /
+                          (weight_a + weight_b) -
+                      func;
     } else {
-      return -function_of_time->func(time)[0];
+      control_error *= -1.0;
     }
+
+    return control_error;
```
wthrowe: Either `return func;` or `return std::move(control_error);`.  (Lvalue) references act differently in return statements.

wthrowe (changes requested):
Squash.

knelli2:
@wthrowe Done! Thanks for the review

wthrowe (approved):

Issue #3713, opened 2021-12-31
Add regression tests


nilsvu:
# Feature request:

I think it would be useful to add tests that don't run on every PR, but regularly on `develop`. These tests can use more resources than those running on every PR. We could run them on PR merges, on a schedule (daily/weekly/monthly), before releases, and/or on request. Here are some ideas:

- [ ] Multi-node tests that run on a cluster using a self-hosted Github Actions runner.
- [ ] Test more build environments, such as building a full dependency stack with Spack. Or test a `spack install spectre`.
- [ ] Reproduce a selected set of published results, keeping corresponding input files up-to-date.
- [ ] Run a set of tutorial notebooks.

nilsdeppe:
I think this is a great suggestion! I would guess the first (running on a supercomputer) is probably trickiest. Doing a spack test sounds great and not too difficult to do. Would be especially nice before releases. Published results will be a bit tricky too since 3d evolution problems are often absurdly expensive. However, even doing something between are input file tests and a full run would be good. Eg if 10 time steps at a lowered resolution should give `X` as the L2 norm of the density we keep track of that and if things start to differ the build fails. Probably check some norm over the domain of evolved/solved variables is a reasonable way to test most things. CCE/wave extraction would be a bit trickier but should still be possible as well.

I think having tutorials or notebooks have some way of being tested would be really fantastic. Having broken tutorials is something we will want to avoid. At least, having the code in the be correct would be good :)

PR #6490, opened 2025-02-20
BBH pipeline: set time bounds in ecc control

nilsvu:
## Proposed changes

Run an ecc control iteration for 2-3 orbits and determine tmin for the eccentricity fit dynamically.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/Inspiral.py
+++ b/support/Pipelines/Bbh/Inspiral.py
@@ -378,6 +378,14 @@ def start_inspiral(
             refinement_level=refinement_level,
             polynomial_order=polynomial_order,
         )
+
+    # Set final time for eccentricity control to 2-3 orbits. This can be set
+    # more dynamically in the future.
+    if eccentricity_control:
+        inspiral_params["FinalTime"] = (
+            500 + 5 * np.pi / inspiral_params["InitialAngularVelocity"]
+        )
+
     logger.debug(f"Inspiral parameters: {pretty_repr(inspiral_params)}")
 
     # Resolve directories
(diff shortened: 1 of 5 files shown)


support/Pipelines/Bbh/Inspiral.py
```diff
@@ -372,6 +372,14 @@ def start_inspiral(
             refinement_level=refinement_level,
             polynomial_order=polynomial_order,
         )
+
+    # Set final time for eccentricity control to 2-3 orbits. This can be set
+    # more dynamically in the future.
+    if eccentricity_control:
+        inspiral_params["FinalTime"] = (
+            500 + 5 * np.pi * inspiral_params["InitialAngularVelocity"]
```
nilsdeppe: Shouldn't this be divided by Omega to get the units right?

nilsdeppe (changes requested):

support/Pipelines/Bbh/Inspiral.py
nilsvu: Ugh yes thanks for catching. Fixed.

nilsdeppe: Thanks! Is there a way to catch this sort of thing? I'm worried we have other latent bugs similar to this one.

nilsvu: There are tests for some of these input file parameters but they are mostly "write the same thing in the test that's written in src" so they're not super useful. We really need pipeline tests like #3713 to run a BBH pipeline on CI to catch things like this.

nilsdeppe: Okay, sounds reasonable. That's what I figured. Thanks! :)

nilsdeppe (approved):

PR #6492, opened 2025-02-20
Make jemalloc compatible with Pybindings

nilsvu:
## Proposed changes

Gets rid of the `LD_PRELOAD` that we had to do when using Pybindings with jemalloc by compiling jemalloc with the flag `--disable-initial-exec-tls` (see https://github.com/jemalloc/jemalloc/issues/937).

### Upgrade instructions

If you are using jemalloc, compile it with the flag `--disable-initial-exec-tls` (see https://github.com/jemalloc/jemalloc/blob/dev/INSTALL.md). You don't need to do anything if you're using the system allocator (set `-D MEMORY_ALLOCATOR=SYSTEM`), or if you're running on a shared cluster like CaltechHPC or in the container.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


wthrowe:
What problem is this solving?  Forcing everyone to compile their own copy of jemalloc is kind of unpleasant.

cmake/SpectreSetupPythonPackage.cmake
```diff
@@ -33,10 +33,15 @@ configure_or_symlink_py_file(
...
-    " LD_PRELOAD=\${LD_PRELOAD}\${LD_PRELOAD:+:}${JEMALLOC_LIBRARIES}")
-endif()
+# At some point we needed to _preload_ jemalloc for the Python bindings to work,
+# otherwise we got "cannot allocate memory in static TLS block" errors. This is
+# fixed by compiling jemalloc with the flag `--disable-initial-exec-tls`, or by
+# using the system allocator (configure with `-D MEMORY_ALLOCATOR=SYSTEM`).
+# if(BUILD_PYTHON_BINDINGS AND "${JEMALLOC_LIB_TYPE}" STREQUAL SHARED)
+#   string(APPEND PYTHON_EXEC_ENV_VARS
+#     " LD_PRELOAD=\${LD_PRELOAD}\${LD_PRELOAD:+:}${JEMALLOC_LIBRARIES}")
+# endif()
```
knelli2, line 43: We should keep this, but somehow do a check to see if JEMALLOC was compiled with `--disable-initial-exec-tls` and if so, disable the LD_PRELOAD thing

knelli2 (changes requested):

nilsvu:
> What problem is this solving?

I ran into issues where the `LD_PRELOAD` interacted badly with things the `sxs` package is importing, specifically PN functions to get initial orbital parameters for BBH.

> Forcing everyone to compile their own copy of jemalloc is kind of unpleasant.

I think few people actually need to install jemalloc anyway. On clusters we have a shared installation of jemalloc (I installed it on CaltechHPC, and we need to recompile it on Ocean and Mbot). And on dev machines we don't care about the extra performance, so we can just set `MEMORY_ALLOCATOR=SYSTEM`. If you really want jemalloc, it's very easy to compile (`./configure --disable-initial-exec-tls && make && make install`). Maybe we should actually default to the system allocator and explicitly use jemalloc in the cluster envs.

cmake/SpectreSetupPythonPackage.cmake
nilsvu: I ran into issues with `LD_PRELOAD` (see comment below), so getting rid of it is I think the only option.

knelli2: Sure I think that's the right thing to do, but I don't agree with doing this abrupt change. As long as we change it on the clusters (which we have control over), we can keep the current behavior and add a deprecation warning saying we'll change it in a month or so. That'll give people time to update JEMALLOC if they need to.

nilsvu: I thought about doing that, but I think it's not necessary and I didn't find a way to determine if jemalloc was actually compiled with the flag or not. Options I considered were:

1. Try to import pybindings and emit a better error message than "cannot allocate memory in static TLS block". But we don't know which Py module will be imported first, and we can't add this error message to every Py module we have.
2. Add a cmake option like `JEMALLOC_PRELOAD` that is `ON` by default, set it to `OFF` on CaltechHPC, and emit a warning that the option will be removed soon. This seems too much effort for little gain, since people can just set `MEMORY_ALLOCATOR=SYSTEM` on personal machines and the few clusters we're running on we can easily update.
3. Change the default for `MEMORY_ALLOCATOR` to `SYSTEM` and explicitly set it to `JEMALLOC` on clusters. This is the best solution I think.

knelli2: You could compile and run the test case that was used in the original PR and see if it throws an error, similar to how we check `H5Pset_file_locking`:
```cpp
#include <dlfcn.h>
#include <stdio.h>

int main() {
        void *handle = dlopen("lib/libjemalloc.so", RTLD_NOW | RTLD_LOCAL);
        if (!handle) {
                fprintf(stderr,"Couldn't open\n");
                fprintf(stderr, "%s\n", dlerror());
        }
        return 0;
}
```

wthrowe:
OK, that's reasonable.  Defaulting to SYSTEM and telling people developing in that mode is fine sounds like a good solution for personal machines.

cmake/SpectreSetupPythonPackage.cmake
wthrowe: I don't think we need a deprecation period.  I think changing the default to SYSTEM is sufficient to avoid problems on personal machines.  For clusters we should make sure modules and everything are updated before merging this, but it sounds like that's being done.

knelli2:
Ok sounds fine to me. Pinging @nilsdeppe @geoffrey4444 @PunJustice @guilara as people who would likely be the ones to make this change on our various clusters. Anybody I'm forgetting?

nilsvu:
Turns out I found a problem with this approach and exploring a different option now. Stand by.

PR #6497, opened 2025-02-22
Set default memory allocator to system

nilsvu:
## Proposed changes

We list jemalloc as optional dependency, but currently it is the default and cmake will fail if it's not installed. This PR changes the default to `SYSTEM` and enables jemalloc explicitly on all clusters that support it.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (changes requested):
I don't think we should do this. I think the default should be jemalloc. I'm going printing a warning and falling back, but this is a performance regression.

wthrowe:
I also don't see a reason to change the default.  I supported the idea when it looked like we were going to require a custom-built jemalloc, but we're not planning on that anymore.

nilsvu:
Ok that's fine with me

PR #6500, opened 2025-02-25
Add `OrbitRadius` trigger

nikwit:
## Proposed changes

This trigger fires if the worldtube / scalar charge is about to cross any specified radii. It is used to manually adjust the time step during an inspiral.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.cpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.cpp
@@ -1,0 +1,47 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.hpp"
+
+#include <array>
+#include <vector>
+
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "Time/Time.hpp"
+#include "Utilities/Serialization/CharmPupable.hpp"
+
+namespace Triggers {
+
+OrbitRadius::OrbitRadius(const std::vector<double>& radii) : radii_(radii) {}
+
+bool OrbitRadius::operator()(
+    const std::array<tnsr::I<double, 3, Frame::Inertial>, 2>&
+        position_and_velocity,
+    const TimeDelta& time_step) const {
+  const auto& position = position_and_velocity[0];
+  const auto& velocity = position_and_velocity[1];
+  const double current_radius = get(magnitude(position));
+  const double radial_velocity = (get<0>(position) * get<0>(velocity) +
+                                  get<1>(position) * get<1>(velocity) +
+                                  get<2>(position) * get<2>(velocity)) /
+                                 current_radius;
+  // factor 1.2 is for safety because the approximation is just linear
+  // approximation and triggering it multiple times is not a problem
+  const double next_radius =
+      current_radius + 1.2 * radial_velocity * time_step.value();
+  // NOLINTNEXTLINE(readability-use-anyofallof)
+  for (const double radius : radii_) {
+    if ((current_radius - radius) * (next_radius - radius) < 0.) {
+      return true;
+    }
+  }
+  return false;
+}
+
+// NOLINTNEXTLINE(google-runtime-references)
+void OrbitRadius::pup(PUP::er& p) { p | radii_; }
+
+PUP::able::PUP_ID OrbitRadius::my_PUP_ID = 0;  // NOLINT
+}  // namespace Triggers
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.hpp
@@ -1,0 +1,66 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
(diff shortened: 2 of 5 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.hpp
```diff
@@ -0,0 +1,68 @@
...
+ public:
+  /// \cond
+  OrbitRadius() = default;
+  explicit OrbitRadius(CkMigrateMessage* /*unused*/) {}
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(OrbitRadius);  // NOLINT
+  /// \endcond
+
+  struct OptionTags {
+    struct Radii {
```
knelli2: Any reason why `Radii` needs to be within another struct? Otherwise you can just remove `OptionTags`.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.cpp
```diff
@@ -0,0 +1,47 @@
...
+  const auto& velocity = position_and_velocity[1];
+  const double current_radius = get(magnitude(position));
+  const double radial_velocity = (get<0>(position) * get<0>(velocity) +
+                                  get<1>(position) * get<1>(velocity) +
+                                  get<2>(position) * get<2>(velocity)) /
+                                 current_radius;
+  // factor 1.2 is for safety because the approximation is just linear
+  // approximation and triggering it multiple times is not a problem
+  const double next_radius =
+      current_radius + 1.2 * radial_velocity * time_step.value();
```
knelli2, line 33: Should 1.2 be an input option? Will it ever need to be changed?

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.cpp
```diff
@@ -0,0 +1,47 @@
...
+  // NOLINTNEXTLINE(readability-use-anyofallof)
+  for (const double radius : radii_) {
+    if ((current_radius - radius) * (next_radius - radius) < 0.) {
```
knelli2, line 36: [optional] I think this could be subject to roundoff error if the radii being subtracted are different enough in magnitude. Could do `(current_radius / radius - 1.0) * (next_radius / radius - 1.0) < 0.0` instead. But this trigger is already only approximate anyways so not sure how much it matters.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
+#include "Time/Time.hpp"
+#include "Utilities/Serialization/CharmPupable.hpp"
+
+namespace Triggers {
+
+OrbitRadius::OrbitRadius(const std::vector<double>& radii) : radii_(radii) {}
```
knelli2, line 17: Do you need to be worried about what happens if two radii are sufficiently close to each other that within a time step the charge will cross both? Should there be two triggers in that case, or is the one enough?

knelli2 (changes requested):
You can squash anything directly

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.cpp
nikwit: I do not think so. Eventually, I would really like to use an automated time step chooser but did not have any luck with them. I found that 1.2 will always trigger and rarely trigger twice which is completely fine for adjusting the time step.

This is not ideal but it is probably not worth finding a better solution for the scalar case at this point

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.cpp
nikwit: These will always be quite close so I do not think this will cause any issues.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.cpp
nikwit: One is completely fine, as I am just using it to adjust the time step. In a normal scenario there will easily be 1000 time steps between two radii.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.hpp
```diff
@@ -0,0 +1,66 @@
...
+  struct Radii {
+    using type = std::vector<double>;
+    static constexpr Options::String help =
+        "The orbital radii which should trigger when crossed by the "
+        "scalar charge.";
+  };
+
+  static constexpr Options::String help =
+      "Trigger that fires when the scalar charge crosses specified radii.";
+  using options = tmpl::list<typename OptionTags::Radii>;
```
knelli2: Oops, forgot to remove OptionsTags here as well

knelli2 (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/Triggers/OrbitRadius.hpp
nikwit: my bad, I fixed this. Thanks for the review!

knelli2 (approved):

PR #6501, opened 2025-02-26
Send inertial centers to control system and add control system for grid centers

nilsdeppe:
## Proposed changes

This adds the first bit of the grid centers control system needed for BNS. Still need to add the map in the domain and then use the control system in an evolution.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/GridCenters.cpp
+++ b/src/ControlSystem/ControlErrors/GridCenters.cpp
@@ -1,0 +1,38 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "ControlSystem/ControlErrors/GridCenters.hpp"
+
+#include <cstddef>
+
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+
+namespace control_system::ControlErrors {
+GridCenters::GridCenters(const Options::Context& /*context*/) {}
+
+DataVector GridCenters::impl(const DataVector& fot_positions_dv,
+                             const DataVector& measured_grid_position_of_A,
+                             const DataVector& measured_grid_position_of_B) {
+  tnsr::I<DataVector, 3, Frame::Grid> measured_grid_positions_tnsr{2_st, 0.0};
+  for (size_t i = 0; i < 3; i++) {
+    measured_grid_positions_tnsr.get(i)[0] = measured_grid_position_of_A[i];
+    measured_grid_positions_tnsr.get(i)[1] = measured_grid_position_of_B[i];
+  }
+
+  DataVector control_error{6, 0.0};
+  for (size_t i = 0; i < 2; i++) {
+    const auto& measured_grid_position =
+        i == 0 ? measured_grid_position_of_A : measured_grid_position_of_B;
+    for (size_t j = 0; j < 3; j++) {
+      control_error[i * 3 + j] =
+          measured_grid_position[j] - fot_positions_dv[i * 3 + j];
+    }
+  }
+
+  return control_error;
+}
+
+void GridCenters::pup(PUP::er& /*p*/) {}
+}  // namespace control_system::ControlErrors
--- a/src/ControlSystem/ControlErrors/GridCenters.hpp
+++ b/src/ControlSystem/ControlErrors/GridCenters.hpp
@@ -1,0 +1,104 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cstddef>
+#include <pup.h>
+
+#include "ControlSystem/Protocols/ControlError.hpp"
+#include "ControlSystem/Tags/QueueTags.hpp"
+#include "ControlSystem/Tags/SystemTags.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "Domain/Creators/Tags/ObjectCenter.hpp"
+#include "Domain/Structure/ObjectLabel.hpp"
+#include "Options/Context.hpp"
+#include "Options/ParseError.hpp"
+#include "Options/String.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "Utilities/ProtocolHelpers.hpp"
+#include "Utilities/TMPL.hpp"
+#include "Utilities/TaggedTuple.hpp"
+
+/// \cond
+namespace domain::Tags {
+struct FunctionsOfTime;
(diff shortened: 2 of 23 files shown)


src/ControlSystem/ControlErrors/GridCenters.hpp
```diff
@@ -0,0 +1,98 @@
...
+ * grid-frame centers is better than tracking the inertial-frame centers
+ * because the timescale at which the grid frame locations change is much
+ * larger than the inertial ones. This is because in the grid frame only the
+ * radial inspiral really changes because the rotation control system handles
+ * the orbital movement. The orbital time scale is much larger than the
+ * Cartesian coordinate location timescale, so splitting the system into
+ * angular and radial allows for both to be slowly varying.
+ *
+ * The intended use case of the output of this control system is for binary
+ * neutron star simulations where the rotation control needs to be disable at
```
knelli2: [nitpick] "disabled"

src/ControlSystem/ControlErrors/GridCenters.hpp
```diff
@@ -0,0 +1,98 @@
...
+ * gauge near when the neutron stars merge. Finally, it can also be used for
+ * triggering a grid change from the inspiral grid to the remnant grid in a
+ * BNS merger simulation.
+ *
+ * Requirements:
+ * - This control error requires that there be either one or two objects in the
+ *   simulation (typically neutron stars).
+ * - Currently this control error can only be used with the \link
+ *   control_system::Systems::GridCenters GridCenters \endlink control system
+ * - There must exist a `PiecewisePolynomial<2>` map named "GridCenters". It is
```
knelli2: "map" -> "function of time"

src/ControlSystem/ControlErrors/GridCenters.cpp
```diff
@@ -0,0 +1,44 @@
...
+  for (size_t i = 0; i < 3; i++) {
+    measured_grid_positions_tnsr.get(i)[0] = measured_grid_position_of_A[i];
+    measured_grid_positions_tnsr.get(i)[1] = measured_grid_position_of_B[i];
+  }
+
+  tnsr::I<DataVector, 3, Frame::Grid> fot_positions_tnsr{2_st, 0.0};
+  for (size_t i = 0; i < 3; i++) {
+    fot_positions_tnsr.get(i)[0] = fot_positions_dv[i];
+    fot_positions_tnsr.get(i)[1] = fot_positions_dv[3 + i];
+  }
```
knelli2: No need to convert anything to tensors. Just use the `DataVector`s directly

src/ControlSystem/Systems/GridCenters.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+/// \cond
+namespace Frame {
+struct Distorted;
+}  // namespace Frame
+/// \endcond
+
+namespace control_system::Systems {
+/*!
+ * \brief Needs some dox
```
knelli2: Indeed it does

src/ControlSystem/Systems/GridCenters.hpp
```diff
@@ -0,0 +1,133 @@
...
+
+      if (Parallel::get<Tags::Verbosity>(cache) >= ::Verbosity::Verbose) {
+        Parallel::printf("%s, time = %.16f: Received measurement '%s'.\n",
+                         name(), measurement_id.id,
+                         pretty_type::name(submeasurement));
+      }
+    }
+  };
+};
+}  // namespace control_system::Systems
```
knelli2, line 141: [random line] Please also add tests of the control error and system. They can both be simple tests like `Test_Skew.cpp` in their respective directories.

knelli2 (changes requested):
I know I was the one who wrote the first commit, but now it seems like we don't need to send the inertial centers to the control system anymore. Should we still do that in case we need them in the future, or just leave things as is until a need arises?

nilsdeppe:
> I know I was the one who wrote the first commit, but now it seems like we don't need to send the inertial centers to the control system anymore. Should we still do that in case we need them in the future, or just leave things as is until a need arises?

I did think about this, and I would like to keep it 1. because we might need it in the future and 2. (maybe the main reason) then it's easy to plot the BNS trajectories without having to get the right observers coded up.

knelli2:
Ok we can just keep it in then. But for your second point, we already had computed the inertial centers for output so that portion wouldn't have changed.

src/ControlSystem/UpdateControlSystem.hpp
```diff
@@ -109,6 +109,12 @@ struct UpdateControlSystem {
 
     // Begin step 1
     // If this control system isn't active, don't do anything
+    ASSERT(Parallel::get<control_system::Tags::IsActiveMap>(cache).find(
+               function_of_time_name) !=
+               Parallel::get<control_system::Tags::IsActiveMap>(cache).end(),
```
knelli2: [optional] Just use `.contains(function_of_time_name)` ?

tests/Unit/ControlSystem/Systems/Test_GridCenters.cpp
```diff
@@ -0,0 +1,178 @@
...
+       control_system::Tags::MeasurementTimescales::type{}}};
+
+  ActionTesting::emplace_singleton_component<component>(make_not_null(&runner),
+                                                        {0}, {0});
+
+  auto& cache = ActionTesting::cache<component>(runner, 0);
+
+  // The data here doesn't matter except for the excision surface because it's
+  // the only one that is used in any calculations. And even then, the only
+  // restriction is that is must be contained within our domain above.
```
knelli2: Comment was probably left over from another test

src/ControlSystem/Systems/GridCenters.hpp
```diff
@@ -0,0 +1,142 @@
...
+namespace control_system::Systems {
+/*!
+ * \brief Tracks the centers of two neutron stars.
+ *
+ * Controls the function of time with the same name which can be used to
+ * determine (until merger) the centers of the two stars.
+ *
+ * Requirements:
+ * - This control system requires that there be either one or two objects in the
+ *   simulation
```
knelli2: I think this requires exactly two objects

knelli2 (changes requested):
You can squash directly

nilsdeppe:
Done :) Thanks for the review!

knelli2 (approved):

PR #6502, opened 2025-02-27
Add ability to preprocess non-GH evolution data

knelli2:
## Proposed changes

This allows codes that evolve BSSN or a variant of Z4 to dump the extrinsic curvature and auxiliary shift vector rather than the time derivative of the metric, lapse, and shift. We then compute these time derivatives instead, assuming 1+log slicing and the gamma-driver condition for the shift. If different lapse and shift conditions will be useful to the community, they can be added later but this should cover a large number of codes already.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/docs/Tutorials/CCE.md
+++ b/docs/Tutorials/CCE.md
@@ -24,7 +24,7 @@ release). Inside this tarball is
 - an example set of Bondi-Sachs worldtube data in the `Tests/` directory (see
    [Input worldtube data formats](#input_worldtube_data_format) section)
 - example output from CCE in the `Tests/` directory
-- a `PreprocessCceWorldtube` executable and YAML file for converting between
+- a `PreprocessCceWorldtube` executable and YAML files for converting between
    [worldtube data formats](#input_worldtube_data_format) in the
    `PreprocessCceWorldtube/` directory
 - a `WriteCceWorldtubeCoordsToFile` executable that writes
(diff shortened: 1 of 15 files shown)


docs/Tutorials/CCE.md
```diff
@@ -233,6 +233,52 @@ named `Legend` which is an ASCII-encoded null-terminated variable-length string.
...
+Gamma-driver condition from Eq. (4.89) of \cite BaumgarteShapiro or using the
+integrated Gamma-driver condition from Eq. (12) of \cite Hilditch:2012fp.
+
+\warning If your worldtube data is in the ADM metric nodal format but you have
+not used `1+log` slicing and the Gamma-driver conditions specified above, your
+time derivatives will be **wrong**. If you'd like us to support other commonly
+used gauge conditions (or variants of `1+log` or Gamma-driver), please open an
+issue on our
+[GitHub](https://github.com/sxs-collaboration/spectre) or email the core
+developers at [core-devs@spectre-code.org](mailto:core-devs@spectre-code.org).
```
nilsdeppe: I'd restrict this to just filing an issue :)

src/Evolution/Systems/Cce/WorldtubeBufferUpdater.cpp
```diff
@@ -420,13 +480,281 @@ double MetricWorldtubeH5BufferUpdater<T>::update_buffers_for_time(
...
+      get<::Tags::Tempijj<0, 3>>(temp_buffer);
+  auto& extrinsic_curvature = get<::Tags::Tempii<0, 3>>(temp_buffer);
+  auto& trace_extrinsic_curvature = get<::Tags::TempScalar<0>>(temp_buffer);
+  auto& determinant_spatial_metric = get<::Tags::TempScalar<1>>(temp_buffer);
+  auto& inverse_spatial_metric = get<::Tags::TempII<0, 3>>(temp_buffer);
+  auto& auxiliary_shift = get<::Tags::TempI<0, 3>>(temp_buffer);
+  auto& conformal_christoffel = get<::Tags::TempI<1, 3>>(temp_buffer);
+
+  const auto set_radial_deriv =
+      [&](const gsl::not_null<DataVector*> radial_deriv,
```
nilsdeppe: `&cartesian_cauchy_coordinates_buffer`. Capturing everything is generally strongly recommended against since it can trivially lead to dangling references and makes the closure generated a lot more complicated

nilsdeppe (changes requested):
you can squash immediately

src/Evolution/Systems/Cce/WorldtubeBufferUpdater.cpp
knelli2: I made this change, but I think when a default capture is used, only the objects that are actually used are captured. At least according to my reading of the [docs](https://en.cppreference.com/w/cpp/language/lambda#Lambda_capture)

nilsdeppe (approved):

PR #6503, opened 2025-02-27
Begin adding support for domains with non-conforming blocks

kidder:
## Proposed changes

By non-conforming Blocks I mean that the block logical coordinates of the two Blocks on an interface between them are not related by a discrete rotation (represented by an OrientationMap which could be the identity map, i.e. no rotation).  There are two main consequences that break existing code:
- A Block may have multiple neighboring Blocks in a given Direction
- Boundary data communicated to a non-conforming neighbor can not be copied or projected onto the mortar, but instead must be interpolated.   

Specific changes in this PR:
- Add template parameter IdType to Neighbors in order to use it for neighbors of either a Block or Element
- replace BlockNeighbor with Neighbors in order to support multiple block neighbors in a given direction
- introduce InterfaceDataPolicy which will be used to determine the communication pattern for applying boundary corrections
- store the InterfaceDataPolicy in Neighbors along with the (now optional) OrientationMap 

### Upgrade instructions

- replace `BlockNeighbor<Dim>` with `Neighbors<Dim, size_t>`
- when constructing block neighbors, it now needs a `std::unordered_set<size_t>` instead of a `size_t` for the Id
- if the neighboring Blocks/Elements are conforming and aligned, you need only construct the Neighbors with the set of Ids
- if the neighboring Blocks/Elements are conforming, but not aligned, you need to construct the Neighbors with the set of Ids, `domain::InterfaceDataPolicy::OrientCopyProject`, and the OrientationMap
- if you call `Neighbors::orientation()` it now returns a `std::optional` so you instead to call `Neighbors::orientation().value_or(OrientationMap<VolumeDim>::create_aligned())` to get the previous behavior.  (In the future, the code calling this will need to be modified in order to support domains with non-conforming Blocks.)

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Amr/NeighborsOfParent.hpp
+++ b/src/Domain/Amr/NeighborsOfParent.hpp
@@ -24,7 +24,7 @@ template <size_t VolumeDim>
 class ElementId;
 template <size_t VolumeDim>
 class Mesh;
-template <size_t VolumeDim>
+template <size_t VolumeDim, typename IdType>
 class Neighbors;
 /// \endcond
 
@@ -33,7 +33,7 @@ namespace amr {
 /// \brief returns the neighbors and their Mesh%es of the Element with ElementId
 /// `parent_id`, that is created from its `children_elements_and_neighbor_info`
 template <size_t VolumeDim>
-std::pair<DirectionMap<VolumeDim, Neighbors<VolumeDim>>,
+std::pair<DirectionMap<VolumeDim, Neighbors<VolumeDim, ElementId<VolumeDim>>>,
           DirectionalIdMap<VolumeDim, Mesh<VolumeDim>>>
 neighbors_of_parent(
     const ElementId<VolumeDim>& parent_id,
--- a/src/Domain/Block.cpp
+++ b/src/Domain/Block.cpp
@@ -11,7 +11,7 @@
 #include <typeinfo>
 #include <utility>
 
-#include "Domain/Structure/BlockNeighbor.hpp"
+#include "Domain/Structure/BlockNeighbors.hpp"
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/DirectionMap.hpp"
 #include "NumericalAlgorithms/Spectral/Basis.hpp"
@@ -24,7 +24,7 @@ Block<VolumeDim>::Block(
     std::unique_ptr<domain::CoordinateMapBase<
         Frame::BlockLogical, Frame::Inertial, VolumeDim>>&& stationary_map,
     const size_t id,
-    DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>> neighbors,
+    DirectionMap<VolumeDim, BlockNeighbors<VolumeDim>> neighbors,
     std::string name, std::array<domain::Topology, VolumeDim> topologies)
     : stationary_map_(std::move(stationary_map)),
       id_(id),
--- a/src/Domain/CreateInitialElement.cpp
+++ b/src/Domain/CreateInitialElement.cpp
@@ -147,12 +147,11 @@ Element<VolumeDim> create_initial_element(
         SegmentId(perpendicular_segment_id.refinement_level(),
                   direction.side() == Side::Upper ? index + 1 : index - 1);
     return std::make_pair(
-        direction,
-        Neighbors<VolumeDim>(
-            {{ElementId<VolumeDim>{element_id.block_id(),
-                                   std::move(segment_ids_of_neighbor),
-                                   element_id.grid_index()}}},
-            OrientationMap<VolumeDim>::create_aligned()));
+        direction, Neighbors<VolumeDim>(
+                       {ElementId<VolumeDim>{element_id.block_id(),
+                                             std::move(segment_ids_of_neighbor),
+                                             element_id.grid_index()}},
+                       OrientationMap<VolumeDim>::create_aligned()));
   };
 
   typename Element<VolumeDim>::Neighbors_t neighbors_of_element;
--- a/src/Domain/CreateInitialElement.cpp
+++ b/src/Domain/CreateInitialElement.cpp
@@ -42,8 +42,10 @@ Element<VolumeDim> create_initial_element(
 
         // SegmentIds of the current element using the neighbor's axes.
         auto segment_ids_of_unrefined_neighbor = orientation(segment_ids);
+        ASSERT(block_neighbor.size() == 1,
+               "Multiple block neighbors not supported by this function.");
         const auto& refinement_of_neighbor =
-            initial_refinement_levels[block_neighbor.id()];
+            initial_refinement_levels[*block_neighbor.begin()];
 
         // Check whether each dimension will have multiple neighbors
         // because the neighboring block is more refined.  For dimensions
@@ -73,7 +75,7 @@ Element<VolumeDim> create_initial_element(
                     << gsl::at(refinement_of_neighbor, d) << " and "
                     << gsl::at(segment_ids_of_unrefined_neighbor, d)
                            .refinement_level()
-                    << " in blocks " << block_neighbor.id() << " and "
+                    << " in blocks " << *block_neighbor.begin() << " and "
                     << block.id() << " differ by more than one.");
           }
         }
@@ -126,7 +128,7 @@ Element<VolumeDim> create_initial_element(
             }
           }
           neighbor_ids.insert(
-              {block_neighbor.id(), segment_ids_of_neighbor, grid_index});
+              {*block_neighbor.begin(), segment_ids_of_neighbor, grid_index});
         next_index:;
         }
         return std::make_pair(
--- a/src/Domain/DomainHelpers.cpp
+++ b/src/Domain/DomainHelpers.cpp
@@ -380,7 +380,10 @@ void set_cartesian_periodic_boundaries(
               opposite_block.end()) {
             continue;
           } else {
-            opposite_block_id = opposite_block[opposite_block_direction].id();
+            ASSERT(opposite_block[opposite_block_direction].size() == 1,
+                   "Multiple block neighbors not supported by this function.");
+            opposite_block_id =
+                *(opposite_block[opposite_block_direction].begin());
             opposite_block = (*neighbors_of_all_blocks)[opposite_block_id];
             opposite_block_direction =
                 orientations_of_all_blocks[opposite_block_id](direction)
--- a/src/Domain/DomainHelpers.cpp
+++ b/src/Domain/DomainHelpers.cpp
@@ -28,7 +28,7 @@
 #include "Domain/CoordinateMaps/ProductMaps.tpp"
 #include "Domain/CoordinateMaps/Wedge.hpp"
 #include "Domain/Domain.hpp"
-#include "Domain/Structure/BlockNeighbor.hpp"
+#include "Domain/Structure/BlockNeighbors.hpp"
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/DirectionMap.hpp"
 #include "Domain/Structure/OrientationMap.hpp"
@@ -347,7 +347,7 @@ void set_cartesian_periodic_boundaries(
         corners_of_all_blocks,
     const std::vector<OrientationMap<VolumeDim>>& orientations_of_all_blocks,
     const gsl::not_null<
-        std::vector<DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>>>*>
+        std::vector<DirectionMap<VolumeDim, BlockNeighbors<VolumeDim>>>*>
         neighbors_of_all_blocks) {
   if (orientations_of_all_blocks.size() != corners_of_all_blocks.size()) {
     ERROR("Each block must have an OrientationMap relative to an edifice.");
@@ -445,13 +445,13 @@ void set_cartesian_periodic_boundaries(
 template <size_t VolumeDim>
 void set_internal_boundaries(
     const gsl::not_null<
-        std::vector<DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>>>*>
+        std::vector<DirectionMap<VolumeDim, BlockNeighbors<VolumeDim>>>*>
         neighbors_of_all_blocks,
     const std::vector<std::array<size_t, two_to_the(VolumeDim)>>&
         corners_of_all_blocks) {
   for (size_t block1_index = 0; block1_index < corners_of_all_blocks.size();
        block1_index++) {
-    DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>> neighbor;
+    DirectionMap<VolumeDim, BlockNeighbors<VolumeDim>> neighbor;
     for (size_t block2_index = 0; block2_index < corners_of_all_blocks.size();
          block2_index++) {
       if (block1_index != block2_index and
@@ -463,7 +463,7 @@ void set_internal_boundaries(
             obtain_correspondence_between_blocks<VolumeDim>(
                 block1_index, block2_index, corners_of_all_blocks);
         neighbor.emplace(std::move((orientation_helper.first)[0]),
-                         BlockNeighbor<VolumeDim>(
+                         BlockNeighbors<VolumeDim>(
                              block2_index, OrientationMap<VolumeDim>(
                                                orientation_helper.first,
                                                orientation_helper.second)));
@@ -476,12 +476,12 @@ void set_internal_boundaries(
 template <size_t VolumeDim>
 void set_internal_boundaries(
     const gsl::not_null<
-        std::vector<DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>>>*>
+        std::vector<DirectionMap<VolumeDim, BlockNeighbors<VolumeDim>>>*>
         neighbors_of_all_blocks,
     const std::vector<std::unique_ptr<domain::CoordinateMapBase<
         Frame::BlockLogical, Frame::Inertial, VolumeDim>>>& maps) {
   for (size_t block1_index = 0; block1_index < maps.size(); block1_index++) {
-    DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>> neighbor;
+    DirectionMap<VolumeDim, BlockNeighbors<VolumeDim>> neighbor;
     for (size_t block2_index = 0; block2_index < maps.size(); block2_index++) {
       const auto corners_for_blocks1_and_2 =
           corners_from_two_maps(maps[block1_index], maps[block2_index]);
@@ -494,7 +494,7 @@ void set_internal_boundaries(
                 0, 1, corners_for_blocks1_and_2);
 
         neighbor.emplace(std::move((orientation_helper.first)[0]),
-                         BlockNeighbor<VolumeDim>(
+                         BlockNeighbors<VolumeDim>(
                              block2_index, OrientationMap<VolumeDim>(
                                                orientation_helper.first,
                                                orientation_helper.second)));
@@ -510,7 +510,7 @@ void set_identified_boundaries(
     const std::vector<std::array<size_t, two_to_the(VolumeDim)>>&
         corners_of_all_blocks,
     const gsl::not_null<
-        std::vector<DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>>>*>
+        std::vector<DirectionMap<VolumeDim, BlockNeighbors<VolumeDim>>>*>
         neighbors_of_all_blocks) {
   for (const auto& pair : identifications) {
     const auto& face1 = pair.first;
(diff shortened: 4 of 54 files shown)


src/Domain/Amr/NeighborsOfParent.hpp
```diff
@@ -33,7 +33,7 @@ namespace amr {
 /// \brief returns the neighbors and their Mesh%es of the Element with ElementId
 /// `parent_id`, that is created from its `children_elements_and_neighbor_info`
 template <size_t VolumeDim>
-std::pair<DirectionMap<VolumeDim, Neighbors<VolumeDim>>,
+std::pair<DirectionMap<VolumeDim, Neighbors<VolumeDim, ElementId<VolumeDim>>>,
```
nilsvu, line 36: Suggestion: define `using ElementNeighbors<Dim> = Neighbors<Dim, ElementId<Dim>>` in `Neighbors.hpp`, and some for `BlockNeighbors`? Then we don't have to repeat this type everywhere.

src/Domain/Block.cpp
```diff
@@ -23,7 +23,7 @@ Block<VolumeDim>::Block(
     std::unique_ptr<domain::CoordinateMapBase<
         Frame::BlockLogical, Frame::Inertial, VolumeDim>>&& stationary_map,
     const size_t id,
-    DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>> neighbors,
+    DirectionMap<VolumeDim, Neighbors<VolumeDim, size_t>> neighbors,
```
nilsvu: Suggestion to define `using BlockNeighbors<Dim> = Neighbors<VolumeDim, size_t>`

src/Domain/Structure/Neighbors.hpp
```diff
@@ -21,20 +19,25 @@ class ElementId;
...
+/// direction.
 ///
 /// \tparam VolumeDim the volume dimension.
-/// \tparam IdType the type of the Id of the neighbor
+/// \tparam IdType the type of the Id of the neighbor (either ElementId or
+/// size_t for a Block)
 template <size_t VolumeDim, typename IdType = ElementId<VolumeDim>>
 class Neighbors {
+  static_assert(std::is_same_v<IdType, size_t> or
+                std::is_same_v<IdType, IdType>);
```
nilsvu: one `IdType` -> `ElementId`?

src/Domain/Amr/NewNeighborIds.cpp
```diff
@@ -49,7 +49,8 @@ std::unordered_map<ElementId<VolumeDim>, Mesh<VolumeDim>> new_neighbor_ids(
       new_neighbors_in_direction;
 
   const OrientationMap<VolumeDim>& orientation_map_from_me_to_neighbors =
-      previous_neighbors_in_direction.orientation();
+      previous_neighbors_in_direction.orientation().value_or(
+          OrientationMap<VolumeDim>::create_aligned());
```
nilsvu: This is used all over the code, can you keep it as a non-optional that is just `aligned`? That's what the missing value means anyway, so this change just makes the calling code more complicated.

src/Domain/Amr/UpdateAmrDecision.cpp
```diff
@@ -40,7 +40,9 @@ bool update_amr_decision(
       const Direction<VolumeDim> direction_to_neighbor =
           direction_neighbors.first;
       const OrientationMap<VolumeDim>& orientation_of_neighbor =
-          neighbors_in_dir.orientation();
+          neighbors_in_dir.orientation().value_or(
+              OrientationMap<VolumeDim>::create_aligned());
+      ;
```
nilsvu: extra semicolon

src/Domain/DomainHelpers.cpp
```diff
@@ -460,11 +460,18 @@ void set_internal_boundaries(
...
-                                                 orientation_helper.second)));
+        if (orientation_helper.first == orientation_helper.second) {
+          neighbor.emplace(std::move((orientation_helper.first)[0]),
+                           Neighbors<VolumeDim, size_t>({block2_index}));
+        } else {
+          neighbor.emplace(
+              std::move((orientation_helper.first)[0]),
+              Neighbors<VolumeDim, size_t>(
+                  {block2_index},
+                  domain::InterfaceDataPolicy::OrientCopyProject,
```
nilsvu: It seems that `OrientCopyProject` is just `CopyProject and not orientation.is_aligned()`. That's how we already handle these interfaces (checking if the orientation is aligned before orientating), so I'd remove `OrientCopyProject` and just keep doing the conditional orientation. The extra enum case just has the potential to be inconsistent with the value of `orientation.is_aligned()`.

src/Domain/Structure/Neighbors.hpp
```diff
@@ -90,8 +104,12 @@ class Neighbors {
   }
 
  private:
+  void maximum_neighbor_check();
+
   std::unordered_set<IdType> ids_;
-  OrientationMap<VolumeDim> orientation_;
+  domain::InterfaceDataPolicy interface_data_policy_{
+      domain::InterfaceDataPolicy::Uninitialized};
+  std::optional<OrientationMap<VolumeDim>> orientation_{std::nullopt};
```
nilsvu: Just make this non-optional and default-construct to aligned (or not default-construct at all, like currently). Has the same meaning and is easier to work with. I understand that you don't store the orientation map at all if you want to do interpolation between the faces, but I wonder if we want to do 1d interpolation at some point, and then we need the orientation anyway. I'd just keep the orientation here instead of making it an optional.

nilsvu (changes requested):

src/Domain/Structure/Neighbors.hpp
kidder: oops

src/Domain/Amr/NewNeighborIds.cpp
kidder: I did this on purpose to identify places that will need to depend upon InterfaceDataPolicy instead of the OrientationMap in order to support non-conforming blocks.  An OrientationMap has no meaning for the non-conforming cases as the logical coordinate axes on the interface are not simply related.

src/Domain/Amr/UpdateAmrDecision.cpp
kidder: oops

src/Domain/DomainHelpers.cpp
kidder: The OrientationMap will only be used for OrientCopyProject.  Instead of checking orientation.is_aligned(), code will need to execute a switch statement based on the InterfaceDataPolicy.

src/Domain/Structure/Neighbors.hpp
kidder: I'm not sure what you mean by this.  Interpolation is done on the boundary (so a 2d interpolation for a 3d domain).  There is no relation between the logical axes tangent on each side of the interface that can be represented by an OrientationMap

src/Domain/Amr/NewNeighborIds.cpp
nilsvu: Ok how about this then:

```cpp
const OrientationMap<Dim>& orientation() {
  ASSERT(interface_data_policy == CopyProject, "Orientation only has meaning if the interface data policy is CopyProject");
  return orientation_.value_or(OrientationMap<Dim>::create_aligned());
}
```

src/Domain/DomainHelpers.cpp
nilsvu: Ok, can we do this (pseudocode):

```cpp
switch (neighbors.interface_data_policy) {
  case CopyProject:
    if (neigbors.orientation().is_aligned()) {
      orient();
    }
    project();
  // don't need separate OrientCopyProject case
  case Nonconforming:  // etc
    // don't need orientation
    interpolate();
```

src/Domain/Structure/Neighbors.hpp
nilsvu: Makes sense, then the assert when calling `orientation()` would be the easiest to work with I think (see comment above). To avoid the `value_or` you could default-construct an aligned map here.

src/Domain/Structure/Neighbors.hpp
```diff
@@ -34,29 +35,29 @@ class Neighbors {
   /// \param orientation This OrientationMap takes objects in the logical
   /// coordinate frame of the host Element and maps them to the logical
   /// coordinate frame of the neighbor Element.
-  Neighbors(std::unordered_set<ElementId<VolumeDim>> ids,
+  Neighbors(std::unordered_set<IdType> ids,
```
nilsvu, line 41: Suggestion: add a constructor with a single `IdType` value? Then you don't have to add curly braces in so many places.

src/Domain/Amr/NeighborsOfParent.hpp
nilsdeppe: If we want to be able to forward declare those we need to use a strong typedef instead of a using. I have no objections to the suggestion, just expanding on it.

src/Domain/Structure/Neighbors.cpp
```diff
@@ -1,3 +1,4 @@
+
```
nilsdeppe: RIP

src/Domain/Structure/Neighbors.hpp
nilsdeppe: I can't think of a reason where that would subtly give unexpected results. As long as we document it, I think it'll work well :)

src/Domain/Structure/InterfaceDataPolicy.hpp
```diff
@@ -0,0 +1,72 @@
...
+/// logical coordinates.  Therefore a DataVector must be interpolated to the
+/// points of the neighboring Mesh.  The cases differ in which Element does the
+/// interpolation:
+/// - NonconformingBothInterpolate:  in this case both the Element and its
+///   neighbor interpolate data to the grid points of each others Mesh.
+/// - NonconformingSelfInterpolates:  in this case the Element will receive the
+///   neighbor's boundary data and will interpolate it to the Mesh of the
+///   Element.  The Element will then need to send boundary correction data to
+///   the neighbor.
+/// - NonconformingNeighborInterpolates:  in this case the Element send its
```
nilsdeppe: `send` -> `sends`?

src/Domain/Structure/InterfaceDataPolicy.hpp
```diff
@@ -0,0 +1,72 @@
...
+///   Element.
+/// The Element and its neighbor will need to use consistent values of this
+/// enum:
+/// - In the cases CopyProject, OrientCopyProject, and
+///   NonconformingBothInterpolate, neighboring elements should agree on the
+///   values.
+/// - For NonconformingSelfInterpolates and NonconformingNeighborInterpolates
+///   neighboring elements should have different values.  These cases should be
+///   used when one Element has many neighboring Elements (e.g. when a single
+///   spherical shell abuts a cubes sphere).  In this case it should be more
```
nilsdeppe: `cubes sphere` -> `cubed sphere`

src/Domain/Structure/InterfaceDataPolicy.hpp
```diff
@@ -0,0 +1,72 @@
...
+///   boundary data to the neighbor who will then interpolate it to its own
+///   Mesh. The neighbor will need to send boundary correction data back to the
+///   Element.
+/// The Element and its neighbor will need to use consistent values of this
+/// enum:
+/// - In the cases CopyProject, OrientCopyProject, and
+///   NonconformingBothInterpolate, neighboring elements should agree on the
+///   values.
+/// - For NonconformingSelfInterpolates and NonconformingNeighborInterpolates
+///   neighboring elements should have different values.  These cases should be
```
nilsdeppe: maybe `different` -> `opposite`?

src/Domain/Structure/InterfaceDataPolicy.cpp
```diff
@@ -0,0 +1,33 @@
+// Distributed under the MIT License.
```
nilsdeppe: Don't you need to add this to a CMakeLists.txt??

src/Domain/DomainHelpers.cpp
nilsdeppe: I don't see the benefit of the suggestion. I find it clearer with the enum describing what to do. I find the pseudo code:
```cpp
switch (neighbors.interface_data_policy) {
  case CopyProject:
    project();
  case OrientCopyProject:
    project(orient());
  case Nonconforming: 
    interpolate();
```
a lot clearer and self-documenting.

src/Domain/Amr/NewNeighborIds.cpp
nilsdeppe: I don't have a strong preference one way or the other. I agree that having to use `value_or` in a bunch of places is less readable. I don't see how changing to the `ASSERT`-based approach would make using the code more error prone than people misusing `value_or` to begin with in this case. I expect the suggestion will actually be safer than the current code because the class itself verifies consistency.

nilsdeppe (commented):
I like the overall design. I think Nils has some great suggestions. I've given my thoughts on this in general and some minor requests

src/Domain/Amr/NeighborsOfParent.hpp
kidder: - made and used a strong typedef for BlockNeighbors
-  did not make one for ElementNeighbors as that would require a global replacement.  If you want me to do so, I will, but in a a separate PR

src/Domain/Block.cpp
kidder: used strong typedef instead

src/Domain/Amr/NewNeighborIds.cpp
kidder: okay, implemented the ASSERT based approach

src/Domain/Structure/Neighbors.hpp
kidder: done

kidder:
- @nilsdeppe @nilsvu I implemented (most of) your suggestions.  
- In addition I kept the old constructors for Neighbors and set the InterfaceDataPolicy appropriately so as to touch less code in existing tests.  
- The first and fifth commits are essentially the same.  The others touched less existing code and are shorter (and I think less confusing) than fixup commits would have been.

src/Domain/Structure/Neighbors.hpp
```diff
@@ -38,6 +38,15 @@ class Neighbors {
...
             OrientationMap<VolumeDim> orientation);
 
+  /// Construct with the id and orientation of a single neighbor relative to the
+  /// host.
+  ///
+  /// \param id the id of the neighbors.
+  /// \param orientation This OrientationMap takes objects in the logical
+  /// coordinate frame of the host Element and maps them to the logical
+  /// coordinate frame of the neighbor Element.
+  Neighbors(const IdType& id, OrientationMap<VolumeDim> orientation);
```
nilsvu: As usual take `id` by value and move into new set?

src/Domain/CreateInitialElement.cpp
```diff
@@ -43,7 +43,7 @@ Element<VolumeDim> create_initial_element(
         // SegmentIds of the current element using the neighbor's axes.
         auto segment_ids_of_unrefined_neighbor = orientation(segment_ids);
         const auto& refinement_of_neighbor =
-            initial_refinement_levels[block_neighbor.id()];
+            initial_refinement_levels[*block_neighbor.begin()];
```
nilsvu, line 48: Assert that `block_neighbor.size() == 1` here and in other places you assume this?

src/Domain/Structure/Neighbors.cpp
```diff
@@ -67,16 +68,37 @@ bool operator!=(const Neighbors<VolumeDim, IdType>& lhs,
 
 template <size_t VolumeDim, typename IdType>
 void Neighbors<VolumeDim, IdType>::pup(PUP::er& p) {
-  p | ids_;
-  p | orientation_;
+  if constexpr (std::is_same_v<IdType, size_t>) {
+    size_t version = 1;
+    p | version;
```
nilsvu, line 69: Add a comment that version 0 is deserializing the old `BlockNeighbor` class. That's what's happening here, right?

src/Domain/Structure/InterfaceDataPolicy.hpp
```diff
@@ -0,0 +1,72 @@
...
+
+namespace domain {
+/// \brief Label for a neighboring Element (or Block) that determines how
+/// information is exchanged between neighboring Elements
+///
+/// \details The specific label is determined by the relationship between the
+/// block logical coordinates of the neighboring Elements (Blocks) for the
+/// points on the interface between them.  In two cases there is a simple
+/// relationship between the coordinates:
+/// - CopyProject: in this case the block logical coordinates are identical.
```
nilsvu: This is a bit misleading, please clarify that the block logical coordinate _frame_ is the same, or something like that, but the two sides can still have different collocation points (different meshes).

src/Domain/Structure/Neighbors.cpp
```diff
@@ -26,6 +26,14 @@ Neighbors<VolumeDim, IdType>::Neighbors(std::unordered_set<IdType> ids,
          "Cannot use a default-constructed OrientationMap in Neighbors.");
 }
 
+template <size_t VolumeDim, typename IdType>
+Neighbors<VolumeDim, IdType>::Neighbors(const IdType& id,
+                                        OrientationMap<VolumeDim> orientation)
+    : ids_(std::unordered_set{id}), orientation_(std::move(orientation)) {
```
nilsvu: I think you can just delegate to the unordered_set constructor here, no need to reimplement it

src/Domain/Structure/Neighbors.hpp
```diff
@@ -98,9 +118,19 @@ class Neighbors {
...
-  std::unordered_set<IdType> ids_;
-  OrientationMap<VolumeDim> orientation_;
+  void maximum_neighbor_check();
+  template <size_t LocalVolumeDim, typename LocalIdType>
+  // NOLINTNEXTLINE(readability-redundant-declaration)
+  friend bool operator==(const Neighbors<LocalVolumeDim, LocalIdType>& lhs,
+                         const Neighbors<LocalVolumeDim, LocalIdType>& rhs);
+
+  std::unordered_set<IdType> ids_{};
+  domain::InterfaceDataPolicy interface_data_policy_{
```
nilsvu: I'm a bit hesitant about storing the interface data policy here because it couples details of our DG scheme into the domain structure. The domain structure doesn't need to know about things like projecting and interpolating. What do you think about storing this interface data policy in an option tag instead, which goes into the discretization group? We probably want to do the same policy on all cube/sphere boundaries anyway, so there's no need to store it separately at each interface. Then I think on the domain side of things we just need to store a `domain::BlockGeometry` enum with cases `Cube` and `SphericalShell` to have all the information we need.

nilsvu (changes requested):

src/Domain/Structure/Neighbors.hpp
kidder: - It has nothing to do with DG specifically.  It just enables a switch based on how the coordinates on the two sides of the interface are related.
- The information needs to be in the Element just like the orientation as that is what is available at the decision points.
- No there is not a global value for the policy.  It is boundary dependent.  Especially as we are adding cylinders as well.
- There is no need for a BlockGeometry enum as the Basis and Quadrature already encode the information.

nilsvu: Just to quickly clarify, I mean there's a global value for how cube/sphere interfaces are handled, and another global value for how cylinder interfaces are handled. We don't want to make this decision differently at each interface of the same type.

kidder:
@nilsvu I pushed fixups, please look at

src/Domain/Structure/Neighbors.cpp
nilsvu: ping on this

src/Domain/Block.cpp
```diff
@@ -23,7 +23,7 @@ Block<VolumeDim>::Block(
     std::unique_ptr<domain::CoordinateMapBase<
         Frame::BlockLogical, Frame::Inertial, VolumeDim>>&& stationary_map,
     const size_t id,
-    DirectionMap<VolumeDim, BlockNeighbor<VolumeDim>> neighbors,
+    DirectionMap<VolumeDim, BlockNeighbors<VolumeDim>> neighbors,
```
nilsvu, line 27: When you rebase you can assert here that the number of neighbors in each dimension is consistent with the topology of the block

nilsvu (dismissed):

src/Domain/Block.cpp
kidder: this is done in the next PR which also correctly sets the number of external boundaries based on the topology and input neighbors

nilsvu (dismissed):

nilsvu:
gcc9 has a complaint (maybe a missing include?)

nilsvu (approved):

PR #6505, opened 2025-02-28
Control ADM energy and angular momentum in BBH ID

iago-mendes:
## Proposed changes

Add support for controlling $E_{ADM}$ and $J^{ADM}_z$ in BBH initial data. This is useful in generating initial data for hyperbolic encounters.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


support/Pipelines/Bbh/ControlId.py
```diff
@@ -278,6 +291,21 @@ def Residual(u):
     # Initialize Jacobian as an identity matrix
     J = np.identity(len(u))
 
+    # Adjust non-unitity components of the Jacobian
```
nilsvu: typo "non-unity"

support/Pipelines/Bbh/ControlId.py
```diff
@@ -278,6 +291,21 @@ def Residual(u):
...
+    q = target_params["MassRatio"]
+    param_index = 0
+    for param in control_params:
+        if param == "TotalEnergy":
+            J[param_index, param_index] = (
```
nilsvu, line 306: Can you put short comments here how to get these expressions for the Jacobian?

support/Pipelines/Bbh/ControlId.py
```diff
@@ -278,6 +291,21 @@ def Residual(u):
...
+                q
+                / (q + 1) ** 2
+                * initial_free_data["radial_expansion_velocity"]
+                * separation**2
+            )
+        elif param == "AngularMomentumZ":
+            J[param_index, param_index] = q / (q + 1) ** 2 * separation**2
```
nilsvu: Do you think it would help to set cross terms here as well to make the control less reliant on the initial guess? You could look at the final Jacobian to see if the cross terms have been activated by Broyden's method, and if their value is somewhat close to the Newtonian approximation.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -31,11 +31,22 @@
     "DimensionlessSpinB": "conformal_spin_b",
     "CenterOfMass": "center_of_mass_offset",
     "LinearMomentum": "linear_velocity",
+    "TotalEnergy": "radial_expansion_velocity",
```
nilsvu: Maybe name `AdmMass`, `AdmLinearMomentum` and `AdmAngularMomentumZ` to make clear what quantities are being controlled?

support/Pipelines/Bbh/ControlId.py
```diff
@@ -300,7 +333,10 @@ def Residual(u):
                 delayed_indices, [key in delayed_params] * 3
             )
 
-    while iteration < max_iterations and np.max(np.abs(F)) > residual_tolerance:
+    while iteration < max_iterations and (
+        np.max(np.abs(F)) > residual_tolerance
+        or iteration <= (max_delayed_iteration + 1)
```
nilsvu: Is this `+ 1` correct? Currently you delay by 1 iteration and want to make sure that zeroing out the residual doesn't prematurely stop the algorithm, right? Meaning that this condition should be the same as line 349 below?

Alternatively, you could also keep only the `iteration < max_iterations` here (or even make this `while True`) and `break` the loop below line 348 (right after computing the residual) if the stopping conditions are met.

nilsvu (changes requested):

support/Pipelines/Bbh/ControlId.py
iago-mendes: The +1 here is only because we have `iteration += 1` at the beginning of the loop. I like the idea of `break`ing because it also avoids an unnecessary update of the jacobian at the last iteration. I've changed the code below to do this.

support/Pipelines/Bbh/ControlId.py
iago-mendes: As we discussed offline, the cross terms indeed get activated by Broyden's method. Since we might also want to experiment with how we initialize the mass terms, I'll let this for a future PR. I've added a comment regarding this.

support/Pipelines/Bbh/ControlId.py
```diff
@@ -333,10 +344,10 @@ def Residual(u):
                 delayed_indices, [key in delayed_params] * 3
             )
 
-    while iteration < max_iterations and (
-        np.max(np.abs(F)) > residual_tolerance
-        or iteration <= (max_delayed_iteration + 1)
-    ):
+    temp_file = open(f"{id_run_dir}/../ControlJacobian.txt", "w")
```
nilsvu: Do you prefer this over writing the Jacobian into each iteration directory?

support/Pipelines/Bbh/ControlId.py
```diff
@@ -349,9 +360,20 @@ def Residual(u):
         if iteration <= max_delayed_iteration:
             F[delayed_indices] = 0.0
 
+        # Check stopping condition
+        if (
+            np.max(np.abs(F)) < residual_tolerance
+            and iteration > max_delayed_iteration
```
nilsvu: If you move this stopping criteria up, right below `F = Residual(u)`, then you don't need this line. That's better because it avoids running always at least `max_delayed_iteration` iterations, even if the residual is small already.

nilsvu (changes requested):

support/Pipelines/Bbh/ControlId.py
iago-mendes: That's a good point! I've moved the stopping condition as you suggested.

support/Pipelines/Bbh/ControlId.py
iago-mendes: I actually think we should combine as much of the control diagnostics as possible. For instance, if someone encounters a weird control behavior, it would be nice if they could send us a `ControlParams.h5` file, which would contain maybe `Residuals.dat` and `Jacobian.dat`. This would also help organize what we currently save into `ControlParamsData.txt`, which was supposed to be a temporary solution.

I think we can do all this along with the PR that addresses the Jacobian cross-terms because there's no reason to output the Jacobian in this PR. To that end, I've now removed this part of the code. Do you think that makes sense?

nilsvu (dismissed):

iago-mendes:
I've fixed `Test_InitialData.py` and squashed the commits. Thanks for the review!

nilsvu (approved):

PR #6506, opened 2025-02-28
Enable self force evolution of worldtube

nikwit:
## Proposed changes

This PR enables the self force evolution of the worldtube executable. It also changes the observation from an axis to an extraction sphere so waveform modes can be computed.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
@@ -16,6 +16,7 @@
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/SendAccelerationTerms.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/SendToElements.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/UpdateAcceleration.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/UpdateFunctionsOfTime.hpp"
 #include "IO/Observer/Actions/RegisterSingleton.hpp"
 #include "Options/String.hpp"
 #include "Parallel/Algorithms/AlgorithmSingleton.hpp"
@@ -25,6 +26,7 @@
 #include "Parallel/PhaseDependentActionList.hpp"
 #include "Parallel/Tags/ResourceInfo.hpp"
 #include "ParallelAlgorithms/Actions/AddComputeTags.hpp"
+#include "ParallelAlgorithms/Actions/FunctionsOfTimeAreReady.hpp"
 #include "ParallelAlgorithms/Actions/InitializeItems.hpp"
 #include "ParallelAlgorithms/Actions/MutateApply.hpp"
 #include "ParallelAlgorithms/Actions/TerminatePhase.hpp"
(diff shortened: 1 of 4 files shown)


tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
```diff
@@ -147,14 +146,11 @@ Worldtube:
...
-    End: [0., 0., 400.]
-    NumberOfPoints: 1000
-  PsiAlongAxis2:
-    Begin: [0., 0., 0.]
-    End: [400., 0., 0.]
-    NumberOfPoints: 1000
+  Spheres:
+    LMax: 10
+    Center: [0.,0.,0.]
+    Radius: [300., 400.]
```
knelli2: So with the `ObserveSurfaceData` callback, it can only handle a single radius at the moment. I'm not sure what happens if you use multiple radii. I don't think it's too much work to allow it to handle multiple radii, I just haven't taken the time to do it. So in the meantime, choose only one radius.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
```diff
@@ -87,14 +89,16 @@ struct WorldtubeSingleton {
...
+                 Actions::ReceiveElementData,
                  ::Actions::MutateApply<IterateAccelerationTerms>,
                  Actions::SendAccelerationTerms<Metavariables>,
                  ::Actions::MutateApply<UpdateAcceleration>,
                  ::Actions::RecordTimeStepperData<worldtube_system>,
                  ::Actions::UpdateU<worldtube_system>,
                  ::Actions::CleanHistory<worldtube_system, false>,
-                 Actions::SendToElements<Metavariables>>;
+                 Actions::SendToElements<Metavariables>,
+                 domain::Actions::CheckFunctionsOfTimeAreReady<Dim>>;
```
knelli2, line 101: Why check if the functions of time are ready at the end of the actions?

knelli2 (changes requested):
You can squash

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
nikwit: They should be checked after `UpdateFunctionsOfTime` but before `AdvanceTime` so this is the most efficient place for it.

tests/InputFiles/CurvedScalarWave/WorldtubeKerrSchild.yaml
nikwit: Hm, perhaps this is related to the weird error I have seen where it would just not write to file sometimes.

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonChare.hpp
knelli2: Are the functions of time needed in any of the actions between `UpdateFunctionsOfTime` and `CheckFoTAreReady`? If so, how do you guarantee they are up to date?

nikwit: No, they are not needed. The singleton computes all the orbit parameters of the particle and then sets the functions of time accordingly. All the actions on the singleton simply use the orbit parameters directly rather than going through the functions of time.

knelli2 (approved):

PR #6507, opened 2025-02-28
Print worldtube status

nikwit:
## Proposed changes

This PR prints a handy, short update about the worldtube orbit to stdout when an observation is done.

The PR also removes an obsolete tag, corrects a wrong descritption and fixes a random test failure.

### DIFF
--- a/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
+++ b/src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
@@ -10,6 +10,7 @@
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "DataStructures/DataBox/Prefixes.hpp"
 #include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
 #include "Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp"
 #include "IO/Observer/Helpers.hpp"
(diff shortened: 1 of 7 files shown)


src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
```diff
@@ -118,6 +120,14 @@ struct ObserveWorldtubeSolution {
         }
       }();
       const auto current_time = db::get<::Tags::Time>(box);
+      Parallel::printf(MakeString{}
+                       << "Time: " << std::setprecision(16) << current_time
```
knelli2: You should add a tag for observation verbosity and wrap this in an `if` so people can turn this off if it gets annoying or is printing too much data.
```cpp
if (Parallel::get<Tags::WorldtubeVerbosity>(cache) >= ::Verbosity::Verbose) {
  ...
}
```

See the control system example in `src/ControlSystem/Tags/SystemTags.hpp{OptionTags.hpp}`

knelli2 (changes requested):

src/Evolution/Systems/CurvedScalarWave/Worldtube/SingletonActions/ObserveWorldtubeSolution.hpp
wthrowe: [drive-by comment] Also, use printf properly with a format string.  If you want to use MakeString, you can just use the format "%s".  The reason for this is that if you do it the way you have and something you're printing happens to have a % in its output your code will crash.

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
```diff
@@ -229,6 +230,16 @@ struct ExpansionOrder {
   static size_t upper_bound() { return 1; }
   using group = Worldtube;
 };
+
+/*!
+ * \brief The value of the scalar charge in units of the black hole mass M.
+ */
```
knelli2: Oops, dox are for the wrong tag

knelli2 (changes requested):
One small thing

src/Evolution/Systems/CurvedScalarWave/Worldtube/Tags.hpp
nikwit: Good catch!

knelli2 (approved):

PR #6510, opened 2025-03-01
Store the DG Basis in the Block.

kidder:
The dg_basis will be used to determine the topology of the Block as well as the Spectral::Basis used for a Mesh on an Element of the Block.  It is default initialized to Spectral::Basis::Legendre as that has been the previous choice.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/tests/Unit/Domain/Test_Block.cpp
+++ b/tests/Unit/Domain/Test_Block.cpp
@@ -24,7 +24,9 @@
 #include "Domain/Structure/DirectionMap.hpp"
 #include "Domain/Structure/OrientationMap.hpp"
 #include "Framework/TestHelpers.hpp"
+#include "NumericalAlgorithms/Spectral/Basis.hpp"
 #include "Utilities/GetOutput.hpp"
+#include "Utilities/MakeArray.hpp"
 #include "Utilities/Serialization/CharmPupable.hpp"
 
 namespace domain {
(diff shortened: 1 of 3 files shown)


tests/Unit/Domain/Test_Block.cpp
```diff
@@ -351,5 +357,13 @@ SPECTRE_TEST_CASE("Unit.Domain.Block", "[Domain][Unit]") {
...
     CHECK(block != rhs);
   }
+  {
+    const Block<2> rhs(
+        identity_map.get_clone(), 3, neighbors, "Identity",
+        {{Spectral::Basis::Chebyshev, Spectral::Basis::Legendre}});
+    CHECK(rhs.dg_basis() ==
+          std::array{Spectral::Basis::Chebyshev, Spectral::Basis::Legendre});
+    CHECK(block != rhs);
+  }
```
nilsdeppe, line 367: Could you add a test for Ylm basis? Mostly just to provide an example for folks in the future with a 2d basis.

nilsdeppe (dismissed):
LGTM, one request. Please squash immediately

tests/Unit/Domain/Test_Block.cpp
kidder: Such a test does not make sense until the next PR which fixes assumptions about external boundaries.   So I can either add the commits for that PR to this one, or you can wait, doesn't matter to me...

nilsdeppe:
Just wait then!!

PR #6514, opened 2025-03-03
Elliptic DG: support modifying boundary data

nilsvu:
## Proposed changes

This is used to solve for different variables in some regions of the domain to absorb singularities.
For example, when solving $-\Delta u = f$ we could define $u=u_R + u_P$ in some region of the domain with a known (often singular) field $u_P$ and solve only for the regular field $u_R$ in this region. So, when receiving boundary data from outside this region, we subtract $u_P$, and when receiving boundary data from inside this region, we add $u_P$. We also add $\Delta u_P$ to the volume fixed sources $f$ inside the regularized region.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/DiscontinuousGalerkin/DgOperator.hpp
+++ b/src/Elliptic/DiscontinuousGalerkin/DgOperator.hpp
@@ -517,11 +517,11 @@ struct DgOperatorImpl<System, Linearized, tmpl::list<PrimalFields...>,
 
   // --- This is essentially a break to communicate the mortar data ---
 
-  template <bool AllDataIsZero, typename... OperatorTags,
-            typename... PrimalVars, typename... PrimalFluxesVars,
-            typename... PrimalMortarVars, typename... PrimalMortarFluxes,
-            typename TemporalId, typename... FluxesArgs,
-            typename... SourcesArgs, typename DataIsZero = NoDataIsZero,
+  template <typename... OperatorTags, typename... PrimalVars,
+            typename... PrimalFluxesVars, typename... PrimalMortarVars,
+            typename... PrimalMortarFluxes, typename TemporalId,
+            typename... FluxesArgs, typename... SourcesArgs,
+            typename DataIsZero = NoDataIsZero,
             typename DirectionsPredicate = AllDirections>
   static void apply_operator(
       const gsl::not_null<Variables<tmpl::list<OperatorTags...>>*>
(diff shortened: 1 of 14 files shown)


src/Elliptic/DiscontinuousGalerkin/DgOperator.hpp
```diff
@@ -1016,23 +1012,23 @@ struct DgOperatorImpl<System, Linearized, tmpl::list<PrimalFields...>,
...
-      const DirectionMap<Dim, std::tuple<FluxesArgs...>>&
-          fluxes_args_on_faces) {
+      const DirectionMap<Dim, std::tuple<FluxesArgs...>>& fluxes_args_on_faces,
+      const std::tuple<ModifyBoundaryDataArgs...>& modify_boundary_data_args) {
     // We just feed zero variables through the nonlinear operator to extract the
     // constant contribution at external boundaries. Since the variables are
     // zero the operator simplifies quite a lot. The simplification is probably
     // not very important for performance because this function will only be
     // called when solving a linear elliptic system and only once during
     // initialization, but we specialize the operator for zero data nonetheless
```
wthrowe, line 1028: It looks like you've removed this specialization, so update the comment.

tests/Unit/Elliptic/DiscontinuousGalerkin/Test_DgOperator.cpp
```diff
@@ -253,6 +253,91 @@ struct Metavariables {
...
+  using Poisson::Solutions::ProductOfSinusoids<Dim>::ProductOfSinusoids;
+  std::unique_ptr<elliptic::analytic_data::AnalyticSolution> get_clone()
+      const override {
+    return std::make_unique<ModifiedPoissonSolution>(*this);
+  }
+  using PUP::able::register_constructor;
+  WRAPPED_PUPable_decl_template(ModifiedPoissonSolution);  // NOLINT
+  template <typename... RequestedTags>
+
+  tuples::TaggedTuple<RequestedTags...> variables(
```
wthrowe: Formatting.  I was looking all over for where `RequestedTags` was defined.

wthrowe (changes requested):

src/Elliptic/DiscontinuousGalerkin/DgOperator.hpp
nilsvu: The `prepare_mortar_data` (first half of the DG operator) still has the specialization. So keep comment I guess?

wthrowe: Ah, yep, missed that.  Good as-is, then.

tests/Unit/Elliptic/DiscontinuousGalerkin/Test_DgOperator.cpp
nilsvu: fixed

wthrowe (approved):

PR #6518, opened 2025-03-11
Send more quantities to size control for state 3

knelli2:
## Proposed changes

These quantities will be needed for state 3 once it is added in order to compute the total derivative of the comoving characteristic speeds with respect to the map parameter $\lambda_{0,0}$.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Tags/QueueTags.hpp
+++ b/src/ControlSystem/Tags/QueueTags.hpp
@@ -6,6 +6,7 @@
 #include "DataStructures/DataBox/Prefixes.hpp"
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Tags.hpp"
+#include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
 #include "Utilities/TaggedTuple.hpp"
 
 /// \cond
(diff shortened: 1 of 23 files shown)


tests/Unit/ParallelAlgorithms/ApparentHorizonFinder/Test_ComputeExcisionBoundaryVolumeQuantities.cpp
```diff
@@ -443,15 +532,24 @@ SPECTRE_TEST_CASE(
...
       tmpl::list<gr::Tags::SpacetimeMetric<DataVector, 3>,
                  gh::Tags::Pi<DataVector, 3>, gh::Tags::Phi<DataVector, 3>>,
       tmpl::list<gr::Tags::SpacetimeMetric<DataVector, 3>,
                  gr::Tags::SpatialMetric<DataVector, 3>,
-                 gr::Tags::Shift<DataVector, 3>>>();
+                 gr::Tags::Shift<DataVector, 3>,
+                 ::Tags::deriv<gr::Tags::Shift<DataVector, 3, Frame::Inertial>,
+                               tmpl::size_t<3>, Frame::Inertial>
+                 //   domain::Tags::InverseJacobian<3, Frame::Grid,
+                 //   Frame::Inertial>
```
nilsdeppe: Remove? Uncomment?

src/ControlSystem/Tags/QueueTags.hpp
```diff
@@ -88,6 +89,34 @@ struct InverseSpatialMetricOnExcisionSurface {
...
+  using type = tnsr::iJ<DataVector, 3, Frame>;
+};
+
+/// \ingroup ControlSystemGroup
+/// Holds the inverse jacobian between frames on the `ExcisionSurface`
+template <typename SrcFrame, typename DestFrame>
+struct InverseJacobian {
+  using type = ::InverseJacobian<DataVector, 3, SrcFrame, DestFrame>;
+};
+
```
nilsdeppe, line 119: Add tests for these tags?

nilsdeppe (changes requested):
LGTM, you can squash immediately

src/ControlSystem/Tags/QueueTags.hpp
knelli2: These aren't DataBox tags. Though I guess I could just use the existing GR tags for this

nilsdeppe: Oh, right. I would say just leave as-is? Could also just inherit off `db::SimpleTag`? I don't have a preferences

knelli2: I think I'll leave as is so it's the same as the other Queue tags in this file

knelli2:
Squashed. Thanks!

nilsdeppe (approved):

PR #6521, opened 2025-03-15
Correctly identify external Block boundaries for Topologies other than I1

kidder:
## Proposed changes

- Add a function has_boundary that returns whether or not a Topology has a boundary (external or internal) on a Side
- Use the function to correctly set external boundaries of a Block, and to check that input neighbors are in valid directions

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Block.cpp
+++ b/src/Domain/Block.cpp
@@ -14,6 +14,7 @@
 #include "Domain/Structure/BlockNeighbors.hpp"
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/DirectionMap.hpp"
+#include "Domain/Structure/HasBoundary.hpp"
 #include "NumericalAlgorithms/Spectral/Basis.hpp"
 #include "Utilities/Algorithm.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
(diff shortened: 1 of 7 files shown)


src/Domain/Block.cpp
```diff
@@ -34,8 +35,17 @@ Block<VolumeDim>::Block(
   // Loop over Directions to search which Directions were not set to neighbors_,
   // set these Directions to external_boundaries_.
   for (const auto& direction : Direction<VolumeDim>::all_directions()) {
+    const bool has_boundary_in_this_direction = has_boundary(
+        gsl::at(topologies_, direction.dimension()), direction.side());
     if (neighbors_.find(direction) == neighbors_.end()) {
-      external_boundaries_.emplace(direction);
+      if (has_boundary_in_this_direction) {
+        external_boundaries_.emplace(direction);
+      }
```
nilsvu, line 43: You can also assert here that the number of neighbors is 1 for cubes and either 1 or 6 for shells

src/Domain/Structure/HasBoundary.hpp
```diff
@@ -0,0 +1,21 @@
...
+
+/// \cond
+namespace domain {
+enum class Topology : uint8_t;
+}  // namespace domain
+enum class Side : uint8_t;
+/// \endcond
+
+namespace domain {
+/// \brief Whether or not a Topology has a boundary on a given
```
nilsvu: ... side

tests/Unit/Domain/Test_Block.cpp
```diff
@@ -291,9 +291,73 @@ void test_block_time_dependent_distorted() {
       translation_distorted_to_inertial_map.get_clone());
   test_move_semantics(std::move(original_block), block_copy);
 }
+
+void test_spherical_shell() {
+  const Block<3> spherical_shell(
+      nullptr, 4, DirectionMap<3, BlockNeighbors<3>>{}, "Shell",
+      std::array{domain::Topology::I1, domain::Topology::S2Colatitude,
```
nilsvu, line 298: Would it help to define this array in Topology.hpp as `spherical_shell_topology`? It will come up more often

nilsvu (changes requested):

src/Domain/Block.cpp
kidder: no, that is not true in general (e.g. in a modified CylindricalCompactBinaryObject domain)

tests/Unit/Domain/Test_Block.cpp
kidder: possibly, though I'm not sure that is the right place to put it.  We can discuss it at the next tech call.

nilsvu (approved):

PR #6531, opened 2025-03-27
Correctly treat external boundaries in Elements with new topologies

kidder:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Structure/Element.cpp
+++ b/src/Domain/Structure/Element.cpp
@@ -3,16 +3,20 @@
 
 #include "Domain/Structure/Element.hpp"
 
+#include <array>
 #include <ostream>
 #include <pup.h>
 
+#include "Domain/Structure/HasBoundary.hpp"
 #include "Domain/Structure/MaxNumberOfNeighbors.hpp"
+#include "Domain/Structure/Topology.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/StdHelpers.hpp"
 
 template <size_t VolumeDim>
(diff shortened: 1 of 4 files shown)


src/Domain/Structure/Element.cpp
```diff
@@ -22,22 +26,22 @@ Element<VolumeDim>::Element(ElementId<VolumeDim> id, Neighbors_t neighbors)
...
-        return external_boundaries;
-      }()),
-      internal_boundaries_([this]() {
-        std::unordered_set<Direction<VolumeDim>> internal_boundaries;
-        for (const auto& direction_neighbors : neighbors_) {
-          internal_boundaries.insert(direction_neighbors.first);
-        }
-        return internal_boundaries;
-      }()) {
+      topologies_(std::move(topologies)) {
```
nilsvu, line 29: I believe it's not necessary to store the topology in every element because you always have access to it in the block. So you can either keep this constructor like you have it but just not store the topology, or you can pass in the `external_boundaries` instead of the topology.

nilsdeppe (approved):
LGTM.

I think we should store the topology in the element, at least right now. Having to pass the block around to a bunch of places will be annoying, and I think there are cases where you can conceivably have different topologies inside the same block. E.g. if you do h-refinement on a B3 in the radial direction you have two elements, one is B3 and is I1xS2 and they do not have to be different blocks, I think (both use Ylm angular collocation).

nilsvu (approved):
ok fine with me

nilsdeppe:
Nils and I discussed this offline and we will try this for now. It offers the most flexibility and we can remove/change it if necessary.

PR #6534, opened 2025-03-30
Allow CCE ObserveFields to write volume data synchronously

knelli2:
## Proposed changes

When running CCE on one core, volume data that is sent to the ObserverWriter for writing could be held in memory for a long time (until the end) because no core is free. This could cause an OOM error. This PR forces CCE to write ObserveFields data synchronously if running on a single core to avoid that.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/Events/ObserveFields.hpp
+++ b/src/Evolution/Systems/Cce/Events/ObserveFields.hpp
@@ -30,6 +30,8 @@
 #include "Options/ParseError.hpp"
 #include "Options/String.hpp"
 #include "Parallel/GlobalCache.hpp"
+#include "Parallel/Info.hpp"
+#include "Parallel/Invoke.hpp"
 #include "ParallelAlgorithms/EventsAndTriggers/Event.hpp"
 #include "Time/Tags/Time.hpp"
 #include "Utilities/Gsl.hpp"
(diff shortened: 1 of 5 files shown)


src/Evolution/Systems/Cce/Events/ObserveFields.hpp
```diff
@@ -130,7 +131,9 @@ std::string name() {
  * spin weighted quantities above (and time as the first column).
  *
  * All data will be written into the `observers::OptionTags::ReductionFileName`
- * file.
+ * file. If `WriteSynchronously` is true, then this will write the volume data
+ * immediately instead of sending it to the ObserverWriter to be written
+ * asynchronously.
  */
```
nilsdeppe: Should we just check if running on 1 core then do sync output? Less thought required on the user part

nilsdeppe (changes requested):

src/Evolution/Systems/Cce/Events/ObserveFields.hpp
knelli2: Yeah that's probably good. Even if it's slower, probably better than running out of memory

nilsdeppe (commented):
LGTM, please squash!

nilsdeppe (approved):

PR #6536, opened 2025-03-30
Disable black formatting in external python files

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,9 +14,17 @@ required-version = "23.3.0"
 # this feature is tracked in this issue:
 # https://github.com/psf/black/issues/2188
 preview = true
+# Skip external code since modifying it can be an issue with GPL-licensed code.
+extend-exclude = '''
+(
+  ^/external
+)
+'''
 
 [tool.isort]
 profile = "black"
 line_length = 80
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -14,6 +14,11 @@ required-version = "23.3.0"
 # this feature is tracked in this issue:
 # https://github.com/psf/black/issues/2188
 preview = true
+exclude = '''
```
nilsvu: Use `extend-exclude` so that other directories like `.git` remain excluded. Also, the patterns are checked recursively, so add `^/`. And should we just ignore all of `^/external`? This seems to work:

```
extend-exclude = '''
(
  ^/external
)
'''
```

Also add this to the `isort` config below:

```
[tool.isort]
# ...
extend_skip = ["external"]
```

nilsvu (changes requested):

pyproject.toml
nilsdeppe: Okay, done! Thanks for the review :)

nilsvu (approved):

PR #6537, opened 2025-03-30
Add GridCenters control system support to BCO for tracking NS centers

nilsdeppe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/TimeDependentOptions/GridCenters.cpp
+++ b/src/Domain/Creators/TimeDependentOptions/GridCenters.cpp
@@ -1,0 +1,101 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/Creators/TimeDependentOptions/GridCenters.hpp"
+
+#include <array>
+#include <fstream>
+#include <iterator>
+#include <optional>
+#include <regex>
+#include <string>
+#include <variant>
+
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/PiecewisePolynomial.hpp"
+#include "Options/ParseError.hpp"
+#include "Utilities/FileSystem.hpp"
+
+namespace domain::creators::time_dependent_options {
+GridCentersOptions::GridCentersOptions() = default;
+
+GridCentersOptions::GridCentersOptions(
+    const std::string& spec_evolution_parameters_perl_file,
+    const std::optional<double> in_scale_inspiral_rate_by,
+    const Options::Context& context)
+    : initial_values{DataVector{6}, DataVector{6}, DataVector{6, 0.0}},
+      scale_inspiral_rate_by(in_scale_inspiral_rate_by) {
+  if (scale_inspiral_rate_by.has_value() and
+      scale_inspiral_rate_by.value() <= 0.0) {
+    PARSE_ERROR(context,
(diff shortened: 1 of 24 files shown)


src/Domain/Creators/TimeDependentOptions/GridCenters.cpp
```diff
@@ -0,0 +1,101 @@
...
+GridCentersOptions::GridCentersOptions(
+    const std::string& spec_evolution_parameters_perl_file,
+    const std::optional<double> in_scale_inspiral_rate_by,
+    const Options::Context& context)
+    : initial_values{DataVector{6}, DataVector{6}, DataVector{6, 0.0}},
+      scale_inspiral_rate_by(in_scale_inspiral_rate_by) {
+  if (scale_inspiral_rate_by.has_value() and
+      scale_inspiral_rate_by.value() <= 0.0) {
+    PARSE_ERROR(context,
+                "The inspiral rate must be scale by a positive number but got "
```
knelli2: [nitpick] "scale" -> "scaled"

src/Domain/Creators/TimeDependentOptions/GridCenters.hpp
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * \details This class can also be used as an option tag with the \p type type
+ * alias, `name()` function, and \p help string.
+ */
+struct GridCentersOptions {
+  using type = Options::Auto<GridCentersOptions, Options::AutoLabel::None>;
+  static std::string name() { return "GridCenters"; }
+
+  struct SpecEvolutionParametersPerlFile {
+    using type = std::string;
```
knelli2, line 32: Do you have a plan for what to do once we have BNS ID in spectre?

tests/InputFiles/GrMhd/GhValenciaDivClean/BinaryNeutronStar.yaml
```diff
@@ -408,6 +408,22 @@ ControlSystems:
       DecreaseFactor: 0.98
     ControlError:
   Translation: None
+  GridCenters:
+    IsActive: true
```
knelli2: Remove the `IsActive:` option.

knelli2 (changes requested):
You can squash these in directly.

I know we discussed this offline, but just to have a record of it somewhere. Having to have the `GridCenters:` option in the BCO domain for all systems besides the BNS that will likely never use the grid centers is a tad annoying. But I also don't have a good solution for this besides adding yet another template parameter to the BCO class so this is alright for now.

src/Domain/Creators/TimeDependentOptions/GridCenters.hpp
nilsdeppe: No, because I don't know what that will output. Probably I will request that the same quantities are dumped somewhere that I can parse them out, then have an option that's a variant or something to support both SpEC and SpECTRE ID. I'm not worried about it, worst case a pipeline would inject these numbers.

nilsdeppe (commented):
Squashed and rebased. Thanks for the review :)

And in regards to it being in the domain creator block, I agree. BNS is also currently stuck with BBH options:
```yaml
      SkewMap: None
      ShapeMapA: None
      ShapeMapB: None
```

knelli2 (approved):

PR #6539, opened 2025-03-31
Support disabling control systems in UpdateFunctionOfTime

nilsdeppe:
## Proposed changes

This is needed because in BNS simulations we disable the rotation control system once the two NSs are close enough to merger.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/UpdateFunctionOfTime.hpp
+++ b/src/ControlSystem/UpdateFunctionOfTime.hpp
@@ -10,6 +10,7 @@
 #include <unordered_set>
 #include <utility>
 
+#include "ControlSystem/Tags/IsActiveMap.hpp"
 #include "DataStructures/DataBox/DataBox.hpp"
 #include "DataStructures/DataVector.hpp"
 #include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
@@ -109,8 +110,12 @@ struct UpdateAggregator {
   /*!
    * \brief Checks if `insert` has been called for all control systems that this
    * class was constructed with.
+   *
+   * The `active_map` is used to inform the aggregator which controls systems
+   * are currently active. This is separate so that individual control
+   * systems can be enabled and disabled during an evolution.
    */
-  bool is_ready() const;
+  bool is_ready(const std::unordered_map<std::string, bool>& active_map) const;
 
   /*!
    * \brief Returns a sorted concatenation of the control system names this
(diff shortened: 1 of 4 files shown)


src/ControlSystem/UpdateFunctionOfTime.hpp
```diff
@@ -109,8 +110,12 @@ struct UpdateAggregator {
   /*!
    * \brief Checks if `insert` has been called for all control systems that this
    * class was constructed with.
+   *
+   * The `active_map` is used to inform the aggregator which controls systems
+   * are currently active. This is separate so that individual controls
```
knelli2: [nitpick] "controls" -> "control". Same below

tests/Unit/ControlSystem/Test_UpdateFunctionOfTime.cpp
```diff
@@ -208,14 +214,14 @@ void test_update_aggregator() {
 #endif
   aggregator.insert("California", DataVector{5.0}, 6.0,
                     signals.at("California"), 1.0);
-  CHECK(aggregator.is_ready());
+  CHECK(aggregator.is_ready(active_map));
```
knelli2, line 221: Before the call to `.insert` can you edit the `active_map` to turn `California` off and check that the aggregator is ready, and then turn `California` back on for the rest of the test? Just so we test dynamically turning things on and off.

knelli2 (changes requested):

tests/Unit/ControlSystem/Test_UpdateFunctionOfTime.cpp
nilsdeppe: I think I did the requested thing right?? 🤞 Let me know if any other changes would be useful to help test :)

knelli2: Sorry, this was my bad for not being specific. Could you do
```cpp
#ifdef
...
#endif
active_map.at("California") = false;
CHECK(aggregator.is_ready(active_map));
active_map.at("California") = true;
CHECK_FALSE(aggregator.is_ready(active_map));
aggregator.insert("California", DataVector{5.0}, 6.0,
                  signals.at("California"), 1.0);
CHECK(aggregator.is_ready(active_map));
```
Since the control system will be turned off and not sending anything to the aggregator, it's important to check that we can not receive a measurement for a disabled control system and still be ready.

nilsdeppe: Ah okay, thanks! Done :)

knelli2 (approved):

PR #6540, opened 2025-03-31
Support tracking ObjectCenters in TimeDependentTripleGaussian

nilsdeppe:
## Proposed changes

This is necessary to have the Gaussian centers move radially with the BNS. Rotation control already takes care of the angular movement, but since BNS doesn't use an expansion map, we need to track the object centers explicitly.

### Upgrade instructions

The time dependent triple gaussian input file now needs:
```
        MovementMethod: ExpansionFactor
```
to preserve the same behavior.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GeneralizedHarmonic/ConstraintDamping/TimeDependentTripleGaussian.hpp
+++ b/src/Evolution/Systems/GeneralizedHarmonic/ConstraintDamping/TimeDependentTripleGaussian.hpp
@@ -12,6 +12,7 @@
 
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "Evolution/Systems/GeneralizedHarmonic/ConstraintDamping/DampingFunction.hpp"
+#include "Options/Auto.hpp"
 #include "Options/String.hpp"
 #include "Utilities/Serialization/CharmPupable.hpp"
 #include "Utilities/TMPL.hpp"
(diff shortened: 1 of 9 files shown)


src/Evolution/Systems/GeneralizedHarmonic/ConstraintDamping/TimeDependentTripleGaussian.hpp
```diff
@@ -85,14 +90,30 @@ class TimeDependentTripleGaussian : public DampingFunction<3, Frame::Grid> {
   template <typename Group>
   struct Center {
     using group = Group;
-    using type = std::array<double, 3>;
+    using type = tmpl::conditional_t<
+        std::is_same_v<Group, Gaussian<3>>, std::array<double, 3>,
+        Options::Auto<std::array<double, 3>, Options::AutoLabel::None>>;
```
knelli2: Why `None` here instead of `Auto`? Not really sure what a "none" center is for a Gaussian

knelli2 (changes requested):
You can squash directly

src/Evolution/Systems/GeneralizedHarmonic/ConstraintDamping/TimeDependentTripleGaussian.hpp
nilsdeppe: Agreed, changed.

knelli2 (approved):

PR #6541, opened 2025-03-31
Support GridCenters BNS tracking in SeparationLessThan

nilsdeppe:
## Proposed changes

I need a trigger to use when the stars get too close and we need to disable rotation control.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
+++ b/src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
@@ -104,6 +104,7 @@
 #include "Evolution/Systems/GrMhd/ValenciaDivClean/System.hpp"
 #include "Evolution/Systems/GrMhd/ValenciaDivClean/Tags.hpp"
 #include "Evolution/Tags/Filter.hpp"
+#include "Evolution/Triggers/SeparationLessThan.hpp"
 #include "Evolution/TypeTraits.hpp"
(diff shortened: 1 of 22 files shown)


src/Evolution/Executables/GrMhd/GhValenciaDivClean/GhValenciaDivCleanBase.hpp
```diff
@@ -645,8 +646,10 @@ struct GhValenciaDivCleanTemplateBase<
         tmpl::pair<TimeSequence<std::uint64_t>,
                    TimeSequences::all_time_sequences<std::uint64_t>>,
         tmpl::pair<TimeStepper, TimeSteppers::time_steppers>,
-        tmpl::pair<Trigger, tmpl::append<Triggers::logical_triggers,
-                                         Triggers::time_triggers>>>;
+        tmpl::pair<
+            Trigger,
+            tmpl::append<Triggers::logical_triggers, Triggers::time_triggers,
+                         tmpl::list<Triggers::SeparationLessThan<true>>>>>;
```
knelli2: This will add the trigger for all GhValencia execs. We probably only want this for the BNS since it doesn't really make sense for other systems. Use the `use_control_systems` bool to determine if you should add this trigger or not (that's kind of been the sentinel bool for BNS in these metavars)

knelli2 (changes requested):

nilsdeppe:
Sounds good! Did that, squashed. Thanks for the review :)

knelli2 (approved):

PR #6543, opened 2025-03-31
Add deadlock analysis to Worldtube CurvedScalarWave

knelli2:
## Proposed changes

Also adds `ExecutePhaseChange` to the Worldtube CSW singleton component.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
+++ b/src/Evolution/Executables/CurvedScalarWave/EvolveWorldtubeCurvedScalarWave.hpp
@@ -22,6 +22,7 @@
 #include "Evolution/DiscontinuousGalerkin/DgElementArray.hpp"
 #include "Evolution/DiscontinuousGalerkin/Initialization/Mortars.hpp"
 #include "Evolution/DiscontinuousGalerkin/Initialization/QuadratureTag.hpp"
+#include "Evolution/Executables/GeneralizedHarmonic/Deadlock.hpp"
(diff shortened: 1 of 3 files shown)


guilara:
Thank you @knelli2. Adding the `ExecutePhaseChange` action to the worldtube singleton has solved the problem.

guilara (approved):
I've tested this on a failing run and this fix has solved the problem.

knelli2:
> There's a subtlety you should be aware of. So the reason the error happened is likely because
> 1. Elements check for phase change by halting the algorithm and saying they may want to go into a different phase. This > leaves the elements in a valid state
> 2. The singleton, however, doesn't know this so it just requests the elements for data and waits, leaving it in an invalid >state for switching phases
> 3. When the algorithm is restarted on the singleton, we get an error because we didn't exit in a valid state.
>
> Putting the ExecutePhaseChange action into the list makes the singleton be in a valid state when switching phases. The subtlety comes in because the trigger for the phase change is checked on both the elements and the singleton which have different time steps/slabs and such. Now the singleton happens to have the same slab size as the elements and is also doing lock-step with the WT boundary, so triggers based on slab number or time value will work for both. But if the slab sizes become different (or you don't do lock-step anymore so the times are out of sync) then you may get a deadlock/error because the elements would try to phase change, but the singleton wouldn't just like currently

Putting this here as record from a conversation with @guilara.

nilsdeppe (approved):

PR #6546, opened 2025-04-01
Exporter: split functions for reuse

nilsvu:
## Proposed changes

Split up the interpolation functions so that they can be reused by the new `PointwiseInterpolator` class, which holds volume data in memory and supports repeated interpolation to arbitrary points. This can be used to load data once and then interpolate many times, e.g. for ray tracing. A follow-up PR will do the time interpolation that's needed for ray tracing, which will use the `PointwiseInterpolator` class as well.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
@knelli2 on your request I dropped all but the first 3 commits for now. These just split up the functions so they can be reused. Mind taking a look?

src/IO/Exporter/Exporter.cpp
```diff
@@ -0,0 +1,52 @@
...
+        volume_files_or_glob,
+    const std::string& subfile_name, const ObservationVariant& observation,
+    const std::vector<std::string>& tensor_components,
+    const std::array<std::vector<double>, Dim>& target_points,
+    const bool extrapolate_into_excisions,
+    const std::optional<size_t> num_threads) {
+  tnsr::I<DataVector, Dim, Frame::Inertial> target_points_dv{};
+  for (size_t d = 0; d < Dim; ++d) {
+    target_points_dv.get(d).set_data_ref(
+        const_cast<double*>(gsl::at(target_points, d).data()),
```
knelli2: Thow some NOLINTs in there or use `make_const_view`

src/IO/Exporter/PointwiseInterpolator.cpp
```diff
@@ -89,8 +97,21 @@ void interpolate_to_points(
       tensor_data.push_back(std::move(double_component_data));
     }
   }
-  h5file.close();
-#pragma omp parallel num_threads(num_threads)
+  return tensor_data;
+}
+
+template <typename ResultDataType, size_t Dim>
+void _interpolate_to_points(
```
knelli2: Why the underscores at the beginning of the function names? I thought our convention with functions in cpp files was just an anonymous namespace

src/IO/Exporter/PointwiseInterpolator.cpp
```diff
@@ -129,6 +149,23 @@ void interpolate_to_points(
   }  // omp parallel
 }
 
+template <size_t Dim>
+void _interpolate_to_point(
```
knelli2: This function doesn't seem to be used anywhere. Is it necessary?

knelli2 (changes requested):
Clang-tidy has some reasonable errors as well

src/IO/Exporter/Exporter.cpp
nilsvu: I don't think we have `make_const_view` for `double*`, so I added the nolint

src/IO/Exporter/PointwiseInterpolator.cpp
nilsvu: changed underscore to `_impl`

src/IO/Exporter/PointwiseInterpolator.cpp
nilsvu: Oh it will be used by the interpolator class in the next PR. I can remove it for now

nilsvu:
Thanks for reviewing! I pushed two fixups

knelli2 (commented):
Looks good, you can squash

nilsvu:
Done 👍

knelli2 (approved):

PR #6548, opened 2025-04-02
Add disabling rotation control system to BNS

nilsdeppe:
## Proposed changes

Adds an event to disable the rotation control system for BNS. Also updates the input files to what we are doing nowadays in simulations.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Actions/GridCenters.hpp
+++ b/src/ControlSystem/Actions/GridCenters.hpp
@@ -1,0 +1,232 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <cmath>
+#include <limits>
+#include <memory>
+
+#include "ControlSystem/Tags/IsActiveMap.hpp"
+#include "ControlSystem/Tags/OptionTags.hpp"
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataBox/Tag.hpp"
+#include "Domain/FunctionsOfTime/FunctionOfTime.hpp"
+#include "Domain/FunctionsOfTime/SettleToConstantQuaternion.hpp"
+#include "Domain/FunctionsOfTime/Tags.hpp"
+#include "Domain/Structure/ElementId.hpp"
+#include "Domain/Tags.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/GlobalCache.hpp"
+#include "ParallelAlgorithms/EventsAndTriggers/EventsAndTriggers.hpp"
+#include "ParallelAlgorithms/EventsAndTriggers/Tags.hpp"
+#include "ParallelAlgorithms/EventsAndTriggers/WhenToCheck.hpp"
+#include "Time/Tags/Time.hpp"
+#include "Time/Tags/TimeStepId.hpp"
+#include "Utilities/ConstantExpressions.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
+
+/// \cond
+namespace Parallel {
+template <typename Metavariables>
+class GlobalCache;
+}  // namespace Parallel
+namespace Tags {
+struct TimeStep;
+struct TimeStepId;
+template <typename StepperInterface>
+struct TimeStepper;
+}  // namespace Tags
+namespace tuples {
+template <class... Tags>
+class TaggedTuple;
+}  // namespace tuples
+/// \endcond
+
+namespace control_system {
+/*!
+ * \brief Holds options used to control when to start disabling the rotation
+ * control system in a BNS simulation.
+ */
+struct DisableRotationWhen {
+  /// The separation at which we start turning off rotation.
+  struct DisableAtSeparation {
+    using type = double;
+    static type lower_bound() { return 2.0; }
+    static constexpr Options::String help{
+        "The separation at which we start turning off rotation."};
+  };
+
+  /// The timescale in code units over which grid rotation is disabled. A
+  /// reasonable value is 30M-60M.
+  struct RotationDecayTimescale {
+    using type = double;
+    static type lower_bound() { return 30.0; }
+    static type upper_bound() { return 200.0; }
+    static constexpr Options::String help{
+        "The timescale in code units over which grid rotation is disabled. A "
+        "reasonable value is 40M-60M."};
+  };
+
+  using options = tmpl::list<DisableAtSeparation, RotationDecayTimescale>;
+
+  static constexpr Options::String help{
+      "Constrols the separation at which the rotation control system is "
+      "disabled and the rotation function of time starts settling to a "
+      "constant."};
+
+  void pup(PUP::er& p);
+
+  double disable_at_separation{std::numeric_limits<double>::signaling_NaN()};
+  double rotation_decay_timescale{std::numeric_limits<double>::signaling_NaN()};
+};
+
+namespace OptionTags {
+/// Option tag for controlling when and how the rotation map is disabled.
+struct DisableRotationWhen {
+  static constexpr Options::String help =
+      "Options for controlling how the rotation control system stops as the "
+      "two neutron stars merge.";
+  using type = control_system::DisableRotationWhen;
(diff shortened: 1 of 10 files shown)


tests/Unit/Framework/MockDistributedObject.hpp
```diff
@@ -443,7 +443,8 @@ class MockDistributedObject {
 
   /// Invoke the next action in the action list for the current phase,
   /// returning whether the action was ready.
-  bool next_action_if_ready();
+  template <typename CatchMatcher = int>
+  bool next_action_if_ready(const CatchMatcher* matcher = nullptr);
```
knelli2, line 447: Question: Could you not strongly type the match (possibly to a base class) and avoid the `int` placeholder?

src/ControlSystem/Actions/GridCenters.hpp
```diff
@@ -0,0 +1,217 @@
...
+
+  /// The timescale in code units over which grid rotation is disabled. A
+  /// reasonable value is 30M-60M.
+  struct RotationDecayTimescale {
+    using type = double;
+    static type lower_bound() { return 30.0; }
+    static type upper_bound() { return 200.0; }
+    static constexpr Options::String help{
+        "The timescale in code units over which grid rotation is disabled. A "
+        "reasonable value is 20M-60M."};
```
knelli2: This lower reasonable value is below the lower bound

src/ControlSystem/Actions/GridCenters.hpp
```diff
@@ -0,0 +1,217 @@
...
+      tmpl::list<control_system::Tags::DisableRotationWhen>;
+
+  template <typename DbTags, typename... InboxTags, typename Metavariables,
+            typename ActionList, typename ParallelComponent>
+  static Parallel::iterable_action_return_t apply(
+      db::DataBox<DbTags>& box, tuples::TaggedTuple<InboxTags...>& /*inboxes*/,
+      Parallel::GlobalCache<Metavariables>& cache,
+      const ElementId<3>& /*array_index*/, const ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    if (not is_zeroth_element(db::get<domain::Tags::Element<3>>(box).id())) {
```
knelli2: Could just use the array index here

src/ControlSystem/Actions/GridCenters.hpp
```diff
@@ -0,0 +1,217 @@
...
+          "function of time to a SettleToConstant. Current TimeStepId is "
+          << time_step_id);
+    }
+
+    const auto& bns_rotation_control =
+        db::get<control_system::Tags::DisableRotationWhen>(box);
+
+    const double time = db::get<::Tags::Time>(box);
+    const domain::FunctionsOfTime::FunctionOfTime& grid_centers_fot =
+        *get<domain::Tags::FunctionsOfTime>(cache).at("GridCenters");
```
knelli2, line 200: May want to `ASSERT` or even just outright check that `GridCenters` is in the FoTs since it comes from an input-file option that people could forget to add.

src/ControlSystem/Actions/GridCenters.hpp
```diff
@@ -0,0 +1,217 @@
...
+  static constexpr bool pass_metavariables = false;
+  static control_system::DisableRotationWhen create_from_options(
+      const control_system::DisableRotationWhen& disable_rotation_when);
+};
+}  // namespace Tags
+
+namespace Actions {
+/*!
+ * \brief Checks if the Rotation function of time has been updated because the
+ * separation between the neutron star grid centers is small enough.
```
knelli2, line 110: Can you add some dox that this should only be run during the `DisableRotationControl` phase? The code below errors if the separation isn't below the input separation which would happen immediately in the `Evolve` phase.

tests/InputFiles/GrMhd/GhValenciaDivClean/BinaryNeutronStar.yaml
```diff
@@ -311,55 +367,73 @@ EventsAndTriggersAtSlabs:
...
       - ObserveFields:
-          SubfileName: VolumeData
+          SubfileName: VolumeSpacetimeData
           VariablesToObserve:
             - SpacetimeMetric
-            - Lapse
-            - Shift
             - DetSpatialMetric
+            # - Pi
+            # - Phi
```
knelli2: Keep/comment or remove?

tests/InputFiles/GrMhd/GhValenciaDivClean/BinaryNeutronStar.yaml
```diff
@@ -287,6 +330,19 @@ EventsAndTriggersAtSlabs:
           - Name: PointwiseL2Norm(GaugeConstraint)
             NormType: L2Norm
             Components: Sum
+          - Name: PointwiseL2Norm(ThreeIndexConstraint)
```
knelli2, line 333: Want to also add the constraint energy? This trigger seems to now be run infrequently enough that it shouldn't affect performance

knelli2 (changes requested):

tests/Unit/Framework/MockDistributedObject.hpp
nilsdeppe: I tried this, and unfortunately there is not a sufficiently unique base class in Catch, and it would make us subject to catch internals :(

nilsdeppe: Oh, there's the `Catch::Matchers::MatcherBase<std::string>` which at least works for the one case I've tried. The restriction is just current further up in the call stack as a static_assert

src/ControlSystem/Actions/GridCenters.hpp
nilsdeppe: I'm going with an `ERROR` since this will run once in a simulation and therefore I'd rather make sure it's caught with more description :)  ALso checking for `Rotation`, just in case! I'd rather get a better backtrace!

tests/Unit/Framework/MockDistributedObject.hpp
knelli2: Do you want to try and use that, or just leave as-is?

nilsdeppe (commented):
Okay, pushed fixups and rebased onto develop. Thanks for the review!

tests/Unit/ControlSystem/Actions/Test_GridCenters.cpp
```diff
@@ -55,6 +55,15 @@ struct Metavariables {
   using component_list = tmpl::list<Component<Metavariables>>;
 };
 
+struct UpdateFot {
+  using FoT_t = std::unordered_map<
+      std::string, std::unique_ptr<domain::FunctionsOfTime::FunctionOfTime>>;
+
+  static void apply(gsl::not_null<FoT_t*> current_fots, const FoT_t& new_fots) {
```
knelli2: [optional] We have the `domain::FunctionsOfTimeMap` alias from `FunctionOfTime.hpp`

tests/Unit/ControlSystem/Actions/Test_GridCenters.cpp
```diff
@@ -153,11 +162,52 @@ void test_action() {
...
+              const domain::FunctionsOfTime::SettleToConstantQuaternion* const>(
+              get<domain::Tags::FunctionsOfTime>(
+                  ActionTesting::cache<component>(runner, id))
+                  .at("Rotation")
+                  .get()) == nullptr);
+  }
+
+  // Remove things we need from FunctionsOfTime map to test ERRORs
+  auto& cache = ActionTesting::cache<component>(runner, element_id_zero);
+  cache.template mutate<FunctionsOfTimeTag, UpdateFot>([&functions_of_time]() {
```
knelli2: Use `Parallel::mutate` here. It should be supported in the test

knelli2 (commented):
Just a couple small things that can be squashed in

nilsdeppe:
Rebased and squashed. Thanks for the review :D

knelli2 (approved):

PR #6551, opened 2025-04-02
Add a Restart phase and regenerate registration data on restart

### LINKED ISSUE #6485: New events are not registered on overlaying
We can now update the events and triggers on restart, but we don't update the registration information, so adding new events can fail to work.  We need to do something along the lines of rerunning the registration phase after doing an overlay.

Thoughts I have:

1. Perhaps have a special "Restart" phase that runs on each restart, after the overlaying if applicable.
2. Do we want to try to update the checkpointed registration info, or can we just completely regenerate it?  The latter might be simpler, but if anything actually has important state that needs to be preserved across restarts it won't work.  (We don't think writers and such ever have pending data on a restart.)
3. This all has to tie into the phase arbitration, but that's probably not any more complicated than the current overlay logic.

wthrowe:
Fixes #6485.

Marking this as a draft until I've run a BBH inspiral to make sure that still works, but otherwise this is ready for review.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


wthrowe:
I have successfully run a BBH inspiral, so I think this is ready.

kidder (approved):

nilsdeppe (approved):

PR #6552, opened 2025-04-02
PySphericalHarmonics: bind Strahlkorper in grid frame

nilsvu:
## Proposed changes

Will be useful to do interpolations in the grid frame

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
@@ -17,45 +17,40 @@
 #include "NumericalAlgorithms/SphericalHarmonics/IO/ReadSurfaceYlm.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/IO/StrahlkorperCoordsToTextFile.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
+#include "Utilities/GetOutput.hpp"
 
 namespace py = pybind11;
 
 namespace ylm::py_bindings {
-void bind_strahlkorper(pybind11::module& m) {  // NOLINT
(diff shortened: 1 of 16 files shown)


nilsvu:
@sxs-collaboration/spectre-core-devs can someone take this one?

src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
```diff
@@ -74,6 +69,12 @@ void bind_strahlkorper(pybind11::module& m) {  // NOLINT
         return std::make_pair(legend, data);
       },
       py::arg("strahlkorper"), py::arg("time"), py::arg("max_l"));
+}
+}  // namespace
+
+void bind_strahlkorper(py::module& m) {
+  bind_strahlkorper_impl<Frame::Inertial>(m);
+  bind_strahlkorper_impl<Frame::Grid>(m);
```
knelli2, line 77: While you're at it, just do `Distorted` and `NoFrame` as well. And also add it to the other instantiations.

knelli2 (changes requested):
You can squash

src/NumericalAlgorithms/SphericalHarmonics/Python/Strahlkorper.cpp
nilsvu: Are those useful for something?

knelli2: It just always goes 1) add inertial thing, 2) realize we need grid and add that, 3) realize we nee distorted and add that, so figured why not just add Distorted now. I guess `NoFrame` we don't really need; that's only used in specific scenarios

nilsvu: Ok added distorted

knelli2 (approved):

PR #6553, opened 2025-04-03
Update create_initial_element for a Domain with non-conforming Blocks

kidder:
## Proposed changes

New logic is needed to determine the neighbors of an Element on the boundary of a Block if the Blocks of a Domain are non-conforming (i.e. the block logical coordinates on the interface of neighboring blocks are not related by a discrete rotation).

For an S2 boundary topology next to multiple wedges of a cubed sphere (which have boundary topology I1 x I1) there is no one-to-one mapping between the logical axes on the interface.  (There is a one-to-one mapping for the logical axis normal to the interface.)  In such a case the OrientationMap will use `Direction<Dim>::self()` as the angular mapped direction.

I do not allow arbitrary non-conforming Blocks.  It is a requirement that neighboring Blocks be conforming if their oriented topologies are the same in the interface dimensions.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CreateInitialElement.cpp
+++ b/src/Domain/CreateInitialElement.cpp
@@ -25,9 +25,11 @@
 namespace domain::Initialization {
 template <size_t VolumeDim>
 Element<VolumeDim> create_initial_element(
-    const ElementId<VolumeDim>& element_id, const Block<VolumeDim>& block,
+    const ElementId<VolumeDim>& element_id,
+    const std::vector<Block<VolumeDim>>& blocks,
     const std::vector<std::array<size_t, VolumeDim>>&
         initial_refinement_levels) {
+  const auto& block = blocks[element_id.block_id()];
   const auto& neighbors_of_block = block.neighbors();
   const auto segment_ids = element_id.segment_ids();
 
@@ -187,10 +189,10 @@ Element<VolumeDim> create_initial_element(
 
 #define DIM(data) BOOST_PP_TUPLE_ELEM(0, data)
 
-#define INSTANTIATE(_, data)                                 \
-  template Element<DIM(data)>                                \
-  domain::Initialization::create_initial_element<DIM(data)>( \
-      const ElementId<DIM(data)>&, const Block<DIM(data)>&,  \
+#define INSTANTIATE(_, data)                                             \
+  template Element<DIM(data)>                                            \
+  domain::Initialization::create_initial_element<DIM(data)>(             \
+      const ElementId<DIM(data)>&, const std::vector<Block<DIM(data)>>&, \
       const std::vector<std::array<size_t, DIM(data)>>&);
 
 GENERATE_INSTANTIATIONS(INSTANTIATE, (1, 2, 3))
--- a/src/Domain/CreateInitialElement.cpp
+++ b/src/Domain/CreateInitialElement.cpp
@@ -12,6 +12,7 @@
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/ElementId.hpp"
+#include "Domain/Structure/NeighborIsConforming.hpp"
 #include "Domain/Structure/Neighbors.hpp"
 #include "Domain/Structure/OrientationMap.hpp"
 #include "Domain/Structure/SegmentId.hpp"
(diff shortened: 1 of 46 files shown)


src/Domain/CreateInitialElement.cpp
```diff
@@ -33,109 +111,70 @@ Element<VolumeDim> create_initial_element(
   const auto& neighbors_of_block = block.neighbors();
   const auto segment_ids = element_id.segment_ids();
 
-  // Declare two helper lambdas for setting the neighbors of an element
```
kidder, line 34: This lambda is completely rewritten and reviewers should ignore the interleaved diffs

src/Domain/CreateInitialElement.cpp
```diff
@@ -33,109 +111,70 @@ Element<VolumeDim> create_initial_element(
...
           }
-          neighbor_ids.insert(
-              {*block_neighbor.begin(), segment_ids_of_neighbor, grid_index});
-        next_index:;
         }
-        return std::make_pair(
-            direction,
-            Neighbors<VolumeDim>(std::move(neighbor_ids), orientation));
+
+        return std::make_pair(direction, std::move(element_neighbors));
```
kidder, line 177: This is the end of the rewritten lambda

src/Domain/Structure/OrientationMap.cpp
```diff
@@ -51,8 +52,12 @@ OrientationMap<VolumeDim>::OrientationMap(
       set_aligned(false);
     }
   }
-  ASSERT(set_of_dimensions().size() == VolumeDim,
-         "This OrientationMap fails to map Directions one-to-one.");
+  if (not alg::found(mapped_directions, Direction<VolumeDim>::self())) {
+    ASSERT(set_of_dimensions().size() == VolumeDim,
```
nilsvu: Maybe assert something like `set_of_dimensions().size() == (Dim - count(mapped_directions, Direction<VolumeDim>::self()))` (peusocode) (more below)

src/Domain/CreateInitialElement.cpp
```diff
@@ -12,6 +12,7 @@
 #include "Domain/Structure/Direction.hpp"
 #include "Domain/Structure/Element.hpp"
 #include "Domain/Structure/ElementId.hpp"
+#include "Domain/Structure/NeighborIsConforming.hpp"
```
nilsvu, line 15: typo in last commit message "create_initial_element"

nilsvu (commented):
I reviewed all but the last commit, so if you rebase and push those before I get the chance to review the last commit we can merge them right away.

src/Domain/Structure/NeighborIsConforming.hpp
```diff
@@ -0,0 +1,39 @@
...
+/// \note If the neighbors are conforming, they can exchange boundary data via
+/// copy or projection, taking into account the discrete rotation if necessary.
+/// If the neighbors are not conforming, boundary data will need to be
+/// interpolated.
+template <size_t VolumeDim>
+bool neighbor_is_conforming(
+    const std::array<Topology, VolumeDim>& self_topologies,
+    const std::array<Topology, VolumeDim>& neighbor_topologies,
+    const Direction<VolumeDim>& direction_to_neighbor,
+    const OrientationMap<VolumeDim>& orientation_of_neighbor);
```
nilsdeppe, line 38: This seems to ignore refinement. I think that should be stated explicitly

nilsdeppe (commented):
No need for me to approve before merging

src/Domain/Structure/NeighborIsConforming.hpp
kidder: is_conforming is with respect to the block, so refinement is irrelevant

nilsdeppe: The documentation literally says " (or Elements)"

kidder: I agree.  But it also says that it is the block logical coordinates (of the Block or Element) which define if the neighbors are conforming...   So please suggest what wording would clarify the documentation for you...

nilsdeppe: Either remove "(or Elements)" or add something like "\note This does not take into account h or p refinement between neighbor elements, so projection may still be necessary."

kidder: Okay after discussion, will update to 
```
/// \note If neighboring Elements are conforming, they can exchange boundary data via
/// either copy (if they have the same h- and p-refinement) or projection (if they
/// don't), taking into account the discrete rotation if necessary.
/// If the neighbors are not conforming, boundary data will need to be
/// interpolated.
```

nilsdeppe (approved):

PR #6560, opened 2025-04-05
Use topology to create initial mesh

kidder:
## Proposed changes

The Topology is used to choose the Basis and Quadrature of the initial Mesh

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Structure/CreateInitialMesh.hpp
+++ b/src/Domain/Structure/CreateInitialMesh.hpp
@@ -10,11 +10,13 @@
 
 /// \cond
 template <size_t Dim>
+struct Block;
+template <size_t Dim>
+struct Element;
+template <size_t Dim>
 struct ElementId;
 template <size_t Dim>
 class Mesh;
-template <size_t Dim>
-struct OrientationMap;
 namespace Spectral {
 enum class Quadrature : uint8_t;
 }  // namespace Spectral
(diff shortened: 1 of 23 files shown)


src/Domain/Structure/Topology.hpp
```diff
@@ -46,4 +50,24 @@ enum class Topology : uint8_t {
...
 /// Output operator for a Topology.
 std::ostream& operator<<(std::ostream& os, Topology topology);
+
+namespace topologies {
+template <size_t VolumeDim>
+static constexpr auto hypercube = make_array<VolumeDim>(Topology::I1);
+
+static constexpr auto annulus = std::array{Topology::I1, Topology::S1};
+
+static constexpr auto circle =
```
nilsvu: I think this is more often called "disk" (like our 2d domain creator)

src/Domain/Structure/CreateInitialMesh.hpp
```diff
@@ -24,19 +26,29 @@ namespace domain::Initialization {
...
-/// of the neighbor.
+/// \param initial_extents initial extents for Elements in each Block of the
+///        Domain
+/// \param element Element
+/// \param legendre_quadrature the quadrature rule/grid point distribution for
+///        dimensions that use Spectral::Basis::Legendre
+template <size_t Dim>
+Mesh<Dim> create_initial_mesh(
+    const std::vector<std::array<size_t, Dim>>& initial_extents,
+    const Element<Dim>& element, Spectral::Quadrature legendre_quadrature);
```
nilsvu, line 37: Just one overload that takes the topologies array is enough I think? Plus initial extents, element id, quadrature as before. Then also you don't need to construct full elements/blocks for testing.

src/Domain/Structure/Topology.hpp
kidder: okay

src/Domain/Structure/CreateInitialMesh.hpp
kidder: I considered this, but am worried that some user will pass in the block topologies instead of the element topologies which will be incorrect once refining B2Radial is supported

nilsvu: Ok

nilsvu (approved):

PR #6561, opened 2025-04-07
Add check for negative expansion boundary condition.

iago-mendes:
## Proposed changes

Check if the expansion at the excisions is negative. We found that this boundary condition is currently violated in two test initial data runs: high mass ratio ($q = 10$) and high spin ($\chi = 0.9999$).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
+++ b/src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
@@ -253,6 +253,9 @@ void apparent_horizon_impl(
         make_not_null(&expansion_of_solution), make_not_null(&beta_orthogonal),
         *solution_for_negative_expansion, x, face_normal, face_normal_magnitude,
         deriv_unnormalized_face_normal);
+    ASSERT(max(get(expansion_of_solution)) <
+               100.0 * std::numeric_limits<double>::epsilon(),
+           "Expansion of solution is not negative everywhere.");
   }
 
   // Shift


src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
```diff
@@ -253,6 +253,13 @@ void apparent_horizon_impl(
         make_not_null(&expansion_of_solution), make_not_null(&beta_orthogonal),
         *solution_for_negative_expansion, x, face_normal, face_normal_magnitude,
         deriv_unnormalized_face_normal);
+#ifdef SPECTRE_DEBUG
+    const auto& expansion = get(expansion_of_solution);
+    const bool negative_expansion = std::all_of(
+        expansion.begin(), expansion.end(), [](double t) { return t < 0.0; });
```
nilsvu:
```cpp
ASSERT(blaze::min(get(expansion_of_solution)) < 0.0, "...");
```

nilsvu (changes requested):

src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
iago-mendes: I've used `blaze::min` now. For some reason, this change made the assert fail only for the `q10` case but not for the `Spin0.9999` case anymore. It also fixed `Test_ApparentHorizon`, which was previously failing. This is confusing because the two approaches seem equivalent.

By the way, `Spin0.9999` now violates some `roots.has_value()` assert. It's possible that `Spin0.9999` can't find horizons for some reason other than the negative expasion BC not working. It's just strange because `Spin0.9999` fails in SpEC due to negative expansion BC (while `q10` has no issues in SpEC ID).

nilsvu: Sorry this was my mistake, this assert should be `max(...) < 0.0` because you want _all_ points to have negative expansion.

iago-mendes: Ups, I should've given it some thought before copying and pasting. Sorry.

Great, so now both `q10` and `Spin0.9999` trigger this assert failure.

Now, `Test_ApparentHorizon` is again triggering this assert failure. I tried to investigate why this was happening, but didn't make much progress. The spin is not too big (`\chi = 0.8`), so that shouldn't be the issue. Perhaps it's a bug in the BC?

nilsvu: You can delete `#ifdef SPECTRE_DEBUG` now (asserts only run in debug mode)

nilsvu: The test failure is just evaluating the AH boundary conditions _at_ the horizon, where the expansion is above zero by roundoff error. Try `max(...) < 100.0 * std::numeric_limits<double>::epsilon()` instead of exactly zero.

src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
```diff
@@ -253,6 +253,10 @@ void apparent_horizon_impl(
         make_not_null(&expansion_of_solution), make_not_null(&beta_orthogonal),
         *solution_for_negative_expansion, x, face_normal, face_normal_magnitude,
         deriv_unnormalized_face_normal);
+#ifdef SPECTRE_DEBUG
+    ASSERT(blaze::max(get(expansion_of_solution)) < 0.0,
+           "Expansion of solution is not negative everywhere.");
+#endif
```
nilsdeppe: You don't need the ifdef.

I don't think you need the namespace `blaze::`, just `max` should work. ADL should take care of that.

src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
iago-mendes: That worked. Thank you!

nilsvu (approved):
Thanks @iago-mendes !

PR #6550, opened 2025-04-02
Communication of coupling terms in MC

ffoucart:
## Proposed changes

Add capability for the Monte Carlo code to gather the contribution of the MC packets to the evolution of the fluid, and communicate interactions happening in ghost zones back to the live points evolving the fluid.
This PR does NOT use the communicated data yet (we do not have a fully coupled MC+fluid executable yet).

Test_CommunicationTags include tests of this infrastructure.

### Upgrade instructions

NA

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #6564, opened 2025-04-07
Ghost zone comm mc sum fluid coupling

ffoucart:
## Proposed changes

Add a mutator responsible for summing the contributions to the coupling between fluid and Monte-Carlo gathered from ghost zones to the values stored locally from the evolution of MC packets in live zones.

### Upgrade instructions

NA

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

Depends on #6550

### DIFF
--- a/src/Evolution/Particles/MonteCarlo/GhostZoneCommunication.hpp
+++ b/src/Evolution/Particles/MonteCarlo/GhostZoneCommunication.hpp
@@ -197,6 +197,103 @@ struct GhostDataMcPackets {
   }
 };
 
+// Take the data for coupling MC to hydro gathered from neighboring
+// ghost zones (stored after communication in GhostZoneCouplingDataTag)
+// and add them to the coupling terms in live zones. After calling this
+// action, the coupling terms CouplingTildeTau, CouplingTildeRhoYe,
+// CouplingTildeS should contain the changes to the energy, RhoYe, and
+// momentum of the fluid due to MC packets evolved locally and those
+// that evolved within ghost zones on neighboring elements.
+// THIS FUNCTION IS CURRENTLY ONLY IMPLENTED IN 3D!
+// As other parts of the MC communication, it is also incompatible with
+// mesh refinement.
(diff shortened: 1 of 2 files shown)


ffoucart:
The first commit of this PR is just #6550 . Only the second commit is relevant. There is a minor change to the test that checks work from 6550, but as the test didn't catch any error, i just kept it in this PR.

src/Evolution/Particles/MonteCarlo/GhostZoneCommunication.hpp
```diff
@@ -197,6 +197,94 @@ struct GhostDataMcPackets {
...
+            for (size_t k = 0; k < ghost_zone_extents[2]; ++k) {
+              Index<Dim> index_3d{i, j, k};
+              // Index of point in ghost zone data
+              const size_t ghost_index =
+                  collapsed_index(index_3d, ghost_zone_extents);
+              // Index of corresponding live point in full
+              // mesh, with ghost zones.
+              index_3d[dimension] =
+                  (side == Side::Lower)
+                      ? index_3d[dimension] + num_ghost_zones
```
kidder, line 272: why isn't this just `index_3d[dimension]`?

ffoucart: I added a much longer comment. I assume that the difference with what you imagine is that we are copying from the ghost zones of the neighbor onto the live points of the current mesh (rather than the other way around, usually), and all meshes used here include ghost points in their indexing.

ffoucart:
Updated after rebasing on merged PR #6550 and adding a comment to address Larry's question.

src/Evolution/Particles/MonteCarlo/GhostZoneCommunication.hpp
kidder: ah, thanks, the comment helps a lot

kidder (approved):

PR #6565, opened 2025-04-07
Add coefficient calculations for variable-order Adams LTS

wthrowe:
Support in the time steppers themselves will be in a subsequent PR.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/TimeSteppers/AdamsBashforth.cpp
+++ b/src/Time/TimeSteppers/AdamsBashforth.cpp
@@ -251,8 +251,7 @@ void AdamsBashforth::add_boundary_delta_impl(
   const auto step_start = local_times.back().step_time();
   const auto lts_coefficients = adams_lts::lts_coefficients(
       local_times, remote_times, step_start, step_start + time_step,
-      adams_lts::SchemeType::Explicit, adams_lts::SchemeType::Explicit,
-      adams_lts::SchemeType::Explicit, 0, 0);
+      adams_lts::SchemeType::Explicit, adams_lts::SchemeType::Explicit, 0, 0);
   adams_lts::apply_coefficients(result, lts_coefficients, coupling);
 }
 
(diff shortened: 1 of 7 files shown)


wthrowe:
I don't understand the clang-tidy errors.  The operators in the highlighted expressions have no negated operands, and the check in question is documented to only flag expressions with negated operands (unless the "Relaxed" version is explicitly enabled)

nilsdeppe (approved):
A few general questions not related to the code, but adaptive order (AO) in general:
1. Is there any overhead from the code changes to the fixed-order versions?
2. I'm assuming there is some overhead when AO is enabled. Do you know where that overhead comes from? I ask because if it's from deciding what order to use, it could be useful to only check if we should switch the order every N time steps. For DG-FD I implemented a change so we only check every N time steps if we should go from FD back to DG and that helped a lot in terms of cutting down overhead.


Edit: I have mixed feelings on these clang-tidy warnings. I sometimes find the "simpler" version harder to read because it's not the algorithmic way of thinking about it. I'm happy to ignore and merge as-is. @wthrowe Please make a decision on what you prefer, if that decision is "as-is" please go ahead and merge this :)
Similarly, no need to wait to merge for a reply to my questions since they won't impact any code in this PR :)

wthrowe:
1. I have not benchmarked, but there should be essentially no overhead for fixed-order evolutions.  A few more cases of integer math and pulling the orders from the histories, but that had really better be negligible.

2. I haven't redone the runs yet since optimizing, but there was noticeable overhead for turning on adaptivity, and I expect at least some of that to remain.  I believe the most expensive part is calculating the lower-order error estimates (future PR), each of which is comparable in cost to the main update of the variables (not the RHS evaluation, just multiplying history terms by coefficients and summing).  I suspect there is significant improvement to be had there from prefetching and inverting some loops.

    Skipping the error estimate on steps that we don't need it (and then only checking every few steps) is a good idea.  The TimeStepper interface already allows this.  We'd just have to figure out the logic in the actions.  I'd thought about that a little for the slab-size adjustment in the past, since the benefit is bigger for some RK time steppers, but it's hard there because it ties into the events and triggers.  Doing it for LTS would be easier.

I think I finally understand what's going on with the clang-tidy warning.  The warning has about two pages of documentation, but the part on the De-Morgan transformations is some combination of nonexistant and wrong.  I'm going to leave them as-is, and suggest we consider disabling `SimplifyDeMorgan` at some point in the future.

nilsdeppe:
Thanks for the detailed explanation!!

Yea, running the adaptivity only every few steps could be a huge win. For DG-FD TCI things that helps a lot because the checks are actually somewhat expensive on the FD grid

PR #6567, opened 2025-04-08
Exporter: add SpacetimeInterpolator

nilsvu:
## Proposed changes

Adds the `SpacetimeInterpolator` class that can load multiple slices of volume data and then interpolate to arbitrary points in both space and time within this time range. The time range can be advanced sequentially. This is useful for ray tracing.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/IO/Exporter/SpacetimeInterpolator.cpp
+++ b/src/IO/Exporter/SpacetimeInterpolator.cpp
@@ -1,0 +1,293 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "IO/Exporter/SpacetimeInterpolator.hpp"
+
+#include <array>
+#include <cstddef>
+#include <mutex>
+#include <optional>
+#include <string>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Domain/BlockLogicalCoordinates.hpp"
+#include "IO/Exporter/PointwiseInterpolator.hpp"
+#include "IO/Exporter/SelectObservation.hpp"
+#include "IO/H5/File.hpp"
+#include "IO/H5/VolumeData.hpp"
+#include "NumericalAlgorithms/Interpolation/PolynomialInterpolation.hpp"
+#include "Utilities/ErrorHandling/Assert.hpp"
+#include "Utilities/FileSystem.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Gsl.hpp"
+#include "Utilities/Overloader.hpp"
+
+namespace spectre::Exporter {
+
+namespace {
+
+auto load_observation_ids(const std::variant<std::vector<std::string>,
+                                             std::string>& volume_files_or_glob,
+                          const std::string& subfile_name) {
+  const std::vector<std::string> filenames =
+      std::visit(Overloader{[](const std::vector<std::string>& volume_files) {
+                              return volume_files;
+                            },
+                            [](const std::string& volume_files_glob) {
+                              return file_system::glob(volume_files_glob);
+                            }},
+                 volume_files_or_glob);
+  const h5::H5File<h5::AccessType::ReadOnly> first_h5file(filenames.front());
+  const auto& first_volfile = first_h5file.get<h5::VolumeData>(subfile_name);
+  const auto all_observation_ids = first_volfile.list_observation_ids();
+  std::vector<double> all_observation_values(all_observation_ids.size());
+  std::transform(all_observation_ids.begin(), all_observation_ids.end(),
+                 all_observation_values.begin(),
+                 [&first_volfile](const size_t obs_id) {
+                   return first_volfile.get_observation_value(obs_id);
+                 });
+  first_h5file.close();
+  return std::make_pair(all_observation_ids, all_observation_values);
+}
+
+}  // namespace
+
+template <size_t Dim, typename Frame>
+SpacetimeInterpolator<Dim, Frame>::SpacetimeInterpolator(
+    std::variant<std::vector<std::string>, std::string> volume_files_or_glob,
+    std::string subfile_name, std::vector<std::string> tensor_components)
+    : volume_files_or_glob_(std::move(volume_files_or_glob)),
+      subfile_name_(std::move(subfile_name)),
+      tensor_components_(std::move(tensor_components)) {
+  std::tie(all_observation_ids_, all_observation_values_) =
+      load_observation_ids(volume_files_or_glob_, subfile_name_);
+  if (all_observation_ids_.empty()) {
+    ERROR("No observation IDs found in the volume data files.");
+  }
+  if (all_observation_ids_.size() < time_interpolation_order_ + 1) {
+    ERROR(
+        "The number of available observations must be at least the time "
+        "interpolation order plus one (num observations >= "
+        << time_interpolation_order_ + 1 << "), but "
+        << all_observation_ids_.size() << " observations were found.");
+  }
+}
+
+template <size_t Dim, typename Frame>
+std::array<double, 2> SpacetimeInterpolator<Dim, Frame>::max_time_bounds()
+    const {
+  return {*(all_observation_values_.begin() + num_ghost_slices_),
+          *(all_observation_values_.end() - num_ghost_slices_ - 1)};
+}
+
+template <size_t Dim, typename Frame>
+void SpacetimeInterpolator<Dim, Frame>::load_time_bounds(
+    std::array<double, 2> time_bounds) {
+  ASSERT(time_bounds[0] < time_bounds[1],
+         "The lower time bound must be smaller than the upper time bound.");
+  ASSERT(time_bounds[0] >= max_time_bounds()[0] and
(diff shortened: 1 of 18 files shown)


src/IO/Exporter/SpacetimeInterpolator.hpp
```diff
@@ -0,0 +1,134 @@
...
+ * This class reads the volume data at multiple time steps into memory and can
+ * then interpolate it to any number of target points at any time within the
+ * time bounds of the data. The time bounds can be advanced forward or backward
+ * in time.
+ *
+ * \details The interpolation is first done in space and then in time. For the
+ * interpolation in space, the PointwiseInterpolator class is used, which does
+ * spectral interpolation within the elements. For the interpolation in time, a
+ * polynomial interpolation with fixed order is used (see
+ * `time_interpolation_order_`).
```
wthrowe: Documentation shouldn't require people to look at private member variables to find out what's going on.

src/IO/Exporter/SpacetimeInterpolator.hpp
```diff
@@ -0,0 +1,134 @@
...
+      std::string subfile_name, const ObservationVariant& initial_observation,
+      std::vector<std::string> tensor_components, bool forward_in_time,
+      size_t num_time_slices);
+
+  bool forward_in_time() const { return forward_in_time_; }
+
+  /// Advance the time bounds by loading new volume data slices. The new time
+  /// bounds will exactly follow the old time bounds (no overlap in time). This
+  /// function is not generally thread safe because it loads new data into
+  /// memory.
```
wthrowe: Is it unsafe to call this in parallel with other member functions, or just to call this function from multiple threads?

src/IO/Exporter/SpacetimeInterpolator.hpp
```diff
@@ -0,0 +1,134 @@
...
+  SpacetimeInterpolator(
+      std::variant<std::vector<std::string>, std::string> volume_files_or_glob,
+      std::string subfile_name, const ObservationVariant& initial_observation,
+      std::vector<std::string> tensor_components, bool forward_in_time,
+      size_t num_time_slices);
+
+  bool forward_in_time() const { return forward_in_time_; }
+
+  /// Advance the time bounds by loading new volume data slices. The new time
+  /// bounds will exactly follow the old time bounds (no overlap in time). This
```
wthrowe: This seems really bad from a performance point of view.  You have to block the entire evolution on I/O.  And, depending on what integrator you're using, the evaluations may not even be monotonic in time, in which case this won't work at all.

tests/Unit/IO/Exporter/Test_SpacetimeInterpolator.cpp
```diff
@@ -0,0 +1,174 @@
...
+}
+
+void test_time_interpolation(
+    const SpacetimeInterpolator<3, Frame::Inertial>& interpolator) {
+  auto [tmin, tmax] = interpolator.time_bounds();
+  if (interpolator.forward_in_time()) {
+    CHECK(tmin < tmax);
+  } else {
+    CHECK(tmin > tmax);
+    std::tie(tmin, tmax) = std::pair{tmax, tmin};
```
wthrowe: `swap` is probably clearer.

src/IO/Exporter/SpacetimeInterpolator.cpp
```diff
@@ -0,0 +1,206 @@
...
+  }
+  const auto obs_steps_bounds =
+      forward_in_time_ ? std::pair{static_cast<size_t>(current_obs_step_),
+                                   static_cast<size_t>(current_obs_step_) +
+                                       num_new_slices - 1}
+                       : std::pair{static_cast<size_t>(current_obs_step_) -
+                                       num_new_slices + 1,
+                                   static_cast<size_t>(current_obs_step_)};
+  if (obs_steps_bounds.first < 0 or obs_steps_bounds.second >= num_obs_steps_) {
+    ERROR("The requested time steps "
```
wthrowe: This should handle reaching the end of the data more gracefully.  Instead of erroring, it should decrease the end time to the maximum allowable value.

wthrowe (changes requested):

src/IO/Exporter/SpacetimeInterpolator.hpp
nilsvu: This function is removed now. I added more details on thread safety to `load_time_bounds`.

src/IO/Exporter/SpacetimeInterpolator.hpp
nilsvu: I removed this function and added `load_time_bounds` instead. That allows for a sliding time window, and concurrently loading new data. I have tested this in a ray tracing simulation where I slide the window backwards in time while rendering multiple frames from a camera.

src/IO/Exporter/SpacetimeInterpolator.cpp
nilsvu: Handled in `load_time_bounds()` now. Only time bounds within the range of available data is allowed.

nilsvu:
@wthrowe I updated this PR. I replaced `advance_time` with the more flexible `load_time_bounds`. That allows for a sliding time window and concurrent loading of new data.

src/IO/Exporter/SpacetimeInterpolator.cpp
```diff
@@ -43,88 +43,118 @@ std::pair<int, size_t> get_observation_step(
...
-    } else {
-      --current_obs_step_;
+
+  // Load the new data. This can happen while other threads are still accessing
+  // the existing data.
+  std::vector<std::pair<size_t, PointwiseInterpolator<Dim, ::Frame::Grid>>>
+      new_interpolators{};
+  for (size_t i = 0; i < num_obs; ++i) {
+    if (i < interpolators_.size() and
+        interpolators_[i].obs_id() == obs_ids[i]) {
```
wthrowe: The indices in `interpolators_` and `obs_ids` don't necessarily correspond, so this can cause you to add copies of interpolators you already have.  I added a statement at the end of the function to print `interpolators_.size()` and then had the test loop over
```c++
interpolator.load_time_bounds({{1.5, 6.5}});
interpolator.load_time_bounds({{2.5, 6.5}});
```
and found that the number of interpolators grows by 8 every cycle.

src/IO/Exporter/SpacetimeInterpolator.cpp
```diff
@@ -43,88 +43,118 @@ std::pair<int, size_t> get_observation_step(
...
+  {
+    // Lock while we are modifying the vector
+    const std::unique_lock lock(interpolators_mutex_);
+    for (auto it = interpolators_.begin(); it != interpolators_.end(); ++it) {
+      if (std::find(obs_ids.begin(), obs_ids.end(), it->obs_id()) ==
+          obs_ids.end()) {
+        it = interpolators_.erase(it);
+        --it;
+      }
     }
```
wthrowe: This can be done much more efficiently since you know that both `interpolators_` and `obs_ids` are contiguous sequences of observations.  Both `find` and `erase` are O(N) operations being done in a loop.  Normally I wouldn't worry about this too much, but here this is done while holding an exclusive lock.  What you want to do is figure out what the range that needs to be deleted off each end of the vector is, and then do range-`erase` calls, which is O(N) total.  (Figuring out how much to delete can even be done without holding the lock.)

A similar thing holds for the other exclusive section, but that has to be rewritten anyway because of the bug causing multiple copies of the same data to be held.

wthrowe (changes requested):

src/IO/Exporter/SpacetimeInterpolator.cpp
nilsvu: Thanks for catching this bug! I rewrote the function to use iterators, limits locking time further, and fixes this bug.

src/IO/Exporter/SpacetimeInterpolator.cpp
```diff
@@ -93,67 +93,109 @@ void SpacetimeInterpolator<Dim, Frame>::load_time_bounds(
...
     const std::unique_lock lock(interpolators_mutex_);
-    for (auto it = interpolators_.begin(); it != interpolators_.end(); ++it) {
-      if (std::find(obs_ids.begin(), obs_ids.end(), it->obs_id()) ==
-          obs_ids.end()) {
-        it = interpolators_.erase(it);
-        --it;
-      }
+    if (num_erase_front > 0) {
+      interpolators_.erase(interpolators_.begin(),
+                           interpolators_.begin() + num_erase_front);
```
wthrowe, line 165: This can go past `end` if the new and old valid ranges are disjoint and far enough apart.  I think an easy fix is to first check something like
```c++
if (num_erase_front >= interpolators_.size() or num_erase_back >= interpoaltors_.size()) {
  interpolators_.clear();
  // copy the "interpolators_.empty()" block from above
}
```

wthrowe (changes requested):
This looks much better.  One last bug commented below, but should be an easy fix.  You can squash, including the new fix as long as it's not any more complicated than it looks from my comment.

nilsvu:
Done and squashed. Thanks for reviewing @wthrowe !

wthrowe (approved):

PR #6573, opened 2025-04-12
Add domain creator for S2 spherical shells

kidder:
## Proposed changes

- Miscellaneous fixes to enable evolutions on a domain of S2 spherical shells
- Domain creator to create the domain of S2 spherical shells

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/SphericalShells.cpp
+++ b/src/Domain/Creators/SphericalShells.cpp
@@ -1,0 +1,276 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/Creators/SphericalShells.hpp"
+
+#include <array>
+#include <cstddef>
+#include <memory>
+#include <optional>
+#include <unordered_map>
+#include <utility>
+#include <variant>
+#include <vector>
+
+#include "Domain/Block.hpp"
+#include "Domain/BoundaryConditions/BoundaryCondition.hpp"
+#include "Domain/BoundaryConditions/None.hpp"
+#include "Domain/BoundaryConditions/Periodic.hpp"
+#include "Domain/CoordinateMaps/CoordinateMap.hpp"
+#include "Domain/CoordinateMaps/CoordinateMap.tpp"
+#include "Domain/CoordinateMaps/Identity.hpp"
+#include "Domain/CoordinateMaps/Interval.hpp"
+#include "Domain/CoordinateMaps/ProductMaps.hpp"
+#include "Domain/CoordinateMaps/ProductMaps.tpp"
+#include "Domain/CoordinateMaps/SphericalToCartesianPfaffian.hpp"
+#include "Domain/Creators/DomainCreator.hpp"
+#include "Domain/Creators/ShellDistribution.hpp"
+#include "Domain/Creators/TimeDependence/None.hpp"
+#include "Domain/Domain.hpp"
+#include "Domain/Structure/BlockNeighbors.hpp"
+#include "Domain/Structure/Direction.hpp"
+#include "Domain/Structure/DirectionMap.hpp"
+#include "Domain/Structure/Topology.hpp"
+#include "Options/Context.hpp"
+#include "Options/ParseError.hpp"
+
+namespace domain::creators {
+
+SphericalShells::SphericalShells(
+    const double inner_radius, const double outer_radius,
+    const size_t initial_radial_refinement,
+    const size_t initial_number_of_radial_grid_points,
+    const size_t initial_spherical_harmonic_l,
+    std::vector<double> radial_partitioning,
+    const typename RadialDistribution::type& radial_distribution,
+    std::optional<TimeDepOptionType> time_dependent_options,
+    std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
+        inner_boundary_condition,
+    std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>
+        outer_boundary_condition,
+    const Options::Context& context)
+    : inner_radius_(inner_radius),
+      outer_radius_(outer_radius),
+      initial_radial_refinement_(initial_radial_refinement),
+      initial_number_of_radial_grid_points_(
+          initial_number_of_radial_grid_points),
+      initial_spherical_harmonic_l_(initial_spherical_harmonic_l),
+      radial_partitioning_(std::move(radial_partitioning)),
+      time_dependent_options_(std::move(time_dependent_options)),
+      inner_boundary_condition_(std::move(inner_boundary_condition)),
+      outer_boundary_condition_(std::move(outer_boundary_condition)),
+      grid_anchors_{{{"Center", tnsr::I<double, 3, Frame::Grid>{
+                                    std::array{0.0, 0.0, 0.0}}}}} {
+  if (inner_radius_ > outer_radius_) {
+    PARSE_ERROR(context,
+                "Inner radius must be smaller than outer radius, but inner "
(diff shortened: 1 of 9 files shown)


src/Domain/Creators/SphericalShells.hpp
```diff
@@ -0,0 +1,260 @@
...
+/// to the domain. Currently, the first (inner-most) shell has maps between
+/// `Frame::Grid`, `Frame::Distorted`, and `Frame::Inertial` while all
+/// subsequent shells only have maps between `Frame::Grid` and
+/// `Frame::Inertial`.
+///
+/// ##### None
+/// To not have any time dependent maps, pass a `std::nullopt` as the
+/// appropriate argument in the constructor. In the input file, simply have
+/// `TimeDependentMaps: None`.
+class SphericalShells : public DomainCreator<3> {
```
knelli2, line 81: Drive by comment. I don't think I'll hold up this PR over this, but I strongly recommend adding spherical shells to the current Sphere domain, rather than creating a new domain creator. This is twofold

1. Maintainability. There are a lot of features in the Sphere domain that we'd have to keep feature parity with this new creator.
2. Interoperability. Long term, I'd like to mix and match spherical shells with cubed shells. Even if right now we only require one or the other for the entire domain. It'll be easier to extend if everything is in one domain creator.

Let me know if there were discussions offline about this design choice, as I could just be out of the loop.

tests/Unit/Helpers/Domain/DomainTestHelpers.cpp
```diff
@@ -33,6 +33,7 @@
 #include "Domain/Structure/Neighbors.hpp"
```
nilsdeppe, line 33: I'm worried that for domains people will call these functions but no tests will happen. They'll then deduce "All tests pass" and move on. Can we add `ERROR`s to guard against that?

src/Domain/Creators/TimeDependentOptions/Sphere.hpp
```diff
@@ -162,7 +162,8 @@ struct TimeDependentMapOptions {
    * blocks, this returns `nullptr`.
    */
   MapType<Frame::Distorted, Frame::Inertial> distorted_to_inertial_map(
-      size_t block_number, bool is_inner_cube) const;
+      size_t block_number, bool is_inner_cube,
+      size_t num_blocks_per_shell = 6) const;
```
nilsdeppe: Why default this? Seems risky and asking for bugs...

src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -110,23 +110,27 @@ block_logical_coordinates_single_point(
   }
 
   for (size_t d = 0; d < Dim; ++d) {
```
nilsdeppe: factor into separate commit. It looks like they might not be easy to test without the shells, though. I'm okay with that.

src/Domain/Creators/SphericalShells.hpp
nilsdeppe: I agree with Kyle. What is your reasoning for making it separate? Or maybe it's easier to just discuss after the spectre call on Monday. I've reviewed the rest of the PR, so if you want to split and get that merged first, I'm fine with that.

nilsdeppe (changes requested):

tests/Unit/Helpers/Domain/DomainTestHelpers.cpp
kidder: no; this just skips a subset of the tests that use corner iterators which make no sense for domains with no corners.  I will rewrite these test functions not to use corner iterators before I put in additional domains, but not now...

src/Domain/Creators/SphericalShells.hpp
kidder: - For just S2 spherical shells, this domain creator is simple and doesn't have to deal with multiple options that make no sense for such a domain.
- I will modify the existing domain when I get non-conforming shells working so that it can have an outer layer of spherical shells.

src/Domain/Creators/SphericalShells.hpp
```diff
@@ -0,0 +1,260 @@
...
+}  // namespace CoordinateMaps
+
+template <typename SourceFrame, typename TargetFrame, typename... Maps>
+class CoordinateMap;
+}  // namespace domain
+/// \endcond
+
+namespace domain::creators {
+/// \brief A set of concentric spherical shells
+
```
nilsdeppe: do you need a `///` here to note break dox?

src/Domain/Creators/SphericalShells.hpp
```diff
@@ -0,0 +1,260 @@
...
+  std::vector<DirectionMap<
+      3, std::unique_ptr<domain::BoundaryConditions::BoundaryCondition>>>
+  external_boundary_conditions() const override;
+
+  std::vector<std::array<size_t, 3>> initial_extents() const override;
+
+  std::vector<std::array<size_t, 3>> initial_refinement_levels() const override;
+
+  std::vector<std::string> block_names() const override;
```
nilsdeppe, line 232: Can you list what the different block names and groups are in the main documentation? I've found it annoying not knowing what they are and the dox not helping :(

src/Domain/Creators/SphericalShells.cpp
```diff
@@ -0,0 +1,275 @@
...
+#include "Domain/Structure/Topology.hpp"
+#include "Options/Context.hpp"
+#include "Options/ParseError.hpp"
+
+namespace domain::creators {
+
+SphericalShells::SphericalShells(
+    double inner_radius, double outer_radius, size_t initial_radial_refinement,
+    size_t initial_number_of_radial_grid_points,
+    size_t initial_spherical_harmonic_l,
```
nilsdeppe: `const`

tests/Unit/Domain/Creators/Test_SphericalShells.cpp
```diff
@@ -0,0 +1,606 @@
...
+  for (const auto& item : t) {
+    if (not first) {
+      result += ", ";
+    }
+    result += get_output(item);
+    first = false;
+  }
+  result += "]";
+  return result;
+}
```
nilsdeppe: How does this differ from `get_output(t)`? Is it that it's brackets instead of parens? If so, why not `get_output(t).replace(0, 1, "[")` and then `str.replace(str.length() -1, 1, "]")`?

src/Domain/Creators/SphericalShells.hpp
```diff
@@ -0,0 +1,260 @@
...
+    static constexpr Options::String help = {
+        "Initial number of radial grid points."};
+  };
+
+  struct InitialSphericalHarmonicL {
+    using type = size_t;
+    static constexpr Options::String help = {
+        "Initial spherical harmonic resolution specified as the highest "
+        "spherical harmonic represented on the grid."};
+  };
```
nilsdeppe: I think we should have a lower bound of 6 or 8 on this. I've never ran a stable simulation with an ell_max lower than that, even for spherically symmetric scalar wave data

nilsdeppe (changes requested):

src/Domain/Creators/SphericalShells.hpp
kidder: apparently not, but I'll add it anyway

nilsdeppe:
please squash

nilsdeppe (approved):

PR #6575, opened 2025-04-13
Support some empty volume files in GenerateXdmf

nilsdeppe:
Since we can now observe only on some blocks, not all volume files are guaranteed to have all blocks. Previously this caused an error. I've needed this for BNS visualization where the hydro data is only dumped at the stars.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Visualization/Python/GenerateXdmf.py
+++ b/src/Visualization/Python/GenerateXdmf.py
@@ -2,7 +2,6 @@
 
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
-
 """Tools to generate XDMF files that ParaView and VisIt can read.
 
 The XDMF file format is documented here:
@@ -13,7 +12,7 @@ import logging
 import os
 import sys
 import xml.etree.ElementTree as ET
-from typing import Optional
+from typing import List, Optional, Sequence, Tuple
 
 import click
 import h5py
@@ -281,6 +280,26 @@ def _xmf_grid(
     return xmf_grid
 
 
+def get_files_with_subfile(
+    h5file_names: Sequence[str], subfile_name: str
+) -> List[Tuple[h5py.File, str]]:
+    """Get the h5files and their name that contain a subfile with the name
+    subfile_name
+
+        \f
+    Arguments:
+      h5file_names: List of H5 file names of files to open and check if they
+        have the subfile.
+      subfile_name: The name of the subfile to check for.
+    """
+    result = list()
+    for filename in h5file_names:
+        h5file = h5py.File(filename, "r")
+        if subfile_name in h5file:
+            result.append((h5file, filename))
+    return result
+
+
 def generate_xdmf(
     h5files,
     output: str,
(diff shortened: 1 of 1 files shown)


kidder (dismissed):

src/Visualization/Python/GenerateXdmf.py
```diff
@@ -281,6 +280,26 @@ def _xmf_grid(
     return xmf_grid
 
 
+def get_files_with_subfile(
+    h5file_names: list[str], subfile_name: str
```
nilsvu: This `list[str]` syntax was added in Py 3.9, so this won't run with Py3.8 (currently our lowest supported version). Turns out our Py 3.8 test on CI is broken (probably my fault, I just fixed it in #6577). You can fix it like this:

```py
from typing import Sequence, List, Tuple

# ...
  h5file_names: Sequence[str]  # more general than List
  -> List[Tuple[...]]
```

src/Visualization/Python/GenerateXdmf.py
```diff
@@ -319,11 +338,12 @@ def generate_xdmf(
       use_tetrahedral_connectivity: Optional. Use "tetrahedral_connectivity".
         Default: False
     """
-    h5files = [(h5py.File(filename, "r"), filename) for filename in h5files]
+    h5file_names = h5files
 
     if not subfile_name:
         subfiles = available_subfiles(
-            (h5file for h5file, _ in h5files), extension=".vol"
+            (h5py.File(filename, "r") for filename in h5file_names),
```
nilsvu: You can just pass the list of filenames to `available_subfiles`:

```py
subfiles = available_subfiles(h5file_names, extension=".vol")
```

This is safer because I don't know if files are automatically closed if you open them in an iterator.

src/Visualization/Python/GenerateXdmf.py
```diff
@@ -342,6 +362,20 @@ def generate_xdmf(
...
 
+    h5files = get_files_with_subfile(h5file_names, subfile_name)
+
+    if len(h5files) == 0:
+        raise ValueError(
+            f"Could not open subfile name '{subfile_name}' in any h5 "
+            "files, {h5file_names}. Available subfiles: "
+            + str(
+                available_subfiles(
+                    (h5py.File(filename, "r") for filename in h5file_names),
```
nilsvu: same here

nilsvu (changes requested):

nilsdeppe:
Done! Thanks for the review!

src/Visualization/Python/GenerateXdmf.py
nilsvu: Still need to change `list` -> `List` etc so this works in Py 3.8.

nilsdeppe: Oh sorry, I misunderstood and didn't read carefully enough. Should be right now! Thanks!

nilsvu (approved):

src/Visualization/Python/GenerateXdmf.py
```diff
@@ -342,6 +362,20 @@ def generate_xdmf(
     if not subfile_name.endswith(".vol"):
         subfile_name += ".vol"
 
+    h5files = get_files_with_subfile(h5file_names, subfile_name)
+
+    if len(h5files) == 0:
+        raise ValueError(
+            f"Could not open subfile name '{subfile_name}' in any h5 "
+            "files, {h5file_names}. Available subfiles: "
```
nilsvu: Just noticed there's a `f"` prefix missing here

nilsvu (changes requested):

src/Visualization/Python/GenerateXdmf.py
nilsdeppe: 🙌 good catch! Should be fixed!

nilsvu (approved):

PR #6587, opened 2025-04-17
Fix macOS build on CI

nilsvu:
### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #6589, opened 2025-04-18
Add states DeltaRDriftInward, DeltaRNoDrift

markscheel:
These correspond to State 3 and State 4 in
SpEC.  These states can prevent the excision
boundary from getting too close to the horizon.

This involves:
 - Adding the two new states.
 - Adding new options DeltaRDriftInwardOptions
 - Passing in a new quantity into control_error.
 - Changing the three current states to allow transitions into the new states.
 - Adding new tests to Test_SizeControlStates that mirror the same tests in SpEC that test the logic of transitioning between states.
 - Updating the other tests to handle the new states.

### Upgrade instructions

Add this option to your input file to have the DetalRDriftInward state disabled by default.
```yaml
ControlSystems:
  Size:
    ...
    DeltaRDriftInwardOptions: None
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/ControlErrors/Size.hpp
+++ b/src/ControlSystem/ControlErrors/Size.hpp
@@ -235,9 +235,47 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
     double outward_drift_timescale{};
   };
 
-  using options = tmpl::list<MaxNumTimesForZeroCrossingPredictor,
-                             SmoothAvgTimescaleFraction, SmootherTuner,
-                             InitialState, DeltaRDriftOutwardOptions>;
+  struct DeltaRDriftInwardOptions {
+    using type =
+        Options::Auto<DeltaRDriftInwardOptions, Options::AutoLabel::None>;
+    static constexpr Options::String help{
+        "Options for State DeltaRDriftInward. Specify 'None' to disable State "
+        "DeltaRDriftInward."};
+    struct MinAllowedRadialDistance {
+      using type = double;
+      static constexpr Options::String help{
+          "Drift excision boundary inward if distance from horizon to "
+          "excision is less than this."};
+    };
+    struct MinAllowedCharSpeed {
+      using type = double;
+      static constexpr Options::String help{
+          "Drift excision boundary inward if min char speed is less than "
+          "this."};
+    };
+    struct InwardDriftVelocity {
+      using type = double;
+      static constexpr Options::String help{
+          "Maximum value of drift velocity term, if State DeltaRDriftInward is "
+          "triggered by MinAllowedRadialDistance or MinAllowedCharSpeed."};
+    };
+    using options = tmpl::list<MinAllowedRadialDistance, MinAllowedCharSpeed,
+                               InwardDriftVelocity>;
+    DeltaRDriftInwardOptions();
+    DeltaRDriftInwardOptions(double min_allowed_radial_distance_in,
+                             double min_allowed_char_speed_in,
+                             double inward_drift_velocity_in);
+    void pup(PUP::er& p);
+
+    double min_allowed_radial_distance{};
+    double min_allowed_char_speed{};
+    double inward_drift_velocity{};
+  };
+
+  using options =
+      tmpl::list<MaxNumTimesForZeroCrossingPredictor,
+                 SmoothAvgTimescaleFraction, SmootherTuner, InitialState,
+                 DeltaRDriftOutwardOptions, DeltaRDriftInwardOptions>;
   static constexpr Options::String help{
       "Computes the control error for size control. Will also write a "
       "diagnostics file if the control systems are allowed to write data to "
@@ -265,7 +303,8 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
   Size(const int max_times, const double smooth_avg_timescale_frac,
        TimescaleTuner<true> smoother_tuner,
        std::unique_ptr<size::State> initial_state,
-       std::optional<DeltaRDriftOutwardOptions> delta_r_drift_outward_options);
+       std::optional<DeltaRDriftOutwardOptions> delta_r_drift_outward_options,
+       std::optional<DeltaRDriftInwardOptions> delta_r_drift_inward_options);
 
   /// Returns the internal `control_system::size::Info::suggested_time_scale`. A
   /// std::nullopt means that no timescale is suggested.
@@ -456,6 +495,21 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
                                         delta_r_drift_outward_options_.value()
                                             .outward_drift_timescale)
             : std::nullopt;
+    const std::optional<double> inward_drift_velocity =
+        delta_r_drift_inward_options_.has_value()
+            ? std::optional<double>(
+                  delta_r_drift_inward_options_.value().inward_drift_velocity)
+            : std::nullopt;
+    const std::optional<double> min_allowed_radial_distance =
+        delta_r_drift_inward_options_.has_value()
+            ? std::optional<double>(delta_r_drift_inward_options_.value()
+                                        .min_allowed_radial_distance)
+            : std::nullopt;
+    const std::optional<double> min_allowed_char_speed =
+        delta_r_drift_inward_options_.has_value()
+            ? std::optional<double>(
+                  delta_r_drift_inward_options_.value().min_allowed_char_speed)
+            : std::nullopt;
 
     // Currently we don't do anything with the derivative of the comoving char
     // speed. Eventually, we will pass it to the computation of the control
@@ -489,15 +543,18 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
     const size::ErrorDiagnostics error_diagnostics = size::control_error(
         make_not_null(&info_), make_not_null(&char_speed_predictor_),
         make_not_null(&comoving_char_speed_predictor_),
-        make_not_null(&delta_radius_predictor_), time, control_error_delta_r,
-        control_error_delta_r_outward,
+        make_not_null(&delta_radius_predictor_),
+        make_not_null(&drift_limit_char_speed_predictor_),
+        make_not_null(&drift_limit_delta_radius_predictor_), time,
+        control_error_delta_r, control_error_delta_r_outward,
         delta_r_drift_outward_options_.has_value()
             ? std::optional<double>(delta_r_drift_outward_options_.value()
                                         .max_allowed_radial_distance)
             : std::nullopt,
-        dt_lambda_00, apparent_horizon, excision_surface, lapse,
-        shifty_quantity, spatial_metric_on_excision,
-        inverse_spatial_metric_on_excision);
+        inward_drift_velocity, min_allowed_radial_distance,
+        min_allowed_char_speed, horizon_00, dt_lambda_00, apparent_horizon,
+        excision_surface, lapse, shifty_quantity, spatial_metric_on_excision,
+        inverse_spatial_metric_on_excision, deriv_comoving_char_speed);
 
     state_history_.store(time, info_, error_diagnostics.control_error_args);
 
@@ -555,10 +612,13 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
   intrp::ZeroCrossingPredictor char_speed_predictor_{};
   intrp::ZeroCrossingPredictor comoving_char_speed_predictor_{};
   intrp::ZeroCrossingPredictor delta_radius_predictor_{};
+  intrp::ZeroCrossingPredictor drift_limit_char_speed_predictor_{};
+  intrp::ZeroCrossingPredictor drift_limit_delta_radius_predictor_{};
   size::StateHistory state_history_{};
   std::vector<std::string> legend_{};
   std::string subfile_name_{};
   std::optional<DeltaRDriftOutwardOptions> delta_r_drift_outward_options_{};
+  std::optional<DeltaRDriftInwardOptions> delta_r_drift_inward_options_{};
   db::compute_databox_type<tmpl::list<
       gr::Tags::InverseSpatialMetric<DataVector, 3, Frame::Distorted>,
       ylm::Tags::Strahlkorper<Frame::Distorted>,
--- a/src/ControlSystem/ControlErrors/Size/DeltaR.cpp
+++ b/src/ControlSystem/ControlErrors/Size/DeltaR.cpp
@@ -107,7 +107,8 @@ std::string DeltaR::update(const gsl::not_null<Info*> info,
   } else if (update_args.average_radial_distance.has_value() and
              update_args.average_radial_distance.value() >
                  non_oscillation_drift_outward_factor *
-                     update_args.max_allowed_radial_distance.value()) {
+                     update_args.max_allowed_radial_distance.value_or(
+                         std::numeric_limits<double>::infinity())) {
     info->discontinuous_change_has_occurred = true;
     info->state = std::make_unique<States::DeltaRDriftOutward>();
     ss << "Current state DeltaR. "
--- a/src/ControlSystem/ControlErrors/Size/DeltaR.cpp
+++ b/src/ControlSystem/ControlErrors/Size/DeltaR.cpp
@@ -9,6 +9,7 @@
 #include <string>
 
 #include "ControlSystem/ControlErrors/Size/AhSpeed.hpp"
+#include "ControlSystem/ControlErrors/Size/DeltaRDriftInward.hpp"
 #include "ControlSystem/ControlErrors/Size/DeltaRDriftOutward.hpp"
 #include "Utilities/StdHelpers.hpp"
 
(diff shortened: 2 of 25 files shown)


markscheel:
Failing test is Unit.PointwiseFunctions.AnalyticSolutions.Xcts.Schwarzschild (only for gcc-9 Release, passes all other cases). Something is really wrong with macOS x86; dies in spack.

wthrowe:
The macOS build is fixed in #6587, which I've just merged.

src/ControlSystem/ControlErrors/Size/DeltaRDriftOutward.cpp
```diff
@@ -72,7 +72,8 @@ std::string DeltaRDriftOutward::update(
     ss << " Suggested timescale = " << info->suggested_time_scale;
   } else if (update_args.average_radial_distance.has_value() and
              update_args.average_radial_distance.value() <
-                 update_args.max_allowed_radial_distance.value()) {
+                 update_args.max_allowed_radial_distance.value_or(
+                     std::numeric_limits<double>::infinity())) {
```
knelli2, line 77: This seems like a bug fix and not necessarily related to the other changes. Maybe move into it's own commit.

src/ControlSystem/ControlErrors/Size/DeltaRDriftInwardHelpers.hpp
```diff
@@ -0,0 +1,42 @@
...
+#include "ControlSystem/ControlErrors/Size/State.hpp"
+
+namespace control_system::size {
+struct StateUpdateArgs;
+}  // namespace control_system::size
+
+namespace control_system::size::States {
+
+/// Value of target_char_speed when state DeltaRDriftInward is in effect.
+double target_speed_for_inward_drift(
```
knelli2: I would just put all these helpers in the `DeltaRDriftInward.hpp` file. Everywhere you include the helpers you also include state 3 so seems like a logical place to put everything.

src/ControlSystem/ControlErrors/Size/DeltaRDriftInward.cpp
```diff
@@ -0,0 +1,141 @@
...
+  std::stringstream ss{};
+
+  if (char_speed_is_in_danger) {
+    ss << "Current state DeltaRDriftInward. Char speed in danger."
+       << " Switching to AhSpeed.\n";
+    // Switch to AhSpeed mode. Note that we don't check ComovingCharSpeed
+    // like we do in state DeltaR; this behavior agrees with SpEC.
+
+    // This factor prevents oscillating between states Initial and
+    // AhSpeed.  It needs to be slightly greater than unity, but the
```
knelli2: Between Initial and AhSpeed? Or between DeltaR{Inward/Outward/NoDrif} and AhSpeed?

src/ControlSystem/ControlErrors/Size/DeltaRNoDrift.cpp
```diff
@@ -0,0 +1,125 @@
...
+  if (char_speed_is_in_danger) {
+    ss << "Current state DeltaRNoDrift. Char speed in danger.";
+    if (crossing_time_info.t_comoving_char_speed.has_value() or
+        update_args.min_comoving_char_speed < 0.0) {
+      // Comoving char speed is negative or threatening to cross zero, so
+      // staying in DeltaRNoDrift mode will not work.  So switch to AhSpeed
+      // mode.
+
+      // This factor prevents oscillating between states Initial and
+      // AhSpeed.  It needs to be slightly greater than unity, but the
```
knelli2: Same question about Initial

src/ControlSystem/ControlErrors/Size/Info.hpp
```diff
@@ -79,6 +81,23 @@ struct CrossingTimeInfo {
   /// predicted to cross zero (or nullopt if the minimum distance is
   /// increasing).
   std::optional<double> t_delta_radius;
+  /// t_drift_limit_char_speed is the time (relative to the current time) when
+  /// the minimum characteristic speed is predicted to cross the
+  /// min_allowed_char_speed associated with state DeltaRDriftInward (or nullopt
+  /// if the minimum characteristic speed is increasing).
+  std::optional<double> t_drift_limit_char_speed;
+  /// t_delta_radius is the time (relative to the current time) when the
+  /// minimum distance between the horizon and the excision boundary is
```
knelli2: `t_drift_limit_delta_radius`

src/ControlSystem/ControlErrors/Size/Info.hpp
```diff
@@ -79,6 +81,23 @@ struct CrossingTimeInfo {
   /// predicted to cross zero (or nullopt if the minimum distance is
   /// increasing).
   std::optional<double> t_delta_radius;
+  /// t_drift_limit_char_speed is the time (relative to the current time) when
+  /// the minimum characteristic speed is predicted to cross the
+  /// min_allowed_char_speed associated with state DeltaRDriftInward (or nullopt
+  /// if the minimum characteristic speed is increasing).
+  std::optional<double> t_drift_limit_char_speed;
```
knelli2: It doesn't look like `t_drift_limit_char_speed` is used outside this class. Do you think we should still keep it exposed or just use it in the constructor to initialize `t_drift_limit` and discard?

src/ControlSystem/ControlErrors/Size/DeltaRDriftInward.cpp
```diff
@@ -0,0 +1,141 @@
...
+  } else if (should_transition_from_state_inward_drift_to_delta_r_no_drift(
+                 crossing_time_info.t_drift_limit, info->damping_time,
+                 update_args)) {
+    ss << "Current state DeltaRDriftInward. Switching to DeltaRNoDrift.\n";
+    info->discontinuous_change_has_occurred = true;
+    info->state = std::make_unique<States::DeltaRNoDrift>();
+    info->suggested_time_scale = crossing_time_info.t_drift_limit;
+  } else if (crossing_time_info.t_delta_radius.has_value() and
+             info->damping_time >
+                 2.0 * spherepack_factor * update_args.horizon_00 * Y00) {
```
knelli2, line 99: What's this factor of 2 for? This seems to be comparing the damping time to twice the average radius of the horizon

src/ControlSystem/ControlErrors/Size/State.hpp
```diff
@@ -44,6 +47,26 @@ struct StateUpdateArgs {
...
+  /// fast the horizon and the excision boundary move apart in state
+  /// DeltaRDriftInward.  If std::nullopt, then DeltaRDriftInward will
+  /// never be triggered.
+  std::optional<double> inward_drift_velocity;
+  /// min_allowed_radial_distance is the minimum distance between the horizon
+  /// and the excision boundary that will trigger state
+  /// Label::DeltaRDriftInward.
+  std::optional<double> min_allowed_radial_distance;
+  /// min_allowed_char_speed is the minimum char speed that will
+  /// trigger state Label::DeltaRDriftInward.  If both
```
knelli2: I don't think you need `Label::` here

src/ControlSystem/ControlErrors/Size/DeltaR.cpp
```diff
@@ -104,10 +118,27 @@ std::string DeltaR::update(const gsl::not_null<Info*> info,
...
+    info->suggested_time_scale = crossing_time_info.t_drift_limit_delta_radius;
+    ss << "Current state DeltaR. Inward drift limit in danger. Staying "
+          "in DeltaR.\n";
+    ss << " Suggested timescale = " << info->suggested_time_scale;
   } else if (update_args.average_radial_distance.has_value() and
              update_args.average_radial_distance.value() >
                  non_oscillation_drift_outward_factor *
-                     update_args.max_allowed_radial_distance.value()) {
+                     update_args.max_allowed_radial_distance.value_or(
+                         std::numeric_limits<double>::infinity())) {
```
knelli2, line 140: This also seems like a bugfix that should go in a separate commit

src/ControlSystem/ControlErrors/Size/Initial.cpp
```diff
@@ -59,25 +61,33 @@ std::string Initial::update(const gsl::not_null<Info*> info,
...
+                                     : "Comoving char speed positive")
+       << ". Switching to "
+       << (drift_inward ? "DeltaRDriftInward." : "DeltaR.")
+       << " with target char speed " << info->target_char_speed;
   } else if (update_args.average_radial_distance.has_value() and
              update_args.average_radial_distance.value() >
                  non_oscillation_drift_outward_factor *
-                     update_args.max_allowed_radial_distance.value()) {
+                     update_args.max_allowed_radial_distance.value_or(
+                         std::numeric_limits<double>::infinity())) {
```
knelli2, line 89: Also bugfix

src/ControlSystem/ControlErrors/Size/Initial.cpp
```diff
@@ -59,25 +61,33 @@ std::string Initial::update(const gsl::not_null<Info*> info,
...
+          update_args.inward_drift_velocity.value());
+    } else {
+      info->state = std::make_unique<States::DeltaR>();
+    }
+    ss << "Current state Initial. "
+       << (delta_radius_is_in_danger ? "DeltaR is in danger"
+                                     : "Comoving char speed positive")
+       << ". Switching to "
+       << (drift_inward ? "DeltaRDriftInward." : "DeltaR.")
+       << " with target char speed " << info->target_char_speed;
```
knelli2: [nitpick] Print the char speed like the other times it's printed. i.e.

```cpp
   << ". Switching to "
   << (drift_inward ? "DeltaRDriftInward." : "DeltaR.") << "\n";
ss << " Target char speed = " << info->target_char_speed;
```

src/ControlSystem/ControlErrors/Size/Error.cpp
```diff
@@ -186,10 +214,31 @@ ErrorDiagnostics control_error(
...
+      min_allowed_char_speed.has_value()
+          ? predictor_drift_limit_char_speed->min_positive_zero_crossing_time(
+                time)
+          : std::nullopt;
 
   // Compute average radial distance for state DeltaRDriftOutward.
+  // NOTE: This choice corresponds to SpEC's "DeltaRPolicy=Absolute"
+  // and SpEC's "FunctionVsTimeMinDeltaRNoLam00=<NONE>".
+  // However, the default in SpEC is "DeltaRPolicy=Relative" and
+  // "FunctionVsTimeMinDeltaRNoLam00" being an actual ComputeItem.
```
knelli2: So just for clarification, this means we won't be doing what spec does by default. But it's possible to add that in?

src/ControlSystem/ControlErrors/Size.hpp
```diff
@@ -235,9 +235,47 @@ struct Size : tt::ConformsTo<protocols::ControlError> {
...
+      using type = double;
+      static constexpr Options::String help{
+          "Drift excision boundary inward if min char speed is less than "
+          "this."};
+    };
+    struct InwardDriftVelocity {
+      using type = double;
+      static constexpr Options::String help{
+          "Constant drift velocity term, if triggered by "
+          "MinAllowedRadialDistance."};
```
knelli2: What velocity is used if triggered by `MinAllowedCharSpeed`? Or is this used for both? If I'm reading things right, the code does a minimum between this and another quantity but I don't think that needs to be explained in this help text.

src/ControlSystem/ControlErrors/Size/Error.hpp
```diff
@@ -72,6 +77,19 @@ struct ErrorDiagnostics {
...
+ * \param min_allowed_radial_distance the minimum average radial distance
+ *        between the horizon and the excision boundary that is allowed without
+ *        triggering the DeltaRDriftInward state.  If std::nullopt, then
+ *        DeltaRDriftInward will not be used.
+ * \param min_allowed_char_speed the minimum characteristic speed on the
+ *        excision boundary that is allowed without triggering the
+ *        DeltaRDriftInward state.  If std::nullopt, then
+ *        DeltaRDriftInward will not be used.
+ * \param horizon_00 The l=0,m=0 component of the spherepack decomposition
+ *        of the apparent horizon.
```
knelli2: Please make a note that you pass `horizon_00` separately from the full strahlkorper below because this one is time-averaged.

src/ControlSystem/ControlErrors/Size/DeltaRDriftInwardHelpers.hpp
```diff
@@ -0,0 +1,42 @@
...
+/// Returs true if we should transition from state DeltaR to state
+/// DeltaRDriftInward.
+bool should_transition_from_state_delta_r_to_inward_drift(
+    const std::optional<double>& crossing_time_state_3,
+    double damping_time, const StateUpdateArgs& update_args);
+
+/// Returns true if we should transition from state DeltaRDriftInward
+/// to state DeltaRNoDrift.
+bool should_transition_from_state_inward_drift_to_delta_r_no_drift(
+    const std::optional<double>& crossing_time_state_3,
```
knelli2: I think `crossing_time_state_3` should be either just `crossing_time` or `drift_crossing_time` (similar to the variable name that's usually passed in) so we avoid have `state_3` in any variable names.

tests/Unit/ControlSystem/ControlErrors/Test_SizeControlStates.cpp
```diff
@@ -123,6 +153,97 @@ void do_test(const TestParams& test_params,
...
+  //     min_allowed_char_speed==0
+  //  3. avg_radial_distance <= 0.9*min_allowed_radial_distance or
+  //     min_allowed_radial_distance==0
+  //  4. comoving_char_speed_increasing_inward is true
+  //  5. min_allowed_char_speed!=0 or min_allowed_radial_distance!=0
+  //
+  // should_transition_from_state_delta_r_to_inward_drift is true iff
+  // all of the following are true:
+  // A. should_activate_inward_drift is true
+  // B. crossing_time_state_3 >= damping time or crossing_time_state_3 == 0
```
knelli2: Again with `state_3` in the name

tests/Unit/ControlSystem/ControlErrors/Test_SizeControlStates.cpp
```diff
@@ -123,6 +153,97 @@ void do_test(const TestParams& test_params,
...
+  //    or inward_drift_velocity==0
```
knelli2: You comments below only pertain to A. here (which combined 1-5 above). Is there anything we should know about B. in your comments below for each test?

knelli2 (changes requested):
Looks good! Please add a fixup commit for the current changes and you can do the reordering while squashing.

As for organization I think we have two options, since I think you'll need a new commit for some bugfixes that were there before. Option A is

1. Commit for bugfixes/fixing existing clang-tidy issues
2. Commit for everything else

since everything is so closely coupled. Option B is

1. Commit for bugfixes/fixing existing clang-tidy issues
2. Commit that adds DeltaRDrifInward.?pp, DeltaRNoDrift.?pp, Info, State, and StateHistory. Basically everything needed for adding the state 3/4 files, but nothing needed for the logic of switching in the other states. No tests are added in this commit which should be stated in the message.
3. Commit that edits all the other states, adds the new options to Size control error, edits Error.?pp, Factory.hpp, register derived, input files, and adds the tests.

I only have a slight preference for option B, so if you have a strong aversion to doing the reordering that's ok and we can go with A.

tests/Unit/ControlSystem/ControlErrors/Test_SizeControlStates.cpp
markscheel: Yes, added a comment that B. is satisfied on entry.  And simplified B. because if inward_drift_velocity doesn't have a value (which used to be part of B) then 1. is false.  Also changed the !=0 and ==0 to "has a value" and "does not have a value".  (the ==0 is how SpEC does it, and where I copied these comments from, but for SpECTRE these are std::optionals).

src/ControlSystem/ControlErrors/Size.hpp
markscheel: Clarified help text.

src/ControlSystem/ControlErrors/Size/DeltaRDriftInwardHelpers.hpp
markscheel: Ok I am calling it crossing_time_drift_limit because it should be the same as info.t_drift_limit

src/ControlSystem/ControlErrors/Size/Error.cpp
markscheel: Yes, that could be added.  It could even be an option.  We may need an extra FunctionOfTime though, so I should think about how exactly to do that.

src/ControlSystem/ControlErrors/Size/DeltaRDriftInward.cpp
markscheel: Adding a comment.

src/ControlSystem/ControlErrors/Size/Info.hpp
markscheel: I am removing it.

markscheel:
Made changes in 2 fixup commits.
Did not split/reorder commits yet; will do when squashing.

markscheel:
Unit.Evolution.Systems.GhValenciaDivClean.TimeDerivativeTerms is the failed test on gcc-10 Debug.

knelli2 (commented):
Looks good! You can squash/rebase and reorder the commits. I still need to test on a singleBH and BBH run.

markscheel:
Ok, squashed and split commits!  (No, I'm not that fast. I already had done this on a local branch).

knelli2 (approved):
I've been trying to test this with a single BH, but haven't been able to actually get the control system to go into the DeltaRDrifInward state. We'll probably have to run more complicated binaries in order to test this.

For now, since this defaults to not using DeltaRDriftInward, I'll merge this and we can test as we go.

PR #6595, opened 2025-04-21
Retry AH interpolation with new surface after failure

knelli2:
## Proposed changes

If the iterations of a horizon find move the surface such that it's within an excision, this will now retry the interpolation at a surface half the distance between the current failing one and the previous successful one for a certain number of retries.

Unfortunately the test for the horizon finder just does a full horizon find and invokes actions in a random order so it's not straightforward to check that reinterpolation happened for a specific iteration. However, I have tested this on a BBH run and it does work.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
+++ b/src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
@@ -69,14 +69,22 @@ struct ApparentHorizon {
     static constexpr Options::String help = {"Verbosity"};
     using type = ::Verbosity;
   };
-  using options = tmpl::list<InitialGuess, FastFlow, Verbosity>;
+  struct MaxInterpolationRetries {
+    static constexpr Options::String help = {
+        "Number of times to retry the interpolation where, with each retry, "
+        "the two previous surfaces are averaged and that new surface is used."};
+    using type = size_t;
+  };
+  using options =
+      tmpl::list<InitialGuess, FastFlow, Verbosity, MaxInterpolationRetries>;
   static constexpr Options::String help = {
       "Provide an initial guess for the apparent horizon surface\n"
       "(Strahlkorper) and apparent-horizon-finding-algorithm (FastFlow)\n"
       "options."};
 
   ApparentHorizon(ylm::Strahlkorper<Frame> initial_guess_in,
-                  ::FastFlow fast_flow_in, ::Verbosity verbosity_in);
+                  ::FastFlow fast_flow_in, ::Verbosity verbosity_in,
+                  size_t max_interpolation_retries_in);
 
   ApparentHorizon() = default;
   ApparentHorizon(const ApparentHorizon& /*rhs*/) = default;
@@ -91,6 +99,7 @@ struct ApparentHorizon {
   ylm::Strahlkorper<Frame> initial_guess{};
   ::FastFlow fast_flow{};
   ::Verbosity verbosity{::Verbosity::Quiet};
+  size_t max_interpolation_retries{};
 };
 
 template <typename Frame>
--- a/src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
+++ b/src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
@@ -154,7 +154,9 @@ struct ApparentHorizon : tt::ConformsTo<intrp::protocols::ComputeTargetPoints> {
   using common_tags =
       tmpl::push_back<ylm::Tags::items_tags<Frame>, ::ah::Tags::FastFlow,
                       logging::Tags::Verbosity<InterpolationTargetTag>,
-                      ylm::Tags::PreviousStrahlkorpers<Frame>>;
+                      ylm::Tags::PreviousStrahlkorpers<Frame>,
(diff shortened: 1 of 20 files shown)


src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
```diff
@@ -97,37 +99,50 @@ struct ReceivePoints {
...
+          // points:
           //
           //  1. There are no current target points at the temporal_id, OR
           //  2. There are target points already at this temporal_id, but the
           //     iteration of the new target points is greater than the
           //     iteration of the current target points.
+          //  3. There are target points already at this temporal_id, but the
+          //     iteration of the new target points is equal to the iteration of
+          //     the current target points, and we are retrying interpolation at
+          //     the same iteration.
```
wthrowe: This sounds like a message-order race condition.  Tracing through the logic, it looks like if messages arrive in the wrong order you will silently discard the updated one.

wthrowe (changes requested):

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
knelli2: Yes I believe you're right. I pushed a fixup which I think addresses this. Basically, if the interpolator is sent a new iteration than is currently stored, we just always accept it, rather than what I had before where I checked if it was a reinterpolation or not. Then if we have multiple messages of a new iteration floating around, it won't matter which is received first.

wthrowe: That's better, but I think there's still a race if you have to retry twice.  Could you just consider these retries as new iterations so we don't need a new set of logic for them?

knelli2: The `FastFlow` class is in charge of the iterations and has internal logic regarding this, so I don't think trying to externally modify that is a good idea and could cause potential issues.

But yes I believe there still is a race condition. I think I just need to store retry iterations as well. This will all be made a lot easier once we consolidate the horizon finders to just singletons rather than having it as this group+singleton combo.

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
```diff
@@ -97,37 +99,51 @@ struct ReceivePoints {
...
+              vars_infos.at(temporal_id).iteration < iteration or
+              (vars_infos.at(temporal_id).iteration == iteration and
+               vars_infos.at(temporal_id).reinterpolation_iteration <
+                   reinterpolation_iteration)) {
             vars_infos.insert_or_assign(
                 temporal_id,
                 intrp::Vars::Info<VolumeDim, typename InterpolationTargetTag::
                                                  vars_to_interpolate_to_target>{
                     std::move(block_logical_coords), iteration});
           } else if (vars_infos.at(temporal_id).iteration == iteration) {
```
wthrowe: I think this works now, except that this error branch can be entered incorrectly.  It needs to also check the `reinterpolation_iteration`s are the same.

wthrowe (changes requested):

src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
```diff
@@ -69,14 +69,22 @@ struct ApparentHorizon {
     static constexpr Options::String help = {"Verbosity"};
     using type = ::Verbosity;
   };
-  using options = tmpl::list<InitialGuess, FastFlow, Verbosity>;
+  struct MaxInterpolationRetries {
+    static constexpr Options::String help = {
+        "Number of time to retry the interpolation where, with each retry, the "
```
nilsdeppe: `time` -> `times`

nilsdeppe (commented):
My only additional comment is a minor typo. No need for me to approve the typo fix.

knelli2:
Squashed everything in. Thanks!

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
```diff
@@ -97,37 +99,54 @@ struct ReceivePoints {
...
-                << iteration
-                << " twice. Only one set of points per iteration is allowed.");
+          } else if (vars_infos.at(temporal_id).iteration == iteration and
+                     vars_infos.at(temporal_id).reinterpolation_iteration ==
+                         reinterpolation_iteration) {
+            ERROR("Interpolator received target points at iteration "
+                  << iteration << " (and reinterpolation iteration "
+                  << reinterpolation_iteration
+                  << ") twice but we are not retrying an interpolation. Only "
+                     "one set of points per iteration is allowed.");
```
wthrowe: Remove (or reword) everything after "twice".

wthrowe (changes requested):

src/ParallelAlgorithms/Interpolation/Actions/InterpolatorReceivePoints.hpp
knelli2: Done!

wthrowe (approved):

PR #6598, opened 2025-04-22
Generalize DerivLapse in the CCZ4 system to allow more tensor symmetry

hen-w:
Generalize DerivLapse in the CCZ4 system to allow tensor symmetry used in second-order CCZ4 system.

In the first-order CCZ4 system, $A_i=\partial_i \ln\alpha$ is an evolved field, so $\partial_i A_j$ has no symmetry. However, in the second-order CCZ4 system, $\partial_i A_j=\partial_i \partial_j \ln\alpha$ is not evolved; it should have symmetry in i and j, since second partial derivatives commute for $C^2$ functions. This is also required to use `second_partial_derivatives()` in `SecondPartialDerivatives.hpp`. Hence, we generalize the `grad_grad_lapse()` within `DerivLapse.hpp` to allow for symmetric $\partial_i A_j$.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Ccz4/DerivLapse.cpp
+++ b/src/Evolution/Systems/Ccz4/DerivLapse.cpp
@@ -22,7 +22,7 @@ void grad_grad_lapse(
     const tnsr::i<DataType, Dim, Frame>& field_a, const TensorType& d_field_a) {
   if constexpr (std::is_same_v<TensorType, tnsr::ij<DataType, Dim, Frame>>) {
     // We keep this for loop specialization for faster speed when no symmetry
-    // of field_a is present.
+    // of d_field_a is present.
     for (size_t i = 0; i < Dim; ++i) {
       for (size_t j = 0; j < Dim; ++j) {
         result->get(i, j) = field_a.get(i) * field_a.get(j) +
--- a/src/Evolution/Systems/Ccz4/DerivLapse.cpp
+++ b/src/Evolution/Systems/Ccz4/DerivLapse.cpp
@@ -4,6 +4,7 @@
 #include "Evolution/Systems/Ccz4/DerivLapse.hpp"
 
 #include <cstddef>
+#include <type_traits>
 
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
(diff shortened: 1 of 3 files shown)


src/Evolution/Systems/Ccz4/DerivLapse.cpp
```diff
@@ -4,6 +4,7 @@
 #include "Evolution/Systems/Ccz4/DerivLapse.hpp"
```
nilsdeppe, line 4: For commit messages, can you keep the first line to 60 chars? You can go into detail later. E.g.
```
Support symmetric grid_grad_lapse in CCZ4

Any additional things worth noting in the commit message.
```

src/Evolution/Systems/Ccz4/DerivLapse.cpp
```diff
@@ -13,32 +14,41 @@
...
+                    lapse() * christoffel_second_kind(ti::K, ti::i, ti::j) *
+                        field_a(ti::k) +
+                    lapse() * 0.5 *
+                        (d_field_a(ti::i, ti::j) + d_field_a(ti::j, ti::i)));
   }
 }
 
-template <typename DataType, size_t Dim, typename Frame>
+template <typename DataType, size_t Dim, typename Frame, typename TensorType>
 tnsr::ij<DataType, Dim, Frame> grad_grad_lapse(
```
nilsdeppe, line 48: I'm confused, the return type is still non-symmetric. Don't you want an overload where you actually make the tensor `tnsr::ii`?

nilsdeppe (changes requested):

src/Evolution/Systems/Ccz4/DerivLapse.cpp
hen-w: Done.

src/Evolution/Systems/Ccz4/DerivLapse.cpp
hen-w: The return type is not important. I just need `d_field_a` to be symmetric in the second-order CCZ4 system to use `second_partial_derivatives()`, which takes the second derivatives of $\ln \alpha$ and feeds the result into a symmetric tensor, `d_field_a`. I can make the return type symmetric too but the change will propagate to more places than I want.

src/Evolution/Systems/Ccz4/DerivLapse.cpp
```diff
@@ -13,32 +14,41 @@
...
-    for (size_t j = 0; j < Dim; ++j) {
-      result->get(i, j) = field_a.get(i) * field_a.get(j) +
-                          0.5 * (d_field_a.get(i, j) + d_field_a.get(j, i));
-      for (size_t k = 0; k < Dim; ++k) {
-        result->get(i, j) -=
-            christoffel_second_kind.get(k, i, j) * field_a.get(k);
+    const tnsr::i<DataType, Dim, Frame>& field_a, const TensorType& d_field_a) {
+  if constexpr (std::is_same_v<TensorType, tnsr::ij<DataType, Dim, Frame>>) {
+    // We keep this for loop specialization for faster speed when no symmetry
+    // of field_a is present.
```
hen-w: Typo: `field_a` should be `d_field_a`

src/Evolution/Systems/Ccz4/DerivLapse.cpp
nilsdeppe: Discussed in a call: we will leave as-is for now. Once we've got everything working we can do this optimization.

nilsdeppe (approved):

PR #6601, opened 2025-04-22
Enable BNS Translation control, plot control systems with different line styles

nilsdeppe:
## Proposed changes

iTerm2 93L5pQ Control

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ControlSystem/Systems/Translation.hpp
+++ b/src/ControlSystem/Systems/Translation.hpp
@@ -78,7 +78,12 @@ struct Translation : tt::ConformsTo<protocols::ControlSystem> {
   static_assert(
       tt::conforms_to_v<measurement, control_system::protocols::Measurement>);
 
-  using control_error = ControlErrors::Translation<NumberOfObjects>;
+  // For BNS we only control the Translation using the CoM, not the individual
+  // objects like for BBH.
+  using control_error = ControlErrors::Translation<
+      std::is_same_v<measurement, measurements::BothNSCenters>
+          ? 1
+          : NumberOfObjects>;
   static_assert(tt::conforms_to_v<control_error,
                                   control_system::protocols::ControlError>);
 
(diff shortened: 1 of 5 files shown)


src/ControlSystem/Systems/Translation.hpp
```diff
@@ -78,7 +78,10 @@ struct Translation : tt::ConformsTo<protocols::ControlSystem> {
   static_assert(
       tt::conforms_to_v<measurement, control_system::protocols::Measurement>);
 
-  using control_error = ControlErrors::Translation<NumberOfObjects>;
+  using control_error = ControlErrors::Translation<
+      std::is_same_v<measurement, measurements::BothNSCenters>
+          ? 1
+          : NumberOfObjects>;
```
knelli2, line 86: Add a comment above this that even though BNS has two objects, we want to only control the translation via the CoM, not the individual objects, so the control error is only for 1 object.

src/Visualization/Python/PlotControlSystem.py
```diff
@@ -194,6 +194,8 @@ def extract_relevant_columns(
     # Plot all components for all systems. The name for the legend differs
     # slightly from the name in the DataFrame, just to make it prettier for the
     # user. Since both figures will have the same lines, we only show one legend
+    line_styles = ["-", "--", ":", "-."]
+    line_style_cycler = 0
     for system in control_system_components:
```
knelli2: Instead of the cycler, you can do
```py
line_styles = ["-", "--", ":", "-."]
for i, system in enumerate(control_system_components):
    ...
    linestyle=linestyles[ i % len(line_styles)]
```

knelli2 (changes requested):
You can squash directly.

nilsdeppe:
Rebased, squashed. Thanks for the review! :)

knelli2 (approved):

PR #6610, opened 2025-04-27
Generalize neighbors for nonconforming blocks

kidder:
## Proposed changes

With non-conforming Blocks, a Block may have multiple neighbors in a direction, and those neighbors need not share a common Orientation.  Thus store an OrientationMap indexed by the id of the Block.  This is sufficient for Elements as well since all Elements in a given Block are aligned with each other.

Note this requires a block id to be passed to fetch the correct OrientationMap from the Neighbors.

Also, for convenience, store a boolean on whether or not the neighbors are conforming to the host.

### Upgrade instructions

You must pass the ID of the Block or Neighbor to the orientation member function of BlockNeighbors or Neighbors.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Structure/Neighbors.cpp
+++ b/src/Domain/Structure/Neighbors.cpp
@@ -3,21 +3,75 @@
 
 #include "Domain/Structure/Neighbors.hpp"
 
+#include <algorithm>
 #include <ostream>
 #include <pup.h>
 #include <pup_stl.h>
 
 #include "Domain/Structure/ElementId.hpp"
+#include "Utilities/Algorithm.hpp"
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/StdHelpers.hpp"
 
+namespace {
+template <size_t VolumeDim>
+bool ids_and_orientations_are_consistent(
+    const std::unordered_set<size_t>& block_ids,
+    const std::unordered_map<size_t, OrientationMap<VolumeDim>>& orientations) {
+  return alg::none_of(block_ids, [&orientations](size_t block_id) {
+    return orientations.count(block_id) == 0;
+  });
+}
+
+template <size_t VolumeDim>
+bool ids_and_orientations_are_consistent(
+    const std::unordered_set<ElementId<VolumeDim>>& element_ids,
+    const std::unordered_map<size_t, OrientationMap<VolumeDim>>& orientations) {
+  return alg::none_of(element_ids,
+                      [&orientations](const ElementId<VolumeDim>& element_id) {
+                        return orientations.count(element_id.block_id()) == 0;
+                      });
+}
+}  // namespace
+
+template <size_t VolumeDim, typename IdType>
+Neighbors<VolumeDim, IdType>::Neighbors(
+    std::unordered_set<IdType> ids,
+    std::unordered_map<size_t, OrientationMap<VolumeDim>> orientations,
+    const bool are_conforming)
+    : ids_(std::move(ids)),
+      orientations_(std::move(orientations)),
+      are_conforming_(are_conforming) {
+  ASSERT(alg::none_of(orientations_,
+                      [](const auto& orientation) {
+                        return orientation.second ==
+                               OrientationMap<VolumeDim>{};
+                      }),
+         "Cannot use a default-constructed OrientationMap in Neighbors.");
+  ASSERT(orientations_.size() == 1 or not are_conforming_,
+         "Conforming neighbors must all be in the same block, but multiple "
+         "orientations were specified "
+             << orientations_);
+  ASSERT(ids_and_orientations_are_consistent(ids_, orientations_),
+         "Not all Ids " << ids_ << " have orientations " << orientations_);
+}
+
 template <size_t VolumeDim, typename IdType>
 Neighbors<VolumeDim, IdType>::Neighbors(std::unordered_set<IdType> ids,
                                         OrientationMap<VolumeDim> orientation)
-    : ids_(std::move(ids)), orientation_(std::move(orientation)) {
-  ASSERT(orientation_ != OrientationMap<VolumeDim>{},
+    : ids_(std::move(ids)) {
+  ASSERT(orientation != OrientationMap<VolumeDim>{},
          "Cannot use a default-constructed OrientationMap in Neighbors.");
+  if constexpr (std::is_same_v<size_t, IdType>) {
+    orientations_.emplace(*ids_.begin(), std::move(orientation));
+  } else {
+    orientations_.emplace(ids_.begin()->block_id(), std::move(orientation));
+  }
+  ASSERT(ids_and_orientations_are_consistent(ids_, orientations_),
+         "Conforming neighbors must all be in the same block, but the "
+         "following Ids were specified: "
+             << ids_);
 }
 
 template <size_t VolumeDim, typename IdType>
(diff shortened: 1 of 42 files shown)


src/Domain/Structure/Neighbors.hpp
```diff
@@ -31,8 +31,22 @@ class Neighbors {
...
 
  public:
+  /// Construct with the ids, the orientation of the neighbors relative to the
+  /// host, and whether the neighbors are conforming.
+  ///
+  /// \param ids the ids of the neighbors.
+  /// \param orientations An OrientationMap (which takes objects in the logical
+  /// coordinate frame of the host and maps them to the logical coordinate frame
+  /// of the neighbor) for each neighboring Block (Elements within a Block share
+  /// the same orientation).
```
nilsdeppe: Is the key in this may the block ID or the `IdType`?

src/Domain/Structure/Neighbors.hpp
```diff
@@ -31,8 +31,22 @@ class Neighbors {
...
+  /// \param are_conforming whether or not the block logical coordinates of the
+  /// neighbors conform to those of the host
```
nilsdeppe: Could you add examples of conforming and not conforming? This way it's abundantly clear

src/Domain/Structure/Neighbors.hpp
```diff
@@ -60,15 +74,32 @@ class Neighbors {
...
+  /// The orientations of the neighbors for each neighboring Block.
+  ///
+  /// \note All Elements within a Block share the same orientation.
+  const std::unordered_map<size_t, OrientationMap<VolumeDim>>& orientations()
+      const {
+    return orientations_;
+  }
+
+  /// Whether or not the block logical coordinates of the neighbors conform to
+  /// those of the host
```
nilsdeppe: Again, give examples of what it means to conform or not

src/Domain/Structure/Neighbors.hpp
```diff
@@ -60,15 +74,32 @@ class Neighbors {
...
+  /// \note This should only be called to reset Element Neighbors after
+  /// h-refinement
+  void set_ids_to(std::unordered_set<IdType> new_ids);
 
   /// Add ids of neighbors.
-  /// Adding an existing neighbor is allowed.
+  ///
+  /// \note Adding an existing neighbor is allowed.
+  /// \note The additional ids must be from Blocks with the existing
+  /// orientations.
```
nilsdeppe, line 103: why does this not have a not about post h-refinemnt? Naively I would've guessed this should only be called after h-refinement?

src/Domain/Structure/Neighbors.cpp
```diff
@@ -12,19 +13,56 @@
...
+    return orientations_.at(id);
+  } else {
+    return orientations_.at(id.block_id());
+  }
+}
+
+template <size_t VolumeDim, typename IdType>
+void Neighbors<VolumeDim, IdType>::set_ids_to(
+    const std::unordered_set<IdType> new_ids) {
+  ids_ = std::move(new_ids);
```
nilsdeppe, line 95: add `ASSERT` that the blocks being added have orientation maps stored in the class like the dox says is required

src/Domain/Structure/Neighbors.cpp
```diff
@@ -12,19 +13,56 @@
...
 
+template <size_t VolumeDim, typename IdType>
+Neighbors<VolumeDim, IdType>::Neighbors(
+    std::unordered_set<IdType> ids,
+    std::unordered_map<size_t, OrientationMap<VolumeDim>> orientations,
+    const bool are_conforming)
+    : ids_(std::move(ids)),
+      orientations_(std::move(orientations)),
+      are_conforming_(are_conforming) {
+  ASSERT(std::ranges::none_of(orientations_,
```
nilsdeppe: Ranges!!! :heart:

src/Domain/Structure/Neighbors.cpp
```diff
@@ -21,12 +43,18 @@ Neighbors<VolumeDim, IdType>::Neighbors(
     : ids_(std::move(ids)),
       orientations_(std::move(orientations)),
       are_conforming_(are_conforming) {
-  ASSERT(std::ranges::none_of(orientations_,
```
nilsdeppe: ranges? :( Do ranges not work for this?

src/Domain/Structure/Neighbors.cpp
```diff
@@ -61,6 +93,11 @@ template <size_t VolumeDim, typename IdType>
 void Neighbors<VolumeDim, IdType>::set_ids_to(
     const std::unordered_set<IdType> new_ids) {
   ids_ = std::move(new_ids);
+  ASSERT(ids_and_orientations_are_consistent(ids_, orientations_),
+         "Some of the new Ids "
+             << ids_
```
nilsdeppe: Bruh, this is _all_ ids, not just new ones. Someone's gonna hate you for this at some point, and it may be you :P

src/Domain/Structure/Neighbors.cpp
```diff
@@ -69,6 +106,11 @@ void Neighbors<VolumeDim, IdType>::add_ids(
   for (const auto& id : additional_ids) {
```
nilsdeppe: Why not take the set by value and move-insert?

src/Domain/Structure/Neighbors.cpp
```diff
@@ -9,10 +9,32 @@
...
 #include "Utilities/ErrorHandling/Assert.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 #include "Utilities/StdHelpers.hpp"
 
+namespace {
+template <size_t VolumeDim>
+bool ids_and_orientations_are_consistent(
+    const std::unordered_set<size_t>& block_ids,
+    const std::unordered_map<size_t, OrientationMap<VolumeDim>>& orientations) {
+  return alg::none_of(block_ids, [&orientations](size_t block_id) {
```
nilsdeppe, line 22: Ranges..??? :( :(

nilsdeppe (changes requested):

src/Domain/Structure/Neighbors.cpp
kidder: can't use ranges until we get rid of gcc9

src/Domain/Structure/Neighbors.cpp
kidder: was done in the next commit

src/Domain/Structure/Neighbors.cpp
kidder: they are equivalent in this function, but I'll reword...

src/Domain/Structure/Neighbors.hpp
kidder: I'll state that it is a size_t for the block ID

src/Domain/Structure/Neighbors.hpp
kidder: can I point to the documentation of neighbor_is_conforming or do you want something else?

src/Domain/Structure/Neighbors.hpp
kidder: It is also called by create_initial_element

src/Domain/Structure/Neighbors.hpp
nilsdeppe: That's fine :) Just so that future-me can avoid fuming about you ;)

src/Domain/Structure/Neighbors.hpp
nilsdeppe: Ah, that's what I figured. I don't think you need to change anything, because unless someone is reviewing code I doubt they would ask this.

src/Domain/Structure/Neighbors.cpp
nilsdeppe: 💔

nilsdeppe (approved):

nilsdeppe:
Yet again the XCTS timed out in CI...

PR #6611, opened 2025-04-28
Only interpolate for horizon finds from certain blocks

knelli2:
## Proposed changes

These are currently taken as options from the input file. Eventually
we could have this be more automatic, e.g. the domain creators know
which blocks could potentially hold a horizon, but that would be
a future optimization.

This is the first step in a lot of backend work to combine the horizon finders and get rid of the interpolator group component.

I've tested this on a single BH while interpolating only some blocks and all blocks. And also tried on a BBH

### Upgrade instructions

Each entry under the `ApparentHorizons:` block of the input file gets a new option `BlocksForInterpolation:`. This could either be a list of block groups or `All` to indicate all blocks in the doman.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/Disk.cpp
+++ b/src/Domain/Creators/Disk.cpp
@@ -60,6 +60,17 @@ Disk::Disk(typename InnerRadius::type inner_radius,
   if (boundary_condition_ != nullptr and is_periodic(boundary_condition_)) {
     PARSE_ERROR(context, "Cannot have periodic boundary conditions on a disk.");
   }
+
+  block_groups_["Shell0"];
+  block_names_.reserve(5);
+  const std::array<std::string, 4> wedge_directions{"UpperX", "UpperY",
+                                                    "LowerX", "LowerY"};
+  for (const std::string& name : wedge_directions) {
+    block_names_.emplace_back(name);
+    block_groups_["Shell0"].insert(name);
+  }
+  block_names_.emplace_back("CenterSquare");
+  block_groups_["CenterSquare"].insert("CenterSquare");
 }
 
 Domain<2> Disk::create_domain() const {
@@ -131,7 +142,8 @@ Domain<2> Disk::create_domain() const {
     boundary_conditions_all_blocks.emplace_back();
   }
 
-  return Domain<2>{std::move(coord_maps), corners};
+  return Domain<2>{std::move(coord_maps), corners,      {}, {},
+                   block_names_,          block_groups_};
 }
 
 std::vector<DirectionMap<
--- a/src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.cpp
+++ b/src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.cpp
@@ -5,8 +5,10 @@
 
 #include <algorithm>
 #include <cstddef>
+#include <unordered_set>
 
 #include "IO/Logging/Verbosity.hpp"
+#include "Options/Context.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
 
 namespace Frame {
@@ -17,11 +19,13 @@ namespace intrp::OptionHolders {
 template <typename Frame>
 ApparentHorizon<Frame>::ApparentHorizon(
     ylm::Strahlkorper<Frame> initial_guess_in, ::FastFlow fast_flow_in,
-    ::Verbosity verbosity_in, const size_t max_interpolation_retries_in)
+    ::Verbosity verbosity_in, const size_t max_interpolation_retries_in,
+    std::optional<std::vector<std::string>> blocks_for_interpolation_in)
     : initial_guess(std::move(initial_guess_in)),
       fast_flow(std::move(fast_flow_in)),  // NOLINT
       verbosity(std::move(verbosity_in)),  // NOLINT
-      max_interpolation_retries(max_interpolation_retries_in) {}
+      max_interpolation_retries(max_interpolation_retries_in),
+      blocks_for_interpolation(std::move(blocks_for_interpolation_in)) {}
 // clang-tidy std::move of trivially copyable type.
 
 template <typename Frame>
@@ -30,6 +34,7 @@ void ApparentHorizon<Frame>::pup(PUP::er& p) {
   p | fast_flow;
   p | verbosity;
   p | max_interpolation_retries;
+  p | blocks_for_interpolation;
 }
 
 template <typename Frame>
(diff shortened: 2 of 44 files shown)


src/Domain/Creators/Disk.cpp
```diff
@@ -60,6 +60,17 @@ Disk::Disk(typename InnerRadius::type inner_radius,
   if (boundary_condition_ != nullptr and is_periodic(boundary_condition_)) {
```
nilsdeppe, line 60: Only the important domains ;) :D

src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.cpp
```diff
@@ -17,11 +19,24 @@ namespace intrp::OptionHolders {
 template <typename Frame>
 ApparentHorizon<Frame>::ApparentHorizon(
     ylm::Strahlkorper<Frame> initial_guess_in, ::FastFlow fast_flow_in,
-    ::Verbosity verbosity_in, const size_t max_interpolation_retries_in)
+    ::Verbosity verbosity_in, const size_t max_interpolation_retries_in,
+    const std::optional<std::vector<std::string>>& blocks_for_interpolation_in,
```
nilsdeppe: take by value and move

src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.cpp
```diff
@@ -30,14 +45,16 @@ void ApparentHorizon<Frame>::pup(PUP::er& p) {
   p | fast_flow;
   p | verbosity;
   p | max_interpolation_retries;
+  p | blocks_for_interpolation;
```
nilsdeppe, line 37: I assume this will break checkpoint restarts? (I don't care, just asking :) )

src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
```diff
@@ -138,6 +165,96 @@ struct ApparentHorizon : db::SimpleTag {
...
+      ::domain::OptionTags::DomainCreator<Metavariables::volume_dim>>;
+
+  static constexpr bool pass_metavariables = true;
+  template <typename Metavariables, typename... HorizonOptions>
+  static type create_from_options(
+      const std::unique_ptr<::DomainCreator<Metavariables::volume_dim>>&
+          domain_creator,
+      const HorizonOptions&... all_horizon_options) {
+    return create_from_options_impl<Metavariables>(
+        std::forward_as_tuple(all_horizon_options...), domain_creator);
```
nilsdeppe: Why not just pass `std::make_index_sequence<sizeof...(all_horizon_options>{}` here?

src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
```diff
@@ -138,6 +165,96 @@ struct ApparentHorizon : db::SimpleTag {
...
+      const std::unique_ptr<::DomainCreator<Metavariables::volume_dim>>&
+          domain_creator) {
+    std::unordered_map<std::string, std::unordered_set<std::string>> result{};
+
+    const auto block_names = domain_creator->block_names();
+    const auto block_groups = domain_creator->block_groups();
+
+    const auto append_to_result = [&](const std::string& name,
+                                      const auto& horizon_options) {
+      result[name];
```
nilsdeppe: I don't understand this. Are you just getting the `name` out of `result` and doing nothing?

src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
```diff
@@ -138,6 +165,96 @@ struct ApparentHorizon : db::SimpleTag {
...
+          // Insert all blocks for this block group
+          result[name].insert(block_groups.at(block_group_name).begin(),
+                              block_groups.at(block_group_name).end());
+        }
+      } else {
+        // Insert all blocks
+        result.at(name).insert(block_names.begin(), block_names.end());
+      }
+    };
+
```
nilsdeppe, line 232: And then
```cpp
expand_pack(append_to_result(
           tmpl::at<option_tags<Metavariables>, tmpl::size_t<Is + 1>>::name(),
           std::get<Is>(all_horizon_options)))...);
```
here? You might need to return `'0'` or some type other than `void` from the lambda, though I think the fold expression with the comma operator is also fine:
```cpp
((append_to), ...);
```

src/ParallelAlgorithms/Interpolation/Actions/InitializeInterpolator.hpp
```diff
@@ -69,8 +71,21 @@ struct InitializeInterpolator {
...
+            Metavariables>;
+    db::mutate<Tags::NumberOfElements<Dim>>(
+        [](const gsl::not_null<std::unordered_map<
+               std::string, std::unordered_set<ElementId<Dim>>>*>
+               num_elements) {
+          tmpl::for_each<sequential_targets>([&](auto target_v) {
+            using target = tmpl::type_from<decltype(target_v)>;
+            const std::string& target_name = pretty_type::name<target>();
+
+            (*num_elements)[target_name];
```
nilsdeppe: `= 0;`? This is uninitialized, or best case default-constructed?

nilsdeppe (changes requested):

src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.cpp
knelli2: Yes this will break charm checkpoints because it's in the cache but not versioned.

knelli2:
@nilsdeppe Pushed fixups

src/ParallelAlgorithms/ApparentHorizonFinder/InterpolationTarget.hpp
```diff
@@ -200,57 +199,44 @@ struct BlocksForInterpolation : db::SimpleTag, BlocksForInterpolationBase {
...
-       ...);
+      // Needed for the expand_pack below
+      return 0;
     };
 
-    expand_tuple(
-        std::make_index_sequence<
-            std::tuple_size_v<std::decay_t<decltype(all_horizon_options)>>>{});
+    expand_pack(append_to_result(
+        tmpl::at<option_tags<Metavariables>, tmpl::size_t<Is + 1>>::name(),
```
nilsdeppe: you can use `at_c` and then don't need the `tmpl::size_t`

nilsdeppe (changes requested):
LGTM, please squash :)

nilsdeppe (approved):

PR #6615, opened 2025-04-30
Fix bugs in AMR TruncationError criterion

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Amr/Criteria/TruncationError.cpp
+++ b/src/ParallelAlgorithms/Amr/Criteria/TruncationError.cpp
@@ -5,7 +5,6 @@
 
 #include <array>
 #include <cstddef>
-#include <optional>
 
 #include "DataStructures/DataVector.hpp"
 #include "Domain/Amr/Flag.hpp"
@@ -21,8 +20,8 @@ void max_over_components(
     const gsl::not_null<std::array<Flag, Dim>*> result,
     const gsl::not_null<std::array<DataVector, Dim>*> power_monitors_buffer,
     const DataVector& tensor_component, const Mesh<Dim>& mesh,
-    const std::optional<double> target_abs_truncation_error,
-    const std::optional<double> target_rel_truncation_error) {
+    const double target_abs_truncation_error,
+    const double target_rel_truncation_error) {
   // We take the highest-priority refinement flag in each dimension, so if any
   // tensor component has a truncation error above the target, the element will
   // increase p refinement in that dimension. And only if all tensor components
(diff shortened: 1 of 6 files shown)


src/ParallelAlgorithms/Amr/Criteria/TruncationError.cpp
```diff
@@ -64,11 +61,8 @@ void max_over_components(
           -PowerMonitors::relative_truncation_error(modes, modes.size() - 1));
       const double abs_truncation_error_coarsened =
           umax * rel_truncation_error_coarsened;
-      if ((target_rel_truncation_error.has_value() and
-           rel_truncation_error_coarsened <=
-               target_rel_truncation_error.value()) and
```
nilsdeppe, line 69: I don't generally disagree with people needing to set both, but just for a record of it: would the but be fixed by doing `value_or(std::numeric_limits<double>::inf()`?

src/ParallelAlgorithms/Amr/Criteria/TruncationError.cpp
```diff
@@ -61,7 +61,7 @@ void max_over_components(
           -PowerMonitors::relative_truncation_error(modes, modes.size() - 1));
       const double abs_truncation_error_coarsened =
           umax * rel_truncation_error_coarsened;
-      if (rel_truncation_error_coarsened <= target_rel_truncation_error and
+      if (rel_truncation_error_coarsened <= target_rel_truncation_error or
           abs_truncation_error_coarsened <= target_abs_truncation_error) {
```
nilsdeppe: Could we recast this and the above check into something like:
```
err = abs(error)/(atol + rtol * abs(u))
```
I ask because I find that much easier to understand than the logic here, and it nicely maps onto 17.2.7-17.2.9 in Numerical Recipes. I think this does effectively that, if I understand correctly.

Up to you, just an idea on how to eliminate the logic, because I think the logic is subtle and requires more thought.

nilsdeppe (changes requested):

wthrowe:
@nilsvu has said he may disagree with this change, and has asked that we wait until he has a change to look at it next week.  (And has pointed out that the docs need updating.)

nilsvu:
Thanks @wthrowe. No I don't disagree with the change, you can go ahead. When I wrote this logic I did intend to have the criterion satisfy _both_ rel and abs target if both are specified, with the idea that you want to stay below those targets if you set them. This is also what the option helpstrings say, so I don't consider this a bugfix. You're changing the behavior to satisfy _either_ the rel or abs target. If you think that makes more sense then go ahead with the change, but also change the docs.

nilsdeppe:
Okay, so my (rather strong) preference is to use
$\mathcal{E}=\frac{|\Delta|}{\epsilon_{\mathrm{abs}}+|u|\epsilon_{\mathrm{rel}}}$.
Then we require $\mathcal{E}<1$ to not refine, and the same for the "one fewer point" in order to coarsen. The goal is that you smoothly transition to "not caring" about a variable whose value is too small if you set an absolute & relative tolerance. If you set an absolute tolerance of $0$ then you only use the relative tolerance, and same with setting a relative tolerance of $0$ to only care about the absolute value. I believe this is what the time steppers already do.

nilsvu:
Yep I agree that it makes more sense to transition between rel and abs tol for this criterion, rather than satisfy both.

src/ParallelAlgorithms/Amr/Criteria/TruncationError.cpp
wthrowe: Yes.  Infinity before the second commit changed to zero after would also have worked.

wthrowe:
Added a fixup and a new commit changing the interface to the power monitor error because it was confusing.

nilsdeppe (commented):
LGTM! Please squash

nilsdeppe (approved):

PR #6617, opened 2025-05-01
Add additional error estimates needed for variable-order stepping

wthrowe:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Time/TimeSteppers/AdamsBashforth.cpp
+++ b/src/Time/TimeSteppers/AdamsBashforth.cpp
@@ -151,8 +151,9 @@ template <bool DenseOutput, typename T>
 std::optional<StepperErrorEstimate> AdamsBashforth::update_u_common(
     const gsl::not_null<T*> u, const ConstUntypedHistory<T>& history,
     const tmpl::conditional_t<DenseOutput, ApproximateTime, Time>& time,
-    const std::optional<StepperErrorTolerances>& tolerances) const {
-  ASSERT(not(DenseOutput and tolerances.has_value()),
+    const StepperErrorTolerances& tolerances) const {
+  ASSERT(not DenseOutput or
+             tolerances.estimates == StepperErrorTolerances::Estimates::None,
          "Can't compute errors in dense output.");
   ASSERT(history.size() >= history.integration_order(),
          "Incorrect data to take an order-" << history.integration_order()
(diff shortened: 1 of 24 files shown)


src/Time/TimeSteppers/AdamsBashforth.cpp
```diff
@@ -173,12 +176,29 @@ std::optional<StepperErrorEstimate> AdamsBashforth::update_u_common(
...
+                  static_cast<decltype(control_times)::difference_type>(
+                      error_order),
+              control_times.end(), step_start, time);
+          gsl::at(error->errors, error_order)
+              .emplace(evaluate_error(u, history, *tolerances,
+                                      higher_order_coefficients,
+                                      lower_order_coefficients));
+        }
+      }
     }
```
nilsdeppe: A naive question: I assume currently this is done over the entire grid, over all variables. Is that the case? If so, would it make sense to collapse that to a single double before computing all the error estimates? That could help reduce the cost of the error estimate. One way of doing this is assume that the grid point with the worst error will be the same for all steppers, and just compute it at that grid point for all the ones that aren't the current order. I wouldn't feel great always doing that for the lowest or highest order, since I worry there may be some noise if you are far from the current order.

 Since this is all templated on `T` it's not clear to me what exactly is a `Variables` vs a `double, sorry :)

src/Time/TimeSteppers/AdamsMoultonPc.cpp
```diff
@@ -243,6 +243,26 @@ std::optional<StepperErrorEstimate> AdamsMoultonPc<Monotonic>::update_u_common(
                     history.integration_order() - 1,
                     evaluate_error(u, history, *tolerances, update_coefficients,
                                    predictor_coefficients));
+
+      if (not order_.has_value()) {
+        for (size_t error_order = history.integration_order() - 2;
```
nilsdeppe, line 250: Same comment/question about if this is done over a `Variables` as for AB

nilsdeppe (changes requested):
LGTM, I just have a question :)

src/Time/TimeSteppers/AdamsBashforth.cpp
wthrowe: Finally getting back to this.  Here `T` can be `double`, `DataVector`, or complex equivalents, so concern about cost is definitely reasonable, and it is unclear to me at the moment whether that cost is significant or not.

Since I'm having trouble coming up with a good algorithm for choosing the step order, my plan for the moment is to add a new commit here to make whether to provide lower-order estimates an external-to-the-TimeStepper choice that will be disabled for now, and just use variable-order as a way to avoid the history-initialization phase.  I think I will still need low-order estimates for LTS h-refinement, so when I get to that I can only request them at slab boundaries, which should be infrequent enough so as not to be a performance problem.

If we later try for more truly adaptive-order methods, maybe for higher mass-ratios, then we can return to the performance questions.

nilsdeppe: Okay, thank you for the detailed and thoughtful explanation! That sounds great to me :)

nilsdeppe (approved):

nilsdeppe:
CI failure was in Poisson ProductOfSinusoids, so unrelated.

PR #6618, opened 2025-05-02
Add first FD spacetime_derivatives() for SoCcz4

hen-w:
Add the first spatial derivatives by finite differencing for calculating the RHS of the evolution equations of the second-order Ccz4 system (see eq. 4(a)-4(i) of this [paper](https://journals.aps.org/prd/pdf/10.1103/PhysRevD.97.084053)). 

We create a new library SoCcz4 in the `Ccz4/FiniteDifference` directory to better separate the finite difference related code from the Ccz4 library which uses DG code. We use the 4th order first partial derivatives stencils implemented in `NumericalAlgorithm` in `Derivatives.cpp` to calculate the spatial derivatives of all the evolved variable defined by `variables_tag` in `FiniteDifference/System.hpp`. We use 4th order finite differencing here to keep the same order as the second partial derivatives which only have a 4th order finite difference stencil.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Ccz4/FiniteDifference/CMakeLists.txt
+++ b/src/Evolution/Systems/Ccz4/FiniteDifference/CMakeLists.txt
@@ -1,9 +1,14 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
+set(LIBRARY SoCcz4)
+
+add_spectre_library(${LIBRARY})
+
 spectre_target_sources(
   ${LIBRARY}
   PRIVATE
+  Derivatives.cpp
   SecondPartialDerivatives.cpp
   )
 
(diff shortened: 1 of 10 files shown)


teukolsky:
Did you mean spatial derivatives or spacetime derivatives?

hen-w:
> Did you mean spatial derivatives or spacetime derivatives?

@teukolsky I mean spatial derivatives. But the original code in `GrMhd/GhValenciaDivClean/FiniteDifference/Derivatives.cpp`, from which I adapted my code, calls it `spacetime_derivatives()` :sweat:. So I am trying to keep the same convention here.

nilsdeppe:
Sorry, I thought I'd commented on this weeks ago. The name originates from the fact that in the GHMHD these are the derivatives of the spacetime variables, not the MHD variables. Given that at some point someone will inevitably couple Z4 to MHD, I'm in favor of staying consistent with GHMHD :shrug: But making sure the documentation is clear :)

nilsdeppe (approved):

PR #6622, opened 2025-05-08
Add combined event for AhC

knelli2:
## Proposed changes

Adds a new event `FindCommonHorizon` that is a wrapper around the Interpolate event which starts a horizon find, and the ObserveFields event which writes volume data to disk for the ringdown restart.

This is the first of 3 PRs that avoid writing volume data for a common horizon find that fails.

### Upgrade instructions

In BBH input files, the part of the events and triggers at slabs block which had the ringdown events goes from this
```yaml
      - ObservationAhC
      - ObserveFields:
          SubfileName: ForContinuation
          VariablesToObserve:
            - SpacetimeMetric
            - Pi
            - Phi
          InterpolateToMesh: None
          # This volume data is for ringdown, so double precision is needed.
          CoordinatesFloatingPointType: Double
          FloatingPointTypes: [Double]
          BlocksToObserve: All
```
to this
```yaml
      - ObservationAhC:
          SubfileName: ForContinuation
          VariablesToObserve:
            - SpacetimeMetric
            - Pi
            - Phi
          InterpolateToMesh: None
          # This volume data is for ringdown, so double precision is needed.
          CoordinatesFloatingPointType: Double
          FloatingPointTypes: [Double]
          BlocksToObserve: All
```

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #6623, opened 2025-05-08
Simplify volume actions, add dependency to interp framework

knelli2:
## Proposed changes

This is the second of 3 PRs that avoid writing volume data for a common horizon find that fails.

This PR is some miscellaneous changes to `VolumeActions.hpp` that are needed and also adds what is called a "dependency" to the interpolation framework which the combined common horizon event in #6622 will make use of eventually. Currently, this dependency does nothing.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

Technically depends on #6622, but only for a single test so I didn't include those commits here. Whichever is merged first, I'll just rebase.

### DIFF
--- a/src/IO/Observer/VolumeActions.hpp
+++ b/src/IO/Observer/VolumeActions.hpp
@@ -157,6 +157,86 @@ void write_data(const std::string& h5_file_name,
                 const std::string& subfile_path,
                 const observers::ObservationId& observation_id,
                 std::vector<ElementVolumeData>&& volume_data);
+
+template <typename ParallelComponent, typename Metavariables,
+          typename VolumeDataAtObsId>
+void write_combined_volume_data(
+    Parallel::GlobalCache<Metavariables>& cache,
+    const observers::ObservationId& observation_id,
+    const VolumeDataAtObsId& volume_data,
+    const gsl::not_null<Parallel::NodeLock*> volume_file_lock,
+    const std::string& subfile_name) {
+  ASSERT(not volume_data.empty(),
+         "Failed to populate volume_data before trying to write it.");
+
+  std::vector<ElementVolumeData> volume_data_to_write;
+
+  if constexpr (std::is_same_v<tmpl::at_c<VolumeDataAtObsId, 1>,
+                               ElementVolumeData>) {
+    volume_data_to_write.reserve(volume_data.size());
+    for (const auto& [id, element] : volume_data) {
+      (void)id;  // avoid compiler warnings
+      volume_data_to_write.push_back(element);
+    }
+  } else {
+    size_t total_size = 0;
+    for (const auto& [id, vec_elements] : volume_data) {
+      (void)id;  // avoid compiler warnings
+      total_size += vec_elements.size();
+    }
+    volume_data_to_write.reserve(total_size);
+
+    for (const auto& [id, vec_elements] : volume_data) {
+      (void)id;  // avoid compiler warnings
+      volume_data_to_write.insert(volume_data_to_write.end(),
+                                  vec_elements.begin(), vec_elements.end());
+    }
+  }
+
+  // Write to file. We use a separate node lock because writing can be
+  // very time consuming (it's network dependent, depends on how full the
+  // disks are, what other users are doing, etc.) and we want to be able
+  // to continue to work on the nodegroup while we are writing data to
+  // disk.
+  const std::lock_guard hold_lock(*volume_file_lock);
+  {
+    // Scoping is for closing HDF5 file before we release the lock.
+    const auto& file_prefix = Parallel::get<Tags::VolumeFileName>(cache);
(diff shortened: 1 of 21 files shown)


src/ParallelAlgorithms/Interpolation/Actions/AddTemporalIdsToInterpolationTarget.hpp
```diff
@@ -57,7 +57,8 @@ struct AddTemporalIdsToInterpolationTarget {
   static void apply(db::DataBox<DbTags>& box,
                     Parallel::GlobalCache<Metavariables>& cache,
                     const ArrayIndex& array_index,
-                    const TemporalId& temporal_id) {
+                    const TemporalId& temporal_id,
+                    std::optional<std::string>&& dependency) {
```
wthrowe: Take by value.

src/ParallelAlgorithms/Interpolation/InterpolationTargetDetail.hpp
```diff
@@ -503,7 +512,8 @@ bool flag_temporal_id_for_interpolation(
 /// - AddTemporalIdsToInterpolationTarget (called by Events::Interpolate)
 template <typename InterpolationTargetTag, typename DbTags, typename TemporalId>
 void flag_temporal_id_as_pending(const gsl::not_null<db::DataBox<DbTags>*> box,
-                                 const TemporalId& temporal_id) {
+                                 const TemporalId& temporal_id,
+                                 std::optional<std::string>&& dependency) {
```
wthrowe: Take by value.

tests/Unit/ParallelAlgorithms/Interpolation/Test_Interpolate.cpp
```diff
@@ -113,10 +114,13 @@ struct MockAddTemporalIdsToInterpolationTarget {
       Parallel::GlobalCache<Metavariables>& /*cache*/,
       const ArrayIndex& /*array_index*/,
       const typename Metavariables::InterpolatorTargetA::temporal_id::type&
-      /*temporal_id*/) {
+      /*temporal_id*/,
+      std::optional<std::string>&& dependency) {
```
wthrowe: By value.

wthrowe (changes requested):

knelli2:
@wthrowe I squashed directly since it was small

src/IO/Observer/VolumeActions.hpp
```diff
@@ -291,116 +291,85 @@ struct ContributeVolumeDataToWriter {
                          VolumeDataAtObsId received_volume_data) {
     // The below gymnastics with pointers is done in order to minimize the
     // time spent locking the entire node, which is necessary because the
-    // DataBox does not allow any functions calls, both get and mutate, during
```
nilsdeppe, line 214: The code already exists and works, why do this change at all?  Moves _can_ be expensive, there's nothing that guarantees they won't. Moves can even fall back to copies.

nilsdeppe (changes requested):

src/IO/Observer/VolumeActions.hpp
knelli2: In #6624, I add a new threaded action that adds the dependency to the ObserverWriter. Since this can edit/use the same box quantities as `ContributeVolumeDataToWriter` we have to keep them consistent somehow. The easiest way I saw to do that was to just grab the node lock for both so that everything is guaranteed to be consistent. Along with the likelihood that we aren't writing volume data that often during a simulation compared to the amount of work being done, the simplicity of the node lock seemed a better option than the (comparatively) small performance gain we'd get having all these different locks.

nilsdeppe: Okay, reasonable. If it shows up on profiles we can change it. Thanks for the explanation!

nilsdeppe (dismissed):
I didn't do a detailed review.

ping @wthrowe

wthrowe:
Needs a rebase.

Also, now that I think of it, if we're not using the volume data lock anymore it would be good to remove it from the initialization action.

knelli2:
@wthrowe Rebased and I removed the tag for the volume lock pointer altogether since that was the only place it was used (squashed into last commit)

wthrowe:
Did you forget to add the file?  I don't see any changes to `src/IO/Observer/Tags.hpp`.

knelli2:
Oops, only searched the repo for `Tags::VolumeDataLock` and so forgot the definition of the tag. Removed it and the test now

wthrowe (approved):

wthrowe (changes requested):
The clang-tidy failure is a reasonable complaint.  Either changing it or marking NOLINT would be fine with me.

knelli2:
@wthrowe I added a NOLINT

wthrowe (approved):

PR #6624, opened 2025-05-08
Prevent writing volume data when AhC fails

knelli2:
## Proposed changes

This is the third of 3 PRs that avoid writing volume data for a common horizon find that fails.

Adds optional dependencies to the ObserverWriter so that it is aware it needs to wait for an external message, then edits ObserveFields to send this optional dependency (not available through options so no current input files are affected). Then, finally adds a callback to the common horizon finder that will couple AhC to the ObserverWriter. The very last commit is the one that actually causes the logic to behave differently than before.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

~Depends on and includes #6622 and #6623~

### DIFF
--- a/src/IO/Observer/VolumeActions.hpp
+++ b/src/IO/Observer/VolumeActions.hpp
@@ -7,6 +7,7 @@
 #include <iterator>
 #include <mutex>
 #include <optional>
+#include <string>
 #include <unordered_map>
 
 #include "DataStructures/DataBox/DataBox.hpp"
@@ -28,6 +29,7 @@
 #include "Parallel/Info.hpp"
 #include "Parallel/Invoke.hpp"
 #include "Parallel/Local.hpp"
+#include "Parallel/NodeLock.hpp"
 #include "Parallel/ParallelComponentHelpers.hpp"
 #include "Utilities/Algorithm.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
@@ -60,16 +62,17 @@ namespace Actions {
 struct ContributeVolumeData {
   template <typename ParallelComponent, typename DbTagsList,
             typename Metavariables, typename ArrayIndex>
-  static void apply(db::DataBox<DbTagsList>& box,
-                    Parallel::GlobalCache<Metavariables>& cache,
-                    const ArrayIndex& array_index,
-                    const observers::ObservationId& observation_id,
-                    const std::string& subfile_name,
-                    const Parallel::ArrayComponentId& sender_array_id,
-                    ElementVolumeData&& received_volume_data) {
+  static void apply(
+      db::DataBox<DbTagsList>& box, Parallel::GlobalCache<Metavariables>& cache,
+      const ArrayIndex& array_index,
+      const observers::ObservationId& observation_id,
+      const std::string& subfile_name,
+      const Parallel::ArrayComponentId& sender_array_id,
+      ElementVolumeData&& received_volume_data,
+      const std::optional<std::string>& dependency = std::nullopt) {
     db::mutate<Tags::TensorData, Tags::ContributorsOfTensorData>(
         [&array_index, &cache, &received_volume_data, &observation_id,
-         &sender_array_id, &subfile_name](
+         &sender_array_id, &subfile_name, &dependency](
             const gsl::not_null<std::unordered_map<
                 observers::ObservationId,
                 std::unordered_map<Parallel::ArrayComponentId,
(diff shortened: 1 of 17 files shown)


src/IO/Observer/VolumeActions.hpp
```diff
@@ -350,17 +360,140 @@ struct ContributeVolumeDataToWriter {
...
           observations_registered_with_id) {
-        perform_write = true;
-        volume_data = std::move(all_volume_data[observation_id]);
+        // Check if
+        //  1. there is an external dependencies
+        if (dependency.has_value()) {
+          //  2. if there is, that we have received something at this time
+          //  3. that the dependencies are the same
+          if (box_dependencies.contains(observation_id) and
+              box_dependencies.at(observation_id).first == dependency.value()) {
```
wthrowe: What does it mean if the `contains` is true but the equality is false?  Should that be an error?

src/ParallelAlgorithms/Events/ObserveFields.hpp
```diff
@@ -181,13 +182,15 @@ class ObserveFields<VolumeDim, tmpl::list<Tensors...>,
...
+  // The dependency isn't available through Options
   ObserveFields(
       const std::string& subfile_name,
       FloatingPointType coordinates_floating_point_type,
       const std::vector<FloatingPointType>& floating_point_types,
       const std::vector<std::string>& variables_to_observe,
       std::optional<std::vector<std::string>> active_block_or_block_groups = {},
       std::optional<Mesh<VolumeDim>> interpolation_mesh = {},
+      std::optional<std::string> dependency = {},
       const Options::Context& context = {});
```
wthrowe, line 194: This breaks passing the context.  I think you need to define another constructor.

wthrowe (changes requested):

src/IO/Observer/VolumeActions.hpp
knelli2: Oh this was a remnant from when I allowed there to be multiple dependencies for a single observation ID. I decided against that since we just don't need to allow multiple dependencies at the moment so I'll remove this when I rebase.

knelli2:
@wthrowe Rebased and no longer dependent. Not sure what's going on with the documentation CI though. I didn't edit that file, nor did I touch the `[ ]` in the test file.

tests/Unit/ParallelAlgorithms/ApparentHorizonFinder/Test_ApparentHorizonFinder.cpp
```diff
@@ -774,6 +826,24 @@ void test_apparent_horizon(const gsl::not_null<size_t*> test_horizon_called,
...
+    const size_t expected_number_of_actions =
+        MakeHorizonFinderFailOnPurpose ? 1 : 3;
+    // Two nodes
+    for (int node = 0; node < 2; node++) {
+      CHECK(ActionTesting::number_of_queued_threaded_actions<
+                obs_writer_component>(runner, node) ==
+            expected_number_of_actions);
+      for (size_t i = 0; i < expected_number_of_actions; i++) {
+        // Set before we check it
+        expected_write_volume_data = !MakeHorizonFinderFailOnPurpose;
```
wthrowe: `not`

tests/Unit/ParallelAlgorithms/ApparentHorizonFinder/Test_ApparentHorizonFinder.cpp
```diff
@@ -831,49 +901,53 @@ SPECTRE_TEST_CASE("Unit.NumericalAlgorithms.Interpolator.ApparentHorizonFinder",
 
   // Time-dependent tests in distorted frame using both translation and shape
   // maps.
+  const std::optional<std::string> dependency{"FakeVolumeSubfile"};
   tmpl::for_each<tmpl::list<std::true_type, std::false_type>>(
-      [](auto use_shape_map) {
+      [&](auto use_shape_map) {
```
wthrowe, line 906: Re doxygen: It is broken because the snippet commands are split across several lines.  Doxygen was previously interpreting the empty lambda captures as snippet boundaries, but this PR removed one of them.  (Details seem to be doxygen-version dependent: locally I see the captures suppressing the warning, but not actually causing the include, while the deployed version accepts them fully.)

wthrowe (changes requested):

knelli2:
Squashed in the `not` and added a new commit that puts the snippets on a single line

wthrowe (approved):

PR #6630, opened 2025-05-13
AMR: add keep_coarse_grids option

nilsvu:
## Proposed changes

Makes multigrid work with AMR. To do this, adds the compile-time option `keep_coarse_grids` to AMR, which makes AMR create a new grid at each AMR step with an incremented grid index and keep the old grid around. Then, multigrid can use this grid hierarchy.

### Upgrade instructions

In new executables, add the following line to `Metavariables::amr` to keep the standard AMR behavior:
```cpp
    static constexpr bool keep_coarse_grids = false;
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
+++ b/src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
@@ -30,6 +30,7 @@
 #include "Parallel/Phase.hpp"
 #include "Parallel/Printf/Printf.hpp"
 #include "Parallel/Tags/DistributedObjectTags.hpp"
+#include "Parallel/Tags/Section.hpp"
 #include "ParallelAlgorithms/Amr/Actions/CreateChild.hpp"
 #include "ParallelAlgorithms/Amr/Actions/CreateParent.hpp"
 #include "ParallelAlgorithms/Amr/Projectors/Mesh.hpp"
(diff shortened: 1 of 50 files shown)


src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
```diff
@@ -183,94 +198,91 @@ struct AdjustDomain {
...
-              } catch (std::exception& e) {
-                ERROR("Error in AMR projector '"
-                      << pretty_type::get_name<projector>() << "':\n"
-                      << e.what());
-              }
-            });
+      // Determine new neighbors and update the Element
+      const auto& amr_info_of_neighbors =
+          db::get<amr::Tags::NeighborInfo<volume_dim>>(box);
+      auto [new_neighbors, new_neighbor_meshes] =
```
kidder, line 227: please add a comment to neighbors_of_child that states it can also be used to determine the new neighbors in this situation as well if passed in the appropriate arguments.

kidder (changes requested):
you can squash the change

src/ParallelAlgorithms/Amr/Actions/AdjustDomain.hpp
nilsvu: fixed 👍

kidder (approved):

PR #6638, opened 2025-05-21
Add callback to dump scalar variables together with metric variables on worldtube

guilara:
## Proposed changes

Write CurvedScalarWave variables together with the BondiSachs quantities for Cce.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Executables/ScalarTensor/EvolveScalarTensorSingleBlackHole.hpp
+++ b/src/Evolution/Executables/ScalarTensor/EvolveScalarTensorSingleBlackHole.hpp
@@ -15,6 +15,7 @@
 #include "Domain/Structure/ObjectLabel.hpp"
 #include "Evolution/Actions/RunEventsAndTriggers.hpp"
 #include "Evolution/Executables/ScalarTensor/ScalarTensorBase.hpp"
+#include "Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp"
 #include "Evolution/Systems/ScalarTensor/Actions/SetInitialData.hpp"
 #include "Options/FactoryHelpers.hpp"
 #include "Options/Protocols/FactoryCreation.hpp"
@@ -137,14 +138,34 @@ struct EvolutionMetavars : public ScalarTensorTemplateBase<EvolutionMetavars> {
   static constexpr bool use_control_systems =
       tmpl::size<control_systems>::value > 0;
 
+  struct BondiSachs;
+
   using interpolation_target_tags = tmpl::push_back<
       control_system::metafunctions::interpolation_target_tags<control_systems>,
-      AhA, ExcisionBoundaryA, SphericalSurface>;
+      AhA, ExcisionBoundaryA, SphericalSurface, BondiSachs>;
   using interpolator_source_vars = ::ah::source_vars<volume_dim>;
 
   using scalar_charge_interpolator_source_vars =
       detail::ObserverTags::scalar_charge_vars_to_interpolate_to_target;
+  using source_vars_no_deriv = tmpl::list<
+      gr::Tags::SpacetimeMetric<DataVector, volume_dim>,
+      gh::Tags::Pi<DataVector, volume_dim>,
+      gh::Tags::Phi<DataVector, volume_dim>, CurvedScalarWave::Tags::Psi,
+      CurvedScalarWave::Tags::Pi, CurvedScalarWave::Tags::Phi<volume_dim>,
+      gr::Tags::Lapse<DataVector>, gr::Tags::Shift<DataVector, volume_dim>>;
 
+  struct BondiSachs : tt::ConformsTo<intrp::protocols::InterpolationTargetTag> {
+    static std::string name() { return "BondiSachsInterpolation"; }
+    using temporal_id = ::Tags::Time;
+    using vars_to_interpolate_to_target = source_vars_no_deriv;
+    using compute_target_points =
+        intrp::TargetPoints::Sphere<BondiSachs, ::Frame::Inertial>;
+    using post_interpolation_callbacks = tmpl::list<
+        intrp::callbacks::DumpBondiSachsOnWorldtube<BondiSachs, true>>;
+    using compute_items_on_target = tmpl::list<>;
+    template <typename Metavariables>
+    using interpolating_component = typename Metavariables::st_dg_element_array;
+  };
   // The interpolator_source_vars need to be the same in both the
   // Interpolate event and the InterpolateWithoutInterpComponent event.  The
   // Interpolate event interpolates to the horizon, and the
@@ -161,16 +182,18 @@ struct EvolutionMetavars : public ScalarTensorTemplateBase<EvolutionMetavars> {
                         LtsTimeStepper>,
             tmpl::pair<LtsTimeStepper,
                        TimeSteppers::monotonic_lts_time_steppers>>,
-        tmpl::pair<Event,
-                   tmpl::flatten<tmpl::list<
-                       intrp::Events::Interpolate<volume_dim, AhA,
-                                                  interpolator_source_vars>,
-                       intrp::Events::InterpolateWithoutInterpComponent<
-                         volume_dim, ExcisionBoundaryA,
-                           interpolator_source_vars>,
-                       intrp::Events::InterpolateWithoutInterpComponent<
-                           volume_dim, SphericalSurface,
-                           scalar_charge_interpolator_source_vars>>>>,
+        tmpl::pair<
+            Event,
+            tmpl::flatten<tmpl::list<
+                intrp::Events::Interpolate<volume_dim, AhA,
+                                           interpolator_source_vars>,
+                intrp::Events::InterpolateWithoutInterpComponent<
+                    3, BondiSachs, source_vars_no_deriv>,
+                intrp::Events::InterpolateWithoutInterpComponent<
+                    volume_dim, ExcisionBoundaryA, interpolator_source_vars>,
+                intrp::Events::InterpolateWithoutInterpComponent<
+                    volume_dim, SphericalSurface,
+                    scalar_charge_interpolator_source_vars>>>>,
         tmpl::pair<DenseTrigger,
                    control_system::control_system_triggers<control_systems>>>;
   };
--- a/src/Evolution/Systems/Cce/BoundaryData.hpp
+++ b/src/Evolution/Systems/Cce/BoundaryData.hpp
@@ -566,34 +566,65 @@ void du_j_worldtube_data(
     const Scalar<SpinWeighted<ComplexDataVector, 0>>& bondi_r,
     const tnsr::I<ComplexDataVector, 2, Frame::RadialNull>& dyad);
 
+/*!
+ * \brief Convert real scalar to complex spinweighted scalar
+ */
+void klein_gordon_psi_worldtube_data(
+    gsl::not_null<Scalar<SpinWeighted<ComplexDataVector, 0>>*> kg_psi,
+    const Scalar<DataVector>& csw_psi);
+
+/*!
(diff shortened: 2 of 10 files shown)


guilara:
@knelli2 I think this is ready for review

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -168,50 +192,136 @@ struct DumpBondiSachsOnWorldtube
...
+
+          recorder.append_modal_data<spin>(
+              Cce::dataset_label_for_tag<typename tag::tag>(), time,
+              bondi_nodal_data, l_max);
+        });
       }
+    } else {
 
-      offset += num_points_single_sphere;
+      const auto& all_csw_psi = get<CurvedScalarWave::Tags::Psi>(all_gh_vars);
```
knelli2: Instead of copying a lot of code when writing klein gordon, just interject where needed. So the rough outline for this whole function would be
```cpp
const tnsr::aa<DataVector, 3, ::Frame::Inertial> spacetime_metric;
const tnsr::aa<DataVector, 3, ::Frame::Inertial> pi;
const tnsr::iaa<DataVector, 3, ::Frame::Inertial> phi;
// Put KG vars here like so
const Scalar<DataVector> csw_psi;
const Scalar<DataVector> csw_pi;
const tnsr::i<DataVector, 3, ::Frame::Inertial> csw_phi;
const Scalar<DataVector> lapse;
const tnsr::I<DataVector, 3, ::Frame::Inertial> shift;
// In case inclue_klein_gordon is false
(void)csw_psi;
(void)csw_pi;
(void)csw_phi;
(void)lapse;
(void)shift;

for (radius : radii) {
  // Regular GH stuff stays as is

  if constexpr (include_klein_gordon) {
    // Grab all the variables you need and set the non-owning stuff as needed

    // Call Cce::create_klein_gordon_boundary_data with necessary vars
  }
  
  // Write data as usual with the proper `include_klein_gordon` template param
}
```

This should cut down on the changes significantly in this file

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -22,6 +22,7 @@
 #include "Evolution/Systems/Cce/OptionTags.hpp"
 #include "Evolution/Systems/Cce/Tags.hpp"
 #include "Evolution/Systems/Cce/WorldtubeModeRecorder.hpp"
+#include "Evolution/Systems/CurvedScalarWave/Tags.hpp"
```
knelli2: We don't want to have the Cce library depend on CSW. Please forward declare the tags instead.

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
```diff
@@ -74,30 +82,44 @@ namespace intrp::callbacks {
...
  * - `Cce::Tags::BondiW`
  *
- * \note For all real quantities (Beta, DuR, R, W) we omit writing the
- * negative m modes, and the imaginary part of the m = 0 mode.
+ * If IncludeKleinGordon is true, also writes:
+ * - `Cce::Tags::KleinGordonPsi`
+ * - `Cce::Tags::KleinGordonPi`
+ *
+ * \note For all real quantities (Beta, DuR, R, W) (as well as Psi an Pi if
+ * included) we omit writing the negative m modes, and the imaginary part of the
```
knelli2: `as well as the Klein-Gordon Psi and Pi`

src/Evolution/Executables/ScalarTensor/EvolveScalarTensorSingleBlackHole.hpp
```diff
@@ -137,14 +138,35 @@ struct EvolutionMetavars : public ScalarTensorTemplateBase<EvolutionMetavars> {
...
+  struct BondiSachs : tt::ConformsTo<intrp::protocols::InterpolationTargetTag> {
+    static std::string name() { return "BondiSachsInterpolation"; }
+    using temporal_id = ::Tags::Time;
+    using vars_to_interpolate_to_target = kg_source_vars_no_deriv;
+    using compute_target_points =
+        intrp::TargetPoints::Sphere<BondiSachs, ::Frame::Inertial>;
+    using post_interpolation_callbacks = tmpl::list<
+        intrp::callbacks::DumpBondiSachsOnWorldtube<BondiSachs, true>>;
+    // May add here the computation of dt Psi
+    using compute_items_on_target = tmpl::list<>;
```
knelli2: Do you need the computation of dt Psi?

src/Evolution/Systems/Cce/BoundaryData.hpp
```diff
@@ -566,19 +566,48 @@ void du_j_worldtube_data(
...
 
+/*!
+ * \brief Compute the time derivative of the scalar for cce
+ */
+void klein_gordon_psi_worldtube_data(
+    gsl::not_null<Scalar<SpinWeighted<ComplexDataVector, 0>>*> kg_psi,
+    const Scalar<DataVector>& csw_psi);
+
+/*!
+ * \brief Compute the time derivative of the scalar for cce
```
knelli2: Are these both time derivatives? It doesn't seem like it

src/Evolution/Systems/Cce/BoundaryData.hpp
```diff
@@ -587,13 +616,52 @@ using characteristic_worldtube_boundary_tags = db::wrap_tags_in<
...
+                 tmpl::conditional_t<IncludeKleinGordon, csw_worldtube_tags,
+                                     tmpl::list<>>>>;
+
+/*!
+ * \brief The collection of tags mutated by `create_bondi_boundary_data`
+ *
+ * \details This list is used in the evolution of `CharacteristicExtract`
+ */
+template <template <typename> class BoundaryPrefix>
+using klein_gordon_characteristic_worldtube_boundary_tags =
```
knelli2: I don't think `create_bondi_boundary_data` mutates these tags

src/Evolution/Systems/Cce/BoundaryData.hpp
```diff
@@ -587,13 +616,52 @@ using characteristic_worldtube_boundary_tags = db::wrap_tags_in<
...
+
+/*!
+ * \brief The collection of tags for worldtube quantities that need to be
+ * written to disk during the Cauchy evolution that the `CharacateristicExtract`
+ * need.
+ *
+ * \details This list is used when writing Bondi quantities to disk.
+ */
+template <template <typename> class BoundaryPrefix = Cce::Tags::BoundaryValue>
+using klein_gordon_worldtube_boundary_tags_for_writing =
```
knelli2: Add to the dox that this is for the combined KG system

tests/Unit/Evolution/Systems/Cce/Test_DumpBondiSachsOnWorldtube.cpp
```diff
@@ -235,6 +245,154 @@ void test(const std::string& filename_prefix,
   }
 }
 
+// Case when Klein Gordon scalar variables are also dumped on the worldtube
+void test_kg(const std::string& filename_prefix,
+                const std::vector<double>& radii) {
```
knelli2: Again, instead of copying, just template the original test function on `IncludeKleinGordon` and interject `if constexpr`s where needed.

knelli2 (changes requested):
I would split this commit into several others because there are way too many disjoint things happening at the same time. The commits I'd do are

- Changes to `BoundaryData.?pp` and corresponding tests
- Changes to `DumpBondiSachs` and corresponding test
- Adding the BondiSachs target to the CSW exec (plus input file)

guilara:
I've cleaned this up addressing the changes requested. Waiting for the tests to run

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
guilara: Sorry what does `(void)csw_psi;` do?

I've changed it to look like this. Kept `all_csw_psi` inside the `if constexpr` .

src/Evolution/Systems/Cce/WorldtubeModeRecorder.cpp
```diff
@@ -64,6 +64,16 @@ std::string dataset_label_for_tag<Cce::Tags::Du<Cce::Tags::BondiR>>() {
...
 }
 
+template <>
+std::string dataset_label_for_tag<Cce::Tags::KleinGordonPsi>() {
+  return "KGPsi";
+}
+
+template <>
+std::string dataset_label_for_tag<Cce::Tags::KleinGordonPi>() {
+  return "dtKGPsi";
```
knelli2, line 74: Question: Why not name this KGPi instead?

tests/Unit/Evolution/Systems/Cce/Test_DumpBondiSachsOnWorldtube.cpp
```diff
@@ -164,22 +178,42 @@ void test(const std::string& filename_prefix,
...
+      const auto& csw_psi =
+          get<CurvedScalarWave::Tags::Psi>(single_spacetime_variables);
+      const auto& csw_pi =
+          get<CurvedScalarWave::Tags::Pi>(single_spacetime_variables);
+      const auto& csw_phi =
+          get<CurvedScalarWave::Tags::Phi<3>>(single_spacetime_variables);
+      const auto& lapse =
+          get<gr::Tags::Lapse<DataVector>>(single_spacetime_variables);
+      const auto& shift =
+          get<gr::Tags::Shift<DataVector, 3>>(single_spacetime_variables);
```
knelli2, line 216: I think some of these quantities are non-zero, but aren't set in the `make_spacetime_variables` function above.

knelli2 (changes requested):

src/Evolution/Systems/Cce/Callbacks/DumpBondiSachsOnWorldtube.hpp
knelli2: It "uses" the variable when `include_klein_gordon` is false so that the compiler doesn't complain about unused variables. But casting to void doesn't actually do anything.

src/Evolution/Systems/Cce/WorldtubeModeRecorder.cpp
guilara: Well @Sizheng-Ma uses the simple derivative of the scalar, which is different from the Csw Pi so it would be better to avoid confusion. But I also remember this is the variable name the KG Cce expects.

tests/Unit/Evolution/Systems/Cce/Test_DumpBondiSachsOnWorldtube.cpp
guilara: I think I found the mistake. I had the `offset += num_points_single_sphere;` line in the wrong place in `DumpBondiSachsOnWorldtube.hpp`.

src/Evolution/Systems/Cce/WorldtubeModeRecorder.cpp
knelli2: Ah yes the KG CCE does expect this name. Ok then!

knelli2 (approved):
Looks good, you can squash! Remember to not squash in the "fix" commit messages though

guilara:
Alright!

knelli2:
Oh sorry @guilara, I wasn't clear. I meant to keep the 3 separate commits and only squash the fixup commits into their respective commits

guilara:
Alright, no worries. Fixed that

PR #6642, opened 2025-05-22
Add Strahlkorper power monitor

geoffrey4444:
## Proposed changes

Add a function in `StrahlkorperFunctions.cpp` that takes a Strahlkorper and computes a power monitor, using the same method as SpEC does in SpEC's `SurfacePowerDiagnostics.cpp`. This function is an ingredient for the "Shape" criteria for adaptive horizon finding. The existing `power_monitor` functionality in `NumericalAlgorithms/LinearOperators/PowerMonitors.?pp`  works on meshes, but not on Strahlkorpers.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.cpp
+++ b/src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.cpp
@@ -3,6 +3,7 @@
 
 #include "NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp"
 
+#include <algorithm>
 #include <deque>
 #include <utility>
 
@@ -11,6 +12,7 @@
 #include "DataStructures/Tensor/TypeAliases.hpp"
 #include "NumericalAlgorithms/FiniteDifference/NonUniform1D.hpp"
 #include "NumericalAlgorithms/Interpolation/LinearLeastSquares.hpp"
+#include "NumericalAlgorithms/SphericalHarmonics/SpherepackIterator.hpp"
 #include "NumericalAlgorithms/SphericalHarmonics/Strahlkorper.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
@@ -500,6 +502,29 @@ void time_deriv_of_strahlkorper(
   time_deriv->coefficients() = std::move(new_coefficients);
 }
 
+template <typename Frame>
+void power_monitor(gsl::not_null<DataVector*> result,
+                   const Strahlkorper<Frame>& strahlkorper) {
+  result->destructive_resize(strahlkorper.l_max() + 1);
+  *result = 0.0;
+  for (SpherepackIterator it(strahlkorper.l_max(), strahlkorper.m_max()); it;
+       ++it) {
+    (*result)[it.l()] += square(strahlkorper.coefficients()[it()]);
+  }
+  for (size_t l = 0; l <= strahlkorper.l_max(); ++l) {
+    const double normalization =
+        2.0 * static_cast<double>(std::min(l, strahlkorper.m_max())) + 1.0;
+    (*result)[l] = sqrt((*result)[l] / normalization);
+  }
+}
+
+template <typename Frame>
+DataVector power_monitor(const Strahlkorper<Frame>& strahlkorper) {
+  DataVector result{strahlkorper.l_max() + 1, 0.0};
+  power_monitor(make_not_null(&result), strahlkorper);
+  return result;
+}
+
 #define FRAME(data) BOOST_PP_TUPLE_ELEM(0, data)
 
 #define INSTANTIATE(_, data)                                                  \
@@ -615,7 +640,12 @@ void time_deriv_of_strahlkorper(
       const std::vector<Strahlkorper<FRAME(data)>>& strahlkorpers);           \
   template void ylm::time_deriv_of_strahlkorper(                              \
       const gsl::not_null<Strahlkorper<FRAME(data)>*>,                        \
-      const std::deque<std::pair<double, Strahlkorper<FRAME(data)>>>&);
+      const std::deque<std::pair<double, Strahlkorper<FRAME(data)>>>&);       \
+  template void ylm::power_monitor(                                           \
+      gsl::not_null<DataVector*> result,                                      \
+      const Strahlkorper<FRAME(data)>& strahlkorper);                         \
+  template DataVector ylm::power_monitor(                                     \
+      const Strahlkorper<FRAME(data)>& strahlkorper);
 
 GENERATE_INSTANTIATIONS(INSTANTIATE,
                         (Frame::Distorted, Frame::Grid, Frame::Inertial))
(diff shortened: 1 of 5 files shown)


geoffrey4444:
The macOS arm64 unit test failure `tests/Unit/Visualization/Python/Test_TransformVolumeData.py` looks unrelated to this PR to me

src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp
```diff
@@ -397,4 +397,47 @@ void time_deriv_of_strahlkorper(
...
     const std::deque<std::pair<double, Strahlkorper<Frame>>>&
         previous_strahlkorpers);
+
+/// @{
+/*!
+ * \brief Compute the power monitor of a Strahlkorper
+ *
+ * \details Computes a Strahlkorper represented as a spectral expansion
+ * in scalar spherical harmonics \f$Y_{lm}(\theta,\phi)\f$ up to
+ * \f$l=l_{\rm max}\f$ and \f$m=m_{\rm max}\f$. Then this function
```
knelli2: Use the more modern method for equations in dox, which is just `$Y_{lm}(\theta,\phi)$` (no `\f`)

src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp
```diff
@@ -397,4 +397,47 @@ void time_deriv_of_strahlkorper(
...
+ * computes the power in each \f$l\f$. If
+ * a Strahlkorper's radius \f$r(\theta,\phi)\f$ is expanded as
+ * \f[
+ * r(\theta,\phi) = \sum_{l=0}^{l_{\rm max}}
+ *   \sum_{m=-\mathcal{M}}^{\mathcal{M}} C_{lm} Y_{lm}(\theta,\phi) + \ldots,
+ * \f]
```
knelli2: Here, you can also just use
```tex
\begin{equation}
...
\end{equation}
```

Also, I think you can remove `\ldots`?

knelli2 (commented):
Looks good. You can squash in the dox changes.

Don't worry about the macOS tests. It's something unrelated that we haven't looked into yet.

src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp
geoffrey4444: I did not know this was a thing...thanks! I was following [our instructions for writing good documentation](https://spectre-code.org/writing_good_dox.html#autotoc_md213) which don't yet have these improvements

geoffrey4444:
Thanks for the quick review, Kyle! Just squashed my docs changes, as you requested

src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.cpp
```diff
@@ -500,6 +501,30 @@ void time_deriv_of_strahlkorper(
   time_deriv->coefficients() = std::move(new_coefficients);
 }
 
+template <typename Frame>
+void power_monitor(gsl::not_null<DataVector*> result,
+                   const Strahlkorper<Frame>& strahlkorper) {
+  *result = DataVector(strahlkorper.l_max() + 1, 0.0);
```
nilsvu: use `destructive_resize()`

src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.cpp
```diff
@@ -500,6 +501,30 @@ void time_deriv_of_strahlkorper(
...
+  for (SpherepackIterator it(strahlkorper.l_max(), strahlkorper.m_max()); it;
+       ++it) {
+    (*result)[it.l()] += square(strahlkorper.coefficients()[it()]);
+  }
+  for (size_t l = 0; l <= strahlkorper.l_max(); ++l) {
+    const double normalization =
+        2.0 * static_cast<double>(
+                  l < strahlkorper.m_max() ? l : strahlkorper.m_max()) +
```
nilsvu: is this `std::min(l, m_max)`?

src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp
```diff
@@ -397,4 +397,47 @@ void time_deriv_of_strahlkorper(
...
+ * selecting the angular resolution $l_{\rm max}$ when finding
+ * apparent horizons. But note that because internally the Strahlkorper is
+ * represented as a Spherepack expansion, the implementation of this function
+ * uses that expansion rather than a literal implementation of the
+ * above equations. As a result, it's the Spherepack spectral coefficients
+ * that get squared and (for each $l$) summed over.
+ * \param strahlkorper The Strahlkorper surface.
+ */
+template <typename Frame>
+DataVector power_monitor(const Strahlkorper<Frame>& strahlkorper);
```
nilsvu, line 432: Please return `ModalVector`. We should have the DG power monitors also do that. (not sure if we have bound ModalVector in Python yet)

nilsvu (changes requested):

src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp
knelli2: We should be consistent with what the DG power monitors do right now. So I'd say either leave this as DataVector and switch the power monitor return type in a separate PR, or add a separate commit that does the switch before adding this power monitor. I prefer the former as it's fairly unrelated to this PR.

geoffrey4444: I used DataVector for consistency with AMR, so I could potentially reuse AMR functions. I'd rather change mesh and domain power monitors to `ModalVector` all at once in a separate PR

geoffrey4444:
Thanks for the review @nilsvu ... I pushed and squashed the two small changes you requested but (as I said above) would prefer to put switching this and the existing AMR power monitors to modal vectors in a different PR

nilsvu (approved):

knelli2 (approved):

src/NumericalAlgorithms/SphericalHarmonics/StrahlkorperFunctions.hpp
geoffrey4444: I'm also concerned that `ModalVector` doesn't support as many math operations as DataVector. Are we sure that we even can change the power monitors for AMR to ModalVectors without just having to convert them back to do whatever operations we do on them? I think we should discuss on a future spectre call

nilsvu: We can leave things as DataVector in this PR. It doesn't matter very much, so let's not worry about it too much.

Power monitors are modes, so they should be stored as ModalVector, so they are not confused as data on grid points (that's a DataVector). The math operations are only restricted to avoid the mixing between modal and nodal data.

PR #6647, opened 2025-05-24
Add ramp up function

guilara:
## Proposed changes

This function helps turn on smoothly the coupling terms at the start of a simulation.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/ScalarTensor/RampUpFunction.cpp
+++ b/src/PointwiseFunctions/ScalarTensor/RampUpFunction.cpp
@@ -1,0 +1,32 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "PointwiseFunctions/ScalarTensor/RampUpFunction.hpp"
+
+#include "Utilities/ConstantExpressions.hpp"
+
+double ScalarTensor::nonic_ramp_function(const double time,
+                                         const double start_time,
+                                         const double ramp_time) {
+  double ramp_factor = 1.0;
+
+  if (time < start_time) {
+    // Before ramping on
+    ramp_factor = 0.0;
+  } else if (time < ramp_time + start_time) {
+    // Ramping on with ramp function given in Eq. B4 of 1911.02588
+    const double t = (time - start_time) / ramp_time;
+    // Evaluate the polynomial using Horner's method
+    ramp_factor =
+        pow<5>(t) * (126. + t * (-420. + t * (540. + t * (-315. + 70. * t))));
+  }
+  // Otherwise the ramp_factor is 1.0 (the default value set above)
+  return ramp_factor;
+}
+
+double ScalarTensor::nonic_ramp_function(
+    const double time, const std::pair<double, double> start_and_ramp_times) {
+  const double start_time = start_and_ramp_times.first;
+  const double ramp_time = start_and_ramp_times.second;
+  return ScalarTensor::nonic_ramp_function(time, start_time, ramp_time);
+}
--- a/src/PointwiseFunctions/ScalarTensor/RampUpFunction.hpp
+++ b/src/PointwiseFunctions/ScalarTensor/RampUpFunction.hpp
@@ -1,0 +1,27 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <utility>
+
+namespace ScalarTensor {
+/// @{
+/*!
+ * \brief Function to smoothly turn on the coupling terms.
+ *
+ * \details Implements Eq. B4 of \cite Okounkova:2019zjf :
+ * \begin{align}
+ *    t_*  &\equiv (t - t_\mathrm{s})/t_\mathrm{ramp} \,, \\
+ *    F(t) &= t_*^5  (126 + t_* (-420 + t_* (540 + t_* (-315 + 70 t_*))))~,
+ * \end{align}
+ * where $ t_\mathrm{s} $ and $ t_\mathrm{ramp} $ are parameters that
+ control the start time and duration of the turn-on period.
+ *
+ */
+double nonic_ramp_function(double time, double start_time, double ramp_time);
+
+double nonic_ramp_function(double time,
+                             std::pair<double, double> start_and_ramp_times);
+/// @}
+}  // namespace ScalarTensor
--- a/src/PointwiseFunctions/ScalarTensor/SourceTags.hpp
+++ b/src/PointwiseFunctions/ScalarTensor/SourceTags.hpp
@@ -1,0 +1,58 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/Systems/ScalarTensor/Tags.hpp"
+#include "Options/String.hpp"
+
+/// \cond
+namespace ScalarTensor::OptionTags {
+struct Group;
+}  // namespace ScalarTensor::OptionTags
+/// \endcond
+
+namespace ScalarTensor {
+namespace OptionTags {
+/*!
+ * \brief Start time for ramp up function.
+ */
+struct RampUpStart {
+  static std::string name() { return "RampUpStart"; }
+  using type = double;
+  static constexpr Options::String help{"Start time for ramp up function"};
+  using group = ::ScalarTensor::OptionTags::Group;
+};
+
+/*!
+ * \brief Start time for ramp up function.
+ */
+struct RampUpDuration {
+  static std::string name() { return "RampUpDuration"; }
+  using type = double;
+  static constexpr Options::String help{"Duration time for ramp up function"};
+  using group = ::ScalarTensor::OptionTags::Group;
+};
+
+}  // namespace OptionTags
+
+namespace Tags {
+/*!
+ * \brief Start and duration time for ramp up function.
+ */
+struct RampUpParameters : db::SimpleTag {
+  using type = std::pair<double, double>;
+  using option_tags =
+      tmpl::list<OptionTags::RampUpStart, OptionTags::RampUpDuration>;
+  static constexpr bool pass_metavariables = false;
+  static std::pair<double, double> create_from_options(
+      const double start_time, const double duration_time) {
+    return std::pair<double, double> {start_time, duration_time};
+  }
+};
+
+}  // namespace Tags
+}  // namespace ScalarTensor
--- a/docs/References.bib
+++ b/docs/References.bib
@@ -2121,6 +2121,22 @@ archivePrefix = {arXiv},
        adsurl = {https://ui.adsabs.harvard.edu/abs/2015ApJS..219...24O}
 }
 
+@article{Okounkova:2019zjf,
+    author = "Okounkova, Maria and Stein, Leo C. and Moxon, Jordan and Scheel,
+     Mark A. and Teukolsky, Saul A.",
+    title = "{Numerical relativity simulation of GW150914 beyond general
+     relativity}",
+    eprint = "1911.02588",
+    archivePrefix = "arXiv",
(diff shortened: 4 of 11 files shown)


duetosymmetry:
@guilara why is the name `fifth_derivative_ramp_function`? The first through 4th derivatives vanish at each endpoint, but the 5th derivative does not.

guilara:
@duetosymmetry I just chose to keep the name used in spec (`spec/Utils/Math/RampFunctions.cpp`). But we can change the name. Since there is no other ramp function it could just be called ramp_function (or ramp_up_function).

duetosymmetry:
@guilara yikes, ok but I think we should not perpetuate some potential confusion. I know some people prefer ramp functions with an infinite number of vanishing derivatives (based on things like exp(-1/x^2)) so I would leave room for other ramp function names in the future.

src/PointwiseFunctions/ScalarTensor/RampUpFunction.hpp
```diff
@@ -0,0 +1,29 @@
...
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <utility>
+
+namespace ScalarTensor {
+/// @{
+/*!
+ * \brief Function to smoothly turn on the couplings terms.
```
knelli2: [nitpick] `coupling terms`

src/PointwiseFunctions/ScalarTensor/RampUpFunction.hpp
```diff
@@ -0,0 +1,29 @@
...
+ * \details Implements Eq. B4 of \cite Okounkova:2019zjf :
+ * \f{align*}
+ *    t_*  &\equiv (t - t_\mathrm{s})/t_\mathrm{ramp} \,, \\
+ *    F(t) &= t_*^5  (126 + t_* (-420 + t_* (540 + t_* (-315 + 70 t_*))))~,
+ * \f}
+ * where \f$ t_\mathrm{s} \f$ and \f$ t_\mathrm{ramp} \f$ are parameters that
+ control the start time and duration of the turn-on period.
+ *
+ */
+double fifth_derivative_ramp_function(const double time,
```
knelli2: Maybe `quintic_ramp_function`?

src/PointwiseFunctions/ScalarTensor/RampUpFunction.hpp
```diff
@@ -0,0 +1,29 @@
...
+/// @{
+/*!
+ * \brief Function to smoothly turn on the couplings terms.
+ *
+ * \details Implements Eq. B4 of \cite Okounkova:2019zjf :
+ * \f{align*}
+ *    t_*  &\equiv (t - t_\mathrm{s})/t_\mathrm{ramp} \,, \\
+ *    F(t) &= t_*^5  (126 + t_* (-420 + t_* (540 + t_* (-315 + 70 t_*))))~,
+ * \f}
+ * where \f$ t_\mathrm{s} \f$ and \f$ t_\mathrm{ramp} \f$ are parameters that
```
knelli2: For dox, just use `$t_\mathrm{s}$` and
```tex
\begin{align}
F(t) &= ...
\end{align}
```
No need for the `\f` anymore. We need to update our dev guide for the updated method of adding equations to dox

src/PointwiseFunctions/ScalarTensor/RampUpFunction.cpp
```diff
@@ -0,0 +1,31 @@
...
+  double ramp_factor = 1.0;
+
+  if (time < start_time) {
+    // Before ramping on
+    ramp_factor = 0.0;
+  } else if (time < ramp_time + start_time) {
+    // Ramping on with ramp function given in Eq. B4 of 1911.02588
+    const double t = (time - start_time) / ramp_time;
+    ramp_factor =
+        pow<5>(t) * (126. + t * (-420. + t * (540. + t * (-315. + 70. * t))));
```
knelli2, line 21: We have a function `integer_pow` in `src/Utilities/Math.hpp` which I think will be more efficient than the `pow<5>(t)` which I think just expands to `t * t * t * t * t`

src/PointwiseFunctions/ScalarTensor/SourceTags.hpp
```diff
@@ -0,0 +1,47 @@
...
+struct Group;
+}  // namespace ScalarTensor::OptionTags
+/// \endcond
+
+namespace ScalarTensor {
+namespace OptionTags {
+/*!
+ * \brief Start and duration time for ramp up function.
+ */
+struct RampUpParameters {
```
knelli2: I would make two different option tags for this. One `RampUpStart` and the other `RampUpDuration`. As it currently stands, if I see `RampUpParameters` in an input file, I'll have no clue what they mean (and I definitely won't 2 years from now)

src/PointwiseFunctions/ScalarTensor/SourceTags.hpp
```diff
@@ -0,0 +1,47 @@
...
+  using group = ::ScalarTensor::OptionTags::Group;
+};
+
+}  // namespace OptionTags
+
+namespace Tags {
+/*!
+ * \copydoc ScalarTensor::OptionTags::RampUpParameters
+ */
+struct RampUpParameters : db::SimpleTag {
```
knelli2, line 46: This can remain a single simple tag, just take both option tags as arguments

knelli2 (changes requested):
Clang-tidy errors are valid and should be fixed

nilsdeppe:
> @guilara yikes, ok but I think we should not perpetuate some potential confusion. I know some people prefer ramp functions with an infinite number of vanishing derivatives (based on things like exp(-1/x^2)) so I would leave room for other ramp function names in the future.

I'm actually worried that the temporal roll-on functions based on Gaussians are bad because they are not smooth. I think we should use functions from the sigmoid family instead, since those are truly smooth. I think in SpEC it's fairly common to try only adding terms once the function is above numerical epsilon. However, some care does need to be taken (it isn't in SpEC!) because the derivatives are wider than the function itself, so it's very reasonable to have the derivatives be above epsilon while the function value isn't. The function added here is a classic example of this not being done, so I agree with Leo, and you've already done this, having room for other functions is important. In Fig. 8 of 1911.02588 it's even visible by eye that there isn't smoothness. I worry that with sufficient accuracy this will leave artifacts in the waveforms. We even have seen this in SpEC with the Gaussian roll-on functions!

src/PointwiseFunctions/ScalarTensor/RampUpFunction.hpp
duetosymmetry: Except it's a degree 9 polynomial 😅

duetosymmetry:
> I'm actually worried that the temporal roll-on functions based on Gaussians are bad because they are not smooth.

Just to clarify, I was not talking about Gaussians — I was talking about smooth bump functions with all derivs vanishing at their endpoints, built out of things like exp(-1/x^2).

> In Fig. 8 of 1911.02588 it's even visible by eye that there isn't smoothness.

Can you clarify? This function was chosen so that 4 derivs match at each endpoint, since going past 4 derivs seems overkill if we only have 4th order convergence in time. I think that's pretty hard to see by eye, though:
image

src/PointwiseFunctions/ScalarTensor/RampUpFunction.cpp
```diff
@@ -0,0 +1,31 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "PointwiseFunctions/ScalarTensor/RampUpFunction.hpp"
+
+#include "Utilities/Math.hpp"
+
+double ScalarTensor::quintic_ramp_function(const double time,
```
duetosymmetry: Not a quintic: it's a degree 9 polynomial

src/PointwiseFunctions/ScalarTensor/RampUpFunction.hpp
knelli2: Oops, how about `nonic_ramp_function`?

duetosymmetry: Sure, as long as it's accurate ;) The polynomial comes from the incomplete Beta function (see https://duetosymmetry.com/notes/a-family-of-ramp-functions-and-the-Beta-function/ for an explanation)

src/PointwiseFunctions/ScalarTensor/RampUpFunction.hpp
```diff
@@ -8,22 +8,20 @@
...
  *
  */
-double fifth_derivative_ramp_function(const double time,
-                                      const double start_time,
-                                      const double ramp_time);
+double quintic_ramp_function(double time, double start_time, double ramp_time);
 
-double fifth_derivative_ramp_function(
-    const double time, const std::pair<double, double> start_and_ramp_times);
+double quintic_ramp_function(double time,
```
knelli2: Yeah change this to `nonic` since I miscounted the powers of t.

knelli2 (changes requested):
You can squash everything together.

I'm good with all the code changes, so if there needs to be a change to the functional form of the ramp function I'll let @nilsdeppe and @duetosymmetry go over that.

nilsdeppe:
My experience with gauge changes suggests the time scales shown are too short and so this will likely act like a discontinuity in the higher modes. That's what we at least see for gauge rolls. 

We also have 6th order convergence in time with dormand prince 5. How do you match 8 conditions with only 5 coefficients?

duetosymmetry:
You don't, you pick the correct ramp function with the desired number of matched derivatives. See https://duetosymmetry.com/notes/a-family-of-ramp-functions-and-the-Beta-function/

guilara:
These comments are interesting. I've only started using this ramp function recently for the code with backreaction (but I'm adding it now because it's simple enough) and haven't yet noticed artefacts that I could trace back to the form of this function. For now I'll pick the short name `nonic_ramp_function`.

nilsdeppe:
My counting was wrong, you have 10 coefficients, which gives you enough to match the derivatives. Fair enough. To be clear, we see artifacts in the the 4,4 mode and up, so that's what would be worth keeping an eye on. It might also not happen at all here since this isn't as direct of a change to the GH evolution as changing the gauge source function

src/PointwiseFunctions/ScalarTensor/RampUpFunction.cpp
```diff
@@ -0,0 +1,31 @@
...
+                                           const double ramp_time) {
+  double ramp_factor = 1.0;
+
+  if (time < start_time) {
+    // Before ramping on
+    ramp_factor = 0.0;
+  } else if (time < ramp_time + start_time) {
+    // Ramping on with ramp function given in Eq. B4 of 1911.02588
+    const double t = (time - start_time) / ramp_time;
+    ramp_factor = integer_pow(t, 5) *
```
nilsdeppe: Use `pow<5>(t)` since you know the exponent at compile time

src/PointwiseFunctions/ScalarTensor/RampUpFunction.cpp
```diff
@@ -0,0 +1,31 @@
...
+                  (126. + t * (-420. + t * (540. + t * (-315. + 70. * t))));
```
nilsdeppe: Can you add a comment like `// Evaluate the polynomial using Horner's method` so people don't rewrite it without thinking about it?

guilara:
@nilsdeppe I've made the changes you requested

nilsdeppe (commented):
Code LGTM, can you remove the two "Fix" statements from the commit message?

guilara:
I've fixed the commit message @nilsdeppe

knelli2 (approved):

duetosymmetry:
I know this commit has already been merged, but the ramp function implemented here would have been `smoothstep<4>()` (see [docs](https://spectre-code.org/group__UtilitiesGroup.html#gafce557042e1335e19b1bf3dbb7a75d79); which note is evaluated with `evaluate_polynomial()` which enforces using Horner's method).

PR #6648, opened 2025-05-24
Cce add volume np spin coeffs

duetosymmetry:
## Proposed changes

Add NP spin coeffs in the volume

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/NewmanPenrose.cpp
+++ b/src/Evolution/Systems/Cce/NewmanPenrose.cpp
@@ -14,6 +14,268 @@
 
 namespace Cce {
 namespace {
+void newman_penrose_alpha_impl(
+    const gsl::not_null<SpinWeighted<ComplexDataVector, -1>*> np_alpha,
+    const SpinWeighted<ComplexDataVector, 2>& bondi_j,
+    const SpinWeighted<ComplexDataVector, 3>& eth_j,
+    const SpinWeighted<ComplexDataVector, 1>& ethbar_j,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_k,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_r,
+    const SpinWeighted<ComplexDataVector, 1>& bondi_q,
+    const SpinWeighted<ComplexDataVector, 1>& eth_beta,
+    const SpinWeighted<ComplexDataVector, 0>& one_minus_y) {
+
+  // Intentionally using expressions with auto. This code is only used for
+  // observing, and should be profiled if used more often
+  const auto one_plus_k = 1. + bondi_k;
+  const auto sqrt_one_plus_k = sqrt(one_plus_k);
+  const auto q_plus_two_eth_beta = bondi_q + 2. * eth_beta;
+
+  *np_alpha =
+      one_minus_y / (32. * bondi_r)
+      * (
+          1. / sqrt_one_plus_k *
+          (
+            (square(conj(bondi_j)) * eth_j) / (bondi_k * one_plus_k)
+            +
+            1. / bondi_k *
+            ( bondi_j * conj(eth_j) + conj(bondi_j) * ethbar_j
+              - conj(ethbar_j))
+            + ( 2. * conj(bondi_j) * q_plus_two_eth_beta
+                - 3. * conj(ethbar_j) )
+          )
+          - 2. * sqrt_one_plus_k * conj(q_plus_two_eth_beta)
+        );
+}
+
+void newman_penrose_beta_impl(
+    const gsl::not_null<SpinWeighted<ComplexDataVector, +1>*> np_beta,
+    const SpinWeighted<ComplexDataVector, 2>& bondi_j,
+    const SpinWeighted<ComplexDataVector, 3>& eth_j,
+    const SpinWeighted<ComplexDataVector, 1>& ethbar_j,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_k,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_r,
+    const SpinWeighted<ComplexDataVector, 1>& bondi_q,
+    const SpinWeighted<ComplexDataVector, 1>& eth_beta,
+    const SpinWeighted<ComplexDataVector, 0>& one_minus_y) {
+
+  // Intentionally using expressions with auto. This code is only used for
+  // observing, and should be profiled if used more often
+  const auto one_plus_k = 1. + bondi_k;
+  const auto sqrt_one_plus_k = sqrt(one_plus_k);
+  const auto q_plus_two_eth_beta = bondi_q + 2. * eth_beta;
+
+  *np_beta =
+      one_minus_y / (32. * bondi_r)
+      * (
+          1. / sqrt_one_plus_k *
+          (
+            ( - square(bondi_j) * conj(eth_j) / (bondi_k * one_plus_k)
+              +
+              1. / bondi_k *
+              ( - bondi_j * conj(ethbar_j) - conj(bondi_j) * eth_j
+                + ethbar_j)
+              + ( 2. * bondi_j * conj(q_plus_two_eth_beta)
+                  - 3. * ethbar_j )
+            )
+          )
+          - 2. * sqrt_one_plus_k * q_plus_two_eth_beta
+       );
+}
+
+void newman_penrose_gamma_impl(
+    const gsl::not_null<SpinWeighted<ComplexDataVector, 0>*> np_gamma,
+    const SpinWeighted<ComplexDataVector, 2>& bondi_j,
+    const SpinWeighted<ComplexDataVector, 2>& dy_j,
+    const SpinWeighted<ComplexDataVector, 3>& eth_j,
+    const SpinWeighted<ComplexDataVector, 1>& ethbar_j,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_k,
+    const SpinWeighted<ComplexDataVector, 2>& bondi_h,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_r,
+    const SpinWeighted<ComplexDataVector, 1>& bondi_u,
+    const SpinWeighted<ComplexDataVector, 2>& eth_u,
+    const SpinWeighted<ComplexDataVector, 0>& ethbar_u,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_w,
+    const SpinWeighted<ComplexDataVector, 0>& dy_w,
+    const SpinWeighted<ComplexDataVector, 0>& exp_2_beta,
+    const SpinWeighted<ComplexDataVector, 0>& one_minus_y) {
+
+  // Intentionally using expressions with auto. This code is only used for
+  // observing, and should be profiled if used more often
+  const auto one_plus_k = 1. + bondi_k;
+
+  *np_gamma =
+    1. / ( sqrt(32.) * exp_2_beta )
+      * ( 0.5 / one_plus_k
+          * ( one_minus_y * ( 0.5 * one_minus_y / bondi_r + bondi_w )
+              * ( conj(bondi_j) * dy_j - bondi_j * conj(dy_j) )
+              + ( 2. * conj(bondi_h) * bondi_j - 2. * bondi_h * conj(bondi_j)
+                  + bondi_u * ( bondi_j * conj(eth_j)
+                                - conj(bondi_j) * ethbar_j)
+                  + conj(bondi_u) * ( bondi_j * conj(ethbar_j)
+                                      - conj(bondi_j) * eth_j )
+                )
+             )
+          + 2. * one_minus_y * dy_w
+          + ( 2. * bondi_w + bondi_j * conj(eth_u) - conj(bondi_j) * eth_u
+              + bondi_k * ( ethbar_u - conj(ethbar_u) ) )
+    );
+}
+
+void newman_penrose_epsilon_impl(
+    const gsl::not_null<SpinWeighted<ComplexDataVector, 0>*> np_epsilon,
+    const SpinWeighted<ComplexDataVector, 2>& bondi_j,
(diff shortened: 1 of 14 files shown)


duetosymmetry:
The first two commits are adding a feature to CheckOutputFiles.py so that it can call additional executables (this feature is used in the last commit, which adds tests). I can add a `pypp::check_with_random_values` regression test for the spin coefficients if desired.

duetosymmetry:
@nilsvu requested I pull the first two commits out, now in PR #6677

duetosymmetry:
Rebased to combine some commits and make it easier to review

keefemitman:
Ok I reviewed the mathematica notebook up to `Regularity preserving coordinate transformations` and did not find anything obviously wrong with the calculation of the NP scalars in numerical coordinates (though the notebook could certainly use some better documentation *cough cough Jordan cough cough*). @knelli2 let me know if you want me to just check the coded up expressions as well or if you want to handle it.

duetosymmetry:
Rebased after #6677 was merged

src/Evolution/Systems/Cce/SwshDerivatives.hpp
```diff
@@ -639,4 +640,58 @@ void mutate_all_swsh_derivatives_for_tag(
...
+  using return_type = typename base::type;
+
+  using ApplySwshJac = ApplySwshJacobianInplace<base>;
+
+  using argument_tags =
+      tmpl::append<
+    tmpl::list<Tags::LMax, Tags::NumberOfRadialPoints, ArgumentTag>,
+    typename ApplySwshJac::argument_tags,
+    typename ApplySwshJac::on_demand_argument_tags
+    >;
```
nilsdeppe: Could you run clang-format on your code? I use `git clang-format -f` regularly to format only lines I currently have changes on :)

src/Evolution/Systems/Cce/SwshDerivatives.hpp
```diff
@@ -639,4 +640,58 @@ void mutate_all_swsh_derivatives_for_tag(
...
+ * place by `ApplySwshJacobianInplace`. However, that infrastructure is specific
+ * to the Cce evolution system itself, so it won't work for other uses
+ * (e.g. observers that are invoked periodically). This compute tag instead
+ * manually computes a certain swsh derivative (at fixed $r$).  Right now this
+ * will only work when `DerivativeKind` is either `Spectral::Swsh::Tags::Eth` or
+ * `Spectral::Swsh::Tags::Ethbar`. If you want to extend it to second
+ * derivatives, there will have to be more template specializations.
+ */
+template <typename ArgumentTag, typename DerivativeKind>
+struct ManualSwshDerivativeCompute :
```
nilsdeppe: Could you add a test to `tests/Unit/Evolution/Systems/Cce/Test_SwshDerivatives.cpp` for this?

src/Evolution/Systems/Cce/Tags.hpp
```diff
@@ -374,6 +374,81 @@ struct EndTime : db::BaseTag {};
 
 struct StartTime : db::BaseTag {};
 
+/// The (adapted) Newman-Penrose spin coefficient $\alpha^{SW}$. See
+/// documentation of `newman_penrose_alpha()` for definition.
+struct NewmanPenroseAlpha : db::SimpleTag {
```
nilsdeppe, line 379: Could you add `TestHelpers::db::test_simple_tag` cases for these to `tests/Unit/Evolution/Systems/Cce/Test_Tags.cpp`?

src/Evolution/Systems/Cce/NewmanPenrose.cpp
```diff
@@ -14,6 +14,252 @@
...
+    const SpinWeighted<ComplexDataVector, 1>& ethbar_j,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_k,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_r,
+    const SpinWeighted<ComplexDataVector, 1>& bondi_q,
+    const SpinWeighted<ComplexDataVector, 1>& eth_beta,
+    const SpinWeighted<ComplexDataVector, 0>& one_minus_y) {
+
+  const auto one_plus_k = 1. + bondi_k;
+  const auto sqrt_one_plus_k = sqrt(one_plus_k);
+  const auto q_plus_two_eth_beta = bondi_q + 2. * eth_beta;
```
nilsdeppe, line 32: `auto` is dumb with expression templates. This is actually not the computed value but the _expression_ to compute the value. Can you change to `ComplexDataVector` if you want to precompute the value or add a comment that you intentionally want the expression (which gets reevaluated each use)? Same for the functions below. Because of the cost of memory allocations it's not clear to me that it would be better to precompute. One option is to use a `Variables` or `TempBuffer` to create one memory allocation that all all three use. We have temp tags that can be used in the Variables. E.g.
```cpp
#include "DataStructures/Tags/TempTensor.hpp"

Variables<tmpl::list<::Tags::TempScalar<0, SpinWeighted<ComplexDataVector, 0>>,
  ::Tags::TempScalar<1, SpinWeighted<ComplexDataVector, 0>>,
  ::Tags::TempScalar<2, SpinWeighted<ComplexDataVector, 1>>>> buffer{bondi_j.size()};
auto& [one_plus_k, sqrt_one_plus_k, q_plus_two_eth_beta] = buffer;
  one_plus_k = 1. + bondi_k;
  sqrt_one_plus_k = sqrt(one_plus_k);
  q_plus_two_eth_beta = bondi_q + 2. * eth_beta;
```

src/Evolution/Systems/Cce/NewmanPenrose.hpp
```diff
@@ -26,6 +27,695 @@ template <typename Tag>
...
+ * \end{align}
+ *
+ * It does not have a well-defined spin weight; but by using a reference
+ * spin-connection, the adapted spin coefficient $\alpha^{SW} \equiv \alpha -
+ * \alpha_{ref}$ does have a well-defined spin weight of -1. In the language
+ * of \cite Moxon2020gha, that is equivalent to setting $\Theta=0$. This
+ * gives the expression
+ *
+ * \begin{align}
+ *   \alpha^{SW} = \frac{1-y}{32R} \Bigg\{
```
nilsdeppe, line 48: For my edification, why not `\left\{`, etc.?

src/Evolution/Systems/Cce/NewmanPenrose.cpp
```diff
@@ -14,6 +14,252 @@
...
+
+  *np_beta =
+      one_minus_y / (32. * bondi_r)
+      * (
+          1. / sqrt_one_plus_k *
+          (
+            ( - bondi_j * bondi_j * conj(eth_j) / (bondi_k * one_plus_k)
```
nilsdeppe: [optional] `square(bondi_j)`

src/Evolution/Systems/Cce/NewmanPenrose.cpp
```diff
@@ -14,6 +14,252 @@
...
+  const auto q_plus_two_eth_beta = bondi_q + 2. * eth_beta;
+
+  *np_alpha =
+      one_minus_y / (32. * bondi_r)
+      * (
+          1. / sqrt_one_plus_k *
+          (
+            (conj(bondi_j) * conj(bondi_j) * eth_j) / (bondi_k * one_plus_k)
+            +
+            1. / bondi_k *
```
nilsdeppe, line 41: [optiona] here and it looks like in a few other functions I think you could factor out the `1/bondi_k` to have `1. / (sqrt_one_plus_k * bondi_k)` out front and then end up with 1 fewer divisions and one extra multiplication. Division is ~25x more expensive than multiplication on modern hardware. Since this won't be used during the evolution, performance isn't critical, but it's easier to do now rather than to discover it later :shrug:

src/Evolution/Systems/Cce/NewmanPenrose.cpp
```diff
@@ -14,6 +14,252 @@
 
 namespace Cce {
 namespace {
+void newman_penrose_alpha_impl(
+    gsl::not_null<SpinWeighted<ComplexDataVector, -1>*> np_alpha,
```
nilsdeppe: `const gsl::not_null` in the cpp. Also for functions below

src/Evolution/Systems/Cce/NewmanPenrose.cpp
```diff
@@ -14,6 +14,252 @@
...
+    gsl::not_null<SpinWeighted<ComplexDataVector, -1>*> np_nu,
+    const SpinWeighted<ComplexDataVector, 2>& bondi_j,
+    const SpinWeighted<ComplexDataVector, 0>& bondi_k,
+    const SpinWeighted<ComplexDataVector, 1>& eth_w,
+    const SpinWeighted<ComplexDataVector, 0>& exp_2_beta) {
+
+  const auto one_plus_k = 1. + bondi_k;
+  const auto sqrt_one_plus_k = sqrt(one_plus_k);
+
+  *np_nu = 1. / ( 2. * exp_2_beta ) *
```
nilsdeppe: `0.5 / exp_2_beta`??

src/Evolution/Systems/Cce/NewmanPenrose.hpp
```diff
@@ -26,6 +27,695 @@ template <typename Tag>
...
+ * It has a spin weight of -2. This gives the expression
+ *
+ * \begin{align}
+ *   \lambda = \frac{e^{-2\beta}}{4\sqrt{2}} \Bigg\{ &
+ *       \left[\frac{1-y}{R}+2W\right] \frac{1-y}{2(1+K)}
+ *       \left[ \frac{\bar{J}^2\partial_y J - \partial_y \bar{J}}{K} -
+ *         (2+K) \partial_y\bar{J} \right] \\ \nonumber
+ *       &{}+ 2(1+K)\bar{\eth}\bar{U}  + \frac{\bar{I}}{K}
+ *       +\left[ \bar{I} +2\bar{J}(\bar\eth U - \eth\bar U) \right]
+ *       + \frac{\bar{J}^2}{K(1+K)}(I + 2K\eth U)
```
nilsdeppe: I think this line and line 259 in the input file differ by a minus sign?

src/Evolution/Systems/Cce/NewmanPenrose.hpp
```diff
@@ -26,6 +27,695 @@ template <typename Tag>
...
+ * \end{align}
+ *
+ * where we have defined the temporary quantity
+ *
+ * \begin{align}
+ *   I = 2H+U\bar\eth{J}+\bar{U}\eth J
+ * \end{align}
+ *
+ */
+void newman_penrose_lambda(
```
nilsdeppe, line 388: Could you add the python tests in `Test_NewmanPenrose.cpp` and `NewmanPenrose.py` to serve as additional regression tests?

src/Evolution/Systems/Cce/NewmanPenrose.hpp
```diff
@@ -26,6 +27,695 @@ template <typename Tag>
...
+    const Scalar<SpinWeighted<ComplexDataVector, 0>>& exp_2_beta,
+    const Scalar<SpinWeighted<ComplexDataVector, 0>>& one_minus_y);
+
+namespace Tags {
+/*!
+ * \brief Compute tag for $\alpha^{SW}$ in the volume.
+ *
+ * \details See documentation of `newman_penrose_alpha()` for definition.
+ */
+struct NewmanPenroseAlphaCompute : Tags::NewmanPenroseAlpha, db::ComputeTag {
```
nilsdeppe, line 410: Could you add to `Test_NewmanPenrose.cpp` tests like
```cpp
TestHelpers::db::test_compute_tag<Tags::NewmanPenroseAlphaCompute>("NewmanPenroseAlpha");
```

nilsdeppe (changes requested):

src/Evolution/Systems/Cce/NewmanPenrose.cpp
duetosymmetry: I mainly used those to make the code easier to read and validate. No idea what is actually more computationally efficient ... I'll just follow your recommendation of whether you want a ComplexDataVector or a Temp*.

src/Evolution/Systems/Cce/NewmanPenrose.hpp
duetosymmetry: The open and close { / } are on different lines, which LaTeX can't manage

src/Evolution/Systems/Cce/NewmanPenrose.hpp
duetosymmetry: Should these go in `Test_NewmanPenrose.cpp` or `Test_Tags.cpp`?

src/Evolution/Systems/Cce/NewmanPenrose.hpp
duetosymmetry: Thanks for catching this!! I checked with the Mma code — the cpp was right, this doc was wrong.

src/Evolution/Systems/Cce/NewmanPenrose.hpp
nilsdeppe: I've done this with `\right.` and `\left.` but sizing may be weird in that case. Anyway, not a change request, I just wanted to understand :)

src/Evolution/Systems/Cce/NewmanPenrose.hpp
nilsdeppe: What you have is fine. I think putting the test in the same-named file in the test directory makes them easier to find :shrug:

src/Evolution/Systems/Cce/NewmanPenrose.cpp
nilsdeppe: I would just add a comment like `// Intentionally using expressions with auto. This code is only used for observing, and should be profiled if used more often`

nilsdeppe (approved):

PR #6649, opened 2025-05-25
Support single precision in interpolation, set initial block order

nilsvu:
### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/BlockLogicalCoordinates.cpp
+++ b/src/Domain/BlockLogicalCoordinates.cpp
@@ -16,6 +16,8 @@
 #include "Utilities/EqualWithinRoundoff.hpp"
 #include "Utilities/ErrorHandling/Error.hpp"
 #include "Utilities/GenerateInstantiations.hpp"
+#include "Utilities/Literals.hpp"
+#include "Utilities/Numeric.hpp"
 
 template <size_t Dim, typename Fr>
 std::optional<tnsr::I<double, Dim, ::Frame::BlockLogical>>
@@ -148,6 +150,12 @@ BlockLogicalCoords<Dim> block_logical_coordinates_single_point(
   // the smallest block_id).
   // In case a block_order is provided, it is no longer guaranteed that the
   // block with the smallest block_id is chosen.
+  if (block_order.has_value() and block_order.value()->empty()) {
+    // If the block order is empty, fill it with the list of blocks in the
+    // domain
+    block_order.value()->resize(domain.blocks().size());
+    alg::iota(*block_order.value(), 0_st);
+  }
   const size_t num_blocks = block_order.has_value()
                                 ? block_order.value()->size()
                                 : domain.blocks().size();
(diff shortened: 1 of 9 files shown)


src/IO/Exporter/PointwiseInterpolator.cpp
```diff
@@ -131,18 +120,34 @@ void interpolate_to_points_impl(
...
+                                            num_element_target_points);
+          const auto input_data = gsl::make_span(
+              std::get<DataVector>(component).data() + offset, length);
+          interpolant.interpolate(make_not_null(&output_data), input_data);
+          for (size_t j = 0; j < num_element_target_points; ++j) {
+            (*result)[i][points.offsets[j]] = interpolated_data_double[j];
+            (*filled_data)[points.offsets[j]] = true;
+          }
+        } else {
+          interpolated_data_float.resize(num_element_target_points);
```
nilsdeppe, line 140: Maybe only resize of size is below what you need? I'm not sure what guarantees there are on vector if you decrease size. Probably would be good to avoid reallocating inside a loop when possible

src/Domain/BlockLogicalCoordinates.cpp
```diff
@@ -148,6 +148,12 @@ BlockLogicalCoords<Dim> block_logical_coordinates_single_point(
   // the smallest block_id).
   // In case a block_order is provided, it is no longer guaranteed that the
   // block with the smallest block_id is chosen.
+  if (block_order.has_value() and block_order.value()->size() == 0) {
+    // If the block order is empty, fill it with the list of blocks in the
+    // domain
+    block_order.value()->resize(domain.blocks().size());
+    std::iota(block_order.value()->begin(), block_order.value()->end(), 0);
```
nilsdeppe: we do have `alg::iota(block_order.value(), 0);` in `Utilities/Numeric.hpp`

tests/Unit/Domain/Test_BlockAndElementLogicalCoordinates.cpp
```diff
@@ -566,14 +566,18 @@ void test_block_and_element_logical_coordinates(
...
     for (size_t d = 0; d < Dim; d++) {
       inertial_coords_double.get(d) = inertial_coords.get(d)[i];
     }
-    // Reset the block order so we always find the same block
-    std::iota(block_order.begin(), block_order.end(), 0);
+    if (i > 0) {
+      // Reset the block order so we always find the same block. In the first
+      // iteration the block order should be initially set to the list of blocks
+      // in the domain.
+      std::iota(block_order.begin(), block_order.end(), 0);
```
nilsdeppe: `alg::iota`?

nilsdeppe (changes requested):
you can squash

src/IO/Exporter/PointwiseInterpolator.cpp
nilsvu: Ok I thought `std::vector` keeps its allocation when shrinking in size. I think that's even guaranteed.

nilsdeppe: Yea, I was unable to check cppreference because it was down while I was reviewing :(

nilsvu:
Squashed in the `alg::iota` change. Thanks for reviewing !

nilsdeppe (approved):

PR #6650, opened 2025-05-25
Add coupling functions and sources for ScalarTensor

guilara:
## Proposed changes

Add scalar curvature coupling functions and sources. 

Some reorganisation: Moved more ScalarTensor source items to `PointwiseFunctions/`. Moved Elliptic items for beyond GR in `PointwiseFunctions/ScalarGaussBonnet/` to `PointwiseFunctions/ScalarTensor/Xcts/`. Created `ScalarTensor/ScalarGaussBonnet` directory to add specific things related to the theory (like the sources for the scalar and tensor fields to be added later) .

Depends on: #6647 

@PunJustice Do you mind if I move `ScalarGaussBonnet` in PointwiseFunctions to the new location (see second commit)?  I'm trying to avoid potential confusion.

### Upgrade instructions
In include statements change `ScalarGaussBonnet/` -> `ScalarTensor/Xcts/`
Renamed library `ScalarGaussBonnetPointwiseFunctions` -> `ScalarTensorXcts`, and analogously for the `Test_` libraries.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/PointwiseFunctions/ScalarTensor/ScalarSource.hpp
```diff
@@ -74,8 +76,156 @@ void add_scalar_source_to_dt_pi_scalar(
...
+    gsl::not_null<Scalar<DataVector>*> gb_scalar,
+    const tnsr::ii<DataVector, 3>& weyl_electric,
+    const tnsr::ii<DataVector, 3>& weyl_magnetic,
+    const tnsr::II<DataVector, 3>& inverse_spatial_metric);
+
+Scalar<DataVector> pontryagin_scalar_in_vacuum(
+    const tnsr::ii<DataVector, 3>& weyl_electric,
+    const tnsr::ii<DataVector, 3>& weyl_magnetic,
+    const tnsr::II<DataVector, 3>& inverse_spatial_metric);
+/// @}
```
nilsdeppe: Why not have this in `GeneralRelativity`? It's not clear why the Potryagin is specific to ScalarTensor.

src/PointwiseFunctions/ScalarTensor/ScalarSource.hpp
```diff
@@ -74,8 +76,156 @@ void add_scalar_source_to_dt_pi_scalar(
...
+ * \param weyl_magnetic_scalar Magnetic scalar \f$ B_{ab} B^{ab} \f$
+ */
+void gauss_bonnet_scalar_in_vacuum(
+    gsl::not_null<Scalar<DataVector>*> gb_scalar,
+    const Scalar<DataVector>& weyl_electric_scalar,
+    const Scalar<DataVector>& weyl_magnetic_scalar);
+
+Scalar<DataVector> gauss_bonnet_scalar_in_vacuum(
+    const Scalar<DataVector>& weyl_electric_scalar,
+    const Scalar<DataVector>& weyl_magnetic_scalar);
```
nilsdeppe: While `GeneralRelativity` is not quite the "right" place, but this also seems like it's more general than ScalarTensor. If so, maybe we move it to `GeneralRelativity` for now too?

src/PointwiseFunctions/ScalarTensor/ScalarSource.cpp
```diff
@@ -16,9 +17,128 @@ void add_scalar_source_to_dt_pi_scalar(
...
-void mass_source(
-    const gsl::not_null<Scalar<DataVector>*> scalar_source,
-    const Scalar<DataVector>& psi, const double mass_psi) {
+void mass_source(const gsl::not_null<Scalar<DataVector>*> scalar_source,
+                 const Scalar<DataVector>& psi, const double mass_psi) {
   get(*scalar_source) = square(mass_psi) * get(psi);
 }
+
+void pontryagin_scalar_in_vacuum(
+    gsl::not_null<Scalar<DataVector>*> pontryagin_scalar,
```
nilsdeppe: `const gsl`

src/PointwiseFunctions/ScalarTensor/ScalarSource.cpp
```diff
@@ -16,9 +17,128 @@ void add_scalar_source_to_dt_pi_scalar(
...
+                                         inverse_spatial_metric(ti::J, ti::K) *
+                                         weyl_magnetic(ti::k, ti::l) *
+                                         inverse_spatial_metric(ti::L, ti::I));
+}
+
+Scalar<DataVector> pontryagin_scalar_in_vacuum(
+    const tnsr::ii<DataVector, 3>& weyl_electric,
+    const tnsr::ii<DataVector, 3>& weyl_magnetic,
+    const tnsr::II<DataVector, 3>& inverse_spatial_metric) {
+  Scalar<DataVector> pontryagin_scalar{};
```
nilsdeppe: can you do `Scalar<DataVector> pontryagin_scalar{get<0, 0>(weyl_electric).size()};`?

src/PointwiseFunctions/ScalarTensor/ScalarSource.cpp
```diff
@@ -16,9 +17,128 @@ void add_scalar_source_to_dt_pi_scalar(
...
+    const Scalar<DataVector>& weyl_magnetic_scalar) {
+  // Compute the Riemann squared scalar in vacuum
+  gb_scalar->get() =
+      8.0 * (weyl_electric_scalar.get() - weyl_magnetic_scalar.get());
+}
+
+Scalar<DataVector> gauss_bonnet_scalar_in_vacuum(
+    const Scalar<DataVector>& weyl_electric_scalar,
+    const Scalar<DataVector>& weyl_magnetic_scalar) {
+  Scalar<DataVector> gb_scalar{};
```
nilsdeppe: can you do `Scalar<DataVector> gb_scalar{get(weyl_electric_scalar).size()};`?

src/PointwiseFunctions/ScalarTensor/ScalarSource.cpp
```diff
@@ -16,9 +17,128 @@ void add_scalar_source_to_dt_pi_scalar(
...
+  scalar_source->get() += square(mass_psi) * get(psi);
+}
+
+Scalar<DataVector> gauss_bonnet_scalar_source(
+    const Scalar<DataVector>& weyl_electric_scalar,
+    const Scalar<DataVector>& weyl_magnetic_scalar,
+    const Scalar<DataVector>& psi,
+    const std::array<double, 3>& coupling_parameters, const double mass_psi,
+    const std::pair<double, double> start_and_ramp_times, const double time) {
+  Scalar<DataVector> result{};
```
nilsdeppe: can you do `Scalar<DataVector> gb_scalar{get(weyl_electric_scalar).size()};`?

src/PointwiseFunctions/ScalarTensor/ScalarSource.cpp
```diff
@@ -16,9 +17,128 @@ void add_scalar_source_to_dt_pi_scalar(
...
+  // Ramp up factor
+  const double ramp_factor =
+      fifth_derivative_ramp_function(time, start_and_ramp_times);
+
+  const double linear_coupling_psi_over_four =
+      ramp_factor * linear_coupling_psi / 4.0;
+  const double first_coupling_psi_over_four =
+      ramp_factor * first_coupling_psi / 4.0;
+  const double second_coupling_psi_over_four =
+      ramp_factor * second_coupling_psi / 4.0;
```
nilsdeppe: In this case it's not a big deal, but in general, multiplying by the inverse is better because division is about 25x slower than multiplication. The compiler is not allowed to automatically do this optimization since it violates IEEE. Here it would then be `0.25 * ramp_factor * second_coupling_psi` etc.

nilsdeppe (commented):
Not a complete review, but I notice you pushed things so I wanted to post comments before anything disappeared. Ping once you've rebased and done these and I'll do a full review :)

guilara:
@nilsdeppe I hadn't noticed you commented on this until not long ago. But I've just addressed your comments. I also thought this morning of moving the curvature scalars to the namespace gr.

src/PointwiseFunctions/ScalarTensor/ScalarGaussBonnet/Tags.hpp
```diff
@@ -0,0 +1,72 @@
...
+                                 OptionTags::QuarticCouplingParameter>;
+  static constexpr bool pass_metavariables = false;
+  static std::array<double, 3> create_from_options(
+      const double linear_parameter, const double quadratic_parameter,
+      const double quartic_parameter) {
+    return std::array<double, 3>{linear_parameter, quadratic_parameter,
+                                 quartic_parameter};
+  }
+};
+
```
nilsdeppe: Why not have a struct:
```cpp
struct CouplingParameters {
  double linear{std::numeric_limits<double>::signaling_NaN()};
  double quadratic{std::numeric_limits<double>::signaling_NaN()};
  double  quartic{std::numeric_limits<double>::signaling_NaN()};

  void pup(PUP::er& p); // def'n in cpp
};
```
and use that? I worry that having an array with `N` numbers like this is more difficult to read and understand. We've had several attempts (myself included!) of trying to use things like `array` or `tuple` to denote "collection of related things" and it's always been a mistake and a struct would've been better.

Additionally, you could also turn the struct `CouplingParameters` into being option createable and then the input file would look like:
```yaml
ScalarGaussBonnet:
  CouplingParameters:
    Linear: 1.0
    Quadratic: 1.0
    Quartic: 1.0
```
If people want to add more in the future, they could use the `Auto<None>` approach and then something like
```yaml
ScalarGaussBonnet:
  CouplingParameters:
    Linear: 1.0
    Quadratic: 1.0
    Quartic: 1.0
    Quintic: None
```
would be possible. 

My suggestion would be to factor the tags into a separate PR, with the above suggestion included. Then we can merge the rest of this current PR since I think that's good to go.

tests/Unit/PointwiseFunctions/ScalarTensor/ScalarGaussBonnet/Test_ScalarSource.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include "Framework/TestCreation.hpp"
+#include "Framework/TestHelpers.hpp"
+#include "Helpers/DataStructures/DataBox/TestHelpers.hpp"
+#include "Helpers/Domain/DomainTestHelpers.hpp"
+#include "Helpers/PointwiseFunctions/GeneralRelativity/TestHelpers.hpp"
+#include "PointwiseFunctions/ScalarTensor/ScalarGaussBonnet/ScalarSource.hpp"
+
+namespace {
+namespace detail {
+// Just remove std::pair for test
```
nilsdeppe: Can this comment be removed or is there still something to do?

nilsdeppe (changes requested):
I have one bigger suggestion, as I say in the comment there, I would recommend splitting this PR into 2 so most of the code gets merged sooner :)

src/PointwiseFunctions/ScalarTensor/ScalarGaussBonnet/Tags.hpp
guilara: I like the suggestion. I have removed changes to tags here and will add them in another PR.

tests/Unit/PointwiseFunctions/ScalarTensor/ScalarGaussBonnet/Test_ScalarSource.cpp
guilara: Removed. Nothing else to do.

src/PointwiseFunctions/ScalarTensor/ScalarGaussBonnet/Tags.hpp
guilara: I've opened up another PR for the tags (#6676). But I need to rebase once this gets merged since it depends on directories created here.

nilsdeppe (approved):

PR #6155, opened 2024-07-16
Allow to read numerical initial data for ScalarTensor

guilara:
## Proposed changes

Requires:  #6154

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #6651, opened 2025-05-25
Update ScalarTensor executable

guilara:
## Proposed changes

- Add extra 3+1 quantities to compute curvature scalars.
- Disable backreaction for consistency with the test-field approximation.
- Change `ScalarSource` compute tag to include the coupling to curvature.
- Update evolution executable for single black hole (among these changes, including control systems).

Depends on: #6155 #6647 #6638 #6650  

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/ScalarTensor/Initialize.hpp
+++ b/src/Evolution/Systems/ScalarTensor/Initialize.hpp
@@ -36,8 +36,8 @@
 #include "PointwiseFunctions/GeneralRelativity/WeylMagnetic.hpp"
 #include "PointwiseFunctions/ScalarTensor/ConstraintDampingTags.hpp"
 #include "PointwiseFunctions/ScalarTensor/ConstraintGammas.hpp"
+#include "PointwiseFunctions/ScalarTensor/ScalarGaussBonnet/ScalarSource.hpp"
 #include "PointwiseFunctions/ScalarTensor/ScalarGaussBonnet/Tags.hpp"
-#include "PointwiseFunctions/ScalarTensor/ScalarSource.hpp"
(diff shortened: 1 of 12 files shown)


src/Evolution/Systems/ScalarTensor/Initialize.hpp
```diff
@@ -76,4 +85,70 @@ using scalar_tensor_3plus1_compute_tags = tmpl::list<
...
+using scalar_tensor_3plus1_compute_tags =
+    tmpl::append<scalar_tensor_basic_compute_tags<Dim, Fr>,
+                 sgb_extra_compute_tags<Dim, Fr>>;
+}  // namespace Initialization
+
+namespace Actions {
+struct InitializeGhAnd3Plus1Variables {
+  static constexpr size_t volume_dim = 3;
+  using frame = Frame::Inertial;
+  using compute_tags = db::AddComputeTags<
```
nilsdeppe, line 128: Not asking for a change, just a note: we've been trying to have fewer compute tags during the evolution (during observing is fine!) because the RAM/memory used by the compute tags is persistent. Instead, we've tried to move more (ideally all) of the computation that is only needed for the time derivative to that computation. This reduces the amount of persistent memory significantly and will be very important for GPUs where there's a lot less RAM for the amount of given compute.

nilsdeppe (approved):

src/Evolution/Systems/ScalarTensor/Initialize.hpp
guilara: Noted. Its probably not too complicated here but wouldn't do it right now because I need to work out how to move the compute tags for the full system with backreaction.

PR #6652, opened 2025-05-26
Fix bug in AH finder initial guess

knelli2:
## Proposed changes

Previously there was a bug that made us not do any extrapolation for the initial guess of the horizon finder. This PR fixes that. There's a significant reduction in the number of iterations taken:

ah_iterations

It also attempts to recover from failed interpolation on the first iteration which previously would immediately fail. I've tested that this allows the ObservationAh{A/B} horizon finder to succeed robustly now instead of failing if the time between horizon finds was too big.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe (approved):
Dang!! I think that puts us roughly at where SpEC is with number of iterations, right?

knelli2:
@nilsdeppe Yes I believe so!

nilsvu (approved):
Nice find!

PR #6665, opened 2025-05-28
Add constructor for replay FoT in FromVolumeFile

AlexCarpenter46:
## Proposed changes

Adds a constructor to replay a function of time from volume data allowing access to all the history and updates of the function of time. This is needed for the transition to ringdown translation function of time.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/Creators/TimeDependentOptions/FromVolumeFile.hpp
+++ b/src/Domain/Creators/TimeDependentOptions/FromVolumeFile.hpp
@@ -44,7 +44,9 @@ struct FromVolumeFile {
       "Read function of time coefficients from a volume subfile of an H5 file.";
 
   FromVolumeFile() = default;
-  FromVolumeFile(std::string h5_filename, std::string subfile_name);
+  // The replay option cannot be set from options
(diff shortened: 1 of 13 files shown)


src/Domain/Creators/TimeDependentOptions/FromVolumeFile.hpp
```diff
@@ -45,6 +45,8 @@ struct FromVolumeFile {
 
   FromVolumeFile() = default;
   FromVolumeFile(std::string h5_filename, std::string subfile_name);
+  FromVolumeFile(std::string h5_filename, std::string subfile_name, bool replay,
+                 const Options::Context& context = {});
```
knelli2: Oh I didn't realize earlier today that the original constructor didn't have a context. In that case, you can just add to the original constructor a `bool replay = false` and that's it. Then you don't need to check anything in the cpp.

Also when you do that, add a comment that the `replay` option cannot be set from options

knelli2 (changes requested):
You can squash in the change. Also just NOLINT the clang-tidy error

knelli2 (approved):

PR #6666, opened 2025-05-29
Add function to compute the convergence rate of a power monitor

geoffrey4444:
## Proposed changes

Add a function to compute the convergence rate of a power monitor, the same way SpEC computes this convergence rate (Eq. (53) of Szilagyi (2014)). This is necessary to implement a function to compute how many pile up modes are in a power monitor, which I plan to add in a follow-up PR rot this one.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
+++ b/src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
@@ -3,14 +3,17 @@
 
 #include "NumericalAlgorithms/LinearOperators/PowerMonitors.hpp"
 
+#include <algorithm>
 #include <array>
 #include <cmath>
+#include <limits>
 
 #include "DataStructures/ComplexDataVector.hpp"
 #include "DataStructures/ComplexModalVector.hpp"
 #include "DataStructures/DataVector.hpp"
 #include "DataStructures/ModalVector.hpp"
 #include "DataStructures/SliceIterator.hpp"
+#include "NumericalAlgorithms/Interpolation/LinearRegression.hpp"
 #include "NumericalAlgorithms/LinearOperators/CoefficientTransforms.hpp"
 #include "NumericalAlgorithms/Spectral/Mesh.hpp"
 #include "Utilities/ConstantExpressions.hpp"
@@ -51,7 +54,7 @@ void power_monitors(const gsl::not_null<std::array<DataVector, Dim>*> result,
           slice_sum += square(mode);
         }
       }
-      slice_sum /= n_slice;
+      slice_sum /= static_cast<double>(n_slice);
       slice_sum = sqrt(slice_sum);
 
       gsl::at(*result, sliced_dim)[index] = slice_sum;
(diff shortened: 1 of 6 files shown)


geoffrey4444:
To-do: ~~get rid of the not null version~~ done

src/NumericalAlgorithms/LinearOperators/PowerMonitors.hpp
```diff
@@ -119,4 +119,48 @@ template <typename VectorType, size_t Dim>
...
+ */
+double convergence_rate(const DataVector& power_monitor,
+                        size_t number_of_filtered_modes = 0);
+/*!
+ * \copydoc convergence_rate
+ * \param result The convergence rate.
+ */
+void convergence_rate(gsl::not_null<double*> result,
+                      const DataVector& power_monitor,
+                      size_t number_of_filtered_modes = 0);
```
knelli2: Remove

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -147,6 +149,56 @@ std::array<double, Dim> absolute_truncation_error(
...
+          << power_monitor.size() << " with " << number_of_filtered_modes
+          << " filtered modes");
+
+  DataVector mode_numbers_for_fit{};
+  DataVector mode_powers_for_fit{};
+  const size_t n_tilde_minus_one =
+      power_monitor.size() - number_of_filtered_modes - 1;
+  std::vector<double> slopes{};
+  std::vector<double> delta_slopes{};
+  intrp::LinearRegressionResult regression_result{};
```
knelli2: No need to have the `regression_result` outside the loop. Just put it inside

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -147,6 +149,56 @@ std::array<double, Dim> absolute_truncation_error(
...
+      slopes.push_back(regression_result.slope);
+      delta_slopes.push_back(regression_result.delta_slope);
+    }
+  }
+  const auto max_delta =
+      std::max_element(delta_slopes.begin(), delta_slopes.end());
+  // Scale the small factor to be a few orders of magnitude smaller than
+  // the max slope delta, as SpEC does. But in case the fit happens to have
+  // identically zero errors, make sure eps is still nonzero.
+  const double eps = 1.e-3 * std::max(*max_delta, 1.e-15);
```
knelli2: The dox have the `1e-3` multiplying the first term in the `std::max`, not the result. Which is correct?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -147,6 +149,56 @@ std::array<double, Dim> absolute_truncation_error(
...
+  double num = 0.0;
+  double denom = 0.0;
+  double denom_this_term = 0.0;
+  for (size_t i = 0; i < slopes.size(); ++i) {
+    denom_this_term = eps + gsl::at(delta_slopes, i);
+    denom += 1.0 / denom_this_term;
```
knelli2: Just define `one_over_denom_this_term` inside the loop and also avoid some divisions.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -147,6 +149,56 @@ std::array<double, Dim> absolute_truncation_error(
...
+      "rate, but power monitor has size "
+          << power_monitor.size() << " with " << number_of_filtered_modes
+          << " filtered modes");
+
+  DataVector mode_numbers_for_fit{};
+  DataVector mode_powers_for_fit{};
+  const size_t n_tilde_minus_one =
+      power_monitor.size() - number_of_filtered_modes - 1;
+  std::vector<double> slopes{};
+  std::vector<double> delta_slopes{};
```
knelli2: To avoid actually resizing the vectors over and over, I think I computed their max possible sizes (though please check my math):
```cpp
const size_t n_tilde =
    power_monitor.size() - number_of_filtered_modes;
const size_t n_tilde_minus_one = n_tilde - 1;
std::vector<double> mode_numbers_for_fit{};
std::vector<double> mode_powers_for_fit{};
mode_numbers_for_fit.reserve(n_tilde);
mode_powers_for_fit.reserve(n_tilde);

const size_t max_slope_size = n_tilde > 6 ? 3*(n_tilde_minus_one)-14 ? 2;
std::vector<double> slopes{};
std::vector<double> delta_slopes{};
slopes.reserve(max_slope_size);
delta_slopes.reserve(max_slope_size);
```

Then below, just use `.resize(M)` on the modes, and `.push_back()` for the slopes and no new allocations should be done.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -1,16 +1,18 @@
 // Distributed under the MIT License.
 // See LICENSE.txt for details.
-
 #include "NumericalAlgorithms/LinearOperators/PowerMonitors.hpp"
```
knelli2: [nitpick] Add this blank line back in.

knelli2 (changes requested):

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
geoffrey4444: Sure, but I thought I was minimizing allocations having it outside the loop. Is it that compiler optimizations will make inside the loop not cost more, or some other reason?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
geoffrey4444: Docs are correct

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
knelli2: It only has `double`s as member variables, so no runtime allocation actually happens

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
geoffrey4444: I just wrote a minimal c++ program to count the number of terms in the loop (see below). Running it shows that 
```c++
size_t max_slope_size = n_tilde > 6 ? 3*(n_tilde-5) ? 4
```

(actually, it's 3 for n_tilde == 4 or 5, 4 for n_tilde == 6). But otherwise I agree with this.

```c++
#include <iostream>

void number_of_terms(const size_t n_tilde) {
  const size_t n_tilde_minus_one = n_tilde - 1;
  size_t terms = 0;
  for (size_t k1 = 0; k1 < 3; ++k1) {
    for (size_t k2 = std::min(k1 + 4, n_tilde_minus_one);
         k2 <= n_tilde_minus_one; ++k2) {
        ++terms;
    }
  }
  const size_t expected_terms = 3 * (n_tilde - 5);
  std::cout << n_tilde << " " << terms << " " << expected_terms
            << " " << terms - expected_terms << "\n";
}

int main() {
  for (size_t n_tilde = 4; n_tilde < 100; ++n_tilde) {
    number_of_terms(n_tilde);
  }
  return 0;
}
```

geoffrey4444:
Thanks for the review, @knelli2 ! Ready for another look when you have a chance

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -187,14 +200,14 @@ double convergence_rate(const DataVector& power_monitor,
   // Scale the small factor to be a few orders of magnitude smaller than
   // the max slope delta, as SpEC does. But in case the fit happens to have
   // identically zero errors, make sure eps is still nonzero.
-  const double eps = 1.e-3 * std::max(*max_delta, 1.e-15);
+  const double eps = std::max(*max_delta * 1.e-3, 1.e-15);
   double num = 0.0;
   double denom = 0.0;
-  double denom_this_term = 0.0;
+  double one_over_denom_this_term = 0.0;
   for (size_t i = 0; i < slopes.size(); ++i) {
```
knelli2: Move `one_over_denom_this_term` into the loop. It doesn't need to persist outside each iteration

knelli2 (changes requested):
You can squash everything

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
geoffrey4444: I did this to avoid an extra allocation, but I guess the cost when it's just a double is negligible?

knelli2: There is no cost since it's a `double`

knelli2 (approved):

nilsdeppe (approved):

PR #6667, opened 2025-05-29
Add action modifying fluid variables for MC coupling

ffoucart:
## Proposed changes

Add an action that
(a) modifies the fluid evolved variables to account for coupling with MC
(b) resets the variables containing the MC coupling to zero

### Upgrade instructions

None

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

None

### DIFF
--- a/src/Evolution/Particles/MonteCarlo/Actions/FluidCouplingAction.hpp
+++ b/src/Evolution/Particles/MonteCarlo/Actions/FluidCouplingAction.hpp
@@ -1,0 +1,110 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "DataStructures/DataBox/DataBox.hpp"
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Index.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Evolution/DgSubcell/ActiveGrid.hpp"
+#include "Evolution/DgSubcell/Tags/ActiveGrid.hpp"
+#include "Evolution/DgSubcell/Tags/Jacobians.hpp"
+#include "Evolution/DgSubcell/Tags/Mesh.hpp"
+#include "Evolution/Particles/MonteCarlo/CellVolume.hpp"
+#include "Evolution/Particles/MonteCarlo/Tags.hpp"
+#include "Evolution/Systems/GrMhd/ValenciaDivClean/Tags.hpp"
+#include "Parallel/AlgorithmExecution.hpp"
+#include "Parallel/GlobalCache.hpp"
+
+namespace Particles::MonteCarlo {
+
+/// Mutator adding the Monte-Carlo contribution
+/// to the evolution of the fluid.
+struct FluidCouplingMutator {
+  static const size_t Dim = 3;
+
+  // We modify the fluid evolved variables, and reset the coupling
+  // terms to zero.
+  using return_tags =
+      tmpl::list<grmhd::ValenciaDivClean::Tags::TildeTau,
+                 grmhd::ValenciaDivClean::Tags::TildeYe,
+                 grmhd::ValenciaDivClean::Tags::TildeS<Frame::Inertial>,
+                 Particles::MonteCarlo::Tags::CouplingTildeTau<DataVector>,
+                 Particles::MonteCarlo::Tags::CouplingTildeRhoYe<DataVector>,
+                 Particles::MonteCarlo::Tags::CouplingTildeS<DataVector, Dim>>;
+  using argument_tags = tmpl::list<
+      evolution::dg::subcell::Tags::Mesh<Dim>,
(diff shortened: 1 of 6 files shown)


src/Evolution/Particles/MonteCarlo/Actions/FluidCouplingAction.hpp
```diff
@@ -0,0 +1,122 @@
...
+      const ArrayIndex& /*array_index*/, const ActionList /*meta*/,
+      const ParallelComponent* const /*meta*/) {
+    ASSERT(db::get<evolution::dg::subcell::Tags::ActiveGrid>(box) ==
+               evolution::dg::subcell::ActiveGrid::Subcell,
+           "MC assumes that we are using the Subcell grid!");
+
+    db::mutate_apply(FluidCouplingMutator{}, make_not_null(&box));
+    return {Parallel::AlgorithmExecution::Continue, std::nullopt};
+  }
+};
```
nilsdeppe: You can use the action `::Actions::MutateApply<Particles::MonteCarlo::FluidCouplingMutator>` where the action is in `src/ParallelAlgorithms/Actions/MutateApply.hpp`, then you don't need the action you have here at all. You could move the `ASSERT` to the mutator as well.

src/Evolution/Particles/MonteCarlo/Actions/FluidCouplingAction.hpp
```diff
@@ -0,0 +1,122 @@
...
+          }
+        }
+      }
+    }
+    // Reset coupling terms to 0 after use
+    get(*coupling_tilde_tau) = 0.0;
+    get(*coupling_tilde_rho_ye) = 0.0;
+    for (size_t d = 0; d < 3; d++) {
+      coupling_tilde_s->get(d) = 0.0;
+    }
```
nilsdeppe, line 106: Why do these get reset instead of just letting them be overwritten the next time step? Normally the pattern we have is:
Action/Mutator 1: Set values (accounting for the fact that they may be non-zero from the last time step)
Action/Mutator 2: Use values (do not modify)
while you have (2) also checking the values.

src/Evolution/Particles/MonteCarlo/Actions/FluidCouplingAction.hpp
```diff
@@ -0,0 +1,122 @@
...
+          // the fluid variables are on the grid without GZ points
+          // Index without GZ
+          const size_t local_idx =
+              i + j * extents[0] + k * extents[0] * extents[1];
+          // Index with GZ
+          const size_t extended_idx =
+              num_ghost_zones + i +
+              (num_ghost_zones + j) * (extents[0] + 2 * num_ghost_zones) +
+              (num_ghost_zones + k) * (extents[0] + 2 * num_ghost_zones) *
+                  (extents[1] + 2 * num_ghost_zones);
```
nilsdeppe: Do or should we factor out these computations into a free function in a header file? That way if they get used in multiple places we only need to update one location and it gets updated everywhere.

nilsdeppe (changes requested):

src/Evolution/Particles/MonteCarlo/Actions/FluidCouplingAction.hpp
ffoucart: All other actions modifying these fields add to them (each packet passing through a cell contributes individually to that tensor, they are not set by a single calculation).
I reset to zero after coupling to the fluid, then start accumulating changes in energy/momentum/lepton number again. To me that's the most natural place to reset to zero, as it makes sure that we reset right after moving the right amount of energy/... to the fluid.

src/Evolution/Particles/MonteCarlo/Actions/FluidCouplingAction.hpp
ffoucart: This already exists, I keep forgetting about it. I'll use the existing function.

ffoucart:
@nilsdeppe A few additional comments:

(1) I switched from an ASSERT that the grid is Subcell to an if statement skipping fluid coupling if the grid is not subcell. This is because that's what I ended up using when combining Subcell and Dg

(2) I took advantage of your comment on indexing to call `collapsed_index` more consistently in the MC code, so there are a few more files now calling that function instead of recalculating indices by hand.

(3) Let me know if you want to move the zeroing of the coupling terms to another function.

src/Evolution/Particles/MonteCarlo/Actions/FluidCouplingAction.hpp
nilsdeppe: Okay, that makes sense to me. Let's leave the code as-is. I don't see any reason to change anything at this stage :)

nilsdeppe:
Fixup LGTM, please rebase and squash :)

ffoucart:
Merged and rebased against develop.

nilsdeppe (approved):

PR #6670, opened 2025-05-30
Factor Spectral.hpp and Spectral.cpp into multiple files

kidder:
No code was changed, except for putting some code that was in Spectral.cpp into a detail namespace when it now appears in a header file.

## Proposed changes

Factors Spectral.?pp into multiple files in preparation for adding support for multiple new bases.
I changed no code with the exception of putting a couple of things into a detail namespace.

### Upgrade instructions

If you included Spectral.hpp you will now need to include the appropriate new header(s).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/NumericalAlgorithms/Spectral/BasisFunctionNormalizationSquare.hpp
```diff
@@ -0,0 +1,21 @@
...
+/// \cond
+namespace Spectral {
+enum class Basis : uint8_t;
+}  // namespace Spectral
+/// \endcond
+
+namespace Spectral {
+/*!
+ * \brief Compute the normalization square of the basis function \f$\Phi_k\f$
+ * (zero-indexed), i.e. the weighted definite integral over its square.
```
nilsvu, line 17: Please add to the docs that this function is specialized for each basis (can be in another commit or another PR). Also for the other specialized functions. Or move these function templates together into `Basis.hpp` or something like that so it's clear which functions need to be specialized when implementing a basis.

src/NumericalAlgorithms/Spectral/CollocationPoints.cpp
```diff
@@ -0,0 +1,44 @@
...
+             BasisType, QuadratureType,
+             detail::CollocationPointsAndWeightsGenerator<BasisType,
+                                                          QuadratureType>>(
+             num_points)
+      .first;
+}
+
+SPECTRAL_QUANTITY_FOR_MESH(collocation_points, DataVector)
+
+#undef SPECTRAL_QUANTITY_FOR_MESH
```
nilsvu, line 30: don't need to undef this I think?

nilsvu (approved):

nilsvu:
I'll just merge this end of day today to avoid conflicts with all the changed includes. Updates to things within Spectral/ can be done in a follow-up.

wthrowe (approved):

PR #6674, opened 2025-06-03
Add function to get number of pile up modes

geoffrey4444:
## Proposed changes

Add a function to compute the number of pile up modes in a power monitor, and ensures that the power monitor `convergence_rate()` does not fail with a floating point exception if a mode has zero power.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

I have used [Cursor](https://cursor.com) with generative-AI features, including tab completion, on this PR. I have examined each proposed change, and to the best of my knowledge, they do not contain code that is copyrighted or not licensed for our use.

### DIFF
--- a/src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
+++ b/src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
@@ -150,16 +150,17 @@ std::array<double, Dim> absolute_truncation_error(
   return result;
 }
 
-double convergence_rate(const DataVector& power_monitor,
-                        const size_t number_of_filtered_modes) {
+ConvergenceInfo convergence_rate_and_number_of_pile_up_modes(
+    const DataVector& power_monitor, const size_t number_of_filtered_modes) {
   // Need enough unfiltered modes to compute the convergence rate. Here,
   // require at least 4 unfiltered modes.
   ASSERT(
       power_monitor.size() > number_of_filtered_modes + 3,
       "Power monitor needs at least 4 unfiltered modes to compute convergence "
-      "rate, but power monitor has size "
+      "rate and number of pile up modes, but power monitor has size "
           << power_monitor.size() << " with " << number_of_filtered_modes
           << " filtered modes");
+  ConvergenceInfo result{};
 
   const size_t n_tilde = power_monitor.size() - number_of_filtered_modes;
   std::vector<double> mode_numbers_for_fit{};
@@ -172,14 +173,29 @@ double convergence_rate(const DataVector& power_monitor,
   std::vector<double> delta_slopes{};
 
   // It turns out (as can be verified empirically) that the number of terms
-  // in this loop is 3 * (n_tilde - 5) for n_tilde > 6, 4 for n_tilde == 5,
-  // and 3 for n_tilde == 4 or n_tilde == 3. Here reserve the correct number of
-  // elements in terms of n_tilde, except don't use an extra if to
-  // distinguish between the 3 and 4 special cases (no harm in reserving space
-  // for a single extra double).
+  // in the loop to compute the convergence rate is 3 * (n_tilde - 5) for
+  // n_tilde > 6, 4 for n_tilde == 5, and 3 for n_tilde == 4 or n_tilde == 3.
+  // Here reserve the correct number of elements in terms of n_tilde, except
+  // don't use an extra if to distinguish between the 3 and 4 special cases (no
+  // harm in reserving space for a single extra double).
   const size_t max_slope_size = n_tilde > 6 ? 3 * (n_tilde - 5) : 4;
   slopes.reserve(max_slope_size);
   delta_slopes.reserve(max_slope_size);
+
+  // Compute log of the power monitor only for unfiltered modes, and
+  // ensure that log10 never causes a floating point exception here.
+  constexpr double eps_for_log = 100.0 * std::numeric_limits<double>::min();
+  const double log_floor = log10(eps_for_log);
+  DataVector log_power = power_monitor;
+  for (size_t i = 0; i < power_monitor.size() - number_of_filtered_modes; ++i) {
+    if (abs(log_power[i]) < eps_for_log) {
+      log_power[i] = log_floor;
+    } else {
+      log_power[i] = log10(abs(log_power[i]));
+    }
+  }
+
+  // Compute convergence rate
   for (size_t k1 = 0; k1 < 3; ++k1) {
     for (size_t k2 = std::min(k1 + 4, n_tilde_minus_one);
          k2 <= n_tilde_minus_one; ++k2) {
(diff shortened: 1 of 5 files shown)


geoffrey4444:
I will figure out the FPEs before this is worth taking a look at

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -212,6 +236,77 @@ double convergence_rate(const DataVector& power_monitor,
   return -num / denom;
 }
 
+double number_of_pile_up_modes(const DataVector& power_monitor,
```
nilsdeppe: This seems to do a decent amount of math that heavily overlaps with the convergence rate computation. It seems at least all the logs, would it make sense to combine the functions, or maybe offer a wrapper that does the shared math, for these implementations?

Probably a future issue, but I worry a bit about the number of memory allocations being done here because this is call per dim (and if I'm reading right per tensor component)?? Between this and the `convergence_rate`, that's a lot of memory allocations being done. Sure, this is only once every 50 time steps or something, but if this code alone does as many memory allocations as those 50 time steps, that's not great...

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -212,6 +236,77 @@ double convergence_rate(const DataVector& power_monitor,
...
+          (mode_powers_for_local_fit[1] - mode_powers_for_local_fit[0]) /
+          (mode_numbers_for_local_fit[1] - mode_numbers_for_local_fit[0]);
+    } else {
+      // cannot measure slope, so skip this term in sum
+      continue;
+    }
+    const double conv_ratio = square(local_convergence_rate / conv_rate);
+    // Avoid underflow: if conv_ratio < 16, just add zero.
+    // exp(-32.0*16.0) ~ 1.0e-195, which is still large enough to
+    // avoid underflow.
```
nilsdeppe: Nice comment :D Very helpful and future us will thank you :)

nilsdeppe (changes requested):
I'm worried about the overall design of these functions being, as far as I can tell, at the by-tensor component per dimension level since each function does a lot of memory allocations. It does look like at least the linear regression is allocation free, so that's fine, I think. It's really the `std::vector` and `DataVector` allocations that are being done.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
geoffrey4444: I've thought about this some. Since `number_of_pile_up_modes()` always needs the output of `convergence_rate()`, and anyone interested in `convergence_rate()` is likely to be interested also in `number_of_pile_up_modes`, I think it makes the most sense to rework this PR to merge these into a single function that returns both. That will let me reuse the `std::vector` and `DataVector` allocations.

I don't try to combine multiple dimensions or tensor components because for the adaptive horizon finder, you get one power monitor from a single power monitor.

geoffrey4444:
@nilsdeppe Thank you very much for the suggestions! To minimize allocations, I've now made a combined function that gets convergence rate and number of pile up modes, to reuse as many allocations as I could. I'm not sure how to respond to the concern about this function working on the by-component, by-dimension level, because for use in adaptive horizon finding, I need it to work for the single scalar power monitor of the Strahlkorper. If you'd like further changes along those lines, could we please discuss the best approach?

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -233,76 +235,49 @@ double convergence_rate(const DataVector& power_monitor,
...
-          << power_monitor.size() << " with " << number_of_filtered_modes
-          << " filtered modes");
-  const double conv_rate =
-      convergence_rate(power_monitor, number_of_filtered_modes);
-  // If the convergence rate is nearly zero, return zero pile up modes.
+  // Compute number of pile up modes
+  // First, if the convergence rate is nearly zero, return zero pile up modes.
   // This avoids dividing by zero, and is reasonable in the sense that if
   // the convergence rate is very small, it's not realistic to distinguish
   // whatever convergence you might have left and pile up modes.
```
nilsdeppe: I realize this is an old comment, but maybe giving an example would be good? E.g. "This means that the function is not at all resolved, eg a step function, and so the power spectrum is approximately flat." I know you're saying pretty much the same thing but it took me a bit of thinking to figure it out, and I've thought about this a lot.

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
```diff
@@ -233,76 +235,49 @@ double convergence_rate(const DataVector& power_monitor,
...
+          std::numeric_limits<double>::signaling_NaN();
+      if (mode_numbers_for_fit.size() > 2) {
+        local_convergence_rate =
+            -intrp::linear_regression(mode_numbers_for_fit, mode_powers_for_fit)
+                 .slope;
+      } else if (mode_numbers_for_fit.size() == 2 and
+                 mode_numbers_for_fit[1] != mode_numbers_for_fit[0]) {
+        local_convergence_rate =
+            (mode_powers_for_fit[1] - mode_powers_for_fit[0]) /
+            (mode_numbers_for_fit[1] - mode_numbers_for_fit[0]);
```
nilsdeppe, line 271: With ternary? I'll let you decide what you think is cleaner!
```cpp
const double local_convergence_rate =                                                                                                          
    mode_numbers_for_fit.size() > 2                                                                                                      
        ? -intrp::linear_regression(mode_numbers_for_fit, mode_powers_for_fit).slope                                                     
        : (mode_numbers_for_fit.size() == 2 and mode_numbers_for_fit[1] != mode_numbers_for_fit[0]                                        
            ? (mode_powers_for_fit[1] - mode_powers_for_fit[0]) /                                                                        
              (mode_numbers_for_fit[1] - mode_numbers_for_fit[0])                                                                        
            : std::numeric_limits<double>::signaling_NaN());
```

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
```diff
@@ -36,13 +36,18 @@ void bind_power_monitors(py::module& m) {
...
-  m.def("number_of_pile_up_modes",
-        py::overload_cast<const DataVector&, const size_t>(
-            &number_of_pile_up_modes),
-        py::arg("power_monitor"), py::arg("number_of_filtered_modes") = 0);
+  m.def(
+      "convergence_rate_and_number_of_pile_up_modes",
+      [](const DataVector& power_monitor,
+         const size_t number_of_filtered_modes) {
+        py::dict result{};
+        ConvergenceInfo info = convergence_rate_and_number_of_pile_up_modes(
```
nilsdeppe: make const as clang-tidy suggests?

nilsdeppe (commented):
LGTM, please squash!

src/NumericalAlgorithms/LinearOperators/Python/PowerMonitors.cpp
geoffrey4444: Yep pretty sure I caught that one late (missed it when going through to make the fix), I will make sure clang tidy passes. EDIT: yep, fixed

src/NumericalAlgorithms/LinearOperators/PowerMonitors.cpp
geoffrey4444: I find it cleaner as is (decoding ternary in parentheses  is the deciding factor for me), but if other reviewers disagree, I'm happy to make the switch.

nilsdeppe: Sounds good! I don't think I'm the right person to evaluate here what is "easier to read" :)

geoffrey4444:
[Test failure](https://github.com/sxs-collaboration/spectre/actions/runs/15767337821/job/44446155249?pr=6674) was 771 - Unit.Evolution.Systems.GrMhd.ValenciaDivClean.Fd.PpaoPrim (Timeout), unrelated to this PR

nilsdeppe (approved):

PR #6679, opened 2025-06-06
Add smoothstep_deriv function

nilsvu:
## Proposed changes

Implements the derivative of the `smoothstep` function. I'm using this for the self-force problem.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/ScalarTensor/SourceTags.hpp
+++ b/src/PointwiseFunctions/ScalarTensor/SourceTags.hpp
@@ -8,6 +8,7 @@
 #include "DataStructures/Tensor/Tensor.hpp"
 #include "Evolution/Systems/ScalarTensor/Tags.hpp"
 #include "Options/String.hpp"
+#include "Utilities/ErrorHandling/Error.hpp"
 
 /// \cond
 namespace ScalarTensor::OptionTags {
@@ -35,6 +36,7 @@ struct RampUpDuration {
   using type = double;
   static constexpr Options::String help{"Duration time for ramp up function"};
   using group = ::ScalarTensor::OptionTags::Group;
+  static double lower_bound() { return 0.0; }
 };
 
 }  // namespace OptionTags
@@ -50,7 +52,11 @@ struct RampUpParameters : db::SimpleTag {
   static constexpr bool pass_metavariables = false;
   static std::pair<double, double> create_from_options(
       const double start_time, const double duration_time) {
-    return std::pair<double, double> {start_time, duration_time};
+    if (duration_time <= 0.0) {
+      ERROR("Ramp up duration time must be greater than zero, but is "
+            << duration_time);
+    }
+    return std::pair<double, double>{start_time, duration_time};
   }
 };
 
--- a/src/Utilities/ConstantExpressions.hpp
+++ b/src/Utilities/ConstantExpressions.hpp
@@ -92,6 +92,25 @@ KOKKOS_FUNCTION constexpr uint64_t factorial(const uint64_t n) {
   return falling_factorial(n, n);
 }
 
+/*!
+ * \ingroup ConstantExpressionsGroup
+ * \brief Compute the binomial coefficient $\binom{n}{k}$
+ *
+ * \details The binomial coefficient is defined as
+ * $\binom{n}{k} = \frac{n!}{k!(n-k)!}$
+ */
+KOKKOS_FUNCTION constexpr uint64_t binomial(const uint64_t n,
+                                            const uint64_t k) {
+  assert(n < 63);  // NOLINT
+  assert(k <= n);  // NOLINT
+  uint64_t result = 1;
+  for (uint64_t j = 1; j <= std::min(k, n - k); ++j) {
+    result *= n + 1 - j;
+    result /= j;
+  }
+  return result;
+}
+
 /// \ingroup ConstantExpressionsGroup
 namespace ConstantExpressions_detail {
 
--- a/src/Utilities/Math.hpp
+++ b/src/Utilities/Math.hpp
@@ -76,29 +76,39 @@ constexpr T step_function(const T& arg) {
 /*!
  * \ingroup UtilitiesGroup
  * \brief Smoothly interpolates from 0 to 1 between `lower_edge` and
- * `upper_edge` with a Hermite polynomial of degree `2 * N + 1`.
+ * `upper_edge` with Hermite interpolation of polynomial degree `2 * N + 1`.
  *
  * The smoothstep function is
  *
- * \f{align*}
+ * \begin{align*}
  * S_N(x) = \begin{cases}
  * 0 &\quad \text{for} \quad x\leq x_0 \\
  * \tilde{S}_N((x - x_0) / (x_1 - x_0))
  * &\quad \text{for} \quad x_0 \leq x\leq x_1 \\
  * 1 &\quad \text{for} \quad x_1\leq x \\
  * \end{cases}
- * \f}
+ * \end{align*}
  *
- * where \f$x_0\f$ is `lower_edge`, \f$x_1\f$ is `upper_edge`, and, up to
- * \f$N=3\f$,
+ * where \f$x_0\f$ is `lower_edge` and \f$x_1\f$ is `upper_edge`. The general
+ * form of the polynomial \f$\tilde{S}_N(x)\f$ is:
  *
- * \f{align*}
+ * \begin{equation}
+ * \tilde{S}_N(x) = x^{N+1} \sum_{k=0}^{N} \binom{N+k}{k}
+ *   \binom{2N+1}{N-k} (-x)^k
+ * \end{equation}
+ *
+ * The first few polynomials are:
+ *
+ * \begin{align*}
  * \tilde{S}_0(x) &= x \\
  * \tilde{S}_1(x) &= 3x^2 - 2x^3 \\
  * \tilde{S}_2(x) &= 10x^3 - 15x^4 + 6x^5 \\
  * \tilde{S}_3(x) &= 35x^4 - 84x^5 + 70x^6 - 20x^7
  * \text{.}
- * \f}
+ * \end{align*}
+ *
+ * This function is $C^N$ continuous at the edges, i.e. the first $N$
+ * derivatives are continuous at the edges.
  */
 template <size_t N, typename DataType>
 DataType smoothstep(const double lower_edge, const double upper_edge,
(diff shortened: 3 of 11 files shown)


duetosymmetry:
Noting for posterity that it may be simpler (depending on who you ask) to write this function as (2 n + 1)!/n!^2 (x (1-x))^n

tests/Unit/Utilities/Test_Math.cpp
```diff
@@ -23,6 +24,21 @@ void test_smoothstep() {
   CHECK_ITERABLE_APPROX(
       smoothstep<N>(-1., 1., DataVector({-2., -1., 0., 1., 2.})),
       DataVector({0., 0., 0.5, 1., 1.}));
+  if constexpr (N > 0) {
+    INFO("Test smoothstep_deriv");
+    CHECK(smoothstep_deriv<N>(0., 1., 0.) == approx(0.));
+    CHECK(smoothstep_deriv<N>(0., 1., 0.) == approx(0.));
```
duetosymmetry: These two checks are the same... is the second one supposed to evaluate at arg=1 instead of arg=0?
```suggestion
    CHECK(smoothstep_deriv<N>(0., 1., 1.) == approx(0.));
```

src/Utilities/Math.hpp
```diff
@@ -120,6 +124,43 @@ DataType smoothstep(const double lower_edge, const double upper_edge,
...
+          0., 1.)));
+}
+
+/*!
+ * \ingroup UtilitiesGroup
+ * \brief Derivative of the $N$-th order smoothstep function
+ *
+ * Since the smoothstep function is $C^N$ continuous at the edges, this
+ * function is $C^{N-1}$ continuous at the edges. It is only implemented for
+ * $N > 0$ and $N \leq 4$.
```
guilara: Looks good to me. I agree with @duetosymmetry , it could be convenient to have some explicit expressions (or references) in the documentation in case anybody wants to go to higher order, compute derivatives, etc. For instance, I was just trying to construct them myself and found the documentation of smoothstep a bit unclear. Leo's expression of (2 n + 1)!/n!^2 (x (1-x))^n works for the derivative here. But I couldn't construct the function itself from Hermite polynomials (probably a combination? or did the documentation actually mean Hermite interpolation and not polynomials?). I found an expression by explicit integration but don't know if there is anything simpler: S_n = x^n * Sum_{k=0}^{n} (-1)^k Binomial[n+k, k] Binomial[2n+1, n-k] x^(k+1) .

guilara (approved):

src/Utilities/Math.hpp
duetosymmetry: @guilara I didn't check, does your expression agree with https://duetosymmetry.com/notes/a-family-of-ramp-functions-and-the-Beta-function/ ? There might be some recursion to compute the polynomial coeffs, but I don't think it's going to get much simpler. We could just compute them directly from the binomial coeffs, which themselves satisfy a simple recursion (better than evaluating them with factorials)

guilara: @duetosymmetry Yep, its matches Eq. 7 actually

duetosymmetry: We can say it's the regularized incomplete beta function and give a ref... There are some recurrences at https://dlmf.nist.gov/8.17#E12 , but I'm not sure they would make anything simpler

nilsvu:
@duetosymmetry @guilara I generalized the function to higher order so @guilara you should be able to just increase `N` now if you want to test higher order.

duetosymmetry (dismissed):

src/Utilities/Math.hpp
```diff
@@ -76,55 +76,85 @@ constexpr T step_function(const T& arg) {
...
  *
- * where \f$x_0\f$ is `lower_edge`, \f$x_1\f$ is `upper_edge`, and, up to
- * \f$N=3\f$,
+ * where \f$x_0\f$ is `lower_edge` and \f$x_1\f$ is `upper_edge`. The general
+ * form of the polynomial \f$\tilde{S}_N(x)\f$ is:
  *
- * \f{align*}
+ * \begin{equation}
+ * \tilde{S}_N(x) = x^{N+1} \sum_{k=0}^{N} \mathrm{binom}(N+k, k)
+ *   \mathrm{binom}(2N+1, N-k) (-x)^k
```
duetosymmetry: I think this doesn't really matter... but in the docs for binomial, you used `\binom{n}{k}` whereas here you write `\mathrm{binom}(n, k)`

nilsvu: fixed

duetosymmetry (dismissed):

guilara:
> @duetosymmetry @guilara I generalized the function to higher order so @guilara you should be able to just increase `N` now if you want to test higher order.

Great! For the failing ScalarTensor test maybe you can change the numbers to for the ramp function in the input file to Start = -2.0 and Duration = 1.0, since its complaining about a duration of zero I think.

guilara:
@nilsvu Is there anything else to do here to merge it?

nilsvu:
@sxs-collaboration/spectre-core-devs can one of you take this one?

src/Utilities/Math.hpp
```diff
@@ -76,55 +76,85 @@ constexpr T step_function(const T& arg) {
...
-          return {0., 0., 0., 0., 35., -84., 70., -20.};
-        }
-      }(),
-      static_cast<DataType>(clamp(
-          static_cast<DataType>((arg - lower_edge) / (upper_edge - lower_edge)),
-          0., 1.)));
+  const DataType x = clamp(
+      static_cast<DataType>((arg - lower_edge) / (upper_edge - lower_edge)), 0.,
+      1.);
+  return (2 * N + 1) * binomial(2 * N, N) * pow<N>(x - square(x));
```
nilsdeppe: [optional] you could force constexpr computation using `constexpr auto coeff = (2 * N + 1) * binomial(2 * N, N);`
Please let me know whether or not you plan on making this change.

tests/InputFiles/ScalarTensor/KerrSchildSphericalHarmonic.yaml
```diff
@@ -97,8 +97,8 @@ EvolutionSystem:
   ScalarTensor:
     ScalarMass: 0.0
     # Ramp up effectively disabled
-    RampUpStart: -1.0
-    RampUpDuration: 0.0
+    RampUpStart: -2.0
+    RampUpDuration: 1.0
```
nilsdeppe, line 103: It is not obvious to me why these values are changing. Why is that?

nilsdeppe (changes requested):

src/Utilities/ConstantExpressions.hpp
```diff
@@ -92,6 +92,22 @@ KOKKOS_FUNCTION constexpr uint64_t factorial(const uint64_t n) {
...
+ * \details The binomial coefficient is defined as
+ * $\binom{n}{k} = \frac{n!}{k!(n-k)!}$
+ *
+ * \note The largest representable factorial is 20!. It is
+ * up to the user to ensure this is satisfied
+ */
+KOKKOS_FUNCTION constexpr uint64_t binomial(const uint64_t n,
+                                            const uint64_t k) {
+  assert(k <= n);  // NOLINT
+  return falling_factorial(n, k) / factorial(k);
```
kidder: - isn't it better to pass min(k, n-k) instead of k?
- do we really want to rely on users and not assert when this will fail?

tests/InputFiles/ScalarTensor/KerrSchildSphericalHarmonic.yaml
nilsvu: `RampUpDuration: 0` is not valid anymore because I didn't implement a special case for that in `RampUpFunction.cpp` (@guilara told me to just do this change instead where the ramp-up time is set to 1).

src/Utilities/Math.hpp
nilsvu: done

src/Utilities/ConstantExpressions.hpp
nilsvu: did the first. The second is not so clear for the falling factorial (the factorial function already has the assert)

duetosymmetry: The range of binomial coeffs you could compute without overflow could be extended all the way up to binomial(67,k) (for uint64) by using their recurrence, hence only adding, instead of multiplying and then dividing back down. But then you'd probably want to pre-populate a table and lookup instead of computing on the fly.

kidder:
```
constexpr uint64_t binomial(const uint64_t n, const uint64_t k) {
  assert(n < 63);
  assert(k <=n);
  uint64_t result = 1;
  for (uint64_t j = 1; j <= std::min(k, n-k); ++j) {
    result *= n+1-j;
    result /= j;
  }
  return result;
}
```

tests/InputFiles/ScalarTensor/KerrSchildSphericalHarmonic.yaml
nilsdeppe: Should we add an ASSERT to the input file parsing code that checks this?

src/Utilities/ConstantExpressions.hpp
nilsvu: Ok I took this implementation. I think we're obsessing about this function a bit too much :)

nilsdeppe (commented):
LGTM, add `CHECK_THROWS_WITH` test for `create_from_options` function?

nilsdeppe:
@nilsvu this is waiting on you

nilsvu:
Squashed and rebased. Thanks for reviewing!

src/PointwiseFunctions/ScalarTensor/SourceTags.hpp
```diff
@@ -50,7 +52,11 @@ struct RampUpParameters : db::SimpleTag {
   static constexpr bool pass_metavariables = false;
   static std::pair<double, double> create_from_options(
       const double start_time, const double duration_time) {
-    return std::pair<double, double> {start_time, duration_time};
+    if (duration_time <= 0.0) {
+      ERROR("Ramp up duration time must be greater than zero, but is "
+            << duration_time);
+    }
```
nilsdeppe, line 58: I can't find the `CHECK_THROWS_WITH` test I requested. Could you add that?

nilsvu: Squashed this in

nilsdeppe:
Looks reasonable, but some detail seems to be wrong :(

nilsdeppe (approved):

PR #6682, opened 2025-06-09
Allow reconstructing rho*T instead of just T

nilsdeppe:
## Proposed changes

This is necessary to get reasonable results with problems like the KH instability. We'll want to experiment and compare what we reconstruct for BNS as well. We probably only need to initially test the first orbit or so.

The improvement in the KH instability comes from that rho*T is basically the pressure. The temperature is very large outside of the central strip and very small inside the central strip, while the product `rho*T` is essentially constant across the domain.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/DgSubcell/Actions/TciAndSwitchToDg.hpp
+++ b/src/Evolution/DgSubcell/Actions/TciAndSwitchToDg.hpp
@@ -285,7 +285,7 @@ struct TciAndSwitchToDg {
         not cell_is_troubled and
 
         (((is_substep_method and
-           tci_history.size() == subcell_options.min_clear_tci_before_dg() - 1)
+           tci_history.size() >= subcell_options.min_clear_tci_before_dg() - 1)
 
           or
 
--- a/src/Evolution/Systems/GrMhd/GhValenciaDivClean/FiniteDifference/ReconstructWork.tpp
+++ b/src/Evolution/Systems/GrMhd/GhValenciaDivClean/FiniteDifference/ReconstructWork.tpp
@@ -60,6 +60,7 @@ void reconstruct_prims_work(
         neighbor_data,
     const Mesh<3>& subcell_mesh, const size_t ghost_zone_size,
     const bool compute_conservatives,
+    const bool reconstruct_density_times_temperature,
     const VariableFixing::FixToAtmosphere<3>* fix_to_atmosphere) {
   ASSERT(Mesh<3>(subcell_mesh.extents(0), subcell_mesh.basis(0),
                  subcell_mesh.quadrature(0)) == subcell_mesh,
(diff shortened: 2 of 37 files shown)


src/Evolution/DgSubcell/Actions/TciAndSwitchToDg.hpp
```diff
@@ -285,7 +285,7 @@ struct TciAndSwitchToDg {
         not cell_is_troubled and
 
         (((is_substep_method and
-           tci_history.size() == subcell_options.min_clear_tci_before_dg() - 1)
+           tci_history.size() >= subcell_options.min_clear_tci_before_dg() - 1)
```
wthrowe, line 288: This looks reasonable, although I don't fully understand what the code is doing.  Should the same change be made for the multistep case 5 lines down?

src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/BoundaryConditionGhostData.hpp
```diff
@@ -267,6 +270,87 @@ void BoundaryConditionGhostData::apply(
...
         });
+    if (dynamic_cast<const BoundaryConditions::DirichletAnalytic*>(
+            &boundary_condition_at_direction) != nullptr and
+        reconstructor.reconstruct_rho_times_temperature()) {
+      // If we reconstruct rho*T we end up having to divide by rho to compute
+      // T. In some cases, like when evolving a TOV star with an analytic
+      // boundary condition, the boundary condition sets rho=0. While
+      // unphysical in general, this is how we have implemented the
+      // solutions. We deal with this by applying our atmosphere treatment to
+      // the reconstructed stated.
```
wthrowe: Typo "stated" -> "state"

src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/ReconstructWork.tpp
```diff
@@ -218,35 +240,66 @@ void reconstruct_prims_work(
...
+                                      (2 * direction.dimension() +
+                                       (direction.side() == Side::Upper ? 1
+                                                                        : 0)) *
+                                          neighbor_num_pts],
+              number_of_variables * neighbor_num_pts);
+
+          const auto& data_in_dir = neighbor_data.at(id);
+          DataVector view{
+              /// NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+              const_cast<double*>(ghost_cell_vars[direction].data()),
```
wthrowe: Do you need the const_cast here?  It looks like the pointer isn't const, and if it is assigning to it in the next expression seems fishy.

src/Evolution/Systems/GrMhd/GhValenciaDivClean/FiniteDifference/ReconstructWork.tpp
```diff
@@ -124,37 +143,69 @@ void reconstruct_prims_work(
...
+                                      (2 * direction.dimension() +
+                                       (direction.side() == Side::Upper ? 1
+                                                                        : 0)) *
+                                          neighbor_num_pts],
+              number_of_variables * neighbor_num_pts);
+
+          const auto& data_in_dir = neighbor_data.at(id);
+          DataVector view{
+              /// NOLINTNEXTLINE(cppcoreguidelines-pro-type-const-cast)
+              const_cast<double*>(ghost_cell_vars[direction].data()),
```
wthrowe: Same comments as previous commit.

wthrowe (changes requested):

src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/ReconstructWork.tpp
nilsdeppe: It's because we use `gsl::span<const double>` in the reconstruction code. However, I just realized I can reorder the code a bit to make it clearer :)

src/Evolution/DgSubcell/Actions/TciAndSwitchToDg.hpp
nilsdeppe: From testing the counting is correct for multi step methods but not for substep methods. This is because in multistep methods we actually push back the right number of TCI histories, but for substep this is a bit more annoying so it's easier to adjust the counting.

nilsdeppe (commented):
Thanks for the review! I rebased and pushed fixups.

src/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/ReconstructWork.tpp
wthrowe: I see, I missed that.  But the new version is better, anyway.

wthrowe (approved):
Looks good.  Squash.

nilsdeppe:
Squashed!

PR #6684, opened 2025-06-10
Add GravitationalEffectiveSource code

nilsvu:
## Proposed changes

This is the GR version of the self-force effective source code bundled in external/. The repo (https://github.com/barrywardell/GravitationalEffectiveSource) is currently private, with invitation on request, but the code is MIT licensed.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


external/GravitationalEffectiveSource/README.md
```diff
@@ -0,0 +1,9 @@
+This is a modified version of https://github.com/barrywardell/GravitationalEffectiveSource
```
wthrowe, line 1: Mention that the repo is private, since people will generally expect a URL to work if given with no context.

external/GravitationalEffectiveSource/kerr-circular.c
```diff
@@ -0,0 +1,1801 @@
+/*******************************************************************************
+ * Copyright (C) 2012-2013 Barry Wardell
+ *******************************************************************************/
+
+#include <math.h>
+#include "effsource.h"
```
wthrowe: Have you tried compiling this?

wthrowe (changes requested):
While I can't see the source repo, the similar https://github.com/barrywardell/EffectiveSource has some basic documentation in the README.  Is there anything similar for the GR version?  If not that's fine, but if it exists it would be good to include it.  If not, maybe reference the similar public repo so people have a better idea what's going on.

external/GravitationalEffectiveSource/README.md
nilsvu: Ok I added some information to the readme (squashed in)

external/GravitationalEffectiveSource/kerr-circular.c
nilsvu: good catch, fixed and squashed in the rename to `effsource_gr.h` (this got lost when I pulled in the most recent revision of the file from the source repo).

wthrowe (approved):

PR #6685, opened 2025-06-10
Update bbh pipeline to use higher resolution in initial data

geoffrey4444:
## Proposed changes

Update the default p refinement in the BBH pipeline to match the choice used in [arXiv:2410.00265](https://arxiv.org/abs/2410.00265). Currently, the default is 3 points coarser (P-6), which I found led to 10^4 worse constraint energy at t=0 in an evolution than the choice used in the paper (P=9).

Until we know to set this to something better, let's at least set it to by default match what was done in the paper.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/InitialData.py
+++ b/support/Pipelines/Bbh/InitialData.py
@@ -154,7 +154,7 @@ def generate_id(
     linear_velocity: Sequence[float] = [0.0, 0.0, 0.0],
     # Resolution
     refinement_level: int = 1,
-    polynomial_order: int = 6,
+    polynomial_order: int = 9,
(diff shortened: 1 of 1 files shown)


support/Pipelines/Bbh/InitialData.py
```diff
@@ -409,7 +409,7 @@ def generate_id(
     "-P",
     type=click.IntRange(1, None),
     help="p-refinement level.",
-    default=6,
```
nilsvu, line 412: Also change this in line 157

nilsvu (approved):

PR #6686, opened 2025-06-11
Change polytropic to tab EOS for CCSN initialize

pajkosmi:
## Proposed changes

Proper read in of tabulated EOS data to initialize CCSN executable.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
+++ b/src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
@@ -18,6 +18,9 @@
 #include "DataStructures/Tensor/EagerMath/DotProduct.hpp"
 #include "DataStructures/Tensor/EagerMath/Magnitude.hpp"
 #include "DataStructures/Tensor/Tensor.hpp"
+#include "IO/H5/AccessType.hpp"
+#include "IO/H5/EosTable.hpp"
+#include "IO/H5/File.hpp"
 #include "NumericalAlgorithms/Interpolation/PolynomialInterpolation.hpp"
 #include "PointwiseFunctions/GeneralRelativity/ExtrinsicCurvature.hpp"
 #include "PointwiseFunctions/GeneralRelativity/Tags.hpp"
@@ -77,8 +80,11 @@ ProgenitorProfile::ProgenitorProfile(const std::string& filename) {
   }
   radius_ *= c2g_length_;
   rest_mass_density_ *= c2g_dens_;
-  temperature_ *= c2g_temp_;
+  // Kelvin to MeV conversion.
+  const double K_to_MeV = 8.621738e-11;
+  temperature_ = temperature_ * K_to_MeV;
   fluid_velocity_ /= speed_of_light_cgs_;
+  fluid_velocity_ = abs(fluid_velocity_);
   maximum_radius_ = max(radius_);
 }
 
@@ -153,8 +159,9 @@ std::array<double, 6> ProgenitorProfile::interpolate(
   }
   using std::abs;
 
-  auto radius_index =
-      static_cast<size_t>(num_radial_points_ * target_radius / maximum_radius_);
+  const auto radius_index = static_cast<size_t>(std::distance(
+      radius_.begin(),
+      std::lower_bound(radius_.begin(), radius_.end(), target_radius)));
 
   if (const size_t grid_index = num_angular_points_ * radius_index;
       UNLIKELY(grid_index > num_grid_points_)) {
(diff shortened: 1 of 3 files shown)


src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -153,8 +156,14 @@ std::array<double, 6> ProgenitorProfile::interpolate(
...
-  auto radius_index =
-      static_cast<size_t>(num_radial_points_ * target_radius / maximum_radius_);
+  size_t radius_index = 0;
+
+  for (size_t i = 0; i < num_radial_points_; i++) {
+    if (radius_[i] > target_radius) {
+      radius_index = i - 1;
+      break;
+    }
+  }
```
nilsdeppe: You can do this with a binary search, I think. https://en.cppreference.com/w/cpp/algorithm/lower_bound.html and then use https://en.cppreference.com/w/cpp/iterator/distance.html
```cpp
const auto radius_index = static_cast<size_t>(std::distance(radius_.begin(),
    std::lower_bound(radius.begin(), radius.end(), target_radius));
```

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -286,18 +295,20 @@ void compute_angular_coordinates(
 }  // namespace detail
 
 CcsnCollapse::CcsnCollapse(std::string progenitor_filename,
-                           double polytropic_constant, double adiabatic_index,
-                           double central_angular_velocity,
-                           double diff_rot_parameter,
-                           double max_dens_ratio_interp)
+                           const double& central_angular_velocity,
+                           const double& diff_rot_parameter,
+                           const double& max_dens_ratio_interp,
```
nilsdeppe: For basic types (double, float, int, etc.) taking by value is better because a pointer (which is what a reference is) is 8 bytes while a double is 8 bytes, but with the pointer you have to resolve the pointer. Not a huge deal here, but in perf critical regions this can make a big difference.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -530,39 +530,55 @@ CcsnCollapse::variables(
...
+    const gsl::not_null<IntermediateVariables<DataType>*> vars,
+    const tnsr::I<DataType, 3>& x,
+    tmpl::list<hydro::Tags::Pressure<DataType>> /*meta*/) const {
+  const auto rest_mass_density = get<hydro::Tags::RestMassDensity<DataType>>(
+      variables(vars, x, tmpl::list<hydro::Tags::RestMassDensity<DataType>>{}));
+  const auto temperature = get<hydro::Tags::Temperature<DataType>>(
+      variables(vars, x, tmpl::list<hydro::Tags::Temperature<DataType>>{}));
+  const auto electron_fraction =
+      get<hydro::Tags::ElectronFraction<DataType>>(variables(
+          vars, x, tmpl::list<hydro::Tags::ElectronFraction<DataType>>{}));
```
nilsdeppe, line 554: Use `const auto&` here. `const auto` does a copy instead of a reference.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -530,39 +530,55 @@ CcsnCollapse::variables(
     tmpl::list<hydro::Tags::SpecificInternalEnergy<DataType>> /*meta*/) const {
   const auto rest_mass_density = get<hydro::Tags::RestMassDensity<DataType>>(
       variables(vars, x, tmpl::list<hydro::Tags::RestMassDensity<DataType>>{}));
+  const auto electron_fraction =
+      get<hydro::Tags::ElectronFraction<DataType>>(variables(
+          vars, x, tmpl::list<hydro::Tags::ElectronFraction<DataType>>{}));
+  const auto temperature = get<hydro::Tags::Temperature<DataType>>(
+      variables(vars, x, tmpl::list<hydro::Tags::Temperature<DataType>>{}));
```
nilsdeppe, line 535: Use `const auto&` here. `const auto` does a copy instead of a reference.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
```diff
@@ -530,39 +530,55 @@ CcsnCollapse::variables(
...
+    const tnsr::I<DataType, 3>& x,
+    tmpl::list<hydro::Tags::SpecificEnthalpy<DataType>> /*meta*/) const {
+  const auto rest_mass_density = get<hydro::Tags::RestMassDensity<DataType>>(
+      variables(vars, x, tmpl::list<hydro::Tags::RestMassDensity<DataType>>{}));
+  const auto specific_internal_energy =
+      get<hydro::Tags::SpecificInternalEnergy<DataType>>(variables(
+          vars, x,
+          tmpl::list<hydro::Tags::SpecificInternalEnergy<DataType>>{}));
+  const auto pressure = get<hydro::Tags::Pressure<DataType>>(
+      variables(vars, x, tmpl::list<hydro::Tags::Pressure<DataType>>{}));
```
nilsdeppe, line 573: Use `const auto&` here. `const auto` does a copy instead of a reference.

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.hpp
```diff
@@ -9,14 +9,28 @@
...
+#include "IO/H5/AccessType.hpp"
+#include "IO/H5/CheckH5.hpp"
+#include "IO/H5/EosTable.hpp"
+#include "IO/H5/File.hpp"
+#include "IO/H5/Header.hpp"
+#include "IO/H5/Helpers.hpp"
+#include "IO/H5/OpenGroup.hpp"
+#include "IO/H5/SourceArchive.hpp"
+#include "IO/H5/Version.hpp"
+#include "IO/H5/Wrappers.hpp"
```
nilsdeppe: I don't think you need any of these includes in the hpp because the cpp file reads in the data, but the class doesn't store the h5 file, it stores the contents. Phrased differently, the hpp doesn't use any of the h5 stuff, so it's only needed in the cpp. Putting it in the cpp will help keep compile times down. In the cpp you only need
```cpp
#include "IO/H5/AccessType.hpp"
#include "IO/H5/EosTable.hpp"
#include "IO/H5/File.hpp"
```

nilsdeppe (changes requested):
You can squash immediately

src/PointwiseFunctions/AnalyticData/GrMhd/CcsnCollapse.cpp
pajkosmi: _Clang tidy was giving me trouble on these lines, saying they should be references; do you suggest taking by value and overriding it with `//NOLINT`?_

edit: disregard above

nilsdeppe (approved):

PR #6696, opened 2025-06-15
Add the mass term to the scalar wave and related potential

SunPeike:
Added the Mass_Squared term and the potential term for the ScalarWave system,. Wrote a small unit test for the potential term.


nilsdeppe:
Please open this to merge into this branch: https://github.com/nilsdeppe/spectre/tree/for_peike_mass_term

It should also compile and pass tests.

PR #6697, opened 2025-06-16
Add Psi_2 in CCE volume

duetosymmetry:
## Proposed changes

This is dependent on #6648, as we need NP spin coeffs to have a (simple)
calculation for Psi_2.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/Cce/LinearOperators.hpp
+++ b/src/Evolution/Systems/Cce/LinearOperators.hpp
@@ -5,13 +5,11 @@
 
 #include <cstddef>
 
+#include "Evolution/Systems/Cce/Tags.hpp"
+#include "NumericalAlgorithms/Spectral/Basis.hpp"
+#include "NumericalAlgorithms/Spectral/Quadrature.hpp"
+#include "NumericalAlgorithms/SpinWeightedSphericalHarmonics/SwshCollocation.hpp"
 #include "Utilities/Gsl.hpp"
-
-/// \cond
-template <size_t Dim>
-class Mesh;
-class ComplexDataVector;
-/// \endcond
 
 namespace Cce {
 /*!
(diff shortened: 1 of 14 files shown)


duetosymmetry:
`ManualDyCompute` should probably go elsewhere... maybe in LinearOperators.hpp?

nilsdeppe (changes requested):
Equation matches documentation. I'll request all the same tests as in your previous PR :)

duetosymmetry:
The main thing I'd like feedback on is where should `ManualDyCompute` live? Also, its code is just copy/pasted (I added possible resizing of the output) from `PreSwshDerivatives<Tags::Dy<Tag>>::apply()`, so there is code duplication. I don't see how to avoid that, though.

duetosymmetry:
I added tests and moved `ManualDyCompute` to `LinearOperators.hpp`. Think this is ready for review.

kidder:
Why name it ManualDyCompute?  I think DyCompute is a sufficient.

duetosymmetry:
@kidder In #6648 I similarly added `ManualSwshDerivativeCompute` . These are to be contrasted with all the automatic stuff that happens in `PreSwshDerivatives.hpp`. I can change the names of both if you want.

kidder:
yes, I would shorten that name as well

duetosymmetry:
@kidder ok, changed both removing 'Manual'.

src/Evolution/Systems/Cce/LinearOperators.hpp
```diff
@@ -29,4 +27,53 @@ namespace Cce {
...
+namespace Tags {
+/// \cond
+struct LMax;
+/// \endcond
+
+/*!
+ * \brief Compute tag for a manually-handled partial y derivative in the volume.
+ *
+ * \details Most of the partial y derivatives are handled by PreSwshDerivatives.
+ * However, that infrastructure is specific to the Cce evolution system itself,
```
nilsdeppe: [optional] `Cce` -> `CCE`?

nilsdeppe (changes requested):
LGTM, one small suggestion.
Could you also squash the first two commits into one so that the code & test are added together? It was easier to review as separate commits, so thank you :)

duetosymmetry:
Squashed the first two and Cce -> CCE in that one place

nilsdeppe (approved):

PR #6702, opened 2025-06-17
Use AMR grids in multigrid solver

nilsvu:
## Proposed changes

AMR grids are kept around in elliptic executables and used by the multigrid solver. This makes AMR and multigrid work very well together.

### Upgrade instructions

In elliptic input files, rename the option `LinearSolver.Multigrid.MaxLevels: Auto` to `InitialCoarseLevels: Auto`. Also add the option `Amr.MaxCoarseLevels: Auto`. Add the AMR phase `UpdateSections` like this:
```yaml
PhaseChangeAndTriggers:
  - Trigger:
      EveryNIterations:
        N: 1
        Offset: 1
    PhaseChanges:
      - VisitAndReturn(EvaluateAmrCriteria)
      - VisitAndReturn(AdjustDomain)
      - VisitAndReturn(UpdateSections)  # < add this line
      - VisitAndReturn(CheckDomain)
```

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
@kidder would you be willing to review this? I'll also rebase as soon as I get the chance (traveling atm)

src/Domain/Creators/RegisterDerivedWithCharm.cpp
```diff
@@ -63,16 +64,20 @@ struct to_grid_map<
...
+
+template <size_t Dim>
+using composition_map = domain::CoordinateMaps::Composition<
+    tmpl::list<Frame::BlockLogical, Frame::Grid, Frame::Inertial>, Dim>;
 }  // namespace
 
 void register_derived_with_charm() {
   using all_maps = tmpl::remove_duplicates<tmpl::append<
-      maps_from_creators<1>, maps_from_creators<2>, maps_from_creators<3>>>;
+      maps_from_creators<1>, maps_from_creators<2>, maps_from_creators<3>,
```
kidder, line 75: why do you need to do this, when it hasn't been needed before?

kidder (commented):
needs a rebase; just one question; otherwise looks good

src/Domain/Creators/RegisterDerivedWithCharm.cpp
nilsvu: I don't remember, I removed it for now and will add it back in when it's needed. It's a map used in some places, so for example it will be needed when deserializing an ElementMap that includes a grid frame, e.g. in AMR projectors with the BBH domain.

nilsvu: The Kerr test failed without this commit, so I added it back in. It's needed for serializing the ElementMap when a grid frame is present for AMR projectors, which is the case for the elliptic BBH domain.

kidder: okay, thanks for checking, and explains why it wasn't hit before

kidder (approved):

PR #6705, opened 2025-06-17
Allow observation of scalar on Ah

guilara:
## Proposed changes

This is what I've done to interpolate the scalar field at the horizon. However, this implementation breaks things in gh since it adds the `CurvedScalarWave::Tags::Psi` tag to certain lists. 

It seems to me that currently, any extra tags to be interpolated to the horizon (regardless of whether additional spatial gradients in the volume are required) have to go through `ParallelAlgorithms/ApparentHorizonFinder/ComputeHorizonVolumeQuantities.hpp`. There is no additional computations needed for the scalar field so maybe it could be avoided.  In the docx of `ComputeHorizonVolumeQuantities`  it even says that "...most computations (e.g. for observers) should be done only on the horizon surface (i.e. after interpolation) as opposed to in the volume; only those computations that require data in the volume (e.g. volume numerical derivatives) should be done here."

I also could not isolate these computations only to the observed Ah (avoiding to do them in the control systems). I tried to do this by passing the relevant source variables to interpolate with template parameter in the apparent horizon aliases (`ParallelAlgorithms/ApparentHorizonFinder/HorizonAliases.hpp)` and in the `interpolator_source_vars` in the executable. This fails to compile because of an assert in Events/Interpolate.hpp:
```
    static_assert(
        std::is_same_v<typename Metavariables::interpolator_source_vars,
                       tmpl::list<InterpolatorSourceVarTags...>>);
```

Maybe one isolate such computations to the ScalarTensor system by passing a flag everywhere in the code where such the horizon aliases are used (including control systems)? 

I don't know much about the interpolation infrastructure so any suggestions would be helpful.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/ParallelAlgorithms/ApparentHorizonFinder/ComputeHorizonVolumeQuantities.hpp
+++ b/src/ParallelAlgorithms/ApparentHorizonFinder/ComputeHorizonVolumeQuantities.hpp
@@ -23,6 +23,9 @@ namespace gsl {
 template <typename T>
 class not_null;
 }  // namespace gsl
+namespace CurvedScalarWave::Tags {
+struct Psi;
+}
 /// \endcond
 
 namespace ah {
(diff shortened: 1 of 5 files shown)


knelli2:
As part of my horizon finder changes I plan to add support for observing arbitrary tensors/scalars on the horizon by just adding a tag to a `tmpl::list` just like observation is done. However, this is still a ways away as there are some other more pressing changes to be made first. I'd have to think about how to do this before those changes if needed.

PR #6708, opened 2025-06-18
Add polar to cartesian coordinate map

kidder:
## Proposed changes

Adds a polar to cartesian coordinate map.  This will be needed for annuli, disks, cylindrical shells, and cylinders that use topologies S1 or B2.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Domain/CoordinateMaps/PolarToCartesian.cpp
+++ b/src/Domain/CoordinateMaps/PolarToCartesian.cpp
@@ -1,0 +1,101 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#include "Domain/CoordinateMaps/PolarToCartesian.hpp"
+
+#include <cmath>
+#include <cstddef>
+#include <pup.h>
+
+#include "DataStructures/DataVector.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/DereferenceWrapper.hpp"
+#include "Utilities/GenerateInstantiations.hpp"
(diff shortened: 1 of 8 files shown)


src/Domain/CoordinateMaps/PolarToCartesian.cpp
```diff
@@ -0,0 +1,100 @@
...
+
+// NOLINTNEXTLINE(readability-convert-member-functions-to-static)
+std::optional<std::array<double, 2>> PolarToCartesian::inverse(
+    const std::array<double, 2>& target_coords) const {
+  const auto& [x, y] = target_coords;
+  if (UNLIKELY(y == 0.0 and x == 0.0)) {
+    return std::array{0.0, 0.0};
+  } else {
+    const double r = std::hypot(x, y);
+    return std::array{r, y < 0.0 ? atan2(y, x) + 2.0 * M_PI : atan2(y, x)};
```
wthrowe: Issues when `x < 0` and `y` is negative 0.  Instead, do
```c++
const double phi = atan2(y, x);
return std::array{r, phi < 0.0 ? phi + 2.0 * M_PI : phi};
```
For the test:
```c++
test_map_at_point(map, {{1.0, M_PI}}, {{-1.0, std::copysign(0.0, -1.0)}});
```
(There are more natural ways to get negative zero, but for a test best to be obvious.)

Same for the pfaffian version.

wthrowe (changes requested):

wthrowe (approved):

PR #6710, opened 2025-06-20
some necessary changes for building on frontera

jyoo1042:
## Proposed changes

- The default gcc version on Frontera is too outdated for building latest version of SpECTRE.
- More recent versions of gcc are available on Frontera; however, they do not have any recent version of Python installed.
- Add a new option ENABLE_PYTHON (default is ON) so that we can easily disable usage of Python
  for minimal building on Frontera without recent installation of Python. We still need some version of
Python (>=2.7) for linking purpose.
- Modify default version of modules and default cmake options for the Frontera environment file.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SpectreSetupPythonPackage.cmake
+++ b/cmake/SpectreSetupPythonPackage.cmake
@@ -1,120 +1,121 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
-spectre_define_test_timeout_factor_option(PYTHON "Python")
+if (ENABLE_PYTHON)
+  spectre_define_test_timeout_factor_option(PYTHON "Python")
 
-option(PY_DEV_MODE "Enable development mode for the Python package, meaning \
-that Python files are symlinked rather than copied to the build directory" OFF)
-function(configure_or_symlink_py_file SOURCE_FILE TARGET_FILE)
-  if(PY_DEV_MODE)
-    get_filename_component(_TARGET_FILE_DIR ${TARGET_FILE} DIRECTORY)
-    execute_process(COMMAND
-      ${CMAKE_COMMAND} -E make_directory ${_TARGET_FILE_DIR})
-    execute_process(COMMAND
-      ${CMAKE_COMMAND} -E create_symlink ${SOURCE_FILE} ${TARGET_FILE})
-  else()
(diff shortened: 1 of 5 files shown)


cmake/SpectreSetupPythonPackage.cmake
```diff
@@ -1,6 +1,7 @@
 # Distributed under the MIT License.
 # See LICENSE.txt for details.
 
+if (ENABLE_PYTHON)
 spectre_define_test_timeout_factor_option(PYTHON "Python")
```
wthrowe: Fix indentation.

support/Environments/frontera_gcc.sh
```diff
@@ -76,14 +76,15 @@ spectre_run_cmake() {
...
           -D MEMORY_ALLOCATOR=SYSTEM \
-          -D BUILD_PYTHON_BINDINGS=ON \
-          -D Python_EXECUTABLE=`which python3` \
+          -D BUILD_TESTING=OFF \
+          -D BUILD_PYTHON_BINDINGS=OFF \
+          -D BUILD_DOCS=OFF \
           -D USE_LD=ld \
-          -D SPECTRE_TEST_RUNNER="$(pwd)/bin/charmrun" \
           "$@" \
           $SPECTRE_HOME
```
wthrowe: You presumably want to set the new ENABLE_PYTHON option here.

wthrowe (changes requested):
@sxs-collaboration/spectre-core-devs I was pretty closely involved in writing this, so it should probably get a second review.  All my requests here are trivial changes.

cmake/SpectreSetupPythonPackage.cmake
jyoo1042: done.

support/Environments/frontera_gcc.sh
jyoo1042: done.

wthrowe:
Fixups look good.  Go ahead and squash.

wthrowe:
You've squashed some of the fixups into the wrong commit.  Should be an easy fix, since the commits touch different files.

wthrowe (approved):

PR #6711, opened 2025-06-20
Xcts: allow superposed lapse as horizon boundary condition

nilsvu:
## Proposed changes

This is SpEC's choice, so we also allow it to compare to SpEC.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
+++ b/src/Elliptic/Systems/Xcts/BoundaryConditions/ApparentHorizon.cpp
@@ -17,6 +17,7 @@
 #include "Elliptic/BoundaryConditions/BoundaryCondition.hpp"
 #include "Elliptic/Systems/Xcts/Geometry.hpp"
 #include "NumericalAlgorithms/DiscontinuousGalerkin/NormalDotFlux.hpp"
+#include "PointwiseFunctions/AnalyticData/Xcts/Binary.hpp"
 #include "PointwiseFunctions/AnalyticSolutions/Xcts/Factory.hpp"
 #include "PointwiseFunctions/GeneralRelativity/Christoffel.hpp"
 #include "PointwiseFunctions/InitialDataUtilities/InitialGuess.hpp"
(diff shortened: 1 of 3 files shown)


src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
```diff
@@ -267,7 +267,7 @@ class KerrSchild : public AnalyticSolution<3_st>,
       "Black hole in Kerr-Schild coordinates"};
 
   KerrSchild(double mass, const std::array<double, 3>& dimensionless_spin,
-             const std::array<double, 3>& center,
+             const std::array<double, 3>& center = {{0., 0., 0.}},
```
nilsdeppe: Why do we need a default? In general we've had a theme of minimizing defaults to increase transparency.

nilsdeppe (changes requested):

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/KerrSchild.hpp
nilsvu: The center seems like an optional thing to move away from zero and constructing a Kerr solution as `KerrSchild{mass, spin}` seems rather obvious, so I think the default makes sense here. The boost velocity also has a default value (zero) for a similar reason.

nilsdeppe: I think the boost has a default because it was added later and so changing it everywhere would be tedious. Here we already have the code without the default, so I don't think it's quite the same reasoning behind it

nilsvu: ok, I do think the default makes sense, but I can remove it if you prefer

nilsdeppe: I have a slight preference, but would like Mark & Will's input. Maybe ping them on Slack to see how they feel? Mark write KS, I think.

nilsvu: I just removed the default to avoid holding up this PR

nilsdeppe (approved):

PR #6699, opened 2025-06-16
Change horizon finder to be a singleton

knelli2:
## Proposed changes

Opening as a draft for the moment to share code. Will split this into many smaller PRs later on.

### Code review checklist

- [ ] The code is documented and the documentation renders correctly. Run
  `make doc` to generate the documentation locally into `BUILD_DIR/docs/html`.
  Then open `index.html`.
- [ ] The code follows the stylistic and code quality guidelines listed in the
  [code review guide](https://spectre-code.org/code_review_guide.html).
- [ ] The PR lists upgrade instructions and is labeled `bugfix` or
  `new feature` if appropriate.

PR #6712, opened 2025-06-23
Fix registration for interpolator component

knelli2:
## Proposed changes

Fixes a bug where, because the interpolator wasn't checkpointed, a tag in the interpolator was registered incorrectly causing an error down the line during the evolution. Now the interpolator is checkpointed and registered properly.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### Further comments

Technical details: The tag `intrp::Tags::NumberOfElements` is using in `src/ParallelAlgorithms/Interpolation/Actions/TryToInterpolate.hpp:207` to ensure the local interpolator core has received data from all elements it expects to for each target. It must have a key for each target in the metavars, even if the value is an empty set. This is enforced by the initialization action for the interpolator. Then the registration action, is what actually populates the tag. However, before, the interpolator wasn't checkpointed so upon a restart, the tag was empty and the registration wouldn't add keys for all targets in the metavars, only the ones specifically necessary for that core. This then caused the check in `TryToInterpolate.hpp` to fail. The solution is to 1) checkpoint the interpolator, and 2) allow the registration action to be run again. This is just a quick fix. There's probably a cleaner solution which involves changing the logic in `TryToInterpolate`, but a) I unfortunately don't have the time to look at this right now and won't for another week or more, b) this actively is preventing people from running BBHs, and c) this will all be mute once #6699 is merged. So I decided not to implement a "cleaner" solution.

### DIFF
--- a/src/ParallelAlgorithms/Interpolation/Actions/InterpolatorRegisterElement.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Actions/InterpolatorRegisterElement.hpp
@@ -87,13 +87,7 @@ struct RegisterElement {
             }
 
             if (blocks_to_interpolate.at(target_name).contains(block_name)) {
-              auto inserted = (*num_elements)[target_name].insert(element_id);
-              if (not inserted.second) {
-                ERROR("Unable to insert element "
-                      << element_id << " into interpolator core "
-                      << Parallel::my_proc<size_t>(cache) << " for target "
-                      << target_name);
-              }
+              (*num_elements)[target_name].insert(element_id);
             }
           });
         },
--- a/src/ParallelAlgorithms/Interpolation/Actions/TryToInterpolate.hpp
+++ b/src/ParallelAlgorithms/Interpolation/Actions/TryToInterpolate.hpp
@@ -201,7 +201,7 @@ void try_to_interpolate(
 
   // Send interpolated data only if interpolation has been done on all
   // of the local elements.
-  const auto all_num_elements =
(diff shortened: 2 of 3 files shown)


nilsvu (approved):

wthrowe:
If you want me to review this, you're going to have to give me more to go on than "a tag in the interpolator was registered incorrectly causing an error down the line during the evolution".

knelli2:
@wthrowe Updated the description.

nilsvu:
For a bit more context @wthrowe : I think your recent change to run registration actions in a new restart phase means that registration actions have to be able to rerun over previously registered state (not sure whether or not that was true before). So I think we have to decide whether we want to deregister before checkpointing and then register at restart, or guarantee that registration actions are able to handle previously registered state. Do you have thoughts on that?

knelli2:
@guilara Confirmed this fixed the issue by doing a checkpoint/restart 5 times without issue.

wthrowe (approved):
I think a cleaner solution would have been to rerun `InitializeInterpolator` in the `Restart` phase, although it would need a modification to not overwrite existing entries (just removing ` = std::unordered_set<ElementId<Dim>>{}` would do it), but if this stuff is all going to be deleted anyway this is fine as a temporary fix.

On the topic of repeated registration in general, not checkpointing the registration data is intentional, because it is potentially wrong after a restart.  The only other way to handle it safely would be to deregister everything before a checkpoint, but this would require running another two phases during checkpointing (for deregistering and reregistering).

PR #6717, opened 2025-06-27
Add branch runs command to pipeline

guilara:
## Proposed changes

Add command to set up directories for branching after junk.

Need further changes in the eccentricity reduction procedure to save volume data needed for restarting after PB&J time.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/BranchRuns.py
+++ b/support/Pipelines/Bbh/BranchRuns.py
@@ -158,7 +158,10 @@ def branch_runs(
     "--requested-levs",
     type=click.IntRange(-2, 10),
     nargs=2,
-    help="Minimum and maximum Levs used for branching.",
+    help=(
+        "Minimum and maximum Levs used for branching. LevN corresponds to a"
+        " polynomial order of P=N+7."
+    ),
 )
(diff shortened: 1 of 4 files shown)


support/Pipelines/Bbh/BranchRuns.py
```diff
@@ -0,0 +1,228 @@
...
+logger = logging.getLogger(__name__)
+
+from spectre.Pipelines.Bbh.Inspiral import (
+    INSPIRAL_INPUT_FILE_TEMPLATE,
+    start_inspiral,
+)
+from spectre.support.Schedule import scheduler_options
+
+# Levs defined in terms of p-refinement
+ALL_LEVS = [
```
nilsvu, line 19: I think this is a good idea for defining Levs. I suggest you move this definition into `Inspiral.py` and (as a first commit and ideally even as a separate PR) add a `--lev` option to `start-inspiral` as an alternative to the `-L / -P` options.

Note that we probably want to define Levs in terms of AMR tolerances eventually, but we can make this change later.

PR #6722, opened 2025-07-01
Define resolution levels and use in inspiral command

guilara:
## Proposed changes

Define resolution levels as in the bbh paper (arXiv: 2410.00265): For N integer, LevN corresponds to h-refinement L = 1 and p-refinement P = 7 + N (to be replaced once AMR is used). Use in inspiral command.  Specify default in terms of a level when the 'evolve' flag is passed to 'generate-id'.

Shall I apply these changes to 'start-ringdown'?

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/Inspiral.py
+++ b/support/Pipelines/Bbh/Inspiral.py
@@ -20,6 +20,17 @@ logger = logging.getLogger(__name__)
 
 INSPIRAL_INPUT_FILE_TEMPLATE = Path(__file__).parent / "Inspiral.yaml"
 
+# Resolution levels defined in terms of p-refinement
+# To be replaced once AMR is used.
+INSPIRAL_LEVS = {
+    lev_number: {
+        "label": f"Lev{lev_number}",
+        "refinement_level": 1,
+        "polynomial_order": 7 + lev_number,
+    }
+    for lev_number in range(-2, 11)
+}
+
 
 # These parameters come from empirically tested values in SpEC and SpECTRE
 def _control_system_params(
@@ -363,8 +374,9 @@ def inspiral_parameters_spec(
 
 def start_inspiral(
     id_input_file_path: Union[str, Path],
-    refinement_level: int = 1,
-    polynomial_order: int = 8,
+    lev: Optional[int] = None,
+    refinement_level: Optional[int] = None,
+    polynomial_order: Optional[int] = None,
     id_run_dir: Optional[Union[str, Path]] = None,
     inspiral_input_file_template: Union[
         str, Path
@@ -400,6 +412,22 @@ def start_inspiral(
         "Cannot enable both 'continue_with_ringdown' and"
         " 'eccentricity_control'. Choose which of the two to perform next."
     )
+
+    if lev is not None:
+        assert (refinement_level is None) and (polynomial_order is None), (
+            "The option 'lev' is mutaully exclusive with 'refinement_level' and"
+            " 'polynomial_order'."
+        )
+        selected_lev = INSPIRAL_LEVS[lev]
+        refinement_level = selected_lev["refinement_level"]
+        polynomial_order = selected_lev["polynomial_order"]
+    else:
+        assert (refinement_level is not None) and (
+            polynomial_order is not None
+        ), (
+            "Resolution not specified. Provide either 'lev' or both"
+            " 'refinement_level' and 'polynomial_order'."
+        )
 
     # Determine inspiral parameters from initial data
     if id_run_dir is None:
@@ -542,20 +570,26 @@ def start_inspiral(
     ),
 )
 @click.option(
+    "--lev",
+    type=int,
+    help=(
+        "Resolution levels defined in terms of h and p refinement. Can be"
+        " specified multiple times. For integer N, LevN corresponds to"
+        " refinement level L = 1 and polynomial order P = 7 + N. Mutually"
+        " exclusive with options '-L' and '-P'"
+    ),
+)
+@click.option(
     "--refinement-level",
     "-L",
     type=int,
     help="h-refinement level.",
-    default=1,
-    show_default=True,
 )
 @click.option(
     "--polynomial-order",
     "-P",
     type=int,
     help="p-refinement level.",
-    default=8,
-    show_default=True,
 )
 @click.option(
     "--continue-with-ringdown",
--- a/support/Pipelines/Bbh/PostprocessId.py
+++ b/support/Pipelines/Bbh/PostprocessId.py
@@ -182,6 +182,7 @@ def postprocess_id(
             continue_with_ringdown=not eccentricity_control,
             eccentricity_control=eccentricity_control,
             pipeline_dir=pipeline_dir,
+            lev=1,
             **scheduler_kwargs,
         )
 
--- a/tests/support/Pipelines/Bbh/Test_Inspiral.py
+++ b/tests/support/Pipelines/Bbh/Test_Inspiral.py
@@ -121,10 +121,6 @@ class TestInspiral(unittest.TestCase):
     def test_cli(self):
         common_args = [
             str(self.id_dir / "InitialData.yaml"),
-            "--refinement-level",
-            "1",
-            "--polynomial-order",
-            "5",
             "--id-horizons-path",
             str(self.horizons_filename),
             "-E",
@@ -139,6 +135,10 @@ class TestInspiral(unittest.TestCase):
             start_inspiral_command(
                 common_args
                 + [
+                    "--refinement-level",
+                    "1",
+                    "--polynomial-order",
+                    "5",
                     "-O",
                     str(self.test_dir / "Inspiral"),
                     "--no-submit",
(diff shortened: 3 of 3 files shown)


support/Pipelines/Bbh/Inspiral.py
```diff
@@ -20,6 +20,25 @@
...
+# To be replaced once AMR is used.
+INSPIRAL_LEVS = [
+    {
+        "Lev number": lev_number,
+        "Label": f"Lev{lev_number}",
+        "refinement level": 1,
+        "polynomial order": 7 + lev_number,
+    }
+    for lev_number in range(-2, 11)
+]
```
nilsvu: I suggest you make this a dict with the lev_number as key, then you can look up levs directly like `INSPIRAL_LEVS[lev]` without searching in a loop

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -20,6 +20,25 @@
...
+
+# If no resolution is specified, use a default level
+DEFAULT_LEV = None
```
nilsvu: Maybe we don't actually need a default lev. I suggest we try without a default for now and just require people to either specify lev or both L and P.

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -401,6 +421,50 @@ def start_inspiral(
...
+        assert (
+            selected_lev is not None
+        ), "Could not select lev from 'INSPIRAL_LEVS'. Possibily undefined."
+        refinement_level = selected_lev["refinement level"]
+        polynomial_order = selected_lev["polynomial order"]
+
+    assert (refinement_level is not None) and (polynomial_order is not None), (
+        "Either 'refinement_level' or 'polynomial_order' is not properly"
+        " specified."
+    )
```
nilsvu: I think most of this added code and the asserts can be trimmed down to this:

```py
if lev is not None:
    assert refinement_level is None and polynomial_order is None, "..."
    selected_lev = INSPIRAL_LEVS[lev]
    # set L, P
else:
    assert refinement_level is not None and polynomial_order is not None, "..."
```

I find this if-else with nested asserts easier to understand.

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -541,21 +605,26 @@ def start_inspiral(
         " not needed if you are starting from a SpEC ID_Params.perl file."
     ),
 )
+@click.option(
+    "--lev",
+    type=int,
+    help=(
+        "Resolution levels defined in terms of h and p refinement. For integer"
+        " N, LevN corresponds to refinement level L = 1 and polynomial order P"
+        " = 7 + N."
```
nilsvu: Add "mutually exclusive with options '-L' and '-P'".

support/Pipelines/Bbh/PostprocessId.py
```diff
@@ -173,14 +173,15 @@ def postprocess_id(
...
-        from spectre.Pipelines.Bbh.Inspiral import start_inspiral
+        from spectre.Pipelines.Bbh.Inspiral import DEFAULT_LEV, start_inspiral
 
         start_inspiral(
             id_input_file_path,
             id_run_dir=id_run_dir,
             continue_with_ringdown=not eccentricity_control,
             eccentricity_control=eccentricity_control,
             pipeline_dir=pipeline_dir,
+            lev=DEFAULT_LEV["Lev number"],
```
nilsvu: Just `lev=1` should be fine here for now

nilsvu (changes requested):

nilsvu:
Thanks for adding this @guilara!

guilara:
@nilsvu  I've done the changes requested. It does look nice this way indeed.

guilara:
@guilara But, should we use these levs for the ringdown as well?

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -22,22 +22,17 @@
...
+for lev_number in range(-2, 11):
+    INSPIRAL_LEVS.update(
+        {
+            lev_number: {
+                "Label": f"Lev{lev_number}",
+                "refinement level": 1,
+                "polynomial order": 7 + lev_number,
+            }
+        }
+    )
```
nilsvu:
```suggestion
INSPIRAL_LEVS = {
     lev_number: {
         "label": f"Lev{lev_number}",
         "refinement_level": 1,
         "polynomial_order": 7 + lev_number,
     }
     for lev_number in range(-2, 11)
 }
```

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -421,49 +416,21 @@ def start_inspiral(
...
-        ), (
-            "Specify a resolution to run. Either 'lev' or both"
-            " 'refinement_level' and 'polynomial_order'."
-        )
-
-    selected_lev = None
     if lev is not None:
         assert (refinement_level is None) and (polynomial_order is None), (
-            "Cannot specify both 'lev' together with 'refinement_level' and"
+            "The option 'lev' is mutaully exclusive with 'refinement_level' and"
```
nilsvu, line 418: typo "mutually"

nilsvu (changes requested):
LGTM please squash (including this).

Now to branch levs I wonder if the easiest solution is to allow for `start-inspiral --lev=1,2,3` (allow a list in addition to a single lev for this option, or add a second option `--levs` that takes a list). Has the advantage that we don't need an extra CLI command like `branch-levs` which has to forward/repeat all the options for `start-inspiral`. I think it's enough to handle the list of levs in the CLI command (`start_inspiral_command`), not in the `start_inspiral()` Python function, because in Python you can just easily call `start_inspiral` in a loop.

support/Pipelines/Bbh/Inspiral.py
```diff
@@ -363,8 +377,9 @@ def inspiral_parameters_spec(
 
 def start_inspiral(
     id_input_file_path: Union[str, Path],
-    refinement_level: int = 1,
-    polynomial_order: int = 8,
+    lev: Optional[int] = None,
```
nilsvu, line 377: In line 487 below I suggest you add the lev to the directory label:

```py
lev_suffix = ("_" + selected_lev["label"]) if lev is not None else ""
segments_dir = pipeline_steps[-1].next(label="Inspiral" + lev_suffix).path
```

guilara: I had in mind to generate subpipelines `Lev1`, etc . separate from the pipeline steps, so that within those subdirectories they can have their own independent steps (Inspiral, Ringdown, Cce).

guilara: I would worry one would end up with a messy directory: e.g. Pip:..., 007_Inspiral_Lev1, 008_Inspiral_Lev2, 009_Ringdown_Lev1, 010_Ringdown_Lev2, 011_Cce, 012_Ringdown_Lev3, 013_Cce, ...

nilsvu: Ah I see. That works too. I think I'd prefer to keep directory structure simpler for now and just label the pipeline step with the Lev. Not set in stone of course.

nilsvu: > I would worry one would end up with a messy directory: e.g. Pip:..., 007_Inspiral_Lev1, 008_Inspiral_Lev2, 009_Ringdown_Lev1, 010_Ringdown_Lev2, 011_Cce, 012_Ringdown_Lev3, 013_Cce, ...

Yes this would be the outcome for now. We could simplify this by not incrementing the pipeline step number for each lev:
```
000_InitialData_Ecc0
001_Inspiral_Ecc0  # will be at some lower lev
002_InitialData_Ecc1
003_Inspiral_Ecc1  # will be at highest lev
004_Inspiral_Lev2
004_Inspiral_Lev1
004_Inspiral_Lev0
005_Ringdown_Lev2
005_Ringdown_Lev1
005_Ringdown_Lev0
```

guilara: I'm just saying bc I'm facing this issue already. Got a run with 8 steps of ecc reduction and  saving PB&J data and then 5 levs (for a two dimensional convergence test). Planning to follow to ringdown and cce.

nilsvu: Hm ok, I was trying to avoid making the directory structure too complicated with nested subdirectories. I think what you are suggesting is:

```
000_InitialData_Ecc0
001_Inspiral_Ecc0  # will be at some lower lev
002_InitialData_Ecc1
003_Inspiral_Ecc1  # will be at highest lev
004_Lev0/
    000_Inspiral
    001_Ringdown
    002_Cce
005_Lev1/
    000_Inspiral
    001_Ringdown
    002_Cce
006_Lev2/
    000_Inspiral
    001_Ringdown
    002_Cce
007_Archive
```

I think this can work too, I'm just a bit worried about adding this new complexity of "subpipelines" and having to go back up to the main pipeline for running steps like archiving.

nilsvu (commented):
@guilara could you update this PR with the small changes requested above and squash, so we can merge it? Skip the directory name comment for now.

guilara:
> @guilara could you update this PR with the small changes requested above and squash, so we can merge it? Skip the directory name comment for now.

Yes, will get back to it probably by the end of the week. Have a grant application deadline soon.

guilara:
@nilsvu I've rebased this and squashed just with the change to the lev dict and ignoring the directory stuff.

support/Pipelines/Bbh/Inspiral.py
guilara: Yes, it works, just tried it. I can make set multiple=True for the lev option and then deal with it in the cli command. But then I need to make a choice about the directory structure, so I've skipped it for now.

nilsvu (approved):
Thanks @guilara !

PR #6726, opened 2025-07-02
Add TrumpetSchwarzschild GR analytic solution

hen-w:
## Proposed changes
Add the [trumpet Schwarzchild solution](https://journals.aps.org/prd/abstract/10.1103/PhysRevD.78.064020) in isotropic coordinates for evolution test of SO-CCZ4. This solution is not intended to be used with GH.

Following the above paper, we first solve for the lapse and the Schwarzchild R on a selected source grid labeled by the isotropic r by eq. (54)-(56), where the Schwarzchild R as a function of lapse is solved using eq. (39) and choosing the physical branch with the correct asymptotic behaviors. Note that eq. (43) is missing a factor of mass^4. Also note that the integrals in eq. (54)-(56) are not very well-behaved so we use a tanh-sinh integrator. We then interpolate the lapse and the Schwarzchild R to the user-specified grid of isotropic r and calculate all the 3+1 quantities per the paper.

We use dimensionless units (in unit of the black hole mass) in the above process to insulate our implementation from different mass parameter and only restore units when returning 3+1 quantities to the user. The other parameter n in the 1+log slicing condition should be chosen sensibly (around n=2, the classic choice in 1+log slicing).

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsvu:
Quick drive-by comment: it's `Schwarzschild` with an `s` :)

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.hpp
```diff
@@ -0,0 +1,246 @@
...
+        "Mass of the Schwarzchild black hole"};
+    static type lower_bound() { return 0.1; };
+  };
+
+  // currently we have only tested for N=2; for other N eq. (39) may need to be
+  // solved differently
+  struct N {
+    using type = double;
+    static constexpr Options::String help = {
+        "Parameter of the trumpet solution family"};
```
nilsdeppe: Add caveat about `N=2` in the help string

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.hpp
```diff
@@ -0,0 +1,246 @@
...
+
+  struct Mass {
+    using type = double;
+    static constexpr Options::String help = {
+        "Mass of the Schwarzchild black hole"};
+    static type lower_bound() { return 0.1; };
+  };
+
+  // currently we have only tested for N=2; for other N eq. (39) may need to be
+  // solved differently
```
nilsdeppe: Promote to doxygen and `C`

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.hpp
```diff
@@ -0,0 +1,246 @@
...
+                     std::array<DataVector, 2>& data_on_source_grid);
+    // the lapse, Schwarzchild R, and isotropic r on isotropic grid points
+    DataType lapse_on_user_grid{};
+    DataType schwarzchild_r_on_user_grid{};
+    DataType d_lapse_d_schwarzchild_r_on_user_grid{};
+    DataType isotropic_r_on_user_grid{};
+  };
+
+  double mass_{1.0};
+  double n_{2.0};
```
nilsdeppe: default to `std::numeric_limits<double>::signaling_NaN()`, then if you have a bug in serialization, etc. it'll be very obvious

hen-w:
> Quick drive-by comment: it's `Schwarzschild` with an `s` :)

Thanks! I will change that later along with other suggestions.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+
+// Element placement in isotropic coordinates for interpolation to user grid
+// It is assumed that element_lower_bounds and element_upper_bounds are sorted
+// in asending order, and element_lower_bounds[i] < element_upper_bounds[i]
+constexpr std::array element_lower_bounds{0.,  1e-4, 1e-3, 1e-2, 1e-1,  0.5,
+                                          1.,  2.,   4.,   8.,   16.,   32.,
+                                          64., 128., 256., 512., 1024., 2048.};
+constexpr std::array element_upper_bounds{
+    1e-4, 1e-3, 1e-2, 1e-1, 0.5,  1.,   2.,    4.,    8.,
+    16.,  32.,  64.,  128., 256., 512., 1024., 2048., max_isotropic_r};
```
nilsdeppe: Can you set this using a lambda so that someone only needs to change the `element_lower_bounds` instead of trying to keep both in sync?

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+                          Spectral::Quadrature::GaussLobatto};
+
+using Affine = domain::CoordinateMaps::Affine;
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+boost::math::quadrature::tanh_sinh<double> de_integrator;
+
+// Get the critical lapse from eq. (42)
+// for n=2 critical lapse ~0.16
+double get_crit_lapse(const double n) {
+  return sqrt((sqrt(4 + 9 * n * n) - 3 * n) / (sqrt(4 + 9 * n * n) + 3 * n));
```
nilsdeppe: we prefer `4.` or `4.0` (either is fine, just be consistent throughout a file) since this makes it clear you want to be doing floating point math (it signals intent)

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+}
+
+// Get the critical Schwarzchild radius R_c from eq. (41)
+double get_crit_schwarzchild_r(const double n) {
+  return ((3 * n * n + sqrt(4 * pow(n, 2) + 9 * pow(n, 4))) / (4 * n * n));
```
nilsdeppe: using `pow<2>(n)` and `pow<4>(n)` will be considerably faster since they convert to multiplies at compile time. We also have `integer_pow` if you don't know that it's integer math until runtime

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+double get_c_n_squared(const double n) {
+  return pow(3 * n + sqrt(4 + 9 * n * n), 3) / (128 * pow(n, 3)) *
+         exp(-2 * get_crit_lapse(n) / n);
+}
+
+// Get the function value of the transedental equation eq. (39)
+double function_of_lapse_and_schwarzchild_r(const double lapse,
+                                            const double schwarzchild_r,
+                                            const double n) {
+  ASSERT((lapse >= 0) and (lapse < 1), "ERROR: invalid lapse range.");
```
nilsdeppe: The `ERROR:` part is printed as part of the error message by `ASSERT` (it's actually an ASSERT message not an ERRROR message). It's also important to give the expected range and the lapse. E.g. `, "Lapse must be in range [0, 1) but is " << lapse);`

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+    16.,  32.,  64.,  128., 256., 512., 1024., 2048., max_isotropic_r};
+static_assert(element_lower_bounds.size() == element_upper_bounds.size());
+constexpr size_t num_of_elements = element_lower_bounds.size();
+
+const Mesh<1> source_mesh{num_of_pts, Spectral::Basis::Legendre,
+                          Spectral::Quadrature::GaussLobatto};
+
+using Affine = domain::CoordinateMaps::Affine;
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+boost::math::quadrature::tanh_sinh<double> de_integrator;
```
nilsdeppe: Is it actually expensive to inline construct the integrator? I have mixed feelings since this is in an anonymous namespace and so cannot be accessed outside of this translation unit. It's also not clear to me what tolerances are being used. Could you add a comment about that? And I assume this is a default constructor call, so doing that explicitly via `de_integrator{};` would be useful to aid readability

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+                                  const double target_isotropic_r,
+                                  const double crit_lapse,
+                                  const double crit_schwarzchild_r,
+                                  const double c_0) {
+  ASSERT(target_isotropic_r >= 0.0,
+         "ERROR: isotropic r must be "
+         "non-negative");
+  ASSERT(target_isotropic_r <= max_isotropic_r,
+         "ERROR: we do not support"
+         " trumpet initial data with isotropic r greater than "
```
nilsdeppe: remove `ERROR:`. If you want this to be an ERROR instead of an ASSERT, make it so :) Same for below

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+  } else {
+    // define the upper bound of lapse for root finding
+    const double max_lapse = sqrt(1.0 - 1.0 / max_isotropic_r);
+    const auto isotropic_r_minus_target = [&](const double lapse) -> double {
+      return (get_isotropic_r_from_lapse(n, lapse, crit_lapse,
+                                         crit_schwarzchild_r, c_0) -
+              target_isotropic_r);
+    };
+    return RootFinder::toms748(isotropic_r_minus_target, 0.0, max_lapse, 1e-15,
+                               1e-15);
```
nilsdeppe: `1e` -> `1.0e`

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+                                  const double crit_schwarzchild_r,
+                                  const double c_0) {
+  ASSERT(target_isotropic_r >= 0.0,
+         "ERROR: isotropic r must be "
+         "non-negative");
+  ASSERT(target_isotropic_r <= max_isotropic_r,
+         "ERROR: we do not support"
+         " trumpet initial data with isotropic r greater than "
+             << max_isotropic_r);
+  if (target_isotropic_r == 0) {
```
nilsdeppe: `0.0`

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+  const DataVector used_for_size(num_of_elements * num_of_pts,
+                                 std::numeric_limits<double>::signaling_NaN());
+  auto source_grid = make_with_value<tnsr::I<DataVector, 1, Frame::Inertial>>(
+      used_for_size, 0.0);
+  const auto logical_coords = logical_coordinates(source_mesh);
+
+  for (size_t i = 0; i < num_of_elements; ++i) {
+    const auto coord_map =
+        domain::make_coordinate_map<Frame::ElementLogical, Frame::Inertial>(
+            Affine{-1., 1., element_lower_bounds.at(i),
```
nilsdeppe: `1.0` (or change rest of file to `1.` type)

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+  const double c_0 = get_c_0(n, crit_lapse, crit_schwarzchild_r);
+
+  data_on_source_grid_.at(0) =
+      get_lapse_on_grid(n, source_grid_, crit_lapse, crit_schwarzchild_r, c_0);
+  data_on_source_grid_.at(1) = get_schwarzchild_r_on_grid(
+      n, data_on_source_grid_.at(0), crit_lapse, crit_schwarzchild_r);
+}
+
+void TrumpetSchwarzchild::pup(PUP::er& p) {
+  p | mass_;
```
nilsdeppe: I think you should call the base class pup functions too

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+  return isotropic_r_on_grid;
+}
+}  // namespace
+
+namespace gr::Solutions {
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+tnsr::I<DataVector, 1, Frame::Inertial> TrumpetSchwarzchild::source_grid_ =
+    set_source_grid();
+
+TrumpetSchwarzchild::TrumpetSchwarzchild(CkMigrateMessage* /*msg*/) {}
```
nilsdeppe: pass `msg` to base class' migration constructor

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+    const double mass) {
+  tnsr::I<DataType, 1, Frame::Inertial> isotropic_r_on_grid{
+      sqrt(pow(get<0>(x), 2) + pow(get<1>(x), 2) + pow(get<2>(x), 2)) / mass};
+  return isotropic_r_on_grid;
+}
+}  // namespace
+
+namespace gr::Solutions {
+// NOLINTNEXTLINE(cppcoreguidelines-avoid-non-const-global-variables)
+tnsr::I<DataVector, 1, Frame::Inertial> TrumpetSchwarzchild::source_grid_ =
```
nilsdeppe: Can you not put this in the anonymous namespace?

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.hpp
```diff
@@ -0,0 +1,246 @@
...
+/*!
+ * \brief Trumpet Schwarzchild solution in isotropic coordinates
+ *
+ * \details
+ * This solution is a trumpet Schwarzchild black hole in the isotropic
+ * coordinates. It is time-independent puncture solution in 1+log slicing.
+ * The solution cannot be written down analytically in Schwarzchild
+ * coordinates or isotropic coordinates. Refer to \cite Hannam2008
+ * for equations and details.
+ *
```
nilsdeppe: Given that going from the paper to an algorithm was non-trivial, could you give a several paragraph overview of how this solution is being implemented and solved for?

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+}
+
+// We first compute lapse, schwarzchid_r, d_lapse_d_schwarzchild_r
+// on a source isotropic grid before interpolating to the user-
+// specified isotropic grid corresponding to x
+template <typename DataType>
+TrumpetSchwarzchild::IntermediateVars<DataType>::IntermediateVars(
+    const double mass, const double n,
+    const tnsr::I<DataType, volume_dim, Frame::Inertial>& x, const double /*t*/,
+    std::array<DataVector, 2>& data_on_source_grid) {
```
nilsdeppe: I feel like this ought to be a `const &`, no?

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+    const IntermediateVars<DataType>& vars,
+    tmpl::list<DerivLapse<DataType>> /*meta*/) const
+    -> tuples::TaggedTuple<DerivLapse<DataType>> {
+  tnsr::i<DataType, volume_dim, Frame::Inertial> d_lapse;
+  for (size_t i = 0; i < volume_dim; ++i) {
+    // note that we divide by mass_ to get the correct dimension
+    d_lapse.get(i) = vars.d_lapse_d_schwarzchild_r_on_user_grid *
+                     vars.lapse_on_user_grid *
+                     vars.schwarzchild_r_on_user_grid * x.get(i) / mass_ /
+                     pow(vars.isotropic_r_on_user_grid, 2) / mass_;
```
nilsdeppe: it's better to precompute `1/mass_` and then multiply by it. In this case compute `1/mass_^2`. Division is 25x slower than multiplication and here you are doing two extra divisions per grid point. Also `pow<2>()`.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
```diff
@@ -0,0 +1,781 @@
...
+auto TrumpetSchwarzchild::variables(
+    const tnsr::I<DataType, volume_dim, Frame::Inertial>& x, const double /*t*/,
+    const IntermediateVars<DataType>& vars,
+    tmpl::list<gr::Tags::Shift<DataType, volume_dim>> /*meta*/) const
+    -> tuples::TaggedTuple<gr::Tags::Shift<DataType, volume_dim>> {
+  tnsr::I<DataType, volume_dim, Frame::Inertial> shift;
+  for (size_t i = 0; i < volume_dim; ++i) {
+    shift.get(i) = x.get(i) / mass_ * sqrt(get_c_n_squared(n_)) *
+                   exp(vars.lapse_on_user_grid / n_) /
+                   pow(vars.schwarzchild_r_on_user_grid, 3);
```
nilsdeppe: You might want to also cache `1/schwarzchild_r_on_user_grid` and then you can just multiply by that. Same for `1/n_` and `1/mass_`

nilsdeppe (changes requested):

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.hpp
hen-w: Done.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.hpp
hen-w: Done.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.hpp
hen-w: Done.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
hen-w: Done.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
hen-w: Neither of the base classes has member variables to pup or member pup functions.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.cpp
hen-w: Same as above.

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzchild.hpp
nilsdeppe: ping

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzschild.cpp
```diff
@@ -40,79 +41,88 @@ namespace {
...
 
   return RootFinder::toms748(
       [n, lapse](double schwarzschild_r) -> double {
         return function_of_lapse_and_schwarzschild_r(lapse, schwarzschild_r, n);
       },
       (lapse < crit_lapse) ? min_schwarzschild_r : crit_schwarzschild_r,
-      (lapse < crit_lapse) ? crit_schwarzschild_r : max_schwarzschild_r, 1e-15,
-      1e-15);
+      (lapse < crit_lapse) ? crit_schwarzschild_r : max_schwarzschild_r,
+      1.0e-15, 1.0e-15);
```
nilsdeppe: 1.e-15 to be consistent in file

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzschild.hpp
```diff
@@ -32,14 +32,48 @@ namespace gr::Solutions {
...
  * This solution is a trumpet Schwarzschild black hole in the isotropic
- * coordinates. It is time-independent puncture solution in 1+log slicing.
+ * coordinates. It is a time-independent puncture solution in 1+log slicing.
  * The solution cannot be written down analytically in Schwarzschild
  * coordinates or isotropic coordinates. Refer to \cite Hannam2008
  * for equations and details.
  *
+ * We first set up a source grid in isotropic radial coordinate r, on which
+ * we compute the lapse and Schwarzschild radial coordinate R. The lapse
+ * is computed by solving eq. (54)-(56) using the toms748 algorithm, where
```
nilsdeppe, line 42: Would it be easy to include the equations directly in the dox? You should be able to download the tex file from arxiv to make it easier. Then at least this class is fairly standalone :)

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzschild.hpp
```diff
@@ -32,14 +32,48 @@ namespace gr::Solutions {
...
+ * the integrals in eq. (54)-(56) are evaluated by a tanh_sinh integrator.
+ * A standard Gauss quadratures will fail due to singular
+ * behaviors of the integrands near the integration limit. The Schwarzschild R
+ * is then computed by solving eq. (39) using again the toms748 algorithm. Note
+ * that for a value of n near 2. (corresponding to standard 1+log slicing),
+ * eq. (39) has two positive roots for a fixed lapse in [0., 1.). We choose
```
nilsdeppe: [optional] Maybe include your Mathematica plot as a png to show the two solutions?

src/PointwiseFunctions/AnalyticSolutions/GeneralRelativity/TrumpetSchwarzschild.hpp
```diff
@@ -32,14 +32,48 @@ namespace gr::Solutions {
...
+ * the physical root, i.e. with the correct asymptotic behaviors. By using
```
nilsdeppe: Can you make mathematically precise what the correct asymptotic behavior is and how it's numerically selected?

nilsdeppe (commented):
Looks great! I've got a few more suggestions on the documentation. Please rebase and squash everything in right away :)

nilsdeppe (approved):

PR #6728, opened 2025-07-03
Adapt BBH ID for high spins

iago-mendes:
## Proposed changes

This PR makes 3 changes to BBH ID so that it is capable of reproducing a solution from SpEC with q = 10 and spin ~ 0.99.

1. Set the (l,m)=(0,0) coefficient of the shape maps to the transformation from Boyer-Lindquist to Kerr-Schild. Before, this was set to 0, which meant that our excisions didn't have a size distortion.
2. Use a higher `l_max` for horizon finding and shape maps when spin > 0.9. In the run mentioned above, SpEC uses Lmesh = 36. Ideally, this should be chosen based on some automatic criterion, but here we set it to 40 as a temporary solution.
3. Increase the maximum number of horizon finder iterations. When running the case above with L=1 and P=10, the horizon finder needed 219 iterations. So, here we default to a maximum of 300 iterations.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/support/Pipelines/Bbh/FindHorizon.py
+++ b/support/Pipelines/Bbh/FindHorizon.py
@@ -167,7 +167,7 @@ def find_horizon(
             truncation_tol=0.01,
             divergence_tol=1.2,
             divergence_iter=5,
-            max_its=100,
+            max_its=300,
         )
     strahlkorper = initial_guess
     while True:
--- a/support/Pipelines/Bbh/InitialData.py
+++ b/support/Pipelines/Bbh/InitialData.py
@@ -103,6 +103,9 @@ def id_parameters(
     extra_radial_refinement_p = (
         round(mass_ratio / 5.0) if (mass_ratio > 5.0) else 0
     )
+    horizon_l_max = (
+        40 if max(np.linalg.norm(chi_A), np.linalg.norm(chi_B)) > 0.9 else 20
+    )
     return {
         "ConformalMassRight": conformal_mass_a,
         "ConformalMassLeft": conformal_mass_b,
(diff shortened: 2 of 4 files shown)


nilsvu (approved):
@iago-mendes looks good, and nice job figuring this out.

Do you know how spec chooses the horizon lmax and the max num iterations? In cases where spec has a simple choice for things it's easiest to take that over for now with a brief comment like "chosen same as spec".

iago-mendes:
SpEC's `ParallelStrahlkorperFinderWork` has a while loop that keeps incrementing `Lmesh` by 1 until a horizon is found or a limit of `80` is reached. We could easily do this, but I think this seems very inefficient. The time it takes to run SpECTRE's horizon finder with `l_max = 35` or `l_max = 40` seems very similar (haven't measured it quantitatively), so it might be best to just run at the highest resolution needed for extreme cases.

What do you think? Should we do a similar loop?

nilsvu:
Nope sounds good to run at high l. And is the number of iterations comparable to spec?

iago-mendes:
Hmm, the output files don't show any number of horizon finder iterations. That said, the default `FastFlow` maximum number of iterations is 100, so I'm guessing that it was still under that? However, it does use the previous low-resolution strahlkorper as an initial guess for the higher-resolution horizon find. So, I guess it takes somewhere between 100 and 200 iterations, which is what we have for most cases in SpECTRE (unless we go to P>=10).

nilsvu (approved):
Ok sounds good let's go with these choices!

PR #6738, opened 2025-07-07
Add Wigner 3j symbols.

markscheel:
## Proposed changes

Adds Wigner 3j symbols and an external fortran library that computes them.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


src/NumericalAlgorithms/SphericalHarmonics/WignerThreeJ.cpp
```diff
@@ -0,0 +1,70 @@
...
+void drc3jj_(const double& l2, const double& l3, const double& m2,
+             const double& m3, double& l1min, double& l1max, double* answer,
+             const int& answersize, int& error_code, const double& doublemax);
+}
+
+WignerThreeJ::WignerThreeJ(const size_t l2, const int m2, const size_t l3,
+                           const int m3)
+    : l2_(double(l2)),
+      m2_(m2),
+      l3_(double(l3)),
```
knelli2: Use `static_cast<double>(l2)` here

src/NumericalAlgorithms/SphericalHarmonics/WignerThreeJ.cpp
```diff
@@ -0,0 +1,70 @@
...
+          std::max(std::abs(static_cast<int>(l2) - static_cast<int>(l3)),
+                   std::abs(m2 + m3)))),
+      l1_max_(l2 + l3),
+      up_to_date_(false),
+      coefs_(l1_max_ - l1_min_ + 1) {
+  if (UNLIKELY(std::abs(m2) > static_cast<int>(l2))) {
+    ERROR("WignerThreeJ: m2 is " << m2 << " but l2 is " << l2);
+  }
+  if (UNLIKELY(std::abs(m3) > static_cast<int>(l3))) {
+    ERROR("WignerThreeJ: m3 is " << m3 << " but l3 is " << l3);
```
knelli2: These should also say something like "abs(m2) must be larger than l2" so the user knows what to change if they get this error

knelli2 (changes requested):

src/NumericalAlgorithms/SphericalHarmonics/WignerThreeJ.hpp
```diff
@@ -0,0 +1,59 @@
...
+#include <cstddef>
+#include <vector>
+
+/*!
+ * \brief Computes Wigner 3J symbols.
+ *
+ * Given l2, m2, l3, and m3, computes and caches the Wigner 3J symbols
+ *
+ * ( l1     l2 l3 )
+ * ( -m3-m2 m2 m3 )
```
nilsdeppe: I doubt this renders the way you want. Can you check it? You might have to use latex mode

knelli2: No it didn't really render properly. I would put this and all the `l2, m2, ...` in latex as well (like `$l_2$`)

image

src/NumericalAlgorithms/SphericalHarmonics/WignerThreeJ.hpp
```diff
@@ -9,16 +9,23 @@
 /*!
  * \brief Computes Wigner 3J symbols.
  *
- * Given l2, m2, l3, and m3, computes and caches the Wigner 3J symbols
+ * Given \f$l_2\f$, \f$m_2\f$, \f$l_3\f$, and \f$m_3\f$,
+ * computes and caches the Wigner 3J symbols
```
knelli2: You can use the more modern way we have of adding latex to dox, which is just dollar signs. No `\f` necessary. Applies to the rest of the dox as well.

src/NumericalAlgorithms/SphericalHarmonics/WignerThreeJ.hpp
```diff
@@ -9,16 +9,23 @@
...
  *
- * ( l1     l2 l3 )
- * ( -m3-m2 m2 m3 )
+ * \f{align}
+ * \left(
```
knelli2: And here, this can instead be
```
 * \begin{align}
 * ...
 * \end{align}
```

knelli2 (changes requested):
Docs look correct now. You can squash everything including these next changes

knelli2 (approved):

PR #6753, opened 2025-07-12
Support linking device code in static libraries

nilsvu:
## Proposed changes

Allows compiling device code into static libraries and link them into Kokkos programs.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.


nilsdeppe:
The code looks fine.
We should come up with a way of handling Kokkos stuff with clang-tidy. Depending on how extensively Kokkos gets used (if we just call double overloads it won't matter) we might have every PR fail clang tidy making it less useful

nilsvu:
Good point, I'll look into clang-tidy for kokkos

PR #6754, opened 2025-07-12
Support Kokkos-parallel tensor functions

nilsvu:
## Proposed changes

Support `Tensor<Kokkos::View<double*>>` to do tensor math on GPUs.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/Tensor/AtIndex.hpp
+++ b/src/DataStructures/Tensor/AtIndex.hpp
@@ -1,0 +1,69 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include <concepts>
+#include <cstddef>
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/Tensor/Metafunctions.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+#include "Utilities/Kokkos/KokkosCore.hpp"
+
+/*!
+ * \brief Get the `Tensor` at a specific grid point index.
+ *
+ * Works with a `Tensor` of `Kokkos::View`s. Can also be made to work with
+ * `DataVector` if needed (use `operator[]` instead of `operator()`).
+ */
+template <
+    typename TensorType, std::integral... Is,
+    typename ValueType = typename TensorType::type::value_type,
+    typename ResultType = TensorMetafunctions::swap_type<ValueType, TensorType>>
+KOKKOS_FUNCTION ResultType make_at_index(const TensorType& tensor,
+                                         const Is&... i) {
+  ResultType result{};
+  for (size_t component = 0; component < TensorType::size(); ++component) {
+    result[component] = tensor[component](i...);
+  }
+  return result;
+}
+
+/*!
+ * \brief Set the `Tensor` at a specific grid point index.
+ *
+ * Works with a `Tensor` of `Kokkos::View`s. Can also be made to work with
+ * `DataVector` if needed (use `operator[]` instead of `operator()`).
+ */
+template <typename TensorType, std::integral... Is,
+          typename ValueType = typename TensorType::type::value_type>
+KOKKOS_FUNCTION void set_at_index(
+    const gsl::not_null<TensorType*> tensor,
+    const TensorMetafunctions::swap_type<ValueType, TensorType>& value,
+    const Is&... i) {
+  for (size_t component = 0; component < value.size(); ++component) {
+    (*tensor)[component](i...) = value[component];
+  }
+}
+
+namespace Tags {
+
+/*!
+ * \brief Tag representing a specific grid point index of a tensor.
+ *
+ * This tag replaces the `Tensor` data type with `double` so that it can be
+ * used with pointwise operations, e.g. in a `Kokkos::parallel_for` kernel.
+ */
+template <typename Tag>
+struct AtIndex : db::PrefixTag {
+ private:
+  using TensorType = typename Tag::type;
+  using ValueType = typename TensorType::type::value_type;
+
+ public:
+  using tag = Tag;
+  using type = TensorMetafunctions::swap_type<ValueType, TensorType>;
+};
+
+}  // namespace Tags
--- a/src/DataStructures/Tensor/Tensor.hpp
+++ b/src/DataStructures/Tensor/Tensor.hpp
@@ -11,11 +11,6 @@
 #include <pup.h>
 #include <pup_stl.h>
 
-#include "DataStructures/ComplexDataVector.hpp"
-#include "DataStructures/ComplexModalVector.hpp"
-#include "DataStructures/DataVector.hpp"
-#include "DataStructures/ModalVector.hpp"
-#include "DataStructures/SpinWeighted.hpp"
 #include "DataStructures/Tensor/Expressions/AddSubtract.hpp"
 #include "DataStructures/Tensor/Expressions/Contract.hpp"
 #include "DataStructures/Tensor/Expressions/DataTypeSupport.hpp"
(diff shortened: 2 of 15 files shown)


src/Utilities/Array.hpp
```diff
@@ -10,15 +10,14 @@
 #include <ostream>
```
nilsdeppe, line 10: Is this true for both CUDA and HIP? I'm worried we did all the work to add them, now do the work to remove them, only to have to do the work to add them. Is there any overhead to keeping them for now?

src/DataStructures/Tensor/Tensor.hpp
```diff
@@ -163,21 +163,21 @@ class Tensor<X, Symm, IndexList<Indices...>> {
   using reverse_iterator = typename storage_type::reverse_iterator;
```
nilsdeppe, line 163: It would be good to verify these work in a constexpr context in the tests. It can be misleading if they are marked `constexpr` but aren't actually valid in constexpr contexts. We've had this in the past. Of course for templates this can depend on the template type, but we can just check `double`

src/DataStructures/Tensor/AtIndex.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+#include <cstddef>
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/Tensor/Metafunctions.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+
+/// Get the `Tensor` at a specific grid point index.
+template <typename TensorType, typename... Is>
+KOKKOS_FUNCTION TensorMetafunctions::swap_type<double, TensorType>
```
nilsdeppe: missing include for `KOKKOS_FUNCTION`

src/DataStructures/Tensor/Tensor.hpp
```diff
@@ -482,20 +497,40 @@ Tensor<X, Symm, IndexList<Indices...>>::component_suffix(
...
+KOKKOS_FUNCTION Tensor<X, Symm, IndexList<Indices...>>::Tensor(
+    storage_type data)
+  requires(sizeof...(Indices) <= 1 and std::is_fundamental_v<X>)
+    : data_(std::move(data)) {}
 template <typename X, typename Symm, template <typename...> class IndexList,
           typename... Indices>
 Tensor<X, Symm, IndexList<Indices...>>::Tensor(storage_type data)
-  requires(sizeof...(Indices) <= 1)
+  requires(sizeof...(Indices) <= 1 and not std::is_fundamental_v<X>)
     : data_(std::move(data)) {}
```
nilsdeppe, line 489: I'm very confused. Are these implementations identical? Same below? Some explanation is needed why we need the two seemingly identical functions :) Maybe write that in the doxygen comments for the two sets of constructors?

src/DataStructures/Tensor/AtIndex.hpp
```diff
@@ -0,0 +1,48 @@
...
+
+/*!
+ * \brief Tag representing a specific grid point index of a tensor
+ *
+ * This tag replaces the `Tensor` data type with `double` so that it can be
+ * used with pointwise operations, e.g. in a `Kokkos::parallel_for` kernel.
+ */
+template <typename Tag>
+struct AtIndex : db::PrefixTag {
+  using tag = Tag;
```
nilsdeppe, line 65: Should we `require` that the type the tensor holds is a `DataVector`? Then we don't do things like `ModalVector` or `Complex*Vector` no accident

tests/Unit/DataStructures/Tensor/Test_TensorKokkos.cpp
```diff
@@ -0,0 +1,48 @@
+// Distributed under the MIT License.
```
nilsdeppe, line 1: Could you please add the `cpp` file to `CMakeLists` so the tests are run?

src/DataStructures/Tensor/AtIndex.hpp
```diff
@@ -0,0 +1,48 @@
...
+#include <cstddef>
+
+#include "DataStructures/DataBox/Tag.hpp"
+#include "DataStructures/Tensor/Metafunctions.hpp"
+#include "DataStructures/Tensor/Tensor.hpp"
+
+/// Get the `Tensor` at a specific grid point index.
+template <typename TensorType, typename... Is>
+KOKKOS_FUNCTION TensorMetafunctions::swap_type<double, TensorType>
+make_at_index(const TensorType& tensor, const Is&... i) {
```
nilsdeppe: made `require`/`static_assert` that all the `Is` are integer types?

src/DataStructures/Tensor/AtIndex.hpp
```diff
@@ -0,0 +1,48 @@
...
+  }
+  return result;
+}
+
+/// Set the `Tensor` at a specific grid point index.
+template <typename TensorType, typename... Is>
+KOKKOS_FUNCTION void set_at_index(
+    const gsl::not_null<TensorType*> tensor,
+    const TensorMetafunctions::swap_type<double, TensorType>& value,
+    const Is&... i) {
```
nilsdeppe, line 44: Same about type constraints.

src/DataStructures/Tensor/AtIndex.hpp
nilsdeppe: Could we constrain the type, maybe `const Tensor<DataVector, ...>&`?

src/DataStructures/Tensor/AtIndex.hpp
```diff
@@ -0,0 +1,48 @@
...
+    result[component] = tensor[component](i...);
+  }
+  return result;
+}
+
+/// Set the `Tensor` at a specific grid point index.
+template <typename TensorType, typename... Is>
+KOKKOS_FUNCTION void set_at_index(
+    const gsl::not_null<TensorType*> tensor,
+    const TensorMetafunctions::swap_type<double, TensorType>& value,
```
nilsdeppe: Could we constrain the types more here too?

tests/Unit/DataStructures/Tensor/Test_TensorKokkos.cpp
```diff
@@ -0,0 +1,48 @@
...
+
+KOKKOS_FUNCTION Scalar<double> tensor_func(const Scalar<double>& input) {
+  return Scalar<double>(2.0 * get(input));
+}
+
+}  // namespace
+
+SPECTRE_TEST_CASE("Unit.DataStructures.Tensor.Kokkos",
+                  "[Unit][DataStructures]") {
+  {
```
nilsdeppe, line 24: why extra braces?

nilsdeppe (changes requested):

src/Utilities/Array.hpp
nilsvu: I believe so: https://rocm.docs.amd.com/projects/HIP/en/docs-6.1.2/reference/cpp_language_support.html

> [...] any functions using the `constexpr` qualifier [...] are implicitly considered to have the __host__ and __device__ execution space specifier.

However, I haven't tried compiling with HIP yet.

I think this treatment of `constexpr` functions as implicit device code is necessary to use e.g. `std::array` (and by extension `Tensor`) on device.

src/DataStructures/Tensor/Tensor.hpp
nilsvu: The difference was the `KOKKOS_FUNCTION` in the `is_fundamental = true` specialization. However, I changed this to just be `constexpr` and added a test that it works as constexpr. That compiles it for device as well.

tests/Unit/DataStructures/Tensor/Test_TensorKokkos.cpp
nilsvu: Oh it is added with `if(SPECTRE_KOKKOS)`

src/DataStructures/Tensor/AtIndex.hpp
nilsvu: The tensor type should actually be `Tensor<Kokkos::View>` for the functions in this file. I added some docs. The functions can also be made to work with `DataVector` (use `operator[]` instead of `operator()`) but we currently don't need that. I also added some code to get the value type of the Kokkos view and use that when swapping out the type.

src/DataStructures/Tensor/AtIndex.hpp
nilsvu: Should be `Tensor<Kokkos::View>` (see above). Constraining to that requires including Kokkos, which only works when Kokkos in installed. Instead, I added a SFINAE constraint that gets the value type from the Kokkos view and use that for swapping out types.

tests/Unit/DataStructures/Tensor/Test_TensorKokkos.cpp
nilsvu: Oh I use these to add `INFO` messages in tests and scope test cases, but forgot that here. Added.

nilsvu:
Ready for another look @nilsdeppe

nilsvu:
Pushed a commit that removes an unused variable that CI didn't like (seems unrelated to this PR, so I don't understand why CI didn't trigger on this in other PRs).

nilsdeppe (commented):
LGTM, please rebase and squash

nilsdeppe (approved):

PR #6756, opened 2025-07-13
Get rid of more base tags

kidder:
## Proposed changes

Eliminate several uses of base tags whose derived simple tags had the same `type` alias.
- Allow derived mutable tags.  This allows several tags to be derived from the same simple tag by adding a `base` type alias.  This allows choosing  at compile time one of the different methods that a simple tag can be created from options.
- Some base tags had one simple tag that could not be created from options, and one that could.  In this case the base and non-creatable simple tag can be removed, and replaced with the creatable simple tag.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/DataStructures/DataBox/DataBox.hpp
+++ b/src/DataStructures/DataBox/DataBox.hpp
@@ -870,8 +870,9 @@ auto DataBox<tmpl::list<Tags...>>::compute_tag_graphs() -> TagGraphs {
       });
 
   // Set mutation function
-  tmpl::for_each<mutable_item_tags>([&result](auto tag_v) {
-    using tag = tmpl::type_from<decltype(tag_v)>;
+  tmpl::for_each<mutable_item_tags>([&result]<typename Tag>(
+                                        tmpl::type_<Tag> /*meta*/) {
+    using tag = detail::get_base<Tag>;
     const std::string tag_name = pretty_type::get_name<tag>();
     if (result.tag_mutate_functions.find(tag_name) ==
         result.tag_mutate_functions.end()) {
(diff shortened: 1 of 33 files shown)


src/DataStructures/DataBox/DataBox.hpp
```diff
@@ -871,7 +871,7 @@ auto DataBox<tmpl::list<Tags...>>::compute_tag_graphs() -> TagGraphs {
 
   // Set mutation function
   tmpl::for_each<mutable_item_tags>([&result](auto tag_v) {
-    using tag = tmpl::type_from<decltype(tag_v)>;
+    using tag = detail::get_base<tmpl::type_from<decltype(tag_v)>>;
     const std::string tag_name = pretty_type::get_name<tag>();
```
nilsdeppe: Do
```cpp
  tmpl::for_each<mutable_item_tags>([&result]<typename Tag>(tmpl::type_<Tag> /*meta*/) {
    using tag = detail::get_base<Tag>;
```

nilsdeppe (changes requested):
You can squash

src/ParallelAlgorithms/ApparentHorizonFinder/Tags.hpp
```diff
@@ -122,16 +126,13 @@ struct FailedInterpolationIterations : db::SimpleTag {
...
 
-/// Base tag for whether or not to write the centers of the horizons to disk.
-/// Most likely to be used in the `ObserveCenters` post horizon find callback
-///
-/// Other things can control whether the horizon centers are output by defining
-/// their own simple tag from this base tag.
-struct ObserveCentersBase : db::BaseTag {};
-
 /// Simple tag for whether to write the centers of the horizons to disk.
 /// Currently this tag is not creatable by options
```
wthrowe: Remove this line of the docs.

nilsdeppe (approved):

src/ParallelAlgorithms/ApparentHorizonFinder/Tags.hpp
nilsdeppe: I've checked that this was done :)

PR #6759, opened 2025-07-14
Recommend charm version 8

wthrowe:
This does not update containers, CI, or clusters, just the docs and cmake files.

### Upgrade instructions

Charm version 8 is now recommended over version 7, but 7 is still supported for applications not using dynamic h-refinement or similar post-startup element creation.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/cmake/SetupCharm.cmake
+++ b/cmake/SetupCharm.cmake
@@ -38,9 +38,9 @@ find_package(Charm ${SPECTRE_REQUIRED_CHARM_VERSION} REQUIRED
   EveryLB
   ${SCOTCHLB_COMPONENT}
   )
-if(CHARM_VERSION VERSION_GREATER 7.0.1)
-  message(WARNING "Builds with Charm++ versions greater than 7.0.1 are \
-considered experimental. Please file any issues you encounter.")
+if(CHARM_VERSION VERSION_LESS 8.0.0)
+  message(NOTICE "Charm++ versions less than 8.0.0 have known bugs with \
+element creation.  Dynamic h-refinement is disabled.")
 endif()
 
 if (USE_SCOTCH_LB)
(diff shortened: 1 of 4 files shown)


src/ParallelAlgorithms/Amr/Actions/CreateChild.hpp
```diff
@@ -50,6 +51,12 @@ struct CreateChild {
       const size_t index_of_child_id,
       const std::unordered_map<Parallel::Phase, size_t>
           parent_phase_bookmarks) {
+    if (CHARM_VERSION_MAJOR < 8 and
+        Parallel::number_of_procs<size_t>(cache) > 1) {
+      ERROR_NO_TRACE(
+          "Dynamically creating elements in parallel executables is broken "
+          "until charm 8.");
```
nilsvu: It's not quite this bad, is it? I've been using AMR with h-refinement in a separate phase in elliptic executables for a while now with no issues.

wthrowe: @kidder Do you know what operations can trigger the hang?

kidder: I believe it occurs when one of the initial elements is destroyed and then recreated in a later iteration on a PE other than PE 0.   If the current elliptic code usage only refines, and never coarsens, this condition won't occur.

See https://github.com/kidder/charm_spectre_tests/tree/main/dynamic_array_hang

nilsdeppe (dismissed):
Larry and Nils, any followup on the discussion so we can merge this?

src/ParallelAlgorithms/Amr/Actions/CreateChild.hpp
nilsvu: The elliptic code only refines, yes.

kidder: okay, then probably only erroring in CreateParent would be adequate

wthrowe: I've dropped the changes to CreateChild.

kidder (approved):

PR #6769, opened 2025-07-17
Add helper function to fill fluxes from primitive data

ncorsobh:
## Proposed changes

Adds a helper function to `ValenciaDivClean` which takes primitive variable data, feeds it through `ConservativeFromPrimitive`, then runs `ComputeFluxes`.

When using higher order finite difference, it is necessary to compute cell centered flux information from the ghost data when handling boundary conditions. Because of this, there are several occasions when these two functions are run in sequence (_e.g._ in the `fd_ghost` functions in the boundary condition files). This helper function is purely meant to improve readability and reduce the number of lines of code in the boundary condition files. This will result in a notable reduction in the number of lines added when higher order finite difference is implemented for `GhValenciaDivClean` as well.

_N.B._ there are some instances when `ConservativeFromPrimitive` is immediately followed by `ComputeFluxes` which were not converted to the helper function (_e.g._ in the `dg_ghost` functions in the boundary condition files). These are instances when the conservative data is also saved for other purposes, so it makes sense to keep these functions separate.

### Code review checklist

  Then open `index.html`.
  `new feature` if appropriate.

### DIFF
--- a/src/Evolution/Systems/GrMhd/ValenciaDivClean/ComputeFluxesFromPrimitives.hpp
+++ b/src/Evolution/Systems/GrMhd/ValenciaDivClean/ComputeFluxesFromPrimitives.hpp
@@ -1,0 +1,61 @@
+// Distributed under the MIT License.
+// See LICENSE.txt for details.
+
+#pragma once
+
+#include "DataStructures/Variables.hpp"
+#include "Evolution/Systems/GrMhd/ValenciaDivClean/ConservativeFromPrimitive.hpp"
+#include "Evolution/Systems/GrMhd/ValenciaDivClean/Fluxes.hpp"
+#include "Utilities/Gsl.hpp"
+
+namespace grmhd::ValenciaDivClean {
+namespace detail {
+template <typename FluxTags, typename BoundaryVarsType, typename... ReturnTags,
+          typename... ConservativeTags, typename... ArgumentTags,
+          typename... FluxArgumentTags>
+void compute_fluxes_from_primitives_impl(
+    const gsl::not_null<Variables<FluxTags>*> flux_vars,
+    const BoundaryVarsType& boundary_vars, tmpl::list<ReturnTags...> /*meta*/,
+    tmpl::list<ConservativeTags...> /*meta*/,
+    tmpl::list<ArgumentTags...> /*meta*/,
+    tmpl::list<FluxArgumentTags...> /*meta*/) {
+  Variables<typename tmpl::list<ConservativeTags...>> conserved_vars{
+      flux_vars->number_of_grid_points()};
+  grmhd::ValenciaDivClean::ConservativeFromPrimitive::apply(
+      make_not_null(&get<ConservativeTags>(conserved_vars))...,
+      get<ArgumentTags>(boundary_vars)...);
+  grmhd::ValenciaDivClean::ComputeFluxes::apply(
+      make_not_null(&get<ReturnTags>(*flux_vars))...,
+      get<ConservativeTags>(conserved_vars)...,
+      get<FluxArgumentTags>(boundary_vars)...);
+}
+}  // namespace detail
+
+/*!
+ * \brief Helper function that computes fluxes from a given set of primitive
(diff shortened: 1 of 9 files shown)


tests/Unit/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Test_BoundaryConditionGhostData.cpp
```diff
@@ -390,8 +345,30 @@ void test(const BoundaryConditionType& boundary_condition,
...
+              gr::Tags::Lapse<DataVector>, gr::Tags::Shift<DataVector, 3>,
+              gr::Tags::SqrtDetSpatialMetric<DataVector>,
+              gr::Tags::InverseSpatialMetric<DataVector, 3>>;
+          Variables<fluxes_tags> boundary_data{get(rest_mass_density).size()};
+
+          get<RestMassDensity>(boundary_data) = rest_mass_density;
+          get<ElectronFraction>(boundary_data) = electron_fraction;
+          get<SpecificInternalEnergy>(boundary_data) = specific_internal_energy;
+          get<Temperature>(boundary_data) = local_temperature;
+          get<SpatialVelocity>(boundary_data) = spatial_velocity;
```
ncorsobh, line 360: If there is a more efficient way to go about creating/filling this `Variables`, I'd like to know. This part feels a little clunky to me.

ncorsobh:
@nilsdeppe changes have been made, so it is good to review again. Less changed than I expected, but it's still more polished now.

src/Evolution/Systems/GrMhd/ValenciaDivClean/ComputeFluxesFromPrimitives.hpp
```diff
@@ -0,0 +1,58 @@
...
+      get<ConservativeTags>(conserved_vars)...,
+      get<FluxArgumentTags>(boundary_vars)...);
+}
+}  // namespace detail
+
+/*!
+ * \brief Helper function that computes fluxes from a given set of primitive
+ * variables. Primarily used for filling neighbor data when flux information is
+ * needed for boundaries.
+ */
```
nilsdeppe, line 41: Can you add (on a new line after a blank line) `\warning This computes the conservative variables internally. If you also need those, you shouldn't call this function.` so that people don't accidentally compute conservative variables twice?

tests/Unit/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Test_BoundaryConditionGhostData.cpp
nilsdeppe: Have you checked if `assign_subset` would work? That's a member function of variables that can do some of these things, but maybe not this exact thing.

nilsdeppe (changes requested):
Overall LGTM, one suggestion and an answer of "maybe" to your question :) You can squash any changes immediately

src/Evolution/Systems/GrMhd/ValenciaDivClean/ComputeFluxesFromPrimitives.hpp
ncorsobh: Done! This was in fact relevant for a few parts of the code, so I agree this is a useful statement to include.

tests/Unit/Evolution/Systems/GrMhd/ValenciaDivClean/FiniteDifference/Test_BoundaryConditionGhostData.cpp
ncorsobh: For posterity: it was discussed in a call that `assign_subset` would not help here. If optimization were pursued, the lambda function `check_fluxes` should be changed to take a `Variables` as an input rather than individual tensor. Since this is just a test file, we decided it was not necessary to make this change.

nilsdeppe (approved):

